PR #3806, opened 2021-05-27
Fix mesh groupshared atomics test bug

PrestonHamlin:
Mesh shader groupshared initialization and read were asymmetric, resulting in read of uninitialized data. 

The uninitialized data caused nondeterministic behaviour, so it would sometimes pass. It passes consistently with the fix.


pow2clk (approved):
Thanks for catching this!

PR #3809, opened 2021-06-01
SROA: Skip arrays of object/matrix as well.

tex3d:
We do not want this SROA pass to break down an object or matrix arrays either.  Otherwise, this code will break down things like an aggregate zeroinitializer into a bunch of stores of zero to the struct elements inside HLSL objects, which are supposed to be treated more like abstract opaque types here.  If this happens, it can lead to broken assumptions down the line, such as during GEP replacement in resource legalization.  `scalarrepl-param-hlsl` should handle these array cases correctly.

This change also refactors the HLSL object/matrix detection code for clarity and simplicity.


tex3d:
Was part of another change that was merged.

PR #3810, opened 2021-06-02
Fix issues with resource array aliasing support

tex3d:
SROA: Skip arrays of object/matrix as well.

Fix issues with ReplaceUseOfZeroInit*

  - Fix ReplaceUseOfZeroInitPostDom bailing on use of memcpy dest.
    When memcpy not in entry block, any use of dest after memcpy would abort
    replacement, as if it was unsafe to replace dest with src.  But uses
    of dest dominated by memcpy *are* safe to replace.
  - Fix ReplaceUseOfZeroInit* misuse of post-dom.
    Post-dom was used to detect whether it was safe to replace uses of dest
    before the memcpy with zero when dest was zeroinitialized.  But post-dom
    is not the right way to tell if this is safe.  It is unsafe if any uses
    *could* follow the memcpy.  So the new test is to gather a set of blocks
    that could be reachable from the successors of the memcpy block (where
    memcpy ends the block because we split it there).  If so, it is not safe
    to replace the use with zeroinitializer, and because it was not dominated
    by the memcpy either, it's not safe to replace dest with src here either,
    so memcpy replacement must abort.
  - Re-merge blocks after splitting in ReplaceUseOfZeroInitBeforeDef.
    Restore CFG after ReplaceUseOfZeroInitBeforeDef, rather than leaving it
    in more of a mess to be cleaned up much later.
    This has the side benefit of preserving more trivial entry-block
    replacement opportunities.

DFE should apply to internal functions in lib target as well.

  - Also, iterating over RemoveUnusedFunctions will allow removal of more
    functions only called by other internal functions being removed.

Add LowerStaticGlobalIntoAlloca before SROA for LowerMemcpy opportunities

  - Existing location after SROA must also be there for non-resource aggregates
  - Don't preserve zeroinitializer when moving static GV to alloca for objects
  - Minor opt in LowerStaticGlobalIntoAlloca: skip fn decls more cheaply

### DIFF
--- a/lib/Transforms/Scalar/SROA.cpp
+++ b/lib/Transforms/Scalar/SROA.cpp
@@ -627,6 +627,17 @@ static Value *foldPHINodeOrSelectInst(Instruction &I) {
   return foldSelectInst(cast<SelectInst>(I));
 }
 
+// HLSL Change - Detect HLSL Object or Matrix [array] type
+// These types should be SROA'd elsewhere as necessary.
+bool SkipHLSLType(Type *Ty, bool SkipHLSLMat) {
+  if (Ty->isPointerTy())
+    Ty->getPointerElementType();
+  while (Ty->isArrayTy())
+    Ty = Ty->getArrayElementType();
+  return (SkipHLSLMat && hlsl::HLMatrixType::isa(Ty)) ||
+         hlsl::dxilutil::IsHLSLObjectType(Ty);
+}
+
 /// \brief Builder for the alloca slices.
 ///
 /// This class builds a set of alloca slices by recursively visiting the uses
@@ -697,21 +708,10 @@ private:
     if (BC.use_empty())
       return markAsDead(BC);
     // HLSL Change Begin - not sroa matrix type.
-    if (PointerType *PT = dyn_cast<PointerType>(BC.getType())) {
-      Type *EltTy = PT->getElementType();
-      if ((SkipHLSLMat && hlsl::HLMatrixType::isa(EltTy)) ||
-          hlsl::dxilutil::IsHLSLObjectType(EltTy)) {
-        AS.PointerEscapingInstr = &BC;
-        return;
-      }
-      if (PointerType *SrcPT = dyn_cast<PointerType>(BC.getSrcTy())) {
-        Type *SrcEltTy = SrcPT->getElementType();
-        if ((SkipHLSLMat && hlsl::HLMatrixType::isa(SrcEltTy)) ||
-            hlsl::dxilutil::IsHLSLObjectType(SrcEltTy)) {
-          AS.PointerEscapingInstr = &BC;
-          return;
-        }
-      }
+    if (SkipHLSLType(BC.getType(), SkipHLSLMat) ||
+        SkipHLSLType(BC.getSrcTy(), SkipHLSLMat)) {
+      AS.PointerEscapingInstr = &BC;
+      return;
     }
     // HLSL Change End.
     return Base::visitBitCastInst(BC);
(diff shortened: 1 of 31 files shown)


python3kgae (approved):

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3621,23 +3630,42 @@ static bool ReplaceUseOfZeroInitPostDom(Instruction *I, Value *V,
...
   if (&F->getEntryBlock() == I->getParent()) {
-    return ReplaceUseOfZeroInitEntry(I, GV);
+    bSuccess = ReplaceUseOfZeroInitEntry(I, GV);
   } else {
-    // Post dominator tree.
-    PostDominatorTree PDT;
-    PDT.runOnFunction(*F);
-    return ReplaceUseOfZeroInitPostDom(I, GV, PDT);
+    DominatorTree DT;
+    DT.recalculate(*F);
```
pow2clk, line 3658: We ran into compile time performance issues recalculating this redundantly elsewhere. Since this is called from LowerMemCpy, which already has a DT for this function, you might as well pass it in and reuse it.

pow2clk (commented):
Looks good overall. There is one typo that will affect correctness though

pow2clk:
Looks like the Ubuntu build timed out

lib/Transforms/Scalar/SROA.cpp
```diff
@@ -627,6 +627,17 @@ static Value *foldPHINodeOrSelectInst(Instruction &I) {
   return foldSelectInst(cast<SelectInst>(I));
 }
 
+// HLSL Change - Detect HLSL Object or Matrix [array] type
+// These types should be SROA'd elsewhere as necessary.
+bool SkipHLSLType(Type *Ty, bool SkipHLSLMat) {
+  if (Ty->isPointerTy())
+    Ty->getPointerElementType();
```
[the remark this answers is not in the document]
tex3d: Thanks for the catch!

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
tex3d: Unfortunately, it will only work if updated for the split block, then again for the re-merge.  That is possible, but I will need to be cautious since I haven't written code that updates the Dom this way.

The other possibility is get rid of the block split operation and do instruction iteration inside the block with the memcpy to determine before/after status for instructions in that block.  I think I would prefer instruction iteration if we are going to try optimizing this, since splitting and joining blocks will have to iterate the instructions in the block in any case (at least the ones following the memcpy).  Additionally, the full dom tree is not necessary, nor is full reachable list.  These could be built manually and cached on the way from the blocks containing users by following predecessors.  One then just keeps track of reachable and dominated properties for each cached block.  Even if we use the existing dom tree, we still need reachability, which could have been computed in parallel.

Perhaps we can leave optimization of this to another step (this was already computing post-dom tree on the fly before), and then there are other higher-level refactor changes that might make the local exercise moot anyway.

pow2clk: Fair enough. If it was easy I wanted to suggest it. It's clearly not and, as you say, the price was being paid before so we're no worse off

pow2clk (approved):
Looks good!

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3587,32 +3588,40 @@ static bool ReplaceUseOfZeroInitEntry(Instruction *I, Value *V) {
...
-                                    PostDominatorTree &PDT) {
+// If a V user is dominated by memcpy (I),
+//    skip it - memcpy dest can simply alias to src for this user.
+// If the V user may follow the memcpy (I),
+//    return false - memcpy dest not safe to replace with src.
+// Otherwise,
+//    replace use with zeroinitializer.
+static bool ReplaceUseOfZeroInit(Instruction *I, Value *V,
+                                 DominatorTree &DT,
+                                 SmallPtrSet<BasicBlock*, 8> &Successors) {
```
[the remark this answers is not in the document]
tex3d: Meant to, but missed this function for some reason, thanks for the catch.

tex3d:
Ubuntu build timed out again after supposedly 90 min, but with only 28 seconds in the log... Everything else passed, so I'm overriding.

Issue #3807, opened 2021-05-31
[SPIR-V] Sample and Load offsets fail to build


pearcewf:
The following code using sampling offsets fails with the latest tagged release, as well as the latest builds from Appveyor. It is also present with Load offsets. This worked in previous releases of DXC.

Simple replication case:

```
Texture2D textureToLoad	: register(t0);

SamplerState bilinearClamp	: register(s0);

struct VertexOut
{
	float4 posH : SV_POSITION;
	float2 tex : TEXCOORD;
};

float4 sampleNeighbor(float2 uv, int2 offset)
{
	return textureToLoad.SampleLevel(bilinearClamp, uv, 0, offset);
}

float4 main(VertexOut pIn) : SV_TARGET
{
	//float4 initialColor = textureToLoad.Load(int3(pIn.posH.xy, 0));
	//float4 color = initialColor;

        // fine
	float4 a = textureToLoad.SampleLevel(bilinearClamp, pIn.tex, 0, int2(1, 1));
	return a;

        // fails
	//const int2 offset = int2(1, 1);
	//float4 b = textureToLoad.SampleLevel(bilinearClamp, pIn.tex, 0, offset);
	//return b;

        // fails
	//float4 c = sampleNeighbor(pIn.tex, int2(1, 1));
	//return c;
}
```

Error message from latest tagged release (v1.6.2104)

```
Error Message:
Shaders/Test_PS.hlsl:25:21: error: Use constant value for offset (SPIR-V spec does not accept a variable offset for OpImage* instructions other than OpImage*Gather)
```

Error message from Appveyor build

```
Error Message:
fatal error: generated SPIR-V is invalid: [VUID-StandaloneSpirv-Offset-04663] Image Operand Offset can only be used with OpImage*Gather operations
  %58 = OpImageSampleExplicitLod %v4float %57 %54 Lod|Offset %float_0 %55

note: please file a bug report on https://github.com/Microsoft/DirectXShaderCompiler/issues with source code if possible
```

Was this a work-around that has been disabled recently? I haven't updated my local DXC since mid-last year, so it's hard to tell when this might have been introduced.

jaebaek:
spirv-val recently added a validation rule to restrict only the constant offset to be used as `Offset` Operand of `OpImageSampleExplicitLod`.
[...]

PR #3820, opened 2021-06-10
[spirv] set needsLegalization as true when variable offset is found

jaebaek:
When we find a variable offset used for the texture sampling, we have to
- Run legalization if `-fcgl` is not enabled.
- Enable `--before-legalize-hlsl` for spirv-val if `-fcgl` is enabled.
to match the behavior of the DXIL backend.

In order to do that, we set SpirvEmitter::needsLegalization as true when
we find a variable offset used for the texture sampling. SpirvEmitter
will determine enabling `--before-legalize-hlsl` or running the
legalization depending on `-fcgl`.


jaebaek:
For the motivation of this PR, please check https://github.com/microsoft/DirectXShaderCompiler/issues/3807#issuecomment-858244516

sudonatalie (approved):

PR #3825, opened 2021-06-14
QuadRead Test fix dependencies and groupIndex

pow2clk:
Add needed check for wave ops needed for QuadRead* operations.

Correct use of GroupIndex to use WaveLaneIndex since that's how quad
relationships can be ensured. That required changing some other aspects
of the test as well.


hekota (approved):
LGTM!

Issue #3016, opened 2020-07-01
Assert/hang in SROA_HLSL pass related to empty base struct


simontaylor81:
We're hitting an assert (which manifests as a hang outside of the debugger) in the SROA_HLSL pass in one of our (generated) shaders. It appears to be related to having a struct with a base struct that is empty (but not completely empty -- it must have members, but those members are empty, so the total size is zero).

It's something related to the clang empty-base-class optimisation, which makes the struct layouts of the base and child look different, which confuses the code in `SROA_Helper::RewriteBitCast`. Afraid I got a bit lost after that :).

Anyway, to repro, compile the following with `-E main -T vs_6_0`:

```
struct EmptyStruct
{
};

struct BaseStruct
{
	EmptyStruct emptyStructMember;

	//float makeMeNotEmpty;	// Uncomment to prevent assert
};

struct ChildStruct : BaseStruct
{
	float4 m_childMember;

	float4 func() { return m_childMember; }
};

float4 main() : SV_Position
{
	ChildStruct var_1;
	EmptyStruct var_2;
	var_1.emptyStructMember = var_2;	// Comment out to prevent assert
	var_1.m_childMember = 0;
	return var_1.func();
}
```

I've marked with comments a couple of possible workarounds, which perhaps help explain what's going on to someone with a greater understanding of what that pass is trying to do :). As I said the code is generated, so using the workarounds in our actual shader is not totally straightforward, but I can probably hack something together.

Here's the callstack of the assert:
```
dxcompiler.dll!`anonymous namespace'::SROA_Helper::RewriteBitCast(llvm::BitCastInst * BCI) Line 2714
	at e:\dev\git\dxcpackage\external\directxshadercompiler\lib\transforms\scalar\scalarreplaggregateshlsl.cpp(2714)
dxcompiler.dll!`anonymous namespace'::SROA_Helper::RewriteForScalarRepl(llvm::Value * V, llvm::IRBuilder<1,llvm::ConstantFolder,llvm::IRBuilderDefaultInserter<1> > & Builder) Line 2964
	at e:\dev\git\dxcpackage\external\directxshadercompiler\lib\transforms\scalar\scalarreplaggregateshlsl.cpp(2964)
[...]

Issue #3770, opened 2021-05-14
dxc crashes on empty embedded struct


danbrown-amd:
Confirmed for [b44890fe1e](https://github.com/microsoft/DirectXShaderCompiler/tree/b44890fe1e0c02669b7290d7eab573ea35f896b2).

```hlsl
// empty_embedded_struct.hlsl
struct s0 {};
struct s1 { s0 a; uint b; };
s1 main() : OUT { s1 s; return s; }

// $ dxc -T vs_6_2 -E main -Fc empty_embedded_struct.dxil empty_embedded_struct.hlsl
// Internal compiler error: Terminal Error 0x80000003
```

The crash occurs during the SROA pass.  The lookup near the end of `SROA_Parameter_HLSL::allocateSemanticIndex()` (`Type *semanticTy = semanticTypeMap[semantic];`) fails because the value of `semantic` is the pseudo-semantic `*` rather than `OUT`, and `AllocateSemanticIndex()` attempts to dereference the resulting null pointer.

PR #3827, opened 2021-06-15
Handles empty embedded struct as first field of struct.

danbrown-amd:
Addresses #3770.


pow2clk (approved):
This also fixes #3016, which was assigned to me! Thanks!

PR #3828, opened 2021-06-15
Enables HLSL bitfields.

danbrown-amd:


tools/clang/test/HLSL/cpp-errors-hv2015.hlsl
```diff
@@ -51,10 +51,6 @@ __real g___real; // expected-error {{expected unqualified-id}}
 __FUNCTION__ g___FUNCTION; // expected-error {{expected unqualified-id}}
 __PRETTY__ g___PRETTY; // expected-error {{unknown type name '__PRETTY__'}}
 
-struct s_with_bitfield {
-  int f_bitfield : 3; // expected-error {{bitfields are not supported in HLSL}}
-};
-
```
pow2clk, line 57: We need to keep this failure for 2015 as it's the fxc compatibility target. 2018 is fuzzier. I don't anticipate any conflicts with existing code, so maybe it's okay to enable there, but since we need to check the version anyway, maybe we just check for 2021

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields_error.hlsl
```diff
@@ -1,6 +1,70 @@
-// RUN: %dxc /T vs_6_0 /E main %s | FileCheck %s
+// RUN: %dxc /T vs_6_2 /E main %s | FileCheck %s
```
tex3d, line 1: This file seems more suitable as a VerifierTest.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields_error.hlsl
```diff
@@ -1,6 +1,70 @@
...
+
+int a[sizeof(S) == 1 ? 1 : -1];
+
+enum E {};
+
+struct Z {};
+typedef int Integer;
+
+struct X {
+  // clang: error: anonymous bit-field
```
tex3d, line 31: Various errors are formatted "clang:" rather than "CHECK:", meaning these will not be checked.  Is this intentional?  Why if so?

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields_error.hlsl
```diff
@@ -1,6 +1,70 @@
...
+  enum E : 1;
+  // CHECK: error: non-integral type 'Z' is an invalid underlying type
+  enum E : Z;
```
tex3d, line 34: One of my primary concerns is the potential collision with semantic parsing.  We need tests that explore this area of intersection in the syntax.  Such as when someone does:
```
static const int bfsize = 3;
...
int foo : bfsize;
```
`bfsize` could be interpreted as a semantic, or a const bitfield size.  I think we wanted to disallow all cases except the obvious parsed literal integers for bitfields in HLSL in order to avoid the ambiguity.

tex3d: Note: this could also have a bad interaction with templates for when the bitfield size is a template argument.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields_error.hlsl
```diff
@@ -1,6 +1,70 @@
...
+};
+
+struct Y {
+  // clang: error: anonymous bit-field
+  enum E : int(2);
```
tex3d, line 39: Is `int(2)` supported?  I see this case should already be illegal due to the anonymous bitfield, but there's no other test that exercises this `int(2)` case without the other error getting in the way.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields_error.hlsl
danbrown-amd: Yes.  I left those to point out where clang expects an error for C++ that does not show up in the HLSL test.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
```diff
@@ -0,0 +1,139 @@
...
+// CHECK: %struct.P1 = type { i32, i32 }
+
+RWByteAddressBuffer BufferOut : register(u0);
+
+// The store of p.y into the temporary was not
+// getting extended to 32 bits, so uninitialized
+// bits of the temporary were used.  7366161.
+struct foo {
+  int x : 8;
+  int : 8;
```
tex3d, line 14: Wait, so anonymous bitfield *is* supported?  What's with the (unchecked) errors in the other test that read `clang: error: anonymous bit-field`?

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields_error.hlsl
danbrown-amd: I've preserved the status quo, in which an identifier after a colon in that context is always interpreted as a semantic.  That seems okay, though perhaps a warning could be issued if there is a homonymous identifier.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
danbrown-amd: I suspect that dxc, unlike clang, does implicit type conversions (to the basis type) in those cases.  Would there be a legitimate reason for that?

tools/clang/test/HLSL/cpp-errors-hv2015.hlsl
danbrown-amd: Should I revert cpp-errors as well, or specify -HV 2021?

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields_error.hlsl
danbrown-amd: The replacement PR will make it a VerifierTest.

danbrown-amd:
Superseded by #3847.

PR #3833, opened 2021-06-17
Fix atomics resource declaration issue

PrestonHamlin:
Resources used as raw RWByteAddressBuffer but created as structured could see conflicting allocation and stride values, resulting in OOB access in some configurations.

Setting the stride to 0 (I like being explicit) and adding the raw/typeless qualifiers, I no longer see OOB accesses.


pow2clk (approved):
LGTM!

PR #3838, opened 2021-06-18
Replace Travis CI with Azdo

pow2clk:
Travis went from org to com. In spite of the perfectly reasonable
assumption and explicitly stated expectation of the simplicity that
transition would involve, multiple engineers from multiple companies
failed to decipher their arcane instructions and permissions. Instead,
we've switched from leeching off of a well-meaning small outfit trying
to promote open source projects at no charge to leeching off Microsoft's
own ample resources.

This replaces all of the testing of travis with Azure Dev Ops (azdo).
The version of Ubuntu is bumped from 14.04 to 18.04, but that was
overdue. The corresponding clang and gcc versions are incremented as
well from 5 to 9 and 5 to 7 respectively with an eye toward the lowest
supported version in the current OS. The MacOS version is similarly
incremented with the clang version jumping to 11. In all cases, these
versions are the lowest azdo provides on the respective platforms.

Left out for now is the address sanitizer configuration, which was
never made blocking. The longevity of this decision is pending cross
company discussions.


azure-pipelines.yml
```diff
@@ -3,32 +3,50 @@
 # Add steps that build, run tests, deploy, and more:
```
pow2clk, line 3: In spite of the fact that at least some of this file is pre-existing, it was never subjected to review. Feel free to comment on any part of it

hekota (approved):
LGTM! If it works well maybe be can replace AppVeyor as well?

pow2clk:
The idea is to replace Appveyor as well, but the lack of travis was the clear and present danger

PR #3843, opened 2021-06-22
Added a way to specify resource binding through an external file.

adam-yang:


include/dxc/DXIL/DxilResourceBinding.h
```diff
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 //                                                                           //
-// DxilResourceBinding.h                                                     //
+// DxcBindingTable.h                                                        //
```
tex3d, line 3: Is this an accidental change?

adam-yang: Yes Oops.

lib/DxcBindingTable/DxcBindingTable.cpp
```diff
@@ -0,0 +1,462 @@
...
+      bool hasQuote = false;
+      if (Peek() == '"') {
+        hasQuote = true;
+        Advance();
+      }
+
+      while (!ReachedEnd()) {
+        if (IsDelimiter(Peek())) {
+          if (hasQuote)
+            return Error("Expected end quote '\"'.");
```
tex3d, line 160: So you can't have a comma inside the quotes?  I feel like only newline should cause an error when inside quotes.

adam-yang: That's a mistake. It's been fixed now and tested.

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -515,7 +524,8 @@ class DxilLowerCreateHandleForLib : public ModulePass {
     if (DM.GetCBuffers().size())
       bChanged |= PatchTBuffers(DM);
 
-
+    // Assign resource binding overrides.
+    hlsl::ApplyBindingTableFromMetadata(DM);
```
pow2clk, line 528: Because you use this here, you need to add a dependency to the new DxcBindingTable LLVM lib in the LLVMBuild.txt file in the lib/HLSL directory. 

The linux builds are failing because cmake generates linking commands according to library interdependencies and the linux linker is persnickety about the order they are in while other platforms don't care provided everything you need is in there somewhere.

dmpots:
Why use a custom csv-like format here? Why not a standard format like json or yaml both of which are supported by existing tools and can be parsed using the llvm yaml library?

lib/HLSL/DxilCondenseResources.cpp
adam-yang: Thank you Greg! Worked this time!

I thought I just had to add the dependency in the target binary's CMakeLists. It didn't occur to me I had to add a dependency in LLVMBuild.txt.

adam-yang:
> Why use a custom csv-like format here? Why not a standard format like json or yaml both of which are supported by existing tools and can be parsed using the llvm yaml library?

I considered using JSON, but decided it was too complicated and permissive for this purpose. It's also an additional dependency to pull in. The LLVM YAML parser is also ~2.5k lines of code to compile into the binary, compared to the far simpler custom CSV parser.

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -380,6 +381,14 @@ static bool LegalizeResourceArrays(Module &M, DxilValueCache *DVC) {
   return Changed;
 }
 
+typedef std::unordered_map<std::string, DxilResourceBase *> ResourceMap;
+template<typename T>
+static inline void GatherResources(const std::vector<std::unique_ptr<T> > &List, ResourceMap *Map) {
```
tex3d, line 386: Where is this used?  I see a copy of this in DxcBindingTable, so is this leftover from a refactor?

tex3d (approved):
Looks good.  Only one seemingly unused template function that was added, likely leftover from before a refactor.  See my most recent comment.

Issue #2319, opened 2019-07-09
File system handle has scaling limitations and other problems


tex3d:
`DxcArgsFileSystemImpl` implementation with `DxcArgsHandle` has a number of limitations that need addressing, for reliability and efficiency.

1. `DxcArgsHandle` uses union of `HANDLE` (typedef'd to `void*`) and `HandleBits` as file handle.  `HANDLE` is pointer sized, but is frequently cast to/from `int` by file system APIs, since this is used as the `int fd` file descriptor.
2. `struct HandleBits` is defined as three `unsigned` bitfields.  The first is `unsigned Offset : 8;` which is used for the index into a single vector of all files and directories ever opened/added.  At one point, the code wouldn't add more than 200 due to the value of `MaxIncludedFiles`, but that value was upped to 1000 without changing the code that couldn't handle more than 256 values due to this 8-bit field.  This will result in silent wrap-around of this index in the file handle, causing who-knows-what undefined behavior later.
3. The second field of `HandleBits` is `unsigned Length : 8`.  This 8-bit field holds directory path lengths.  8-bits seems too small for this as well.  Even through it may seem rare to see paths this long, it's not out of spec and certainly could happen, and once again we have a silent overflow that could result in who-knows-what behavior.
4. Casing and path separators are not being normalized, so we don't match paths or filenames that are the same like we should if these differ - leading to additional handle space usage.
5. Linear searches abound here, and need to be replaced with a better system using DenseMap or something.
6. File handles are 1:1 with a `ReadOnlyBlobStream` instance, and the handle returned for `CreateFileW` will be the same for the same filename (on the same thread), even though different calls should return different handles.  Handles should not be a single value tied to a single stream, but each file handle should be tied to a different stream instance, even if it's the same file (they could both ref the same underlying file contents blob though).

In general, this system needs some real attention/refactor work.

otanter-at-ubi:
Hi,
regarding point 2, I could indeed experiment the wrap around, as described in issue #2731 (where you can find an easy repro). The end result is that the shader compilation fails and returns rather unclear errors.

PR #3846, opened 2021-06-23
Fixed HandleBits struct limitations due to 8bits fields

### LINKED ISSUE #2731: Shader compilation fail with 256 includes or more.
Hello,

I have some generated shaders that fail to compile because they have more than 255 includes. 
A fix was done related to MaxIncludeFiles (= 1000 currently), see issue [#866](https://github.com/microsoft/DirectXShaderCompiler/issues/866).
But it seems we also need to update the Offset bit size in HandleBits struct in dxcfilesystem.cpp: 
```
struct HandleBits {
  unsigned Offset : 8;
  unsigned Length : 8;
  unsigned Kind : 4;
};
```
Am I right?

issue can be easilty tested with this cpp that generate a simple main.fx that include 260 header files with default parameters : [GenerateHlslManyHeaders.txt](https://github.com/microsoft/DirectXShaderCompiler/files/4274842/GenerateHlslManyHeaders.txt)

Thanks,
Olivier

edit: we have tested with 16 bits Offset and it indeed fixes the shader compilation.

otanter-at-ubi:
- use 10 bits for Offset to support MaxIncludedFiles include files.
- use 16 bits for Length to support nearly arbitrary path length.

Partially addresses #2319 (point 2 and 3)
Fixes #2731

### DIFF
--- a/tools/clang/tools/dxcompiler/dxcfilesystem.cpp
+++ b/tools/clang/tools/dxcompiler/dxcfilesystem.cpp
@@ -81,9 +81,12 @@ enum class SpecialValue {
   Source = 3,
   Output = 4
 };
+
+// We use 10 bits for Offset to support MaxIncludedFiles include files,
+// and we use 16 bits for Length to support nearly arbitrary path length.
 struct HandleBits {
-  unsigned Offset : 8;
-  unsigned Length : 8;
+  unsigned Offset : 10;
+  unsigned Length : 16;
   unsigned Kind : 4;
 };
 struct DxcArgsHandle {


meeths (approved):

otanter-at-ubi:
Is appveyor waiting on the missing review or is it something unrelated?

pow2clk (approved):
Looks good! Thanks for the contribution!

pow2clk:
The lack of appveyor respons is odd. The test passed, so I'm ignoring it

PR #3847, opened 2021-06-23
Enables HLSL bitfields.

danbrown-amd:
Supersedes #3828.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
@@ -1,0 +1,135 @@
+// RUN: %dxc /Zi /T vs_6_2 /E main %s | FileCheck %s
+
+RWByteAddressBuffer BufferOut : register(u0);
+
+// The store of p.y into the temporary was not
+// getting extended to 32 bits, so uninitialized
+// bits of the temporary were used.  7366161.
+struct foo {
+  int x : 8;
+  int : 8;
+  int y : 16;
+};
+
+int main(struct foo p : IN0, int x : IN1) : OUT {
+  // CHECK: %[[val:.*]] = {{.*}}loadInput{{.*}} i32 0, i32 0, {{.*}}
+  // CHECK:               ashr i32 %[[val]]
+  BufferOut.Store(p.x, asuint(p.y));
+  x = (p.y > x ? x : p.y);
+  return x;
+}
+
+struct M {
+  uint f0;
+  uint bf0 : 8;
+  uint bf1 : 24;
+  uint f1;
+  uint : 8;
+  uint bf2 : 16;
+  uint : 8;
+  uint f2;
+};
+
+void m(uint3 xyz : IN) {
+  M m;
+
+  m.f0 = 0;
+  m.bf1 = xyz.x;
+  m.f1 = xyz.y;
+  m.bf2 = xyz.z;
+  m.f2 = 1;
+  BufferOut.Store(m.f1*16, asuint(m.bf1 + m.bf2));
+  return;
+}
+
+struct S {
+  uint : 8;
+  uint : 8;
+  uint x : 8;
+} s;
+// CHECK: !DIDerivedType(tag: DW_TAG_member, name: "x", {{.*}}, size: 8, align: 32, offset: 16)
+
+// Below adapted from clang test suite.
+
+// PR14638; make sure this doesn't crash.
+struct A {
+  bool m_sorted : 1;
+};
+
+void func1(bool b, A a1)
+{
+  if ((a1.m_sorted = b)) {}
+}
+
+struct TEST2
+{
+  int subid:32;
+  int :0;
+};
+
+typedef struct _TEST3
+{
+  TEST2 foo;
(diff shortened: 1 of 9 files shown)


tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
```diff
@@ -0,0 +1,137 @@
...
+struct foo {
+  int x : 8;
+  int : 8;
+  int y : 16;
+};
+
+int main(struct foo p : IN0, int x : IN1) : OUT {
+  // CHECK: %1 = {{.*}}loadInput{{.*}} i32 0, i32 0, {{.*}}
+  // CHECK:      ashr i32 %1
+  // %[[val:.*]] = {{.*}}loadInput{{.*}} i32 0, i32 0, {{.*}}
```
tex3d, line 17: Above, your CHECK is strict about the value label, which could cause failures depending on whether this is run on a Release or Debug version of the compiler.  Here you have a more general version, but it's not used.  Was this intended to replace the above CHECK: contents?

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
```diff
@@ -0,0 +1,137 @@
+// RUN: %dxc /Zi /T vs_6_2 /E main %s | FileCheck %s
```
tex3d, line 1: When using CHECK statements with /Zi you must be careful add a label for the embedded source, otherwise your CHECK statement might match itself from the source instead of the IR you intended to check.
See https://github.com/microsoft/DirectXShaderCompiler/blob/master/tools/clang/test/HLSLFileCheck/Readme.md

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
danbrown-amd: Yes, that was a temporary debugging change I'd forgotten about.  I've added the check to exclude the quoted source file as well.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
pow2clk: FWIW, if all your checks use wildcards, and I think yours do, this isn't an issue.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/bitfields.hlsl
```diff
@@ -0,0 +1,138 @@
...
+};
+
+bool f() : OUT
+{
+  F f = { true, true };
+
+  return f.b1 != 1;
+}
+
+// Disabled until #3770 gets fixed.
```
pow2clk, line 129: I have merged your fix for #3770. Could we try enabling this?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -4981,10 +4990,12 @@ void SROA_Parameter_HLSL::flattenArgument(
...
           }
 
-          // Inherit semantic from parent, but only preserve it for the first element.
+          // Inherit semantic from parent, but only preserve it for the first
+          // non-zero-sized element.
           // Subsequent elements are noted with a special value that gets resolved
           // once the argument is completely flattened.
-          EltAnnotation.SetSemanticString(i == 0 ? semantic : ContinuedPseudoSemantic);
+          EltAnnotation.SetSemanticString(
+            i == firstNonEmptyIx ? semantic : ContinuedPseudoSemantic);
```
pow2clk, line 4998: oh shoot! I merged this into the main branch, but I need to merge this change into hlsl2021. I forgot that this was in hlsl2021

danbrown-amd: I've cherry-picked it into this PR already.

pow2clk: yeah, but I don't want those conflicts when we do merge. I've created a merge here: #3850  sorry for leading you down this path only to draw things out 😣

danbrown-amd: No problem, I'll rebase again.

pow2clk: I've  updated the hlsl-2021 branch. You should be able to rebase now

danbrown-amd: Done, but the Ubuntu build failed due to timeout.

pow2clk:
The Ubuntu test that timed out is redundant for purposes of verification. It remains (for now) for artifact retention. We can ignore it. 

I'm satisfied with the change as is, but I wanted to give @tex3d a chance to weigh in if he still has concerns.

I don't think I've yet said how much we appreciate this contribution! This is a feature that has been long requested and it's very nice to see it so near completion!

danbrown-amd:
Thanks, Greg.  Glad to contribute!

PR #3848, opened 2021-06-24
Fix command line parsing on empty string.

LukasBanana:
This bug caused one of our cooks in UE5 to produce several of these error messages:
```
: CommandLine Error: Option '
```
I presume the last changed line was intended to be `Opt->ArgStr[0] != 0` instead of `!Opt->ArgStr[0]` as the error message is cropped. I replaced this and two other conditions with the appropriate function `hasArgStr()`.

### DIFF
--- a/lib/Support/CommandLine.cpp
+++ b/lib/Support/CommandLine.cpp
@@ -120,7 +120,7 @@ public:
 
   void addOption(Option *O) {
     bool HadErrors = false;
-    if (O->ArgStr[0]) {
+    if (O->hasArgStr()) {
       // Add argument to the argument map!
       if (!OptionsMap.insert(std::make_pair(O->ArgStr, O)).second) {
         errs() << ProgramName << ": CommandLine Error: Option '" << O->ArgStr
@@ -153,7 +153,7 @@ public:
   void removeOption(Option *O) {
     SmallVector<const char *, 16> OptionNames;
     O->getExtraOptionNames(OptionNames);
-    if (O->ArgStr[0])
+    if (O->hasArgStr())
       OptionNames.push_back(O->ArgStr);
     for (auto Name : OptionNames)
       OptionsMap.erase(StringRef(Name));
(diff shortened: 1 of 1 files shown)


lib/Support/CommandLine.cpp
```diff
@@ -859,7 +859,7 @@ void CommandLineParser::ParseCommandLineOptions(int argc,
               "error - this positional option will never be matched, "
               "because it does not Require a value, and a "
               "cl::ConsumeAfter option is active!");
-      } else if (UnboundedFound && !Opt->ArgStr[0]) {
+      } else if (UnboundedFound && Opt->hasArgStr()) {
```
pow2clk, line 862: I think this should be `!Opt->hasArgStr()`?

LukasBanana: Then line 871 would terminate the error message, which is exactly what happend on our side:
https://github.com/microsoft/DirectXShaderCompiler/blob/8d9e93c8c9dc95ddb5c708b1dc8af5bb33709fa7/lib/Support/CommandLine.cpp#L871
This is why I presume it's a bug and my change fixes it.

pow2clk: I see your point. I looked at the current code in upstream llvm since this is a common file and while they have switched to `Opt->hasArgStr()` as you have here, the negation is still present: 

https://github.com/llvm/llvm-project/blob/76f734040a54622b847729e5acb4480e667d2c74/llvm/lib/Support/CommandLine.cpp#L1407

As I see it, the intent of this error is to identify when an option with an arbitrary number of arguments is followed by an option with no required argument. The `!Opt->hasArgStr()` check is what determines that the second option doesn't require an argument. I think it's correct.

There is still a bug here of course in the attempt to print out what is known to be an empty string. Instead, I think it should be printing the name of the option itself.  Maybe ` << Opt->getName()`. This bug seems to be in upstream LLVM as well.

LukasBanana: Good finding. I can update the PR next week when I'm back from vacation. Thanks for taking a closer look.

LukasBanana: I updated the PR and now print the entire command line when this error occurs. There is no such thing as `Option::getName` and without an argument string it would be hard to see what caused the issue, so displaying the entire command line might help track down what caused the issue especially when DXC is executed many thousand times like in our case.

pow2clk (approved):
I like this solution!

PR #3851, opened 2021-06-25
Update version to 1.6.2106

hekota:


pow2clk (approved):
*sigh* sorry about the mac build. I haven't added the provisions to detect whether it's a legit failure or just the flaky macos agents. You can safely ignore it if you want to

PR #3858, opened 2021-07-01
Add /bigobj flag on big spirvtest object

pow2clk:
This build was failing on windows debug because recent additions pulled
it over a size threshold. This flag is used elsewhere in the compile
already and works fine here.

### DIFF
--- a/tools/clang/unittests/SPIRV/CMakeLists.txt
+++ b/tools/clang/unittests/SPIRV/CMakeLists.txt
@@ -44,6 +44,10 @@ target_include_directories(clang-spirv-tests
 set_output_directory(clang-spirv-tests
   ${LLVM_RUNTIME_OUTPUT_INTDIR} ${LLVM_LIBRARY_OUTPUT_INTDIR})
 
+if(WIN32)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj") # otherwise will hit fatal error C1128 on windows x64
+endif(WIN32)
+
 add_test(NAME test-spirv-codegen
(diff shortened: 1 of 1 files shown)


hekota (approved):
Looks good!

pow2clk:
@jaebaek just FYI, an alternative approach would be to break up CodeGenSpirvTest.cpp into a few files. It's not a big deal to use this flag, but I thought I'd let you know our options

jaebaek:
Thank you for heads-up. I will consider the idea of slitting CodeGenSpirvTest.cpp next week. This change looks good to me. If you think it is urgent issue, please submit this PR.

jaebaek (approved):

pow2clk:
It's not terribly urgent. I'll leave it for now. I encountered this when trying to add a debug windows build to automated testing, but that's being held up for other reasons

jaebaek:
If the sizes of object files matter, I guess not only `CodeGenSpirvTest.cpp` but also `tools/clang/lib/SPIRV/SpirvEmitter.cpp` matters. Note that `SpirvEmitter.cpp` has 12473 LoC.

While splitting `CodeGenSpirvTest.cpp` into multiple files is a relatively simple task, breaking up `SpirvEmitter.cpp` into multiple files is not trivial. If it is possible, we want to delay it because it would be a significant refactoring task.

If it is not critical, can we leave it for now?

pow2clk:
It's not critical. We can definitely use this solution. The selection of `CodeGenSpirvTest.cpp` wasn't arbitrary, that's the one that fails to build in debug mode on windows. Thanks for looking into it

jaebaek:
Thanks. I just opened #3867 to handle it later.

PR #3864, opened 2021-07-07
Amend hctdb_instrhelp.py for case-sensitive filesystems

tcorringham:
Trivial changes to the case of some file names used by hctdb_instrhelp.py
to match the actual case of the files in the repo. This fixes problems
experienced when using the script on Linux.


pow2clk (approved):
Looks good! Thanks!

PR #3866, opened 2021-07-07
Warns when semantic name same as identifier.

danbrown-amd:
Intent is to avoid confusion where C++ syntax would allow named constant as bitfield width, which is disallowed in HLSL because of conflict with syntax for semantics.

### DIFF
--- a/tools/clang/lib/Parse/ParseDecl.cpp
+++ b/tools/clang/lib/Parse/ParseDecl.cpp
@@ -29,6 +29,7 @@
 #include "llvm/ADT/StringSwitch.h"
 #include "dxc/Support/Global.h"       // HLSL Change
 #include "clang/Sema/SemaHLSL.h"      // HLSL Change
+#include "clang/AST/UnresolvedSet.h"  // HLSL Change
 #include "dxc/DXIL/DxilShaderModel.h" // HLSL Change
 #include "dxc/DXIL/DxilConstants.h"   // HLSL Change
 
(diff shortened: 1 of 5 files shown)


tools/clang/lib/Parse/ParseDecl.cpp
```diff
@@ -619,6 +620,23 @@ bool Parser::MaybeParseHLSLAttributes(std::vector<hlsl::UnusualAnnotation *> &ta
...
+                       Sema::LookupOrdinaryName);
+        if (Actions.LookupName(R, getCurScope())) {
+          for (UnresolvedSetIterator I = R.begin(), E = R.end(); I != E; ++I) {
+            NamedDecl *D = (*I)->getUnderlyingDecl();
+            if (D->getKind() == Decl::Kind::Var) {
+              VarDecl *VD = static_cast<VarDecl *>(D);
+              if (VD->getType()->isIntegralOrEnumerationType()) {
+                unsigned diagId = context.getDiagnostics().getCustomDiagID(
+                    DiagnosticsEngine::Level::Warning,
+                    "'%0' interpreted as semantic; previous definition(s) ignored");
```
tex3d, line 634: Great to have the warning!

I have a couple thoughts:
- Shouldn't we call a function implemented in SemaHLSL.cpp to do name collision lookup and warning reporting?
- I think we need an explicit warning defined in DiagnosticParseKinds.td or DiagnosticSemaKinds.td (depending on who's responsible for the warning based on previous point), with a new group to allow disabling the warning with a compiler argument or pragma (see end of DiagnosticGroups.td for group definitions, and use `InGroup<HLSLSemanticIdentifierCollision>` in warning def in the .td file).

New group could be:
```
def HLSLSemanticIdentifierCollision: DiagGroup<"semantic-identifier-collision">;
```

danbrown-amd: To me, this seems like more of a parsing issue, in that it potentially produces an unexpected AST, and so it makes sense to have it in ParseDecl.  However, I don't claim to have a deep understanding of the division of labor between Parse and Sema, so I could perhaps be convinced otherwise.

tex3d: Regardless, my primary concern is to have a new warning group and explicit new warning defined in the appropriate .td file for whichever side handles it, rather than an on-the-fly warning with no group to allow disabling.

danbrown-amd: OK, done.

tex3d (approved):
Looks good to me!

PR #3879, opened 2021-07-14
Add 'fourf16' as a valid OpFunc (for future changes)

bfavela:


pow2clk (approved):
Doesn't do much yet, but it looks good!

PR #3888, opened 2021-07-26
Re-enable a swath of DXIL tests for *nix builds

pow2clk:
These were disabled at a point that the DXC CI didn't reject builds for
failing on Linux, now that they do, we can enable these tests sure in
the knowledge that any changes to them that break on Linux will be
detected and fixed quickly

Fixed *nix error code return. The previous caused an infinite loop of retries.
This definition more accurately identifies the error that has occurred as not
benefiting from repeated attempts

Skip tests and filechecker elements that entirely depend on reflection or
linker interfaces that aren't implemented on *nix yet.

Correct capitalization in a few places to satisfy a case-sensitive filesystem

Detect directories from test include handler. *nix virtual filesystem will try to
load the file using the include handler when performing stat() on a directory.
The test include handler is pretty basic and would provide the next file, leaving
it without a file to offer when the real include came along.
By detecting a name that looks like a dir and returning failure, it
correctly indicates that the file isn't found by the handler and saves
the virtual file for when its needed.

Don't mark non-debug builds as such. The DEBUG define determines which
tests get run. Some of which depend on consistencies that can only be relied
upon in debug builds. So DEBUG is disabled for all non-debug builds.

Correct free of new() memory. Incidental regex allocations were changed for
HLSL to use the global new() operator, but were still freed using the standard
free(), but replacing the default free() with regex_free() allows the default new()
operator to be paired with the default delete() operator

Correct misnamed test flag. The *nix flag was mistakenly called InputFile instead
of InputPath. By renaming it, the ManualFile test can target a specific HLSL file.

Fix misused ArrayRef in legacy GV replacement. The replacement of the GV with
the legacy form used in libraries collected the arguments into an arrayref without
an array to ref, which caused garbage in release builds. Also moved to using end
vars for loops where the elements can get deleted during the loop. 

Fix *nix conversion to utf8. When an unknown codepage is found, it tries to convert
to utf16, but that isn't' actually helpful for *nix for a shader codepage derived from
the BOM, which is correctly identified as utf32


pow2clk (commented):
A few inline comments to try to explain the need for some of these changes where not obvious

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -1821,18 +1821,17 @@ bool UpdateStructTypeForLegacyLayout(DxilResourceBase &Res,
       Function *NF = hlslOP->GetOpFunc(hlsl::OP::OpCode::CreateHandleForLib, UpdatedST);
 
       // Replace old GV.
-      for (auto UserIt = Symbol->user_begin(); UserIt != Symbol->user_end();) {
+      for (auto UserIt = Symbol->user_begin(), userEnd = Symbol->user_end(); UserIt != userEnd;) {
```
llvm-beanz, line 1824: Is there a reason for this transformation? I feel like the loop was cleaner to read before this change.

llvm-beanz (approved):
This all looks reasonable to me. I had one stylistic comment inline.

lib/HLSL/DxilCondenseResources.cpp
pow2clk: It was an attempt to track down a different bug, but the user can get deleted in this loop and that is a condition where this form of the loop has been required before. Even though I couldn't get a crash or a memory misuse complaint here, this seems the safer route.

PR #3891, opened 2021-07-30
-strict-udt-casting prevents implicit casts between unrelated structs

### LINKED ISSUE #3563: [templates] ambigous function override
Hello

Templates are awesome. Just reporting a nices to have to minimize surprise when using template in HLSL.

Template instentiations allow to do function override through template specialisation, but the C++ syntax for function override is still a lot simpler. Can we have -enable-template to also allow C++ function override?

```
// RUN: %dxc -E main -T ps_6_0 -enable-templates %s

struct LinearRGB
{
	float3 RGB;
};

struct LinearYCoCg
{
	float3 YCoCg;
};

#if 1 // error: call to 'GetLuma4' is ambiguous

float GetLuma4(LinearRGB In)
{
	return In.RGB.g * 2.0 + In.RGB.r + In.RGB.b;
}

float GetLuma4(LinearYCoCg In)
{
	return In.YCoCg.x;
}

#else // works but code getting too complicated
	
template<typename InputColorSpace>
float GetLuma4(InputColorSpace In)
{
	//static_assert(sizeof(InputColorSpace) == 0, "GetLuma4() not implemented");
	float Array[!sizeof(InputColorSpace)];
	return Array[0];
}

template<>
float GetLuma4<LinearRGB>(LinearRGB In)
{
	return In.RGB.g * 2.0 + In.RGB.r + In.RGB.b;
}

template<>
float GetLuma4<LinearYCoCg>(LinearYCoCg In)
{
	return In.YCoCg.x;
}

#endif

float4 main() : SV_TARGET {
	LinearRGB Color;
	Color.RGB = float3(0, 1, 0);

	return float4(GetLuma4(Color), 0.0, 0.0, 0.0);
}
```

This produce this error:
```
ambigous_function_override.hlsl:54:16: error: call to 'GetLuma4' is ambiguous
        return float4(GetLuma4(Color), 0.0, 0.0, 0.0);
                      ^~~~~~~~
ambigous_function_override.hlsl:15:7: note: candidate function
float GetLuma4(LinearRGB In)
      ^
ambigous_function_override.hlsl:20:7: note: candidate function
float GetLuma4(LinearYCoCg In)
      ^
```

### LINKED ISSUE #3564: [template] have -enable-templates to disallow structure casting
Hello

Templates are awesome. Just reporting a nices to have to minimize surprise when using template in HLSL.

Legacy HLSL allow to cast structure that have same amount of register. But with the use of templated function, this can silently happen deep in a templated function. Could we have -enable-templates to start throwing errors instead?
```
// RUN: %dxc -E main -T ps_6_0 -enable-templates %s | FileCheck %s

struct LinearRGB
{
	float3 RGB;
};

struct LinearYCoCg
{
	float3 YCoCg;
};


LinearYCoCg RGBToYCoCg(LinearRGB In)
{
	float Y = dot(In.RGB, float3(1, 2, 1));
	float Co = dot(In.RGB, float3(2, 0, -2));
	float Cg = dot(In.RGB, float3(-1, 2, -1));

	LinearYCoCg Out;
	Out.YCoCg = float3(Y, Co, Cg);
	return Out;
}

LinearRGB YCoCgToRGB(LinearYCoCg In)
{
	float Y =  In.YCoCg.x * float(0.25);
	float Co = In.YCoCg.y * float(0.25);
	float Cg = In.YCoCg.z * float(0.25);

	float R = Y + Co - Cg;
	float G = Y + Cg;
	float B = Y - Co - Cg;

	LinearRGB Out;
	Out.RGB = float3(R, G, B);
	return Out;
}

template<typename OutputColorSpace, typename InputColorSpace>
OutputColorSpace TransformColor(InputColorSpace In)
{
	//static_assert(sizeof(InputColorSpace) == 0, "TransformColor() not implemented");
	float Array[!sizeof(InputColorSpace)];
	Array[0];
	return In;
}

template<>
LinearYCoCg TransformColor<LinearYCoCg, LinearRGB>(LinearRGB In)
{
	return RGBToYCoCg(In);
}


template<>
LinearRGB TransformColor<LinearRGB, LinearYCoCg>(LinearYCoCg In)
{
	return YCoCgToRGB(In);
}

float4 main() : SV_TARGET {
	LinearRGB Color;
	Color.RGB = float3(0, 1, 0);

	LinearYCoCg Color2 = TransformColor<LinearYCoCg>(Color);
	LinearRGB Color3 = TransformColor<LinearRGB>(Color2);
	
	// ISSUE: TransformColor<LinearRGB>() returns LinearRGB, but get cast silently in LinearYCoCg without error.
	LinearYCoCg Color4 = TransformColor<LinearRGB>(Color2);

	return float4(Color4.YCoCg, 0.0);
}
```

tex3d:
TODO: Enable this by default under -HV 2021

Fixes #3564.
Fixes #3563.

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -8631,7 +8631,7 @@ bool HLSLExternalSource::CanConvert(
     && sourceExpr->getStmtClass() != Expr::StringLiteralClass;
 
   bool targetRef = target->isReferenceType();
-  bool TargetUnnamed = false;
+  bool TargetIsAnonymous = false;
 
   // Initialize the output standard sequence if available.
   if (standard != nullptr) {
@@ -8723,7 +8723,7 @@ bool HLSLExternalSource::CanConvert(
         }
         // There is no way to cast to anonymous structures.  So we allow legacy
         // HLSL implicit casts to matching anonymous structure types.
-        TargetUnnamed = !targetRD->hasNameForLinkage();
+        TargetIsAnonymous = !targetRD->hasNameForLinkage();
       }
     }
 
@@ -8755,7 +8755,7 @@ bool HLSLExternalSource::CanConvert(
     } else if (m_sema->getLangOpts().StrictUDTCasting &&
                (SourceInfo.ShapeKind == AR_TOBJ_COMPOUND ||
                 TargetInfo.ShapeKind == AR_TOBJ_COMPOUND) &&
-               !TargetUnnamed) {
+               !TargetIsAnonymous) {
       // Not explicit, either are struct/class, not derived-to-base,
       // target is named (so explicit cast is possible),
       // and using strict UDT rules: disallow this implicit cast.
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -8631,6 +8631,7 @@ bool HLSLExternalSource::CanConvert(
     && sourceExpr->getStmtClass() != Expr::StringLiteralClass;
 
   bool targetRef = target->isReferenceType();
+  bool TargetUnnamed = false;
 
   // Initialize the output standard sequence if available.
   if (standard != nullptr) {
@@ -8694,7 +8695,6 @@ bool HLSLExternalSource::CanConvert(
   // Structure cast.
   SourceIsAggregate = SourceInfo.ShapeKind == AR_TOBJ_COMPOUND || SourceInfo.ShapeKind == AR_TOBJ_ARRAY;
   TargetIsAggregate = TargetInfo.ShapeKind == AR_TOBJ_COMPOUND || TargetInfo.ShapeKind == AR_TOBJ_ARRAY;
-  bool TargetUnnamed = false;
   if (SourceIsAggregate || TargetIsAggregate) {
     // For implicit conversions, FXC treats arrays the same as structures
     // and rejects conversions between them and numeric types
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -8694,6 +8694,7 @@ bool HLSLExternalSource::CanConvert(
   // Structure cast.
   SourceIsAggregate = SourceInfo.ShapeKind == AR_TOBJ_COMPOUND || SourceInfo.ShapeKind == AR_TOBJ_ARRAY;
   TargetIsAggregate = TargetInfo.ShapeKind == AR_TOBJ_COMPOUND || TargetInfo.ShapeKind == AR_TOBJ_ARRAY;
+  bool TargetUnnamed = false;
   if (SourceIsAggregate || TargetIsAggregate) {
     // For implicit conversions, FXC treats arrays the same as structures
     // and rejects conversions between them and numeric types
@@ -8720,6 +8721,9 @@ bool HLSLExternalSource::CanConvert(
           Second = ICK_HLSL_Derived_To_Base;
           goto lSuccess;
         }
+        // There is no way to cast to anonymous structures.  So we allow legacy
+        // HLSL implicit casts to matching anonymous structure types.
+        TargetUnnamed = !targetRD->hasNameForLinkage();
       }
     }
 
@@ -8748,6 +8752,14 @@ bool HLSLExternalSource::CanConvert(
           break;
         }
       }
+    } else if (m_sema->getLangOpts().StrictUDTCasting &&
+               (SourceInfo.ShapeKind == AR_TOBJ_COMPOUND ||
+                TargetInfo.ShapeKind == AR_TOBJ_COMPOUND) &&
+               !TargetUnnamed) {
+      // Not explicit, either are struct/class, not derived-to-base,
+      // target is named (so explicit cast is possible),
+      // and using strict UDT rules: disallow this implicit cast.
+      return false;
     }
 
     FlattenedTypeIterator::ComparisonResult result =
--- a/tools/clang/test/HLSLFileCheck/hlsl/functions/arguments/outputArray-strictudt.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/functions/arguments/outputArray-strictudt.hlsl
@@ -1,0 +1,26 @@
+// RUN: %dxc -E main -T vs_6_0 -strict-udt-casting %s | FileCheck %s
+
+// CHECK: switch
+
+struct Vertex
+{
+    float4 position     : POSITION0;
+    float4 color        : COLOR0;
+    float4 a[3]         : A;
+};
+
+struct Interpolants
+{
+    float4 position     : SV_POSITION0;
+    float4 color        : COLOR0;
+    float4 a[3]         : A;
+};
+
+uint i;
+
+void main( Vertex In, int j : J, out Interpolants output )
+{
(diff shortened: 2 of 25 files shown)


adam-yang (approved):
LGTM!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -8720,6 +8721,9 @@ bool HLSLExternalSource::CanConvert(
           Second = ICK_HLSL_Derived_To_Base;
           goto lSuccess;
         }
+        // There is no way to cast to anonymous structures.  So we allow legacy
+        // HLSL implicit casts to matching anonymous structure types.
+        TargetUnnamed = !targetRD->hasNameForLinkage();
```
pow2clk, line 8726: Minor, but I find boolean variables with negations in them confusing. `TargetNamed` and flip the logic?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -8748,6 +8752,14 @@ bool HLSLExternalSource::CanConvert(
           break;
         }
       }
+    } else if (m_sema->getLangOpts().StrictUDTCasting &&
+               (SourceInfo.ShapeKind == AR_TOBJ_COMPOUND ||
+                TargetInfo.ShapeKind == AR_TOBJ_COMPOUND) &&
+               !TargetUnnamed) {
```
pow2clk, line 8758: are there any builtin compound objects that could get implicitly cast?

tools/clang/test/HLSLFileCheck/hlsl/functions/arguments/outputArray-strictudt.hlsl
```diff
@@ -0,0 +1,26 @@
+// RUN: %dxc -E main -T vs_6_0 -strict-udt-casting %s | FileCheck %s
+
+// CHECK: switch
```
pow2clk, line 3: Not sure what this is meant to check

tools/clang/test/HLSLFileCheck/hlsl/types/cast/StructCast-strictudt.hlsl
```diff
@@ -0,0 +1,48 @@
...
+
+uint   RTIndex      : SV_RenderTargetArrayIndex;
+};
+
+struct Interpolants
+{
+    float4 position     : SV_POSITION0;
+    float4 color        : COLOR0;
+
+uint   RTIndex      : SV_RenderTargetArrayIndex;
```
pow2clk, line 18: nit: weird indentation

tools/clang/test/HLSLFileCheck/hlsl/types/struct/anonymous-strictudt.hlsl
```diff
@@ -0,0 +1,23 @@
...
+struct Outer
+{
+    struct { int x; } field;
+};
+
+int main(Outer input : IN) : OUT
+{
+    struct { int x; } local = input.field;
+    typedefed retval = (typedefed)local;
+    return retval.x + global.x;
```
pow2clk, line 22: I think that implicitly casting from a named struct to an unnamed should succeed. Is that correct? Is there a test for that?

tools/clang/test/HLSLFileCheck/shader_targets/library/inout_struct_mismatch-strictudt.hlsl
```diff
@@ -0,0 +1,36 @@
...
+void modify_ext(inout ParamStruct s);
+
+CallStruct g_struct;
+
+float4 main() : SV_Target {
+  CallStruct local = g_struct;
+  modify((ParamStruct)local);
+  modify_ext((ParamStruct)local);
+  return local.f;
+}
```
pow2clk, line 36: I'm curious how these test variants were chosen. Did you try running it with the flag enabled by default and identified the shaders that failed?

pow2clk (approved):
One question and a few comments. Looks good!

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d: Sure, I think originally had thought of it as "is anonymous" but it's based on this name for linkage property, so "Unnamed".  Alternatively, it could be "TargetAnon" Instead and don't flip the logic?

tex3d: Actually, I kind of like `TargetIsAnonymous` for clarity.

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d: Do you mean unnamed builtin compound objects?  No, I don't think so.  But if there were, I think we would want to preserve this behavior, since otherwise there would be no way to pass something to an intrinsic that used that type.

If that's not what you mean, I'm not sure what you're getting at.

tools/clang/test/HLSLFileCheck/hlsl/functions/arguments/outputArray-strictudt.hlsl
tex3d: It's a modified version of an existing test for something else, that runs with this flag on because it relied on implicit UDT casting.  I thought leaving the original version is helpful to make sure the prior behavior continues to work, and the new test makes sure we didn't break something when forced to explicitly cast in the same scenario.

tools/clang/test/HLSLFileCheck/hlsl/types/cast/StructCast-strictudt.hlsl
tex3d: Another copy of an existing test that has exactly the same contents, besides the extra flag and necessary explicit casting.  I could fix this, but should I fix the original as well?  I kind of don't want to crack it open right now though.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/anonymous-strictudt.hlsl
tex3d: Good catch, I don't think there is one.  I thought I hit this case, but perhaps it was some manual testing I was doing when making the change.  I'll see if I can add something here.

tools/clang/test/HLSLFileCheck/shader_targets/library/inout_struct_mismatch-strictudt.hlsl
tex3d: Yes, I ran it with the flag enabled by default, then added copies of failing ones using the flag with added explicit casts.

tools/clang/test/HLSLFileCheck/hlsl/types/struct/anonymous-strictudt.hlsl
tex3d: Actually, there are cases in the `conversions-between-type-shapes-strictudt.hlsl` VerifierTest:

```
    // implicit casting to anonymous struct of identical layout ok
    struct { int a, b; } anon2_1 = s2;
```
and
```
    // implicit casting to anonymous struct of subset layout is not legal
    struct { int a, b; } anon2_2;
    anon2_2 = derivedfrom_s2;                               /* expected-error {{cannot implicitly convert from 'DerivedFrom_S2' to 'struct (anonymous struct at }} */
    struct { int a, b; } anon2_3;
    anon2_3 = s4;                                           /* expected-error {{cannot implicitly convert from 'S4' to 'struct (anonymous struct at }} */
```

PR #3892, opened 2021-07-30
Add -enable-short-circuit option for operators &&, ||, :?

tex3d:
And add and(), or(), and select(), to replace vector logical operators that don't short circuit.


tools/clang/test/HLSLFileCheck/hlsl/template/BitwiseOps2.hlsl
```diff
@@ -0,0 +1,125 @@
...
+  lshift(b,i);
+  lshift(i,b);
+  rshift(b,i);
+  rshift(i,b);
+
+  return  r1 + r2 + r3.x + r4[3].y + r5 + r6;
+
+#endif
+
+}
```
pow2clk, line 125: The only difference between this and BitwiseOps.hlsl I can see is the capitalization of the user functions to prevent clashes with the new intrinsics. I see the need for a test that works with all the hlsl2021 features, but I don't know why the original version needs to be retained. Couldn't we just replace BitwiseOps.hlsl with this?

pow2clk (approved):
I have very little to say! It looks good!

tools/clang/test/HLSLFileCheck/hlsl/template/BitwiseOps2.hlsl
tex3d: Maybe, but this is the only place where it verifies the other names still work when you don't use the option.  We could add another test for that I suppose.

adam-yang (approved):
LGTM!

PR #3895, opened 2021-08-03
Add attr::HLSLShader to list of HLSL attributes in IsHLSLAttr()

LukasBanana:
`IsHLSLAttr(clang::attr::Kind AttrKind)` must return `true` for `clang::attr::HLSLShader` to support SM6.3 `[shader("...")]` attributes.


pow2clk (approved):
Looks good!

PR #3900, opened 2021-08-07
Fix issues in DxilMutateResourceToHandle.

tex3d:
- Don't clobber HLSL type on resource if already handle
- Don't walk GV types that are already handle
- Add all non-handle CreateHandleForLib functions and map arg to handle


python3kgae (approved):

dmpots (commented):
LGTM!

dmpots (approved):

PR #3901, opened 2021-08-09
Execute optimizations on functions to be inlined to avoid code explosion

mkroll-nv:
The DirectXShaderCompiler always inlines all functions. But it does so before applying any optimizations. Thus unoptimized code is inlined into functions which are also inlined into other functions quickly leading to code explosion for complex call graphs, which results in very slow compile time.
With these changes, a few optimizations are applied before inlining to reduce the compile time.

- Allow to execute DxilConditionalMem2Reg pass without touching functions
  requiring special care due to DXIL signatures, inputs, outputs, etc.,
  i.e. skipping entry functions and other functions with Dxil function
  properties ("exported" functions).
- Fixes TryRemoveUnusedAlloca() skipping casts while comparing the source
  of a MemCpyInst with a BitCast.
- Disable conversion of MemTransfer or MemSet instructions to
  i1/2/4/7 load/stores in InstCombine causing SROA_Helper::RewriteBitCast
  to assert.
- Disable elimination of casts of casts in InstCombine to avoid
  SROA_Helper::LowerMemcpy to fail due to non-matching types.
- Adapted tests:
  - lib_arg_flatten: add a nonnull in the expected call arguments

### DIFF
--- a/include/llvm/Transforms/Scalar.h
+++ b/include/llvm/Transforms/Scalar.h
@@ -126,7 +126,7 @@ void initializeSROA_Parameter_HLSLPass(PassRegistry&);
 Pass *createDxilFixConstArrayInitializerPass();
 void initializeDxilFixConstArrayInitializerPass(PassRegistry&);
 
-Pass *createDxilConditionalMem2RegPass(bool NoOpt);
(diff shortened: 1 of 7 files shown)


damyanp:

Issue #3824, opened 2021-06-14
Assertion when calling atan2() with vector


ben-clayton:
[According to the HLSL documentation](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-isnan) `atan2()` should accept parameters of type scalar or a vector:

Name | Template Type | Component Type | Size
-- | -- | -- | --
y | same as input&nbsp;x | float | same dimension(s) as input&nbsp;x
x | scalar,&nbsp;vector, or&nbsp;matrix | float | any
ret | same as input&nbsp;x | float | same dimension(s) as input&nbsp;x

However attempting to call `atan2()` with a vector produces the following ICE:

```hlsl
void main() {
  float3 _ = atan2(float3(0.0f, 0.0f, 0.0f), float3(0.0f, 0.0f, 0.0f));
}
```

```bash
> ./dxc test.hlsl -Tps_6_0
Internal Compiler error: cast<X>() argument of incompatible type!
```

This might be a regression, as the version of DXC used by shader compiler [appears to handle this](http://shader-playground.timjones.io/fe1cc117f9207abdfeba4617e320a297) (`Updated from trunk on 2020-12-16`).

jaebaek:
Fixed by #3908

PR #3908, opened 2021-08-16
Support const vector argument for isnan(), atan2()

### LINKED ISSUE #3823: Assertion when calling isnan() with vector
[According to the HLSL documentation](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-isnan) `isnan()` should accept a scalar or a vector:

Name | Template Type | Component Type | Size
-- | -- | -- | --
x | scalar,&nbsp;vector, or&nbsp;matrix | float | any
ret | scalar,&nbsp;vector, or&nbsp;matrix | bool | as input

However attempting to call `isnan()` with a vector produces the following ICE:

```hlsl
void main() {
  bool3 _ = isnan(float3(0.0f, 0.0f, 0.0f));
}
```

```bash
> ./dxc test.hlsl -Tps_6_0
Internal Compiler error: cast<X>() argument of incompatible type!
```

This might be a regression, as the version of DXC used by shader compiler [appears to handle this](http://shader-playground.timjones.io/0800e300a92f9cdf26404549ef54fb0d) (`Updated from trunk on 2020-12-16`).

jaebaek:
Fixes #3823, #3824

### DIFF
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -188,6 +188,8 @@ public:
   TEST_METHOD(CodeGenRootSigProfile2)
   TEST_METHOD(CodeGenRootSigProfile5)
   TEST_METHOD(CodeGenWaveSize)
+  TEST_METHOD(CodeGenVectorIsnan)
+  TEST_METHOD(CodeGenVectorAtan2)
   TEST_METHOD(PreprocessWhenValidThenOK)
   TEST_METHOD(LibGVStore)
   TEST_METHOD(PreprocessWhenExpandTokenPastingOperandThenAccept)
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -2877,6 +2877,14 @@ TEST_F(CompilerTest, CodeGenWaveSize) {
   CodeGenTestCheck(L"attributes_wavesize.hlsl");
 }
 
+TEST_F(CompilerTest, CodeGenVectorIsnan) {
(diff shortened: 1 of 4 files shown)


jaebaek:
It is not related to SPIR-V. It is related to the constant folding for the intrinsic function calls with constant arguments. PTAL!

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2877,6 +2877,14 @@ TEST_F(CompilerTest, CodeGenWaveSize) {
...
 }
 
+TEST_F(CompilerTest, CodeGenVectorIsnan) {
+  CodeGenTestCheck(L"isnan_vector_argument.hlsl");
+}
+
+TEST_F(CompilerTest, CodeGenVectorAtan2) {
+  CodeGenTestCheck(L"atan2_vector_argument.hlsl");
+}
+
```
pow2clk, line 2887: The TAEF version of this file requires TEST_METHOD invocations in the class above around line 158. That's why the windows builds are failing

pow2clk (commented):
This is great! Thanks for doing it!

With the small change to the CompilerTest.cpp file, this should be good to go.

I suppose the advantage of adding these files to CompilerTest instead of just dumping them in the HLSLFileCheck directory is that they will be run on Linux. That's fine for now. I have a 80% completed task to re-enable the filecheck test on Linux, but I can't justify blocking this behind that.

tools/clang/unittests/HLSL/CompilerTest.cpp
jaebaek: done

pow2clk (approved):
Looks good!

Thanks again for doing this!

jaebaek:
My pleasure!

PR #3912, opened 2021-08-23
Add --version flag to print dxc version info

gracejennings:
- Added a --version flag to display the same dxc version info displayed from the -? flag
- Added --version flag description to the help text displayed from the -? flag


pow2clk (approved):
Looks good!

PR #3914, opened 2021-08-25
Allow host to be passed to CMake for cross compile

pdodzweit:
When cross compiling for Android on Windows, GetHostTriple.cmake will end up running Linux shell commands.  This small change lets the user pass in LLVM_INFERRED_HOST_TRIPLE via the cmake command line and bypass the need to invoke the GetHostTriple logic.

With this change I was able to build for Android on Windows using the ninja generator and passing -DLLVM_INFERRED_HOST_TRIPLE=x86_64-pc-win32 to cmake.

### DIFF
--- a/cmake/config-ix.cmake
+++ b/cmake/config-ix.cmake
@@ -318,8 +318,10 @@ endif()
 
 # By default, we target the host, but this can be overridden at CMake
 # invocation time.
-include(GetHostTriple)
-get_host_triple(LLVM_INFERRED_HOST_TRIPLE)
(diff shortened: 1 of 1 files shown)


llvm-beanz:
Upstream LLVM resolved this issue with the following commit:

https://github.com/llvm/llvm-project/commit/2874d285b9f35af8ca6c0268d3aadea4fe671c68

Might be worth just cherry-picking that back instead of diverging.

raghavthind2005 (approved):

PR #3915, opened 2021-08-26
[SPIRV] Add linkageAttributes decorates

jiaolu:
Add linkageAttributes OpDecorate as required by the spirv spec for the
exported/imported values of Linkage capability

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -641,6 +641,10 @@ public:
   void decorateCoherent(SpirvInstruction *target, SourceLocation);
 
   /// \brief Decorates the given target with LinkageAttributes
+  /// We have to set targetInst as nullptr when it is an imported or exported
+  /// function.
+  /// We have to set targetFunc as nullptr when it is an imported or
+  /// exported global variable.
   void decorateLinkage(SpirvInstruction *targetInst, SpirvFunction *targetFunc,
                        llvm::StringRef name, spv::LinkageType linkageType,
                        SourceLocation);
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -640,6 +640,11 @@ public:
   /// \brief Decorates the given target with Coherent
   void decorateCoherent(SpirvInstruction *target, SourceLocation);
 
+  /// \brief Decorates the given target with LinkageAttributes
+  void decorateLinkage(SpirvInstruction *targetInst, SpirvFunction *targetFunc,
+                       llvm::StringRef name, spv::LinkageType linkageType,
+                       SourceLocation);
+
   /// --- Constants ---
   /// Each of these methods can acquire a unique constant from the SpirvContext,
   /// and add the context to the list of constants in the module.
--- a/tools/clang/include/clang/SPIRV/SpirvInstruction.h
+++ b/tools/clang/include/clang/SPIRV/SpirvInstruction.h
@@ -470,10 +470,8 @@ public:
                   spv::Decoration decor,
                   llvm::ArrayRef<SpirvInstruction *> params);
 
-  SpirvDecoration(SourceLocation loc, SpirvInstruction *target,
-                  SpirvFunction *targetFunc, spv::Decoration decor,
-                  llvm::ArrayRef<uint32_t> params = {},
-                  llvm::Optional<uint32_t> index = llvm::None);
+  SpirvDecoration(SourceLocation loc, SpirvFunction *targetFunc,
+                  spv::Decoration decor, llvm::ArrayRef<uint32_t> params);
 
   DEFINE_RELEASE_MEMORY_FOR_CLASS(SpirvDecoration)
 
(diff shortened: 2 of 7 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1402,6 +1402,10 @@ SpirvFunction *DeclResultIdMapper::getOrRegisterFn(const FunctionDecl *fn) {
       spvBuilder.createSpirvFunction(fn->getReturnType(), fn->getLocation(),
                                      fn->getName(), isPrecise, isNoInline);
 
+  if (fn->getAttr<HLSLExportAttr>())
+    spvBuilder.decorateLinkage(nullptr, spirvFunction, fn->getName(),
+                               spv::LinkageType::Export, fn->getLocation());
```
jaebaek, line 1407: could you please wrap the body of if-statement with `{}`?

e.g.,
```
  if (fn->getAttr<HLSLExportAttr>()) {
    spvBuilder.decorateLinkage(nullptr, spirvFunction, fn->getName(),
                               spv::LinkageType::Export, fn->getLocation());
  }
```

tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -640,6 +640,11 @@ class SpirvBuilder {
   /// \brief Decorates the given target with Coherent
   void decorateCoherent(SpirvInstruction *target, SourceLocation);
 
+  /// \brief Decorates the given target with LinkageAttributes
```
jaebaek, line 643: Could you please explain
- We have to set `targetInst` as `nullptr` when it is an imported or exported function
- We have to set `targetFunc` as `nullptr` when it is an imported or  exported global variable

in this comment?
Without this description, someone can misuse it.

tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -1379,6 +1380,22 @@ void SpirvBuilder::decorateCoherent(SpirvInstruction *target,
   mod->addDecoration(decor);
 }
 
+void SpirvBuilder::decorateLinkage(SpirvInstruction *targetInst,
+                                   SpirvFunction *targetFunc,
+                                   llvm::StringRef name,
+                                   spv::LinkageType linkageType,
+                                   SourceLocation srcLoc) {
```
jaebaek, line 1387: I'd add `assert((targetInst == nullptr) != (targetFunc == nullptr));` and explain we have to set a decoration for the linkage of a global variable or a function, but we cannot set them at the same time.

Note that `(targetInst == nullptr) != (targetFunc == nullptr)` is the same as `(targetInst != nullptr || targetFunc != nullptr) && (targetInst == nullptr || targetFunc == nullptr)`

tools/clang/lib/SPIRV/EmitVisitor.cpp
```diff
@@ -591,7 +591,14 @@ bool EmitVisitor::visit(SpirvModuleProcessed *inst) {
 
 bool EmitVisitor::visit(SpirvDecoration *inst) {
   initInstruction(inst);
-  curInst.push_back(getOrAssignResultId<SpirvInstruction>(inst->getTarget()));
+  if (inst->getTarget())
```
jaebaek, line 594: Wrap this if-statement body with `{}`

```
if (inst->getTarget()) {
  curInst.push_back(getOrAssignResultId<SpirvInstruction>(inst->getTarget()));
} ￼else {
 ...
}
```

tools/clang/include/clang/SPIRV/SpirvInstruction.h
```diff
@@ -470,6 +470,11 @@ class SpirvDecoration : public SpirvInstruction {
                   spv::Decoration decor,
                   llvm::ArrayRef<SpirvInstruction *> params);
 
+  SpirvDecoration(SourceLocation loc, SpirvInstruction *target,
+                  SpirvFunction *targetFunc, spv::Decoration decor,
+                  llvm::ArrayRef<uint32_t> params = {},
+                  llvm::Optional<uint32_t> index = llvm::None);
```
jaebaek, line 476: Since we have
```
  SpirvDecoration(SourceLocation loc, SpirvInstruction *target,
                  spv::Decoration decor,
                  llvm::ArrayRef<SpirvInstruction *> params);
```

I think the new one can be simply
```
  SpirvDecoration(SourceLocation loc, SpirvFunction *targetFunc, spv::Decoration decor,
                  llvm::ArrayRef<uint32_t> params = {},
                  llvm::Optional<uint32_t> index = llvm::None);
```

In addition, just out of curiosity, do we have a specific case that we need `llvm::Optional<uint32_t> index = llvm::None`?

I am not sure SPIR-V `OpMemberDecorate` has a function operand.
Could you please remove `llvm::Optional<uint32_t> index = llvm::None` parameter if you do not have a specific example?
We can add it later when we actually meet the case.

jaebaek (commented):
@jiaolu Thanks a lot for your contribution! It helps a lot to improve DXC/SPIR-V.

I have a few comments but not critical. After you handle them, I will submit this PR.

jaebaek (approved):

PR #3918, opened 2021-08-28
Convert cmake-predefined-config-params to cache

llvm-beanz:
CMake supports passing in CMake scripts via the `-C` command line
option, which can set CMake cache variables to initialize options
before the first CMakeLists file is processed. This is a portable and
shell-agnostic way of supporting what the
`cmake-predefined-config-params` file is used for.


pow2clk (commented):
My only reservation is that we should eliminate the old cmake-predefined-config-params file if we can. @jaebaek or @sudonatalie will have to weigh it for us to be sure on that

llvm-beanz:
@pow2clk  100% understand your point. As I mentioned in IM, I only left the file there because I was unsure if people were relying on it for testing or workflows that aren't visible on GitHub.

I've pushed for many years in the LLVM community (with some success) to move common or complex build configurations into CMake cache scripts.

CMake cache scripts are as portable as CMake itself, which is a big win, but they also provide a mechanism for simplifying a configuration command from potentially dozens (or more) CMake variables to a single command line flag and value. We use them extensively now in LLVM for configuring toolchain builds and other complex builds.

I've also advocated (with less success) that we have a CMake cache script that corresponds to each CI build configuration. Thus simplifying the CI tasks, but also making it easy for developers to replicate CI build configurations locally if a bug only appears in certain configurations.

I am totally happy to delete `cmake-predefined-config-params` if there are no objections.

jaebaek:
@llvm-beanz Thank you for sending this PR. This PR looks good to me. Please delete `cmake-predefined-config-params` file. It will not be used anywhere if we merge this PR.

PR #3935, opened 2021-09-07
Fix issue when updateCounter for dynamic resource

python3kgae:
Fix issue when updateCounter for dynamic resource, also remove createHandleForLib on handle type when not a library.


tex3d (approved):
LGTM!

PR #3941, opened 2021-09-09
Allow calling DxcInitThreadMalloc again after calling DxcCleanupThreadMalloc

dmpots:
Without this change we cannot call a sequence of

    DxcInitThreadMalloc()
    DxcCleanupThreadMalloc()
    ...
    DxcInitThreadMalloc() // <- This will trigger an assert

because we hit an assert that assumes the default malloc pointer is nullptr.


pow2clk (approved):
Looks good!

I think I ran into this a couple times when working on the Linux port :)

PR #3942, opened 2021-09-10
NoUnwind fix. HLExtIntrinsic collision fix.

adam-yang:
- Fixed HL operations and DXIL operations not having the NoUnwind attribute, and therefore not DCE'ed
- Fixed LegalizeEvalOp not checking for the right intrinsic group and incorrectly removing Allocas for HLExtIntrinsic intrinsics

### DIFF
--- a/lib/HLSL/HLOperations.cpp
+++ b/lib/HLSL/HLOperations.cpp
@@ -421,6 +421,8 @@ HLBinaryOpcode GetUnsignedOpcode(HLBinaryOpcode opcode) {
 
 static void SetHLFunctionAttribute(Function *F, HLOpcodeGroup group,
                                        unsigned opcode) {
+  F->addFnAttr(Attribute::NoUnwind);
+
   switch (group) {
   case HLOpcodeGroup::HLUnOp:
   case HLOpcodeGroup::HLBinOp:
@@ -428,14 +430,12 @@ static void SetHLFunctionAttribute(Function *F, HLOpcodeGroup group,
   case HLOpcodeGroup::HLSubscript:
     if (!F->hasFnAttribute(Attribute::ReadNone)) {
       F->addFnAttr(Attribute::ReadNone);
-      F->addFnAttr(Attribute::NoUnwind);
     }
     break;
   case HLOpcodeGroup::HLInit:
     if (!F->hasFnAttribute(Attribute::ReadNone))
       if (!F->getReturnType()->isVoidTy()) {
         F->addFnAttr(Attribute::ReadNone);
-        F->addFnAttr(Attribute::NoUnwind);
       }
     break;
   case HLOpcodeGroup::HLMatLoadStore: {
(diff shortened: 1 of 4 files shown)


tex3d (approved):
LGTM!

tools/clang/unittests/HLSL/ExtensionTest.cpp
```diff
@@ -1076,6 +1082,114 @@ TEST_F(ExtensionTest, SamplerExtensionIntrinsic) {
   CheckMsgs(disassembly.c_str(), disassembly.length(), expected, 1, true);
 }
 
+// Takes a string to match, and a regex pattern string, returns the first match at index [0],
+// as well as sub expressions starting at index [1]
+static std::vector<std::string> Match(const std::string &str, const std::string pattern) {
```
pow2clk, line 1087: It's minor, but I don't see the need to convert the output to a std::vector rather than keep it as a SmallVector

lib/HLSL/HLOperations.cpp
```diff
@@ -421,21 +421,21 @@ HLBinaryOpcode GetUnsignedOpcode(HLBinaryOpcode opcode) {
 
 static void SetHLFunctionAttribute(Function *F, HLOpcodeGroup group,
                                        unsigned opcode) {
+  F->addFnAttr(Attribute::NoUnwind);
```
pow2clk, line 424: I'd like to understand better why this can be unconditionally included where it wasn't before

tools/clang/unittests/HLSL/ExtensionTest.cpp
```diff
@@ -1076,6 +1082,114 @@ TEST_F(ExtensionTest, SamplerExtensionIntrinsic) {
...
+            return ret;
+        }
+      )",
+      {L"/Vd", L"/Od"}, {}
+    );
+
+    std::string disassembly = c.Disassemble();
+
+    auto match1 = Match(disassembly, std::string("%([0-9.a-zA-Z]*) = call float @collide_proc\\(i32 ") + std::to_string(Intrinsic.hlsl.Op));
+    VERIFY_ARE_EQUAL(match1.size(), 2);
```
pow2clk, line 1134: Just changinge `2` to `2U` should fix the Linux failure

tools/clang/unittests/HLSL/ExtensionTest.cpp
```diff
@@ -1076,6 +1082,114 @@ TEST_F(ExtensionTest, SamplerExtensionIntrinsic) {
...
+
+  std::string disassembly = c.Disassemble();
+
+  /*
+  * We're looking for this:
+  *   declare float @test_proc(i32, float) #1
+  *   attributes #1 = { nounwind }
+  */
+  auto m1 = Match(disassembly, std::string("declare float @test_proc\\(i32, float\\) #([0-9]*)"));
+  VERIFY_ARE_EQUAL(m1.size(), 2);
```
pow2clk, line 1164: Here too

pow2clk (commented):
Two questions and suggestions to fix the *Nix builds

lib/HLSL/HLOperations.cpp
adam-yang: I think we just added them as needed. Some of the HLOpcodeGroup's get replaced up pretty early with Matrix lowering, so inability to do DCE wouldn't really have impacted them very much.

tools/clang/unittests/HLSL/ExtensionTest.cpp
adam-yang: I just didn't want the caller to always have to think about how big of a SmallVector to use. I could have used a SmallVectorImpl arg but that's kind of clumsy instead of having a return.

PR #3947, opened 2021-09-13
Added unused output target warning and moved usage mask fill location

gracejennings:
Added a warning for output targets that are not fully written in a shader by checking the signature element usage mask. 

Also moved the location where the signature element usage masks are filled to more accurately reflect which channels have been assigned for each element. The signature element usage masks were previously completely filled for a target if at least one of the channels were assigned.


tools/clang/test/HLSLFileCheck/hlsl/diagnostics/warnings/unused-output-target-warning.hlsl
```diff
@@ -0,0 +1,18 @@
...
+// CHECK-NOT: warning: Declared output SV_Target0 not fully written in shader.
+// CHECK: warning: Declared output SV_Target1 not fully written in shader.
+// CHECK: warning: Declared output SV_Target2 not fully written in shader.
+// CHECK: warning: Declared output SV_Target3 not fully written in shader.
+
+void main(out float4 outRT0
+    : SV_Target0, out float4 outRT1
+    : SV_Target1, out float4 outRT2
+    : SV_Target2, out float4 outRT3
+    : SV_Target3) {
```
pow2clk, line 12: I suspect this fell victim to some autoformatting that makes it confusing to read. I think it's worth changing it so that the SV_Target*s are on the same line as the parameters they apply to.

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -742,6 +737,29 @@ class DxilFinalizeModule : public ModulePass {
...
+        if (DM.GetShaderModel()->IsHS())
+          MarkUsedSignatureElements(DM.GetPatchConstantFunction(), DM);
+      }
+
+      // Adding warning for pixel shader with unassigned target
+      if (DM.GetShaderModel()->IsPS()) {
+        DxilSignature &sig = DM.GetOutputSignature();
+        for (auto &Elt : sig.GetElements()) {
+          if (Elt->GetKind() == Semantic::Kind::Target) {
+            if (Elt->GetUsageMask() != Elt->GetColsAsMask()) {
```
pow2clk, line 752: you might consider combining these conditionals with && to reduce the indenting, which for me hinders readability

pow2clk (commented):
No functional suggestions. just some things that might improve readability. 

As to the test failure, it's falling prey to the devops issue with macos-10 and python3. It seems to be pulling the yaml file from your local branch instead of the main branch. If you rebase your local branch and push the change here, I think the test will pass

pow2clk (approved):
Looks good!

PR #3948, opened 2021-09-14
Add literal eval property to intrinsics; improve literal arg conversion

### LINKED ISSUE #3380: WaveActive* and WavePrefix* ops emitted as 64-bit for constant int
WaveActive* and WavePrefix* ops are being emitted as 64-bit DXIL instructions on constant int.

```
[numthreads(3,1,1)]
void main() {
    WaveActiveMax(10);
    WavePrefixSum(5);
}
````

```
define void @main() {
  %1 = call i64 @dx.op.waveActiveOp.i64(i32 119, i64 10, i8 3, i8 0)  ; WaveActiveOp(value,op,sop)
  %2 = call i64 @dx.op.wavePrefixOp.i64(i32 121, i64 5, i8 0, i8 0)  ; WavePrefixOp(value,op,sop)
  ret void
}
```

We should be defaulting to 32-bits for literal values.

### PR DESCRIPTION
**Pull request by tex3d**
- Change independent BOOL flag arguments in HLSL_INTRINSIC to Flags
- Add flag for literal eval
- Avoid preserving literal types when literal eval is not supported
  - This will help prevent 64-bit ops from being generated by accident
- Move literal conversions to after first pass at argument resolution
- Prefer int over uint for positive literal that fits

Note: introduces breaking change to HLSL_INTRINSIC structure, so external intrinsics will have to be adjusted.

Fixes #3380

### CODE REVIEW

File: tools/clang/lib/Sema/SemaHLSL.cpp [hunk 1]
```diff
@@ -5889,6 +5896,33 @@ bool HLSLExternalSource::MatchArguments(
...
+
+    // Merge ComponentTypes
+    if (AR_BASIC_UNKNOWN == ComponentType[pIntrinsicArg->uComponentTypeId]) {
+      ComponentType[pIntrinsicArg->uComponentTypeId] = TypeInfoEltKind;
+    }
+    else {
+      if (!CombineBasicTypes(
+            ComponentType[pIntrinsicArg->uComponentTypeId],
+            TypeInfoEltKind,
+            &ComponentType[pIntrinsicArg->uComponentTypeId])) {
```
tex3d, line 5920: Ideally, I think in the type-combine case, we would want to preserve the original literal when combining types here, only converting the literal to a concrete type when no type was ever assigned to this slot after the first pass (the AR_BASIC_UNKNOWN case).  I'll update the PR with this change.

llvm-beanz:
@tex3d, we discussed this PR and the associated issues in the morning triage meeting. Since this issue will be resolved with the HLSL 202x [Conforming Literals feature](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0017-conforming-literals.md), we do not want to continue pursuing this change. As a result I'm closing this PR per the triage discussion.

PR #3957, opened 2021-09-18
Fix check in dxc extra output processing for non-blob output

dmpots:
When processing the DXC_OUT_EXTRA_OUTPUTS output from a compile there
is an attempt to skip non-blob outputs. The check is written incorrectly
because the call to `GetOutput` will fail (when it tries to QueryInterface
on the blob output).

The fix is to check the HRESULT returned from GetOutput and skip the output
if it failed.

Found when using the dxclib from an external project. I did not see an easy
way to add a test for this.


adam-yang (approved):
LGTM. Thanks!

PR #3959, opened 2021-09-19
Expose the DxcCompilerAdapter through the com interface

dmpots:
This commit exposes the DxcCompilerAdapter class through the standard
Dxc com interface. The DxcCompilerAdapter is used to implement
the IDxcCompiler and IDxcCompiler2 interfaces by backing it with
an instance of the IDxcCompiler3 interface.

The class is used internally by the DxcCompiler class to support the
older compiler interfaces. We have an external project that could
take advantage of this support as well so we now expose it through
the com interface.

### DIFF
--- a/tools/clang/tools/dxcompiler/dxcompileradapter.h
+++ b/tools/clang/tools/dxcompiler/dxcompileradapter.h
@@ -51,6 +51,11 @@ public:
   ULONG STDMETHODCALLTYPE Release() override;
   HRESULT STDMETHODCALLTYPE QueryInterface(REFIID iid, void **ppvObject) override;
 
+  void SetDxcCompiler(IDxcCompiler3* pCompilerImpl)
+  {
+      m_pCompilerImpl = pCompilerImpl;
+  }
+
   // ================ IDxcCompiler ================
 
   // Compile a single entry point to the target shader model
--- a/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
+++ b/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
@@ -1777,7 +1777,7 @@ public:
     return hr;
   }
 
-  HRESULT SetDxcCompiler(IUnknown *pCompiler)
+  HRESULT SetDxcCompiler(IUnknown *pCompiler) override
   {
       HRESULT hr = pCompiler->QueryInterface(IID_PPV_ARGS(&m_pDxcCompiler3));
       if (SUCCEEDED(hr))
--- a/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
+++ b/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
@@ -1749,6 +1749,95 @@ HRESULT DxcCompilerAdapter::WrapCompile(
 }
 //////////////////////////////////////////////////////////////
 
+namespace
+{
+class DxcCompilerAdapterInterface : public IDxcCompilerAdapter, public IDxcCompiler3
+{
+private:
+  DXC_MICROCOM_TM_REF_FIELDS()
+  CComPtr<IDxcCompiler3> m_pDxcCompiler3;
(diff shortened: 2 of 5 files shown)


tools/clang/tools/dxcompiler/dxcompileradapter.h
```diff
@@ -51,6 +51,11 @@ class DxcCompilerAdapter: public IDxcCompiler2
   ULONG STDMETHODCALLTYPE Release() override;
   HRESULT STDMETHODCALLTYPE QueryInterface(REFIID iid, void **ppvObject) override;
 
+  void SetDxcCompiler(IDxcCompiler3* pCompilerImpl)
+  {
+      m_pCompilerImpl = pCompilerImpl;
```
pow2clk, line 56: should we do something about the references to any previously set m_pCompilerImpl? Depending on when this is done, it could result in extra Releases on the DxcCompilerAdapter object which will be passed to this one possibly erroneously. Maybe verify that there are no references instead?

It doesn't match the form of the other SetDxcCompiler functions because it returns nothing, but I don't see this being called anywhere so I can't tell if that's for good reason

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1749,6 +1749,91 @@ HRESULT DxcCompilerAdapter::WrapCompile(
...
+      , m_DxcCompilerAdapter(this, pMalloc) 
+  {
+  }
+  
+  DXC_MICROCOM_TM_ALLOC(DxcCompilerAdapterInterface)
+  
+  HRESULT STDMETHODCALLTYPE QueryInterface(REFIID iid, void **ppvObject) override {
+    HRESULT hr = DoBasicQueryInterface<IDxcCompilerAdapter, IDxcCompiler3>(this, iid, ppvObject);
+    if (FAILED(hr)) {
+        return DoBasicQueryInterface<IDxcCompiler, IDxcCompiler2>(&m_DxcCompilerAdapter, iid, ppvObject);
```
pow2clk, line 1775: I suppose this may be where the mismatching SetDxcCompiler call might cause a problem?

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -830,6 +831,43 @@ TEST_F(CompilerTest, CompileWhenWorksThenDisassembleWorks) {
...
+
+  LPCWSTR args[] = { L"-Wno-parentheses-equality" };
+
+  CComPtr<IDxcOperationResult> pResult;
+  VERIFY_SUCCEEDED(pCompiler->Compile(pSource, L"source.hlsl", L"main",
+                                      L"ps_6_0", args, _countof(args), nullptr,
+                                      0, nullptr, &pResult));
+  HRESULT hr = E_FAIL;
+  VERIFY_SUCCEEDED(pResult->GetStatus(&hr));
+  VERIFY_SUCCEEDED(hr);
```
pow2clk, line 868: Maybe another run using the legacy interface? that is the same, but with ` CComPtr<IDxcCompiler2> pCompiler2;`?

pow2clk (commented):
I have a concern about the legacy implementation

tools/clang/tools/dxcompiler/dxcompileradapter.h
dmpots: Yeah, I think the design here is too specific to the intended use case (an implementation of IDxcCompiler3 that contains an instance of IDxcCompilerAdapter and passes itself as the IDxcCompiler3).

I think its not a good general purpose thing to expose in the dxcapi so I will move it to a private implementation where the use case is more restricted and well defined.

dmpots:
I think its not a good general purpose thing to expose in the dxcapi so I will move it to a private implementation where the use case is more restricted and well defined.

PR #3960, opened 2021-09-20
PIX: Disambiguate mesh shader activity based on full XYZ dispatch paramete…

jeffnn:
PIX's mesh shader output instrumentation needs to disambiguate groups of mesh shader invocations, so that PIX can reconstruct the intended mesh after the fact. In the case where an amplification shader is present, the previous code was incomplete: it included only the X value of the mesh shader's group ID relative to the AS DispatchMesh call. The full X*Y*Z value is needed. Unfortunately, this value does not necessarily fit into a single DWORD for returning to PIX via the instrumentation, but various spec constraints mean that the product of the group ID cube's dimensions will fit in a DWORD.
So this change expands the custom data added to the AS->MS payload to include the Y and Z group counts that the AS passed to DispatchMesh. The mesh shader can then multiply its group ID's component values by these counts in order to come up with a single unique group ID value.

### DIFF
--- a/lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
+++ b/lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
@@ -35,17 +35,22 @@ public:
   const char *getPassName() const override { return "DXIL Add flat thread id to payload from AS to MS"; }
   bool runOnModule(Module &M) override;
   void applyOptions(PassOptions O) override;
-
-private:
-
-    void EmitInstructionsToCopyStructContents(llvm::IRBuilder<> &B,
-        llvm::Value *NewStructAlloca,
-        llvm::Value*OldStructAlloca);
 };
 
 void DxilPIXAddTidToAmplificationShaderPayload::applyOptions(PassOptions O) {
   GetPassOptionUInt32(O, "dispatchArgY", &m_DispatchArgumentY, 1);
   GetPassOptionUInt32(O, "dispatchArgZ", &m_DispatchArgumentZ, 1);
+}
+
+void AddValueToExpandedPayload(OP* HlslOP, llvm::IRBuilder<> &B, ExpandedStruct & expanded, AllocaInst* NewStructAlloca, unsigned int expandedValueIndex, Value* value) {
+  Constant *Zero32Arg = HlslOP->GetU32Const(0);
+  SmallVector<Value *, 2> IndexToAppendedValue;
+  IndexToAppendedValue.push_back(Zero32Arg);
+  IndexToAppendedValue.push_back(HlslOP->GetU32Const(expandedValueIndex));
+  auto *PointerToEmbeddedNewValue =
+      B.CreateInBoundsGEP(expanded.ExpandedPayloadStructType, NewStructAlloca,
+                          IndexToAppendedValue, "PointerToEmbeddedNewValue" + std::to_string(expandedValueIndex));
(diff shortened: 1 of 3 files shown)


lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
```diff
@@ -35,19 +35,24 @@ class DxilPIXAddTidToAmplificationShaderPayload : public ModulePass {
...
 void DxilPIXAddTidToAmplificationShaderPayload::applyOptions(PassOptions O) {
   GetPassOptionUInt32(O, "dispatchArgY", &m_DispatchArgumentY, 1);
   GetPassOptionUInt32(O, "dispatchArgZ", &m_DispatchArgumentZ, 1);
 }
 
+void AddValueToExpandedPayload(OP* HlslOP, llvm::IRBuilder<> &B, ExpandedStruct & expanded, AllocaInst* NewStructAlloca, unsigned int expandedValueIndex, Value* value) {
+  Constant *Zero32Arg = HlslOP->GetU32Const(0);
+  SmallVector<Value *, 2> IndexToAppendedValue;
+  IndexToAppendedValue.push_back(Zero32Arg);
+  IndexToAppendedValue.push_back(HlslOP->GetU32Const(expandedValueIndex));
```
pow2clk, line 49: Pre-existing, but since you have the values you want to populate this SmallVector with,  you could have the same effect with ` SmallVector<Value *, 2> IndexToAppendedValue(Zero32Arg, HlslOP->GetU32Const(expandedValueIndex));` if you wanted. Entirely unnecessary though.

lib/DxilPIXPasses/DxilPIXMeshShaderOutputInstrumentation.cpp
```diff
@@ -196,21 +206,34 @@ SmallVector<Value*, 2> DxilPIXMeshShaderOutputInstrumentation::
...
+    auto * GroupIdX =
+        Builder.CreateCall(GroupIdFunc, {Opcode, Zero32Arg}, "GroupIdX");
+    auto * GroupIdY =
+        Builder.CreateCall(GroupIdFunc, {Opcode, One32Arg}, "GroupIdY");
+    auto * GroupIdZ =
+        Builder.CreateCall(GroupIdFunc, {Opcode, Two32Arg}, "GroupIdZ");
+
+    auto *XxY =
+      Builder.CreateMul(GroupIdX, ASDispatchMeshYCount);
+    auto *XandY = Builder.CreateAdd(GroupIdY, XxY);
```
pow2clk, line 230: minor, but perhaps it should be `XaddY`? `XandY` makes me think of `X&Y`

pow2clk (approved):
Looks good to go! I have a couple minor suggestions you can take or leave

lib/DxilPIXPasses/DxilPIXMeshShaderOutputInstrumentation.cpp
jeffnn: I went with XplusY (and fixed another instance of this too). Thanks!

lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
jeffnn: I think I'll leave it due to the testing rigamarole that would ensue. Thanks for the education.

PR #3961, opened 2021-09-21
Fix bugs in signature reflection, add signatures to dumper and tests

### LINKED ISSUE #3952: Reflection API returns wrong "Register" for some mesh shader output signatures
According to the disassembly, this shader's output sig should be:

'''
; Output signature:
;
; Name                 Index   Mask Register SysValue  Format   Used
; -------------------- ----- ------ -------- -------- ------- ------
; BLOAT                    0   x           0     NONE   float   x   
; TEX                      0    yz         0     NONE   float    yz 
; BLOAT                    1   x           1     NONE   float   x   
; BLOAT                    2   x           2     NONE   float   x   
; BLOAT                    3   x           3     NONE   float   x   
; BLOAT                    4   x           4     NONE   float   x   
; BLOAT                    5   x           5     NONE   float   x   
; BLOAT                    6   x           6     NONE   float   x   
; BLOAT                    7   x           7     NONE   float   x   
; SV_Position              0   xyzw        8      POS   float   xyzw
;
;
'''
Note the non-zero values for BLOAT in the "Register" column. However, if you call ID3D12ShaderReflection::GetOutputParameterDesc for these, the API will return 0 for Register. Curiously, it will return correct values for the other entries in the list. It also returns correct values for all other mesh shaders I've tried. It infer that the bug is related to the extreme semantic index values.
[ConfShader.zip](https://github.com/microsoft/DirectXShaderCompiler/files/7171909/ConfShader.zip)

tex3d:
- Fix Barycentric system value translation
- Increment Register for array elements
- Add signature dumping to D3DReflectionDumper
- Dump PatchConstantParameters when mesh shader

Fixes #3952

### DIFF
--- a/tools/clang/unittests/HLSLTestLib/D3DReflectionDumper.cpp
+++ b/tools/clang/unittests/HLSLTestLib/D3DReflectionDumper.cpp
@@ -10,6 +10,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 
 #include "dxc/Support/Global.h"
+#include "dxc/DXIL/DxilConstants.h"
 #include "dxc/test/D3DReflectionDumper.h"
 #include "dxc/DxilContainer/DxilContainer.h"
 #include <sstream>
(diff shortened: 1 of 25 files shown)


tools/clang/unittests/HLSLTestLib/D3DReflectionDumper.cpp
```diff
@@ -347,6 +348,96 @@ LPCSTR ToString(D3D_PARAMETER_FLAGS Flag) {
...
+LPCSTR ToString(D3D_MIN_PRECISION MinPrec) {
+  switch (MinPrec) {
+  case D3D_MIN_PRECISION_DEFAULT: return "D3D_MIN_PRECISION_DEFAULT";
+  case D3D_MIN_PRECISION_FLOAT_16: return "D3D_MIN_PRECISION_FLOAT_16";
+  case D3D_MIN_PRECISION_FLOAT_2_8: return "D3D_MIN_PRECISION_FLOAT_2_8";
+  case D3D_MIN_PRECISION_RESERVED: return "D3D_MIN_PRECISION_RESERVED";
+  case D3D_MIN_PRECISION_SINT_16: return "D3D_MIN_PRECISION_SINT_16";
+  case D3D_MIN_PRECISION_UINT_16: return "D3D_MIN_PRECISION_UINT_16";
+  case D3D_MIN_PRECISION_ANY_16: return "D3D_MIN_PRECISION_ANY_16";
+  case D3D_MIN_PRECISION_ANY_10: return "D3D_MIN_PRECISION_ANY_10";
```
pow2clk, line 410: loath as I am to bring it up, many of these don't have tests

tools/clang/unittests/HLSLTestLib/D3DReflectionDumper.cpp
```diff
@@ -411,7 +502,7 @@ void D3DReflectionDumper::Dump(D3D12_SHADER_BUFFER_DESC &Desc) {
 }
 void D3DReflectionDumper::Dump(D3D12_SHADER_INPUT_BIND_DESC &resDesc) {
   SetLastName(resDesc.Name);
-  WriteLn("D3D12_SHADER_BUFFER_DESC: Name: ", m_LastName);
+  WriteLn("D3D12_SHADER_INPUT_BIND_DESC: Name: ", m_LastName);
```
pow2clk, line 505: This and the other output change account for most of the changed files, but I can't really contrive an excuse not to make this change

pow2clk (approved):
Quite a lot in this change! as ever, more testing is possible, but probably not necessary just now

PR #3963, opened 2021-09-21
Limit Sema intrinsic checks to builtin "op" table

pow2clk:
The check for short curcuit intrinsics failed to account for
intrinsic tables other than the default. This ensures that assumption is
correct before performing the check

As an incidental, added code to catch a few other cases too


dmpots (approved):
LGTM

tex3d (approved):
Looks good, thanks!

PR #3967, opened 2021-09-23
Expand AttrKind from 8bit to 16 bits

jiaolu:
Attributes size exceeds to the limit of the 255, use 16 bit AttrKind

### DIFF
--- a/tools/clang/include/clang/Sema/AttributeList.h
+++ b/tools/clang/include/clang/Sema/AttributeList.h
@@ -122,7 +122,7 @@ private:
   /// True if this has a ParsedType
   unsigned HasParsedType : 1;
 
-  unsigned AttrKind : 8;
+  unsigned AttrKind : 16;
 
   /// \brief The location of the 'unavailable' keyword in an
   /// availability attribute.
--- a/tools/clang/include/clang/Sema/AttributeList.h
+++ b/tools/clang/include/clang/Sema/AttributeList.h
@@ -121,8 +121,10 @@ private:
 
   /// True if this has a ParsedType
   unsigned HasParsedType : 1;
+  
+  unsigned padding : 8;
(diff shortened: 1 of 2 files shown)


jaebaek:
@jiaolu Nice catch!

@pow2clk We met this issue while adding new attributes, which makes existing attribute unit tests fail because it exceeds the attribute size to the limit of 255. Could you please review this change?

tools/clang/include/clang/Sema/AttributeList.h
```diff
@@ -122,7 +122,7 @@ class AttributeList { // TODO: This should really be called ParsedAttribute
   /// True if this has a ParsedType
   unsigned HasParsedType : 1;
 
-  unsigned AttrKind : 8;
+  unsigned AttrKind : 16;
```
pow2clk, line 125: To match with a change this will eventually collide with, It would be helpful if you used this fix instead:

```c++
  unsigned padding : 8;

  unsigned AttrKind;
```

tools/clang/lib/Sema/AttributeList.cpp
```diff
@@ -138,6 +138,7 @@ AttributeList::Kind AttributeList::getKind(const IdentifierInfo *Name,
     std::string lower = FullName.str().lower();
     Result = ::getAttrKind(StringRef(lower), SyntaxUsed);
   }
+  assert (static_cast<unsigned>(Result) < 65536);
```
pow2clk, line 141: Trying to imagine 65536 different attributes 😆

pow2clk (approved):
Looks good! 

I didn't mean to tease about the 65K attrib count. It was a reasonable thing to check just for sanity's sake, but the thought of that many attributes was boggling

Thanks!

PR #3974, opened 2021-09-27
Fixed an assert when a dxil error happens with global variable

adam-yang:
`DILocation` can't actually point to any non-local scopes, so can't be used to represent location of a global variable. Must use `DIGlobalVariable` instead.

Changed `DiagnosticInfoDxil` to hold the actual location info instead of a pointer to a `DILocation`.


pow2clk (approved):
Looks good!

PR #3977, opened 2021-09-28
Always update end scope when structurize returns.

python3kgae:
When structurize return like return 1 in
  if (b == 77)
    return 0;
  else if (b != 3)
    return 1;

the end scope of if (b == 77) should be updated to bReturned.Cmp.false

Final cfg should be the right one, left one is wrong cfg without the fix.
cfg
to


tex3d (approved):
LGTM!

PR #3981, opened 2021-09-29
Fixed structurized return when there are phi's from shortcircuiting.

adam-yang:


python3kgae (approved):
Need a test and make sure incoming block of phi is updated if required.

PR #3982, opened 2021-09-29
Enable -HV 2021

pow2clk:
Enables the associated feature flags when 2021 is selected. Also adds a
feature flag for bitfields.

Add to all feature tests a variant that uses -HV 2021

Made some small changes to a template test that relied on vector
operands with binary logical operators

### DIFF
--- a/lib/DxcSupport/HLSLOptions.cpp
+++ b/lib/DxcSupport/HLSLOptions.cpp
@@ -521,7 +521,7 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
       if (opts.EnableBitfields)
         errors << "/enable-bitfields is not supported with HLSL Version " << opts.HLSLVersion;
 
-        return 1;
+      return 1;
     }
   }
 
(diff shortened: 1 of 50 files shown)


lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -446,7 +446,7 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
   } else {
     try {
       opts.HLSLVersion = std::stoul(std::string(ver));
-      if (opts.HLSLVersion < 2015 || opts.HLSLVersion > 2018) {
+      if (opts.HLSLVersion < 2015 || opts.HLSLVersion > 2021) {
```
tex3d, line 449: Maybe we should update the help text for -HV (look for `hlsl_version`) to indicate 2021 is a valid option.  Also, since there are only specific values that are valid (2015, 2016, 2018, 2021), maybe we should use a switch statement with the valid versions here instead of a range?

pow2clk: I intended to do the first, but forgot and was considering the second

tex3d: Note: help text doesn't mention 2015 (that's because that's an intellisense-only target), and does mention 2017, even though I don't recall if there is any differentiation for that version.

tex3d: Oh, nevermind, 2017 is valid, enums were introduced there.

tex3d (approved):
Looks good with minor comment re. -HV help text and validation.

jaebaek (commented):
Sorry for being late. LGTM!

PR #3983, opened 2021-09-30
Fix missing RDAT function resource usage by recursing constant users

tex3d:
- Before module serialization, constant users of resource globals could themselves have multiple constant users.  These could be consolidated through module serialization/deserialization, leading to a difference when only the first constant user was followed.  The difference could lead to a validation failure only with external validator (DXIL.dll) since reconstructed RDAT would not match the original RDAT with the missing use.  This would happen in cases such as identical constant indexing of resource arrays from multiple entries.
- Also, fix groupshared usage detection when setting function compat flags for RDAT.

### DIFF
--- a/lib/DxilContainer/DxilContainerAssembler.cpp
+++ b/lib/DxilContainer/DxilContainerAssembler.cpp
@@ -1133,7 +1133,7 @@ private:
   }
 
   void
-  FindUsingFunction(const llvm::Value *User,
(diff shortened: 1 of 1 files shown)


pow2clk (approved):
Looks good!

I don't think it's worth iterating on this again, but if you find reason to do so anyway, the function might be more accurately named "FindUsingFunctions" now

PR #3985, opened 2021-10-01
PIX: Changes to run PIX passes on libraries. Also add raygen shader debugging prolog

jeffnn:
These changes are a result of prototype work in PIX to support shader debugging for DXR,, which was an exercise to figure out how to support libraries in general in PIX. These changes are the result. It's all about iterating over all the library functions, rather than "the" entry point function. 
In addition, to have something concrete to submit to a driver, there's the addition of codegen for the shader debugging "prolog", which chooses at runtime whether or not to emit instrumentation data for a given shader invocation based on its (in this case) thread id. This last is not yet ready for production in PIX, in that it is necessary but not sufficient to enable DXR shader debugging.


tools/clang/test/HLSLFileCheck/pix/DebugRayGen.hlsl
```diff
@@ -0,0 +1,75 @@
...
+
+// CHECK:  %RayX = call i32 @dx.op.dispatchRaysIndex.i32(i32 145, i8 0)
+// CHECK:  %RayY = call i32 @dx.op.dispatchRaysIndex.i32(i32 145, i8 1)
+// CHECK:  %RayZ = call i32 @dx.op.dispatchRaysIndex.i32(i32 145, i8 2)
+
+// CHECK: %CompareToThreadIdX
+// CHECK: %CompareToThreadIdY
+// CHECK: %CompareToThreadIdZ
+// CHECK: %CompareXAndY
+// CHECK: %CompareAll
```
pow2clk, line 14: I'm a bit surprised that you can expect these register names to persist. Maybe I'm missing something?

pow2clk (approved):
Looks good!

One small possible omission for hull shaders and a question

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -869,26 +918,58 @@ void DxilDebugInstrumentation::addStepDebugEntryValue(
...
+  bool modified = false;
+  if (shaderKind == DXIL::ShaderKind::Library) {
+    for (llvm::Function& F : M.functions()) {
+      modified = modified | RunOnFunction(M, DM, &F); 
+      return modified;
+    }
+  }
+  else {
+    llvm::Function *entryFunction = PIXPassHelpers::GetEntryFunction(DM);
+    modified = RunOnFunction(M, DM, entryFunction);
```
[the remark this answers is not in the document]
jeffnn: Well spotted, but PIX doesn't support debugging the tessellation stages yet anyway, so that's for another day.

tools/clang/test/HLSLFileCheck/pix/DebugRayGen.hlsl
jeffnn: You are of course correct. This tickled my mind when I coded it and I ignored the tickle. Bad Jeffrey.

PR #3989, opened 2021-10-04
[NFC] Silence a few new clang warnings

llvm-beanz:
Clang has added new warnings for possible unintended string
concatenations. When concatonating strings in an array literal warpping
the concatenation in parentesis quiets the warning.


pow2clk (approved):
Looks good!

PR #3990, opened 2021-10-04
Od fixes sep2021

adam-yang:
Improved code simplification for Od to prevent holding on to 

- Improved DxilValueCache's ability to deduce values
  - Fixed some bugs that caused reachability analysis to fail (which leads to phi value deduction to fail).
  - Added a callback to allow caller to exclude some values from being evaluated where necessary.
- Made DxilEraseDeadRegion do some simple DCE
- Taught DxilEraseDeadRegion about certain metadata on branches that previously pretended it from simplifying it.
- Made RemoveDeadBlock try to remove values iteratively, and also delete store-only allocas and remove any values not contributing to shader output.
- Running DxilEraseDeadRegion even on Od


tools/clang/test/HLSLFileCheck/dxil/debug/unbound_res_arithmetic.hlsl
```diff
@@ -0,0 +1,17 @@
+// RUN: %dxc -T ps_6_6 -Od %s | FileCheck %s
+
+// Regression check for when resources are used by
+// arithmetic that can be trivially replaced by constants.
+
+// CHECK: @main
```
tex3d, line 6: Doesn't make sure the resource isn't used, so what's it checking for?

lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
```diff
@@ -175,6 +274,119 @@ static bool EraseDeadBlocks(Function &F, DxilValueCache *DVC) {
...
+}
+
+static bool IsDxBreak(Instruction *I) {
+  CallInst *CI = dyn_cast<CallInst>(I);
+  if (!CI) return false;
+  Function *CalledFunction = CI->getCalledFunction();
+  return CalledFunction && CalledFunction->getName() == hlsl::DXIL::kDxBreakFuncName;
+}
+
+static bool IsIsHelperLane(Instruction *I) {
```
tex3d, line 299: You can use `IsDxilOpFuncCallInst(I, hlsl::DXIL::OpCode::IsHelperLane)` instead.

tex3d (approved):
LGTM!

Issue #3973, opened 2021-09-27
Type coercion into scalar fails with internal compiler error


sndels:
Implicit narrowing conversions into a scalar type seem to cause `Internal Compiler error: llvm::cast<X>() argument of incompatible type!`. Doesn't appear to happen when coercing into a vector type.

Minimal example, reproduces with dndxc.exe from de6a8ed4's AppVeyor build
Fails: `-enable-templates -DFAILURE_CASE`
Compiles: `-enable-templates`
```
#ifdef FAILURE_CASE
template<typename T>
T Fn(T input)
{
    T ret = input - 1.f;

    return ret;
}

#else

template<typename T>
T Fn(T input)
{
    T ret = input - (T)1.f;

    return ret;
}

#endif

void main()
{
    Fn(1.0);
}
```

llvm-beanz:
This is also respond by PR #3991

PR #3327, opened 2020-12-18
Fix crash in Sema::CheckDerivedToBaseConversion when Paths is empty.

python3kgae:

PR #3991, opened 2021-10-05
Fix cast construction in template instantiation for Constant and Texture buffers

### LINKED ISSUE #3972: Reading from a ConstantBuffer within a template fails compilation in a non-library profile
Reading from a ConstantBuffer within a template causes `error: External function used in non-library profile: : \01??$Fn@N@@YANN@Z`. Making the read within a separate function seems to be a workaround. This could be tied to #3731, but the error is quite different.

Minimal example, reproduces with dndxc.exe from de6a8ed4's AppVeyor build
Fails: `-enable-templates -DFAILURE_CASE`
Compiles: `-enable-templates`
```
struct ConstantsStruct
{
    float Val : Val;
};
ConstantBuffer<ConstantsStruct> Constants;

#ifdef FAILURE_CASE
template<typename T>
T Fn(T input)
{
    // Also tried to explicitly cast into T here without luck
    T ret = input - Constants.Val;

    return ret;
}

#else

float ReadConstant()
{
    return Constants.Val;
}
template<typename T>
T Fn(T input)
{
    // Need explicit cast here, type coercion fails with internal compiler error
    // Separate issue for this
    T ret = input - (T)ReadConstant();

    return ret;
}

#endif

void main()
{
    Fn(1.0);
}
```

llvm-beanz:
ConstantBuffers and TextureBuffers need to behave at the source and AST levels as if they implicitly convert to the underlying buffer data type.

When generating the AST we generate FlatConversion casts, but we don't have the proper overrides in place in `Sema::PerformObjectMemberConversion` to generate the same conversions on AST transformations that occur during template instantiation.

This is the root cause of issue #3327, so this change also includes reverting 2903170ac.

This change resolves #3972.

### DIFF
--- a/tools/clang/lib/Sema/SemaExpr.cpp
+++ b/tools/clang/lib/Sema/SemaExpr.cpp
@@ -11,6 +11,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "clang/AST/OperationKinds.h"
 #include "clang/Sema/SemaInternal.h"
 #include "TreeTransform.h"
 #include "clang/AST/ASTConsumer.h"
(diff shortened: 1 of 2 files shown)


tools/clang/unittests/HLSL/OptionsTest.cpp
```diff
@@ -198,7 +198,7 @@ TEST_F(OptionsTest, ReadOptionsWhenInvalidThenFail) {
   const wchar_t *ArgsNoArg[] = {L"exe.exe", L"hlsl.hlsl", L"/E", L"main",
                                 L"/T"};
   const wchar_t *ArgsUnknown[] = { L"exe.exe", L"hlsl.hlsl", L"/E", L"main",
-    L"/T" L"ps_6_0", L"--unknown"};
+    (L"/T" L"ps_6_0"), L"--unknown"};
```
pow2clk, line 201: Looks like your other change got mixed in here. Might want to rebase to avoid confusion

llvm-beanz: Should be fixed now :)

llvm-beanz:
This also resolves #3973

llvm-beanz:
> This also resolves #3973

Or not... there a different error being surfaced in that case with this patch applied. Will address that in a separate PR.

PR #4014, opened 2021-10-15
Fix bug with HLSL case-insensitive attribute parsing

tex3d:
During parsing, some properties about attributes are determined by
matching the identifier used in the source with the attribute spelling
in Attr.td. The assumption here was that attributes are all
case-sensitive.  But HLSL attributes are not case-sensitive, and even
though code that matches these attributes to kind and determines
whether the attribute is valid has been adjusted to handle
case-insensitivity, the code that looks up properties had not.

One impact was that identifier arguments were not being recognized and
stored in the attributes properly (attributeHasIdentifierArg).

This change makes normalizeAttrName use lower() to force lowercase for
the incoming attribute name before it gets used in the StringSwitch.
normalizeAttrName now must return std::string instead of StringRef
since it's potentially changing the string, not just slicing it.

It would seem better to use the AttributeList::Kind in a switch instead
of matching the string used, but that would be a much more significant
change.

Note: for attributes to work now, all spellings for valid attributes
must be lower case in Attr.td (fortunately they currently are).

### DIFF
--- a/tools/clang/lib/Parse/ParseDecl.cpp
+++ b/tools/clang/lib/Parse/ParseDecl.cpp
@@ -644,10 +644,10 @@ bool Parser::MaybeParseHLSLAttributes(std::vector<hlsl::UnusualAnnotation *> &ta
 // HLSL Change Ends
 
 /// \brief Normalizes an attribute name by dropping prefixed and suffixed __.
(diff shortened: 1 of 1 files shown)


pow2clk (approved):
Simple enough 😄

PR #4017, opened 2021-10-18
Enable printing dependencies of compilation target

jaebaek:
This commit adds `-dump-dependencies` option to print dependencies of
compilation target.

For example, when a HLSL file X includes Y, Z, W files,

$ ./bin/dxc -dump-dependencies -T ps_6_0 -E main X

will print:
X
Y
Z
W

The first line i.e., X is the name of the main compilation target.
Following lines are dependencies for X.


tools/clang/tools/dxclib/dxc.cpp
```diff
@@ -264,6 +264,14 @@ static void WriteDxcExtraOuputs(IDxcResult *pResult) {
   }
 }
 
+std::string getDependencyOutputFileName(llvm::StringRef inputFileName) {
+  size_t extensionOffset = inputFileName.rfind('.');
+  if (extensionOffset == llvm::StringRef::npos)
+    return inputFileName;
```
tex3d, line 270: So this will cause the dependency dump to be written back to the input file if it didn't have an extension?  Is that really what we intend?  It seems like we would instead want to set `extensionOffset` to `inputFileName.size()` so `.d` would be appended without removing any extension.

I think the desired behavior can be within in one expression:
```
return inputFileName.substr(0, inputFileName.rfind('.')).str() + ".d";
```

llvm-beanz:
Can we get some test cases to verify the flag behaviors?

tools/clang/tools/dxclib/dxc.cpp
jaebaek: Ah .. this was a mistake. Thank you for letting me know!

jaebaek:
@llvm-beanz I am going to add some tests under `tools/clang/test/HLSLFileCheck/info/dump_dependency/` and they will be parts of `CompilerTest` in tools/clang/unittests/HLSL/CompilerTest.cpp. If you have any better suggestions, please let me know!

include/dxc/Support/HLSLOptions.td
```diff
@@ -229,6 +229,12 @@ def _version : Flag<["--"], "version">, Group<hlslcore_Group>, Flags<[DriverOpti
 
 def ast_dump : Flag<["-", "/"], "ast-dump">, Flags<[CoreOption, DriverOption, HelpHidden]>,
   HelpText<"Dumps the parsed Abstract Syntax Tree.">; // should not be core, but handy workaround until explicit API written
+def dump_dependencies : Flag<["-", "/"], "M">, Flags<[CoreOption, DriverOption]>,
+  HelpText<"Dumps the list of the compilation target dependencies.">;
+def write_dependencies : Flag<["-", "/"], "MD">, Flags<[CoreOption, DriverOption]>,
+  HelpText<"Write a file with .d extension that will contain the list of the compilation target dependencies.">;
+def write_dependencies_to : JoinedOrSeparate<["-", "/"], "MF">, MetaVarName<"<file>">, Flags<[CoreOption, DriverOption]>,
```
tex3d, line 236: Both `write_dependencies` and `write_dependencies_to` appear to only be implemented by dxc.exe, so I don't think they should include `CoreOption` here.

Also, just as an aside about having these at all: we don't have these sorts of extra options for automatically writing to a file based on the input filename or specifying an output filename specifically for other analogous options (like ast-dump), so while this is consistent with clang's driver options, it feels a bit like option bloat.  It's not hard to redirect stdout.  But that's just a minor quibble, and we should probably keep these for convenience if that's what people want - just limited to `DriverOption` since they are only in dxc.exe, not the DxCompiler API.

tools/clang/test/HLSLFileCheck/info/dump_dependency/dump_dependency.hlsl
```diff
@@ -0,0 +1,20 @@
+// RUN: %dxc -E main -T ps_6_0 -M %s | FileCheck %s
+// RUN: %dxc -E main -T ps_6_0 -MF%s.deps %s && cat %s.deps | FileCheck %s
```
tex3d, line 2: How does this succeed on Windows in TAEF?  These commands are not run in a shell (see FileCheckerTest.cpp), so `&&` won't separate commands, and `cat` wouldn't be understood.  Additionally, -MF and -MD are not handled by the DxCompiler API, so this would have to actually be run through dxc.exe command line interface in order to test this.

For command line tests on Windows targets, we use `utils\hct\hcttestcmds.cmd`.

tex3d: Of course removing the options would simplify things, since it would resolve the test issue I raised without requiring you to implement separate tests in hcttestcmds.cmd. 😃

include/dxc/Support/HLSLOptions.td
jaebaek: Done.

> Also, just as an aside about having these at all: we don't have these sorts of extra options for automatically writing to a file based on the input filename or specifying an output filename specifically for other analogous options (like ast-dump), so while this is consistent with clang's driver options, it feels a bit like option bloat. It's not hard to redirect stdout. But that's just a minor quibble, and we should probably keep these for convenience if that's what people want - just limited to DriverOption since they are only in dxc.exe, not the DxCompiler API.

I do not have a strong opinion on it. As far as I remember, `-MF` and `-MD` was suggested by Chris in our meeting. I agreed on it because `-MD` is what CMake uses to keep/check the dependencies.

tools/clang/test/HLSLFileCheck/info/dump_dependency/dump_dependency.hlsl
jaebaek: I moved this unittests to `utils\hct\hcttestcmds.cmd`. Thank you for clarifying it!

jaebaek:
@llvm-beanz Could you please review this PR?

In terms of the unit test, I tried to add a test for `/MD` but it was not easy because I am not experienced on the batch script.
```
rem ============================================
rem Check that file exists and find text
:check_file
rem echo check_file %*
if "%1"=="log" (
  set check_file_pattern=%OutputLog%
) else (
  set check_file_pattern=%CD%\%1
)
if not exist %check_file_pattern% (
  if !Failed! equ 0 (
    call :set_failed
    echo Failed to find output file %check_file_pattern%
  )
  exit /b 1
)
shift /1
```

`utils/hct/hcttestcmds.cmd` contains the above code and I wanted to check the prefix of `%1` before `set check_file_pattern=%CD%\%1`.
I used the following code:
```
set check_file_pattern=%CD%\%1
if not exist !check_file_pattern! (
  set check_file_pattern=%1
)
```
but the test failed. If you are versed on the batch script, please give me some advice. Thanks!

llvm-beanz:
Hey @jaebaek, sorry it took me so long to get back to you here.

I think everything in your PR is actually good to go. The PR test failure is actually related to a test that broke in master which I fixed at the end of last week. If you rebase the branch and push it up, you should be good to go.

llvm-beanz (approved):
LGTM!

PR #4046, opened 2021-10-29
Fix issues in intrinsic properties in hctdb.py

tex3d:
Two changes that impact RDAT, and thus validation:
- AnnotateHandle and CreateHandleFromBinding can be translated down to SM 6.0
- Sampler Feedback operations available in additional stages
  The translation issue here is the same as for other sampler/deriv ops.
  Currently, we mask off the additional stages for RDAT, while allowing
  those instructions on those stages for the validator.  RDAT will have
  to be fixed to describe the more complicated feature requirements of
  these operations, but that's not necessary now since RDAT will only be
  used for DXR stage for now, meaning that this deficiency won't block
  any valid scenario.

Couple of minor edits with no functional impact:
- sync stage order, enabling merging of some generated conditions
- fix a couple bad descriptions for return values on dxil ops

### DIFF
--- a/lib/DXIL/DxilOperations.cpp
+++ b/lib/DXIL/DxilOperations.cpp
@@ -672,11 +672,6 @@ void OP::GetMinShaderModelAndMask(OpCode C, bool bWithTranslation,
     mask = SFLAG(Hull);
     return;
   }
-  // Instructions: QuadReadLaneAt=122, QuadOp=123
-  if ((122 <= op && op <= 123)) {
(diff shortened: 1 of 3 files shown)


pow2clk (approved):
Looks good!

tex3d:
I'm abandoning this, since first, it turns out SamplerFeedback sample ops weren't part of the spec adding sampler ops to compute-like shader stages, and second, the issue with handle down-level conversion for RDAT isn't important enough to make a validation compatibility breaking change when we don't actually support the actual downlevel conversion of the 6.6 handle operations in the linker yet.  Both these will have to be revisited in the future.  The other non-functional changes can be cleaned up separately later as well.

Issue #3767, opened 2021-05-12
[SPIR-V] Invalid OpLoad when using vk::image_format


nyorain:
Using `vk::image_format` introduced in #3395 on RWTexture results in invalid SPIR-V being generated. Judging by the tests, it looks like the attribute was only tested for RWBuffer.

Minimal example:
```hlsl
[[vk::image_format("r8")]][[vk::binding(0,0)]] RWTexture2D<unorm float> testTex;

[numthreads(4,4,1)]
void csTest(uint3 dtID : SV_DispatchThreadID)
{
	testTex[uint2(0 , 0)] = 0.f;
}
```

Produces the following error:
> fatal error: generated SPIR-V is invalid: OpLoad Result Type <id> '5[%type_2d_image_0]' does not match Pointer <id> '4[%testTex]'s type.  %18 = OpLoad %type_2d_image_0 %testTex
note: please file a bug report on https://github.com/Microsoft/DirectXShaderCompiler/issues with source code if possible

It seems that DXC doesn't perform HLSL legalization in this case, even though it would be needed. When manually forcing legalization via `-Oconfig` or passing `-fspv-flatten-resource-arrays` (which just always forces the legalization pass to be done), the error is gone. 

Maybe @seppala2 knows more (thanks a lot for this feature btw, we need it).

seppala2:
Seems like the issue is just the lack of legalization pass, since the implementation assumes that the type is propagated also for all OpLoads as it is in the legalization. I initially assumed the legalization pass was always done, but apparently it is only done when it is determined as necessary (?). More complex use cases and the unit test work fine since they have legalization enabled due to textures being passed as variables etc.

So I guess there are two options for fixing this:
1) Enable legalization whenever an image with vk::image_format is used for OpLoad
**or**
2) Add a new pass that propagates the image format of the image to all OpLoads. Something like that was actually initially in my pull request (except it was embedded in some other visitor), but then I thought that it wouldn't be needed at all due to legalization already taking care of it

@jaebaek Do you think one of these options could be appropriate?

jaebaek:
Let me first check what happens. I prefer the first option because adding new passes for specific cases would make spirv-opt too big, but I should first check if we have to consider the option of adding a new pass. I will get back after testing some code.

PR #4048, opened 2021-10-29
[spirv] Enable legalization when the vk::image_format attribute is used

seppala2:
Legalization is required to propagate the correct type of the OpLoad even when
the resource variables are not being passed or assigned

This fixes a validation error in simple shaders using vk::image_format, where legalization
would not be needed otherwise

This fixes https://github.com/microsoft/DirectXShaderCompiler/issues/3767


jaebaek (approved):
Thanks for your contribution!

PR #4052, opened 2021-11-01
Support phi on dynamic resource.

python3kgae:


lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2781,12 +2783,329 @@ ModulePass *llvm::createDxilAllocateResourcesForLibPass() {
...
+            }
+          }
+
+          Value *res = annot.get_res();
+          sel->setOperand(2, res);
+        }
+      }
+
+      // Insert after sel.
+      IRBuilder<> B(sel->getNextNode());
```
tex3d, line 3091: Will creating the builder on the next inst break the connection with the right debug info?

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2781,12 +2783,329 @@ ModulePass *llvm::createDxilAllocateResourcesForLibPass() {
...
+      CallInst *annotCI = B.CreateCall(annotHdlFn, {OpArg, sel, props});
```
tex3d, line 3092: Initial value for resource here could be undef to prevent replacement of use with self before it's re-set below.

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2813,9 +3139,13 @@ class DxilCleanupAnnotateHandle : public ModulePass {
...
               continue;
             CallInst *CRes = dyn_cast<CallInst>(Res);
+            if (!CRes)
+              continue;
             DxilInst_AnnotateHandle PrevAH(CRes);
             if (PrevAH) {
-              DXASSERT(AH.get_props() == PrevAH.get_props(), "otherwise, AnnotateHandle chain with inconsistent props");
+              DXASSERT(
+                  AH.get_props() == PrevAH.get_props(),
+                  "otherwise, AnnotateHandle chain with inconsistent props");
```
tex3d, line 3148: Where does this happen now?  If we are still merging redundant annotate handles, shouldn't we be using tryMergeProps() here too?

tex3d (commented):
I have some comments, and I think we still need to add the new pass into `DxilLinkJob::RunPreparePass` in DxilLinker.cpp.

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2781,12 +2783,329 @@ ModulePass *llvm::createDxilAllocateResourcesForLibPass() {
...
+      return getIsSampler(Sel->getTrueValue());
+    } else if (CallInst *CI = dyn_cast<CallInst>(Hdl)) {
+      if (!hlsl::OP::IsDxilOpFuncCallInst(CI))
+        return nullptr;
+      DxilInst_CreateHandleFromHeap createHdl(CI);
+      if (createHdl)
+        return createHdl.get_samplerHeap();
+      DxilInst_AnnotateHandle annotHdl(CI);
+      if (annotHdl)
+        return getDynamicResourceIndex(annotHdl.get_res());
```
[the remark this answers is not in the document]
python3kgae: Here is go check dynamic resource index for the input handle of annotateHandle.
But since we already sink annotateHandle after phi, we can just remove annotateHandle check here.

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2781,12 +2783,329 @@ ModulePass *llvm::createDxilAllocateResourcesForLibPass() {
...
+          }
+        }
+      }
+      // Insert after phi.
+      IRBuilder<> B(phi->getParent()->getFirstNonPHI());
+      CallInst *annotCI = B.CreateCall(annotHdlFn, {OpArg, phi, props});
+      phi->replaceAllUsesWith(annotCI);
+      annotCI->setArgOperand(DxilInst_AnnotateHandle::arg_res, phi);
+    } else {
+      SelectInst *sel = dyn_cast<SelectInst>(Hdl);
```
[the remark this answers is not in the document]
python3kgae: The set will only insert two types of instructions, phi and select.

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2781,12 +2788,409 @@ ModulePass *llvm::createDxilAllocateResourcesForLibPass() {
...
+    Function *createHdlFromHeap =
+        hlslOP->GetOpFunc(DXIL::OpCode::CreateHandleFromHeap, pVoidTy);
+    Value *hdlFromHeapOP = hlslOP->GetI32Const(
+        static_cast<unsigned>(DXIL::OpCode::CreateHandleFromHeap));
+    for (auto &Select : HandleSelects) {
+      if (PHINode *Phi = dyn_cast<PHINode>(Select)) {
+        unsigned numIncoming = Phi->getNumIncomingValues();
+        CreateHandleFromHeapArgs &args = HandleToArgs[Phi];
+        PHINode *newPhi = cast<PHINode>(args.Index);
+        bool bAllIndexResolved = true;
```
tex3d, line 2943: All checking related to `bAllIndexResolved` and additional merging seems like it should be unnecessary now that isResolved is on the args for this phi.  If not resolved, we can't replace it.  If it is resolved, we can, and all incoming args with values must be there.  Right?

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2781,12 +2788,409 @@ ModulePass *llvm::createDxilAllocateResourcesForLibPass() {
...
+        }
+        if (bAllIndexResolved) {
+          IRBuilder<> B(Phi->getParent()->getFirstNonPHI());
+          B.SetCurrentDebugLocation(Phi->getDebugLoc());
+          Value *isSampler = hlslOP->GetI1Const(args.isSampler);
+          // TODO: or args.IsNonUniform with !isUniform(Phi).
+          Value *isNonUniform = hlslOP->GetI1Const(args.isNonUniform);
+          CallInst *newCI =
+              B.CreateCall(createHdlFromHeap,
+                           {hdlFromHeapOP, newPhi, isSampler, isNonUniform});
```
tex3d, line 2968: Couldn't the new call become unused if it ultimately was only used by other phi/select instructions that we later replace?  Also, the original CreateHandleFromHeap calls that become dead won't be cleaned up without subsequent DCE.  Is that just what you were expecting to happen after this?  Since DCE will only occur when optimizing, perhaps we shouldn't leave extra instructions behind when we know they should be cleaned up?

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2781,12 +2787,397 @@ ModulePass *llvm::createDxilAllocateResourcesForLibPass() {
...
+            DXASSERT(it != HandleToArgs.end(),
+                     "args.isResolved should be false");
+            CreateHandleFromHeapArgs &itArgs = it->second;
+            newPhi->setIncomingValue(j, itArgs.Index);
+          }
+
+          IRBuilder<> B(Phi->getParent()->getFirstNonPHI());
+          B.SetCurrentDebugLocation(Phi->getDebugLoc());
+          Value *isSampler = hlslOP->GetI1Const(args.isSampler);
+          // TODO: or args.IsNonUniform with !isUniform(Phi).
```
tex3d, line 2957: Or emit warning or error if `(isUniform(Phi) && !args.IsNonUniform)`, since we should encourage explicit nonuniform marking and sinking of indexing in these cases, rather than silently try to fix it for them with performance implications.

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2781,12 +2787,397 @@ ModulePass *llvm::createDxilAllocateResourcesForLibPass() {
...
+      continue;
+    for (auto U = F->user_begin(); U != F->user_end();) {
+      CallInst *CI = dyn_cast<CallInst>(*(U++));
+      annotHdls.emplace_back(CI);
+    }
+  }
+  if (annotHdls.empty())
+    return false;
+
+  DenseSet<Instruction *> selectAnnotHdls;
```
tex3d, line 3081: Need to use SetVector, otherwise iteration will be pointer-dependent, leading to potentially different orderings of replacement calls after phi nodes on different runs (nondeterminism).

tex3d (commented):
Mostly minor comments, but one source of nondeterminism (`DenseSet<Instruction *> selectAnnotHdls;`).

tex3d (approved):

PR #4053, opened 2021-11-01
Added source encoding option

gracejennings:
Using the existing -encoding option to also specify source encoding. Includes support for UTF-8 and UTF-16 encoded sources with and without BOMs by using specified encoding if a BOM is not detected in a source file. 

TO-DO: Add a specific error for UTF-16 encoded file without a BOM used with -encoding utf8 option

### DIFF
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -83,6 +83,9 @@ public:
     UINT32 codePage;
     LoadSourceCallResult() : hr(E_FAIL), codePage(0) { }
     LoadSourceCallResult(const char *pSource, UINT32 codePage = CP_UTF8) : hr(S_OK), source(pSource), codePage(codePage) { }
+    LoadSourceCallResult(const void *pSource, size_t size,
+                         UINT32 codePage = CP_ACP)
+        : hr(S_OK), source((const char *)pSource, size), codePage(codePage) {}
   };
   std::vector<LoadSourceCallResult> CallResults;
   size_t callIndex;
@@ -157,6 +160,17 @@ public:
   TEST_METHOD(CompileWhenODumpThenPassConfig)
   TEST_METHOD(CompileWhenODumpThenOptimizerMatch)
   TEST_METHOD(CompileWhenVdThenProducesDxilContainer)
+
+  void TestEncodingImpl(
+      const void *sourceData,
+      size_t sourceSize,
+      UINT32 codePage,
+      const void *includedData,
+      size_t includedSize,
+      bool succ,
+      const WCHAR *encoding = nullptr
+  );
+  TEST_METHOD(CompileWithEncodeFlagTestSource)
 
 #if _ITERATOR_DEBUG_LEVEL==0 
   // CompileWhenNoMemThenOOM can properly detect leaks only when debug iterators are disabled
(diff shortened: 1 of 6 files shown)


tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2589,6 +2603,80 @@ TEST_F(CompilerTest, CompileWhenVdThenProducesDxilContainer) {
...
+  std::string includeUtf8 = "// Comment\n";
+  std::string utf8BOM = "\xEF" "\xBB" "\xBF"; // UTF-8 BOM
+  std::string includeUtf8BOM = utf8BOM + includeUtf8;
+
+  TestEncodingImpl(sourceUtf8.data(), sourceUtf8.size(), DXC_CP_UTF8,
+                   includeUtf8BOM.data(), includeUtf8BOM.size(),
+                   true, L"utf16");
+
+  TestEncodingImpl(sourceUtf8.data(), sourceUtf8.size(), DXC_CP_UTF8,
+                   includeUtf8.data(), includeUtf8.size(), false, L"utf16");
```
tex3d, line 2652: So this is testing an invalid case where files are utf8 without BOM, and you say expect utf16.  Are you including it here because you get an encoding error?  Or is the size odd?  I think mismatch cases like this and the one below for the reverse can't be handled in any universal way really, so we should perhaps just skip the test?  Though if it currently fails slightly more gracefully due to your change to the error handling, I suppose this one could be ok.  I just feel like it shouldn't always be expected to detect mismatch for any arbitrary input.

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2589,6 +2603,80 @@ TEST_F(CompilerTest, CompileWhenVdThenProducesDxilContainer) {
...
+
+  TestEncodingImpl(sourceUtf16.data(), sourceUtf16.size() * sizeof(L'A'),
+                   DXC_CP_UTF16,
+                   includeUtf16BOM.data(), includeUtf16BOM.size() * sizeof(L'A'), true, L"utf8");
+
+   //TO-DO: not hitting a custom error here, need to add a test that checks if a file specified
+   //as utf8 is actually utf encoded earlier in compilation
+  //TestEncodingImpl(sourceUtf16.data(), sourceUtf16.size() * sizeof(L'A'),
+  //                 DXC_CP_UTF16,
+  //                 includeUtf16.data(), includeUtf16.size() * sizeof(L'A'), false, L"utf8");
```
tex3d, line 2672: I think it's ok to remove this test because it's testing a case we can't detect reliably in the general case.

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -3007,7 +3095,7 @@ TEST_F(CompilerTest, CompileHlsl2015ThenFail) {
   VERIFY_ARE_EQUAL(status, E_INVALIDARG);
   VERIFY_SUCCEEDED(pResult->GetErrorBuffer(&pErrors));
   LPCSTR pErrorMsg = "HLSL Version 2015 is only supported for language services";
-  CheckOperationResultMsgs(pResult, &pErrorMsg, 1, false, false);
+  CheckOperationResultMsgs(pResult, &pErrorMsg, 1, false, false); /// use this for the failing cases
```
tex3d, line 3098: Not sure why this random comment was added (it's harmless) but is a change outside the context of other changes you are making.  You could clean it up on next revision when you delete the commented out test case.

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2589,6 +2603,80 @@ TEST_F(CompilerTest, CompileWhenVdThenProducesDxilContainer) {
...
+   //TO-DO: not hitting a custom error here, need to add a test that checks if a file specified
+   //as utf8 is actually utf encoded earlier in compilation
+  //TestEncodingImpl(sourceUtf16.data(), sourceUtf16.size() * sizeof(L'A'),
+  //                 DXC_CP_UTF16,
+  //                 includeUtf16.data(), includeUtf16.size() * sizeof(L'A'), false, L"utf8");
+
+  TestEncodingImpl(sourceUtf16.data(), sourceUtf16.size() * sizeof(L'A'),
+                   DXC_CP_UTF16,
+                   includeUtf16.data(), includeUtf16.size() * sizeof(L'A'),
+                   true, L"utf16");
```
tex3d, line 2677: One thing that might be missing is the mixed case where the input buffer is one type, but the default and included file is a different type.  That seems like it could be interesting to test in case there are any assumptions between the main source file input format and the included format.

tools/clang/unittests/HLSL/CompilerTest.cpp
gracejennings: Previously there was a failed assert in checking the size, so this one of the locations where I changed it to throw a more specific error instead of a failed assert. That makes sense that it shouldn't be expected to detect mismatch for any arbitrary input, so I will remove it from the test cases.

PR #4054, opened 2021-11-02
Fix capitalization of RUN in SPIR-V tests

llvm-beanz:
The SPIR-V tests have different capitalization in their `RUN` directives from the rest of the HLSL tests or from LLVM-LIT.

This unifies all the tests and infrastructure to use `RUN`.

The first commit in this PR is the logic changes to the testing infrastructure. The second commit is the giant mass-modify of the test cases.


jaebaek (approved):
LGTM

PR #4057, opened 2021-11-03
Require Windows 7 Everywhere

llvm-beanz:
Some of the LLVM APIs were expecting Windows XP support, while other
places in the code were going with whatever your SDK configured.
This should get things consistent and avoid mismatches in the Windows
headers.


hekota (approved):
LGTM!

PR #4060, opened 2021-11-04
[WSL] Support using one checkout for Win & WSL

llvm-beanz:
This change updates the HCTGen code for unix systems to match the line
endings of the checked out file for in-tree generation. It also updates
the SPIR-V whole file tests to strip trailing carriage returns.


pow2clk (approved):
Thanks! This will help windows devs debug Linux breaks

PR #4062, opened 2021-11-04
Add back HlslIntrinsicOp.h

llvm-beanz:
This header is also packaged up as part of the interface for handling
DXIL, so it needs to live in-tree.


hekota (approved):
LGTM!

Issue #4005, opened 2021-10-13
Assignment operator cannot be overloaded with types other than the same struct


pow2clk:
A slight modification of the assign operator overloading test (https://github.com/microsoft/DirectXShaderCompiler/blob/master/tools/clang/test/HLSLFileCheck/hlsl/operator_overloading/operator.overloading.assign.hlsl)  reveals a limitation of its implementation:

```c++
struct Number {
    int n;

    void operator=(float x) {
        n = x;
    }
};

int main(float4 pos: SV_Position) : SV_Target {
    Number a = {pos.x};
    Number b = {pos.y};
    a = pos.x;
    return a.n;
}
```
error: cannot implicitly convert from 'float' to 'Number'
    a = pos.x;

This is permitted in C++ and would be useful here as well.

jaebaek:
I am taking a look at this issue. It seems to be related to both #4004 and #4066

jaebaek:
This was fixed by #4063

PR #4063, opened 2021-11-04
Handle implicit operator overloading

### LINKED ISSUE #4004: Structs with overloaded operators cannot be copied using default assign operator
Using overloaded operators of any kind causes an attempt to copy that struct.

Consider this fairly silly example:
```c++
#define MAX_SIZE 100

struct MyArray {
  float4 A[MAX_SIZE];
  
  void splat(float f) {
    for (int i = 0; i < MAX_SIZE; i++)
      A[i] = f;
  };

#if ERROR

  MyArray operator+(MyArray RHS) {
    MyArray OutArray;
    for (int i = 0; i < MAX_SIZE; i++)
      OutArray.A[i] = A[i] + RHS.A[i];
    return OutArray;
  };
#endif
};


float4 main(float4 col1: COLOR0, float4 col2: COLOR2, int ix : I) : SV_Target {
  MyArray A, B;
  B.splat(col1);
  A = B;
  return A.A[ix];
}
```
The result when compiled with ERROR defined is Internal compiler error: access violation. Attempted to read from address 0x0000000000000020

jaebaek:
Currently, for a binary operator, we check whether the RHS expr defines
an operator using the operator overloading or not regardless the binary
operator to handle. If the RHS expr defines such operator, we build the
operator using the defined one. However, it can result in trying to
build an operator that is not actually defined by the RHS expr.
This commit lets it decide whether it has to build a new operator or use
some pre-defined one based on the type of the binary operator to handle
and the operator defined by the RHS expr.

Fixes #4004


jaebaek:
Will fix #4066

jaebaek:
Will fix #4005

jaebaek:
@pow2clk @tex3d 
I am a little bit busy for other internal tasks so I am delaying issues other than this one. Since this looks critical for hlsl-2021, I handled this first. I do not want to let the bugs I made have negative impact on the entire hlsl-2021.

Could you please review this PR?

pow2clk (approved):
Looks good! Thanks for taking this on!

PR #4065, opened 2021-11-07
Minor fix for dependencies of generated sources

llvm-beanz:
For the `CODE_TAG` sources we need to ensure that they are dependencies
of the generation command. This also requires creating an extra
temporary file so that the dependencies aren't circular.

### DIFF
--- a/cmake/modules/HCT.cmake
+++ b/cmake/modules/HCT.cmake
@@ -69,7 +69,7 @@ function(add_hlsl_hctgen mode)
       # source copying is enabled for CODE_TAG sources. That means we need to
       # create an extra temporary to key the copy step on.
       set(output ${temp_output}.2)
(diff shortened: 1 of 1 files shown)


cmake/modules/HCT.cmake
```diff
@@ -54,6 +58,15 @@ function(add_hlsl_hctgen mode)
...
       endif()
     endif()
+
+    list(APPEND hct_dependencies ${full_output})
+    if (HLSL_COPY_GENERATED_SOURCES)
+      # The generation command both depends on and produces the final output if
+      # source copying is enabled for CODE_TAG sources. That means we need to
+      # create an extra temporary to key the copy step on.
+      set(output ${temp_output}.2)
+      set(secomd_copy COMMAND ${CMAKE_COMMAND} -E copy_if_different ${temp_output} ${temp_output}.2)
```
pow2clk, line 68: minor typo? secomd?

pow2clk (approved):
Minor spelling issue you can take or leave.

Thanks for resolving this!

Issue #3772, opened 2021-05-16
[SPIR-V] Device Buffer Address- feature


Eichenherz:
Hello ! 
In Vulkan it's possible to query the gpu address of a buffer and pass that as a uint64_t, pretty much like a pointer in C/C++.
```cpp
inline u64 VkGetBufferDeviceAddress( VkDevice vkDevice, VkBuffer hndl )
{
	VkBufferDeviceAddressInfo deviceAddrInfo = {};
	deviceAddrInfo.sType = VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO;
	deviceAddrInfo.buffer = hndl;

	return vkGetBufferDeviceAddress( vkDevice, &deviceAddrInfo );
```
```glsl
layout( buffer_reference, scalar, buffer_reference_align = 4 ) readonly buffer mtl_ref{
	material_data materials[];
};
layout( buffer_reference, scalar, buffer_reference_align = 4 ) readonly buffer light_ref{
	light_data lights[];
};

.....................

void main()
{
	material_data mtl = mtl_ref( g.addr + g.materialsOffset ).materials[ mtlIdx ];

	vec4 baseCol = texture( sampler2D( sampledImages[ nonuniformEXT( mtl.baseColIdx ) ], samplers[ nonuniformEXT( 0 ) ] ), uv );
	...................
}
```

As far as I know, DX12 doesn't have this kind of capabilities, but Vulkan does, and since compiling HLSL to SPIR-V is possible, even used in the game industry, it would be very nice to have this. ( Maybe even in DX12, why not ? )  
Thanks !

StarsX:
I think the key point is HLSL hasn’t supported 64-bit indexing for raw buffer yet.

sudonatalie:
This is now supported by [`vk::RawBufferLoad`](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#rawbufferload-and-rawbufferstore) when targeting SPIR-V from HLSL!

Issue #3919, opened 2021-08-30
[SPIR-V] HLSL version of GL_EXT_spirv_intrinsics


jaebaek:
Similar to https://github.com/KhronosGroup/glslang/pull/2625, we allow HLSL shader developers to test arbitrary SPIR-V code (I want to call it "inlined SPIR-V").

### New intrinsics, types and attributes
- vk::ext_execution_mode(uint _execution_mode_, ...);
  - Stand-alone intrinsic function to emit an `OpExecutionMode`.
  - To specify extensions and/or capabilities needed by the `OpExecutionMode`, we have to use `vk::ext_extension(string)` and `vk::ext_capability(uint)` attributes (see below).
  - Return type: void
  - uint _execution_mode_ must be a constant expression.
  - Extra parameters must be constant expressions. Some [execution modes](https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#_execution_mode) e.g., `Invocations`, `LocalSize`, .. have extra parameters. Since extra parameters of `OpExecutionMode` are literals, DXC will emit these extra parameters as literals as well.
- [[vk::ext_extension(string _extension_name_)]]
  - An attribute to specify `OpExtension` instruction.
  - It can be an attribute of `vk::ext_execution_mode(...)`, `vk::ext_execution_mode_id(...)`, a function declared with `vk::ext_instruction(...)`, or a function declared with `[[vk::ext_type_def(...)]]`. We can use multiple `[[vk::ext_extension(string extension_name)]]` for a single function.
  - string _extension_name_ must be a constant string.
- [[vk::ext_capability(uint _capability_)]]
  - An attribute to specify `OpCapability` instruction.
  - It can be an attribute of `vk::ext_execution_mode(...)`, `vk::ext_execution_mode_id(...)`, a function declared with `vk::ext_instruction(...)`, or a function declared with `[[vk::ext_type_def(...)]]`. We can use multiple `[[vk::ext_extension(string extension_name)]]` for a single function.
  - uint _capability_ must be a constant expression.

Example:
```
// HLSL
[[vk::ext_capability(/* StencilExportEXT */ 5013)]]
[[vk::ext_extension("SPV_EXT_shader_stencil_export")]]
vk::ext_execution_mode(/* StencilRefReplacingEXT */ 5027);

// SPIR-V
OpCapability StencilExportEXT
...
OpExtension "SPV_EXT_shader_stencil_export"
...
[...]

PR #4069, opened 2021-11-08
[spirv] Add Vk::RawBufferLoad intrinsic

yuriy-odonnell-epic:
This PR adds a basic implementation of memory loads through `buffer_reference` mechanism in SPIR-V. Partially addresses #3772.

Current implementation only handles basic `uint` loads, with similar semantics as HLSL `ByteAddressBuffer` (4 byte alignment).

Example shader:
```
uint64_t Address;
float4 main() : SV_Target0 {
  uint Value = vk::RawBufferLoad(Address);
  return asfloat(Value);
}
```

Proposed change implements the minimal functionality required for Unreal Engine and I would like to keep the scope of this particular PR minimal. Custom alignment, large/vector loads, templated load syntax, stores, etc. would also be great to implement _(and I could do the work, if needed)_, but I would prefer to do them separately.

**Update:** The PR is now rebased and squashed, ready for review.

### DIFF
--- a/include/dxc/HlslIntrinsicOp.h
+++ b/include/dxc/HlslIntrinsicOp.h
@@ -222,6 +222,9 @@ enum class IntrinsicOp {  IOP_AcceptHitAndEndSearch,
   IOP_unpack_u8u16,
   IOP_unpack_u8u32,
 #ifdef ENABLE_SPIRV_CODEGEN
+  IOP_VkRawBufferLoad,
+#endif // ENABLE_SPIRV_CODEGEN
+#ifdef ENABLE_SPIRV_CODEGEN
   IOP_VkReadClock,
 #endif // ENABLE_SPIRV_CODEGEN
   MOP_Append,
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -212,6 +212,8 @@ public:
   /// the instruction pointer for the result.
   SpirvUnaryOp *createUnaryOp(spv::Op op, QualType resultType,
                               SpirvInstruction *operand, SourceLocation loc);
+  SpirvUnaryOp *createUnaryOp(spv::Op op, const SpirvType *resultType,
+                              SpirvInstruction *operand, SourceLocation loc);
 
   /// \brief Creates a binary operation with the given SPIR-V opcode. Returns
   /// the instruction pointer for the result.
@@ -686,6 +688,10 @@ public:
 
   SpirvString *getString(llvm::StringRef str);
 
+  const HybridPointerType *getPhysicalStorageBufferType(QualType pointee);
+  const SpirvPointerType *
+  getPhysicalStorageBufferType(const SpirvType *pointee);
+
 public:
   std::vector<uint32_t> takeModule();
 
@@ -790,8 +796,14 @@ private:
 
 void SpirvBuilder::requireCapability(spv::Capability cap, SourceLocation loc) {
   auto *capability = new (context) SpirvCapability(loc, cap);
-  if (!mod->addCapability(capability))
+  if (mod->addCapability(capability)) {
+    if (cap == spv::Capability::PhysicalStorageBufferAddresses) {
+      mod->promoteAddressingModel(
+          spv::AddressingModel::PhysicalStorageBuffer64);
+    }
+  } else {
     capability->releaseMemory();
+  }
 }
 
 void SpirvBuilder::requireExtension(llvm::StringRef ext, SourceLocation loc) {
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -687,6 +687,7 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
   }
 
   // Addressing and memory model are required in a valid SPIR-V module.
+  // It may be promoted based on features used by this shader.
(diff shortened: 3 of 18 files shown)


yuriy-odonnell-epic:
Potentially, custom alignment tagging on load ops could be implemented using SPIR-V intrinsics that are being added in #3919.

yuriy-odonnell-epic:
The functionality, docs and test are now implemented. I am not planning to make any significant further changes, so now would be a good time to review.

tools/clang/test/CodeGenSPIRV/intrinsics.vkrawbufferload.hlsl
```diff
@@ -0,0 +1,15 @@
+// RUN: %dxc -T ps_6_0 -E main
+
+// CHECK: OpCapability PhysicalStorageBufferAddresses
+// CHECK: OpExtension "SPV_KHR_physical_storage_buffer"
+// CHECK: OpMemoryModel PhysicalStorageBuffer64 GLSL450
+
+uint64_t Address;
+float4 main() : SV_Target0 {
+  // CHECK: OpTypePointer PhysicalStorageBuffer %uint
+  // CHECK: OpBitcast %_ptr_PhysicalStorageBuffer_uint %24
```
sudonatalie, line 10: We avoid relying on the numerical result-ids (`%24`, `%26`, `%27`, etc) in tests since they can change and don't affect the correctness of the test. Switch these to use pattern matching ([example](https://github.com/microsoft/DirectXShaderCompiler/blob/e4ddf78452be1354d16d3a8cddb73eb6e252a1ca/tools/clang/test/CodeGenSPIRV/binary-op.assign.opaque.array.hlsl#L27-L30)).

tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -213,6 +213,9 @@ class SpirvBuilder {
   SpirvUnaryOp *createUnaryOp(spv::Op op, QualType resultType,
                               SpirvInstruction *operand, SourceLocation loc);
```
sudonatalie, line 215: nit: Remove newline to match surrounding formatting

include/dxc/HlslIntrinsicOp.h
```diff
@@ -221,6 +221,9 @@ enum class IntrinsicOp {  IOP_AcceptHitAndEndSearch,
   IOP_unpack_s8s32,
   IOP_unpack_u8u16,
   IOP_unpack_u8u32,
+#ifdef ENABLE_SPIRV_CODEGEN
```
sudonatalie, line 224: nit: Combine with ifdef below

yuriy-odonnell-epic: This is an auto-generated file.

tools/clang/test/CodeGenSPIRV/intrinsics.vkrawbufferload.hlsl
yuriy-odonnell-epic: Done.

tools/clang/include/clang/SPIRV/SpirvBuilder.h
yuriy-odonnell-epic: Done.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -12525,6 +12529,32 @@ SpirvEmitter::processSpvIntrinsicCallExpr(const CallExpr *expr) {
   return retVal;
 }
 
+SpirvInstruction *SpirvEmitter::processRawBufferLoad(const CallExpr *callExpr) {
+  clang::SourceLocation loc = callExpr->getExprLoc();
+  const clang::Expr *addressExpr = callExpr->getArg(0);
+  SpirvInstruction *address = doExpr(addressExpr);
+
+  const HybridPointerType *bufferType =
+      spvBuilder.getPhysicalStorageBufferType(astContext.UnsignedIntTy);
```
jaebaek, line 12538: Just out of a curiosity, is there any reason that you do not use
```
  const SpirvPointerType *bufferType =
      spvBuilder.getPhysicalStorageBufferType(spvContext.getUIntType(32));
```
?

It is a minor issue, but we prefer `SpirvPointerType` because `const SpirvPointerType *SpirvContext::getPointerType(..)` internally keeps the pointers we already created and reuses it while `const HybridPointerType *SpirvContext::getPointerType(..)` does not.

If there is no special reason, could you please replace it with `SpirvPointerType`?

jaebaek (commented):
Looks good to me. Just a nit.

By the way, this PR clearly and precisely addresses a non-trivial issue, which is pretty impressive. I appreciate your contribution!

jaebaek:
By the way, I observed that running clang-format slightly changes the code format.
Could you please apply this patch?
[diff.txt](https://github.com/microsoft/DirectXShaderCompiler/files/7549016/diff.txt)

I got it by running
```
git clang-format master
git diff > diff.txt
```

tools/clang/lib/SPIRV/SpirvEmitter.cpp
yuriy-odonnell-epic: Done.

yuriy-odonnell-epic:
Applied the `getUIntType` and clang format changes.

jaebaek (approved):
Thanks again for your contribution!
This PR clearly addresses the buffer address support!

yuriy-odonnell-epic:
Thanks for reviewing and merging! I will work on the vectorized, templated and aligned loads next.

PR #4071, opened 2021-11-08
[spirv] Improve lines and columns for DebugLine (pt 1)

greg-lunarg:
In particular, add support for start and end lines and columns for
DebugLine. Previously, end line and column was same as start line and
column.

This is just the first installment of this change. It covers binary ops,
OpAccessChain, OpLoad, OpStore, DebugDeclare, OpSample and OpReturn. A
second commit will complete support.

This commit primarily adds SourceRange to the appropriate emit and
generation functions. This is necessary since sometimes neither value
in the SourceRange matches the accompanying SourceLocation.

Note that while the end column is more correct than it was previously, it
is not perfectly correct. It currently points to the beginning of the last
token of the computation rather than the end of that token. We will
attempt to improve this with a followup commit.

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -173,9 +173,9 @@ public:
   /// <result-type> from the given pointer. Returns the instruction pointer for
   /// the loaded value.
   SpirvLoad *createLoad(QualType resultType, SpirvInstruction *pointer,
-                        SourceLocation loc);
+                        SourceLocation loc, SourceRange range = {});
   SpirvLoad *createLoad(const SpirvType *resultType, SpirvInstruction *pointer,
-                        SourceLocation loc);
+                        SourceLocation loc, SourceRange range = {});
 
   /// \brief Creates an OpCopyObject instruction from the given pointer.
   SpirvCopyObject *createCopyObject(QualType resultType,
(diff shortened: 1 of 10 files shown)


greg-lunarg:
@jaebaek PTAL

greg-lunarg:
@dnovillo Is @jaebaek on vacation? Any estimate on when he might be able to review this?

dnovillo:
Not on vacation, but with a fairly intense load.  How urgent is this?

greg-lunarg:
It is not urgent as in a bugfix. It is a new "feature" that will improve the experience for users of shader source debugging.

jaebaek (commented):
Sorry for being late.
I was busy for intensive workloads.

Looks good to me, but I am just concerned that allocating `SourceRange` for all `SpirvInstruction` objects consumes meaningful amount of memory while it is used by only the non-semantic debug info emission.

I do not have a clear idea how to reduce the memory, but I want to ask you whether it is negligible or not. If it is not small, do you have an idea to reduce the memory usage?

greg-lunarg:
A SourceRange is only two unsigned words long. Consequently, I think the impact on memory usage as well as compile time is negligible.

jaebaek (approved):

greg-lunarg:
Thanks for the review! @jaebaek @dnovillo

jaebaek:
Thank you for your contribution!

PR #4072, opened 2021-11-08
Workaround autocrlf=input

llvm-beanz:
We've seen some lingering odd issues with autocrlf. I think this will
resolve them all because our gitattrbutes sets text=auto which is only
overridden if autocrlf=input.

We'll know if this works based on whether or not the AppVeyor build
works.

### DIFF
--- a/cmake/modules/HCT.cmake
+++ b/cmake/modules/HCT.cmake
@@ -11,9 +11,13 @@ if (WIN32 AND NOT DEFINED HLSL_AUTOCRLF)
                   OUTPUT_VARIABLE output
                   OUTPUT_STRIP_TRAILING_WHITESPACE)
   if( result EQUAL 0 )
-    set(val Off)
-    if (output STREQUAL "true")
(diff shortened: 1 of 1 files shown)


pow2clk (approved):
If appveyor is happy, I'm happy

pow2clk:
I've recently thought that if we used `git diff` instead of `cmake -E compare_files` the line endings are ignored

llvm-beanz:
I don't think that helps. `git diff` only ignores line endings if the line ending settings tell it to (same as the underlying options we're dealing with here). Also, it doesn't compare files outside the source tree. In the build configuration where we copy the files into the source tree we don't emit the `compare_files` calls.

jeremyong:
What crlf settings are expected for use on Windows? After updating recently, suddenly I get a lot of build errors related to file comparisons, and the difference AFAICT is that line endings for temporary copies in the build output use LF endings compared to the originals

PR #4073, opened 2021-11-09
[spirv] isRelaxedPrecisionType() check for referenceTypes as well.

sindney:
Hi, i found that by default, input/output variables 
won't be marked as RelaxedPrecision when using [min16float types](https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#id58). 
So the generated essl for example don't have mediump qualifier.
By considering ReferenceTypes in isRelaxedPrecisionType() function will add the correct decoration: 

```
OpDecorate %out_var_SV_Target0 RelaxedPrecision
OpDecorate %out_var_SV_Target1 RelaxedPrecision
OpDecorate %out_var_SV_Target2 RelaxedPrecision
OpDecorate %38 RelaxedPrecision
OpDecorate %39 RelaxedPrecision
```

```glsl
layout(location = 0) in highp vec4 in_var_TEXCOORD10_centroid;
layout(location = 1) in highp vec4 in_var_TEXCOORD11_centroid;
layout(location = 2) in highp vec4 in_var_TEXCOORD8;
layout(location = 0) out vec4 out_var_SV_Target0;
layout(location = 1) out vec4 out_var_SV_Target1;
layout(location = 2) out vec4 out_var_SV_Target2;
```

### DIFF
--- a/tools/clang/lib/SPIRV/AstTypeProbe.cpp
+++ b/tools/clang/lib/SPIRV/AstTypeProbe.cpp
@@ -1142,6 +1142,14 @@ bool isRelaxedPrecisionType(QualType type, const SpirvCodeGenOptions &opts) {
     }
   }
 
+  // Reference types
+  if (const auto *refType = type->getAs<ReferenceType>())
+    return isRelaxedPrecisionType(refType->getPointeeType(), opts);
+
+  // Pointer types
+  if (const auto *ptrType = type->getAs<PointerType>())
+    return isRelaxedPrecisionType(ptrType->getPointeeType(), opts);
+
   return false;
 }
 
--- a/tools/clang/test/CodeGenSPIRV/spirv.stage-io.16bit.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/spirv.stage-io.16bit.hlsl
@@ -16,6 +16,9 @@
 // CHECK: OpDecorate %out_var_D Location 7
 // CHECK: OpDecorate %out_var_E Location 8
 
+// CHECK: %half = OpTypeFloat 16
+// CHECK-NOT: %float = OpTypeFloat 32
+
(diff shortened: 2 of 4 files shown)


tools/clang/lib/SPIRV/AstTypeProbe.cpp
```diff
@@ -1142,6 +1142,15 @@ bool isRelaxedPrecisionType(QualType type, const SpirvCodeGenOptions &opts) {
     }
   }
 
+  // Handles reference type, input/output variable for example.
+  {
+    QualType pointee = {};
+    if (type->isReferenceType() || type->isPointerType()) {
```
sudonatalie, line 1148: For consistency with how we handle pointee types elsewhere in this file, can you change this to:

```
  // Reference types
  if (const auto *refType = type->getAs<ReferenceType>())
    return isRelaxedPrecisionType(refType->getPointeeType(), opts);

  // Pointer types
  if (const auto *ptrType = type->getAs<PointerType>())
    return isRelaxedPrecisionType(ptrType->getPointeeType(), opts);
```

sudonatalie (changes requested):
This looks like a good change overall, thank you for your contribution!

Could you also add a couple tests to validate the behavior. I believe this will also affect how stage vars are handled with the `-enable-16bit-types` flag, so we should add a test to both CHECK for `RelaxedPrecision` when the flag is off, and CHECK for  `OpTypeFloat 16` when the flag is on.

sindney:
Got it, I'll add some test to cover possible cases, it might take sometime.

tools/clang/lib/SPIRV/AstTypeProbe.cpp
sindney: Hi @sudonatalie, I've committed your suggestions and added a test case to this pr.

sudonatalie (approved):
LGTM!

PR #4074, opened 2021-11-09
Fix leak from HLMatrixLowerPass when cleaning up temp functions

tex3d:


pow2clk (approved):
Looks good!

It's the small bugs that are  hardest to catch

PR #4076, opened 2021-11-12
PIX: Instrument instructions with !dbg

jeffnn:
This is part of work to improve PIX shader debugging breakpoint binding.
Previously, llvm pointer types weren't considered interesting values for PIX because they have no meaning to shader authors, so the pass being modified here did not add PIX's instrumentation to pointer-type instructions. However, lack of instrumentation meant that sometimes PIX can't map an HLSL line to the corresponding instruction, because PIX is only aware of the instrumented ones.
(Also added a teeny little & so we don't take a copy of the instructions in dxil_debug_info::DxcPixDxilDebugInfo::FindInstruction)


tex3d (approved):
LGTM!

raghavthind2005 (approved):

jeffnn:
This wasn't necessary, after all.

PR #4077, opened 2021-11-12
PIX: Use DebugLoc for filenames instead of file list

jeffnn:
This fix is for shaders (in particular those for materials from Unreal Engine) that feature #line directives.

In the UE case, for example, there is one source file called hlsl.hlsl, with many #line directives pointing to the original HLSL files from which the source was assembled via UE's material system.

At the start of shader debugging, all the filenames from DebugLoc are returned from code near here to PIX. Those filenames are the ones from the #line directives.

The previous code would iterate over source file names (which in the UE case would be just hlsl.hlsl), and fail to find the filename that PIX is now asking to find the instructions for.

Instead, we now gather the filenames from the DebugLoc. Note that the filename comparison is done after the line number comparison for performance's sake.

### DIFF
--- a/lib/DxilDia/DxcPixDxilDebugInfo.cpp
+++ b/lib/DxilDia/DxcPixDxilDebugInfo.cpp
@@ -164,40 +164,32 @@ static bool CompareFilenames(const wchar_t * l, const char * r)
 }
 
 dxil_debug_info::DxcPixDxilInstructionOffsets::DxcPixDxilInstructionOffsets(
-  IMalloc *pMalloc,
-  dxil_dia::Session *pSession,
-  const wchar_t *FileName,
+  IMalloc* pMalloc,
+  dxil_dia::Session* pSession,
+  const wchar_t* FileName,
   DWORD SourceLine,
-  DWORD SourceColumn) 
+  DWORD SourceColumn)
 {
   assert(SourceColumn == 0);
   (void)SourceColumn;
-
-  auto files = pSession->Contents()->operands();
-  for (const auto& file : files)
-  {
-    auto candidateFilename = llvm::dyn_cast<llvm::MDString>(file->getOperand(0))
-        ->getString();
-
-    if (CompareFilenames(FileName, candidateFilename.str().c_str()))
-    {
-
-      auto Fn = pSession->DxilModuleRef().GetEntryFunction();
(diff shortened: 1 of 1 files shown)


lib/DxilDia/DxcPixDxilDebugInfo.cpp
```diff
@@ -172,32 +172,24 @@ dxil_debug_info::DxcPixDxilInstructionOffsets::DxcPixDxilInstructionOffsets(
...
+  auto Fn = pSession->DxilModuleRef().GetEntryFunction();
+  auto &Blocks = Fn->getBasicBlockList();
+  for (auto& CurrentBlock : Blocks) {
+    auto& Is = CurrentBlock.getInstList();
+    for (auto& Inst : Is) {
+      auto & debugLoc = Inst.getDebugLoc();
+      if (debugLoc)
+      {
+        unsigned line = debugLoc.getLine();
+        if (line == SourceLine)
```
adam-yang, line 184: Is SourceLine the actual line number in the preprocessed shader or the line in the #line directive?

jeffnn: It's the line (and filename) from the #line directive

adam-yang: I thought the lines in the DebugLoc are actually lines in the flattened preprocessed file, wouldn't that not match with the lines from #line?

jeffnn: More accurately, if the trace is from an older compiler (like, e.g. the UE4 shaders I've been working with), then it's the #lines. If it's a new compiler, or a recompiled (PIX edit-and-continue) one, then it's the preprocessed file.

jeffnn: So for new compiler output, this change doesn't make any difference because the older outer loop over filenames should be the same as the one from the debugloc

adam-yang (approved):

lib/DxilDia/DxcPixDxilDebugInfo.cpp
adam-yang: Okay gotcha. Thank you!

PR #4080, opened 2021-11-15
Code cleanup to help build with newer version of llvm.

python3kgae:


tex3d (approved):
LGTM!

lib/DXIL/DxilModule.cpp
```diff
@@ -31,10 +31,18 @@
...
 #include "llvm/IR/DiagnosticPrinter.h"
 #include "llvm/Support/raw_ostream.h"
+#ifndef _WIN32
 #include "llvm/ADT/STLExtras.h"
+#endif
 #include "llvm/ADT/SetVector.h"
 #include <unordered_set>
 
+#ifndef _WIN32
+using llvm::make_unique;
```
llvm-beanz, line 41: This seems wrong and limiting. Requiring llvm::make_unique on non-Windows means that the intent of your change (to make libDXIL compatible with newer LLVM) is bound to Windows.

llvm-beanz (changes requested):
What needs to change to make use of std::make_unique on non-Windows platforms?

lib/DXIL/DxilModule.cpp
tex3d: Oh yeah, that's a good point.

The original reason for using llvm::make_unique to work around some issue with older versions of the CRT, right? 
 Shouldn't this be chosen on that basis instead?

Additionally, it could use an independent define that determines whether to use this particular workaround (such as `USE_LLVM_MAKE_UNIQUE`), which would be switched on based on whatever CRT/compiler testing logic is necessary in some header elsewhere (maybe `dxc/Support/Global.h`?) - removing that logic from this file.

tex3d: Looks like you addressed this by simply requiring c++14 instead, so ignore my comment above.

llvm-beanz: The issue is `std::make_unique` was introduced in c++ 14, and we build with c++ 11 by default. This only built with Visual Studio because of LLVM's build system not knowing how to properly specify the language version to MSVC.

I suggested to @python3kgae that we just bump to C++14, because I think that's pretty safe at this point. GCC 5.0 and Clang 3.4 and later have full C++14 support. I doubt anyone is using compilers older than that and our survey seems to indicate the same.

llvm-beanz:
`std::random_shuffle` was depreciated in C++14 and removed in C++17. You probably just need to migrate to `std::shuffle` in those cases. LLVM did that for C++14, then eventually migrated to its own shuffle implementation.

llvm-beanz (approved):
LGTM. Thank you for cleaning this up!

PR #4081, opened 2021-11-16
[NFC] HLSL doesn't really support digraphs

llvm-beanz:
Digraphs are support in C for non-ASCII conforming keyboards. Pretty
sure we don't really need that, and it doesn't work anyways.


pow2clk (approved):
Simple enough!

PR #4082, opened 2021-11-16
[HLSL21] Use C++ for loop scoping rules #584

### LINKED ISSUE #584: Spurious warnings about duplicated loop indices
The example below emits spurious redefinition warnings about the loop iterator.  It appears that permissive scoping rules are  in use.  This may be intentional, to match old compiler behavior, but if so, it might be better to hide it behind a compatibility switch, since C++ has moved past this.

```
uint g_count1;
uint g_count2;
Buffer<float4> Things;

float4 main() : SV_Target
{
	float4 data=0;
	for( uint i=0; i<g_count1; i++ )
	{
		data += Things[i];
	}

	for( uint i=0; i<g_count2; i++ )
	{
		data += Things[g_count1+i];
	}

	return data;
}
```

```
C:\scratch\hlsl.bin\Debug\bin>dxc -T ps_6_0 test.txt
test.txt:14:12: warning: redefinition of 'i' shadows declaration in the outer scope; most recent declaration will be used
        for( uint i=0; i<g_count2; i++ )
                  ^
test.txt:9:12: note: previous definition is here
        for( uint i=0; i<g_count1; i++ )
                  ^
```

llvm-beanz:
This should resolve #584 by adopting C++ for loop scoping rules for
HLSL 2021.

### DIFF
--- a/tools/clang/lib/Parse/ParseStmt.cpp
+++ b/tools/clang/lib/Parse/ParseStmt.cpp
@@ -1582,7 +1582,7 @@ StmtResult Parser::ParseForStatement(SourceLocation *TrailingElseLoc) {
     ScopeFlags = Scope::DeclScope | Scope::ControlScope;
 
   // HLSL Change Starts - leak declarations in for control parts into outer scope
-  if (getLangOpts().HLSLVersion < 2020) {
+  if (getLangOpts().HLSLVersion < 2021) {
     ScopeFlags = Scope::ForDeclScope;
   }
   // HLSL Change Ends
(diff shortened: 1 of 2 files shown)


tools/clang/lib/Parse/ParseStmt.cpp
```diff
@@ -1582,7 +1582,7 @@ StmtResult Parser::ParseForStatement(SourceLocation *TrailingElseLoc) {
     ScopeFlags = Scope::DeclScope | Scope::ControlScope;
 
   // HLSL Change Starts - leak declarations in for control parts into outer scope
-  if (getLangOpts().HLSL) {
+  if (getLangOpts().HLSLVersion < 2020) {
```
pow2clk, line 1585: It makes little difference to the end result, but referencing non-existent version 2020 makes me just a bit uncomfortable. The way these kind of checks are done elsewhere is of the form `HLSLVersion < 2021`.

The only reason I can think of is if someone wants to search for all the places that HLSLVersion 2021 changes the behavior and would miss this.

pow2clk (approved):
Nice to squeeze this in!

I have an NFC alternative suggestion

tools/clang/lib/Parse/ParseStmt.cpp
llvm-beanz: My brain is still slightly stuck in March 2020... I think today's date is 3/626/2020...

PR #4086, opened 2021-11-17
[SPIRV] Add support vk::ext_execution_mode

jiaolu:
[[vk::ext_storage_class(uint storage_class)]]
this is part of PRs for the
https://github.com/microsoft/DirectXShaderCompiler/issues/3919

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -7637,13 +7637,11 @@ SpirvEmitter::processIntrinsicCallExpr(const CallExpr *callExpr) {
   case hlsl::IntrinsicOp::IOP_VkReadClock:
     retVal = processIntrinsicReadClock(callExpr);
     break;
-<<<<<<< HEAD
   case hlsl::IntrinsicOp::IOP_VkRawBufferLoad:
     retVal = processRawBufferLoad(callExpr);
-=======
+    break;
   case hlsl::IntrinsicOp::IOP_Vkext_execution_mode:
     retVal = processIntrinsicExecutionMode(callExpr);
->>>>>>> daae8b9c0 ([SPIRV] Add support vk::ext_execution_mode)
     break;
   case hlsl::IntrinsicOp::IOP_saturate:
     retVal = processIntrinsicSaturate(callExpr);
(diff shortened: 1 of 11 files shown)


tools/clang/test/CodeGenSPIRV/spv.intrinsicStorageClass.hlsl
```diff
@@ -0,0 +1,13 @@
+// RUN: %dxc -T ps_6_0 -E main -spirv -Vd
+
+//CHECK: {{%\w+}} = OpVariable {{%\w+}} RayPayloadNV
+//CHECK: {{%\w+}} = OpVariable {{%\w+}} CrossWorkgroup
```
jaebaek, line 4: Could you please check the type of variable as well?
We want to make sure that `RayPayloadNV` is used for `payload` but it is not used for `foo` and vice versa. I am just worried in the future some developers can update the code and break your implementation.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -12539,6 +12545,29 @@ SpirvEmitter::processSpvIntrinsicCallExpr(const CallExpr *expr) {
   return retVal;
 }
 
+SpirvInstruction *
+SpirvEmitter::processIntrinsicExecutionMode(const CallExpr *expr) {
+  llvm::SmallVector<uint32_t, 2> execModesParams;
+  uint32_t exeMode = 0;
+  const auto args = expr->getArgs();
+  for (uint32_t i = 0; i < expr->getNumArgs(); ++i) {
```
jaebaek, line 12553: It seems that I missed the extra operands of `OpExecutionMode` when I wrote the spec for `vk::ext_execution_mode(uint execution_mode);`. I am sorry about that.

This code looks good but I am concerned about
```
    SpirvConstantInteger *argInst =
        dyn_cast<SpirvConstantInteger>(doExpr(args[i]));
```

because they is no guarantee that `args[i]` is an expression to generate a constant integer. In addition, I am not sure we conduct the constant folding in DXC (definitely, spirv-opt does ..).

Why don't we just check whether `args[i]` is a constant integer or not here?
If it is not a constant integer, we can just report error and returns this method with `nullptr`. Please see `emitError(..)` in this file for the error report.

jaebaek (commented):
I realize that the [spec](https://github.com/microsoft/DirectXShaderCompiler/issues/3919) I wrote about `vk::ext_execution_mode(uint execution_mode);` is wrong. I forgot to mention the extra literal parameters (Thank you for catching it!).

Your implementation looks fine. I will update the spec based on your implementation soon.

jaebaek (approved):
LGTM. Thanks again for your contribution!

PR #4087, opened 2021-11-17
Fix finding Dia SDK on non-VS generators

gongminmin:
On non-VS generator, e.g., Ninja generator, DIASDK_GUIDS_LIBRARY can't be found. To fix this, the VS2019/2022 code path is generalized to support other single configuration generators.


hekota (approved):
Thank you for the fix!

PR #4089, opened 2021-11-18
Propagate ParameterModifiers in templates

### LINKED ISSUE #4084: Template partial specialization trips up around `inout` qualifier
The following code fails to compile with `-HV 2021`:

```
template<typename T>
void apply(inout T Val) {
  Val += 1.0;
}

#define VECTOR_TYPE vector<float,3>
//#define VECTOR_TYPE float3

template<>
void apply<VECTOR_TYPE>(inout VECTOR_TYPE Val) {
  Val.x += 1;
}
```

If the template is specialized to `float3`, the error complains that `float3` can't be substituted for `vector<float,3>`, and if the specialization is `vector<float,3>`, it complains that `vector<float,3>` can't be substituted for `vector<float,3>` (no that isn't a typo).

See:

```
test.hlsl:11:6: error: no function template matches function template specialization 'apply'
void apply<VECTOR_TYPE>(inout VECTOR_TYPE Val) {
     ^
test.hlsl:3:6: note: candidate template ignored: could not match 'void (vector<float, 3> &__restrict)' against 'void (vector<float, 3> &__restrict)'
void apply(inout T Val) {
```

And
```
test.hlsl:11:6: error: no function template matches function template specialization 'apply'
void apply<VECTOR_TYPE>(inout VECTOR_TYPE Val) {
     ^
test.hlsl:3:6: note: candidate template ignored: could not match 'void (vector<float, 3> &__restrict)' against 'void (float3 &__restrict)'
void apply(inout T Val) {
```

llvm-beanz:
This addresses two issues that blocked template instantiation of
functions with ParameterModifiers.

The first issue is that template deduction was not propagating
ParameterModifiers, but since the modifiers do influence overload
resolution, matches were never being found.

The second issue is that template instantiation was not propagating
ParameterModifiers, so even if dedution managed to succeed, it would
produce an instantiaton that wouldn't match.

The silly issue that still remains after this is that our
ParameterModifiers sholud be attached to the QualType for the
parameters. We already bake in the effect of the ParameterModifier to
the QualType, so we were getting overload mismatches on identical
functions. If we move the ParameterModifiers to the QualTypes similar
to other type qualifiers, we can actually get rid of this patch.

This resolves #4084

### DIFF
--- a/tools/clang/include/clang/Sema/Sema.h
+++ b/tools/clang/include/clang/Sema/Sema.h
@@ -1254,10 +1254,12 @@ public:
   /// \returns A suitable function type, if there are no errors. The
   /// unqualified type will always be a FunctionProtoType.
   /// Otherwise, returns a NULL type.
-  QualType BuildFunctionType(QualType T,
-                             MutableArrayRef<QualType> ParamTypes,
+  // HLSL Change - FIX - We should move param mods to parameter QualTypes
+  QualType BuildFunctionType(QualType T, MutableArrayRef<QualType> ParamTypes,
                              SourceLocation Loc, DeclarationName Entity,
-                             const FunctionProtoType::ExtProtoInfo &EPI);
+                             const FunctionProtoType::ExtProtoInfo &EPI,
+                             ArrayRef<hlsl::ParameterModifier> ParamMods);
+  // HLSL Change - End
 
   QualType BuildMemberPointerType(QualType T, QualType Class,
                                   SourceLocation Loc,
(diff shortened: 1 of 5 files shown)


tex3d (approved):
I think this looks good, and your advice about moving the parameter modifiers sounds right.  I guess we can do that in a future change, since it seems like a larger amount of work.

llvm-beanz:
That was my thought. This is minimally invasive and should be good for the HLSL 2021 release.

PR #4092, opened 2021-11-18
PIX: Check for null HS entry point (pass-through case)

jeffnn:
This happens in the 11on12 case...

### DIFF
--- a/lib/DxilPIXPasses/PixPassHelpers.cpp
+++ b/lib/DxilPIXPasses/PixPassHelpers.cpp
@@ -203,10 +203,12 @@ std::vector<llvm::BasicBlock*> GetAllBlocks(hlsl::DxilModule& DM) {
     std::vector<llvm::BasicBlock*> ret;
     auto entryPoints = DM.GetExportedFunctions();
     for (auto& fn : entryPoints) {
-        auto & blocks = fn->getBasicBlockList();
-        for (auto & block : blocks) {
-            ret.push_back(&block);
+      if (fn != nullptr) {
+        auto& blocks = fn->getBasicBlockList();
+        for (auto& block : blocks) {
+          ret.push_back(&block);
         }
+      }
     }
     return ret;
 }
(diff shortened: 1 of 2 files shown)


lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -203,10 +203,12 @@ std::vector<llvm::BasicBlock*> GetAllBlocks(hlsl::DxilModule& DM) {
     std::vector<llvm::BasicBlock*> ret;
     auto entryPoints = DM.GetExportedFunctions();
```
tex3d, line 204: `GetExportedFunctions` doesn't do what it says.  It only returns entry point functions.
It appears that it is only used by two PIX passes.  And in each, I suspect you would want more than just the entry functions, you would want exports, but also non-inlined functions defined in the module.  In other words, I think what you want is all functions with bodies.  You can get this from the `llvm::Module` and filter by `!isDeclaration()`.  So I don't think we need the misleading DxilModule method.  But I could be misinterpreting what you want to do here.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -203,10 +203,12 @@ std::vector<llvm::BasicBlock*> GetAllBlocks(hlsl::DxilModule& DM) {
...
     for (auto& fn : entryPoints) {
-        auto & blocks = fn->getBasicBlockList();
-        for (auto & block : blocks) {
-            ret.push_back(&block);
+      if (fn != nullptr) {
```
tex3d, line 206: There are no null patch constant functions, this would be the null hull shader case (pass-through).  But see comment above about getting functions differently in the first place.

adam-yang (approved):
LGTM! Thanks!

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: That's good info- thanks Tex. I'll certainly do that in future since we need to be able to operate on all functions in libraries.

tex3d: You could probably make the change as part of this PR, unless you're really worried about making a more minimal patch and know libraries aren't used with this code.

I believe you can just replace `entryPoints` in the for loop with `DM.GetModule()->functions()` and add `if (fn.isDeclaration()) continue;` to skip declarations with no body.

tex3d: Note: The only other use of `GetExportedFunctions` is in `lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp`, and it seems like you would want the same change there to pick up all functions.

PR #4099, opened 2021-11-24
[HLSL2021] Fix template parameter for buffer types

llvm-beanz:
This is a bit of an oddball situation caused by a two factors.

(1) We verify that template parameters for buffer types are complete
(2) The template expansion for the HLSL buffer types doesn't actually
require instantiation of the parameter type

We could, skip verification of template parameters, but doing so causes
the backend to crash because we need the buffer type expanded for code
generation. Skipping verification might work in some common cases, but
will fail if buffers are unused, and if nothing else forces the
instantiation.

The safe approach is to force instantiation ourselves, and that's what
this change does. Basically if the parameter of a buffer type is a
template specialization, we force the specialization to instantiate by
creating the specialization in the AST.

This allows us to continue verifing that the types are complete,
because after instantiation the template type is complete.

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -4830,6 +4830,17 @@ public:
               << argType;
           return true;
         }
+        if (auto *TST = dyn_cast<TemplateSpecializationType>(argType)) {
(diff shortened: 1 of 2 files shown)


tex3d (approved):
LGTM!

PR #4100, opened 2021-11-24
[HLSL 2021] Allow dependent non-type parameters

llvm-beanz:
Non-type template parameters need more complicated constant resolution
because they need to look through layers of templates. This change
makes it so that when evaluating integer constants we use C++ 11
constant expression evaluation to look through template instantiations.


tex3d (approved):
LGTM!

PR #4101, opened 2021-11-24
Add additional template test case

llvm-beanz:
Test case authored by @tex3d!

This test case covers a whole bunch of edge conditions around using
templates with builtin types. Huge thank you to @tex3d for writing this
up and giving me a great basis to write fixes based on.

This test case covers a bunch of super common use cases for templates
interacting with builtin types.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/template/DependentWithBuiltinTemplate.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/template/DependentWithBuiltinTemplate.hlsl
@@ -1,0 +1,121 @@
+// RUN: %dxc -E main -T ps_6_0 -enable-templates %s | FileCheck -check-prefixes=CHECK,BUFFER %s
+// RUN: %dxc -E main -T ps_6_0 -HV 2021 %s | FileCheck -check-prefixes=CHECK,BUFFER %s
+// RUN: %dxc -E main -T ps_6_0 -HV 2021 -DERANGE1 %s | FileCheck %s -check-prefix=ERANGE1
+// RUN: %dxc -E main -T ps_6_0 -HV 2021 -DERANGE2 %s | FileCheck %s -check-prefix=ERANGE2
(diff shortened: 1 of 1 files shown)


tools/clang/test/HLSLFileCheck/hlsl/template/DependentWithBuiltinTemplate.hlsl
```diff
@@ -0,0 +1,121 @@
...
+// BUFFER: %CB0 = type { %"struct.Wrapper<Foo>" }
+// BUFFER: %"struct.Wrapper<Foo>" = type { %struct.Foo }
+
+// WRAPPER: %"hostlayout.$Globals" = type { %"hostlayout.struct.WrappedCB<Foo>" }
+// WRAPPER: %"hostlayout.struct.WrappedCB<Foo>" = type {}
+
+// CHECK: define void @main
+
+// BUFFER: !6 = !{i32 0, %CB0* undef, !"CB0", i32 0, i32 0, i32 1, i32 4, null}
+// WRAPPER: !6 = !{i32 0, %"hostlayout.$Globals"* undef, !"$Globals", i32 0, i32 0, i32 1, i32 0, null}
```
tex3d, line 21: Since `WrappedCB` has no data other than the `ConstantBuffer<>` inside, declaring a global of this type should not actually result in any `$Global` cbuffer use.  I think the fact that this is found is an indication of a bug.  As you can see from `%"hostlayout.struct.WrappedCB<Foo>" = type {}`, the buffer type is empty (as it should be), so ultimately, it should not have been included in the cbuffer resource list at all.  Instead, we should have a cbuffer type that indicates the layout of the ConstantBuffer inside the structure, which appears to be missing, if I'm reading this right.  This isn't necessarily something to fix right here, but the test case is checking for something that should not actually be there, and not checking for something that should, if we were to fix the bug.

I will check what the rest of the DXIL/metadata looks like for this case.

tex3d (approved):

PR #4105, opened 2021-11-26
[spirv] Use sequential container for ivars

### LINKED ISSUE #4088: [ SPIR-V ] Non-determinism in output for raytracing shaders
Using dxc June 2021 release (https://github.com/microsoft/DirectXShaderCompiler/releases/download/v1.6.2106/dxc_2021_07_01.zip) I am see a non-determinism in compiling raytracing shaders.  The non-determinism seems to be in the ordering of the OpEntryPoint.  For example:

```
<                OpEntryPoint AnyHitNV %main "main" %payload %gl_InstanceID %g_Instances %g_Materials
---
>                OpEntryPoint AnyHitNV %main "main" %g_Instances %payload %gl_InstanceID %g_Materials
```

This can also impact downstream SPIR-V instructions (although this particular shader doesn't show that because I stripped it down to share).  The following shader reproduces the issue:

```
struct Instance_t
{
	uint m_nTriangleGeometry;
	uint m_nMaterialIndex;
	uint m_nFlags; // FCACHETRI_...
	float4 m_vTintColor;
	row_major float3x4 m_mTransform;
	row_major float3x4 m_mTransformIT;
};

struct PayloadShadow_t
{
	float m_flVisibility;
};

struct GeometryInfo_t
{
	int m_nVertexBufferBindOffsetBytes;
	int m_nIndexBufferBindOffsetBytes;
	int m_nVertexStrideInBytes;
	int m_nIndexStrideInBytes;
	int m_nNormalOffset;
	int m_nNormalFormat;
	int m_nColorOffset;
	int m_nColorFormat;
	int m_nTexCoordOffset;
	int m_nTexCoordFormat;
};

struct Material_t
{
	float3 m_vAlbedo;
	float m_flOpacity;
	int m_nAlbedoTxRGB;
	int m_nOpacityTxA;
};

[[vk::binding( 1, 0 )]] StructuredBuffer< GeometryInfo_t > g_Geometries;
[[vk::binding( 2, 0 )]] ByteAddressBuffer g_Vertices[];
[[vk::binding( 3, 0 )]] ByteAddressBuffer g_Indices[];
[[vk::binding( 6, 0 )]] StructuredBuffer< Material_t > g_Materials;
[[vk::binding( 7, 0 )]] StructuredBuffer< Instance_t > g_Instances;


[shader("anyhit")]
void main( inout PayloadShadow_t payload, in BuiltInTriangleIntersectionAttributes attrs )
{
	Instance_t instance = g_Instances[InstanceIndex()];
	Material_t material = g_Materials[ instance.m_nMaterialIndex ];

	uint nGeometry = instance.m_nTriangleGeometry;
	GeometryInfo_t g = g_Geometries[NonUniformResourceIndex(nGeometry)];

	float flOpacity = material.m_flOpacity;

	if ( material.m_nOpacityTxA != -1 )
	{
		flOpacity = 0.5;
	}

	payload.m_flVisibility *= max( 0.0, 1.0 - flOpacity );

	if ( payload.m_flVisibility == 0.0 )
	{
		AcceptHitAndEndSearch();
	}
	else
	{
		IgnoreHit();
	}
}

```

I run:

```
H:\dev\source2\main\src\utils\vrad3\shaders>..\..\..\thirdparty\dxc\bin\x64\dxc.exe -T lib_6_4 generic_shadow.rahit.hlsl -spirv -fspv-target-env=vulkan1.2 -fvk-use-scalar-layout
```

Each time the output can differ in the way shown above.

sudonatalie:
To ensure a deterministic order of the interface variable in
OpEntryPoint, use a SetVector to hold them.

Fixes #4088


sudonatalie:
I tested this manually with the issue shader provided in #4088. Before this change, I saw the non-deterministic ordering after only 3 runs. After this change, I did 50 runs without seeing any change. (Still not a complete guarantee, but this is certainly one potential source of non-determinism that we should remove.)

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -622,7 +623,7 @@ SpirvEmitter::getInterfacesForEntryPoint(SpirvFunction *entryPoint) {
   // declIdMapper keeps the mapping between variables with Input or Output
   // storage class and their storage class, we have to rely on
   // declIdMapper.collectStageVars() to collect them.
-  llvm::DenseSet<SpirvVariable *> interfaces;
+  llvm::SetVector<SpirvVariable *> interfaces;
```
jaebaek, line 626: nit: do you think `SmallSetVector` would have benefits in terms of the memory usage?

jaebaek (commented):
LGTM. Thanks!

tools/clang/lib/SPIRV/SpirvEmitter.cpp
sudonatalie: `SmallSetVector` is a "`SetVector` that performs no allocations if smaller than a certain size." but that has to be a const size (since it's part of the type). I wouldn't want to guess at what an appropriate number would be there to ensure it's actually an optimization without data to back it up.

I did switch to initializing with the `stageVars` iterator though, I missed that that was a constructor before.

jaebaek (approved):
LGTM. Thanks!

PR #4107, opened 2021-11-29
Update version to 1.6.2112

hekota:


pow2clk (approved):
2112 is better than 2110. That's just math!

PR #4110, opened 2021-11-29
Missed a bit of generated code

llvm-beanz:
This code should be part of the generated file not the hard coded bits.


hekota (approved):
Thanks! :)

PR #4111, opened 2021-11-30
All source files in PDBs interpreted as UTF8

adam-yang:


tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -1139,12 +1139,24 @@ static void VerifyPdbUtil(dxc::DxcDllSupport &dllSupport,
       VERIFY_SUCCEEDED(pPdbUtils->GetSourceName(i, &pFileName));
       VERIFY_SUCCEEDED(pPdbUtils->GetSource(i, &pFileContent));
       if (0 == wcscmp(pFileName, pMainFileName)) {
-        VERIFY_IS_TRUE(pFileContent->GetBufferSize() == MainSource.size());
-        VERIFY_IS_TRUE(0 == std::memcmp(pFileContent->GetBufferPointer(), MainSource.data(), MainSource.size()));
+        llvm::StringRef FileContentRef((const char *)pFileContent->GetBufferPointer(), pFileContent->GetBufferSize());
```
tex3d, line 1142: We should probably just QI for IDxcBlobUtf8 (verify that succeeds) and then use GetStringPointer() and GetStringLength() here (which will exclude the null terminator).

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -1139,12 +1139,24 @@ static void VerifyPdbUtil(dxc::DxcDllSupport &dllSupport,
...
       }
       else {
         VERIFY_IS_TRUE(0 == std::memcmp(pFileContent->GetBufferPointer(), IncludedFile.data(), IncludedFile.size()));
       }
+
+      // Make sure encoding is what we expect
+      {
+        BOOL bEncodingKnown = false;
+        UINT32 CodePage = CP_ACP;
+        VERIFY_SUCCEEDED(pFileContent->GetEncoding(&bEncodingKnown, &CodePage));
```
tex3d, line 1156: If QI for IDxcBlobUtf8 succeeded above, this is unnecessary.  If it failed, then we have a bug.

tex3d (commented):
Looks good, but I think the test could just QI for IDxcBlobUtf8 instead, as we intend that to work.

tex3d (approved):

PR #4112, opened 2021-11-30
Make `this` more consistently a reference type

### LINKED ISSUE #3732: Template type on struct breaks `this` pointer (hlsl-2021)
```
template<typename T>
struct SimpleBuffer {
    uint handle;

    T Load() {
        ByteAddressBuffer buf = ResourceDescriptorHeap[this.handle];
        return (T)0;
    }
};

[numthreads(1,1,1)]
void main() {
    SimpleBuffer<uint> b;
    b.Load();
}
```

Regular structs with `this` in their methods (with access through the `.` operator) do work, however, once a template argument is introduced DXC yields this error message:

`dxc -Tcs_6_6 -Emain -enable-templates test_bindless.hlsl`

```
test_bindless.hlsl:6:60: error: member reference type 'SimpleBuffer<unsigned int> *' is a pointer; did you mean to use '->'?
        ByteAddressBuffer buf = ResourceDescriptorHeap[this.handle];
                                                       ~~~~^
                                                           ->
test_bindless.hlsl:14:7: note: in instantiation of member function 'SimpleBuffer<unsigned int>::Load' requested here
    b.Load();
```

llvm-beanz:
Prior to this change explicit `this` access inside a struct was treated
as a reference, but implicit `this` was still a pointer type. With this
change `this` is always a reference type.

This has no impact on code generation since references and pointers all
digress to addresses, but it does make the AST more accurately reflect
the types and it fixes #3732.

### DIFF
--- a/tools/clang/lib/Sema/SemaExprMember.cpp
+++ b/tools/clang/lib/Sema/SemaExprMember.cpp
@@ -1049,7 +1049,13 @@ Sema::BuildMemberReferenceExpr(Expr *BaseExpr, QualType BaseExprType,
     if (SS.getRange().isValid())
       Loc = SS.getRange().getBegin();
     CheckCXXThisCapture(Loc);
-    BaseExpr = new (Context) CXXThisExpr(Loc, BaseExprType,/*isImplicit=*/true);
+
+    // HLSL Change Starts - adjust this from T* to T&-like
+    if (getLangOpts().HLSL && BaseExprType->isPointerType())
+      BaseExpr = genereateHLSLThis(Loc, BaseExprType, /*isImplicit=*/true);
+    else
+      BaseExpr = new (Context) CXXThisExpr(Loc, BaseExprType,/*isImplicit=*/true);
+    // HLSL Change Ends
   }
 
   bool ShouldCheckUse = true;
@@ -1762,12 +1768,17 @@ Sema::BuildImplicitMemberExpr(const CXXScopeSpec &SS,
     if (SS.getRange().isValid())
       Loc = SS.getRange().getBegin();
     CheckCXXThisCapture(Loc);
-    baseExpr = new (Context) CXXThisExpr(loc, ThisTy, /*isImplicit=*/true);
+    if (getLangOpts().HLSL && ThisTy->isPointerType()) {
+      baseExpr = genereateHLSLThis(Loc, ThisTy, /*isImplicit=*/true);
+      ThisTy = ThisTy->getAs<PointerType>()->getPointeeType();
+    } else
+      baseExpr = new (Context) CXXThisExpr(loc, ThisTy, /*isImplicit=*/true);
   }
 
   return BuildMemberReferenceExpr(baseExpr, ThisTy,
                                   /*OpLoc*/ SourceLocation(),
-                                  /*IsArrow*/ true,
+                                  // HLSL Change - this is a reference
+                                  /*IsArrow*/ !getLangOpts().HLSL,
                                   SS, TemplateKWLoc,
                                   /*FirstQualifierInScope*/ nullptr,
                                   R, TemplateArgs);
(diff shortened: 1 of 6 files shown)


llvm-beanz:
This ended up not being as bad as I thought it was going to be.

tools/clang/lib/Sema/SemaExprMember.cpp
```diff
@@ -1762,12 +1768,17 @@ Sema::BuildImplicitMemberExpr(const CXXScopeSpec &SS,
     if (SS.getRange().isValid())
       Loc = SS.getRange().getBegin();
     CheckCXXThisCapture(Loc);
-    baseExpr = new (Context) CXXThisExpr(loc, ThisTy, /*isImplicit=*/true);
+    if (getLangOpts().HLSL && ThisTy->isPointerType()) {
+      baseExpr = genereateHLSLThis(Loc, ThisTy, /*isImplicit=*/true);
+      ThisTy = ThisTy->getAs<PointerType>()->getPointeeType();
+    } else
+      baseExpr = new (Context) CXXThisExpr(loc, ThisTy, /*isImplicit=*/true);
```
pow2clk, line 1775: This leaves just a couple places where we create a new `CXXThisExpr` without calling `generateHLSLThis()` when the language is HLSL and it's a pointer. Should we do it everywhere?

tools/clang/test/HLSLFileCheck/hlsl/template/ThisAccess.hlsl
```diff
@@ -0,0 +1,48 @@
...
+// CHECK-NEXT: `-ReturnStmt 0x{{[0-9a-zA-Z]+}} <line:10:5, col:17>
+// CHECK-NEXT: `-ImplicitCastExpr 0x{{[0-9a-zA-Z]+}} <col:12, col:17> 'int':'int' <LValueToRValue>
+// CHECK-NEXT: `-MemberExpr 0x{{[0-9a-zA-Z]+}} <col:12, col:17> 'int':'int' lvalue .First 0x{{[0-9a-zA-Z]+}}
+// CHECK-NEXT: `-CXXThisExpr 0x{{[0-9a-zA-Z]+}} <col:12> 'pair<int, float>' lvalue this
+// CHECK-NEXT: `-CXXMethodDecl 0x{{[0-9a-zA-Z]+}} <line:13:3, line:15:3> line:13:5 used second 'float ()'
+// CHECK-NEXT: `-CompoundStmt 0x{{[0-9a-zA-Z]+}} <col:14, line:15:3>
+// CHECK-NEXT: `-ReturnStmt 0x{{[0-9a-zA-Z]+}} <line:14:5, col:12>
+// CHECK-NEXT: `-ImplicitCastExpr 0x{{[0-9a-zA-Z]+}} <col:12> 'float':'float' <LValueToRValue>
+// CHECK-NEXT: `-MemberExpr 0x{{[0-9a-zA-Z]+}} <col:12> 'float':'float' lvalue .Second 0x{{[0-9a-zA-Z]+}}
+// CHECK-NEXT: `-CXXThisExpr 0x{{[0-9a-zA-Z]+}} <col:12> 'pair<int, float>' lvalue this
```
pow2clk, line 48: In addition to these, I think we should have a test that produces an error similar to that reported in #3732. However, I can't reproduce that exact error even without your change 😕

tools/clang/lib/Sema/SemaExprMember.cpp
llvm-beanz: I was hesitant to change the code in code paths that I wasn't sure were reachable. Like lambda captures. I don't particularly like making changes that I can't write tests for.

pow2clk: That's reasonable. The only one that I found not involved in lambdas is this: https://github.com/microsoft/DirectXShaderCompiler/blob/39dd31f3552632de480fcc575de452f673915737/tools/clang/lib/Sema/SemaExpr.cpp#L1900

Not sure exactly how to get there either though and it is for a failure diagnostic, so it's pretty minor.

tex3d (approved):
As far as I understand it, I think this change looks good.

tools/clang/test/HLSLFileCheck/hlsl/classes/ThisReference.hlsl
```diff
@@ -0,0 +1,42 @@
+// RUN: %dxc -T cs_6_6 -E main -ast-dump %s | FileCheck %s
+// RUN: %dxc -T cs_6_6 -E main -DERROR_CASE 2>&1 %s | FileCheck -check-prefix=FAIL %s
```
pow2clk, line 2: This could also be accomplished with `FileCheck -input-file=stderr`, but perhaps that's something we want to get away from to better integrate with the LLVM style testing.

pow2clk (approved):

Issue #3556, opened 2021-03-09
[templates] Issues with dependent template names


pow2clk:
DXC has some issues with dependent template names. Sometimes dependent template names are inevitable to leverage the usage of templates.
 
The following example reports the correct error that is also reported by Clang in C++:
```hlsl
template <typename T>
struct Vec2 {
    T x, y;
    T Length() {
        return sqrt(x*x + y*y);
    }
    typedef T value_type;
};
template <typename TVec>
TVec::value_type Length(TVec v) { // <-- Must be "typename TVec::value_type"  
    return v.Length();
}
void VSMain() {
    Vec2<int> v_i;
    Vec2<float> v_f;
   
    int   a = Length(v_i);
    float b = Length(v_f);
}
```
The error would be this one:
```
.\TemplateTypename.hlsl:12:1: error: missing 'typename' prior to dependent type name 'TVec::value_type'
TVec::value_type Length(TVec v) {
^~~~~~~~~~~~~~~~
typename
```
 
If I add the missing typename keyword to forward declare TVec::value_type to be a type name, as shown here:
```hlsl
template <typename TVec>
typename TVec::value_type Length(TVec v) {
    return v.Length();
}
```
 
the following compile error is produced:
```
.\TemplateTypename.hlsl:12:16: error: use 'template' keyword to treat 'value_type' as a dependent template name
typename TVec::value_type Length(TVec v) {
               ^
               template
.\TemplateTypename.hlsl:12:39: error: expected ')'
typename TVec::value_type Length(TVec v) {
                                      ^
.\TemplateTypename.hlsl:12:33: note: to match this '('
typename TVec::value_type Length(TVec v) {
                                ^
.\TemplateTypename.hlsl:12:34: warning: variable templates are a C++14 extension [-Wc++14-extensions]
typename TVec::value_type Length(TVec v) {
                                 ^
.\TemplateTypename.hlsl:12:42: warning: effect state block ignored - effect syntax is deprecated. To use braces as an initializer use them with equal signs. [-Weffects-syntax]
typename TVec::value_type Length(TVec v) {
                                         ^
.\TemplateTypename.hlsl:14:2: error: expected ';' at end of declaration
}
 ^
 ;
.\TemplateTypename.hlsl:20:13: error: use of undeclared identifier 'Length'
    int a = Length(v_i);
            ^
.\TemplateTypename.hlsl:21:15: error: use of undeclared identifier 'Length'
    float b = Length(v_f);
[...]

PR #4114, opened 2021-11-30
[HLSL21] Allow HLSLDecls in dependent DeclContexts

### LINKED ISSUE #3731: ResourceDescriptorHeap doesn't work properly with templates (hlsl-2021)
```
#ifdef FAILURE_CASE
struct SimpleBuffer {
    uint handle;

    template<typename T>
    T Load() {
        ByteAddressBuffer buf = ResourceDescriptorHeap[this.handle];
        return buf.Load<T>(0);
    }
};
#endif

[numthreads(1,1,1)]
void main() {
#ifdef FAILURE_CASE
    SimpleBuffer b = (SimpleBuffer)0;
    b.Load<uint>();
#endif

    ByteAddressBuffer buf = ResourceDescriptorHeap[0];
    buf.Load<uint>(0);
}
```

Works: 
`dxc -Tcs_6_6 -Emain -enable-templates test_bindless.hlsl`

Doesn't work:
`dxc -Tcs_6_6 -Emain -enable-templates -DFAILURE_CASE test_bindless.hlsl`

Errors generated:

```
test_bindless.hlsl:8:20: error: no member named 'Load' in 'ByteAddressBuffer'
        return buf.Load<T>(0);
               ~~~ ^
test_bindless.hlsl:8:25: error: 'T' does not refer to a value
        return buf.Load<T>(0);
                        ^
test_bindless.hlsl:5:23: note: declared here
    template<typename T>
                      ^
test_bindless.hlsl:21:5: warning: ignoring return value of function that only reads data [-Wunused-value]
    buf.Load<uint>(0);
    ^~~~~~~~~~~~~~ ~
```

Notice, this doesn't just affect ByteAddressBuffer but all resource types it seems (StructuredBuffer, Texture etc).

danbrown-amd:
Fixes #3731.

- Error for dependent context decl (template decl) now only occurs when templates are not enabled
- Amended faulty test cases revealed by bug fix (tools\clang\test\HLSLFileCheck\hlsl\payload_qualifier\combination.hlsl, tools\clang\test\HLSLFileCheck\hlsl\template\InstantiateHLSLAttributes.hlsl, tools\clang\test\HLSLFileCheck\hlsl\template\missing_typename.hlsl)
- Added test case to reflect bug fix
(tools\clang\test\HLSLFileCheck\hlsl\template\3731-resource-descriptor-heap.hlsl)

Attribution:
Chris Bieneman
Timothy Corringham
Samantha Del Rosario


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12426,11 +12426,11 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
          "otherwise this is called without checking language first");
 
   // NOTE: some tests may declare templates.
-  if (DC->isDependentContext()) return true;
+  if (DC->isDependentContext() && !getLangOpts().EnableTemplates) return true;
 
   DeclSpec::SCS storage = D.getDeclSpec().getStorageClassSpec();
   assert(!DC->isClosure() && "otherwise parser accepted closure syntax instead of failing with a syntax error");
-  assert(!DC->isDependentContext() && "otherwise parser accepted a template instead of failing with a syntax error");
+  assert((!DC->isDependentContext() || getLangOpts().EnableTemplates) && "otherwise parser accepted a template instead of failing with a syntax error");
```
llvm-beanz, line 12433: We should just delete this assert. It is impossible to hit because of the early return, and it is meaningless.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12426,11 +12426,11 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
          "otherwise this is called without checking language first");
 
   // NOTE: some tests may declare templates.
-  if (DC->isDependentContext()) return true;
+  if (DC->isDependentContext() && !getLangOpts().EnableTemplates) return true;
```
llvm-beanz, line 12429: I think in this case we should return `false` not `true`. I know that's not how the code originally was, but we should be surfacing an error if template dependent contexts are generated without templates being enabled.

tools/clang/test/HLSLFileCheck/hlsl/template/missing_typename.hlsl
```diff
@@ -7,12 +7,12 @@
 // PRESENT: define void @VSMain() {
 
 template <typename T>
+typedef T value_type;
 struct Vec2 {
     T x, y;
     T Length() {
         return sqrt(x*x + y*y);
     }
-    typedef T value_type;
```
llvm-beanz, line 15: Did you mean to change this? I don't think this test case should be getting changed. If something in your change requires this change to be made, that is concerning.

tools/clang/lib/Sema/SemaTemplate.cpp
```diff
@@ -2880,7 +2880,7 @@ ExprResult Sema::BuildTemplateIdExpr(const CXXScopeSpec &SS,
   //       vs template<class T, class U> void f(U);
 
   // These should be filtered out by our callers.
-  assert(!R.empty() && "empty lookup results when building templateid");
+  assert((!R.empty() || RequiresADL) && "empty lookup results when building templateid");
```
llvm-beanz, line 2883: I don't think you should need this change.

llvm-beanz (commented):
Overall mostly looks good. We also need a test case for the bug being resolved.

tools/clang/test/HLSLFileCheck/hlsl/template/missing_typename.hlsl
sam-sam98: The test case did need changes; however, it looks like I've included the wrong version in the patch - this will be updated soon. In the meantime the correct text can be found in the [attached text file](https://github.com/microsoft/DirectXShaderCompiler/files/7629930/missing_typename.txt).

As for why it needed to be changed, the original test was not valid and was only passing because of the very issue being dealt with here. The example code in the original issue would fail because there was an early exit in the parsing of the struct with templates enabled - as a result, not all contents of the struct would be processed, and so trying to access them later in the code would cause an error. On the flip side, for the above test case, the early exit actually prevented the compiler from catching an error (nested typedef declarations are illegal) so the test had to be modified to be able to test the same concept without the nested typedef. Once again, thought, this is the wrong version of the test case, and the correct one will be added soon. @tcorringham can better explain the amended test case if need be.

tools/clang/lib/Sema/SemaTemplate.cpp
sam-sam98: Omitting this change results in an internal assert error when compiling the code provided in the original issue. Tracing the stack shows that all asserts encountered prior to this one reflect the structure of ```assert((!R.empty() || RequiresADL) && ... );``` and the assert shown here was failing because it receives ```R.empty() == true && RequiresADL == true```, which was accepted by the previous asserts.

tools/clang/test/HLSLFileCheck/hlsl/template/missing_typename.hlsl
llvm-beanz: I'd really like to understand why you needed changes in this test file. That test file's code is valid C++, and the attached text file is not at all equivalent.

sam-sam98: This set of test cases is written in HLSL, which does not have all of the support and capabilities of C++ just yet. From my understanding, nested typedefs are excluded from the current version of the compiler by design, as it has been explicitly enumerated as an error. As such it is fairly certain that the original test case should not have been usable and requires changes.

I'll defer to Tim for any explanation on the equivalency of the tests - he authored both the original and amended versions of this test case, so he has a better understanding of what particular scenario it is meant to verify.

tcorringham: Just some background to this test case. I added this test as part of a fix for #3556 - where a cascade of unhelpful errors were produced rather than the expected diagnostic as in C++.
The test was taken from the example C++ code in the issue report - and was intended only to demonstrate the different behaviour before and after the fix for #3556. The use of the nested typedef is not supported in HLSL, although previously no diagnostic was produced in this case.
The fix for #3731 does elicit an error for the nested typedef - I suggested removing the typedef to preserve this test as a valid test for the #3556 fix, as it is incidental to that. The amended test still demonstrates the expected before/after differences.
So I suggest that this amended test is valid - although it could be argued that the lack of a diagnostic for the nested typdef previously was a bug.

llvm-beanz: Before this change, the test case is valid C++, that should produce an error of some kind (useful or otherwise) in HLSL. With this change or the proposed change in the text file, the test case is invalid code in C++ and HLSL. IMO, that reduces the usefulness of the test case.

Clearly the test is wrong, in that it had a success case where it allowed the code to compile, (no arguments fixing that), and maybe the test case produces a different error (useful or otherwise, I'm okay with that). But changing the valid C++ to invalid C++ to produce the same error and allow strange code to compile seems like the wrong change to the test case.

llvm-beanz: So, I ran the test locally (unmodified), and can confirm that the different error message produced is actually the error message we want to produce (HLSL doesn't support nested typedefs). I think the only problem with this test case is that it checks for the wrong result.

tools/clang/lib/Sema/SemaTemplate.cpp
llvm-beanz: I don't locally hit this assert with the code for the original issue or in any of our test cases.

llvm-beanz (commented):
This change actually fixes several latent bugs (Thanks!).

I've put updated test cases (and the minimal fix) into a branch on my fork here:
https://github.com/llvm-beanz/DirectXShaderCompiler/tree/cbieneman/3731

Feel free to grab the new test case and existing test case updates from there.

tools/clang/lib/Sema/SemaTemplate.cpp
llvm-beanz: Just as a note: the reason I'm concerned about this change is because this isn't an assert written by the HLSL compiler team, it is an assert inherited from Clang (I think authored by John McCall). If we're hitting this assert as written, we're probably doing something else wrong and in for a world of pain.

sam-sam98: Ah I see! Well it was necessary when I initially wrote the bug fix, but it is possible that something else has been changed since then that resolves the issue otherwise. Especially if you no longer encounter the problem with the original code from the issue.

llvm-beanz:
I updated the title of the PR to reflect the actual change being made. @danbrown-amd, can you please update the PR description to provide more context. PR descriptions end up in the final git commit, and we like to have meaningful descriptions that show up in the history.

llvm-beanz (approved):
Looks good to me. Thank you!

llvm-beanz:
Just please update the description before merging so that we have a useful commit message.

danbrown-amd:
[AMD Official Use Only]

Hope I did that in the right place.  If it fixes other issues, could you add those?

-- Dan.

llvm-beanz:
@danbrown-amd, I think this change is good to go. There seems to have been some sort of hiccough with the automation not reporting that it completed successfully. If you're fine with it we can override and force the merge through.

Issue #1234, opened 2018-04-16
[SPIR-V] Global array with initializer list does not appear in SPIR-V code


Cry-Mory:
I have a test shader code which contains a global array with an initializer list. It reads the vertex id from the input stream and reads the data from this array. But it seems the array values are completely ignored in SPIR-V code. As a test shader you can use [shader.hlsl.txt](https://github.com/Microsoft/DirectXShaderCompiler/files/1916920/shader.hlsl.txt) which can be compiled with the following command:

`.\dxc.exe -spirv -Od -Zpr .\shader.hlsl -Fo .\shader.spv -Fc .\shader.h_spv -T vs_5_0 -E TestVS`

The human-readable output is [shader.h_spv.txt](https://github.com/Microsoft/DirectXShaderCompiler/files/1916931/shader.h_spv.txt).

Update 1: I tried to generate DXIL code and it seems, the directx compiler is also ignoring the initializer list ([shader.dxil.txt](https://github.com/Microsoft/DirectXShaderCompiler/files/1916935/shader.dxil.txt)) and not SPIR-V related.

Update 2: At least the SPIR-V Compiler should give a warning or an error.

antiagainst:
External-visible stand-alone global variables are collected into the `$Globals` cbuffer, which is translated into a Vulkan Uniform Buffer. IIRC, Vulkan currently doesn't support populating uniform buffer's data inside a shader. As a workaround for the above specific case, you can declare `inputs` as `static` to retain the data in the shader.

But yes, we should emit an warning. Actually we are already doing that for normal cbuffer field initializers.

antiagainst:
But for the `$Globals` cbuffer, we are not doing that yet. Will add.

Cry-Mory:
Thanks for the hint! Using `static` identifier for the global array fixed the problem.

antiagainst:
If you want to see the "raw" SPIR-V generated, you can specify `-fcgl`. `-Od` only turns off optimization, not legalization. To learn more about the differences, see [the doc here](https://github.com/Microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#legalization-optimization-validation).

I did a quick check with `-fcgl`, the arrays are there.  I think they are just transformed away during legalization so that we are using the array elements directly in the generated SPIR-V code? But I haven't followed the logic closely. Please definitely report if you think the transformations are incorrect.

Cry-Mory:
[...]

PR #4116, opened 2021-12-01
[Doc] Updating guidelines for Pull Requests

llvm-beanz:
This more clearly codifies and explains the guidelines we want for PR
titles and descriptions, and how they relate to final git commit
messages.

This isn't really a change from existing policy, but rather a
codification of the current expectations so that we have something clear
to point to.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ Summary of the changes (Less than 80 chars)
  - Detail 1
  - Detail 2
 
-Addresses #bugnumber (in this specific format)
+Fixes #bugnumber (in this specific format)
 ```
 
 Your pull request should:
@@ -59,6 +59,8 @@ Your pull request should:
 * Include a description of what your change intends to do
   * The title of your PR should be a very brief description of the change, and
     can use tags to allow for speedy categorization.
+    * Titles under 76 characters print nicely in unix terminals under `git log`.
+      This is not a hard requirement, but is good guidance.
     * Title tags are generally one word or acronym enclosed in square brackets.
       Limiting to one or two tags is ideal to keep titles short. Some examples
       of common tags are:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -57,11 +57,30 @@ Addresses #bugnumber (in this specific format)
 Your pull request should:
 
 * Include a description of what your change intends to do
+  * The title of your PR should be a very brief description of the change, and
```
pow2clk, line 60: I would like to encourage an 80 character limit without being too insistent on it. If it's 90 characters, it's not worth rejecting a change over, if it's 120, we start to have a problem . . .

pow2clk: I see that a few lines up it says 80 characters, but perhaps we can add the nuances I suggested here since it is a rule oft ignored.

CONTRIBUTING.md
```diff
@@ -57,11 +57,30 @@ Addresses #bugnumber (in this specific format)
...
+      * [SPIRV] - Changes related to SPIR-V
+      * [HLSL2021] - Changes related to HLSL 2021 features
+      * Other tags in use: [Linux], [mac], [Win], [PIX], etc...
+      * Tags aren't formalized or any specific limited set. If you're unsure of
+        a reasonable tag to use, just don't use any. If you want to invent a new
+        tag, go for it! These are to help categorize changes at a glance.
+  * The PR description should include a more detailed description of the change,
+    an explanation for the motivation of the change, and links to any relevant
+    Issues. This does not need to be a dissertation, but should leave
+    breadcrumbs for the next person debugging your code (who might be you).
```
pow2clk, line 77: Also mentioned in the template above, but I can't comment that far outside of a change. 

I like how if you inclide `fixes #1234` then it will close that issue. I don't think `addresses #1234` does anything except link to the issue, which is good. Do we want to encourage that in cases where it is relevant?

pow2clk (approved):
Well I'm happy anyway!

pow2clk:
Dunno what's going on with automated testing, but it's obviously not necessary here anyway

PR #4117, opened 2021-12-02
Unions support in HLSL.

danbrown-amd:
Unions support in HLSL.

### DIFF
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -1137,6 +1137,9 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(DxilStructAnnotation *annota
         fieldAnnotation.SetPrecise();
     }
 
+    if (RD->isUnion())
+      break;
+
     // Update offset.
     CBufferSize = std::max(CBufferSize, CBufferOffset + size);
     CBufferOffset = CBufferSize;
@@ -1277,6 +1280,17 @@ unsigned CGMSHLSLRuntime::AddTypeAnnotation(QualType Ty,
     unsigned size = ConstructStructAnnotation(annotation, payloadAnnotation, RD, dxilTypeSys);
     // Resources don't count towards cbuffer size.
     return IsHLSLResourceType(Ty) ? 0 : size;
+  } else if (const RecordType *RT = paramTy->getAsUnionType()) {
+    RecordDecl *RD = RT->getDecl();
+    llvm::StructType *ST = CGM.getTypes().ConvertRecordDeclType(RD);
+    DxilStructAnnotation *annotation = dxilTypeSys.AddStructAnnotation(
+        ST, GetNumTemplateArgsForRecordDecl(RT->getDecl()), true);
+    DxilPayloadAnnotation *payloadAnnotation = nullptr;
+    if (ValidatePayloadDecl(RT->getDecl(), *m_pHLModule->GetShaderModel(),
+                            CGM.getDiags(), CGM.getCodeGenOpts()))
+      payloadAnnotation = dxilTypeSys.AddPayloadAnnotation(ST);
+    return ConstructStructAnnotation(annotation, payloadAnnotation, RD,
+                                     dxilTypeSys);
   } else if (const RecordType *RT = dyn_cast<RecordType>(paramTy)) {
     // For this pointer.
     RecordDecl *RD = RT->getDecl();
--- a/tools/clang/lib/Format/TokenAnnotator.cpp
+++ b/tools/clang/lib/Format/TokenAnnotator.cpp
@@ -1667,7 +1667,7 @@ unsigned TokenAnnotator::splitPenalty(const AnnotatedLine &Line,
   if (Left.is(tok::coloncolon) ||
       (Right.is(tok::period) && Style.Language == FormatStyle::LK_Proto))
     return 500;
-  if (Left.isOneOf(tok::kw_class, tok::kw_struct))
+  if (Left.isOneOf(tok::kw_class, tok::kw_struct, tok::kw_union))
     return 5000;
 
   if (Left.isOneOf(TT_RangeBasedForLoopColon, TT_InheritanceColon))
@@ -2260,7 +2260,7 @@ bool TokenAnnotator::canBreakBefore(const AnnotatedLine &Line,
        Left.getPrecedence() == prec::Assignment))
     return true;
   return Left.isOneOf(tok::comma, tok::coloncolon, tok::semi, tok::l_brace,
-                      tok::kw_class, tok::kw_struct) ||
+                      tok::kw_class, tok::kw_struct, tok::kw_union) ||
          Right.isMemberAccess() ||
          Right.isOneOf(TT_TrailingReturnArrow, TT_LambdaArrow, tok::lessless,
                        tok::colon, tok::l_square, tok::at) ||
--- a/tools/clang/lib/Format/UnwrappedLineParser.cpp
+++ b/tools/clang/lib/Format/UnwrappedLineParser.cpp
@@ -1525,7 +1525,8 @@ void UnwrappedLineParser::parseEnum() {
     nextToken();
 
   // Eat up enum class ...
-  if (FormatTok->Tok.is(tok::kw_class) || FormatTok->Tok.is(tok::kw_struct))
+  if (FormatTok->Tok.is(tok::kw_class) || FormatTok->Tok.is(tok::kw_struct) ||
+      FormatTok->Tok.is(tok::kw_union))
     nextToken();
 
   while (FormatTok->Tok.getIdentifierInfo() ||
--- a/tools/clang/lib/Parse/ParseDeclCXX.cpp
+++ b/tools/clang/lib/Parse/ParseDeclCXX.cpp
@@ -1580,7 +1580,9 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
 
   if (!Name && !TemplateId && (DS.getTypeSpecType() == DeclSpec::TST_error ||
                                TUK != Sema::TUK_Definition)) {
-    if (DS.getTypeSpecType() != DeclSpec::TST_error) {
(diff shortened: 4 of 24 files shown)


danbrown-amd:
Windows Debug failed but Release succeeded.  Can someone trigger a retry?

tools/clang/lib/Format/TokenAnnotator.cpp
```diff
@@ -1667,7 +1667,7 @@ unsigned TokenAnnotator::splitPenalty(const AnnotatedLine &Line,
   if (Left.is(tok::coloncolon) ||
       (Right.is(tok::period) && Style.Language == FormatStyle::LK_Proto))
     return 500;
-  if (Left.isOneOf(tok::kw_class, tok::kw_struct))
+  if (Left.isOneOf(tok::kw_class, tok::kw_struct, tok::kw_union))
```
llvm-beanz, line 1670: This change isn't required for C++ to support unions, so I don't understand why it is required for HLSL.

tools/clang/lib/Format/TokenAnnotator.cpp
```diff
@@ -2260,7 +2260,7 @@ bool TokenAnnotator::canBreakBefore(const AnnotatedLine &Line,
        Left.getPrecedence() == prec::Assignment))
     return true;
   return Left.isOneOf(tok::comma, tok::coloncolon, tok::semi, tok::l_brace,
-                      tok::kw_class, tok::kw_struct) ||
+                      tok::kw_class, tok::kw_struct, tok::kw_union) ||
```
llvm-beanz, line 2263: Same as above. C++ doesn't need this, so HLSL shouldn't either.

tools/clang/lib/Format/UnwrappedLineParser.cpp
```diff
@@ -1525,7 +1525,8 @@ void UnwrappedLineParser::parseEnum() {
     nextToken();
 
   // Eat up enum class ...
-  if (FormatTok->Tok.is(tok::kw_class) || FormatTok->Tok.is(tok::kw_struct))
+  if (FormatTok->Tok.is(tok::kw_class) || FormatTok->Tok.is(tok::kw_struct) ||
+      FormatTok->Tok.is(tok::kw_union))
```
llvm-beanz, line 1529: I assume we don't support enum unions, so we probably don't need this change either.

tools/clang/lib/Parse/ParseDeclCXX.cpp
```diff
@@ -1580,7 +1580,9 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
 
   if (!Name && !TemplateId && (DS.getTypeSpecType() == DeclSpec::TST_error ||
                                TUK != Sema::TUK_Definition)) {
-    if (DS.getTypeSpecType() != DeclSpec::TST_error) {
+    if (getLangOpts().HLSL && TagTokKind == tok::kw_union) {
+      Diag(Tok, diag::err_anonymous_unions);
```
llvm-beanz, line 1584: I'm not sure this is the right place for this check. `Sema::BuildAnonymousStructOrUnion` seems more appropriate. You can basically just add `if (getLangOpts.HLSL) { Invalid=true; Diag(...); }` around line 4108, and it should do the right thing.

llvm-beanz (changes requested):
This mostly looks good. I have a few areas where I'm concerned that changes shouldn't be required because they aren't required for C++.

I'm super excited to see this coming in!

hekota:
> Windows Debug failed but Release succeeded. Can someone trigger a retry?

The build failed because of crash in test. Can you please try if it repros for you locally?

```
StartGroup: D:\a\1\s\tools\clang\test\HLSL\..\HLSLFileCheck\hlsl\types\union\embeddedEmptyStruct.hlsl
Error: TAEF: Capturing minidump of crashing host process.
Saved minidump to "D:\a\1\Debug\test\WexLogFileOutput\000001_~_embeddedEmptyStruct.hlsl_MiniDump.dmp".
Error: TAEF: [HRESULT 0x800706BE] A failure occurred while running a test operation: 'CompilerTest::BatchHLSL'. (A crash with exception code 0x80000003 occurred in module "dxcompiler.dll" in the process hosting the test code while invoking a test operation.)
Error: Wex.Logger Mismatched Group: A log grouping named "CompilerTest::BatchHLSL" was ended when the active group was "D:\a\1\s\tools\clang\test\HLSL\..\HLSLFileCheck\hlsl\types\union\embeddedEmptyStruct.hlsl".
EndGroup: CompilerTest::SubobjectCodeGenErrors [Passed]
EndGroup: Wex.Logger Mismatched Group: D:\a\1\s\tools\clang\test\HLSL\..\HLSLFileCheck\hlsl\types\union\embeddedEmptyStruct.hlsl [Failed]
EndGroup: CompilerTest::BatchHLSL [Failed]
TestSkipped: TAEF: The cleanup method 'TestModuleCleanup' will not be run as TAEF has stopped communicating with the test host process due to a previous failure.

Summary of Non-passing Tests:
    CompilerTest::BatchHLSL [Failed]
        D:\a\1\s\tools\clang\test\HLSL\..\HLSLFileCheck\hlsl\types\union\embeddedEmptyStruct.hlsl [Failed]
```

danbrown-amd:
[AMD Official Use Only]

All tests had passed locally.  The patch author will be submitting an update soon anyway, so no need to rerun at this point.

-- Dan.

tools/clang/lib/Sema/SemaDecl.cpp
```diff
@@ -5820,7 +5820,7 @@ Sema::ActOnVariableDeclarator(Scope *S, Declarator &D, DeclContext *DC,
 
         // C++98 [class.union]p1: If a union contains a static data member,
         // the program is ill-formed. C++11 drops this restriction.
-        if (RD->isUnion())
+        if (RD->isUnion() && !getLangOpts().HLSL)
```
llvm-beanz, line 5823: Something else that occurred to me. Is this change right? Do we want to allow static data members inside a union? Does that even make sense?

llvm-beanz:
> All tests had passed locally. The patch author will be submitting an update soon anyway, so no need to rerun at this point.

Did you have asserts enabled? I'm seeing that crash locally with this PR.

llvm-beanz:
The crash is caused by creating duplicate struct annotations for the union type. You can resolve the crash with the following patch:

```patch
diff --git a/tools/clang/lib/CodeGen/CGHLSLMS.cpp b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
index 812254c53..50ba2e156 100644
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -1283,6 +1283,10 @@ unsigned CGMSHLSLRuntime::AddTypeAnnotation(QualType Ty,
   } else if (const RecordType *RT = paramTy->getAsUnionType()) {
     RecordDecl *RD = RT->getDecl();
     llvm::StructType *ST = CGM.getTypes().ConvertRecordDeclType(RD);
+    if (DxilStructAnnotation *annotation = dxilTypeSys.GetStructAnnotation(ST)) {
+      unsigned structSize = annotation->GetCBufferSize();
+      return structSize;
+    }
     DxilStructAnnotation *annotation = dxilTypeSys.AddStructAnnotation(
         ST, GetNumTemplateArgsForRecordDecl(RT->getDecl()), true);
     DxilPayloadAnnotation *payloadAnnotation = nullptr;
```

llvm-beanz:
There are other asserts that are firing in other tests once this one clears. Please make sure you are building and running tests with asserts enabled.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1137,6 +1137,9 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(DxilStructAnnotation *annota
         fieldAnnotation.SetPrecise();
     }
 
+    if (RD->isUnion())
```
llvm-beanz, line 1140: Looking at this deeper, I think this bit is a problem.

This change as it is makes unions empty structs in the DXIL StructAnnotation, which causes them to have size 0 in cbuffers, and is the first of _many_ issues that cause HLSL's SROA pass to fall over.

Once you fix that, SROA currently restricts how bitcasts can be used in ways that are fundamentally incompatible with unions. Overall I think this change is going to need a lot more baking.

llvm-beanz (changes requested):
I suspect your development here may have hit a series of unfortunate problems with testing.

The HLSLFileCheck tests don't currently run on non-Windows platforms. So unless you're developing on Windows, they aren't running, and if you're not building and running with asserts enabled you're not going to see the issues that are popping up.

Additionally, your verifier tests don't seem to be connected in VerifierTest.cpp, so they're not running either. Connecting them locally, they don't pass for me.

tools/clang/lib/Format/TokenAnnotator.cpp
mthatish: Done. New pull request has been created. Here is the link :https://github.com/microsoft/DirectXShaderCompiler/pull/4132

tools/clang/lib/Format/TokenAnnotator.cpp
mthatish: Done

tools/clang/lib/Format/UnwrappedLineParser.cpp
mthatish: Done

tools/clang/lib/Parse/ParseDeclCXX.cpp
mthatish: I tried this change, but anonymous unions in HLSL don't hit this function. I understood your motive, though. Ill see what I can do here.

tools/clang/lib/Sema/SemaDecl.cpp
mthatish: Yes, Standard says that static data members are allowed inside a union.

mthatish:
Thanks for all the reviews :) The tests are passing locally and I'm assuming they will all pass this time in the new pull request.

PR #4121, opened 2021-12-03
Avoid crash when updating unused DIGV info

pow2clk:
When a Global Variable is replaced with another, that includes the debug
info for the variable. As a result, if the global variable is flattened
and an attempt is made to update the debug info, none is found, which
results in a crash when the null pointer is dereferneced.

Add test that prompts a memcpy replacement and check that the
replacement has been made without crashing

### DIFF
--- a/tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/memcpy_digv_replacement.hlsl
+++ b/tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/memcpy_digv_replacement.hlsl
@@ -1,0 +1,47 @@
+// RUN: %dxc -T vs_6_0 -Zi %s | FileCheck %s
+
+// Test replacement of static global in a case where SROA may try to
+// process a static global replaced by memcpy lowering.
+// This processing includes creating flattened GVs for it
+// and then updating the debug info for them with the original DIGV,
+// which no longer matches because of the replacement, so null is found
+// and unnecessary flattened variables were created.
+// Trivially dead GVs are now removed in SROA
+
+// Ensure the replacement var is flattened
+// CHECK: ReplacementVar.0 = internal unnamed_addr constant [2 x float]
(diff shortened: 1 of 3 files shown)


tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/memcpy_digv_replacement.hlsl
```diff
@@ -0,0 +1,30 @@
...
+
+float4 main(int ix : I) : SV_Position
+{
+    UnusedVar = ReplacementVar;
+    return  UnusedVar[ix];
+}
+
+
+// Exclude quoted source file (see readme)
+// CHECK-LABEL: {{!"[^"]*\\0A[^"]*"}}
```
adam-yang, line 29: What if we check for the DIGV pointing to `ReplacementVar`?

adam-yang: Are there two sets of DIGV's both pointing to `ReplacementVar.*`?

pow2clk: No, the change means that the DIGV doesn't get flattened. So there is one lingering reference to the unusedvar

pow2clk: I can add checks for the DIGVs

tex3d: So does this test cause the assert to fire on debug build?  That means we can't pass the tests on a debug build.
If we have an identifiable case that hits an assert, we should see what we can change to fix that, shouldn't we?

PR #4125, opened 2021-12-03
[NFC] Fix broken template test case

llvm-beanz:
This test case worked in PR tests, but was broken on merge to master
due to other changes that were merged between when the PR went up and
when it was merged.


tex3d (approved):
LGTM!

PR #4131, opened 2021-12-06
[spirv] handle hull shader OutputPatch

jaebaek:
The existing code creates a temporary variable with Output storage class
for the parameter with OutputPatch type for the patch constant
functions. However, this adds an additional output stage variable and
consumes more locations. In addition, it can also change the rendering
result depending on the driver.

This commit removes the additional output stage variable and use the
actual output stage variables for the argument passing for the
OutputPatch. This is correctly working because the output stage variable
keeps the output value for all invocation id and we can simply reuse it.


jaebaek:
@tex3d This PR just cherry-picks b6360c6c0ce83c7c8ada3d8909f1d2103736b109 for release-1.6.2112

PR #4132, opened 2021-12-06
Unions support in HLSL.

danbrown-amd:
Unions support in HLSL.


tools/clang/lib/Parse/ParseDeclCXX.cpp
```diff
@@ -1580,7 +1580,9 @@ void Parser::ParseClassSpecifier(tok::TokenKind TagTokKind,
 
   if (!Name && !TemplateId && (DS.getTypeSpecType() == DeclSpec::TST_error ||
                                TUK != Sema::TUK_Definition)) {
-    if (DS.getTypeSpecType() != DeclSpec::TST_error) {
+    if (getLangOpts().HLSL && TagTokKind == tok::kw_union) {
```
mthatish, line 1583: Leaving it as syntax error makes more sense to me than the semantic error. Please let me know if you people think differently.

llvm-beanz (changes requested):
Not all of your unit tests are running. You need to add the following to the end of `tools\clang\unittest\HLSL\VerifierTest.cpp` to run the HLSL test cases that you've added:

```cpp
TEST_F(VerifierTest, UnionAnon) {
  CheckVerifiesHLSL(L"union_anon.hlsl");
}

TEST_F(VerifierTest, UnionDerivedToBase) {
  CheckVerifiesHLSL(L"union-derived-to-base.hlsl");
}

TEST_F(VerifierTest, UnionSizeOf) {
  CheckVerifiesHLSL(L"union-sizeof.hlsl");
}

TEST_F(VerifierTest, Unions0) {
  CheckVerifiesHLSL(L"unions_0.hlsl");
}
```

mthatish:
> Not all of your unit tests are running. You need to add the following to the end of `tools\clang\unittest\HLSL\VerifierTest.cpp` to run the HLSL test cases that you've added:
> 
> ```c++
> TEST_F(VerifierTest, UnionAnon) {
>   CheckVerifiesHLSL(L"union_anon.hlsl");
> }
> 
> TEST_F(VerifierTest, UnionDerivedToBase) {
>   CheckVerifiesHLSL(L"union-derived-to-base.hlsl");
> }
> 
> TEST_F(VerifierTest, UnionSizeOf) {
>   CheckVerifiesHLSL(L"union-sizeof.hlsl");
> }
> 
> TEST_F(VerifierTest, Unions0) {
>   CheckVerifiesHLSL(L"unions_0.hlsl");
> }
> ```

Done

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1137,6 +1137,9 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(DxilStructAnnotation *annota
         fieldAnnotation.SetPrecise();
     }
 
+    if (RD->isUnion())
+      break;
+
```
llvm-beanz, line 1142: If you put a union in a constant buffer, the size of the union comes back as 0, which isn't correct.

llvm-beanz (changes requested):
Thank you for working on this. This is really shaping up!

A few thoughts.

Should we even be disabling anonymous unions? They can be super useful, and are frequently used for vector and SIMD code.

We should have tests that test some of the interactions between unions and other HLSL features. I don't think the struct annotations are correct, and one test that demonstrates it is adding a union to a ConstantBuffer.

Additionally we should test applying semantics to union elements (which should be an error).

tools/clang/lib/CodeGen/CGHLSLMS.cpp
mthatish: Your comment above::
I don't think the struct annotations are correct, and one test that demonstrates it is adding a union to a ConstantBuffer.

Can you tell me which test case is this about?

llvm-beanz: There isn’t a test case in the PR that covers this, but any shader using a `ConstantBuffer<T>`, where `T` is or contains a union type will generate incorrect cbuffer layout information.

mthatish:
Can I please get another review? Thanks

llvm-beanz:
The PR still doesn't have a test case verifying unions in cbuffers. I checked out the PR locally built and tried to run the following code and the compiler crashes:

```c
// RUN: %dxc -E main -T ps_6_0 -HV 2021 %s | FileCheck %s

union U {
  uint x;
  float4 v;
};

cbuffer Foo
{
  U g;
};


float4 main(int2 a : A) : SV_TARGET
{
  return g.v;
}
```

include/llvm/IR/Type.h
```diff
@@ -71,7 +71,8 @@ class Type {
     StructTyID,      ///< 12: Structures
     ArrayTyID,       ///< 13: Arrays
     PointerTyID,     ///< 14: Pointers
-    VectorTyID       ///< 15: SIMD 'packed' format, or other vector type
+    VectorTyID,      ///< 15: SIMD 'packed' format, or other vector type
+    UnionTyID        ///< 16: Unions
```
llvm-beanz, line 75: This seems unnecessary. We shouldn't need a new union type in the LLVM IR layer, and I don't think this is even used.

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -499,6 +499,8 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
     opts.EnableOperatorOverloading = true;
     // Enable template support
     opts.EnableTemplates = true;
+    // Enable Unions support
+    opts.EnableUnions = true;
```
llvm-beanz, line 503: This should be gated on HLSL 202x rather than 2021. You'll need to rebase your branch to get the new version support.

tools/clang/test/HLSLFileCheck/hlsl/types/union/union_cbuffer.hlsl
```diff
@@ -0,0 +1,16 @@
+// RUN: %dxc -E main -T ps_6_0 -enable-unions -HV 2021 %s | FileCheck %s
+
+// CHECK:define void @main
```
llvm-beanz, line 3: It is not sufficient to test that `main` is defined in the generated IR. Tests should test that the output is correct as expected.

In this test case verifying the union member access IR is appropriate.

Other tests in this PR also need revision.

llvm-beanz (changes requested):

lib/DxcSupport/HLSLOptions.cpp
mthatish: Done

tools/clang/test/HLSLFileCheck/hlsl/types/union/union_cbuffer.hlsl
mthatish: Done

mthatish:
Can i please get another review?

tools/clang/lib/CodeGen/CGRecordLayoutBuilder.cpp
```diff
@@ -261,8 +261,10 @@ void CGRecordLowering::lower(bool NVBaseType) {
   // 8) Format the complete list of members in a way that can be consumed by
   //    CodeGenTypes::ComputeRecordLayout.
   CharUnits Size = NVBaseType ? Layout.getNonVirtualSize() : Layout.getSize();
+#if 0
```
llvm-beanz, line 264: This seems incorrect.

include/dxc/Support/HLSLOptions.td
```diff
@@ -300,6 +300,8 @@ def disable_lifetime_markers : Flag<["-", "/"], "disable-lifetime-markers">, Gro
   HelpText<"Disable generation of lifetime markers where they would be otherwise (6.6+)">;
 def enable_templates: Flag<["-", "/"], "enable-templates">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Enable template support for HLSL.">;
+def enable_unions: Flag<["-", "/"], "enable-unions">, Group<hlslcomp_Group>, Flags<[CoreOption]>,
```
llvm-beanz, line 303: @pow2clk or @tex3d may have a different opinion from me, but I'd prefer if we didn't add individual feature flags for HLSL 202x features. I think they complicate the testing configurations, and we won't want to support them on release.

include/llvm/IR/Type.h
llvm-beanz, line 75: I don't understand why this is necessary. Clang has never needed an IR Union type before, so I think adding one is concerning. Additionally, if we do require this, it changes DXIL so we'll need to evaluate the impact of this on other hardware vendors.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1146,16 +1147,18 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(DxilStructAnnotation *annota
...
     }
-
     // Update offset.
     CBufferSize = std::max(CBufferSize, CBufferOffset + size);
     CBufferOffset = CBufferSize;
 
     ++Field;
   }
 
+  if (RD->isUnion() && CBufferSize == 0)
```
llvm-beanz, line 1157: How do we handle empty structs here? Unions and structs should be interchangeable in this case.

llvm-beanz (commented):
I want to start by apologizing because I think you've done a lot of work here (much of it great), but we're moving the goal posts on you 😭.

There is a set of problems relating to `union` types that are a bit challenging to address as we look at how unions interact with other HLSL features. We didn't do a great job for HLSL 2021 of assessing the impact of new features and the way they interact (sometimes poorly) with existing features.

With HLSL 202x we're trying to address that by being a bit more thoughtful up front about how these features interact, and for unions I think the simplest approach is to just disallow certain use cases.

A good example of one of the challenges is if you use a union inside a structure that is encapsulated in a resource, how do you interpolate the resource data? For this reason we think it is best to restrict how unions are used to cases that are clear and understandable.

I started drafting a feature proposal in our new proposal process for unions:
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0004-unions.md

While I do have some concrete suggestions that I've included inline, I think we also need to work out some of the details about where unions can and can't be used.

include/llvm/IR/Type.h
Alexander-Johnston: As mentioned this was unnecessary and we've since removed it.

include/dxc/Support/HLSLOptions.td
Alexander-Johnston: We still have this flag in the PR, but enable unions by default with LangStd::v202x. We can remove the flag entirely if you'd still like.

llvm-beanz: We should remove the option entirely, and add test cases to verify that use of unions causes an error with `-HV 2021`.

I think the code still enables unions under HLSL 2021 because the comparison is `>= 2021`, and should probably be `>= 202x`.

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -523,6 +523,43 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
     opts.EnableFXCCompatMode = true;
   }
 
+  // If the HLSL version is 2021, allow the 2021 features by default.
+  // If the HLSL version is 2016 or 2018, allow them only
+  // when the individual option is enabled.
+  // If the HLSL version is 2015, dissallow these features
+  if (opts.HLSLVersion >= hlsl::LangStd::v2021) {
```
llvm-beanz, line 530: See here, where this is enabled for 2021.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -957,112 +957,87 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(
...
+    for (FieldDecl *Field : RD->fields()) {
+      CBufferOffset = 0;
+
+      QualType fieldTy = Field->getType();
+
+      CBufferOffset = AlignBaseOffset(fieldTy, CBufferOffset, bDefaultRowMajor,
+                                      CGM, dataLayout);
+
+      const hlsl::ConstantPacking *packOffset = nullptr;
+      for (const hlsl::UnusualAnnotation *it : Field->getUnusualAnnotations()) {
```
llvm-beanz, line 977: It doesn't look like any of the tests apply unusual annotations to union members, and it seems to me like you shouldn't be able to specify a pack offset on a union member because all union members should start at the same pack offset.

The (extremely inadequate) spec that I wrote [here](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0004-unions.md) explicitly disallowed semantics on union fields, we should probably also extend that to disallow all HLSL annotations (semantics, pack offset, register, etc).

We should make sure to have parsing tests to verify that we get errors too.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -957,112 +957,87 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(
...
+          packOffset = cast<hlsl::ConstantPacking>(it);
+          CBufferOffset = packOffset->Subcomponent << 2;
+          CBufferOffset += packOffset->ComponentOffset;
+          // Change to byte.
+          CBufferOffset <<= 2;
+        } break;
         }
+      }
 
-        llvm::Type *Type = Types.ConvertTypeForMem(Field->getType());
```
llvm-beanz, line 989: I think something just tripped up the diff here. I think this code just all slid lower in the file rather than being removed, but there is a lot of change highlighted here that isn't related to unions.

llvm-beanz (commented):
One area that I'm concerned about is how we handle unions in `groupshared`, and how we might handle atomic operations on unions in `groupshared` memory. IIRC bit casts are disallowed on pointers to group shared memory, so we may need to design a way to handle that (or disallow it).

Nielsbishere:
Will this ever be merged? This is really nice.

danbrown-amd:
Not until there is a SPIR-V implementation, I'm told, and that has been set aside pending approval of an enabling SPIR-V extension.

llvm-beanz:
I'm not convinced that we need a SPIR-V extension to implement this, but we do require all language features be supported by both DXIL and SPIR-V.

In addition to the SPIR-V implementation I think we had some outstanding design considerations we needed to be sure to address. Specifically we need to make sure to understand all the ways this interfaces with data storage (groupshared device and local memory), as well as put in place appropriate limitations for shader inputs and outputs.

devshgraphicsprogramming:
SPIR-V bitcast can do most stuff you'd want for a union as long as the type you're casting between has the same size.

Theoretically for every "unioned" load, one could generate a synthetic (ofc hash-consed) SPIR-V struct with a member  of the unioned type at the correct manually declared offset.

Ofc this kinda hinges on **scalar layout**, IMHO I'd gatekeep a lot of interesting features behind `-fvk-use-scalar-layout`.

P.S. It might be useful to "legalize" bitcasts between the same type, otherwise the generated SPIR-V has issue with passing legalization now or in the future.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
jaykrell: Or a union with one member? :)

devshgraphicsprogramming:
> SPIR-V bitcast can do most stuff you'd want for a union as long as the type you're casting between has the same size.
> 
> Theoretically for every "unioned" load, one could generate a synthetic (ofc hash-consed) SPIR-V struct with a member of the unioned type at the correct manually declared offset.
> 
> Ofc this kinda hinges on **scalar layout**, IMHO I'd gatekeep a lot of interesting features behind `-fvk-use-scalar-layout`.
> 
> P.S. It might be useful to "legalize" bitcasts between the same type, otherwise the generated SPIR-V has issue with passing legalization now or in the future.

Errata correction, `bit_cast` can only cast between fundamental types (scalar or vector) and pointers of PhysicalStorageBuffer address space.

There's also some stuff about casting to pointers of integers, but the spec is soo weirdly worded I can't make out if its just convolutese for `Integer[]`.

s-perron:
It is possible to (partially) implement unions the same method used for byte address buffers. The union is represented using as an array of 32-bit integers, and when you need to load something of a particular type you extract the required bits, and build the object.

This has a few problems:

1. It is impossible to do atomic operations on anything other than 32-bit integers.
2. The size of the object has to be a multiple of 32-bits otherwise loading and storing will access memory that was not allocated. (Could be avoided if we use array of int8).
3. Potentially many extra instructions to load/store the various parts and combine them using bit operations.

To properly implement unions in HLSL using SPIR-V, we need [SPV_KHR_untyped_pointers](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_untyped_pointers.html). If we don't, we will end up with the same problems we have for byte address buffers. There is no Vulkan extension yet that enabled untyped pointers in vulkan yet.

devshgraphicsprogramming:
> It is possible to (partially) implement unions the same method used for byte address buffers. The union is represented using as an array of 32-bit integers, and when you need to load something of a particular type you extract the required bits, and build the object.
> 
> This has a few problems:
> 
>     1. It is impossible to do atomic operations on anything other than 32-bit integers.
> 
>     2. The size of the object has to be a multiple of 32-bits otherwise loading and storing will access memory that was not allocated. (Could be avoided if we use array of int8).
> 
>     3. Potentially many extra instructions to load/store the various parts and combine them using bit operations.
> 
> 
> To properly implement unions in HLSL using SPIR-V, we need [SPV_KHR_untyped_pointers](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_untyped_pointers.html). If we don't, we will end up with the same problems we have for byte address buffers. There is no Vulkan extension yet that enabled untyped pointers in vulkan yet.

Nice extension, this is basically OpenCL's `generic`, right?

Also I think with non-restrict BDA and SSBO (via magic of SSBO aliasing) you don't end up with that "only 32 bit" issue. The only downside is aliasing/norestrict whenever you use a `union`.

I think UBO can be aliased too, but there's no perf downside because they're readonly so aliasing doesn't affect them.


By aliasing I mean
```cpp
layout(set=S,binding=B) buffer AsUint8
{
   uint8_t data[];
} bindingBu8;

layout(set=S,binding=B) buffer AsUint16
{
   uint16_t data[];
} bindingBu16;

layout(set=S,binding=B) buffer AsUint32
{
   uint32_t data[];
} bindingBu32;

layout(set=S,binding=B) buffer AsUint64
{
   uint64_t data[];
} bindingBu64;
```
its allowed and works on SSBOs in Vulkan (but not in GL).

llvm-beanz:
As we've worked through more details implementing HLSL in Clang and writing down the language rules in the specification (in particular casting and initialization rules), I'm deeply concerned about the addition of `union` types to HLSL without some significant additional changes to clean up ambiguous cases that `union` objects would introduce.

For example, HLSL's initialization list syntax has no way to initialize a union that wouldn't be potentially ambiguous. For example:

```hlsl
union U {
  int X;
  float F;
  int2 Y;
};

U u = {5} // Is this an int or a float, or did you forget a second int to be an int vector?
U u2 = {{4}}; // Is this a single int or a vector with a missing second initializer?
```

HLSL requires all scalar fields in an initializer list to have a value, and ignores intermediate braces in the initializer list, so these cases are ambiguous with no clear way to resolve the ambiguity.

The second bad area is HLSL's "flat conversion", which allows an object of some type `T` to convert to a different type `U` if `T` and `U` have the same number of scalar members. This allows conversions like structures to vectors, or structures to different structures.

Unions also pose a big problem in this conversion for two reasons:
1) The conversion is element-wise and respects proper data conversions (i.e. a structure of 2 integers, converts to a structure of 2 floats by integer->float conversion, not bitcasts), this is problematic for `union` types because the actual type of the stored data is unknown at compile time.
2) The conversion is only valid when the number of scalar elements in the source type and the destination type are the same. The number of scalar elements in a `union` type is dependent on the active field of the union, which varies at runtime, so we cannot determine if this cast is valid at compile time.

Without the ability to initialize or cast `union` types, it is unlikely that union types would be particularly useful, so before we consider adding `union` types to the language we'll need to address underling problems with initialization and casting.

PR #4133, opened 2021-12-07
[HLSL2021] Disable operators that rely on references

llvm-beanz:
After lots of discussion, we think it is best to disable operators that
should rely on reference parameters or return types for HLSL 2021. The
main driver of this is that we believe that references will get added to
HLSL in the near future, and we'd like to make it easier to maintain
compatability between HLSL 2021 and future versions of HLSL where
references are supported.


tex3d (approved):
LGTM!

llvm-beanz:
I realized that I somehow managed to not include the test cases and updates...

PR #4134, opened 2021-12-07
[NFC] Ignore all build* directories

llvm-beanz:
This allows build directories to be ignored as long as they start with
the name `build`.


hekota (approved):
LGTM!

PR #4147, opened 2021-12-15
[SPIR-V] Improve lines and columns for DebugLine (pt 2)

greg-lunarg:
The continues the work of pt 1. In paticular, it improved DebugLine for
constructors; array refs; swizzle; increments; intrinsics; builtins;
if; loops; switch; variable initialization; compound assigns; zero casts;
texture samples; vector and matrix swizzle LHS assignment;
matrix, struct and array initializers; function call args;
matrix element stores; matrix multiply.

### DIFF
--- a/tools/clang/include/clang/AST/Stmt.h
+++ b/tools/clang/include/clang/AST/Stmt.h
@@ -923,6 +923,7 @@ class IfStmt : public Stmt {
 
   SourceLocation IfLoc;
   SourceLocation ElseLoc;
+  SourceLocation MergeLoc;
 
 public:
   IfStmt(const ASTContext &C, SourceLocation IL, VarDecl *var, Expr *cond,
@@ -964,6 +965,8 @@ public:
   void setIfLoc(SourceLocation L) { IfLoc = L; }
   SourceLocation getElseLoc() const { return ElseLoc; }
   void setElseLoc(SourceLocation L) { ElseLoc = L; }
+  SourceLocation getMergeLoc() const { return MergeLoc; }
+  void setMergeLoc(SourceLocation L) { MergeLoc = L; }
 
   SourceLocation getLocStart() const LLVM_READONLY { return IfLoc; }
   SourceLocation getLocEnd() const LLVM_READONLY {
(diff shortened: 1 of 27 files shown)


tools/clang/tools/dotnetc/OptEditorForm.cs
```diff
@@ -71,8 +71,8 @@ private void LeftButton_CheckedChanged(object sender, EventArgs e)
         private void ApplyChangesButton_Click(object sender, EventArgs e)
```
jaebaek, line 71: I do not think you actually changed this file. For the future readers of this commit, could you please drop the change of this file?

jaebaek (approved):
LGTM.

Personally, the order of 

<id> Line Start | <id> Line End | <id> Column Start | <id> Column End
-- | -- | -- | --


for [DebugLine](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugLine) is confusing.
<id> Line Start | <id> Column Start | <id> Line End | <id> Column End
-- | -- | -- | --

looks more intuitive. Anyway, it is not related to this PR.

greg-lunarg:
> Could you please roll back the change of tools/clang/tools/dotnetc/OptEditorForm.cs?
> It is not related to this commit, right?

Right. That change does not represent any intentional or functional change to that file. One day after a pull that file started showing a phantom change and nothing I tried would let me get rid of it. I suspect there was some git-based line-ending inconsistency that was not visible. I had been successfully working around it and avoiding checking it in, but it somehow it still ended up in my final squash and now I don't know how to get it out. I don't know what the change was in the first place so I don't know how to undo it. The good news is that with this commit, I am no longer seeing the problem. Could I just add a comment that this commit does not functionally change that file?

jaebaek:
Well .. you can just download the source code and copy only that file. Or you can use `git reset master tools/clang/tools/dotnetc/OptEditorForm.cs`. Since `tools/clang/tools/dotnetc/OptEditorForm.cs` is particularly not under our ownership (SPIR-V code gen contributors), I would avoid updating it if possible.

greg-lunarg:
OK. I think I know how to do it...

greg-lunarg:
@jaebaek I have removed the change to the non-SPIRV file, so I believe I have addressed all your issues to this point.

jaebaek:
@greg-lunarg Thank you for your contribution!

PR #4152, opened 2021-12-15
Fix verifier tests for HLSL 2021

llvm-beanz:
This fixes two issues in the verifier tests that impacted HLSL 2021.

(1) The verifier test runner only handles one `RUN` line, which results in the tests skipping subsequent lines. This becomes a problem with our pattern of features tests for HLSL 2021, where we verify both the individual feature flag and `-HV 2021`

(2) Turns out, the clang option parser had not been updated for HLSL 2021, and was not able to parse the option at all. We missed this because the second run line is the one that we put `-HV 2021` on, and because of the previous issue, we weren't running those tests. Additionally HLSL 2021 wasn't implemented with the same feature set in the clang driver, so I copied over the feature settings.

### DIFF
--- a/tools/clang/lib/Frontend/CompilerInvocation.cpp
+++ b/tools/clang/lib/Frontend/CompilerInvocation.cpp
@@ -1727,7 +1727,7 @@ static void ParseLangArgs(LangOptions &Opts, ArgList &Args, InputKind IK,
   else {
     try {
       Opts.HLSLVersion = std::stoi(std::string(ver));
-      if (Opts.HLSLVersion < 2015 || Opts.HLSLVersion > 2018) {
+      if (Opts.HLSLVersion < 2015 || Opts.HLSLVersion > 2021) {
        Diags.Report(diag::err_drv_invalid_value)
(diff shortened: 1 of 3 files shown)


tools/clang/lib/Frontend/CompilerInvocation.cpp
```diff
@@ -1727,7 +1727,7 @@ static void ParseLangArgs(LangOptions &Opts, ArgList &Args, InputKind IK,
   else {
     try {
       Opts.HLSLVersion = std::stoi(std::string(ver));
-      if (Opts.HLSLVersion < 2015 || Opts.HLSLVersion > 2018) {
+      if (Opts.HLSLVersion < 2015 || Opts.HLSLVersion > 2021) {
```
bfavela, line 1730: Theoretically, aren't 2019 and 2020 also errors?

llvm-beanz: Sadly, we just range check this value... We totally should make this an enum rather than an integer and do a real conversion. I've got that on my mental to do list.

gracejennings (approved):
LGTM!

PR #4154, opened 2021-12-16
FixIt diagnostic for HLSL 2021 ?: operator

llvm-beanz:
This updates the diagnostic for the HLSL 2021 short-circuiting ternary operator to issue a fixit suggesting using the select intrinsic. I don't have a good way right now to execute the FixIt, which makes it difficult to write a test that it applies cleanly. I'm working on that in a subsequent patch, but it's not going well...

This patch does add a test that verifies the error emission, and the diagnostic provided will include a FixIt resulting in the error output:

```
./vector-select.hlsl:14:17: error: condition for short-circuiting ternary operator must be scalar, for non-scalar types use 'select'
  rw[0] = rw[0] ? rw[0] : c.f;
          ~~~~~~^~~~~~~~~~~~~
          select(rw[0], rw[0], c.f)
```


jaebaek (approved):

gracejennings (approved):
LGTM

PR #4156, opened 2021-12-17
Remove DXC_BUILD_ARCH variable

llvm-beanz:
This commit simplifies the cross-targeting build support to not require
specifying the build architecture explicitly. Instead it is derived from
the build tools used in the generation.

I've tested this change locally for Win32, x64 and arm64.


cmake/modules/FindTAEF.cmake
```diff
@@ -27,15 +27,7 @@ macro(find_taef_libraries targetplatform)
...
-if ("${DXC_BUILD_ARCH}" STREQUAL "x64" )
-  find_taef_libraries(x64)
-elseif (CMAKE_GENERATOR MATCHES "Visual Studio.*ARM" OR "${DXC_BUILD_ARCH}" STREQUAL "ARM")
-  find_taef_libraries(arm)
-elseif (CMAKE_GENERATOR MATCHES "Visual Studio.*ARM64" OR "${DXC_BUILD_ARCH}" MATCHES "ARM64.*")
-  find_taef_libraries(arm64)
-elseif ("${DXC_BUILD_ARCH}" STREQUAL "Win32" )
-  find_taef_libraries(x86)
-endif ("${DXC_BUILD_ARCH}" STREQUAL "x64" )
+find_taef_libraries(${CMAKE_C_COMPILER_ARCHITECTURE_ID})
```
hekota, line 30: Could you please check if hctbuild -arm64ec works? I suspect it might not in this case. There is no arm64ec or arm64x directory in WinSDK, we should use the arm64 one.

llvm-beanz: I'm struggling to get arm64ec to build locally without this change. I'm sure I don't have my system setup correctly :(

gongminmin:
CMAKE_<LANG>_COMPILER_ARCHITECTURE_ID is introduced in cmake 3.10. The cmake_minimum_required(VERSION 2.8.12.2) in root cmakelists may need to be changed, too.

pow2clk (approved):
Looks good. Can you confirm that arm is working as Helena was worried about?

llvm-beanz:
Arm64 works fine for me locally, but I can't get Arm64EC to build without this change, and with it I get the same failure.

hekota (approved):

hekota:
Go ahead and complete it, the ARM64EC should work down the line.

PR #4157, opened 2021-12-17
Removing invalid hlsl from test

llvm-beanz:
This test case contained some invalid HLSL (dependent typedef), and
slipped in because there was a bug in templates that allowed dependent
typedefs at the time this test was posted. The bug was fixed before the
PR was approved and merged, so we are now broken :(.


hekota (approved):
LGTM!

PR #4158, opened 2021-12-18
Disable NRVO for `precise` decls

llvm-beanz:
Named return value optimization prevents allocating objects when
returning objects by value. This optimization trips up our
implementation of `precise` whihc requires the attribute be applied to
an `alloca`.

Disabling NRVO for Decls with the HLSLPreciseAttr attached avoids a
crash in the compiler.

Disabling this optimization should have no performance impact on most
existing HLSL code because function calls are inlined, and HLSL doesn't
support side-effecting constructors. Those two factors should allow the
optimizer to eliminte redundant copies even without NRVO enabled.


gracejennings (approved):
lgtm!

PR #4159, opened 2021-12-20
Preserve qualifiers on array subscript

### LINKED ISSUE #2860: DXC allows assigning to members of a const struct variable
Hello, I just noticed that DXC will compile the following code without any errors:

```
struct MyData {
   float3 x[4];  
};

StructuredBuffer<MyData> DataIn;
RWStructuredBuffer<MyData> DataOut;

[numthreads(64, 1, 1)]
void CSMain(uint3 dispatchid : SV_DispatchThreadID)
{
    const MyData data = DataIn[dispatchid.x];
    data.x[0] = 1.0f;
    DataOut[dispatchid.x] = data;
}
```

I would expect the `data.x[0] = 1.0f` assignment to cause a compiler error since the `data` variable is marked as `const`.

llvm-beanz:
This change prevents dropping const qualifiers on array subscript
operators. Without this change the array subscript operator converts
from an lvalue to an rvalue that drops the type qualifiers. Since
arrays are actually addresses to the start of the array, it results in
dropping the qualifiers on the underlying array, making the array
modifiable.

This resolves #2860

### DIFF
--- a/tools/clang/test/HLSL/InputPatch-const.hlsl
+++ b/tools/clang/test/HLSL/InputPatch-const.hlsl
@@ -1,9 +1,4 @@
-// RUN: %dxc -E main -T gs_6_0 %s | FileCheck %s
-
-// CHECK: InputPrimitive=patch2
-// CHECK: emitStream
-// CHECK: cutStream
-// CHECK: i32 24}
+// RUN: %clang_cc1 -fsyntax-only -ffreestanding -verify %s
 
 struct GSOut {
   float2 uv : TEXCOORD0;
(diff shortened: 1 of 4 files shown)


gracejennings (approved):
LGTM!

tools/clang/test/HLSL/InputPatch-const.hlsl
```diff
@@ -21,9 +16,9 @@ cbuffer b : register(b0) {
 [instance(24)]
 void main(InputPatch<GSOut, 2>points, inout PointStream<GSOut> stream) {
 
-  points[0].norm[0] = 1;
-  points[0].norm[1] = 2;
+  points[0].norm[0] = 1; // expected-error {{read-only variable is not assignable}}
+  points[0].norm[1] = 2; // expected-error {{read-only variable is not assignable}}
```
tex3d, line 20: I don't think points should be considered const.  Or is our subscript operator returning the wrong type here?  I believe this is incompatible with fxc with this change.

llvm-beanz: The docs for InputPatch’s operator[] say “ Gets a read-only resource variable.”, and our implementation returns const.

If that’s not correct we can fix our InputPatch implementation, but that should be orthogonal to this change.

See: https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-object-inputpatch

tex3d: I just checked and it turns out FXC doesn't like this code either.  Strange as I thought I'd seen this pattern somewhere, and thought this came from some sample (since it's under a path with that implication).  So it's all good!

PR #4163, opened 2021-12-25
Fix the order of shader profiles in hct code gen

### LINKED ISSUE #4138: Requires Python 3.6+?
Happen to find a building failure on an old computer with Python 3.5 installed. The problem is in HLSLOptions.td. The order of shader profiles from hctgen is different than the hand written one. Traced more on this, the issue is in utils/hct/hctdb_instrhelp.py:

```
 def getShaderProfiles():
    # order match DXIL::ShaderKind.
    profiles = {"ps":"4_0", "vs":"4_0", "gs":"4_0", "hs":"5_0", "ds":"5_0", "cs":"4_0", "lib":"6_1", "ms":"6_5", "as":"6_5"}
    return profiles;
```

Starting from Python 3.6, dict preserves the order of insertion. But not on 3.5. In README, the requirement of Python is 3.x, makes it not accurate.

Of course, another fix is change the dict here to OrderedDict. But still, OrderedDict was added in 3.1, not 3.0. Change the README to 3.6+ may be the easiest way.

gongminmin:
Fix #4138 . Change the dict to tuple, because dict is not order preserved.

### DIFF
--- a/utils/hct/hctdb_instrhelp.py
+++ b/utils/hct/hctdb_instrhelp.py
@@ -1200,7 +1200,7 @@ shader_profiles = [ shader_profile(0, "ps", "Kind::Pixel", "4_0", 32, 8),
 
 def getShaderProfiles():
     # order match DXIL::ShaderKind.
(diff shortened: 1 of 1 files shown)


llvm-beanz (approved):
Looks good to me! Thank you for the fix.

PR #4167, opened 2021-12-31
Fix the linking warnings about multiple imported DxcCreateInstance2

gongminmin:


pow2clk (commented):
First, we require a description of the change.

Second, perhaps that description could explain why this is necessary. dxcapi.h already defines DXC_API_IMPORT this way and that just before `DxcCreateInstance2` is declared with that macro prepended. What do these additions do that the macro in the header doesn't?

gongminmin:
> First, we require a description of the change.
> 
> Second, perhaps that description could explain why this is necessary. dxcapi.h already defines DXC_API_IMPORT this way and that just before `DxcCreateInstance2` is declared with that macro prepended. What do these additions do that the macro in the header doesn't?

dxcapi.h has DXC_API_IMPORT defined as dllimport. It's for the consumer of the DxcCreateInstance2. In dxcapi.cpp, DXC_API_IMPORT is defined as dllexport, which is for the provider of the DxcCreateInstance2.

Now take dxcpdbutils.cpp as an example. It includes dxcapi.h, but doesn't define DXC_API_IMPORT before it. As a result, DxcCreateInstance2 in dxcpdbutils.cpp is dllimport, in dxcapi.cpp is dllexport. They conflict with each other. In linking, you can see warnings about this conflict in those 3 cpp compilation units.

Normally a project with CMake uses this kind of export/import macros:

```
#if defined(_WIN32)
#define xxx_SYMBOL_EXPORT __declspec(dllexport)
#define xxx_SYMBOL_IMPORT __declspec(dllimport)
#else
#define xxx_SYMBOL_EXPORT __attribute__((visibility("default")))
#define xxx_SYMBOL_IMPORT
#endif

#ifdef xxx_EXPORTS
#define xxx_API xxx_SYMBOL_EXPORT
#else
#define xxx_API xxx_SYMBOL_IMPORT
#endif

```

For a project named xxx, CMake defines a xxx_EXPORTS macro to indicate it's building the project. Then the xxx_API is export for this project. But for the consumer of this project, xxx_EXPORTS is not defined. The same xxx_API becomes import.

damyanp:

PR #4171, opened 2022-01-05
Exclude errs and outs from per-thread FS

llvm-beanz:
The per-thread filesystem is useful if you're hitting the filesystem. If
you are writing to stdout/stderrs, multi-threading is going to cause all
sorts of crazy issues anyways the per-thread filesystem gets us nothing
but bugs.

This change adds an option to exclude raw_fd_ostreams from the
per-thread filesystem on creation. If opted-out they will fallback to
using the normal POSIX C interfaces. This change also removes the RAII
construct used to flush stdout and stderr and close stdout. Closing
stdout is not required on any platform, modern LLVM also excludes
closing stdout & stderr on all platforms.

### DIFF
--- a/include/llvm/Support/raw_ostream.h
+++ b/include/llvm/Support/raw_ostream.h
@@ -376,6 +376,8 @@ class raw_fd_ostream : public raw_pwrite_stream {
 
   bool SupportsSeeking;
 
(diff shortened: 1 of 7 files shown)


llvm-beanz:

PR #4172, opened 2022-01-05
[spirv] use OpIsHelperInvocation for [[vk::HelperInvocation]]

jaebaek:
For Vulkan 1.2 or less, we must not use HelperInvocation BuiltIn
decoration. Instread, for a variable with `[[vk::HelperInvocation]]`
attribute, we have to use `OpIsHelperInvocation` instruction:
  - Create a variable (with `Private` storage class)
  - In the module initialization, execute `OpIsHelperInvocation`
    instruction and store the result in the variable

### DIFF
--- a/docs/SPIR-V.rst
+++ b/docs/SPIR-V.rst
@@ -259,7 +259,13 @@ language. To support them, ``[[vk::builtin("<builtin>")]]`` is introduced.
 Right now the following ``<builtin>`` are supported:
 
 * ``PointSize``: The GLSL equivalent is ``gl_PointSize``.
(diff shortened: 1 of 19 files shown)


sudonatalie (approved):
LGTM!

sudonatalie (approved):
Actually, I just noticed that `vk::HelperInvocation` is also not documented in [Vulkan specific attributes](https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#vulkan-specific-attributes). Can you update it here?

PR #4174, opened 2022-01-05
Wave intrinsics inside of dead loops are not properly deleted

bfavela:
Summary of the changes
 - Issue was identified that dead loops with wave intrinsics were not being deleted which are marked as having side effects (due to code motion issues elsewhere).
 - The code modified here was checking for wave intrinsics, but only for "self-loops". The code was updated to merge the check for safety (no side effects) and the self-loop check which are functionally equivalent in that both have to bail if the block is not safe to delete.
 - During testing of the original change, some tests failed due to blocks not being properly deleted. The root cause of this was the walk of the successors is incorrect, since it would attempt to delete instructions that had uses later in the region. The code now does a post-order walk to satisfy these requirements.

### DIFF
--- a/lib/Transforms/Scalar/DxilEraseDeadRegion.cpp
+++ b/lib/Transforms/Scalar/DxilEraseDeadRegion.cpp
@@ -81,49 +81,17 @@ struct DxilEraseDeadRegion : public FunctionPass {
   }
 
   std::unordered_map<BasicBlock *, bool> m_SafeBlocks;
-  std::unordered_map<BasicBlock *, bool> m_HasSideEffect;
   MiniDCE m_DCE;
 
- bool HasSideEffects(BasicBlock *BB) {
-    auto FindIt = m_HasSideEffect.find(BB);
-    if (FindIt != m_HasSideEffect.end()) {
-      return FindIt->second;
+  // Replace all uses of every instruction in a block with undefs
+  void UndefBasicBlock(BasicBlock* BB) {
+    while (BB->begin() != BB->end()) {
+      Instruction *I = &BB->back();
+      if (!I->user_empty())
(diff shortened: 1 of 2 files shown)


lib/Transforms/Scalar/DxilEraseDeadRegion.cpp
```diff
@@ -245,16 +252,32 @@ struct DxilEraseDeadRegion : public FunctionPass {
     m_DCE.EraseAndProcessOperands(Common->getTerminator());
     BranchInst::Create(BB, Common);
 
-    // Delete the region
-    for (BasicBlock *BB : Region) {
+    // Delete the region by walking post-order (starting from the predecessors of BB)
```
adam-yang, line 255: Just curious, how did the original order cause problems?

bfavela: The "Region" data structure was filled out by walking the successors of the common dominator. If this code identified a chain of blocks such that the first block had instruction defs that were used in the second, it would try to delete the source defs first, which crashing because there are still pending uses. I believe deleting a region needs to be done in post order in order to satisfy those dependencies.

bfavela: Thanks for enticing me to have another look. I ended up discovering the issue was related to the "self-loop" cleanup done earlier. The shader I was chasing had a nested self-loop inside another loop that was dead, but wasn't cleaning the instructions inside of the self-loop, causing the use vectors to be wrong for the branch to the loop. This was because the self-loop was 'disconnected' in the graph, and not marked to be cleaned by the region builder.

adam-yang (approved):

PR #4175, opened 2022-01-06
Fix frequent warnings on gcc and clang

gongminmin:


pow2clk (approved):
Looks good! Thanks @gongminmin!

PR #4177, opened 2022-01-06
Adding shader model version promotion warning

### LINKED ISSUE #4038: Question: Detecting driver support for dxc compiled shaders
We're using dxc to compile shaders with SM5.1. Unfortunately, on some older AMD FirePro drivers, we're seeing errors from `CreateGraphicsPipelineState` such as:

> D3D12 ERROR: ID3D12Device::CreateVertexShader: Shader is corrupt or in an unrecognized format

It's easily resolved by updating the drivers to the most recent version (from AMD's site, not from Windows AutoUpdate). But, we would like to diagnose this issue ahead of time, and present the user with relevant help. Is there any way to check driver support for dxc shaders, other than just attempting to create a pipeline state with a shader and checking for this specific error result?

FWIW, our command line flags expand to something similar to (in debug):
```
dxc.exe input.ps -T ps_5_1 -Fh output.h -Vn output -E main -Zi -Qembed_debug
```

gracejennings:
Adding a warning when the shader model version is promoted from an older version to the 6.0 version. 

The motivation for adding this warning is to make this version promotion more explicit, clearing up errors if features from the newer SM are not supported on the driver.  Fixes #4038

### DIFF
--- a/tools/clang/tools/dxclib/dxc.cpp
+++ b/tools/clang/tools/dxclib/dxc.cpp
@@ -766,6 +766,9 @@ int DxcContext::Compile() {
       const hlsl::ShaderModel *SM = hlsl::ShaderModel::GetByName(m_Opts.TargetProfile.str().c_str());
       if (SM->IsValid() && SM->GetMajor() < 6) {
         TargetProfile = hlsl::ShaderModel::Get(SM->GetKind(), 6, 0)->GetName();
+        std::string versionWarningString =
+            "warning: Promoting older shader model profile to 6.0 version.";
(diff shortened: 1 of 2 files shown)


pow2clk (approved):
It looks good, but it would be nice to have a filecheck test. You'll want to use a run line similer to 

``` RUN: %dxc -T ps_5_1 %s | FileCheck -input-file=stderr %s```

The `-input-file=stderr` is required to get the warning output

pow2clk (approved):
Looks good!

tex3d (commented):
LGTM!

PR #4178, opened 2022-01-09
proposal: improve source compatibility with MinGW and Zig compilers

slimsag:
Hello! In [Mach engine](https://hexops.com/mach/) we are consuming the C/C++ sources of DirectXShaderCompiler directly (without Visual Studio, CMake, etc.) via the [Zig programming language](https://ziglang.org) (which is also a C/C++ compiler) in order to build Google's Dawn / WebGPU implementation.

Zig utilizes the MinGW windows headers by default, rather than the Windows SDK / official Microsoft headers. While MinGW headers are mostly compatible with those in the Windows SDKs, there are some notable omissions such as missing ATL support currently. Luckily, I found `WinIncludes.h` etc. in this repository has some emulated support for non-Windows targets.

Thus, in order to build dxcompiler with Zig, [we're maintaining these patches on our end](https://github.com/hexops/mach/issues/151). I'd love to upstream these patches so others can benefit from improved compatibility with Zig and MinGW, if you'd have it.

Would you be open to accepting such contributions? I'm happy to continue ensuring this works well into the future, is maintained, etc. and the changes are pretty minimal. I am also happy to send smaller PRs with more details on each patch if desired (I sent all patches in this PR so you can easily get the full picture of what is involved)

### DIFF
--- a/include/dxc/Support/WinAdapter.h
+++ b/include/dxc/Support/WinAdapter.h
@@ -15,7 +15,13 @@
 #ifndef LLVM_SUPPORT_WIN_ADAPTER_H
 #define LLVM_SUPPORT_WIN_ADAPTER_H
 
-#ifndef _WIN32
+#if !defined(_WIN32) || defined(__MINGW32__)
+
+// For MinGW, clang in GNU mode, Zig, etc. compilers we will emulate ATL types which are not
+// available in the MinGW headers yet.
+#ifdef __MINGW32__
+#include <windows.h>
+#endif // __MINGW32__
 
 #ifdef __cplusplus
 #include <atomic>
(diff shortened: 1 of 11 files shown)


lib/DxcSupport/WinAdapter.cpp
```diff
@@ -7,11 +7,45 @@
...
+
+//===--------------------------- CAllocator -------------------------------===//
+
+void *CAllocator::Reallocate(void *p, size_t nBytes) throw() {
+  return realloc(p, nBytes);
+}
+void *CAllocator::Allocate(size_t nBytes) throw() { return malloc(nBytes); }
+void CAllocator::Free(void *p) throw() { free(p); }
+
+#ifndef __MINGW32__
```
slimsag, line 48: Note the additions above here were merely moved from below in the file to here, so that we can utilize them when `__MINGW32__` is defined. Not modified at all.

lib/DxcSupport/dxcapi.use.cpp
```diff
@@ -33,7 +63,7 @@ static std::string GetWin32ErrorMessage(DWORD err) {
   DWORD formattedMsgLen =
       FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
                     nullptr, err, 0, formattedMsg, _countof(formattedMsg), 0);
-  if (formattedMsg > 0 && formattedMsgLen < _countof(formattedMsg)) {
+  if (formattedMsgLen > 0 && formattedMsgLen < _countof(formattedMsg)) {
```
slimsag, line 66: The Zig compiler caught this as it enables a number of additional clang safety checks by default. `formattedMsg` is `char formattedMsg[200]` and was being illegally coerced into an integer here. This was not a warning, but a hard error: clang thinks it is illegal C++ code.

I'm not sure how this would've behaved in other compilers (UB for sure I imagine), but it seems like a simple mistake where the intent was to check `formattedMsgLen` instead.

slimsag:
Hi team, just wanted to ping this and ask again if such patches would be welcomed or not?

include/dxc/Support/WinAdapter.h
```diff
@@ -38,21 +44,49 @@
...
+    struct __declspec(uuid(spec)) type;                                       \
+    template<> const GUID &__mingw_uuidof<type>() {                           \
+        static constexpr IID __uuid_inst = guid_from_string(spec);            \
+        return __uuid_inst;                                                   \
+    }                                                                         \
+    template<> const GUID &__mingw_uuidof<type*>() {                          \
+        return __mingw_uuidof<type>();                                        \
+    }                                                                         \
+    }
+
```
lygstate, line 70: @kusma This is not using constexpr and would case compiling issues when using __uuidof in static_assert

lib/DxcSupport/dxcapi.use.cpp
```diff
@@ -16,6 +16,36 @@
 #include "dxc/Support/FileIOHelper.h"
 #include "dxc/Support/WinFunctions.h"
 
+// For MinGW, export specializations for our public COM interface. See WinAdapter.h for more info.
+#ifdef __MINGW32__
+MINGW_UUIDOF(IDxcBlob, "8BA5FB08-5195-40e2-AC58-0D989C3A0102")
+MINGW_UUIDOF(IDxcBlobEncoding, "7241d424-2646-4191-97c0-98e96e42fc68")
```
lygstate, line 22: with constexpr, this won't be need anymore

include/llvm/Support/Atomic.h
```diff
@@ -16,6 +16,17 @@
...
+#define NOMINMAX 1
+#include <windows.h>
+// HACK: Pretend we're CYGWIN when including intrin.h, otherwise the MinGW headers will not know
+// that strcat and other conflicting symbols are already defined.
+#define __CYGWIN__
+#include <intrin.h>
+#undef __CYGWIN__
+#endif // __MINGW32__
+
 namespace llvm {
```
lygstate, line 30: This is  a upstream header, should never changed.

lib/DxcSupport/dxcapi.use.cpp
lygstate: This is a bug fix can be in seprate PR

PR #4180, opened 2022-01-11
Remove redundant uav load/store caused by partial update or copyin/out

python3kgae:
For the repro, only u[i].b need to be load and store.

struct S {

 float4x4 a[10];
 float4 b;
};

RWStructuredBuffer<S> u;

void foo(inout S s) {

  s.b += sin(s.b);
  
}

float4 main(uint i:I) : SV_Target {
  foo(u[i]);
  return u[i].b;
}


llvm-beanz (changes requested):
At a minimum this change needs a whole lot more testing, but I'm not sure if we're going to commit to getting this ready.

@python3kgae, if this isn't going to jump up the priority list please close the PR.

damyanp:

PR #4181, opened 2022-01-12
Fix cbuffer alignment for u64, add tests

tex3d:
When using uint64_t fields in structures, the member was not being aligned to 8-bytes in constant buffers like it should be.

Constant buffer offsets come from type annotation information, rather than raw data layout.  The CBufferOffset for FieldAnnotation was not being properly aligned for u64 field types, though it was ok for i64.  This was because the offset alignment code was missing ULongLong when checking the basic type.

This change fixes the issue with a minimal change by adding ULongLong to the if condition.

Modified and added tests to check this case for StructuredBuffer, cbuffer/tbuffer (both inside and outside struct member), ConstantBuffer, and TextureBuffer.


vcsharma (approved):

llvm-beanz (approved):
Patch looks good. Please update the PR description with a brief explanation before merging.

PR #4182, opened 2022-01-12
Remove redundant dbg.value instructions

adam-yang:
- Added a pass to remove redundant dbg.value's
- Removed known dead dbg.declares right at the end of SROA_Parameter_HLSL
- Added helper option `-print-after <pass>`

### DIFF
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -284,6 +284,8 @@ def validator_version : Separate<["-", "/"], "validator-version">, Group<hlslcom
   HelpText<"Override validator version for module.  Format: <major.minor> ; Default: DXIL.dll version or current internal version.">;
 def print_after_all : Flag<["-", "/"], "print-after-all">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Print LLVM IR after each pass.">;
+def print_after : Separate<["-", "/"], "print-after">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
+  HelpText<"Print LLVM IR after a specific pass.">;
 def ignore_opt_semdefs : Flag<["-", "/"], "ignore-opt-semdefs">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Ignore optional semantic defines which are not required to ensure program correctness.">;
 def ignore_semdef : Separate<["-", "/"], "ignore-semdef">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -284,8 +284,6 @@ def validator_version : Separate<["-", "/"], "validator-version">, Group<hlslcom
   HelpText<"Override validator version for module.  Format: <major.minor> ; Default: DXIL.dll version or current internal version.">;
 def print_after_all : Flag<["-", "/"], "print-after-all">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Print LLVM IR after each pass.">;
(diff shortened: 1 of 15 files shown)


include/dxc/Support/HLSLOptions.td
```diff
@@ -284,6 +284,8 @@ def validator_version : Separate<["-", "/"], "validator-version">, Group<hlslcom
   HelpText<"Override validator version for module.  Format: <major.minor> ; Default: DXIL.dll version or current internal version.">;
 def print_after_all : Flag<["-", "/"], "print-after-all">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Print LLVM IR after each pass.">;
+def print_after : Separate<["-", "/"], "print-after">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
+  HelpText<"Print LLVM IR after a specific pass.">;
```
tex3d, line 288: This is nice, but I immediately wonder if this could be made to work when supplied multiple times on the command line, collecting a set of passes you are interested in, instead of being limited to one.

tex3d: Also, a `-print-on-change` (or some better name) where it prints before and after for a matching pass only when it changes the module would be useful (and would require printing before the pass into a stream that's unused if unchanged).  Of course, this is just an idea for potential future use, not really relevant to this particular PR.

lib/HLSL/DxilDeleteRedundantDebugValues.cpp
```diff
@@ -0,0 +1,155 @@
...
+        else
+          NumNonDbgInsts++;
+      }
+    }
+  }
+
+  // Here is an arbitrary threshold where we decide there's too many
+  // debug instructions.
+  // TODO: Issue a warning here about removing debug instructions.
+  bool TryReduceDbgInsts = NumDbgInsts >= 10*NumNonDbgInsts && NumDbgInsts >= 100000;
```
tex3d, line 70: When is it not valuable to remove redundant debug insts?   Does this make debug info worse?

lib/HLSL/DxilDeleteRedundantDebugValues.cpp
```diff
@@ -0,0 +1,155 @@
...
+  SmallDenseMap<VarPair, DbgValueInst *> SeenVar;
+  SmallVector<DILocalScope *, 8> ScopeWorklist;
+
+  for (Function &F : M) {
+    SeenScopes.clear();
+    SeenVar.clear();
+
+    // Collect a set of all scopes that non-debug instructions
+    // are attached to. There's no need to keep debug info for
+    // any variables for any scopes that
```
tex3d, line 83: Finish sentence?

lib/HLSL/DxilDeleteRedundantDebugValues.cpp
adam-yang: I suppose that could be worded clearer. We can remove more debug insts by ignoring the debug locations' inlinedAt locations when collecting the set of scopes. Doing that will makes it impossible to look at some variables' values in parent scopes.

adam-yang:
```
void foo(inout float arg) {
  arg += 10; // While on this line, you should be able to look at variables in function `bar`, unless inlinedAt is ignored.
}

void bar(float val) {
  foo(val);
}

```

include/dxc/Support/HLSLOptions.td
adam-yang: Moving all this to a different PR now.

lib/HLSL/DxilDeleteRedundantDebugValues.cpp
adam-yang: As per our offline discussion, it's actually not really necessary to do this.

lib/HLSL/DxilDeleteRedundantDebugValues.cpp
adam-yang: Done.

tex3d (approved):
LGTM, but would be nice to have a test of some sort.

tex3d (approved):

PR #4191, opened 2022-01-17
Fix build failure caused by clang version 11.1.0 in Linux

jaebaek:


jaebaek:
I met the following build error:
```
... /lib/Support/Unix/Process.inc:98:10: error: 'mallinfo' is deprecated [-Werror,-Wdeprecated-declarations]
  mi = ::mallinfo();
         ^
/usr/include/malloc.h:118:48: note: 'mallinfo' has been explicitly marked deprecated here
extern struct mallinfo mallinfo (void) __THROW __MALLOC_DEPRECATED;
                                               ^
/usr/include/malloc.h:31:30: note: expanded from macro '__MALLOC_DEPRECATED'
# define __MALLOC_DEPRECATED __attribute_deprecated__
                             ^
/usr/include/x86_64-linux-gnu/sys/cdefs.h:261:51: note: expanded from macro '__attribute_deprecated__'
# define __attribute_deprecated__ __attribute__ ((__deprecated__))
                                                  ^
1 error generated.
```

sudonatalie:
Should we just use `mallinfo2` when it is available instead of ignoring?

sudonatalie:
This version is failing because it's checking whether `mallinfo` exists, but then using `mallinfo2` (which may not exist). It should be:

```
#if defined(HAVE_MALLINFO2)
  struct mallinfo2 mi;
  mi = ::mallinfo2();
  return mi.uordblks;
#elif defined(HAVE_MALLINFO)
  struct mallinfo mi;
  mi = ::mallinfo();
  return mi.uordblks;
```

I haven't tried it but I expect that should build without errors on both versions.

sudonatalie (approved):
LGTM, thanks for the changes!

PR #4192, opened 2022-01-17
[spirv] add `vk::ext_result_id<T>` type

jaebaek:
vk::ext_result_id<T> type. We must use it for a variable definition or
a function parameter. It means we do not consider it as a physical
storage. Instead, it will be a result id of the instruction.
    
Related to #3919

### DIFF
--- a/lib/Support/Unix/Process.inc
+++ b/lib/Support/Unix/Process.inc
@@ -99,7 +99,10 @@ size_t Process::GetMallocUsage() {
   return mi.uordblks;
 #elif defined(HAVE_MALLINFO)
   struct mallinfo mi;
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
   mi = ::mallinfo();
+#pragma clang diagnostic pop
   return mi.uordblks;
 #elif defined(HAVE_MALLOC_ZONE_STATISTICS) && defined(HAVE_MALLOC_MALLOC_H)
   malloc_statistics_t Stats;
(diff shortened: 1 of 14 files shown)


jaebaek:
@jiaolu Could you please review this PR?

jiaolu:
> @jiaolu Could you please review this PR?

sure, will take a look

jiaolu (commented):
very cool implementation, 
looks great to me

lib/Support/Unix/Process.inc
```diff
@@ -95,7 +95,10 @@ unsigned Process::getPageSize() {
 size_t Process::GetMallocUsage() {
 #if defined(HAVE_MALLINFO)
   struct mallinfo mi;
+#pragma clang diagnostic push
```
sudonatalie, line 98: nit: Guessing this unrelated piece should be rebased onto your other PR?

tools/clang/include/clang/SPIRV/AstTypeProbe.h
```diff
@@ -324,6 +324,12 @@ bool isOrContainsNonFpColMajorMatrix(const ASTContext &,
                                      const SpirvCodeGenOptions &, QualType type,
                                      const Decl *decl);
 
+/// \bried Returns true if the given type is `vk::ext_result_id<T>`.
```
sudonatalie, line 327: nit: Looks like a copy-paste error from below, these should all be "\brief"

sudonatalie (approved):
LGTM % nits

lib/Support/Unix/Process.inc
jaebaek: Removed.

tools/clang/include/clang/SPIRV/AstTypeProbe.h
jaebaek: Nice catch!

PR #4194, opened 2022-01-18
[spirv] support [[vk::ext_decorate_id]] attribute

jaebaek:
vk::ext_decorate_id attribute.
    
Related to #3919

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvInstruction.cpp
+++ b/tools/clang/lib/SPIRV/SpirvInstruction.cpp
@@ -234,14 +234,16 @@ SpirvDecoration::SpirvDecoration(SourceLocation loc,
 SpirvDecoration::SpirvDecoration(SourceLocation loc,
                                  SpirvInstruction *targetInst,
                                  spv::Decoration decor,
-                                 llvm::StringRef strParam,
+                                 llvm::ArrayRef<llvm::StringRef> strParams,
                                  llvm::Optional<uint32_t> idx)
     : SpirvInstruction(IK_Decoration, getDecorateStringOpcode(idx.hasValue()),
                        /*type*/ {}, loc),
       target(targetInst), targetFunction(nullptr), decoration(decor),
       index(idx), params(), idParams() {
-  const auto &stringWords = string::encodeSPIRVString(strParam);
-  params.insert(params.end(), stringWords.begin(), stringWords.end());
+  for (llvm::StringRef str : strParams) {
+    const auto &stringWords = string::encodeSPIRVString(str);
+    params.insert(params.end(), stringWords.begin(), stringWords.end());
+  }
 }
 
 SpirvDecoration::SpirvDecoration(SourceLocation loc,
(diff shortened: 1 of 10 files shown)


jaebaek:
@jiaolu This is another PR for #3919. Could you please review the code?

tools/clang/lib/SPIRV/SpirvInstruction.cpp
```diff
@@ -234,14 +234,16 @@ SpirvDecoration::SpirvDecoration(SourceLocation loc,
...
+                                 llvm::ArrayRef<llvm::StringRef> strParams,
                                  llvm::Optional<uint32_t> idx)
     : SpirvInstruction(IK_Decoration, getDecorateStringOpcode(idx.hasValue()),
                        /*type*/ {}, loc),
       target(targetInst), targetFunction(nullptr), decoration(decor),
       index(idx), params(), idParams() {
-  const auto &stringWords = string::encodeSPIRVString(strParam);
-  params.insert(params.end(), stringWords.begin(), stringWords.end());
+  for (llvm::StringRef str : strParams) {
+    const auto &stringWords = string::encodeSPIRVString(str);
```
jiaolu, line 244: so you want to have with multiple strings input to the ext_decorate_string
vk::ext_decorate_string(5635, "return", "variable", "kk0")]]

and concatenate all the string together at the end .

It looks like syntax sugar to me.

jiaolu (approved):

tools/clang/lib/SPIRV/SpirvInstruction.cpp
jaebaek: This is a minor change because the only decoration I know that we use for `OpDecorateString` is `UserSemantic` and it has only a single parameter, which means your initial implementation looks good for the current SPIR-V spec.

However, I checked [OpDecorateString](https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpDecorateString) and just wanted to make it the same as its actual operands in the spec. It has "Optional Literals See Decoration." as the last operand that specifies zero or multiple string literals at the end.

sudonatalie (approved):

tools/clang/lib/SPIRV/SpirvInstruction.cpp
jiaolu: > This is a minor change because the only decoration I know that we use for `OpDecorateString` is `UserSemantic` and it has only a single parameter, which means your initial implementation looks good for the current SPIR-V spec.
> 
> However, I checked [OpDecorateString](https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpDecorateString) and just wanted to make it the same as its actual operands in the spec. It has "Optional Literals See Decoration." as the last operand that specifies zero or multiple string literals at the end.


The operands here means multiple literals, not multiple strings.

Actually, the spirv specifically says,
"
OpDecorateString (OpDecorateStringGOOGLE)

**Add a string Decoration to another <id>.** 
"
 "Add A string", i guess it means one string.

jaebaek: Oops. Ok. I will doulecheck it and open an issue for it.

PR #4202, opened 2022-01-21
FixIt for HLSL 2021 logical operations && and ||

gracejennings:
Adding FixIts to the diagnostics for short-circuiting logical binary operations && and || on non-scalar values, to suggest logical intrinsics _and()_  and _or()_.

### DIFF
--- a/tools/clang/test/HLSL/vector-or.hlsl
+++ b/tools/clang/test/HLSL/vector-or.hlsl
@@ -1,0 +1,10 @@
+// RUN: %clang_cc1 -HV 2021 -fsyntax-only -ffreestanding -verify %s
+
+RWStructuredBuffer<int3> rw;
+
+[numthreads(1, 1, 1)]
+void main()
+{
+// expected-error@+1 {{operands for short-circuiting logical binary operator must be scalar, for non-scalar types use the logical intrinsic}}
+    rw[0] = rw[0] || rw[0];
+}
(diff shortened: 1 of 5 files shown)


tools/clang/test/HLSL/vector-or.hlsl
```diff
@@ -0,0 +1,10 @@
+// RUN: %clang_cc1 -HV 2021 -fsyntax-only -ffreestanding -verify %s
+
+RWStructuredBuffer<int3> rw;
+
+[numthreads(1, 1, 1)]
+void main()
+{
+// expected-error@+1 {{operands for short-circuiting logical binary operator must be scalar, for non-scalar types use 'or'}}
```
pow2clk, line 8: Can the Fixit component be verified here?

pow2clk (approved):
Looks good! We should do more of these!

Let me know if you think of any other places where this is possible. 

I did make a minor testing suggestion

tools/clang/test/HLSL/vector-or.hlsl
llvm-beanz: I don't think so. We could probably use a FileCheck test for it, but what I'd really like to do is get the rewriter plumbed through so that we could write FileCheck tests that verify the applied FixIt, but that's a bigger change.

llvm-beanz (approved):

tools/clang/test/HLSL/vector-or.hlsl
gracejennings: Does it make sense to add a FileCheck test now or leave it as is and address tests to verify applied Fixits in a later change?

llvm-beanz: I think it is fine as it is

PR #4210, opened 2022-01-26
PIX: Handle zeroInitializer for SM6.6 resource handles.

jeffnn:
Under -Od, a resource handle (passed to a function?) can be zero-initialized.

### DIFF
--- a/lib/DxilPIXPasses/DxilShaderAccessTracking.cpp
+++ b/lib/DxilPIXPasses/DxilShaderAccessTracking.cpp
@@ -24,6 +24,7 @@
 #include "llvm/IR/PassManager.h"
 #include "llvm/Support/FormattedStream.h"
 #include "llvm/Transforms/Utils/Local.h"
+
 #include <deque>
 
 #include "PixPassHelpers.h"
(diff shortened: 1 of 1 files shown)


lib/DxilPIXPasses/DxilShaderAccessTracking.cpp
```diff
@@ -586,38 +592,39 @@ DxilShaderAccessTracking::GetResourceFromHandle(Value *resHandle,
...
-          {
-              drb.HeapIndex = constInt->getLimitedValue();
-          }
-          m_dynamicResourceBindings.emplace_back(std::move(drb));
-
-          return ret;
-      } else {
-          DXASSERT_NOMSG(false);
+      auto* handleCreation = dyn_cast<CallInst>(annotateHandle.get_res());
+      if (handleCreation != nullptr) {
```
jeffnn, line 596: The diff in this block is smaller than github makes it look. The only change is cast->dyn_cast on 595, the if on 596, and the extra closing brace on 627

adam-yang (approved):
LGTM!

PR #4213, opened 2022-01-27
Fix build failure for Linux clang 13.0.0-9+build1

jaebaek:


jaebaek:
I just met build failures caused by the updated clang version.
This commit contains cmake file changes to turn off a few warnings.

Do you have any concerns and suggestions?

In particular, there are so many variables that we set but not actually use.
I turned off `-Wno-unused-but-set-variable`, but I am not sure it is too strong or it would be fine.

In terms of `-Wno-deprecated-copy`, only googletest matters. If we remove `utils/unittest/googletest` (because we have `external/googletest/` as a submodule), I do not think this change is needed.

pow2clk (approved):
Hmm. I'd like to fix these eventually, but let's get it building again for now!

PR #4214, opened 2022-01-28
Correct validation error for col/row of storeoutput

pow2clk:
The validation error was missing the arguments that gave important
information on the problem.


jenatali (approved):
Thanks!

PR #4215, opened 2022-01-28
User/jbatista/dyn uniform indexing hlk00

bob80905:


tools/clang/unittests/HLSL/OptionsTest.cpp
```diff
@@ -349,3 +353,10 @@ TEST_F(OptionsTest, ReadOptionsJoinedWithSpacesThenOK) {
     VERIFY_ARE_EQUAL_STR("foo.dll", o->ExternalLib.data());
   }
 }
+*/
+TEST_F(OptionsTest, ReadOptionsImplicitInput) {
+  const wchar_t *Args[] = {L"/Zpc", L"/Qstrip_debug", L"/Qstrip_priv", L"/O3"};
+  MainArgsArr ArgsArr(Args);
+  std::unique_ptr<DxcOpts> o = ReadOptsTest(ArgsArr, DxcFlags, true, true);
+  VERIFY_ARE_EQUAL_STR("hlsl.hlsl", o->InputFile.data());
+}
```
pow2clk, line 362: I think all these changes were included in error

include/dxc/Support/HLSLOptions.td
```diff
@@ -490,7 +490,7 @@ def setprivate : JoinedOrSeparate<["-", "/"], "setprivate">, Flags<[DriverOption
 def getprivate : JoinedOrSeparate<["-", "/"], "getprivate">, Flags<[DriverOption]>, MetaVarName<"<file>">, Group<hlslutil_Group>,
   HelpText<"Save private data from shader blob">;
 
-def nologo : Flag<["-", "/"], "nologo">, Group<hlslcore_Group>, Flags<[DriverOption]>,
+def nologo : Flag<["-", "/"], "nologo">, Group<hlslcore_Group>, Flags<[DriverOption, HelpHidden]>,
```
pow2clk, line 493: I'm a bit unsure how this is showing up here since this change is already in ToT

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -8361,7 +8361,45 @@ TEST_F(ExecutionTest, DynamicResourcesTest) {
     return;
   }
 
-  RunResourceTest(pDevice, pShader, L"cs_6_6", /*isDynamic*/true);
+  RunResourceTest(pDevice, pShader, L"cs_6_6", /*isDynamic*/false);
```
pow2clk, line 8364: No reason to change this to false. Probably a temp change that didn't get reverted

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -8361,7 +8361,45 @@ TEST_F(ExecutionTest, DynamicResourcesTest) {
...
+      "{"
+      "    Texture2D Input = ResourceDescriptorHeap[InputIndex];" // Dynamic
+      "    float3 color = Input.Sample(PointClamp, IN.Texture).rgb;"
+      "    color = color+1;"
+      "    return float4(color, 1.0f);"
+      "}\n";
+
+  CComPtr<ID3D12Device> pDeviceDynamic;
+  if (!CreateDevice(&pDeviceDynamic, D3D_SHADER_MODEL_6_6))
+    return;
```
pow2clk, line 8386: You should be able to reuse the pDevice already present.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -8361,7 +8361,45 @@ TEST_F(ExecutionTest, DynamicResourcesTest) {
...
+  D3D12_FEATURE_DATA_D3D12_OPTIONS devOptionsDynamic;
+  VERIFY_SUCCEEDED(
+      pDevice->CheckFeatureSupport((D3D12_FEATURE)D3D12_FEATURE_D3D12_OPTIONS, &devOptionsDynamic,
+      sizeof(devOptionsDynamic)));
+  if (devOptionsDynamic.ResourceBindingTier < D3D12_RESOURCE_BINDING_TIER_3) {
+    WEX::Logging::Log::Comment(
+        L"Device does not support Resource Binding Tier 3");
+    WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+    return;
+  }
```
pow2clk, line 8399: This is actually checking the previous device handle, but it shouldn't be necessary.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -8361,7 +8361,45 @@ TEST_F(ExecutionTest, DynamicResourcesTest) {
...
+      "struct VSOutput {"
+      "  float2 Texture : TEXCOORD0;"
+      "};"
+      "float4 PSMain(VSOutput IN) : SV_TARGET"
+      "{"
+      "    Texture2D Input = ResourceDescriptorHeap[InputIndex];" // Dynamic
+      "    float3 color = Input.Sample(PointClamp, IN.Texture).rgb;"
+      "    color = color+1;"
+      "    return float4(color, 1.0f);"
+      "}\n";
```
pow2clk, line 8382: This is the right idea, but we need to match the expected outputs to reuse the verification code. For that, you'll want to model your shader after the `pShader` defined here https://github.com/microsoft/DirectXShaderCompiler/blob/ba2069d67a7045be8719e3a720c30e8a0e5bebcb/tools/clang/unittests/HLSL/ExecutionTest.cpp#L8329

The difference is that instead of the hard-coded literals to index into the `ResourceDescriptorHeap` and `SamplerDescriptorHeap`, you'll want to retrieve those same values from a cbuffer that you create by adding a function similar to `CreateSRV` https://github.com/microsoft/DirectXShaderCompiler/blob/ba2069d67a7045be8719e3a720c30e8a0e5bebcb/tools/clang/unittests/HLSL/ExecutionTest.cpp#L950 but for a constant buffer view (CBV). This page will be helpful: https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-createconstantbufferview

You'll want to populate that CBV with the same indices as are hard coded in `pShader` and use them just as they are in that shader.

You'll need to do the same with parameters to PMain. You could also dump them in `VSOutput` struct. You'll need to give them some annotation, you can use arbitrary annotations. I suggest you get the cbuffer variant done first though.

pow2clk (commented):
I'm not sure what's going on with the title, but it could use some editing. A description would also be helpful. 

The intent here is to use more variable indices to verify them on actual hardware. We'll need to test it on 6.6 capable devices to verify that the test is right.

PR #4216, opened 2022-01-28
[dxil2spv] Construct minimal SPIR-V module

sudonatalie:
Modify SpirvBuilder to allow unset astContext, because astContext is
unavilable when translating from DXIL. Initialize a SpirvContext and
SpirvBuilder to generate a basic module, and output as disassembled
SPIR-V to STDOUT.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvBuilder.cpp
+++ b/tools/clang/lib/SPIRV/SpirvBuilder.cpp
@@ -1639,17 +1639,14 @@ std::vector<uint32_t> SpirvBuilder::takeModule() {
   addModuleInitCallToEntryPoints();
 
   // Run necessary visitor passes first
-  if (astContext) {
     LiteralTypeVisitor literalTypeVisitor(*astContext, context, spirvOptions);
     mod->invokeVisitor(&literalTypeVisitor, true);
-  }
 
   // Propagate NonUniform decorations
   NonUniformVisitor nonUniformVisitor(context, spirvOptions);
   mod->invokeVisitor(&nonUniformVisitor);
 
   // Lower types
-  if (astContext) {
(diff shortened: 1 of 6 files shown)


sudonatalie:
The output currently looks like:

```
$ ./build/bin/dxil2spv shader.ps.ll  
; SPIR-V
; Version: 1.0
; Generator: Google spiregg; 0
; Bound: 1
; Schema: 0
               OpCapability Shader
               OpCapability Linkage
               OpMemoryModel Logical GLSL450
```

Next PR will add FileCheck testing so it's easier to review the progression of each patch.

tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -1633,48 +1639,53 @@ std::vector<uint32_t> SpirvBuilder::takeModule() {
...
-
-  // Generate debug types (if needed)
-  if (spirvOptions.debugInfoRich) {
-    DebugTypeVisitor debugTypeVisitor(astContext, context, spirvOptions, *this,
-                                      lowerTypeVisitor);
-    SortDebugInfoVisitor sortDebugInfoVisitor(context, spirvOptions);
-    mod->invokeVisitor(&debugTypeVisitor);
-    mod->invokeVisitor(&sortDebugInfoVisitor);
+  if (astContext) {
+    LowerTypeVisitor lowerTypeVisitor(*astContext, context, spirvOptions);
```
jaebaek, line 1653: I am worried that this can be confusing.

If we do not set `astContext` for dxil-to-spirv, just returning this method when `astContext == nullptr` (with an error message explaining that this method expects `astContext != nullptr`).

I'd create a separate function for `EmitVisitor` and other visitors for dxil-to-spirv.
e.g., `takeModuleForDxilToSpv()` ??

jaebaek (approved):

PR #4219, opened 2022-02-01
User/jbatista/hlktest dynamic index resource desc01

bob80905:
hlk test now uses a cbuffer to run an additional dynamic indexing test on the resource descriptor


pow2clk (commented):
Looking better. We'll want to test this. At least make sure that the pipeline creation succeeds

PR #4220, opened 2022-02-01
Allow HLSLOptimizationToggles to control lifetime marker.

python3kgae:


tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -346,7 +346,10 @@ void EmitAssemblyHelper::CreatePasses() {
                         !CodeGenOpts.HLSLOptimizationToggles.count("debug-nops") ||
                         CodeGenOpts.HLSLOptimizationToggles.find("debug-nops")->second;
 
-  PMBuilder.HLSLEnableLifetimeMarkers = CodeGenOpts.HLSLEnableLifetimeMarkers;
+  PMBuilder.HLSLEnableLifetimeMarkers =
+      CodeGenOpts.HLSLEnableLifetimeMarkers &
```
tex3d, line 350: Why bitwise `&` instead of logical `&&`?

tex3d (approved):

Issue #2193, opened 2019-05-16
ByteAddressBuffer templated loads/stores with optional alignment argument


sebbbi:
SM 6.2 added templated ByteAddressBuffer loads and templated stores were added in https://github.com/microsoft/DirectXShaderCompiler/pull/2176. With these new additions the usability of raw buffers has increased drastically. Code looks nice:

float4 v = buffer.Load<float4>(idx);
buffer.Store<float4>(idx, v);

There's one remaining problem in DirectX raw buffer: Both the original raw load API and the new templated one use alignment of 4 bytes for all load widths. There's no way to define a custom alignment. This results in poor codegen on some GPUs which require natural alignment (8 or 16 bytes) for 2/4 wide load/store instructions. Seems that their internal compiler chops these wide loads into 2/3/4 individual 32 bit loads, clearly decreasing performance in mem issue bound cases.

I propose a new optional syntax to provide alignment as second template parameter to give compilers more knowledge, allowing them to emit wide load instructions on all architectures:

float4 v = buffer.Load<float4, 16>(idx);
http://buffer.Store <float4, 16>(idx, v);

**Investigation:**

I analyzed the performance of raw loads with my L1$ load benchmark:
https://github.com/sebbbi/perftest 

On AMD GPUs Load2/3/4 instructions only require alignment of 4 bytes, matching DirectX specification. This results in peak throughput for all widths: 
_ByteAddressBuffer.Load random: 22.153ms 1.004x
ByteAddressBuffer.Load2 random: 22.435ms 0.992x
ByteAddressBuffer.Load4 random: 23.508ms 0.947x_

On Nvidia GPUs (Kepler, Maxwell, Pascal), we get clear linear regression with load width:
_ByteAddressBuffer.Load random: 24.336ms 1.479x
ByteAddressBuffer.Load2 random: 48.744ms 0.738x
ByteAddressBuffer.Load4 random: 209.280ms 0.172x_

Structured buffer doesn't show the same regression on Nvidia GPUs, as StructuredBuffer guarantees alignment. Now we get 100% performance with float2 (64 bit) loads. Nvidia has half rate 128 bit loads (including textures). That explains half rate for float4 case. Still much better than raw load case:
StructuredBuffer<float>.Load random: 34.092ms 1.056x
StructuredBuffer<float2>.Load random: 34.125ms 1.055x
StructuredBuffer<float4>.Load random: 67.950ms 0.530x

[...]

PR #4226, opened 2022-02-02
[spirv] support 'T vk::RawBufferLoad<T = uint>(uint64_t address [, uint alignment = 4] )'

jaebaek:


godlikepanos:
I don't fully understand the change but is the 4 byte alignment enough? What if the address points to an array of doubles?

jaebaek:
@yuriy-odonnell-epic Actually I have the same question. I cannot find the alignment rule from the Vulkan spec. Must we assume that it always loads 32-bit data (and 4 bytes alignment) ?

If the size of data to load varies, we have to discuss the memory layout rule.

godlikepanos:
Not an expert but for GLSL the GL_EXT_buffer_reference that enables device_buffer_address has the ability to accept alignment:

```
layout(buffer_reference, scalar, buffer_reference_align = 16) buffer Ref
{
    MyType value;
};
```
More complete example here: https://shader-playground.timjones.io/e4943ae927233add59b42aca85d85845

Since HLSL is a bit more advanced here maybe it could use the base alignment of the typedef to determine the alignment of the OpLoad.

yuriy-odonnell-epic:
Thanks for doing the work on this. 

When it comes to alignment, by default we should be following the exact same rules as apply to `ByteAddressBuffer` in HLSL and the templated load syntax for it (wherever possible).

I imagine that the additional alignment should be provided at call site via SPIR-V intrinsic tag/attribute (via the mechanism that was added not too long ago) or as a separate `LoadRawBufferAligned`. Theoretically, additional layout rules could also be tagged via the intrinsic mechanism. By default the rules should follow _exactly_ what HLSL does for raw buffers.

> What if the address points to an array of doubles?

By default this should translate to just multiple independent loads in the back-end compiler for some GPUs. However, some other GPUs (such as AMD GCN & RDNA) can and do combine loads like that into a wide instructions.

docs/SPIR-V.rst
```diff
@@ -3753,6 +3753,10 @@ implicit ``vk`` namepsace.
   
     uint64_t ReadClock(in uint scope);
     uint     RawBufferLoad(in uint64_t deviceAddress);
+    void     RawBufferLoadToParam(out T output, in uint64_t deviceAddress);
```
yuriy-odonnell-epic, line 3756: One idea for alternative name: `RawBufferLoadInto`.

jaebaek:
I found https://github.com/microsoft/DirectXShaderCompiler/issues/2193 that is the similar alignment issue for `ByteAddressBuffer`.

Initially, I thought loading "raw buffer address" must be similar to `ByteAddressBuffer` that literally support loading "bytes", because the shader language does not have any information about what the "raw buffer address" is and has to rely on the developer's self-correctness-check. Therefore, we assume that it simply loads "4" bytes for whatever types. In addition, we do not set any memory layout and just give freedom to developers (and rely on their own error checks).

I still believe we must give enough freedom for loading data from a random pointer to developers, but I agree it has too much freedom and probably an optional alignment argument will be useful. So .. I want to propose the following intrinsics:

```
void vk::RawBufferLoadInto(out T output, in uint64_t deviceAddress,
                           in uint alignment = 4)

// Defined only when HLSL template support is enabled
template<typename T>
T vk::LoadRawBuffer<T>(uint64_t address, uint alignment = 4) {
        T output;
        RawBufferLoadInto(output, address, alignment);
        return output;
}

// Or
T vk::LoadRawBuffer<T, ALIGNMENT=4>(uint64_t address) {
        T output;
        RawBufferLoadInto(output, address, ALIGNMENT);
        return output;
}
```

Note that if `T` is an aggregate type instead of scalar, because of the lack of memory layout, we cannot guarantee what memory layout will be used i.e., undefined behavior.


Additionally, if developers want a "structured data" load, we have to provide separate intrinsics e.g,
```
// Use the same memory layout rules as the StructuredBuffer to get the alignment.
void vk::RawStructuredBufferLoadInto(out T output, in uint64_t deviceAddress)

// Defined only when HLSL template support is enabled
template<typename T>
T vk::LoadRawStructuredBuffer<T>(uint64_t address) {
        T output;
        RawBufferLoadInto(output, address);
        return output;
}
```

Similarly, we can consider intrinsics for store.

@yuriy-odonnell-epic What do you think?

godlikepanos:
@jaebaek Would it make sense to also support versions that take uint2 instead of uint64_t? Similar to `GL_EXT_buffer_reference_uvec2` extension. Some mobile platforms don't support 64bit integer arithmetic just yet.

jaebaek:
@godlikepanos Good point. I am not versed on the mobile GPU architecture, so I have a question about how it should work i.e., what SPIR-V we have to generate with `uint2`.

```
// 1. Convert uint2 to an address
%address = OpVariable %_ptr_UniformBuffer_v2uint UniformBuffer
...
%ptr = OpBitCast %_ptr_Function_SomeType %address
%value = OpLoad %SomeType %ptr <given alignment>

// 2. Convert first element of uint2 to an address
%address = OpVariable %_ptr_UniformBuffer_v2uint UniformBuffer
...
%load_addr = OpLoad %v2uint %address
%first_elem = OpCompositeExtract %uint %load_addr 0
%ptr = OpBitCast %_ptr_Function_SomeType %first_elem
%value = OpLoad %SomeType %ptr <given alignment>
```

I guess the approach "1. Convert uint2 to an address" would work .. is it right?

godlikepanos:
@jaebaek Good question but I'm not sure. Maybe SPIRV generated from GLSL can give a hint: https://shader-playground.timjones.io/ebf587b72a6aecfe26d0810c154a6626

jaebaek:
Based on the SPIRV generated from GLSL (without optimization),
```
...
%load_vu2_vf4_ = OpFunction %void None %12
    %address = OpFunctionParameter %_ptr_Function_v2uint
          %o = OpFunctionParameter %_ptr_Function_v4float
         %16 = OpLabel
         %17 = OpLoad %v2uint %address
         %20 = OpBitcast %_ptr_PhysicalStorageBuffer__Ref2vec4 %17
         %24 = OpAccessChain %_ptr_PhysicalStorageBuffer_v4float %20 %int_0
         %25 = OpLoad %v4float %24 Aligned 16
               OpStore %o %25
               OpReturn
               OpFunctionEnd
```

"1. Convert uint2 to an address" is the correct approach.
Let me wait for the response from @yuriy-odonnell-epic.
If @yuriy-odonnell-epic agrees with the memory layout and alignment rules, I will first slightly update this PR and submit it.
For the load for "structured data" and the store intrinsics, I will open issues for them, but the implementation will be slightly delayed.

tools/clang/lib/Frontend/InitPreprocessor.cpp
```diff
@@ -1038,7 +1036,14 @@ void clang::InitializePreprocessor(
   // Instruct the preprocessor to skip the preamble.
   PP.setSkipMainFilePreamble(InitOpts.PrecompiledPreambleBytes.first,
                              InitOpts.PrecompiledPreambleBytes.second);
-                          
+
+#ifdef ENABLE_SPIRV_CODEGEN
+  if (LangOpts.SPIRV) {
+    spirv::BuildPredefinesForSPIRV(Predefines,
+                                   PP.getLangOpts().EnableTemplates);
```
jaebaek, line 1043: TODO: When HLSL 2021 is released, we will drop `EnableTemplates`. Use HLSL version instead of it.

yuriy-odonnell-epic:
> T vk::LoadRawBuffer<T, ALIGNMENT=4>

I think I personally slightly prefer this version compared to an alignment argument, as it makes things very clear that alignment must be specified at compile time.

_Except probably vk::RawBufferLoad instead of LoadRawBuffer, following the existing function naming_

Originally, the name "RawBufferLoad" was chosen to match a semantically-similar DXIL op (https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/DXIL.rst#rawbufferload). This op does take an alignment parameter. Though, of course, the SPIR-V intrinsic is not a 100% match to the DXIL op, we could try to follow its rules and semantics as closely as we can (when in doubt).

> we cannot guarantee what memory layout will be used i.e., undefined behavior.
> if developers want a "structured data" load similar to StructuredBuffer, we have to provide separate intrinsics e.g,

Would it be possible to specify the memory layout using a function attribute instead of adding more parameters / overloads / function variants? If yes, then what would be the downsides? I can think of "discoverability" as one, but anything else? Alternatively, `RawBufferLoad<T, ALIGNMENT, LAYOUT>` could potentially also be considered.

In my mind, adding support for a `uint2` address should be a separate PR, as it's independent from templated load or alignment concerns.

jaebaek:
> Would it be possible to specify the memory layout using a function attribute instead of adding more parameters / overloads / function variants? If yes, then what would be the downsides? I can think of "discoverability" as one, but anything else? Alternatively, RawBufferLoad<T, ALIGNMENT, LAYOUT> could potentially also be considered.

Well .. setting a memory layout to the data loaded by `RawBufferLoad<T, ALIGNMENT>` sounds weird to me. It loads "unknown" data from an "arbitrary" address.

If its name is `StructuredBufferLoad<T, ALIGNMENT>` or `BufferLoad<T, ALIGNMENT>` instead of `RawBufferLoad<T, ALIGNMENT>`, supporting memory layout makes sense.

Now we need a discussion about what memory layout rules we must use. Currently, we have std140/std430/DX/scalar [layout rules](https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#memory-layout-rules).
Since it is a complicated issue, I want to first support `RawBufferLoad<T, ALIGNMENT>` without memory layout support. What do you think?

yuriy-odonnell-epic:
I would perhaps suggest also seeking some opinions from other people on a separate function vs extra parameter.

> I want to first support RawBufferLoad<T, ALIGNMENT> without memory layout support. 
Definitely adding the alignment stuff seems like a topic for a separate patch / PR.

docs/SPIR-V.rst
jaebaek: Done

tools/clang/lib/Frontend/InitPreprocessor.cpp
jaebaek: Done.

jaebaek:
I updated this PR based on the discussion with @yuriy-odonnell-epic and @godlikepanos.
Main updates:
1. Add `void vk::RawBufferLoadInto(out T output, in uint64_t address, in uint alignment = 4)`
2. Replace the existing `uint vk::RawBufferLoad(uint64_t addr)` with `T vk::RawBufferLoad<T, alignment>(in uint64_t address)`.
   - Note that I do not set a default alignment for `T vk::RawBufferLoad<T, alignment>(in uint64_t address)`. If I set a default one like `T vk::RawBufferLoad<T, alignment=4>(in uint64_t address)`, we will get a warning message "`warning: default template arguments for a function template are a C++11 extension [-Wc++11-extensions]`" for **all shaders** (not only shaders using this intrinsic function but also all other shaders). It is a limitation of the current template support and the way how we add the predefined macros. I will discuss this issue with Microsoft to figure out if we can consider dropping the warning message.

We have to prepare a few more PRs for the following issues:
1. For mobile GPUs, we will add `void vk::RawBufferLoadInto(out T output, in uint2 address, in uint alignment = 4)` and `T vk::RawBufferLoad<T, alignment>(in uint2 address)`
   - Based on the glslang result, we should directly convert `uint2` to a pointer:
```
         %17 = OpLoad %v2uint %address
         %20 = OpBitcast %_ptr_PhysicalStorageBuffer__Ref2vec4 %17
         %24 = OpAccessChain %_ptr_PhysicalStorageBuffer_v4float %20 %int_0
```
2. We will support loading data with a specified memory layout. Since we need a sophisticated design for the memory layout rules, it may be delayed.

yuriy-odonnell-epic:
@jaebaek I think the inconsistency between RawBufferLoadInto taking alignment as a function argument and RawBufferLoad taking it as a template argument is unfortunate. Is there any it could be unified one way or another?

> Replace the existing uint vk::RawBufferLoad(uint64_t addr)
Does it mean that existing shaders would stop compiling when the PR lands?

jaebaek:
@yuriy-odonnell-epic
> I think the inconsistency between RawBufferLoadInto taking alignment as a function argument and RawBufferLoad taking it as a template argument is unfortunate. Is there any it could be unified one way or another?

Sure, we can let `RawBufferLoad` take a function argument. Unfortunately, as I explained, the purpose of `RawBufferLoadInto` is to support the data load with a random type **without the template support** and we cannot use a template parameter for `RawBufferLoadInto`, which means we have to use a function argument for the alignment of `RawBufferLoadInto`. Since you said you slightly prefer a template parameter for `RawBufferLoad` because it explicitly requires a constant value for the alignment and I agree with the idea, I used a template parameter for it. However, if you think the inconsistency is more critical, definitely we can let `RawBufferLoad` take a function argument for the alignment.

> Does it mean that existing shaders would stop compiling when the PR lands?

Yes. I tried to name the function as `vk::RawBufferLoad` instead of `LoadRawBuffer` based on what you recommended in the previous comment.
> I think I personally slightly prefer this version compared to an alignment argument, as it makes things very clear that alignment must be specified at compile time.
Except probably vk::RawBufferLoad instead of LoadRawBuffer, following the existing function naming
Originally, the name "RawBufferLoad" was chosen to match a semantically-similar DXIL op (https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/DXIL.rst#rawbufferload). This op does take an alignment parameter. Though, of course, the SPIR-V intrinsic is not a 100% match to the DXIL op, we could try to follow its rules and semantics as closely as we can (when in doubt).

We cannot support both `uint vk::RawBufferLoad(uint64_t address)` and `T vk::RawBufferLoad<T>(uint64_t address)` because they have the same name and the same function parameter.

If you want to keep `uint vk::RawBufferLoad(uint64_t address)`, we have to change the name of `T vk::RawBufferLoad<T>(uint64_t address)` e.g., `T vk::LoadRawBuffer<T>(uint64_t address)`.

yuriy-odonnell-epic:
Potentially `RawBufferLoadInto` should just not exist at all then and the templated load syntax should just require template language features.

I would not recommend having inconsistent style in how alignment is specified in different functions. I would not recommend having functions with very similar behavior but slightly different names (i.e. LoadRawBuffer and RawBufferLoad).

jaebaek:
I spent a long time to understand the front-end of DXC and in particular, the template creation/deduction.

Now this PR allows us to use the previous API i.e., `uint vk::RawBufferLoad(uint64_t address)` and support `T vk::RawBufferLoad<T = uint>(uint64_t address, [uint alignment])` that satisfies

- loading data with an arbitrary type
- loading data with an alignment
- the same builtin function name (`RawBufferLoad`)
- compatibility

PR #4230, opened 2022-02-03
Support -setprivate, -setrootsignature, and -Qstrip_priv to compile API

gracejennings:
When you call the compiler from the command line you are able to provide a source with a serialized root signature and a source with private data using the `-setrootsignature` and `-setprivate` options. This change creates the same behavior when you call compile from the API. File names are added to the arguments along with the options and root signature/private data are added through the include handler. The `-Qstrip_priv` option was also added to the compile API calls with a warning if you try to set and strip private data in the same compilation. 

Adding the root signature part in the initial serialization allows for root signature validation. The private data is added in a later serialization because there is no such validation for private data.


tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1218,6 +1238,51 @@ class DxcCompiler : public IDxcCompiler3,
...
       } // Write PDB
 
+      bool updatePartRequired =
+          !opts.PrivateSource.empty() || opts.StripPrivate;
+
+      if (!hasErrorOccurred && updatePartRequired) {
+
+        CComPtr<IDxcContainerBuilder> pContainerBuilder;
+        DxcCreateInstance2(this->m_pMalloc, CLSID_DxcContainerBuilder,
+                           IID_PPV_ARGS(&pContainerBuilder));
```
tex3d, line 1248: I believe the parts should be added/omitted inside `SerializeDxilContainerForModule`.  Which parts to include in the container is communicated through the `SerializeDxilFlags`.  That will avoid serialization of the container, then rebuilding of the container here, which was part of the point of moving these into the API.  Private Data will also have to be added to the `dxcutil::AssembleInputs` and `SerializeDxilContainerForModule` arguments.  If root signature is supplied from above, it would have to override the one being serialized to pRootSigStream and placed into the container (unless being stripped at the same time).

tex3d: Quick followup: I see you're resetting the root signature in the module when there's an incoming one, so that's ok, and means it doesn't need to be passed down to `SerializeDxilContainerForModule`.  However, we should add the private part while constructing the original container still, rather than modifying it and reserializing it after.

include/dxc/DxilContainer/DxilContainer.h
```diff
@@ -581,6 +581,7 @@ enum class SerializeDxilFlags : uint32_t {
   StripReflectionFromDxilPart = 1 << 3, // Strip Reflection info from DXIL part.
   IncludeReflectionPart       = 1 << 4, // Include reflection in STAT part.
   StripRootSignature          = 1 << 5, // Strip Root Signature from main shader container.
+  SetRootSignature            = 1 << 6, // Set Root Signature in the container from provided source.
```
tex3d, line 584: I don't think we need this new flag do we?  We include the root signature during container assembly unless StripRootSignature is provided.

tools/clang/tools/dxcompiler/dxcutil.cpp
```diff
@@ -108,6 +111,9 @@ void GetValidatorVersion(unsigned *pMajor, unsigned *pMinor) {
 void AssembleToContainer(AssembleInputs &inputs) {
   CComPtr<AbstractMemoryStream> pContainerStream;
   IFT(CreateMemoryStream(inputs.pMalloc, &pContainerStream));
+  if (inputs.SerializeFlags & SerializeDxilFlags::SetRootSignature) {
```
tex3d, line 114: If removing SetRootSignature flag, this can change to:
```cpp
if (!(inputs.SerializeFlags & SerializeDxilFlags::StripRootSignature) && inputs.pRootSigBlob) {
```

include/dxc/DxilContainer/DxilContainer.h
gracejennings: Ya you are right, the same thing can be accomplished checking the inputs.pRootSigBlob

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
gracejennings: Ok yes that makes sense, I had discussed adding the private data in the initial serialization with Xiang but he had recommended I try in a reserialization since there is no validation. But it does makes sense to move the setprivate to avoid the reserialization, I'll go ahead and change that.

pow2clk (approved):
Looks good to me. I do suggest removing the unused variable.

Feel free to wait for Tex to sign off if you like.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1770,6 +1770,16 @@ void hlsl::SerializeDxilContainerForModule(DxilModule *pModule,
     }
   }
 
+  if (pPrivateData) {
+    CComPtr<AbstractMemoryStream> pPrivateStream;
```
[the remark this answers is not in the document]
gracejennings: Ya , you're right, I'll take it out!

tex3d (approved):
Looks good, but you'll have to wait for #4239 to go in before your test will pass.

PR #4232, opened 2022-02-04
Globally disable NRVO for HLSL; Merge resprops for resource return

tex3d:
NRVO can drop type attributes not captured into QualType, like glc.  HLSL does not need NRVO, and it only poses problems in various areas, so this change turns it off globally for HLSL.

The test required an additional fix to resource return resprops merging, provided by @python3kgae.


llvm-beanz (approved):
Looks good to me. Can you please make sure the description reflects the merged-in change since that is the majority of the change here?

Issue #4146, opened 2021-12-14
__HLSL_VERSION is defined to 2018 even if the flag "-HV 2021" is passed to the compiler


zigguratvertigo:
Hi,

We are compiling shaders with `-HV 2021`. Checking the value of `__HLSL_VERSION` in shaders, the value seems to be still 2018 instead of 2021 [as per the documentation](https://github.com/microsoft/DirectXShaderCompiler/wiki/Predefined-Version-Macros#hlsl-language-version-define). 

This issue can be verified by simply writing the define in a shader and letting the compiler output its value as an error.

In HLSL:
```
__HLSL_VERSION
```

Compilation: 
```
error: expected unqualified-id
  2018
  ^
```

zigguratvertigo:
In the event where someone uses both the preprocessor and the compiler, the issue can be resolved by making sure "-HV 2021" is passed to both.

zigguratvertigo:
Is this expected behaviour @tex3d ?

llvm-beanz:
This doesn't sound like a bug to me. `__HLSL_VERSION` is a preprocessor define, so it is only set by the preprocessor stage. If you don't set `-HV 2021` for the preprocessor call, the preprocessor runs in the 2018 language mode (which is the default language mode). By the time you get to the compile all the places that you used `__HLSL_VERSION` are already expanded, so there is nothing the compiler can do to alter that.

PR #4234, opened 2022-02-05
Pass compiler options to preprocessor

llvm-beanz:
If we don't pass the compiler options to the preprocessor the
preprocessor doesn't get initialized correctly.

Without this change the preprocessor step identifies the shader model
version as 0.0, pipeline stage as invalid, and language version as the
default.


tools/clang/tools/dxclib/dxc.cpp
```diff
@@ -924,7 +924,14 @@ void DxcContext::Preprocess() {
...
   CComPtr<IDxcOperationResult> pPreprocessResult;
   CComPtr<IDxcBlobEncoding> pSource;
+  
+  std::vector<std::wstring> argStrings;
+  CopyArgsToWStrings(m_Opts.Args, CoreOption, argStrings);
+
   std::vector<LPCWSTR> args;
+  args.reserve(argStrings.size());
+  for (const std::wstring &a : argStrings)
+    args.push_back(a.data());
```
pow2clk, line 934: There are a couple places below where `m_Opts.Args` are explicitly added to the preprocessor `args`. They would now add duplicates.

pow2clk (commented):
I'd like to avoid the duplicate arguments.

I can't come up with any reason why including all the arguments for the preprocessor would cause a problem, but I also don't know why they were excluded in the first place. Maybe @Tex3d does?

llvm-beanz:
I was also curious if there was a reason for not passing all arguments through. I think if there is, we should filter out specific arguments rather than only passing along specific ones.

Running the preprocessor with different arguments than the compile call can (and does) result in unexpected behavior differences when preprocessing as a distinct step.

Having an accurate preprocess-only step is a precursor to our compiler being used by object caching and compilation distribution tools like ccache, sccache, distcc and incredibuild.

This also might be related to the issue reported in #4146, which I closed because the comment made it sound like user error that was corrected.

pow2clk:
I was trying to find #4146, but couldn't because it got closed. I don't want to give you busy work, but I wonder if we should add a test using the API interface as well

llvm-beanz:
> I was trying to find #4146, but couldn't because it got closed. I don't want to give you busy work, but I wonder if we should add a test using the API interface as well

If you think it is worth the extra test I can add it, but since DXC calls the same `IDxcCompiler::Preprocess` method, I wasn't sure it was worth it.

llvm-beanz:
Discussed with Tex offline, he was unsure of any reason why we needed to filter the options for the preprocessor. If something pops up we can adjust this, but let's just go with this for now :)

PR #4237, opened 2022-02-07
Fix DxilProgramSignature size alignment.

### LINKED ISSUE #4236: DXIL container signatures require potentially-unaligned data accesses
This is a regression from FXC to DXC. The FXC container assembler produced 32-bit-aligned container sub-parts for every part. However, DXC, and by extension the DXIL validator, can produce 8-bit-aligned container parts, specifically for the I/O signatures, where the semantic string table is written without any padding.

Any component that wishes to produce valid DXIL which can be signed by the validator must adhere to this tight alignment. This can be problematic for handling DXIL containers on some architectures, such as ARM32. I've seen instances where the ARM compiler assumes alignment and generates 64-bit accesses, which require 32-bit data alignment, resulting in unhandled exceptions due to alignment violation when reading this data out of a DXIL container.

tex3d:
- Maintain compatibility for validator version < 1.7.
- DxilContainerWriter_impl verifies size alignment for all parts, unless created with bUnaligned=true, used for backward compatibility.
- DxilValidation uses NewProgramSignatureWriter(), so bUnaligned is based on validator version in module.

Fixes #4236

### DIFF
--- a/include/dxc/DxilContainer/DxilContainerAssembler.h
+++ b/include/dxc/DxilContainer/DxilContainerAssembler.h
@@ -51,7 +51,8 @@ DxilPartWriter *NewFeatureInfoWriter(const DxilModule &M);
 DxilPartWriter *NewPSVWriter(const DxilModule &M, uint32_t PSVVersion = UINT_MAX);
 DxilPartWriter *NewRDATWriter(const DxilModule &M);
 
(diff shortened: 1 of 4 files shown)


pow2clk (approved):
Looks good!

PR #4239, opened 2022-02-08
Allow unaligned private data size, by keeping it at end of container.

tex3d:
Since all other blob parts are expected to be aligned, make sure we always add private data at the end of the container.
Update DxcContainerBuilder to keep private data at the end as well.

### DIFF
--- a/lib/DxilContainer/DxcContainerBuilder.cpp
+++ b/lib/DxilContainer/DxcContainerBuilder.cpp
@@ -86,6 +86,9 @@ HRESULT STDMETHODCALLTYPE DxcContainerBuilder::RemovePart(_In_ UINT32 fourCC) {
         [&](DxilPart part) { return part.m_fourCC == fourCC; });
     IFTBOOL(it != m_parts.end(), DXC_E_MISSING_PART);
     m_parts.erase(it);
+    if (fourCC == DxilFourCC::DFCC_PrivateData) {
+      m_HasPrivateData = false;
+    }
     return S_OK;
   }
   CATCH_CPP_RETURN_HRESULT();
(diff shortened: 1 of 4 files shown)


lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -208,4 +200,20 @@ HRESULT DxcContainerBuilder::UpdateParts(AbstractMemoryStream *pStream) {
...
+}
+
+void DxcContainerBuilder::AddPart(DxilPart&& part) {
+  PartList::iterator it = std::find_if(m_parts.begin(), m_parts.end(), [&](DxilPart checkPart) {
+    return checkPart.m_fourCC == part.m_fourCC;
+  });
+  IFTBOOL(it == m_parts.end(), DXC_E_DUPLICATE_PART);
+  if (m_HasPrivateData) {
+    // Keep PrivateData at end, since it may have unaligned size.
+    m_parts.insert(&m_parts.back(), std::move(part));
```
jenatali, line 212: That's not standard C++, is it? Perhaps `.end() - 1` instead of `.back()`?

jenatali (approved):

lib/DxilContainer/DxcContainerBuilder.cpp
tex3d: Since m_parts is a SmallVector, iterator is actually just `typedef T *iterator;` where T is the element type, so an address to an element of the vector is fine as an iterator here.  But `end() - 1` is fine too of course.

pow2clk (approved):
Looks good!

Sorry about missing support for ContainerBuilder. I'll add that to the list :)

PR #4241, opened 2022-02-08
[linux] Fix incorrect REGDB_E_CLASSNOTREG value

tex3d:
REGDB_E_CLASSNOTREG was defined to 1, which is a success code for HRESULT.
This value is returned when trying to create an unsupported object with DxcCreateInstance.
Since create appears to have succeeded, it would lead to a deref of a null pointer - segfault.


llvm-beanz (approved):

pow2clk:
Incidentally, this was part of my change here #4244, which still needs a review 😄

PR #4242, opened 2022-02-08
Rename UTF16 to Wide

pow2clk:
In many cases, systems where wchar was represented as 32-bits were being
referred to by types and variables affixed with "utf16" or similar
consistent with what they are on windows. To avoid confusion, this
renames all such places to "wide".

In a few cases, conversions were explicitly 16-bit, which wasn't the
intent. They have been converted to platform-specific wide char
conversions

### DIFF
--- a/include/dxc/Support/DxcLangExtensionsCommonHelper.h
+++ b/include/dxc/Support/DxcLangExtensionsCommonHelper.h
@@ -39,7 +39,7 @@ private:
     try {
       IFTPTR(name);
       std::string s;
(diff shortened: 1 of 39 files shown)


pow2clk:
Missed out some Windows fixups, plus rebasing required a few

pow2clk:
Tagging @Jaebaek as this pertains to his interests

jaebaek (approved):
Wow this is a huge change. It becomes much more clear now. Thanks!
I tried to review it, but probably I missed some issues. I want to see @tex3d 's reviews.

PR #4244, opened 2022-02-09
Enable Container Builder on *nix

pow2clk:
This never got enabled in dxcapi.cpp in spite of all the code to support
it being in place. This enables the creation of the interface and
enables the tests that were disabled for want of it.

As an incidental, correctly defines the REGDB_E_CLASSNOTREG which made
the lack of this interface on Linux hard to diagnose.

### DIFF
--- a/include/dxc/Support/WinAdapter.h
+++ b/include/dxc/Support/WinAdapter.h
@@ -88,13 +88,12 @@
 #define FALSE 0
 #define TRUE 1
 
-#define REGDB_E_CLASSNOTREG 1
-
 // We ignore the code page completely on Linux.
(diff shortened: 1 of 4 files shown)


adam-yang:
There's a test for PdbUtils in CompilerTest.cpp that's also disabled on anything but windows, can we enable that too?

tex3d (approved):
LGTM!

pow2clk:
In response to @adam-yang's comment,

Everything is on the table, but I don't think enabling that support is quite as trivial. Given how much testing is missing behind it, my next goal is to enable reflection, but that's blocked behind another pending change and also integrating the DirectX-Headers submodule

PR #4245, opened 2022-02-09
Fix ReplaceMemcpy IR corruption on constant user replacement.

tex3d:
When replacing constant users, those users could be in functions other than the one being worked on.
The code would happily replace uses of the constant with instructions that weren't in the same function.
This would lead to bizarre consequences due to corrupt IR down the line.
This change prevents that from happenning.

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -3215,9 +3215,12 @@ bool SROA_Helper::DoScalarReplacement(GlobalVariable *GV,
 }
 
 static void ReplaceConstantWithInst(Constant *C, Value *V, IRBuilder<> &Builder) {
+  Function *F = Builder.GetInsertBlock()->getParent();
   for (auto it = C->user_begin(); it != C->user_end(); ) {
     User *U = *(it++);
     if (Instruction *I = dyn_cast<Instruction>(U)) {
+      if (I->getParent()->getParent() != F)
+        continue;
       I->replaceUsesOfWith(C, V);
     } else {
       // Skip unused ConstantExpr.


dmpots:
Can we get a test case?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3215,9 +3215,12 @@ bool SROA_Helper::DoScalarReplacement(GlobalVariable *GV,
 }
 
 static void ReplaceConstantWithInst(Constant *C, Value *V, IRBuilder<> &Builder) {
+  Function *F = Builder.GetInsertBlock()->getParent();
```
dmpots, line 3218: Should we be grabbing the function from the value V (in the case where V is an instruction). There is no obvious guarantee from this context that V was created by builder at that location, but maybe it is enforced by the caller? 

Seems to be a potentially fragile assumption.

tex3d: The builder assumption must be true, it's what's used to create instruction replacements for constant expressions, like GEP.  It's based on the location of the memcpy instruction being replaced.

tex3d: For extra clarity: We are attempting to replace a memcpy instruction.  Any replacements should only occur within the same function as that memcpy instruction.  The builder is based on that memcpy instruction location.  If V is constant, I still don't think we should be replacing users in functions other than the one with the memcpy instruction being replaced.  If we conditionally skipped only when V was an instruction and in a different function, we would replace users of constants in other functions.  I'm just not positive that constant user replacement in other functions is what we want here.

dmpots (approved):
Would be great to include a test case here.

PR #4246, opened 2022-02-09
Taught DxilEraseDeadRegion to handle more general loops.

adam-yang:
- Replaced the self-loop code with simplifying more general loops.
- Made some safety checks more general.
- Changed some tests to explicitly use values produced by loops that don't get used so the loops don't get deleted.


bfavela (approved):
LGTM!

tex3d (approved):
LGTM!

PR #4248, opened 2022-02-10
[spirv] Append template params to UserTypeGOOGLE

sudonatalie:


tools/clang/lib/SPIRV/AstTypeProbe.cpp
```diff
@@ -1061,7 +1061,7 @@ std::string getHlslResourceTypeName(QualType type) {
         name == "RWTexture2DArray" || name == "Buffer" || name == "RWBuffer" ||
         name == "SubpassInput" || name == "SubpassInputMS" ||
         name == "InputPatch" || name == "OutputPatch") {
-      return name;
+      return type.getLocalUnqualifiedType().getAsString();
```
jaebaek, line 1064: Could you please add a comment for this to explain that `getLocalUnqualifiedType()` uses `getCommonPtr()->BaseType` which causes segfault when `type.isNull()` is true but we return this method for that case in the above line?

jaebaek (approved):
Could you please briefly explain this change in the SPIR-V code gen document?

sudonatalie:
> Could you please briefly explain this change in the SPIR-V code gen document?

Done

tools/clang/lib/SPIRV/AstTypeProbe.cpp
jaebaek: I suspected that `type.getLocalUnqualifiedType()` can crash inside and realized that the only case it has crashes is when it is null. I just wanted to clarify it for readers like me. Obviously, it is not null at this point.

I still want to explain calling `type.getLocalUnqualifiedType()` will be fine at this point.

PR #4250, opened 2022-02-11
Uniquely detect low 16-bit usage properly for tbuffer

tex3d:
Uniquely detect low 16-bit usage properly for tbuffer, while detecting the case where full 32-bits are used that map to two 16-bit fields, requiring both fields to be marked.

### DIFF
--- a/lib/HLSL/DxilCondenseResources.cpp
+++ b/lib/HLSL/DxilCondenseResources.cpp
@@ -2529,11 +2529,15 @@ static unsigned GetCBOffset(Value *V, OffsetForValueMap &visited) {
 
 typedef std::map<unsigned, DxilFieldAnnotation*> FieldAnnotationByOffsetMap;
 
-static void MarkCBUse(unsigned offset, FieldAnnotationByOffsetMap &fieldMap) {
+// Returns size in bits of the field if it's a basic type, otherwise 0.
+static unsigned MarkCBUse(unsigned offset, FieldAnnotationByOffsetMap &fieldMap) {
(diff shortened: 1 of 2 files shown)


lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -2585,9 +2589,17 @@ static void MarkCBUsesForExtractElement(
...
+  // If field marked is < 32 bits, we still need to mark the high 16-bits as
+  // used, in case there is another 16-bit field.
+  // Since MarkCBUse could return 0 on non-basic type field, look for 16
+  // when determining whether we still need to mark high word as used.
+  bool highUnmarked = EV->getType()->getScalarSizeInBits() == 32;
   if (!lowHighWordUsage || 0 != (lowHighWordUsage & kLowWordUsed))
-    MarkCBUse(offset + evOffset, fieldMap);
-  if (lowHighWordUsage & kHighWordUsed)
+    highUnmarked &= MarkCBUse(offset + evOffset, fieldMap) == 16;
+  if (highUnmarked && (!lowHighWordUsage || 0 != (lowHighWordUsage & kHighWordUsed)))
```
python3kgae, line 2602: For !lowHighWordUsage || 0 != (lowHighWordUsage & kHighWordUsed)), could it just be !lowHighWordUsage?

python3kgae (approved):

lib/HLSL/DxilCondenseResources.cpp
tex3d: They're not equivalent.

PR #4254, opened 2022-02-12
PIX: avoid crash on enumerated types without int base type

jeffnn:
HLSL 2017's enum and enum class compile to a type without an integer base.
So in this case we'll call it an unknown type for the purposes of PIX debug instrumentation.
Further work will be required to return the enumerants to PIX and to record an actual value during shader debug instrumentation.


adam-yang (approved):
LGTM!

PR #4257, opened 2022-02-14
PIX: Cherry-pick: DS/HS debugging, enum types crash fix

jeffnn:
Cherry-pick from master:
PIX: avoid crash on enumerated types without int base type (#4254)
PIX: Shader debugging instrumentation for HS+DS (#4251)


hekota (approved):
LGTM!

PR #4258, opened 2022-02-14
[spirv] Generate correct Shader.DebugInfo for runtime arrays.

greg-lunarg:
This is a common construct for bindless shaders.

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/shader.debug.runtimearray.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/shader.debug.runtimearray.hlsl
@@ -1,0 +1,18 @@
+// RUN: %dxc -T ps_6_0 -E main -fspv-debug=vulkan
+
+struct PSInput
+{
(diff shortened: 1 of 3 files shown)


jaebaek (approved):
Thank you for your contribution!

tools/clang/test/CodeGenSPIRV/shader.debug.runtimearray.hlsl
```diff
@@ -0,0 +1,18 @@
+// Run: %dxc -T ps_6_0 -E main -fspv-debug=vulkan
```
jaebaek, line 1: Please use `// RUN:` (upper cases) instead of "// Run:`.

greg-lunarg: Oh yes. Thanks for the reminder! I was copying from an old version of the tree!

PR #4271, opened 2022-02-16
[dxil2spv] Add stage IO variables

sudonatalie:


tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -607,6 +607,9 @@ class SpirvBuilder {
   SpirvVariable *addStageIOVar(QualType type, spv::StorageClass storageClass,
                                std::string name, bool isPrecise,
                                SourceLocation loc);
+  SpirvVariable *addStageIOVar(const SpirvType *type,
+                               spv::StorageClass storageClass, std::string name,
```
jaebaek, line 611: nit:
I have no idea why we initially pass `std::string name`, but it can slightly consume more memory.
Could you please change it to `llvm::StringRef name` or `const std::string &name`?

jaebaek (approved):

tools/clang/include/clang/SPIRV/SpirvBuilder.h
sudonatalie: I'll do this in a follow up so I can replace all uses in this class without cluttering up this PR.

PR #4272, opened 2022-02-17
Delete dbg.declare in scopes where the alloca is unused

adam-yang:
Delete all dbg.declare instructions for allocas that are never touched in scopes. This could greatly improve compilation speed and binary size at the cost of in-scope but unused (part of) variables being invisible in certain functions. For example:
```
struct Context {
   float a, b, c;
};
void bar(inout Context ctx) {
  ctx.b = ctx.a * 2;
}
void foo(inout Context ctx) {
  ctx.a = 10;
  bar(ctx);
}

float main() : SV_Target {
  Context ctx = (Context)0;
  foo(ctx);
  return ctx.a + ctx.b + ctx.c;
}
```
Before running this, shader would generate dbg.declare for members `a`, `b`, and `c` for variable `ctx` in every scope (`main`, `foo`, and `bar`). In the call stack with `foo` and `bar`, member `c` is never used in any way, so it's a waste to generate dbg.declare for member `c` for the `ctx` in scope `foo` and scope `bar`, so they can be removed.

### DIFF
--- a/include/dxc/DXIL/DxilUtil.h
+++ b/include/dxc/DXIL/DxilUtil.h
@@ -162,6 +162,10 @@ namespace dxilutil {
   /// This can enhance SROA and other transforms that want type-safe pointers,
   /// and enables merging with other getelementptr's.
   llvm::Value *TryReplaceBaseCastWithGep(llvm::Value *V);
+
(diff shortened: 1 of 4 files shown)


tex3d (approved):
Great!

PR #4276, opened 2022-02-18
Only set ResMayNotAlias flag when on DXIL 1.7

tex3d:
Avoid failure in existing driver on unrecognized raw flags.


llvm-beanz (approved):
LGTM

PR #4277, opened 2022-02-18
ExecutionTest: Autofill Path/Version for AgilitySDK

tex3d:
Default D3D12SDKPath to ".\D3D12\"
Autodetect version when not specified or "/p:D3DSDKVersion=1"

If D3DSDKVersion specified, fail if it can't use Agility SDK.

### DIFF
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -1370,7 +1370,7 @@ public:
       // this case shouldn't be hit, fail if it is.
       return HRESULT_FROM_WIN32(GetLastError());
     }
-    return S_OK; //pD3D12EnableExperimentalFeatures(0, nullptr, nullptr, nullptr);
+    return pD3D12EnableExperimentalFeatures(0, nullptr, nullptr, nullptr);
   }
 
   static HRESULT EnableExperimentalShaderModels(HMODULE hRuntime) {
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -52,6 +52,7 @@
 #include <d3dcompiler.h>
 #include <wincodec.h>
 #include "ShaderOpTest.h"
+#include <libloaderapi.h>
 
 #pragma comment(lib, "d3dcompiler.lib")
 #pragma comment(lib, "windowscodecs.lib")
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -1286,76 +1311,153 @@ class ExecutionTest {
...
-      IFR(pD3D12SDKConfiguration->SetSDKVersion(SDKVersion, SDKPath));
+  static UINT GetD3D12SDKVersion(std::wstring SDKPath) {
+    // Try to automatically get the D3D12SDKVersion from the DLL
+    UINT SDKVersion = 0;
+    std::wstring D3DCorePath = ComputeSDKFullPath(SDKPath);
+    D3DCorePath.append(L"D3D12Core.dll");
+    HMODULE hCore = LoadLibraryW(D3DCorePath.c_str());
+    if (hCore) {
+      if (UINT *pSDKVersion = (UINT*)GetProcAddress(hCore, "D3D12SDKVersion"))
+        SDKVersion = *pSDKVersion;
```
hekota, line 1341: Missing () after *pSDKVersion?

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -1286,76 +1311,153 @@ class ExecutionTest {
...
+    D3D12EnableExperimentalFeaturesFn pD3D12EnableExperimentalFeatures =
+        (D3D12EnableExperimentalFeaturesFn)GetProcAddress(
+            hRuntime, "D3D12EnableExperimentalFeatures");
+    if (pD3D12EnableExperimentalFeatures == nullptr) {
+      // If this failed, D3D12 must be too old for AgilitySDK.  But if that's
+      // the case, creating D3D12SDKConfiguration should have failed.  So while
+      // this case shouldn't be hit, fail if it is.
+      return HRESULT_FROM_WIN32(GetLastError());
     }
+    return S_OK; //pD3D12EnableExperimentalFeatures(0, nullptr, nullptr, nullptr);
```
hekota, line 1372: Maybe remove the comment after return.

tools/clang/unittests/HLSL/ExecutionTest.cpp
tex3d: The export is a UINT*, not a function to be called.

tools/clang/unittests/HLSL/ExecutionTest.cpp
tex3d: Oops, this should have called the API.  I had an issue and thought it was because of this call, but in reality it was a different issue.  Apparently, I forgot to uncomment the call!  Good catch!

I'll re-test this merged with latest master to be sure, then update this PR.

tools/clang/unittests/HLSL/ExecutionTest.cpp
hekota: Oh I see :)

tools/clang/unittests/HLSL/ExecutionTest.cpp
tex3d: Fixed.

hekota (approved):
LGTM!

PR #4279, opened 2022-02-18
ShaderOpTest: Use DXC to compile root signature instead of FXC

tex3d:


tex3d:
> how is that a faithful translation

The original was using the macros to pass the root signature text in by defining `RootSigVal` to it, but then also passed the same string in as the entry point name, which was weird.  I guess that worked, but that all seems unnecessary.

If you define main to the string in the shader contents, then the default entry point of "main" selects that as the root signature to use.

If we tried to set it through the defines array, we'd have to convert the string to a wide string as well.  Same with providing that string as the entry name.  I think this way is better.

Anyway, it doesn't really matter, as long as it works through the API.

PR #4281, opened 2022-02-19
ShaderOpTest: Add sampler parsing and desc support

tex3d:


tools/clang/unittests/HLSL/ShaderOpTest.cpp
```diff
@@ -1795,6 +1884,32 @@ void ShaderOpParser::ParseDescriptor(IXmlReader *pReader, ShaderOpDescriptor *pD
       CHECK_HR(ReadAttrUINT(pReader, L"WSize", &pDesc->UavDesc.Texture3D.WSize));
       break;
     }
+  } else if (isCBV) {
```
bob80905, line 1887: Is there nothing to do in this case? or is there more to fill out here? isCBV case.

tex3d: Correct.

PR #4285, opened 2022-02-22
ExecutionTest Programmable Sample Offsets

pow2clk:
A texture containing an encoded value representing the X,Y location is
used to retrieve values at various samples using the 8 methods that
newly allow programmable values. Each result is compared to what is
expected based on the given coords and offsets.

### DIFF
--- a/tools/clang/test/HLSL/ShaderOpArith.xml
+++ b/tools/clang/test/HLSL/ShaderOpArith.xml
@@ -662,7 +662,8 @@
 
         uint4 DoTests( int2 coord, int2 offset ) {
 #if  __SHADER_TARGET_MAJOR > 6 || (__SHADER_TARGET_MAJOR == 6 && __SHADER_TARGET_MINOR >= 7)
-          float cmp = (coord.y + offset.y)*1000 + coord.x + offset.x;
+          int2 coffset = ClampOffset(offset);
+          float cmp = (coord.y + coffset.y)*1000 + coord.x + coffset.x;
           float2 fcoord = (coord + 0.5)/1000.0;
           return uint4( g_tex.Sample(g_samp, fcoord, offset),
                         g_tex.SampleCmp(g_sampCmp, fcoord, cmp, offset),
--- a/tools/clang/test/HLSL/ShaderOpArith.xml
+++ b/tools/clang/test/HLSL/ShaderOpArith.xml
@@ -581,6 +581,361 @@
       ]]>
     </Shader>
   </ShaderOp>
+  <ShaderOp Name="SampleCmpLevel" PS="PS" VS="VS" CS="CS" AS="AS" MS="MS" TopologyType="TRIANGLE">
+    <RootSignature>
+      RootFlags(ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT),
+      DescriptorTable(SRV(t0,numDescriptors=1), UAV(u0), UAV(u1), UAV(u2)),
+      StaticSampler(s0, addressU = TEXTURE_ADDRESS_WRAP, addressV = TEXTURE_ADDRESS_WRAP, filter = FILTER_MIN_MAG_LINEAR_MIP_POINT),
+      StaticSampler(s1, addressU = TEXTURE_ADDRESS_WRAP, addressV = TEXTURE_ADDRESS_WRAP, filter = FILTER_COMPARISON_MIN_MAG_LINEAR_MIP_POINT)
+    </RootSignature>
+    <Resource Name="VBuffer" Dimension="BUFFER" InitialResourceState="COPY_DEST" Init="FromBytes" Topology="TRIANGLELIST">
+      { { -1.0f, 1.0f, 0.0f }, { 0.0f, 0.0f } },
+      { { 1.0f, 1.0f, 0.0f }, { 1.0f, 0.0f } },
+      { { -1.0f, -1.0f, 0.0f }, { 0.0f, 1.0f } },
+
+      { { -1.0f, -1.0f, 0.0f }, { 0.0f, 1.0f } },
+      { { 1.0f, 1.0f, 0.0f }, { 1.0f, 0.0f } },
+      { { 1.0f, -1.0f, 0.0f }, { 1.0f, 1.0f } }
+    </Resource>
+    <Resource Name="T0" Dimension="Texture2D" Width="336" Height="336" MipLevels="7" InitialResourceState="COPY_DEST" Init="ByName" Format="R32_FLOAT" />
+    <Resource Name="RTarget" Dimension="TEXTURE2D" Width="84" Height="4" Format="R32G32B32A32_FLOAT" Flags="ALLOW_RENDER_TARGET" InitialResourceState="COPY_DEST" />
+    <Resource Name="U0" Dimension="BUFFER" Width="8192"
+              Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST"
+              Init="Zero" ReadBack="true" />
+    <Resource Name="U1" Dimension="BUFFER" Width="1024"
+              Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST"
+              Init="Zero" ReadBack="true" />
+    <Resource Name="U2" Dimension="BUFFER" Width="1024"
+              Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST"
+              Init="Zero" ReadBack="true" />
+
+    <RootValues>
+      <RootValue HeapName="ResHeap" />
+    </RootValues>
+    <DescriptorHeap Name="ResHeap" Type="CBV_SRV_UAV">
+      <Descriptor Name='T0' Kind='SRV' ResName='T0' />
+      <Descriptor Name='U0' Kind='UAV' ResName='U0'
+                  NumElements="336" StructureByteStride="8" />
+      <Descriptor Name='U1' Kind='UAV' ResName='U1'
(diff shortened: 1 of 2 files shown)


tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -581,6 +581,361 @@
       ]]>
     </Shader>
   </ShaderOp>
+  <ShaderOp Name="SampleCmpLevel" PS="PS" VS="VS" CS="CS" AS="AS" MS="MS" TopologyType="TRIANGLE">
```
jholewinski, line 584: Is this from a different execution test?

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -581,6 +581,361 @@
...
+                        g_tex.SampleCmpLevelZero(g_sampCmp, fcoord, cmp));
+#endif
+        }
+
+        uint4 DoMoarTests( int2 coord, int2 offset ) {
+          uint3 lcoord = uint3(coord, 0);
+#if __SHADER_TARGET_MINOR == 7
+          float2 fcoord = (coord + 0.5)/1000.0;
+          return uint4( g_tex.Load(lcoord, offset),
+                        g_tex.SampleBias(g_samp, coord, -1.0, offset),
```
jholewinski, line 833: Shouldn't this use `fcoord` instead of `coord`?

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -581,6 +581,361 @@
...
+          if (offset.x < -8 || offset.x > 7)
+            offset.x = (offset.x << 28) >> 28;
+          if (offset.y < -8 || offset.y > 7)
+            offset.y = (offset.y << 28) >> 28;
+          return offset;
+        }
+
+        uint4 DoTests( int2 coord, int2 offset ) {
+#if __SHADER_TARGET_MINOR == 7
+          float cmp = (coord.y + offset.y)*1000 + coord.x + offset.x;
```
jholewinski, line 811: `offset` should be clamped here so only the lower 4 bits are considered

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -581,6 +581,361 @@
...
+          result.position = float4(position, 1.0);
+          result.uv = uv;
+          return result;
+        }
+
+        float4 PSMain(PSInput input) : SV_TARGET {
+          uint ix = uint(input.uv.x * 1000);
+          float2 coord = float2(coords[input.uv.x*19], coords[input.uv.y*19]);
+          uint2 offset = uint2(offsets[input.uv.x*19], offsets[input.uv.y*19]);
+          g_bufMain[2*ix] = 1;//DoTests(coord, offset);
```
jholewinski, line 923: Debugging code?

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -581,6 +581,361 @@
...
+        PSInput VSMain(float3 position : POSITION, float2 uv : TEXCOORD) {
+          PSInput result;
+          result.position = float4(position, 1.0);
+          result.uv = uv;
+          return result;
+        }
+
+        float4 PSMain(PSInput input) : SV_TARGET {
+          uint ix = uint(input.uv.x * 1000);
+          float2 coord = float2(coords[input.uv.x*19], coords[input.uv.y*19]);
```
jholewinski, line 921: What is the meaning of 19 here?

tools/clang/test/HLSL/ShaderOpArith.xml
pow2clk: Nope. That's the point of this test, to verify that behavior when bounds are exceeded.

pow2clk: sorry. I missed the context here. Yeah. the Clamp call is missing in the 6.7 variant

pow2clk: Okay. Deleted my previous comments and I'm going to try again. 

The intent is that the clamping takes place by the implementation. Since the 6.6 code has no implementation, the test does the clamping. The results testing assumes that the clamping was performed either by the implementation or the test.

I considered performing the clamping within the compiler to make sure the limit was respected, but I didn't want to generate possibly unneeded code that would slow the operation down. So we have this HLK test to make sure the implementation performs the clamps.

jholewinski: But this particular expression is computing a reference value without using texture ops. So there is no clamping that will be performed by the HW. As far as its concerned, it's just a standard 32-bit integer computation.

pow2clk: I'm with you now

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -581,6 +581,216 @@
...
+
+        Texture2D<float> g_tex : register(t0);
+        RWStructuredBuffer<uint4> g_bufMain : register(u0);
+        RWStructuredBuffer<uint4> g_bufMesh : register(u1);
+        RWStructuredBuffer<uint4> g_bufAmp : register(u2);
+
+        SamplerState g_samp : register(s0);
+        SamplerComparisonState g_sampCmp : register(s1);
+
+        int2 ClampOffset(int2 offset) {
```
tex3d, line 655: "Clamp" seems misleading here, since it actually wraps, right?

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -581,6 +581,216 @@
...
+            verts[ix%6].position = g_Verts[ix%6];
+            verts[ix%6].uv = g_UV[ix%6];
+            tris[ix&1] = uint3((ix&1)*3, (ix&1)*3 + 1, (ix&1)*3 + 2);
+            // Bunching up work to compensate for small numthreads limits
+            for (int y = 0; y < 9; y++) {
+              for (int x = 0; x < 9; x++) {
+                float2 coord = float2(coords[9*id.x+x], coords[9*id.y+y]);
+                int2 offset = uint2(offsets[9*id.x+x], offsets[9*id.y+y]);
+                g_bufAmp[2*9*9*ix + 9*y + x + 0] = DoTests(coord, offset);
+                g_bufAmp[2*9*9*ix + 9*y + x + 1] = DoMoarTests(coord, offset);
```
tex3d, line 762: I think you meant to use g_bufMesh here?

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -3538,6 +3539,92 @@ TEST_F(ExecutionTest, ComputeSampleTest) {
...
+  pShaderOp->CS = nullptr;
+
+  if (DoesDeviceSupportMeshAmpDerivatives(pDevice)) {
+    test = RunShaderOpTestAfterParse(pDevice, m_support, "ProgOffset", SampleInitFn, ShaderOpSet);
+
+    test->Test->GetReadBackData("U0", &data);
+    VerifyProgOffsetResults((UINT*)data.data());
+
+    test->Test->GetReadBackData("U1", &data);
+    VerifyProgOffsetResults((UINT*)data.data());
```
tex3d, line 3613: How does this pass if it doesn't appear to be written to?

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -581,6 +581,216 @@
...
+        }
+
+        uint4 DoMoarTests( int2 coord, int2 offset ) {
+          uint3 lcoord = uint3(coord, 0);
+#if  __SHADER_TARGET_MAJOR > 6 || (__SHADER_TARGET_MAJOR == 6 && __SHADER_TARGET_MINOR >= 7)
+          float2 fcoord = (coord + 0.5)/1000.0;
+          return uint4( g_tex.Load(lcoord, offset),
+                        g_tex.SampleBias(g_samp, fcoord, -1.0, offset),
+                        g_tex.SampleGrad(g_samp, fcoord, 1.0, 1.0, offset),
+                        g_tex.SampleLevel(g_samp, fcoord, 0, offset));
```
tex3d, line 690: `g_tex` only has one mip level, so I'm not sure these are adequately testing Bias/Grad/Level.

pow2clk: Simpler that way. It's very hard to anticipate exactly where a mipmapped result will land. I didn't suppose that mip levels had much effect here.

I suppose I could retrieve two values, one without an offset and one with and verify that the differences are as expected. I'd prefer to do that in a followup.

tex3d (approved):
Looks good now!

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -3640,6 +3641,164 @@ TEST_F(ExecutionTest, ComputeSampleTest) {
...
+                                         D3D_SHADER_MODEL_6_6,
+                                         D3D_SHADER_MODEL_6_7};
+  for (unsigned i = 0; i < _countof(TestShaderModels); i++) {
+    D3D_SHADER_MODEL sm = TestShaderModels[i];
+
+    CComPtr<ID3D12Device> pDevice;
+    if (!CreateDevice(&pDevice, sm, /*skipUnsupported*/false)) {
+      LogCommentFmt(L"Device does not support shader model 6.%1u",
+                    ((UINT)sm & 0x0f));
+      continue;
```
tex3d, line 3717: Oh, should have made this `break;` instead of `continue;` so it doesn't keep trying higher shader models.

PR #4287, opened 2022-02-22
Extending QuadAnyAll test for MS and AS

llvm-beanz:
This extends the existing compute test to cover Mesh and Amplifiction
shaders.

### DIFF
--- a/tools/clang/test/HLSL/ShaderOpArith.xml
+++ b/tools/clang/test/HLSL/ShaderOpArith.xml
@@ -2933,23 +2933,87 @@
 ]]>
     </Shader>
   </ShaderOp>
-  <ShaderOp Name="QuadAnyAll" CS="CS" DispatchX="8" DispatchY="8">
-    <RootSignature>RootFlags(0), UAV(u0)</RootSignature>
-    <Resource Name="UAVBuffer0" Dimension="BUFFER" Width="1024" InitialResourceState="COPY_DEST" Init="Zero" Flags="ALLOW_UNORDERED_ACCESS" ReadBack="true" />
+  <ShaderOp Name="QuadAnyAll" CS="CS" VS="VS" PS="PS" DispatchX="8" DispatchY="8">
+    <RootSignature>RootFlags(ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT), UAV(u0)</RootSignature>
+    <Resource Name="UAVBuffer0" Dimension="BUFFER" Width="2048" Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST" TransitionTo="UNORDERED_ACCESS" Init="ZERO" ReadBack="true"/>
+    <Resource Name="RTarget" Dimension="TEXTURE2D" Width="16" Height="16" Format="R32G32B32A32_UINT" Flags="ALLOW_RENDER_TARGET" InitialResourceState="COPY_DEST" ReadBack="true" />
+    <Resource Name="VBuffer" Dimension="BUFFER" InitialResourceState="COPY_DEST" Init="FromBytes" Topology="TRIANGLELIST">
+      { { -1.0f, 1.0f, 0.0f }, { 0.0f, 0.0f } },
+      { { 1.0f, 1.0f, 0.0f }, { 1.0f, 0.0f } },
+      { { -1.0f, -1.0f, 0.0f }, { 0.0f, 1.0f } },
+
+      { { -1.0f, -1.0f, 0.0f }, { 0.0f, 1.0f } },
+      { { 1.0f, 1.0f, 0.0f }, { 1.0f, 0.0f } },
+      { { 1.0f, -1.0f, 0.0f }, { 1.0f, 1.0f } }
+    </Resource>
     <RootValues>
-      <RootValue Index="0" ResName="UAVBuffer0"/>
+      <RootValue Index="0" ResName="UAVBuffer0" />
     </RootValues>
-    <Shader Name="CS67" Target="cs_6_7" EntryPoint="main" Text="@CS"/>
-    <Shader Name="CS" Target="cs_6_0" EntryPoint="main">
+    <DescriptorHeap Name="RtvHeap" NumDescriptors="1" Type="RTV">
+      <Descriptor Name="RTarget" Kind="RTV"/>
+    </DescriptorHeap>
+    <InputElements>
+      <InputElement SemanticName="POSITION" Format="R32G32B32_FLOAT" AlignedByteOffset="0" />
+      <InputElement SemanticName="TEXCOORD" Format="R32G32_FLOAT" AlignedByteOffset="12" />
+    </InputElements>
+    <RenderTargets>
+      <RenderTarget Name="RTarget">
+        <Viewport Width="2.0" Height="2.0" MaxDepth="1.0"/>
+      </RenderTarget>
+    </RenderTargets>
+    
+    <Shader Name="PS" Target="ps_6_5" EntryPoint="PSMain" Text="@CS"/>
+    <Shader Name="VS" Target="vs_6_5" EntryPoint="VSMain" Text="@CS"/>
+    <Shader Name="AS67" Target="as_6_7" EntryPoint="ASMain" Text="@CS"/>
+    <Shader Name="AS" Target="as_6_5" EntryPoint="ASMain" Text="@CS"/>
+    <Shader Name="MS67" Target="ms_6_7" EntryPoint="MSMain" Text="@CS"/>
+    <Shader Name="MS" Target="ms_6_5" EntryPoint="MSMain" Text="@CS"/>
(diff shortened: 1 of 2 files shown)


tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9573,45 +9573,64 @@ struct int2 {
...
+      pShaderOp->PS = "PS";
+      pShaderOp->VS = "VS";
+      pShaderOp->MS = "MS";
+      pShaderOp->AS = "AS";
+    } else if (sm == D3D_SHADER_MODEL_6_7) {
+      pShaderOp->PS = "PS";
+      pShaderOp->VS = "VS";
+      pShaderOp->AS = "AS67";
+      pShaderOp->MS = "MS67";
       pShaderOp->CS = "CS67";
```
pow2clk, line 9634: In my experience, the checks for these were pointer compares, so I had to search for the matching names in the shaders in ShaderOp. Maybe that's changed?

jholewinski: Yup, just ran into this during testing. See #4293 for the fix for the CS67 case. The same logic can be applied here for the other shader stages.

tex3d: I merged a more general fix #4296, so you can use that pattern here instead, like: `pShaderOp->CS = pShaderOp->GetString("CS67");`.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9573,45 +9573,64 @@ struct int2 {
...
   if (args[0]) {
-    for (st::ShaderOpShader& S : pShaderOp->Shaders)
+    for (st::ShaderOpShader &S : pShaderOp->Shaders)
       S.Arguments = args;
   }
 
   bool Skipped = true;
-  D3D_SHADER_MODEL TestShaderModels[] = { D3D_SHADER_MODEL_6_0, D3D_SHADER_MODEL_6_7 };
+  D3D_SHADER_MODEL TestShaderModels[] = {D3D_SHADER_MODEL_6_0,
+                                         D3D_SHADER_MODEL_6_5};
```
jholewinski, line 9618: Accidental removal of `D3D_SHADER_MODEL_6_7`?

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9573,45 +9573,64 @@ struct int2 {
...
-    LogCommentFmt(L"\r\nVerifying QuadAny/QuadAll using Wave intrinsics in shader model 6.%1u", ((UINT)sm & 0x0f));
-
-    if (sm == D3D_SHADER_MODEL_6_7) {
+    LogCommentFmt(L"\r\nVerifying QuadAny/QuadAll using Wave intrinsics in "
+                  L"shader model 6.%1u",
+                  ((UINT)sm & 0x0f));
+    if (sm == D3D_SHADER_MODEL_6_5) {
+      pShaderOp->PS = "PS";
+      pShaderOp->VS = "VS";
+      pShaderOp->MS = "MS";
```
jholewinski, line 9627: Unfortunately the framework seems to only run the compute version since a compute shader is bound. I believe the test will need to be run once with `CS` set, and again with `CS` nulled out to force the AS/MS path. Do we need the VS/PS shaders?

llvm-beanz:
The force push there is a rebase on master to adjust to @tex3d's change.

tools/clang/unittests/HLSL/ExecutionTest.cpp
jholewinski: Nevermind, I somehow missed the `CS = nullptr` part of the test below.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -2933,23 +2933,87 @@
...
+  // This output doesn't actually matter
+  return input.position;
+}
+
+PSInput VSMain(float3 position : POSITION, float2 uv : TEXCOORD) {
+  PSInput result;
+  result.position = float4(position, 1.0);
+  result.uv = uv;
+  return result;
+}
```
pow2clk, line 3010: I don't think this is currently used.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -2933,23 +2933,87 @@
...
+  { 1.0f, 0.0f },
+  { 0.0f, 1.0f },
+
+  { 0.0f, 1.0f },
+  { 1.0f, 0.0f },
+  { 1.0f, 1.0f }};
+
+float4 PSMain(PSInput input) : SV_TARGET {
+  // This output doesn't actually matter
+  return input.position;
```
pow2clk, line 3002: I think we should test QuadAny/All in PS too. It needn't be in this PR though.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -2933,23 +2933,87 @@
...
+      <InputElement SemanticName="TEXCOORD" Format="R32G32_FLOAT" AlignedByteOffset="12" />
+    </InputElements>
+    <RenderTargets>
+      <RenderTarget Name="RTarget">
+        <Viewport Width="2.0" Height="2.0" MaxDepth="1.0"/>
+      </RenderTarget>
+    </RenderTargets>
+    
+    <Shader Name="PS" Target="ps_6_5" EntryPoint="PSMain" Text="@CS"/>
+    <Shader Name="VS" Target="vs_6_5" EntryPoint="VSMain" Text="@CS"/>
```
pow2clk, line 2966: As discussed offline, since we could use this with a device created as 6.0, these should be 6.0 too. Perhaps that's part of the followup PR

pow2clk (approved):
Chris is aware of the omitted 6_7 variant and will remedy it. With that understanding, I'm approving this with comments that I expect are already being addressed in a change to come

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9578,18 +9594,34 @@ TEST_F(ExecutionTest, QuadAnyAll) {
...
+    std::shared_ptr<ShaderOpTestResult> test = RunShaderOpTestAfterParse(
+        pDevice, m_support, "QuadAnyAll", CleanUAVBuffer0Buffer, ShaderOpSet);
 
     MappedData uavData;
     test->Test->GetReadBackData("UAVBuffer0", &uavData);
-    bool Result = VerifyQuadAnyAllResults((int2*)uavData.data());
+    bool Result = VerifyQuadAnyAllResults((int2 *)uavData.data());
+    VERIFY_IS_TRUE(Result);
+
+    if (!DoesDeviceSupportMeshShaders(pDevice))
```
tex3d, line 9613: This will still try to run AS/MS on 6.0, and get all 0's in readback, failing the test.  Need to add ` || sm < D3D_SHADER_MODEL_6_5`.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -2966,6 +3030,36 @@ void main(uint TID: SV_GroupIndex) {
...
+}
+
+struct Payload {
+  uint nothing;
+};
+
+[numthreads(8, 8, 1)]
+void ASMain(uint GID : SV_GroupIndex) {
+  QuadAnyAllTest(GID, 0);
+  Payload P = {0};
```
tex3d, line 3046: Please move payload to groupshared global.  We don't want to encourage this pattern in any of our code.  I think we'll want to prevent it in the future even.

tex3d: I see that this is a pattern throughout this file, so perhaps that can be cleaned up with another change...

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9680,18 +9697,34 @@ TEST_F(ExecutionTest, QuadAnyAll) {
...
-    std::shared_ptr<ShaderOpTestResult> test = RunShaderOpTestAfterParse(pDevice, m_support, "QuadAnyAll",
-      CleanUAVBuffer0Buffer, ShaderOpSet);
+    std::shared_ptr<ShaderOpTestResult> test = RunShaderOpTestAfterParse(
+        pDevice, m_support, "QuadAnyAll", CleanUAVBuffer0Buffer, ShaderOpSet);
 
     MappedData uavData;
     test->Test->GetReadBackData("UAVBuffer0", &uavData);
-    bool Result = VerifyQuadAnyAllResults((int2*)uavData.data());
+    bool Result = VerifyQuadAnyAllResults((int2 *)uavData.data());
+    VERIFY_IS_TRUE(Result);
```
tex3d, line 9714: When these fail, it's not easy to tell which thing failed.  I think we need to have some more detail here if possible.  Perhaps this could be a future improvement though.

tex3d (approved):
These pass for me now on fallback path - up through SM 6.5 with AS/MS enabled.

PR #4292, opened 2022-02-23
Missing break in CreateHandleFromHeap case caused unwanted shader flags

tex3d:
hasWriteableMSAATextures and hasAdvancedTextureOps were being set if you used CreateHandleFromHeap because of a missing break.


bob80905 (approved):
Got rid of the options for me when running dxc

PR #4293, opened 2022-02-23
Fix shader selection in QuadAnyAll test

jholewinski:
The `CS` field (and all shader type fields) in ShaderOp are used for
pointer comparisons, not string comparisons. Therefore, we must locate
the original pointer to the shader name we want from the `Shaders` field
instead of passing a string literal.


jholewinski:
Abandoning in favor of #4296

PR #4296, opened 2022-02-23
Exec tests: change string entry setting pattern

tex3d:
Because the LPCSTR set to shader stages in ShaderOp are used as a key,
the string pointer has to match an entry.  Normally this works because
all strings are added to a string map and the pointer from this is used
for all LPCSTR pointers.

ExecutionTest groups were trying to override the entries by searching each
entry point for the desired name, then setting the stage to the string
pointer of that entry, so the pointer would match.
Another way is to just use insert() on the string map, which is what all
the other code uses (during loading from XML, for instance).

This change adds ShaderOp::GetString() to make it convenient to get the
appropriate string pointer for whatever string you have, so you may simply
set this pointer to the shader stage, without iterating through all shaders,
looking for the matching name.

This adds a check to detect when an entry name is set for a stage, but that
entry name doesn't match any shader in the ShaderOp.

It also updates the code in ExecutionTest to use this pattern instead, so
future tests won't copy the old pattern.

One additional advantage to this pattern is error detection: if you used
the old pattern, you could end up not finding the entry, setting nullptr to
the stage, then what you run might be a different stage than you thought,
without any error or indication otherwise.  Many tests expect the same
results written from different stages, so if the wrong stage produced the
results, the test would not detect this.  Now you will get a failure during
execution in this case.

This is an alternative fix to #4293.

### DIFF
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -8859,20 +8859,11 @@ TEST_F(ExecutionTest, Atomics64Test) {
 
   // Reassign shader stages to 64-bit versions
   // Collect 64-bit shaders
-  LPCSTR CS64 = nullptr, VS64 = nullptr, PS64 = nullptr;
-  LPCSTR AS64 = nullptr, MS64 = nullptr;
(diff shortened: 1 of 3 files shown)


hekota (approved):
LGTM!

PR #4297, opened 2022-02-24
New HLSL_AGILITYSDK_DIR and HLSLHost dependency copying help.

tex3d:
Use environment variable HLSL_AGILITYSDK_DIR which you can point to
extracted contents of an AgilitySDK.
Automatically copy binaries to appropriate location when set.

Look for MinTe binary set from TAEF.  If found, set a local TAEF binary
location.
If this is set when running execution tests, copy MinTe there, and if the
HLSL_AGILITYSDK_DIR was set, also copy agility SDK binaries under the
.\D3D12\ subdir there, so the test can automatically pick it up.

Add hctbins.cmd to copy extra bins necessary for running things out of
the ...\bin or ...\test directories such as HLSLHost.exe.  This includes
AgilitySDK bins if set.

### DIFF
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -491,7 +491,6 @@ if "%HLSL_TAEF_DIR%"=="" (
   echo HLSL_AGILITYSDK_DIR set, but no HLSL_TAEF_DIR set, no AgilitySDK will be copied
   exit /b 1
 )
-echo MARK 4
 set FULL_AGILITY_PATH=
 if exist "%HLSL_AGILITYSDK_DIR%\build\native\bin\%BUILD_ARCH:Win32=x86%\D3D12Core.dll" (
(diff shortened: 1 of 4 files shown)


utils/hct/hcttest.cmd
```diff
@@ -472,3 +481,28 @@ if not "%2"=="" (
...
+:copyagility
+if "%HLSL_AGILITYSDK_DIR%"=="" (
+  exit /b 0
+)
+set USE_AGILITY_SDK=/p:D3D12SDKVersion=1
+if "%HLSL_TAEF_DIR%"=="" (
+  echo HLSL_AGILITYSDK_DIR set, but no HLSL_TAEF_DIR set, no AgilitySDK will be copied
+  exit /b 1
+)
+echo MARK 4
```
hekota, line 494: Forgotten debugging mark ;)

hekota (approved):

utils/hct/hcttest.cmd
tex3d: Oops, thanks!

PR #4298, opened 2022-02-24
Forward compiler arguments to rewriter to support SPIR-V and HLSL 2021 features

LukasBanana:
This is a proposal to forward compiler arguments such as `-spirv` and `-HV 2021` to the rewriter which we use in UE5.
It introduces a breaking change in the interface of `RemoveUnusedGlobals`, `RewriteUnchanged`, and `RewriteUnchangedWithInclude` but I'm open for discussions about getting this in as a non-breaking change, perhaps by just adding alternative interface functions.

### DIFF
--- a/tools/clang/tools/libclang/dxcrewriteunused.cpp
+++ b/tools/clang/tools/libclang/dxcrewriteunused.cpp
@@ -546,6 +546,15 @@ void SetupCompilerCommon(CompilerInstance &compiler,
   compiler.getLangOpts().UseMinPrecision = !opts.Enable16BitTypes;
   compiler.getLangOpts().EnableDX9CompatMode = opts.EnableDX9CompatMode;
   compiler.getLangOpts().EnableFXCCompatMode = opts.EnableFXCCompatMode;
+  compiler.getLangOpts().EnableTemplates = opts.EnableTemplates;
+  compiler.getLangOpts().EnableOperatorOverloading = opts.EnableOperatorOverloading;
+  compiler.getLangOpts().StrictUDTCasting = opts.StrictUDTCasting;
+  compiler.getLangOpts().EnablePayloadAccessQualifiers = opts.EnablePayloadQualifiers;
+  compiler.getLangOpts().EnableShortCircuit = opts.EnableShortCircuit;
+  compiler.getLangOpts().EnableBitfields = opts.EnableBitfields;
+#ifdef ENABLE_SPIRV_CODEGEN
+  compiler.getLangOpts().SPIRV = opts.GenSPIRV;
+#endif
   compiler.getDiagnostics().setIgnoreAllWarnings(!opts.OutputWarnings);
   compiler.getCodeGenOpts().MainFileName = pMainFile;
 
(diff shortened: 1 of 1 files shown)


LukasBanana:
@tex3d would you be able to take a look at this PR? We need the rewriter and these configurations being forwarded for UE5.
Thanks in advance.

include/dxc/dxctools.h
```diff
@@ -29,19 +29,22 @@ struct IDxcRewriter : public IUnknown {
                                                         _In_z_ LPCWSTR entryPoint,
                                                         _In_count_(defineCount) DxcDefine *pDefines,
                                                         _In_ UINT32 defineCount,
+                                                        _In_count_(argCount) LPCSTR *pArgs, _In_ UINT32 argCount,
```
tex3d, line 32: Can't just change existing interface.  Need to use the new IDxcRewriter2::RewriteWithOptions instead, which supports passing options.

tools/clang/tools/libclang/dxcrewriteunused.cpp
```diff
@@ -546,6 +546,15 @@ void SetupCompilerCommon(CompilerInstance &compiler,
   compiler.getLangOpts().UseMinPrecision = !opts.Enable16BitTypes;
   compiler.getLangOpts().EnableDX9CompatMode = opts.EnableDX9CompatMode;
   compiler.getLangOpts().EnableFXCCompatMode = opts.EnableFXCCompatMode;
+  compiler.getLangOpts().EnableTemplates = opts.EnableTemplates;
```
tex3d, line 549: These additions are fine.

tools/clang/tools/libclang/dxcrewriteunused.cpp
```diff
@@ -838,7 +847,9 @@ HRESULT ReadOptsAndValidate(hlsl::options::MainArgs &mainArgs,
   IFT(CreateMemoryStream(GetGlobalHeapMalloc(), &pOutputStream));
   raw_stream_ostream outStream(pOutputStream);
 
-  if (0 != hlsl::options::ReadDxcOpts(table, hlsl::options::HlslFlags::RewriteOption,
+  if (0 != hlsl::options::ReadDxcOpts(table,
+                                      hlsl::options::HlslFlags::RewriteOption |
+                                          hlsl::options::HlslFlags::CoreOption,
```
tex3d, line 852: This is fine.

tex3d (changes requested):
We can't simply modify the existing interface methods.

There's already a new interface with a method that takes options (`IDxcRewriter2::RewriteWithOptions`) you can use instead, and you use the rewriter options to target the same functionality that the prior interface methods provided.  See `dxc -?` options under "Rewriter Options:".

The change to set compiler langopts in SetupCompilerCommon, and adding CoreOption to ReadOptsAndValidate are good, but I think the rest of the changes need to be reverted.

LukasBanana:
@tex3d I updated the PR according to your review. I'll utilize the `IDxcRewriter2::RewriteWithOptions` interface for our purpose instead. Thanks

damyanp:
@LukasBanana - sorry for the long gap here.  If you're able to get this building again and add some test coverage for it, then I think we'd be better positioned to take the change.  Thank you!

damyanp (changes requested):
As it's been a while since anyone looked at this, we need to get it building again.

We'd also need to see some test coverage for this functionality before we can accept it.

Thank you!

LukasBanana:
@damyanp Apologies for the late response, this went under my radar. We don't use the DXC rewriter anymore as we switched to SPIRV-Tools to handle all of the removal of unused globals and functions for our SPIR-V based backends. So this doesn't have priority for us at the moment to write unit tests for. I'm fine with closing this PR.

Thanks
Laura

damyanp:
Thanks for the update @LukasBanana - closing this now.

PR #4299, opened 2022-02-25
ExecutionTest for Writable MSAA

pow2clk:
Writes sample values to a RWTexture2DMS that are the sample index plus 1
in the shader. Reads that data back and verifies that it matches

Incidentally fixes a ShaderOpTest issue when initializing a texture with
depth or array dimensions.

### DIFF
--- a/tools/clang/test/HLSL/ShaderOpArith.xml
+++ b/tools/clang/test/HLSL/ShaderOpArith.xml
@@ -2538,6 +2538,93 @@
     </Shader>
   </ShaderOp>
 
+  <ShaderOp Name="WriteMSAA" PS="PS" VS="VS" CS="CS">
+    <RootSignature>
+      RootFlags(ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT),
+      DescriptorTable(UAV(u0) ,UAV(u1)),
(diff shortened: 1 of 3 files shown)


tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -2814,6 +2814,93 @@
...
+    <InputElements>
+      <InputElement SemanticName="POSITION" Format="R32G32B32_FLOAT" AlignedByteOffset="0" />
+      <InputElement SemanticName="TEXCOORD" Format="R32G32_FLOAT" AlignedByteOffset="12" />
+    </InputElements>
+    <RenderTargets>
+      <RenderTarget Name="RTarget"/>
+    </RenderTargets>
+    <Shader Name="VS"   Target="vs_6_6" EntryPoint="VSMain" Text="@CS"/>
+    <Shader Name="PS"   Target="ps_6_6" EntryPoint="PSMain" Text="@CS"/>
+    <Shader Name="CS"   Target="cs_6_6" EntryPoint="CSMain">
```
tex3d, line 2861: SM 6.6?

pow2clk:
I ended up completely rewriting this. Adding the changes here would be more confusing than enlightening. So here is the new PR: #4387

Issue #2472, opened 2019-09-19
DXC: Compilation fails if a Buffer/Texture resource defined inside a namespace is sampled/used


Sparkrai:
```
namespace test
{
    Texture2D<float4> colorMap : register(t0);
}

SamplerState pointSampler : register(s0);

float4 PSMain(in float2 uv : TEXCOORD0) : SV_TARGET0
{
    return test::colorMap.Sample(pointSampler, uv);
}
```

The provided code, even if entirely valid, will not compile. When run dxc will output this error:

> test.hlsl:10:26: error: no member named 'Sample' in 'Texture2D<vector<float, 4> >'
    return test::colorMap.Sample(pointSampler, uv);
            ~~~~~~~~~~~~~ ^

With FXC the shader gets compiled correctly without any errors/warnings. This is the assembly output by FXC with shader model ps_5_1:

> Microsoft (R) Direct3D Shader Compiler 10.1 (using D:\p4\HMD2\Development\LsTools\Bin\pipeline\D3DCOMPILER_47.dll)
Copyright (C) 2013 Microsoft. All rights reserved.
//
// Generated by Microsoft (R) HLSL Shader Compiler 10.1
//
//
// Resource Bindings:
//
// Name                                 Type  Format         Dim      ID      HLSL Bind  Count
// ------------------------------ ---------- ------- ----------- ------- -------------- ------
// pointSampler                      sampler      NA          NA      S0             s0      1
// test::colorMap                    texture  float4          2d      T0             t0      1
//
//
//
// Input signature:
//
// Name                 Index   Mask Register SysValue  Format   Used
// -------------------- ----- ------ -------- -------- ------- ------
// TEXCOORD                 0   xy          0     NONE   float   xy
//
//
// Output signature:
//
// Name                 Index   Mask Register SysValue  Format   Used
// -------------------- ----- ------ -------- -------- ------- ------
// SV_TARGET                0   xyzw        0   TARGET   float   xyzw
//
ps_5_1
dcl_globalFlags refactoringAllowed
dcl_sampler S0[0:0], mode_default, space=0
dcl_resource_texture2d (float,float,float,float) T0[0:0], space=0
dcl_input_ps linear v0.xy
dcl_output o0.xyzw
sample o0.xyzw, v0.xyxx, T0[0].xyzw, S0[0]
ret
// Approximately 2 instruction slots used

Sparkrai:
[...]

PR #4302, opened 2022-02-28
Enable 'using' keyword.

python3kgae:
Enable 'using' keyword to help move hlsl intrinsic inside hlsl namespace.

This is just reenable existing c++ feature by remove check on HLSL.


tex3d:
Looks like you forgot to update verifier tests: `tools/clang/test/HLSL/cpp-errors*.hlsl`
```
using MyNS; // expected-error {{'using' is a reserved keyword in HLSL}}
```

pow2clk (approved):
LGTM. I was worried about some of the pitfalls that we've encountered with namespaces and builtins such as #2472, but I tested the shader I added for that modified to use "using" instead of the explicit namespaces and it worked fine.

I'm a bit surprised it's this easy. Why didn't we do this before? 🤔

tex3d (approved):
LGTM!

PR #4303, opened 2022-02-28
Fix AtomicsShared64 ExecutionTest

pow2clk:
The test was assigning a pixel shader that didn't exist due to the
conversion of mistaken code that was harmless in the previous form, but
results in an error message in the new.


hekota (approved):
LGTM!

PR #4305, opened 2022-03-01
Clean up debug preprocessor macros

llvm-beanz:
This change removes the `DBG` macro value in favor of `NDEBUG`, and
also converts many uses of `_DEBUG` with `NDEBUG`.

`_DEBUG` is set by MSVC when debug information generation is enabled.
`NDEBUG` is set by CMake debug configurations and by LLVM on all
configurations when `LLVM_ENABLE_ASSERTIONS` is set.

Aligning with LLVM's `NDEBUG` usage allows a consistent pattern
throughout the codebase.

### DIFF
--- a/cmake/modules/HandleLLVMOptions.cmake
+++ b/cmake/modules/HandleLLVMOptions.cmake
@@ -57,6 +57,8 @@ endif()
 
 if( LLVM_ENABLE_ASSERTIONS )
   # MSVC doesn't like _DEBUG on release builds. See PR 4379.
+  # HLSL Note: the above comment referrs to llvm.org problem, not pull request: 
+  #            https://bugs.llvm.org/show_bug.cgi?id=4379
   if( NOT MSVC )
     add_definitions( -D_DEBUG )
   endif()
--- a/cmake/modules/HandleLLVMOptions.cmake
+++ b/cmake/modules/HandleLLVMOptions.cmake
@@ -56,9 +56,10 @@ int main() { return (float)x; }"
 endif()
 
 if( LLVM_ENABLE_ASSERTIONS )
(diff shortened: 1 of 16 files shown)


pow2clk (approved):
Looks good! The sanity of future debuggers will improve as a result!

cmake/modules/HandleLLVMOptions.cmake
```diff
@@ -56,9 +56,10 @@ int main() { return (float)x; }"
 endif()
 
 if( LLVM_ENABLE_ASSERTIONS )
-  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDBG") # HLSL Change
-  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -UNDEBUG") # HLSL Change
-  if (0) # HLSL Change Starts
+  # MSVC doesn't like _DEBUG on release builds. See PR 4379.
```
tex3d, line 59: What PR is this referencing?  This PR number doesn't exist yet here.

lib/DXIL/DxilModule.cpp
```diff
@@ -124,7 +124,7 @@ DxilModule::DxilModule(Module *pModule)
   DXASSERT_NOMSG(m_pModule != nullptr);
   SetDxilHook(*m_pModule);
 
-#if defined(_DEBUG) || defined(DBG)
+#ifdef NDEBUG
```
tex3d, line 127: Shouldn't this be `ifndef`?

cmake/modules/HandleLLVMOptions.cmake
llvm-beanz: This is referencing the LLVM bug report (why bugzilla used to call them PRs I have no idea). The URL is: https://bugs.llvm.org/show_bug.cgi?id=4379

PR #4312, opened 2022-03-04
Add warning for 'using' keyword before hlsl2021.

python3kgae:


tools/clang/lib/Parse/ParseDeclCXX.cpp
```diff
@@ -584,7 +584,13 @@ Decl *Parser::ParseUsingDeclaration(unsigned Context,
     }
 
     ConsumeToken();
-
+    // HLSL change begin -warning ext before HLSL2021.
+    if (getLangOpts().HLSL) {
```
tex3d, line 588: This looks like it's only under type alias declaration.  Does it cover the `using namespace n;` case?

tools/clang/test/HLSLFileCheck/hlsl/diagnostics/warnings/using.hlsl
```diff
@@ -0,0 +1,27 @@
...
+// HV2021:error: control reaches end of non-void function
+namespace n {
+    using f = float;
+    namespace n2 {
+       f foo(f a) { return sin(a); }
+    }
+
+}
+
+using namespace n;
```
tex3d, line 18: Does the warning fire for this line?  It's not clear that it would, and the test does not check that it does.  A verifier test should be better for testing scenarios like this.

tools/clang/lib/Parse/ParseDeclCXX.cpp
python3kgae: Fixed.

tools/clang/include/clang/Basic/DiagnosticParseKinds.td
```diff
@@ -1027,6 +1027,8 @@ def warn_hlsl_semantic_identifier_collision : Warning <
   InGroup< HLSLSemanticIdentifierCollision >;
 def err_hlsl_enum : Error<
   "enum is unsupported in HLSL before 2017">;
+def warn_hlsl_new_feature : Warning <
+  "%0 may not be supported by compilers prior to %1.">;
```
llvm-beanz, line 1031: being a bit nicpicky, maybe the warning message could be something like:

%0 is an HLSL %1 feature, and is available in older versions as a non-portable extension

llvm-beanz (approved):
Looks good, I did have a bit of a nitpick on wording for the warning. Feel free to ignore if you want.

PR #4317, opened 2022-03-07
Make external sema source handle type completion

### LINKED ISSUE #4315: Template instantiation fails to populate object methods
DXC lazy-initializes the member methods for built-in objects. This triggers during parsing when objects of an HLSL type are provided, and fails to trigger during template instantiation as shown in the example below. This test case was reduced from the test provided in #4263, but is a separate issue from the "partial specialization" error.

```cpp
template <typename T> struct MyTex2D {
  uint heapId;

  template <typename Arg0> T Load(Arg0 arg0) { return Get().Load(arg0); }

  Texture2D<T> Get() { return (Texture2D<T>)ResourceDescriptorHeap[heapId]; }
};

cbuffer constantBuffer : register(b0) {
  MyTex2D<float4> tex;
};

float4 main() : SV_Target {
  float4 output = tex.Load(float3(0, 0, 0));
  return output;
}
```

llvm-beanz:
Prior to this change the HLSL built in types are inserted into the AST
during initialization as complete types missing their methods. The
method implementations are then inserted during parsing when handling
declarators and parameter declarations of the built in types.

This method of lazy expansion does not work with HLSL templates because
template declarators in dependent contexts are not handled by the
parser, so they bypass the code path that generates method definitions
(see #4315).

This change alters how we lazy initialize types. Instead of
lazy-initializing in parsing, this change lazy initializes when the
type is required to be complete. This is accomplished by creating the
built in type objects as incomplete decls, and adding an override for
`CompleteType` to the `HLSLExternalSource`.

This change simplifies our code because we can use the type's
`IsCompleteDefinition` flag to track types that have been completed
instead of our own bitmask system.

Fixes #4315

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -3098,8 +3098,6 @@ private:
   // Map from object decl to the object index.
   using ObjectTypeDeclMapType = std::array<std::pair<CXXRecordDecl*,unsigned>, _countof(g_ArBasicKindsAsTypes)+_countof(g_DeprecatedEffectObjectNames)>;
   ObjectTypeDeclMapType m_objectTypeDeclsMap;
-  // Mask for object which not has methods created.
-  uint64_t m_objectTypeLazyInitMask;
 
   UsedIntrinsicStore m_usedIntrinsics;
 
@@ -3550,7 +3548,6 @@ private:
 
     QualType float4Type = LookupVectorType(HLSLScalarType_float, 4);
     TypeSourceInfo *float4TypeSourceInfo = m_context->getTrivialTypeSourceInfo(float4Type, NoLoc);
-    m_objectTypeLazyInitMask = 0;
     unsigned effectKindIndex = 0;
     const auto *SM =
         hlsl::ShaderModel::GetByName(m_sema->getLangOpts().HLSLProfile.c_str());
@@ -3656,7 +3653,6 @@ private:
       }
       m_objectTypeDecls[i] = recordDecl;
       m_objectTypeDeclsMap[i] = std::make_pair(recordDecl, i);
-      m_objectTypeLazyInitMask |= ((uint64_t)1)<<i;
     }
 
     // Create an alias for SamplerState. 'sampler' is very commonly used.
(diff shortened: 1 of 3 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5117,20 +5113,21 @@ class HLSLExternalSource : public ExternalSemaSource {
...
+    if(Tag->isCompleteDefinition() || !isa<CXXRecordDecl>(Tag))
       return;
-    CXXRecordDecl *recordDecl = const_cast<CXXRecordDecl *>(GetRecordDeclForBuiltInOrStruct(qt->getAsCXXRecordDecl()));
+
+    CXXRecordDecl *recordDecl = cast<CXXRecordDecl>(Tag);
+    if (auto TDecl = dyn_cast<ClassTemplateSpecializationDecl>(recordDecl))
+      recordDecl = TDecl->getSpecializedTemplate()->getTemplatedDecl();
+
+    if (recordDecl->isCompleteDefinition())
+      return;
```
pow2clk, line 5125: Is this necessary? I think there's only one implementation of `isCompleteDefinition()`. So if it didn't bail above, it won't bail here.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12688,16 +12684,12 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
...
-    // Add methods if not ready.
-    hlslSource->AddHLSLObjectMethodsIfNotReady(qt);
   } else if (qt->isArrayType()) {
     QualType eltQt(qt->getArrayElementTypeNoTypeQual(), 0);
     while (eltQt->isArrayType())
       eltQt = QualType(eltQt->getArrayElementTypeNoTypeQual(), 0);
 
     if (hlsl::IsObjectType(this, eltQt, &bDeprecatedEffectObject)) {
-      // Add methods if not ready.
-      hlslSource->AddHLSLObjectMethodsIfNotReady(eltQt);
```
pow2clk, line 12700: Where do these get "completed" now? Looks like RequireCompleteType is called all over the place?

pow2clk (approved):
Looks good, just a small suggestion and an ignorant question

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: It’s a different Decl. The first call might be safe to omit because we shouldn’t get there without it being incomplete. The second call is on the template decl rather than the specialization decl. Because of the way clang recursively resolves templates the two don’t always agree.

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: This is the best part. Clang enforces C rules for requiring completeness. Any use of a type that requires completeness in C also requires it in HLSL. Whenever a type is required to be complete, but is not, clang calls out to the External AST source (if available) to complete the type. This preserves our laziness, but through clang’s built in methods.

pow2clk:
sounds good!

llvm-beanz:
It occurred to me that I could nest the second `isCompleteDefinition` check under the if statement that gates the `recordDecl` being different. In the most common cases that if block is executed, but it isn't always.

PR #4320, opened 2022-03-09
Add -arm64 argument to hcttest; exclude tests with issues on ARM64 & WARP

hekota:
- add arm64 argument to hcttest; use x86 as BUILD_ARCH instead of Win32 
- temporarily disable tests that have issues on ARM64, ARM64EC and WARP; bugs filed to track.
- remove obsoleted /labmode argument to TE.exe
- do not run arm64 tests in parallel (new TAEF has bug when -parallel is used, TAEF team has been notified)


tex3d (approved):
LGTM!

tex3d (approved):
LGTM!

PR #4323, opened 2022-03-10
[dxil2spv] Implement simple entry function creation

sudonatalie:
Translate a DXIL entry function to a SPIR-V entry function. This is
currently only supporting the few types and instructions needed to
create the entry function for a simple passthrough pixel shader.

### DIFF
--- a/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
+++ b/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
@@ -28,6 +28,7 @@
 #include "clang/CodeGen/CodeGenAction.h"
 #include "clang/Frontend/CodeGenOptions.h"
 #include "clang/Frontend/TextDiagnosticPrinter.h"
+#include "clang/SPIRV/SpirvUtils.h"
 
 namespace clang {
 namespace dxil2spv {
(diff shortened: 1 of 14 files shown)


tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -130,6 +124,153 @@ int Translator::Run(CComPtr<IDxcBlobEncoding> blob) {
...
+  // Handle Call instructions.
+  if (isa<llvm::CallInst>(instruction)) {
+    llvm::CallInst &callInstruction = cast<llvm::CallInst>(instruction);
+    // Many shader operations are represented by call instructions that call
+    // external functions, whose name is prefixed with dx.op. (i.e.
+    // dx.op.loadInput, dx.op.storeOutput).
+    hlsl::DXIL::OpCode dxilOpcode =
+        hlsl::OP::GetDxilOpFuncCallInst(&instruction);
+    switch (dxilOpcode) {
+    case hlsl::DXIL::OpCode::LoadInput: {
```
jaebaek, line 198: nit:
This method will become very long and complicated in long term. I'd like to split each case as a method e.g., `createLoadInstructionForInput()`.

jaebaek (approved):
LGTM with a nit. Congratulation!

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
sudonatalie: Done

PR #4324, opened 2022-03-12
User/jbatista/pass correct enumkind00

bob80905:
Swapped Filter to Comparison_Func, the correct EnumKind for this specific function.


tex3d (approved):
LGTM!

PR #4326, opened 2022-03-15
Remove LLVMHLSL dependency from IPO and ScalarOpts

hekota:
It causes circular target dependency when embedding the DXC in an internal project and it does not seem like it is necessary (builds fine without it).


hekota:
Windows build with SPIRV seems to be fine without these additional dependencies. Let's see if it is true for Linux as well.

PR #4328, opened 2022-03-15
[dxil2spv] SPIR-V validation and related fixes

sudonatalie:
Add SPIR-V validation after the module is generated, and apply some
fixes required for validation to succeed with a passthrough pixel
shader:
- Add execution mode
- Decorate with SPIR-V Locations
- Refactor some LowerTypeVisitor logic to a visitor that is shared with
  dxil2spv
- Remove extra pointerness on stage IO vars now that it is handled by
  the above pass

### DIFF
--- a/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
+++ b/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
@@ -14,6 +14,7 @@
 #include "clang/AST/HlslTypes.h"
 #include "clang/SPIRV/AstTypeProbe.h"
 #include "clang/SPIRV/SpirvFunction.h"
+#include "clang/SPIRV/SpirvUtils.h"
 
 namespace {
 /// Returns the :packoffset() annotation on the given decl. Returns nullptr if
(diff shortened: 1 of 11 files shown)


tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
```diff
@@ -176,6 +157,11 @@ bool LowerTypeVisitor::visitInstruction(SpirvInstruction *instr) {
     break;
   }
 
+  // Apply additional SPIR-V type transformations, including ensuring variables
+  // and function parameters have a pointer type.
+  SpirvTypeVisitor spirvTypeVisitor(context, opts);
+  spirvTypeVisitor.visitInstruction(instr);
```
jaebaek, line 163: I am worried that readers cannot expect `bool LowerTypeVisitor::visitInstruction(SpirvInstruction *instr)` uses `SpirvTypeVisitor` object as a local variable.

I'd like to place this in `std::vector<uint32_t> SpirvBuilder::takeModule()` right after
```
LowerTypeVisitor lowerTypeVisitor(*astContext, context, spirvOptions);
mod->invokeVisitor(&lowerTypeVisitor);
// here
```

Is there any issue in that change?

jaebaek: I am still concerned about this because we have an implicit rule to call the visitors only in the `takeModule()` method of SpirvBuilder. I know it can degrade the performance because it will visit all the instructions, but doing it in the `takeModule()` will be better for the long term maintenance.

I also took some time to understand this part. I believe later more readers would have the same issue.

I am sorry if I bother you but I'd like to update it
```
LowerTypeVisitor lowerTypeVisitor(*astContext, context, spirvOptions);
mod->invokeVisitor(&lowerTypeVisitor);
SpirvTypeVisitor spvTypeVisitor(context, spirvOptions);
mod->invokeVisitor(&spvTypeVisitor);
```

(if it does not have technical issue).

sudonatalie: As discussed offline, this change doesn't work in the current state, so I've moved it to a util function as requested.

jaebaek (approved):

PR #4329, opened 2022-03-15
Implement Integer Sampling

pow2clk:
Add support for integer sampling as part of advanced texture ops in
Shader Model 6.7. This includes adding static border colors and allowing
integer resource types to be used by all Sample* operations through
Sema, Dxil generation, and validation when 6.7 is available.

Adds root signature compilation and filecheck testing.


hekota (approved):
LGTM!

PR #4330, opened 2022-03-16
DxilConv: Pad copied IO signature parts

tex3d:
DXBC containers from some sources may have signature parts with unaligned sizes, such as from 9on12.

This change pads these signature parts when this is the case to avoid any
problems down the line.


hekota (approved):
LGTM!

PR #4332, opened 2022-03-16
Rework name lookup

### LINKED ISSUE #4263: [HLSL 2021] function template partial specialization is not allowed when declaring two functions with the same name
The following snippet is failing with "function template partial specialization is not allowed"

    template <typename T>
    struct MyTex2D
    {
        uint heapId;

        template <typename Arg0>
        T Load(Arg0 arg0)
        {
            return Get().Load(arg0);
        }

        template <typename Arg0, typename Arg1>
        T Load(Arg0 arg0, Arg1 arg1)
        {
            return Get().Load(arg0, arg1);
        }

        Texture2D<T> Get()
        {
            return (Texture2D<T>)ResourceDescriptorHeap[heapId];
        }
    };

    cbuffer constantBuffer : register(b0)
    {
        MyTex2D<float4> tex;
    };

    float4 main() : SV_Target
    {
        float4 output = tex.Load(int3(0, 0, 0));
        return output;
    }

llvm-beanz:
This is a big change... A lot of the change here is actually reverting
back to clang 3.7 code. Specifically around template name resoltuion.

The root cause of issue #4263 is that in DXC (before this change) HLSL
routes all name lookup through template name resolution. This results
in annotating some things as templates that shouldn't be. Specifically
in the test case in #4263, the second `Load` gets annotated as a
template-id tracking to the first `Load` when it should be just a
method specialization.

This fix is a bit involved.

I revert a bunch of the template handling code back to clang-3.7. This
causes two problems:

(1) HLSL allows omiting empty template argument lists for default
template cases (i.e. `matrix` is valid in HLSL, but in C++ it would
need to be `matrix<>`). This patch should fully resolve that case.
(2) Since DXC routes all names through template lookup, some
diagnostics were only emitted in template name lookup. This patch fixes
_some_ of those cases, but there are test failures that I'll resolve in
a subsequent commit on this PR.

I'm posting this change before it is fully ready to go because it is
large and may take some time to review.

Fixes #4263.


llvm-beanz:
There is one significant side effect of these changes that users will notice. In diagnostics instead of referring to ​`Texture2D​` errors will now show the template name ​`Texture2D<vector<float, 4> >​​`. This is the result of actually differentiating between template and non-template types.

pow2clk:
I confess I somewhat liked that you could refer to template object with default parameters without any angle brackets.

By `matrix` above, I hope you aren't referring to the built-in matrix type which defaults to float4x4?  I don't think we can afford to lose that unless it's limited to HLSL2021. Even then, I'm not sure it's desirable.

llvm-beanz:
In source you can still omit the empty template argument list. The change is that diagnostics will report the default template parameters.

pow2clk (approved):
A question about why we lost a warning and a comment suggestion, but nothing that need hold this up

PR #4333, opened 2022-03-17
Remove test output files; update .gitignore

hekota:
- added *.dxo.converted to the list of extensions ignored by git
- renamed 3 test outputs from .fxc to .dxbc (which will make them to be ignored by git)


tex3d (approved):
LGTM!

PR #4335, opened 2022-03-17
[spirv] Improve ray tracing capability error

### LINKED ISSUE #4313: [SPIR-V] Ray tracing incorrectly requires Vulkan 1.2
Duplicate of #2984, now with added urgency ;-)

Now that the S22 Navi2 phones are out, there are actual use-cases for this, since Android will only support Vulkan 1.1

sudonatalie:
When the specific target environment required is exactly
SPV_ENV_VULKAN_1_1_SPIRV_1_4, then make the error message explicitly
note that it is met by Vulkan 1.1 with Spirv 1.4 (rather than simply
requiring Vulkan 1.2).

Fixes #4313


jaebaek (approved):
Thanks!

PR #4336, opened 2022-03-17
Added a DXC_OUT_REMARKS output kind for display include process

### LINKED ISSUE #3663: IDxcResult->GetOutput(DXC_OUT_ERRORS, ...) gives false error information when using -H
When using a custom include handler, after a successful compile, the GetOutput(DXC_OUT_ERRORS,...) gives me Include information when the -H argument is used, e.g. false positives, unless I am understanding this wrong. All my HLSL files compile correctly using both fxc and the DX Compiler Editor. 

My custom include handler builds the path and filename required, loads the file, then passes back the IDxcBlob.

The error output reads:
**Opening file [./aabb_include.hlsl], stack top [0]
Opening file [./LightingUtil.hlsl], stack top [0]**
There is no other information.

This doesn't happen when the -H argument is omitted.

Shouldn't this information be available in GetOutput(DXC_OUT_TEXT,...) ? 

If this is correct behavior, what is the correct way to check for errors?  Many thanks.

gracejennings:
Adding an additional output kind for compile remarks that are not warnings/errors or part of the disassembly. The motivation for this was to display the include process added with the `-Vi` or `-H` options. Also set the remarks to print to console in the command line interface as comments to preserve valid LLVM IR in stdout. 
Fixes #3663


tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -3599,10 +3598,6 @@ TEST_F(CompilerTest, CodeGenFloatingPointEnvironment) {
 
 #endif  //  defined(_WIN32) && !defined(_M_ARM64)
 
-TEST_F(CompilerTest, CodeGenInclude) {
-  CodeGenTestCheck(L"Include.hlsl");
-}
-
```
pow2clk, line 3605: Why was this test removed?

gracejennings: This test was removed because it was actually checking stderr for the displayed include process text. The File Check tests are checking the disassembly and the remarks don't go there, so Adam suggested I just add a command test instead.

adam-yang: We should probably add a new test in CompilerTest to check for the remark output.

pow2clk (approved):

adam-yang (approved):
LGTM!

PR #4342, opened 2022-03-21
Pix hull shader debug

jeffnn:
A couple of minor changes to support PIX shader debugging for hull shaders generated by the 11on12 mapping layer for DXBC originals.


adam-yang (approved):
LGTM!

PR #4354, opened 2022-03-25
[dxil2spv] Add support for dx.op.threadId

sudonatalie:
Translate the dx.op.threadId instruction to a load of a single element
of the SPIR-V BuiltIn GlobalInvocationID.

### DIFF
--- a/tools/clang/test/Dxil2Spv/passthru-cs.ll
+++ b/tools/clang/test/Dxil2Spv/passthru-cs.ll
@@ -86,15 +86,16 @@ attributes #2 = { nounwind }
 ; ; SPIR-V
 ; ; Version: 1.0
 ; ; Generator: Google spiregg; 0
-; ; Bound: 13
+; ; Bound: 20
 ; ; Schema: 0
 ;                OpCapability Shader
 ;                OpMemoryModel Logical GLSL450
-;                OpEntryPoint GLCompute %main "main"
+;                OpEntryPoint GLCompute %main "main" %gl_GlobalInvocationID
 ;                OpExecutionMode %main LocalSize 1 1 1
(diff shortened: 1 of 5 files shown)


tools/clang/test/Dxil2Spv/passthru-cs.ll
```diff
@@ -86,38 +86,45 @@ attributes #2 = { nounwind }
...
 ; %_ptr_Uniform_type_ByteAddressBuffer = OpTypePointer Uniform %type_ByteAddressBuffer
 ; %type_RWByteAddressBuffer = OpTypeStruct %_runtimearr_uint
 ; %_ptr_Uniform_type_RWByteAddressBuffer = OpTypePointer Uniform %type_RWByteAddressBuffer
 ;        %void = OpTypeVoid
-;          %11 = OpTypeFunction %void
-;           %6 = OpVariable %_ptr_Uniform_type_ByteAddressBuffer Uniform
-;           %9 = OpVariable %_ptr_Uniform_type_RWByteAddressBuffer Uniform
-;        %main = OpFunction %void None %11
-;          %12 = OpLabel
+;          %15 = OpTypeFunction %void
```
jaebaek, line 115: I am worried that this whole SPIR-V code check can fail in the future e.g., changing the instruction emission order can assign different result ids.
So .. will we switch this to the one based on the pattern matching e.g., Filecheck tests?

jaebaek (approved):

tools/clang/test/Dxil2Spv/passthru-cs.ll
jaebaek: Sounds good.

Issue #3870, opened 2021-07-12
[SPIR-V] Auto-packing of outputs using "Component" decoration


scottkircher:
I had posted this issue awhile ago:

https://github.com/microsoft/DirectXShaderCompiler/issues/2618

About the fact that the SPIR-V backend for DXC does not pack outputs/interpolators the way that the DXIL backend does, so Vulkan shaders end up using more outputs/interpolators and may exceed hardware limits where DX12 shaders do not. We are, in particular, hitting this issue with complex tessellation shaders that have many outputs between hull (tessellation control) and domain (tessellation evaluation) shaders.

In that thread, it was suggested that the issue could be fixed by use of the SPIR-V "Component" decoration, but as far as I'm aware, no action has been taken to that effect.

I attempted to revive the original issue, but received no response, so I'm posting a new issue. This is becoming more important to us to have fixed now so I am hoping someone can take a look. Thanks!

jaebaek:
Sorry for being late to handle this issue. After carefully reading what happened, I have questions about packing the signature supported by the DXIL code gen:
1. Are `-pack_optimized` and `-pack_prefix_stable` (or `-pack-optimized` / `-pack-prefix-stable`) the flags to enable the packing?
2. I am not versed on the rule of packing the signature in DXIL. I checked a few unittests like [tools/clang/test/HLSLFileCheck/hlsl/compile_options/pack_optimized/optimized.hlsl](https://github.com/microsoft/DirectXShaderCompiler/tree/master/tools/clang/test/HLSLFileCheck/hlsl/compile_options/pack_optimized/optimized.hlsl), but the packing rule is not clear to me. I want to just use [the existing functions in DXIL code gen](https://github.com/microsoft/DirectXShaderCompiler/blob/master/include/dxc/HLSL/DxilSignatureAllocator.inl#L279). I want to ask you whether following the DXIL's packing rule would have some issues or not.

scottkircher:
1. I believe those are the related flags, however the DXIL backend appears to pack by default without any options specified. We are currently not using those options, so I don't actually know which packing behavior we're getting. But, the interpolators are definitely getting packed. Hopefully you can find in the code where the default is specified.
2. I think following DXIL's packing rule (i.e. reusing the existing packing functionality) would be ideal. We'd like to have consistent behavior between DXIL and SPIR-V, here.

jaebaek:
I will get back here after digging into it more.

jaebaek:
[...]

PR #4361, opened 2022-03-28
[spirv] support signature packing

### LINKED ISSUE #2618: [spirv] Interpolator (shader stage output location) packing for SPIR-V target?
We are using dxc to compile HLSL to SPIR-V for the Vulkan port of our current project. Recently we have encountered a serious issue related to vertex shader outputs. In particular, it appears that no packing of interpolators is performed by the compiler. A declaration like this:

	float particle_user_Fire_Light_Scatter_Intensity : ATTR_FIRE_LIGHT_SCATTER_INTENSITY;
	float particle_user_Tempature_Crush : ATTR_TEMPATURE_CRUSH;
	float particle_user_ParticleOpacity : ATTR_PARTICLEOPACITY;

Turns into SPIR-V like this:

OpDecorate %out_var_ATTR_FIRE_LIGHT_SCATTER_INTENSITY Location 10
OpDecorate %out_var_ATTR_TEMPATURE_CRUSH Location 11
OpDecorate %out_var_ATTR_PARTICLEOPACITY Location 12

So, each single float is taking an entire float4 output location.

Therefore, we exceed the maximum number of vertex shader outputs far faster with Vulkan shaders than we do with our current DX SM5.0 shaders (which we are still compiling with fxc). We have a somewhat large number of shaders where this will be an issue. Manual packing is a possibility, but it makes the code less readable and is labor intensive, so we'd like to avoid it if possible.

Am I missing something? Is there a way to enable interpolator packing akin to what we have gotten for years with HLSL? If not, is this in the plans for dxc when targeting SPIR-V?

Thanks!

jaebaek:
Currently, DXC assigns new locations for each stage variable. When there
are lots of stage variables, the number of locations we use can exceed the
maximum location limit. This commit uses `Component` decoration to pack
signature and reduce the number of locations DXC uses.

Fixes #2618 and #3870

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvContext.h
+++ b/tools/clang/include/clang/SPIRV/SpirvContext.h
@@ -424,6 +424,10 @@ public:
            instructionsWithLoweredType.end();
   }
 
+  /// Getter/setter of signaturePacking.
+  bool isSignaturePackingEnabled() const { return signaturePacking; }
+  void enableSignaturePacking() { signaturePacking = true; }
+
 private:
   /// \brief The allocator used to create SPIR-V entity objects.
   ///
@@ -482,6 +486,8 @@ private:
   // Major/Minor hlsl profile version.
   uint32_t majorVersion;
   uint32_t minorVersion;
+
+  bool signaturePacking; ///< Whether signature packing is enabled or not
 
   /// File name to rich debug info map. When the main source file
   /// includes header files, we create an element of debugInfo for
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1719,10 +1719,8 @@ private:
 class PackedLocationAndComponentSet {
 public:
   PackedLocationAndComponentSet(SpirvBuilder &spirvBuilder,
-                                llvm::function_ref<uint32_t(uint32_t)> nextLocs,
-                                bool forInput_)
-      : spvBuilder(spirvBuilder), assignLocs(nextLocs),
-        forInput(forInput_) {}
+                                llvm::function_ref<uint32_t(uint32_t)> nextLocs)
+      : spvBuilder(spirvBuilder), assignLocs(nextLocs) {}
 
   bool assignLocAndComponent(const StageVar *var) {
     if (tryReuseLocations(var)) {
@@ -1798,7 +1796,6 @@ private:
   llvm::SmallVector<uint32_t, 8>
       nextUnusedComponent; ///< A vector to keep the starting unused component
                            ///< number in each assigned location
-  bool forInput;
 };
 
 /// A class for managing resource bindings to avoid duplicate uses of the same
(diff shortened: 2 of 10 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1691,6 +1714,90 @@ class LocationSet {
   uint32_t nextLoc[kMaxIndex];              ///< Next available location
 };
 
+/// A class for managing stage input/output packed locations to avoid duplicate
```
sudonatalie, line 1717: Can this class be moved to its own file? I know `DeclResultIdMapper` is already out of hand but still better not to make it worse.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1691,6 +1714,90 @@ class LocationSet {
...
+    if (tryReuseLocations(var)) {
+      return true;
+    }
+    return assignNewLocations(var);
+  }
+
+private:
+  // Try to pack signature of |var| into unused component slots of already
+  // assigned locations. Returns whether it succeeds or not.
+  bool tryReuseLocations(const StageVar *var) {
```
sudonatalie, line 1735: I've read over this function a few times, and I'm still struggling to understand it at a high-level. Can you include a comment describing what the condition(s) needing to be satisfied are that determine whether "trying to pack" succeeds or not?

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1691,6 +1714,90 @@ class LocationSet {
...
+      assignedLocs.push_back(loc + i);
+      nextUnusedComponent.push_back(componentCount);
+    }
+    return true;
+  }
+
+private:
+  SpirvBuilder &spvBuilder;
+  llvm::function_ref<uint32_t(uint32_t)>
+      assignLocs; ///< A function to assign a new location number.
```
sudonatalie, line 1793: nit: Move these comments to their own lines above vars since they don't fit as EOL comments

tools/clang/include/clang/SPIRV/SpirvContext.h
```diff
@@ -483,6 +487,8 @@ class SpirvContext {
   uint32_t majorVersion;
   uint32_t minorVersion;
 
+  bool signaturePacking; ///< Whether signature packing is enabled or not
```
sudonatalie, line 490: I think we should make this its own `SpirvCodeGenOptions` and set and access it like any other SPIR-V option. I don't think we handle any other options that are this granular by setting a member variable on `SpirvContext`.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
jaebaek: I tried to describe it in detail, but not sure it is easier to understand. Please let me know if it is still not clear.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
jaebaek: done

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -533,6 +533,8 @@ SpirvEmitter::SpirvEmitter(CompilerInstance &ci)
   spvContext.setCurrentShaderModelKind(shaderModel->GetKind());
   spvContext.setMajorVersion(shaderModel->GetMajor());
   spvContext.setMinorVersion(shaderModel->GetMinor());
+  if (ci.getCodeGenOpts().HLSLSignaturePackingStrategy != 0)
```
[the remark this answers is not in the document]
jaebaek: done

tools/clang/include/clang/SPIRV/SpirvContext.h
jaebaek: done

sudonatalie (approved):

PR #4367, opened 2022-03-31
Update test for helperlane wave on VS.

python3kgae:


tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3250,7 +3250,7 @@
         
         PSInput VSMain(float3 pos : POSITION) {
             HelperLaneWaveTestResult60 tr60 = RunHelperLaneWaveTests60();
-            if (WaveGetLaneIndex() == 2) { // last lane writes results
+            if ( WaveIsFirstLane()) { // last lane writes results
```
tex3d, line 3253: For prefix/multiprefix ops, don't we want the values from the last lane, not the first?  In which case the fix here would be to do a WaveActiveMax on WaveGetLaneIndex() and use the highest lane index to write results, wouldn't it?

python3kgae: Updated to WaveGetLaneIndex() == WaveActiveMax(WaveGetLaneIndex()).

tex3d (approved):
LGTM!

PR #4370, opened 2022-04-01
Omit BufferBlock on SPIR-V 1.4 (~ Vulkan 1.1) instead of Vulkan 1.2

MarijnS95:
According to the SPIR-V validator this annotation was last usable on SPIR-V 1.3:

    fatal error: generated SPIR-V is invalid: 2nd operand of Decorate: operand BufferBlock(3) requires SPIR-V version 1.3 or earlier
        OpDecorate %type_ByteAddressBuffer BufferBlock

The visitor was originally built to remove/omit it on Vulkan 1.2 and higher only, but since supporting Vulkan 1.1 with SPIR-V 1.4 (in #4364)
- which conveniently sits inbetween SPIR-V 1.3 and Vulkan 1.2 - the check must be made more stict and apply from SPIR-V 1.4 universal onwards.

### DIFF
--- a/tools/clang/lib/SPIRV/EmitVisitor.cpp
+++ b/tools/clang/lib/SPIRV/EmitVisitor.cpp
@@ -2438,9 +2438,9 @@ uint32_t EmitTypeHandler::emitType(const SpirvType *type) {
     // Emit Block or BufferBlock decorations if necessary.
     auto interfaceType = structType->getInterfaceType();
     if (interfaceType == StructInterfaceType::StorageBuffer)
-      // BufferBlock decoration is deprecated in Vulkan 1.2 and later.
+      // The BufferBlock decoration requires SPIR-V version 1.3 or earlier.
       emitDecoration(id,
-                     featureManager.isTargetEnvVulkan1p2OrAbove()
+                     featureManager.isTargetEnvSpirv1p4OrAbove()
                          ? spv::Decoration::Block
                          : spv::Decoration::BufferBlock,
                      {});
--- a/tools/clang/lib/SPIRV/RemoveBufferBlockVisitor.cpp
+++ b/tools/clang/lib/SPIRV/RemoveBufferBlockVisitor.cpp
@@ -14,16 +14,16 @@
 namespace clang {
 namespace spirv {
 
-bool RemoveBufferBlockVisitor::isBufferBlockDecorationDeprecated() {
-  return featureManager.isTargetEnvVulkan1p2OrAbove();
+bool RemoveBufferBlockVisitor::isBufferBlockDecorationAvailable() {
+  return !featureManager.isTargetEnvSpirv1p4OrAbove();
 }
 
 bool RemoveBufferBlockVisitor::visit(SpirvModule *mod, Phase phase) {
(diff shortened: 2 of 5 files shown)


tools/clang/lib/SPIRV/EmitVisitor.cpp
```diff
@@ -2438,9 +2438,9 @@ uint32_t EmitTypeHandler::emitType(const SpirvType *type) {
     // Emit Block or BufferBlock decorations if necessary.
     auto interfaceType = structType->getInterfaceType();
     if (interfaceType == StructInterfaceType::StorageBuffer)
-      // BufferBlock decoration is deprecated in Vulkan 1.2 and later.
+      // BufferBlock decoration is deprecated in SPIR-V 1.4 and later.
```
sudonatalie, line 2441: To match the spec explicitly: "BufferBlock decoration is missing after SPIR-V 1.3." (since it was probably deprecated earlier, but still valid).

tools/clang/lib/SPIRV/RemoveBufferBlockVisitor.cpp
```diff
@@ -15,7 +15,7 @@ namespace clang {
 namespace spirv {
 
 bool RemoveBufferBlockVisitor::isBufferBlockDecorationDeprecated() {
```
sudonatalie, line 17: optional style nit: This should preferably be rewritten as `isBufferBlockDecorationAvailable()` then negate its usages (so as not to equate "deprecated" and "removed", and to avoid double negatives).

sudonatalie (changes requested):

tools/clang/lib/SPIRV/RemoveBufferBlockVisitor.cpp
MarijnS95: Would you then negate it here? We can't check for `isTargetEnvSpirv1.3OrBelow` (if that function is added) since `SPV_ENV_VULKAN_1_1` (etc) is a higher enum-value than `SPIRV_1_3_UNIVERSAL`, but still counts valid for `1_3`.

In other words, what would you turn this function body into? `return !featureManager.isTargetEnvSpirv1p4OrAbove();`?

tools/clang/lib/SPIRV/EmitVisitor.cpp
MarijnS95: I can drive-by update the original comment if desired, yes.

MarijnS95: Fwiw we could also follow `spirv-val` and use `requires SPIR-V version 1.3 or earlier`?

sudonatalie: Sure, both "is missing/removed after SPIR-V 1.3" or "requires SPIR-V 1.3 or earlier" are fine.

MarijnS95: Thanks, updated.

sudonatalie (approved):
Looks great, thanks!

PR #4371, opened 2022-04-01
[dxil2spv] Fix build warning

sudonatalie:
Address the "Policy CMP0023 is not set: Plain and keyword target_link_libraries signatures cannot be mixed." warning.


sudonatalie:
Just happened to notice this when the appveyor build was failing this morning.

jaebaek (approved):

PR #4378, opened 2022-04-06
Added a pass using GVN to eliminate redundant code without doing code motion.

adam-yang:
Without GVN, some basic ability to eliminate redundancy was lost. This change adds a pass that uses value numbering to eliminate redundant code without code motion.

### DIFF
--- a/lib/HLSL/DxilSimpleGVNHoist.cpp
+++ b/lib/HLSL/DxilSimpleGVNHoist.cpp
@@ -6,7 +6,7 @@
 // License. See LICENSE.TXT for details.                                     //
 //                                                                           //
 // A simple version of GVN hoist for DXIL.                                   //
-// Based on GVNHoist in LLVM 6.0.                                            //                                                                           //
+// Based on GVNHoist in LLVM 6.0.                                            //
 ///////////////////////////////////////////////////////////////////////////////
 
 #include "dxc/HLSL/DxilGenerationPass.h"
@@ -20,6 +20,10 @@
 #include "llvm/ADT/DenseSet.h"
 #include "llvm/IR/IntrinsicInst.h"
 #include "llvm/IR/CFG.h"
+
+#include "llvm/Analysis/PostDominators.h"
+
+#pragma optimize( "", off )
 
 using namespace llvm;
 using namespace hlsl;
@@ -442,6 +446,7 @@ namespace {
 //    r = tex.Sample(ss, uv) + 3;
 // }
 class DxilSimpleGVNHoist : public FunctionPass {
+  bool bOnlyAllowCompleteHoist = false;
 
 public:
   static char ID; // Pass identification, replacement for typeid
@@ -478,6 +483,10 @@ bool DxilSimpleGVNHoist::tryToHoist(BasicBlock *BB, BasicBlock *Succ0,
   for (Instruction &I : *Succ0) {
     uint32_t V = VT.lookupOrAdd(&I);
     VNtoInsts[V].emplace_back(&I);
+    // Also assign value numbers to the operands
+    for (Value *Operand : I.operands()) {
+      VT.lookupOrAdd(Operand);
+    }
   }
 
   std::vector<uint32_t> HoistCandidateVN;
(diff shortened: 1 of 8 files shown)


python3kgae:
If feels like we need to do all optimizations which can remove unused resource even when the optimizations are disabled thru command line option.

tools/clang/test/HLSLFileCheck/passes/hl/gvn-eliminate-region/gvn-eliminate-region.hlsl
```diff
@@ -0,0 +1,45 @@
+// RUN: %dxc -T ps_6_0 -E main /opt-disable gvn %s | FileCheck %s
```
dmpots, line 1: It would be good to add a few more tests here. 

1. Cases where the syntax is not identical, but is equivalent (e.g. foo=a+b, bar=b+a).
2. Cases where there are memory operations. Do we expect these to be eliminated?

It would also be good to add some negative tests. Ideally anywhere we have a `return false` in the `CompareOrBuild` we would have a test that exercises that condition.

lib/HLSL/DxilSimpleGVNHoist.cpp
```diff
@@ -585,3 +587,180 @@ FunctionPass *llvm::createDxilSimpleGVNHoistPass() {
...
+
+};
+
+bool DxilSimpleGVNEliminateRegion::runOnFunction(Function &F) {
+  PostDominatorTree *PDT = &getAnalysis<PostDominatorTree>();
+
+  ValueTable VT;
+
+  bool bChanged = false;
+  for (BasicBlock &BB : F) {
```
dmpots, line 724: I'm worried about the time complexity of this whole walk. It looks like an N^2 walk since we can visit a block multiple times (so if we have highly nested ifs we are going to be vising the same block many times).

It would be good to think of a way to structure it so that we only visit each block once.

lib/HLSL/DxilSimpleGVNHoist.cpp
```diff
@@ -585,3 +587,180 @@ FunctionPass *llvm::createDxilSimpleGVNHoistPass() {
...
+      continue;
+    if (!HasOnePred(Succ0))
+      continue;
+    if (!HasOnePred(Succ1))
+      continue;
+
+    BasicBlock *End = PDT->findNearestCommonDominator(Succ0, Succ1);
+
+    ValueNumberComparator Comparison;
+    if (!Comparison.BuildOrCompare(Succ0, End, VT, /*compare*/false))
```
dmpots, line 744: A nicer api here would be to expose `Build()` and `Compare()` functions. Internally they could call `BuildOrCompare()`

lib/Transforms/IPO/PassManagerBuilder.cpp
```diff
@@ -501,7 +501,10 @@ void PassManagerBuilder::populateModulePassManager(
     }
     // HLSL Change Ends
   }
+
   // HLSL Change Begins.
+  // Use value numbering to figure out if regions are equivalent, and branch to only one.
+  MPM.add(createDxilSimpleGVNEliminateRegionPass());
```
dmpots, line 507: If this is redundant with GVN, should we only do it when GVN is disabled?

dmpots (commented):
My main concern is the time complexity. We should think about if there is a better way to avoid repeated processing.

Also, needs an update to the PR description (the comment in the test case is good).

lib/Transforms/IPO/PassManagerBuilder.cpp
adam-yang: Done.

tools/clang/test/HLSLFileCheck/passes/hl/gvn-eliminate-region/gvn-eliminate-region.hlsl
```diff
@@ -0,0 +1,45 @@
...
+
+// CHECK: @main
```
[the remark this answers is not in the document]
adam-yang: Done.

adam-yang:
> If feels like we need to do all optimizations which can remove unused resource even when the optimizations are disabled thru command line option. Maybe we can skip this when -all_resources_bound is enabled?

The bug actually happens when gvn was explicitly disabled for performance reasons. I'm not sure if there's any downsides to running this pass anyway, it could find some stuff to delete.

lib/Transforms/IPO/PassManagerBuilder.cpp
adam-yang: I think it's fine to run them both, there's not much downsides to running this, it might catch some redundant code.

lib/HLSL/DxilSimpleGVNHoist.cpp
adam-yang: In this new version, the expensive thing, which is to go through every instruction, is limited to running only once (twice if a block is modified), so it's N at this point.

I don't know if it' actually worth it to try to not re-walk blocks when looking for regions.

tools/clang/test/HLSLFileCheck/passes/hl/gvn-eliminate-region/gvn-eliminate-region.hlsl
adam-yang: Added some new tests. Interestingly, simple commutative expressions work fine, but we can't really handle expressions that are identical from associativity. Maybe there's more opportunities there that we can explore later.

PR #4379, opened 2022-04-06
Complete builtin types when used as array params

llvm-beanz:
This is an odd edge case.

C doesn't require array element types to be complete because they decay
to pointers. In HLSL we don't decay to pointers, so we need them to be
complete, otherwise things trip up.


tools/clang/lib/Sema/SemaType.cpp
```diff
@@ -6734,6 +6734,17 @@ bool Sema::RequireCompleteTypeImpl(SourceLocation Loc, QualType T,
   while (const ConstantArrayType *Array
            = Context.getAsConstantArrayType(MaybeTemplate))
     MaybeTemplate = Array->getElementType();
+
+  // HLSL Change - Begin
+  // In HLSL we don't decay pointers to arrays, so we need to require
```
anupamachandra, line 6739: you meant decay arrays to pointers :)

pow2clk (approved):
Looks good!

PR #4381, opened 2022-04-08
Reenable NRVO when no precise and global coherent.

python3kgae:
Enable NRVO when no precise and global coherent to avoid extra memcpy created in clang codeGen.
Also propagate precise from the function to named return value.

### DIFF
--- a/tools/clang/lib/Sema/SemaStmt.cpp
+++ b/tools/clang/lib/Sema/SemaStmt.cpp
@@ -2710,7 +2710,7 @@ VarDecl *Sema::getCopyElisionCandidate(QualType ReturnType,
     // NRVO breaks on bool component type due to diff between
     // i32 memory and i1 register representation
     if (hlsl::IsHLSLVecMatType(ReturnType))
-      return false;
+      return nullptr;
     QualType ArrayEltTy = ReturnType;
     while (const clang::ArrayType *AT =
                Context.getAsArrayType(ArrayEltTy)) {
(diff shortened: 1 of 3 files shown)


tools/clang/lib/Sema/SemaStmt.cpp
```diff
@@ -2709,6 +2701,37 @@ VarDecl *Sema::getCopyElisionCandidate(QualType ReturnType,
...
+  // HLSL Change Begins: NRVO unsafe for a variety of cases in HLSL
+  // - vectors/matrix with bool component types
+  // - attributes not captured to QualType, such as precise and globallycoherent
+  // NRVO is also unnecessary for HLSL correctness or performance.
+  if (getLangOpts().HLSL) {
+    // exclude vectors/matrix (not treated as record type)
+    // NRVO breaks on bool component type due to diff between
+    // i32 memory and i1 register representation
+    if (hlsl::IsHLSLVecMatType(ReturnType))
+      return false;
```
llvm-beanz, line 2713: Returning `false` from a function that returns a pointer is not ideal, these should be `nullptr`.

llvm-beanz:
Can you please provide a change description that explains the motivation for this change. I don't see anything wrong with it, but I'm also unclear what the motivation or benefit is.

tools/clang/lib/Sema/SemaStmt.cpp
```diff
@@ -2709,6 +2701,38 @@ VarDecl *Sema::getCopyElisionCandidate(QualType ReturnType,
   if (!VD)
     return nullptr;
 
+  // HLSL Change Begins: NRVO unsafe for a variety of cases in HLSL
+  // - vectors/matrix with bool component types
+  // - attributes not captured to QualType, such as precise and globallycoherent
+  // NRVO is also unnecessary for HLSL correctness or performance.
```
tex3d, line 2707: Comment seems out-of-date now, since it's required to avoid extra memcpy and memcpy elimination.  I wish there was a way to fix that part (handle extra memcpy) though.

tex3d (approved):
LGTM

tools/clang/lib/Sema/SemaStmt.cpp
llvm-beanz: SROA should make the memcpy go away, but I suspect the version we're using in DXC doesn't match the case.

PR #4382, opened 2022-04-08
PIX: Incorrect index for CreateHandleFromBinding

jeffnn:
This mistake was upsetting a certain driver.  The UAV being declared is always u0 in space -2. Also changed the name of one struct type to make it more obvious what PIX had added.


pow2clk (approved):
Simple enough!

PR #4383, opened 2022-04-08
Added a pass to use GVN to remove redundant branches without doing code motion.

adam-yang:
Sometimes GVN is turned off deliberately to avoid code motion, which can lead to certain redundant code being left behind. This change adds a pass that uses value numbering to identify and remove identical code from two sides of a conditional branch.

### DIFF
--- a/lib/HLSL/DxilSimpleGVNHoist.cpp
+++ b/lib/HLSL/DxilSimpleGVNHoist.cpp
@@ -594,11 +594,32 @@ FunctionPass *llvm::createDxilSimpleGVNHoistPass() {
 INITIALIZE_PASS(DxilSimpleGVNHoist, "dxil-gvn-hoist",
                 "DXIL simple gvn hoist", false, false)
 
-
-///////////////////////////////////////////////////////////////////////////////
+//================================================================================
 //
-// Use value numbering to compare regions. If they are equivalent, branch
-// to only one side.
+// This pass tries to turn conditional branches to unconditional branches by
+// proving two sides of branch are equivalent using ValueTable and dominator
+// trees.
+// 
+// The algorithm:
+//
+// - Find any conditional branch 'Br' with successors 'Succ0' and 'Succ1'
+// - Find the common destination 'End' of the branches.
+// - Look at predecessors of 'End'. Find two predecessors 'P0' and 'P1',
+//   where 'Succ0' dominates 'P0', and 'Succ1' dominates 'P1'
+//     + This means the condition of 'Br' determinates whether we end up taking
+//       'P0' or 'P1' to 'End'
+// - Using ValueTable, compare if every pair of incoming values from P0 and P1
+//   is identical for any PHIs in 'End'
+// - Make sure there is no side effect or loop between 'Br' and 'End'
+// - If all above checks succeed, replace 'Br' with with an unconditional branch
+//   to Succ0
+// 
+// The current state of the pass is pretty limited. If incoming values from P0
+// and P1 are dependent on any PHIs defined between Br and End, then the pass
+// will fail to simplify the branch. If there are any side effects within the
+// region, the pass will also fail. It's possible to handling these cases, but
+// require proving the two sides of the branch have equivalent control flow,
+// which is non-trivial, and will be left to a later date.
 //
 namespace {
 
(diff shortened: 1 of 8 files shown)


lib/HLSL/DxilSimpleGVNHoist.cpp
```diff
@@ -585,3 +593,197 @@ FunctionPass *llvm::createDxilSimpleGVNHoistPass() {
...
+
+  bool CheckRegionForSideEffectsOrLoops(BasicBlock *Begin, BasicBlock *End /*Non inclusive*/);
+
+  std::unordered_map<BasicBlock *, bool> BlockHasSideEffects;
+  bool MayHaveSideEffects(BasicBlock *BB) {
+    auto It = BlockHasSideEffects.find(BB);
+    if (It != BlockHasSideEffects.end())
+      return It->second;
+
+    for (Instruction &I : *BB) {
```
dmpots, line 639: Can simplify this so that there is only one return and write to the map like

```
bool HasSideEffects = false;
for (...)
  if (...)
   HasSideEffects = true;
   break;   
}
BlockHasSideEffects[BB] = HasSideEffects;
return HasSideEffects;
```

lib/HLSL/DxilSimpleGVNHoist.cpp
```diff
@@ -585,3 +593,197 @@ FunctionPass *llvm::createDxilSimpleGVNHoistPass() {
...
+
+  void getAnalysisUsage(AnalysisUsage &AU) const override {
+    AU.addRequired<PostDominatorTree>();
+    AU.addRequired<DominatorTreeWrapperPass>();
+  }
+};
+
+char DxilSimpleGVNEliminateRegion::ID = 0;
+
+bool DxilSimpleGVNEliminateRegion::CheckRegionForSideEffectsOrLoops(BasicBlock *Begin, BasicBlock *End /*Non inclusive*/) {
```
dmpots, line 669: Since this returns a bool a better name might be `RegionHasSideEffectOrLoop`

lib/HLSL/DxilSimpleGVNHoist.cpp
```diff
@@ -585,3 +593,197 @@ FunctionPass *llvm::createDxilSimpleGVNHoistPass() {
...
+      // Goes back into the region. Give up.
+      if (Seen.count(Succ))
+        return false;
+      Worklist.push_back(Succ);
+    }
+  }
+  return true;
+}
+
+bool DxilSimpleGVNEliminateRegion::ProcessBB(BasicBlock &BB, ValueTable &VT, DominatorTree *DT, PostDominatorTree *PDT) {
```
dmpots, line 694: This seems to still have the problem of repeatedly walking the CFG for each region. If we have a graph that contains a bunch of nested diamond graphs then it seems like would could run into pathological compile times.


```
                                A
                             /     \
                            C        G
                          /   \    /   \
                         D     E  H     I
                         \     /   \   /
                            F        J
                              \     /
                                 K
```

lib/HLSL/DxilSimpleGVNHoist.cpp
```diff
@@ -585,3 +593,197 @@ FunctionPass *llvm::createDxilSimpleGVNHoistPass() {
...
+    if (BB == End)
+      continue;
+
+    if (MayHaveSideEffects(BB))
+      return false;
+
+    Seen.insert(BB);
+    for (BasicBlock *Succ : successors(BB)) {
+      // Goes back into the region. Give up.
+      if (Seen.count(Succ))
```
dmpots, line 686: If the region is wholly contained in a loop you would remove it, but there is no test case for that. We should add one.

lib/HLSL/DxilSimpleGVNHoist.cpp
adam-yang: Done.

lib/HLSL/DxilSimpleGVNHoist.cpp
adam-yang: Done.

lib/HLSL/DxilSimpleGVNHoist.cpp
adam-yang: Done.

dmpots (approved):

lib/HLSL/DxilSimpleGVNHoist.cpp
adam-yang: I did some stress test, using a diamond graph 13 levels deep (8192 branches). The runtime is completely dominated by the linear initial attempt to identify diamond graphs. We'll have to push the program size pretty far to have that N^2 walk be a problem.

PR #4388, opened 2022-04-11
[dxil2spv] Support shl instruction

sudonatalie:
Add support for translating the first simple binary operator
instruction, shift left, from DXIL to SPIR-V.

### DIFF
--- a/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
+++ b/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
@@ -385,6 +385,14 @@ void Translator::createBinaryOpInstruction(llvm::BinaryOperator &instruction) {
   case llvm::Instruction::Shl: {
     // Value to be shifted.
     spirv::SpirvInstruction *val = instructionMap[instruction.getOperand(0)];
+    if (!val) {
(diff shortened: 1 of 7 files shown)


tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -361,6 +376,34 @@ void Translator::createThreadIdInstruction(llvm::CallInst &instruction) {
...
 
+void Translator::createBinaryOpInstruction(llvm::BinaryOperator &instruction) {
+  llvm::Instruction::BinaryOps opcode = instruction.getOpcode();
+  spirv::SpirvBinaryOp *result;
+
+  switch (opcode) {
+  // Shift left instruction.
+  case llvm::Instruction::Shl: {
+    // Value to be shifted.
+    spirv::SpirvInstruction *val = instructionMap[instruction.getOperand(0)];
```
jaebaek, line 387: Do you think we have to add an error check here?
I am just worried the case we don't have a key `instruction.getOperand(0)` in `instructionMap`.

jaebaek (approved):
LGTM with a nit

PR #4389, opened 2022-04-12
[dxil2spv] Translate createHandle and bufferLoad

sudonatalie:
Add support for translating createHandle and bufferLoad DXIL operations
to SPIR-V instructions. The most significant limitation of this current
implementation is the naive translation of descriptor set and binding
numbers, but it is sufficient for simple passthrough shaders.

Some error handling is also simplified with a wrapper function for the
frequently used instructionMap accesses.

### DIFF
--- a/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
+++ b/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
@@ -573,7 +573,7 @@ DiagnosticBuilder Translator::emitError(const char (&message)[N],
   std::string str;
   llvm::raw_string_ostream os(str);
   value.print(os);
-  emitError(message) << os.str();
+  return emitError(message) << os.str();
 }
 
 } // namespace dxil2spv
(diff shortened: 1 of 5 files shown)


sudonatalie:
@jaebaek This is now ready for review

jaebaek (approved):

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -412,6 +413,88 @@ void Translator::createBinaryOpInstruction(llvm::BinaryOperator &instruction) {
...
+      resourceMap[{resourceClass, resourceRangeId}];
+  if (!inputVar) {
+    emitError("No resource found corresponding to handle: %0", instruction);
+    return;
+  }
+
+  instructionMap[&instruction] = inputVar;
+}
+
+void Translator::createBufferLoadInstruction(llvm::CallInst &instruction) {
```
jaebaek, line 437: nit: if you provided an example DXIL instruction and its translation result i.e., output SPIR-V as a comment, it would be easier to understand.

sudonatalie: Done

PR #4390, opened 2022-04-12
[SPIR-V] Add option to rename SPIR-V entry point

### LINKED ISSUE #2972: [SPIR-V] SPIR-V module entrypoint renaming
In glslang, there is an option to rename the entrypoint in the shader module.  For example, if you have a vertex shader with the name "MainVs", the entrypoint can be renamed to "main" using:

--source-entrypoint MainVs -e main

The OpEntryPoint will then be "main".  For us, this simplifies a bit of runtime code.  It would be useful if this feature could be supported in dxc.

### LINKED ISSUE #4356: [SPIR-V] Ability to control OpEntryPoint export name
When using -E foo, it seems like the entry point name is also forced to foo, and I can't find any option to export all entry points as just "main" (target engine in question does not use multiple entry point SPIR-V).

If it does not exist, it would be good to have an option to control the output entry point name, since I wouldn't have to reflect out the entry point name from SPIR-V to be able to use it in a pipeline.

sudonatalie:
Allow renaming the SPIR-V entry point name from the default HLSL entry
point name with `-fspv-entrypoint-name`.

Fixes #2972
Fixes #4356

### DIFF
--- a/include/dxc/Support/SPIRVOptions.h
+++ b/include/dxc/Support/SPIRVOptions.h
@@ -84,6 +84,7 @@ struct SpirvCodeGenOptions {
   llvm::SmallVector<llvm::StringRef, 4> optConfig;
   std::vector<std::string> bindRegister;
   std::vector<std::string> bindGlobals;
+  llvm::StringRef entrypointName;
 
   bool signaturePacking; ///< Whether signature packing is enabled or not
 
--- a/include/dxc/Support/SPIRVOptions.h
+++ b/include/dxc/Support/SPIRVOptions.h
@@ -84,7 +84,7 @@ struct SpirvCodeGenOptions {
   llvm::SmallVector<llvm::StringRef, 4> optConfig;
   std::vector<std::string> bindRegister;
   std::vector<std::string> bindGlobals;
-  llvm::StringRef entrypointName;
+  std::string entrypointName;
 
   bool signaturePacking; ///< Whether signature packing is enabled or not
 
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -549,8 +549,8 @@ SpirvEmitter::SpirvEmitter(CompilerInstance &ci)
     : theCompilerInstance(ci), astContext(ci.getASTContext()),
       diags(ci.getDiagnostics()),
       spirvOptions(ci.getCodeGenOpts().SpirvOptions),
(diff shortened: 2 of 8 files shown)


include/dxc/Support/SPIRVOptions.h
```diff
@@ -84,6 +84,7 @@ struct SpirvCodeGenOptions {
   llvm::SmallVector<llvm::StringRef, 4> optConfig;
   std::vector<std::string> bindRegister;
   std::vector<std::string> bindGlobals;
+  llvm::StringRef entrypointName;
```
jaebaek, line 87: I am worried that later this `llvm::StringRef` loses the actual string object i.e., dangling pointer.
I am surprised other members here use `llvm::StringRef`.

If there is no strong guarantee that it will not be a dangling pointer, could you please change it to `std::string`?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -748,9 +748,18 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
     // TODO: assign specific StageVars w.r.t. to entry point
     const FunctionInfo *entryInfo = workQueue[i];
     assert(entryInfo->isEntryFunction);
+
```
jaebaek, line 751: Oh .. this method becomes long. Could you please separate this part?
e.g.,
```
llvm::StringRef SpirvEmitter::getEntryPointName() {
    llvm::StringRef entrypointName = entryInfo->funcDecl->getName();
    // If this is the -E HLSL entrypoint and -fspv-entrypoint-name was set,
    // rename the SPIR-V entrypoint.
    if (entrypointName ==
            theCompilerInstance.getCodeGenOpts().HLSLEntryFunction &&
        !spirvOptions.entrypointName.empty()) {
      entrypointName = spirvOptions.entrypointName;
    }
  return entrypointName;
}
```

jaebaek: (sorry for making it long before .. let's make it shorter)

include/dxc/Support/SPIRVOptions.h
sudonatalie: Done

tools/clang/lib/SPIRV/SpirvEmitter.cpp
sudonatalie: Done. Also noticed there was already a member variable for `HLSLEntryFunction` so I cleaned that usage up.

jaebaek (approved):

PR #4395, opened 2022-04-16
Od fixes for April 2022

adam-yang:
- Reordered passes so now ExilEraseDeadRegions happens AFTER DxilRemoveDeadBlocks. ExilEraseDeadRegions may depend on simplifications made by DxilRemoveDeadBlocks more than the reverse is true. The only thing that DxilRemoveDeadBlocks needed from ExilEraseDeadRegions was removing dead allocas that only had stores but no loads. This change also factors it out to a helper that both passes can call.
- Fixed the issue that DxilValueCache wasn't able to handle switch statements where each case had more than on basic block. This change adds reachability analysis for it similar to normal branches.
- Fixed the issue that DxilRemoveDeadBlocks removes all instructions that could be replaced with constants, but did not perform DCE on those instructions' operands, which are not guaranteed to also be constants. For example, `fmul %value, 0`. Those instructions could hold on to PHIs and prevent ExilEraseDeadRegions from deleting certain regions.

### DIFF
--- a/include/dxc/DXIL/DxilUtil.h
+++ b/include/dxc/DXIL/DxilUtil.h
@@ -169,6 +169,10 @@ namespace dxilutil {
     return mdv_users_begin(V) == mdv_users_end(V);
   }
 
+  /// Finds all allocas that only have stores and delete them.
+  /// These allocas hold on to values that do not contribute to the
+  /// shader's results.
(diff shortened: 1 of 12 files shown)


dmpots:
Can we split these into separate smaller PRs for easier reviewing? Or do all the changes depend on each other?

PR #4397, opened 2022-04-19
Moved DxilEraseDeadRegions to after DxilRemoveDeadBlocks

adam-yang:
Reordered passes so now ExilEraseDeadRegions happens AFTER DxilRemoveDeadBlocks.

ExilEraseDeadRegions may depend on simplifications made by DxilRemoveDeadBlocks more than the reverse. The only thing that DxilRemoveDeadBlocks needed from ExilEraseDeadRegions was removing dead allocas that only had stores but no loads. This change also factors it out to a helper that both passes can call.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_dead_region_after_remove_dead_blocks.hlsl
+++ b/tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_dead_region_after_remove_dead_blocks.hlsl
@@ -12,7 +12,7 @@
 //
 
 // CHECK: @main
-// CHECK-NOT: call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(
+// CHECK-NOT: call %dx.types.CBufRet.f32 @dx.op.cbufferLoad.f32(
 
 cbuffer cb : register(b0) {
   float foo;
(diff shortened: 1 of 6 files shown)


dmpots (approved):

tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_dead_region_after_remove_dead_blocks.hlsl
```diff
@@ -0,0 +1,36 @@
...
+// The way the Od dead code removal passes were arranged was unable to clean up this pattern.
+// EraseDeadRegion would run first, but unable to delete 'if (foo)' because of outflowing
+// value 'cond' still being used.
+//
+// If we run RemoveDeadBlocks first, however, it would delete the branch `if (!my_cond)` and
+// relieve the use of 'cond', allowing EraseDeadRegion to delete `if (foo)`.
+//
+
+// CHECK: @main
+// CHECK-NOT: call %dx.types.CBufRet.f32 @dx.op.cbufferLoad.f32(
```
dmpots, line 15: Wouldn't this test pass even without your change (since it uses the Legacy load by default).

adam-yang: Oops

PR #4399, opened 2022-04-19
RawGather ExecutionTest

pow2clk:
Creates a texture interface that represents an anonymous texture to be
populated with valid values then raw gathered from a shader and copied
to a UAV whose contents are compared with the original to verify correct
behavior. Derived textures use templating to share code that varies
depending on the type, type size, format characteristics (norm, signed,
etc), and number of channels.

This includes a fallback mechanism that uses a standard gather or loads
to retrieve the expected results from the aliased texture.

Some incidental refinements and corrections to the resource and view
creation functions

### DIFF
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -308,6 +308,7 @@ public:
   TEST_METHOD(ATOProgOffset);
   TEST_METHOD(ATOSampleCmpLevelTest);
   TEST_METHOD(ATOWriteMSAATest);
+  TEST_METHOD(ATORawGather);
   TEST_METHOD(AtomicsTest);
   TEST_METHOD(Atomics64Test);
   TEST_METHOD(AtomicsRawHeap64Test);
@@ -559,6 +560,20 @@ public:
     return GetTestParamBool(L"SaveImages");
   }
 
+  // Base class used by raw gather test for polymorphic assignments
+  struct RawGatherTexture {
+    // Set Element <i> to a format-appropriate value derived from 2D coords <x,y>
+    virtual void SetElement(int i, int x, int y) = 0;
+    // Retrieve pointer to the elements
+    virtual void *GetElements() = 0;
+    // Get dimensions/format
+    virtual unsigned GetXDim() = 0;
+    virtual unsigned GetYDim() = 0;
+    virtual DXGI_FORMAT GetFormat() = 0;
+  };
+
+  template<typename GatherType>
+  void DoRawGatherTest(ID3D12Device *pDevice, RawGatherTexture *rawTex, DXGI_FORMAT viewFormat);
   void RunResourceTest(ID3D12Device *pDevice, const char *pShader, const wchar_t *sm, bool isDynamic);
 
   template <class T1, class T2>
@@ -893,7 +908,7 @@ public:
 
   void CreateTestResources(ID3D12Device *pDevice,
                            ID3D12GraphicsCommandList *pCommandList, LPCVOID values,
-                           UINT32 valueSizeInBytes, D3D12_RESOURCE_DESC resDesc,
+                           UINT64 valueSizeInBytes, D3D12_RESOURCE_DESC resDesc,
                            ID3D12Resource **ppResource,
                            ID3D12Resource **ppUploadResource,
                            ID3D12Resource **ppReadBuffer = nullptr) {
@@ -949,7 +964,7 @@ public:
 
   void CreateTestUavs(ID3D12Device *pDevice,
                       ID3D12GraphicsCommandList *pCommandList, LPCVOID values,
-                      UINT32 valueSizeInBytes, ID3D12Resource **ppUavResource,
+                      UINT64 valueSizeInBytes, ID3D12Resource **ppUavResource,
                       ID3D12Resource **ppUploadResource = nullptr,
                       ID3D12Resource **ppReadBuffer = nullptr) {
     D3D12_RESOURCE_DESC bufferDesc = CD3DX12_RESOURCE_DESC::Buffer(valueSizeInBytes, D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS);
@@ -1037,8 +1052,8 @@ public:
   }
 
   void CreateTex2DSRV(ID3D12Device *pDevice, CD3DX12_CPU_DESCRIPTOR_HANDLE &heapStart,
-                      UINT numElements, DXGI_FORMAT format, const CComPtr<ID3D12Resource> pResource) {
-    CreateSRV(pDevice, heapStart, format, D3D12_SRV_DIMENSION_TEXTURE2D, numElements, 0, pResource);
+                      DXGI_FORMAT format, const CComPtr<ID3D12Resource> pResource) {
+    CreateSRV(pDevice, heapStart, format, D3D12_SRV_DIMENSION_TEXTURE2D, 0/*numElements*/, 0/*stride*/, pResource);
   }
 
   void CreateUAV(ID3D12Device *pDevice, CD3DX12_CPU_DESCRIPTOR_HANDLE &baseHandle,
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -8929,15 +9504,15 @@ void ExecutionTest::RunResourceTest(ID3D12Device *pDevice, const char *pShader,
...
+  CreateTex2DSRV(pDevice, baseHandle, DXGI_FORMAT_R32_FLOAT, pSRVResources[2]);
   // Create UAVs
   CreateRawUAV(pDevice, baseHandle, valueSize, pUAVResources[0]);
   CreateStructUAV(pDevice, baseHandle, valueSize, sizeof(float), pUAVResources[1]);
   CreateTypedUAV(pDevice, baseHandle, valueSize, DXGI_FORMAT_R32_FLOAT, pUAVResources[2]);
   CreateTex1DUAV(pDevice, baseHandle, DXGI_FORMAT_R32_FLOAT, pUAVResources[3]);
 
   D3D12_FILTER filters[] = {D3D12_FILTER_MIN_MAG_LINEAR_MIP_POINT, D3D12_FILTER_COMPARISON_MIN_MAG_LINEAR_MIP_POINT};
-  float borderColors[] = {30.0, 31.0};
+  float borderColors[] = {30.0, 31.0, 0.0, 0.0};
```
tex3d, line 9515: It's not obvious, but borderColors is indexed by sampler (count = NumSamplers), not RGBA index.  So, what's the purpose of expanding this array?  Though this *is* a bit confusing when using this "CreateDefaultSamplers" function.  Perhaps `borderColors` should be renamed to `perSamplerBorderColorScalars`, or some such.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -4146,7 +4162,566 @@ TEST_F(ExecutionTest, ATOSampleCmpLevelTest) {
...
+// that has the same element size and initializes them with various values,
+// The shader code copies the results of raw gather to an unsigned integer UAV
+// The UAV contents are compared to the values assigned to the texture
+TEST_F(ExecutionTest, ATORawGather) {
+
+  WEX::TestExecution::SetVerifyOutput verifySettings(WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
+
+  CComPtr<ID3D12Device> pDevice;
+#ifdef RAWGATHER_FALLBACK
+  D3D_SHADER_MODEL sm = D3D_SHADER_MODEL_6_6;
```
tex3d, line 4583: Is it possible to make fallback work on an earlier shader model?  What SM 6.6 feature does the fallback path require?

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -4146,7 +4162,566 @@ TEST_F(ExecutionTest, ATOSampleCmpLevelTest) {
...
+
+  // Create shaders
+#ifdef RAWGATHER_FALLBACK
+  const wchar_t *target = L"cs_6_6";
+#else
+  const wchar_t *target = L"cs_6_7";
+#endif
+
+  CComPtr<ID3D12PipelineState> pPSO;
+  CreateComputePSO(pDevice, pRootSignature, pShader, target, &pPSO);
```
tex3d, line 4497: Missing `-enable-16bit-types` option - how does this compile?

tools/clang/unittests/HLSL/ExecutionTest.cpp
tex3d: I was successfully able to run the fallback with SM 6.2 (when adding `-enable-16bit-types` option).

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -4146,7 +4162,566 @@ TEST_F(ExecutionTest, ATOSampleCmpLevelTest) {
...
+  CComPtr<ID3D12Device> pDevice;
+#ifdef RAWGATHER_FALLBACK
+  D3D_SHADER_MODEL sm = D3D_SHADER_MODEL_6_6;
+#else
+  D3D_SHADER_MODEL sm = D3D_SHADER_MODEL_6_7;
+#endif
+  if (!CreateDevice(&pDevice, sm))
+      return;
 
+  if (DoesDeviceSupportAdvancedTexOps(pDevice)) {
```
tex3d, line 4590: Missing `!`, and exclude check in fallback?

tex3d: And return from block here when skipping.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -4146,7 +4162,566 @@ TEST_F(ExecutionTest, ATOSampleCmpLevelTest) {
...
+    }
+    RGBA[i].SetChannels((UINT)fr, (UINT)fg, (UINT)fb, (UINT)fa);
+  }
+  virtual void *GetElements() { return (void*)RGBA; }
+  virtual unsigned GetXDim() { return xdim; }
+  virtual unsigned GetYDim() { return ydim; }
+  virtual DXGI_FORMAT GetFormat() override { return m_format; };
+};
+
+#define RAWGATHER_FALLBACK // Enable to use pre-6.7 fallback mechanisms to vet raw gather tests
```
tex3d: Fallback worked for me with the following changes, but I don't think you want to commit with fallback enabled.
```
diff --git a/tools/clang/unittests/HLSL/ExecutionTest.cpp b/tools/clang/unittests/HLSL/ExecutionTest.cpp
index ee61606ff..18defc085 100644
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -4488,13 +4488,14 @@ void ExecutionTest::DoRawGatherTest(ID3D12Device *pDevice, RawGatherTexture *raw

   // Create shaders
 #ifdef RAWGATHER_FALLBACK
-  const wchar_t *target = L"cs_6_6";
+  const wchar_t *target = L"cs_6_2";
 #else
   const wchar_t *target = L"cs_6_7";
 #endif
+  const wchar_t *options[] = {L"-enable-16bit-types"};

   CComPtr<ID3D12PipelineState> pPSO;
-  CreateComputePSO(pDevice, pRootSignature, pShader, target, &pPSO);
+  CreateComputePSO(pDevice, pRootSignature, pShader, target, &pPSO, options, _countof(options));

   // Reset commandlist to shader PSO
   VERIFY_SUCCEEDED(pCommandList->Reset(pCommandAllocator, pPSO));
@@ -4580,17 +4581,20 @@ TEST_F(ExecutionTest, ATORawGather) {

   CComPtr<ID3D12Device> pDevice;
 #ifdef RAWGATHER_FALLBACK
-  D3D_SHADER_MODEL sm = D3D_SHADER_MODEL_6_6;
+  D3D_SHADER_MODEL sm = D3D_SHADER_MODEL_6_2;
 #else
   D3D_SHADER_MODEL sm = D3D_SHADER_MODEL_6_7;
 #endif
   if (!CreateDevice(&pDevice, sm))
       return;

-  if (DoesDeviceSupportAdvancedTexOps(pDevice)) {
+#ifndef RAWGATHER_FALLBACK
+  if (!DoesDeviceSupportAdvancedTexOps(pDevice)) {
     WEX::Logging::Log::Comment(L"Device does not support Advanced Texture Operations.");
     WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+    return;
   }
+#endif

   static const int NumThreadsX = 32;
   static const int NumThreadsY = 32;
```

tex3d (changes requested):
The few changes I that are needed before merging:
- `-enable-16bit-types` option for shader compilation
- `!` before `DoesDeviceSupportAdvancedTexOps(pDevice)`, and return in block after skip.
- comment out `#define RAWGATHER_FALLBACK`

Other nits/preferences:
- change fallback to use SM 6.2 for maximum compatibility
- rename borderColors argument to avoid confusion (and potentially reverse unnecessary array size change in other code).

tools/clang/unittests/HLSL/ExecutionTest.cpp
pow2clk: `D3D12_SAMPLER_DESC` has the element `BorderColor` which is an array of 4. `CreateDefaultSamplers` tried to initialize all four from this parameter, leading to garbage that didn't harm anything to my knowledge, but it was wrong.

I don't think I follow what you're saying about indexing by sampler and not RGBA index. I'm happy to rename it, but I don't think I see the confusion yet. Probably because I created the function 😓

tex3d: Because this array doesn't map to the 4 BorderColor array elements in D3D12_SAMPLER_DESC.  In `CreateDefaultSamplers`, this array is interpreted as an array of scalars, one per sampler.

In this assignment: `sampDesc.BorderColor[j] = borderColors[i];`, `i` is the sampler index, while `j` is the RGBA BorderColor array index.  Each scalar is replicated across all RGBA components for one sampler's BorderColor.  This code should not have been reading off the end of the array with `NumSamplers == 2`.

That's what's confusing about the name.  Intuitively, this looks like it would be RGBA, but it's not used that way.

pow2clk: ah yes! I misunderstood my own code. I'll fix it

pow2clk:
appveyor timed out 😒

tex3d (approved):
LGtM!

PR #4401, opened 2022-04-19
Removed all instruction made dead by their users being simplified in DxilRemoveDeadBlocks

adam-yang:
In some cases, DxilValueCache can simplify instructions where not all its operands can be simplified, for example:
```
%my_actually_dead_inst = ...
%my_const = fmul 0.0, %my_actually_dead_inst
```
`%my_const` can be simplified to 0, which the simplification pass removes, but `%my_actually_dead_inst` is incorrectly marked as being used by the output.

This change fixes this bug, and also takes extra care to preserve debug info for all deleted instructions that may be reduced to constant.

### DIFF
--- a/lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
+++ b/lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
@@ -278,8 +278,11 @@ struct ValueDeleter {
         LLVMContext &Ctx = F.getContext();
         for (User *U : DbgValueF->users()) {
           DbgValueInst *DbgVal = cast<DbgValueInst>(U);
-          if (Constant *C = DVC->GetConstValue(DbgVal->getValue())) {
-            DbgVal->setArgOperand(0, MetadataAsValue::get(Ctx, ValueAsMetadata::get(C)));
+          Value *Val = DbgVal->getValue();
+          if (Val) {
+            if (Constant *C = DVC->GetConstValue(DbgVal->getValue())) {
+              DbgVal->setArgOperand(0, MetadataAsValue::get(Ctx, ValueAsMetadata::get(C)));
+            }
           }
         }
       }
--- a/lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
+++ b/lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
@@ -22,6 +22,7 @@
 #include "llvm/Analysis/DxilValueCache.h"
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/ADT/SetVector.h"
+#include "llvm/IR/DebugInfo.h"
 
 #include "dxc/DXIL/DxilMetadataHelper.h"
 #include "dxc/DXIL/DxilOperations.h"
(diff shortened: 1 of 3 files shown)


tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_constant_dce.hlsl
```diff
@@ -0,0 +1,27 @@
+// RUN: %dxc %s -T ps_6_0 -Od | FileCheck %s
+
+// This test verifies the fix for a deficiency in RemoveDeadBlocks where:
```
dmpots, line 3: Seems like this is a good candidate to also have a "unit" test for to validate that the DxilRemoveDeadBlocks pass is working as expected.

lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
```diff
@@ -247,9 +248,39 @@ struct ValueDeleter {
...
+              Add(OpI);
+          }
+        }
+      }
+    }
+
+    // Go through all dbg.value and see if we can replace their value with constant.
+    // As we go and delete all non-contributing values, we want to preserve as much debug
+    // info as possible.
+    if (llvm::hasDebugInfo(*F.getParent())) {
```
dmpots, line 275: This part seems to not be covered by a test. Can we add a test to verify its working as expected?

dmpots (approved):

lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
adam-yang: Since we're deleting more instructions now, without this part some of the other tests started failing.

tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_constant_dce.hlsl
adam-yang: Like an .ll test using opt?

dmpots: Yeah, that's the idea.

adam-yang: Done.

lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
```diff
@@ -247,9 +248,39 @@ struct ValueDeleter {
...
-            Add(OpI);
+        for (unsigned i = 0; i < I->getNumOperands(); i++) {
+          Value *op = I->getOperand(i);
+          if (Instruction *OpI = dyn_cast<Instruction>(op)) {
+            // If this operand could be reduced to a constant, stop adding all
+            // its operands. Otherwise, we could unintentionally hold on to
+            // dead instructions like:
+            // 
+            //   %my_actually_dead_inst = ...
+            //   %my_const = fmul 0.0, %my_actually_dead_inst
```
tex3d, line 259: Presumably, this would only reduce when `fast` flags used, right?

tex3d (approved):
LGTM!

lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
```diff
@@ -304,6 +330,18 @@ static bool DeleteNonContributingValues(Function &F, DxilValueCache *DVC) {
   return Changed;
 }
 
+static void EnsureDxilModule(Module *M) {
+  if (M->HasDxilModule())
+    return;
+  for (Function &F : *M) {
+    if (OP::IsDxilOpFunc(&F)) {
```
tex3d, line 337: Wait, so this only happens if there are dxil intrinsic calls?  This seems a bit subtle.  Is that intentional?  Is this to avoid doing so on HL module or something?

lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
adam-yang: Yes.

lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
adam-yang: > Is this to avoid doing so on HL module or something?
That's right. The thought is maybe this pass should be usable at any stage. The real reason dxil module is needed is because instruction simplify asserts when it hits a dx op but there's no dxil module. Maybe that's what needs to be fixed instead.

PR #4403, opened 2022-04-20
Taught DxilValueCache to handle complex switch statements.

adam-yang:
- Taught DxilValueCache to handle switch statements where each case has more than just a single block.
- Fixed `DxilValueCache::ResetUnknowns` just setting value association to nullptr, which was entirely incorrect. Now it correctly deletes the value association that have been marked as unknown.
- Deleted the idea of 'AlwaysReachable' from DxilValueCache, which is unused, ill-defined, and potentially dangerous.
- Improved the dump function for 'DxilValueCache' for better debugging.

### DIFF
--- a/lib/Analysis/DxilValueCache.cpp
+++ b/lib/Analysis/DxilValueCache.cpp
@@ -56,18 +56,8 @@ bool IsEntryBlock(const BasicBlock *BB) {
   return BB == &BB->getParent()->getEntryBlock();
 }
 
-void DxilValueCache::MarkAlwaysReachable(BasicBlock *BB) {
-  ValueMap.Set(BB, ConstantInt::get(Type::getInt1Ty(BB->getContext()), 1));
-}
 void DxilValueCache::MarkUnreachable(BasicBlock *BB) {
   ValueMap.Set(BB, ConstantInt::get(Type::getInt1Ty(BB->getContext()), 0));
-}
-
-bool DxilValueCache::IsAlwaysReachable_(BasicBlock *BB) {
-  if (Value *V = ValueMap.Get(BB))
-    if (IsConstantTrue(V))
-      return true;
-  return false;
 }
 
 bool DxilValueCache::MayBranchTo(BasicBlock *A, BasicBlock *B) {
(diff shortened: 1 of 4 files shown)


tools/clang/test/HLSLFileCheck/samples/d3d11/TessellatorCS40_TessellateIndicesCS.hlsl
```diff
@@ -581,6 +581,7 @@ void main( uint3 DTid : SV_DispatchThreadID, uint3 Gid : SV_GroupID, uint GI : S
                 else
                 {
                     int ring = GetRingFromIndexStitchRegular(true, DIAGONALS_MIRRORED, processedTessFactors.numPointsForOutsideInside.w, index_id - num.z);
+                    DebugOutput[DTid.x+1] = int4(ring,0,0,0); // HLSL Change: Use that output so the value defines it doens't get deleted.
```
tex3d, line 584: I don't understand the need for this added line.

lib/Analysis/DxilValueCache.cpp
```diff
@@ -180,6 +172,40 @@ Value *DxilValueCache::ProcessAndSimplify_PHI(Instruction *I, DominatorTree *DT)
...
+        FoundCase = true;
+      }
+      else {
+        if (Succ->getSinglePredecessor())
+          MarkUnreachable(Succ);
+      }
+    }
+
+    BasicBlock *DefaultSucc = Sw->getDefaultDest();
+    if (FoundCase && DefaultSucc->getSinglePredecessor()) {
```
tex3d, line 201: Could the default dest be the same as the successor for the found case in some instance?  At which point it would be invalid to mark it unreachable here?

adam-yang: If the switch has multiple cases branching to the same BB, BB->getSinglePredecessor() returns NULL, because it checks if BB has a single Use, which in this case it would have multiple, each being an operand of the switch statement. But this code is a bit confusing... I'll revise it.

tools/clang/test/HLSLFileCheck/samples/d3d11/TessellatorCS40_TessellateIndicesCS.hlsl
adam-yang: The test was checking for a Sqrt in this function, but with this new change, this calls turns out to be not needed, so i'm artificially creating a use for the call so it sticks around.

adam-yang: Undid this change. Turns out it was broken by an earlier iteration, which was incorrect.

tex3d (approved):
LGTM!

PR #4405, opened 2022-04-20
Handle UDT ptr collisions in FixOverloadNames

tex3d:
FixOverloadNames renames intrinsics to match renamed overload types,
which happen as a result of loading modules with conflicting types into
the same llvm context.  A conflicting type would be one that has the same
name but is a different type.

UDT arguments are always pointer types, so when the intrinsic overload is
based on a UDT argument, the type retrieved by GetOverloadType will be a
pointer.  The code did not handle this case, skipping the overload name fix.

This change adds the check for pointer type so UDT overloads will be fixed.

The test links two libraries with a conflicting UDT by the same name, used
in a TraceRay call.  This previously crashed during bitcode writing because
of invalid IR.

### DIFF
--- a/lib/DXIL/DxilOperations.cpp
+++ b/lib/DXIL/DxilOperations.cpp
@@ -1050,7 +1050,7 @@ void OP::FixOverloadNames() {
       CallInst *CI = cast<CallInst>(*F.user_begin());
       DXIL::OpCode opCode = OP::GetDxilOpFuncCallInst(CI);
       llvm::Type *Ty = OP::GetOverloadType(opCode, &F);
-      if (isa<StructType>(Ty)) {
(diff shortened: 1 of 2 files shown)


dmpots:
LGTM. I don't know what udt means though so it would be good to define it once in your message like 

"blah blah Ufdsa Ddfsa Tdfa (UDT) blah blah blah"

dmpots (approved):

tex3d:
> LGTM. I don't know what udt means though so it would be good to define it once in your message like
> 
> "blah blah Ufdsa Ddfsa Tdfa (UDT) blah blah blah"

Oh, sorry, UDT = User Defined Type 😄

PR #4410, opened 2022-04-23
Add pass to remove redundant UAV copy caused by copy-in copy-out.

python3kgae:
Transform
  Tmp = uav[i]; // UavLdCpy
  ...
  Update Tmp; // partial update.
  ...
  uav[i] = Tmp; // UavStCpy
into
  Tmp = uav[i]; // UavLdCpy
  ...
  Update Tmp; // partial update.
  ...
  uav[i].updated_part = Tmp.updated_part;

This will remove the store on the part which not been updated.
The load will be removed later in DCE if it is not used.
Only support basic case now. Disabled by default.
Matrix element access, write Tmp with bitcast address is not supported.
Overlapped store on Tmp will be transfomred into overlapped store on uav[i] now.


damyanp:

PR #4412, opened 2022-04-25
Fix Writable MSAA ExecutionTest device checks

sooknarine:
Writable MSAA ExecutionTest was setting TestResults::Skipped for devices
that have the necessary support and continuing to run the test.

Follow the style of the RawGather ExecutionTest.


tex3d (approved):
LGTM!

PR #4414, opened 2022-04-25
Fix warning (x86-only build break)

hekota:


bob80905 (approved):

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -947,8 +947,8 @@ class ExecutionTest {
         D3D12_RESOURCE_STATE_COPY_DEST, nullptr, IID_PPV_ARGS(&pReadBuffer)));
 
     transferData.pData = values;
-    transferData.RowPitch = valueSizeInBytes/resDesc.Height;
-    transferData.SlicePitch = valueSizeInBytes;
+    transferData.RowPitch = (LONG_PTR)(valueSizeInBytes/resDesc.Height);
```
dmpots, line 950: Using LONG_PTR for a non-pointer value seems a bit strange. If we need a pointer-sized arithmetic type, then SIZE_T would make more sense here I think.

pow2clk (approved):
Looks good. I think a sentence in the comment to explain what is being fixed would be helpful.

dmpots (approved):
Fix looks reasonable, but I think a better type could be used here.

tools/clang/unittests/HLSL/ExecutionTest.cpp
hekota: I used LONG_PTR here because it is the type of the [D3D12_SUBRESOURCE_DATA ](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_subresource_data) structure RowPitch and SlicePitch members.

dmpots: Well that makes it a very good choice indeed here :)

PR #4416, opened 2022-04-26
[SM6.7][WritableMSAA] Front-end and test fixes

jholewinski:
This change fixes DXC to properly pass the SampleCount attribute to DXIL
metadata. A few test issues are also fixed, namely:

- Do not create upload resource for MSAA textures, since this will cause
  a sample count mismatch in the runtime and we do not need the upload
  resource anyway.
- Fix inverted logic to detect support for ATO and WritableMSAA.
- Pass D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET resource flag when
  creating MSAA resource, which is required by the runtime.


jholewinski:
It looks like this includes the same fix as #4412, so hopefully this will still merge cleanly once that PR lands.

tex3d (approved):
Looks good, but should keep the ifndef around the feature checks from #4412 when merging.

raghavthind2005 (approved):

PR #4418, opened 2022-04-26
[ATORawGather] Remove unreferenced local variable

jholewinski:
This is causing VS2022 builds to fail with warnings-as-errors enabled.


tex3d (approved):
Yes!

PR #4419, opened 2022-04-27
Fix interface breaks in IDxcUtils and IDxcLibrary from GetBlobAsWide

tex3d:
GetBlobAsWide was added as a new method to the existing interfaces,
but the existing GetBlobAsUtf16 should have simply been renamed to
GetBlobAsWide instead, since that's what GetBlobAsUtf16 actually did.
Otherwise interface compatibility is broken between builds/releases.

If we want to add a GetBlobAsUtf16 which guarantees UTF16 across platforms,
we can add that in a future interface revision.

### DIFF
--- a/include/dxc/dxcapi.h
+++ b/include/dxc/dxcapi.h
@@ -257,8 +257,8 @@ struct IDxcLibrary : public IUnknown {
     _In_ IDxcBlob *pBlob, _COM_Outptr_ IStream **ppStream) = 0;
   virtual HRESULT STDMETHODCALLTYPE GetBlobAsUtf8(
     _In_ IDxcBlob *pBlob, _COM_Outptr_ IDxcBlobEncoding **pBlobEncoding) = 0;
-  virtual HRESULT STDMETHODCALLTYPE GetBlobAsUtf16(
-    _In_ IDxcBlob *pBlob, _COM_Outptr_ IDxcBlobEncoding **pBlobEncoding) = 0;
+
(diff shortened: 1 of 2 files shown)


include/dxc/dxcapi.h
```diff
@@ -257,8 +257,8 @@ struct IDxcLibrary : public IUnknown {
     _In_ IDxcBlob *pBlob, _COM_Outptr_ IStream **ppStream) = 0;
   virtual HRESULT STDMETHODCALLTYPE GetBlobAsUtf8(
     _In_ IDxcBlob *pBlob, _COM_Outptr_ IDxcBlobEncoding **pBlobEncoding) = 0;
-  virtual HRESULT STDMETHODCALLTYPE GetBlobAsUtf16(
```
dmpots, line 260: Isn't this still a breaking change from the original because we no longer have the `GetBlobAsUtf16` method?

tex3d: It is a breaking change, but intentional.  The code gives (and has always given) you utf16 on win32, but utf32 on *nix.  This is the reason for the rename.  Code that updates to a new header and uses the old method must be changed to use the renamed method.  This is why there's a comment indicating that the method has simply been renamed.  It's an intentional breaking change.

pow2clk (approved):
Lesser evil in forcing the users of the interface to change a few characters. Let's remember to mention this in the release notes.

PR #4423, opened 2022-04-28
[dxil2spv] Translate SV_Position to BuiltIn::Position

sudonatalie:
SV_Position elements in DXIL input signature should be treated as SPIR-V
BuiltIn variables with the Position type rather than regular stage IO
variables with a Location.

### DIFF
--- a/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
+++ b/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
@@ -189,8 +189,8 @@ void Translator::createStageIOVariable(hlsl::DxilSignatureElement *elem) {
   const spirv::SpirvType *spirvType = toSpirvType(elem);
   spv::StorageClass storageClass =
       elem->IsInput() ? spv::StorageClass::Input : spv::StorageClass::Output;
-  unsigned id = elem->GetID();
-  spirv::SpirvVariable *var;
+  const unsigned id = elem->GetID();
+  spirv::SpirvVariable *var = nullptr;
   switch (elem->GetKind()) {
   case hlsl::Semantic::Kind::Position: {
     var = spvBuilder.addStageBuiltinVar(spirvType, storageClass,
--- a/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
+++ b/tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
@@ -195,14 +195,16 @@ void Translator::createStageIOVariable(hlsl::DxilSignatureElement *elem) {
   case hlsl::Semantic::Kind::Position: {
     var = spvBuilder.addStageBuiltinVar(spirvType, storageClass,
                                         spv::BuiltIn::Position, false, {});
-  } break;
+    break;
+  }
   default: {
     var = spvBuilder.addStageIOVar(spirvType, storageClass,
                                    elem->GetSemanticName(), false, {});
 
     // Use unique DXIL signature element ID as SPIR-V Location.
     spvBuilder.decorateLocation(var, id);
-  } break;
+    break;
+  }
   }
   interfaceVars.push_back(var);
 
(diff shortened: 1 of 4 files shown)


tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -175,29 +175,41 @@ void Translator::createStageIOVariables(
...
+}
+
+void Translator::createStageIOVariable(hlsl::DxilSignatureElement *elem) {
+  const spirv::SpirvType *spirvType = toSpirvType(elem);
+  spv::StorageClass storageClass =
+      elem->IsInput() ? spv::StorageClass::Input : spv::StorageClass::Output;
+  unsigned id = elem->GetID();
+  spirv::SpirvVariable *var;
+  switch (elem->GetKind()) {
+  case hlsl::Semantic::Kind::Position: {
```
kuhar, line 195: nit: why not if-else if you have only 1 non-default case?

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -175,29 +175,41 @@ void Translator::createStageIOVariables(
...
-    outputSignatureElementMap[id] = var;
+    createStageIOVariable(elem.get());
+  }
+}
+
+void Translator::createStageIOVariable(hlsl::DxilSignatureElement *elem) {
+  const spirv::SpirvType *spirvType = toSpirvType(elem);
+  spv::StorageClass storageClass =
+      elem->IsInput() ? spv::StorageClass::Input : spv::StorageClass::Output;
+  unsigned id = elem->GetID();
```
kuhar, line 192: nit: `const unsigned`?

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -175,29 +175,41 @@ void Translator::createStageIOVariables(
...
+  spirv::SpirvVariable *var;
```
kuhar, line 193: nit: to be super-save, you could immediately initialize this to `nullptr` here, in case the code gets modified and this pointer becomes accidentally uninitialized later on

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -175,29 +175,41 @@ void Translator::createStageIOVariables(
...
+    var = spvBuilder.addStageBuiltinVar(spirvType, storageClass,
+                                        spv::BuiltIn::Position, false, {});
+  } break;
+  default: {
+    var = spvBuilder.addStageIOVar(spirvType, storageClass,
+                                   elem->GetSemanticName(), false, {});
 
     // Use unique DXIL signature element ID as SPIR-V Location.
     spvBuilder.decorateLocation(var, id);
+  } break;
```
jaebaek, line 205: nit: this indentation looks weird. Did you run clang-format?

jaebaek (commented):
LGTM with a nit.

jaebaek (approved):

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -175,29 +175,41 @@ void Translator::createStageIOVariables(
...
-        elem->GetSemanticName(), false, {});
-    unsigned id = elem->GetID();
-    interfaceVars.push_back(var);
-    outputSignatureElementMap[id] = var;
+    createStageIOVariable(elem.get());
+  }
+}
+
+void Translator::createStageIOVariable(hlsl::DxilSignatureElement *elem) {
+  const spirv::SpirvType *spirvType = toSpirvType(elem);
```
kuhar, line 189: Could this be a `nullptr`? I can see that `toSpirvType` emits a diagnostic, but I'm not sure if that's fatal or not.

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
sudonatalie: Yeah, this is what clang-format does.

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
sudonatalie: Because we should handle many more HLSL semantics with SPIR-V BuiltIns in future, this is just the first one.

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
kuhar: I think it's more common to put a break inside the braces. Then it's on its own line.

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
sudonatalie: It can be nullptr and the error emitting is not fatal. Since this is a problem throughout already that hasn't really been handled well (while I was debating introducing exceptions) I'll file a follow-up bug to address it uniformly.

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
sudonatalie: Done

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
sudonatalie: Done

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
kuhar: SG

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
kuhar: @sudonatalie :^

sudonatalie: Agreed, this was one of those style things I kept for consistency with the existing SPIR-V backend code, although now that I go to find examples I see that it's not even necessarily consistent within a file (i.e. https://github.com/microsoft/DirectXShaderCompiler/blob/master/tools/clang/lib/SPIRV/SpirvEmitter.cpp), so I'll switch to the more typical way, thanks!

PR #4427, opened 2022-04-29
enable optimization passes to throw hlsl exceptions

bob80905:
Previously, during optimization passes, 
either exceptions would be thrown and diagnostic data would be lost, 
or fatal errors discovered during the passes would not prevent later passes from executing when execution should stop.

This change's purpose is to allow optimization passes the option to abort further compilation 
(stop later passes from executing), and capture relevant error data in the Diagnostic object before terminating.

This was tested with a cherry-pick here: bece3d4fafe4d7d50f5d0df59597a467d4f89f85,
and an accompanying hlsl file with invalid semantics. The thrown exception looked like this in HLSLSignatureLower.cpp:
```
void HLSignatureLower::Run() {

  // Generate error and exit if semantic type
  // is not one of the allowed types
  if (!ValidateSemanticType(Entry))
    throw hlsl::Exception(DXC_E_NOT_SUPPORTED, "Semantic type invalid");
```

Output to the console window was:

```
D:\DirectXShaderCompiler\tools\clang\test\HLSLFileCheck\hlsl\diagnostics\errors\semantics_type_checking\unsupported_types_cs_input_semantics.hlsl:41:1: error: invalid type used for 'SV_GroupThreadID' semantic.
void main(TID_TY tid : SV_DispatchThreadID,
^
error: 'Semantic type invalid'
Fatal error during optimization, aborting.


D:\hlsl.bin\Debug\bin\dxc.exe (process 187836) exited with code -2147467259.
To automatically close the console when debugging stops, enable Tools->Options->Debugging->Automatically close the console when debugging stops.
Press any key to close this window . . .
```

### DIFF
--- a/lib/HLSL/DxilCondenseResources.cpp
+++ b/lib/HLSL/DxilCondenseResources.cpp
@@ -567,7 +567,7 @@ public:
     bChanged |= RemovePhiOnResource();
 
     if (m_bLegalizationFailed)
-      throw hlsl::Exception(DXC_E_OPTIMIZATION_FAILED, "Resource legalization failed");
+      return bChanged;
 
     if (m_bIsLib) {
       if (DM.GetOP()->UseMinPrecision())
--- a/lib/HLSL/DxilCondenseResources.cpp
+++ b/lib/HLSL/DxilCondenseResources.cpp
@@ -567,7 +567,7 @@ public:
     bChanged |= RemovePhiOnResource();
 
     if (m_bLegalizationFailed)
-      return bChanged;
+      throw hlsl::Exception(DXC_E_OPTIMIZATION_FAILED, "Resource legalization failed");
 
     if (m_bIsLib) {
       if (DM.GetOP()->UseMinPrecision())
--- a/tools/clang/lib/CodeGen/BackendUtil.cpp
+++ b/tools/clang/lib/CodeGen/BackendUtil.cpp
@@ -763,18 +763,15 @@ void clang::EmitBackendOutput(DiagnosticsEngine &Diags,
                               raw_pwrite_stream *OS) {
   EmitAssemblyHelper AsmHelper(Diags, CGOpts, TOpts, LOpts, M);
 
-  // Catch any fatal errors during optimization passes here
-  // so that future passes can be skipped.
-  try {
+
+  try { // HLSL Change Starts
+    // Catch any fatal errors during optimization passes here
+    // so that future passes can be skipped.
     AsmHelper.EmitAssembly(Action, OS);
   } catch (const ::hlsl::Exception &hlslException) {
-    const char *msg = hlslException.what();
-    const std::string msgStr(msg);
-    unsigned DiagID = Diags.getCustomDiagID(
-        DiagnosticsEngine::Error,
-        "'%0'\nFatal error during optimization, aborting.");
-    Diags.Report(DiagID) << msgStr;
-  }
+    Diags.Report(Diags.getCustomDiagID(DiagnosticsEngine::Error, "%0"))
+        << StringRef(hlslException.what());
+  } // HLSL Change Ends
 
   // If an optional clang TargetInfo description string was passed in, use it to
   // verify the LLVM TargetMachine's DataLayout.
--- a/tools/clang/lib/CodeGen/BackendUtil.cpp
+++ b/tools/clang/lib/CodeGen/BackendUtil.cpp
@@ -769,7 +769,7 @@ void clang::EmitBackendOutput(DiagnosticsEngine &Diags,
     // so that future passes can be skipped.
     AsmHelper.EmitAssembly(Action, OS);
   } catch (const ::hlsl::Exception &hlslException) {
-    Diags.Report(Diags.getCustomDiagID(DiagnosticsEngine::Error, "%0"))
+    Diags.Report(Diags.getCustomDiagID(DiagnosticsEngine::Error, "%0\nFatal error during optimization, aborting"))
         << StringRef(hlslException.what());
   } // HLSL Change Ends
 
--- a/tools/clang/lib/CodeGen/BackendUtil.cpp
+++ b/tools/clang/lib/CodeGen/BackendUtil.cpp
@@ -7,6 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "dxc/Support/Global.h"
 #include "clang/CodeGen/BackendUtil.h"
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/LangOptions.h"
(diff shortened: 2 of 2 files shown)


tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -762,7 +763,18 @@ void clang::EmitBackendOutput(DiagnosticsEngine &Diags,
                               raw_pwrite_stream *OS) {
   EmitAssemblyHelper AsmHelper(Diags, CGOpts, TOpts, LOpts, M);
 
-  AsmHelper.EmitAssembly(Action, OS);
+  // Catch any fatal errors during optimization passes here
```
tex3d, line 766: When we add/modify code in existing llvm files, we surround added code blocks with `// HLSL Change Starts` and `// HLSL Change Ends`, or use `// HLSL Change` at end of single line additions.

For this change, this could be simpler by adding the `// HLSL Change` to the end of the `try {` line, surrounding the catch block, and moving the comment into the catch block explaining the reason for the change.

tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -762,7 +763,18 @@ void clang::EmitBackendOutput(DiagnosticsEngine &Diags,
...
+  // so that future passes can be skipped.
+  try {
+    AsmHelper.EmitAssembly(Action, OS);
+  } catch (const ::hlsl::Exception &hlslException) {
+    const char *msg = hlslException.what();
+    const std::string msgStr(msg);
```
tex3d, line 772: I believe a `llvm::StringRef` will suffice here instead of the `std::string`, avoiding a potential string copy.

tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -762,7 +763,18 @@ void clang::EmitBackendOutput(DiagnosticsEngine &Diags,
...
+    unsigned DiagID = Diags.getCustomDiagID(
+        DiagnosticsEngine::Error,
+        "'%0'\nFatal error during optimization, aborting.");
```
tex3d, line 775: I don't think surrounding the msgStr in quotes `'` is necessary.
Seeing this in the example, I'm thinking that we probably don't really need the extra text here either.
Since you can create single-use things on the fly, I don't see why we couldn't just reduce the code in the block to:
```
Diags.Report(Diags.getCustomDiagID(DiagnosticsEngine::Error, "%0")) << StringRef(hlslException.what());
```

tex3d:
For a test case that doesn't require applying other large changes, the `DxilLowerCreateHandleForLib` pass will emit errors for overlapping resource binding requirements during automatic binding.  So, with a change to abort from that pass on error, a modified version of the following could suffice as a test case:
```
// RUN: %dxc /T vs_6_0 /E main %s | FileCheck %s

// CHECK: error: more than one unbounded resource (Buf1 and Buf2) in space 0

ByteAddressBuffer Buf1[];
ByteAddressBuffer Buf2[];

uint main() : OUT {
  return Buf1[27].Load(27) + Buf2[15].Load(15);
}
```

tex3d:
Actually, the ideal place to place a throw to abort in that pass would be to replace `return bChanged` here with the throw:
```
    if (m_bLegalizationFailed)
      return bChanged;
```
But this won't catch all of the error types that this pass can emit, so a test case like this instead could be used to trigger this error:
```
// RUN: %dxc /T vs_6_0 /E main %s | FileCheck %s

// CHECK: error: local resource not guaranteed to map to unique global resource.

ByteAddressBuffer Buf1;
ByteAddressBuffer Buf2;
bool b;

uint main() : OUT {
  ByteAddressBuffer Buf = Buf1;
  if (b)
    Buf = Buf2;
  return Buf.Load(27);
}
```

bob80905:
After the latest commit, when running with these arguments:
`/E main -T vs_6_0 -E main C:\Users\jbatista\Desktop\legalization.hlsl`
The console output is:
```
C:\Users\jbatista\Desktop\legalization.hlsl:11:7: error: local resource not guaranteed to map to unique global resource.
  if (b)
      ^
error: Resource legalization failed
Fatal error during optimization, aborting


D:\hlsl.bin\Debug\bin\dxc.exe (process 201432) exited with code -2147467259.
To automatically close the console when debugging stops, enable Tools->Options->Debugging->Automatically close the console when debugging stops.
Press any key to close this window . . .
```

lib/HLSL/DxilCondenseResources.cpp
```diff
@@ -567,7 +567,7 @@ class DxilLowerCreateHandleForLib : public ModulePass {
     bChanged |= RemovePhiOnResource();
 
     if (m_bLegalizationFailed)
-      return bChanged;
+      throw hlsl::Exception(DXC_E_OPTIMIZATION_FAILED, "Resource legalization failed");
```
tex3d, line 570: I understand that this was used to test the change, but let's not change the behavior of this pass just yet.  If we do want to abort compilation on an error in this pass, there is more work to do to capture more failure cases that `m_bLegalizationFailed` doesn't capture.  We can save that for a different PR.

tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -7,6 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "dxc/Support/Global.h"
```
tex3d, line 10: I believe minimal include is actually this header:
```
#include "dxc/Support/exception.h"  // HLSL Change
```

tex3d, line 10: I believe minimal include is actually this header:
```
#include "dxc/Support/exception.h"  // HLSL Change
```

tex3d: Also, this should probably be included at the bottom of the list with the other dxc includes.

tex3d: Also, this should probably be included at the bottom of the list with the other dxc includes.

tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -762,7 +763,15 @@ void clang::EmitBackendOutput(DiagnosticsEngine &Diags,
...
   EmitAssemblyHelper AsmHelper(Diags, CGOpts, TOpts, LOpts, M);
 
-  AsmHelper.EmitAssembly(Action, OS);
+
+  try { // HLSL Change Starts
+    // Catch any fatal errors during optimization passes here
+    // so that future passes can be skipped.
+    AsmHelper.EmitAssembly(Action, OS);
+  } catch (const ::hlsl::Exception &hlslException) {
+    Diags.Report(Diags.getCustomDiagID(DiagnosticsEngine::Error, "%0\nFatal error during optimization, aborting"))
```
tex3d, line 772: I think we can get rid of the extra error text added here.  We already print an error with the exception-supplied reason, so I think the extra text is unnecessary and out of place now.

lib/HLSL/DxilCondenseResources.cpp
tex3d: This change is also causing the test failure, so let's back it out for now.

tools/clang/lib/CodeGen/BackendUtil.cpp
bob80905: There are several build errors when just including exception, that are resolved by including global:
Error	C3646	'hr': unknown override specifier	clangCodeGen	D:\DirectXShaderCompiler\include\dxc\Support\exception.h	28	
Error	C4430	missing type specifier - int assumed. Note: C++ does not support default-int	clangCodeGen	D:\DirectXShaderCompiler\include\dxc\Support\exception.h	28	
Error	C2061	syntax error: identifier 'HRESULT'	clangCodeGen	D:\DirectXShaderCompiler\include\dxc\Support\exception.h	31

tex3d (approved):
LGTM!

PR #4428, opened 2022-04-30
add non-uniform dynamic indexing option to dynamic resources uniform indexing test

bob80905:
Previously, this test failed to test dynamic resources with non-uniform indices.
Now, the test runs with multiple threads, and each thread may use 1 of 2 different indices
to access a resource, depending on pixel location or thread id.
#define NON_UNIFORM 
is used to force the test to enter the Non-uniform path, which (maybe)? is a stricter test than the uniform variant.
Although, the goal is to allow the test suite to define whether or not to run this test in Uniform or NON_uniform mode depending on a define like /DNON_UNIFORM
The default mode is Uniform. 
All resources were duplicated to allow each thread the option to select one or the other duplicate. 
At this point, the Non-Fallback path is passing on NVD hardware. More testing should be done before this gets checked in.

### DIFF
--- a/tools/clang/test/HLSL/ShaderOpArith.xml
+++ b/tools/clang/test/HLSL/ShaderOpArith.xml
@@ -3494,7 +3494,7 @@ void MSMain(uint GID : SV_GroupIndex,
     </Shader>
   </ShaderOp>
 
-  <ShaderOp Name="DynamicResourcesUniformAndNonUniformIndexing" CS="CS66" VS="VS66" PS="PS66">
+  <ShaderOp Name="DynamicResourcesDynamicIndexing" CS="CS66" VS="VS66" PS="PS66">
 
     <Resource Name="ConstantBuffer" Dimension="BUFFER" Width="256" InitialResourceState="COPY_DEST" Init="FromBytes" ReadBack="true">
       { 0u, 1u, 2u, 3u, 4u, 5u, 6u, 7u, 8u, 9u, 10u, 11u}
@@ -3505,40 +3505,40 @@ void MSMain(uint GID : SV_GroupIndex,
     </Resource>
 
     <Resource Name="T1"            Dimension="BUFFER"     Width="32"            InitialResourceState="COPY_DEST" Init="FromBytes" >
-      10.0
+      11.0
     </Resource>
 
     <Resource Name="T2"            Dimension="BUFFER"     Width="32"            InitialResourceState="COPY_DEST" Init="FromBytes" >
-      11.0
+      12.0
     </Resource>
 
     <Resource Name="T3"            Dimension="BUFFER"     Width="32"            InitialResourceState="COPY_DEST" Init="FromBytes" >
-      11.0
+      13.0
     </Resource>
 
     <Resource Name="T4"            Dimension="Texture2D"  Width="4" Height="4"  InitialResourceState="COPY_DEST" Init="FromBytes" Format="R32_FLOAT">
-      12.0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
+      14.0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
     </Resource>
     <Resource Name="T5"            Dimension="Texture2D"  Width="4" Height="4"  InitialResourceState="COPY_DEST" Init="FromBytes" Format="R32_FLOAT">
-      12.0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
+      15.0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
     </Resource>
     
     <Resource Name="U0"            Dimension="BUFFER"     Width="32"            InitialResourceState="COPY_DEST" Init="FromBytes" Flags="ALLOW_UNORDERED_ACCESS" TransitionTo="UNORDERED_ACCESS">
-      23.0
+      20.0
     </Resource>
     <Resource Name="U1"            Dimension="BUFFER"     Width="32"            InitialResourceState="COPY_DEST" Init="FromBytes" Flags="ALLOW_UNORDERED_ACCESS" TransitionTo="UNORDERED_ACCESS">
-      23.0
+      21.0
     </Resource>
     
     <Resource Name="U2"            Dimension="BUFFER"     Width="32"            InitialResourceState="COPY_DEST" Init="FromBytes" Flags="ALLOW_UNORDERED_ACCESS" TransitionTo="UNORDERED_ACCESS">
-      24.0
+      22.0
     </Resource>
     <Resource Name="U3"            Dimension="BUFFER"     Width="32"            InitialResourceState="COPY_DEST" Init="FromBytes" Flags="ALLOW_UNORDERED_ACCESS" TransitionTo="UNORDERED_ACCESS">
-      24.0
+      23.0
     </Resource>
     
     <Resource Name="U4"            Dimension="TEXTURE1D"  Width="4"             InitialResourceState="COPY_DEST" Init="FromBytes" Flags="ALLOW_UNORDERED_ACCESS" Format="R32_FLOAT" TransitionTo="UNORDERED_ACCESS">
-      25.0 0 0 0
+      24.0 0 0 0
     </Resource>
     <Resource Name="U5"            Dimension="TEXTURE1D"  Width="4"             InitialResourceState="COPY_DEST" Init="FromBytes" Flags="ALLOW_UNORDERED_ACCESS" Format="R32_FLOAT" TransitionTo="UNORDERED_ACCESS">
       25.0 0 0 0
@@ -3643,9 +3643,9 @@ void MSMain(uint GID : SV_GroupIndex,
 
     ConstantBuffer<Constants> CB               : register(b12);
     // Result buffers
-    RWBuffer<float> g_result   : register(u13);
-    RWBuffer<float> g_resultVS : register(u14);
-    RWBuffer<float> g_resultPS : register(u15);
+    RWStructuredBuffer <float> g_result   : register(u13);
+    RWStructuredBuffer <float> g_resultVS : register(u14);
+    RWStructuredBuffer <float> g_resultPS : register(u15);
     
     #else // NO FALLBACK
 
@@ -3661,8 +3661,10 @@ void MSMain(uint GID : SV_GroupIndex,
     
     #endif // FALLBACK
     
-    #ifndef FALLBACK
+    
     void TestResources(RWStructuredBuffer<float> result, uint ix) {
+        
+    #ifndef FALLBACK
         ByteAddressBuffer rawBuf = ResourceDescriptorHeap[INDEXER(CB.idx0,ix)];
         StructuredBuffer<float> structBuf = ResourceDescriptorHeap[INDEXER(CB.idx2,ix)];
         Texture2D<float> tex = ResourceDescriptorHeap[INDEXER(CB.idx4,ix)];
@@ -3673,8 +3675,8 @@ void MSMain(uint GID : SV_GroupIndex,
         static SamplerState g_samp = SamplerDescriptorHeap[INDEXER(CB.idx0, ix)];
         static SamplerComparisonState g_sampCmp  = SamplerDescriptorHeap[INDEXER(CB.idx2, ix)];
     #endif
+    
     #ifdef FALLBACK
-    void FallBackTestResources(RWBuffer<float> result, uint ix) {
         ByteAddressBuffer rawBuf = g_fallback_rawBuf[INDEXER(0,ix)];
         StructuredBuffer<float> structBuf = g_fallback_structBuf[INDEXER(0,ix)];
         Texture2D<float> tex = g_fallback_tex[INDEXER(0,ix)];
@@ -3699,11 +3701,7 @@ void MSMain(uint GID : SV_GroupIndex,
     ROOT_SIG
     [NumThreads(2, 2, 1)]
     void main(uint ix : SV_GroupIndex) {
-        #ifndef FALLBACK
         TestResources(g_result, ix);
-        #else
-        FallBackTestResources(g_result, ix);
-        #endif
     }
     
     struct PSInput {
@@ -3713,22 +3711,14 @@ void MSMain(uint GID : SV_GroupIndex,
     ROOT_SIG
     float4 PSMain(PSInput input) : SV_TARGET {
         int ix = WaveGetLaneIndex();
-        #ifndef FALLBACK
-        TestResources(g_resultPS, ix);
-        #else
-        FallBackTestResources(g_resultPS, ix);
-        #endif       
+        TestResources(g_resultPS, ix);      
         // This output doesn't actually matter
         return input.position;
     }
     
     ROOT_SIG
     PSInput VSMain(float3 pos : POSITION, uint ix : SV_VertexID) {
-        #ifndef FALLBACK
         TestResources(g_resultVS, ix);
-        #else
-        FallBackTestResources(g_resultVS, ix);
-        #endif
         PSInput r;
         r.position = float4(pos, 1); 
         return r;
--- a/tools/clang/test/HLSL/ShaderOpArith.xml
+++ b/tools/clang/test/HLSL/ShaderOpArith.xml
@@ -3631,15 +3631,15 @@ void MSMain(uint GID : SV_GroupIndex,
     #define ROOT_SIG [RootSignature("RootFlags(ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT), DescriptorTable( SRV(t0, numDescriptors=6), UAV(u6, numDescriptors=6), CBV(b12), UAV(u13,numDescriptors=3) ), \
     DescriptorTable(Sampler(s0, numDescriptors=2), Sampler(s2, numDescriptors=2))")]
     
-    ByteAddressBuffer fallback_g_rawBuf[2]              : register(t0);
-    StructuredBuffer<float> fallback_g_structBuf[2]     : register(t2);
-    Texture2D<float> fallback_g_tex[2]                  : register(t4);
-    RWByteAddressBuffer fallback_g_rwRawBuf[2]          : register(u6);
-    RWStructuredBuffer<float> fallback_g_rwStructBuf[2] : register(u8);
-    RWTexture1D<float> fallback_g_rwTex[2]              : register(u10);
+    ByteAddressBuffer g_fallback_rawBuf[2]              : register(t0);
+    StructuredBuffer<float> g_fallback_structBuf[2]     : register(t2);
+    Texture2D<float> g_fallback_tex[2]                  : register(t4);
+    RWByteAddressBuffer g_fallback_rwRawBuf[2]          : register(u6);
+    RWStructuredBuffer<float> g_fallback_rwStructBuf[2] : register(u8);
+    RWTexture1D<float> g_fallback_rwTex[2]              : register(u10);
 
-    SamplerState fallback_g_samp[2]              : register(s0);
-    SamplerComparisonState fallback_g_sampCmp[2] : register(s2);
+    SamplerState g_fallback_samp[2]              : register(s0);
+    SamplerComparisonState g_fallback_sampCmp[2] : register(s2);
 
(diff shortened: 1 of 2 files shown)


tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3616,29 +3642,39 @@ void MSMain(uint GID : SV_GroupIndex,
...
     static SamplerComparisonState g_sampCmp = SamplerDescriptorHeap[CB.idx1];
 
     #endif // FALLBACK
 
     ROOT_SIG
-    [NumThreads(1, 1, 1)]
+    [NumThreads(8, 8, 1)]
     void main(uint ix : SV_GroupIndex) {
+    
+        ByteAddressBuffer g_rawBuf = ResourceDescriptorHeap[INDEXER(CB.idx0) + OFFSET(ix)];
```
tex3d, line 3671: Non-uniform resource indexes must use the `NonUniformResourceIndex()` intrinsic around the index inside the indexing operation, like:
```
ByteAddressBuffer g_rawBuf = ResourceDescriptorHeap[NonUniformResourceIndex(INDEXER(CB.idx0) + OFFSET(ix))];
```

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3655,6 +3691,17 @@ void MSMain(uint GID : SV_GroupIndex,
...
+        int ix = x+y;
+        
+        ByteAddressBuffer g_rawBuf = ResourceDescriptorHeap[INDEXER(CB.idx0) + OFFSET(ix)];
+        StructuredBuffer<float> g_structBuf = ResourceDescriptorHeap[INDEXER(CB.idx2) + OFFSET(ix)];
+        Texture2D<float> g_tex = ResourceDescriptorHeap[INDEXER(CB.idx4) + OFFSET(ix)];
+        RWByteAddressBuffer g_rwRawBuf = ResourceDescriptorHeap[INDEXER(CB.idx6) + OFFSET(ix)];
+        RWStructuredBuffer<float> g_rwStructBuf = ResourceDescriptorHeap[INDEXER(CB.idx8) + OFFSET(ix)];
+        RWTexture1D<float> g_rwTex = ResourceDescriptorHeap[INDEXER(CB.idx10) + OFFSET(ix)];
         
         g_resultPS[0] = g_rawBuf.Load<float>(0);
```
tex3d, line 3706: If the resource is non-uniform, the value loaded here will be non-uniform, so you'll need to change the location written based on your `OFFSET(ix)` value so you can capture the results independently.  You probably want to add `(OFFSET(ix)*8)` since you have 8 result values per index read.  Then you'll need to expand and modify the corresponding verification code in the test.  Since you'll be adding more expected result values, at this point, might I suggest an expected result array that you loop through along with the output values, instead of a bunch of independent `VERIFY_ARE_EQUAL` statements?

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3655,6 +3691,17 @@ void MSMain(uint GID : SV_GroupIndex,
 
     ROOT_SIG
     float4 PSMain(PSInput input) : SV_TARGET {
+    
+        int x = (int)input.position.x;
+        int y = (int)input.position.y;
+        int ix = x+y;
```
tex3d, line 3697: Double check that this doesn't end up being zero for all/most of your pixels based on the incoming range for position x and y.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3616,29 +3642,39 @@ void MSMain(uint GID : SV_GroupIndex,
...
+    #else
+    #define INDEXER(ix) NonUniformResourceIndex(ix)
+    #define OFFSET(ix) ix % 2
+    #endif
 
     // Result buffers
-    static RWStructuredBuffer<float> g_result = ResourceDescriptorHeap[7];
-    static RWStructuredBuffer<float> g_resultVS = ResourceDescriptorHeap[8];
-    static RWStructuredBuffer<float> g_resultPS = ResourceDescriptorHeap[9];
+    static RWStructuredBuffer<float> g_result = ResourceDescriptorHeap[13];
```
tex3d, line 3658: Recommend declaring a `static const int` constant for first index used for constant buffer above, then adding 1/2/3 here instead.  Should be easier to follow/maintain.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3497,29 +3497,49 @@ void MSMain(uint GID : SV_GroupIndex,
   <ShaderOp Name="DynamicResourcesUniformIndexing" CS="CS66" VS="VS66" PS="PS66">
```
tex3d, line 3497: Shouldn't this be renamed now?

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3616,29 +3642,39 @@ void MSMain(uint GID : SV_GroupIndex,
...
 
-    static ConstantBuffer<Constants> CB = ResourceDescriptorHeap[6];
-
-    static ByteAddressBuffer g_rawBuf = ResourceDescriptorHeap[CB.idx0];
-    static StructuredBuffer<float> g_structBuf = ResourceDescriptorHeap[CB.idx1];
-    static Texture2D<float> g_tex = ResourceDescriptorHeap[CB.idx2];
-    static RWByteAddressBuffer g_rwRawBuf = ResourceDescriptorHeap[CB.idx3];
-    static RWStructuredBuffer<float> g_rwStructBuf = ResourceDescriptorHeap[CB.idx4];
-    static RWTexture1D<float> g_rwTex = ResourceDescriptorHeap[CB.idx5];
+    #define NON_UNIFORM
```
tex3d, line 3648: Option should be driven by a compilation argument I suspect, then code in test added to try both cases.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3616,29 +3642,39 @@ void MSMain(uint GID : SV_GroupIndex,
...
+    #ifndef NON_UNIFORM
+    #define INDEXER(ix) ix
+    #define OFFSET(ix) 0
+    #else
+    #define INDEXER(ix) NonUniformResourceIndex(ix)
+    #define OFFSET(ix) ix % 2
```
tex3d, line 3654: `NonUniformResourceIndex` needs to surround the entire indexing expression, not just the first part coming from the constant buffer.  So, a change to integrate the offset into one indexing macro would be better, like:
```
#define INDEXER(base, offset) NonUniformResourceIndex(base + (offset % 2))
```

tex3d, line 3654: `NonUniformResourceIndex` needs to surround the entire indexing expression, not just the first part coming from the constant buffer.  So, a change to integrate the offset into one indexing macro would be better, like:
```
#define INDEXER(base, offset) NonUniformResourceIndex(base + (offset % 2))
```

bob80905:
Need to add non-uniform indexing for samplers, add a definition option in the test, and update the fallback path.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9756,130 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
+        pShaderOp->RootValues.clear();
+      }
 
-      MappedData resultData;
-      test->Test->GetReadBackData("g_result", &resultData);
-      const float *resultFloats = (float *)resultData.data();
+      // Update shader target in xml.
+      for (st::ShaderOpShader &S : pShaderOp->Shaders){
+        if (sm==D3D_SHADER_MODEL_6_0){
+          S.Arguments = compilerOptions;
```
tex3d, line 9826: You can't just safely assign compilerOptions to S.Arguments here like this.  Make sure you do this instead:
```
  S.Arguments = pShaderOp->GetString(compilerOptions);
```
This will assign a string that pShaderOp owns rather than having it refer to a string on the stack here.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9756,130 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
+          vector<char> s;
+          for (unsigned int j = 0; j < sizeof(S.Target); j++){
+            s.push_back(S.Target[j]);
+          }
+          // change Target from ps_6_6 to ps_6_0, for example
+          s[sizeof(S.Target)-1] = '0';
+          strcpy_s(s.data(), sizeof(S.Target), S.Target);
```
tex3d, line 9833: It's not good to modify the strings owned by ShaderOp in-place.  These are not expected to be mutable.  Plus sizeof() isn't going to give you the string length, it gives you the size of the type (LPCSTR), which for 64-bit architectures is 8 bytes.  You're thinking of `strlen()`, not `sizeof()`.

You can use this instead:
```
std::string Target(S.Target);
Target[Target.length() - 1] = '0';
S.Target = pShaderOp->GetString(Target.c_str());
```

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9756,130 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
+      {
+        pShaderOp->RootValues.clear();
+      }
 
-      MappedData resultData;
-      test->Test->GetReadBackData("g_result", &resultData);
-      const float *resultFloats = (float *)resultData.data();
+      // Update shader target in xml.
+      for (st::ShaderOpShader &S : pShaderOp->Shaders){
+        if (sm==D3D_SHADER_MODEL_6_0){
```
tex3d, line 9825: Logic for whether to assign arguments appears to be based only on shader model, yet you want to set NON_UNIFORM for non-fallback case as well, right?

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9756,130 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
-      WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
-      return;
-    }
+      // Add compile options
+      char compilerOptions[256];
+      if (non_uniform_bit && sm==D3D_SHADER_MODEL_6_0) {
+        sprintf_s(compilerOptions, sizeof(compilerOptions),
+                                "-D FALLBACK=%d, -D NON_UNIFORM=%d", 1, 1);
+      }
+      else if (non_uniform_bit || sm==D3D_SHADER_MODEL_6_0){
```
tex3d, line 9805: All this logic seems a bit much.  Simple string concatenation would do:
```
std::string compilerOptions;
if (sm==D3D_SHADER_MODEL_6_0)
  compilerOptions += " -DFALLBACK=1";
if (non_uniform_bit)
  compilerOptions += " -DNON_UNIFORM=1";
```

tex3d (changes requested):

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -3576,77 +3604,106 @@ void MSMain(uint GID : SV_GroupIndex,
...
-    RWBuffer<float> g_result : register(u7);
-    RWBuffer<float> g_resultVS : register(u8);
-    RWBuffer<float> g_resultPS : register(u9);
-
-    SamplerState g_samp : register(s0);
-    SamplerComparisonState g_sampCmp : register(s1);
-
+    RWBuffer<float> g_result   : register(u13);
+    RWBuffer<float> g_resultVS : register(u14);
+    RWBuffer<float> g_resultPS : register(u15);
```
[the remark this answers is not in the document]
bob80905: I don't think I can merge the 2 functions into one because Non-Fallback has to access resources via ResourceDescriptorHeap, and fallback needs to access resources using the explicitly bound register array. If Fallback is given ResourceDescriptorHeap, it says "use of undeclared identifier".

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9749,141 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
   st::ParseShaderOpSetFromStream(pStream, ShaderOpSet.get());
   st::ShaderOp *pShaderOp =
-      ShaderOpSet->GetShaderOp("DynamicResourcesUniformIndexing");
+      ShaderOpSet->GetShaderOp("DynamicResourcesDynamicIndexing");
 
   bool Skipped = true;
 
   //D3D_SHADER_MODEL TestShaderModels[] = {D3D_SHADER_MODEL_6_0}; // FALLBACK
-  D3D_SHADER_MODEL TestShaderModels[] = {D3D_SHADER_MODEL_6_6};
+  D3D_SHADER_MODEL TestShaderModels[] = {D3D_SHADER_MODEL_6_0, D3D_SHADER_MODEL_6_6};
```
tex3d, line 9757: I am a bit torn with what to do here.  I don't think we should be checking in the test in a state that will silently pass without SM 6.6 support, without some special selection for the fallback mode.  That selection could be made with a define, or maybe even with a test parameter.  We should standardize this mechanism across tests somehow as well.

Perhaps the cleanest approach is to add a method like `bool ExecutionTest::IsFallbackPathEnabled()` that does something like:
```
// Enable fallback paths with: /p:"EnableFallback=1"
UINT EnableFalbackValue = 0;
WEX::TestExecution::RuntimeParameters::TryGetValue(L"EnableFallback", EnableFallbackValue);
return EnableFalbackValue != 0;
```
In a future change, we can update other tests that silently support fallback paths or use a define to use this instead.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9749,141 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
-      VERIFY_ARE_EQUAL(resultPSFloats[3], 23.0F);
-      VERIFY_ARE_EQUAL(resultPSFloats[4], 24.0F);
-      VERIFY_ARE_EQUAL(resultPSFloats[5], 25.0F);
-      VERIFY_ARE_EQUAL(resultPSFloats[6], 30.0F);
-      VERIFY_ARE_EQUAL(resultPSFloats[7], 31.0F);
+        MappedData resultData;
+        test->Test->GetReadBackData("g_result", &resultData);
+        const float *resultCSFloats = (float *)resultData.data();
+
+        TestReadbackDynamicResourcesDynamicIndexing(non_uniform_bit, resultCSFloats, expectedResultsArray, expectedResultsSize);
```
tex3d, line 9852: Could simply index `expectedResultsArray` here (or set a ref at the beginning of the loop), making it unnecessary to pass `non_uniform_bit` to the function, and for the function to even have to know anything about this.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9749,141 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
+        // VS
+        TestReadbackDynamicResourcesDynamicIndexing(non_uniform_bit, resultVSFloats, expectedResultsArray, expectedResultsSize);
+
+        // PS
+        TestReadbackDynamicResourcesDynamicIndexing(non_uniform_bit, resultPSFloats, expectedResultsArray, expectedResultsSize);
+      }
+      Skipped = false;
+    }
+    if (Skipped) {
+      WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
```
tex3d, line 9883: Shouldn't this be outside both loops, and at the end of the test?

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9749,141 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
+    22.0, 23.0,
+    24.0, 25.0, 
+    30.0, 31.0, 
+    32.0, 33.0};
+  float *expectedResultsArray[2];
+  expectedResultsArray[0] = expectedResultsUniform;
+  expectedResultsArray[1] = expectedResultsNonUniform;
+    
+  for (unsigned int non_uniform_bit = 0; non_uniform_bit < 2; non_uniform_bit++) {
+    for (unsigned i = 0; i < _countof(TestShaderModels); i++) {
```
tex3d, line 9784: I think the nested order of these loops should be reversed, then device and feature detection moved outside the uniformity loop.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9740,99 +9749,141 @@ TEST_F(ExecutionTest, DynamicResourcesUniformIndexingTest) {
...
-    {
-      pShaderOp->CS = pShaderOp->GetString("CS66");
-      std::shared_ptr<ShaderOpTestResult> test = RunShaderOpTestAfterParse(
-          pDevice, m_support, "DynamicResourcesUniformIndexing", nullptr,
-          ShaderOpSet);
+      std::string compilerOptions = "";
+      if (sm==D3D_SHADER_MODEL_6_0)
+        compilerOptions += "-D FALLBACK=1";
+      if (non_uniform_bit && sm==D3D_SHADER_MODEL_6_0)
+        compilerOptions += ", -D NON_UNIFORM=1";
```
tex3d, line 9811: You don't want a `,` in the string between options here, only a space.  If you simply add a leading space for the `" -D NON_UNIFORM=1"` string (as I had written in my comment), you don't need this branch at all.  Was there a problem running with the options from my previous comment?  Was the leading space a problem?  If so, just try trailing space instead.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9776,51 +9783,60 @@ TEST_F(ExecutionTest, DynamicResourcesDynamicIndexingTest) {
...
+          L"Device does not support Resource Binding Tier 3");
+      WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+      return;
+    }
 
-      // Add compile options
+    for (unsigned int non_uniform_bit = 0; non_uniform_bit < 2; non_uniform_bit++) {
+      float *expectedResults = non_uniform_bit ? expectedResultsNonUniform : expectedResultsUniform;
+
+      LogCommentFmt(L"\r\nnon uniform bit is %1u",
```
tex3d, line 9816: Why `\r\n` in front of the message?  Plus this message isn't very consistent with conventions here.  What about:
```
LogCommentFmt(L"Testing %s Resource Indexing.", non_uniform_bit ? L"NonUniform" : L"Uniform");
```

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9750,11 +9756,12 @@ TEST_F(ExecutionTest, DynamicResourcesDynamicIndexingTest) {
   st::ParseShaderOpSetFromStream(pStream, ShaderOpSet.get());
   st::ShaderOp *pShaderOp =
       ShaderOpSet->GetShaderOp("DynamicResourcesDynamicIndexing");
+  vector<st::ShaderOpRootValue> oldRootValues = pShaderOp->RootValues;
```
tex3d, line 9759: Good catch!  I think the name `fallbackRootValues` would be a bit more clear.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9776,51 +9783,60 @@ TEST_F(ExecutionTest, DynamicResourcesDynamicIndexingTest) {
...
       }
+      else
+      {
+         pShaderOp->RootValues = oldRootValues;
+      }
 
       // Update shader target in xml.
       for (st::ShaderOpShader &S : pShaderOp->Shaders){
+        S.Arguments = NULL;
         if (compilerOptions != ""){
```
tex3d, line 9840: `!compilerOptions.empty()` should also work from default initialized `std::string`.

tex3d (approved):
Only minor comments, take or leave them.

PR #4426, opened 2022-04-28
[dxil2spv] Translate DXIL constants to SPIR-V

sudonatalie:
Add intial handling of DXIL float constants, translating to SPIR-V float
constant instructions. Emit errors for other constant types not yet
handled.

PR #4431, opened 2022-05-02
[dxil2spv] Add FileCheck testing

sudonatalie:
Add FileCheck testing to dxil2spv, and a first static vertex shader test case to test #4426. Also includes documentation for how to
add additional tests, and replaces the existing WholeFileTestFixture with regular FileCheck.


kuhar (commented):
Why can't we use LIT and FileCheck directly?

sudonatalie:
All of the existing SPIR-V codegen tests are currently with `effcee`, so I was just opting for consistency with that.

tools/clang/unittests/Dxil2Spv/LitTest.cpp
```diff
@@ -7,21 +7,54 @@
...
+  TEST_METHOD_PROPERTY(L"Priority", L"0")
+  END_TEST_CLASS()
 
-TEST_F(WholeFileTest, PassThruVertexShader) {
-  runWholeFileTest("passthru-vs.ll");
-}
+  TEST_CLASS_SETUP(InitSupport);
+
+  dxc::DxcDllSupport m_dllSupport;
+  VersionSupportInfo m_ver;
```
jaebaek, line 30: nit: This is confusing. `m_version` is more readable.

tools/clang/unittests/Dxil2Spv/LitTest.cpp
```diff
@@ -7,21 +7,54 @@
...
+
+  void runFileTest(std::string name) {
+    std::string fullPath =
+        clang::dxil2spv::utils::getAbsPathOfInputDataFile(name);
+    FileRunTestResult t =
```
jaebaek, line 35: nit: `t` --> `result`

jaebaek (approved):

CMakeLists.txt
```diff
@@ -132,6 +132,9 @@ endif()
 
 # Building dxil2spv off by default while it's in early development.
 option(ENABLE_DXIL2SPV "Enables DXIL to SPIR-V translation." OFF)
+if (${ENABLE_DXIL2SPV})
+  add_definitions(-DENABLE_DXIL2SPV)
```
kuhar, line 136: Would it be possible to make this a target definition instead of a global one?

tools/clang/test/Dxil2Spv/passthru-cs.ll
```diff
@@ -82,94 +82,93 @@ attributes #2 = { nounwind }
...
-;          %54 = OpAccessChain %_ptr_Function_int %24 %uint_4
-;          %55 = OpBitcast %int %53
-;                OpStore %54 %55
-;          %56 = OpAccessChain %_ptr_Function_int %24 %uint_0
-;          %57 = OpAccessChain %_ptr_Uniform_uint %4 %uint_0 %28
-;          %58 = OpBitcast %uint %56
-;                OpStore %57 %58
-;                OpReturn
-;                OpFunctionEnd
+; CHECK:      ; SPIR-V
```
kuhar, line 85: Do you plan to clean up tests in future PRs? I would not expect them to try to match every single output line

tools/clang/test/Dxil2Spv/passthru-cs.ll
```diff
@@ -1,4 +1,4 @@
-; RUN: %dxil2spv
+; RUN: %dxil2spv | %FileCheck %s
```
kuhar, line 1: If the filecheck if new enough, you may want to use `--match-full-lines`: https://llvm.org/docs/CommandGuide/FileCheck.html#:~:text=By%20default%2C%20FileCheck%20allows%20matches,affected%20by%20this%20option!).
If the copy in this repo doesn't have this mode, it's also fine as-is.

tools/clang/unittests/HLSLTestLib/CMakeLists.txt
```diff
@@ -19,10 +22,24 @@ add_clang_library(HLSLTestLib
   DxcTestUtils.cpp
   FileCheckerTest.cpp
   FileCheckForTest.cpp
+  HLSLTestOptions.cpp
 )
 include_directories(${LLVM_MAIN_SRC_DIR}/utils/unittest/googletest/include)
 include_directories(${LLVM_MAIN_SRC_DIR}/utils/unittest/googlemock/include)
 
+if(ENABLE_DXIL2SPV)
+include_directories(
```
kuhar, line 31: Can we use `target_include_directories`?

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
```diff
@@ -1091,6 +1103,59 @@ FileRunCommandResult FileRunCommandPart::RunDxilVer(dxc::DxcDllSupport& DllSuppo
...
+  bool success = true;
+
+  std::string stdoutString;
+  std::string stderrString;
+  llvm::raw_string_ostream stdoutStream(stdoutString);
+  llvm::raw_string_ostream stderrStream(stderrString);
+
+  try {
+    // Configure filesystem.
+    llvm::sys::fs::MSFileSystem *msf;
```
kuhar, line 1121: nit: can you initialize this pointer to null to be super-safe across all branches?

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
```diff
@@ -1091,6 +1103,59 @@ FileRunCommandResult FileRunCommandPart::RunDxilVer(dxc::DxcDllSupport& DllSuppo
...
+  } catch (...) {
+    success = false;
+  }
+
+  stdoutStream.flush();
+  stderrStream.flush();
+
+  FileRunCommandResult result = {};
+  result.StdOut = stdoutString;
+  result.StdErr = stderrString;
```
kuhar, line 1152: nit: You can do `stdoutStream.str()` and it will `flush` automatically for you

tools/clang/unittests/Dxil2Spv/LitTest.cpp
sudonatalie: Done

tools/clang/test/Dxil2Spv/passthru-cs.ll
sudonatalie: I hadn't. The plan was that we have "WholeFileTests" for just the 3 passthrough shaders, and their small benefit was that they do provide a reference for the entire output (which was probably more helpful for the initial development up until now). We have the same on the SPIR-V backend and I haven't yet seen a change that required modifications to the handful of them, so I don't expect the churn to be too bad but if it is, then changing them makes sense. If they seem like more harm than good though let me know and I'll track a follow up.

tools/clang/test/Dxil2Spv/passthru-cs.ll
sudonatalie: Fails with "Invalid argument", unfortunately

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
sudonatalie: Done

kuhar (approved):
LGTM, thanks for the changes!

tools/clang/test/Dxil2Spv/passthru-cs.ll
kuhar: I think this is good enough for what we need now, but some things seem too specific. For example, I don't think we care about the exact ID numbers that are printed, e.g., `%19 = OpTypeFunction %void`. If this becomes a problem in the future and we need to update these files often, we can revisit this at that time.

PR #4432, opened 2022-05-03
Align RawGather Exe Test with dependencies

pow2clk:
This introduces the correct mechanism for creating resources that might
be cast to a different format of the same bit size when enhanced
variables are found to be available. It also limits the formats to those
that are natively castable unless the relaxed format casting feature is
available.

To allow this, the resource creation utility function now has the
ability to provide these potential casting targets if supported, new
queries for these features are added, and some of the formats were shuffled
to make it easier to isolate the relevant ones.

Also, retreive advanced tex ops caps bits from from proper options struct
The previous implementation drew from options13, but the settings ended
up on options14

### DIFF
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -908,12 +908,19 @@ public:
     }
   }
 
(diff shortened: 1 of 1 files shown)


tex3d (approved):
LGTM!

PR #4440, opened 2022-05-06
[dxil2spv] Add additional error checking

sudonatalie:
Add some missing error checking for possible nullptrs, and use result of
diagnostic client for setting a succes/failure return value.

### DIFF
--- a/tools/clang/tools/dxil2spv/dxil2spvmain.cpp
+++ b/tools/clang/tools/dxil2spv/dxil2spvmain.cpp
@@ -48,17 +48,14 @@ int __cdecl wmain(int argc, const wchar_t **argv_) {
 #else
 int main(int argc, const char **argv_) {
 #endif // _WIN32
-  // Use same error return code for all dxil2spv failures.
-  const int ERROR_RETURN = 1;
-
   // Configure filesystem for llvm stdout and stderr handling.
   if (llvm::sys::fs::SetupPerThreadFileSystem())
-    return ERROR_RETURN;
+    return EXIT_FAILURE;
   llvm::sys::fs::AutoCleanupPerThreadFileSystem auto_cleanup_fs;
   llvm::sys::fs::MSFileSystem *msfPtr;
   HRESULT hr;
   if (!SUCCEEDED(hr = CreateMSFileSystemForDisk(&msfPtr)))
-    return ERROR_RETURN;
+    return EXIT_FAILURE;
   std::unique_ptr<llvm::sys::fs::MSFileSystem> msf(msfPtr);
   llvm::sys::fs::AutoPerThreadSystem pts(msf.get());
   llvm::STDStreamCloser stdStreamCloser;
@@ -66,7 +63,7 @@ int main(int argc, const char **argv_) {
   // Check input arguments.
   if (argc < 2) {
     llvm::errs() << "Required input file argument is missing\n";
-    return ERROR_RETURN;
+    return EXIT_FAILURE;
   }
 
   // Setup a compiler instance with diagnostics.
@@ -87,5 +84,6 @@ int main(int argc, const char **argv_) {
   clang::dxil2spv::Translator translator(instance);
   translator.Run();
 
-  return instance.getDiagnosticClient().getNumErrors() > 0 ? ERROR_RETURN : 0;
+  return instance.getDiagnosticClient().getNumErrors() > 0 ? EXIT_FAILURE
+                                                           : EXIT_SUCCESS;
 }
--- a/tools/clang/tools/dxil2spv/dxil2spvmain.cpp
+++ b/tools/clang/tools/dxil2spv/dxil2spvmain.cpp
@@ -48,14 +48,17 @@ int __cdecl wmain(int argc, const wchar_t **argv_) {
 #else
 int main(int argc, const char **argv_) {
 #endif // _WIN32
+  // Use same error return code for all dxil2spv failures.
+  const int ERROR_RETURN = 1;
+
   // Configure filesystem for llvm stdout and stderr handling.
   if (llvm::sys::fs::SetupPerThreadFileSystem())
-    return DXC_E_GENERAL_INTERNAL_ERROR;
+    return ERROR_RETURN;
   llvm::sys::fs::AutoCleanupPerThreadFileSystem auto_cleanup_fs;
   llvm::sys::fs::MSFileSystem *msfPtr;
   HRESULT hr;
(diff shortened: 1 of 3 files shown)


tools/clang/tools/dxil2spv/dxil2spvmain.cpp
```diff
@@ -48,22 +48,25 @@ int __cdecl wmain(int argc, const wchar_t **argv_) {
 #else
 int main(int argc, const char **argv_) {
 #endif // _WIN32
+  // Use same error return code for all dxil2spv failures.
+  const int ERROR_RETURN = 1;
```
kuhar, line 52: AFAIK it's idiomatic to use EXIT_SUCCESS and EXIT_FAILURE in `main`: https://en.cppreference.com/w/cpp/utility/program/EXIT_status

tools/clang/tools/dxil2spv/lib/dxil2spv.h
```diff
@@ -115,6 +115,8 @@ class Translator {
   template <unsigned N>
   DiagnosticBuilder emitError(const char (&message)[N],
                               llvm::Value &instruction);
+  template <unsigned N>
+  DiagnosticBuilder emitError(const char (&message)[N], llvm::Type &type);
```
kuhar, line 119: These `emitError` seem very awkward to me. None of the files including dxil2spv.h will be able to use `emitError` except dxil2spv.cpp, because they won't be able to instantiate the function template without its definition.

I'd expect the message argument to be `llvm::StringRef`/`llvm::Twine` or `const char *`, which wouldn't require any template arguments.

kuhar: Also, please mention in the comment above the function if the message is just any string or a format string.

It seems like the underlying function `getCustomDiagID` takes a fixed format string and internally converts it to a `StringRef`: https://sourcegraph.com/github.com/microsoft/DirectXShaderCompiler@45500c12174457e73d2b3276b5c77af9fb979c8e/-/blob/tools/clang/include/clang/Basic/Diagnostic.h?L603. Could we add a second overload that takes StringRef directly and get rid of all the template stuff here?

kuhar (changes requested):

tools/clang/tools/dxil2spv/lib/dxil2spv.h
kuhar: If we can't modify `Diagnostic.h` for some reason, wouldn't it be enough to have a fixed set of diagnostic IDs with a generic format string? I'm thinking something like:
```c++
emitSpecificError(StringRef message, Instruction *I) {
 emitError("Specific error: %s", <cat-message-and-I>);
}
```

kuhar: One last idea: we could also have `emitError` that takes the char array for the specific diagnostic id and returns you an object that you can append additional content to, e.g., `emitError("Error category") << "something failed: " << *I;`, and then emit the diagnostic on destruction. Similar to what we did here: https://github.com/googlestadia/performance-layers/blob/main/layer/logging.h#L29

kuhar: One thing missing for me is what the consequence of having different diagnostic IDs is. I think the intention is that there will be one ID generated for each format string. But what is the use of these IDs? Do we care at all or are only interested in error messages? If it's the latter, maybe we could consider not generating any diagnostics at all and use something like `report_fatal_error`: https://llvm.org/doxygen/ErrorHandling_8cpp_source.html#l00090.

I'd follow what other llvm tools do (e.g., llvm-as/-dis/-mc, etc.) do, instead of copying clang diagnostics 1:1. My guess is that error handling would be simpler there.

sudonatalie: I'm open to considering that this is just a poor design, but in early development it's been really helpful to have non-fatal diagnostics because we can run a new shader through and quickly see a list of issues (all the not-yet-supported opcodes, instructions, types), rather than just the first one encountered. I'm only familiar with the clang-way this is done in existing DXC (and we don't do anything with the IDs AFAIK) but is there a more idiomatic LLVM way to do this?

kuhar: I think you can do something like:
```c++
void emitError(const Twine &message) {
  if (FatalErrors())
    report_fatal_error(message);
  errs() << "Error: " << message << "\n";
}
```

where `FatalErrors()` reads a global variable set by command line arguments or a field in `Context`.

sudonatalie: Sorry for the loads of back and forth. After further discussion and trying all the solutions laid out here, I ultimately decided to keep this as-is if that's alright with the two of you. The `DiagnosticsEngine` is a good layer of abstraction for testing and passing to other future dxil2spv classes that may need to emit diagnostics, and although the `emitSpecificError` prefix solution seemed like a good way to get the templates out of the header file, it forces all callers to a single StringRef arg, so we lose a lot of of our error messaging niceties like being able to stream multiple arguments and automatically translate ints to strings. That means many more cases require a few extra lines of boilerplate sprintf-ing at the `emitError` callsite, which seems worse than the status quo of being to do this:
```
emitError("Error reading %0: %1") << filename << errorOrInputFile.getError().message();
```
or this:
```
emitError("Failed to translate DXIL resource to SPIR-V: %0") << resource->GetID();
```
Hope that makes sense.

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
```diff
@@ -336,11 +350,21 @@ void Translator::createLoadInputInstruction(llvm::CallInst &instruction) {
                                   hlsl::DXIL::OperandIndex::kLoadInputIDOpIdx))
           ->getLimitedValue();
   spirv::SpirvVariable *inputVar = inputSignatureElementMap[inputID];
+  if (!inputVar) {
+    emitError(
+        "No matching SPIR-V input variable found for load instruction: %0",
+        instruction);
```
jaebaek, line 356: Does LLVM support formatting `llvm::CallInst &instruction`?

kuhar (approved):

tools/clang/tools/dxil2spv/lib/dxil2spv.cpp
sudonatalie: Yes, this `emitError` wrapper calls the `llvm::Value::print` function to pretty-print it.

jaebaek (approved):

PR #4451, opened 2022-05-11
Teach llvm-dis to read DXIL out of DXBC files

llvm-beanz:
This change has no impact on the DXC codebase since we don't use
llvm-dis here, but in LLVM upstream I'm using llvm-dis from DXC to
verify bitcode compatability.

The change here is to intercept the file passed into llvm-dis to detect
DXIL Container files, and read the bitcode out of the DXIL part.


tex3d (commented):
Just some error-handling comments.

tex3d (approved):
LGTM!

PR #4452, opened 2022-05-12
Fix use constant corruption from ReplaceUsesForLoweredUDT

### LINKED ISSUE #4421: Simple Amplification Shader internal compiler error repro (using 2021-12-08 dxc.exe release)
Hello.  I've got a quite small reproduction of an "Internal compiler error: access violation ..." using the 2021-12-08 released dxc as follows -

> dxc.exe test.hlsl -E main -T as_6_5 -Od

where 'test.hlsl' contains -

```
struct SharedPayload
{
  uint2 m_a;

  void Foo( in float3 v3 )
  {
    uint3 f16Vec3 = f32tof16(v3);
    m_a.x = f16Vec3.x | (f16Vec3.y<<16);
    m_a.y = f16Vec3.z;
  }
};

groupshared SharedPayload g_payload;

[numthreads(8, 8, 1)]
void main()
{
  g_payload.Foo( 1.0.xxx );
  DispatchMesh(1,1,1,g_payload);
}
```

I'm not claiming this is supposed to compile without errors but I would expect it to report a sensible error message rather than get a memory access violation.

Cheers

Dan

tex3d:
ReplaceUsesForLoweredUDT was calling use.set(NewV), or calling dropAllReferences() on constant user.
This results in modification of the constant operand, which is illegal and led to corruption of constants.

Loop in ReplaceUsesForLoweredUDT relies on all uses being eliminated to terminate, so constant or potentially constant users are now handled with legal replacements and removeDeadConstantUsers at the end as needed.

Fixes #4421

### DIFF
--- a/lib/HLSL/HLLowerUDT.cpp
+++ b/lib/HLSL/HLLowerUDT.cpp
@@ -201,8 +201,9 @@ void hlsl::ReplaceUsesForLoweredUDT(Value *V, Value *NewV) {
   while (!V->use_empty()) {
     Use &use = *V->use_begin();
     User *user = use.getUser();
-    // Clear use to prevent infinite loop on unhandled case.
-    use.set(UndefValue::get(V->getType()));
+    if (Instruction *I = dyn_cast<Instruction>(user)) {
+      use.set(UndefValue::get(I->getType()));
+    }
 
     if (LoadInst *LI = dyn_cast<LoadInst>(user)) {
       // Load for non-matching type should only be vector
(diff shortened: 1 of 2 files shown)


lib/HLSL/HLLowerUDT.cpp
```diff
@@ -288,11 +289,13 @@ void hlsl::ReplaceUsesForLoweredUDT(Value *V, Value *NewV) {
         } else {
           // Could be i8 for memcpy?
           // Replace bitcast argument with new value
-          use.set(NewV);
+          CE->replaceAllUsesWith(
+              ConstantExpr::getBitCast(cast<Constant>(NewV), CE->getType()));
         }
       } else {
         DXASSERT(0, "unhandled constant expr for lowered UTD");
```
pow2clk, line 296: Pre-existing of course, but *UDT

pow2clk (approved):
Looks good! Just a pre-existing typo in an assert no less that you can leave if you prefer

PR #4453, opened 2022-05-12
Disallow illegal Atomic targets + propagate consts

### LINKED ISSUE #4319: Compiler allows to use atomics on bitfields, despite not supporting it
This code compiles
```
struct ColorRGBA {
  uint R : 8;
  uint G : 8;
  uint B : 8;
  uint A : 8;
};

StructuredBuffer<ColorRGBA> testBuffer;

float4 PSMain() : SV_TARGET
{
    InterlockedAdd(testBuffer[0].G, 1);
	return 0.0f;
}
```
But generates 32 bit atomics
```
%3 = call i32 @dx.op.atomicBinOp.i32(i32 78, %dx.types.Handle %2, i32 0, i32 0, i32 0, i32 undef, i32 1)  ; AtomicBinOp(handle,atomicOp,offset0,offset1,offset2,newValue)
```
It should not compile, since there are no instruction that could allow arbitrary bit width atomics

Tested on https://github.com/microsoft/DirectXShaderCompiler/releases/tag/v1.6.2112

### LINKED ISSUE #4377: There's no compile error when doing InterlockedAdd on (readonly) StructuredBuffer
This code compiles
```
StructuredBuffer<int> SRVBuffer : register(t0);

float4 PSMain() : SV_TARGET
{
    InterlockedAdd(SRVBuffer[0], 1);
    return 0.0f;
}
```
https://shader-playground.timjones.io/37a9dfb3c90c42c072fb2f1d70d78f90
Expected behavior is compile error, since you can't modify StructuredBuffer

### LINKED ISSUE #4437: Interlock ops on RO StructuredBuffers don't produce errors
Discovered this by accident, but it appears that DXC does not notice and report the illegality of performing atomic operations (say, InterlockedExchange) on read-only structured buffers.  Code doing this should produce should produce an error.

For example, the following code compiles fine on ShaderPlayground using a cs6_6 profile, instead of reporting an illegal use of writing to _g_input_.  Note, if _g_input_ is changed to a regular buffer, the error is correctly reported.:

    StructuredBuffer<uint> g_input : register(t0, space0);

    [numthreads(32, 1, 1)]
    void CsEntry(uint dtid : SV_DispatchThreadID)
    {
        uint value;
        InterlockedExchange(g_input[dtid], 0, value);
    }

pow2clk:
This disallows a number of illegal destination parameters to atomic
operations. This includes SRVs and other const values, non-groupshared
and non-resourse variables, members of typed resources, and bitfield
members of any resource.

In addition, this correctly propagates const information so that they
can be properly rejected and also the information is reflected.
This involves the consolidation and earlier detection of a number of
these failures as well as an expansion of that detection.
Also adds validation checks that the targets are not const and either
UAVs or groupshared address space.

Add compilation tests as well as validation tests for all.

Fixes #4319
Fixes #4377
Fixes #4437


lib/HLSL/HLOperationLower.cpp
```diff
@@ -7698,50 +7735,10 @@ void TranslateDefaultSubscript(CallInst *CI, HLOperationLowerHelper &helper,  HL
...
-              userCall, "Typed resources used in atomic operations must have a scalar element type.");
-          return;
-        } break;
-        default:
-          // Invalid operations.
-          Translated = false;
-          dxilutil::EmitErrorOnInstruction(userCall, "Invalid operation on typed buffer.");
-          return;
-          break;
-        }
```
pow2clk, line 7744: All of this was just to generate different errors when atomic ops were performed on typed resources had non-scalar elements. Because this is handled elsewhere, the distinction need not be done here anymore

lib/HLSL/HLOperationLower.cpp
```diff
@@ -7754,29 +7751,8 @@ void TranslateDefaultSubscript(CallInst *CI, HLOperationLowerHelper &helper,  HL
...
-          case IntrinsicOp::IOP_InterlockedCompareExchangeFloatBitwise: {
-            dxilutil::EmitErrorOnInstruction(
-                userCall, "Atomic operation targets must be groupshared or UAV.");
-            return;
-          } break;
-          default:
-            dxilutil::EmitErrorOnInstruction(userCall, "Invalid operation on typed buffer.");
-            return;
-            break;
-          }
```
pow2clk, line 7779: As above, this was just to distinguish the error for atomic ops, it's now handled elsewhere

tools/clang/lib/Sema/SemaExpr.cpp
```diff
@@ -4285,7 +4285,7 @@ Sema::CreateBuiltinArraySubscriptExpr(Expr *Base, SourceLocation LLoc,
       // We need to make sure to preserve qualifiers on array types, since these
       // are in effect references.
       if (LHSTy.hasQualifiers())
-        ResultType.setLocalFastQualifiers(LHSTy.getLocalFastQualifiers());
+        ResultType.setLocalFastQualifiers(LHSTy.getQualifiers().getFastQualifiers());
```
pow2clk, line 4288: It's a bit clunky, but it passes along the full const and other qualifiers even if they are not local.

tools/clang/test/HLSLFileCheck/hlsl/intrinsics/atomic/atomic_restypes.hlsl
```diff
@@ -0,0 +1,131 @@
+// RUN: %dxc -T cs_6_0 -DDEST=ro_structbuf -DIDX=    %s | %FileCheck %s
+// RUN: %dxc -T cs_6_0 -DDEST=ro_structbuf -DIDX=[0] %s | %FileCheck %s
```
python3kgae, line 2: What is the rule for linking?
Should we report error on foo when compile to lib or only report error when linking with illegal argument a?
export void foo(inout int a, int i) {
  InterlockedAdd(a, i);
}

pow2clk: This will fail because it is neither a resource nor a groupshared variable. The only valid entries for atomics currently are resources or groupshared. We have some open questions about passing groupshared parameters to functions that I think are outside the scope of this change. We might provide a way to do this in the future though.

What we can do is pass in resource params, which might be something useful to test.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3180,11 +3184,19 @@ static void ValidateFunctionBody(Function *F, ValidationContext &ValCtx) {
...
+        PointerType *ptrType = cast<PointerType>(Ptr->getType());
+        Type *elType = ptrType->getElementType();
         const ShaderModel *pSM = ValCtx.DxilMod.GetShaderModel();
-        if ((T->isIntegerTy(64)) && !pSM->IsSM66Plus())
+        if ((elType->isIntegerTy(64)) && !pSM->IsSM66Plus())
           ValCtx.EmitInstrFormatError(&I, ValidationRule::SmOpcodeInInvalidFunction,
                                       {"64-bit atomic operations", "Shader Model 6.6+"});
+        if (ptrType->getAddressSpace() != DXIL::kTGSMAddrSpace) {
+          ValCtx.EmitInstrError(&I, ValidationRule::InstrAtomicOpNonGroupshared);
+        } else if (GlobalVariable *GV = dyn_cast<GlobalVariable>(Ptr)) {
```
tex3d, line 3196: Multiple thoughts:
1. Why `else`?  Shouldn't a tgsm constant global be a problem too?  Perhaps other validation should fire for that case, since it shouldn't be possible to declare constant groupshared in DXIL I think.
2. The pointer must ultimately come from a global TGSM.
3. The pointer may not itself be a `GlobalVariable`, but it could instead be a GEP of `GlobalVariable`, so you may need to drill through a GEP here.
4. If/when we support unions, I imagine we would also have to drill through pointer bitcast.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -4421,6 +4421,40 @@ static Value* SkipAddrSpaceCast(Value* Ptr) {
   return Ptr;
 }
 
+// For known non-groupshared, verify that the destination param is valid
+void ValidateAtomicDestination(CallInst *CI, HLObjectOperationLowerHelper *pObjHelper) {
+  Value *dest = CI->getArgOperand(HLOperandIndex::kInterlockedDestOpIndex);
+  GetElementPtrInst *gep = nullptr;
+
+  // If we encounter a gep, we may provide a more specific error message
+  while (isa<GetElementPtrInst>(dest)) {
```
tex3d, line 4430: If it's at all possible for the base pointer to be global, this should use `GEPOperator` instead, but comment indicates that this function is only called for non-groupshared case, so that ptr should always originate from an intrinsic call I think.
This was just a passing thought, probably nothing to change here.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -4421,6 +4421,40 @@ static Value* SkipAddrSpaceCast(Value* Ptr) {
...
+    hlsl::HLOpcodeGroup group = hlsl::GetHLOpcodeGroup(destCall->getCalledFunction());
+    while (group == HLOpcodeGroup::HLSubscript) {
+      dest = destCall->getArgOperand(HLOperandIndex::kSubscriptObjectOpIdx);
+      if (!isa<CallInst>(dest))
+        break;
+      destCall = cast<CallInst>(dest);
+      group = hlsl::GetHLOpcodeGroup(destCall->getCalledFunction());
+    }
+    DXIL::ResourceKind RK = pObjHelper->GetRK(destCall);
+    if (DXIL::IsStructuredBuffer(RK))
```
tex3d, line 4446: Shouldn't this also fail if not UAV?  Or was that caught elsewhere?  Perhaps while translating the subscript operation?

lib/HLSL/HLOperationLower.cpp
```diff
@@ -4421,6 +4421,40 @@ static Value* SkipAddrSpaceCast(Value* Ptr) {
...
+    gep = cast<GetElementPtrInst>(dest);
+    dest = gep->getPointerOperand();
+  }
+
+  // Confirm that dest is a UAV
+  if (CallInst *destCall = dyn_cast<CallInst>(dest)) {
+    hlsl::HLOpcodeGroup group = hlsl::GetHLOpcodeGroup(destCall->getCalledFunction());
+    while (group == HLOpcodeGroup::HLSubscript) {
+      dest = destCall->getArgOperand(HLOperandIndex::kSubscriptObjectOpIdx);
+      if (!isa<CallInst>(dest))
```
tex3d, line 4440: I'm not sure what scenarios are possible here.  Could there be subscript->something else->subscript?  What's this operand when not another subscript?  What scenario has multiple layers of subscript operators?

I guess the general concern is about doing this on the HL ops.  Could it be performed before/during CodeGen (earlier is always better for the error reporting), or is that even more complicated?  Or could checking the DXIL translated ops simplify the check?

lib/HLSL/HLOperationLower.cpp
```diff
@@ -7754,29 +7751,8 @@ void TranslateDefaultSubscript(CallInst *CI, HLOperationLowerHelper &helper,  HL
...
-            dxilutil::EmitErrorOnInstruction(
-                userCall, "Atomic operation targets must be groupshared or UAV.");
-            return;
-          } break;
-          default:
-            dxilutil::EmitErrorOnInstruction(userCall, "Invalid operation on typed buffer.");
-            return;
-            break;
-          }
+          dxilutil::EmitErrorOnInstruction(userCall, "Invalid operation on typed buffer.");
```
tex3d, line 7754: Don't you mean on SRV?  I think this error already existed in the deleted block from which this was copied.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6155,14 +6155,23 @@ bool HLSLExternalSource::MatchArguments(
...
+      if (pType.isConstant(actx)) {
         // Can't use a const type in an out or inout parameter.
         badArgIdx = std::min(badArgIdx, iArg);
       }
     }
 
+    // Catch invalid atomic dest parameters
+    if (iArg == kAtomicDstOperandIdx &&
+        IsAtomicOperation(static_cast<IntrinsicOp>(pIntrinsic->Op))) {
+      // bitfield error is confusing
```
tex3d, line 6170: Still confusing?  Any suggestion to improve it?

tools/clang/test/HLSLFileCheck/hlsl/diagnostics/errors/write_const_arrays.hlsl
```diff
@@ -0,0 +1,61 @@
+// RUN: %dxc -T cs_6_0 _HV 2021 %s | FileCheck %s
```
tex3d, line 1: Can we use verifier tests for front-end error checks?

tools/clang/test/HLSLFileCheck/hlsl/diagnostics/errors/write_const_arrays.hlsl
```diff
@@ -0,0 +1,61 @@
...
+  gs_val[0] = 0;
+  g_cbuf[0] = 0;
+  g_robuf[0] = 0;
+
+  // Assigning using out param of builtin function
+  //CHECK: error: no matching function for call to 'sincos'
+  //CHECK: error: no matching function for call to 'sincos'
+  //CHECK: error: no matching function for call to 'sincos'
+  //CHECK: error: no matching function for call to 'sincos'
+  sincos(0.0, local[0], local[0]);
```
tex3d, line 33: But is it because it's `const` or because it's `uint`?  Maybe try with `const float`?  I suppose subsequent tests with GetDimensions demonstrate that `const` at least produces the same error (as unhelpful as it is).

tools/clang/test/HLSLFileCheck/hlsl/intrinsics/atomic/atomic_restypes.hlsl
tex3d: Right, we really can't support atomics on arguments until we support proper references or pointers to the right memory type.

lib/HLSL/HLOperationLower.cpp
tex3d: Obviously, there is already some checking earlier, and a bit later.  So perhaps this is just about my uncertainty around what kinds of things are possible here that may need drilling through or not.

tools/clang/test/DXILValidation/atomics.hlsl
```diff
@@ -0,0 +1,53 @@
...
+RWStructuredBuffer<uint> rw_structbuf;
+RWBuffer<uint> rw_buf;
+RWTexture1D<uint> rw_tex;
+
+// SRVs to plug into atomics
+StructuredBuffer<uint> ro_structbuf;
+Buffer<uint> ro_buf;
+Texture1D<uint> ro_tex;
+
+const groupshared uint cgs_var;
```
tex3d, line 15: What does this produce?  Should we even allow const groupshared global?

lib/HLSL/DxilValidation.cpp
pow2clk: 1. The else case is what catches the tgsm constant global. It is possible to declare a constant groupshared and it validates fine.
2. The validation error below detects when a TGSMAddrSpace pointer doesn't point to a globalvariable. I didn't want to duplicate or confuse the error for this, which is a subset of the ways that could theoretically be encountered.
3. Valid. I'll address that and it'll probably end up removing the else anyway
4. Getting to where there's more and more duplication with the validation below for TGSM pointers. Maybe I can consolidate instead of duplicating

lib/HLSL/HLOperationLower.cpp
pow2clk: There's no way for correct code to make the base pointer global here. It shouldn't be possible at all, but if it does somehow, this will throw an error at least.

lib/HLSL/HLOperationLower.cpp
pow2clk: The while loop over subscripts that I think is troubling you is needed to handle cases such as this:
https://github.com/microsoft/DirectXShaderCompiler/blob/50c0199c17b0b7530cbc0bdddadc3697eaece695/tools/clang/test/HLSLFileCheck/hlsl/intrinsics/atomic/atomic_structuredbuf_i64.hlsl#L36

Someday I'd like to refactor this code such that this is detectable in codegen, but it's not simple. The information to determine these errors is not readily available there and the errors were previously generated during this stage because of that. I decided not to take on that task for now, but we will have to move all this error checking to earlier eventually.

lib/HLSL/HLOperationLower.cpp
pow2clk: Non-UAV resources are considered const at codegen and able to be caught there. Non-resource parameters will fail all these checks and hit the catchall at the end of this function

lib/HLSL/HLOperationLower.cpp
pow2clk: I suppose so. As you say, this is a pre-existing error message that I've preserved. The only change was to generate the more specific error elsewhere. and avoid all this convolution here. That said, the only way I've managed to produce this error is through atomic ops.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Well it was non-existent before, so it's less confusing now. It's appears as an overload mismatch that says `note: candidate function not viable: no known conversion from 'uint' to 'unsigned int &' for 1st argument`, which is confusing, but the mechanism doesn't have a mismatch option for bitfields specifically.

tools/clang/test/DXILValidation/atomics.hlsl
pow2clk: It produces as valid code as can be involving an uninitialized variable. Should we? We certainly have in the past. Fxc accepts it if it's initialized. Of course, that produces an error in dxc because groupshared inits aren't supported.

tools/clang/test/HLSLFileCheck/hlsl/diagnostics/errors/write_const_arrays.hlsl
pow2clk: I suppose so.

tools/clang/test/HLSLFileCheck/hlsl/diagnostics/errors/write_const_arrays.hlsl
pow2clk: Actually, this error is just fine. The note below the error line makes it clear that a const param isn't acceptable. Prior to the qualifier propagation that I added here, we got crashes.

pow2clk (commented):
Respond to feedback

lib/HLSL/HLOperationLower.cpp
pow2clk: Actually, this is the one that I couldn't generate. I suspect it's here as a failsafe. It could probably be converted to an assert and/or internal error, though this gives us more information about where the problem is.

lib/HLSL/HLOperationLower.cpp
tex3d: We have a property right here identifying this as a UAV or not.  Perhaps we should at least DXASSERT that it is a UAV, if the error should normally be caught elsewhere (with comment stating that to explain why it's just an assert here).

lib/HLSL/HLOperationLower.cpp
tex3d: I wonder about more complex cases, such as ones with other types of accesses in between.  Like, what does `simpArrBuf[a].ss[b].longEnding[c][d].x` end up looking like (assuming `d` is defined, and the matrix type is uint)?  Is it subscripts all the way down?  There's a matrix subscript here, but I guess that's an HLSubscript here too at this point.  What about the intermediate struct member, same?

I figured I might as well experiment to see what it looks like just before dxilgen, and I see:
`hl.subscript.[] -> GEP -> GEP -> GEP -> GEP -> hl.subscript.colMajor[] -> GEP -> hl atomic op`
Perhaps that's not what it looks like when it reaches this particular code though?  Does this code handle that gracefully as well?

lib/HLSL/HLOperationLower.cpp
tex3d: maybe your const-ness fixes catch the SRV case earlier, preventing this from being hit.

tools/clang/test/HLSLFileCheck/hlsl/diagnostics/errors/write_const_arrays.hlsl
tex3d: I wasn't aware of the note when I made the comment.  Sounds good.

tools/clang/test/DXILValidation/atomics.hlsl
tex3d: In FXC, while it accepts it, does it actually produce a groupshared declaration in the bytecode for it?  That was my concern.  const groupshared could always just be turned into const global data which could either become immediate values or an immediate constant buffer if dynamically indexed.  Hence the groupshared would basically get ignored, so if you declared it this way, it might be useful to at least warn that the groupshared will basically be ignored.

For uninitialized constant (internal-linkage global or local) variable, we should probably fail on declaration or use of such, since it's impossible for it to be anything but undef.

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d: I see, it's better than nothing!  I guess I was just confused at what this comment in the code was trying to tell the reader of this code.

lib/HLSL/HLOperationLower.cpp
pow2clk: The DXASSERT would have to be pretty complicated because while SRVs can't be here, the resource could be invalid because it's not actually a resource or possibly other values. I'll just add a mostly redundant check

lib/HLSL/HLOperationLower.cpp
pow2clk: This would be simpler after more lowering in valid usages because the lowered atomic op would have a direct pointer to the resource. However, that lowering is performed dependent on the destination argument being a subscripted resource. If it isn't, lowering is missed entirely leaving an unlowered hl intrinsic. That leaves this as the best place to do it even though it requires some "drilling"

tools/clang/test/DXILValidation/atomics.hlsl
pow2clk: I have an issue on me to correctly initialize groupshared, which would make this code valid. I'm not sure what we should do, but I think it's outside the scope of this change.

lib/HLSL/DxilValidation.cpp
pow2clk: I didn't end up merging any of the code since they were dealing with different variables at different stages anyway. I drill through GEP and also BC since we'd rather not rev the validator for an HLSL version that might have unions.

tools/clang/test/HLSL/write-const-arrays.hlsl
```diff
@@ -0,0 +1,50 @@
...
+  g_tex.GetDimensions(g_robuf[0]);                          /* expected-error {{no matching member function for call to 'GetDimensions'}} expected-note {{candidate function template not viable: requires 3 arguments, but 1 was provided}} */
+
+
+  // Assigning using dest param of atomics
+  // Distinct because of special handling of atomics dest param
+  InterlockedAdd(local[0], 1);                              /* expected-error {{no matching function for call to 'InterlockedAdd'}} expected-note {{candidate function not viable: 1st argument ('const uint') would lose const qualifier}} expected-note {{candidate function not viable: no known conversion from 'const uint' to 'unsigned long long &' for 1st argument}} */
+  InterlockedAdd(gs_val[0], 1);                             /* expected-error {{no matching function for call to 'InterlockedAdd'}} expected-note {{candidate function not viable: 1st argument ('const uint') would lose const qualifier}} expected-note {{candidate function not viable: no known conversion from 'const uint' to 'unsigned long long' for 1st argument}} */
```
tex3d, line 46: Strange, why does this one (and next two) say `to 'unsigned long long'` instead of `'unsigned long long &'` like the first one?
And why would it be `unsigned long long` in the first place, since I presume this should take a 32-bit reference?

tools/clang/test/HLSL/atomics-on-bitfields.hlsl
```diff
@@ -0,0 +1,30 @@
...
+  InterlockedCompareStore(str[tid.y].q, 3, 1);              /* expected-error {{no matching function for call to 'InterlockedCompareStore'}} expected-note {{candidate function not viable: no known conversion from 'uint' to 'unsigned int' for 1st argument}} expected-note {{candidate function not viable: no known conversion from 'uint' to 'unsigned long long' for 1st argument}} */
+
+  InterlockedOr(gs.q, 2);                                   /* expected-error {{no matching function for call to 'InterlockedOr'}} expected-note {{candidate function not viable: 1st argument ('__attribute__((address_space(3))) uint') is in address space 3, but parameter must be in address space 0}} expected-note {{candidate function not viable: no known conversion from '__attribute__((address_space(3))) uint' to 'unsigned long long' for 1st argument}} */
```
tex3d, line 28: It's a shame about the address space error, since it's misleading.

tex3d (approved):
Comments about strange/misleading error messages are really just to note, not for fixing in this PR.  Makes sense to add new verifier tests to VerifierHelper though.

tools/clang/test/HLSL/write-const-arrays.hlsl
pow2clk: Another "quirk" of how atomics are lowered. The dest param is not made a reference until the candidate function is first added to the used intrinsics. Subsequent candidates are found to match the existing so their params don't get converted. It's only an issue for matching failures. We should talk about why it was designed this way in the first place.

tools/clang/test/HLSL/atomics-on-bitfields.hlsl
pow2clk: Less so than the previous I'd say. Bitfields just make the compiler do your shifting and casting for you. The result of that operation is a local. Granted, shader authors may not know that. We could consider creating some custom mismatch types at some point. That would be useful in all versions of LLVM unless they've seriously overhauled this system

PR #4456, opened 2022-05-12
Fix memcpy replacement removing memcpy to output argument

### LINKED ISSUE #4455: Incorrect DXIL for library functions returning arrays
When a library function returns an array, the DXIL generated has `%agg.result` but the body of the function is void and does not write to it. Further, the the callers of said function do not make use of the allocated input after the function call.

```
export int Test0(float f[4]) [4]
{
    // return { int(f[0]), int(f[1]), int(f[2]), int(f[3]) };
    int ret[4];
    ret[0] = int(f[0]);
    ret[1] = int(f[1]);
    ret[2] = int(f[2]);
    ret[3] = int(f[3]);
    return ret;
}
```
```
; Function Attrs: nounwind readnone
define void @"\01?Test0@@YA$$BY03HY03M@Z"([4 x i32]* noalias nocapture sret %agg.result, [4 x float]* nocapture %f) #0 {
  ret void
}
```

[library-return-arrays.zip](https://github.com/microsoft/DirectXShaderCompiler/files/8682246/library-return-arrays.zip)

tex3d:
Fixes #4455

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/functions/return_type/array/lib_function.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/functions/return_type/array/lib_function.hlsl
@@ -1,0 +1,18 @@
+// RUN: %dxc -T lib_6_3 %s | FileCheck %s
+// RUN: %dxc -T lib_6_6 %s | FileCheck %s
+
+// Tests that non-entry point functions can return arrays in library profile.
+
+// CHECK: define void
+// CHECK-SAME: getA
+// CHECK-SAME: ([2 x i32]* noalias nocapture sret %[[result:.*]])
+export int getA() [2]
+{
+  // CHECK: %[[gep1:.*]] = getelementptr inbounds [2 x i32], [2 x i32]* %[[result]], i32 0, i32 0
+  // CHECK: store i32 1, i32* %[[gep1]]
+  // CHECK: %[[gep2:.*]] = getelementptr inbounds [2 x i32], [2 x i32]* %[[result]], i32 0, i32 1
+  // CHECK: store i32 2, i32* %[[gep2]]
+  int a[2] = { 1, 2 };
+  return a;
+}
+// CHECK: ret void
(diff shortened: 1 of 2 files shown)


tools/clang/test/HLSLFileCheck/hlsl/functions/return_type/array/lib_function.hlsl
```diff
@@ -0,0 +1,18 @@
...
+// CHECK-SAME: ([2 x i32]* noalias nocapture sret %[[result:.*]])
+export int getA() [2]
+{
+  // CHECK: %[[gep1:.*]] = getelementptr inbounds [2 x i32], [2 x i32]* %[[result]], i32 0, i32 0
+  // CHECK: store i32 1, i32* %[[gep1]]
+  // CHECK: %[[gep2:.*]] = getelementptr inbounds [2 x i32], [2 x i32]* %[[result]], i32 0, i32 1
+  // CHECK: store i32 2, i32* %[[gep2]]
+  int a[2] = { 1, 2 };
+  return a;
+}
```
pow2clk, line 17: Not essential, but another way to encounter this would be with an out param that gets assigned wholesale from a local variable. Something like 
```hlsl
export int getB(out uint b[2]) {
  int a[2] = {1,2};
  b = a;
}
```

pow2clk (approved):
I got curious about other ways this could be encountered and did some experiments. structs work around this, but out arrays will hit it to if they are assigned in a way that produces a memcpy.

tools/clang/test/HLSLFileCheck/hlsl/functions/return_type/array/lib_function.hlsl
tex3d: Yup!  But it would need some slight fixes there (void return type, match out type with local):
```
export void getB(out int b[2]) {
  int a[2] = {1,2};
  b = a;
}
```

Interestingly, if b doesn't match a in component type, I think HLSL should be able to implicitly convert, but instead we get this error:
```
error: cannot implicitly convert from 'int [2]' to 'uint __restrict[2]'
  b = a;
    ^
```
This looks like another bug 😒

At least it works with a cast:
```
  b = (uint[2])a;
```

tex3d: Filed #4458 for the above issue.

PR #4457, opened 2022-05-13
Enable control flow guard

hekota:


CMakeLists.txt
```diff
@@ -105,6 +105,11 @@ endif()
 if(NOT CMAKE_VERSION VERSION_LESS "3.13" AND WIN32)
   add_link_options(/DEBUGTYPE:CV,FIXUP,PDATA /INCREMENTAL:NO)
 endif()
+
+# enable control flow guard
+add_compile_options(/guard:cf)
```
tex3d, line 110: Need to guard option in:
```
if(WIN32)
...
endif()
```

tex3d (approved):
Just needs `if(WIN32)` around the options so it doesn't break linux/mac builds.

tex3d (approved):
LGTM!

PR #4459, opened 2022-05-16
instcombine pass before jumpthread for lifetimes

pow2clk:
Lifetime markers require a cleanup pass to eliminate some unnecessary
code, however one of those passes was JumpThreading, which only
eliminated certain dead blocks, which eliminated phi instructions
leaving looped self increments unconditionally incrementing themselves,
which ended up with infinite recursion when trying to process that
instruction.

By adding Instcombining first, all unreachable blocks are elimianted
first so the partial elimination doesn't leave such bad phi users. To
avoid losing convergence for DXIL specific operations, we must first run
the convergent marker pass, so these passes are moved a little later
than before.


pow2clk:
@rkarrenberg for visibility since he added this pass in the first place

PR #4460, opened 2022-05-16
Fix broken ternary on matrix return type (#4434)

### LINKED ISSUE #4434: Internal Compiler Error with ternary operations on matrices
December 2021 DXC crashes when compiling this shader:

```
float2x2 crashingFunction( bool b )
{
    float2x2 x = { 0.0, 0.0, 0.0, 0.0 };
    float2x2 y = { 0.0, 0.0, 0.0, 0.0 };
    return b ? x : y; // <-- this is the issue
}

float2x2 workingFunction( bool b )
{
    float2x2 x = { 0.0, 0.0, 0.0, 0.0 };
    float2x2 y = { 0.0, 0.0, 0.0, 0.0 };
    if( b ) { return x; } else { return y; }
}

[numthreads( 1, 1, 1 )]
void CSMain()
{
    if( crashingFunction(true)[0][0] > 0 )
        return;
    if( workingFunction(true)[0][0] > 0 )
        return;
} 
```
Compile command:
`dxc.exe /Ges -HV 2021 /T cs_6_0 /Fo repro.obj /E CSMain repro.hlsl`

Output:
`Internal compiler error: access violation. Attempted to read from address 0x0000000000000018`

llvm-beanz:
HLSL ternary operators that result in vector or matrix types need
special handling even if the condition is not a vector. This change
allows vector and matrix result types for ternary operators even if
matrix and vector conditions are not allowed.

Preventing matrix and vector conditions is done in Sema, so we should
never be able to reach this code path with a matrix or vector condition
unless the HLSL language version is < 2021.

Fixes #4434

### DIFF
--- a/tools/clang/lib/CodeGen/CGExprScalar.cpp
+++ b/tools/clang/lib/CodeGen/CGExprScalar.cpp
@@ -20,6 +20,7 @@
 #include "TargetInfo.h"
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/DeclObjC.h"
+#include "clang/AST/HlslTypes.h"
 #include "clang/AST/RecordLayout.h"
 #include "clang/AST/StmtVisitor.h"
 #include "clang/Basic/TargetInfo.h"
@@ -3701,7 +3702,7 @@ VisitAbstractConditionalOperator(const AbstractConditionalOperator *E) {
   }
   // HLSL Change Starts
   if (CGF.getLangOpts().HLSL && !CGF.getLangOpts().EnableShortCircuit) {
-    // HLSL does not short circuit by default.
+    // HLSL does not short circuit by default before HLSL 2021
     if (hlsl::IsHLSLVecType(E->getType()) || E->getType()->isArithmeticType()) {
       llvm::Value *CondV = CGF.EmitScalarExpr(condExpr);
       llvm::Value *LHS = Visit(lhsExpr);
@@ -3728,6 +3729,13 @@ VisitAbstractConditionalOperator(const AbstractConditionalOperator *E) {
       return CGF.CGM.getHLSLRuntime().EmitHLSLMatrixOperationCall(
           CGF, E, LHS->getType(), {Cond, LHS, RHS});
     }
+  }
+
+  llvm::Value *ResultAlloca = nullptr;
+  if (CGF.getLangOpts().HLSL && CGF.getLangOpts().EnableShortCircuit &&
+      hlsl::IsHLSLMatType(E->getType())) {
+    llvm::Type *MatTy = CGF.ConvertType(E->getType());
+    ResultAlloca = Builder.CreateAlloca(MatTy);
   }
   // HLSL Change Ends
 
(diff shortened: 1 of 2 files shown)


tools/clang/lib/CodeGen/CGExprScalar.cpp
```diff
@@ -3700,7 +3700,13 @@ VisitAbstractConditionalOperator(const AbstractConditionalOperator *E) {
     return tmp5;
   }
   // HLSL Change Starts
-  if (CGF.getLangOpts().HLSL && !CGF.getLangOpts().EnableShortCircuit) {
+  // This code generates ternary results for vector and matrix types, and needs
+  // to execute if the result of the expression is a vector or matrix type even
+  // if the condition is not. Conditions are not allowed to be vector or matrix
```
tex3d, line 3705: But don't we still want to avoid emitting these outside the conditional when short-circuiting?
```cpp
      llvm::Value *LHS = Visit(lhsExpr);
      llvm::Value *RHS = Visit(rhsExpr);
```
I'm not sure the fix is this simple, or I'm missing something here.

tex3d: In fact, I suspect we are missing tests for short-circuiting to ensure side-effects from dead branches are not seen for functions that return matrix and vector.

llvm-beanz: Nope, you're right. We need to generate that under a branch instead of outside the branch.

tex3d:
Test failures are due to our preservation of value names in debug builds.

tools/clang/lib/CodeGen/CGExprScalar.cpp
```diff
@@ -3729,6 +3730,13 @@ VisitAbstractConditionalOperator(const AbstractConditionalOperator *E) {
           CGF, E, LHS->getType(), {Cond, LHS, RHS});
     }
   }
+
+  llvm::Value *ResultAlloca = nullptr;
+  if (CGF.getLangOpts().HLSL && CGF.getLangOpts().EnableShortCircuit &&
+      hlsl::IsHLSLMatType(E->getType())) {
+    llvm::Type *MatTy = CGF.ConvertType(E->getType());
+    ResultAlloca = Builder.CreateAlloca(MatTy);
```
pow2clk, line 3738: We previously, perhaps superstitiously, tried to push all the allocas at the top of the code block. Is this going to maintain that? Need it?

elsewhere, we've used `dxilutil::FindAllocaInsertionPoint(F)`, which would also allow this operation to be done below where the alloca is used

tex3d: At the very least, allocas must be inserted in the entry block.  You should use `CGF.CreateTempAlloca` instead, that will place it correctly.

tex3d: Alsok, shouldn't we do this after the `isCheapEnoughToEvaluateUnconditionally` part when we know we'll use it?

pow2clk (approved):
All comments addressed. Let's get it in!

PR #4462, opened 2022-05-18
[cmake] Allow to set C/C++ linker

kuhar:
Backport an LLVM cmake option to specify the C/C++ linker to use.
This implementation is simplified and unlike the llvm-project one does not contain any linker-specific tweaks.

On my system, setting `-DLLVM_USE_LINKER=mold` reduces the full build time from 81s to 43s.


llvm-beanz (approved):
LGTM, even if I tend to prefer `lld` over `mold`...

jaebaek:
LGTM

PR #4467, opened 2022-05-18
Correctly copy to resource_desc1 in rawgather test

pow2clk:
The previous approach was assigning garbage to the final field of desc1
due to alignment padding in desc0 where that field would be.
This just duplicates the utility function behavior to copy one desc to
the new one

### DIFF
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -4651,7 +4651,7 @@ void ExecutionTest::DoRawGatherTest(ID3D12Device *pDevice, RawGatherTexture *raw
   for (UINT y = 0; y < yDim; y++)
     for (UINT x = 0; x < xDim; x++)
       rawTex->SetElement(ix++, x, y);
-  D3D12_RESOURCE_DESC tex2dDesc = CD3DX12_RESOURCE_DESC::Tex2D(resFormat, xDim, yDim);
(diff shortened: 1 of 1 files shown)


tex3d (approved):
LGTM!

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -939,7 +956,7 @@ class ExecutionTest {
       CComPtr<ID3D12Device10> pDevice10;
       // Copy resDesc0 to resDesc1 zeroing anything new
       D3D12_RESOURCE_DESC1 resDesc1 = {0};
-      memcpy(&resDesc1, &resDesc, sizeof(resDesc));
+      CopyDesc0ToDesc1(resDesc1, resDesc) {
```
sooknarine, line 959: Looks like there was a typo here, should be `CopyDesc0ToDesc1(resDesc1, resDesc);` but has an open curly brace :(

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -908,6 +908,23 @@ class ExecutionTest {
     }
   }
 
+#if defined(NTDDI_WIN10_CU) && WDK_NTDDI_VERSION >= NTDDI_WIN10_CU
+  // Copy common fields from desc0 to desc1 and zero out the new one
+  void CopyDesc0ToDesc1(D3D12_RESOURCE_DESC1 &desc1, const D3D12_RESOURCE_DESC1 &desc0) {
```
sooknarine, line 913: Also, I think the second argument needs to `D3D12_RESOURCE_DESC` to compile.

PR #4468, opened 2022-05-19
Fix lifetime marker detection causing other users to be missed

tex3d:
The codepath to detect lifetime-marker only uses (bitcast -> lifetime intrinsic), failed to check the instruction itself, which could be a user function call directly using the alloca ptr instead of the expected bitcast for the lifetime marker scenario.
If that user function call was unused (for instance, if it returns void), then it will improperly assume that the instruction is inconsequential and only used by lifetime markers, or is unused, thus doesn't matter.

The fix is to only check onlyUsedByLifetimeMarkers if the instruction is a bitcast, since that's the pattern for lifetime marker use on the alloca.  If the bitcast is unused, that's also safe to continue, since it does nothing on its own.

BreakUpArrayAllocas during DxilLoopUnroll had the same issue.  Targeting and testing this led to several issues (crashes) around multi-dimensional arrays and library paths.  For now, there is no simple way to pass multi-dimensional arrays directly to function calls across library boundaries, without either relaxing validation or lowering function parameter types.  However, it's legal to use the offline-linking target with these, and that scenario shouldn't crash or generate bad code.

Tests have been added for multi-dim resource array passing and passing of a sub-array type to a user function, which both used to crash or assert.

### DIFF
--- a/lib/Transforms/Scalar/LowerTypePasses.cpp
+++ b/lib/Transforms/Scalar/LowerTypePasses.cpp
@@ -540,15 +540,15 @@ protected:
 // Recurse users, looking for any direct users of array or sub-array type,
 // other than lifetime markers:
 bool MultiDimArrayToOneDimArray::isSafeToLowerArray(Value *V) {
-  if (isa<BitCastOperator>(V))
-    if (!onlyUsedByLifetimeMarkers(V))
-      return false;
   if (!V->getType()->getPointerElementType()->isArrayTy())
     return true;
   for (auto it = V->user_begin(); it != V->user_end();) {
     User *U = *it++;
-    if (isa<GEPOperator>(U) || isa<BitCastOperator>(V) ||
-        isa<AddrSpaceCastInst>(U) || isa<ConstantExpr>(U)) {
+    if (isa<BitCastOperator>(U)) {
+      // Bitcast is ok because source type can be changed.
+      continue;
+    } else if (isa<GEPOperator>(U) || isa<AddrSpaceCastInst>(U) ||
+               isa<ConstantExpr>(U)) {
       if (!isSafeToLowerArray(U))
         return false;
     } else {
--- a/lib/Transforms/Scalar/LowerTypePasses.cpp
+++ b/lib/Transforms/Scalar/LowerTypePasses.cpp
@@ -136,7 +136,6 @@ bool LowerTypePass::runOnModule(Module &M) {
     Finder.processModule(M);
   }
 
-  std::vector<AllocaInst*> multiDimAllocas;
   for (Function &F : M.functions()) {
     if (F.isDeclaration())
       continue;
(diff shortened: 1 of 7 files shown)


lib/Transforms/Scalar/LowerTypePasses.cpp
```diff
@@ -535,8 +534,39 @@ class MultiDimArrayToOneDimArray : public LowerTypePass {
...
+      if (!isSafeToLowerArray(U))
+        return false;
+    } else if (BitCastInst *BCI = dyn_cast<BitCastInst>(U)) {
+      if (onlyUsedByLifetimeMarkers(U))
+        continue;
+    } else if (isa<AddrSpaceCastInst>(U)) {
+      if (!isSafeToLowerArray(U))
+        return false;
+    } else if (ConstantExpr *CE = dyn_cast<ConstantExpr>(U)) {
+      if (CE->getOpcode() == Instruction::BitCast && onlyUsedByLifetimeMarkers(U))
```
adam-yang, line 558: Don't we have BitCastOperator?

tools/clang/test/HLSLFileCheck/hlsl/control_flow/attributes/unroll/lib_2d_array_call_export.hlsl
```diff
@@ -0,0 +1,53 @@
...
+// Ensure BreakUpArrayAllocas doesn't misfire and ignore the function call
+// users, causing everything but the calls to be moved to separate scalar
+// arrays, then all that code, being dead, getting deleted, leaving just the
+// passing of uninitialized arrays to the function calls.
+
+// CHECK-DAG: %[[h0:.*]] = load %dx.types.Handle, %dx.types.Handle* @{{.*}}buf0
+// CHECK-DAG: %[[h1:.*]] = load %dx.types.Handle, %dx.types.Handle* @{{.*}}buf1
+// CHECK-DAG: %[[h2:.*]] = load %dx.types.Handle, %dx.types.Handle* @{{.*}}buf2
+// CHECK-DAG: %[[h3:.*]] = load %dx.types.Handle, %dx.types.Handle* @{{.*}}buf3
+// CHECK-DAG: %[[gep0:.*]] = getelementptr inbounds [2 x [2 x %dx.types.Handle]], [2 x [2 x %dx.types.Handle]]* %buffers, i32 0, i32 0, i32 0
```
adam-yang, line 14: Is it okay to just hardcode the name of the value to be %buffer?

adam-yang:
Should we mention the validator change with the 2D array?

tools/clang/test/HLSLFileCheck/hlsl/control_flow/attributes/unroll/lib_2d_array_call_export.hlsl
tex3d: Names will/may be stripped in release build, which reminds me this uses `%buffers` and that could be stripped (I think), so I should fix that at least.

tex3d:
> Should we mention the validator change with the 2D array?

Good point.  I was planning on updating the title/description to reflect expanded changes anyway, and will make sure those make it to the commit when merging.

PR #4470, opened 2022-05-23
dxc: Use relative include path instead of absolute path

lygstate:
Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>


include/dxc/Support/WinAdapter.h
```diff
@@ -15,8 +15,70 @@
...
+#ifndef _In_opt_count_
+#define _In_opt_count_(s)
+#endif
+#ifndef _In_bytecount_
+#define _In_bytecount_(s)
+#endif
+#ifndef _Maybenull_
+#define _Maybenull_
+#endif
+constexpr unsigned char nybble_from_hex(char c) {
```
kusma, line 39: I don't think you can use constexpr outside of the __cplusplus guard, no?

lygstate: Good catch, you are right

lygstate: __CRT_UUID_DECL only for C++, and define to nothing for C

kusma:
Is this solving the same problem as #4178?

lygstate:
> Is this solving the same problem as #4178?

Something alike, I quck look at #4178, I found it's `#include <windows.h>` in `include/dxc/Support/WinAdapter.h` and that's will expose the whole `windows.h` and would easily cause all sorts of compiling errors.

gongminmin:
CMAKE_CXX_COMPILER_FRONTEND_VARIANT is introduced in cmake 3.14. Maybe CMAKE_CXX_COMPILER_ID is a more compatible option?

damyanp:

PR #4471, opened 2022-05-23
Fix crash for llvm-dis for windows build.

python3kgae:


tex3d (approved):

llvm-beanz:
Have you verified that this works on non-Windows builds and with LLVM_ENABLE_TESTS=On?

llvm-dis is only used if LLVM_ENABLE_TESTS=On, and that build mode requires being able to opt out of the MS file system abstractions.

python3kgae:
> 

Tested llvm/test/tools/dxil-dis on llvm/main.
6 pass, 1 fail.  The fail not related to this change.

PR #4472, opened 2022-05-23
Correct const check for atomic methods

pow2clk:
The check for a const destination param for atomic functions mistakenly
encompassed the first param of atomic methods, which is an offset and
can very well be const. This changes the check to account for methods
when relevant.

Adds a test for this and a few other aspects of atomic methods


tex3d (approved):
LGTM!

PR #4473, opened 2022-05-24
change filter comparison type for samplers, update test name

bob80905:
Previously, no filter was specified for the samplers that perform a comparison. So, the filter defaulted to D3D_FILTER_ANISOTROPIC. The filter type should be D3D_FILTER_COMPARISON_ANISOTROPIC instead.


sooknarine:
Thanks!  This fixes the issue I saw with the results from `tex.SampleCmpLevelZero()`.

https://github.com/microsoft/DirectXShaderCompiler/blob/9cd1f68c46d4fcf822a3e6468d33a0cbabeda38f/tools/clang/test/HLSL/ShaderOpArith.xml#L3698

tex3d (approved):
LGTM!

PR #4474, opened 2022-05-25
Replaced some code in dxil-cond-mem2reg with a dxilutils helper. Fixed a crash.

adam-yang:
- Replaced the code to delete unread allocas in dxil-cond-mem2reg with calling the helper in dxilutils.
  - Fixed a crash in the helper caused by walking the entry in forward order, which could invalidate next iterator.
  - Added handling of memcpy's in the helper.
  - Added tests that actually exercise the memcpy path.


tex3d (approved):
LGTM!

PR #4476, opened 2022-05-26
Skip trivially dead GVs during SROA

pow2clk:
Similar to how allocas are skipped when they are popped off of the
worklist and found to have no users, GVs are similarly removed and no
further processing is performed. That processing flattened the unused
GV, creating new child GVs and associating the debug info from the
parent with the children. Since the unused GV had all its uses replaced
with another, no debug info is found and the compiler crashed.

Also includes more graceful failure if debug info isn't found

Add a test that prompts that memcpy replacement order and verifies that
it doesn't crash and doesn't flatten the unused GV

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -1704,7 +1704,7 @@ bool SROAGlobalAndAllocas(HLModule &HLM, bool bHasDbgInfo) {
   // alloca. Big alloca will be split to smaller piece first, when process the
   // alloca, it will be alloca flattened from big alloca instead of a GEP of
   // big alloca.
-  auto size_cmp = [&DL](const Value *a0, const Value *a1) -> bool {
+  auto size_cmp = [&DL, bHasDbgInfo](const Value *a0, const Value *a1) -> bool {
     Type *a0ty = a0->getType()->getPointerElementType();
     Type *a1ty = a1->getType()->getPointerElementType();
     bool isUnitSzStruct0 =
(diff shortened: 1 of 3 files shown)


lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -1713,9 +1713,12 @@ bool SROAGlobalAndAllocas(HLModule &HLM, bool bHasDbgInfo) {
...
     auto sz1 = DL.getTypeAllocSize(a1ty);
-    if (sz0 == sz1 && (isUnitSzStruct0 || isUnitSzStruct1))
-      return getNestedLevelInStruct(a0ty) < getNestedLevelInStruct(a1ty);
-    return sz0 < sz1;
+    if (sz0 == sz1 && (isUnitSzStruct0 || isUnitSzStruct1)) {
+      sz0 = getNestedLevelInStruct(a0ty);
+      sz1 = getNestedLevelInStruct(a1ty);
+    }
+    // If sizes are equal, tiebreak with alphabetically lesser at higher priority
+    return sz0 < sz1 || (bHasDbgInfo && sz0 == sz1 && a0->getName() > a1->getName());
```
tex3d, line 1721: Are a0 and a1 always guaranteed to have a name?

pow2clk: That method is guaranteed to return a StringRef. If there is no name, it returns an empty string. In debugging case, it's at least very likely, but I didn't think it was worth prechecking the presence of a name since there is this acceptable fallback.

adam-yang: Don't we have some weirdness where Release build of the compiler vs Debug build of the compiler have different behaviours with giving values names? Not sure if it's limited to certain types of values or if it affects allocas.

adam-yang (approved):
👍

PR #4482, opened 2022-05-27
Rename struct members of memcpy pointer analysis

pow2clk:
Clarify some comments and rename some members to more consistent
terminology to make the functionality clearer

This includes removing StoredOnce status from memcpy pointer status
It was never differentiated. This removal allows for some code
simplification, which gains little perf, but reduces confusion


damyanp:

PR #4483, opened 2022-05-27
Fix leak of FixedSizeMemoryStream object from DxilConv

tex3d:
Thanks @jenatali for noticing the errant Detach()!  I think it was intended to be `->Detach()`, but was `.Detach()`.  The first would have simply cleared the pointer to the buffer that the stream object didn't own, the second tells CComPtr not to clean up the object itself, so nobody was cleaning it up.

However, `->Detach()` is also unnecessary because the stream can't and won't free the buffer that we want to keep anyway.


hekota (approved):
LGTM!

jenatali (approved):

Issue #4447, opened 2022-05-10
Disassembly feature request: Annotations for entrypoint properties


jenatali:
Looking at HLSL source for a compute shader, the annotation for thread group size is pretty prominent, e.g.:
```hlsl
[numthreads(32, 1, 1)]
```

But looking at the disassembly of the generated DXIL, it can be really difficult to find that same information, since it just ends up as unlabeled metadata, e.g.:
```
!16 = !{i32 32, i32 1, i32 1}
```

It'd be great if metadata that contributes to entrypoint properties could be surfaced more explicitly, like the signature or resource information is in comments before the actual shader code.

bob80905:
The other entry point properties were added and tested with:
https://github.com/microsoft/DirectXShaderCompiler/pull/4491

PR #4484, opened 2022-05-27
Add numthreads to disassembly output

bob80905:
Previously, for CS, MS, and AS shaders, the metadata at the top of the dxil disassembly would not mention anything about numthreads attributes for the respective shader, and there would be an obscure instruction in the dxil reflecting the numthreads configuration. 
Now, numthreads configurations per shader (that supports it) will be seen in the disassembly spew at the top of the dxil output.


pow2clk (approved):
Thanks for updating the description!

bob80905:
This PR is for [#4447](https://github.com/microsoft/DirectXShaderCompiler/issues/4447)

tools/clang/tools/dxcompiler/dxcdisassembler.cpp
```diff
@@ -1294,6 +1294,7 @@ void PrintPipelineStateValidationRuntimeInfo(const char *pBuffer,
 
   const unsigned offset = sizeof(unsigned);
   const PSVRuntimeInfo0 *pInfo = (const PSVRuntimeInfo0 *)(pBuffer + offset);
+  const PSVRuntimeInfo2 *pInfo2 = (const PSVRuntimeInfo2 *)(pBuffer + offset);
```
tex3d, line 1297: The first DWORD that is skipped here by `offset = sizeof(unsigned)` is the size of the `PSVRuntimeInfo` structure.  Only if it is `>= sizeof(PSVRuntimeInfo2)` should you be casting it to that.

In fact, this should use `DxilPipelineStateValidation`, since it handles all this.  Here's how:
```
DxilPipelineStateValidation PSV;
PSV.InitFromPSV0(pBuffer, uBufferSize);
const PSVRuntimeInfo0 *pInfo0 = PSV.GetPSVRuntimeInfo0();
const PSVRuntimeInfo2 *pInfo2 = PSV.GetPSVRuntimeInfo2();
...
if (pInfo2) {
  OS << comment << " Numthreads: (" << pInfo2->NumThreadsX << "," << pInfo2->NumThreadsY << "," << pInfo2->NumThreadsZ << ")\n";
}
```

This requires that size be passed in here, which should be available from the part size when `PSV0` part was looked up.

PR #4491, opened 2022-05-31
dxil disassembly: update check convention, add entry point properties on topology for mesh shaders

bob80905:
DXIL disassembly now additionally checks topology values for mesh / geometry shaders, in addition to cleaning up the convention / format for checking the numthreads attribute.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/disassembler/geometrytest.hlsl
+++ b/tools/clang/test/HLSLFileCheck/disassembler/geometrytest.hlsl
@@ -1,7 +1,17 @@
-// RUN: %dxc -E main -T gs_6_0 %s | FileCheck %s
+// RUN: %dxc -E main1 -T gs_6_0 %s | FileCheck %s -check-prefix=CHECK1
+// CHECK1: OutputTopology=point
+// RUN: %dxc -E main2 -T gs_6_0 %s | FileCheck %s -check-prefix=CHECK2
+// CHECK2: OutputTopology=line
+// RUN: %dxc -E main3 -T gs_6_0 %s | FileCheck %s -check-prefix=CHECK3
+// CHECK3: OutputTopology=triangle
+// RUN: %dxc -E main4 -T gs_6_0 %s | FileCheck %s -check-prefix=CHECK4
+// CHECK4: OutputTopology=line
+// RUN: %dxc -E main5 -T gs_6_0 %s | FileCheck %s -check-prefix=CHECK5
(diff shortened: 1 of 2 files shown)


tools/clang/test/HLSLFileCheck/disassembler/amplificationtest.hlsl
```diff
@@ -1,6 +1,6 @@
 // RUN: %dxc /T as_6_6 /E ASMain %s | FileCheck %s
 // CHECK: Amplification Shader
-// CHECK: Numthreads: (2,4,1)
+// CHECK: NumThreads=(2,4,1)
```
pow2clk, line 3: I don't prefer the formatting, but I suppose it's consistent with what was already there. So let's be consistent.

tools/clang/test/HLSLFileCheck/disassembler/geometrytest.hlsl
```diff
@@ -0,0 +1,23 @@
...
+{
+  float4 r = array[0];
+  MyStruct a = (MyStruct)0;
+
+  a.pos = array[r.x] + 1;
+
+  OutputStream0.Append(a);
+  OutputStream0.RestartStrip();
+
+}
```
pow2clk, line 23: This is now a duplicate of gs_precise_output.hlsl. We could have a dedicated test that verifies all the GS pipeline runtime info. I think testing the full list and using different explicit topologies to verify that they are printed properly would be useful. If you were to convert this shader into that test, that would be alright.

tools/clang/test/HLSLFileCheck/shader_targets/mesh/mesh.hlsl
```diff
@@ -2,7 +2,8 @@
 // RUN: %dxc -E main -T ms_6_5 %s | %D3DReflect %s | FileCheck -check-prefix=REFL %s
 
 // CHECK: Mesh Shader
-// CHECK: Numthreads: (32,1,1)
+// CHECK: MeshOutputTopology=triangle
+// CHECK: NumThreads=(32,1,1)
```
pow2clk, line 6: As with GS, it would be useful to have a mesh shader test that verifies all of the possible topologies. It needn't be this file though. a dedicated one might make more sense.

pow2clk (commented):
I think some additional testing would be useful

pow2clk (approved):
Looks good! thanks for the quick turnaround!

PR #4495, opened 2022-06-01
Validate no gaps in UAV store writemask

### LINKED ISSUE #4161: Missing validation for UAV store writemasks that must not contain holes
It seems that DXIL inherited the D3D11 requirements around UAV raw/structured stores, both of which have this text in the D3D11 functional spec:
```
                dst0 can only have a write mask that is one of the
                following: .x, .xy, .xyz, .xyzw.   The writemask determines
                the number of 32bit components to write – without gaps.
```

Attempting to compile HLSL that doesn't follow this restriction will produce multiple stores in the resulting DXIL. Attempting to run DXIL that violates this rule doesn't work correctly. It seems this should be added to DXIL validation for future validator versions.

gracejennings:
This change adds a validation rule for the inherited D3D11 requirement that a UAV store write mask must not contain gaps and therefore be one of the following: .x, .xy, .xyz, .xyzw.

Fixes #4161

### DIFF
--- a/tools/clang/unittests/HLSL/ValidationTest.cpp
+++ b/tools/clang/unittests/HLSL/ValidationTest.cpp
@@ -1199,11 +1199,13 @@ TEST_F(ValidationTest, UAVStoreMaskMatch) {
 }
 
 TEST_F(ValidationTest, UAVStoreMaskGap) {
+  if (m_ver.SkipDxilVersion(1, 6))
+    return;
   RewriteAssemblyCheckMsg(
       L"..\\CodeGenHLSL\\uav_store.hlsl", "ps_6_0",
       "i32 2, i32 2, i32 2, i32 2, i8 15)",
       "i32 undef, i32 2, i32 undef, i32 2, i8 10)",
-      "uav write mask must be continuous and not contain gaps.");
+      "UAV write mask must be contiguous and without gaps: .x, .xy, .xyz, or .xyzw.");
 }
 
 TEST_F(ValidationTest, Recursive) {
--- a/tools/clang/unittests/HLSL/ValidationTest.cpp
+++ b/tools/clang/unittests/HLSL/ValidationTest.cpp
@@ -72,6 +72,7 @@ public:
   TEST_METHOD(TypedUAVStoreFullMask0)
   TEST_METHOD(TypedUAVStoreFullMask1)
   TEST_METHOD(UAVStoreMaskMatch)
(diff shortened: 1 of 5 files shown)


tex3d (approved):
LGTM!

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1197,6 +1198,16 @@ TEST_F(ValidationTest, UAVStoreMaskMatch) {
...
 
+TEST_F(ValidationTest, UAVStoreMaskGap) {
+  if (m_ver.SkipDxilVersion(1, 6))
+    return;
+  RewriteAssemblyCheckMsg(
+      L"..\\CodeGenHLSL\\uav_store.hlsl", "ps_6_0",
+      "i32 2, i32 2, i32 2, i32 2, i8 15)",
+      "i32 undef, i32 2, i32 undef, i32 2, i8 10)",
+      "UAV write mask must be contiguous and without gaps: .x, .xy, .xyz, or .xyzw.");
+}
```
pow2clk, line 1209: I don't think we need to test all 11 invalid combinations, but maybe a couple more such as all set save x (0xE) and only W set (0x8)

utils/hct/hctdb.py
```diff
@@ -2582,6 +2582,7 @@ def build_valrules(self):
         self.add_valrule("Instr.BarrierModeNoMemory", "sync must include some form of memory barrier - _u (UAV) and/or _g (Thread Group Shared Memory).  Only _t (thread group sync) is optional.")
         self.add_valrule("Instr.BarrierModeForNonCS", "sync in a non-Compute/Amplification/Mesh Shader must only sync UAV (sync_uglobal).")
         self.add_valrule("Instr.WriteMaskForTypedUAVStore", "store on typed uav must write to all four components of the UAV.")
+        self.add_valrule("Instr.WriteMaskGapForUAV", "UAV write mask must be contiguous and without gaps: .x, .xy, .xyz, or .xyzw.")
```
pow2clk, line 2585: I suppose another characteristic that is checked here is that it starts at x. The explicit examples make that clear though.

pow2clk (commented):
I'm ambivalent about adding more detail to the validation message. I think with the examples, it's clear what's required, but I would welcome a few more tests just in case

pow2clk (approved):

PR #4505, opened 2022-06-06
Isnormal hlktest

bob80905:
This change adds an hlktest to test the isSpecialFloat dxil call. This test facilitated by the implementation of a new feature for the shader op test framework: shader call back functions. This specific test was implemented by taking the dxil generated from an isnan call in the source, and replacing it with the "isSpecialFloat" dxil call using various disassembly replacement functions.

### DIFF
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -24,6 +24,7 @@
 #include <iomanip>
 #include "dxc/Test/CompilationResult.h"
 #include "dxc/Test/HLSLTestData.h"
+#include "dxc/DxilContainer/DxilContainer.h"
 #include <Shlwapi.h>
 #include <atlcoll.h>
 #include <locale>
@@ -3163,10 +3164,12 @@ struct SPrimitives {
   float f_float2_o;
 };
 
+
 std::shared_ptr<ShaderOpTestResult>
 RunShaderOpTestAfterParse(ID3D12Device *pDevice, dxc::DxcDllSupport &support,
                           LPCSTR pName,
                           st::ShaderOpTest::TInitCallbackFn pInitCallback,
+                          st::ShaderOpTest::TShaderCallbackFn pShaderCallback,
                           std::shared_ptr<st::ShaderOpSet> ShaderOpSet) {
   st::ShaderOp *pShaderOp;
   if (pName == nullptr) {
@@ -3198,6 +3201,7 @@ RunShaderOpTestAfterParse(ID3D12Device *pDevice, dxc::DxcDllSupport &support,
   std::shared_ptr<st::ShaderOpTest> test = std::make_shared<st::ShaderOpTest>();
   test->SetDxcSupport(&support);
   test->SetInitCallback(pInitCallback);
+  test->SetShaderCallback(pShaderCallback);
   test->SetDevice(pDevice);
   test->RunShaderOp(pShaderOp);
 
@@ -3207,6 +3211,14 @@ RunShaderOpTestAfterParse(ID3D12Device *pDevice, dxc::DxcDllSupport &support,
   result->Test = test;
   result->ShaderOp = pShaderOp;
   return result;
+}
+
+std::shared_ptr<ShaderOpTestResult>
+RunShaderOpTestAfterParse(ID3D12Device *pDevice, dxc::DxcDllSupport &support,
+                          LPCSTR pName,
+                          st::ShaderOpTest::TInitCallbackFn pInitCallback,
+                          std::shared_ptr<st::ShaderOpSet> ShaderOpSet) {
+return RunShaderOpTestAfterParse(pDevice, support, pName, pInitCallback, nullptr, ShaderOpSet);
 }
 
 std::shared_ptr<ShaderOpTestResult>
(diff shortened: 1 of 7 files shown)


tools/clang/unittests/HLSLTestLib/DxcTestUtils.cpp
```diff
@@ -195,6 +195,77 @@ void AssembleToContainer(dxc::DxcDllSupport &dllSupport, IDxcBlob *pModule,
...
+                (unsigned)disassembly.size(), disassembly.data()));
+          }
+          VERIFY_IS_TRUE(found);
+        }
+      }
+    }
+  }
+}
+
+void ReplaceDisassemblyText(llvm::ArrayRef<LPCSTR> pLookFors,
```
tex3d, line 260: This overload is odd.  It's only used by one caller.  Can we just move the blob creation into the caller and get rid of this one?  In fact, the caller itself (`CompileAndRewriteAssemblyToText`) only has one caller, so that call could be replaced with the compile, disassemble, ReplaceText, and a CreateBlobFromPinned, and the `CompileAndRewriteAssemblyToText` function can also be eliminated.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -11330,6 +11343,124 @@ TEST_F(ExecutionTest, QuadAnyAll) {
...
+    VERIFY_SUCCEEDED(m_support.CreateInstance(CLSID_DxcUtils, &pUtils));
+    // Compile original HLSL with op to replace, and disassemble.
+    CComPtr<IDxcBlob> compiledShader;
+    std::vector<LPCWSTR> args; // Add args here
+    VerifyCompileOK(m_support, pText, L"cs_6_0", args, &compiledShader);
+    std::string disassembly = DisassembleProgram(m_support, compiledShader);
+    // Replace op
+    ReplaceText(
+      { "@dx.op.isSpecialFloat.f32(i32 8," },
+      { "@dx.op.isSpecialFloat.f32(i32 11," },
```
tex3d, line 11398: This function only replaces the f32 operation.  Double will require another replacement, like so:
```
      { "@dx.op.isSpecialFloat.f32(i32 8,",  "@dx.op.isSpecialFloat.f64(i32 8," },
      { "@dx.op.isSpecialFloat.f32(i32 11,", "@dx.op.isSpecialFloat.f64(i32 11," },
```

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -11330,6 +11343,124 @@ TEST_F(ExecutionTest, QuadAnyAll) {
...
+      pDevice->CheckFeatureSupport((D3D12_FEATURE)D3D12_FEATURE_D3D12_OPTIONS,
+                                    &devOptions, sizeof(devOptions)));
+  if (devOptions.ResourceBindingTier < D3D12_RESOURCE_BINDING_TIER_3) {
+    WEX::Logging::Log::Comment(
+        L"Device does not support Resource Binding Tier 3");
+    WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+    return;
+  }
+
+  auto ShaderInitFn = [&](LPCSTR Name, LPCSTR pText, IDxcBlob **ppShaderBlob, st::ShaderOp *pShaderOp) {
```
tex3d, line 11385: This function definition could be more generally useful if wrapped in a function that takes arguments for compilation and replacements, like so:
```
// Copies input strings to local storage, so it doesn't rely on lifetime of input string pointers.
st::ShaderOpTest::TShaderCallbackFn MakeShaderReplacementCallback(
    llvm::ArrayRef<LPCWSTR> dxcArgs,
    llvm::ArrayRef<LPCSTR> LookFors,
    llvm::ArrayRef<LPCSTR> Replacements,
    dxc::DxcDllSupport &dllSupport) {
  // place ArrayRef arguments in std::vector locals, and copy them by value into the callback function lambda
  std::vector<std::wstring> ArgsStorage(dxcArgs.size());
    for (unsigned i = 0; i < dxcArgs.size(); ++i)
      ArgsStorage[i] = dxcArgs[i];
  std::vector<std::string> LookForsStorage(LookFors.size());
    for (unsigned i = 0; i < LookFors.size(); ++i)
      LookForsStorage[i] = LookFors[i];
  std::vector<std::string> ReplacementsStorage(Replacements.size());
    for (unsigned i = 0; i < Replacements.size(); ++i)
      ReplacementsStorage[i] = Replacements[i];
  auto ShaderInitFn = 
      [ArgsStorage, LookForsStorage, ReplacementsStorage, &dllSupport]
      (LPCSTR Name, LPCSTR pText, IDxcBlob **ppShaderBlob, st::ShaderOp *pShaderOp) {
    // Create pointer vectors from local storage to supply API needs
    std::vector<LPCWSTR> Args(ArgsStorage.size());
    for (unsigned i = 0; i < ArgsStorage.size(); ++i)
      args[i] = ArgsStorage[i].c_str();
    std::vector<LPCSTR> LookFors(LookForsStorage.size());
    for (unsigned i = 0; i < LookForsStorage.size(); ++i)
      LookFors[i] = LookForsStorage[i].c_str();
    std::vector<LPCSTR> Replacements(ReplacementsStorage.size());
    for (unsigned i = 0; i < ReplacementsStorage.size(); ++i)
      Replacements[i] = ReplacementsStorage[i].c_str();
    // Do the rest of the work for compilation, replacement, and root sig copying, if present
  }
  return ShaderInitFn;
}
```

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -11330,6 +11343,124 @@ TEST_F(ExecutionTest, QuadAnyAll) {
...
+
+  std::shared_ptr<st::ShaderOpSet> ShaderOpSet =
+      std::make_shared<st::ShaderOpSet>();
+  st::ParseShaderOpSetFromStream(pStream, ShaderOpSet.get());
+  st::ShaderOp *pShaderOp =
+      ShaderOpSet->GetShaderOp("IsNormal");
+  vector<st::ShaderOpRootValue> fallbackRootValues = pShaderOp->RootValues;
+
+  const int expectedResultsSize = 12;
+  // false represents 0, true represents a non-zero value
```
tex3d, line 11360: What's this comment for?  It doesn't seem to clarify anything.  Perhaps a comment like: "Result of IsNormal operation"?

`isnan`, and thus the `isnormal` replacement, returns bool, which translates to 0 or 1 when converted to int or uint.  So it doesn't make a difference in the form `isnan(...) ? 1 : 0`.  The ternary operation doesn't do anything extra.

There isn't any guide to see what each expected result maps to for the input of the operation here either.  That would actually be helpful (or if the inputs were defined here alongside the outputs).

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -5798,6 +5811,16 @@ static TableParameter PackUnpackOpParameters[] = {
     { L"Validation.Input", TableParameter::UINT32_TABLE, true },
 };
 
+static TableParameter IsNormalParameters[] = {
```
tex3d, line 5814: This is unused.  I imagine copied/adapted from the UnaryFloatOpTest.  This test isn't data driven like UnaryFloatOpTest is.

tex3d (changes requested):

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -11330,6 +11353,202 @@ TEST_F(ExecutionTest, QuadAnyAll) {
...
+
+    CComPtr<IDxcContainerBuilder> pBuilder;
+    VERIFY_SUCCEEDED(dllSupport.CreateInstance(CLSID_DxcContainerBuilder, &pBuilder));
+    VERIFY_SUCCEEDED(pBuilder->Load(assembledShader));
+
+    // If the Root Signautre is defined as a node in the XML, then 
+    // the Root Signature will be retained by ShaderOpTest, and won't be 
+    // permanently lost when assembling the container
+    // Otherwise, it's necessary to add the root signature back into the assembled dxil container
+    if (!bRootSigInXML)
```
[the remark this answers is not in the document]
bob80905: For your last point, I don't think I can move the creation of DxcContainerBuilder. The pbuilder object is used outside of the "if (!bRootSigInXML)" branch, and it's also used at the bottom inside the branch.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -11425,7 +11415,11 @@ st::ShaderOpTest::TShaderCallbackFn MakeShaderReplacementCallback(
       VERIFY_SUCCEEDED(hlsl::IsValidDxilContainer(pContainerHeader, compiledShader->GetBufferSize()));
       hlsl::DxilPartHeader *pPartHeader = hlsl::GetDxilPartByType(
           pContainerHeader, hlsl::DxilFourCC::DFCC_RootSignature);
-      VERIFY_IS_NOT_NULL(pPartHeader);
+      if (!pPartHeader) {
```
tex3d, line 11414: I meant that this condition can be used instead of `if (!bRootSigInXML)`, so that condition goes away, and the pBuilder doesn't need to be used, so that code can be moved after this branch.

Like, after removing the `if (!bRootSigInXML)`, the braces surrounding this code, and moving builder init, you basically have:
```
// Find root signature part in container
...
if (!pPartHeader) {
  // No root signature to copy, use the assembledShader.
  *ppShaderBlob = assembledShader.Detach();
  return;
}

CComPtr<IDxcContainerBuilder> pBuilder;
VERIFY_SUCCEEDED(dllSupport.CreateInstance(CLSID_DxcContainerBuilder, &pBuilder));
VERIFY_SUCCEEDED(pBuilder->Load(assembledShader));

// Wrap root signature in blob
...
```

Then there's no need for the bRootSigInXML parameter.  It just copies the root sig if there is one.

tex3d (approved):
LGTM.

Issue #4506, opened 2022-06-08
[SPIR-V] Multidimensional array of constant buffers crashes dxc


kuhar:
```shell
$ bin/dxc -T ps_6_0 -E psmain ./test.hlsl -spirv 
dxc: /usr/local/google/home/kubak/dxc/DirectXShaderCompiler/include/llvm/Support/Casting.h:96: static bool llvm::isa_impl_cl<clang::FieldDecl, const clang::Decl *>::doit(const From *) [To = clang::FieldDecl, From = const clang::Decl *]: Assertion `Val && "isa<> used on a null pointer"' failed.
[1]    318156 IOT instruction  bin/dxc -T ps_6_0 -E psmain ./test5.hlsl -spirv
```

Repro:
```hsls
struct Foo {
  float4 a;
  int2 b;
};
ConstantBuffer<Foo> myCB2[2][3]: register(b0, space1);

struct VSOutput {
  float2 TexCoord : TEXCOORD;
};

float4 psmain(VSOutput input) : SV_TARGET {
  return 0;
}
```

This is based on the example in https://docs.microsoft.com/en-us/windows/win32/direct3d12/resource-binding-in-hlsl#constant-buffers.

kuhar:
Fixed by https://github.com/microsoft/DirectXShaderCompiler/pull/4507

PR #4507, opened 2022-06-10
Added handling for multi-dimensional CT buffers

miseydl:
Previously the usage of multi dimensional ConstantBuffers would have just crashed the SPIRV generation due to a flawed assumption. This implementation is unrolling the array and is emitting the correct SPIRV. 

Remark: When generating SPIRV for any Vulkan target environment the validation will fail due to a violation of Vulkan spec section 14.5.2, which requires uniform buffers to be of type struct or "array of struct". Apparently "array of array of struct" is upsetting the validator.


kuhar (commented):
Thanks for working on this!

Could you add a test (under `tools/clang/test/CodeGenSPIRV`)? I recently opened an issue with a minimized repro that could serve as a test case: https://github.com/microsoft/DirectXShaderCompiler/issues/4506

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -600,6 +600,52 @@ bool containOnlyVecWithFourFloats(QualType type, bool use16Bit) {
   return false;
 }
 
+// Evaluates whether a given QualType is in fact an array and unroll accordingly.
+// Will return the the appropriate RecordType for the provided QualType. Will be 
+// the actual type of the array element, if it's really an array.
+template <size_t SmallVectorSize>
+QualType unrollMultiDimensionalArray(
+    const ASTContext &astContext, const QualType &startType,
+    llvm::SmallVector<int, SmallVectorSize> *arraySizes
```
kuhar, line 609: Use `llvm::SmallVectorImpl<int>&`, see: https://llvm.org/docs/ProgrammersManual.html#:~:text=Prefer%20to%20use%20ArrayRef%3CT%3E%20or%20SmallVectorImpl%3CT%3E%20as%20a%20parameter%20type.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -600,6 +600,52 @@ bool containOnlyVecWithFourFloats(QualType type, bool use16Bit) {
...
+) {
```
kuhar, line 610: Can you please format this whole PR with clang-format? You can do that with something like `git diff main -U0 | clang-format-diff.py -i -p1`. See https://clang.llvm.org/docs/ClangFormat.html#script-for-patch-reformatting

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -600,6 +600,52 @@ bool containOnlyVecWithFourFloats(QualType type, bool use16Bit) {
...
+
+  QualType innerQualType = startType;
+
+  // Unroll a multidimensional array.
+  if (const auto *arrayType = startType->getAsArrayTypeUnsafe()) {
```
kuhar, line 615: Use an early exit: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -600,6 +600,52 @@ bool containOnlyVecWithFourFloats(QualType type, bool use16Bit) {
   return false;
 }
 
+// Evaluates whether a given QualType is in fact an array and unroll accordingly.
+// Will return the the appropriate RecordType for the provided QualType. Will be
```
kuhar, line 604: nit: Will return -> Returns

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -600,6 +600,52 @@ bool containOnlyVecWithFourFloats(QualType type, bool use16Bit) {
   return false;
 }
 
+// Evaluates whether a given QualType is in fact an array and unroll accordingly.
+// Will return the the appropriate RecordType for the provided QualType. Will be 
+// the actual type of the array element, if it's really an array.
```
kuhar, line 605: Can you briefly mention that arraySizes is an output parameter and explain the order of array dimensions?

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1132,7 +1178,8 @@ DeclResultIdMapper::createOrUpdateStringVar(const VarDecl *var) {
 }
 
 SpirvVariable *DeclResultIdMapper::createStructOrStructArrayVarOfExplicitLayout(
-    const DeclContext *decl, int arraySize, const ContextUsageKind usageKind,
+    const DeclContext *decl, const llvm::ArrayRef<int> &arraySize,
```
kuhar, line 1181: Pass `arraySize` by value, this is cheap to copy (ptr and size only)

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1223,6 +1271,14 @@ SpirvVariable *DeclResultIdMapper::createStructOrStructArrayVarOfExplicitLayout(
   return var;
 }
 
+SpirvVariable *DeclResultIdMapper::createStructOrStructArrayVarOfExplicitLayout(
+    const DeclContext *decl, int arraySize, const ContextUsageKind usageKind,
+    llvm::StringRef typeName, llvm::StringRef varName) {
+  return createStructOrStructArrayVarOfExplicitLayout(
+      decl, arraySize == 0 ? llvm::SmallVector<int, 2>{} : llvm::SmallVector<int, 2>{arraySize},
```
kuhar, line 1278: I'd just create a local variable for this array. Also, since you have 0 or 1 elements, small size of 1 would be enough.
```c++
llvm::SmallVector<int, 1> arraySizes;
if (arraySize > 0)
  arraySizes.push_back(arraySize);
return createStructOrStructArrayVarOfExplicitLayout(decl, arraySizes, usageKind, typeName, varName);
```

tools/clang/test/CodeGenSPIRV/type.constant-buffer.multiple-dimensions.hlsl
```diff
@@ -0,0 +1,18 @@
...
+ConstantBuffer<Foo> myCB2[2][3] : register(b0, space1);
+
+struct VSOutput {
+  float2 TexCoord : TEXCOORD;
+};
+
+float4 main(VSOutput input) : SV_TARGET {
+  // CHECK: OpAccessChain %_ptr_Uniform_v4float %myCB2 %int_1 %int_0 %int_0
+  return float4(1.0, 1.0, 1.0, 1.0) * myCB2[1][0].a;
+}
```
kuhar, line 18: nit: add a new line at the end

tools/clang/test/CodeGenSPIRV/type.constant-buffer.multiple-dimensions.hlsl
```diff
@@ -0,0 +1,18 @@
+// RUN: %dxc -T ps_6_0 -E main -fspv-target-env=universal1.5
+
+struct Foo {
+  float4 a;
+  int2 b;
+};
+
+// CHECK: %_ptr_Uniform__arr__arr_type_ConstantBuffer_Foo_uint_3_uint_2 = OpTypePointer Uniform %_arr__arr_type_ConstantBuffer_Foo_uint_3_uint_2
```
kuhar, line 8: Relying on exact names may be quite brittle, I don't think we guarantee the naming scheme will remain as-is. I think it would be more robust to match each Op that creates this final type step by step.

tools/clang/test/CodeGenSPIRV/type.constant-buffer.multiple-dimensions.hlsl
```diff
@@ -0,0 +1,18 @@
...
+ConstantBuffer<Foo> myCB2[2][3] : register(b0, space1);
+
+struct VSOutput {
+  float2 TexCoord : TEXCOORD;
+};
+
+float4 main(VSOutput input) : SV_TARGET {
+  // CHECK: OpAccessChain %_ptr_Uniform_v4float %myCB2 %int_1 %int_0 %int_0
```
kuhar, line 16: similar for constants

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
kuhar: It was updated to python3 over a year ago, so it should be in llvm 13 and newer: https://github.com/llvm/llvm-project/blame/5179f885d159a01e00c208e3255052fe0fde06bf/clang/tools/clang-format/clang-format-diff.py#L1

kuhar (approved):
LGTM. Thanks for the changes.

One tiny nit: could you prepend `[spirv]` to the commit / PR title?

PR #4508, opened 2022-06-13
Add support for separate ARM64 and ARM64EC testing

hekota:
If tests are built as ARM64X, the test dll contains two versions of each test - the native ARM64 code and the emulation-ready x64 code (== ARM64EC). By default, TAEF will run both versions of the tests. The tests have the same name, so if one is failing, it is not clear if it is the ARM64 or ARM64EC version. 

It is preferable to run the tests for each platform separately. This PR adds @Architecture parameter to the TAEF tests invocation, which enables exactly that. The hcttest.cmd script has a new argument -arm64ec for running ARM64EC-only tests, and similarly the -arm64 will run only the native ARM64 tests.

Also includes couple of fixes in hctstart.cmd.

### DIFF
--- a/tools/opt/CMakeLists.txt
+++ b/tools/opt/CMakeLists.txt
@@ -39,12 +39,13 @@ add_llvm_tool(opt
 )
 
 if(WIN32 AND HLSL_BUILD_DXILCONV)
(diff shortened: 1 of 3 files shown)


tex3d (approved):
I think this looks good.

PR #4509, opened 2022-06-14
Enable ChangeStdin/outToBinary for llvm-dis on Windows.

python3kgae:
This will fix issue when read input from stdin.
Without ChangeStdinToBinary, llvm-dis will get wrong size for stdin input in some case. That will cause 'Invalid bitcode size' error when parse bitcode.


tex3d (approved):
LGTM.

While this was added a long time ago during initial file system virtualization work, I don't think the API should hit these functions under virtualized paths (and testing so far confirms this).  We should keep an eye out still for a regression where the global stdin or stdout mode is unexpectedly overridden when using the API.

PR #4515, opened 2022-06-16
[spirv] Translate HLSL round to RoundEven

### LINKED ISSUE #4368: [SPIR-V] HLSL round intrinsic is translated to "Round" instead of to "RoundEven" which is incorrect based on the HLSL docs (works with the DXIL backend and FXC)
### Title
[SPIR-V] HLSL round intrinsic is translated to "Round" instead of to "RoundEven" which is incorrect based on the HLSL docs (works with the DXIL backend and FXC)

### Functional impact
Functional parity when using the HLSL `round` intrinsic is lost between the DXIL and SPIR-V backends for floating-point arguments exactly in between integers.

Based on the [HLSL docs for round](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-round): `Halfway cases are rounded to the nearest even.`
and indeed FXC and the DXIL backend generate [round_ne](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/round-ne--sm4---asm-) (which means "round to nearest even")

### Minimal repro steps

1. Compile the following shader with `dxc.exe -T ps_6_0 -E PSMain -spirv repro_shader.txt`:
```
float4 PSMain(float4 color : COLOR) : SV_TARGET
{
	return round(color);
}
```

### Expected result
The generated SPIR-V should use the `RoundEven` instruction from the `GLSL.std.450` extended instruction set

### Actual result
The generated SPIR-V generates the following instructions:
```
 %1 = OpExtInstImport "GLSL.std.450"
...
%13 = OpExtInst %v4float %1 Round %12
```

### Further technical details
Used DXC binary from the latest official release [v1.6.2112](https://github.com/microsoft/DirectXShaderCompiler/releases/tag/v1.6.2112)

sudonatalie:
The specification for the HLSL round intrinsic function states "Rounds
the specified value to the nearest integer. Halfway cases are rounded to
the nearest even.", so translate it to RoundEven SPIR-V extended
instruction rather than Round.

Fixes #4368

### DIFF
--- a/docs/SPIR-V.rst
+++ b/docs/SPIR-V.rst
@@ -2364,7 +2364,7 @@ HLSL Intrinsic Function   GLSL Extended Instruction
 ``pow``                 ``Pow``
 ``reflect``             ``Reflect``
 ``refract``             ``Refract``
-``round``               ``Round``
+``round``               ``RoundEven``
 ``rsqrt``               ``InverseSqrt``
 ``saturate``            ``FClamp``
 ``sign``                ``SSign``/``FSign``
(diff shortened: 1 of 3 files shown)


kuhar (approved):
LGTM

PR #4516, opened 2022-06-17
Fix several bugs in code

gongminmin:
1. Wrong usage of memset
2. Member initializer list in order of declaration
3. Add explicit to MSFileSystemHandle's single parameter constructors

### DIFF
--- a/lib/Support/MSFileSystemBasic.cpp
+++ b/lib/Support/MSFileSystemBasic.cpp
@@ -107,10 +107,10 @@ struct MSFileSystemHandle {
       stream; // For a file or console file handle, the stream interface.
   int fd;     // For a file handle, its file descriptor.
 
-  MSFileSystemHandle(int knownFD)
+  explicit MSFileSystemHandle(int knownFD)
       : kind(MSFileSystemHandleKind_FileHandle), fd(knownFD) {}
 
-  MSFileSystemHandle(IUnknown *pMapping)
+  explicit MSFileSystemHandle(IUnknown *pMapping)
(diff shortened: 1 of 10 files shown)


python3kgae (approved):

lib/Support/MSFileSystemBasic.cpp
```diff
@@ -125,7 +125,7 @@ struct MSFileSystemHandle
   {
   }
 
-  MSFileSystemHandle(IEnumSTATSTG *pEnumSTATG)
+  explicit MSFileSystemHandle(IEnumSTATSTG *pEnumSTATG)
```
pow2clk, line 128: This isn't mentioned in the commit message. Is it strictly necessary? Additionally, I don't see any instance of a pointer being assigned false, which was mentioned in the commit message

pow2clk (commented):
Thanks for this! Sorry it fell by the wayside

lib/Support/MSFileSystemBasic.cpp
gongminmin: Just don't want surprisingly construct a MSFileSystemHandle object from a IEnumSTATSTG pointer. Using 'explicit' on any single parameter constructor is a good convention. If implicit construct is intentional, it needs to be commented.

gongminmin: The commit rebased and update the message.

The pointer = false has been fixed some time ago.

pow2clk (approved):
Looks good

PR #4517, opened 2022-06-17
[spirv] Translate HLSL min/max to NMin/NMax

### LINKED ISSUE #3221: [SPIR-V] FMin/FMax/FClamp NaN handling contradicts HLSL specification
Currently the SPIR-V emitter generates FMin, FMax and FClamp instructions for min, max, clamp and saturate. The [specification of the GLSL.std.450 instruction set](https://www.khronos.org/registry/spir-v/specs/1.0/GLSL.std.450.html) states that "which operand is the result is undefined if one of the operands is a NaN". This is in contrast with the [DXIL](https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/DXIL.rst#fmax), [DXBC](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/max--sm4---asm-) and [HLSL](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-max) specification of those instructions, which, if one of the operands is NaN, give the other operand as the result (and clamp/saturate gives the lower bound, as if max was performed first).

While HLSL is normally optimized as if there were no special floating-point values at all, with the `precise` qualifier or the "force IEEE strictness" compiler option, the compiler must honor the defined behavior for specials. Clamping operations may also be used with the implication of sanitizing the input, making sure the resulting value is strictly in the bounds and not NaN.

The correct implementation, which maps 1:1 to the DXIL/DXBC behavior, is NMax/NMin/NClamp from the GLSL.std.450 instruction set. Are those safe to use in DXC (such as, can we assume that NClamp to 0 and 1, on existing drivers or in the near future, will surely map to the hardware saturation modifier in all the same cases as when FClamp to 0 and 1 is used; and whether they're compatible with wave reduction)?

sudonatalie:
The specification for the HLSL min and max intrinsic functions states
that if one of the values is NaN, the other will be given as the result,
which is correctly represented by the GLSL.std.450 instructions NMin and
NMax, respectively. By the semantics of the previously used FMin and
FMax instructions, this would be undefined behavior.

Fixes #3221

### DIFF
--- a/docs/SPIR-V.rst
+++ b/docs/SPIR-V.rst
@@ -2357,8 +2357,8 @@ HLSL Intrinsic Function   GLSL Extended Instruction
 ``log10``               ``Log2`` (scaled by ``1/log2(10)``)
 ``log2``                ``Log2``
 ``mad``                 ``Fma``
-``max``                 ``SMax``/``UMax``/``FMax``
-``min``                 ``SMin``/``UMin``/``FMin``
+``max``                 ``SMax``/``UMax``/``NMax``
+``min``                 ``SMin``/``UMin``/``NMin``
 ``modf``                ``ModfStruct``
 ``normalize``           ``Normalize``
 ``pow``                 ``Pow``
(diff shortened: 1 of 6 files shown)


kuhar (approved):
LGTM

Issue #4496, opened 2022-06-02
[SPIR-V] Compile and generate reflection information for shader in a single command.


wookie41:
Hello.

I've been trying to get `dxc` to compile to SPIRV, stripping reflection infro from the shader and generate reflection info for SPIRV-reflect using a single command like this:

```
dxc -Qstrip_reflect -spirv -fspv-reflect -fspv-target-env=vulkan1.3 -HV 2021 -T ps_6_7 -Fre reflect/base.frag.ref -Fo bin/base.frag.spv base.frag.hlsl
```

But instead of two files - one `.ref` and `.spv`, I only get the `.spv` one, with (what it looks like) reflection info in it, as SPIRV-Reflect is able to read the reflection info from it.
I've been trying to do it according to the description in [this issue](https://github.com/microsoft/DirectXShaderCompiler/issues/2749), but wasn't able to and this issue was closed, as the OP was using the SDK and I have to use the commandline.

Any suggestions?

sudonatalie:
The `-Qstrip_reflect` and `-Fre` options are not currently supported by the SPIR-V backend and we don't have any current plans to implement them, although PR contributions are welcome and will be reviewed.

As a workaround, you can follow up your compilation with `-fspv-reflect` by running `spirv-opt --strip-nonsemantic` on the results to output a SPIR-V module with all reflection information removed. `spirv-opt` is distributed as part of the Vulkan SDK and at [KhronosGroup/SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools).

s-perron:
Considering this complete because there is a working solution using spirv-opt as a post processor to remove the reflection information. We do not intend to support the other options.

Issue #3111, opened 2020-09-03
Looks like -Fd doesn't work with -spirv


siliconvoodoo:
I get "Unable to find required part in blob" if I attempt the acrobat move of building with -Zi -Fd -spirv

siliconvoodoo:
I agree with your questioning. I was imagining you'd know better than me, but from a very rough mind image I was expecting there may be some sort of DWARF-like format (or whatever) that RenderDoc supports. I guess in the end it's a matter of what debugging tools want as inputs. This could be a case for a discussion with Pix developers. In the meantime you could just implement a hard clear error "not supported"?

PR #4518, opened 2022-06-17
[spirv] Fail when unsupported options are used

sudonatalie:
Some DXC options are not supported in combination with the SPIR-V
backend. Rather than silently ignore them, we should emit a clear error
message. This change adds appropriate error messages for a couple
options that have been mentioned in issues #3111 and #4496, but more
should be added in future.

### DIFF
--- a/lib/DxcSupport/HLSLOptions.cpp
+++ b/lib/DxcSupport/HLSLOptions.cpp
@@ -322,6 +322,18 @@ static bool handleVkShiftArgs(const InputArgList &args, OptSpecifier id,
   return true;
 }
 
+// Check if the given unsupported option is used with SPIR-V.
+static bool checkUnsupportedSpirvOption(const InputArgList &args,
+                                        OptSpecifier id, const char *name,
+                                        llvm::raw_ostream &errors) {
+  if (args.hasFlag(OPT_spirv, OPT_INVALID, false) &&
+      !args.getLastArgValue(id).empty()) {
+    errors << "-" << name << " is not supported with -spirv";
+    return true;
+  }
+  return false;
+}
+
 namespace {
 
 /// Maximum size of OpString instruction minus two operands
@@ -329,7 +341,7 @@ static const uint32_t kDefaultMaximumSourceLength = 0xFFFDu;
 static const uint32_t kTestingMaximumSourceLength = 13u;
 
 }
-#endif
+#endif // ENABLE_SPIRV_CODEGEN
 // SPIRV Change Ends
 
 namespace hlsl {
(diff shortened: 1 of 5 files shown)


lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -1071,6 +1071,21 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
   opts.SpirvOptions.entrypointName =
       Args.getLastArgValue(OPT_fspv_entrypoint_name_EQ);
 
+  // Options not currently implemented in the SPIR-V backend.
+  // The options checked here are non-exhaustive. A thorough audit of available
+  // options and their current compatibility is needed to generate a complete
+  // list.
+  if (Args.hasFlag(OPT_spirv, OPT_INVALID, false)) {
```
kuhar, line 1078: This function is very long. Could we move these checks to a new helper function?

tools/clang/test/CodeGenSPIRV/spirv.opt.fd.hlsl
```diff
@@ -0,0 +1,5 @@
+// RUN: %dxc -T ps_6_0 -E main -spirv -Fd file.ext
+
+void main() {}
+
+// CHECK: -Fd is not currently supported with -spirv
```
kuhar, line 5: I'm not sure if we should include 'currently'. I'd read this as an incomplete feature that will be or may be added in the future. What about saying that something is just not supported?

lib/DxcSupport/HLSLOptions.cpp
sudonatalie: Done

tools/clang/test/CodeGenSPIRV/spirv.opt.fd.hlsl
sudonatalie: Done

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -322,14 +322,26 @@ static bool handleVkShiftArgs(const InputArgList &args, OptSpecifier id,
   return true;
 }
 
+// Check if the given unsupported option is used with SPIR-V.
+static bool checkUnsupportedSpirvOption(const InputArgList &args,
```
kuhar, line 326: Maybe: `hasUnsupportedSpirvOption` or `containsUnsuportedSpirvOption`? The issue with `check` and `bool` as the return type is that it's not clear what the success/failure values are.

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -1071,20 +1083,13 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
...
-      errors << "-Fre is not currently supported with -spirv";
-      return 1;
-    }
-  }
+  // Check for use of options not implemented in the SPIR-V backend.
+  // Note: The options checked here are non-exhaustive. A thorough audit of
+  // available options and their current compatibility is needed to generate a
+  // complete list.
+  if (checkUnsupportedSpirvOption(Args, OPT_Fd, "Fd", errors) ||
+      checkUnsupportedSpirvOption(Args, OPT_Fre, "Fre", errors))
```
kuhar, line 1091: What do you think about passing a list of options to check, or just moving the list of unsupported options to the helper?

lib/DxcSupport/HLSLOptions.cpp
sudonatalie: Done

lib/DxcSupport/HLSLOptions.cpp
sudonatalie: I think this is tidier now, but let me know what you think!

sudonatalie: Also added one more unsupported arg that's just a bool flag to make sure it functions with both opt types.

kuhar (approved):

PR #4521, opened 2022-06-18
More conforming to C++ standards

gongminmin:
1. Can't mix char* and const char*
2. Remove redundant const and std::move
3. Use standard types and macros to replace platform dependent ones

### DIFF
--- a/tools/dxexp/dxexp.cpp
+++ b/tools/dxexp/dxexp.cpp
@@ -187,7 +187,7 @@ static const char *ShaderModelToStr(D3D_SHADER_MODEL SM) {
   }
 }
 
-static char *ViewInstancingTierToStr(D3D12_VIEW_INSTANCING_TIER Tier) {
+static const char *ViewInstancingTierToStr(D3D12_VIEW_INSTANCING_TIER Tier) {
   switch (Tier) {
   case D3D12_VIEW_INSTANCING_TIER_NOT_SUPPORTED:
     return "NO";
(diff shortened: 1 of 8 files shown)


llvm-beanz (approved):
These all look like reasonable changes to me.

Is there a set of compiler warnings you enabled when developing this patch?
Might be worth enabling extra warnings in the future to prevent regression.

python3kgae (approved):

pow2clk (commented):
Looks good. 

I don't see any example of some of the things mentioned in the commit message. 

2. Use this-> to represent member from parent template classes
I don't see anything similar to this.

3. Can't do double implicit conversion

Maybe not this either? There are some places that lose casting. I don't know if this is in reference to double type casting or two casts that form a (double)

Perhaps these comments represent an earlier version of the change?

gongminmin:
> Looks good.
> 
> I don't see any example of some of the things mentioned in the commit message.
> 
>     2. Use this-> to represent member from parent template classes
>        I don't see anything similar to this.
> 
>     3. Can't do double implicit conversion
> 
> 
> Maybe not this either? There are some places that lose casting. I don't know if this is in reference to double type casting or two casts that form a (double)
> 
> Perhaps these comments represent an earlier version of the change?

Right. 2) and 3) are already fixed in main branch in the last 15 months. I've updated the commit message and the PR description.

pow2clk (approved):
Thanks for updating the commit message!

tools/dxexp/dxexp.cpp
```diff
@@ -296,7 +296,7 @@ static HRESULT PrintAdapters() {
     hr = e.m_hr;
     json_printf("%c { \"err\": \"unable to print information for adapters - "
                 "0x%08x\" }\n",
-                comma, hr);
+                comma, (unsinged int)hr);
```
python3kgae, line 299: need to be unsigned

But doesn't have access to push fix:(

gongminmin: Fixed.

PR #4522, opened 2022-06-18
Fix version skip for UAVStoreMaskGap to exclude v1.6

tex3d:
SkipDxilVersion skips test if the either the DXIL version (DxCompiler) or
Validator version (DXIL.dll if external) is less than the specified version.
This message is being added in version 1.7, so the skip value needs to be
1.7 to skip unless it's version 1.7 or higher.

Without this fix, these tests fail when run with DXIL.dll version 1.6.


dmpots (approved):
LGTM

PR #4524, opened 2022-06-20
Add missing `_COM_Outptr_` SAL to `DxcCreateInstance` and friends

MarijnS95:
Ever since [windows-rs started requiring SAL] for properly processing functions that return a COM instance these annotations had to be added.  They seem to be present on most functions and were [worked around for `DxcCreateInstance(2)` in `win32`metadata] (yet their `*Proc` `typedef` variants were omitted).

[windows-rs started requiring SAL]: https://github.com/microsoft/windows-rs/pull/1088#issuecomment-907808506
[worked around for `DxcCreateInstance(2)` in `win32`metadata]: https://github.com/microsoft/win32metadata/pull/890/files

This includes more random functions that were spotted (particularly the `_Out_ IDxc*` pattern) but I have been unable to check the nullability for every variant. As per the documentation these `_COM_Outptr_` annotations already include the notion of leaving/setting this output pointer to `NULL` in an error case, in which case a non-`S_OK` `HRESULT` should also be returned.

### DIFF
--- a/include/dxc/DxilContainer/DxcContainerBuilder.h
+++ b/include/dxc/DxilContainer/DxcContainerBuilder.h
@@ -36,7 +36,7 @@ public:
   HRESULT STDMETHODCALLTYPE RemovePart(UINT32 fourCC) override;
   // Builds a container of the given container builder state
   HRESULT STDMETHODCALLTYPE
-  SerializeContainer(IDxcOperationResult **ppResult) override;
+  SerializeContainer(_COM_Outptr_ IDxcOperationResult **ppResult) override;
 
   DXC_MICROCOM_TM_ADDREF_RELEASE_IMPL()
   DXC_MICROCOM_TM_CTOR(DxcContainerBuilder)
--- a/include/dxc/dxcapi.internal.h
+++ b/include/dxc/dxcapi.internal.h
@@ -342,7 +342,7 @@ CROSS_PLATFORM_UUIDOF(IDxcContainerEventsHandler,
 struct IDxcContainerEventsHandler : public IUnknown {
 public:
   virtual HRESULT STDMETHODCALLTYPE
-  OnDxilContainerBuilt(IDxcBlob *pSource, IDxcBlob **ppTarget) = 0;
+  OnDxilContainerBuilt(IDxcBlob *pSource, _COM_Outptr_ IDxcBlob **ppTarget) = 0;
 };
 
 CROSS_PLATFORM_UUIDOF(IDxcContainerEvent,
--- a/include/dxc/dxcisense.h
+++ b/include/dxc/dxcisense.h
@@ -662,6 +662,12 @@ struct IDxcCodeCompleteResults;
 struct IDxcCompletionResult;
 struct IDxcCompletionString;
 
+// TODO for nullonfailure:
+// https://learn.microsoft.com/en-us/cpp/code-quality/annotating-function-parameters-and-return-values
+// > Certain interface conventions presume that output parameters are nullified
+// on failure. Except for explicitly COM code, the forms in the following table
+// are preferred. For COM code, use the corresponding COM forms that are listed
+// in the previous section.
 CROSS_PLATFORM_UUIDOF(IDxcCursor, "1467b985-288d-4d2a-80c1-ef89c42c40bc")
 struct IDxcCursor : public IUnknown {
   virtual HRESULT STDMETHODCALLTYPE
@@ -725,7 +731,7 @@ struct IDxcCursor : public IUnknown {
   /// cursor.</summary>
   virtual HRESULT STDMETHODCALLTYPE
   GetSnappedChild(_In_ IDxcSourceLocation *location,
-                  _Outptr_result_maybenull_ IDxcCursor **pResult) = 0;
+                  _COM_Outptr_result_maybenull_ IDxcCursor **pResult) = 0;
 };
 
 CROSS_PLATFORM_UUIDOF(IDxcDiagnostic, "4f76b234-3659-4d33-99b0-3b0db994b564")
@@ -736,6 +742,7 @@ struct IDxcDiagnostic : public IUnknown {
   virtual HRESULT STDMETHODCALLTYPE
   GetSeverity(_Out_ DxcDiagnosticSeverity *pResult) = 0;
   virtual HRESULT STDMETHODCALLTYPE
+  // COM_Outptr_?
   GetLocation(_Outptr_result_nullonfailure_ IDxcSourceLocation **pResult) = 0;
   virtual HRESULT STDMETHODCALLTYPE
   GetSpelling(_Outptr_result_maybenull_ LPSTR *pResult) = 0;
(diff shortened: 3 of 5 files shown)


damyanp:

MarijnS95:
@damyanp A contributor cannot reopen a PR after it has been closed by a collaborator or member. If I were to rebase and (force)push to the branch GitHub dissociates it from the PR, never allowing it to be reopened again.

Would you be able to help out reviewing and getting this PR merged if I rebase it? Is this repository even the right place (source of truth) for most of these headers before they ultimately end up in the Windows SDK (which is the ultimate goal to get picked up by `win32metadata` and `windows-rs`)?

damyanp:
I'll help shepherd this one through.  This repo is the source of truth for dxcapi.h and DxbcConverter.h, so if we get it merged they'll eventually flow through to where win32metadata / windows-rs will see it.

I think you're saying that you'll need to create a new PR if you rebase this?  That should be fine.

MarijnS95:
> I'll help shepherd this one through. This repo is the source of truth for dxcapi.h and DxbcConverter.h, so if we get it merged they'll eventually flow through to where win32metadata / windows-rs will see it.

Awesome, thanks!

> I think you're saying that you'll need to create a new PR if you rebase this? That should be fine.

I would be able to reuse this PR and "context" if you reopen it _before_ I rebase and force-push :)

MarijnS95:
I've very quickly gone through a first rebase to clean up conflicts. Lots of existing SAL seems to have been removed (either because it was redundant or not relevant in internal files). TODO:

- Need to figure out if some files should be excluded, i.e. `.cpp` files and "internal" headers (though those existing `_Out_` ones might still be relevant to replace);
- Difference between `_COM_Outptr_` vs `_Outptr_result_nullonfailure_`;
- Check internal implementations when `_opt` and `_result_maybenull` should be used.

See also https://learn.microsoft.com/en-us/cpp/code-quality/annotating-function-parameters-and-return-values.

damyanp:
Looks like some formatting stuff needed.  In the meantime, I was wondering what you'd think about limiting this change to just the header files that are included in the SDK, just to be super cautious of churn?  I think that would just be dxcapi.h?

damyanp:
Sorry, I didn't see your last comment when I posted my previous message where you'd already called out the point I was raising!

In places where the SAL has been removed we shouldn't add it back.  It should have been removed from all internal uses - only the external headers that get included in releases / the SDK should be annotated.

MarijnS95:
Cool, so `tools/clang/tools/dxcompiler/dxcapi.cpp` shouldn't be changed.

`tools/clang/tools/dxcompiler/dxcapi.cpp` however has what I believe to be wrong `_Out_` annotations that should be `_COM_Outptr_`.

`projects/dxilconv/lib/DxbcConverter/DxbcConverterImpl.h` as a header misses all annotations too (but I might not need to repeat declaration annotations at definition time in `projects/dxilconv/lib/DxbcConverter/DxbcConverter.cpp`).

Everything in the root `include/` folder should be annotated however, as I assume those become the public SDK.

Finally, `include/dxc/dxcisense.h` seems to have a bunch of "mistakes" that need to be clarified.

---

Regarding formatting, my local `clang-format` seems to be a different/newer version; will patch this up by hand.

damyanp:
Generally, we're trying to avoid SAL annotations as much as possible - so we're only including them where there are clear needs for tooling to use them.

My preference here would be for a minimal change - just enough so that the workarounds in win32metadata can be removed.  So this would just involve updating DxcCreateInstance and DxcCreateInstance2.

If there are reasons other than aiming for completeness that are motivating updating more than this then please let me know :).


> Cool, so `tools/clang/tools/dxcompiler/dxcapi.cpp` shouldn't be changed.

If there are implementations of functions defined in the header then we should make sure the SAL matches - so if we're changing dxcapi.h we may need to change the cpp as well.

> `projects/dxilconv/lib/DxbcConverter/DxbcConverterImpl.h` as a header misses all annotations too (but I might not need to repeat declaration annotations at definition time in `projects/dxilconv/lib/DxbcConverter/DxbcConverter.cpp`).

We shouldn't modify this file.


> Everything in the root `include/` folder should be annotated however, as I assume those become the public SDK.

You'd think so, but it turns out that there are many files in that include folder that aren't published.  The only file that goes into the SDK is dxcapi.h.

Some additional files are included in the nuget / zip release packages:

* dxcerrors.h
* dxcisense.h

But I'm not sure these are worth doing anything with since win32metadata isn't touching them (and long term future of dxcisense.h is TBD).


> Regarding formatting, my local `clang-format` seems to be a different/newer version; will patch this up by hand.

There's a check box in the comment that reported the error, if you click that then it'll just fix it for you.  The confusing thing is that the same comment / check box is reused every time you push new changes, so you need to scroll back up to that comment if you need to have it fixed again.

MarijnS95:
> If there are implementations of functions defined in the header then we should make sure the SAL matches - so if we're changing dxcapi.h we may need to change the cpp as well.

Right, but the cpp also has as few `_Out_` annotations (for internal forward-declared methods) that appear to be `_COM_Outptr_`; hard to tell if those should change?

>> projects/dxilconv/lib/DxbcConverter/DxbcConverter.cpp
> We shouldn't modify this file.

Ok, reverted. But keeping the `_Out_` -> `_COM_Outptr_` fix in `projects/dxilconv/include/DxbcConverter.h`?

---

My last commit already added a bunch of TODOs to `dxcisense.h`, need to investigate their applicability.

include/dxc/DxilContainer/DxcContainerBuilder.h
```diff
@@ -36,7 +36,7 @@ class DxcContainerBuilder : public IDxcContainerBuilder {
   HRESULT STDMETHODCALLTYPE RemovePart(UINT32 fourCC) override;
   // Builds a container of the given container builder state
   HRESULT STDMETHODCALLTYPE
-  SerializeContainer(IDxcOperationResult **ppResult) override;
+  SerializeContainer(_COM_Outptr_ IDxcOperationResult **ppResult) override;
```
damyanp, line 39: please revert this

include/dxc/dxcapi.internal.h
```diff
@@ -342,7 +342,7 @@ CROSS_PLATFORM_UUIDOF(IDxcContainerEventsHandler,
 struct IDxcContainerEventsHandler : public IUnknown {
 public:
   virtual HRESULT STDMETHODCALLTYPE
-  OnDxilContainerBuilt(IDxcBlob *pSource, IDxcBlob **ppTarget) = 0;
+  OnDxilContainerBuilt(IDxcBlob *pSource, _COM_Outptr_ IDxcBlob **ppTarget) = 0;
```
damyanp, line 345: please revert this

include/dxc/dxcdxrfallbackcompiler.h
```diff
@@ -107,8 +107,7 @@ CLSID_SCOPE const GUID CLSID_DxcDxrFallbackCompiler = {
     0x4b72,
     {0x9e, 0x10, 0x63, 0xcd, 0x97, 0xdf, 0x57, 0xf0}};
 
-typedef HRESULT(__stdcall *DxcCreateDxrFallbackCompilerProc)(REFCLSID rclsid,
-                                                             REFIID riid,
-                                                             LPVOID *ppv);
+typedef HRESULT(__stdcall *DxcCreateDxrFallbackCompilerProc)(
```
damyanp, line 110: please revert this

include/dxc/dxcisense.h
```diff
@@ -662,6 +662,12 @@ struct IDxcCodeCompleteResults;
 struct IDxcCompletionResult;
 struct IDxcCompletionString;
 
+// TODO for nullonfailure:
```
damyanp, line 665: please revert the changes to this file

projects/dxilconv/include/DxbcConverter.h
```diff
@@ -17,17 +17,15 @@
 #ifndef _MSC_VER
 extern "C"
 #endif
-    DXC_API_IMPORT HRESULT __stdcall DxcCreateInstance(_In_ REFCLSID rclsid,
-                                                       _In_ REFIID riid,
-                                                       _Out_ LPVOID *ppv);
+    DXC_API_IMPORT HRESULT __stdcall DxcCreateInstance(
```
damyanp, line 20: please revert the changes to this file

Issue #3370, opened 2021-01-20
[SPIR-V] ByteAddressBuffer doesn't emit RowMaj or ColMaj decorations for templated loads


Jasper-Bekkers:
I'm not sure if this is done explicitly or not, however when converting some of our ConstantBuffer's to ByteAddressBuffers we noticed that Load<float4x4> doesn't OpDecorate the matrix it's loading, nor do `column_major` or `row_major` attributes get respected when used like this: `.Load<column_major float4x4>(...)`.

Example: http://shader-playground.timjones.io/e8c49c3276e111e220873d91c937bb2b

I'm not sure if it's intentional or not, but at least I wanted to raise the issue that it's inconsistent.

dneto0:
Looking at the generated SPIR-V, the matrix type is only used for an intermediate value.  It's not used to describe the contents of a buffer. So there is no obligation to apply layout decorations to it.  In fact, it *should not* have layout decorations.

I think this is not a bug.  I recommend closing.

MarijnS95:
@dneto0 Note that the loads should probably get vectorized as per #3372, then this becomes relevant?

Generated SPIR-V remains identical regardless of `column_major` or `row_major` annotations, which should be a bug if I'm not mistaken?

Jasper-Bekkers:
> Looking at the generated SPIR-V, the matrix type is only used for an intermediate value. It's not used to describe the contents of a buffer. So there is no obligation to apply layout decorations to it. In fact, it should not have layout decorations.

Notice that not doing this leads to subtle SPIR-V specific bugs when converting for example from a StructuredBuffer<float4x4> to a ByteAddressBuffer. Though this could also point to a codegen bug for the SPIRV path that just loads the matrices in the incorrect order.

gjaegy:
Hi, I'm actually facing the exact same issue... 
We are also switching to bindless and all our matrices are now stored into a ByteAdressBuffer. When loading them from the buffer (using load<float4x4>()), an additional transpose() call is required, which I really want to avoir.
May I ask, what is the status of that issue ? Is it a bug or not, will it be fixed or not ?
Thanks a lot !

jbmcgee:
Also ran into this issue. Matrices are stored column-major in ByteAddressBuffer, but the load results in row-major. 

[...]

PR #4526, opened 2022-06-21
[spirv] Treat matrix load/stores as column major in raw buffers

kuhar:
Assume that matrices are stored in the column major order in raw buffers, e.g., `ByteAddressBuffer` and `RWByteAddressBuffer`.

Add a new flag,`-fspv-use-legacy-buffer-matrix-order`, so that shaders that depend on the previous matrix order (row major) can opt-out of this change.

Fixes: https://github.com/microsoft/DirectXShaderCompiler/issues/3370

### DIFF
--- a/tools/clang/lib/SPIRV/RawBufferMethods.cpp
+++ b/tools/clang/lib/SPIRV/RawBufferMethods.cpp
@@ -11,9 +11,11 @@
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/CharUnits.h"
 #include "clang/AST/RecordLayout.h"
+#include "clang/AST/Type.h"
 #include "clang/SPIRV/AstTypeProbe.h"
 #include "clang/SPIRV/SpirvBuilder.h"
 #include "clang/SPIRV/SpirvInstruction.h"
+#include <cstdint>
 
 namespace {
 /// Rounds the given value up to the given power of 2.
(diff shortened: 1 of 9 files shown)


tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.templated-load.matrix.hlsl
```diff
@@ -33,12 +32,14 @@ void main(uint3 tid : SV_DispatchThreadId)
 // CHECK:         [[word2:%\d+]] = OpLoad %uint [[ptr]]
 // CHECK: [[shifted_word2:%\d+]] = OpShiftRightLogical %uint [[word2]] %uint_16
 // CHECK:          [[val5:%\d+]] = OpUConvert %ushort [[shifted_word2:%\d+]]
+// CHECK:          [[row0:%\d+]] = OpCompositeConstruct %v3ushort [[val0]] [[val1]] [[val2]]
 // CHECK:          [[row1:%\d+]] = OpCompositeConstruct %v3ushort [[val3]] [[val4]] [[val5]]
 // CHECK:        [[matrix:%\d+]] = OpCompositeConstruct %_arr_v3ushort_uint_2 [[row0]] [[row1]]
 // CHECK:                          OpStore %u16 [[matrix]]
-  uint16_t2x3 u16 = buf.Load<uint16_t2x3>(tid.x);
+  uint16_t2x3 u16 = buf.Load<column_major uint16_t2x3>(tid.x);
```
sudonatalie, line 39: Could you duplicate this test with these changes rather than modifying the existing one (to help verify that the default behavior hasn't changed)?

kuhar: I wanted to decouple the testing of defaults and matrix orders. In this test, I set everything explicitly so that we can check that we can handle various matrix shapes. In the order two new test, I tested that both defaults and overrides work as expected.

kuhar:
@sudonatalie

> Can you also add this flag to the documentation here: https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#vulkan-specific-options

Done, thanks!

vettoreldaniele (approved):

tools/clang/lib/SPIRV/RawBufferMethods.cpp
```diff
@@ -290,22 +292,43 @@ SpirvInstruction *RawBufferHandler::processTemplatedLoadFromBuffer(
...
+      // major. If HLSL/DXIL support the `column_major` and `row_major`
+      // attributes in the future, we will have to check for them here and
+      // override the behavior.
+      //
+      // The assume buffer matrix order is controlled by the
+      // `-fspv-use-legacy-buffer-matrix-order` flag:
+      //   (a) false --> assume the matrix is stored column major
+      //   (b) true  --> assume the matrix is stored row major
+      //
+      // We profide (b) for compatibility with legacy shaders that depend on
```
vettoreldaniele, line 308: nit: typo `profide` -> `provide`

vettoreldaniele (approved):

sudonatalie (approved):
LGTM!

PR #4531, opened 2022-06-24
Pass parameters to CMake in a correct way

gongminmin:
According to Azure pipeline's document, the `$(...)` is used in bash to reference a variable. The original `${...}` doesn't really do it. `CC`, `CXX`, `CXX_FLAGS` are fine because they are also environment variables. But the `configuration` doesn't pass into CMAKE_BUILD_TYPE. The CI is actually build Debug build twice, not a Debug and a Release.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -109,7 +109,7 @@ stages:
     - bash: |
         mkdir build
         cd build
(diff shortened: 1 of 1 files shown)


hekota (approved):
LGTM!

Issue #3377, opened 2021-01-21
No compiler errors displayed when pixel shader fails to build


Dwedit:
This is a port of a D3D9 trivial shader that just samples a texture.  It successfully compiles on FXC as Pixel Shader 5.0, but fails to build on DXC as Pixel Shader 6.0.

No compile errors are displayed at all.  `echo %ERRORLEVEL%` returns `-1073741819` (0xC0000005, Access Violation)

There are probably mistakes in this code, but I have no clue what they are, as no compiler errors are being displayed.

```
SamplerState PointSampler { Filter = MIN_MAG_MIP_POINT; AddressU = Clamp; AddressV = Clamp; };

void main_vertex
(
	float4 position	: POSITION,
	float4 color	: COLOR,
	float2 texCoord : TEXCOORD0,

	uniform float4x4 modelViewProj,

	out float4 oPosition : POSITION,
	out float4 oColor    : COLOR,
	out float2 otexCoord : TEXCOORD
)
{
	oPosition = mul(modelViewProj, position);
	oColor = color;
	otexCoord = texCoord;
}

float4 main_fragment(float2 texCoord : TEXCOORD0, uniform Texture2D<float4> decal : TEXUNIT0) : SV_Target
{
	return decal.Sample(PointSampler, texCoord);
}
```

Dwedit:
Additional information:

Crashes in a memory copy, Source address = FEEEFEEEABABABAB

Stack trace from a debugger:

```
dxcompiler.std::vector<unsigned int,std::allocator<unsigned int> >::_Emplace_reallocate<unsigned int &>+CA from dxcompiler.memmove                                                                                                                                                                                                User
dxcompiler.operator new+1B from ???                                                                                                                                                                                                                                                                                               User
dxcompiler.hlsl::DxilParameterAnnotation::AppendSemanticIndex+2E from dxcompiler.public: unsigned int * __ptr64 __cdecl std::vector<unsigned int,class std::allocator<unsigned int> >::_Emplace_reallocate<unsigned int & __ptr64>(unsigned int * __ptr64 const,unsigned int & __ptr64) __ptr64                                   User
dxcompiler.AllocateSemanticIndex+16D from dxcompiler.public: void __cdecl hlsl::DxilParameterAnnotation::AppendSemanticIndex(unsigned int) __ptr64                                                                                                                                                                                User
[...]

PR #4538, opened 2022-07-01
Add extra type null checking to prevent AV

bob80905:
For unsupported semantic types, grabbing the semantic type would result in a nullptr. There was never a check to see if the type returned was null, and so this PR will prevent AV from happening by skipping the appropriate statements when the semantic returned is null and reporting the issue to the user.
This PR should address https://github.com/microsoft/DirectXShaderCompiler/issues/3377


lib/HLSL/HLSignatureLower.cpp
```diff
@@ -1145,10 +1145,12 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
...
       Ty = i32Ty;
     }
-    if (!hlslOP->IsOverloadLegal(opcode, Ty)) {
+  
+    if (!Ty || !hlslOP->IsOverloadLegal(opcode, Ty)) {
       std::string O;
       raw_string_ostream OSS(O);
-      Ty->print(OSS);
+      if (Ty)
+        Ty->print(OSS);
```
pow2clk, line 1153: This is definitely an improvement to crashing. It would be better if we could print out the type even though it's not one of the basic types. I realize that information may not be readily available at this location. Perhaps it can be caught when the signature element for the texture type is created?

We definitely need at least one, probably a few tests for this. I suggest trying Textures, buffers, samplers, any HLSL types. Those tests will remain useful indefinitely.

lib/HLSL/HLSignatureLower.cpp
```diff
@@ -1145,10 +1145,27 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
...
+        llvm::Type *generic_Ty = nullptr;
+        while (Ty_iterator != HLM.GetEntryFunction()->getArgumentList().end()) {
+          if (argCount == i) {
+            generic_Ty = (*Ty_iterator).getType();
+            break;
+          }
+          Ty_iterator++;
+          argCount += 1;
+        }
+        generic_Ty->print(OSS);
```
pow2clk, line 1166: I suspect generic_Ty might still be null, which is causing the crash in automated testing. As we discussed, I think it's acceptable to do without the type for this error since it gives sufficient indication of where the problem is.

tools/clang/test/HLSLFileCheck/hlsl/semantics/legacy/D3D9_effect_ps_6_0Tex2D.hlsl
```diff
@@ -0,0 +1,27 @@
+// RUN: %dxc /E main_fragment -T ps_6_0 %s | FileCheck %s
```
pow2clk, line 1: I think these three files could be merged into one that tests all of these possibilities. That would make maintaining them simpler.

You could either include a define which specifies the type of the "decal" parameter and specify it on the command line using something like:
```
// RUN: %dxc -D PARAMTYPE=Texture2D<float4> ...
// RUN: %dxc -D PARAMTYPE=SamplerState ...
...
int main_fragment(float2 texCoord : TEXCOORD0, uniform PARAMTYPE decal: TEXUNIT0
...
```
or you could just have an entry function that includes all of the invalid parameters provided the compiler produces errors for each one.

pow2clk (commented):
Just a few more suggestions while figuring out the testing failure.

As we discussed, the clang built tests are failing due to the unsigned vs signed comparison

PR #4539, opened 2022-07-02
Splitting simple array during mem2reg.

adam-yang:
Added a simple transformation before mem2reg to change all allocas of scalar arrays to individual scalar allocas. This could make value deduction easier in Od compilation.

### DIFF
--- a/lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
+++ b/lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
@@ -110,6 +110,134 @@ public:
     AU.setPreservesCFG();
   }
 
+  // Replace simple array allocas with individual scalar allocas.
+  // Only handle if:
+  // - All the alloca's users are geps
+  // - The geps all have only constant indices
+  // - The geps are indexing to just the scalar elements
+  //
+  bool SplitSimpleAllocas(llvm::Function &F) {
+    llvm::SmallVector<AllocaInst *, 10> ScalarAllocas;
+
+    if (F.empty()) return false;
+    BasicBlock *Entry = &F.getEntryBlock();
+
+    bool Changed = false;
+
+    LLVMContext &Ctx = F.getContext();
+    Module *M = F.getParent();
+    IRBuilder<> Builder(Ctx);
+    DIBuilder DIB(*M);
+    const DataLayout &DL = F.getParent()->getDataLayout();
+
+    for (Instruction *it = &Entry->back(); it != nullptr;) {
+      Instruction *I = it;
+      it = (it == &Entry->front()) ? nullptr : it->getPrevNode();
+
+      AllocaInst *AI = dyn_cast<AllocaInst>(I);
+      if (!AI) continue;
+      Type *AllocType = AI->getAllocatedType();
+      if (!AllocType->isArrayTy())
+        continue;
+      Type *ArrayElemType = AllocType->getArrayElementType();
+      if (!ArrayElemType->isSingleValueType())
+        continue;
+
+      unsigned MaxSize = 0;
+      bool Giveup = false;
+      for (User *U : AI->users()) {
+        GEPOperator *Gep = dyn_cast<GEPOperator>(U);
+        if (!Gep) {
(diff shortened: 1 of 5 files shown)


lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
```diff
@@ -356,6 +484,7 @@ class DxilConditionalMem2Reg : public FunctionPass {
 
     Changed |= RewriteOutputArgsDebugInfo(F);
     Changed |= dxilutil::DeleteDeadAllocas(F);
+    Changed |= SplitSimpleAllocas(F);
```
pow2clk, line 487: I originally wondered if we were missing an opportunity by placing this after dead alloca deletion, but I think that no unused allocas will be created by the split, so none would be deleted.

lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
```diff
@@ -110,6 +110,134 @@ class DxilConditionalMem2Reg : public FunctionPass {
     AU.setPreservesCFG();
   }
 
+  // Replace simple array allocas with individual scalar allocas.
+  // Only handle if:
+  // - All the alloca's users are geps
+  // - The geps all have only constant indices
```
pow2clk, line 116: Ah! This is why you changed the existing tests to use non-constant indices right?

lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
```diff
@@ -110,6 +110,134 @@ class DxilConditionalMem2Reg : public FunctionPass {
...
+
+    bool Changed = false;
+
+    LLVMContext &Ctx = F.getContext();
+    Module *M = F.getParent();
+    IRBuilder<> Builder(Ctx);
+    DIBuilder DIB(*M);
+    const DataLayout &DL = F.getParent()->getDataLayout();
+
+    for (Instruction *it = &Entry->back(); it != nullptr;) {
```
pow2clk, line 133: Why start at the back? IIRC, allocas are still all at the front of the block

tools/clang/test/HLSLFileCheck/passes/hl/mem2reg_hlsl/mem2reg_simple_split.ll
```diff
@@ -0,0 +1,182 @@
+; RUN: %opt %s -dxil-cond-mem2reg -S | FileCheck %s
+
+; CHECK: @main
+; CHECK-NOT: alloca [2 x float]
+
+; CHECK-DAG: !DIExpression(DW_OP_bit_piece, 0, 32)
+; CHECK-DAG: !DIExpression(DW_OP_bit_piece, 32, 32)
```
pow2clk, line 7: Maybe checks to verify that the GEPs are gone and replaced with the new allocas? I know that might get complicated, but it would more fully vet the behavior of this pass, making it a more complete unit test.

pow2clk (approved):
Looks good. I made a few comments to ensure I was following along.

Just one suggestion to enhance testing

lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
adam-yang: Yep.

lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
adam-yang: Because we could be deleting GEPs, which could be the next instruction after the alloca. Starting at the back lets me delete instructions on the fly.

tools/clang/test/HLSLFileCheck/passes/hl/mem2reg_hlsl/mem2reg_simple_split.ll
adam-yang: Thanks! Modified this test to make sure there's no allocas at all. The point of split array that early was to get get rid of the allocas.

PR #4547, opened 2022-07-11
Fixed assert in delete dead region with simple loops.

adam-yang:
Fixed an assert caused by DxilDeleteDeadRegion making the incorrect assumption that loops with no outgoing values must have no PHI nodes in its exit block. Exit blocks could have phis anyways by referencing incoming values from outside the loop.


pow2clk (approved):
Looks good modulo a comment typo

tex3d (approved):
LGTM!

PR #4551, opened 2022-07-13
Add checks to advanced tex ops exectests

pow2clk:
Two of the Advanced Texture Ops feature tests were missing the check for
the feature in addition to the check for 6.7 support. This adds those
checks

### DIFF
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -4086,6 +4086,10 @@ TEST_F(ExecutionTest, ATOProgOffset) {
                     ((UINT)sm & 0x0f));
       break;
     }
+    if (sm >= D3D_SHADER_MODEL_6_7 && !DoesDeviceSupportAdvancedTexOps(pDevice)) {
+      LogCommentFmt(L"Device does not support Advanced Texture Ops");
+      break;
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -4181,6 +4185,12 @@ TEST_F(ExecutionTest, ATOSampleCmpLevelTest) {
   if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_7))
       return;
 
+  if (!DoesDeviceSupportAdvancedTexOps(pDevice)) {
+    WEX::Logging::Log::Comment(L"Device does not support Advanced Texture Operations.");
```
bfavela, line 4189: Not sure if these get parsed by any automated scripts, but there's minor difference between this text and the previous check ('Ops' vs 'Operations.')

bfavela (approved):

tools/clang/unittests/HLSL/ExecutionTest.cpp
pow2clk: It's also a slightly different case since this is not a skip, but just an FYI note that rarely gets read. I'd have changed it for consistency anyway, but I set it to automerge. oh well!

PR #4554, opened 2022-07-13
param validation for uniform / resources in entry point functions

bob80905:
There was some lacking parameter validation for arguments to entry point functions that either had the uniform attribute, or were HLSL resource types. This was causing AV problems as described in #3377.
Now, when each parameter is processed, checking for the uniform attribute and that the parameter type is not a resource takes place, to prevent implicit assumptions from being incorrect. Additionally, some extra nullptr checking to prevent the underlying AV problem.

### DIFF
--- a/lib/DXIL/DxilUtil.cpp
+++ b/lib/DXIL/DxilUtil.cpp
@@ -491,6 +491,27 @@ std::pair<bool, DxilResourceProperties> GetHLSLResourceProperties(llvm::Type *Ty
   return FalseRet;
 }
 
+bool IsHLSLObjectStreamType(llvm::Type *Ty) {
+  if (llvm::StructType *ST = dyn_cast<llvm::StructType>(Ty)) {
+    if (!ST->hasName()) {
+      return false;
+    }
+
+    StringRef name = ST->getName();
+
+    ConsumePrefix(name, "class.");
+    ConsumePrefix(name, "struct.");
+
+    if (name.startswith("TriangleStream<"))
+      return true;
+    if (name.startswith("PointStream<"))
+      return true;
+    if (name.startswith("LineStream<"))
+      return true;
+  }
+  return false;
+}
+
 bool IsHLSLObjectType(llvm::Type *Ty) {
   if (llvm::StructType *ST = dyn_cast<llvm::StructType>(Ty)) {
     if (!ST->hasName()) {
@@ -511,15 +532,7 @@ bool IsHLSLObjectType(llvm::Type *Ty) {
     if (IsHLSLResourceType(Ty))
       return true;
 
-    ConsumePrefix(name, "class.");
-    ConsumePrefix(name, "struct.");
-
-    if (name.startswith("TriangleStream<"))
-      return true;
-    if (name.startswith("PointStream<"))
-      return true;
-    if (name.startswith("LineStream<"))
-      return true;
+    return IsHLSLObjectStreamType(Ty);
   }
   return false;
 }
--- a/lib/HLSL/DxilPreparePasses.cpp
+++ b/lib/HLSL/DxilPreparePasses.cpp
@@ -197,7 +197,7 @@ static Function *StripFunctionParameter(Function *F, DxilModule &DM,
   FunctionType *FT = FunctionType::get(VoidTy, false);
   for (auto &arg : F->args()) {
     if (!arg.user_empty())
-      return nullptr;
+      return nullptr; // If this would return here, we really should produce an error somewhere.
     DbgDeclareInst *DDI = llvm::FindAllocaDbgDeclare(&arg);
     if (DDI) {
       DDI->eraseFromParent();
--- a/lib/HLSL/HLSignatureLower.cpp
+++ b/lib/HLSL/HLSignatureLower.cpp
@@ -1145,10 +1145,12 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
       bI1Cast = true;
       Ty = i32Ty;
     }
-    if (!hlslOP->IsOverloadLegal(opcode, Ty)) {
+    if (!Ty || !hlslOP->IsOverloadLegal(opcode, Ty)) {
       std::string O;
       raw_string_ostream OSS(O);
-      Ty->print(OSS);
+      if (Ty)
+        Ty->print(OSS);
+
       OSS << "(type for " << SE->GetName() << ")";
       OSS << " cannot be used as shader inputs or outputs.";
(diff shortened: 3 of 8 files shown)


lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -197,7 +197,7 @@ static Function *StripFunctionParameter(Function *F, DxilModule &DM,
   FunctionType *FT = FunctionType::get(VoidTy, false);
   for (auto &arg : F->args()) {
     if (!arg.user_empty())
-      return nullptr;
+      return nullptr; // If this would return here, we really should produce an error somewhere.
```
tex3d, line 200: This comment was intended to identify something we needed to do before merging the changes I had (since it silently fails to remove arguments if they are still used).  I didn't intend to just add the comment here.

lib/DXIL/DxilUtil.cpp
```diff
@@ -511,15 +532,7 @@ bool IsHLSLObjectType(llvm::Type *Ty) {
...
-    ConsumePrefix(name, "class.");
-    ConsumePrefix(name, "struct.");
-
-    if (name.startswith("TriangleStream<"))
-      return true;
-    if (name.startswith("PointStream<"))
-      return true;
-    if (name.startswith("LineStream<"))
-      return true;
+    return IsHLSLObjectStreamType(Ty);
```
tex3d, line 535: This change is no longer used, right?  So it seems almost unrelated to other things in the PR now, and not really necessary.

tools/clang/test/HLSLFileCheck/hlsl/semantics/legacy/resource_in_entry_point_ps.hlsl
```diff
@@ -0,0 +1,34 @@
+// RUN: %dxc /E main -T ps_6_0 %s | FileCheck %s
+// CHECK: warning: effect state block ignored - effect syntax is deprecated. To use braces as an initializer use them with equal signs. [-Weffects-syntax]
```
tex3d, line 2: This test contains a bunch of unrelated code.  The root sig is unnecessary, the SamplerState declaration is unused, the global texture declaration is unused and may only confuse someone reading the test, since the name is the same as the input resource in main(), and the main_vertex entry point is unused.

It seems that the test could be reduced to this:
```
int main(float2 texCoord : TEXCOORD0, Texture2D<float> tex  : TEXUNIT0) : SV_Target {
  return 3;
}
```

Your next test also tests SamplerState, which should also be tested without explicit `uniform` attribute, and fail in the same way.  Though perhaps the error message would need to be updated to make more sense with SamplerState as well.

tools/clang/test/HLSLFileCheck/hlsl/semantics/legacy/uniform_in_entry_point_ps.hlsl
```diff
@@ -0,0 +1,43 @@
+// RUN: %dxc -D PARAMTYPE=Tex2D /E main_fragment -T ps_6_0 %s | FileCheck %s -check-prefix=CHECK1
+// CHECK1: warning: effect state block ignored - effect syntax is deprecated. To use braces as an initializer use them with equal signs. [-Weffects-syntax]
```
tex3d, line 2: See prior comment regarding extra irrelevant test content.

tools/clang/test/HLSLFileCheck/hlsl/semantics/legacy/uniform_in_entry_point_ps.hlsl
```diff
@@ -0,0 +1,43 @@
...
+// CHECK1: attribute uniform only valid for non-entry-point functions.
+
+// RUN: %dxc -D PARAMTYPE=RW /E main_fragment -T ps_6_0 %s | FileCheck %s -check-prefix=CHECK2
+// CHECK2: warning: effect state block ignored - effect syntax is deprecated. To use braces as an initializer use them with equal signs. [-Weffects-syntax]
+// CHECK2: attribute uniform only valid for non-entry-point functions.
+
+
+// RUN: %dxc -D PARAMTYPE=SamplerState /E main_fragment -T ps_6_0 %s | FileCheck %s -check-prefix=CHECK3
```
tex3d, line 10: This pattern for varying type could be adopted for prior test as well, though I'm not sure how much value the extra level of indirection through "Tex2D" provides.  Also, if the expected errors are the same, there's no need to have independent CHECK prefixes and separate blocks to check each case.

lib/HLSL/HLSignatureLower.cpp
```diff
@@ -1145,10 +1145,12 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
       bI1Cast = true;
       Ty = i32Ty;
     }
-    if (!hlslOP->IsOverloadLegal(opcode, Ty)) {
+    if (!Ty || !hlslOP->IsOverloadLegal(opcode, Ty)) {
       std::string O;
       raw_string_ostream OSS(O);
-      Ty->print(OSS);
+      if (Ty)
```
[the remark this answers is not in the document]
bob80905: for (unsigned i = 0; i < Sig.GetElements().size(); i++) {
    DxilSignatureElement *SE = &Sig.GetElement(i);
    llvm::Type *Ty = SE->GetCompType().GetLLVMType(HLM.GetCtx());
    
  This is how Ty is set. I'm not sure about the codegen check, but this code looks like it's looping through each argument in the function signature. Before, there was a resource in the signature (Texture2D) and I suppose since that's not an LLVM type, Ty was set to null. As to how we got to this point, I imagine the uniform semantic threw off SROA. I'll do more research, but this was the hlsl to reproduce this behavior:
  void main_vertex
(
	float4 position	: POSITION,
	float4 color	: COLOR,
	float2 texCoord : TEXCOORD0,

	uniform float4x4 modelViewProj,

	out float4 oPosition : POSITION,
	out float4 oColor    : COLOR,
	out float2 otexCoord : TEXCOORD
)
{
	oPosition = mul(modelViewProj, position);
	oColor = color;
	otexCoord = texCoord;
}

float4 main_fragment(float2 texCoord : TEXCOORD0, uniform Texture2D<float4> decal : TEXUNIT0) : SV_Target
{
	return decal.Sample(PointSampler, texCoord);
}

tex3d (changes requested):

tex3d (changes requested):

bob80905:
Merge conflicts, and according to @Tex3D it seems like this is the wrong direction

PR #4555, opened 2022-07-14
PIX Run value-to-declare per function, not per module

jeffnn:
This is a change to enable PIX to run shader debugging on libraries.
Previously, it was assumed there was a single entry point in the module.
The significant change is to add "m_Registers.clear();" as the value-to-declare pass runs over each function in the module. Without this, references to allocas would be added that straddled functions and broke validation (of course).
As mentioned in the comments, there is more work to do here, but this gets us to a place where we can debug a single function within a library.


lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -402,26 +402,48 @@ bool DxilDbgValueToDbgDeclare::runOnModule(
...
+    // function, so the instrumentation here gets a bit confused that the same
+    // variable is present in two functions and ends up pointing one function
+    // to allocas in another function. (This is easy to repro: comment out the
+    // above clear(), and run PixTest::PixStructAnnotation_Lib_DualRaygen.)
+    // Not sure what the right path forward is: might be that we have to tag
+    // m_Registers with the exported function, and maybe write out a function
+    // identifier during debug instrumentation...
+    auto &blocks = fn->getBasicBlockList();
+    for (auto &block : blocks) {
+      std::vector<Instruction *> instructions;
```
adam-yang, line 430: A common pattern that we do to safely delete while iterating is to increment the iterator right after getting the Instruction * pointer. No copying to a std::vector needed. (Unless there's danger that the next iterator could be deleted too.)

Not a big deal, just fyi.

adam-yang (approved):

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
jeffnn: Thanks for the tip

PR #4556, opened 2022-07-14
Use VertexID to check last lane.

python3kgae:
For Vertex Shader, use VertexID to check last lane instead of WaveGetLaneIndex() == WaveActiveMax(WaveGetLaneIndex()).
This is for case where 3 vertices not in same wave and g_TestResults[VS_INDEX] will be write more than once with different value.


hekota (approved):
LGTM!

PR #4559, opened 2022-07-15
Barycentric check ordering

bob80905:
Add additional check to pre-existing barycentric HLK test to ensure that GetAttributeAtVertex returns 0 for whichever vertex is first in the vertex buffer.

### DIFF
--- a/tools/clang/test/HLSL/ShaderOpArith.xml
+++ b/tools/clang/test/HLSL/ShaderOpArith.xml
@@ -1543,7 +1543,7 @@
         return result;
       }
     
-      float4 PSMain(PSInput input, float3 bary : SV_Barycentrics, uint vid : VertexID) : SV_Target {
+      float4 PSMain(PSInput input, float3 bary : SV_Barycentrics) : SV_Target {
         float4 pos = input.position;
         // check to make sure that the pixel shader will see the barycentric weight associated
         // with the first vertex in the x component of the SV_Barycentric vector, and likewise for subsequent vertices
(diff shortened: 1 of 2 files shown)


tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -1491,50 +1491,72 @@
...
+    
+      float4 PSMain(PSInput input, float3 bary : SV_Barycentrics) : SV_Target {
+        float4 pos = input.position;
+        // check to make sure that the pixel shader will see the barycentric weight associated
+        // with the first vertex in the x component of the SV_Barycentric vector, and likewise for subsequent vertices
+        float4 color = input.color;
+     
+        g_result[0] = GetAttributeAtVertex(input.svid, 0);
+        g_result[1] = GetAttributeAtVertex(input.svid, 1);
+        g_result[2] = GetAttributeAtVertex(input.svid, 2);
```
tex3d, line 1551: I think it's important for each pixel to get the right values here.  Since all pixels are writing to the same locations, we have a data race.  While we expect the values to be the same, that's something we should probably be testing, not assuming.

Some approaches to solve this:
- use AppendStructuredBuffer<uint3> and check that all resulting values are the same expected result.  Drawback: this might require additional setup the framework doesn't support easily.
- output results to render target and check those.  Since this is already outputting the interpolated color using the barycentrics, I think it would be best to just split this into a new shader if this approach is chosen, making it easier to test both things independently.

Issue #3584, opened 2021-03-15
Misleading error message for missing semantic on return value


dmpots:
```
// dxc /Tps_6_0 coverage.hlsl
[RootSignature("")]
float main(float a : A, uint b : SV_Coverage)
{
    return a;
}
```
I get the following error from dxc

```
Function: main: error: Semantic must be defined for all parameters of an entry function or patch constant function Use /Zi for source location.
```

I am probably using coverage incorrectly, but the error message does not make sense (I did give it a semantic).

dmpots:
This was my bug in the shader. It was missing an `SV_Target` on the output. It would be nice if dxc would tell me its the return value that is missing the semantic instead of the parameter.

bob80905:
Closed by https://github.com/microsoft/DirectXShaderCompiler/pull/4562

PR #4562, opened 2022-07-19
Add semantic clarification to error on missing output semantic

bob80905:
When an entry point function is missing a semantic on its output / return type, an error is thrown saying that all parameters must have a semantic. In this case the error is inaccurate since the problem is with the return argument, not any actual function parameters. So, the error message was updated to reflect this difference when the difference arises.
This PR should address #3584


tex3d (approved):
LGTM.

PR #4572, opened 2022-08-03
Replace -fcolor-diagnostics by -fdiagnostics-color

### LINKED ISSUE #3730: `-fcolor-diagnostics` fails with GCC when building llvm-tblgen
The command line argument `-fcolor-diagnostics` is not supported in GCC (at least not the version of my Ubuntu 20.04 LTS default installation):
https://github.com/microsoft/DirectXShaderCompiler/blob/e621158fbff597c978836210803d6e41d2e5550e/cmake/modules/HandleLLVMOptions.cmake#L531

The other occurance of this option is already replaced to something that works with both GCC and Clang, as Clang is fine with both `-fcolor-diagnostics` and `-fdiagnostics-color`:
https://github.com/microsoft/DirectXShaderCompiler/blob/e621158fbff597c978836210803d6e41d2e5550e/cmake/modules/HandleLLVMOptions.cmake#L380

This issue only became apparent after I ran CMake manually inside its cache folder.

Keenuts:
Replace this option with a similar one more compatible with different
compiler versions.

Fixes #3730

### DIFF
--- a/cmake/modules/HandleLLVMOptions.cmake
+++ b/cmake/modules/HandleLLVMOptions.cmake
@@ -541,7 +541,7 @@ add_llvm_definitions( -D__STDC_LIMIT_MACROS )
 if (UNIX AND
     CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
     CMAKE_GENERATOR STREQUAL "Ninja")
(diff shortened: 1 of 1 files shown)


llvm-beanz:
I don't understand how the root issue occurs and how this fixes it. This is under an `if` block that only runs for clang and the flag is a valid clang flag. This code path shouldn't be reachable if you build with gcc.

PR #4575, opened 2022-08-05
Support extension KHR_fragment_shading_barycentric

ShchchowAMD:
Add SPV support in DXC to support extension SPV_KHR_fragment_shading_barycentric.

Including new built-ins and built-in function.

Ref:
https://github.com/microsoft/DirectXShaderCompiler/wiki/SV_Barycentrics


ShchchowAMD:
Re-create to check expired build log.

PR #4576, opened 2022-08-05
[SPIRV] Fix crash when returning void

### LINKED ISSUE #3596: [SPIR-V] Return a "void" in a void function causes crash
```cpp
void A(out float a)
{
    a = 1.;
}

void B()
{
    float a;
    return A(a);
}

[shader("raygeneration")]
void MainRG() 
{
	B();
}
```

This is the shader to reproduce the issue, it causes the dxc to crash in spirvtools::opt

compile arguments as below:
`dxc.exe  -T lib_6_5 E:\test.txt -exports MainRG-spirv -fspv-target-env=vulkan1.2`

Tested on dxc of VulkanSDK 1.2.154.1, 1.2.170.0,  and revision: 259ce00a7868fc1f4585d95c305aa3fb71c6b964

Chaning the "return A(a);" to "A(a); return;" would fix the issue.

cassiebeckley:
This fixes a bug where invalid SPIR-V would be emitted when returning a void value from a void function. According to the spec, the `Value` of `OpReturnValue` must not have type `OpTypeVoid`; therefore, `OpReturn` will now be emitted instead.

Fixes #3596

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -2275,7 +2275,8 @@ void SpirvEmitter::doIfStmt(const IfStmt *ifStmt,
 void SpirvEmitter::doReturnStmt(const ReturnStmt *stmt) {
   const auto *retVal = stmt->getRetValue();
   bool returnsVoid = curFunction->getReturnType().getTypePtr()->isVoidType();
-  if (!returnsVoid && retVal) {
+  if (!returnsVoid) {
+    assert(retVal);
     // Update counter variable associated with function returns
     tryToAssignCounterVar(curFunction, retVal);
 
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -2273,7 +2273,9 @@ void SpirvEmitter::doIfStmt(const IfStmt *ifStmt,
 }
 
 void SpirvEmitter::doReturnStmt(const ReturnStmt *stmt) {
-  if (const auto *retVal = stmt->getRetValue()) {
+  const auto *retVal = stmt->getRetValue();
+  bool returnsVoid = curFunction->getReturnType().getTypePtr()->isVoidType();
+  if (!returnsVoid && retVal) {
     // Update counter variable associated with function returns
     tryToAssignCounterVar(curFunction, retVal);
 
--- a/tools/clang/test/CodeGenSPIRV/cf.return.void.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/cf.return.void.hlsl
@@ -1,0 +1,9 @@
+// RUN: %dxc -T ps_6_0 -E main
+
+void A() {
+}
+
+void main() {
+  // CHECK: OpReturn
+  return A();
+}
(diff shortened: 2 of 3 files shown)


tools/clang/test/CodeGenSPIRV/cf.return.void.hlsl
```diff
@@ -0,0 +1,9 @@
+// RUN: %dxc -T ps_6_0 -E main
+
+void A() {
+}
+
+void main() {
+  // CHECK: OpReturn
+  return A();
+}
```
s-perron, line 9: Did we double check with the DXIL developers if this should be allowed?  In C++ this is considered a syntax error, so I would have expected the same for HLSL.  I guess they are following C.  From the bug, it seems like it is being used.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -2273,7 +2273,9 @@ void SpirvEmitter::doIfStmt(const IfStmt *ifStmt,
 }
 
 void SpirvEmitter::doReturnStmt(const ReturnStmt *stmt) {
-  if (const auto *retVal = stmt->getRetValue()) {
+  const auto *retVal = stmt->getRetValue();
+  bool returnsVoid = curFunction->getReturnType().getTypePtr()->isVoidType();
+  if (!returnsVoid && retVal) {
```
s-perron, line 2278: Is it possible that a function does not return void and `retVal` is null?  If not, I would only check `returnsVoid` and assert that `retVal` is not null in the if-statement.

If it is possible, then you might need to handle that case as well.  You'll need to generate an `OpReturnValue` with an undef value for that case.  I don't think it is correct to have an `OpReturn` is a function with a non-void return type in spir-v.  See https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpReturn

s-perron (changes requested):
Looks good.  That is a good fix for the problem in the bug.  However, there might be another case to consider.

tools/clang/test/CodeGenSPIRV/cf.return.void.hlsl
cassiebeckley: DXIL does allow it, but I don't know if it _should_.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: I checked and the frontend doesn't allow non-void functions to not return anything (`error: non-void function 'B' should return a value [-Wreturn-type`), so I switched to an assert. I'm not sure if there's a formal specification for HLSL – who should I talk to about HLSL structure?

s-perron (approved):

tools/clang/test/CodeGenSPIRV/cf.return.void.hlsl
gongminmin: > Did we double check with the DXIL developers if this should be allowed? In C++ this is considered a syntax error, so I would have expected the same for HLSL. I guess they are following C. From the bug, it seems like it is being used.

No, in C++ it's allowed. "In a function returning (possibly cv-qualified) void, the return statement with expression can be used, if the expression type is (possibly cv-qualified) void. "

PR #4585, opened 2022-08-10
Add vs2022 support to hctstart

pow2clk:
vs2022 installs in a different location, so finding cmake requires a
slightly different path. Variable assignment in batch files is ugly.
Duplication of code slightly less ugly.

Mostly done by @tex3d

### DIFF
--- a/utils/hct/hctstart.cmd
+++ b/utils/hct/hctstart.cmd
@@ -115,20 +115,18 @@ echo.
 goto :eof
 
 :findcmake 
-for %%v in (2022 2019 2017) do (
-  for %%e in (Community Professional Enterprise) do (
-    if "%%v"=="2022" (
-      if exist "%programfiles%\Microsoft Visual Studio\%%v\%%e\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin" (
-        set "PATH=%PATH%;%programfiles%\Microsoft Visual Studio\%%v\%%e\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin"
-        echo Path adjusted to include cmake from Visual Studio %%v %%e.
-        exit /b 0
-      )
-    ) else (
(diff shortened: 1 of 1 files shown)


utils/hct/hctstart.cmd
```diff
@@ -115,12 +115,20 @@ echo.
...
 
 :findcmake 
-for %%v in (2019 2017) do (
+for %%v in (2022 2019 2017) do (
   for %%e in (Community Professional Enterprise) do (
-    if exist "%programfiles(x86)%\Microsoft Visual Studio\%%v\%%e\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin" (
-      set "PATH=%PATH%;%programfiles(x86)%\Microsoft Visual Studio\%%v\%%e\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin"
-      echo Path adjusted to include cmake from Visual Studio %%v %%e.
-      exit /b 0
+    if "%%v"=="2022" (
```
tex3d, line 120: I suppose it would make sense to just separate the 2022 check out of the loop instead of special casing in the loop.  For future versions installed under 64-bit programfiles path, a separate loop can search those first without extra conditions inside one loop.
In any case, this whole thing *is* a bit ugly of course.

Side note: The default build is 2019, not 2022, and you have to remember to supply -vs2022 every time you use hctbuild.cmd.  It would be better if we could detect the vs version found and auto-supply the default in hctbuild.cmd.

Side note 2: if you use VsDevCmd.bat to initialize vs environment, this detection is unnecessary if cmake is included in that installation, but it can/does mess up testing by allowing DxCompiler.dll to find DXIL.dll in the SDK path, when you didn't expect to use that DLL.  We need another workaround for that in hcttest.cmd (and I have an ugly patch I use locally).

tex3d (approved):
LGTM!

Issue #3355, opened 2021-01-14
Some compile time const expressions can't be used to specify const array lengths (works with FXC)


Dredhog:
### Title
Some compile time const expressions can't be used to specify const array lengths (works with FXC)

### Functional impact
Certain shaders which use expressions using arithmetic operations and intrinsics involving constants that compile with FXC cannot be compiled with DXC. Lack of support for the `min` intrinsic while `max` works seem like an obvious oversight.

### Minimal repro steps

1. Compile the following shader with `dxc.exe -T ps_6_0 -E PSMain issue_shader.txt`:
```
float4 BrokenTestArray0[int(.3*12)];
float4 BrokenTestArray1[min(2, 3)];
float4 WorkingTestArray0[max(2, 3)];

float4 PSMain(float4 color : COLOR) : SV_TARGET
{
    return BrokenTestArray0[int(color.x)] +
           BrokenTestArray1[int(color.y)] +
           WorkingTestArray0[int(color.z)];
}
```
2. Observe the compiler errors:
```
C:\issue_shader.txt:1:24: error: variable length arrays are not supported in HLSL
float4 BrokenTestArray0[int(.3*12)];
                       ^
C:\issue_shader.txt:2:24: error: variable length arrays are not supported in HLSL
float4 BrokenTestArray1[min(2, 3)];
                       ^
```
### Expected result
Casting to int and also the `min()` intrinsic with constant arguments are allowed when used in describing const array lengths as these operations work with FXC and since the `max()` intrinsic already works with DXC.

### Actual result
Casting to int and the `min()` intrinsic with constant arguments produce compiler errors when used to define const array lengths.

### Further technical details
There are likely other untested/unsupported operations which I haven't noticed yet. The use of `min` was an actual use case from our Unity shader codebase and I found the casting issue with some quick testing. It seems that `mad` and other intrinsics don't work as well but I didn't list them as some of them also fail with FXC, though it would be nice if all const expressions could be supported when describing array lengths.

pow2clk:
Hi @Dredhog!

We're looking into this issue, but it looks unlikely for the next release. What is the priority for you?

Dredhog:
Hi @pow2clk, thanks for taking it on!

It is not a critiacal/blocking issue for us ATM as I've only noticed it affect one shader in our SriptableRenderPipeline shader codebase, I was able to work around the issue from the shader using a macro.

[...]

PR #4591, opened 2022-08-11
allow array sizes defined by min ops to not be treated as variable length

bob80905:
added an extra case to use the comparison function for "min" when handling intrinsics that define the size of an array.
This PR should partially address #3355

### DIFF
--- a/tools/clang/test/HLSLFileCheck/validation/NonVLAArrays.hlsl
+++ b/tools/clang/test/HLSLFileCheck/validation/NonVLAArrays.hlsl
@@ -1,13 +1,13 @@
 // RUN: %dxc -E PSMain -T ps_6_0 %s | FileCheck %s
 
-// CHECK: define void @PSMain()
+// CHECK: %"$Globals" = type { [2 x <4 x float>], [3 x <4 x float>] }
 
 // fix me: float4 BrokenTestArray0[int(.3)*12]; 
-float4 BrokenTestArray1[min(2, 3)];
-float4 WorkingTestArray0[max(2, 3)];
+float4 WorkingTestArray0[min(2, 3)];
+float4 WorkingTestArray1[max(2, 3)];
 
 float4 PSMain(float4 color : COLOR) : SV_TARGET
 {
-    return BrokenTestArray1[int(color.y)] +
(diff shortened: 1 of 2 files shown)


tools/clang/test/HLSLFileCheck/validation/NonVLAArrays.hlsl
```diff
@@ -0,0 +1,15 @@
+// RUN: %dxc -E PSMain -T ps_6_0 %s | FileCheck %s
+
+// CHECK: error: variable length arrays are not supported in HLSL
+// CHECK-NEXT: float4 BrokenTestArray0[int(.3)*12];
```
pow2clk, line 4: If we are going to check for this failure, it would be better to check it independently of the checks for lack of failure. If the compiler failed due to this failure and also produced failures for the min and or the max, then this test would still pass. You could conditionally include the expected failure array or you could split this file into one that passes and one that fails. 

That said, I'm not sure we landed on wanting this to fail.

tools/clang/test/HLSLFileCheck/validation/NonVLAArrays.hlsl
```diff
@@ -0,0 +1,15 @@
...
+
+float4 BrokenTestArray0[int(.3)*12];
+float4 BrokenTestArray1[min(2, 3)];
```
pow2clk, line 7: This isn't broken anymore

pow2clk: Since the array isn't broken, I recommend naming it something else

tools/clang/test/HLSLFileCheck/validation/NonVLAArrays.hlsl
```diff
@@ -0,0 +1,13 @@
+// RUN: %dxc -E PSMain -T ps_6_0 %s | FileCheck %s
+
+// CHECK: define void @PSMain()
```
pow2clk, line 3: We are determined to get away from tests that only check for an entry function. In this case, I think the most important part would be to verify that these cbuffers show up in the $Globals with the proper sizes.

pow2clk (approved):
Looks good!

PR #4597, opened 2022-08-15
Fix Layer & ViewportIndex ext in 1.2+ 

Keenuts:
Back in Vulkan1.1, ShaderViewportIndexLayer was required to use SV_RenderTargetArrayIndex. This was promoted in VK1.2:
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_EXT_shader_viewport_index_layer.html
Sadly, DXC didn't checked the target SPV environment, and always requested the extension. This commit addresses this.

Additionally, when DXC finishes generating SPIR-V, the binary is disassembled and printed to stdout.
The target env was hardcoded to Vulkan1.1, meaning disassembly would fail if we tried disassembling Vulkan1.2+ SPIR-V.

### DIFF
--- a/lib/Support/MemoryBuffer.cpp
+++ b/lib/Support/MemoryBuffer.cpp
@@ -86,6 +86,11 @@ public:
     init(InputData.begin(), InputData.end(), RequiresNullTerminator);
   }
 
+  // Disable sized deallocation for MemoryByfferMem, because it has
+  // tail-allocated data.
+  // (See llvm commit 21c303e9eadfbd2d685665176159f5f4738169b1)
+  void operator delete(void *p) { ::operator delete(p); }
+
   const char *getBufferIdentifier() const override {
      // The name is stored after the class itself.
     return reinterpret_cast<const char*>(this + 1);
@@ -211,6 +216,11 @@ public:
       init(Start, Start + Len, RequiresNullTerminator);
     }
   }
+
+  // Disable sized deallocation for MemoryByfferMem, because it has
+  // tail-allocated data.
+  // (See llvm commit 21c303e9eadfbd2d685665176159f5f4738169b1)
+  void operator delete(void *p) { ::operator delete(p); }
 
   const char *getBufferIdentifier() const override {
     // The name is stored after the class itself.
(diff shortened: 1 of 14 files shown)


tools/clang/test/CodeGenSPIRV/raytracing.target-env-error.hlsl
```diff
@@ -1,6 +1,6 @@
 // RUN: %dxc -T lib_6_3
 
-// CHECK: error: Vulkan 1.1 with SPIR-V 1.4 is required for Raytracing but not permitted to use
+// CHECK: error: vulkan1.1spirv1.4 is required for Raytracing but not permitted to use
```
s-perron, line 3: This is not a good change to the error message.

s-perron (changes requested):
Does not look too bad, but I do not like the change to the error messages.  Those should be human readable, and using official spelling for names.  For example "Vulkan" is capitalized.

tools/clang/test/CodeGenSPIRV/raytracing.target-env-error.hlsl
Keenuts: Right, "reverted" this part. (Split the string to target env function with env to human readable).
Reverted test file changes.

Keenuts:
Thanks for the review. Updated the PR to not change the tests.

lib/Support/MemoryBuffer.cpp
```diff
@@ -86,6 +86,11 @@ class MemoryBufferMem : public MemoryBuffer {
     init(InputData.begin(), InputData.end(), RequiresNullTerminator);
   }
 
+  // Disable sized deallocation for MemoryByfferMem, because it has
+  // tail-allocated data.
+  // (See llvm commit 21c303e9eadfbd2d685665176159f5f4738169b1)
+  void operator delete(void *p) { ::operator delete(p); }
+
```
s-perron, line 93: The changes outside of SPIRV should be reviewed by some from Microsoft.

s-perron (commented):
LGTM.  Some changes should be reviewed by others.

Keenuts:
@tex3d would you be available to review those DXC changes?

Keenuts:
@tex3d cherry-picked fix from llvm upstream for MemoryBuffer also fixes a bug when using tcmalloc. If that's what's blocking this review, I could split it, but it would also be nice to have this additional fix.

pow2clk (approved):
The upstream MemoryBuffer changes look fine and the changes to dxc.cpp are strictly SPIRV related and look fine too.

PR #4599, opened 2022-08-16
Add diagnostic tests

gracejennings:
This change cleans up the diagnostics emitted in the /lib/HLSL and /lib/DXIL. This started as an exploration of moving post codegen and pre validation diagnostics earlier in compilation. Most of the functional change here is to add testing for untested errors and change unreachable errors to internal compiler errors. For each diagnostic: 

1. Changed to the updated dxilutil error emitting functions
2. Added a test if untested
3. Changed to an internal compiler error if unreachable

### DIFF
--- a/lib/HLSL/DxilGenerationPass.cpp
+++ b/lib/HLSL/DxilGenerationPass.cpp
@@ -210,7 +210,7 @@ public:
     if (!SM->IsLib()) {
       Function *EntryFn = m_pHLModule->GetEntryFunction();
       if (!m_pHLModule->HasDxilFunctionProps(EntryFn)) {
-        llvm_unreachable("Entry function doesn't have property.");
+        llvm_unreachable("Entry function doesn't have any properties.");
         return false;
       }
       DxilFunctionProps &props = m_pHLModule->GetDxilFunctionProps(EntryFn);
--- a/lib/HLSL/DxilGenerationPass.cpp
+++ b/lib/HLSL/DxilGenerationPass.cpp
@@ -210,7 +210,7 @@ public:
     if (!SM->IsLib()) {
       Function *EntryFn = m_pHLModule->GetEntryFunction();
       if (!m_pHLModule->HasDxilFunctionProps(EntryFn)) {
-        dxilutil::EmitErrorOnFunction(M.getContext(), EntryFn, "Entry function don't have property.");
+        llvm_unreachable("Entry function doesn't have property.");
         return false;
       }
       DxilFunctionProps &props = m_pHLModule->GetDxilFunctionProps(EntryFn);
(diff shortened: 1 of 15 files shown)


lib/HLSL/DxilGenerationPass.cpp
```diff
@@ -210,7 +210,7 @@ class DxilGenerationPass : public ModulePass {
     if (!SM->IsLib()) {
       Function *EntryFn = m_pHLModule->GetEntryFunction();
       if (!m_pHLModule->HasDxilFunctionProps(EntryFn)) {
-        dxilutil::EmitErrorOnFunction(M.getContext(), EntryFn, "Entry function don't have property.");
+        llvm_unreachable("Entry function doesn't have property.");
```
pow2clk, line 213: As long as we're correcting the grammar, perhaps "doesn't have any properties"?

lib/HLSL/DxilPromoteResourcePasses.cpp
```diff
@@ -219,7 +220,8 @@ bool DxilPromoteStaticResources::PromoteStaticGlobalResources(
       Insts.clear();
     }
     if (!bUpdated) {
-      M.getContext().emitError(dxilutil::kResourceMapErrorMsg);
+      llvm_unreachable(
+          "local resource not guaranteed to map to unique global resource.");
```
pow2clk, line 224: I'd like to understand better how these were identified as unreachable. They are in a pass that comes before the only untouched reporting of this error in LowerCreateHandleForLib. This is a rather squirrelly error that often only fires with no optimizations enabled. I worry that these might be triggerable here after all.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -2387,7 +2388,7 @@ Value *TranslatePrintf(CallInst *CI, IntrinsicOp IOP, DXIL::OpCode opcode,
                        HLObjectOperationLowerHelper *pObjHelper,
                        bool &Translated) {
   Translated = false;
-  CI->getContext().emitError(CI, "use of undeclared identifier 'printf'");
+  dxilutil::EmitErrorOnInstruction(CI, "use of undeclared identifier 'printf'");
```
pow2clk, line 2391: It's pre-existing, but I feel that this message could be more helpful. It's not undeclared nor unrecognized, it's just not supported by DXC (yet)

pow2clk (commented):
A few small suggestions and a question

tools/clang/test/HLSLFileCheck/hlsl/resource_binding/unbound_overlap.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: %dxc -E main -T ps_6_0 %s  | FileCheck -input-file=stderr %s
+
+
+// CHECK: error: more than one unbounded resource (Tex1 and Tex2) in space 0
+
+
+Texture2D<float4> Tex1[] : register(t5);  // unbounded
+Texture2D<float4> Tex2[] : register(t6);  // unbounded
```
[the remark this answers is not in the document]
gracejennings: I'll add some broader testing here

pow2clk (approved):
Looks good! Nice to get our diagnostics (more) in order!

PR #4600, opened 2022-08-16
[SPIRV] Generate DebugSource for API invocation with in-memory source

greg-lunarg:


greg-lunarg:
@Keenuts Can you please review?

tools/clang/lib/SPIRV/EmitVisitor.cpp
```diff
@@ -1437,10 +1437,11 @@ void EmitVisitor::generateDebugSourceContinued(uint32_t textId,
 void EmitVisitor::generateChoppedSource(uint32_t fileId, SpirvDebugSource *inst) {
   // Chop up the source into multiple segments if it is too long.
   llvm::SmallVector<std::string, 2> choppedSrcCode;
-  std::string text;
+  std::string text = inst->getContent();
```
Keenuts, line 1440: text seem to only be used in the if(spcOptions.debugInfoSource) scope.
Would suggest moving declaration into the scope.

Keenuts (changes requested):
Thanks for the work!
Would it be possible to write tests for this addition?

greg-lunarg:
> Would it be possible to write tests for this addition?

Possible? Maybe. Easy? I don't think so.

The testing framework is based on the source of the test coming from a file. The bug is a problem when the source is  in-memory and passed through the API.

I would have to hack the testing framework to take in-memory source. It would be a one-off and would not be used again. I am not sure it is worth it.

This is why the bug was missed in the first place: it cannot be easily tested for with the current framework.

So I respectfully ask for an exemption from submitting a test.

greg-lunarg:
@Keenuts I have addressed all concerns besides the testing. Can you please review?

Keenuts:
> The testing framework is based on the source of the test coming from a file. The bug is a problem when the source is in-memory and passed through the API.

Couldn't this be done without the framework tho? Tried to setup a quick unittest using the SpirvBuilder, and this is what I have:

```diff
diff --git a/tools/clang/lib/SPIRV/EmitVisitor.cpp b/tools/clang/lib/SPIRV/EmitVisitor.cpp
index 8a003dc4b..48fbc6ded 100644
--- a/tools/clang/lib/SPIRV/EmitVisitor.cpp
+++ b/tools/clang/lib/SPIRV/EmitVisitor.cpp
@@ -1440,7 +1440,10 @@ void EmitVisitor::generateChoppedSource(uint32_t fileId, SpirvDebugSource *inst)
   std::string text;
   uint32_t textId = 0;
   if (spvOptions.debugInfoSource) {
-    text = ReadSourceCode(inst->getFile());
+    text = inst->getContent();
+    if (text.empty()) {
+      text = ReadSourceCode(inst->getFile());
+    }
     if (!text.empty()) {
       // Maximum characters for DebugSource and DebugSourceContinued
       // OpString literal minus terminating null.
diff --git a/tools/clang/unittests/SPIRV/SpirvDebugInstructionTest.cpp b/tools/clang/unittests/SPIRV/SpirvDebugInstructionTest.cpp
index c003604a7..992750270 100644
--- a/tools/clang/unittests/SPIRV/SpirvDebugInstructionTest.cpp
+++ b/tools/clang/unittests/SPIRV/SpirvDebugInstructionTest.cpp
@@ -7,10 +7,12 @@
 //
 //===---------------------------------------------------------------------===//

+#include "FileTestUtils.h"
 #include "SpirvTestBase.h"
 #include "clang/SPIRV/SpirvBuilder.h"
 #include "clang/SPIRV/SpirvInstruction.h"
 #include "clang/SPIRV/SpirvType.h"
+#include "effcee/effcee.h"

 using namespace clang::spirv;

@@ -19,11 +21,22 @@ namespace {
 class SpirvDebugInstructionTest : public SpirvTestBase {
 public:
   SpirvDebugInstructionTest()
-      : spirvBuilder(getAstContext(), getSpirvContext(), {},
-                     getFeatureManager()) {}
+   : spirvCodeGenOptions(getDefaultOptions()),
+     spirvBuilder(getAstContext(), getSpirvContext(),
+         spirvCodeGenOptions, getFeatureManager()) {}
+
+  static SpirvCodeGenOptions getDefaultOptions() {
+    SpirvCodeGenOptions options = { };
+    options.debugInfoVulkan = true;
+    options.debugInfoSource = true;
+    return options;
+  }
+
   SpirvBuilder *GetSpirvBuilder() { return &spirvBuilder; }
+  const SpirvCodeGenOptions& GetOptions() const { return spirvCodeGenOptions; };

 private:
+  SpirvCodeGenOptions spirvCodeGenOptions;
   SpirvBuilder spirvBuilder;
 };

@@ -42,4 +55,28 @@ TEST_F(SpirvDebugInstructionTest, DynamicTypeCheckDebugTypeTemplateParameter) {
   EXPECT_TRUE(llvm::isa<SpirvInstruction>(i));
 }

+TEST_F(SpirvDebugInstructionTest, ReadSourceFromMemory) {
+  const std::string checkCommands =
+  "CHECK: [[s0:%\\d+]] = OpString \"file\"\n"
+  "CHECK: [[s1:%\\d+]] = OpString \"return x;\"\n"
+  "CHECK: [[r0:%\\d+]] = OpExtInst %void [[e0:%\\d+]] DebugSource [[s0]] [[s1]]";
+
+  auto builder = GetSpirvBuilder();
+  builder->setMemoryModel(spv::AddressingModel::Logical, spv::MemoryModel::VulkanKHR);
+  SpirvDebugInstruction *i = builder->createDebugSource("file", "return x;");
+  i->setAstResultType(getAstContext().VoidTy);
+  i->setResultType(getSpirvContext().getVoidType());
+  i->setInstructionSet(builder->getDebugInfoExtInstSet(true));
+
+  auto binary = builder->takeModule();
+
+  std::string assembly;
+  ASSERT_TRUE(utils::disassembleSpirvBinary(binary, &assembly, false, SPV_ENV_VULKAN_1_1));
+  effcee::Result result = effcee::Match(assembly, checkCommands);
+  if (result.status() != effcee::Result::Status::Ok) {
+    fprintf(stderr, "%s\n", result.message().c_str());
+  }
+  ASSERT_EQ(result.status(), effcee::Result::Status::Ok);
+}
+
 } // anonymous namespace
```


This adds a test generating a DebugSource instruction with what I believe is in-memory source, and checks that the generated code is indeed containing the source.

Keenuts (changes requested):
Are the submodule updates expected?
SPIRV-Tools does show a valid commit, but effcee cannot show me the diff by ex.

greg-lunarg:
> Are the submodule updates expected?

No. I accidentally picked up a different SPIRV-Tools. I will fix that.

s-perron (commented):
I agree that a test would be good.  I don't know enough about the testing framework to know how it could be done.  I'll leave this to Nathan.

greg-lunarg:
> This adds a test generating a DebugSource instruction with what I believe is in-memory source, and checks that the generated code is indeed containing the source.

Unfortunately, this is not what is needed here.

This code creates SPIR-V in memory and passes it immediately to a disassembler.

What is needed is code which passes in-memory HLSL to the dxc API and collects the binary SPIR-V from that.

So I stand by my original assessment that the current framework does not support creating a test for this, and that it is not clear that the effort is worth the benefit.

Normally, testing is meant to test very complex algorithms such parsing or code generation to guard against regression after frequent changes in these delicate areas.  The functionality I am fixing here is just a simple question of plumbing. This area is not complex and is not frequently changed.

Nonetheless, this PR fixes a very important and show-stopping bug for a very real and important Vulkan user with approximately a half-dozen lines of code.

Can we please merge this?

Keenuts:
Hi! Sorry for the delay, was out-of-office.
I see. Don't know enough about the whole DXC testing framework, and if it would be doable in a clean manner, will keep that in mind for future tasks, in the meantime, let's unblock this.

Keenuts (approved):

PR #4603, opened 2022-08-18
PIX: stabilize block enumeration order for PIX's instruction numbering pass.

jeffnn:
PIX has a pass that adds "instruction ordinals" in order to have a common communication mechanism between, e.g., shader debugging instrumentation and the PIX UI. Unfortunately, the enumeration order of basic blocks in a module isn't necessarily stable, resulting in odd things happening in shader debugging. This change attempts to sort basic blocks so that instructions have a consistent ordering.


adam-yang (approved):
LGTM, but it's alarming this is needed. The underlying non-determinism must be fixed.

jeffnn:
This wasn't really a problem. Abandoning PR.

PR #4604, opened 2022-08-18
Fix unroller crash in library targets

adam-yang:
Fixed when there are multiple functions with loops that can't be unrolled at
the point of doing [unroll], the unroll pass crashes because it didn't clean up
memory between runs.


pow2clk (approved):
Looks good!

Not worth remedying now (or maybe ever), but I find `LoopsThatFailed` to be a bit confusing, but in trying to think of alternatives, perhaps I realize why that was chosen, `UnUnrolledLoops`? `RolledLoops`?😆

PR #4581, opened 2022-08-08
Recognize RaserizerOrderedTexure as resource type

pow2clk:
The HLSL type system was failing to recognize the
RasterizerOrderedTexture2D type as a resource type. As a result, it was
not getting annotated properly during UAV codegen and when the time came
to assign a struct member of that type, there was no declaration to use
to do so. This resulted in a validation failure when used as a struct
member.

By simply adding it to the list of detected resource types, the
assignments are successful.

Fixes #4578

Issue #4578, opened 2022-08-08
RasterizerOrderedTexture2D initialization fails when placed within a struct


charles-epic:
Hi,

Placing a **RasterizerOrderedTexture2D** resource within a struct and initializing it with a global resource causes an error.

Here is a simple repro of the issue:

```
RasterizerOrderedTexture2D<uint> GlobalTexture;

static const struct {
	RasterizerOrderedTexture2D<uint> Texture;
} Output = {GlobalTexture};

void MainPS(in float4 SvPosition : SV_Position, out float4 Out : SV_Target0)
{
	Out = float4(0, 0, 0, 0);
	Output.Texture[uint2(SvPosition.xy)] += 1;
}
```

which results into the following output:

> error: validation errors
D:\local\Temp\629636d9-99d6-453a-a41d-098ffb2b2c95.hlsl:10:2: error: Constant values must be in-range for operation.
note: at '%8 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %1, %dx.types.ResourceProperties zeroinitializer)' in block '#0' of function 'MainPS'.
D:\local\Temp\629636d9-99d6-453a-a41d-098ffb2b2c95.hlsl:10:17: error: Instructions must be of an allowed type.
note: at '%6 = insertelement <2 x i32> undef, i32 %4, i64 0' in block '#0' of function 'MainPS'.
Function: dx.hl.subscript.[].i32* (i32, %dx.types.Handle, <2 x i32>): error: External function 'dx.hl.subscript.[].i32* (i32, %dx.types.Handle, <2 x i32>)' is not a DXIL function.
Function: dx.hl.subscript.[].i32* (i32, %dx.types.Handle, <2 x i32>): error: Function 'dx.hl.subscript.[].i32* (i32, %dx.types.Handle, <2 x i32>)' uses resource in function signature.
Validation failed.

Thank you for your help!

/Charles.

PR #4606, opened 2022-08-19
Accept all ROV types as resources

pow2clk:
Includes all the Raster Order Views types in the check for UAVs and add
a test that exercises all of them, verifying they can be assigned
properly.

Followon to fix #4581 for bug #4578


dmpots (approved):
LGTM

PR #4617, opened 2022-08-25
Initialize all DxilModule members

### LINKED ISSUE #4616: DxilModule::m_bResMayAlias can be uninitialized
In the case where the DXIL module is being loaded, if no shader flags are present in the module (`rawShaderFlags` is zero), then `m_bResMayAlias` is never assigned. This causes problems when using the DXIL validator to validate SM6.7 shaders that were compiled with resource aliasing *enabled* (since the default of disabling aliasing produces the no-aliasing shader flag) and no other flag-requiring functionality.

They may sometimes arbitrary fail to validate depending on the value of this uninitialized variable.

pow2clk:
m_bResMayAlias was not initialized, leading to intermittent failures
when it was not explicitly set. The only functional change here is
setting that variable to eliminate that volatility. The other changes
are to move the setting of these simple typed variables to the header to
make it less likely that future added members will suffer the same fate.

Also in the process, I couldn't resist removing an unused variable added
along with the Payload Access Qualifiers in error, removing redundant
private regions, and moving the private methods out of the middle of
all the variable members.

Fixes #4616

### DIFF
--- a/include/dxc/DXIL/DxilModule.h
+++ b/include/dxc/DXIL/DxilModule.h
@@ -344,8 +344,8 @@ private:
   llvm::Module *m_pModule = nullptr;
   llvm::Function *m_pEntryFunc = nullptr;
   std::string m_EntryName = "";
-  std::unique_ptr<DxilMDHelper> m_pMDHelper = nullptr;
-  std::unique_ptr<llvm::DebugInfoFinder> m_pDebugInfoFinder = nullptr;
+  std::unique_ptr<DxilMDHelper> m_pMDHelper;
+  std::unique_ptr<llvm::DebugInfoFinder> m_pDebugInfoFinder;
   const ShaderModel *m_pSM = nullptr;
   unsigned m_DxilMajor = DXIL::kDxilMajor;
   unsigned m_DxilMinor = DXIL::kDxilMinor;
@@ -381,7 +381,7 @@ private:
   uint32_t m_IntermediateFlags = 0;
   uint32_t m_AutoBindingSpace = UINT_MAX;
 
-  std::unique_ptr<DxilSubobjects> m_pSubobjects = nullptr;
+  std::unique_ptr<DxilSubobjects> m_pSubobjects;
 
   // m_bMetadataErrors is true if non-fatal metadata errors were encountered.
   // Validator will fail in this case, but should not block module load.
--- a/include/dxc/DXIL/DxilModule.h
+++ b/include/dxc/DXIL/DxilModule.h
@@ -333,30 +333,28 @@ private:
   std::vector<std::unique_ptr<DxilSampler> > m_Samplers;
 
   // Geometry shader.
-  DXIL::PrimitiveTopology m_StreamPrimitiveTopology;
-  unsigned m_ActiveStreamMask;
+  DXIL::PrimitiveTopology m_StreamPrimitiveTopology = DXIL::PrimitiveTopology::Undefined;
+  unsigned m_ActiveStreamMask = 0;
 
-private:
   enum IntermediateFlags : uint32_t {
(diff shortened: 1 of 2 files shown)


include/dxc/DXIL/DxilModule.h
```diff
@@ -381,24 +395,6 @@ class DxilModule {
...
-  uint32_t m_AutoBindingSpace;
-
-  // porperties infered from the DXILTypeSystem
-  bool m_bHasPayloadQualifiers;
-
-  std::unique_ptr<DxilSubobjects> m_pSubobjects;
-
-  // m_bMetadataErrors is true if non-fatal metadata errors were encountered.
-  // Validator will fail in this case, but should not block module load.
-  bool m_bMetadataErrors;
```
pow2clk, line 401: Well, the way the diff looks now, it seems like I moved these variables above the methods and initialized them. That's accurate enough I guess, but the way I see it, I moved the methods below the variables. The important thing to note is that there were no changes to any of the relocated code apart from giving them their default assignments here and the one removed member mentioned above.

include/dxc/DXIL/DxilModule.h
```diff
@@ -381,24 +395,6 @@ class DxilModule {
...
-  bool m_bUseMinPrecision;
-  bool m_bAllResourcesBound;
-  bool m_bResMayAlias;
-
-  // properties from HLModule that should not make it to the final DXIL
-  uint32_t m_IntermediateFlags;
-  uint32_t m_AutoBindingSpace;
-
-  // porperties infered from the DXILTypeSystem
-  bool m_bHasPayloadQualifiers;
```
pow2clk, line 395: This is the variable that was removed because it was unused @pacxx? Do you remember why it was added in #3171 in the first place? It has never been used.

include/dxc/DXIL/DxilModule.h
```diff
@@ -333,30 +333,28 @@ class DxilModule {
...
-  std::unique_ptr<llvm::DebugInfoFinder> m_pDebugInfoFinder;
-  const ShaderModel *m_pSM;
-  unsigned m_DxilMajor;
-  unsigned m_DxilMinor;
-  unsigned m_ValMajor;
-  unsigned m_ValMinor;
-  bool m_ForceZeroStoreLifetimes;
+  llvm::Module *m_pModule = nullptr;
+  llvm::Function *m_pEntryFunc = nullptr;
+  std::string m_EntryName = "";
```
jenatali, line 346: I'd suggest leaving the STL containers alone, they have default constructors that already initialize themselves.

jenatali: In fact, I think the `= nullptr;` on the `llvm::DebugInfoFinder` is what's breaking the CI build.

jenatali (approved):

lib/DXIL/DxilModule.cpp
```diff
@@ -96,29 +96,13 @@ void ClearDxilHook(Module &M);
...
-, m_ValMinor(0)
-, m_ForceZeroStoreLifetimes(false)
 , m_pOP(make_unique<OP>(pModule->getContext(), pModule))
 , m_pTypeSystem(make_unique<DxilTypeSystem>(pModule))
-, m_bDisableOptimizations(false)
-, m_bUseMinPrecision(true) // use min precision by default
-, m_bAllResourcesBound(false)
-, m_IntermediateFlags(0)
-, m_AutoBindingSpace(UINT_MAX)
 , m_pSubobjects(nullptr)
```
jenatali, line 105: For this and `m_pDebugInfoFinder`, you can remove the explicit `nullptr` initialization - since they're `unique_ptr`s, initialization with `nullptr` is the default.

PR #4623, opened 2022-08-29
Remove clang-format-vs

hekota:
It is an outdated version that no longer works and is not included in DXC build.


pow2clk (approved):
Don't let the door hit clang-format-vs on the way out! ;)

PR #4624, opened 2022-08-30
Change /P to match cl.exe /P behavior.

python3kgae:
/P will write to InputFile.i
/P -Fi filename will write to filename.
/P filename still work with hack.

### DIFF
--- a/lib/DxcSupport/HLSLOptions.cpp
+++ b/lib/DxcSupport/HLSLOptions.cpp
@@ -597,23 +597,28 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
     // Hack to support fxc style /P preprocess_filename.
     // When there're more than 1 Input file, use the input which is after /P as
     // preprocess.
-    std::vector<std::string> Inputs = Args.getAllArgValues(OPT_INPUT);
-    if (Inputs.size() > 1) {
-      llvm::opt::Arg *PArg = Args.getLastArg(OPT_P);
-      std::string LastInput = Inputs.back();
-      llvm::opt::Arg *PrevInputArg = nullptr;
-      for (llvm::opt::Arg *InputArg : Args.filtered(OPT_INPUT)) {
-        // Find Input after /P.
-        if ((PArg->getIndex() + 1) == InputArg->getIndex()) {
-          opts.Preprocess = InputArg->getValue();
-          if (LastInput == opts.Preprocess && PrevInputArg) {
-            // When InputArg is last Input, update it to other Input so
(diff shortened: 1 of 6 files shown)


lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -588,7 +588,35 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
...
+  if (Args.hasFlag(OPT_P, OPT_INVALID, false)) {
+    // Default preprocess filename is InputName.i.
+    llvm::SmallString<128> Path(Args.getLastArgValue(OPT_INPUT));
+    llvm::sys::path::replace_extension(Path, "i");
+    // Try to get preprocess filename from Fi.
+    opts.Preprocess = Args.getLastArgValue(OPT_Fi, Path).str();
+    // Hack to support fxc style /P preprocess_filename.
+    // When there're more than 1 Input file, use the input which is after /P as
+    // preprocess.
+    std::vector<std::string> Inputs = Args.getAllArgValues(OPT_INPUT);
```
pow2clk, line 600: I feel that the code below should only be run if there is no -Fi parameter. Where -P and -Fi is used such as `dxc -Fi out.pp -P in.hlsl`, I think that `in.hlsl` will get treated as the preprocessor output, which probably wasn't the intent.

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -588,7 +588,35 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
...
+        if ((PArg->getIndex() + 1) == InputArg->getIndex()) {
+          opts.Preprocess = InputArg->getValue();
+          if (LastInput == opts.Preprocess && PrevInputArg) {
+            // When InputArg is last Input, update it to other Input so
+            // Args.getLastArgValue(OPT_INPUT) get expect Input.
+            InputArg->getValues()[0] = PrevInputArg->getValues()[0];
+          }
+          break;
+        }
+        PrevInputArg = InputArg;
```
pow2clk, line 616: Overall, this is kind of ugly, but I don't have a better alternative and I like that it maintains the previous behavior. We might consider including a warning suggesting using the new interface. I realize that probably contradicts something I said before. Sorry :(

pow2clk (commented):
I think we should include a few tests in cmd tests if nowhere else. Specifically maybe
* `-Fi out.pp -P in.hlsl`
* `-P -Fi out.pp in.hlsl`
* `-P out.pp in.hlsl`
* `in.hlsl -P out.pp`

pow2clk (approved):
Looks good! Thanks!

PR #4631, opened 2022-09-01
Insert cast when types are identical to HLSL but not to clang

tex3d:
It's possible to get a value of a type `unsigned long` in HLSL, such as
the result of sizeof(), while the uint type is `unsigned int` instead.
This leads to the situation where no cast is inserted because the types
look identical when translated down to ArBasicKind, which can lead to an
assert during codegen, such as when passing sizeof() result to a function
call parameter.

This change preserves the original canonical type ptr from clang and
sets the conversion type when these differ, even when ArBasicKind is
identical.

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -708,6 +708,7 @@ BITWISE_ENUM_OPS(TYPE_CONVERSION_REMARKS)
 struct ArTypeInfo {
   ArTypeObjectKind ShapeKind;      // The shape of the type (basic, matrix, etc.)
   ArBasicKind EltKind;             // The primitive type of elements in this type.
+  const clang::Type *EltTy;        // Canonical element type ptr
   ArBasicKind ObjKind;             // The object type for this type (textures, buffers, etc.)
   UINT uRows;
   UINT uCols;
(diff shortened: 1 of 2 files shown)


pow2clk (approved):
I think this is functionally good. I made the mistake of digging into the various type retrieval helper functions and identifying some potential simplification nitpicks.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -8899,6 +8917,18 @@ static bool ConvertComponent(ArTypeInfo TargetInfo, ArTypeInfo SourceInfo,
           ComponentConversion = ICK_Floating_Integral;
       }
     }
+  } else if (TargetInfo.EltTy != SourceInfo.EltTy) {
+    // Types are identical in HLSL, but not identical in clang,
+    // such as unsigned long vs. unsigned int.
+    // Add conversion based on the type.
+    if (IS_BASIC_AINT(TargetInfo.EltKind))
```
llvm-beanz, line 8924: Is this not something the integer promotion rules should handle? I don't remember which way it goes, but in C `int` either promotes to `long` or vice versa. I think we should be able to use the C promotion rules here to get the result we need.

tex3d: My understanding is that we are overriding the type promotion with our own, so if we don't insert the cast here, nobody will, and codegen will assert due to a technical mismatch in types, even though they are equivalent.

PR #4635, opened 2022-09-03
add barycentrics ordering check onto existing barycentrics test

bob80905:
The existing barycentrics test failed to account for an extra part of the specification.
No matter in which order the vertices are passed to the pixel shader, GetAttributeAtVertex on the svid of each vertex should equal the numbered vertex that's given to the GetAttributeAtVertex function. 
So, this test was added inside of the barycentrics test, with an extra shader to accomplish the test.
If this spec is met, then the associated pixel will write a float4 of all 1.0's to the render target view. These pixels are specifically tested to see if the ordering constraint is met.


tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9001,53 +9001,135 @@ TEST_F(ExecutionTest, BarycentricsTest) {
...
+            LogCommentFmt(L"location  %u %u, value %f, %f, %f, %f", pixelX, pixelY, pPixels[offset], pPixels[offset + 1], pPixels[offset + 2], pPixels[offset + 3]);
+          
+            // If the ordering constraint is met, then this pixel's RGBA should be all 1.0's
+            // since the shader only returns float4<1.0,1.0,1.0,1.0> when this condition is met.
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset]    , 1.0, tolerance));
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset + 1], 1.0, tolerance));
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset + 2], 1.0, tolerance));
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset + 3], 1.0, tolerance));
+        }
+
```
pow2clk, line 9128: A lot of this is duplicated with the above. Making a helper function would make it easier to compare to the old code and also allow any updates to be in just one place.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9001,53 +9001,135 @@ TEST_F(ExecutionTest, BarycentricsTest) {
...
+    {
+        std::shared_ptr<ShaderOpTestResult> test = RunShaderOpTestAfterParse(pDevice, m_support, "Barycentrics", nullptr, ShaderOpSet);
+        MappedData data;
+        D3D12_RESOURCE_DESC &D = test->ShaderOp->GetResourceByName("RTarget")->Desc;
+        UINT width = (UINT)D.Width;
+        UINT height = D.Height;
+        UINT pixelSize = GetByteSizeForFormat(D.Format);
+
+        test->Test->GetReadBackData("RTarget", &data);
+        //const uint8_t *pPixels = (uint8_t *)data.data();
```
pow2clk, line 9018: commented code

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9001,53 +9001,135 @@ TEST_F(ExecutionTest, BarycentricsTest) {
...
+            };
+
+        std::shared_ptr<ShaderOpTestResult> test2 = RunShaderOpTestAfterParse(pDevice, m_support, "BarycentricsCheckOrder", ResourceCallbackFn, ShaderOpSet);
+        MappedData data;
+        D3D12_RESOURCE_DESC &D = test2->ShaderOp->GetResourceByName("RTarget")->Desc;
+        UINT width = (UINT)D.Width;
+        UINT height = D.Height;
+        UINT pixelSize = GetByteSizeForFormat(D.Format);
+        test2->Test->GetReadBackData("RTarget", &data);
+        //const uint8_t *pPixels = (uint8_t *)data.data();
```
pow2clk, line 9088: commented code.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -1539,6 +1539,66 @@
     </Shader>
   </ShaderOp>
 
+
+  <ShaderOp Name="BarycentricsCheckOrder" PS="PS61" VS="VS61">
+    <RootSignature>RootFlags(ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT)</RootSignature>
+    <Resource Name="VBuffer" Dimension="BUFFER" Width="1024" Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST" Init="ByName" ReadBack="true">
+    </Resource>
+    <Resource Name="RTarget" Dimension="TEXTURE2D" Width="64" Height="64" Format="R32G32B32A32_FLOAT" Flags="ALLOW_RENDER_TARGET" InitialResourceState="COPY_DEST" ReadBack="true" />
```
pow2clk, line 1547: If we can make the render targets the same size, you can use different entry point targets to make one ubershader and prevent duplicating the shader code above with slight modifications.

pow2clk (commented):
I still have to look more closely at the changes, but here are a few things that might help make that easier and would improve the final result as well

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9043,13 +9028,77 @@ TEST_F(ExecutionTest, BarycentricsTest) {
...
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset + 1], w1, tolerance));
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset + 2], w2, tolerance));
+        }
+        else{
+            // If the ordering constraint is met, then this pixel's RGBA should be all 1.0's
+            // since the shader only returns float4<1.0,1.0,1.0,1.0> when this condition is met.
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset]    , 1.0, tolerance));
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset + 1], 1.0, tolerance));
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset + 2], 1.0, tolerance));
+            VERIFY_IS_TRUE(CompareFloatEpsilon(pPixels[offset + 3], 1.0, tolerance));
```
pow2clk, line 9042: Am I correct that the old test required looser precision as a result of the changed dimensions?

bob80905: The old test required higher precision because the dimensions were relatively larger.
Now that we want the dimensions to be smaller, it seems the precision should be looser.

tools/clang/test/HLSL/ShaderOpArith.xml
```diff
@@ -1535,7 +1532,47 @@
...
+    float4 PSMain(PSInput input) : SV_Target {
+      // check to make sure that the pixel shader will see the barycentric weight associated
+      // with the first vertex in the x component of the SV_Barycentric vector, and likewise for subsequent vertices
+      
+      if (0 == GetAttributeAtVertex(input.svid, 0) && 
+          1 ==  GetAttributeAtVertex(input.svid, 1) &&
+          2 ==  GetAttributeAtVertex(input.svid, 2))
+      {
+        // special value of all 1's in a float4 will be returned if the ordering is maintained for this pixel.
+        return float4(1.0, 1.0, 1.0, 1.0);
```
tex3d, line 1568: An alternative way to do this (without the `if`) is:
````
return float4(float3(GetAttributeAtVertex(input.svid, 0),
                     GetAttributeAtVertex(input.svid, 1),
                     GetAttributeAtVertex(input.svid, 2)) * 0.5, 1.0)
```
then change the expected result to (0, 0.5, 1.0, 1.0) which makes out-of-order failure case reveal the ordering used in the result.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -9025,33 +9010,106 @@ TEST_F(ExecutionTest, BarycentricsTest) {
...
         float w0 = barycentricWeights[i].x;
         float w1 = barycentricWeights[i].y;
         float w2 = barycentricWeights[i].z;
         float x1 = w0 * p0.x + w1 * p1.x + w2 * p2.x;
         float y1 = w0 * p0.y + w1 * p1.y + w2 * p2.y;
         // map from x1 y1 to rtv pixels
-        int pixelX = (int)((x1 + 1) * (width - 1) / 2);
-        int pixelY = (int)((1 - y1) * (height - 1) / 2);
+        int pixelX = (int)round((x1 + 1) * (width - 1) / 2);
+        int pixelY = (int)round((1 - y1) * (height - 1) / 2); //1.3332 * 63 = 83.999, /2 = 41.999 truncated = 41
```
pow2clk, line 9031: This is going to truncate before it rounds. You'll need to change the `/ 2` to `/ 2.0` or `/ 2f`

I think the comment is obsolete.

pow2clk (commented):
I'm curious if the change to the round() makes any difference in the required precision.

I think tex's suggestion is a good one.

tex3d (approved):
LGTM.

pow2clk (approved):
Looks good! Thanks for sticking with this!

tools/clang/unittests/HLSL/ExecutionTest.cpp
pow2clk: Tex points out that the single floats in this equation would have made the whole of it floating point. Even so, I like to have the divisor make which kind of division is intended clear.

PR #4638, opened 2022-09-05
[SPIR-V] add support for extension KHR_fragment_shading_barycentric (SV_Barycentrics)

ShchchowAMD:
* [SPIR-V] Add support for SPV_KHR_fragment_shading_barycentrics
    (SV_Barycentrics)

  Implement latest `SV_Barycentrics` semantics in DXC, according to DXC wiki spec.

  For variant qualifier decorated baryCoord inputs, as KHR extension has no matched built-in, still mapped to Bary*AMD.

  Add support for using input as first parameter of GetAttributeAtVertex. This would expand input's dimension to meet extension spec.

  Merge some features from AMD_shader_explicit_parameter and current extension together for MSAA usage. Keep generated SPIR-V codes similar to GLSL as possible.

  With concern to HLSL/DXIL part, keep most type modification/variable decoration change in SPIRV trans unit.

Ref:
https://github.com/microsoft/DirectXShaderCompiler/wiki/SV_Barycentrics


docs/SPIR-V.rst
```diff
@@ -1541,7 +1542,8 @@ some system-value (SV) semantic strings will be translated into SPIR-V
 +---------------------------+-------------+----------------------------------------+-----------------------+-----------------------------+
 | SV_StencilRef             | PSOut       | ``FragStencilRefEXT``                  | N/A                   | ``StencilExportEXT``        |
 +---------------------------+-------------+----------------------------------------+-----------------------+-----------------------------+
-| SV_Barycentrics           | PSIn        | ``BaryCoord*AMD``                      | N/A                   | ``Shader``                  |
+| SV_Barycentrics           | PSIn        | ``BaryCoord*KHR`` for linear interp    | N/A                   | ``FragmentBarycentricKHR``  |
+|                           |             | ``BaryCoord*AMD`` for other types      |                       |                             |
```
Tobski, line 1546: What are the "other types" here - did we add extra decorations for AMD interpolation modes, or are we translating core HLSL qualifiers to the AMD extension still? The KHR should support everything that HLSL does so we should be translating to that always unless we added extra bits for our extension.

cassiebeckley (dismissed):

cassiebeckley:
In addition, this PR generates invalid SPIR-V for some inputs. For example, running `dxc -T ps_6_1 -E main foo.hlsl -spirv` on

```hlsl
enum VertexID {
    FIRST = 0,
    SECOND = 1,
    THIRD = 2
};


float3 main( float3 vBaryWeights : SV_Barycentrics,
    nointerpolation float3 Color : COLOR ) : SV_Target
{
    float3 vColor = GetAttributeAtVertex( Color, VertexID::SECOND );
    return Color * vColor.x;
}
```

returns the error message

```
fatal error: generated SPIR-V is invalid: Expected vector operand type to be equal to Result Type: VectorTimesScalar
  %17 = OpVectorTimesScalar %v3float %14 %16

```

tools/clang/test/CodeGenSPIRV/intrinsics.get-attribute-at-vertex.hlsl
```diff
@@ -1,25 +1,30 @@
 // RUN: %dxc -T ps_6_1 -E main
 
-struct PSInput {
-  float4 position : SV_POSITION;
-  nointerpolation float3 color : COLOR;
```
cassiebeckley, line 5: Can you also include a test making sure that having inputs in a struct still works?

docs/SPIR-V.rst
ShchchowAMD: I will reword comment here.
The KHR extension adds two builtins for center_interp: BaryCoordKHR and BaryCoordNoPerspKHR.

From recent discussion, we need to support centroid_interp and sample_interp for MSAA too.
As KHR ext doesn't add extra builtins for this feature, I just want to keep existed AMD* builtins to meet the requirement.

tools/clang/include/clang/AST/ASTContext.h
```diff
@@ -909,6 +911,10 @@ class ASTContext : public RefCountedBase<ASTContext> {
   // HLSL Change Starts
   /// \brief Retrieve the declaration for HLSL string type.
   TypedefDecl *getHLSLStringTypedef() const;
+  /// \brief Record usage of input in GetAttributeAtVertex for lately
+  /// modification to spv variable decl.
```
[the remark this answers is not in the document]
ShchchowAMD: I'll fix in next commit, thanks.

tools/clang/lib/AST/ASTContext.cpp
```diff
@@ -1208,6 +1208,28 @@ ASTContext::getInstantiatedFromUsingDecl(UsingDecl *UUD) {
...
+}
+
+void
+ASTContext::recordBaryCoordInputLoc(SourceLocation srcLoc) {
+  for (unsigned i = 0; i < GetAttribAtVertCallInputLoc.size(); i++) {
+    if (GetAttribAtVertCallInputLoc[i] == srcLoc) {
+      return;
+    }
+  }
+  GetAttribAtVertCallInputLoc.push_back(srcLoc);
```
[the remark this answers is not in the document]
ShchchowAMD: I'll fix in next commit, thanks.

tools/clang/lib/SPIRV/CapabilityVisitor.cpp
```diff
@@ -357,15 +357,18 @@ bool CapabilityVisitor::visit(SpirvDecoration *decor) {
...
-    case spv::BuiltIn::BaryCoordNoPerspAMD:
     case spv::BuiltIn::BaryCoordNoPerspCentroidAMD:
     case spv::BuiltIn::BaryCoordNoPerspSampleAMD:
-    case spv::BuiltIn::BaryCoordSmoothAMD:
     case spv::BuiltIn::BaryCoordSmoothCentroidAMD:
-    case spv::BuiltIn::BaryCoordSmoothSampleAMD:
-    case spv::BuiltIn::BaryCoordPullModelAMD: {
+    case spv::BuiltIn::BaryCoordSmoothSampleAMD: {
       addExtension(Extension::AMD_shader_explicit_vertex_parameter,
                    "SV_Barycentrics", loc);
```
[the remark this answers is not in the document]
ShchchowAMD: I was trying to replace AMD extension with KHR before. But as I commented in Tobski's review, we still need those builtins for non-center interpolation usage:

1. For spv validation part, those *AMD builtins will need AMD_shader_explicit_vertex_parameter extension decl to pass it. I thought that would be better to keep this check in validation step so the extension decl is kept here.

2. For current implementation, they are all for KHR extension in HLSL->SPIRV translation. So yes, I keep those cases fall through to add both extensions here in generated SPV code.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11172,6 +11176,37 @@ void SpirvEmitter::processMeshOutputCounts(const CallExpr *callExpr) {
...
+
+  // Decorate input arg0
+  declIdMapper.decoratePerVertexKHR(arg0NamedDecl, *vectexId);
+
+  // Change to access chain instr
+  const auto arg0Type = (dyn_cast<VarDecl>(arg0ValDecl))->getType()->getAsArrayTypeUnsafe()->getElementType();
+  auto *arg0BaseExpr = doExpr(arg0);
+  auto *accessChainPtr = spvBuilder.createAccessChain(
+      arg0Type, arg0BaseExpr,
+      {spvBuilder.getConstantInt(astContext.UnsignedIntTy,
```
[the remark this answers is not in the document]
ShchchowAMD: Just try to keep same code style with other parts in DXC, should I remove it?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11172,6 +11176,37 @@ void SpirvEmitter::processMeshOutputCounts(const CallExpr *callExpr) {
...
+
+  // arg0
+  const auto *arg0 = expr->getArg(0)->IgnoreParenLValueCasts();
+  const auto *arg0ValDecl = (dyn_cast<DeclRefExpr>(arg0))->getDecl();
+  const auto *arg0NamedDecl = dyn_cast<NamedDecl>(arg0ValDecl);
+
+  // arg1
+  const auto *arg1BaseExpr = doExpr(expr->getArg(1)->IgnoreParenLValueCasts());
+  const auto *arg1ConstExpr = dyn_cast<SpirvConstantInteger>(arg1BaseExpr);
+  const auto *vectexId = arg1ConstExpr->getValue().getRawData();
```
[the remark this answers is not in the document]
ShchchowAMD: I'll fix in next commit, thanks.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -12230,13 +12265,16 @@ bool SpirvEmitter::emitEntryFunctionWrapper(const FunctionDecl *decl,
...
-  for (const auto *param : decl->params()) {
+  for (auto *param : decl->params()) {
     const auto paramType = param->getType();
     std::string tempVarName = "param.var." + param->getNameAsString();
     auto *tempVar =
         spvBuilder.addFnVar(paramType, param->getLocation(), tempVarName,
                             param->hasAttr<HLSLPreciseAttr>());
 
+    if (paramType != tempVar->getAstResultType())
+        param->setType(tempVar->getAstResultType());
```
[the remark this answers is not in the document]
ShchchowAMD: In spvBuilder.addFnVar call here:
```
        spvBuilder.addFnVar(paramType, param->getLocation(), tempVarName,
                            param->hasAttr<HLSLPreciseAttr>());
```
I add code for pervertexExt decorated input (COLOR).
This input will be changed to a three-element array according to extension spec (when builtin function GetAttributeAtVertex has been detected before).
```
if (astContext->validateBaryCoordInputLoc(loc)) {
      if (!valueType->isArrayType())
      {
        QualType qtype = astContext->getConstantArrayType(
            valueType, llvm::APInt(32, 3), clang::ArrayType::Normal, 0);
        var = new (context) SpirvVariable(
              qtype, loc, spv::StorageClass::Function, isPrecise, init);
      }
      else
        var = new (context) SpirvVariable(
            valueType, loc, spv::StorageClass::Function, isPrecise, init);
    }
```
Here is to change param type as same as generated spv variable.
Otherwise, we will get a type mismatched error later (OpLoad for function parameter).

ShchchowAMD:
> In addition, this PR generates invalid SPIR-V for some inputs. For example, running `dxc -T ps_6_1 -E main foo.hlsl -spirv` on
> 
> ```hlsl
> enum VertexID {
>     FIRST = 0,
>     SECOND = 1,
>     THIRD = 2
> };
> 
> 
> float3 main( float3 vBaryWeights : SV_Barycentrics,
>     nointerpolation float3 Color : COLOR ) : SV_Target
> {
>     float3 vColor = GetAttributeAtVertex( Color, VertexID::SECOND );
>     return Color * vColor.x;
> }
> ```
> 
> returns the error message
> 
> ```
> fatal error: generated SPIR-V is invalid: Expected vector operand type to be equal to Result Type: VectorTimesScalar
>   %17 = OpVectorTimesScalar %v3float %14 %16
> 
> note: please file a bug report on https://github.com/Microsoft/DirectXShaderCompiler/issues with source code if possible
> ```

I do have concerns here too.

For current PR, we need to change input COLOR to an three-element array when generating SPIR-V code. The purpose is to set new decorator and change its type to meet new feature.

I tried to keep its influence scope small so this would only happen when COLOR is used as first parameter of GetAttributeAtVertex.

Which means, it would be correct when you use either GetAttributeAtVertex or Color here within a shader. But it will report error when you use them together, as the input has been changed to an array.

It is assumed here when you are using GetAttributeAtVertex, input COLOR is treated as pervertex data array.

tools/clang/test/CodeGenSPIRV/intrinsics.get-attribute-at-vertex.hlsl
ShchchowAMD: I'll have a try.

AntonTokarev:
Any updates on this? Missing this feature :)

ShchchowAMD:
> Any updates on this? Missing this feature :)

I'm waiting for review now. You can have a try with this PR, main features should have been finished.

ShchchowAMD:
Fixed conflict in spvEmitter.
Add comment to SV_BaryCentrics builtin cases in CapatibilityVisitor.

Tobski:
@s-perron looks like this was assigned to you for review, are you able to take a look?

@ShchchowAMD can you rebase this again please?

ShchchowAMD:
> @s-perron looks like this was assigned to you for review, are you able to take a look?
> 
> @ShchchowAMD can you rebase this again please?

I'll resolve the conflict today. Thanks.

ShchchowAMD:
Conflict resolved.

s-perron (approved):

ShchchowAMD:
@cassiebeckley  Hi cassie, current PR still needs your approval as there's a 'change requested' step from last review. Is it okay to merge with current impl?

ShchchowAMD:
@cassiebeckley looks like there's still one 'change requested' needs your approval. Would you please help to verify and let me merge this PR?

someonesomewheredev:
Hi, any updates on this?

llvm-beanz:
I'm a bit concerned about the implementation and testing here.

We have parsing changes, with no testing specifically focused on the parser. It looks like we're hooking into the parser to run call diagnostics, which... seems wrong. This also introduces a stateful-ness in the Sema object...

I'm going to need at least a day to wrap my head around what this change is trying to do.

llvm-beanz (changes requested):
I think there is a design problem with how this change is implemented, and I'm sorry this is just coming to my attention now.

The AST context and Sema objects shouldn't be growing state. In both cases, context-specific state should be stored in the AST where it can be inspected, tested, and verified.

The underlying decls for barrycentrics parameters have a SemanticDecl for SV_Barycentrics attached, so you should always be able to look at the AST and verify the semantic annotations.

I'll try to make some more detailed specific notes, but few general guidelines around Clang's architecture:

* Changes in the parser should produce AST nodes (or diagnostics)
* Changes in Sema should produce diagnostics or mutate the AST
* Changes in CodeGen should _not_ produce diagnostics... by this point we should have a valid AST.

DXC's modifications to clang don't always stick to this, but we're working to fix that so we shouldn't make the problems worse.

llvm-beanz (commented):
Unless I'm misunderstanding something (and there's a lot of code here so I'm probably misunderstanding a lot of things).

I think all the code in the AST, Parse and Sema layers is unnecessary. In the SPIRV code you should be able to look at the Decls to see if barycentric annotations are present and "do the right thing" rather than relying on the ASTContext or Sema storing any data.

Tobski:
@ShchchowAMD in case you haven't seen this, can you look at/respond to @llvm-beanz's comments?

ShchchowAMD:
@Tobski  @llvm-beanz  Sorry for the late response.

I'll refine current implementation and try to avoid problems beanz mentioned above. It may take some time.

Thanks very much for all the advice and notes beanz shared here. I'll try to keep new changes following those points and avoid introducing new issues in above modules.

BTW, in case I misunderstanding, should I try to keep most of changes in SPV level instead of storing data or pass info to AST and Sema?
The reason of using current solution is that:

1. I found I must change type of first parameter (PervertexKHR decorated) to an **array** to meet spec for GetAttributeAt.

2. (If I understand right, I assumed that we will use **KHR** prefix built-ins to **replace** **AMD** built-ins after this change).

As an example, in test file semantic.barycentrics.ps.np-c.hlsl we could found this:

```
float4 main(noperspective centroid float3 bary : SV_Barycentrics) : SV_Target {
    return float4(bary, 1.0);
}
```
If bary is also used as the first parameter of **GetAttributeAt**, it should be an 'array', according to spec.

We could change type of 'bary' to an array in SPV level but this would against the 'return' statement (it is treated as a float3, non-array).

I think that's the reason for current implementation (I agree that I should not add too much things in other modules) to avoid spv level errors.

I'll try to find a new way to avoid those problems, my current idea is to add a new temp array input for this issue but not make sure this is correct. Any ideas about this?

llvm-beanz:
> BTW, in case I misunderstanding, should I try to keep most of changes in SPV level instead of storing data or pass info to AST and Sema?

The point is that if you need to store data, store it _in the AST_ not in the ASTContext or Sema objects. Those objects shouldn't store semantic information about the program.

> 1. I found I must change type of first parameter (PervertexKHR decorated) to an **array** to meet spec for GetAttributeAt.

I'm unclear what exactly you mean here. GetAttributeAtVertex isn't defined at the source level to take an array, so it shouldn't take an array in the AST. If you need to convert that to something else in CodeGen, that's a code generation change, not an AST/Sema change.

> We could change type of 'bary' to an array in SPV level but this would against the 'return' statement (it is treated as a float3, non-array).

I don't know that I understand what you mean here. Can you provide an example?

This code is valid in HLSL and doesn't require changing the return type of the function to an array:

```c++
int4 main(nointerpolation float4 col : COLOR) : SV_Target {
    return GetAttributeAtVertex(col, 2);
}
```

I'm unclear what the SPIR-V issue is, and this is making me concerned that maybe there is a larger design issue with Vulkan barycentrics we need to address.

ShchchowAMD:
> The point is that if you need to store data, store it _in the AST_ not in the ASTContext or Sema objects. Those objects shouldn't store semantic information about the program.

Thanks, I will refine this part.

> I'm unclear what exactly you mean here. GetAttributeAtVertex isn't defined at the source level to take an array, so it shouldn't take an array in the AST. If you need to convert that to something else in CodeGen, that's a code generation change, not an AST/Sema change.

Understand, I'll check whether I could do current changes in CodeGen.

> 
> I don't know that I understand what you mean here. Can you provide an example?
> 
> This code is valid in HLSL and doesn't require changing the return type of the function to an array:
> 
> ```c++
> int4 main(nointerpolation float4 col : COLOR) : SV_Target {
>     return GetAttributeAtVertex(col, 2);
> }
> ```
> 
> I'm unclear what the SPIR-V issue is, and this is making me concerned that maybe there is a larger design issue with Vulkan barycentrics we need to address.

I mean, as an example:

```

float4 main(nointerpolation float4 col : COLOR) : SV_Target {
     return GetAttributeAtVertex(col, 2) + col;
}
```

For PerVertexKHR decorated input 'col':

`PerVertexKHR, which indicates that a fragment shader input will not have interpolated values, but instead must be accessed with an extra array index that identifies one of the vertices of the primitive producing the fragment`

(https://github.com/KhronosGroup/Vulkan-Docs/blob/main/proposals/VK_KHR_fragment_shader_barycentric.adoc)

I mean, when generating spirv code, I tried to add an extra array dimension to 'col' and implemented GetAttributeAtVertex as an array indexing op:

`GetAttributeAtVertex(col, 2)  ==>  col[2]`

Then it would report an error if:

1. I change the declared type of 'col' from **float4** to an **float4 array** to meet above requirement.

2. 'col' will also be used in OpAdd later as a float4 here. But its type has been changed to an array.


I don't make sure whether above example makes it clear.

The point is above extension spec expects input variable to have an extra dimension for access by an index.

I think I'm confused here that whether I should treat 'col' as an array in spirv level but still allow users to use it as a float vector in HLSL.

ShchchowAMD:
I'll double check my codes.
I was trying to redeclare 'col' only when I met a GetAttributeAt during parsing.
This should be done in another way.

llvm-beanz:
> I mean, when generating spirv code, I tried to add an extra array dimension to 'col' and implemented GetAttributeAtVertex as an array indexing op:
> 
> `GetAttributeAtVertex(col, 2) ==> col[2]`

You can't do this transformation on the AST. Doing that would break the language rules.

> I think I'm confused here that whether I should treat 'col' as an array in spirv level but still allow users to use it as a float vector in HLSL.

`col` needs to be usable as a `float4` throughout the program, I'm not sure treating it as an array at any point makes sense. If SPIR-V requires that barycentrics are represented as an array we likely need a bigger discussion about the design.

ShchchowAMD:
> You can't do this transformation on the AST. Doing that would break the language rules.

Yes, I agree as GetAttributAt has a param list as (floatVector, int) in source level. We should not change this.
 
> > I think I'm confused here that whether I should treat 'col' as an array in spirv level but still allow users to use it as a float vector in HLSL.
> 
> `col` needs to be usable as a `float4` throughout the program, I'm not sure treating it as an array at any point makes sense. If SPIR-V requires that barycentrics are represented as an array we likely need a bigger discussion about the design.

I'll discuss with other developers and update here later.

I don't make sure whether current detailed design and my understanding is expected, but I'm glad that above example code seems to make my concern more clear.

Here is the discussion points I would share with other guys, please let me know if any of them looks confused:

1. We need to make generated spirv codes from both GLSL and HLSL as similar as possible for this extension.

2. If HLSL keeps first parameter as a float3, we need to double check whether we could impl GetAttributAt as an indexing op.

3. If we make sure we would use new extension to refine definition of SV_BaryCentrics, is there any chance we could change definition of GetAtAttribute in HLSL, or PerVertex input in GLSL, only when generating spv, to avoid this problem.

I just checked my code today and found if we could determine col's type, no matter float3 or array, I could avoid current paradox and also remove hooking points in AST, Sema and Parser.

llvm-beanz:
> 1. We need to make generated spirv codes from both GLSL and HLSL as similar as possible for this extension.

This is not a general goal for the HLSL compiler. We need to generate valid SPIR-V, but I don't see why similarity between HLSL and GLSL code generation is a goal.

> 2. If HLSL keeps first parameter as a float3, we need to double check whether we could impl GetAttributAt as an indexing op.

This shipped in HLSL with Shader Model 6.1. Changing it isn't possible without connecting the change to a new shader model, so I don't think this is an actionable path for HLSL. That doesn't mean that SPIR-V can't CodeGen to something different, but the language can't change here.

> 3. If we make sure we would use new extension to refine definition of SV_BaryCentrics, is there any chance we could change definition of GetAtAttribute in HLSL, or PerVertex input in GLSL, only when generating spv, to avoid this problem.

You cannot change the definition of `SV_Barycentrics` without causing source compatibility issues. In general we've avoided having general HLSL constructs (like `SV_Barycentrics`) mean something different or be used significantly differently under SPIR-V because part of the goal of having HLSL target SPIR-V is to be able to use the same code when targeting DirectX or Vulkan.

> I just checked my code today and found if we could determine col's type, no matter float3 or array, I could avoid current paradox and also remove hooking points in AST, Sema and Parser.

Yea, I didn't think from reading your initial PR that you needed language changes here, or even that you needed to fundamentally change what you were doing. My thought was that you just needed to not store data in the ASTContext and Sema objects and instead traverse the AST up to the root decl in SPIR-V generation. That would give you the type and ability to verify the barycentrics semantic annotation during code generation and you can generate the conversions you need there.

Where I'm concerned about the language issue is that if a parameter annotated with `SV_Barycentrics` is assumed to be an array, what happens when it is used not as an array? Is that implicitly using the first element of the array, or does it do something different?

In HLSL under DirectX the parameter is a vector and can be used as a vector.

ShchchowAMD:
Thanks for your quick response and detailed explanation.
I'll refine current code according to your advice, removing those hook points and make a new commit later.

For the implicit type issue, it may take me some time to understand your concern for this. Still I fully agree with that we should try to avoid this situation.

s-perron:
I have been silently following this. I think I understand the problem. It might be that HLSL seems to allow a use of `nointerpolation` that is hard (impossible?) to represent in SPIR-V.

Going back to the example above:

```
float4 main(nointerpolation float4 col : COLOR) : SV_Target {
     return GetAttributeAtVertex(col, 2) + col;
}
```

You can see the DXIL [here](url). It seems like the use of `col` in the RHS of the addition will be interpolated. However, `col` is not interpolated when its value is retrieved using `GetAttributeAtVertex`.

The problem in SPIR-V is that a non-interpolated value is passed to the shader as an array, and the array is indexed to get the value at the desired vertex. The interpolated value is simply not available in SPIR-V. Maybe there is a way to calculate it, but it could get tricky in some cases.

If it was not for this problem, generating the SPIR-V should be easy. Any variable declaration marked "nointerpolation" should generate an array with the `PerVertexKHR` decoration. Assuming all accesses are done using `GetAttributeAtVertex`, this can be easily turned into an array access. 

Is that correct?

ShchchowAMD:
Yes, exactly! I can't easily determine its type due to some tricky cases.

ShchchowAMD:
Update: discussed with other developer, seems 'noninterpolation' decorated input here could only be used in GetAttributAtVertex. Hence above problem would not exist. I'll refine my code. Thanks.

s-perron:
> Update: discussed with other developer, seems 'noninterpolation' decorated input here could only be used in GetAttributAtVertex. Hence above problem would not exist. I'll refine my code. Thanks.

@llvm-beanz What is supposed to happen in HLSL when a `noninterpolation` is used outside of `GetAttributeAtVertex`? In any case, for SPIR-V we will probably want to issue an error for that situation.

I see two separate PRs coming out of this.

1. Implement change in the SPIR-V backend to generate correct code assuming that `noninterpolation` inputs are only used in `GetAttributeAtVertex`. This should not have to change the parsing or the AST. The design for this seems pretty straight forward.
2. We may want to issue an error when a `noninterpolation` input is used outside of `GetAttributeAtVertex` and the compiler is generating SPIR-V. Depending on Chris's answer above, we can see what will happen on the DXIL path.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -12601,13 +12710,20 @@ bool SpirvEmitter::emitEntryFunctionWrapper(const FunctionDecl *decl,
 
   // Create temporary variables for holding function call arguments
   llvm::SmallVector<SpirvInstruction *, 4> params;
-  for (const auto *param : decl->params()) {
-    const auto paramType = param->getType();
+  for (auto *param : decl->params()) {
+    QualType paramType = param->getType();
+    if (param->hasAttr<HLSLNoInterpolationAttr>() ||
+        paramType->isStructureType()) {
```
s-perron, line 12716: The function `redeclSpecialVarType` seems to have an odd interface, which make the code harder to read.  If the type is a struct, then you call `redeclSpecialVarType`, and it will check for the attribute.  If the type is not a struct and not an array, then it does not look for the attribute. The use of `redeclSpecialVarType` should be consistent. 

I think you should change `redeclSpecialVarType` it check for the attribute and remove this check.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -1251,6 +1251,41 @@ SpirvInstruction *SpirvEmitter::loadIfAliasVarRef(const Expr *expr,
   return instr;
 }
 
+QualType SpirvEmitter::redeclSpecialVarType(QualType type,
```
s-perron, line 1254: `redecl` is not a word or a common abbreviation. Also "redeclare special variable type" is so vague it is hard to know what it is doing at the call sights.  A better name might be `expandNoInterpolationParamToArray`. Personally, I do not have a problem with long names. Good IDEs can save you from typing it in all of the time.

s-perron (commented):
Generally look go to me. I have some minor issues with readability. Thanks for making the changes.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
ShchchowAMD: OK. Thanks.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
ShchchowAMD: I'll move check point inside this function.

llvm-beanz:
Please avoid force-pushing. Git doesn't track force pushes well so comments lose their attachment and it makes it much harder to review.

The messy history will be cleaned up by squashing on merge. If you need to update from main a merge should suffice rather than rebase.

ShchchowAMD:
> Please avoid force-pushing. Git doesn't track force pushes well so comments lose their attachment and it makes it much harder to review.
> 
> The messy history will be cleaned up by squashing on merge. If you need to update from main a merge should suffice rather than rebase.

Will follow your advice and avoid force pushing.

Changes are kept in SPIRV translation scope according to review. Any suggestions before this could be approved?

s-perron (approved):

ShchchowAMD:
@llvm-beanz  Hi Chirs, would this PR be good for you. I do need your approval now.

llvm-beanz (approved):
Yep, this looks reasonable to me now.

ShchchowAMD:
Thanks.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11539,6 +11592,66 @@ void SpirvEmitter::processMeshOutputCounts(const CallExpr *callExpr) {
...
+
+  if (dyn_cast<CastExpr>(expr->getArg(0)->IgnoreParenLValueCasts())) {
+    // If stage inputs is boolean type, it has been cast to uint type.
+    hasImplicitCast = true;
+  }
+
+  if (dyn_cast<MemberExpr>(arg0NoCast)) {
+    // As a structure field
+    const auto *arg0NamedDecl = (dyn_cast<MemberExpr>(arg0NoCast))->getFoundDecl().getDecl();
+    arg0Type = dyn_cast<ValueDecl>(arg0NamedDecl)->getType()->getAsArrayTypeUnsafe()->getElementType();
```
Keenuts, line 11620: segmentation fault if the nointerpolation attribute is used on the struct:
```hlsl
enum VertexID {
    FIRST = 0,
    SECOND = 1,
    THIRD = 2
};

struct PSInput
{
    //float4 position : SV_POSITION;
    float3 color1 : COLOR1;
    bool3 color2 : COLOR2;
};

float3 main(nointerpolation PSInput input) : SV_Target
{
    float3 vColor0 = GetAttributeAtVertex( input.color1, VertexID::SECOND );
    float3 vColor1 = GetAttributeAtVertex( input.color2, VertexID::THIRD );
    return (vColor0 + vColor1);
}
```

Keenuts: Same crash if I remove the nointerpolation altogether:
```hlsl
enum VertexID {
    FIRST = 0,
    SECOND = 1,
    THIRD = 2
};

struct PSInput
{
    float3 color1 : COLOR1;
    bool3 color2 : COLOR2;
};

float3 main(float4 position : SV_POSITION, PSInput input) : SV_Target
{
    float3 vColor0 = GetAttributeAtVertex(input.color1, VertexID::SECOND );
    float3 vColor1 = GetAttributeAtVertex(input.color2, VertexID::THIRD );
    return (vColor0 + vColor1);
}
```

For the nointerpolation on the struct, I believe it should be valid?
For the second case, seems like there should be an error as I pass an attribute which is not marked as "nointerpolation" no?

Keenuts (commented):
Overall LGTM, just a few issues I believe?

Keenuts:
Also, from the spec, AFAIK we can have only 2 inputs with some SV_Barycentrics semantic:
```
he nointerpolation interpolation modifier is disallowed for barycentric attributes. As an exception to the current attribute specification rules, it is allowed for the pixel shader to declare at most two input attributes with the SV_Barycentrics system semantics, provided that one declaration uses perspective-correct interpolation type, and the other uses noperspective(affine, screen-space) interpolation type. These two system semantics should have different indices with value either 0 or 1.
```

So shouldn't this shader be refused?

```hlsl
enum VertexID {
    FIRST = 0,
    SECOND = 1,
    THIRD = 2
};

struct PSInput
{
    float3 color1 : COLOR1;
    bool3 color2 : COLOR2;
};

float3 main(float4 position : SV_POSITION,
            float3 bary0 : SV_Barycentrics0,
            sample float3 bary1 : SV_Barycentrics1,
            noperspective sample float3 bary : SV_Barycentrics) : SV_Target
{
    return bary + bary0 + bary1;
}
```

Right now, the code compiles, and the `bary0` and `bary1` inputs are merged into one input with both the `BaryCoordKHR` and `Sample` decorator.

Keenuts:
> @llvm-beanz What is supposed to happen in HLSL when a noninterpolation is used outside of GetAttributeAtVertex? In any case, for SPIR-V we will probably want to issue an error for that situation.

Might have missed it but seems like this case is not resolved?

Keenuts:
Btw, also noticed this behavior:

```hlsl
enum VertexID {
    FIRST = 0,
    SECOND = 1,
    THIRD = 2
};

struct PSInput
{
    float3 color1 : COLOR1;
    bool3 color2 : COLOR2;
};

float3 main(float4 position : SV_POSITION,
            nointerpolation float3 color : COLOR,
            nointerpolation float3 bary : SV_Barycentrics0) : SV_Target
{
    float3 a = GetAttributeAtVertex(bary, VertexID::SECOND);
    float3 b = GetAttributeAtVertex(bary, VertexID::THIRD);
    return a + b;
}
```

```
fatal error: generated SPIR-V is invalid: Expected arithmetic operands to be of Result Type: FAdd operand index 2
  %17 = OpFAdd %v3float %14 %16
```

Not sure what a not-interpolated barycentric coordinate would mean in that case, but the error seems undesirable. Maybe a diagnostic should be emitted in that specific case?

Keenuts (dismissed):
marking as request change for the bugs above.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
ShchchowAMD: For case 1, I think I miss this case, would add a new commit for decorated struct.
For case 2, yes, need to report error message for this.

ShchchowAMD:
1. Will add check points and error reports for cases mentioned above.

2. Will add missing part for decorated struct input.

> Might have missed it but seems like this case is not resolved?

Now I assume that this case is invalid, noninterpolation decorated inputs should not be used outside GetAttributeAtVertex. Maybe I miss it, I didn't find description for this case in spec.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
ShchchowAMD: Case 1 : supported.
Case 2 : errors could be observed when running dxc:

```
enum VertexID { 
    FIRST = 0,
    SECOND = 1,
    THIRD = 2
};

struct PSInput
{
    float3 color1 : COLOR1;
    bool3 color2 : COLOR2;
};

float3 main(PSInput input,
            bool3 color2 : COLOR2) : SV_Target
{
    float3 vColor0 = GetAttributeAtVertex( input.color1, VertexID::SECOND );
    float3 vColor1 = GetAttributeAtVertex( color2, VertexID::SECOND );
    return vColor0 + vColor0;
}
12:12: error: First parameter of GetAttributeAtVertex should be decorated with 'nointerpolation'.
17:19: error: First parameter of GetAttributeAtVertex should be decorated with 'nointerpolation'.
```

ShchchowAMD:
> Btw, also noticed this behavior:
> ...
> Not sure what a not-interpolated barycentric coordinate would mean in that case, but the error seems undesirable. Maybe a diagnostic should be emitted in that specific case? (DXIL seems to output a nice `SV_Barycentrics cannot be used with 'nointerpolation' type.`

Fixed.

s-perron:
I'm going to merge this. If @Keenuts has any more comments, we can follow up in a new PR. This one is big, and it is getting hard to track the handling of all of the corner cases.

mklefrancois:
Is there a chance of seeing it coming back? Without this, I don't see how to get access to the built-in when SPV_AMD_shader_explicit_vertex_parameter isn't available, but VK_NV_fragment_shader_barycentric is.

ShchchowAMD:
> Is there a chance of seeing it coming back? Without this, I don't see how to get access to the built-in when SPV_AMD_shader_explicit_vertex_parameter isn't available, but VK_NV_fragment_shader_barycentric is.

Sorry for the late response, will make a PR this week.

PR #4641, opened 2022-09-08
Fix version string in metadata, support optional toolname override

tex3d:
Version string in metadata is no longer the old clang string or stale HLSL_FIXED_VER.  Now it's derived from the defines in dxcversion.inc that's generated by gen_version.py or copied from fixed version file.

Addition of toolname in latest-release.json allows for this to be changed for every build out of a branch or project that sets this.  It defaults to "dxcoob" which matches the dxc "Out Of Box" name used for releases before.

The fixed version file can also define the HLSL_TOOL_NAME to override this.


tools/clang/lib/Basic/Version.cpp
```diff
@@ -136,48 +138,57 @@ std::string getClangFullRepositoryVersion() {
...
 #endif
+
+// HLSL Change Starts
+#ifdef HLSL_TOOL_NAME
+  OS << HLSL_TOOL_NAME << " ";
+#endif
+#ifdef RC_FILE_VERSION
+  OS << RC_VERSION_FIELD_1 << "." << RC_VERSION_FIELD_2 << "."
+     << RC_VERSION_FIELD_3 << "." << RC_VERSION_FIELD_4;
+#else
```
hekota, line 186: Just curious - why did you choose to use the individual RC_VERSION_FIELD* values instead of RC_FILE_VERSION?

hekota (approved):
LGTM!

tools/clang/lib/Basic/Version.cpp
tex3d: That's a good question.  I originally was going to use RC_FILE_VERSION, then was thinking this version should probably be a truncated one without the commit count, then added the commit count back anyway, thinking we could have a discussion about what to include here.

tools/clang/test/HLSLFileCheck/hlsl/resource_binding/resource_cleanup.hlsl
```diff
@@ -3,6 +3,7 @@
 // Make sure only one cbuffer is emitted for the final
 // dxil.
 
+// CHECK-LABEL: ; Buffer Definitions:
```
dmpots, line 6: Why this change? Seems unrelated to the PR.

Can we get a test to verify the new/expected behavior?

dmpots (approved):

tools/clang/test/HLSLFileCheck/hlsl/resource_binding/resource_cleanup.hlsl
tex3d: This randomly matched "cb1" in the shader hash printed in disassembly causing a false-positive test failure because has `CHECK-NOT: cb1` to make sure there is only one constant buffer in the table.  This label makes sure it skips the hash and starts from the buffer definitions when making sure "cb1" isn't present.

tex3d: Regarding a test to verify new/expected behavior, if you're referring to the version being in the IR, that's going to be a bit hard to test, since the version will depend on build settings and whatnot.  I'm not sure how to test that reliably.

hekota (approved):
LGTM!

PR #4642, opened 2022-09-09
Remove lookup of full paths of D3D12 libraries

hekota:
Let the build system find the libraries based on the Windows SDK used by cmake.
Enables use of Windows SDK from nuget packages.


tex3d (approved):
LGTM!

PR #4643, opened 2022-09-12
Move HLSL shift behavior to frontend

llvm-beanz:
The HLSL bit shift behavior was implemented in constant folding which alters the semantics of the IR. Alternatively we can implement it in the front end, leveraging the existing OpenCL code paths.

This results in the same final IR optimization and allows reverting the ConstantFolder and IR tests back to match LLVM 3.7.

The HLSL shift behavior's optimized behavior is verified in existing tests like the shift-fold.hlsl test, which is unimpacted by this change.

The big motiviation for this change is heading off subtle bugs. All our tests cover cases where the shift's size is known at compile time. In those cases the compiler can resolve the masking behavior automatically. We have _no_ tests, nor compiler changes that impact shifts that aren't compile-time known.

This means that prior to this change the behavior of shifting by a compile-time resolvable value and shifting by a non-resolvable value likely differs. Since backends often load DXIL and treat it as LLVM IR, this could cause concerning issues.

### DIFF
--- a/tools/clang/lib/CodeGen/CGExprScalar.cpp
+++ b/tools/clang/lib/CodeGen/CGExprScalar.cpp
@@ -3035,7 +3035,7 @@ Value *ScalarExprEmitter::EmitShl(const BinOpInfo &Ops) {
                       Ops.Ty->hasSignedIntegerRepresentation();
   bool SanitizeExponent = CGF.SanOpts.has(SanitizerKind::ShiftExponent);
   // OpenCL 6.3j: shift values are effectively % word size of LHS.
-  if (CGF.getLangOpts().OpenCL)
+  if (CGF.getLangOpts().OpenCL || CGF.getLangOpts().HLSL) // HLSL Change
     RHS =
         Builder.CreateAnd(RHS, GetWidthMinusOneValue(Ops.LHS, RHS), "shl.mask");
   else if ((SanitizeBase || SanitizeExponent) &&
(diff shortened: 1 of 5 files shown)


tools/clang/test/HLSLFileCheck/hlsl/operators/binary/shift-mask.hlsl
```diff
@@ -0,0 +1,33 @@
+// RUN: %dxc -T lib_6_3 %s -fcgl | FileCheck %s
+
+int shl32(int V, int S) {
+  return V << S;
+}
+
+// CHECK: define internal i32 @"\01?shl32
+// CHECK-DAG:  %[[Masked:[0-9]+]] = and i32 %{{[0-9]+}}, 31
```
pow2clk, line 8: I expect you've figured this out, but for debug builds, you'll need to allow for alphabetic characters here too.

tools/clang/lib/CodeGen/CGExprScalar.cpp
```diff
@@ -3035,7 +3035,7 @@ Value *ScalarExprEmitter::EmitShl(const BinOpInfo &Ops) {
                       Ops.Ty->hasSignedIntegerRepresentation();
   bool SanitizeExponent = CGF.SanOpts.has(SanitizerKind::ShiftExponent);
   // OpenCL 6.3j: shift values are effectively % word size of LHS.
-  if (CGF.getLangOpts().OpenCL)
+  if (CGF.getLangOpts().OpenCL || CGF.getLangOpts().HLSL) // HLSL Change
     RHS =
         Builder.CreateAnd(RHS, GetWidthMinusOneValue(Ops.LHS, RHS), "shl.mask");
   else if ((SanitizeBase || SanitizeExponent) &&
```
pow2clk, line 3041: I find myself thinking the CL behavior exactly matching what we need for HLSL seems too good to be true. I'm not familiar with these sanitizer settings. Where do they get set? How do the checks affect the codegen?

pow2clk (commented):
Some clarifications for my own benefit and a tip you probably don't need :)

pow2clk:
Did you shift the 'f' in the title in the spirit of the functionality involved? 😉

tools/clang/lib/CodeGen/CGExprScalar.cpp
llvm-beanz: That is the undefined behavior sanitizer instrumentation, which isn't supported by HLSL (yet!).

OpenCL's definition for bit shifts matches HLSL's probably more because it matches common hardware implementations.

pow2clk (approved):
Still surprised OpenCL does what we need, but happily so!

tex3d (approved):
I think the change is good.  It may add an extra `and` instruction for dynamic shift value, but a driver *should* eliminate an extra `and` if a target instruction has the assumed wrapping behavior.

PR #3606, opened 2021-03-18
Internal validator error messages don't need /Zi anymore.

adam-yang:

PR #4645, opened 2022-09-12
Add option to force disabling debug info

llvm-beanz:
This is (admittedly) a little hacky. DXC spends a lot of compile time updating debug information because we always generate it so that we can generate diagnostics for late-running validation.

This flag disables that validation resulting in about a 15% compile time improvement on a pessimistic inliner test.

### DIFF
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -168,6 +168,10 @@ def opt_enable : Separate<["-", "/"], "opt-enable">, Group<hlsloptz_Group>, Flag
 def opt_select : MultiArg<["-", "/"], "opt-select", 2>, MetaVarName<"<opt> <variant>">, Group<hlsloptz_Group>, Flags<[CoreOption, DriverOption, HelpHidden]>,
   HelpText<"Select this optimization variant.">;
 
+def fdisable_debug_info : Flag<["-"], "fdisable-debug-info">,
+  Group<hlslcomp_Group>, Flags<[CoreOption]>,
+  HelpText<"Force disabling all debug info. This will break diagnostic generation for late validation.">;
+
 /*
 def fno_caret_diagnostics : Flag<["-"], "fno-caret-diagnostics">, Group<hlslcomp_Group>,
  Flags<[CoreOption]>;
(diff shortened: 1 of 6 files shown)


include/dxc/Support/HLSLOptions.td
```diff
@@ -168,6 +168,10 @@ def opt_enable : Separate<["-", "/"], "opt-enable">, Group<hlsloptz_Group>, Flag
 def opt_select : MultiArg<["-", "/"], "opt-select", 2>, MetaVarName<"<opt> <variant>">, Group<hlsloptz_Group>, Flags<[CoreOption, DriverOption, HelpHidden]>,
   HelpText<"Select this optimization variant.">;
 
+def fdisable_debug_info : Flag<["-"], "fdisable-debug-info">,
+  Group<hlslcomp_Group>, Flags<[CoreOption]>,
+  HelpText<"Force disabling all debug info. This will break diagnostic generation for late validation.">;
```
pow2clk, line 173: I'm not sure the description needs to mention debug info. It might be more confusing since the flag to control the debug info in the output is `-Zi`. In this case, it seems like an implementation detail. Of more interest might be to refer to it as not maintaining line tracking information and mention it improves compile times. The diagnostic helper functions should still give the best information about location they can when they don't have line info.

tools/clang/test/HLSLFileCheck/validation/ResourceResolvingNoDI.hlsl
```diff
@@ -0,0 +1,25 @@
...
+  RWBuffer<int> Local;
+  if (B) Local = Buf[Idx];
+  return Local[Idx];
+}
+
+[numthreads(1,1,1)]
+void main(uint GI : SV_GroupIndex) {
+  Out[GI] = getVal(false, GI);
+  Out[GI+1] = getValASecondTime(false, GI);
+}
```
pow2clk, line 25: A test that verifies that line info isn't available might be useful. You can look at the tests Adam changed in #3606 for some examples.

adam-yang:
What happens you give -Zi and -fdisable-debug-info together?

tools/clang/test/HLSLFileCheck/validation/ResourceResolvingNoDI.hlsl
llvm-beanz: The note this test checks for is only emitted if line information isn't available... does that not cover that case or are you wanting me to verify that the IR has no line information before we get into the backend?

llvm-beanz:
If you pass `-Zi` `-fdisable-debug-info` is ignored.

adam-yang (approved):

pow2clk (approved):
Looks good

PR #4646, opened 2022-09-12
Add option to enable LTO/LTCG

llvm-beanz:
This adds a new option to the build to enable link-time optimzied builds of DXC. On Windows with MSVC this seems to give about a 4% runtime performance boost on my one locally run test case... and on Linux I'm seeing about 6% on that same locally run test case.

It's free performance so yay :D

### DIFF
--- a/utils/hct/hctbuild.cmd
+++ b/utils/hct/hctbuild.cmd
@@ -217,6 +217,10 @@ if "%1"=="-update-generated-sources" (
   set CMAKE_OPTS=%CMAKE_OPTS% -DHLSL_COPY_GENERATED_SOURCES=1
   shift /1 & goto :parse_args
 )
+if "%1"=="-lto" (
+  set CMAKE_OPTS=%CMAKE_OPTS% -DLLVM_ENABLE_LTO=On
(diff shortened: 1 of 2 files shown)


utils/hct/hctbuild.cmd
```diff
@@ -217,6 +217,10 @@ if "%1"=="-update-generated-sources" (
   set CMAKE_OPTS=%CMAKE_OPTS% -DHLSL_COPY_GENERATED_SOURCES=1
   shift /1 & goto :parse_args
 )
+if "%1"=="-lto" (
+  set CMAKE_OPTS=%CMAKE_OPTS% -DLLVM_ENABLE_LTO=On
+  shift /1 & goto :parse_args
+)
```
pow2clk, line 223: My only thought is why not enable this by default?

pow2clk (approved):
Makes sense!

utils/hct/hctbuild.cmd
llvm-beanz: LTO is generally slower to build, so it costs developer time if this is the default build option.

PR #4652, opened 2022-09-14
build: fix include in header files.

Keenuts:
Building DXC with another build system revealed that some includes were dependent on headers only available by chance. Adding missing headers.

### DIFF
--- a/include/dxc/DXIL/DxilInstructions.h
+++ b/include/dxc/DXIL/DxilInstructions.h
@@ -12,11 +12,11 @@
 
 #pragma once
 
+#include "dxc/DXIL/DxilOperations.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/Instruction.h"
 #include "llvm/IR/Instructions.h"
 
-// TODO: add correct include directives
 // TODO: add accessors with values
 // TODO: add validation support code, including calling into right fn
 // TODO: add type hierarchy
--- a/include/dxc/DXIL/DxilFunctionProps.h
+++ b/include/dxc/DXIL/DxilFunctionProps.h
@@ -13,6 +13,7 @@
 
 // for memset dependency:
 #include <cstring>
+#include <vector>
 
 #include "dxc/DXIL/DxilConstants.h"
 
--- a/include/dxc/DXIL/DxilShaderFlags.h
+++ b/include/dxc/DXIL/DxilShaderFlags.h
@@ -11,6 +11,8 @@
 
 #pragma once
 
+#include <stdint.h>
+
 namespace hlsl {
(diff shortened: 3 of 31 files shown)


include/dxc/DXIL/DxilInstructions.h
```diff
@@ -12,11 +12,11 @@
 
 #pragma once
 
+#include "dxc/DXIL/DxilOperations.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/Instruction.h"
 #include "llvm/IR/Instructions.h"
 
-// TODO: add correct include directives
```
pow2clk, line 19: I'm not sure this means what you think it means. Regardless, I think it is out of date.

pow2clk (commented):
As I see it, there are three kinds of changes included here:
1. reshuffling of include directives
2. addition of includes to headers
3. whitespace changes

To what extent are the type 2 changes required as a result of the shuffling done by the type 1? A lot of this looks like the result of automated formatting, which is fine, but I'd like to understand if the missing includes were extant or a result of this formatting correction.

Keenuts:
Hi! Thanks for the review!

My process was to build DXC using bazel, and fixing includes as they failed. Once the include fixed, I sorted them alphabetically, or switched the system vs local headers as I've seen some files doing a block of local first, then a block of system includes.
I haven't touched any files not causing a build failure first, to reduce the size of the patch 🙂 

As I fixed them, I have removed one or 2 trailing whitespaces I saw, and added space between the pragma once and first include as seen in other files. EOF \n removal is an result from saving with vim, can remove that from the patch if that's an issue.

pow2clk:
None of these are really issues. I just wanted to better understand where this was coming from. Did the build you reference fail suddenly? I thought it was one you kept running regularly.

pow2clk:
I ask because, if possible, I'd like to prevent regressions like this. I don't think we can get a CI build for bazel, but perhaps if we understand the source of the discrepancy better, we might be able to tweak our CI builds so they are more similar and more likely to catch failures.

pow2clk (approved):
I'd still like to better understand the source of the failure that required this change, but it is pretty clearly NFC and if nothing else, it is better organization and formatting. So, I approve!

Keenuts:
Thanks!
So we do build it regularly using bazel, but we made some changes internally and it became a bit more sensible to this kind of issues. It will be continuously built from now on, so we should catch regressions on that part. (Same goes for the other PR I sent you https://github.com/microsoft/DirectXShaderCompiler/pull/4597, the memory UB was caught by the CI) 

Why couldn't we have a CI build using bazel btw?

PR #4653, opened 2022-09-14
move entry point default code

bob80905:
The entry point member variable is currently only set if the dxilcontainer contains a DFCC_ShaderSourceInfo header.
This change will set the entry point default to main regardless of which headers can be found in the container.

### DIFF
--- a/tools/clang/tools/dxcompiler/dxcpdbutils.cpp
+++ b/tools/clang/tools/dxcompiler/dxcpdbutils.cpp
@@ -495,11 +495,6 @@ private:
           AddArgPair(std::move(newPair));
         }
 
-        // Entry point might have been omitted. Set it to main by default.
-        if (m_EntryPoint.empty()) {
-          m_EntryPoint = L"main";
-        }
-
         // Sources
         for (unsigned i = 0; i < reader.GetSourcesCount(); i++) {
           hlsl::SourceInfoReader::Source source_data = reader.GetSource(i);
@@ -546,6 +541,12 @@ private:
       } break; // hlsl::DFCC_ShaderDebugInfoDXIL
       } // switch (four_cc)
     } // For each part
+    
+    // Entry point might have been omitted. Set it to main by default.
+    // TODO: Check to see that this DxilContainer is not a library before setting the entry point.
+    if (m_EntryPoint.empty()) {
+      m_EntryPoint = L"main";
+    }
 
     return S_OK;
   }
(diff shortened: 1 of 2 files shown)


tools/clang/tools/dxcompiler/dxcpdbutils.cpp
```diff
@@ -546,6 +541,8 @@ struct DxcPdbUtils : public IDxcPdbUtils, public IDxcPixDxilDebugInfoFactory
       } break; // hlsl::DFCC_ShaderDebugInfoDXIL
       } // switch (four_cc)
     } // For each part
+    
+    SetEntryPointToDefaultIfEmpty();
```
adam-yang, line 545: It's possible `HandleDxilContainer` didn't find anything at all, and then need to look in the debug DXIL module. Better to call this at the end of the `try` block in DxcPdbUtils::Load

tools/clang/tools/dxcompiler/dxcpdbutils.cpp
```diff
@@ -752,6 +749,14 @@ struct DxcPdbUtils : public IDxcPdbUtils, public IDxcPixDxilDebugInfoFactory
     return m_pDebugProgramBlob != nullptr;
   }
 
+  virtual void STDMETHODCALLTYPE SetEntryPointToDefaultIfEmpty() {
+    // Entry point might have been omitted. Set it to main by default.
+    // TODO: Check to see that this DxilContainer is not a library before setting the entry point.
+    if (m_EntryPoint.empty()) {
```
adam-yang, line 755: It's possible that the input is just a non-debug DXIL and have no args at all, in which case we don't want to populate m_EntryPoint. Should also check if m_ArgPairs is not empty.

tools/clang/tools/dxcompiler/dxcpdbutils.cpp
```diff
@@ -770,6 +775,8 @@ struct DxcPdbUtils : public IDxcPdbUtils, public IDxcPixDxilDebugInfoFactory
     }
     CATCH_CPP_RETURN_HRESULT()
 
+    SetEntryPointToDefaultIfEmpty();
```
adam-yang, line 778: `OverrideRootSignature` could also run into the same problem. Should call it there too.

adam-yang:
Tests?

adam-yang (approved):
Thanks! LGTM!

PR #4654, opened 2022-09-14
[SPIR-V] Handle debugInfo for composite types across source files

greg-lunarg:
This fixes a crash when generating source debugInfo for composite types across multiple source files.

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/shader.debug.file.composite.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/shader.debug.file.composite.hlsl
@@ -1,0 +1,30 @@
+// RUN: %dxc -E MainPs -T ps_6_0 -fspv-target-env=vulkan1.1 -fspv-debug=vulkan-with-source
+
+// Just check that compilation completes
+// CHECK: %MainPs = OpFunction %void None
+
+#line 43 "lpv_debug.fxc"
+struct PS_INPUT
+{
+
+    float3 vPositionWithOffsetWs : TEXCOORD0 ;
+
+} ;
+
+#line 430 "csgo_common_ps_code.fxc"
+struct PS_OUTPUT
+{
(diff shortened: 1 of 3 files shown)


greg-lunarg:
@Keenuts or @cassiebeckley Can you please review?

tools/clang/test/CodeGenSPIRV/shader.debug.file.composite.hlsl
```diff
@@ -0,0 +1,30 @@
+// RUN: %dxc -E MainPs -T ps_6_0 -fspv-target-env=vulkan1.1 -fspv-debug=vulkan-with-source -spirv
```
cassiebeckley, line 1: I believe `-spirv` is implicit for these

tools/clang/test/CodeGenSPIRV/shader.debug.file.composite.hlsl
```diff
@@ -0,0 +1,30 @@
...
+
+// Just check that compilation completes
```
cassiebeckley, line 3: Would it be useful to add more detailed checks in order to prevent regressions, or will just checking that it compiles successfully suffice?

tools/clang/test/CodeGenSPIRV/shader.debug.file.composite.hlsl
greg-lunarg: OK.

tools/clang/test/CodeGenSPIRV/shader.debug.file.composite.hlsl
greg-lunarg: Since the issue was that the compiler crashed (and hit an assert under Debug), I believe it is sufficient to just test that the compile completes.

cassiebeckley (approved):

Issue #4141, opened 2021-12-13
[SPIR-V] Logical Operator Short Circuiting


jaebaek:
Based on [the blog posting](https://devblogs.microsoft.com/directx/announcing-hlsl-2021/#logical-operator-short-circuiting), "HLSL 2021 is introducing logical operator short circuiting behavior matching C. This means that if the first operand of a boolean `&&` is evaluated as `false`, the second operand will not be evaluated. Similarly, if the first operand of a boolean `||` is evaluated as `true` the second operand will not be evaluated."

For `A && B`, we have to generate the following SPIR-V:

```
%before = OpLabel
...
%A_value = value of A
OpMergeBranch %merge
OpBranch %A_value %label %merge
%label = OpLabel
%result0 = value of B
%merge = OpLabel
%result = OpPhi %bool %false %before %result0 %label
```

"In addition to adding short circuiting behavior, in HLSL 2021 logical operators can only be used with scalar values. ..."

Therefore, we have to emit errors if the HLSL code contains logical operators with non-scalar type values.

PR #4658, opened 2022-09-15
[SPIR-V] Add short-circuiting logical operators for HLSL 2021

vettoreldaniele:
This implements logical AND and logical OR short-circuiting, which was added with HLSL 2021.

Note: the ternary operator is also short-circuited since HLSL 2021. I will implement that in a follow-up PR.

Issue: #4141


cassiebeckley (approved):
LGTM

PR #4660, opened 2022-09-15
Pulled the zlib compression code into a library

adam-yang:
Refactor the zlib compression code into its own library. This does not change functionality, just a refactor to prepare for further work. No new tests needed.

### DIFF
--- a/include/dxc/DxilCompression/DxilCompression.h
+++ b/include/dxc/DxilCompression/DxilCompression.h
@@ -1,0 +1,37 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DxilCompression.h                                                         //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+//
+// Helper wrapper functions for zlib deflate and inflate. Entirely
+// self-contained, only depends on IMalloc interface.
+//
+
+#pragma once
+
+struct IMalloc;
+namespace hlsl {
+  enum class ZlibResult {
+    Success = 0,
+    InvalidData = 1,
+    OutOfMemory = 2,
+  };
+
+  ZlibResult ZlibDecompress(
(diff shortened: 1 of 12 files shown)


tex3d (approved):
LGTM!

include/dxc/DxilCompression/DxilCompression.h
```diff
@@ -0,0 +1,37 @@
...
+    const void *pCompressedBuffer,
+    size_t BufferSizeInBytes,
+    void *pUncompressedBuffer,
+    size_t UncompressedBufferSize);
+
+  typedef void *ZlibAllocateBufferFn(void *pUserData, size_t RequiredSize);
+  ZlibResult ZlibCompress(IMalloc *pMalloc,
+    const void *pData, size_t pDataSize,
+    void *pUserData,
+    ZlibAllocateBufferFn *AllocateBufferFn,
```
dmpots, line 35: Why do we have a separate allocation function? Why is it not using the already passed IMalloc or a different IMalloc?

lib/DxilCompression/DxilCompression.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "dxc/Support/Global.h"
+#include "dxc/Support/WinIncludes.h"
+
+#include "miniz.h"
+typedef size_t ZlibSize_t;
+
+namespace {
+//
+// A resource managment class for a zlib stream.
+// Class the appropriate init and end routines on the stream.
```
dmpots, line 24: This sentence does not make sense as written.

include/dxc/DxilCompression/DxilCompression.h
adam-yang: It's not really an allocation function, more of a "resize buffer" callback function. Caller code might be writing into a container type like std::vector. I'll rename it and add an explanation in the comments.

dmpots: Ok, that makes sense. A comment with some explanation would be helpful!

dmpots (approved):

Keenuts:
Hi Adam, thanks for refactoring!
Is it in your plans to remove the old miniz implementation in DXC (lib/Miniz, include/miniz) ?

PR #4662, opened 2022-09-16
Separated RDAT creation code into its own library

adam-yang:
- Pulled out RDAT writing from DxilContainerAssembly into its own helper DxilRDATBuilder
- Changed the way RawBytesPart deduplicates inputs. Instead of using the pointer address, use the data's content as keys in a hash table, then writing it to the final buffer in the order they were inserted.


tex3d (approved):
Looks good!  Only one minor optional nit.

tex3d (approved):
LGTM!

PR #4663, opened 2022-09-16
[SPIR-V] Support arrays of RWStructuredBuffers

### LINKED ISSUE #3281: [SPIR-V]  arrays of RW/append/consume structured buffers unsupported
When targetting SPIR-V the following code generates a compile error:

[Shader Playground](http://shader-playground.timjones.io/db9ec964835342b15cbdd150df723ff7)
```hlsl
struct PSInput
{
	float4 color : COLOR;
};

StructuredBuffer<uint> g_buffer[] : register(t0, space1);
RWStructuredBuffer<uint> g_rwbuffer[] : register(u0, space2);

float4 PSMain(PSInput input) : SV_TARGET
{
	return input.color;
}
```

However, in DXIL mode this is perfectly fine - and it's how you would go about implementing bindless UAVs.

It's a little hard to grep for but the relevant error is from here https://github.com/microsoft/DirectXShaderCompiler/blob/4ced9bdf05ba26c6d9c01258ad8914f892e3620a/tools/clang/lib/SPIRV/SpirvEmitter.cpp#L1357

I suspect the check should just be `isAppendStructuredBuffer(type) || isConsumeStructuredBuffer(type)` and allow RWStructuredBuffers to pass through.

cassiebeckley:
I'm not confident this generates the correct code in all cases, so I'd appreciate it if there are any suggestions.

Fixes #3281

### DIFF
--- a/docs/SPIR-V.rst
+++ b/docs/SPIR-V.rst
@@ -1171,6 +1171,10 @@ will be translated into
   ; Variable
   %myCbuffer = OpVariable %_ptr_Uniform_type_ConstantBuffer_T Uniform
 
+You may create arrays of ``RWStructuredBuffer<T>``, but no associated counter
+variables will be created, and you may not use ``.IncrementCounter()`` or
+``.DecrementCounter()`` on the generated array's elements.
+
 ``AppendStructuredBuffer`` and ``ConsumeStructuredBuffer``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1679,6 +1679,17 @@ void DeclResultIdMapper::registerSpecConstant(const VarDecl *decl,
 void DeclResultIdMapper::createCounterVar(
     const DeclaratorDecl *decl, SpirvInstruction *declInstr, bool isAlias,
     const llvm::SmallVector<uint32_t, 4> *indices) {
+
+  if (const auto *varDecl = dyn_cast<VarDecl>(decl)) {
+    if (const Expr *init = varDecl->getInit()) {
+      init = init->IgnoreParenCasts();
+      if (dyn_cast<ArraySubscriptExpr>(init)) {
+        // do not create counter from array subscript
+        return;
+      }
+    }
+  }
+
   std::string counterName = "counter.var." + decl->getName().str();
   if (indices) {
     // Append field indices to the name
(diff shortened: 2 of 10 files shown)


vettoreldaniele (changes requested):
I see two problems with this:
1) The following breaks with `-fspv-flatten-resource-arrays`:
```
RWStructuredBuffer<float4>   localSBuffer = MyRWSBuffer[index];
localSBuffer[0] = 3.0f;
```
```
warning: cannot find the associated counter variable
    RWStructuredBuffer<float4>   localSBuffer = MyRWSBuffer[index];
                                                ^
fatal error: failed to legalize SPIR-V: Variable cannot be replaced: invalid index
  %52 = OpAccessChain %_ptr_Uniform_type_RWStructuredBuffer_v4float %MyRWSBuffer %29
```

2) `MyRWSBuffer[index].IncrementCounter()` correctly generates an error, but the following does not and instead loads an undef value for the non-existent counter variable:
```
RWStructuredBuffer<float4> localSBuffer = MyRWSBuffer[index];
localSBuffer.IncrementCounter();
```

Ipotrick:
Id love to have this merged. This would make my hlsl api much nicer.

GabeRundlett:
Is there a reason this was not merged?

vettoreldaniele:
There are still outstanding comments. @cassiebeckley PTAL when you have a chance. 

I'm ok with the `-fspv-flatten-resource-arrays` behavior being fixed in a follow-up PR , but the problem with the non-existent counter variable should be fixed before this is merged.

cassiebeckley:
Sorry for the slow response on this, things have been busy. I'll try to get to this by the end of the week

Ipotrick:
Thank you!

cassiebeckley:
Quick update: I did make some progress on this, but I'm still working on making sure the correct error messages show up in the correct situations

cshenton:
This would be amazing to have for bindless workflows!

cassiebeckley:
Added a new commit that prevents counters from being created from an array subscript expression, which should fix issue 1. There are some tricky situations which should probably also result in errors that I don't know how to fix - for example, a counter is always created for any buffer returned from a function call. This seems difficult to detect statically - the return value of the function is effectively the buffer + the counter, and the function could have a branch where it either returns a buffer from an array, or a buffer that is not from an array.

As for issue 2, I'll create a tracking issue to follow up.

s-perron:
RWStructuredBuffer<float4> localSBuffer = MyRWSBuffer[index];
localSBuffer.IncrementCounter();

s-perron:
> As for issue 2, I'll create a tracking issue to follow up.

We should discuss this. We should not give a partially working feature until we have a design to make it fully functional. Something like issue 2 is a clear bug that would be nearly impossible to debug. Is there a design doc for this?

cassiebeckley:
(I mixed up issue 1 and 2 in my comment, fixed it)

s-perron:
> (I mixed up issue 1 and 2 in my comment, fixed it)

Cool, what is the error message?

cassiebeckley:
At the moment, 

```
arr.hlsl:5:12: fatal error: cannot find the associated counter variable
    return localSBuffer.IncrementCounter();
           ^
```

PR #4667, opened 2022-09-16
Add `LLVM_OPTIMZIE_SANITIZED_BUILDS` option

llvm-beanz:
I added this upstream a few years ago so that you can control passing `-O1` to the compiler when building debug sanitized builds. Without being able to disable optimizing sanitizer builds it is really tricky to debug sanitizer failures, but without optimizing the build ASan builds take _forever_ to run tests which makes them difficult to use.


hekota (approved):
LGTM, please fix the typo in title ;)

PR #4668, opened 2022-09-16
[NFC] Fix asan error in test case

llvm-beanz:
This test relies on a stack allocation remaining live after the scope of an if block that it is allocated in... Clearly no good.

### DIFF
--- a/tools/clang/unittests/HLSL/AllocatorTest.cpp
+++ b/tools/clang/unittests/HLSL/AllocatorTest.cpp
@@ -102,6 +102,8 @@ bool Align(unsigned &pos, unsigned end, unsigned align) {
 }
 
 struct Element {
+  Element() = default;
+  Element(const Element&) = default;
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSL/AllocatorTest.cpp
```diff
@@ -663,7 +665,7 @@ TEST_F(AllocatorTest, GapFilling) {
 
 TEST_F(AllocatorTest, Allocate) {
   WEX::TestExecution::SetVerifyOutput verifySettings(WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
-  for (auto &&scenario : m_Scenarios) {
+  for (auto &scenario : m_Scenarios) {
```
pow2clk, line 668: You're under no obligation to fix this entire file, but there are a number of other areas where this is done unnecessarily.

pow2clk (approved):
Looks good. 

I feel like this should kind of mistake should be harder to do, but I don't have a proposal for Bjarne . . . yet 🤔

PR #4670, opened 2022-09-17
Fix issue in HelperLaneTestWave ExecutionTest caused by assume last lane has VertexID 2.

python3kgae:
Use the last lane in the wave with VertexID 2 instead of the lane with VertexID 2 when writes results.


tex3d (approved):
LGTM!

PR #4671, opened 2022-09-18
Remove ResourceAttribute metadata node from DxilFieldAnnotation

tex3d:
Currently, ResourceAttribute is only used to capture resource type information during CodeGen for the annotate handle generated during AddOpcodeParamForIntrinsic.

Going through a metadata node is totally unnecessary, just adding code and complexity.

This change just stores the DxilResourceProperties instead, renaming `*ResourceAttribute` methods to `*ResourceProperties`.
This change also emits the resource properties to new field annotaion metadata in validator version 1.8+ to preserve through serialization.


1w02:
@EpicTeamAdmin

python3kgae (approved):

Issue #4673, opened 2022-09-20
Cannot use `select` intrinsic with SamplerState


cassiebeckley:
The [blog post announcing logical operator short circuiting](https://devblogs.microsoft.com/directx/announcing-hlsl-2021/#logical-operator-short-circuiting) seems to suggest that the new `select` intrinsic should behave the same as the ternary operator used to prior to HLSL 2021. However, it does not allow selection between two different `SamplerState`s.

For example,

```hlsl
SamplerState gSS1;
SamplerState gSS2;

Texture2D gTex;

void main() {
  bool cond;
  int intCond;
  float4 l = gTex.Sample(cond ? gSS1 : gSS2, float2(1., 2.));
  SamplerState s = intCond ? gSS1 : gSS2;
}
```

compiles, but

```hlsl
SamplerState gSS1;
SamplerState gSS2;

Texture2D gTex;

void main() {
  bool cond;
  int intCond;
  float4 l = gTex.Sample(select(cond, gSS1, gSS2), float2(1., 2.));
  SamplerState s = select(intCond, gSS1, gSS2);
}
```

produces

```
sample-select.hlsl:9:26: error: no matching function for call to 'select'
  float4 l = gTex.Sample(select(cond, gSS1, gSS2), float2(1., 2.));
                         ^~~~~~
sample-select.hlsl:9:26: note: candidate function not viable: no known conversion from 'SamplerState' to 'float' for 2nd argument
  float4 l = gTex.Sample(select(cond, gSS1, gSS2), float2(1., 2.));
                         ^            ~~~~
sample-select.hlsl:10:20: error: no matching function for call to 'select'
  SamplerState s = select(intCond, gSS1, gSS2);
                   ^~~~~~
sample-select.hlsl:10:20: note: candidate function not viable: no known conversion from 'SamplerState' to 'float' for 2nd argument
  SamplerState s = select(intCond, gSS1, gSS2);
                   ^               ~~~~
```

when compiled with `dxc -T ps_6_0 -E main -HV 2021 sample-select.hlsl`

pow2clk:
Just a note that we will need to fix this for DXIL generation as well 😣

PR #4674, opened 2022-09-20
[SPIR-V] Add support for logical operator intrinsics

### LINKED ISSUE #4148: [SPIR-V] select intrinsic function unimplemented
As mentioned [here](https://devblogs.microsoft.com/directx/announcing-hlsl-2021/#logical-operator-short-circuiting) short circuiting prevents usage of logical operators with non-scalar type values.
The alternative is to use new `and`, `or` and `select` intrinsics.
However using `select` produces "select intrinsic function unimplemented" error, thus there is no way to actually get the old behavior to work.

cassiebeckley:
Implemented `and`, `or`, and `select`. I used the existing tests for the old behavior of `&&`, `||`, and `?:` in order to verify that they behave as expected (with the caveat of #4673, but that has the same behavior for DXIL as for SPIR-V).

Fixes #4148

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -1058,7 +1058,10 @@ SpirvInstruction *SpirvEmitter::doExpr(const Expr *expr,
     if (getCompilerInstance().getLangOpts().EnableShortCircuit) {
       result = doShortCircuitedConditionalOperator(condExpr);
     } else {
-      result = doConditionalOperator(condExpr);
+      const Expr *cond = condExpr->getCond();
+      const Expr *falseExpr = condExpr->getFalseExpr();
+      const Expr *trueExpr = condExpr->getTrueExpr();
+      result = doConditional(condExpr, cond, falseExpr, trueExpr);
     }
   } else if (const auto *defaultArgExpr = dyn_cast<CXXDefaultArgExpr>(expr)) {
     if (defaultArgExpr->getParam()->hasUninstantiatedDefaultArg()) {
@@ -3378,14 +3381,6 @@ SpirvInstruction *SpirvEmitter::doShortCircuitedConditionalOperator(
   SpirvInstruction *result = spvBuilder.createLoad(type, tempVar, loc, range);
   result->setRValue();
   return result;
-}
-
-SpirvInstruction *
-SpirvEmitter::doConditionalOperator(const ConditionalOperator *expr) {
-  const Expr *cond = expr->getCond();
-  const Expr *falseExpr = expr->getFalseExpr();
-  const Expr *trueExpr = expr->getTrueExpr();
-  return doConditional(expr, cond, falseExpr, trueExpr);
 }
 
 SpirvInstruction
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -3382,12 +3382,20 @@ SpirvInstruction *SpirvEmitter::doShortCircuitedConditionalOperator(
 
 SpirvInstruction *
 SpirvEmitter::doConditionalOperator(const ConditionalOperator *expr) {
-  const auto type = expr->getType();
-  const SourceLocation loc = expr->getExprLoc();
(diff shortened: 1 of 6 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.h
```diff
@@ -122,6 +122,10 @@ class SpirvEmitter : public ASTConsumer {
                                SourceRange rangeOverride = {});
   SpirvInstruction *doCompoundAssignOperator(const CompoundAssignOperator *);
   SpirvInstruction *doConditionalOperator(const ConditionalOperator *expr);
+  SpirvInstruction *doConditional(const Expr *expr,
+                                  const Expr *cond,
+                                  const Expr *falseExpr,
+                                  const Expr *trueExpr);
```
Keenuts, line 128: this could remain private to SpirvEmitter no?

tools/clang/test/CodeGenSPIRV/intrinsics.select.hlsl
```diff
@@ -0,0 +1,181 @@
+// RUN: %dxc -T ps_6_0 -E main -HV 2021
+
+// CHECK: [[v3i0:%\d+]] = OpConstantComposite %v3int %int_0 %int_0 %int_0
+SamplerState gSS1;
```
Keenuts, line 4: I might be wrong, but those aren't used right? Not sure I'm following the commit message, are those supposed to be tested/working yet?

Keenuts (changes requested):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -3382,12 +3382,20 @@ SpirvInstruction *SpirvEmitter::doShortCircuitedConditionalOperator(
 
 SpirvInstruction *
 SpirvEmitter::doConditionalOperator(const ConditionalOperator *expr) {
```
vettoreldaniele, line 3384: Maybe we can get rid of `doConditionalOperator` entirely now and just call `doConditional` with the right arguments, like you do for the intrinsics?

tools/clang/test/CodeGenSPIRV/intrinsics.select.hlsl
cassiebeckley: Right, you're correct. Thanks for catching that!

tools/clang/lib/SPIRV/SpirvEmitter.h
cassiebeckley: It can and I believe it is (see line 95 for the `private` label)

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: Done

tools/clang/test/CodeGenSPIRV/intrinsics.select.hlsl
```diff
@@ -0,0 +1,179 @@
+// RUN: %dxc -T ps_6_0 -E main -HV 2021
+
+// CHECK: [[v3i0:%\d+]] = OpConstantComposite %v3int %int_0 %int_0 %int_0
+
+Texture2D gTex;
```
vettoreldaniele, line 5: This isn't used either.

cassiebeckley: Removed

vettoreldaniele (approved):

tools/clang/lib/SPIRV/SpirvEmitter.h
Keenuts: Yes! Sorry, saw the header and assumed it was public.

tools/clang/test/CodeGenSPIRV/intrinsics.select.hlsl
Keenuts: Shouldn't we test them instead of remove them? (since the bug was found with those?)

Keenuts (changes requested):

Keenuts (approved):
Apparently SamplerStates are not supposed to work, so fine in removing them from the tests. LGTM

PR #4677, opened 2022-09-20
dxexp - report advanced texture ops capability

hekota:
- add D3D12_FEATURE_DATA_D3D12_OPTIONS14 definition if not supported by current WinSDK
- move all D3D_SHADER_MODEL defines into one spot

### DIFF
--- a/tools/dxexp/dxexp.cpp
+++ b/tools/dxexp/dxexp.cpp
@@ -45,7 +45,6 @@ static HRESULT AtlCheck(HRESULT hr) {
 
 // Not defined in Creators Update version of d3d12.h:
 #if WDK_NTDDI_VERSION <= NTDDI_WIN10_RS2
-#define D3D_SHADER_MODEL_6_1 ((D3D_SHADER_MODEL)0x61)
 #define D3D12_FEATURE_D3D12_OPTIONS3 ((D3D12_FEATURE)21)
 typedef
 enum D3D12_COMMAND_LIST_SUPPORT_FLAGS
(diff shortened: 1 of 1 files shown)


tools/dxexp/dxexp.cpp
```diff
@@ -229,21 +232,26 @@ static HRESULT PrintAdapters() {
...
       pDevice->CheckFeatureSupport(D3D12_FEATURE_D3D12_OPTIONS4, &DeviceOptions4, sizeof(DeviceOptions4));
       pDevice->CheckFeatureSupport(D3D12_FEATURE_D3D12_OPTIONS5, &DeviceOptions5, sizeof(DeviceOptions5));
+      pDevice->CheckFeatureSupport(D3D12_FEATURE_D3D12_OPTIONS14, &DeviceOptions14, sizeof(DeviceOptions14));
+
       AtlCheck(GetHighestShaderModel(pDevice, DeviceSM));
       const char *Format = IsOutputJson ?
-        "%c { \"name\": \"%S\", \"sm\": \"%s\", \"wave\": %s, \"i64\": %s, \"bary\": %s, \"view-inst\": \"%s\", \"16bit\": %s, \"raytracing\": \"%s\" }\n" :
-        "%c %S - Highest SM [%s] Wave [%s] I64 [%s] Barycentrics [%s] View Instancing [%s] 16bit Support [%s] Raytracing [%s]\n";
+        "%c { \"name\": \"%S\", \"sm\": \"%s\", \"wave\": %s, \"i64\": %s, \"bary\": %s, \"view-inst\": \"%s\", \"16bit\": %s, \"raytracing\": \"%s\", \"ato\":\"%s\" }\n" :
+        "%c %S - Highest SM [%s] Wave [%s] I64 [%s] Barycentrics [%s] View Instancing [%s] 16bit Support [%s] Raytracing [%s] Advanced Texture Ops [%s]\n";
```
tex3d, line 254: Glad to see this updated!

For future work: I think we are still missing more optional feature bits, and this one-line format is getting pretty unwieldy for the command line output.  I wonder if we should review/add the missing caps and reconsider the format?

hekota: Sure, if there are any caps that are important we can add them. Do you have any in mind?

Regarding the output - we can change it, but I personally I don't mind the one-line output. For me it usually lists caps for about 6 devices at a time (including experimental) so it makes the output nicely compressed and still readable. For automation there is an option to get the output formatted as json, which is useful.

pow2clk (approved):
Looks good!

PR #4678, opened 2022-09-20
Fix ShaderModel::Get() when entry does not exist in hashToIdxMap

yuriy-odonnell-epic:
Fixes a bug introduced in commit 8b8956f82be592ed774ee8c59980d6336a67bff0 (cc @pow2clk)


pow2clk (approved):
Nice catch!

PR #4679, opened 2022-09-21
[SPIRV] Support 16-bit Type Cast Intrinsics

gmitrano-unity:
This change adds support for 16-bit type cast intrinsics by routing them into the same logic used by the 32-bit type casts. With this change in place, asfloat16, asint16, and asuint16 are now translated into OpBitcast in SPIR-V.

This change also includes new partial disassembly tests for each function similar to the ones used for 32-bit casts.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -7969,8 +7969,11 @@ SpirvEmitter::processIntrinsicCallExpr(const CallExpr *callExpr) {
     break;
   case hlsl::IntrinsicOp::IOP_asdouble:
   case hlsl::IntrinsicOp::IOP_asfloat:
+  case hlsl::IntrinsicOp::IOP_asfloat16:
   case hlsl::IntrinsicOp::IOP_asint:
+  case hlsl::IntrinsicOp::IOP_asint16:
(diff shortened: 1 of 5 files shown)


cassiebeckley (changes requested):
Could you update the documentation comments in `processIntrinsicAsType` on I believe line 10112 of SpirvEmitter.cpp? Other than that, this looks great!

cassiebeckley (approved):
LGTM, thanks for the contribution!

gmitrano-unity:
> LGTM, thanks for the contribution!

You're very welcome! Thanks for the quick review! 🙂

gmitrano-unity:
Hmmm, did the license/cla job get stuck or did I miss a step? That check has been running for 16+ hrs.

vettoreldaniele:
Looks like there are two CLA jobs, not sure if this is a recent change. Can you try force pushing to retry the checks?

vettoreldaniele:
@pow2clk Do you happen to know what's up with the double CLA check?

vettoreldaniele:
@tex3d Adding more folks from Microsoft to see if they can help with the CLA issue.

pow2clk:
We've been having some issues with the CLA check recently. I'm not sure what the cause is. I'm looking into it.

pow2clk:
Looks like I cleared the logjam! Fire when ready!

PR #4680, opened 2022-09-22
Add RDAT::RecordTraits::MaxRecordSize to default to newest version

tex3d:
Added MaxRecordSize calling DerivedRecordSize that gets specialized for each base in RDAT_STRUCT_TABLE_DERIVED macro to call the derived MaxRecordStride.  The most-derived version of DerivedRecordSize will use the default implementation, which is the size of that type.


adam-yang (approved):
LGTM!

PR #4685, opened 2022-09-23
Remove arbitrary run line length limit

tex3d:
A very strange sequence of copying the run line to a stack char array of size 300, where it would not even null-terminate the line if it hit the limit, seems totally unnecessary, since it's dealing with std::strings in the first place.
This would truncate and pick up garbage from the stack in the stored run line when length >= 300 characters - causing the test to fail, of course. Perhaps some of this is an artifact of some older version of the code.

In any case, this change removes the extra copying step and simply uses the entire std::string result from std::getline() instead.


pow2clk (approved):
Now our RUN lines will be limitless!

PR #4687, opened 2022-09-24
Pull ReplaceDisassemblyText out of DXCTestUtils.cpp

bob80905:
The IsNormal HLK test depends on a function called ReplaceShaderText.cpp.
As it is currently defined, the function takes a dependency on llvm for regex usage.
This dependency prevents successful builds when copied over to another repo.
IsNormal depends on other functions with bad dependencies, so this is the first of several PRs to get IsNormal 
to a state in which it only depends on what is available in the other repo. 
This PR divides ReplaceDisassemblyText  (RDT) into 3 functions, RDT, RDTWithoutRegex, and RDTWithRegex.
RDT will call the other 2 functions depending on whether the caller specified that any of the LookFor strings contain a regex, with the boolean parameter passed to RDT.
Because RDTWithoutRegex will be defined in HLSLTestUtils.h, it will be available in the other repo without any llvm dependencies.
RDTWithRegex will still use the llvm regex library, and this should eventually be switched to use std::regex.

### DIFF
--- a/include/dxc/Test/DxcTestUtils.h
+++ b/include/dxc/Test/DxcTestUtils.h
@@ -175,6 +175,17 @@ std::string DisassembleProgram(dxc::DxcDllSupport &dllSupport, IDxcBlob *pProgra
 void SplitPassList(LPWSTR pPassesBuffer, std::vector<LPCWSTR> &passes);
 void MultiByteStringToBlob(dxc::DxcDllSupport &dllSupport, const std::string &val, UINT32 codePoint, _Outptr_ IDxcBlob **ppBlob);
 void MultiByteStringToBlob(dxc::DxcDllSupport &dllSupport, const std::string &val, UINT32 codePoint, _Outptr_ IDxcBlobEncoding **ppBlob);
+void ReplaceDisassemblyText(llvm::ArrayRef<LPCSTR> pLookFors,
+                            llvm::ArrayRef<LPCSTR> pReplacements, bool bRegex,
+                            std::string &disassembly);
+void ReplaceDisassemblyText(llvm::ArrayRef<LPCSTR> pLookFors,
+                            llvm::ArrayRef<LPCSTR> pReplacements,
+                            _Outptr_ IDxcBlob **pBlob, bool bRegex,
+                            std::string &disassembly,
+                            dxc::DxcDllSupport &dllSupport);
+void ReplaceDisassemblyTextWithRegex(llvm::ArrayRef<LPCSTR> pLookFors,
+                                     llvm::ArrayRef<LPCSTR> pReplacements,
+                                     std::string &disassembly);
 void Utf8ToBlob(dxc::DxcDllSupport &dllSupport, const std::string &val, _Outptr_ IDxcBlob **ppBlob);
 void Utf8ToBlob(dxc::DxcDllSupport &dllSupport, const std::string &val, _Outptr_ IDxcBlobEncoding **ppBlob);
 void Utf8ToBlob(dxc::DxcDllSupport &dllSupport, const char *pVal, _Outptr_ IDxcBlobEncoding **ppBlob);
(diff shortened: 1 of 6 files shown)


include/dxc/Test/HlslTestUtils.h
```diff
@@ -517,6 +518,44 @@ inline bool CompareHalfEpsilon(const uint16_t &fsrc, const uint16_t &fref, float
...
+    if (pLookFor[0] == '?') {
+      bOptional = true;
+      pLookFor++;
+    }
+    LPCSTR pReplacement = pReplacements[i].data();
+    if (pLookFor && *pLookFor) {
+      
+      bool found = false;
+      size_t pos = 0;
+      size_t lookForLen = strlen(pLookFor);
```
tex3d, line 537: `strlen()` is unnecessary since original std::string object has the length.  Only complication is subtracting one when stripping the `?` from the pLookFor string.
So this could be:
```
    LPCSTR pLookFor = pLookFors[i].data();
    size_t lookForLen = pLookFors[i].size();
    bool bOptional = false;
    if (pLookFor[0] == '?') {
      bOptional = true;
      pLookFor++;
      lookForLen--;
    }
```

include/dxc/Test/HlslTestUtils.h
```diff
@@ -517,6 +518,44 @@ inline bool CompareHalfEpsilon(const uint16_t &fsrc, const uint16_t &fref, float
...
+      size_t replaceLen = strlen(pReplacement);
```
tex3d, line 538: There's actually no need to capture LPCSTR and size from `pReplacements[i]`, since the code could just use the object and size directly: `disassembly.replace(pos, lookForLen, pReplacements[i]);` ... `pos += pReplacements[i].size()`.

tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -11351,36 +11351,20 @@ TEST_F(ExecutionTest, QuadAnyAll) {
...
-  std::vector<std::string> LookForsStorage(LookFors.size());
-    for (unsigned i = 0; i < LookFors.size(); ++i)
-      LookForsStorage[i] = LookFors[i];
-  std::vector<std::string> ReplacementsStorage(Replacements.size());
-    for (unsigned i = 0; i < Replacements.size(); ++i)
-      ReplacementsStorage[i] = Replacements[i];
+  
   auto ShaderInitFn = 
-      [ArgsStorage, LookForsStorage, ReplacementsStorage, &dllSupport]
+      [dxcArgs, &LookFors, &Replacements, &dllSupport]
```
tex3d: It's not safe to use reference for `LookFors` and `Replacements` because the objects are on the argument stack of the MakeShaderReplacementCallback call, which will be outlived by the returned callback function.  Just use a copy (remove `&`), as the original did here for the `Storage` vectors.

include/dxc/Test/DxcTestUtils.h
```diff
@@ -175,10 +175,17 @@ std::string DisassembleProgram(dxc::DxcDllSupport &dllSupport, IDxcBlob *pProgra
...
+void ReplaceDisassemblyText(llvm::ArrayRef<LPCSTR> pLookFors,
+                            llvm::ArrayRef<LPCSTR> pReplacements, bool bRegex,
+                            std::string &disassembly);
+void ReplaceDisassemblyText(llvm::ArrayRef<LPCSTR> pLookFors,
+                            llvm::ArrayRef<LPCSTR> pReplacements,
+                            _Outptr_ IDxcBlob **pBlob, bool bRegex,
+                            std::string &disassembly,
+                            dxc::DxcDllSupport &dllSupport);
+void ReplaceDisassemblyTextWithRegex(llvm::ArrayRef<LPCSTR> pLookFors,
+                                     llvm::ArrayRef<LPCSTR> pReplacements,
```
tex3d, line 187: I'm wondering if these functions that use `ArrayRef<LPCSTR>` could be changed to use `const std::vector<std::string>&` for consistency with the `ReplaceDisassemblyTextWithoutRegex` function, and to avoid llvm::ArrayRef dependency here.  I wonder how many other functions would have to be changed in individual test modules for this.  Perhaps in a follow-up change.

hekota (approved):
LGTM!

tex3d (changes requested):

hekota (changes requested):
Please address Tex' comments.

include/dxc/Test/DxcTestUtils.h
bob80905: Yes, this should be left for a follow-up change.

include/dxc/Test/HlslTestUtils.h
```diff
@@ -12,6 +12,7 @@
 #include <sstream>
 #include <fstream>
 #include <atomic>
+#include <regex>
```
tex3d, line 15: Oops, I don't think this is used in this header, is it?

tex3d (approved):
One leftover `#include <regex>` added that I don't think is necessary.  Other than that, looks good to me.

hekota (approved):
LGTM!

PR #4688, opened 2022-09-26
[SPIR-V] Add error for casting from vector to scalar

### LINKED ISSUE #4640: Internal compiler error when targeting spirv
Sometimes when targeting SPIRV the compiler crashes instead of outputing warnings.

The following shader contains an error, a `float` is passed to a `inout float3` argument. 

```
RWTexture2D<float> g_output;

void foo(inout float3 x)
{
    x += float3(1, 2, 3);
}

[numthreads(1, 1, 1)]
void main(uint3 id : SV_DispatchThreadId)
{
    float bar = 1;
    foo(bar);
    g_output[id.xy] = bar;
}
```

Compiling to DXIL produces the expected error:
```
> dxc.exe -E main -Fo out.bin -T cs_6_0 shader.hlsl
shader.hlsl:3:23: warning: implicit truncation of vector type [-Wconversion]
void foo(inout float3 x)
                      ^
```

Targeting SPIRV crashes the compiler:

```
> dxc.exe -spirv -E main -Fo out.bin -T cs_6_0 shader.hlsl
Internal compiler error: access violation. Attempted to read from address 0x0000000000000028
```

Tested on version:

```
> dxc.exe --version
dxcompiler.dll: 1.7 - 1.6.0.3576 (9395376ef)
```

cassiebeckley:
Fixes #4640

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -2539,7 +2539,7 @@ SpirvInstruction *SpirvEmitter::getBaseOfMemberFunction(
 SpirvInstruction *SpirvEmitter::processCall(const CallExpr *callExpr) {
   const FunctionDecl *callee = getCalleeDefinition(callExpr);
 
-  // Note that we always want the defintion because Stmts/Exprs in the
+  // Note that we always want the definition because Stmts/Exprs in the
   // function body references the parameters in the definition.
   if (!callee) {
     emitError("found undefined function", callExpr->getExprLoc());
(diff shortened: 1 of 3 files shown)


vettoreldaniele (commented):
Looks good, but I believe DXIL only outputs a warning and extends the variable with zero values.

Do we want to match DXIL's behavior here? 

I also get two errors from this - is it intended?
```
../tools/clang/test/CodeGenSPIRV/cast.scalar-to-vec.implicit.hlsl:15:7: error: casting from 'float' to 'float3' unsupported
  foo(bar);
      ^
../tools/clang/test/CodeGenSPIRV/cast.scalar-to-vec.implicit.hlsl:15:7: error: casting from 'float3' to 'float' unsupported
```

vettoreldaniele (changes requested):
Also, I just noticed this fix only applies to `float`. But this behavior seems to be independent of type (see https://shader-playground.timjones.io/f1b2e245672750fae10053e504f16e22 for an example with `int`).

s-perron:
https://github.com/microsoft/DirectXShaderCompiler/pull/5249 will make this obsolete.

PR #4692, opened 2022-09-27
User/jbatista/remove isnormal dependencies02

bob80905:
There are some other functions that are found in the IsNormal HLK test that are defined and declared in the DxcTestUtils files. This is bad because DxcTestUtils takes a dependency on LLVM. This PR gets rid of the rest of the functions that don't need to depend on an llvm include, but are necessary to allow IsNormal to function, and moves the functions to HLSLTestUtils.h. Hopefully after this PR, IsNormal will be able to run when copied to other repos.
These functions and all their dependent functions were moved to HLSLTestUtils.h:
-VerifyCompileOK
-DisassembleProgram
-AssembleToContainer

### DIFF
--- a/include/dxc/Test/HlslTestUtils.h
+++ b/include/dxc/Test/HlslTestUtils.h
@@ -8,6 +8,9 @@
 // Provides utility functions for HLSL tests.                                //
 //                                                                           //
 ///////////////////////////////////////////////////////////////////////////////
+
+// *** THIS FILE CANNOT TAKE ANY LLVM DEPENDENCIES  *** //
+
 #include <string>
 #include <sstream>
 #include <fstream>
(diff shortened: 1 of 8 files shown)


hekota (changes requested):
Please create a new HlslTestUtils.cpp files and move all of the function there, including the ReplaceDisassemblyTextWithoutRegex function from previous PR. Having all of these functions inlined would unnecessary grow the test dll binary size.

include/dxc/Test/HlslTestUtils.h
```diff
@@ -82,553 +82,172 @@ using namespace std;
 #define EXPECT_EQ(expected, actual) VERIFY_ARE_EQUAL(expected, actual)
 #endif 
 #endif // VERIFY_ARE_EQUAL
-
 static constexpr char whitespaceChars[] = " \t\r\n";
-
-inline std::string strltrim(const std::string &value) {
+inline std::string strltrim(const std::string& value) {
```
hekota, line 86: Whitespace-only change

hekota (changes requested):
Please undo the whitespace only changes.

hekota (approved):
LGTM!

PR #4694, opened 2022-09-28
Fix failing test and DxilMutateResourceToHandle case hit by updated test

tex3d:
DxilMutateResourceToHandle would skip GEP if result type does not need to be mutated. But then the GEP would not have setSourceElementType called with the new source type leaving invalid IR.

The fix makes sure the GEP is added to MutateValSet when skipping recursion.


python3kgae (approved):
Need an issue to track commented tests.

tex3d:
Note: related issue found when testing this scenario: #4695

PR #4713, opened 2022-10-09
Switch to dlerror() to catch failures of dlopen()

ruiling:
GetLastError() cannot catch the failures happened during dlopen/dlsym.

### DIFF
--- a/include/dxc/Support/dxcapi.use.h
+++ b/include/dxc/Support/dxcapi.use.h
@@ -28,13 +28,17 @@ protected:
 
 #ifdef _WIN32
     m_dll = LoadLibraryW(dllName);
+    if (m_dll == nullptr) return HRESULT_FROM_WIN32(GetLastError());
 #else
     char nameStr[256];
     std::wcstombs(nameStr, dllName, 256);
     m_dll = ::dlopen(nameStr, RTLD_LAZY);
+    if (!m_dll) {
+      fprintf(stderr, "%s\n", ::dlerror());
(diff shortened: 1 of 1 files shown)


ruiling:
@tex3d @pow2clk Would you mind take a look at the change? I don't know why I cannot edit the "Reviewers".

pow2clk (commented):
Hi @ruiling!

Unfortunately, you can't assign reviewers without write permissions. Tagging us is a good way to get our attention it though. I think this is better suited to my background as I was involved in the initial Linux port. I may have even written this code. I'm not sure. 

I see the problem you are trying to solve, but I don't think printing just there is the answer. In some cases, these are called and the results passed down a few levels before it is properly addressed. In such cases, the `dlerror` string could appear well above where it was really relevant. In addition, in places, the error result should end up in a buffer instead of stderr so the mysterious message in stderr would just confuse people or get completely lost.

I think the appropriate solution is to create an error code that indicates a `dlopen`/`dlsym` related error (or errors) and use those within `GetWin32ErrorMessage` in dxcapi.use.cpp to indicate that `dlerror()` is needed to get the appropriate message. This gets tricky with `ERROR_FILE_NOT_FOUND` because it is used and expected in several places. So, we can't just change its value willy nilly, but the appropriate error for a failure of `dlsym()` is actually `ERROR_PROC_NOT_FOUND`. So that could easily be set in WinAdapter.h to the windows value, used to set the `hr` in the `dlsym()` failure case and handled by `GetWin32ErrorMessage` by retrieving the string from `dlerror()`

pow2clk:
This change is being subsumed by #5024

PR #4736, opened 2022-10-21
Enable -ftime-report flag

llvm-beanz:
This just surfaces and plumbs throug the `-ftime-report` clang flag in
dxc.exe.

`-ftime-report` prints timing data for different parts of the compile
to help identify and track down performance issues. When using the
dxcompiler library interface the output for the time report is returned
as a DXC_OUT_TIME_REPORT buffer.

I've set up the -ftime-report tests to run against the DXC command
line. On Windows this usees the hcttestcmds file, on Linux (and windows
if enabled) this will test through a lit shell suite.

### DIFF
--- a/include/dxc/dxcapi.h
+++ b/include/dxc/dxcapi.h
@@ -480,8 +480,13 @@ typedef enum DXC_OUT_KIND {
   DXC_OUT_TIME_REPORT = 12,   // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
 
   DXC_OUT_LAST = DXC_OUT_TIME_REPORT, // Last value for a counter
+
+  DXC_OUT_NUM_ENUMS,
   DXC_OUT_FORCE_DWORD = 0xFFFFFFFF
 } DXC_OUT_KIND;
+
+static_assert(DXC_OUT_NUM_ENUMS == DXC_OUT_LAST + 1,
+              "DXC_OUT_* Enum added and last value not updated.");
 
 CROSS_PLATFORM_UUIDOF(IDxcResult, "58346CDA-DDE7-4497-9461-6F87AF5E0659")
 struct IDxcResult : public IDxcOperationResult {
--- a/include/dxc/dxcapi.h
+++ b/include/dxc/dxcapi.h
@@ -477,7 +477,9 @@ typedef enum DXC_OUT_KIND {
   DXC_OUT_ROOT_SIGNATURE = 9, // IDxcBlob - Serialized root signature output
   DXC_OUT_EXTRA_OUTPUTS  = 10,// IDxcExtraResults - Extra outputs
   DXC_OUT_REMARKS = 11,       // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
+  DXC_OUT_TIME_REPORT = 12,   // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
 
+  DXC_OUT_LAST = DXC_OUT_TIME_REPORT, // Last value for a counter
   DXC_OUT_FORCE_DWORD = 0xFFFFFFFF
 } DXC_OUT_KIND;
 
--- a/include/dxc/dxcapi.h
+++ b/include/dxc/dxcapi.h
@@ -476,8 +476,8 @@ typedef enum DXC_OUT_KIND {
   DXC_OUT_REFLECTION = 8,     // IDxcBlob - RDAT part with reflection data
   DXC_OUT_ROOT_SIGNATURE = 9, // IDxcBlob - Serialized root signature output
   DXC_OUT_EXTRA_OUTPUTS  = 10,// IDxcExtraResults - Extra outputs
-  DXC_OUT_REMARKS = 11,       // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
-  DXC_OUT_TIME_REPORT = 12,   // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
+  DXC_OUT_REMARKS = 11,       // IDxcBlobUtf8 or IDxcBlobWide - text directed at stdout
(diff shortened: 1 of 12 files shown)


tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1831,6 +1833,9 @@ HRESULT DxcCompilerAdapter::WrapCompile(
     pResult->CopyOutputsFromResult(pImplResult);
     pResult->SetStatusAndPrimaryResult(hr, pImplResult->PrimaryOutput());
 
+    if (opts.TimeReport)
+      llvm::TimerGroup::printAll(outStream);
```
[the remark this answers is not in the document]
llvm-beanz: This should dump the output to the same stream that compiler errors and warnings go out, which I think is what we want because it matches the behavior in Clang.

We could also support mapping to a separate file, which would be clang's `-ftime-report=` option.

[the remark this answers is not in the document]
dmpots: I would think it would be nicer to have it written to a separate DXC_OUT_KIND stream. Either a new one like `DXC_OUT_FTIME` or maybe `DXC_OUT_REMARKS`. The dxc.exe behavior would be the same (writing that output to stdout/stderr), but the API usage would be nicer.

I don't think that necessarily needs to be done for the first checkin, but we should decide if we want to do that soon before people start relying on a particular behavior. What do you think @tex3d?

In any case we should add a compile test that passes the `-ftime arg and reads the data from `DXC_OUT_ERRORS` output (or the separate output if we want to go that route).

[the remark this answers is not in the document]
tex3d: I like the idea of a separate output.  Perhaps `DXC_OUT_TIME_REPORT`?  If the option for separate output is `-ftime-report=` in Clang, how do we separate that from a flag that doesn't set an output file for Dxc?  Could we do something here where `-` means print inline to the stderr output for dxc.exe, while placing a filename here would keep the output separate and dump to the specified file?

include/dxc/dxcapi.h
```diff
@@ -477,7 +477,9 @@ typedef enum DXC_OUT_KIND {
   DXC_OUT_ROOT_SIGNATURE = 9, // IDxcBlob - Serialized root signature output
   DXC_OUT_EXTRA_OUTPUTS  = 10,// IDxcExtraResults - Extra outputs
   DXC_OUT_REMARKS = 11,       // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
+  DXC_OUT_TIME_REPORT = 12,   // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
 
+  DXC_OUT_LAST = DXC_OUT_TIME_REPORT, // Last value for a counter
```
dmpots, line 482: We should add a `static_assert(DXC_OUT_LAST = DXC_OUT_TIME_REPORT)` so that when we add a new enum value we will get a compile-generated reminder to update it.

dmpots (approved):
LGTM. Would be great to also support a file output for the report, but that can alwasy be added as an extension later.

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2812,6 +2813,32 @@ TEST_F(CompilerTest, CompileThenCheckDisplayIncludeProcess) {
 
 }
 
+TEST_F(CompilerTest, CompileThenPrintTimeReport) {
```
dmpots, line 2816: Should we also have a FileChecker test to validate that the output gets embedded in the disassembly output?

include/dxc/dxcapi.h
llvm-beanz: Unfortunately all that will do is cause us to update in tow places. If we add an enum but don't update `DXC_OUT_LAST`'s assignment the static assert will still pass :(

We could add another field that is auto-incremented to handle that case. If this wasn't part of the dxcapi header I'd like to redo the enum ordering and put the `NONE` value at the end. Then we could have `NONE` be the number of valid cases and simplify the indexing using the enum value. I'm not sure how much code that could break since these are part of our ABI 😢 .

dmpots: Ah yeah of course.

So why not make DXC_OUT_LAST autoincrementing then? We could still have a static assert if we wanted to make it clear that we are relying on that to work correctly

`static_assert(DXC_OUT_LAST == DXC_OUT_TIME_REPORT`, "you added a new enum, which is ok, but update this assert to make sure the count is still correct")`

dmpots: Oh, I just saw your update. That way looks good too.

tex3d: I think DXC_OUT_NONE should be a stable value, so making it last+1 isn't good.

One thing that can be done here is to define DXC_OUT_LAST after DXC_OUT_ENUMS as:
`DXC_OUT_LAST = DXC_OUT_ENUMS - 1`

That way you don't have to manually update the value.

include/dxc/dxcapi.h
```diff
@@ -477,10 +477,17 @@ typedef enum DXC_OUT_KIND {
   DXC_OUT_ROOT_SIGNATURE = 9, // IDxcBlob - Serialized root signature output
   DXC_OUT_EXTRA_OUTPUTS  = 10,// IDxcExtraResults - Extra outputs
   DXC_OUT_REMARKS = 11,       // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
+  DXC_OUT_TIME_REPORT = 12,   // IDxcBlobUtf8 or IDxcBlobUtf16 - text directed at stdout
```
tex3d, line 480: The comment on this one and the prior one should say `IDxcBlobWide` instead of `IDxcBlobUtf16`.  I'm not sure how the prior one ended up with the older comment either.

PR #4745, opened 2022-10-27
Made shader reflection return thread group size for AS and MS

adam-yang:
Fixed DxilShaderReflection::GetThreadGroupSize returning only thread group size only for CS but not MS and AS.

### DIFF
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -139,8 +139,6 @@ public:
   TEST_METHOD(CompileWhenWorksThenAddRemovePrivate)
   TEST_METHOD(CompileThenAddCustomDebugName)
   TEST_METHOD(CompileThenTestReflectionWithProgramHeader)
-  TEST_METHOD(CompileThenTestReflectionThreadSizeMS)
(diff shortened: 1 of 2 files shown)


tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -1663,6 +1665,107 @@ TEST_F(CompilerTest, CompileThenTestReflectionWithProgramHeader) {
...
+  CComPtr<ID3D12ShaderReflection> pReflection;
+  VERIFY_SUCCEEDED(pUtils->CreateReflection(&buf, IID_PPV_ARGS(&pReflection)));
+
+  UINT x = 0, y = 0, z = 0;
+  VERIFY_SUCCEEDED(pReflection->GetThreadGroupSize(&x, &y, &z));
+  VERIFY_ARE_EQUAL(x, 2);
+  VERIFY_ARE_EQUAL(y, 3);
+  VERIFY_ARE_EQUAL(z, 1);
+}
+
```
pow2clk, line 1768: Since it's lacking, can we get a CS test in here too? Since all that differs between these two tests is the source and target, it seems we could have a common function that does all this for all three shader stages.

pow2clk (approved):
Thanks!

adam-yang:
Thanks!

PR #4748, opened 2022-10-28
PIX: Fix shader-access-tracking pass' reportage of dynamic indexing

jeffnn:
During a refactor this chunk of code was left inside the for loop over functions, which means that many "DynamicallyIndexedBindPoints" were reported, and PIX just assumed there was one and took the first one. The first one is typically a dx.op and so was quite incorrect.


adam-yang (approved):
LGTM!

PR #4749, opened 2022-10-28
Always LValue cast operands in hlsl CheckVectorConditional for now

tex3d:
CheckVectorConditional does not handle setting ExprValueKind or ExprObjectKind, but should be updated to handle this in the future. Normally the function would cast operands to LValues, but these happened late, so that some return paths skipped this LValue cast. When skipped, this would lead to asserts later in clang code.

This change adds arguments for ExprValueKind and ExprObjectKind so that they may be set properly in the future, and moves the LValue casting earlier, so it's never skipped.

RValue is not preserved in this change due to other assumptions which will be broken, leading to crashes. Supporting RValue expression result will require some other code changes.

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -5038,6 +5038,8 @@ public:
     _In_ ExprResult &Cond,
     _In_ ExprResult &LHS,
     _In_ ExprResult &RHS,
+    _In_ ExprValueKind &VK,
+    _In_ ExprObjectKind &OK,
     _In_ SourceLocation QuestionLoc);
 
   clang::QualType ApplyTypeSpecSignToParsedType(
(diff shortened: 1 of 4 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -9748,6 +9752,22 @@ clang::QualType HLSLExternalSource::CheckVectorConditional(
...
   }
 
+  // TBD: Handle LValues instead of always doing RValue cast.
+  VK = VK_RValue;
+  // TBD: Use OK_BitField and OK_VectorComponent when appropriate?
+  OK = OK_Ordinary;
+
+  // Cast condition to RValue
+  if (Cond.get()->isLValue())
+    Cond.set(CreateLValueToRValueCast(Cond.get()));
```
tex3d, line 9762: One thing to note about the existing code is that if Cond is dependent (and LHS and RHS are not), it will return Cond type as the expression type.  I don't think this is correct, but that's not the bug being fixed in this PR.

damyanp:

PR #4752, opened 2022-10-31
Remove assumption that templates are never UDTs

### LINKED ISSUE #4735: ByteAddressBuffer.Load<T> doesn't work if T contains a templated type
Hi, I ran into an issue with templated loads from ByteAddressBuffer using the latest official release (July 2022). Basically the following works as expected:

```
struct MyStructA
{
    float m_0;
};

struct MyStructB
{
    MyStructA m_a;
    float m_1;
    float m_2;
    float m_3;
};

ByteAddressBuffer g_bab;
RWBuffer<float> result;

[numthreads(1,1,1)]
void CSMain()
{
    MyStructB b = g_bab.Load<MyStructB>(0);
    result[0] = b.m_a.m_0;
}
```

But if MyStructA is a templated struct it fails to compile:

```
template<typename T>
struct MyStructA
{
    float m_0;
};

struct MyStructB
{
    MyStructA<float> m_a;
    float m_1;
    float m_2;
    float m_3;
};

ByteAddressBuffer g_bab;
RWBuffer<float> result;

[numthreads(1,1,1)]
void CSMain()
{
    MyStructB b = g_bab.Load<MyStructB>(0);
    result[0] = b.m_a.m_0;
}
```

This will raise the following error: "Explicit template arguments on intrinsic Load must be a single numeric type"

Is this expected and not supported or is this a bug that can be fixed?

llvm-beanz:
There was an assumtion in the HLSL sema code that a template
specialization could never be a UDT. This assumption is incorrect now.
I've reworked the code so that we instead assume built-in types are
marked as `implicit` (which they all should and seem to be).

Correcting this in `IsHLSLNumericUserDefinedType` resulted in some
breakge in raytracing code generation because we used that method to
deterimine if structures could be payloads or attributes. That was an
incorrect API usage because we do have some builtin types that are
allowed.

The change here does the following:
* Introduce `IsHLSLBuiltinRayAttributeStruct` which returns true for the
  builtin raytracing data types that behave like UDTs.
* Introduce `IsHLSLCopyableAnnotatableRecord` returns true for
  user-defined trivially copyable structures and the builtin ray tracing
  types.
* Adjust `IsHLSLNumericUserDefinedType` to do what the name says.
* Consolidates implementations of `IsUserDefinedRecordType` across
  the project.
* Adds new test cases for the ray tracing built in structs to cover
  diagnostic cases missed by the existing tests.

The new `IsHLSLBuiltinRayAttributeStruct` is hacky and uses the type
names (as the old code did). We should in the future insert an internal
attribute on the types that can be used to denote them so that we don't
need to match string names.

Resolves #4735

### DIFF
--- a/tools/clang/lib/AST/HlslTypes.cpp
+++ b/tools/clang/lib/AST/HlslTypes.cpp
@@ -132,7 +132,8 @@ bool IsHLSLBuiltinRayAttributeStruct(clang::QualType QT) {
   const clang::Type *Ty = QT.getTypePtr();
   if (const RecordType *RT = dyn_cast<RecordType>(Ty)) {
     const RecordDecl *RD = RT->getDecl();
-    if (RD->getName() == "BuiltInTriangleIntersectionAttributes")
+    if (RD->getName() == "BuiltInTriangleIntersectionAttributes" || 
+        RD->getName() == "RayDesc")
       return true;
   }
   return false;
--- a/tools/clang/lib/AST/HlslTypes.cpp
+++ b/tools/clang/lib/AST/HlslTypes.cpp
@@ -111,26 +111,12 @@ bool IsHLSLNumericUserDefinedType(clang::QualType type) {
   const clang::Type *Ty = type.getCanonicalType().getTypePtr();
   if (const RecordType *RT = dyn_cast<RecordType>(Ty)) {
     const RecordDecl *RD = RT->getDecl();
-    if (!IsUserDefinedType(type))
+    if (!IsUserDefinedRecordType(type))
       return false;
     for (auto member : RD->fields()) {
       if (!IsHLSLNumericOrAggregateOfNumericType(member->getType()))
         return false;
     }
-    return true;
-  }
-  return false;
-}
-
-bool IsUserDefinedType(clang::QualType QT) {
-  const clang::Type *Ty = QT.getCanonicalType().getTypePtr();
-  if (const RecordType *RT = dyn_cast<RecordType>(Ty)) {
-    const RecordDecl *RD = RT->getDecl();
-    if (RD->isImplicit())
-      return false;
-    if (auto TD = dyn_cast<ClassTemplateSpecializationDecl>(RD))
-      if (TD->getSpecializedTemplate()->isImplicit())
-        return false;
     return true;
   }
   return false;
@@ -153,7 +139,7 @@ bool IsHLSLAggregateType(clang::QualType type) {
   type = type.getCanonicalType();
   if (isa<clang::ArrayType>(type)) return true;
 
-  return IsUserDefinedType(type);
+  return IsUserDefinedRecordType(type);
 }
 
 clang::QualType GetElementTypeOrType(clang::QualType type) {
@@ -602,23 +588,17 @@ bool IsHLSLSubobjectType(clang::QualType type) {
   return GetHLSLSubobjectKind(type, kind, hgType);
 }
 
-bool IsUserDefinedRecordType(clang::QualType type) {
-  if (const auto *rt = type->getAs<RecordType>()) {
-    // HLSL specific types
-    if (hlsl::IsHLSLResourceType(type) || hlsl::IsHLSLVecMatType(type) ||
-        isa<ExtVectorType>(type.getTypePtr()) || type->isBuiltinType() ||
-        type->isArrayType()) {
+bool IsUserDefinedRecordType(clang::QualType QT) {
+  const clang::Type *Ty = QT.getCanonicalType().getTypePtr();
+  if (const RecordType *RT = dyn_cast<RecordType>(Ty)) {
+    const RecordDecl *RD = RT->getDecl();
+    if (RD->isImplicit())
       return false;
-    }
-
-    // SubpassInput or SubpassInputMS type
-    if (rt->getDecl()->getName() == "SubpassInput" ||
-        rt->getDecl()->getName() == "SubpassInputMS") {
-      return false;
-    }
+    if (auto TD = dyn_cast<ClassTemplateSpecializationDecl>(RD))
+      if (TD->getSpecializedTemplate()->isImplicit())
+        return false;
     return true;
   }
-
   return false;
 }
 
--- a/tools/clang/lib/AST/HlslTypes.cpp
+++ b/tools/clang/lib/AST/HlslTypes.cpp
@@ -111,14 +111,7 @@ bool IsHLSLNumericUserDefinedType(clang::QualType type) {
   const clang::Type *Ty = type.getCanonicalType().getTypePtr();
   if (const RecordType *RT = dyn_cast<RecordType>(Ty)) {
     const RecordDecl *RD = RT->getDecl();
-    if (isa<ClassTemplateSpecializationDecl>(RD)) {
-      return false;   // UDT are not templates
-    }
-    // TODO: avoid check by name
-    StringRef name = RD->getName();
-    if (name == "ByteAddressBuffer" ||
-        name == "RWByteAddressBuffer" ||
-        name == "RaytracingAccelerationStructure")
+    if (!IsUserDefinedType(type))
       return false;
     for (auto member : RD->fields()) {
       if (!IsHLSLNumericOrAggregateOfNumericType(member->getType()))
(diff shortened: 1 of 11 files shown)


tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -129,15 +122,38 @@ bool IsHLSLNumericUserDefinedType(clang::QualType type) {
...
+    const RecordDecl *RD = RT->getDecl();
+    if (RD->isImplicit())
+      return false;
+    if (auto TD = dyn_cast<ClassTemplateSpecializationDecl>(RD))
+      if (TD->getSpecializedTemplate()->isImplicit())
+        return false;
+    return true;
+  }
+  return false;
+}
```
pow2clk, line 137: I feel like we have a lot of these. To wit, we have a very similarly named function `IsUserDefinedRecordType` at around line 589 which uses a very different approach, but seems to have the same intent.

Can you explain the method used here and used previously? Maybe we can consolidate the UDT detection functions?

pow2clk (approved):
Looks good! Go HLSL 2021! 🥳

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -129,15 +122,24 @@ bool IsHLSLNumericUserDefinedType(clang::QualType type) {
   return false;
 }
 
+bool IsHLSLBuiltinRayAttributeStruct(clang::QualType QT) {
+  QT = QT.getCanonicalType();
+  const clang::Type *Ty = QT.getTypePtr();
+  if (const RecordType *RT = dyn_cast<RecordType>(Ty)) {
+    const RecordDecl *RD = RT->getDecl();
+    if (RD->getName() == "RaytracingAccelerationStructure" ||
+        RD->getName() == "BuiltInTriangleIntersectionAttributes")
```
tex3d, line 131: These two shouldn't be treated the same.  The first (`RaytracingAccelerationStructure`) is an opaque resource type, while the second (`BuiltInTriangleIntersectionAttributes`) is a pre-defined structure you can use just like any normal numeric structure, in particular, you may use this for the attribute struct "user-defined-type" in a `[shader("closesthit")]` or other shader that takes hit attributes.

tex3d: `RaytracingAccelerationStructure` should not be legal for use as numeric struct, so that's the one that will be a problem here.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2202,7 +2202,8 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
           rayShaderHaveErrors = true;
         }
         if (ArgNo < 2) {
-          if (!IsHLSLNumericUserDefinedType(parmDecl->getType())) {
+          if (!(IsHLSLNumericUserDefinedType(parmDecl->getType()) ||
+                IsHLSLBuiltinRayAttributeStruct(parmDecl->getType()))) {
```
tex3d, line 2206: `BuiltInTriangleIntersectionAttributes` should be allowed here, but not `RaytracingAccelerationStructure`, since that's an opaque resource type.  So this code shouldn't be treating them the same.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -2472,7 +2473,7 @@ SpirvInstruction *SpirvEmitter::doCallExpr(const CallExpr *callExpr,
             dyn_cast<CXXMethodDecl>(operatorCall->getCalleeDecl())) {
       QualType parentType =
           QualType(cxxMethodDecl->getParent()->getTypeForDecl(), 0);
-      if (isUserDefinedRecordType(astContext, parentType)) {
+      if (hlsl::IsUserDefinedRecordType(parentType)) {
```
tex3d, line 2476: So this will not cover the `BuiltInTriangleIntersectionAttributes`, even though that should be treated like any other ordinary struct type.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -586,23 +587,17 @@ bool IsHLSLSubobjectType(clang::QualType type) {
...
-    if (hlsl::IsHLSLResourceType(type) || hlsl::IsHLSLVecMatType(type) ||
-        isa<ExtVectorType>(type.getTypePtr()) || type->isBuiltinType() ||
-        type->isArrayType()) {
-      return false;
-    }
-
-    // SubpassInput or SubpassInputMS type
-    if (rt->getDecl()->getName() == "SubpassInput" ||
-        rt->getDecl()->getName() == "SubpassInputMS") {
+bool IsUserDefinedRecordType(clang::QualType QT) {
```
tex3d, line 590: This changes the behavior of the function, at least for `BuiltInTriangleIntersectionAttributes`, which would have returned true before.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
tex3d: Ok, nevemind, in this case, `BuiltInTriangleIntersectionAttributes` can't have methods, so it can't be parentType.

tools/clang/lib/AST/HlslTypes.cpp
tex3d: Upon review, I think this change is ok for this function based on the name, and based on a survey of uses.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -129,15 +122,28 @@ bool IsHLSLNumericUserDefinedType(clang::QualType type) {
   return false;
 }
 
+bool IsHLSLNumericAggregate(clang::QualType QT) {
```
tex3d, line 125: I don't think the name captures the right intent.

The purpose in uses of this function is to verify that we have a struct or class type (record type that will map to a struct type in llvm to which we can associate type annotation metadata) that only contains numeric types (isn't and doesn't contain a special HLSL object type).  Matrix, vector, and array types should not be included as the top-level type, but are fine as members of the top-level type.

Maybe `IsHLSLNumericRecordType` or `IsHLSLNumericStructOrClassType` would be more appropriate?

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -111,14 +111,7 @@ bool IsHLSLNumericUserDefinedType(clang::QualType type) {
...
     const RecordDecl *RD = RT->getDecl();
-    if (isa<ClassTemplateSpecializationDecl>(RD)) {
-      return false;   // UDT are not templates
-    }
-    // TODO: avoid check by name
-    StringRef name = RD->getName();
-    if (name == "ByteAddressBuffer" ||
-        name == "RWByteAddressBuffer" ||
-        name == "RaytracingAccelerationStructure")
+    if (!IsUserDefinedRecordType(type))
```
tex3d, line 114: This will exclude use of `BuiltInTriangleIntersectionAttributes`, which should return true for `IsHLSLNumericOrAggregateOfNumericType`.

tex3d: Will this also exclude built-in matrix or vector types?  That would not be good.

tools/clang/lib/AST/HlslTypes.cpp
llvm-beanz: That naming doesn't quite work either. Vector and Matrix types are Record types in DXC too.

tools/clang/lib/AST/HlslTypes.cpp
llvm-beanz: I think excluding matrix and vector types here is what we want because we exclude those in some places since they can't be annotated. If the use case wants to allow matrix and vectors, you call `IsHLSLNumericOrAggregateOfNumericType` which will return true for matrix and vector types. If you want to disallow those you call `IsHLSLNumericUserDefinedType`, which returns false for those cases.

tex3d: Right, I think the vector/matrix were excluded by being template specializations before.

tex3d (approved):
I don't love the name `IsHLSLCopyableAnnotatableRecord`, but don't think I have a perfect one in mind to replace it either.

PR #4754, opened 2022-11-01
IsNormalTest - correct Disassemble arguments

hekota:
IDxcResult needs to be used to get to the disassembled text.

### DIFF
--- a/tools/clang/unittests/HLSL/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSL/ExecutionTest.cpp
@@ -11389,9 +11389,11 @@ st::ShaderOpTest::TShaderCallbackFn MakeShaderReplacementCallback(
     // Disassemble
     std::string disassembly;
     {
+      CComPtr<IDxcResult> pDisassemblyResult;
       CComPtr<IDxcBlobUtf8> pDisassembly;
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSL/ExecutionTest.cpp
```diff
@@ -11389,9 +11389,11 @@ st::ShaderOpTest::TShaderCallbackFn MakeShaderReplacementCallback(
     // Disassemble
     std::string disassembly;
     {
+      CComPtr<IDxcResult> pDisassemblyResult;
       CComPtr<IDxcBlobUtf8> pDisassembly;
       DxcBuffer compiledBuffer = {compiledShader->GetBufferPointer(), compiledShader->GetBufferSize(), 0};
-      VERIFY_SUCCEEDED(pCompiler->Disassemble(&compiledBuffer, IID_PPV_ARGS(&pDisassembly)));
+      VERIFY_SUCCEEDED(pCompiler->Disassemble(&compiledBuffer, IID_PPV_ARGS(&pDisassemblyResult)));
```
pow2clk, line 11395: This is clearly right. I wonder why this didn't fail. It looks as though the implementation checks that the parameter is the right type.  Maybe it did?

pow2clk (approved):
Looks good!

PR #4757, opened 2022-11-02
Added constant folding for some DXIL operations with non-finite operands

adam-yang:
Constant folder for DXIL operations with non-finite operands (inf, -inf, NaN) was intentionally left undone. This change adds constant folding for some of the missing cases based on the D3D spec. This change does not include all the operations. Some of them will be added later.


damyanp:

PR #4759, opened 2022-11-02
Bring in code coverage tooling

llvm-beanz:
This cherry-picks a few changes from upstream LLVM, and has one small DXC-specific change on top.

The goal here is to just bring in the basic code coverage tooling so that we can generate coverage reports. With this PR coverage reports on Unix work, but coverage reports on Windows are still WIP.

Looping in @dnovillo and @bogner since they both worked on LLVM's code coverage infrastructure in past lives 😁

### DIFF
--- a/cmake/modules/CoverageReport.cmake
+++ b/cmake/modules/CoverageReport.cmake
@@ -1,0 +1,64 @@
+# if coverage reports are not enabled, skip all of this
+if(NOT LLVM_BUILD_INSTRUMENTED_COVERAGE)
+  return()
+endif()
+
+file(TO_NATIVE_PATH
+     "${LLVM_SOURCE_DIR}/utils/prepare-code-coverage-artifact.py"
+     PREPARE_CODE_COV_ARTIFACT)
+
+# llvm-cov and llvm-profdata need to match the host compiler. They can either be
+# explicitly provided by the user, or we will look them up based on the install
+# location of the C++ compiler.
+get_filename_component(COMPILER_DIRECTORY ${CMAKE_CXX_COMPILER} DIRECTORY)
+find_program(LLVM_COV "llvm-cov" ${COMPILER_DIRECTORY} NO_DEFAULT_PATH)
+find_program(LLVM_PROFDATA "llvm-profdata" ${COMPILER_DIRECTORY} NO_DEFAULT_PATH)
+
+if(NOT LLVM_COV OR NOT LLVM_PROFDATA)
+  message(WARNING "Could not find code coverage tools, skipping generating targets. You may explicitly specify LLVM_COV and LLVM_PROFDATA to work around this warning.")
+  return()
+endif()
+
+set(LLVM_CODE_COVERAGE_TARGETS "" CACHE STRING "Targets to run code coverage on (defaults to all exported targets if empty)")
+mark_as_advanced(LLVM_CODE_COVERAGE_TARGETS)
+
+if(NOT LLVM_CODE_COVERAGE_TARGETS)
+  # by default run the coverage report across all the exports provided
+  get_property(COV_TARGETS GLOBAL PROPERTY LLVM_EXPORTS)
+endif()
+
+file(TO_NATIVE_PATH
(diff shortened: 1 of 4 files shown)


hekota (approved):
LGTM, though it would be good if someone with LLVM's code coverage infrastructure knowledge could take a look as well.

cmake/modules/CoverageReport.cmake
```diff
@@ -0,0 +1,73 @@
...
+  message(WARNING "Could not find code coverage tools, skipping generating targets. You may explicitly specify LLVM_COV and LLVM_PROFDATA to work around this warning.")
+  return()
+endif()
+
+set(LLVM_CODE_COVERAGE_TARGETS "" CACHE STRING "Targets to run code coverage on (defaults to all exported targets if empty)")
+mark_as_advanced(LLVM_CODE_COVERAGE_TARGETS)
+
+# HLSL Change Begin - This is probably worth upstreaming...
+set(LLVM_CODE_COVERAGE_TEST_TARGETS "" CACHE STRING "Targets to run to generate coverage profiles.")
+mark_as_advanced(LLVM_CODE_COVERAGE_TEST_TARGETS)
```
dnovillo, line 27: In the past, I remember using a bootstrap process to generate profiles.  So the target would be cc1/cc1plus itself.  In here, isn't it just `dxc`?

cmake/modules/CoverageReport.cmake
```diff
@@ -0,0 +1,73 @@
...
+
+set(LLVM_COVERAGE_SOURCE_DIRS "" CACHE STRING "Source directories to restrict coverage reports to.")
+mark_as_advanced(LLVM_COVERAGE_SOURCE_DIRS)
+
+foreach(dir ${LLVM_COVERAGE_SOURCE_DIRS})
+  list(APPEND restrict_flags -restrict ${dir})
+endforeach()
+
+# Utility target to clear out profile data.
+# This isn't connected to any dependencies because it is a bit finacky to get
```
dnovillo, line 58: nit: s/finacky/finicky/

llvm-beanz: lol... I spelled that wrong in upstream too... I guess I should fix that 😓

cmake/modules/CoverageReport.cmake
llvm-beanz: We don't (yet) have any build targets inside the CMake build that run `dxc` or its tests. For code coverage the goal is really to get the coverage numbers from running the tests, which at the moment run through external testing harnesses, but will soon run under CMake and Lit (#4762).

llvm-beanz: In my local runs I set `LLVM_CODE_COVERAGE_TARGETS` to `dxc;dxcompiler`, which constrains the coverage report to just DXC and the compiler dynamic library, but we instrument all the binaries so that the test binaries report coverage for the code they load from `dxcompiler`.

dnovillo (approved):

cmake/modules/CoverageReport.cmake
```diff
@@ -61,4 +70,4 @@ add_custom_target(generate-coverage-report
                           ${REPORT_DIR} ${coverage_binaries}
                           --unified-report ${restrict_flags}
                   WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-                  DEPENDS check-llvm) # Run tests
+                  ${COV_DEPENDS}) # Run tests
```
bogner, line 73: This looks like it changes the default behaviour, since `COV_DEPENDS` is empty if `LLVM_CODE_COVERAGE_TEST_TARGETS` isn't set.

cmake/modules/CoverageReport.cmake
bogner: Wait, is `LLVM_CODE_COVERAGE_TARGETS` meant to be targets like `dxc` or something along the lines of `check-dxc`? Your explanation here sounds like the former, but the logic it replaces is more like the latter.

PR #4760, opened 2022-11-02
Allow exception when compile shader for hash test.

python3kgae:
For case the shader is expect to be fail in FileCheck.

### DIFF
--- a/tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
+++ b/tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
@@ -409,8 +409,13 @@ FileRunCommandResult FileRunCommandPart::RunDxcHashTest(dxc::DxcDllSupport &DllS
   CComPtr<IDxcBlob> pOriginalHash;
   // If failed the original compilation, just pass the test. The original test was likely
   // testing for failure.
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
```diff
@@ -409,8 +409,13 @@ FileRunCommandResult FileRunCommandPart::RunDxcHashTest(dxc::DxcDllSupport &DllS
   CComPtr<IDxcBlob> pOriginalHash;
   // If failed the original compilation, just pass the test. The original test was likely
   // testing for failure.
-  if (FAILED(CompileForHash(opts, CommandFileName, DllSupport, original_flags, &pOriginalHash, originalOutput)))
+  try {
+    if (FAILED(CompileForHash(opts, CommandFileName, DllSupport, original_flags,
+                              &pOriginalHash, originalOutput)))
+      return FileRunCommandResult::Success();
+  } catch (...) {
```
tex3d, line 416: I don't think exceptions from test should be translated to success here.  Exceptions at this point indicate broken test infrastructure or bug in API, not the kind of failure condition that a test might be expecting.

tex3d (changes requested):
I'm not convinced this change is necessary or desired at this point.

python3kgae:
Cannot repro the issue after a clean build.

PR #4761, opened 2022-11-02
Require C++ 17 for DXC

llvm-beanz:
This change moves DXC to C++ 17.

The primary motivator for this is enabling building DXC with clang-cl. Some of the Windows headers use C++ 17 language features and MSVC allows those features in older language modes without issuing a diagnostic.

We _could_ enable this only for clang-cl under Windows, but it is probably better to just move the requirement and unlock new language features. C++ 17 does deprecate and remove features from earlier C++ versions. This PR updates the removals but not the deprecations.

### DIFF
--- a/cmake/modules/HandleLLVMOptions.cmake
+++ b/cmake/modules/HandleLLVMOptions.cmake
@@ -261,6 +261,7 @@ endif()
 
 if( MSVC )
   include(ChooseMSVCCRT)
+  set(CMAKE_CXX_STANDARD 17)
 
   if( NOT (${CMAKE_VERSION} VERSION_LESS 2.8.11) )
     # set stack reserved size to ~10MB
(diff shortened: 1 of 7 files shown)


include/llvm/ADT/STLExtras.h
```diff
@@ -200,6 +211,91 @@ inline mapped_iterator<ItTy, FuncTy> map_iterator(const ItTy &I, FuncTy F) {
   return mapped_iterator<ItTy, FuncTy>(I, F);
 }
 
+/// An iterator adaptor that filters the elements of given inner iterators.
```
pow2clk, line 214: I don't follow why this and the tests below were required. It doesn't look like they are used outside of the tests?

pow2clk (commented):
I think we are fine to require 17 at this point. The default compilers we expect people to use all support it now. 

I am confused by some of the inclusions. Perhaps them came along with some of the other code in merged changes?

Looks like this will fix many of the warnings-as-errors failures: https://github.com/llvm/llvm-project/commit/91c8daf3009209a63534b4312d74322e1399bba1

include/llvm/ADT/STLExtras.h
llvm-beanz: I pulled this change in because the other changes conflicted without it. It isn’t necessary, but it made merging easier.

cmake/modules/HandleLLVMOptions.cmake
```diff
@@ -453,23 +454,22 @@ elseif( LLVM_COMPILER_IS_GCC_COMPATIBLE )
...
-      endif()
+  # HLSL Change Begin - Require C++ 17
+  check_cxx_compiler_flag("-std=c++17" CXX_SUPPORTS_CXX17)
+  if (CXX_SUPPORTS_CXX17)
+    # TODO: Remove this next line and fix the issues.
+    add_definitions(-D_LIBCPP_DISABLE_DEPRECATION_WARNINGS)
+    if (CYGWIN OR MINGW)
+      # MinGW and Cygwin are a bit stricter and lack things like
+      # 'strdup', 'stricmp', etc in c++11 mode.
+      append("-std=gnu++17" CMAKE_CXX_FLAGS)
```
pow2clk, line 465: This was blindly updated to 14 from 11 but for the comment. I'm not certain it applies anymore and we don't really support wingw or cygwin builds anyway. Maybe we could just excise this part.

cmake/modules/HandleLLVMOptions.cmake
```diff
@@ -453,23 +454,22 @@ elseif( LLVM_COMPILER_IS_GCC_COMPATIBLE )
...
-        # MinGW and Cygwin are a bit stricter and lack things like
-        # 'strdup', 'stricmp', etc in c++11 mode.
-        append("-std=gnu++14" CMAKE_CXX_FLAGS)
-      else()
-        append("-std=c++14" CMAKE_CXX_FLAGS)
-      endif()
+  # HLSL Change Begin - Require C++ 17
+  check_cxx_compiler_flag("-std=c++17" CXX_SUPPORTS_CXX17)
+  if (CXX_SUPPORTS_CXX17)
+    # TODO: Remove this next line and fix the issues.
```
pow2clk, line 460: Can we get a bug so we don't forget to fix these issues?

pow2clk (approved):
Just a comment you can take or leave and a request for another bug.

PR #4762, opened 2022-11-03
Support taef run in lit.

python3kgae:
lit.formats.taef is added to run taef under lit.

New lit target check-clang-hlsl is added to run clang-hlsl-test taef tests.


tools/clang/test/CMakeLists.txt
```diff
@@ -19,6 +19,11 @@ configure_lit_site_cfg(
   ${CMAKE_CURRENT_BINARY_DIR}/Unit/lit.site.cfg
   )
 
+configure_lit_site_cfg(
+  ${CMAKE_CURRENT_SOURCE_DIR}/clang-hlsl-tests/lit.site.cfg.in
```
llvm-beanz, line 23: nit: rather than calling the suite `clang-hlsl-tests` maybe just `TAEF`?

tools/clang/test/CMakeLists.txt
```diff
@@ -92,3 +97,13 @@ add_lit_target("check-clang-unit" "Running lit suite clang-unit"
           ARGS ${CLANG_TEST_EXTRA_ARGS}
         )
 # HLSL Change End - Add a separate target for clang unit tests
+
+# HLSL Change Begin - Add taef tests
+add_lit_target("check-clang-hlsl" "Running lit suite clang-hlsl"
```
llvm-beanz, line 102: by convention the `check-clang-*` targets refer to the directory containing the lit suite. This is important because there is tooling to auto-generate build targets based on the subdirectory. This name will conflict with an existing subdirectory under tests, so it should probably be `taef` instead.

utils/lit/lit/formats/taef.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import absolute_import
+import os
+import sys
+
+import lit.Test
+import lit.TestRunner
+import lit.util
+from .base import TestFormat
+
+kIsWindows = sys.platform in ['win32', 'cygwin']
```
llvm-beanz, line 10: this is always going to be true for TAEF

utils/lit/lit/formats/taef.py
```diff
@@ -0,0 +1,95 @@
...
+          localConfig: TestingConfig instance"""
+
+        # TE:F:\repos\DxcGitHub\hlsl.bin\TAEF\x64\te.exe
+        # test dll : F:\repos\DxcGitHub\hlsl.bin\Debug\test\clang-hlsl-tests.dll
+        # /list
+
+        try:
+            lines = lit.util.capture([self.te, self.test_dll, '/list'],
+                                     env=localConfig.environment)
+            if kIsWindows:
```
llvm-beanz, line 35: This can collapse since this will always be true.

utils/lit/lit/formats/taef.py
```diff
@@ -0,0 +1,95 @@
...
+              lines = lines.replace('\r', '')
+            lines = lines.split('\n')
+        except:
+            litConfig.error("unable to discover taef in %r" % path)
+            raise StopIteration
+
+        nested_tests = []
```
llvm-beanz, line 42: This is copied from the google test implementation (which I get), but this makes no sense here. Googletest does this because the test listing doesn't print the full test name so it need so combine partial strings. Here, the `nested_tests` variable is always an empty array.

llvm-beanz:
Something else that occurred to me. It might be better if the suite detected files ending in `-tests.dll` similar to how the google test format detects all the executables ending in `Tests[.exe]`. That would make it so that (1) it works for all the TAEF dlls (we have dxil-conv-tests too), and (2) if we add more in the future those would get detected automatically without needing to also modify the testing configuration.

llvm-beanz (approved):
LGTM

PR #4764, opened 2022-11-04
PIX: Switch back to non-raw buffer stores

jeffnn:
A recent change switched to RawBufferStore, because it's a raw UAV at the API. Some GPUs' drivers were upset by this form of store. Spent quite some time trying to modify parameters to please the driver, but to no avail. Since it doesn't really matter which store operand is used, I switched back to BufferStore, and now all (known) drivers are happy.


adam-yang (approved):
LGTM!

PR #4768, opened 2022-11-04
Avoid > in FileCheck run line.

python3kgae:
Things like -DResType=RasterizerOrderedStructuredBuffer<uint> could be workaround with -DResType="RasterizerOrderedStructuredBuffer<uint>" for lit. But clang-hlsl-tests will not work.
So put the template inside the test.


pow2clk (approved):
Looks good! It's kind of a disappointing limitation, but it's one we can deal with.

PR #4776, opened 2022-11-08
misc: fix clang warnings: deprecated,bitwise-logic

Keenuts:
Fixing some clang warnings about mixing logical and bitwise and the use of one deprecated CPP type (Removed in cpp17, same change as seen in https://github.com/microsoft/DirectXShaderCompiler/pull/4761).

Some are using those bitwise operation to avoid operator laziness. It was fixed upstream by an int cast, applying the same fixes.


pow2clk (approved):
Looks good

PR #4781, opened 2022-11-10
bitfields: refactorize struct offset computation

Keenuts:
This is some preliminary work to bitfield addition into the SPIR-V backend. This commit should not change the behavior of our backend.

The next steps of the bitfield feature can be seen here : https://github.com/Keenuts/DirectXShaderCompiler/tree/bitfields
(state of this branch not guaranteed, as it's my working branch)


cassiebeckley (approved):
LGTM! Thanks for making this more readable

PR #4783, opened 2022-11-10
[SPIR-V] Handle structured conditions in for and while loops

### LINKED ISSUE #4714: SPIR-V degradation 1.7.2207.10069 compare to 1.7.2207.10036
After upgrading from _1.7.2207.10036_ to _1.7.2207.10069_ I got pixel shader compilation error:

```txt
C:\Users\Goshido\Desktop\1>"D:\Development\DXC-builds\Release-2022-10-09\bin\dxc.exe" -HV 2021 -spirv -fvk-use-dx-layout -fspv-reduce-load-size -fspv-target-env=vulkan1.1 -enable-16bit-types -WX -O3 -T ps_6_7 -E PS -Fo mandelbrot-lut-color-ps.spv mandelbrot-lut-color.ps
fatal error: generated SPIR-V is invalid: Back-edges (35[%35] -> 32[%32]) can only be formed between a block and a loop header.
  %35 = OpLabel

C:\Users\Goshido\Desktop\1>"D:\Development\DXC-builds\Release-2022-10-09\bin\dxc.exe" -HV 2021 -spirv -fvk-use-dx-layout -fspv-reduce-load-size -fspv-target-env=vulkan1.1 -enable-16bit-types -WX -O3 -T ps_6_7 -E PS -Fo mandelbrot-analytic-color-ps.spv mandelbrot-analytic-color.ps
fatal error: generated SPIR-V is invalid: Back-edges (37[%37] -> 34[%34]) can only be formed between a block and a loop header.
  %37 = OpLabel

```

Compilation to _SPIR-V_ worked previously in version _1.7.2207.10036_.

### Minimal repro steps

I attached minimal reproducible example with pixel shader sources and `bat` script with all compilation parameters. You need to change `DXC_LOCATION` variable inside the `build-test.bat` to _dxc.exe v1.7.2207.10069_ executable.

[Direct link to issue-001.zip](https://github.com/microsoft/DirectXShaderCompiler/files/9741870/issue-001.zip)

### Expected result

Compilation success. _DXC_ produces _SPIR-V_ binaries.

### Actual result

Compilation error:

```txt
C:\Users\Goshido\Desktop\1>"D:\Development\DXC-builds\Release-2022-10-09\bin\dxc.exe" -HV 2021 -spirv -fvk-use-dx-layout -fspv-reduce-load-size -fspv-target-env=vulkan1.1 -enable-16bit-types -WX -O3 -T ps_6_7 -E PS -Fo mandelbrot-lut-color-ps.spv mandelbrot-lut-color.ps
fatal error: generated SPIR-V is invalid: Back-edges (35[%35] -> 32[%32]) can only be formed between a block and a loop header.
  %35 = OpLabel

C:\Users\Goshido\Desktop\1>"D:\Development\DXC-builds\Release-2022-10-09\bin\dxc.exe" -HV 2021 -spirv -fvk-use-dx-layout -fspv-reduce-load-size -fspv-target-env=vulkan1.1 -enable-16bit-types -WX -O3 -T ps_6_7 -E PS -Fo mandelbrot-analytic-color-ps.spv mandelbrot-analytic-color.ps
fatal error: generated SPIR-V is invalid: Back-edges (37[%37] -> 34[%34]) can only be formed between a block and a loop header.
  %37 = OpLabel

```

No _SPIR-V binaries._

### Further technical details

- Commit _SHA-1_: `81ef064a47a191c9fc68a43adca4706244c3687b`
- `git submodule update --init` has been invoked before build
- `hctbuild` build parameters: `-official -rel -x64 -vs2022 -speak-up -no-dxilconv -spirv`
- _Visual Studio 2022 Community_ components were up to date [2022/10/09].
- _Windows 11 Pro 64-bit (10.0 Build 22000, Version 21H2)_

Would you kindly help?

vettoreldaniele:
With the introduction of short-circuited operators in HLSL 2021, using them as a loop condition currently results in invalid SPIR-V.

The problem lies in the fact that short-circuited operators introduce control flow. Basically, `a && b` becomes equivalent to `bool result = false; if (a) result = b;`. 

This creates a problem when such a construct is used as a loop condition. When generating the loop in SPIR-V, we take the basic block that performs the condition check and reuse it as a loop header block instead of having a bespoke header block. This shortcut only works when the condition can be expressed in a single basic block, which for short-circuited conditions, it can't.

This fix reworks how we generate loop conditions and headers (both for `for` and `while` loops) when we detect such conditions. We want to move the condition check within the loop, basically transforming a loop 
```
for (int i = 0; condition; ++i) {
   // loop body
}
```
into
```
for (int i = 0; /* always true */; ++i) {
   if (!condition) {
       break;
   }
   // loop body
}
```

The reason we don't unconditionally apply this transformation, which is technically always legal, is because it prevents loop unrolling in SPIR-V Tools, which does not support unrolling loops with early breaks ([code](https://github.com/KhronosGroup/SPIRV-Tools/blob/d35a78d/source/opt/loop_unroller.cpp#L1027)). As a future improvement, we should probably enhance the unroller to handle this case, and always apply this transformation in DXC (as an example, glslang always does this).

Fixes #4714.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -2000,14 +2000,50 @@ void SpirvEmitter::doWhileStmt(const WhileStmt *whileStmt,
   //     | merge |
   //     +-------+
   //
+  // The only exception is when the condition cannot be expressed in a single
+  // block. Specifically, short-circuited operators end up producing multiple
+  // blocks. In that case, we cannot treat the <check> block as the header
+  // block, and must instead have a bespoke <header> block. The condition is
+  // then moved into the loop. For example, given a loop in the form
+  //   while (a && b) { <body> }
+  // we will generate instructions for the equivalent loop
+  //   while (true) { if (!(a && b)) { break }  <body> }
+  //            +----------+
+  //            |  header  | <------------------+
+  //            +----------+                    |
+  //                 |                          |
+  //                 v                          |
+  //            +----------+                    |
+  //            |  check   |                    |
+  //            +----------+                    |
+  //                 |                          |
+  //         +-------+-------+                  |
+  //         | false         | true             |
+  //         |               v                  |
+  //         |            +------+     +------------------+
+  //         |            | body | --> | continue (no-op) |
+  //         v            +------+     +------------------+
+  //     +-------+
+  //     | merge |
+  //     +-------+
+  // The reason we don't unconditionally apply this transformation, which is
+  // technically always legal, is because it prevents loop unrolling in SPIR-V
+  // Tools, which does not support unrolling loops with early breaks.
   // For more details, see "2.11. Structured Control Flow" in the SPIR-V spec.
 
   const spv::LoopControlMask loopControl =
       attrs.empty() ? spv::LoopControlMask::MaskNone
                     : translateLoopAttribute(whileStmt, *attrs.front());
 
+  const Expr *check = whileStmt->getCond();
+  const Stmt *body = whileStmt->getBody();
+  bool checkHasShortcircuitedOp = stmtTreeContainsShortCircuitedOp(check);
+
   // Create basic blocks
   auto *checkBB = spvBuilder.createBasicBlock("while.check");
+  auto *headerBB = checkHasShortcircuitedOp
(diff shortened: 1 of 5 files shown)


vettoreldaniele:

cassiebeckley (approved):
LGTM

PR #4785, opened 2022-11-11
Fixed bug where MergeGepUse doesn't return true

adam-yang:
Variable `changed` should be set to true as soon as `MergeGEP` returns a value.


tex3d (approved):
Good catch!

PR #4793, opened 2022-11-16
[CMake] Simplify code coverage CMake configuration

llvm-beanz:
This just adds an option processed by the CMake cache file to configure code coverage for DXC. With this patch you can generate a coverage report locally if you build with Clang using the following commnads:

```
cmake -G Ninja -DDXC_COVERAGE=On <other options> -C <path to dxc>/cmake/caches/PredefinedParams.cmake <path to dxc>
ninja
ctest
ninja generate-coverage-report
```

This removes the need to separately specify `LLVM_BUILD_INSTRUMENTED_COVERAGE`, `LLVM_PROFILE_DATA_DIR`, and `LLVM_CODE_COVERAGE_TARGETS` manually as well as increases the robustness of detecting `llvm-cov` and `llvm-profdata` on Linux.


cmake/caches/PredefinedParams.cmake
```diff
@@ -1,3 +1,9 @@
+if (DXC_COVERAGE)
```
dmpots, line 1: Should this have the HLSL change marker comment too?

dmpots (approved):
Might be worth a short comment in the commit message about how its better than before (e.g. what did you have to do before to get code coverage).

cmake/caches/PredefinedParams.cmake
llvm-beanz: That file is HLSL-only.

llvm-beanz: I updated the file with a comment at the top making clear it is an HLSL-only file and describing how CMake cache scripts are used.

PR #4794, opened 2022-11-16
Disable death test for assert in dxc not work with death test.

python3kgae:
If use normal assert, all death tests will pass except ValueHandle.AssertingVH_Asserts.


llvm-beanz:
Can you please provide a list of which test cases fail without this change?

This change doesn't just disable test cases that test for death, but also the ability to test for death more broadly. That is likely okay given how DXC converts death-cases to exceptions, but it would be good to know which tests specifically rely on testing for death to ensure we're not losing something important.

llvm-beanz:
These tests seem to pass on Linux as-is. Maybe we should restrict this change to Windows only?

PR #4796, opened 2022-11-17
Create action to generate code coverage

chrisbieneman:
This is a first stab at generating a github action to generate code coverage reports.

This is built using GitHub actions because it streamlines publishing to GitHub Pages... hopefully this works.


hekota (approved):
I'm not entirely familiar with the GitHub workflow and pages, but it looks to be doing what is says :)

PR #4798, opened 2022-11-18
Enable running unit tests without extra arguments

llvm-beanz:
This change provides a default value based on the build configuration which can be used in the absence of a specified HLSL Data Dir when running the tests. This does not change any existing functionality, but enables the common configuration of running the tests inside the configured source directory to omit specifiying a path.


pow2clk (approved):
I know there was some disagreement about this and we can alter this behavior later if desired, but for now this unblocks some valuable additional testing and I'm not inclined to wait.

PR #4799, opened 2022-11-18
TAEF Tests only work on Windows

llvm-beanz:
Disable the TAEF lit suite on non-Windows platforms.

### DIFF
--- a/tools/clang/test/clang-hlsl/lit.cfg
+++ b/tools/clang/test/clang-hlsl/lit.cfg
@@ -35,3 +35,6 @@ test_dir = os.path.join(config.llvm_obj_root, config.build_mode, 'test')
 #projects\dxilconv\test
 config.test_format = lit.formats.TaefTest(te, test_dll, hlsl_data_dir, test_dir)
 
+# TAEF only runs on Windows
+if platform.system() != 'Windows':
+  config.unsupported = True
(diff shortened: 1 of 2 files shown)


tools/clang/test/clang-hlsl/lit.cfg
```diff
@@ -35,3 +35,6 @@ test_dir = os.path.join(config.llvm_obj_root, config.build_mode, 'test')
 #projects\dxilconv\test
 config.test_format = lit.formats.TaefTest(te, test_dll, hlsl_data_dir, test_dir)
 
+# TAEF only runs on Windows
+if platform.system() != 'Windows':
+  config.unsupported = True
```
pow2clk, line 40: Is this necessary? Won't this get ignored per the changes to the above file?

Will we be using this config to run other tests through lit on non-windows platforms going forward even though, as I read it, all it does for now is TAEF?

llvm-beanz: The CMake change prevents generating the target to run the lit suite, this change marks the suite unsupported so that if the suite is run along other suites (as check-clang or check-all will hopefully work), it is still disabled.

pow2clk (approved):
I'm satisifed

PR #4800, opened 2022-11-18
Fix failing LLVM ArrayRef unit test

llvm-beanz:
The initializer list goes out of scope after the expression it is used in, so the ArrayRef points to invalid memory. This change keeps the initialization list around so that the array ref remains valid through the test.


dmpots (approved):
LGTM. I assume this was fixed upstream as well? It doesn't look like a test we added.

llvm-beanz:
> LGTM. I assume this was fixed upstream as well? It doesn't look like a test we added.

Yep. Not sure at what point it got fixed, but sanitizers would have hit it even if compilers got lucky.

PR #4802, opened 2022-11-18
Enable SPIR-V tests to run without arguments

llvm-beanz:
This specifies a default value for the --spirv-test-root flag based on the source directory at configuration time. This simplifies the command line for running the tests manually, but does not alter the ability to override the test directory explicitly.

### DIFF
--- a/tools/clang/unittests/SPIRV/TestMain.cpp
+++ b/tools/clang/unittests/SPIRV/TestMain.cpp
@@ -103,9 +103,6 @@ int main(int argc, char **argv) {
       // Allow the user set the root directory for test input files.
       if (i + 1 < argc) {
         clang::spirv::testOptions::inputDataDir = argv[++i];
-      } else {
-        fprintf(stderr, "Error: --spirv-test-root requires an argument\n");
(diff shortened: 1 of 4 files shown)


tools/clang/unittests/SPIRV/TestMain.cpp
```diff
@@ -103,9 +103,6 @@ int main(int argc, char **argv) {
       // Allow the user set the root directory for test input files.
       if (i + 1 < argc) {
         clang::spirv::testOptions::inputDataDir = argv[++i];
-      } else {
```
vettoreldaniele, line 106: This check should stay, since it's only triggered if somebody passes `--spirv-test-root` without a following argument. It should be an error not providing an argument to the option (if somebody goes as far as providing the option)

llvm-beanz: doh... of course.

vettoreldaniele (approved):
Thanks!

cassiebeckley (approved):
LGTM

PR #4803, opened 2022-11-18
Don't emit built-in 'vk' namespace in DeclPrinter

LukasBanana:
The rewriter emits the internal `namespace vk` in `DeclPrinter` when `-spirv` is specified and `RewriterOpts::RemoveUnusedGlobals` is enabled.
This results in HLSL parse errors because `long` and `extern` are reserved keywords amongst other follow-up errors:
```
input.hlsl:4:33: error: 'long' is a reserved keyword in HLSL
	extern T RawBufferLoad(unsigned long long addr);
	                                ^
input.hlsl:3:19: warning: default template arguments for a function template are a C++11 extension
	  template <class T = unsigned int>
	                  ^   ~~~~~~~~
input.hlsl:6:1: error: 'extern' is not a valid modifier for a function
	extern T RawBufferLoad(unsigned long long addr, unsigned int alignment);
	^
```
This PR just quits the `DeclPrinter::VisitNamespaceDecl` visitor when the namespace declaration's name is `"vk"` as this is a known built-in name for the SPIR-V backend and there's no need to print implicit namespaces in the rewriter.

### DIFF
--- a/tools/clang/lib/AST/DeclPrinter.cpp
+++ b/tools/clang/lib/AST/DeclPrinter.cpp
@@ -852,6 +852,13 @@ void DeclPrinter::VisitStaticAssertDecl(StaticAssertDecl *D) {
 // C++ declarations
 //----------------------------------------------------------------------------
 void DeclPrinter::VisitNamespaceDecl(NamespaceDecl *D) {
+  // HLSL Change Begin - Don't emit built-in "vk" namespace, it's implicitly
+  // declared when compiling to SPIR-V and would otherwise cause parsing errors
+  // due to unsupported HLSL 2021 features.
+  if (D->getNameAsString() == "vk")
+    return;
+  // HLSL Change End
+
   if (D->isInline())
     Out << "inline ";
   Out << "namespace " << *D << " {\n";
--- a/tools/clang/test/HLSL/rewriter/spirv.hlsl
+++ b/tools/clang/test/HLSL/rewriter/spirv.hlsl
@@ -1,0 +1,16 @@
+
+struct S
+{
+	int4 a;
+	int4 b;
+};
+
+RWStructuredBuffer<S> structuredUAV : register(u0);
+
+RWBuffer<int4> outBufferUAV : register(u1);
+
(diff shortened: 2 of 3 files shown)


tools/clang/lib/AST/DeclPrinter.cpp
```diff
@@ -852,6 +852,11 @@ void DeclPrinter::VisitStaticAssertDecl(StaticAssertDecl *D) {
 // C++ declarations
 //----------------------------------------------------------------------------
 void DeclPrinter::VisitNamespaceDecl(NamespaceDecl *D) {
+  // HLSL Change Begin - Don't emit built-in "vk" namesapce
```
cassiebeckley, line 855: nit: typo "namesapce" -> "namespace"

Would you also be able to add a bit more detail about why we're not emitting this namespace in the comments here?

cassiebeckley (changes requested):
Thanks for the fix! Would you be able to add a test for this?

tools/clang/lib/AST/DeclPrinter.cpp
LukasBanana: I added a SPIR-V rewriter test to this PR, but I have difficulties actually running this test. Can you point me to the docs where the unit test procedure is explained? I can't find a `clangTest` project in my DXC VS solution.

The reason this namespace should not be emitted is because it's an internal declaration and DXC doesn't actually support parsing this declaration without additional arguments such as explicitly enabling template support and other HLSL 2021 features. It is also not necessary to print this declaration when rewriting HLSL source, because it is implicitly declared when we compile with the `-spirv` argument.

cassiebeckley: I think it's `hcttest` in the HLSL console according to https://github.com/microsoft/DirectXShaderCompiler#running-tests, but I don't know for sure since I generally build and test on Linux which has a slightly different setup and just uses CMake's `ctest` command.

Also, could you put the explanation you wrote here in the comments in after line 855? Thanks!

cassiebeckley (approved):
LGTM once tests pass

@chrisbieneman will this interact with the change to tests you're working on? I'd also like your input since this is outside of the SPIR-V directories.

PR #4804, opened 2022-11-19
Cleanup Windows code to conform to stricter C++ 17 compiler

llvm-beanz:
MSVC's C++ language implementation deviates from the standard in a number of ways, this patch cleans up several issues that become errors if you build DXC using clang-cl and C++ 17, which has a more strict interpretation of C++.

This change should not have any functional impact.

Specific changes include:
* static_assert(false...) is removed, this is always ill-formed by the C++ standard.
* Under template instantiations implicit `this` is not always allowed.
* Character array `<` comparision is... probably not what was intended.
* Implicit conversions are stricter than MSVC enforces.
* Template base types need explicit type specialization when referred to.
* It is illegal to initialize more than one member of a union.
* Scope nested enums cannot be forward declared.
* Header paths should be case-sensitive.
* Exception specifications of specific types are disallowed in C++ 17.
* Clang is stricter about const-correctness, especially as it relates to constant c strings.

### DIFF
--- a/include/dxc/DxilContainer/DxilRuntimeReflection.h
+++ b/include/dxc/DxilContainer/DxilRuntimeReflection.h
@@ -238,20 +238,12 @@ template<typename _T>
 class RecordTraits {
 public:
   static constexpr const char *TypeName() {
-#ifdef _WIN32
-    static_assert(false, "");
-#else
     assert(false);
-#endif
     return nullptr;
   }
 
   static constexpr RuntimeDataPartType PartType() {
-#ifdef _WIN32
-    static_assert(false, "");
-#else
     assert(false);
-#endif
     return RuntimeDataPartType::Invalid;
   }
 
@@ -259,11 +251,7 @@ public:
   // RDAT_STRUCT is being used in ref type, which requires the struct to have
   // a table and be defined with RDAT_STRUCT_TABLE instead.
   static constexpr RecordTableIndex TableIndex() {
-#ifdef _WIN32
-    static_assert(false, "");
-#else
     assert(false);
-#endif
     return (RecordTableIndex)-1;
   }
   // RecordSize() is defined in order to allow for use of forward decl type in RecordRef
--- a/include/dxc/DxilContainer/DxilRuntimeReflection.h
+++ b/include/dxc/DxilContainer/DxilRuntimeReflection.h
@@ -237,15 +237,9 @@ public:
 template<typename _T>
 class RecordTraits {
 public:
-  static constexpr const char *TypeName() {
-    assert(false);
-    return nullptr;
-  }
+  static constexpr const char *TypeName();
 
-  static constexpr RuntimeDataPartType PartType() {
-    assert(false);
-    return RuntimeDataPartType::Invalid;
-  }
+  static constexpr RuntimeDataPartType PartType();
 
   // If the following static assert is hit, it means a structure defined with
   // RDAT_STRUCT is being used in ref type, which requires the struct to have
--- a/include/dxc/DxilContainer/DxilRuntimeReflection.h
+++ b/include/dxc/DxilContainer/DxilRuntimeReflection.h
@@ -244,14 +244,12 @@ public:
   // If the following static assert is hit, it means a structure defined with
   // RDAT_STRUCT is being used in ref type, which requires the struct to have
   // a table and be defined with RDAT_STRUCT_TABLE instead.
-  static constexpr RecordTableIndex TableIndex() {
-    assert(false);
-    return (RecordTableIndex)-1;
-  }
+  static constexpr RecordTableIndex TableIndex();
+  
   // RecordSize() is defined in order to allow for use of forward decl type in RecordRef
-  static constexpr size_t RecordSize() { /*static_assert(false, "");*/ return sizeof(_T); }
+  static constexpr size_t RecordSize();
   static constexpr size_t MaxRecordSize() { return RecordTraits<_T>::DerivedRecordSize(); }
-  static constexpr size_t DerivedRecordSize() { return sizeof(_T); }
+  static constexpr size_t DerivedRecordSize();
 };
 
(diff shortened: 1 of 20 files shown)


lib/DxilDia/DxcPixTypes.cpp
```diff
@@ -275,7 +275,7 @@ STDMETHODIMP dxil_debug_info::DxcPixStructType::GetFieldByName(
     _In_ LPCWSTR lpName,
     _Outptr_result_z_ IDxcPixStructField **ppField)
 {
-  std::string name = CW2A(lpName);
+  std::string name = CW2A(lpName).m_psz;
```
bogner, line 278: For these string-related ones, I'm pretty sure the conversion will be allowed if you do `std::string name(CW2A(lpName));`, and that's better code anyway.

projects/dxilconv/lib/DxbcConverter/DxbcConverterImpl.h
```diff
@@ -407,12 +407,9 @@ class DxbcConverter : public IDxbcConverter {
...
     vector<pair<unsigned, BasicBlock*> > SwitchCases;  // Switch
 
-    Scope() : Kind(Kind::Function), pPreScopeBB(nullptr), pPostScopeBB(nullptr), NameIndex(0), 
-              pThenBB(nullptr), pElseBB(nullptr), pCond(nullptr),
-              pLoopBB(nullptr), ContinueIndex(0), LoopBreakIndex(0),
-              pDefaultBB(nullptr), pSelector(nullptr), CaseGroupIndex(0), SwitchBreakIndex(0),
-              LabelIdx(0), CallIdx(0), ReturnTokenOffset(0), ReturnIndex(0), bEntryFunc(false),
-              pHullLoopBB(nullptr), HullLoopBreakIndex(0), pInductionVar(nullptr), HullLoopTripCount(0) {}
+    Scope() : Kind(Kind::Function), pPreScopeBB(nullptr), pPostScopeBB(nullptr), NameIndex(0) {
+      memset(reinterpret_cast<char*>(&pThenBB), '\0', reinterpret_cast<char*>(&SwitchCases) - reinterpret_cast<char*>(&pThenBB));
```
bogner, line 411: This seems pretty fragile, but I guess the other options (while keeping this completely NFC, anyway) aren't really better

bogner (approved):
Looks good other than my comment about string conversions

hekota (approved):
LGTM!

include/dxc/DxilContainer/DxilRuntimeReflection.h
```diff
@@ -238,32 +238,20 @@ template<typename _T>
 class RecordTraits {
 public:
   static constexpr const char *TypeName() {
-#ifdef _WIN32
-    static_assert(false, "");
```
adam-yang, line 242: Do we have a way to do this at compile time?

llvm-beanz: This pattern doesn't work at compile time because a non-dependent static assert is defined to be evaluated at parse, not instantiation. MSVC incorrectly implements the C++ standard.

Looking at what this is doing, you can convert this to a linker error by deleting the method body entirely, which is probably better than what I did here.

projects/dxilconv/lib/DxbcConverter/DxbcConverterImpl.h
llvm-beanz: Yea... The thing that makes this really gross to me is that since the union and contained structs are anonymous you can't just take the sizeof... instead you have to do member offset shenanigans.

I could not think of a better way to accomplish this without just rewriting the code.

include/dxc/DxilContainer/DxilRuntimeReflection.h
llvm-beanz: Update pushed.

include/dxc/DxilContainer/DxilRuntimeReflection.h
```diff
@@ -237,33 +237,15 @@ class RawBytesReader {
...
 
   // If the following static assert is hit, it means a structure defined with
   // RDAT_STRUCT is being used in ref type, which requires the struct to have
   // a table and be defined with RDAT_STRUCT_TABLE instead.
   static constexpr RecordTableIndex TableIndex() {
-#ifdef _WIN32
-    static_assert(false, "");
-#else
     assert(false);
-#endif
```
pow2clk, line 266: Why shouldn't this get the same body deletion treatment as the two above?

lib/DxcSupport/dxcapi.use.cpp
```diff
@@ -33,7 +33,7 @@ static std::string GetWin32ErrorMessage(DWORD err) {
   DWORD formattedMsgLen =
       FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
                     nullptr, err, 0, formattedMsg, _countof(formattedMsg), 0);
-  if (formattedMsg > 0 && formattedMsgLen < _countof(formattedMsg)) {
+  if (formattedMsgLen > 0 && formattedMsgLen < _countof(formattedMsg)) {
```
pow2clk, line 36: 😣

pow2clk (approved):
Looks fine. Just wondering about the last static_assert

include/dxc/DxilContainer/DxilRuntimeReflection.h
adam-yang: ^

adam-yang (approved):

include/dxc/DxilContainer/DxilRuntimeReflection.h
llvm-beanz: Doh! I missed that one, and another two below. Updated now.

PR #4806, opened 2022-11-21
[SPIR-V] get field index from SPIR-V type, not AST

Keenuts:
Before this commit, struct field indices were computed from the AST type. This was correct as SPIR-V types and AST types had the same structure.

When adding bitfields, we started to squash some fields together, as long as some rules were respected (bitfield, same type, doesn't span over multiple type-sized words, etc).
This means indices now diverge between SPIR-V types and AST types.

This requires us to lower the AST type before the lowering pass, when generating instructions like OpAccessChain.
A cleaner alternative would be to lower all types before instruction generation, and only operate on SPIR-V type. But cost of making this refactoring is large, and we don't believe it is worth it since we plan to upstream to clang (hence rewrite this code).

note: this code prevents AST indices to diverge from SPIR-V indices when computed. This is just a safeguard until bitfields are in place in case I made a mistake with the layout rules, allowing us to catch those bugs faster.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -15,6 +15,7 @@
 
 #include "AlignmentSizeCalculator.h"
 #include "InitListHandler.h"
+#include "LowerTypeVisitor.h"
 #include "RawBufferMethods.h"
 #include "dxc/DXIL/DxilConstants.h"
 #include "dxc/HlslIntrinsicOp.h"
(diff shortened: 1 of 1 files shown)


Keenuts:
found a bug when implementing bitfields. Removing reviewers

Keenuts:
The bug was related to inheritance, when the base class was empty. Because I only used the offset value to determine field merging, I was returning the wrong index for such cases.
It will be better to check for bitfield presence, and disallow any kind of field-merge is not a bitfield. Should be more robust. Will add that with the bitfield PRs

cassiebeckley (approved):
Looks great! Thanks for figuring out a way to do this without it being too messy.

PR #4808, opened 2022-11-21
Workarounds for arm64ec build

llvm-beanz:
I encountered two issues building arm64ec. First CMake and/or link.exe seem to have a disagreement about flag ordering and how to handle `/INCREMENTAL`. This change results in disabling kernel-mode debugging on arm64ec builds, but it means we can build... so that's a win.

The other issue is that libClangSema has some files that have gotten too big, so we need `/bigobj` for Sema.


hekota (approved):
LGTM!

PR #4809, opened 2022-11-21
Make clang taef lit config conform to conventions

llvm-beanz:
The new taef lit configuration names don't conform to the normal conventions which makes it difficult for them to have automatically generated targets. This updates them to match LLVM & Clang conventions.

This changes the directory name from `clang-hlsl` to `taef` avoiding future conflicts with `HLSL` test directories, and changes the target from `check-hlsl` to `check-clang-taef` consistent with other clang-based testing directories.


tools/clang/test/CMakeLists.txt
```diff
@@ -20,8 +20,8 @@ configure_lit_site_cfg(
   )
 
 configure_lit_site_cfg(
-  ${CMAKE_CURRENT_SOURCE_DIR}/clang-hlsl/lit.site.cfg.in
-  ${CMAKE_CURRENT_BINARY_DIR}/clang-hlsl/lit.site.cfg
+  ${CMAKE_CURRENT_SOURCE_DIR}/taef/lit.site.cfg.in
+  ${CMAKE_CURRENT_BINARY_DIR}/taef/lit.site.cfg
   )
```
hekota, line 25: I think it might would be slightly better to name this directory clang_taef instead of just taef.

hekota (approved):

tools/clang/test/CMakeLists.txt
llvm-beanz: The convention with LLVM project testing targets is that they are named `check-${project}-${subdirectory}`, where the subdirectory is the directory containing the lit suite with slashes replaced with dashes, and all lowercase. For example: check-llvm-codegen-directx runs the tests living under llvm/tests/CodeGen/DirectX, or check-clang-ast, runs the tests from clang/tests/AST.

Putting clang in the directory name (as the existing code does) makes a redundant target name `check-clang-clang-hlsl` or `check-clang-clang_taef` both are a bit more of a mouthful.

Part of why I'm making this change is so that when we merge #4801 and start using lit's shell tests, we'll have consistent target names for running subsets of our tests.

hekota: Sounds good!

PR #3793, opened 2021-05-22
[Linux] WinAdapter: Remove virtual dtors from IUnknown to fix vtable ABI

MarijnS95:
Fixes #3783

The vtable for `IUnknown` and its subclasses contain two deletion pointers when compiled on non-Windows systems with `IUnknown` from `WinAdapter.h`:

```cpp
vtable for 'DxcLibrary' @ 0x7ffff7cbc5f8 (subobject @ 0x5555556bb9e0):
[0]: 0x7ffff6a56d40 <DxcLibrary::QueryInterface(_GUID const&, void**)>
[1]: 0x7ffff6a56d20 <DxcLibrary::AddRef()>
[2]: 0x7ffff6a56d30 <DxcLibrary::Release()>
[3]: 0x7ffff6b36bc0 <IUnknown::~IUnknown()> // Complete object destructor
[4]: 0x7ffff6a57130 <DxcLibrary::~DxcLibrary()> // Deleting destructor
[5]: 0x7ffff6a56d50 <DxcLibrary::SetMalloc(IMalloc*)>
[6]: 0x7ffff6a56d60 <DxcLibrary::CreateBlobFromBlob(IDxcBlob*, unsigned int, unsigned int, IDxcBlob**)>
... More DxcLibrary virtual functions
```

This shifts the the pointers for functions for all subclasses, and is [annoying] to deal with in otherwise cross-platform applications using DirectXShaderCompiler as library.  `dxcompiler.dll` compiled on/for
Windows without `WinAdapter.h` does not suffer this problem, and only has three function pointers for `IUnknown`.

Fortunately it is easily solved by removing the virtual destructor from `IUnknown`.  LLVM enables `-Wnon-virtual-dtor` that warns against classes with virtual methods but no virtual destructor, though this
warning is best not enabled akin to Windows builds where `IUnknown` from `windows.h` (`unknwn.h`) results in the same warning on MSVC ([1]/[2]).

[annoying]: https://github.com/Traverse-Research/hassle-rs/blob/1e624792fc3a252ac7788e3c1c5feda52887272f/src/unknown.rs
[1]: https://github.com/microsoft/DirectXShaderCompiler/issues/3783#issuecomment-844189358
[2]: https://godbolt.org/z/hKPT6ThEf

---

Tested on Clang `11.1.0` and GCC `11.1.0`. Unfortunately GCC fails compiling on an irrelevant `-Werror=stringop-truncation` in `SPIRV-Tools`, irrespective of this PR.

[...]

PR #4810, opened 2022-11-21
[linux] Enable Reflection on *nix platforms

### LINKED ISSUE #4358: Support IDxcContainerReflection on Linux
IDxcContainerReflection is currently only implemented on the Windows build, since the headers `d3d11shader.h` and `d3d12shader.h` are included. These headers could be added to this repo for Linux support. (We have already done this for `d3d12shader.h` in Rust.)

The use case for this feature is that I would like to run reflection during offline shader compilation on a Linux server.

I am willing to open a PR for this, but would like to know if this is something that the maintainers would like to support or not.

https://github.com/microsoft/DirectXShaderCompiler/blob/2dc067b561f17d09d8012a1ded05bf0f6253fea5/lib/HLSL/DxilContainerReflection.cpp#L39

pow2clk:
This adds the DirectX-Headers repo as a submodule to grant the shader reflection structs
that are needed to enable the functionality on non-windows platforms.  No real
changes related to the reflection code had to be made so much as the newly enabled
code had to be made to work on the platform compilers, namely clang. These changes
come first with one separated out for a particularly gnarly issue with a circular dependency
involving the RDAT dumping code.

Thereafter, the submodule is added, the reflection code is enabled for functionality and
tests and a small incidental improvement to error reporting when the data dir isn't found
was made because it's driven me crazy one too many times.

Fixes #4358


lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -2389,10 +2360,13 @@ HRESULT DxilModuleReflection::_GetResourceBindingDesc(UINT ResourceIndex,
   _Out_ D3D12_SHADER_INPUT_BIND_DESC *pDesc, PublicAPI api) {
   IFRBOOL(pDesc != nullptr, E_INVALIDARG);
   IFRBOOL(ResourceIndex < m_Resources.size(), E_INVALIDARG);
+#ifdef _WIN32
```
llvm-beanz, line 2363: splitting a conditional like this is pretty gross... Is there no way to make the `if` statement valid on Linux?

If not, you could make the if return after the memcpy, which would eliminate the need for an else block.

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -2472,10 +2446,13 @@ HRESULT DxilModuleReflection::_GetResourceBindingDescByName(LPCSTR Name,
 
   for (UINT i = 0; i < m_Resources.size(); i++) {
     if (strcmp(m_Resources[i].Name, Name) == 0) {
+#ifdef _WIN32
       if (api != PublicAPI::D3D12) {
         memcpy(pDesc, &m_Resources[i], sizeof(D3D11_SHADER_INPUT_BIND_DESC));
       }
-      else {
+      else
+#endif // _WIN32
```
llvm-beanz, line 2454: Same here :(

llvm-beanz (commented):
The vast majority of this looks very straightforward to me.

external/CMakeLists.txt
```diff
@@ -7,6 +7,25 @@ if (NOT HLSL_ENABLE_DEBUG_ITERATORS)
   add_definitions(/D_ITERATOR_DEBUG_LEVEL=0)
 endif (NOT HLSL_ENABLE_DEBUG_ITERATORS)
 
+# Need DirectX-Headers module if not on windows
+if (NOT WIN32)
+  set(DXC_DIRECTX_HEADERS_DIR "${DXC_EXTERNAL_ROOT_DIR}/DirectX-Headers"
+      CACHE STRING "Location of DirectX-Headers source")
+  if (NOT DEFINED DirectX-Headers_SOURCE_DIR)
+    if (IS_DIRECTORY ${DXC_DIRECTX_HEADERS_DIR})
+      add_subdirectory(${DXC_DIRECTX_HEADERS_DIR}
```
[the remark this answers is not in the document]
pow2clk: Currently we do not. However, I greatly prefer the approach that DirectX-Headers takes to defining dx guids as it allows the same guid structure to have the same address across multiple files which ours does not.

I chose not to take that step with this change, but I want to take it going forward when I try to better consolidate the efforts to make non-Windows platforms look a bit more like Windows for DirectX purposes, which is a goal both projects share.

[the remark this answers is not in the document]
llvm-beanz: Their GUID handling is all header defined too, right? The source for it just sets a single preprocessor define then includes headers.

If we want to migrate to use it we could define that ourselves, or we depend on their static archive target. The concern I have about their static archive target is that it means filtering target dependencies through the other CMake targets. If we're only doing that on Linux it feels like we might be introducing some extra complexity. Would we shift to using the submodule headers on Windows too?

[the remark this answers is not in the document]
pow2clk: I mean to shift to using submodule headers on Windows as it would garner us certain advantages, not the least of which being easy access to more recent headers.

You make a good point that we could just as easily recreate dxguid.cpp. Regardless of where we land on Windows using these headers, that's a sufficient argument to simplify this.

llvm-beanz (approved):
I think there is a bigger issue with our find_package usage, but this change all looks good to me. We can sort out the find_package issues later.

include/dxc/Support/D3DReflection.h
```diff
@@ -0,0 +1,25 @@
...
+// Provides the needed headers and defines for D3D reflection.               //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#ifndef _WIN32
+// need to disable this as it is voilated by this header
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wnon-virtual-dtor"
```
MarijnS95, line 17: @pow2clk Didn't we disable this warning altogether as discussed in https://github.com/microsoft/DirectXShaderCompiler/pull/3793#pullrequestreview-861757581, or is it now enabled by default on newer compilers (since I only commented out, instead of leaving an unconditional `-Wnon-virtual-dtor`)?

include/dxc/Support/D3DReflection.h
```diff
@@ -0,0 +1,25 @@
...
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Provides the needed headers and defines for D3D reflection.               //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#ifndef _WIN32
+// need to disable this as it is voilated by this header
```
MarijnS95, line 15:
```suggestion
// need to disable this as it is violated by this header
```

MarijnS95:
Thanks for this! Tested on Linux and works a treat, at least for reflecting compute group sizes which we previously couldn't to on our Linux build system!

PR #4812, opened 2022-11-22
Update GitHub workflow to (hopefully) work

llvm-beanz:
There were a few things missing in the initial attempt at this workflow. I left off the file extension for the CMake cache file, and also forgot to initialize the submodules.

Additionally using make is... slow. Using make `-j` causes the VM to run out of memory. To resolve those issues I'm installing ninja and using ninja instead. I'm also setting `LLVM_USE_LINKER=lld` to reduce linker memory usage and `LLVM_PARALLEL_LINK_JOBS=1` to only run one link at a time.

I've also added `--rerun-failed` and `--output-on-failure` to the ctest invocation so that if it fails we get some hopefully helpful output.

### DIFF
--- a/.github/workflows/coverage-gh-pages.yml
+++ b/.github/workflows/coverage-gh-pages.yml
@@ -27,16 +27,20 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          submodules: true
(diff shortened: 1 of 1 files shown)


pow2clk (approved):
It all looks like goodness to me

PR #4813, opened 2022-11-22
Enable running tests under LIT

llvm-beanz:
If `CLANG_INCLUDE_TESTS=On`, the googletest unit tests are built as part of the clang unit test suite. This allows them to run via lit using the `check-clang-unit` target.

This change requires renaming the test binaries to match LLVM conventions. This binary renaming only occurs if `CLANG_INCLUDE_TESTS=On`, so existing workflows are unimpacted. To keep things consistent all the test targets are renamed to match LLVM conventions regardless of whether or not the option is enabled.

With this workflow enabled `check-clang-unit` runs all the HLSL & SPIR-V googletest tests _and_ a handful of Clang unit tests that I had previously fixed. Additionally due to lit's more aggressive parallelization, the tests run almost 2x as fast despite running more tests. Win-win-win.

`check-clang-unit` works on both Linux and Windows, and `check-clang-taef` works on Windows as well.


hekota (approved):
LGTM!

tools/clang/test/taef/lit.cfg
```diff
@@ -28,7 +28,7 @@ if llvm_obj_root is None:
 
 te = os.path.join(config.llvm_obj_root, "TAEF", config.arch, 'te.exe')
 
-test_dll = os.path.join(config.llvm_obj_root, config.build_mode, 'bin', 'clang-hlsl-tests.dll')
+test_dll = os.path.join(config.llvm_obj_root, config.build_mode, 'bin', 'ClangHLSLTests.dll')
```
pow2clk, line 31: Out of belated curiosity, why the rename?

It looks like the previous DLL is still referenced by HLSLHost.cpp and a few powershell and cmd files. Less important, but possibly worth it for consistency, the gtest executable equivalent is named `clang-hlsl-tests` to keep with the previous naming convention.

PR #4815, opened 2022-11-23
Update macos/linux builds to *-latest

pow2clk:
This updates the ADO yaml to select the latest version of the OS. In the case of MacOS, this will still be 11 for now. For Ubuntu, it will silence the warnings that we are using a deprecated version and save us the headache of having to do this again.

The risk is that a new version of either OS will cause the build to break without our intervention, but the real motivation for explicit versions was an earlier impulse to support the minimal build system that is no longer a priority. As for the risk, I feel it is minimal and we can respond in a timely fashion if it should occur.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -43,8 +43,8 @@ stages:
     timeoutInMinutes: 90
 
     variables:
-      macOS: macOS-11
-      linux: Ubuntu-18.04
+      macOS: macOS-latest
+      linux: Ubuntu-latest
 
     strategy:
       matrix:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -64,14 +64,14 @@ stages:
         Linux_Gcc_Release:
           image: ${{ variables.linux }}
           configuration: Release
-          CC: gcc-7
-          CXX: g++-7
+          CC: gcc-9
+          CXX: g++-9
           CXX_FLAGS:
         Linux_Gcc_Debug:
           image: ${{ variables.linux }}
           configuration: Debug
-          CC: gcc-7
-          CXX: g++-7
+          CC: gcc-9
+          CXX: g++-9
           CXX_FLAGS:
         MacOS_Clang_Release:
           image: ${{ variables.macOS }}


pow2clk:
Feel free to talk me out of this. At the very least, we need to bump the ubuntu version as the current version turns into pumpkins in April

pow2clk:
A potentially useful reference: https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml

azure-pipelines.yml
```diff
@@ -64,14 +64,14 @@ stages:
         Linux_Gcc_Release:
           image: ${{ variables.linux }}
           configuration: Release
-          CC: gcc-7
-          CXX: g++-7
+          CC: gcc-9
```
Keenuts, line 67: a reason not to use gcc-12? (latest seems to be jammy, which seems to have gcc-12)

pow2clk:
appveyor timed out ☹️

@dnovillo, I'd like to make sure this is in keeping with your team's current priorities for these builds.

dnovillo:
Timed out after 90mins (the limit Appveyor imposes on our account).  A
subsequent PR succeeded, so I'm hoping it was a fluke.  I've just restarted
the job.

Thanks for the heads up.

pow2clk:
@dnovillo I meant more the direction of this change. It is somewhat related to @Keenuts comment about what version of GCC. Earlier there was an effort to support as early versions as we could. I have the impression that's no longer such a priority. I wanted to know how you prioritize that?

azure-pipelines.yml
pow2clk: gcc 12 is not natively available on this image. GCC 10 is and I could easily be convinced to switch to that. 

In both compiler and OS version, I'm trying to balance the eagerness to benefit from the advantages of the latest versions with the interest of supporting people who might not have the latest versions.

The highest available Ubuntu right now is this: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md which has GCC 11.  However, to get it, we'd have to specify it by number and lose the auto updating advantage of just specifying -latest.

Keenuts:
> Earlier there was an effort to support as early versions as we could. I have the impression that's no longer such a priority.

Since the end goal is to upstream into LLVM, maybe we should align with there policy?
From this, it seems that they want a 3 years old version:
https://discourse.llvm.org/t/rfc-increasing-the-gcc-and-clang-requirements-to-support-c-17-in-llvm/59983?page=2

This would mean:
 - clang 9.0.1 (almost) (https://releases.llvm.org/)
 - gcc... 9.2? (release order is weird) (https://gcc.gnu.org/releases.html)
 - MSVC to 2019

PR #4817, opened 2022-11-23
PIX: Fix instrumentation of libraries; instruction numbering reports

jeffnn:
This PR does a few things to fix PIX's operation on lib targets:
-Remove the check for "use_empty" on a function. Lib entry functions have no uses! This was done by centralizing the discovery of interesting (to PIX) functions in a new helper in PixPassHelpers.cpp
-In DxilDebugInstrumentation.cpp, remove the early "return" inside runOnModule. There are more functions left to instrument!
-Add reporting from the virtual register pass that reveals the start and end instruction ordinal for interesting functions. PIX can use these values to disambiguate different functions within a module for e.g. shader debugging output.


adam-yang (approved):
LGTM!

PR #4821, opened 2022-11-28
Don't cancel in-progress coverage builds

llvm-beanz:
The coverage builds take a _long_ time to run on the standard GitHub action bots. I initially thought cancelling and restarting was a good idea and have since reconsidered...


hekota (approved):
LGTM!

PR #4823, opened 2022-11-29
Fix ARM build break

hekota:


llvm-beanz (approved):
Sorry for breaking this! Thank you for fixing it 😅.

Looks good to me. Can you please update the description before merging to include some context for the change?

PR #4829, opened 2022-11-29
Fix or disable last LLVM tests on Windows

llvm-beanz:
This handles a few last Windows fixes for LLVM tests.

* The YAML praser convertes \r to \n, resulting in repeated newlines in YAML tests on Windows.
* The MS filesystem changes on Windows cause one of the MemoryBuffer tests to fail. I've disabled that test on Windows.
* Because we convert many crash cases to exceptions on Windows the GoogleTest DEATH tests fail to die on Windows.

This change addresses the same issue #4794 solves, but restricts the change to Windows only. I'm only merging this since @python3kgae is out and I don't really want to wait until he returns.

utils/unittest/googletest/include/gtest/internal/gtest-port.h

### DIFF
--- a/test/YAMLParser/spec-09-02.test
+++ b/test/YAMLParser/spec-09-02.test
@@ -11,4 +11,4 @@
 # FIXME: The string below should actually be
 #   "as space trimmed\nspecific\nescaped\tnone", but the parser currently has
 #   a bug when parsing multiline quoted strings.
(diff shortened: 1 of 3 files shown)


pow2clk (approved):
Might be more accurately described as "Fix or disable last LLVM tests on Windows" 😉

PR #4830, opened 2022-11-30
Fixed lifetime marker causing unroll to fail

adam-yang:
Fixed a bug where enabling lifetime markers causes unroll to fail. Running instcombine before loop unroll may cause unroll to fail, because instcombine may rewrite expressions in a way that make ScalarEvolutions unable to evaluate. This change replaces the instcombine with simplify-inst, which serves the same purpose as the instcombine, but leaves the expression easy for ScalarEvolutions to evaluate.

Also moved the original test for instcombine to 

The problematic expression rewrite was as follows:
```
limit = (l % 2) ? 1 : 2;
```
Instcombine rewrites it to
```
limit = 2 - (l & 1);
```
Which stumped ScalarEvolutions.


tex3d (approved):
LGTM!

PR #4831, opened 2022-11-30
[SPIR-V] HLSL2021: initial bitfield implementation

Keenuts:
This commit chain adds initial support for bitfield implementation.
As-is, bitfields can be generated and used. What's missing is proper support on SPIR-V tools. Because bitfield operations are not recognize, constant propagation and all other optimization cannot work with it. This means we would load the same bitfield multiple times, and so on. But hey, it works!

I believe I covered most cases, but I might miss obvious usages!

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -578,7 +578,7 @@ const StructType *lowerStructType(const SpirvCodeGenOptions &spirvOptions,
   const SpirvType *spvType = lowerTypeVisitor.lowerType(
       baseType, spirvOptions.sBufferLayoutRule, llvm::None, SourceLocation());
 
-  const StructType *output = dyn_cast<const StructType *>(spvType);
+  const StructType *output = dyn_cast<StructType>(spvType);
   assert(output != nullptr);
   return output;
 }
@@ -5846,7 +5846,7 @@ SpirvInstruction *SpirvEmitter::doMemberExpr(const MemberExpr *expr,
     return instr;
   }
 
-  const auto *fieldDecl = dyn_cast<FieldDecl*>(expr->getMemberDecl());
+  const auto *fieldDecl = dyn_cast<FieldDecl>(expr->getMemberDecl());
   if (!fieldDecl || !fieldDecl->isBitField()) {
     return derefOrCreatePointerToValue(base->getType(), instr, expr->getType(),
                                        indices, loc, range);
@@ -7721,7 +7721,7 @@ const Expr *SpirvEmitter::collectArrayStructIndices(
       const uint32_t fieldIndex = getFieldIndexInStruct(
           spirvStructType, astStructType,
           /* fieldDecl */
-          dyn_cast<const FieldDecl *>(indexing->getMemberDecl()));
+          dyn_cast<FieldDecl>(indexing->getMemberDecl()));
 
       if (rawIndex) {
         rawIndices->push_back(fieldIndex);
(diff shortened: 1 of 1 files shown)


tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -254,9 +269,32 @@ SpirvLoad *SpirvBuilder::createLoad(const SpirvType *resultType,
 SpirvStore *SpirvBuilder::createStore(SpirvInstruction *address,
                                       SpirvInstruction *value,
                                       SourceLocation loc, SourceRange range) {
-  assert(insertPoint && "null insert point");
```
cassiebeckley, line 257: Any particular reason for removing this?

cassiebeckley (approved):
LGTM

tools/clang/lib/SPIRV/SpirvBuilder.cpp
Keenuts: No, mistake from the change, restoring this!

Keenuts:
@llvm-beanz : The SPIR-V part might not be your cup of tea, but in case you see some cases that would need coverage in the tests?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -7682,8 +7713,15 @@ const Expr *SpirvEmitter::collectArrayStructIndices(
...
-          getFieldIndexInStruct(spirvOptions, lowerTypeVisitor, indexing);
+      const auto &astStructType =
+          /* structType */ indexing->getBase()->getType();
+      const StructType *spirvStructType =
+          lowerStructType(spirvOptions, lowerTypeVisitor, astStructType);
+      assert(spirvStructType != nullptr);
+      const uint32_t fieldIndex = getFieldIndexInStruct(
+          spirvStructType, astStructType,
+          /* fieldDecl */
+          dynamic_cast<const FieldDecl *>(indexing->getMemberDecl()));
```
llvm-beanz, line 7724: Should this be `dyn_cast`? I'm pretty sure we don't build with RTTI enabled.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -13536,7 +13575,9 @@ SpirvEmitter::loadDataFromRawAddress(SpirvInstruction *addressInUInt64,
       spv::Op::OpBitcast, bufferPtrType, addressInUInt64, loc);
   address->setStorageClass(spv::StorageClass::PhysicalStorageBuffer);
 
-  SpirvLoad *loadInst = spvBuilder.createLoad(bufferType, address, loc);
+  SpirvLoad *loadInst = dynamic_cast<SpirvLoad *>(
```
llvm-beanz, line 13578: Same here, I assume this should be `dyn_cast`

llvm-beanz (approved):
This looks good to me with two small comments inline.

On a stylistic note LLVM style conventions omit braces around single-statement conditions and loops. DXC hasn't historically conformed rigidly to LLVM conventions so this is really up to you, but I thought I'd mention it.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts: Oh thanks! Assumed dyn_cast was kind of legacy stuff, didn't knew the diff! Changing that.

Keenuts:
> This looks good to me with two small comments inline.
> 
> On a stylistic note LLVM style conventions omit braces around single-statement conditions and loops. DXC hasn't historically conformed rigidly to LLVM conventions so this is really up to you, but I thought I'd mention it.

Changing to LLVM style then

tools/clang/lib/SPIRV/SpirvEmitter.cpp
llvm-beanz: Ah, yea. LLVM's dyn_cast is a lightweight RTTI implementation. It is basically dynamic casting without typeid. I think the approach dyn_cast uses creates binaries that are around 10-15% smaller.

llvm-beanz:
Force pushes make the incremental reviewing hard, but it looks like there are still some `dynamic_cast` calls in the change. We should correct those.

Keenuts:
> Force pushes make the incremental reviewing hard, 

Ahh, didn't thought about that (tried to keep commit cleans), but make sense! Will add new commits from now on.

> but it looks like there are still some `dynamic_cast` calls in the change. We should correct those.
yes sorry! Should be done by now.

I'm still checking in the internal CI this doesn't add regressions, will let you know 😊

Keenuts:
All good! Our CI is green with this commit 😊

PR #4833, opened 2022-11-30
[NFC] Address some low hanging fruit UBSan failures

llvm-beanz:
Undefined Behavior Sanitizer causes hard failures on runtime conditions that exhibit undefined behavior as defined by C and C++. These issues may not cause immediately identifiable software errors, but the code behavior is dependent on the compiler and runtime implementation. As a result they may become difficult to diagnose or detect errors in the future if the compiler or runtime changes.

This change addresses occurrences of the following undefined behaviors.

* Performing arithmetic operations on null pointers.
* Calling memcpy with a source or destination pointer that is null.
* Calling memset with a destination pointer that is null.
* Assigning an enum a value which is not a valid enum for that enum.
* Reading from unaligned pointers is undefined behavior.

After this change, there are 23 remaining undefined behavior failures in the current DXC tests which are less trivial to resolve.

### DIFF
--- a/include/dxc/DxilContainer/DxilPipelineStateValidation.h
+++ b/include/dxc/DxilContainer/DxilPipelineStateValidation.h
@@ -668,7 +668,8 @@ DxilPipelineStateValidation::CheckedReaderWriter::IncrementPos(size_t size) {
 template <typename _T> inline bool
 DxilPipelineStateValidation::CheckedReaderWriter::Cast(_T **ppPtr, size_t size) {
   PSV_RETB(CheckBounds(size));
-  *ppPtr = reinterpret_cast<_T*>(Ptr + Offset);
+  if (Mode != RWMode::CalcSize)
+    *ppPtr = reinterpret_cast<_T*>(Ptr + Offset);
   return true;
 }
 template <typename _T>
--- a/include/llvm/ADT/BitVector.h
+++ b/include/llvm/ADT/BitVector.h
@@ -534,8 +534,10 @@ private:
     // HLSL Change Starts: Use overridable operator new
     // Bits = (BitWord *)std::realloc(Bits, Capacity * sizeof(BitWord));
     BitWord  *newBits = new BitWord[Capacity];
-    std::memcpy(newBits, Bits, NumBitWords(Size) * sizeof(BitWord));
-    delete[] Bits;
+    if (Bits != nullptr) {
+      std::memcpy(newBits, Bits, NumBitWords(Size) * sizeof(BitWord));
+      delete[] Bits;
+    }
     Bits = newBits;
     // HLSL Change Ends
 
--- a/lib/DxilCompression/miniz.c
+++ b/lib/DxilCompression/miniz.c
@@ -1102,7 +1102,9 @@ static mz_bool tdefl_compress_lz_codes(tdefl_compressor *d)
         if (flags & 1)
         {
             mz_uint s0, s1, n0, n1, sym, num_extra_bits;
-            mz_uint match_len = pLZ_codes[0], match_dist = *(const mz_uint16 *)(pLZ_codes + 1);
+            mz_uint16 read_val;
+            memcpy((void*)&read_val, pLZ_codes + 1, sizeof(mz_uint16));
+            mz_uint match_len = pLZ_codes[0], match_dist = read_val;
             pLZ_codes += 3;
 
             MZ_ASSERT(d->m_huff_code_sizes[0][s_tdefl_len_sym[match_len]]);
@@ -1147,7 +1149,7 @@ static mz_bool tdefl_compress_lz_codes(tdefl_compressor *d)
         if (pOutput_buf >= d->m_pOutput_buf_end)
             return MZ_FALSE;
 
-        *(mz_uint64 *)pOutput_buf = bit_buffer;
+        memcpy((void*)pOutput_buf, (void*)&bit_buffer, sizeof(bit_buffer));
         pOutput_buf += (bits_in >> 3);
         bit_buffer >>= (bits_in & ~7);
         bits_in &= 7;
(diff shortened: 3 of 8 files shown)


lib/DxilRootSignature/DxilRootSignatureSerializer.cpp
```diff
@@ -450,8 +451,8 @@ void DeserializeRootSignatureTemplate(_In_reads_bytes_(SrcDataSizeInBytes) const
   IFTBOOL(((const char*)pInRTS) + s <= pMaxPtr, E_FAIL);
   if (pRootSignature->NumParameters) {
     pRootSignature->pParameters = new T_ROOT_PARAMETER[pRootSignature->NumParameters];
+    memset((void *)pRootSignature->pParameters, 0, s);
```
Keenuts, line 454: nit: size parameter is calculated from another count (pRS). Today this matches the allocation size, but might be source of a bug later no?
What about: `sizeof(T_ROOT_PARAMETER)*p_pRootSignature->NumParameters` would make clear we don't copy more than we allocate? Or maybe static_assert that the template type matches DxilContainerRootParameter?

Keenuts (commented):
Thanks for the fixes. TIL that memcpy with a size=0 is undefined behavior..

llvm-beanz:
> Thanks for the fixes. TIL that memcpy with a size=0 is undefined behavior.

Only if the source or destination pointer is null. If source and destination are both valid it is defined.

include/dxc/DxilContainer/DxilPipelineStateValidation.h
```diff
@@ -668,7 +668,8 @@ DxilPipelineStateValidation::CheckedReaderWriter::IncrementPos(size_t size) {
 template <typename _T> inline bool
 DxilPipelineStateValidation::CheckedReaderWriter::Cast(_T **ppPtr, size_t size) {
   PSV_RETB(CheckBounds(size));
-  *ppPtr = reinterpret_cast<_T*>(Ptr + Offset);
+  if (Mode != RWMode::CalcSize)
+    *ppPtr = reinterpret_cast<_T*>(Ptr + Offset);
```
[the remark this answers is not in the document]
llvm-beanz: It is undefined behavior to apply arithmetic operations to a null pointer. In CalcSize mode Ptr is null.

include/llvm/ADT/BitVector.h
```diff
@@ -534,8 +534,10 @@ class BitVector {
     // HLSL Change Starts: Use overridable operator new
     // Bits = (BitWord *)std::realloc(Bits, Capacity * sizeof(BitWord));
     BitWord  *newBits = new BitWord[Capacity];
-    std::memcpy(newBits, Bits, NumBitWords(Size) * sizeof(BitWord));
-    delete[] Bits;
+    if (Bits != nullptr) {
+      std::memcpy(newBits, Bits, NumBitWords(Size) * sizeof(BitWord));
+      delete[] Bits;
+    }
     Bits = newBits;
```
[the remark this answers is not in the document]
llvm-beanz: The allocation is defined, just not initialized. This is the reallocation for growing the bit vector storage. I think only coping the set bits if bits were set is correct.

lib/DxilCompression/miniz.c
```diff
@@ -1102,7 +1102,9 @@ static mz_bool tdefl_compress_lz_codes(tdefl_compressor *d)
         if (flags & 1)
         {
             mz_uint s0, s1, n0, n1, sym, num_extra_bits;
-            mz_uint match_len = pLZ_codes[0], match_dist = *(const mz_uint16 *)(pLZ_codes + 1);
+            mz_uint16 read_val;
+            memcpy((void*)&read_val, pLZ_codes + 1, sizeof(mz_uint16));
+            mz_uint match_len = pLZ_codes[0], match_dist = read_val;
```
tex3d, line 1107: Unsure why this change is necessary.  Is the issue that `pLZ_codes + 1` (plus 2 bytes read from there) is potentially out of buffer bounds at this point (according to UBSan's knowledge)?  Does the memcpy simply obfuscate things so UBSan doesn't complain?

lib/DxilCompression/miniz.c
```diff
@@ -1147,7 +1149,7 @@ static mz_bool tdefl_compress_lz_codes(tdefl_compressor *d)
         if (pOutput_buf >= d->m_pOutput_buf_end)
             return MZ_FALSE;
 
-        *(mz_uint64 *)pOutput_buf = bit_buffer;
+        memcpy((void*)pOutput_buf, (void*)&bit_buffer, sizeof(bit_buffer));
```
tex3d, line 1152: Similar question here to above.  Is memcpy used to obfuscate so UBSan doesn't complain?  Is there another way?

lib/DxilCompression/miniz.c
llvm-beanz: Not potentially... UBSan only complains on actual occurred runtime errors.

By definition memcpy will perform byte-by-byte copy, it _may_ optimize for aligned data reads but is conservative.

lib/DxilCompression/miniz.c
llvm-beanz: Same answer as above. This doesn't obfuscate, it fixes the unaligned data reads.

tex3d (approved):
LGTM!

include/dxc/DxilContainer/DxilPipelineStateValidation.h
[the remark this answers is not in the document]
pow2clk: That makes sense then. I guess the check wasn't for `if(Ptr != nullptr)` so that if there is ever another mode where it can be nullptr, this will catch it and we'll decide if that's intentional?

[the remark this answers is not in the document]
llvm-beanz: Yea. The pointer should be valid when reading and writing, if that changes in the future we can identify and address. Hopefully by then we'll be able to run ubsan as part of our normal testing flow.

PR #4837, opened 2022-12-01
Added dumpToFile debug functions

adam-yang:


lib/IR/AsmWriter.cpp
```diff
@@ -3366,6 +3366,30 @@ void MDNode::printAsBody(raw_ostream &OS, ModuleSlotTracker &MST, const Module *
...
+  os << *obj;
+  os.flush();
+
+  size_t bytesWritten = 0;
+  FILE *f = fopen(path, "wb");
+  if (f) {
+    bytesWritten = fwrite(s.c_str(), 1, s.size(), f);
+    fclose(f);
+  }
+  fprintf(stderr, "%u bytes written to %s\n", (unsigned)bytesWritten, path);
```
tex3d, line 3383: Does this circumvent the per-thread file system, or do we override the fprintf implementation?  I don't think we want to circumvent the file system stuff we have in place.

tex3d: Oh yeah, and same question for above `fopen` etc...

adam-yang: This is just for calling in the debugger, so it's circumventing the per thread file system deliberately.

adam-yang: That's a good idea. Done.

lib/IR/AsmWriter.cpp
```diff
@@ -3366,6 +3366,32 @@ void MDNode::printAsBody(raw_ostream &OS, ModuleSlotTracker &MST, const Module *
     TypePrinter.incorporateTypes(*M);
   WriteMDNodeBodyInternal(OS, this, &TypePrinter, MST.getMachine(), M);
 }
+
+#if defined(NDEBUG) || defined(LLVM_ENABLE_DUMP)
```
llvm-beanz, line 3370: Sorry, I put this wrong in my comment, it should be `#if !defined(NDEBUG) || defined(LLVM_ENABLE_DUMP)`

include/llvm/IR/Value.h
```diff
@@ -199,6 +199,7 @@ class Value {
 
   /// \brief Support for debugging, callable in GDB: V->dump()
   void dump() const;
+  void dumpToFile(const char *path) const; // HLSL Change
```
llvm-beanz, line 202: We should also wrap the declaration in `#if !defined(NDEBUG) || defined(LLVM_ENABLE_DUMP)`

damyanp:

damyanp:

PR #4838, opened 2022-12-01
Fix ternary operator for enum result type

### LINKED ISSUE #4826: ternary `?` degrades enumerator types to `int`
```
enum Composite
{
    Spec,
    Diff,
};
static const Composite g_c = true ? Spec : Diff; 
```

result

> error: cannot initialize a variable of type 'const Composite' with an rvalue of type 'int'

excepted: successful build. (verified in clang, gcc and msvc on C++)

tex3d:
Note: existing path that limits operator to basic numeric types still allowed enum due to it being AR_TOBJ_BASIC, rather than using a new ArTypeObjectKind. It's questionable whether this was intentional, but changing it would require another special case to allow enums, like exists for object types, and it might have other side effects as well.

Fixes #4826

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -9855,8 +9855,15 @@ clang::QualType HLSLExternalSource::CheckVectorConditional(
     return QualType();
   }
 
-  // Here, element kind is combined with dimensions for result type.
-  ResultTy = NewSimpleAggregateType(AR_TOBJ_INVALID, resultElementKind, 0, rowCount, colCount)->getCanonicalTypeInternal();
+  // Here, element kind is combined with dimensions for primitive types.
+  if (IS_BASIC_PRIMITIVE(resultElementKind)) {
+    ResultTy = NewSimpleAggregateType(AR_TOBJ_INVALID, resultElementKind, 0, rowCount, colCount)->getCanonicalTypeInternal();
+  } else {
+    DXASSERT(rowCount == 1 && colCount == 1,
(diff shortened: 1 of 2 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -9855,8 +9855,15 @@ clang::QualType HLSLExternalSource::CheckVectorConditional(
     return QualType();
   }
 
-  // Here, element kind is combined with dimensions for result type.
-  ResultTy = NewSimpleAggregateType(AR_TOBJ_INVALID, resultElementKind, 0, rowCount, colCount)->getCanonicalTypeInternal();
```
pow2clk, line 9859: Troubling how long this has been here 😣

tools/clang/test/HLSLFileCheck/hlsl/operators/ternary-enum.hlsl
```diff
@@ -0,0 +1,36 @@
...
+#endif
+MyEnum {
+    One,
+    Two,
+};
+
+#ifdef CLASS
+static const MyEnum g_c = true ? MyEnum::One : MyEnum::Two;
+#else
+static const MyEnum g_c = true ? One : Two;
```
pow2clk, line 31: Does it have to be static const? Does it have to be the initialization? Just wondering how narrow this is. Clearly we aren't testing the different things we can plug into the ternary operator (or select()) as well as we should. What do you think about expanding this to include a number of other types? Maybe not in this change, but I think it's a test we need.

pow2clk (approved):
Everything here looks good. I'm just wondering about futher testing.

PR #4841, opened 2022-12-02
Replacing fileIO exceptions with error codes

harshpg:
The exceptions thrown by ReadBinaryFile and WriteBinaryFile give a noticeable performance impact, especially when being called by managed code which automatically tracks and logs exceptions called from native code. The exception logic can be replaced by error code return values. Adding this change greatly improved performance when calling the DXC API directly since exceptions were being thrown and reported all over the place while looking for include files.

### DIFF
--- a/include/dxc/Support/FileIOHelper.h
+++ b/include/dxc/Support/FileIOHelper.h
@@ -116,16 +116,16 @@ public:
   }
 };
 
-void ReadBinaryFile(_In_opt_ IMalloc *pMalloc,
+HRESULT ReadBinaryFile(_In_opt_ IMalloc *pMalloc,
                     _In_z_ LPCWSTR pFileName,
                     _Outptr_result_bytebuffer_(*pDataSize) void **ppData,
-                    _Out_ DWORD *pDataSize);
-void ReadBinaryFile(_In_z_ LPCWSTR pFileName,
+                    _Out_ DWORD *pDataSize) throw();
+HRESULT ReadBinaryFile(_In_z_ LPCWSTR pFileName,
                     _Outptr_result_bytebuffer_(*pDataSize) void **ppData,
-                    _Out_ DWORD *pDataSize);
-void WriteBinaryFile(_In_z_ LPCWSTR pFileName,
+                    _Out_ DWORD *pDataSize) throw();
+HRESULT WriteBinaryFile(_In_z_ LPCWSTR pFileName,
                      _In_reads_bytes_(DataSize) const void *pData,
-                     _In_ DWORD DataSize);
+                     _In_ DWORD DataSize) throw();
 
 ///////////////////////////////////////////////////////////////////////////////
 // Blob and encoding manipulation functions.
@@ -153,8 +153,7 @@ HRESULT DxcCreateBlobEncodingFromBlob(
     IMalloc *pMalloc, IDxcBlobEncoding **ppBlobEncoding) throw();
 
 // Load files
-HRESULT
-DxcCreateBlobFromFile(_In_opt_ IMalloc *pMalloc, LPCWSTR pFileName,
+HRESULT DxcCreateBlobFromFile(_In_opt_ IMalloc *pMalloc, LPCWSTR pFileName,
                       _In_opt_ UINT32 *pCodePage,
                       _COM_Outptr_ IDxcBlobEncoding **pBlobEncoding) throw();
 
--- a/lib/DxcSupport/FileIOHelper.cpp
+++ b/lib/DxcSupport/FileIOHelper.cpp
@@ -107,7 +107,7 @@ HRESULT ReadBinaryFile(IMalloc *pMalloc, LPCWSTR pFileName, void **ppData,
     return HRESULT_FROM_WIN32(GetLastError());
   }
   if (FileSize.u.HighPart != 0) {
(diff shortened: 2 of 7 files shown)


llvm-beanz (approved):
Thank you @harshpg!

This is an awesome change. I have a general dislike for exceptions myself especially as an error reporting pattern.

Before we can merge this PR, you'll need to agree to the Microsoft CLA. The bot that commented in the PR has the text of the CLA in the comment and instructions for how to agree to the CLA.

Thanks!

tex3d:
Looks good, though I'm curious about the performance impact mentioned.  Is this because ReadBinaryFile and WriteBinaryFile are being called directly from managed code, and exceptions are not handled in the code being called before crossing the managed code boundary?  I ask because there are lot of other places where exceptions are used to indicate failure in this codebase, and I expect these should not cause performance problems when the API being called catches and handles the exceptions thrown (translating exceptions to `HRESULT` for instance in `IDxcCompiler3::Compile()`).

harshpg:
@llvm-beanz 

Thank you for approving the changes!

Yes, I also prefer to use error codes instead of exceptions whenever possible.

I have requested approval from my company to be able to sign the license.
Once I get confirmation, I will go ahead and sign it.

Sorry about the delay.

llvm-beanz:
> Looks good, though I'm curious about the performance impact mentioned.

I don’t know if @harshpg has profiled, but in general an exception throw is going to be at least an order of magnitude slower than returning an int. In most ABIs returning an int will result in a register passed result, and RAII cleanup at each level of the stack, but unlike exception throwing there is no stack unwinding. Even unwinding a stack one level is likely an order of magnitude slower than a return instruction, and it can’t be optimized by the compiler. Calls can be inlined, but the stack unwind can’t.

I’ve certainly observed DXC’s file APIs throwing exceptions for recoverable and sometimes expected error propagation. Getting away from that likely has a measurable impact on performance.

harshpg:
@llvm-beanz 

I didn't profile the code directly, but I looked at the general execution time to compile all shaders as well as CPU usage. Before, the CPU usage would spike and drop due to the exceptions being thrown. After changing the logic, the spikes were gone and there was a consistent  90%+ utilization. Due to the spikes and exception handling, compilation was really slow and probably would have taken at least 2 hours to finish. After my changes, compilation takes about 5 - 6 minutes.

harshpg:
I reviewed my changes again, and noticed one mistake I made:

`return HRESULT_FROM_WIN32(DXC_E_INPUT_FILE_TOO_LARGE);`

This should be:

`return DXC_E_INPUT_FILE_TOO_LARGE;`

Should I resubmit the pull request?

harshpg:
@microsoft-github-policy-service agree [company="PlatinumGames Inc."]

harshpg:
@microsoft-github-policy-service agree company="PlatinumGames Inc."

llvm-beanz:
>  the CPU usage would spike and drop due to the exceptions being thrown.

Every exception throw more or less results in worst-case cache misses. Most modern "zero-cost" exception models rely on looking up call stack entries in side tables. Those side tables are big, and pretty much always cold data. The drop in CPU utilization is likely the result of paging in the exception side tables as the unwinder tries to figure out how far to unwind the stack and which cleanup code to run.

> After changing the logic, the spikes were gone and there was a consistent 90%+ utilization.

The before and after here is more or less consistent with what I'd expect removing exceptions. Non-exception error handling is just returning an integer up the stack. That should just be a register-return value which is super fast.

One of the other drawbacks of exception models (even "zero-cost" exceptions) is that they create boundaries the compiler can't optimize across. Eliminating the exceptions from the code should allow more aggressive inlining and other optimizations.

> Due to the spikes and exception handling, compilation was really slow and probably would have taken at least 2 hours to finish. After my changes, compilation takes about 5 - 6 minutes.

That's an awesome speedup! Is there any chance you can share (publicly or privately) more information about your shader(s) or the shader code itself? It would be very interesting for us to analyze the performance and see if there are other things we can do to help.

Overall, this seems like a big win. Thank you so much for posting it!

I'm going to run a couple tests with this change against some local testing changes I have and I'll aim to merge it in a few hours.

harshpg:
@llvm-beanz 

Thank you!

I have also attached the before and after CPU utilization images.

Unfortunately, I can't publicly share any details, but I can ask my manager tomorrow if it would be possibly to share more in private.

CA5795CF-8079-4086-B19C-29F5ECEA4A23_4_5005_c

6B30FC33-CBAF-4013-BA2B-C1D20E017B05

harshpg:
@llvm-beanz 

On a side note, I would be interested in seeing how much this change affects stutters in games that compile shaders at runtime.

llvm-beanz:
> @llvm-beanz
> 
> On a side note, I would be interested in seeing how much this change affects stutters in games that compile shaders at runtime.

I don't know that we have a good way to measure that. A lot of games that ship shader source preprocess it first, so they are shipping a single header or they use a custom include handler to avoid disk IO performance. I would expect this change to have minimal impact on those usage patterns.

harshpg:
@llvm-beanz 

Ah, yes that makes sense.

Since we are on the topic of optimizations, I wonder how much of a gain can be made by using the likely / unlikely compiler attributes added in C++ 20.

Also, thank you for merging the changes.

PR #4842, opened 2022-12-02
Restore impl of DerivedRecordSize in RDAT

tex3d:
This default method implementation is necessary for the base->derived size version mechanism to work. This is only specialized when a base has a derived, and it relies on the newest version to use the default implementation (no specialization).

The implementation was removed in commit 8ee6ed19f3adfe993e91708a24f96efb0f3a00bc.


llvm-beanz (approved):
LGTM.

PR #4843, opened 2022-12-02
Fallthrough

pow2clk:
Fix fallthrough warnings + enable Werror

A common source of bugs. Although many of these were harmless, some were
not. By eliminating them and enabling this warning as an error, we won't
add more in the future.

### DIFF
--- a/include/dxc/DXIL/DxilSigPoint.inl
+++ b/include/dxc/DXIL/DxilSigPoint.inl
@@ -166,9 +166,9 @@ DXIL::SigPointKind SigPoint::GetKind(DXIL::ShaderKind shaderKind, DXIL::Signatur
   if (isSpecialInput) {
     switch (shaderKind) {
     case DXIL::ShaderKind::Hull:
-    if (sigKind == DXIL::SignatureKind::Input)
-      return isPatchConstantFunction ? DXIL::SigPointKind::PCIn : DXIL::SigPointKind::HSIn;
-      break;
+      if (sigKind == DXIL::SignatureKind::Input)
+        return isPatchConstantFunction ? DXIL::SigPointKind::PCIn : DXIL::SigPointKind::HSIn;
+          break;
     case DXIL::ShaderKind::Geometry:
       if (sigKind == DXIL::SignatureKind::Input)
         return DXIL::SigPointKind::GSIn;
--- a/include/dxc/DXIL/DxilSigPoint.inl
+++ b/include/dxc/DXIL/DxilSigPoint.inl
@@ -168,6 +168,7 @@ DXIL::SigPointKind SigPoint::GetKind(DXIL::ShaderKind shaderKind, DXIL::Signatur
     case DXIL::ShaderKind::Hull:
     if (sigKind == DXIL::SignatureKind::Input)
       return isPatchConstantFunction ? DXIL::SigPointKind::PCIn : DXIL::SigPointKind::HSIn;
+      break;
     case DXIL::ShaderKind::Geometry:
       if (sigKind == DXIL::SignatureKind::Input)
         return DXIL::SigPointKind::GSIn;
--- a/include/dxc/DXIL/DxilSigPoint.inl
+++ b/include/dxc/DXIL/DxilSigPoint.inl
@@ -172,6 +172,7 @@ DXIL::SigPointKind SigPoint::GetKind(DXIL::ShaderKind shaderKind, DXIL::Signatur
     case DXIL::ShaderKind::Geometry:
       if (sigKind == DXIL::SignatureKind::Input)
         return DXIL::SigPointKind::GSIn;
+      break;
     default:
       break;
     }
--- a/include/dxc/Support/WinAdapter.h
+++ b/include/dxc/Support/WinAdapter.h
@@ -314,7 +314,7 @@
 #define _Null_terminated_
 
 #if __cplusplus > 201402L
-#define __fallthrough [[fallthrough]
+#define __fallthrough [[fallthrough]]
 #elif defined(__clang__)
 #define __fallthrough [[clang::fallthrough]]
 #else
--- a/include/dxc/Support/WinAdapter.h
+++ b/include/dxc/Support/WinAdapter.h
@@ -313,14 +313,6 @@
 #define _Printf_format_string_
 #define _Null_terminated_
 
-#if __cplusplus > 201402L
-#define __fallthrough [[fallthrough]]
-#elif defined(__clang__)
-#define __fallthrough [[clang::fallthrough]]
-#else
-#define __fallthrough [[gnu::fallthrough]]
-#endif
-
 #define _Field_size_(size)
 #define _Field_size_full_(size)
 #define _Field_size_opt_(size)
--- a/include/dxc/Support/WinAdapter.h
+++ b/include/dxc/Support/WinAdapter.h
@@ -312,7 +312,14 @@
 
 #define _Printf_format_string_
 #define _Null_terminated_
-#define __fallthrough
+
+#if __cplusplus > 201402L
+#define __fallthrough [[fallthrough]
+#elif defined(__clang__)
+#define __fallthrough [[clang::fallthrough]]
+#else
+#define __fallthrough [[gnu::fallthrough]]
+#endif
 
 #define _Field_size_(size)
 #define _Field_size_full_(size)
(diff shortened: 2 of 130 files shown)


tools/clang/lib/CodeGen/CGExprConstant.cpp
```diff
@@ -918,7 +919,7 @@ class ConstExprEmitter :
         llvm::Constant *destVal = llvm::Constant::getNullValue(destType);
         return destVal;
       }
-    }
+    } break;
```
llvm-beanz, line 922: Unreachable or assert here too?

tools/clang/lib/Parse/ParseDecl.cpp
```diff
@@ -4217,6 +4217,8 @@ void Parser::ParseDeclarationSpecifiers(DeclSpec &DS,
         isInvalid = true;
         break;
       };
+      ParseOpenCLQualifiers(DS.getAttributes()); // HLSL Change -
```
llvm-beanz, line 4220: I don't think you need to bring this line up. It shouldn't have been possible to hit.

tools/clang/lib/SPIRV/CapabilityVisitor.cpp
```diff
@@ -559,8 +559,8 @@ bool CapabilityVisitor::visitInstruction(SpirvInstruction *instr) {
       addCapability(spv::Capability::RayTracingKHR);
       addExtension(Extension::KHR_ray_tracing, "SPV_KHR_ray_tracing", {});
     }
+    break;
```
llvm-beanz, line 562: This looks like a behavior change... possibly a correct change, but a behavior change.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5025,6 +5025,7 @@ SpirvEmitter::processIntrinsicMemberCall(const CXXMemberCallExpr *expr,
     break;
   case IntrinsicOp::MOP_CalculateLevelOfDetail:
     retVal = processTextureLevelOfDetail(expr, /* unclamped */ false);
+    break;
```
llvm-beanz, line 5028: This also looks like a behavior change. Probably a bug fix, but a change.

lib/DXIL/DxilOperations.cpp
```diff
@@ -458,7 +458,7 @@ unsigned OP::GetTypeSlot(Type *pType) {
     case 32:              return 7;
     case 64:              return 8;
     }
-  }
+  } break;
```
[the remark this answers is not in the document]
pow2clk: I'm not sure llvm_unreachable would prevent the warning. I wouldn't mind adding it in addition to.

include/dxc/Support/WinAdapter.h
```diff
@@ -312,7 +312,14 @@
 
 #define _Printf_format_string_
 #define _Null_terminated_
-#define __fallthrough
+
+#if __cplusplus > 201402L
+#define __fallthrough [[fallthrough]
```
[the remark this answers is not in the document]
pow2clk: There's a lot to dislike along those lines in this file as it attempts to convert all the windows-isms into *nix equivalents. That said, I don't mind taking an approach similar to LLVM and replace all the platform-specific solutions, including for visual studio, with a bespoke macro. I'll do that in a separate commit to the other suggestions

tools/clang/lib/Parse/ParseDecl.cpp
pow2clk: The conundrum is that clang recognizes that it is impossible, but GCC does not. I'd like to enable warnings as errors on GCC too. So this is toward that end.

tools/clang/lib/SPIRV/CapabilityVisitor.cpp
pow2clk: It is. I'm confident it's correct. I'll ask @keenuts or @cassiebeckley to take a look. I see no reason why RT ops should fallthrough to mesh shader handling.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
pow2clk: As before, I think it's correct. @Keenuts @cassiebeckley?

lib/HLSL/DxilLinker.cpp
```diff
@@ -540,10 +540,9 @@ void DxilLinkJob::AddResourceToDM(DxilModule &DM) {
       *ptr = *(static_cast<DxilSampler *>(res));
       ID = DM.AddSampler(std::move(pSampler));
       basePtr = &DM.GetSampler(ID);
-    }
+
+    } break;
     default:
-      DXASSERT(res->GetClass() == DXIL::ResourceClass::Sampler,
```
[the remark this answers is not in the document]
pow2clk: yup! Caught that with my self-review.

pow2clk (commented):
Followup commits coming

include/dxc/DXIL/DxilSigPoint.inl
```diff
@@ -168,9 +168,11 @@ DXIL::SigPointKind SigPoint::GetKind(DXIL::ShaderKind shaderKind, DXIL::Signatur
     case DXIL::ShaderKind::Hull:
     if (sigKind == DXIL::SignatureKind::Input)
```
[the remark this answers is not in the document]
pow2clk: Looks like this caused a failure in the automated build that I guess wasn't a warning locally.

lib/DXIL/DxilOperations.cpp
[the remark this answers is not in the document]
pow2clk: Looks like it does. Alright then

lib/IR/ConstantFold.cpp
```diff
@@ -1657,7 +1657,7 @@ static ICmpInst::Predicate evaluateICmpRelation(Constant *V1, Constant *V2,
           }
         }
       }
-    }
+    } break;
```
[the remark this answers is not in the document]
pow2clk: Looks like upstream took the same approach I did. The result is returning an error code 🤷

tools/clang/lib/CodeGen/CGExprConstant.cpp
pow2clk: The unreachable on line 949 is reached by breaking here.

pow2clk: I suppose we could have a more specialized message though

tools/clang/lib/SPIRV/SpirvEmitter.cpp
pow2clk: This changed the output of a test, but since the test was expecting two LOD calculations where it only asked for one, I feel good about it. That said, it looks like the generated results were the same suggesting no distinction between `CalculateLevelofDetail` and `CalculateLevelofDetailClamped`. Which isn't right.

tools/clang/lib/Parse/ParseDecl.cpp
llvm-beanz: Bringing the break up to fix the warning seems correct, but the `ParseOpenCLQualifiers` shouldn't need to be copied up.

llvm-beanz (approved):
Looks good to me.

tools/clang/lib/SPIRV/CapabilityVisitor.cpp
Keenuts: Yes, that's a behavior change, but a good one 🙃
This was a bug.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts: TL;DR: thanks for fixing! Break is needed!

Wasn't aware of this at all. Looked into the SPIR-V spec, and I have trouble parsing it:
```
The first component of the result contains the mipmap array layer.
The second component of the result contains the implicit level of detail relative to the base level.
```

The initial commit added the implicit fallthrough, which is surprising, but the test code seems to match what glslang is doing, aka taking the second component of the result: https://github.com/KhronosGroup/glslang/issues/1423

Looking at tests at ce495eb611be06cb7249c174ea93a61ec163eaeb, it does seem to work as expected at least from the test PoV but I believe it didn't. The tests looks for positive, not negative presence. Hence because both were present, even if it was dead code.
Thanks for fixing!

Keenuts (approved):

PR #4845, opened 2022-12-03
dxcopt: Support full container and restore extra data to module

### LINKED ISSUE #4874: Extra container data lost when running IDxcOptimizer on bitcode from final DxilContainer
If one wants to use IDxcOptimizer to run passes on a DxilModule that has been serialized to a container, there are various pieces of data that have been stripped from the module and placed into other container parts which cannot be restored without a more painful manual process.

If IDxcOptimizer could accept a DxilContainer, rather than only module bitcode, it could restore data from the parts to the module, allowing that data to survive the optimization pipeline.  When that module is serialized back to a DxilContainer, it will have the data to place into the various parts again.

Lost data that could be restored to the module includes:
  - Subobjects from RDAT
  - RootSignature from RTS0
  - ViewID and I/O dependency data from PSV0
  - Resource names and types/annotations from STAT

As it turns out, this is just what's needed by PIX, especially for preserving the subobject data from RDAT.

tex3d:
This modifies IDxcOptimizer::RunOptimizier to accept full DxilContainer input. When full container input is used, this restores some data that is stripped from the module and placed in various other container parts.

Data restored:
  - Subobjects from RDAT
  - RootSignature from RTS0
  - ViewID and I/O dependency data from PSV0
  - Resource names and types/annotations from STAT

Serialization of these to metadata in module bitcode output still requires hlsl-dxilemit step.

Fixes #4874


lib/HLSL/DxcOptimizer.cpp
```diff
@@ -249,28 +250,77 @@ HRESULT STDMETHODCALLTYPE DxcOptimizer::RunOptimizer(
...
-    M = parseIR(memBuf->getMemBufferRef(), Err, Context);
-  }
+      reinterpret_cast<const DxilProgramHeader *>(pBlobContent);
+  const DxilContainerHeader *pContainerHeader = IsDxilContainerLike(pBlobContent, blobSize);
+  bool bIsFullContainer = IsValidDxilContainer(pContainerHeader, blobSize);
 
-  if (M == nullptr) {
-    return DXC_E_IR_VERIFICATION_FAILED;
-  }
+  try {
```
adam-yang, line 257: Context might need to be moved in here, since it could throw.

jeffnn (approved):
Confirmed that this fixes the PIX scenario.

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -259,3 +269,273 @@ void OptimizerTest::OptimizerWhenSliceNThenOK(int optLevel, LPCSTR pText, LPCWST
...
+  pAssembler->AssembleToContainer(pDxil, &result);
+
+  CComPtr<IDxcBlob> pOptimizedContainer;
+  result->GetResult(&pOptimizedContainer);
+
+  CComPtr<IDxcContainerReflection> pReflection;
+  VERIFY_SUCCEEDED(m_dllSupport.CreateInstance(CLSID_DxcContainerReflection, &pReflection));
+  VERIFY_SUCCEEDED(pReflection->Load(pOptimizedContainer));
+  UINT32 dxilIndex;
+  VERIFY_SUCCEEDED(pReflection->FindFirstPartKind(hlsl::DFCC_DXIL, &dxilIndex));
```
tex3d, line 399: Is there a reason this uses container reflection here, then the lower-level DxilContainer code below?  Either one works, but it doesn't seem necessary to use a different method when getting a different part.

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -259,3 +269,273 @@ void OptimizerTest::OptimizerWhenSliceNThenOK(int optLevel, LPCSTR pText, LPCWST
...
+
+  const char *pBlobContent =
+      reinterpret_cast<const char *>(pOptimizedContainer->GetBufferPointer());
+  unsigned blobSize = pOptimizedContainer->GetBufferSize();
+  const hlsl::DxilContainerHeader *pContainerHeader =
+      hlsl::IsDxilContainerLike(pBlobContent, blobSize);
```
tex3d, line 405: Just a note: After `IsDxilContainerLike`, you can also use `IsValidDxilContainer` to make sure the container is well-formed (though it's unlikely to be badly formed on output from `AssembleToContainer`).

tools/clang/unittests/HLSL/OptimizerTest.cpp
jeffnn: Just because I had lifted the lower-level code from elsewhere prior to deleting the code in the optimizer that serialized the dxil to container.

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -259,3 +269,273 @@ void OptimizerTest::OptimizerWhenSliceNThenOK(int optLevel, LPCSTR pText, LPCWST
...
+  auto const subObjectTableReader = rdat.GetSubobjectTable();
+
+  // There are 9 subobjects in the HLSL above:
+  VERIFY_ARE_EQUAL(subObjectTableReader.Count(), 9u);
+  for (uint32_t i = 0; i < subObjectTableReader.Count(); ++i) {
+    auto subObject = subObjectTableReader[i];
+    hlsl::DXIL::SubobjectKind subobjectKind = subObject.getKind();
+    switch (subobjectKind) {
+    case hlsl::DXIL::SubobjectKind::StateObjectConfig:
+      VERIFY_IS_TRUE(0 == strcmp(subObject.getName(), "so_StateObjectConfig"));
```
tex3d, line 423: I think it's better to use VERIFY_ARE_EQUAL_STR, since that will be more informative if it fails.

lib/HLSL/DxcOptimizer.cpp
```diff
@@ -221,6 +225,69 @@ HRESULT STDMETHODCALLTYPE DxcOptimizer::GetAvailablePass(
...
+  TargetRes.SetHLSLType(Ty);
+
+  if (isa<PointerType>(Ty))
+    Ty = Ty->getPointerElementType();
+
+  while (isa<ArrayType>(Ty))
+    Ty = Ty->getArrayElementType();
+
+  // TODO: Detect if struct annotation is stripped, then remove it if so, so
+  // it can be replaced.  For now, just always replace it.
```
tex3d, line 254: I don't know if this comment is very clear, and I don't think the case is necessary after all.  I'll remove the comment.

lib/HLSL/DxcOptimizer.cpp
```diff
@@ -221,6 +225,69 @@ HRESULT STDMETHODCALLTYPE DxcOptimizer::GetAvailablePass(
...
+  if (StructType *ST = dyn_cast_or_null<StructType>(Ty)) {
+    if (const DxilStructAnnotation *STAnn = TargetTypeSys.GetStructAnnotation(ST))
+      TargetTypeSys.EraseStructAnnotation(ST);
+  }
+  TargetTypeSys.CopyTypeAnnotation(Ty, SourceTypeSys);
+}
+
+static void CopyMatchingResourceInfo(DxilModule &TargetDM,
```
tex3d, line 262: One thing I was thinking is that this code could be moved to `DxilModule.cpp`.

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -259,3 +303,1152 @@ void OptimizerTest::OptimizerWhenSliceNThenOK(int optLevel, LPCSTR pText, LPCWST
...
+
+  VERIFY_ARE_EQUAL(original.size(), optimized.size());
+  for (size_t i = 0; i < original.size(); ++i) {
+    auto const &originalRes = original.at(i);
+    auto const &optimizedRes = optimized.at(i);
+    VERIFY_ARE_EQUAL(originalRes->GetClass(), optimizedRes->GetClass());
+    VERIFY_ARE_EQUAL(originalRes->GetGlobalName(), optimizedRes->GetGlobalName());
+    VERIFY_ARE_EQUAL(originalRes->GetGlobalSymbol()->getName(),
+                     optimizedRes->GetGlobalSymbol()->getName());
+    VERIFY_ARE_EQUAL(originalRes->GetHLSLType()->getTypeID(),
```
tex3d, line 1232: Might have to be careful in case `GetHLSLType()` returns nullptr.  And, I'm not sure how useful this check is.

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -259,3 +303,1152 @@ void OptimizerTest::OptimizerWhenSliceNThenOK(int optLevel, LPCSTR pText, LPCWST
...
+void CompareResources(const vector<unique_ptr<ResourceType>> &original,
+                      const vector<unique_ptr<ResourceType>> &optimized) {
+
+  VERIFY_ARE_EQUAL(original.size(), optimized.size());
+  for (size_t i = 0; i < original.size(); ++i) {
+    auto const &originalRes = original.at(i);
+    auto const &optimizedRes = optimized.at(i);
+    VERIFY_ARE_EQUAL(originalRes->GetClass(), optimizedRes->GetClass());
+    VERIFY_ARE_EQUAL(originalRes->GetGlobalName(), optimizedRes->GetGlobalName());
+    VERIFY_ARE_EQUAL(originalRes->GetGlobalSymbol()->getName(),
```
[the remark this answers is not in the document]
jeffnn: Fixed

tools/clang/unittests/HLSL/OptimizerTest.cpp
jeffnn: Cuz it'll always be struct?

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -259,3 +303,1152 @@ void OptimizerTest::OptimizerWhenSliceNThenOK(int optLevel, LPCSTR pText, LPCWST
...
+                     optimizedRes->GetGlobalSymbol()->getName());
+    VERIFY_ARE_EQUAL(originalRes->GetHLSLType()->getTypeID(),
+                     optimizedRes->GetHLSLType()->getTypeID());
+    if (originalRes->GetHLSLType()->getTypeID() == Type::TypeID::StructTyID) {
```
tex3d, line 1234: This will probably never be true, since it should normally be a pointer type which points to either a struct type or a potentially nested array type with the struct type as the final array element type.

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -259,3 +303,1152 @@ void OptimizerTest::OptimizerWhenSliceNThenOK(int optLevel, LPCSTR pText, LPCWST
...
+      VERIFY_ARE_EQUAL(originalRes->GetHLSLType()->getStructName(),
```
tex3d, line 1235: `getStructName()` returns a llvm::StringRef.  This might be doing a pointer comparison on a `const char *` that this will implicitly cast to.  If execution ever reaches this code that is.

jeffnn: I checked, and StringRef has an operator ==, so I think VERIFY_ARE_EQUAL is ok

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -259,3 +303,1152 @@ void OptimizerTest::OptimizerWhenSliceNThenOK(int optLevel, LPCSTR pText, LPCWST
...
+    VERIFY_ARE_EQUAL(originalPsv.GetViewIDOutputMask(stream).IsValid(),
+                     optimizedPsv.GetViewIDOutputMask(stream).IsValid());
+    if (originalPsv.GetViewIDOutputMask(stream).IsValid())
+    {
+      VERIFY_ARE_EQUAL(*originalPsv.GetViewIDOutputMask(stream).Mask,
+                       *optimizedPsv.GetViewIDOutputMask(stream).Mask);
+      VERIFY_ARE_EQUAL(originalPsv.GetViewIDOutputMask(stream).NumVectors,
+                       optimizedPsv.GetViewIDOutputMask(stream).NumVectors);
+
+    }
```
tex3d, line 610: This misses comparing the input->output dependency bits, which are set regardless of whether `SV_ViewID` is used in the shader.

tools/clang/unittests/HLSL/OptimizerTest.cpp
tex3d: Ok, but what does it do if it's not equal?  I think the _STR variant is always better since it will write out the mismatching strings when it fails.

tools/clang/unittests/HLSL/OptimizerTest.cpp
tex3d: No, it should always be a `PointerType`, I think.

tools/clang/unittests/HLSL/OptimizerTest.cpp
jeffnn: Fixed to follow pointer

tools/clang/unittests/HLSL/OptimizerTest.cpp
jeffnn: Looks like VERIFY_ARE_EQUAL prints both values anyway, if they're different, and works with std::string (I tested that). Also there's no default conversion to char * for std::string or StringRef, so VERIFY_ARE_EQUAL_STR doesn't work without adding .c_str() anyway.

Example of comparison failure from  VERIFY_ARE_EQUAL operating on StringRef :

Error: Verify: AreEqual(originalPointedType->getStructName(), "Funny") [File: D:\DirectXShaderCompiler\tools\clang\unittests\HLSL\OptimizerTest.cpp, Function: CompareResources, Line: 1254]

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -442,6 +442,180 @@ DxilPartWriter *hlsl::NewFeatureInfoWriter(const DxilModule &M) {
...
 
+
+//////////////////////////////////////////////////////////
+// Utility code for serializing/deserializing ViewID state
+
+// Code for ComputeSeriaizedViewIDStateSizeInUInts copied from
+// ComputeViewIdState. It could be moved into some common location if this
+// ViewID serialization/deserialization code were moved out of here.
+static unsigned RoundUpToUINT(unsigned x) { return (x + 31) / 32; }
+static unsigned ComputeSeriaizedViewIDStateSizeInUInts(
```
adam-yang, line 453: Are we going to make ComputeViewIdState call this?

tools/clang/unittests/HLSL/OptimizerTest.cpp
jeffnn: LMK if what I've done is insufficient.

lib/DxilContainer/DxilContainerAssembler.cpp
tex3d: Ideally, this could live in a common location, but I haven't yet figured out where that should be.  Hence the comment on the function.

tools/clang/unittests/HLSL/OptimizerTest.cpp
tex3d: Number of input and output vectors is not going to be different as long as there is data here, it's the contents of the masks that matter.  If you assume that the mere presence of the ViewID data is enough, this check is enough, but I'm not sure that's enough given the potential to make a mistake in the code that transfers the data from PSV0 to the DxilModule serialized format.

tex3d: A simple comparison of the data might be fine.  The data at `PSVDependencyTable::Table` is an array of `uint32_t` with an array size of `PSVComputeMaskDwordsFromVectors(OutputVectors) * InputVectors * 4`.  You could just to a `memcmp` on that.

tex3d: This also misses any potential table from `GetInputToPCOutputTable` or `GetPCInputToOutputTable`.  The content `memcmp` could be performed when `InputVectors * OutputVectors > 0`.

Likely it'll be best to write a little helper function (or lambda), like this:
```
void VerifyIOTablesMatch(const PSVDependencyTable &original,
                         const PSVDependencyTable &optimized) {
  VERIFY_ARE_EQUAL(original.InputVectors, optimized.InputVectors);
  VERIFY_ARE_EQUAL(original.OutputVectors, optimized.OutputVectors);
  if (original.InputVectors == optimized.InputVectors &&
      original.OutputVectors == optimized.OutputVectors &&
      original.InputVectors * original.OutputVectors > 0) {
    VERIFY_ARE_EQUAL(
        0, memcmp(original.Table, optimized.Table,
                  PSVComputeMaskDwordsFromVectors(original.OutputVectors) *
                      original.InputVectors * 4));
  }
}
```

Then use that on all three possible tables.

tex3d: That won't tell you where it went wrong, but will catch it if it is wrong.

tools/clang/unittests/HLSL/OptimizerTest.cpp
```diff
@@ -1227,22 +1236,32 @@ void CompareResources(const vector<unique_ptr<ResourceType>> &original,
...
     VERIFY_ARE_EQUAL(originalRes->GetID(), optimizedRes->GetID());
     VERIFY_ARE_EQUAL(originalRes->GetKind(), optimizedRes->GetKind());
     VERIFY_ARE_EQUAL(originalRes->GetLowerBound(), optimizedRes->GetLowerBound());
     VERIFY_ARE_EQUAL(originalRes->GetRangeSize(), optimizedRes->GetRangeSize());
-    VERIFY_ARE_EQUAL(originalRes->GetResBindPrefix(), optimizedRes->GetResBindPrefix());
-    VERIFY_ARE_EQUAL(originalRes->GetResClassName(), optimizedRes->GetResClassName());
-    VERIFY_ARE_EQUAL(originalRes->GetResDimName(), optimizedRes->GetResDimName());
-    VERIFY_ARE_EQUAL(originalRes->GetResIDPrefix(), optimizedRes->GetResIDPrefix());
+    VERIFY_ARE_EQUAL_STR(originalRes->GetResBindPrefix(), optimizedRes->GetResBindPrefix());
+    VERIFY_ARE_EQUAL_STR(originalRes->GetResIDPrefix(), optimizedRes->GetResIDPrefix());
```
tex3d, line 1264: These are just the prefix character decoded from the Class, so comparing `GetResBindPrefix` and `GetResIDPrefix` are also unnecessary.  Basically, if Class was wrong, there would be something very fundamentally wrong with which list the resource is in for the DxilModule.  By which I mean we would have much bigger problems, and expect a whole lot of things to break.

tools/clang/unittests/HLSL/OptimizerTest.cpp
jeffnn: Now testing all 3 tables

lib/DxilContainer/DxilContainerAssembler.cpp
tex3d: I do plan to clean this up along with the format inconsistency between module and PSV0 in the near future, so I'd rather clean this up at that point.

adam-yang (approved):
LGTM!

PR #4846, opened 2022-12-03
Enable optimizer to receive full container, and re-emit e.g. RDAT after passes run

jeffnn:
DRAFT!


jeffnn:
Tex took care of this one.

PR #4850, opened 2022-12-06
LLVM_REQUIRES_* variables shound't be set

llvm-beanz:
The `LLVM_REQUIRES_*` variables are internal variables in LLVM's configuration used to determine if a build setting is required, they don't actually control the build setting itself, so we shouldn't be setting them.

`LLVM_ENABLE_*` variables control the build settings, and we do set those correctly.


hekota (approved):
LGTM!

PR #4851, opened 2022-12-06
Remove use of `dynamic_cast`

llvm-beanz:
This is the only instance of a C++ dynamic_cast in the codebase and it can be avoided by just making an interface virtual.


llvm-beanz:
Abandoning because I'm not going to follow up on this. We would also need to eliminate exception usage to turn off RTTI.

PR #4852, opened 2022-12-06
Fix dxcisense.h includes

hekota:
- dxcapi.h is in the same directory and the dxc directory does not exist in shipped packages
- WinAdapter.h not needed on Windows


pow2clk (approved):
Looks good!

PR #4854, opened 2022-12-07
Converting std::set to std::unordered_set when order does not matter

harshpg:
`std::set` uses a red black tree as its internal data structure. This is good when sorting is necessary, but has overhead, and the overhead is not necessary when sorting is not required. In those cases, an `unordered_set` should be used. There could be more that could be replaced, but this change attempts to be more conservative.


Keenuts:
Hi! Thanks for the PR! 😊
Overall, I like the change. Just unsure about some parts, like DominanceFrontier, since work is added following the set order, does it has any signification? Since all tests are passing I'd say no (also passed on our internal CI)
Will let MSFT folks approve since it mostly impacts the core DXC side.

harshpg:
@Keenuts 
Thanks for checking out the PR!

I'm not sure about that either, but I guess its fine since the tests are passing? Might be good to have a few more eyes look over the changes. I also added one more change to `AssertingVH `so that it can also be moved into an unordered set.

Thanks!

harshpg:
I also measured the difference in compile times using a sample of our shaders:

```
Before (ms):
249
59
102
73
88
77
73
70
557
104
71
252
36
51
95
74
1817
```

```
After (ms):
107
57
67
51
58
63
67
69
578
101
57
208
38
42
79
54
1813
```

llvm-beanz:
Have you tried using `llvm::DenseSet` instead. Many of the built-in LLVM data structures are optimized for smaller data sets that are common in compiler algorithms, and the often perform much better than the `std` variants.

harshpg:
@llvm-beanz 

I have not tried `llvm::DenseSet` yet. Does it keep objects sorted inside contiguously?

llvm-beanz:
> @llvm-beanz
> 
> I have not tried `llvm::DenseSet` yet. Does it keep objects sorted inside contiguously?

`llvm::DenseMap` (the data structure `llvm::DenseSet` is built on) is an unordered hash map, but it groups multiple keys and values into buckets in the same allocation. It can have slightly more memory overhead but it is way more cache efficient and usually has faster insertion time since insertion doesn’t always require allocation.

I would expect that `llvm::DenseSet` and `llvm::DenseMap` would provide even more performance over using the `std::unordered` structures.

harshpg:
@llvm-beanz 
Ah, that sounds quite nice! I'll try using `llvm::DenseSet` and see what the performance looks like.

damyanp:

damyanp:

PR #4855, opened 2022-12-07
ConvertFloat32ToFloat16: Use DirectXMath conversion functions

tex3d:
Custom half <-> float conversion functions had problems in multiple scenarios.  This PR changes them into a wrapper, using the DirectXMath conversion functions instead.


tools/clang/unittests/HLSL/ShaderOpTest.cpp
```diff
@@ -40,6 +41,13 @@
 ///////////////////////////////////////////////////////////////////////////////
 // Useful helper functions.
 
+uint16_t ConvertFloat32ToFloat16(float Value) throw() {
+  return DirectX::PackedVector::XMConvertFloatToHalf(Value);
+}
+float ConvertFloat16ToFloat32(uint16_t Value) throw() {
+  return DirectX::PackedVector::XMConvertHalfToFloat(Value);
+}
+
```
pow2clk, line 50: I know you want Helena to comment here and she should, but we talked about this earlier and she plans to move a lot of the ShaderOp code into its own directory. That might resolve any concerns about dependencies involving execution tests.

tex3d: Fortunately, this cpp file is already dependent on DirectXMath.  So, I don't think there will be any trouble with this change.  I still wanted to double check that it doesn't break tests in some unexpected way though.

hekota: I believe this should be fine, as Tex said the dependency is already there, and these headers should be available in the Windows build. It is the LLVM dependencies and additional DXC headers that are problematic.

hekota (approved):
LGTM!

PR #4859, opened 2022-12-08
Fix enum scope used inside template function.

tex3d:
While enums were introduced in HLSL 2017, this code should only be reached on HLSL if we are HLSL 2021 or greater, so an additional language version check does not appear necessary here.


llvm-beanz (approved):
LGTM

PR #4860, opened 2022-12-09
Fixing a [[nodiscard]] error in DXC

Keenuts:
This was found while moving to c++20 by an anonymous internal contributor. Integrating the finding to mainline.

std::unique iterates on elements of the container, and removes all but the first element matching the predicate, but does not resize the container. This means the following code doesn't assert.

```cpp
    std::vector<int> v = { 1, 1, 2, 2, 3, 3 };
    assert(v.size() == 6);
    std::unique(v.begin(), v.end());
    assert(v.size() == 6);
```

The function returns a past-end iterator, and all the undefined elements are at the end of the container, meaning we can safely erase them, and fix the container size.

### DIFF
--- a/tools/clang/include/clang/Serialization/ContinuousRangeMap.h
+++ b/tools/clang/include/clang/Serialization/ContinuousRangeMap.h
@@ -117,14 +117,17 @@ public:
     
     ~Builder() {
       std::sort(Self.Rep.begin(), Self.Rep.end(), Compare());
-      std::unique(Self.Rep.begin(), Self.Rep.end(),
-                  [](const_reference A, const_reference B) {
-        // FIXME: we should not allow any duplicate keys, but there are a lot of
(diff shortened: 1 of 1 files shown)


llvm-beanz (commented):
Please clang-format the block. @rnk made this same change in upstream clang [dd870f6929ee](https://github.com/llvm/llvm-project/commit/dd870f6929ee0b1dfb5fd000c7b826a1bd2d2571).

@rnk, do you know if there was any fallout from this change? I suspect not, but just want to verify.

Keenuts:
RIght! Forgot to run it, thanks!

Keenuts:
A quick look into the commit history, I see no revert and doesn't seem to have related fixes in the following months (checked 2)

llvm-beanz (approved):
LGTM

PR #4864, opened 2022-12-10
Remove template/ByteAddressBufferLoad test dependency on high-level opcode

tex3d:
High-level opcode is not stable, except within a build.  Tests should not take a dependency on these opcode numbers.

Fixed the template/ByteAddressBufferLoad test that was failing due to this dependency.


hekota (approved):
LGTM!

PR #3395, opened 2021-01-26
[spirv] Add vk::image_format attribute for Buffers, RWBuffers and RWTextures

seppala2:
According to Vulkan specification when using `OpImageRead/OpImageWrite`, the `OpTypeImage` (`Buffers`, `RWBuffers`, `RWTextures`) must have a format that matches the format on the API side, unless the StorageImageReadWithoutFormat/StorageImageWriteWithoutFormat is added and `Unknown` is used as the format.

This pull request addressess #2498 for the format part by adding an attribute `[[vk::image_format("<image format as spelled in SPIR-V spec>")]].` Example of the syntax:

```
[[vk::image_format("rgba8")]]
RWBuffer<float4> Buf;

[[vk::image_format("rg16f")]]
RWTexture2D<float2> Tex;

RWTexture2D<float2> Tex2; // Works like before
```

The `image_format` only applies to **global variables** of type `Buffer`, `RWBuffer`, `RWTexture`. For variables and function parameters it is propagated by the inlining pass in legalization. This required a small change to one of the passes in SPIRV-Tools, that should be also checked by someone more familiar with the codebase: https://github.com/KhronosGroup/SPIRV-Tools/pull/4126

Note that this does not fix the handling of unspecified format (that case still works like before, using `R32f`, etc. based on the type in shader), although it should be still fixed to add the
StorageImageReadWithoutFormat and/or StorageImageWriteWithoutFormat and use Undefined. But I think the ability to specify the format is more urgent.

Design note from Jaebaek:
Since the `image_format` attribute only applies to **global variables**, under the DXC architecture
only `DeclResultIdMapper` can check the attribute when it handles `VarDecl`s. It means
we have to pass the `image_format` information to `LowerTypeVisitor` because it cannot access to
`VarDecl`. In order to pass the `image_format`, we use `SpirvContext` that can be accessed by
`SpirvEmitter` and all visitors. We use `SpirvVariable` to `spv::ImageFormat` mapping because the
attribute only applies to **global variables** (not to image types).
See how we use `llvm::DenseMap<const SpirvVariable *, spv::ImageFormat> spvVarToImageFormat`.

PR #4868, opened 2022-12-12
spirv: keep image format to unknown if it is.

### LINKED ISSUE #4773: [SPIR-V] invalid OpImageWrite instruction generated if vk::image_format is not specified.
Issue derived from (https://github.com/microsoft/DirectXShaderCompiler/issues/4772)

HLSL has attributes to specify the texture format (https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#textures).

When not specified, the fallback is Rgba32f (first type of the list).
Given this sample:
```hlsl
RWTexture2D<float3> uavs[2];

[numthreads(1,1,1)]
void maincomp(uint3 threadID : SV_DispatchThreadID)
{
  uavs[threadID.z][threadID.xy] = float3(0.5, 0.2, 1.0);
}
```

and this command:
```
dxc repro.hlsl -E maincomp -T cs_6_0 -spirv -O0
```

The generated code is invalid (breaks VUID-RuntimeSpirv-OpImageWrite-07112)
```
OpImageWrite to any Image whose Image Format is not Unknown must have the Texel operand contain at least as many components as the corresponding [VkFormat](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkFormat.html) as given in the [SPIR-V Image Format compatibility table](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#spirvenv-image-formats)
```

Reason is we default the image format to RGBA32F instead of unknown.
A quick inspection in DXC showed we set it to unknown at first, but then the information is ignored when the type is lowered, and the image format is selected depending on the sampled type (float3), hence RGBA32F, since RGB32F doesn't exists.

Since we emit programs with the `shader` capability, we could enable `StorageImageReadWithoutFormat` and keep the format as `unknown`.

Keenuts:
This commit aim to fix #4773. In SPIR-V, image type must be specified, and GLSL has the syntax for it. In HLSL, we don't, until #3395. When this attribute is not specified, we decided to guess the type, not using Unknown anymore.
But the type guessing was a bit too optimistic, and thus allowed a 3 component layout to be upgraded to 4 component layout.

Changing the guessing function to either give a close type, or return Unknown, adding the capability on load/store to the variable.

### DIFF
--- a/tools/clang/lib/SPIRV/CapabilityVisitor.cpp
+++ b/tools/clang/lib/SPIRV/CapabilityVisitor.cpp
@@ -432,6 +432,22 @@ bool CapabilityVisitor::visit(SpirvImageSparseTexelsResident *instr) {
   return true;
 }
 
+namespace {
+bool isImageOpOnUnknownFormat(const SpirvImageOp *instruction) {
+  if (!instruction->getImage() || !instruction->getImage()->getResultType()) {
+    return false;
+  }
+
+  const ImageType *imageType =
+      dyn_cast<ImageType>(instruction->getImage()->getResultType());
+  if (!imageType || imageType->getImageFormat() != spv::ImageFormat::Unknown) {
+    return false;
+  }
+
+  return imageType->getImageFormat() == spv::ImageFormat::Unknown;
+}
+} // namespace
+
 bool CapabilityVisitor::visit(SpirvImageOp *instr) {
   addCapabilityForType(instr->getResultType(), instr->getSourceLocation(),
                        instr->getStorageClass());
@@ -441,6 +457,12 @@ bool CapabilityVisitor::visit(SpirvImageOp *instr) {
     addCapability(spv::Capability::MinLod);
   if (instr->isSparse())
     addCapability(spv::Capability::SparseResidency);
+
+  if (isImageOpOnUnknownFormat(instr)) {
+    addCapability(instr->isImageWrite()
+                      ? spv::Capability::StorageImageWriteWithoutFormat
+                      : spv::Capability::StorageImageReadWithoutFormat);
+  }
 
   return true;
 }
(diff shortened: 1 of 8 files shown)


Keenuts:
@cassiebeckley anything blocking on your end?

cassiebeckley (approved):

Keenuts:
@godlikepanos sorry, didn't had time to dive back on this specific issue.
That's a valid point. Initially I wanted to be conservative on the change, as I'm still new to this, but you might be right.
I need to check that again.

godlikepanos:
@Keenuts Let me do some spec digging and I'll get back to you.

godlikepanos:
Hi @Keenuts,

Vulkan spec has this table which ties the OpTypeImage's format to the VkFormat of the VkImageView: https://registry.khronos.org/vulkan/specs/1.1-extensions/html/chap49.html#spirvenv-image-formats. So the existing DXC implementation that is guessing the OpTypeImage format is not quite correct. 

I also had a chat with @HansKristian-Work who works on vkd3d-proton and he mentioned to me that HLSL doesn't have the concept of format in the shader anyway.  

We think that the best way forward for DXC is to always set the format to unknown. If the GPU can't work with unknown formats then the author of the shader should use `[[vk::image_format("xxx")]]` to specify it. When it comes to unknown formats almost all GPUs allow them for writing (VkPhysicalDeviceFeatures::shaderStorageImageWriteWithoutFormat) and 70% for reading (VkPhysicalDeviceFeatures::shaderStorageImageReadWithoutFormat). It's mainly Intel that doesn't support shaderStorageImageReadWithoutFormat but not for all formats really since shaderStorageImageReadWithoutFormat is all or nothing. There is now the VK_KHR_format_feature_flags2 extension which exposes the shaderStorageImageWriteWithoutFormat/shaderStorageImageReadWithoutFormat for each format individually.

If you have any thoughts or you need any help please let me know. I can dig deeper.

Keenuts:
Thanks for all the info! Created https://github.com/microsoft/DirectXShaderCompiler/issues/4941 to track this so we don't forget about it.

PR #4869, opened 2022-12-12
Add -enable-lit option to hcttest

llvm-beanz:
This patch attempts to translate HCTTest options to LIT build targets. Since HCTTest and lit break up tests differently the mapping isn't quite the same, but it should be close enough.

Before we can migrate our automation builds to using lit we need to separate out the exec tests so that they can be disabled on the ADO bots.

### DIFF
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -291,9 +291,9 @@ if not exist %TEST_DIR% (mkdir %TEST_DIR%)
 echo Copying binaries to test to %TEST_DIR%:
 if "%CUSTOM_BIN_SET%"=="" (
   if not "%TEST_USE_LIT%"=="1" (
-    set TAEF_TEST_DLL=clang-hlsl-tests.dll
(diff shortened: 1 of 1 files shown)


utils/hct/hcttest.cmd
```diff
@@ -252,11 +255,45 @@ if "%TEST_CLEAN%"=="1" (
...
+      cmake --build %HLSL_BLD_DIR% --config %BUILD_CONFIG% --target check-dxilconv
+      set RES_DXILCONV=!ERRORLEVEL!
+    )
+    if "%TEST_CLANG%"=="1" (
+      cmake --build %HLSL_BLD_DIR% --config %BUILD_CONFIG% --target check-clang
+      set RES_CLANG=!ERRORLEVEL!
+      set RES_EXEC=%RES_CLANG%
+    )
+  )
+  set TEST_CLANG=0
```
hekota, line 283: It might be better to use goto :eof here to skip any further processing.

hekota (approved):

PR #4872, opened 2022-12-12
Unbuffer raw_string_stream

llvm-beanz:
Double buffering a string stream writing in memory is silly. Upstream LLVM made this change a few years ago based on analysis that most string implementations are sane:

https://github.com/llvm/llvm-project/commit/65b1361


dmpots (approved):
LGTM

tex3d (approved):
LGTM.

PR #4873, opened 2022-12-12
Backport -ftime-trace from upstream llvm & clang

llvm-beanz:
This change back-ports the -ftime-trace feature implemented by Aras Pranckevičius (@aras-p) into DXC.

Aras has a great blog post describing the motivaiton for this feature and how to use it (see:
https://aras-p.info/blog/2019/01/16/time-trace-timeline-flame-chart-prof iler-for-Clang/).

As shaders are getting more and more complex we need better timing data for ourselves as compiler writers and our users as HLSL writers to know how to improve compile times of complex shaders.

This implementation is a starting point to give us data to help drive our efforts to improve compile-time performance.

Because DXC's LLVM is very far removed from the version of LLVM the patch was originally written for this change is not strictly a cherry-pick of
https://github.com/llvm/llvm-project/commit/d880de2d19d46f7cfea7aa593602 458440bc1e8d. Instead this change is a manually applied and reconstructed patch with significant alterations along the way. The alaterations are performed by myself, but credit for the patch remains with @aras-p.

Patch by Aras Pranckevičius (@aras-p).

### DIFF
--- a/include/dxc/Support/HLSLOptions.h
+++ b/include/dxc/Support/HLSLOptions.h
@@ -207,6 +207,7 @@ public:
   bool EnableLifetimeMarkers = false; // OPT_enable_lifetime_markers
   bool ForceDisableLocTracking = false; // OPT_fdisable_loc_tracking
   bool TimeReport = false; // OPT_ftime_report
+  std::string TimeTrace = ""; // OPT_ftime_trace[EQ]
 
   // Optimization pass enables, disables and selects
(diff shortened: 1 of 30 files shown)


dmpots (approved):
Touches a lot of places, but looks pretty straightforward. I didn't review the actual TimerProfile code closely since I assume that is an already well tested upstream component.

The actual dxc-specific changes seem reasonable to me.

PR #4876, opened 2022-12-13
PIX: Modify root sigs in place (plus fix root sig memory leak)

jeffnn:
PIX is unique in that it needs to deserialize, _modify_ and then reserialize root sigs. The focus of this checkin is adding such modifications to PixPassHelpers.cpp. (This closes a gap in PIX support: PIX can now support shaders that use dxil-defined and attribute-style root signatures.)

But this required some work outside of the purely PIX-focused areas. Deserialized root sigs are described by a C-like structure with embedded arrays that are new/delete-ed by routines in DxilRootSignature.cpp. Since modifying these structs requires more new/delete calls, I chose to add a new entry point in DxilRootSignature.cpp to do the bare minimum that PIX needs: extend root params by one descriptor. This approach keeps all those raw new/deletes in one file.

I found a leak in DxilRootSignatureSerialzier.cpp, which I fixed. There appear to be no unit tests that exercise this path. I welcome feedback on adding one.

There were other leaks in class CVersionedRootSignatureDeserializer, but this class is unused so I deleted it.

Oh, and there are bazillions of commits because I was cherry-picking from a recent change (https://github.com/microsoft/DirectXShaderCompiler/pull/4845)  as it eveolved, since I needed that change and this to test PIX.


lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -159,7 +165,97 @@ llvm::CallInst *CreateHandleForResource(hlsl::DxilModule &DM,
...
+
+static void AddUAVToDxilDefinedGlobalRootSignatures(DxilModule& DM) {
+  auto *subObjects = DM.GetSubobjects();
+  if (subObjects != nullptr) {
+    for (auto const &subObject : subObjects->GetSubobjects()) {
+      if (subObject.second->GetKind() ==
+          DXIL::SubobjectKind::GlobalRootSignature) {
+        const char *Text;
+        const void *Data;
+        uint32_t Size;
```
pow2clk, line 240: I can't think of any concrete reason why it would cause problems, but I'd feel more comfortable if these were initialized.

lib/DxilRootSignature/DxilRootSignature.cpp
```diff
@@ -190,6 +190,44 @@ void DeleteRootSignature(const DxilVersionedRootSignatureDesc * pRootSignature)
   delete pRootSignature;
 }
 
+template <typename RootSigDesc, typename RootParameterDesc>
+void ExtendRootSig(RootSigDesc &rootSigDesc,
```
pow2clk, line 194: Looks familiar. Was this meant to be consolidated?

lib/DxilRootSignature/DxilRootSignature.cpp
```diff
@@ -190,6 +190,44 @@ void DeleteRootSignature(const DxilVersionedRootSignatureDesc * pRootSignature)
   delete pRootSignature;
 }
 
+template <typename RootSigDesc, typename RootParameterDesc>
+void ExtendRootSig(RootSigDesc &rootSigDesc, 
+                   uint32_t ShaderRegister,
+                   uint32_t RegisterSpace) {
+  auto *existingParams = rootSigDesc.pParameters;
+  auto *newParams = new RootParameterDesc[rootSigDesc.NumParameters + 1];
+  memcpy(newParams, existingParams, rootSigDesc.NumParameters * sizeof(RootParameterDesc));
```
pow2clk, line 199: as above, I think this should be in the conditional.

lib/DxilRootSignature/DxilRootSignature.cpp
```diff
@@ -190,6 +190,44 @@ void DeleteRootSignature(const DxilVersionedRootSignatureDesc * pRootSignature)
...
+    ExtendRootSig<DxilRootSignatureDesc, DxilRootParameter>(rs->Desc_1_0,
+                                                            ShaderRegister, RegisterSpace);
+    break;
+  case DxilRootSignatureVersion::Version_1_1:
+    ExtendRootSig<DxilRootSignatureDesc1, DxilRootParameter1>(rs->Desc_1_1,
+                                                            ShaderRegister, RegisterSpace);
+    rs->Desc_1_1.pParameters[rs->Desc_1_1.NumParameters - 1].Descriptor.Flags =
+        hlsl::DxilRootDescriptorFlags::None;
+    break;
+  }
```
pow2clk, line 227: If this can return rs, then I think the call to AddUAVParamterToRootSignature() can be replaces with:

```c++
auto rs = AddDescriptorParamter();
auto extendedRootSig = SerializeRootSignatureToVector(rs);
```

tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -195,10 +200,6 @@ class PixTest {
   TEST_METHOD(DiaCompileArgs)
   TEST_METHOD(PixDebugCompileInfo)
 
-  TEST_METHOD(CheckSATPassFor66_NoDynamicAccess)
-  TEST_METHOD(CheckSATPassFor66_DynamicFromRootSig)
-  TEST_METHOD(CheckSATPassFor66_DynamicFromHeap)
```
pow2clk, line 200: Curious why these had to go

pow2clk (commented):
Mostly just clarifications. I don't see anything wrong, but I'd like to know more about the memleak

tools/clang/unittests/HLSL/PixTest.cpp
jeffnn: Oh i should have dropped a comment: PIX doesn't use this functionality anymore, so I deleted these tests.

include/dxc/DxilRootSignature/DxilRootSignature.h
```diff
@@ -360,6 +360,9 @@ class RootSignatureHandle {
 
 void DeleteRootSignature(const DxilVersionedRootSignatureDesc *pRootSignature);  
 
+void AddDescriptorParamter(const DxilVersionedRootSignatureDesc *pRootSignature,
```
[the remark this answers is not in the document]
jeffnn: It is indeed

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -159,7 +165,97 @@ llvm::CallInst *CreateHandleForResource(hlsl::DxilModule &DM,
...
+  }
+  rootSigDesc.pParameters = newParams;
+  rootSigDesc.pParameters[rootSigDesc.NumParameters].ParameterType = DxilRootParameterType::UAV;
+  rootSigDesc.pParameters[rootSigDesc.NumParameters].Descriptor.RegisterSpace = toolsRegisterSpace;
+  rootSigDesc.pParameters[rootSigDesc.NumParameters].Descriptor.ShaderRegister = toolsUAVRegister;
+  rootSigDesc.pParameters[rootSigDesc.NumParameters].ShaderVisibility = DxilShaderVisibility::All;
+  rootSigDesc.NumParameters++;
+} 
+
+static std::vector<uint8_t> AddUAVParamterToRootSignature(const void *Data,
```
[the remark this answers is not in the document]
jeffnn: Strange. There must be something about how I type that sequence of characters that makes me miss the first e. Weerd.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -159,7 +165,97 @@ llvm::CallInst *CreateHandleForResource(hlsl::DxilModule &DM,
...
+constexpr uint32_t toolsRegisterSpace = static_cast<uint32_t>(-2);
+constexpr uint32_t toolsUAVRegister = 0;
+
+template<typename RootSigDesc, typename RootParameterDesc>
+void ExtendRootSig(RootSigDesc &rootSigDesc) {
+  auto *existingParams = rootSigDesc.pParameters;
+  auto *newParams = new RootParameterDesc[rootSigDesc.NumParameters + 1];
+  memcpy(newParams, existingParams,
+         rootSigDesc.NumParameters * sizeof(RootParameterDesc));
+  if (existingParams != nullptr) {
```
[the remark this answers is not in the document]
jeffnn: Ooh good catch.

lib/DxilRootSignature/DxilRootSignature.cpp
jeffnn: Sadly, this duplication was a down-side of keeping all the new/delete in the same place...

lib/DxilRootSignature/DxilRootSignature.cpp
jeffnn: Yep

lib/DxilRootSignature/DxilRootSignature.cpp
jeffnn: DxilRootSingature.cpp's public surface uses "const" in all its DxilVersionedRootSignatureDesc pointers. Presumably it's doing this to enforce the contract that nobody outside of DxilRootSignature.cpp owns the memory allocation, nor is allowed to modify the structure. Returning rs (the non-const) would mess with this convention.

tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -1665,81 +1670,35 @@ TEST_F(PixTest, PixDebugCompileInfo) {
...
-Texture1D<float4> tex[5] : register(t3);
-SamplerState SS[3] : register(s2);
-
-[RootSignature("DescriptorTable(SRV(t3, numDescriptors=5)),\
-                DescriptorTable(Sampler(s2, numDescriptors=3))")]
-float4 main(int i : A, float j : B) : SV_TARGET
-{
-  float4 r = tex[i].Sample(SS[i], i);
-  return r;
-}
```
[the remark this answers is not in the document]
jeffnn: I'm sorry I don't follow what you mean: what has an equivalent? The test case?

lib/DxilRootSignature/DxilRootSignatureSerializer.cpp
```diff
@@ -332,93 +332,6 @@ void SerializeRootSignature(const DxilVersionedRootSignatureDesc *pRootSignature
...
-    }
-    return m_pRootSignature11;
-
-  default:
-    IFTBOOL(false, E_FAIL);
-  }
-
-  return nullptr;
-}
-
```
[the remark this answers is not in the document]
jeffnn: Yes, there was a leak in here. There's another fixed leak in DxilValidation.cpp that is actually used.

lib/HLSL/DxilValidation.cpp
```diff
@@ -5861,6 +5861,7 @@ HRESULT ValidateDxilBitcode(
                                                pOutputStream->GetPtr(), pWriter->size(),
                                                DiagStream), DXC_E_INCORRECT_ROOT_SIGNATURE);
     } catch (...) {
+      DeleteRootSignature(pDesc);
       return DXC_E_INCORRECT_ROOT_SIGNATURE;
     }
   }
```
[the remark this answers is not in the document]
jeffnn: Face-palm. Thanks.

lib/HLSL/DxilValidation.cpp
```diff
@@ -5850,8 +5850,8 @@ HRESULT ValidateDxilBitcode(
     IFT(CreateMemoryStream(DxcGetThreadMallocNoRef(), &pOutputStream));
     pOutputStream->Reserve(pWriter->size());
     pWriter->write(pOutputStream);
+    const DxilVersionedRootSignatureDesc* pDesc = nullptr;
```
tex3d, line 5853: Could we create a small object wrapper for this which owns (and automatically frees) the memory, then has a `const DxilVersionedRootSignatureDesc*` cast operator, a `*` operator, and `get()` method (all returning this type)?  That would make it more reliable to work with and avoid memory leaks, right?

tex3d: And, perhaps a non-const `get()` or `get_mutable()` could return the non-const pointer, allowing you to modify it without using const_cast (as long as you promise to be careful with memory management 😉).

jeffnn: Is this stuff visible to the driver? Is that why it's binary-stable? Or is it just because it's trying to echo the API/DDI structures? If it's the latter, is there any reason not to replace the bare arrays with std::vector?

jeffnn: Not so sure about the suggestion of allowing the callers to get non-const access to the bare arrays, but it's your guys' call. It would allow me to isolate the code that modifies the arrays inside the PIX passes.

jeffnn: Given that Tex+Greg advocate doing the memory allocation outside of this file (and the attendant de-duplication of some of this new code), I've done so. And added the wrapper Tex suggested.

include/dxc/DxilRootSignature/DxilRootSignature.h
```diff
@@ -388,6 +385,30 @@ bool VerifyRootSignature(_In_ const DxilVersionedRootSignatureDesc *pDesc,
...
+    return m_pRootSignature;
+  }
+  const DxilVersionedRootSignatureDesc ** get_address_of() {
+    return &m_pRootSignature;
+  }
+  const DxilVersionedRootSignatureDesc* get() const { 
+      return m_pRootSignature;
+  }
+  DxilVersionedRootSignatureDesc* get_mutable() { 
+      return const_cast<DxilVersionedRootSignatureDesc *>(m_pRootSignature);
```
tex3d, line 409: I think it would be preferable for the private member variable to be non-const, then avoid the const_cast here.

lib/DxilRootSignature/DxilRootSignature.cpp
jeffnn: As noted below, I'll bow to the consensus and return a non-const pointer to PIX. THis helps clean this up a bit.

lib/DxilRootSignature/DxilRootSignature.cpp
jeffnn: With the change to allocate memory outside of DxilRootSignature.cpp, I can de-dupe this.

include/dxc/DxilRootSignature/DxilRootSignature.h
jeffnn: Your call, of course. The const_cast either has to be here or in get_address_of and the constructor. That is, if we want to be able to use one of these wrappers with DeserializeRootSignature.

include/dxc/DxilRootSignature/DxilRootSignature.h
```diff
@@ -388,6 +385,30 @@ bool VerifyRootSignature(_In_ const DxilVersionedRootSignatureDesc *pDesc,
...
+  ScopedVersionedRootSignature(
+      const DxilVersionedRootSignatureDesc *pRootSignature)
+  : m_pRootSignature(pRootSignature){}
+  ~ScopedVersionedRootSignature() { 
+      DeleteRootSignature(m_pRootSignature);
+  }
+  const DxilVersionedRootSignatureDesc* operator -> () const {
+    return m_pRootSignature;
+  }
+  const DxilVersionedRootSignatureDesc ** get_address_of() {
```
tex3d, line 402: Can this use the `&` operator?  Or is that now considered bad practice?

Also, could the functions take the new `ScopedVersionedRootSignature` instead of `const DxilVersionedRootSignatureDesc**`?

And, could `ScopedVersionedRootSignature` just be `DxilVersionedRootSignature`?

include/dxc/DxilRootSignature/DxilRootSignature.h
```diff
@@ -388,6 +385,30 @@ bool VerifyRootSignature(_In_ const DxilVersionedRootSignatureDesc *pDesc,
                          _In_ llvm::raw_ostream &DiagStream,
                          _In_ bool bAllowReservedRegisterSpace);
 
+class ScopedVersionedRootSignature {
+  const DxilVersionedRootSignatureDesc *m_pRootSignature;
+
+public:
+  ScopedVersionedRootSignature() : m_pRootSignature(nullptr) {}
+  ScopedVersionedRootSignature(
```
tex3d, line 393: Could have a constructor that takes `ScopedVersionedRootSignature&&` to move ownership, and delete ones that allow copy, since this should act more like `unique_ptr`.

If interfaces that init/modify root sig take the object instead of the ptr, then access could be further protected.

include/dxc/DxilRootSignature/DxilRootSignature.h
jeffnn: I was following the example of ComPtr, which intentionally doesn't allow & access. I presume that's to make it a bit more explicit and so harder to accidentally take the address of and then overwrite a non-zero pointer.

jeffnn: Yes, those functions could be converted but that's a bigger-scope change than I was hoping for here.

include/dxc/DxilRootSignature/DxilRootSignature.h
jeffnn: Good point about approaching unique_ptr semantics

include/dxc/DxilRootSignature/DxilRootSignature.h
tex3d: Re: `&` operator:  Strangely, `CComPtr` supports it, we use that all over the place, and it does throw (or assert?) if the pointer is not null.

On that note, we should probably fail somehow if the pointer is not null here.  Perhaps assert and return nullptr (for NDEBUG build)?  That way we don't silently leak if it's misused.

jeffnn: Should that be a DXASSERT or is a different macro more appropriate in this header?

jeffnn: (I did think of adding one but wasn't sure what macro to use :-)) )

tex3d (approved):
LGTM.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -159,7 +165,97 @@ llvm::CallInst *CreateHandleForResource(hlsl::DxilModule &DM,
...
+  if (subObjects != nullptr) {
+    for (auto const &subObject : subObjects->GetSubobjects()) {
+      if (subObject.second->GetKind() ==
+          DXIL::SubobjectKind::GlobalRootSignature) {
+        const char *Text;
+        const void *Data;
+        uint32_t Size;
+        constexpr bool notALocalRS = false;
+        if (subObject.second->GetRootSignature(notALocalRS, Data, Size,
+                                               &Text)) {
```
tex3d, line 240: Since you don't use Text, it's ok to just pass nullptr there, when you only expect the binary root signature.  The text would only be set for parsing in front end, and won't be available later.

jeffnn: Don. Thanks.

include/dxc/DxilRootSignature/DxilRootSignature.h
jeffnn: Added not-null check

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: Ach missed this comment. Fixed.

PR #4877, opened 2022-12-13
Clean up and rework DXIL library depencencies

llvm-beanz:
This change reworks the dependency specifications for the DXIL* libraries. Many of these libraries had over and under specified dependencies.

This also collapses the DxilRDATBuidlder into the DxilContainer library to remove the cyclic dependency between the two libraries.


adam-yang (approved):

lib/Analysis/Analysis.cpp
```diff
@@ -69,6 +69,18 @@ void llvm::initializeAnalysis(PassRegistry &Registry) {
...
   initializeScopedNoAliasAAPass(Registry);
+
+// HLSL Change Begin - Windows doesn't support __attribute__((used)) so these methods
+// need to be forcibly bound or they could be stripped at build time.
+#if defined(_MSC_VER) && (!defined(NDEBUG) || defined(LLVM_ENABLE_DUMP))
+#pragma optimize("", off)
+  // Pin LLVM dump methods.
+  void (__thiscall Module::*pfnModuleDump)() const = &Module::dump;
+  void (__thiscall Type::*pfnTypeDump)() const = &Type::dump;
+  void (__thiscall Function::*pfnViewCFGOnly)() const = &Function::viewCFGOnly;
```
tex3d, line 80: I think it's intended that the dump functions can be used for debugging in targets that don't link to Analysis (such as DxilConv), though this `Function::viewCFGOnly` dependency should have been problematic in that case.  I wonder if that means we should have the `Function::viewCFGOnly` here and the other two dump method refs kept in DxilModule so the appropriate ones are available with subsets of dependencies?

llvm-beanz: DxilConv does depend on Analysis transitively. It links ScalarOps, which in turn links Analysis.

DxbcConverter depends on DxilConvPasses, which should also depend on ScalarOps, that should also be fine.

I don't think we have any binaries that don't depend on analysis. The issue this likely solves is that MSVC has no equivalent of `__attribute__((used))` to mark that a function that is never used shouldn't be stripped from the final binary.

This should address that case for all our binaries without introducing any additional dependencies. I'm not opposed to moving this somewhere else, but I'm unsure we need to and I'd prefer keeping the code all grouped since it all serves the same purpose.

PR #4878, opened 2022-12-14
Update all the HLSL tests that rely on HLSL 2018

llvm-beanz:
This updates all the tests cases that rely on HLSL 2018 behaviors that will break if the default HLSL version updates to 2021.

These tests are updated to be passed the HLSL 2018 version so they continue testing and working under the same behaviors.


python3kgae (approved):

pow2clk (approved):
This represents 4% of the test shaders we have. It's a big chunk, but I think it's acceptable to declare these as 2018 specific until we have a reason to update them.

PR #4879, opened 2022-12-14
hcttestcmds: Fix error handling and support skip cleanup on failure

tex3d:
Fix problem with expression under :failed subroutine, and preserve Failed variable on endlocal in :cleanup subroutine.

Use environment variable HLSL_TESTCMD_CLEANUP_ON_FAILURE to drive whether hcttestcmd cleans up temporary files on failure. Defaults to 1 (true) unless it's already set.
You can set it to 0 to disable cleanup when the test fails.

Fix originally intended behavior for -Vi test and check files for cleanup.

### DIFF
--- a/utils/hct/hcttestcmds.cmd
+++ b/utils/hct/hcttestcmds.cmd
@@ -16,6 +16,8 @@ set FailingCmdWritten=0
 set OutputLog=%1\testcmd.log
 set LogOutput=1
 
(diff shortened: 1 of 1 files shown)


pow2clk (approved):
Looks good! I wonder if there's anything about how we invoke these that might have caught the invalid operator 😞 Hopefully we can replace these soon anyway

PR #4882, opened 2022-12-14
Enable 2021 operator intrinsics in older modes

llvm-beanz:
This enables the `select`, `and`, and `or` intrinsics in older HLSL language modes to ease transitioning to HLSL 2021.


pow2clk (approved):
I lament that we have so few tests and that the only DXIL-specific one entangles it a bit with templates, but that's where we are. I already have a note to write more such tests.

PR #4883, opened 2022-12-14
Update all the HLSL tests that rely on HLSL 2018 (#4878)

llvm-beanz:
This updates all the tests cases that rely on HLSL 2018 behaviors that will break if the default HLSL version updates to 2021.

These tests are updated to be passed the HLSL 2018 version so they continue testing and working under the same behaviors.


hekota (approved):
LGTM!

llvm-beanz:
Since we've decided not to enable HLSL 2021 by default in this release I'm going to abandon this change.

PR #4886, opened 2022-12-15
Allow select() to work with objects

### LINKED ISSUE #4673: Cannot use `select` intrinsic with SamplerState

pow2clk:
Add any_object builtin parameter type. This will match any object type as the select() built-in added with HLSL 2021 requires.

Adds a new built-in overload that takes the any_object type and ensures that the last two arguments and return type match. Many of the changes were required because we never had to ensure object parameter types matched nor transmit that type to the return type before. Most of the changes to MatchArguments in SemaHLSL are addressing new issues this revealed. Non-basic types can pass the conversion check if they outright match. There is a final fixup to the return type for textures so the actual template type is used instead of just Texture2D<element>

Adds extensive testing to both select() and the ternary operator that it supplements for these usages

Fixes #4673


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6421,8 +6466,9 @@ bool HLSLExternalSource::MatchArguments(
       }
       pNewType = objectElement;
     }
-    else if (pArgument->uLegalComponentTypes == LICOMPTYPE_TEXTURE2D
-      || pArgument->uLegalComponentTypes == LICOMPTYPE_TEXTURE2DARRAY) {
+    else if (i != 0 && Template[pArgument->uTemplateId] == AR_TOBJ_OBJECT) {
```
pow2clk, line 6469: Just a potentially useful historical note: This was originally written for sampler feedback, which were the only previous built-in functions that took objects. They all have the same solution, so this just broadens it out.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -1101,6 +1101,50 @@ static const ArBasicKind g_AnyCT[] =
...
+  AR_OBJECT_TEXTURE1D_ARRAY,
+  AR_OBJECT_TEXTURE2D,
+  AR_OBJECT_TEXTURE2D_ARRAY,
+  AR_OBJECT_TEXTURE3D,
+  AR_OBJECT_TEXTURE2DMS,
+  AR_OBJECT_TEXTURE2DMS_ARRAY,
+  AR_OBJECT_TEXTURECUBE,
+  AR_OBJECT_TEXTURECUBE_ARRAY,
+  AR_OBJECT_STRING_LITERAL,
+  AR_OBJECT_STRING,
```
tex3d, line 1118: String and String literal are only supported in special cases for constant initializer of subobjects.  I don't think they should be included here.

llvm-beanz:
Does this work in template dependent contexts? Something like:

``` c++
template<typename SamplerTy>
SamplerTy SelWrapper(bool C, SamplerTy First, SamplerTy Second) {
  return select(C, First, Second);
}
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6504,6 +6550,11 @@ bool HLSLExternalSource::MatchArguments(
              "have as many arguments and types as the intrinsic template");
   }
 
+  // For object return types that need to match arguments, we need to slot in the full type here
+  // Can't do it sooner because when return is encountered above, the other arg types haven't been set
+  if (Template[pIntrinsic->pArgs[0].uTemplateId] == AR_TOBJ_OBJECT)
+    argTypes[0] = argTypes[pIntrinsic->pArgs[0].uComponentTypeId];
```
tex3d, line 6556: Man, this function's code is scary/tricky.  I thought this shouldn't be necessary at first, but it seems like this is necessary specifically for object types in this case, since it would normally have constructed a type if it was a basic type (which you can't do for the specific object), or taken it from `objectElement` or `functionTemplateTypeArg`.  This is yet another case.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -2156,16 +2201,16 @@ static bool CombineBasicTypes(ArBasicKind LeftKind,
...
-    (RightKind < 0 || RightKind >= AR_BASIC_COUNT)) {
-    return false;
-  }
-
   if (LeftKind == RightKind) {
     *pOutKind = LeftKind;
     return true;
   }
 
+  if ((LeftKind < 0 || LeftKind >= AR_BASIC_COUNT) ||
```
tex3d, line 2209: I think the point here was that this function should only be used to combine "basic" types, not object types.  If we are using this for object types, I think there could be a problem.

For instance, I fear there may not be sufficient checks to ensure that the object types of the two inputs are identical, when they are of the same `ArBasicKind`.  For instance, if you pass a `StructuredBuffer<Foo>` and `StructuredBuffer<Bar>`, will it correctly identify the mismatch?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6421,8 +6466,9 @@ bool HLSLExternalSource::MatchArguments(
       }
       pNewType = objectElement;
     }
-    else if (pArgument->uLegalComponentTypes == LICOMPTYPE_TEXTURE2D
-      || pArgument->uLegalComponentTypes == LICOMPTYPE_TEXTURE2DARRAY) {
+    else if (i != 0 && Template[pArgument->uTemplateId] == AR_TOBJ_OBJECT) {
+      // For object parameters, just use the argument type
+      // Return type is assigned below
```
tex3d, line 6471: I don't think this code will correctly initialize the object type for second matching argument.  Instead, it's likely to just use the input argument type, which will match the input argument of course, but won't be able to detect when two arguments don't exactly match when they are the same `ArBasicKind`.  I made a related comment above regarding the change in CombineBasicTypes.

In order to match, I think we have to set the type to the type of the target argument at index `pArgument->uTemplateId` when this the target argument isn't this one.  Also, assert that the target argument is always `<=` this one (which it is in this case).

PR #4889, opened 2022-12-16
Enable building with ninja through Visual Studio

llvm-beanz:
This change enables using Visual Studio's built-in CMake support for building DXC using Ninja under Visual Studio. To accomplish that this change fixes seveal issues around the TAEF lit test integraiton when using ninja.

The first fix is to alter how the `build_mode` variable was processed. The `build_mode` variable needs to be fed from the build system into LIT at both configuration and runtime so that it can be specialized for multi-configuration generators like Visual Studio or given a default value for generators like Ninja.

The second fix is to extend the FindTAEF module to identify the taef binary. This allows our build configuraiton to find taef even from a shell environment not invoked by the hctstart environment.

Last this change includes a CMakeSettings.json file which configures Visual Studio. With this file in place a user can open the DXC checkout using Visual Studio's "Open Folder" feature and Visual Studio will handle configuring the build from there.

When using Visual Studio the only testing that is exposed is the LIT-based testing. You can run the lit tests by building the `check-*` targets.

Any tests which haven't been migrated to lit (like the hcttestcmd tests) will not be run during the Visual Studio test targets.


dmpots (commented):
Nice!

The changes look fine to me. It would be great to update the "getting started" instructions (if we have any) to point out that you can now just clone and point VS to the source directory (and also that you need ninja for that).

llvm-beanz:
>It would be great to update the "getting started" instructions (if we have any) to point out that you can now just clone and point VS to the source directory (and also that you need ninja for that).

I was a little hesitant to update the build instructions yet because this workflow has some testing gaps. Once we get those resolved I'd like to do a bigger update that consolidates the Windows and Linux instructions together in one place since they'll be very similar.

Thoughts?

dmpots:
> > It would be great to update the "getting started" instructions (if we have any) to point out that you can now just clone and point VS to the source directory (and also that you need ninja for that).
> 
> I was a little hesitant to update the build instructions yet because this workflow has some testing gaps. Once we get those resolved I'd like to do a bigger update that consolidates the Windows and Linux instructions together in one place since they'll be very similar.
> 
> Thoughts?

That's a fair point.

I think it would be ok to just point out these testing gaps if you want to update it now. The PR will always go through full testing via the automated PR builds so I think the risk for us is pretty low.

Also seems reasonable to just wait and update later.

llvm-beanz:
I put a new doc up in #4890. I don't yet want to replace the existing docs, but maybe this is a good middle ground?

cmake/modules/FindTAEF.cmake
```diff
@@ -41,20 +41,21 @@ set(TAEF_INCLUDE_DIRS ${TAEF_INCLUDE_DIR})
...
-elseif(EXISTS "${TAEF_NUGET_BIN}/x64/te.exe" AND EXISTS "${TAEF_NUGET_BIN}/x86/te.exe")
-  set(TAEF_BIN_DIR "${TAEF_NUGET_BIN}")
-elseif(EXISTS "${TAEF_SDK_BIN}/x64/te.exe" AND EXISTS "${TAEF_SDK_BIN}/x86/te.exe")
-  set(TAEF_BIN_DIR "${TAEF_SDK_BIN}")
-elseif(EXISTS "${WINDOWS_KIT_10_PATH}")
-  message(ERROR "Unable to find TAEF binaries under Windows 10 SDK.")
-elseif(EXISTS "${WINDOWS_KIT_81_PATH}")
-  message(ERROR "Unable to find TAEF binaries under Windows 8.1 or 10 SDK.")
+find_program(TAEF_EXECUTABLE te.exe PATHS
+  $ENV{HLSL_TAEF_DIR}/x64
```
python3kgae, line 45: When build for x86, will x86 te.exe be select?

llvm-beanz: Good catch! Will update.

llvm-beanz: In addition to fixing this, I added an x86 configuration to the CMake settings file.

We should be able to add arm, linux, and clang-cl variants to over time.

CMakeSettings.json
```diff
@@ -0,0 +1,48 @@
...
+      "cmakeCommandArgs": "-DDXC_USE_LIT=On -C ${projectDir}\\cmake\\caches\\PredefinedParams.cmake -DLLVM_ENABLE_ASSERTIONS=On -DLLVM_LIT_ARGS=\"-v\"",
+      "buildCommandArgs": "-v",
+      "ctestCommandArgs": "",
+      "inheritEnvironments": [ "msvc_x86_x64" ]
+    },
+    {
+      "name": "x86-Debug",
+      "generator": "Ninja",
+      "configurationType": "Debug",
+      "buildRoot": "${projectDir}\\out\\build\\${name}",
```
python3kgae, line 40: This will make out inside the source directory.
Could we change it to ${projectDir}\\..\\..\\out\\build\\${name} ?

cmake/modules/FindTAEF.cmake
```diff
@@ -41,20 +41,18 @@ set(TAEF_INCLUDE_DIRS ${TAEF_INCLUDE_DIR})
...
-elseif(EXISTS "${WINDOWS_KIT_10_PATH}")
-  message(ERROR "Unable to find TAEF binaries under Windows 10 SDK.")
-elseif(EXISTS "${WINDOWS_KIT_81_PATH}")
-  message(ERROR "Unable to find TAEF binaries under Windows 8.1 or 10 SDK.")
+find_program(TAEF_EXECUTABLE te.exe PATHS
+  $ENV{HLSL_TAEF_DIR}/${CMAKE_C_COMPILER_ARCHITECTURE_ID}
+  ${TAEF_NUGET_BIN}/${CMAKE_C_COMPILER_ARCHITECTURE_ID}
+  ${TAEF_SDK_BIN}/${CMAKE_C_COMPILER_ARCHITECTURE_ID}
+  ${WINDOWS_KIT_10_PATH}
+  ${WINDOWS_KIT_81_PATH}
```
python3kgae, line 49: Is there anybody still install Windows kit 81?

CMakeSettings.json
llvm-beanz: The directory gets ignored by our `.gitignore` file so it shouldn't interfere with git. Writing to an external directory runs the likelihood of conflicting with build directories created by other projects (or other clones of DXC) that share a common parent directory.

I think it is preferable to keep the build outputs nested under the source to not cause issues with user configurations that are out of our control.

cmake/modules/FindTAEF.cmake
llvm-beanz: No idea, but if we want to remove that we should do it in a separate change from this one. That was a searched directory before this change so I don't want to remove it.

PR #4890, opened 2022-12-16
[Docs] Add new docs on unified build workflow

llvm-beanz:
This document tries to capture the basics of a unified build and test workflow for all platforms. It is complete enough to start using but not yet ready to be the default build mode.


dmpots (approved):
LGTM.

Would it make sense to add a link to this page somewhere? Thinking how would someone that stumbles on the project find these notes?

This workflow seems attractive for newcomers particularly on windows where you can just point VS to the cloned directory.

PR #4893, opened 2022-12-19
Add lit test for hcttestcmds.

python3kgae:
Change tests in hcttestcmds into lit shell tests.

echo.bat and del.bat is added to use echo and del thru subprocess.Popen.


tools/clang/test/CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ configure_lit_site_cfg(
   ${CMAKE_CURRENT_BINARY_DIR}/taef/lit.site.cfg
   )
 
+configure_lit_site_cfg(
```
llvm-beanz, line 27: We should not need to add a lit site config for this.

tools/clang/test/CMakeLists.txt
```diff
@@ -111,3 +116,15 @@ if (WIN32)
           )
 endif()
 # HLSL Change End
+
+# HLSL Change Begin - Add cmds tests
+if (WIN32)
```
llvm-beanz, line 121: We shouldn't add an explicit target for this either. #4801, will add test targets for every subdirectory without needing to manually add them.

tools/clang/test/hlsl_cmds_test/NonUniform.test
```diff
@@ -0,0 +1,56 @@
...
+
+// RUN: %dxc %s /T ps_6_0 /DDX12 /Fo %t.NonUniform.cso
+// RUN: %dxc %t.NonUniform.cso /dumpbin /Qstrip_rootsignature /Fo %t.NonUniformNoRootSig.cso
+// RUN:%dxa  -listparts %t.NonUniformNoRootSig.cso | FileCheck %s --check-prefix=NO_RS
+
+// NO_RS:Part count
+// NO_RS-NOT:RTS0
+
+// RUN: %dxc  %t.NonUniform.cso /dumpbin /extractrootsignature /Fo %t.NonUniformRootSig.cso
+// RUN:%dxa  -listparts %t.NonUniformRootSig.cso | FileCheck %s --check-prefix=RS_PART
```
llvm-beanz, line 20: We don't build dxa on Linux at the moment, so we should look at the tests that use it and evaluate if it makes sense to support dxa on Linux, or if we should adjust the test or only run the test on Windows.

tools/clang/test/hlsl_cmds_test/NonUniform.test
```diff
@@ -0,0 +1,56 @@
...
+// Verify mismatched root signatures fail verification
+
+// RUN: %dxc %S/smoke.hlsl  /D "semantic = SV_Position" /T vs_6_0 /Zi /Qembed_debug /DDX12 /Fo %t.smoke.define.cso
+// RUN: %dxc %t.smoke.define.cso /dumpbin /Qstrip_rootsignature /Fo %t.norootsignature.cso
+// RUN: %dxc %t.smoke.define.cso /dumpbin /extractrootsignature /Fo %t.rootsig.cso
+
+// RUN: not %dxc %t.NonUniformNoRootSig.cso /dumpbin /verifyrootsignature %t.rootsig.cso
+// RUN: not %dxc %t.norootsignature.cso /dumpbin /verifyrootsignature %t.NonUniformRootSig.cso
+
+// RUN: %del %t.NonUniform.cso
```
llvm-beanz, line 38: lit requires the Unix shell commands provided by git be on your path, so you could use `rm` instead of a custom delete, but neither should be required since lit deletes temporary files.

tools/clang/test/hlsl_cmds_test/batch_compile.test
```diff
@@ -0,0 +1,3 @@
+// RUN: %batch -lib-link -multi-thread %S\batch_cmds2.txt
```
llvm-beanz, line 1: `dxc_batch` is also Windows-only, so we should evaluate porting it. It also likely makes sense to pull dxc_batch tests into their own subdirectory.

tools/clang/test/hlsl_cmds_test/batch_cmds.txt
```diff
@@ -0,0 +1,5 @@
+-E ps_main -T ps_6_0     lib_entries2.hlsl
```
llvm-beanz, line 1: It is helpful to take files that are used as the inputs to tests rather than tests themselves and move them into an `Inputs` subdirectory.

tools/clang/test/hlsl_cmds_test/dump_dependency.test
```diff
@@ -0,0 +1,24 @@
+// RUN: %dxc /T ps_6_0 %s /M | FileCheck %s
+
+// RUN: %dxc /T ps_6_0 %s /MFdeps | FileCheck --input-file=%S/deps %s
```
llvm-beanz, line 3: In lit `%T` resolves to a temporary directory and `%t` to a temporary file path. Both get cleaned up by lit.

tools/clang/test/hlsl_cmds_test/lit.cfg
```diff
@@ -0,0 +1,91 @@
...
+else:
+    config.available_features.add("nospirv")
+
+
+bin_dir = os.path.join(config.llvm_obj_root, lit_config.params['build_mode'], 'bin')
+dxc_bin = os.path.join(bin_dir, 'dxc.exe')
+
+config.environment['PATH'] = str.format('{};{}',config.environment['PATH'], bin_dir)
+
+config.substitutions.append( ('%dxc',
```
llvm-beanz, line 39: It makes sense to add these substitutions to the top-level lit config (dxc is already there).

tools/clang/test/hlsl_cmds_test/lit.cfg
```diff
@@ -0,0 +1,91 @@
...
+    # configuration hasn't been created by the build system, or we are in an
+    # out-of-tree build situation).
+
+    # Check for 'clang_unit_site_config' user parameter, and use that if available.
+    site_cfg = lit_config.params.get('clang_hlsl_cmds_site_config', None)
+    if site_cfg and os.path.exists(site_cfg):
+        lit_config.load_config(config, site_cfg)
+        raise SystemExit
+
+if config.spirv:
```
llvm-beanz, line 28: Features for spirv should be added to the top-level lit config, not a suite-specific one.

tools/clang/test/hlsl_cmds_test/test_rs.test
```diff
@@ -0,0 +1,18 @@
+// Root Signature target
+// RUN:%echo #define main "CBV(b0)"> %t.test-global-rs.hlsl
```
llvm-beanz, line 2: Why not just have this in a file that serves as the test file?

llvm-beanz (changes requested):
There is a `DXC` lit suite for test that check the DXC command interface. That makes sense for where these tests could be added since they are testing the DXC interface.

Overall I think the test suite is a bit of a mess to look at. Mixing input files and test files in the same directory makes it a bit hard to follow, but also some of the tests have some unnecessary complexity.

Check the file attributes on the utf-16 files you added. It looks like git thinks it is binary rather than text.

Part of the benefit (and driving reason) for moving these to lit is to get test coverage on Linux, so we should make sure as many of these tests as possible work on Linux.

tools/clang/test/hlsl_cmds_test/NonUniform.test
```diff
@@ -0,0 +1,56 @@
+// RUN: %dxc -E main -T ps_6_0 %s | FileCheck %s
```
pow2clk, line 1: Am I correct that these are all copied without alteration from the utils directory? Is there any need for the old versions? Were there any changes?

I see that some were renamed from .hlsl to .test, but some were not. Why is that?

python3kgae: I changed hlsl to test because not all hlsl files have a test use it directly.
I'll move all hlsl into Input directory and keep the test as only lit command as Chris suggested.

We don't need the old version once the lit version is enabled by default.
There should no change on the test itself.

tools/clang/test/hlsl_cmds_test/test_rs.test
python3kgae: Was trying to keep what the original test doing.
Will just create the file.

tools/clang/test/hlsl_cmds_test/NonUniform.test
llvm-beanz: If the files are actual HLSL source we shouldn't be renaming them to `.test`

python3kgae:
Replaced with https://github.com/microsoft/DirectXShaderCompiler/pull/4921

PR #4895, opened 2022-12-20
Enable build for dxa/dxl/dxopt/dxr/dxv on linux

python3kgae:
Only enabled build under linux.
Tests will be enabled by 
https://github.com/microsoft/DirectXShaderCompiler/pull/4893

A helper class WArgV was added to convert argv to wchar.


lib/DxcSupport/WinAdapter.cpp
```diff
@@ -75,4 +76,20 @@ CHandle::CHandle(HANDLE h) { m_h = h; }
 CHandle::~CHandle() { CloseHandle(m_h); }
 CHandle::operator HANDLE() const throw() { return m_h; }
 
+//===--------------------------- WArgV -------------------------------===//
+WArgV::WArgV(int argc, const char **argv)
```
llvm-beanz, line 80: It seems like we could have some unit tests for this code.

tools/clang/tools/dxa/dxa.cpp
```diff
@@ -282,8 +281,13 @@ bool DxaContext::ExtractPart(const char *pName) {
 
   WriteBlobToFile(pContent, StringRefWide(OutputFilename),
                   DXC_CP_UTF8); // TODO: Support DefaultTextCodePage
+#ifdef _WIN32
   printf("%Iu bytes written to %s\n", pContent->GetBufferSize(),
          OutputFilename.c_str());
+#else
+  printf("%zu bytes written to %s\n", pContent->GetBufferSize(),
```
llvm-beanz, line 288: `I` as a length specifier isn't part of the C specification, but `z` is. As a result `z` should be supported on Windows too, and I think the Windows docs indicate that.

tools/clang/tools/dxopt/dxopt.cpp
```diff
@@ -62,25 +63,27 @@ bool isFileInputArg(LPCWSTR arg) {
 }
 
 static HRESULT ReadStdin(std::string &input) {
-  HANDLE hStdIn = GetStdHandle(STD_INPUT_HANDLE);
```
llvm-beanz, line 65: No need to change for this PR, but it looks like we could dramatically simplify this tool by using the LLVM filesystem interfaces.

tools/clang/tools/dxr/dxr.cpp
```diff
@@ -195,19 +121,26 @@ int __cdecl wmain(int argc, const wchar_t **argv_) {
...
     IFT(dxcSupport.CreateInstance(CLSID_DxcRewriter, &pRewriter));
+#ifdef _WIN32
     IFT(pRewriter->RewriteWithOptions(pSource, wName.c_str(),
                                       argv_, argc,
                                       nullptr, 0, pIncludeHandler,
                                       &pRewriteResult));
-
+#else
+    // Convert argv to wchar.
+    WArgV ArgV(argc, argv_);
```
llvm-beanz, line 139: Does it make sense to have a passthrough mode so that the code using it doesn't need to be under ifdefs?

tools/clang/tools/dxr/dxr.cpp
```diff
@@ -195,19 +121,26 @@ int __cdecl wmain(int argc, const wchar_t **argv_) {
     CComPtr<IDxcOperationResult> pRewriteResult;
     CComPtr<IDxcBlobEncoding> pSource;
     std::wstring wName(CA2W(dxcOpts.InputFile.empty()? "" : dxcOpts.InputFile.data()));
-    if (!dxcOpts.InputFile.empty()) {
-      IFT_Data(FileMapDxcBlobEncoding::CreateForFile(wName.c_str(), &pSource), wName.c_str());
-    }
+    if (!dxcOpts.InputFile.empty())
+      ReadFileIntoBlob(dxcSupport, wName.c_str(), &pSource);
```
llvm-beanz, line 125: This looks like a great simplification of the code!

tools/clang/unittests/Basic/CMakeLists.txt
```diff
@@ -8,6 +9,7 @@ add_clang_unittest(BasicTests
   FileManagerTest.cpp
   SourceManagerTest.cpp
   VirtualFileSystemTest.cpp
+  WinAdapterTest.cpp
```
llvm-beanz, line 12: Clang/Basic is an odd place to put this. We should probably have a unit test binary at the LLVM level for DxcSupport rather than put this under Clang.

tools/clang/test/DXC/dxa_tests.hlsl
```diff
@@ -0,0 +1,36 @@
+
```
llvm-beanz, line 1: Probably worth giving this a `.test` extension instead of `.hlsl` and ensuring that `.test` is supported in the suite. This isn't actually a source file.

tools/clang/test/lit.cfg
```diff
@@ -269,6 +269,41 @@ config.substitutions.append( ('%itanium_abi_triple', makeItaniumABITriple(config
...
+config.substitutions.append( ('%batch',
+                            'dxc_batch' ))
+
+config.substitutions.append( ('%dxv',
+                            'dxv' ))
+
+config.substitutions.append( ('%dxa',
+                            'dxa') )
+
+if platform.system() not in ['Windows']:
```
llvm-beanz, line 288: LLVM's lit tests require that the git command line tools be on your path for Windows. It seems reasonable to adopt that requirement for DXC so that we don't need to do things like this.

tools/clang/test/lit.cfg
```diff
@@ -269,6 +269,41 @@ config.substitutions.append( ('%itanium_abi_triple', makeItaniumABITriple(config
...
+    config.substitutions.append( ('%echo',
+                                    os.path.join(cur_dir, "echo.sh") ))
+else:
+    config.substitutions.append( ('%echo',
+                                    os.path.join(cur_dir, "echo.bat") ))
+
+config.substitutions.append( ('%opt',
```
llvm-beanz, line 295: This substitution concerns me. We probably should use `dxopt` rather than `opt` to ensure we don't have confusion between `dxopt` and llvm's `opt`

tools/clang/test/lit.cfg
```diff
@@ -269,6 +269,41 @@ config.substitutions.append( ('%itanium_abi_triple', makeItaniumABITriple(config
...
 config.substitutions.append( ('%dxc', lit.util.which('dxc', llvm_tools_dir)) )
 
+bin_dir = os.path.join(config.llvm_obj_root, lit_config.params['build_mode'], 'bin')
+config.environment['PATH'] = str.format('{};{}',config.environment['PATH'], bin_dir)
+
+
+# Displaying the parent directory of the script
+cur_dir = os.path.dirname(__file__)
+
+config.substitutions.append( ('%batch',
```
llvm-beanz, line 279: Why not `dxc_batch`? That matches the tool name.

tools/clang/test/lit.cfg
```diff
@@ -269,6 +269,41 @@ config.substitutions.append( ('%itanium_abi_triple', makeItaniumABITriple(config
...
+                            'dxc_batch' ))
```
llvm-beanz, line 280: Using `lit.util.which` allows you to leverage lit finding the tool. That might be a more resilient approach.

tools/clang/test/lit.cfg
```diff
@@ -475,6 +510,12 @@ if config.enable_backtrace == "1":
 if lit_config.params.get("run_long_tests", None) == "true":
     config.available_features.add("long_tests")
 
+# Check if spirv is enabled.
+if config.spirv:
+    config.available_features.add("spirv")
+else:
+    config.available_features.add("nospirv")
```
llvm-beanz, line 517: I don't see where we're using the feature here, but I'm also not sure we need an inverse of the feature. Does `UNSUPPORTED` not cover the use case?

python3kgae: Removed.

tools/clang/test/lit.cfg
python3kgae: Removed the bat file.

lib/DxcSupport/WinAdapter.cpp
```diff
@@ -75,4 +77,49 @@ CHandle::CHandle(HANDLE h) { m_h = h; }
...
+// CComBSTR
+CComBSTR::CComBSTR(_In_ int nSize, LPCWSTR sz) : m_needFree(true) {
+  if (nSize < 0) {
+    throw  std::invalid_argument("CComBSTR must have size >= 0");
+  }
+
+  if (nSize == 0) {
+    m_str = NULL;
+  } else {
+    m_str = (BSTR)malloc((nSize + 1) * sizeof(WCHAR));
```
tex3d, line 89: This does not look correct to me.  A BSTR is supposed to start with a 4-byte prefix for the size of the string, then the string pointer points to the first character of the string, after the prefix.

tex3d: See:
https://learn.microsoft.com/en-us/previous-versions/windows/desktop/automat/bstr

tools/clang/tools/dxa/dxa.cpp
```diff
@@ -190,7 +189,7 @@ bool DxaContext::ExtractFile(const char *pName) {
 
   CA2W WideName(pName, CP_UTF8);
   for (UINT32 i = 0; i < uNumSources; i++) {
-    CComBSTR name;
+    BSTR name;
```
tex3d, line 192: Wouldn't this change cause name to leak?

lib/DxcSupport/WinAdapter.cpp
python3kgae: Switch to SysAllocStringLen.

tools/clang/tools/dxa/dxa.cpp
python3kgae: Fixed.

include/dxc/Support/WinAdapter.h
```diff
@@ -1039,8 +1040,62 @@ class CHandle {
...
+
+class CComBSTR
+{
+public:
+    BSTR m_str;
+    bool m_needFree;
+    CComBSTR() : m_str(nullptr), m_needFree(false) {};
+    CComBSTR(_In_ int nSize, LPCWSTR sz);
+    ~CComBSTR() throw() {
+    if (m_needFree)
```
tex3d, line 1055: Isn't this tied to whether `m_str` is non-null?  I don't think the independent variable is needed, since all BSTR strings must be freed with SysFreeString somehow.

include/dxc/Support/WinAdapter.h
```diff
@@ -1039,8 +1040,62 @@ class CHandle {
...
+
+    BSTR* operator&() throw()
+    {
+        return &m_str;
+    }
+
+    BSTR Detach() throw()
+    {
+        BSTR s = m_str;
+        m_str = NULL;
```
tex3d, line 1074: This does not reset `m_needFree`.  However, I don't see why we need `m_needFree` in the first place - see earlier comment.

include/dxc/Support/WinAdapter.h
python3kgae: Removed m_needFree.

tools/clang/tools/dxopt/dxopt.cpp
```diff
@@ -57,30 +58,32 @@ bool isStdIn(LPCWSTR fName) {
...
-  std::vector<unsigned char> buffer(1024);
-  DWORD numBytesRead = -1;
-  BOOL ok = FALSE;
-
-  // Read all data from stdin.
-  while (ok = ReadFile(hStdIn, buffer.data(), buffer.size(), &numBytesRead, NULL)) {
```
pow2clk, line 71: ReadFile is implemented in WinFunctions. I'm curious what was wrong with its use here

tools/clang/test/lit.cfg
```diff
@@ -269,6 +269,31 @@ config.substitutions.append( ('%itanium_abi_triple', makeItaniumABITriple(config
...
 config.substitutions.append( ('%dxc', lit.util.which('dxc', llvm_tools_dir)) )
 
+config.substitutions.append( ('%dxv',
+                            lit.util.which('dxv', llvm_tools_dir)) )
+
+config.substitutions.append( ('%dxa',
+                            lit.util.which('dxa', llvm_tools_dir)) )
+
+# Displaying the parent directory of the script
+config.substitutions.append( ('%echo',
```
llvm-beanz, line 279: We shouldn't need this substitution.

tools/clang/test/lit.cfg
```diff
@@ -269,6 +269,31 @@ config.substitutions.append( ('%itanium_abi_triple', makeItaniumABITriple(config
...
+                                    "echo") )
+config.substitutions.append( ('%del',
```
llvm-beanz, line 281: This one either

tools/clang/tools/dxr/dxr.cpp
pow2clk: Perhaps this suggestion was why WArgV was moved out of the non-windows block, but WinAdapter is only really intended for non-windows platforms. It shouldn't be required for the windows build. I also can't find the passthrough implementation

tools/clang/test/del.bat
```diff
@@ -0,0 +1,2 @@
+
+del %1 %2 %3 %4 %5 %6
```
llvm-beanz, line 2: We should not need this file

tools/clang/test/echo.bat
```diff
@@ -0,0 +1,2 @@
+
+echo %1 %2 %3 %4 %5 %6
```
llvm-beanz, line 2: We should not need this file

tools/clang/tools/dxclib/dxc.cpp
```diff
@@ -770,11 +769,12 @@ void DxcContext::Recompile(IDxcBlob *pSource, IDxcLibrary *pLibrary,
       NewDefines.size(), pIncludeHandler, &pResult));
   }
 
+#ifndef _WIN32
+  // FIXME: fix crash on linux when not detach pCompileSource.
```
pow2clk, line 773: Are you sure this crash still happens? It looks like a number of fixes have gone into CComBSTR since this comment was written. Might those have fixed it?

tools/clang/tools/dxcompiler/dxcpdbutils.cpp
```diff
@@ -48,7 +48,9 @@
 #include <locale>
 #include <codecvt>
 #include <string>
+#ifdef _WIN32
 #include <dia2.h>
```
pow2clk, line 52: Mild preference to group OS-conditional includes together.

include/dxc/Support/WinAdapter.h
```diff
@@ -1039,8 +1040,60 @@ class CHandle {
...
+  std::vector<std::wstring> WStringVector;
+  std::vector<const wchar_t *> WCharPtrVector;
+
+public:
+  WArgV(int argc, const char **argv);
+  WArgV(int argc, const wchar_t **argv);
+  const wchar_t **argv() { return WCharPtrVector.data();}
+};
+#endif
+
```
pow2clk, line 1098: I have looked at the places that `WArgV` is used. They convert args to wchar that are then passed either to the `MainArgs` constructor or to `RewriteWithOptions`. The `MainArgs` constructor can already handle either wchar or char arguemtns and all that `RewriteWithOptions` does is pass it into the `MainArgs` constructor.

I think this is redundant. I think we should change `RewriteWithOptions` to take a `MainArgs` object and make use of the existing code to perform this conversion of arguments. If that code can be improved, I think it should be improved there. https://github.com/microsoft/DirectXShaderCompiler/blob/60530e57ebdcc5833fce48313286f90803b05bb5/lib/DxcSupport/HLSLOptions.cpp#L168

tools/clang/test/DXC/Inputs/inc2.hlsli
```diff
@@ -0,0 +1,4 @@
+int f2(int g)
```
pow2clk, line 1: I don't like having identical copies of the same files. Can the cmdtestfiles variants not be removed? Can the scripts there redirect to this location if nothing else?

tools/clang/test/DXC/dxa_tests.test
```diff
@@ -0,0 +1,36 @@
+
```
pow2clk, line 1: I've been made to understand that `.test` shouldn't be used as a suffix for an hlsl file.  At any rate, it is inconsistently used among these files.

pow2clk (commented):
I don't think WArgV is necessary. That's the biggest point. The others are smaller.

tools/clang/test/DXC/Inputs/inc2.hlsli
pow2clk: Can you confirm that these are all duplicates of what's in `cmdtestfiles`?

tools/clang/test/DXC/dxa_tests.test
python3kgae: This is not a hlsl file. The hlsl is in %S/Inputs/smoke.hlsl. This file just contains commands and FileCheck tags.

tools/clang/test/DXC/Inputs/inc2.hlsli
python3kgae: The plan is to remove cmdtestfiles once https://github.com/microsoft/DirectXShaderCompiler/pull/4893 is done.
https://github.com/microsoft/DirectXShaderCompiler/pull/4893 depends on this PR.

python3kgae: And yes. These are just copy of cmdtestfiles.

tools/clang/tools/dxopt/dxopt.cpp
python3kgae: Cannot find GetStdHandle in WinFunctions :(

tools/clang/tools/dxclib/dxc.cpp
python3kgae: It still happens.

PR #4898, opened 2022-12-21
add test to handle preprocessing an already preprocessed file.

bob80905:
There are some captures that cause problems when preprocessing the extracted source. It appears like some of the source files have already been preprocessed. This change prevents the preprocessor from blowing up when receiving a preprocessed file as input Two tests should be added to test this behavior.
1. Make sure that when the preprocessor receives a preprocessed file, it doesn't blow up.
(this is checked by the new test, PreprocessCheckBuiltinIsOk).
2. Make sure that the preprocessor itself doesn't produce files with the line '#line 1 "<built-in>"' at the top.
(this is already satisfied by PreprocessWhenValidThenOK)

### DIFF
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -238,6 +238,7 @@ public:
   TEST_METHOD(LibGVStore)
   TEST_METHOD(PreprocessWhenExpandTokenPastingOperandThenAccept)
   TEST_METHOD(PreprocessWithDebugOptsThenOk)
+  TEST_METHOD(PreprocessCheckBuiltinIsOk)
   TEST_METHOD(WhenSigMismatchPCFunctionThenFail)
   TEST_METHOD(CompileOtherModesWithDebugOptsThenOk)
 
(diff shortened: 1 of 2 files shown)


tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -3967,6 +3968,31 @@ TEST_F(CompilerTest, PreprocessWithDebugOptsThenOk) {
...
 
+// Make sure that '#line 1 "<built-in>"' won't blow up when preprocessing.
+TEST_F(CompilerTest, PreprocessCheckBuiltinIsOk) {
+  CComPtr<IDxcCompiler> pCompiler;
+  CComPtr<IDxcOperationResult> pResult;
+  CComPtr<IDxcBlobEncoding> pSource;
+  
+  VERIFY_SUCCEEDED(CreateCompiler(&pCompiler));
+  CreateBlobFromText(
+    " \r\n\t \r\n\r\n"
```
llvm-beanz, line 3979: nit: The blank line here probably isn’t needed and you should be able to remove the `\r` characters.

llvm-beanz (approved):
Looks good. A pretty minor nit below.

PR #4900, opened 2022-12-22
Fix copy-in/copy-out parameter passing

llvm-beanz:
This change fixes HLSL's copy-in/copy-out parameter passing. Before this change we skip emitting copies for all local variable parameters, which results in code correctness issues.

With this change, we do a poor-man's alias analysis walking each local variable back to either an argument or an alloca. If it tracks back to an argument marked noalias we can skip the copy. If it tracks to an alloca, we skip the copy for the first parameter that tracks to that alloca, and emit the copies for any subsequent parameters.

This change does have one side-effect that I don't understand. It results in changing the ordering of some of the disassembly output. I suspect that is caused by something in the disassembly being order-dependent but haven't investigated fully.


tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -6018,21 +6019,25 @@ void CGMSHLSLRuntime::EmitHLSLOutParamConversionInit(
...
-        if (GlobalVariable *GV = dyn_cast_or_null<GlobalVariable>(Ptr)) {
-          bConstGlobal = m_ConstVarAnnotationMap.count(GV) | GV->isConstant();
-        }
         // Skip copy-in copy-out when safe.
         // The unsafe case will be global variable alias with parameter.
         // Then global variable is updated in the function, the parameter will
         // be updated silently. For non global variable or constant global
         // variable, it should be safe.
-        if (argAddr &&
-            (isa<AllocaInst>(Ptr) || isa<Argument>(Ptr) || bConstGlobal)) {
```
python3kgae, line 6035: Why const global is not safe?

llvm-beanz: The problem is that we don't capture the const-ness correctly in the source. If you look at the `gep_arg_nocopy.hlsl` test, if we don't copy the array in at the first call it is writable back to the global.

We could omit copying const globals in cases where the source appropriately annotates them as const, but promoting what should be a const global to a non-const pointer is incorrect code generation.

llvm-beanz: Worth noting there is a larger semantic problem that since we don't capture the constness in the AST we don't respect it with reasonable source diagnostics.

In [this example](https://godbolt.org/z/cc8En7Eh8) based on the `gep_arg_nocopy.hlsl` test, we get the right behavior in the end, but I think we only get there by doing an unsafe transformation. If you look at the `-fcgl` IR generation we're generating a store to what should be a constant global. I assume at some point in our legalization passes we drop that store.

python3kgae: We do capture constant variable inside constant buffers.
Those should be safe.

python3kgae (approved):

PR #4901, opened 2022-12-22
Improve time tracing data

llvm-beanz:
This is a bunch of small changes to improve the quality of the time traces. This mostly adds new timers breakign down dxcompilerobj and the always inliner code.


tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -719,6 +719,7 @@ class DxcCompiler : public IDxcCompiler3,
       bool validateRootSigContainer = false;
 
       if (isPreprocessing) {
+        TimeTraceScope TimeScope("PreprocessAction", StringRef(""));
```
bob80905, line 722: Should this also be prefaced with "DXCompiler::"? As above?

lib/Transforms/Utils/CloneFunction.cpp
```diff
@@ -473,6 +475,9 @@ void llvm::CloneAndPruneIntoFromInst(Function *NewFunc, const Function *OldFunc,
                                      const char *NameSuffix, 
                                      ClonedCodeInfo *CodeInfo,
                                      CloningDirector *Director) {
+  TimeTraceScope TimeScope("CloneAndPruneIntoFromInst", [&] {
```
bob80905, line 478: should we prefix the function name with the "llvm::" namespace?

llvm-beanz:

PR #4905, opened 2022-12-29
Refactor ExecutionTests into a separate dll

hekota:
ExecutionTests are currently compiled into clang-hlsl-tests.dll together with all of the other DXC unit tests. Later in the code pipeline the ExecutionTests are getting transformed into HLK tests which have different dependency constraints than the clang HLSL tests. 

Refactoring ExecutionTests into a separate dll allows us to set its dependencies and build settings accordingly and closely matching those of the HLK test dll. This will help HLK authors to implement the tests with those constraints in mind.


include/dxc/Test/HlslTestUtils.h
```diff
@@ -336,8 +371,9 @@ inline bool GetTestParamBool(LPCWSTR name) {
   if (NameValue.IsEmpty()) {
     return false;
   }
-  return Unicode::IsStarMatchWide(ParamValue, ParamValue.GetLength(),
-                                  NameValue, NameValue.GetLength());
+
+  return hlsl_test::IsStarMatchWide(ParamValue, ParamValue.GetLength(),
```
python3kgae, line 375: Why cannot use Unicode::IsStarMatchWide here?

python3kgae: Because we don't want to depend on dxcSupport?

include/dxc/Test/HlslTestUtils.h
```diff
@@ -29,7 +29,8 @@
 #include "dxc/DXIL/DxilConstants.h" // DenormMode
```
python3kgae, line 29: Remove include Unicode.h?

include/dxc/Test/HlslTestUtils.h
hekota: Yes, we want to avoid linking DXC support lib to the ExecutionTests and therefore the HLK tests.

python3kgae (approved):

PR #4908, opened 2023-01-03
build: disable LLVM_ENABLE_TERMINFO by default.

### LINKED ISSUE #4892: Linux: libdxcompiler.so dependency on libncurses5 library
Hello, when I updated the libdxcompiler.so recently, I got this error when trying to load the dynamic library:
```
libtinfo.so.5: cannot open shared object file: No such file or directory
```
`dlopen()` was used to load the libdxcompiler.so, and the error was retrieved by `dlerror()`. I'm not sure when this was introduced, because I updated a quite old version to this commit: 2c3d965b2fc734c09e76631de65f8ffbe4e3e68b

The error can be resolved by installing the `libncurses5` library, but this was not required by an older version of the compiler. 

Best regards,
Janos

Keenuts:
Having this enabled makes dxc depends on ncurses5. This is required by LLVM when supporting colored message output, but DXC doesn't supports it for now.

Fixes #4892


llvm-beanz (approved):
LGTM

Issue #4727, opened 2022-10-17
[SPIR-V] Segmentation fault when compiling to SPIR-V


KYovchevski:
At some point changes were introduced which have caused previously valid HLSL code to result in a segmentation fault when compiling into SPIR-V. It would appear this has occurred after #4535. There are no issues compiling the same code into DXIL.

Below is a minimal case for reproducing the segmentation fault.

```hlsl
[[vk::binding(0, 0)]] RaytracingAccelerationStructure tlas;

bool traceClosestHitRay(RayDesc ray) {

    const RAY_FLAG flags = RAY_FLAG_FORCE_OPAQUE | RAY_FLAG_SKIP_PROCEDURAL_PRIMITIVES;
    RayQuery<flags> q;
    q.TraceRayInline(tlas, flags, 0xff, ray);
    while (q.Proceed()) {
    }

    if (q.CommittedStatus() != COMMITTED_TRIANGLE_HIT) {
        return false;
    }
    return true;
}

// Utility function to get a vector perpendicular to an input vector
// (from "Efficient Construction of Perpendicular Vectors Without Branching", 2009)
float3 getPerpendicularVector(float3 u) {
    float3 a = abs(u);
    uint xm = ((a.x - a.y) < 0 && (a.x - a.z) < 0) ? 1 : 0;
    uint ym = (a.y - a.z) < 0 ? (1 ^ xm) : 0;
    uint zm = 1 ^ (xm | ym);
    return cross(u, float3(xm, ym, zm));
}

[[vk::binding(1, 0)]] RWTexture2D<float> debugOutput;

[numthreads(8, 8, 1)] void main(uint2 threadId
                                : SV_DispatchThreadID) {

    float3 normal = float3(0.0, 1.0, 0.0);

    RayDesc diffuseRay = (RayDesc)0;
    // Commenting out this line stops the seg fault from occurring
    diffuseRay.Direction = getPerpendicularVector(normal);
    diffuseRay.Origin = float3(0.0f, 0.0f, 0.0f);
    diffuseRay.TMin = 0.0;
    diffuseRay.TMax = 10000.0;

    float output = 0.0f;
    if (traceClosestHitRay(diffuseRay))
    {
        output = 1.0f;
    }

    debugOutput[threadId] = output;
}
```

This is also the callstack when compiling this code, with the seg fault occuring because `inst` is `nullptr`
```
[...]

PR #4913, opened 2023-01-05
[SPIR-V] Fix dangling pointer to previous field in LowerTypeVisitor

LukasBanana:
`previousField` was accessing an element from `loweredFields` **after** it has been invalidated through `emplace_back`. This change uses indexed access instead. This _might_ also be related to a segmentation fault mentioned in issue #4727


Keenuts:
Hi! Thanks for the fix! Completely missed the case where emplace_back would move the allocation! And our tests use small structs... Nice catch 😊
Not sure if it's related to the issue #4727: it reaches an assert because we copy a signed int to an unsigned, due to short-circuit, segv in prod because assert.

Keenuts (approved):

PR #4919, opened 2023-01-07
Clean up and rework DXIL library depencencies (#4877)

hekota:
* Clean up and rework DXIL library depencencies

This change reworks the dependency specifications for the DXIL* libraries. Many of these libraries had over and under specified dependencies.

This also collapses the DxilRDATBuidlder into the DxilContainer library to remove the cyclic dependency between the two libraries.

* Break assert dependency between DXIL and Analysis

In assert builds the DXILModule constructor was creating forced bindings to debugging methods that are often used from the debugger. This forced binding is useful, but doesn't need to live in the DXIL library.

To break the dependency I've moved the code into Analysis. I've also made that code only included when building with MSVC. When using other compilers `__attribute__((used))` can be applied to the function via the `LLVM_DUMP_METHOD` annotation to have the same effect.

(cherry picked from commit 2168dcb4fb0ac7841624228f8e186cc28f4cca91)


llvm-beanz (approved):
LGTM

PR #4921, opened 2023-01-09
Change hctcmdtests into lit test.

python3kgae:
Reimplement https://github.com/microsoft/DirectXShaderCompiler/pull/4893

Limited BOM test to windows.


tools/clang/test/DXC/denorm.test
```diff
@@ -0,0 +1,5 @@
+// Test for denorm options
+// RUN: %dxc  %S/Inputs/smoke.hlsl /Tps_6_2 /denorm preserve
```
llvm-beanz, line 2: These are the only test cases that I can find which pass the `/denorm` flag, and they really only verify that we don't reject the flag.

Is there something trivial that we can check for to verify that option is being interpreted correctly. If not, let's file an issue and reference the issue in a comment here to ensure we follow up.

tools/clang/test/DXC/Inputs/bom-main-ascii.hlsl
```diff
@@ -0,0 +1,20 @@
+// Tests main and include files with/without BOM to ensure BOM is stripped.
+
+#include "bom-inc-ascii.hlsli"
+#include "bom-inc-utf8.hlsli"
+#include "bom-inc-utf16le.hlsli"
+
+// TODO: Add support for Big Endian and UTF-32
```
llvm-beanz, line 7: Do we have a GitHub issue for this?

tools/clang/test/DXC/Inputs/inc/include-declarations.h
```diff
@@ -0,0 +1,3 @@
+struct foo {
+  float4x4 MPV;
+} CB;
```
llvm-beanz, line 3: nit: several of the added files don't have newlines at the end.

tools/clang/test/DXC/Inputs/include-main.hlsl
```diff
@@ -0,0 +1,13 @@
+// Test display include process with /Vi
```
llvm-beanz, line 1: Is this an input or a test? If it is a test, it shouldn't be in the Inputs subdirectory. Lit doesn't traverse into directories named `inputs` to identify test cases.

tools/clang/test/DXC/NonUniform.hlsl
```diff
@@ -0,0 +1,49 @@
...
+
+// RUN: %dxc  %t.NonUniform.cso /dumpbin /extractrootsignature /Fo %t.NonUniformRootSig.cso
+// RUN:%dxa  -listparts %t.NonUniformRootSig.cso | FileCheck %s --check-prefix=RS_PART
+
+// RS_PART:Part count: 1
+// RS_PART:RTS0
+
+// Verify root signature for NonUniform.cso
+// RUN: %dxc  %t.NonUniform.cso /dumpbin /verifyrootsignature %t.NonUniformRootSig.cso
+// RUN: %dxc  %t.NonUniformNoRootSig.cso /dumpbin /verifyrootsignature %t.NonUniformRootSig.cso
```
llvm-beanz, line 27: Is there any expected output from these commands, or is the process exit code sufficient?

tools/clang/test/DXC/NonUniform.hlsl
```diff
@@ -0,0 +1,49 @@
...
+
+// Verify mismatched root signatures fail verification
+
+// RUN: %dxc %S/Inputs/smoke.hlsl  /D "semantic = SV_Position" /T vs_6_0 /Zi /Qembed_debug /DDX12 /Fo %t.smoke.define.cso
+// RUN: %dxc %t.smoke.define.cso /dumpbin /Qstrip_rootsignature /Fo %t.norootsignature.cso
+// RUN: %dxc %t.smoke.define.cso /dumpbin /extractrootsignature /Fo %t.rootsig.cso
+
+// RUN: not %dxc %t.NonUniformNoRootSig.cso /dumpbin /verifyrootsignature %t.rootsig.cso
+// RUN: not %dxc %t.norootsignature.cso /dumpbin /verifyrootsignature %t.NonUniformRootSig.cso
```
llvm-beanz, line 36: Is there an error message we should be checking for here?

tools/clang/test/DXC/enable_16_bits.test
```diff
@@ -0,0 +1,4 @@
+// Test /enable-16bit-types option
+// RUN: %dxc  %S/Inputs/smoke.hlsl /Tps_6_2 /enable-16bit-types
+// RUN: not %dxc  %S/Inputs/smoke.hlsl /Tps_6_1 /enable-16bit-types
+// RUN: not %dxc  %S/Inputs/smoke.hlsl /Tps_6_2 /enable-16bit-types /HV 2017
```
llvm-beanz, line 4: Is there a successful value we should look for or an error message we should verify?

tools/clang/test/DXC/extra_output.test
```diff
@@ -0,0 +1,9 @@
+// Test extra DXC outputs together
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /DDX12 /Dcheck_warning /Fh %t.h /Vn g_myvar /Fc %t.ll /Fo %t.cso /Fre %t.reflection /Frs %t.rootsig /Fe %t.err
+// RUN: FileCheck --input-file=%t.h %s --check-prefix=HEADER | FileCheck --input-file=%t.ll %s --check-prefix=IR
```
llvm-beanz, line 3: The `|` gets interpreted as a shell pipe. Probably better to just put the second FileCheck command on a separate line since they are both reading from a file.

tools/clang/test/DXC/fd_strip_dbg.test
```diff
@@ -0,0 +1,8 @@
+// /Fd implies /Qstrip_debug
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Fd %t.fd /Fo %t.dxo
+// RUN:FileCheck --input-file=%t.fd %s --check-prefix=FD | %dxc -dumpbin %t.dxo | FileCheck %s --check-prefix=FO
```
llvm-beanz, line 3: Same thing here with the first `|`.

tools/clang/test/DXC/force_root_sig_ver.test
```diff
@@ -0,0 +1,5 @@
+
+// force_rootsig_ver
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl -force_rootsig_ver rootsig_1_0
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl -force_rootsig_ver rootsig_1_1
+// RUN: not %dxc /T ps_6_0 %S/Inputs/smoke.hlsl -force_rootsig_ver rootsig_2_0
```
llvm-beanz, line 5: Is there some meaningful success or failure output to check for here?

tools/clang/test/DXC/hlsl_ver.test
```diff
@@ -0,0 +1,3 @@
+// HLSL Version
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl  /HV 2016
+// RUN: not %dxc /T ps_6_0 %S/Inputs/smoke.hlsl  /HV 2015
```
llvm-beanz, line 3: If we merge the not line for this into the other HLSL version test, we could just delete this test. IMO the other test is more robust and tests all cases.

tools/clang/test/DXC/include-main.test
```diff
@@ -0,0 +1,13 @@
+// Test display include process with /Vi
```
llvm-beanz, line 1: This file looks like an `hlsl` file, not a test file

tools/clang/test/DXC/private_data.test
```diff
@@ -0,0 +1,26 @@
...
+// RUN: %dxc %t.private.smoke.cso /dumpbin /setprivate %t.private.txt /Fo %t.private.cso
+
+// RUN: %dxc %t.private.cso /dumpbin /Qstrip_priv /Fo %t.noprivate.cso
+
+// RUN: %dxc %t.private.cso /dumpbin /getprivate %t.private1.txt
+// RUN: FileCheck %s --input-file=%t.private1.txt --check-prefix=PRIVATE1
+// PRIVATE1:private data
+
+// Appending and removing private data blob roundtrip
+// RUN: fc %t.private.smoke.cso %t.noprivate.cso
```
llvm-beanz, line 14: The Windows FC command and the Unix FC command do _very_ different things. This line is almost certainly not doing what you think it is.

tools/clang/test/DXC/Inputs/bom-main-ascii.hlsl
```diff
@@ -0,0 +1,20 @@
+// Tests main and include files with/without BOM to ensure BOM is stripped.
```
tex3d, line 1: Are these files really just additions?  Aren't they coming from somewhere else?  If so, shouldn't they be moved instead to preserve the ability to trace them back in history?

tools/clang/test/DXC/spirv/spirv.hlsl
```diff
@@ -0,0 +1,4 @@
+
+// REQUIRES: spirv
+
+// RUN:%dxc %s  /T ps_6_0 -spirv
```
pow2clk, line 4: Should some kind of return value check be here?

tools/clang/test/DXC/Inputs/bom-main-ascii.hlsl
tex3d: I don't think we have anything tracking this.  There probably should be two separate issues.

1. Test BOM on Linux with utf32le in place of utf16le.
2. Expand the compiler's Unicode support to more formats, such as big endian formats, utf-32 on Win32, and utf-16 on Linux/Mac.

tools/clang/test/DXC/Inputs/bom-main-ascii.hlsl
python3kgae: How to move but not remove the original file?
I hope to keep the original file and remove them when hctcmds test disabled.

tools/clang/test/DXC/extra_output.test
python3kgae: The test will fail for cannot find the temp file generated by first line if use '|'.

tools/clang/test/DXC/fd_strip_dbg.test
python3kgae: Use | will fail for temp file cannot find.

tools/clang/test/DXC/extra_output.test
llvm-beanz: Oh, that probably has to do with when lit decides the temporary should be cleaned up. So maybe instead use `&&`.

tools/clang/test/DXC/Inputs/bom-main-ascii.hlsl
llvm-beanz: Could you update hctcmds to refer to the file in a new location?

pow2clk (commented):
I don't know if this is the time to standardize all these files to use either `-` or `_`, but it might be?

tools/clang/test/DXC/PipelineStage.hlsl
```diff
@@ -0,0 +1,31 @@
+// RUN: %dxc -T ps_6_5 -P -Fi %t.ps_65.hlsl.pp %s
+// RUN: FileCheck --input-file=%t.ps_65.hlsl.pp %s --check-prefix=PS_65
+// PS_65:ps 6 5
+
+// RUN: %dxc -T ps_6_6 -P -Fi %t.ps_66.hlsl.pp %s
+// RUN: FileCheck --input-file=%t.ps_66.hlsl.pp %s --check-prefix=PS_66
+// PS_66:ps 6 6
+
+// VS PS GS HS DS CS LIB MS AS
```
tex3d, line 9: With all the stages listed below, was this test intended to test each one, rather than just PS?

python3kgae: Nice catch.
Will fix it.

tools/clang/test/DXC/denorm.test
python3kgae: Added https://github.com/microsoft/DirectXShaderCompiler/issues/4930

tools/clang/test/DXC/Inputs/bom-main-ascii.hlsl
python3kgae: Added https://github.com/microsoft/DirectXShaderCompiler/issues/4932 and https://github.com/microsoft/DirectXShaderCompiler/issues/4931

python3kgae:
> I don't know if this is the time to standardize all these files to use either `-` or `_`, but it might be?

Let's do it in another PR. This PR is already too big.

cmake/modules/FindTAEF.cmake
```diff
@@ -41,18 +41,28 @@ set(TAEF_INCLUDE_DIRS ${TAEF_INCLUDE_DIR})
 set(TAEF_NUGET_BIN ${TAEF_INCLUDE_DIR}/../Binaries/Release)
 set(TAEF_SDK_BIN ${TAEF_INCLUDE_DIR}/../../Runtimes/TAEF)
 
+set(ARCH ${CMAKE_C_COMPILER_ARCHITECTURE_ID})
```
llvm-beanz, line 44: The changes in this file seem unrelated to the rest of the changes in this PR. They should probably be in a separate PR.

tools/clang/test/DXC/root_sig.test
```diff
@@ -0,0 +1,32 @@
...
+// RS_PART:RTS0
+
+
+// Verify root signature for smoke.cso
+// RUN: %dxc %t.define.cso /dumpbin /verifyrootsignature %t.rootsig.cso
+// RUN: %dxc %t.norootsignature.cso /dumpbin /verifyrootsignature %t.rootsig.cso
+
+// Set root signature when already has one should succeed
+// RUN: %dxc %t.define.cso /dumpbin /setrootsignature %t.rootsig.cso /Fo %t.smoke.rsadded.cso
+// RUN: %dxc -dumpbin %t.smoke.rsadded.cso
```
pow2clk, line 29: I don't know what purpose this line serves

tools/clang/test/DXC/shader_model_update.test
```diff
@@ -0,0 +1,9 @@
+
+
+// RUN: %dxc  %S/Inputs/smoke.hlsl  /D "semantic = SV_Position" /T vs_6_0 /Zi /Qembed_debug /DDX12 /Fo %t.define.cso
+
+// dxc.exe shader model upgrade
```
pow2clk, line 5: I recognize this is how this test was described before, but it just seems to verify that we can disassemble rather than anything to do with SM upgrades

tools/clang/test/DXC/strip_dbg.test
```diff
@@ -0,0 +1,18 @@
+// /Fd implies /Qstrip_debug ; path with \ produces auto hash-named .pdb file
```
pow2clk, line 1: nit: well, it's actually `/` now 😄

tools/clang/test/DXC/strip_dbg.test
```diff
@@ -0,0 +1,18 @@
...
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Fd ./ /Fo %t.strip
```
pow2clk, line 2: can we include the check for the generated .pdb file? Should we delete it afterwards since filecheck isn't' aware of its generation?

tools/clang/test/DXC/strip_reflect.test
```diff
@@ -0,0 +1,6 @@
+
+// /Qstrip_reflect strips reflection
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl -D DX12 /Qstrip_reflect /Fo %t.strip.reflection
+// RUN: %dxc -dumpbin %t.strip.reflection | FileCheck  %s --check-prefix=STRIP_REFLECTION
+// STRIP_REFLECTION:define void @main()
+// STRIP_REFLECTION-not:i32 6, !"g"
```
pow2clk, line 6: again, I'm unsure -not is doing the job here

utils/hct/hcttestcmds.cmd
```diff
@@ -407,18 +407,18 @@ call :check_file smoke.opt.prn.txt find MODULE-PRINT del
...
 if %Failed% neq 0 goto :failed
-call :run dxc_batch.exe -lib-link -multi-thread "%testfiles%\batch_cmds.txt"
+call :run dxc_batch.exe -lib-link -multi-thread "%testfiles%\Inputs\batch_cmds.txt"
 if %Failed% neq 0 goto :failed
-call :run dxc_batch.exe -multi-thread "%testfiles%\batch_cmds.txt"
+call :run dxc_batch.exe -multi-thread "%testfiles%\Inputs\batch_cmds.txt"
 if %Failed% neq 0 goto :failed
 
 set testname=Smoke test for dxl command line
 call :run dxc.exe -T lib_6_x "%testfiles%\lib_entry4.hlsl" -Fo lib_entry4.dxbc
```
pow2clk, line 418: It's true that there is a copy of this in `tools/clang/test/DXC`, but I don't know why it's not in the `Inputs/` subdirectory. It seems more consistent to put it there. Nothing is using it except this script.

pow2clk (commented):
Sorry for being nitpicky. I wanted to do a close reading of how what was tested might have been changed. By deleting the moved files, that was made possible. Thanks!

tools/clang/test/DXC/fd_zs.test
```diff
@@ -0,0 +1,7 @@
+// /Fd plus /Zs
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zs /Fd %t.pdb /Fo %t.dxo_fd
+// RUN:FileCheck --input-file=%t.pdb %s --check-prefix=PDB
+// Make sure PDB not empty.
+// PDB:{{.+}}
+// RUN:FileCheck --input-file=%t.dxo_fd %s --check-prefix=FD_FO
```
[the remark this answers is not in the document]
python3kgae: My feeling is temp file not being deleted, but not ready.
Move File check in another line will make the file ready.

tools/clang/test/DXC/fd_strip_dbg.test
```diff
@@ -0,0 +1,8 @@
...
+
+// Make sure FD not empty.
+// FD:{{.+}}
+// FO:shader debug name: {{.*}}fd_strip_dbg.test.tmp.fd
+// FO-not:DICompileUnit
```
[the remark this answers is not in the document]
python3kgae: Guess not, I 'll fix it

tools/clang/test/DXC/fd_zs.test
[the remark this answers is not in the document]
llvm-beanz: I'm not sure I understand how that would be possible. Since these are each different processes, each process starts after the last one completes. The filesystem updates should be flushed before the the next process starts.

I wonder if there is something else going on here that might be a bug we should investigate.

utils/hct/hcttestcmds.cmd
python3kgae: the copy of batch_cmds.txt in tools/clang/test/DXC is already removed.

tools/clang/test/DXC/strip_dbg.test
python3kgae: It was checked by *.pdb in hcttestcmds.cmd.
I'm not sure it is OK to run 
  rm *.pdb in lit or not.

tools/clang/test/DXC/shader_model_update.test
python3kgae: Good catch.
The real test is removed in an old change in hcttestcmds.com.
Will fix it.

tools/clang/test/DXC/root_sig.test
python3kgae: Make sure %t.smoke.rsadded.cso exist.

tools/clang/test/DXC/strip_dbg.test
python3kgae: changed to cat *.pdb | rm *.pdb
linux passed.
Hope this will work.

tools/clang/test/DXC/NonUniform.hlsl
```diff
@@ -0,0 +1,54 @@
...
+// CHECK: createHandle
+// CHECK: i1 true)
+
+// CHECK: createHandle
+// CHECK: i1 true)
+
+// CHECK: createHandle
+// CHECK: i1 true)
+
+// RUN: %dxc %s /T ps_6_0 /DDX12 /Fo %t.NonUniform.cso
```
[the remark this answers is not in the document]
python3kgae: Used cat to check file exist.

tools/clang/test/DXC/embed_dbg.test
```diff
@@ -0,0 +1,13 @@
+// Embed debug info
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
+// RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
+
+
+// Auto-embed debug info when no debug output, and expect warning signifying that this is the case.
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl  /Zi /Fo %t.embedpdb2 /Fe %t.err.embedpdb
+// RUN: FileCheck --input-file=%t.err.embedpdb %s --check-prefix=ERR_EMBED_PDB
+// ERR_EMBED_PDB:warning: no output provided for debug - embedding PDB in shader container.  Use -Qembed_debug to silence this warning.
```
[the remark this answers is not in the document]
python3kgae: not cat *pdb works in this case.

tools/clang/test/DXC/root_sig.test
python3kgae: Changed to cat.

tools/clang/test/DXC/embed_dbg.test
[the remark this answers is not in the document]
pow2clk: In conversation with Chris, I think perhaps `file -E` might be a better method as it doesn't have any issue with binary files and I expect it will be quicker

tools/clang/test/DXC/fd_zs.test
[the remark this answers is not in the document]
pow2clk: I share that concern

tools/clang/test/DXC/strip_dbg.test
pow2clk: In conversation with Chris, I no longer think deleting these files explicitly is necessary. I misunderstood how lit identifies which files need deletion

tools/clang/test/DXC/root_sig.test
python3kgae: Change to file -E.

tools/clang/test/DXC/fd_zs.test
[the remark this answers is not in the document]
python3kgae: Upstream test has some pattern.
Like https://github.com/llvm/llvm-project/blob/main/llvm/test/Analysis/Dominators/print-dot-dom.ll#L2

tools/clang/test/DXC/denorm.test
llvm-beanz: I feel like we can add a trivial check here to verify the `fp32-denorm-mode` attribute is getting generated.

tools/clang/test/DXC/include-main.test
llvm-beanz: I'm not sure I agree with how you changed this. Instead of changing the extension to `hlsl` you moved the code out to a separate file that is only used from one test. That seems unnecessary.

tools/clang/test/DXC/denorm.test
```diff
@@ -0,0 +1,9 @@
+// Test for denorm options
+// RUN: %dxc  %S/Inputs/smoke.hlsl /Tps_6_2 /denorm preserve | FileCheck %s
+// CHECK:"fp32-denorm-mode"="preserve"
+
+// RUN: %dxc  %S/Inputs/smoke.hlsl /Tps_6_2 /denorm ftz | FileCheck %s --check-prefix=ftz
+// CHECK:"fp32-denorm-mode"="ftz"
+
+// RUN: not %dxc  %S/Inputs/smoke.hlsl /Tps_6_2 /denorm abc
```
llvm-beanz, line 8: Should we check for errors here?

Should we add a run line to verify that the `fp32-denorm-mode` metadata doesn't appear if you don't pass the flag?

tools/clang/test/DXC/include-main.test
python3kgae: This file also tests hlsl file not in current directory which includes header file in different directory.
So I want to keep it as a test instead of hlsl.

llvm-beanz: I'm not sure I fully understand the value of that test (since I would hope we have alternative tests covering that case), but it could also be accomplished in a more clearly deliberate way by adding `cd %T && ` in front of the dxc command.

python3kgae: Updated with cd %T.

tools/clang/test/DXC/extra_output.test
llvm-beanz: You didn’t actually fix this…

I think we also need to understand why this causes an issue if the FileCheck lines are separate.

tools/clang/test/DXC/fd_zs.test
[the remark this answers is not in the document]
llvm-beanz: This pattern here is fine, but the test case with the `extra_output` above still seems wrong. We should be able to have those on separate `RUN` lines.

tools/clang/test/DXC/denorm.test
```diff
@@ -1,9 +1,15 @@
 // Test for denorm options
 // RUN: %dxc  %S/Inputs/smoke.hlsl /Tps_6_2 /denorm preserve | FileCheck %s
 // CHECK:"fp32-denorm-mode"="preserve"
+// RUN: %dxc  %S/Inputs/smoke.hlsl /Tps_6_2 | FileCheck %s --check-prefix=CLEAN
+// CLEAN-NOT:"fp32-denorm-mode"
+// CLEAN:define void @main()
```
llvm-beanz, line 6: I don’t think this works the way you think it does. A `NOT` directive just means it can’t be found until the next `CHECK` directive. This verifies that the attribute doesn’t occur before the main declaration, but the attribute list is always later in the file the function declarations.

A check on something early in the file (like `datalayout`, which is always early) followed by the `NOT` line will verify that the string doesn’t occur before the end of the file.

tools/clang/test/DXC/fd_zs.test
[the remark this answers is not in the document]
python3kgae: Could you show what you want with example?
Is the // PDB:{.+} you don't like?

tools/clang/test/DXC/extra_output.test
python3kgae: Fixed.
I don't think it is a problem since upstream use FileCheck in separate RUN lines too.
Maybe we could fix it in upstream first then merge back?

tools/clang/test/DXC/extra_output.test
```diff
@@ -0,0 +1,8 @@
+// Test extra DXC outputs together
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /DDX12 /Dcheck_warning /Fh %t.h /Vn g_myvar /Fc %t.ll /Fo %t.cso /Fre %t.reflection /Frs %t.rootsig /Fe %t.err && FileCheck --input-file=%t.h %s --check-prefix=HEADER && FileCheck --input-file=%t.ll %s --check-prefix=IR
```
pow2clk, line 2: This is the place that was of concern. The concern is not that this won't work, it clearly will, but that we don't understand why having it on two lines as you tried earlier didn't work. Not understanding that worries me. That said, perhaps it's not important at this point. @llvm-beanz, what do you think?

llvm-beanz: Yea, this is exactly where I’m concerned. We should investigate why this doesn’t work on separate lines.

python3kgae: That is issue of lit itself, not about this PR which just move some test to lit.
We could do that in anther PR.

python3kgae: Issue created for check why | not work https://github.com/microsoft/DirectXShaderCompiler/issues/4956

python3kgae: I think we can close this issue based on https://www.baeldung.com/linux/piped-commands-order where it says, "
The shell starts the second process while the first process is still running. Therefore, the processes run concurrently."
So FileCheck will start while dxc has not finish generating the temp files.
'|' only works for things like stdin.

llvm-beanz: Yea, so the feedback here is, let's put these on separate run lines so that they are easier to read.

pow2clk (approved):
I think we understand the run line issues. Thanks for sticking with this!

PR #4922, opened 2023-01-10
PIX: Remove per-instruction disassembly in annotation pass output

jeffnn:
llvm::Instruction::printAsOperand is quite slow. The "DxilAnnotateWithVirtualRegister" calls printAsOperand for each instruction in the shader and returns that string to PIX (in order to allow PIX to match Value names to the ordinals that this pass is adding). In shaders with tens of thousands of instructions, the overhead adds up. Previously, I did work in PIX to gather the same information directly from a one-time disassembly of the shader, obviating the need for this per-assembly print.
The resulting speed up is worth a 10x perf improvement in certain PIX UI scenarios with large shaders.


adam-yang (approved):
LGTM!

PR #4925, opened 2023-01-10
PIX: Fix overactive assert in value-to-declare pass

jeffnn:
This assert fires (fired!) under many PIX tests. No actual functional change required.

### DIFF
--- a/lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
+++ b/lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
@@ -864,8 +864,11 @@ VariableRegisters::VariableRegisters(
 {
   PopulateAllocaMap(Ty);
   m_Offsets.AlignTo(Ty); // For padding.
+
+  // (min16* types can occupy 16 or 32 bits depending on whether or not they are natively supported.
+  // If non-native, the alignment will be 32, but the claimed size will still be 16, hence the "max" here)
   assert(m_Offsets.GetCurrentAlignedOffset() ==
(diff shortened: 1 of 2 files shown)


lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -864,8 +864,11 @@ VariableRegisters::VariableRegisters(
 {
   PopulateAllocaMap(Ty);
   m_Offsets.AlignTo(Ty); // For padding.
+
+  // (min16* types can occupy 16 or 32 bits depending on whether or not they are natively supported.
+  // If non-native, the alignment will be 32, but the claimed size will still be 16, hence the "max" here)
   assert(m_Offsets.GetCurrentAlignedOffset() ==
-         DITypePeelTypeAlias(Ty)->getSizeInBits());
+         std::max<uint64_t>(DITypePeelTypeAlias(Ty)->getSizeInBits(), DITypePeelTypeAlias(Ty)->getAlignInBits()));
```
adam-yang, line 871: Does `getAlignInBits` return 32 for 16 bit types when 16bit types are not natively supported?

jeffnn: When not natively supported, getAlignInBits returns 32, and getSizeInBits returns 16.

adam-yang: Got it. Just making sure.

adam-yang (approved):
LGTM!

PR #4926, opened 2023-01-10
Fix assert when bool field in nested struct during init list casting

tex3d:
The scenario:
- Empty struct field in MyStruct2 causes init list cast path in CodeGen
- DXASSERT that the value type matches the ConvertType for the bool
- value type is i32 based on field in MyStruct from RecordLayout
- ConvertType for bool is i1, while ConvertTypeForMem will be i32.

The assert should instead check both ConvertType and ConvertTypeForMem.


llvm-beanz (approved):
LGTM

PR #4928, opened 2023-01-11
Add HlslDataDir parameter back to execution tests

hekota:
This parameter was removed when the execution tests were split into a separate dll from clang-hlsl-tests.dll. When the tests are run on the same machine that they were built on this parameter is not needed. However, for ARM64 the tests are running on a different machine and the infrastructure is using hct scripts and HLSL environment variables, so the build machine defaults do not work here.


llvm-beanz (approved):
LGTM

PR #4929, opened 2023-01-11
PIX: dxcpdbutils: Don't set default entry name in lib case

jeffnn:
In the lib case, m_EntryPoint will be non-null but empty, but shouldn't be set anyway. 
(Also, not clearing m_EntryPoint first can result in an assert in CComPtr that the poiner is non-null, and then the pointer is overwritten, resulting in a small memory leak.)

### DIFF
--- a/tools/clang/tools/dxcompiler/dxcpdbutils.cpp
+++ b/tools/clang/tools/dxcompiler/dxcpdbutils.cpp
@@ -958,10 +958,13 @@ public:
   HRESULT SetEntryPointToDefaultIfEmpty() {
     // Entry point might have been omitted. Set it to main by default.
     // Don't set entry point if this instance is non-debug DXIL and has no arguments at all.
-    // TODO: Check to see that this DxilContainer is not a library before setting the entry point.
     if ((!m_EntryPoint || m_EntryPoint->GetStringLength() == 0) && !m_ArgPairs.empty()) {
-      m_EntryPoint = nullptr;
-      IFR(Utf8ToBlobWide("main", &m_EntryPoint));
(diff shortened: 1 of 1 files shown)


tools/clang/tools/dxcompiler/dxcpdbutils.cpp
```diff
@@ -958,9 +958,13 @@ struct DxcPdbUtils : public IDxcPdbUtils2
   HRESULT SetEntryPointToDefaultIfEmpty() {
     // Entry point might have been omitted. Set it to main by default.
     // Don't set entry point if this instance is non-debug DXIL and has no arguments at all.
-    // TODO: Check to see that this DxilContainer is not a library before setting the entry point.
     if ((!m_EntryPoint || m_EntryPoint->GetStringLength() == 0) && !m_ArgPairs.empty()) {
-      IFR(Utf8ToBlobWide("main", &m_EntryPoint));
+      // Don't set the name if the target is a lib
+      if (m_TargetProfile->GetStringLength() < 3 ||
```
adam-yang, line 963: We should probably also check `m_TargetProfile` is nullptr. The code around here is a bit messy and it's not clear what m_TargetProfile would be in every code path. I think it'd be best to do:
```
if (!m_TargetProfile ||
    m_TargetProfile->GetStringLength() < 3 ||
    0 != wcsncmp(m_TargetProfile->GetStringPointer(), L"lib", 3)
```

jeffnn: Good point. Will do.

adam-yang (approved):
LGTM!

PR #4935, opened 2023-01-12
spir-v: fix missing typecast on HLSL21 ternary

Keenuts:
In the AST, the type of the ternary in the case `i < 10 ? 1 : 0` is `literal int`.

In the cast logic, we had a check saying "if it's a literal, ignore the sign for now". Then a pass would determine the correct SPIR-V sign depending on the variable the literal was stored to.

When introducing HLSL-2021, short-circuiting was allowed. This means we have to evaluate only the valid side of the ternary. This was done using a temporary register, storing the branch evaluation result.

So instead of having:
`literal -> result variable (sign=U)`
we had:
`literal -> tmp variable (sign=?) -> result variable (sign=U)`

In the first case, we could generate the load correctly, as we had the sign of the variable. In the second one, we could not, because the tmp variable sign was unknown.

Now, why cannot we know the variable sign? because the AST type is 'literal int', and has no sign information. Adding more complexity to the pass determining types could solve it, but doesn't seem like good solution. After all, the AST tells us when to add a cast! So we should simply follow the AST, and generate the cast when asked to.

This is not a perfect fix. I believe a good fix should be to remove this 'literal int' type, and keep sign information, like clang is doing today, but it would impact the DXIL side.


repro.hlsl
```diff
@@ -0,0 +1,4 @@
+[numthreads(8, 8, 1)] void main(uint2 threadId : SV_DispatchThreadID) {
```
vettoreldaniele, line 1: Can you remove these repro files?

tools/clang/lib/SPIRV/AstTypeProbe.cpp
```diff
@@ -735,13 +730,30 @@ bool isSameScalarOrVecType(QualType type1, QualType type2) {
...
+  do { // Vector types
     QualType elemType1 = {}, elemType2 = {};
     uint32_t count1 = {}, count2 = {};
-    if (isVectorType(type1, &elemType1, &count1) &&
-        isVectorType(type2, &elemType2, &count2))
-      return count1 == count2 && canTreatAsSameScalarType(elemType1, elemType2);
-  }
+    if (!isVectorType(type1, &elemType1, &count1) ||
+        !isVectorType(type2, &elemType2, &count2))
+      break;
```
vettoreldaniele, line 738: Can you remove the `do { ... } while (false);` and replace the `break` with `return false` instead?

vettoreldaniele (changes requested):

repro.hlsl
Keenuts: Oh yes, sorry!

tools/clang/lib/SPIRV/AstTypeProbe.cpp
Keenuts: Yes, that's better!

tools/clang/lib/SPIRV/AstTypeProbe.cpp
```diff
@@ -569,11 +569,6 @@ bool canTreatAsSameScalarType(QualType type1, QualType type2) {
           type2->isFloatingType()) ||
          (type2->isSpecificBuiltinType(BuiltinType::LitFloat) &&
           type1->isFloatingType()) ||
-         // Treat 'literal int' and 'int'/'uint' as the same
```
vettoreldaniele, line 572: I'm curious why this was deleted - from what I understand this affects more than just this case?

Keenuts: Yes it does, that's why I had to patch the template code, but what I believe is worst case we generate more instructions, but at least they are correct.
But yes, it could break other paths.
I don't believe this is the best solution, but fixing the AST seems more complex/risky and impacts DXIL.
(https://github.com/microsoft/DirectXShaderCompiler/issues/4727#issuecomment-1373638890)

The internal test suite is still running on this one, but so far looks good.

Keenuts: All tests are green on our end.

vettoreldaniele (approved):

Keenuts:
rebasing on main, hoping to relaunch the 2 stuck presubmits.

Keenuts:
Arf, pushed an old version..

Keenuts:
rebased on main.

Keenuts:
@llvm-beanz looks like azure pipelines are failing at the "update deps" level.

llvm-beanz:
I’ve kicked the failed checks to re-run. That looked like a network issue on the Azure VMs. I don’t know if there are special permissions required to re-run checks (@hekota would know).

When I click the `details` link on a build that had failing steps there’s a button on the right hand side to “Re-run checks” which allows you to re-run either all checks or just the failing ones.

PR #4937, opened 2023-01-12
PIX: Educate type system about samplers and resources, base classes

jeffnn:
These fixes were the result of a single shader from a well-known game engine, all of which resulted in mismatched offsets between PIX's offsets-into-fake-allocas and the offsets into aggregates for dbg.value statements.
-Base class sizes weren't being added to derived classes
-Samplers+Resources weren't being given any size at all
-Raygen payloads (referred to by pointer) were triggering an assert
Added a new set of tests for the type system to exercise these.

### DIFF
--- a/lib/DxilDia/DxilDiaSymbolManager.cpp
+++ b/lib/DxilDia/DxilDiaSymbolManager.cpp
@@ -1558,10 +1558,6 @@ HRESULT dxil_dia::hlsl_symbols::SymbolManagerInit::CreateCompositeType(DWORD dwP
     return S_OK;
   };
 
-      if (CT->getTag() == llvm::dwarf::DW_TAG_structure_type &&
-      CT->getName() == "SamplerState") {
-    CT->getName();
-  }
   IFR(AddType<symbol_factory::UDT>(dwParentID, CT, pNewTypeID, CT->getAlignInBits(), CT, LazyName));
 
   TypeInfo *udtTI;
--- a/lib/DxilDia/DxilDiaSymbolManager.cpp
+++ b/lib/DxilDia/DxilDiaSymbolManager.cpp
@@ -607,6 +607,7 @@ public:
     void Embed(const TypeInfo &TI);
 
     void AddBasicType(llvm::DIBasicType *BT);
+    void PrependBaseClassSize(uint64_t baseSize);
 
   private:
     DWORD m_dwTypeID;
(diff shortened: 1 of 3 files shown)


lib/DxilDia/DxilDiaSymbolManager.cpp
```diff
@@ -1529,13 +1563,30 @@ HRESULT dxil_dia::hlsl_symbols::SymbolManagerInit::CreateCompositeType(DWORD dwP
...
+  if (CT->getElements().size() == 0) {
+    // "Resources" (textures, samplers, etc.) are composite types without any elements,
+    // but they do have a size.
+    udtTI->AppendSize(CT->getSizeInBits());
+  } else {
+    for (llvm::DINode *N : CT->getElements()) {
+      if (auto *Field = llvm::dyn_cast<llvm::DIType>(N)) {
+        DWORD dwUnusedFieldID;
+        if (Field->getName() == "MaterialTextureBilinearWrapedSampler") {
+          Field->getName();
```
adam-yang, line 1575: Is this test code?

lib/DxilDia/DxilDiaSymbolManager.cpp
```diff
@@ -1529,13 +1563,30 @@ HRESULT dxil_dia::hlsl_symbols::SymbolManagerInit::CreateCompositeType(DWORD dwP
...
+        }
+        IFR(CreateType(Field, &dwUnusedFieldID));
+        if (Field->getTag() == llvm::dwarf::DW_TAG_inheritance) {
+          // The base class is a type of its own, so will have contributed to
+          // its own TypeInfo. But we still need to remember the size that it
+          // contributed to this type:
+          auto *DerivedType = llvm::dyn_cast<llvm::DIDerivedType>(Field);
```
adam-yang, line 1582: If we expect this cast to work, we should just use llvm::cast. This way if it fails it'll be a llvm fatal error instead of an AV.

jeffnn: Looks like the base class always has to be DIDerivedType, so yes that should always work. Changed.

lib/DxilDia/DxilDiaSymbolManager.cpp
jeffnn: Eagle eyes!

PR #4903, opened 2022-12-23
PIX: Account for alignment when computing member offset

jeffnn:
This has been a long-standing issue for PIX. Certain structs (with e.g., embedded arrays of structs) don't pay attention to their alignment requirements.

PR #4942, opened 2023-01-13
PIX: Account for alignment when computing member offset (#4903)

jeffnn:
* Account for alignment

* Use llvm::RoundUpToAlignment


jeffnn:
Merged into one PR

PR #4945, opened 2023-01-14
Added a workaround for PDBs with empty defines.

adam-yang:
PDBs compiled with /D "" results in an empty define in the PDB/debug DXIL, which causes IDxcPdbUtils to crash when reading the defines. This change does not fix the empty define, just makes IDxcPdbUtils not crash when reading PDBs like this.

The test consists of a compiled DXIL with embedded debug with an empty define. If we decide to fix the empty define later, this test would continue to test for regression.


tex3d (approved):
I think it would also be good to prevent the argument from creating an empty define in the first place in DXC.

jeffnn (approved):
Great! I'll borrow that shader to test PIX end-to-end as well.

PR #4955, opened 2023-01-19
hlsl, spirv: allow use of ffinit-math-only

### LINKED ISSUE #4954: [SPIR-V] Difficulties with NMin/NMax and lack of `-ffinite-math-only`
Commit 2a0833a (@sudonatalie for visibility) introduced `OpNMin` and `OpNMax` to translate HLSL's `min`/`max` according to the specification. I generally appreciate compiler correctness over performance but in this case it not only introduces a lot of extra dynamic branching in the GLSL backend of SPIRV-Cross but it also exposed a bug in one of our third-party platform compilers which we don't have source access to.

With this ticket I merely want to ask if we can consider resurrecting Clang compiler options like `-ffinite-math-only` to have an option to prefer performance and stability over correctness. I'm sure there are cases were the correct behavior of `min`/`max` with `NaN` are necessary, but in our case, it also causes stability issues.

My proposal for the intrinsics in question at [SpirvEmitter.cpp#L8613](https://github.com/microsoft/DirectXShaderCompiler/blob/bef540d360d5b737c83678ffde572820d1e12ec4/tools/clang/lib/SPIRV/SpirvEmitter.cpp#L8613) would look like this:
```cpp
  case hlsl::IntrinsicOp::IOP_min: {
    glslOpcode = isFloatType
                     ? (getCompilerInstance().getLangOpts().FiniteMathOnly // FMin if '-ffinite-math-only'
                            ? GLSLstd450::GLSLstd450FMin
                            : GLSLstd450::GLSLstd450NMin)
                 : isSintType ? GLSLstd450::GLSLstd450SMin
                              : GLSLstd450::GLSLstd450UMin;
    retVal = processIntrinsicUsingGLSLInst(callExpr, glslOpcode, true, srcLoc,
                                           srcRange);
    break;
  }
// same for IOP_Max
```
Thanks for considering and kind regards

Laura

Keenuts:
This options should match GCC's & Clang's behavior:

From GCC documention:
```
Allow optimizations for floating-point arithmetic that assume that arguments and results are not NaNs or +-Infs.

This option is not turned on by any -O option since it can result in incorrect output for programs that depend on an exact implementation of IEEE or ISO rules/specifications for math functions. It may, however, yield faster code for programs that do not require the guarantees of these specifications.

The default is -fno-finite-math-only.
```

This commit allows the flag to be used again, and makes SPIR-V min/max intrinsics use FMin/FMax instead of NMin/NMax when enabled.

Fixes #4954

### DIFF
--- a/include/dxc/Support/HLSLOptions.h
+++ b/include/dxc/Support/HLSLOptions.h
@@ -165,7 +165,7 @@ public:
   bool EnableDX9CompatMode = false;     // OPT_Gec
   bool EnableFXCCompatMode = false;     // internal flag
   LangStd HLSLVersion = LangStd::vUnset; // OPT_hlsl_version (2015-2021)
-  bool Enable16BitTypes = false;         // OPT_enable_16bit_types
+  bool Enable16BitTypes = false; // OPT_enable_16bit_types
   bool OptDump = false; // OPT_ODump - dump optimizer commands
   bool OutputWarnings = true; // OPT_no_warnings
   bool ShowHelp = false;  // OPT_help
--- a/include/dxc/Support/HLSLOptions.h
+++ b/include/dxc/Support/HLSLOptions.h
@@ -165,7 +165,7 @@ public:
   bool EnableDX9CompatMode = false;     // OPT_Gec
   bool EnableFXCCompatMode = false;     // internal flag
   LangStd HLSLVersion = LangStd::vUnset; // OPT_hlsl_version (2015-2021)
-  bool Enable16BitTypes = false; // OPT_enable_16bit_types
+  bool Enable16BitTypes = false;         // OPT_enable_16bit_types
   bool OptDump = false; // OPT_ODump - dump optimizer commands
   bool OutputWarnings = true; // OPT_no_warnings
   bool ShowHelp = false;  // OPT_help
--- a/include/dxc/Support/HLSLOptions.h
+++ b/include/dxc/Support/HLSLOptions.h
@@ -166,7 +166,6 @@ public:
   bool EnableFXCCompatMode = false;     // internal flag
   LangStd HLSLVersion = LangStd::vUnset; // OPT_hlsl_version (2015-2021)
   bool Enable16BitTypes = false; // OPT_enable_16bit_types
-  bool FiniteMathOnly = false;   // finite_math_only
   bool OptDump = false; // OPT_ODump - dump optimizer commands
   bool OutputWarnings = true; // OPT_no_warnings
   bool ShowHelp = false;  // OPT_help
--- a/include/dxc/Support/HLSLOptions.h
+++ b/include/dxc/Support/HLSLOptions.h
@@ -166,6 +166,7 @@ public:
   bool EnableFXCCompatMode = false;     // internal flag
   LangStd HLSLVersion = LangStd::vUnset; // OPT_hlsl_version (2015-2021)
   bool Enable16BitTypes = false; // OPT_enable_16bit_types
+  bool FiniteMathOnly = false;   // finite_math_only
   bool OptDump = false; // OPT_ODump - dump optimizer commands
   bool OutputWarnings = true; // OPT_no_warnings
   bool ShowHelp = false;  // OPT_help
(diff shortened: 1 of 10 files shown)


vettoreldaniele (commented):
Can we document this behavior in the SPIR-V wiki? At least to keep track of what this option does and what it does not (yet). I'm assuming there are more optimizations that can be enabled through that flag's assumption.

include/dxc/Support/HLSLOptions.td
```diff
@@ -143,8 +143,10 @@ def fno_associative_math : Flag<["-"], "fno-associative-math">, Group<hlsloptz_G
 //  Flag<["-"], "freciprocal-math">, Group<hlsloptz_Group>, Flags<[CoreOption]>,
 //  HelpText<"Allow division operations to be reassociated">;
 def fno_reciprocal_math : Flag<["-"], "fno-reciprocal-math">, Group<hlsloptz_Group>;
-def ffinite_math_only : Flag<["-"], "ffinite-math-only">, Group<hlsloptz_Group>, Flags<[CoreOption]>;
-def fno_finite_math_only : Flag<["-"], "fno-finite-math-only">, Group<hlsloptz_Group>;
+def ffinite_math_only: Flag<["-"], "ffinite-math-only">, Group<hlsloptz_Group>, Flags<[CoreOption]>,
+  HelpText<"Allow optimization for floating-point arithmetic that assume that arguments and results are not NaNs or +-Infs.">;
```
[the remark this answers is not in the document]
Keenuts: > Can we document this behavior in the SPIR-V wiki? At least to keep track of what this option does and what it does not (yet). I'm assuming there are more optimizations that can be enabled through that flag's assumption.

That's a good point, adding this.

vettoreldaniele (commented):
LGTM but please wait for somebody on Microsoft's side.

Keenuts:
just rebased on main, to unblock bots that were failing the last time.

Keenuts:
Hello @llvm-beanz, would you have some time to review this new option?

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -731,6 +731,8 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
   else
     opts.OptLevel = 3;
   opts.OptDump = Args.hasFlag(OPT_Odump, OPT_INVALID, false);
+  opts.FiniteMathOnly =
+      Args.hasFlag(OPT_ffinite_math_only, OPT_fno_finite_math_only, false);
```
llvm-beanz, line 735: Can we instead set this value right into the SpirvOptions so that we don't need a separate bool in the HLSLOptions?

llvm-beanz (commented):
One minor suggestion to simplify the code, but looks good to me.

lib/DxcSupport/HLSLOptions.cpp
Keenuts: The idea was have this option higher-up, as it could impact both SPIRV and DXIL codegen, in case you'd also want to bring some optimizations depending on this flag. (like the general optimization level flag which impacts both backends)
Let me know if you prefer for this option to be spirv only.

llvm-beanz: We already handle this option for DXIL. All flags marked `CoreOption` are identical matches to clang flags and we handle them by passing the flag into Clang's option parsing (search for `CopyArgsToWStrings`).

We don't need any special handling for the option for DXIL.

An alternate approach SPIR-V could take would be to read the flag values out of the Clang Language Options. The parsing code today does store the value into LangOpts so it is probably already available without additional handling.

Keenuts: Ahh I see, completely missed that! Thanks for the explanation!

include/dxc/Support/HLSLOptions.h
```diff
@@ -165,7 +165,7 @@ class DxcOpts {
   bool EnableDX9CompatMode = false;     // OPT_Gec
   bool EnableFXCCompatMode = false;     // internal flag
   LangStd HLSLVersion = LangStd::vUnset; // OPT_hlsl_version (2015-2021)
-  bool Enable16BitTypes = false; // OPT_enable_16bit_types
+  bool Enable16BitTypes = false;         // OPT_enable_16bit_types
```
llvm-beanz, line 168: While I think the formatting change here is a strict improvement, maybe leave it out for this commit since it is unrelated.

llvm-beanz (commented):
One nit, otherwise looks good to me.

llvm-beanz (approved):

PR #4957, opened 2023-01-20
Explicitly use n* version of vsnprintf

pow2clk:
Newest clang makes use of older version a warning, which we get as an error


tools/clang/lib/Parse/HLSLRootSignature.cpp
```diff
@@ -530,7 +530,7 @@ HRESULT RootSignatureParser::Error(uint32_t uErrorNum, LPCSTR pError, ...)
     va_list Args;
     char msg[512];
     va_start(Args, pError);
-    vsprintf_s(msg, pError, Args);
+    vsnprintf_s(msg, _countof(msg), pError, Args);
```
python3kgae, line 533: So the size will always be 512?

python3kgae (approved):

PR #4960, opened 2023-01-24
Fix issue sys::findProgramByName cannot find exe on path

python3kgae:
Merge https://github.com/llvm/llvm-project/commit/87780300f624e5437aee0c5a2d7ca30ea87b5ba2 to fix test issue where not.exe cannot find file.exe in https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/DXC/embed_dbg.test#L4

### DIFF
--- a/lib/Support/Windows/Program.inc
+++ b/lib/Support/Windows/Program.inc
@@ -71,7 +71,7 @@ ErrorOr<std::string> sys::findProgramByName(StringRef Name,
   DWORD Len = MAX_PATH;
   for (StringRef Ext : PathExts) {
     do {
-      U16Result.resize(Len);
+      U16Result.reserve(Len);
       // Lets attach the extension manually. That is needed for files
       // with a point in name like aaa.bbb. SearchPathW will not add extension
       // from its argument to such files because it thinks they already had one.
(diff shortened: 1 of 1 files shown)


lib/Support/Windows/Program.inc
```diff
@@ -70,14 +70,17 @@ ErrorOr<std::string> sys::findProgramByName(StringRef Name,
...
   for (StringRef Ext : PathExts) {
-    SmallVector<wchar_t, MAX_PATH> U16Ext;
-    if (std::error_code EC = windows::UTF8ToUTF16(Ext, U16Ext))
-      return EC;
-
     do {
-      U16Result.reserve(Len);
-      Len = ::SearchPathW(Path, c_str(U16Name),
-                          U16Ext.empty() ? nullptr : c_str(U16Ext),
+      U16Result.resize(Len);
```
tex3d, line 74: I see you changed `reserve` to `resize`, though `set_size` is used later to set it to size `Len`.  This wasn't part of the original change in upstream.  Was there a specific reason for this change?

tex3d (approved):

lib/Support/Windows/Program.inc
python3kgae: changed back to reserve.

PR #4962, opened 2023-01-24
Fix warning caused by redefine NOMINMAX.

python3kgae:


tools/llvm-as/llvm-as.cpp
```diff
@@ -84,10 +84,12 @@ static void WriteOutputFile(const Module *M) {
 }
 
 // HLSL Change Starts
-#define NOMINMAX
 #include "dxc/Support/WinIncludes.h"
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/MSFileSystem.h"
+#ifndef NOMINMAX
+#define NOMINMAX
```
dmpots, line 91: I think this define should be before the WinIncludes.h include. I think its purpose is to prevent windows headers from defining min/max macros.

python3kgae: Moved ifndef to WinIncludes.h.

dmpots (approved):
LGTM

PR #4964, opened 2023-01-24
Add lit test for execution test.

python3kgae:


tools/clang/test/taef_exec/lit.cfg
```diff
@@ -0,0 +1,39 @@
...
+        lit_config.load_config(config, site_cfg)
+        raise SystemExit
+
+# TAEF only runs on Windows
+if platform.system() != 'Windows':
+  config.unsupported = True
+else:
+  test_dll = os.path.join(config.llvm_obj_root, config.llvm_build_mode, 'bin', 'ExecHLSLTests.dll')
+
+  hlsl_data_dir = os.path.join(config.llvm_src_root, 'tools', 'clang', 'test', 'HLSL')
```
tex3d, line 34: Is `hlsl_data_dir` unused?  Shouldn't it contribute to `extra_params`?

tools/clang/test/taef_exec/lit.cfg
```diff
@@ -0,0 +1,39 @@
...
+
+  test_dir = os.path.join(config.llvm_obj_root, config.llvm_build_mode, 'test')
+
+  extra_params = ["/p:\"ExperimentalShaders=*\""]
```
tex3d, line 38: How can parameters be added / customized when running the test with the lit runner, like we support for `hcttest.cmd`?  For instance, how do you specify `/p:"Adapter=..."`?

tools/clang/test/taef_exec/lit.cfg
python3kgae: with hlsl_data_dir, a lot of execution test fails.

tools/clang/test/taef_exec/lit.cfg
python3kgae: Not supported now.
We could add more information to config and send it down as extra_params.

tools/clang/test/taef_exec/lit.cfg
tex3d: Without it, does it silently skip or pass without actually running?  I don't think we want this in this state.

python3kgae: Found the issue, used wrong hlsl_data_dir :(

tools/clang/test/taef_exec/lit.cfg
```diff
@@ -0,0 +1,40 @@
...
+
+# Check that the object root is known.
+te = getattr(config, 'te', None)
+if te is None:
+    # Otherwise, we haven't loaded the site specific configuration (the user is
+    # probably trying to run on a test file directly, and either the site
+    # configuration hasn't been created by the build system, or we are in an
+    # out-of-tree build situation).
+
+    # Check for 'clang_unit_site_config' user parameter, and use that if available.
```
tex3d, line 22: Comment doesn't match code: `'clang_unit_site_config'` should be `'clang_taef_exec_site_config'`.

python3kgae: Fixed.

tools/clang/test/taef_exec/lit.cfg
python3kgae: Added.

cmake/modules/FindTAEF.cmake
```diff
@@ -55,6 +55,16 @@ else()
   message(ERROR "Unable to find TAEF binaries.")
 endif()
 
+set(TAEF_ARCH "x86")
```
tex3d, line 58: Defaults to x86 if other expressions fail?  I'm not sure that's desirable.

tools/clang/test/taef/lit.cfg
```diff
@@ -34,4 +34,12 @@ else:
...
 
   test_dir = os.path.join(config.llvm_obj_root, config.llvm_build_mode, 'test')
-  config.test_format = lit.formats.TaefTest(config.te, test_dll, hlsl_data_dir, test_dir)
+
+  param_hlsl_data_dir = str.format('/p:HlslDataDir={}', hlsl_data_dir)
+  extra_params = [param_hlsl_data_dir]
+
+  verbose = lit_config.params.get('verbose', None)
+  if verbose == None:
+      extra_params.append('/logOutput:LowWithConsoleBuffering')
```
pow2clk, line 43: Does this address the taef tests in the `unittests` directory including `CompilerTest`? I would have thought they'd need a priority parameter too as some of them are restricted based on it

tools/clang/test/taef_exec/lit.cfg
```diff
@@ -0,0 +1,68 @@
...
+  verbose = lit_config.params.get('verbose', None)
+  if verbose == None:
+      extra_params.append('/logOutput:LowWithConsoleBuffering')
+
+  arch = getattr(config, 'taef_arch', None)
+  arch_filter = str.format("@Architecture='{}'", arch)
+  priority_filter = "@Priority<2"
+  future_filter = lit_config.params.get('future_filter', None)
+  if future_filter:
+      priority_filter = "@Priority=2"
```
pow2clk, line 50: I'm not aware of any Execution tests that have a Priority of 2. they all look to be 0

pow2clk (commented):
It looks like a very positive change. I'm just concerned that the priority flag is getting set for the wrong set of tests

tools/clang/test/taef_exec/lit.cfg
tex3d, line 50: In hcttest.cmd, the `exec-future-filter` combines `exec-future` and `exec-filter` mode.  The first sets the `@Priority` filter, but the second is supposed to set the `@Name` filter.

For this, I think it would be best to have separate options: `taef_exec_future` which just sets this priority, and if we want to support name filtering (that would certainly be important if moving off `hcttest.cmd`), then we should have another `taef_exec_filter` option.  This implies we should have a `taef_filter` option for the other non-exec taef test as well.  Then name filtering needs to be incorporated into the `/select:` string option.

tex3d: Priority=2 was intended for in-development features that wouldn't run normally on standard runtime/warp, but needed a specific configuration to run.  So you wouldn't want to run them in normal test pass.

pow2clk: Sure, I don't mind it being here to much as I worry about it not being in the other runner since ther are Execution and DxilContainer tests that make use of it. When not specified with a flag, does it default to 0? Maybe that's fine? If we ever want to promote those tests to more regular testing, maybe we'll just update their priorities.

Incidentally, googletest has no notion of priorities, but you can disable tests that can be run if you explicitly allow disabled tests. So this might have to change going forward anyway

tools/clang/test/taef/lit.cfg
python3kgae: Yes. I'll add the priority parameter for taef/lit.cfg too.

tools/clang/test/taef_exec/lit.cfg
python3kgae: For lit test, each name is tested separately with /name:testName.
The testName is get from te /list.
So @Name filter is not offered for lit test.
Maybe we can keep hcttest.com for case @Name filter is needed?

cmake/modules/FindTAEF.cmake
python3kgae: Fixed.

tools/clang/test/taef_exec/lit.cfg
tex3d: I think `future_filter` isn't the best name to go with here, and could either be made consistent with the other taef test as `no_priority` removing the priority filte, or we could name it `exec_future` if we still want to filter to `@Priority=2`.

Another option for the filtering is to provide a `taef_select` argument (and maybe a separate `taef_exec_select`?) that supplies taef's `/select` argument contents instead of constructing them from other arguments here.

tex3d (approved):

PR #4967, opened 2023-01-25
Allow noinline function has struct argument.

python3kgae:


tools/clang/test/DXC/noinline_struct_arg.hlsl
```diff
@@ -0,0 +1,26 @@
+// RUN: %dxc -E main -T ps_6_0 %s | FileCheck %s
```
llvm-beanz, line 1: I know that native lit tests are easier to write, but until all our testing infrastructure is migrated to lit, we should still prefer writing tests in the older infrastructure.

llvm-beanz (commented):
Please provide a description describing the motivation for this change.

Also, do we need changes to DXL or does this work as expected? If it works as expected can you also add a DXL test case?

tools/clang/test/DXC/noinline_struct_arg.hlsl
python3kgae: Developing under linux.
I hate Taef :(

python3kgae: Added test for linking.
Let's wait for switch to lit test by default.

llvm-beanz: What are you suggesting we do after switching to lit?

I’m not sure we have a clear timeline for moving our testing, so I’m not sure we should be planning anything depending on the switch to lit.

python3kgae: I suggested hold this PR until lit is switched as default way to test.

Once execution test supported to lit, we'll have all we need to switch to lit.

llvm-beanz: I’ll leave it up to you if you want to hold off on the PR until we move to lit, but I don’t think we’re as close as you do.

We can’t add tests that only run under lit until we remove the old test infrastructure. All the current lit-based test cases have non-lit equivalent tests.

When we do switch to writing lit tests, we should also organize our suites in an intuitive way, we shouldn’t just dump all the tests under the `DXC` suite. We probably should have separate suites for dxl and validation tests.

tex3d (changes requested):

damyanp:

damyanp:

PR #4968, opened 2023-01-25
[bug] Fix possible leak when creating empty blobs

dmpots:
In DxcCreateBlob we check to see if the wrapped buffer is empty and throw away the passed in pointer to use nullptr instead.

This is fine except that the passed pointer could have been allocated through IMalloc interface that needs to free the original pointer. The IMalloc docs allow allocating 0 bytes and specify that the returned pointer should be non-null.

This PR updates DxcCreateBlob to free the passed pointer in cases where the resulting blob should take ownership of the passed in buffer.

### DIFF
--- a/lib/DxcSupport/FileIOHelper.cpp
+++ b/lib/DxcSupport/FileIOHelper.cpp
@@ -612,6 +612,15 @@ HRESULT DxcCreateBlob(
 
   // Handle empty blob
   if (emptyString) {
+    // Free the passed in pointer if we are supposed to be taking
+    // ownership of it. It is a bit strange to have allocated a zero-sized
+    // buffer, but the IMalloc docs explicitly allow it:
+    //
+    // "If cb is zero, Alloc allocates a zero-length item and returns a valid pointer to that item"
+    //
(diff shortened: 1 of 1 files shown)


lib/DxcSupport/FileIOHelper.cpp
```diff
@@ -612,6 +612,15 @@ HRESULT DxcCreateBlob(
 
   // Handle empty blob
   if (emptyString) {
+    // Free the passed in pointer if we are supposed to be taking
+    // ownership of it. It is a bit strange to have allocated a zero-sized
+    // buffer, but the IMalloc docs explicitly allow it:
+    //
+    // "If cb is zero, Alloc allocates a zero-length item and returns a valid pointer to that item"
```
tex3d, line 619: This case can happen with non-zero sized buffer when it's just a null-terminator, or BOM + null-terminator (but BOM in this scenario has another bug that would blow this up, which needs fixing).

tex3d (approved):

damyanp:

damyanp:

PR #4970, opened 2023-01-26
Fix leak and BOM handling issues in DxcCreateBlob

tex3d:
A leak would happen when input buffer is considered an emptyString: zero-length or a string with only a BOM and/or null-terminator. Fixed by freeing the supplied buffer in the emptyString case when we are supposed to take ownership of it.

BOM handling was also problematic because it offset the original pPtr input to the function. If we were taking ownership of this buffer, this was not safe, since the pointer could no longer be used with Free.

Fix is:
- Keep track of pContents and contentSize separately, adjusting those according to the BOM.
- Use pContents and contentSize for bPinned and bCopy cases.
- Wrap blob in blob to skip BOM when taking ownership of memory.

Also:
- moved null-termination into copy block, instead of next to blob creation.

### DIFF
--- a/lib/DxcSupport/FileIOHelper.cpp
+++ b/lib/DxcSupport/FileIOHelper.cpp
@@ -587,8 +587,13 @@ HRESULT DxcCreateBlob(
   unsigned bomSize = (encodingKnown && codePage != CP_ACP)
                          ? GetBomLengthFromBytes((const char *)pPtr, size)
                          : 0;
+
+  // pContents may point into buffer, past the BOM.
+  LPCVOID pContents = pPtr;
+  SIZE_T contentSize = size;
+
   if (bomSize) {
-    // Adjust pPtr and size to skip BOM.
+    // Adjust pContents and contentSize to skip BOM.
     // When !encodingKnown or codePage == CP_ACP, BOM will be skipped when
     // interpreting as text and translating to unicode, since at this point,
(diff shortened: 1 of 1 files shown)


lib/DxcSupport/FileIOHelper.cpp
```diff
@@ -668,34 +678,50 @@ HRESULT DxcCreateBlob(
...
-    else
-      memset(pData, 0, size);
+    DXASSERT_NOMSG(pContents && contentSize);
+    memcpy(pData, pContents, contentSize);
+
+    // Guarantee null-termination
+    if (bNullTerminated && newSize > contentSize) {
+      if (codePage == CP_UTF8)
+        ((char*)pData)[newSize - 1] = 0;
+      else if (codePage == DXC_CP_WIDE)
```
dmpots, line 688: What if codePage is neither CP_UTF8 or DXC_CP_WIDE? Is that impossible? Could add an else { assert(false) } in that case.

dmpots (approved):
The leak case will not be caught by the current test infrastructure (requires leak tracking).

But the BOM case could have been (we should have crashed freeing a pointer that was not returned from Alloc()). We should add some test coverage for the BOM case and probably good to flesh out the empty string and zero-length tests (if needed) as well.

damyanp:

damyanp:

PR #4971, opened 2023-01-26
Add a github action to verify PR descriptions

llvm-beanz:
This action just verifies that PRs contain descriptions. Our policy requires PR descriptions, but we don't have consistent application of the policy. This should help correct that.

### DIFF
--- a/.github/workflows/pr-description-checker.yml
+++ b/.github/workflows/pr-description-checker.yml
@@ -1,0 +1,18 @@
+name: 'PR description checker'
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - labeled
+      - unlabeled
+jobs:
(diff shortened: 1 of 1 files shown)


dmpots (approved):
Sounds like a good idea to me.

diegonovillo (approved):

vettoreldaniele (commented):
I saw a few actions in the marketplace that don't require the repo to be checked out (which can add time for such a simple check). I'm ok with this but maybe worth looking into for future improvement.

llvm-beanz:
> I saw a few actions in the marketplace that don't require the repo to be checked out (which can add time for such a simple check). I'm ok with this but maybe worth looking into for future improvement.

That's fair... I think right now our other status checks are slow enough that this one will _always_ finish first.

PR #4972, opened 2023-01-26
Add experimental -fnew-inlining-behavior

llvm-beanz:
This flag is _extremely_ experimental. It is the first step in an effort to clean up and adjust our optimization pipeline ordering, which itself is part of an effort to evolve and formalize DXIL libraries. This first step enables compiling shaders using a more traditional LLVM pass ordering and the normal simple inliner rather than the AlwaysInliner. It also disables annotating `alwaysinline` on functions in library shaders so that heuristic-driven inlining decisions can be made. This will generate incorrect code and surface crashes and other bugs in the compiler, which is why it is opt-in.

There is much more work to do to get this flag to generate valid shaders, but adding the flag so that we can start working on it is the first step.

### DIFF
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -172,9 +172,9 @@ def fdisable_loc_tracking : Flag<["-"], "fdisable-loc-tracking">,
   Group<hlslcomp_Group>, Flags<[CoreOption]>,
   HelpText<"Disable source location tracking in IR. This will break diagnostic generation for late validation. (Ignored if /Zi is passed)">;
 
-def fdisable_always_inline : Flag<["-"], "fdisable-always-inline">,
+def fnew_inlining_behavior : Flag<["-"], "fnew-inlining-behavior">,
   Group<hlslcomp_Group>, Flags<[CoreOption]>,
-  HelpText<"Experimental option for library shaders to disable forcing inlining calls under entry functions.">;
+  HelpText<"Experimental option to use heuristics-driven late inlining and disable alwaysinline annotation for library shaders">;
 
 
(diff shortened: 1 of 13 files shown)


include/dxc/Support/HLSLOptions.td
```diff
@@ -172,6 +172,10 @@ def fdisable_loc_tracking : Flag<["-"], "fdisable-loc-tracking">,
   Group<hlslcomp_Group>, Flags<[CoreOption]>,
   HelpText<"Disable source location tracking in IR. This will break diagnostic generation for late validation. (Ignored if /Zi is passed)">;
 
+def fdisable_always_inline : Flag<["-"], "fdisable-always-inline">,
```
python3kgae, line 175: Shall we add it as something like always_force_no_inline which always add no inline on all functions to expose more issues?

pow2clk (commented):
I don't have a counterproposal because I know what this flag actually does is hard to explain, but I really don't like booleans and flags with negations in the name 😝

Do you have any other proposals?

include/dxc/Support/HLSLOptions.td
tex3d: I would prefer if it was added to our opt enable/disable system if that makes sense, like: `-opt-disable always-inline`

llvm-beanz: It isn't really disabling a pass... it is changing the inlining policy (swapping inliners and the timing of the pass in the pipeline), as well as changing how we generate function declarations.

llvm-beanz: > Shall we add it as something like always_force_no_inline which always add no inline on all functions to expose more issues?

Forcing adding `noinline` isn’t the goal here. The goal is to enable a compiler mode where profitable inlining occurs but we don’t force _unprofitable_ inlining the way that the `alwaysinline` attribute does.

tex3d (approved):

PR #4973, opened 2023-01-27
Fix lit test fails

python3kgae:
Added dxc_batch as dependent for clang tests.
Remove select filter for taef test which is conflict with name.
Add strip_dxil_validator_path to work around outdated dxil.dll.
Disable taef_exec for check-clang so execution test and normal test are separate.


llvm-beanz (changes requested):
We should separate any functional changes from enabling this by default.

python3kgae:
> We should separate any functional changes from enabling this by default.

Will cleanup once default lit passed github test.

utils/lit/lit/TestingConfig.py
```diff
@@ -1,13 +1,39 @@
 import os
 import sys
+import platform
 
 OldPy = sys.version_info[0] == 2 and sys.version_info[1] < 7
 
+def strip_dxil_validator_path(env_path):
```
tex3d, line 7: Looks fine, but in case you're curious, the short version could look like:
```py
def strip_dxil_validator_path(env_path):
    dxil_name, separator = ('dxil.dll', ';') if platform.system() == 'Windows' else ('dxil.so', ';')
    return separator.join([
        p for p in env_path.split(separator)
        if not os.path.isfile(os.path.join(p, dxil_name))
        ])
```

tools/clang/test/CMakeLists.txt
```diff
@@ -122,6 +123,7 @@ if (WIN32)
   add_lit_target("check-clang-taef-exec" "Running lit suite hlsl execution test"
             ${CMAKE_CURRENT_SOURCE_DIR}/taef_exec
             PARAMS ${CLANG_TEST_PARAMS}
+                  need_dxil_validator=True
```
tex3d, line 126: I don't think this is the right mechanism.  You don't technically need the validator to run execution tests if you use experimental shader models.  If you do want to use a validator, you don't necessarily want a random one from the path.  You want to use one specifically selected for testing.  In either case, we don't want a random DXIL.dll from the path to be used.

utils/lit/lit/TestingConfig.py
```diff
@@ -16,9 +42,16 @@ def fromdefaults(litConfig):
         Create a TestingConfig object with default values.
         """
         # Set the environment based on the command line arguments.
+
+        # strip dxil validator dir if not need it.
+        all_path = os.pathsep.join(litConfig.path +
+                                     [os.environ.get('PATH','')])
+        need_dxil_validator = litConfig.params.get('need_dxil_validator', None)
+        if need_dxil_validator == None:
+            all_path = strip_dxil_validator_path(all_path)
```
tex3d, line 51: See previous comment about `need_dxil_validator`.  Do we rely on the path to look up to find the TAEF binaries we are testing?  If not, we should just make sure the DXIL.dll we want to test is next to that binary, and not ever picked up from the path.  If we do, then I'm worried that we need to change the path stripping code, or add the path to the test content after stripping any paths leading to potentially unwanted DXIL.dlls.

utils/lit/lit/formats/taef.py
```diff
@@ -80,15 +82,24 @@ def getTestsInDirectory(self, testSuite, path_in_suite,
...
+        #    select_filter = str.format("{} AND {}", select_filter, self.select_filter)
+
+        select_filter = str.format('/select:{}', select_filter)
+
         cmd = [self.te, test_dll, '/inproc',
+                select_filter,
                 '/miniDumpOnCrash', '/unicodeOutput:false',
-                str.format('/outputFolder:{}', self.test_path),
-                str.format('/name:{}', testName)]
+                str.format('/outputFolder:{}', self.test_path)]
```
tex3d, line 102: I think we need to change this whole function for TAEF execution, because apparently, TAEF custom-processes the command line instead of letting the CRT process it first.  This method of passing arguments will cause them to be quoted if necessary by subprocess.Popen, but the quotes won't be where TAEF wants them, so it won't recognize the arguments properly at all.

Instead, this needs to construct the full command line string and use `shell=True` when invoking `subprocess.Popen()`.  Each injected value that could contain characters that need quoting (path names, `select_filter`) need to be quoted just around the value.  Like:

```py
def execute(self, test, litConfig):

    test_dll = test.getFilePath()

    testPath,testName = os.path.split(test.getSourcePath())

    select_filter = str.format("@Name='{}'", testName)

    if self.select_filter != "":
        select_filter = str.format("{} AND {}", select_filter, self.select_filter)

    select_filter = str.format('/select:"{}"', select_filter)

    cmd = [self.te, test_dll, '/inproc',
        select_filter,
        '/miniDumpOnCrash', '/unicodeOutput:false',
        str.format('/outputFolder:"{}"', self.test_path)]

    cmd.extend(self.extra_params)

    if litConfig.useValgrind:
        cmd = litConfig.valgrindArgs + cmd
    if litConfig.noExecute:
        return lit.Test.PASS, ''

    # TAEF must be run with custom command line string and shell=True
    # because of the way it manually processes quoted arguments in a
    # non-standard way.
    command = ' '.join(cmd)

    print(command)

    p = subprocess.Popen(command, cwd=cwd, shell=True,
                         stdin=subprocess.PIPE,
                         stdout=subprocess.PIPE,
                         stderr=subprocess.PIPE,
                         env=test.config.environment,
                         close_fds=lit.util.kUseCloseFDs)
    out,err = p.communicate()
    exitCode = p.wait()

    # Detect Ctrl-C in subprocess.
    if exitCode == -signal.SIGINT:
        raise KeyboardInterrupt

    # Ensure the resulting output is always of string type.
    out = convert_string(out)
    err = convert_string(err)

    ...
```

Also need to quote a few other variables set in `lit.cfg` files elsewhere:
`param_hlsl_data_dir = str.format('/p:HlslDataDir="{}"', hlsl_data_dir)`
`extra_params.append(str.format('/p:Adapter="{}"', adapter))`

utils/lit/lit/TestingConfig.py
python3kgae: This is much better.

utils/lit/lit/TestingConfig.py
python3kgae: Removed need_dxil_validator.

utils/lit/lit/formats/taef.py
python3kgae: Thanks a lot for work this out.

tools/clang/test/taef_exec/lit.cfg
```diff
@@ -24,45 +25,64 @@ if te is None:
...
+    result = subprocess.run([dxexp], capture_output=True)
+    no_experimental = "Experimental shader model feature failed"
+    out = lit.util.convert_string(result.stdout)
+    if no_experimental in out:
+        print("""Experimental shader model feature is not enabled.
+  Execution test require dxil.dll as validator to sign dxil container or enable experimental shader model feature to avoid signature check.
+              """)
+        config.unsupported = True
+
+skip_taef_exec = lit_config.params.get('skip_taef_exec', None)
```
tex3d, line 49: Why not move this up and the code in the prior else down to when we know exec tests are enabled?  If you're skipping execution tests, it would be odd to get errors about the execution tests still.

python3kgae: Fixed.

llvm-beanz (approved):

PR #4974, opened 2023-01-27
Support taef in external/taef.

python3kgae:
Fix cmake fail for cannot find taef in external/taef which is get from hctgettaef.
Also allow hctgettaef find HLSL_SRC_DIR automatically.


cmake/modules/FindTAEF.cmake
```diff
@@ -55,16 +66,11 @@ else()
...
-elseif ((CMAKE_GENERATOR_PLATFORM MATCHES "ARM64.*") OR ("${CMAKE_C_COMPILER_ARCHITECTURE_ID}" MATCHES "ARM64.*"))
-  set(TAEF_ARCH "arm64")
-elseif ((CMAKE_GENERATOR_PLATFORM MATCHES "ARM.*") OR ("${CMAKE_C_COMPILER_ARCHITECTURE_ID}" MATCHES "ARM.*"))
-  set(TAEF_ARCH "arm")
-endif((CMAKE_GENERATOR_PLATFORM STREQUAL "x64") OR ("${CMAKE_C_COMPILER_ARCHITECTURE_ID}" STREQUAL "x64"))
-
+# taef/build/Binaries use amd64 for x64.
+# For hlsl test, use x64 for x64.
+if (TAEF_ARCH STREQUAL "amd64")
+    set(TAEF_ARCH "x64")
```
tex3d, line 72: I don't like the redefinition of the same variable, that's scary.  Can we use two separate variables if they have to be different?  `TAEF_TEST_ARCH` perhaps?

llvm-beanz: Yea... this is odd. Why not have the if/elseif blob above set `TAEF_ARCH` and `TAEF_EXTERNAL_PATH` as separate variables, then you only have one conditional block.

python3kgae: Will remove the redef.
It will be useless after https://github.com/microsoft/DirectXShaderCompiler/pull/4973 remove the filter for filter conflict with name for taef test.

python3kgae: Removed.

damyanp:

damyanp:

PR #4977, opened 2023-01-27
Pix: Debug instrumentation: Create values per-function (#4976)

jeffnn:
The previous formulation had a single member variable for several Values which would be used in multiple functions of a lib, which of course means that in all but one of those functions, the Values' creation function was not the right one. Funnily enough, I had got this right in the access-tracking pass but somehow neglected to do it in the debugging pass. (Also need to audit the other passes to make sure I haven't been repeating the mistake.)


hekota (approved):
LGTM!

PR #4978, opened 2023-01-28
Explicitly use n* version of vsnprintf (#4957)

pow2clk:
Newest clang makes use of older version a warning, which we get as an error

(cherry picked from commit 79efe5342e2ae15e8b9ebcdf5680098bf3204dd8)


hekota (approved):
LGTM!

PR #4963, opened 2023-01-24
Remove Ignore property from execution tests

hekota:

PR #4980, opened 2023-01-31
Fix hcttest file-check mode with /runIgnoredTests

tex3d:
`hcttest.cmd file-check` runs the `CompilerTest::ManualFileCheckTest` test which is normally Ignored because it's only there to run a custom manual filecheck path. A prior change removed the /runIgnoredTests argument, breaking the testing mode.  This change just adds it back.

This was broken by PR #4963.


hekota (approved):
LGTM! Thanks Tex!

PR #4982, opened 2023-01-31
Fix lower type crash on vector array parameter.

python3kgae:
To allow vector array parameter for extern functions.

When the parameter used as argument for call.
Just bitcast it to the argument type.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/shader_targets/library/vec_array_param.hlsl
+++ b/tools/clang/test/HLSLFileCheck/shader_targets/library/vec_array_param.hlsl
@@ -1,7 +1,7 @@
 // RUN: %dxc -T lib_6_6 %s | FileCheck %s
 
 // Make sure vector array param works.
-// CHECK:%[[Cast:.+]] = bitcast [3 x [3 x float]]* %0 to [3 x <3 x float>]*
+// CHECK:%[[Cast:.+]] = bitcast [9 x float]* %{{.+}} to [3 x <3 x float>]*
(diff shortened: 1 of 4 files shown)


tools/clang/test/HLSLFileCheck/shader_targets/library/vec_array_param.hlsl
```diff
@@ -0,0 +1,11 @@
+// RUN: %dxc -T lib_6_6 %s | FileCheck %s
+
+// Make sure vector array param works.
+// CHECK:call float @"\01?foo@@YAMY02V?$vector@M$02@@@Z"([3 x <3 x float>]* %{{.+}})
```
llvm-beanz, line 4: This test doesn’t actually verify the correct transformation occurs, it just verifies we don’t hit the assert.

We should test things like this using IR tests either through `dxopt` or `opt`.

python3kgae: Fixed.

tools/clang/test/HLSLFileCheck/shader_targets/library/vec_array_param.hlsl
```diff
@@ -0,0 +1,11 @@
+// RUN: %dxc -T lib_6_6 %s | FileCheck %s
+
+// Make sure vector array param works.
```
pow2clk, line 3: Could we have a check for the bitcast operator here?

python3kgae: Sure.

pow2clk (approved):
Looks good!

PR #4984, opened 2023-01-31
Introduce issue templates

### LINKED ISSUE #4975: Add an issue template
Hello!

Regarding recent issues, the amount of information provided by our users vary a lot, and we often miss crucial information.
Would it be possible to add a default template for issues, guiding users to fill a good issue description?
And maybe a workflow to tag them as "need triage", so we quickly know which ones need triage/assign/work.

IMO, bug reports should contains:

- error
- DXC version, git hash
- commandline
- sample shader reproducing the issue

or a godbolt link, some users do use that and it's fine.

Example template we could use:

````
Hi!

I encountered this bug/error:

```
the error message
```

DXC version: `v1.2.3 (or hash 09abcdef if custom build)`
command: `dxc.exe -T ps_6_0 -some-flag shader.hlsl`

shader:

```hlsl
void ps_main(int arg) {
    some_code_reproducing_error();
}
```

````

### PR DESCRIPTION
**Pull request by llvm-beanz**
This change introduces three new issue templates for DXC. The three issue templates are:

* feature_request - Issues filed with this template will get automatically tagged with the `enhancement` label
* bug_report - Issues filed with this template will get automatically get tagged with the `bug` label
* bug_report_spirv - Issues filed with this template will get automatically tagged with the `bug` and `spirv` labels

The bug_report and bug_report_spirv templates are otherwise identical, but could diverge to better meet the needs of the supporting teams.

Fixes #4975.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,25 @@
+---
+name: Bug Report
+about: Report an issue with DXC
+title: ''
+labels: ['bug']
+assignees: ''
+
+---
+
+**Description**
+<!--- Please provide a few sentences describing the issue you encountered. --->
+
+**Steps to Reproduce**
+<!--- Provide a description of how to reproduce the error. If possible please 
+provide source and tool command line options. If the issue reproduces on
+Compiler Explorer (https://godbolt.org/) or Shader Playground
+(https://shader-playground.timjones.io/) please provide a link. --->
+
+
+**Actual Behavior**
+<!--- Please provide error output or a description of the observed issue. --->
+
+**Environment**
+- DXC version <!-- replace with the output of 'dxc --version' -->
+- Host Operating System <!--- Host operating system and version --->
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -14,7 +14,9 @@ assignees: ''
 <!--- Provide a description of how to reproduce the error. If possible please 
 provide source and tool command line options. If the issue reproduces on
 Compiler Explorer (https://godbolt.org/) or Shader Playground
-(https://shader-playground.timjones.io/) please provide a link. --->
+(https://shader-playground.timjones.io/) please provide a link. If the source is
+split across multiple files, please preprocess into a single file using DXC's 
+command line `-P -Fi <path>`. --->
 
 
 **Actual Behavior**
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug Report
 about: Report an issue with DXC
 title: ''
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: .github/ISSUE_TEMPLATE/bug_report.md [hunk 1]
```diff
@@ -0,0 +1,25 @@
+---
+name: Bug Report
+about: Report an issue with DXC
+title: ''
+labels: ['bug']
```
Keenuts, line 5: shall we add a "needs triage" label by default on all templates and clear them if one of us have interacted with it?

Keenuts (commented):
Personal nit/opinion: maybe we should be more strict and short in the wording on the "commandline + shader" part:
 - the longer the higher the change parts will be missed.
 - without those reproducing or fixing the issue can be quite tricky.

What about a short:
"Please provide the a command line and a shader we can use to reproduce the error. " + the part about godbolt/shader-playground after.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,25 @@
...
+assignees: ''
+
+---
+
+**Description**
+<!--- Please provide a few sentences describing the issue you encountered. --->
+
+**Steps to Reproduce**
+<!--- Provide a description of how to reproduce the error. If possible please 
+provide source and tool command line options. If the issue reproduces on
```
dmpots, line 15: Might be worth asking for preprocessed source and specifiying how to generate it (e.g. `/P`) since a preprocessed source can be easier to work with than a zip full of shader/header files.

dmpots (approved):
LGTM

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,23 @@
...
+labels: ['enhancement', 'needs-triage']
+assignees: ''
+
+---
+
+<!--- If this is a language feature request please file the issue against the
+hlsl-specs repository: https://github.com/microsoft/hlsl-specs/issues/new --->
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
pow2clk, line 14: I think we should push more for the detailed description of the problem itself. I would remove the question and say something like "Describe the problem in as much detail that led you to request this feature". 

I don't really want concise. Very often feature requests that we just can't do can be solved another way if we understood the problem.

pow2clk (commented):
I think suggesting linking related PRs and issues would be useful at the bottom of each of these

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Bug Report
+about: Report an issue with DXC
+title: ''
+labels: ['bug', 'needs-triage']
```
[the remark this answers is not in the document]
llvm-beanz: This should work. I believe the automated tagging through templates isn't gated by the permissions to edit tags on an issue.

Keenuts (approved):

PR #4987, opened 2023-02-01
Fix assert when a function been called more than once by entry.

python3kgae:
For better support for noinline.

If a function is processed, just skip in DxilViewIdStateBuilder::ComputeReachableFunctionsRec.

### DIFF
--- a/lib/HLSL/ComputeViewIdStateBuilder.cpp
+++ b/lib/HLSL/ComputeViewIdStateBuilder.cpp
@@ -285,6 +285,8 @@ void DxilViewIdStateBuilder::ComputeReachableFunctionsRec(CallGraph &CG, CallGra
   Function *F = pNode->getFunction();
   // Accumulate only functions with bodies.
   if (F->empty()) return;
+  if (FuncSet.count(F))
+    return;
   auto itIns = FuncSet.emplace(F);
   DXASSERT_NOMSG(itIns.second);
   (void)itIns;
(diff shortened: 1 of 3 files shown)


pow2clk (approved):
Looks good. Can this not be triggered from HLSL? It's fine if not, we want to have more unit testy unit tests. I'm just curious

python3kgae:
> Looks good. Can this not be triggered from HLSL? It's fine if not, we want to have more unit testy unit tests. I'm just curious

It sure can. I thought our new style for test is opt only. :)

llvm-beanz (commented):
I think the gap in test coverage here is that we're not actually verifying that the changed code behaves correctly, we're verifying it doesn't crash/assert.

We should probably have a printer for the View Id state data, and have that connected to a printing pass that we can run to verify the correct result.

python3kgae:
> I think the gap in test coverage here is that we're not actually verifying that the changed code behaves correctly, we're verifying it doesn't crash/assert.
> 
> We should probably have a printer for the View Id state data, and have that connected to a printing pass that we can run to verify the correct result.

Sure. I can do that in another PR.

llvm-beanz:
> Sure. I can do that in another PR.

If you wan tot do that in a separate PR, that's fine, but this PR should be based on that functionality.

You're changing behavior in this PR that you can't accurately test. So we should make sure the required testing support is available before you change the behavior.

python3kgae:
Redo in https://github.com/microsoft/DirectXShaderCompiler/pull/4992  with dump test.

PR #4989, opened 2023-02-01
Support this cast to base class.

python3kgae:
Copy will be generated from this to tmp base class ptr.

### DIFF
--- a/tools/clang/lib/CodeGen/CGExprAgg.cpp
+++ b/tools/clang/lib/CodeGen/CGExprAgg.cpp
@@ -741,10 +741,13 @@ void AggExprEmitter::VisitCastExpr(CastExpr *E) {
         LValue LV;
         if (DeclRefExpr *SrcDecl = dyn_cast<DeclRefExpr>(Src))
           LV = CGF.EmitLValue(SrcDecl);
-        else if (ArraySubscriptExpr *ArraySubExpr = dyn_cast<ArraySubscriptExpr>(Src))
+        else if (ArraySubscriptExpr *ArraySubExpr =
+                     dyn_cast<ArraySubscriptExpr>(Src))
           LV = CGF.EmitLValue(ArraySubExpr);
         else if (ParenExpr *parenExpr = dyn_cast<ParenExpr>(Src))
           LV = CGF.EmitLValue(parenExpr->getSubExpr());
+        else if (isa<CXXThisExpr>(Src))
+          LV = CGF.EmitLValue(Src);
         else
           LV = CGF.EmitAggExprToLValue(Src);
 
--- a/tools/clang/lib/CodeGen/CGExprAgg.cpp
+++ b/tools/clang/lib/CodeGen/CGExprAgg.cpp
@@ -741,8 +741,7 @@ void AggExprEmitter::VisitCastExpr(CastExpr *E) {
         LValue LV;
         if (DeclRefExpr *SrcDecl = dyn_cast<DeclRefExpr>(Src))
           LV = CGF.EmitLValue(SrcDecl);
-        else if (ArraySubscriptExpr *ArraySubExpr =
-                     dyn_cast<ArraySubscriptExpr>(Src))
+        else if (ArraySubscriptExpr *ArraySubExpr = dyn_cast<ArraySubscriptExpr>(Src))
           LV = CGF.EmitLValue(ArraySubExpr);
         else if (ParenExpr *parenExpr = dyn_cast<ParenExpr>(Src))
           LV = CGF.EmitLValue(parenExpr->getSubExpr());
--- a/tools/clang/lib/Sema/SemaOverload.cpp
+++ b/tools/clang/lib/Sema/SemaOverload.cpp
@@ -4966,8 +4966,10 @@ InitCallParamConversions(Sema &S, const FunctionProtoType *Proto,
     }
     Expr *OutFrom = DeclRefExpr::Create(
         S.getASTContext(), NestedNameSpecifierLoc(), SourceLocation(), Param,
-        true, Param->getLocation(), ParamType.getNonReferenceType(), VK_RValue, nullptr);
-    OutConversion = TryCopyInitialization(S, OutFrom, ParamType, SuppressUserConversions,
+        true, Param->getLocation(), ParamType.getNonReferenceType(), VK_RValue,
+        nullptr);
+    OutConversion = TryCopyInitialization(
+        S, OutFrom, Arg->getType(), SuppressUserConversions,
         InOverloadResolution, false, AllowExplicit);
   }
 }
--- a/tools/clang/lib/Sema/SemaOverload.cpp
+++ b/tools/clang/lib/Sema/SemaOverload.cpp
@@ -4967,8 +4967,7 @@ InitCallParamConversions(Sema &S, const FunctionProtoType *Proto,
     Expr *OutFrom = DeclRefExpr::Create(
         S.getASTContext(), NestedNameSpecifierLoc(), SourceLocation(), Param,
         true, Param->getLocation(), ParamType.getNonReferenceType(), VK_RValue, nullptr);
-    OutConversion = TryCopyInitialization(
-        S, OutFrom, Arg->getType(), SuppressUserConversions,
+    OutConversion = TryCopyInitialization(S, OutFrom, ParamType, SuppressUserConversions,
         InOverloadResolution, false, AllowExplicit);
   }
 }
--- a/tools/clang/lib/Sema/SemaOverload.cpp
+++ b/tools/clang/lib/Sema/SemaOverload.cpp
@@ -4966,8 +4966,7 @@ InitCallParamConversions(Sema &S, const FunctionProtoType *Proto,
     }
     Expr *OutFrom = DeclRefExpr::Create(
         S.getASTContext(), NestedNameSpecifierLoc(), SourceLocation(), Param,
-        true, Param->getLocation(), ParamType.getNonReferenceType(), VK_RValue,
-        nullptr);
+        true, Param->getLocation(), ParamType.getNonReferenceType(), VK_RValue, nullptr);
     OutConversion = TryCopyInitialization(
         S, OutFrom, Arg->getType(), SuppressUserConversions,
         InOverloadResolution, false, AllowExplicit);
--- a/tools/clang/test/HLSLFileCheck/hlsl/classes/this_cast_to_base_class.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/classes/this_cast_to_base_class.hlsl
@@ -1,5 +1,36 @@
 // RUN: %dxc -T lib_6_6 -HV 2021 -fcgl %s | FileCheck %s
 
+// CHECK-LABLE: define linkonce_odr void @"\01?foo@Child@@QAAXXZ"(%class.Child* %this)
+// CHECK: %[[OutArg:.+]] = alloca %class.Parent
+// CHECK: %[[OutThisPtr:.+]] = getelementptr inbounds %class.Child, %class.Child* %this, i32 0, i32 0
+// CHECK: call void @"\01?lib_func2@@YAXVParent@@@Z"(%class.Parent* %[[OutArg]])
+// Make sure copy-out.
+// CHECK: %[[OutThisCpyPtr:.+]] = bitcast %class.Parent* %[[OutThisPtr]] to i8*
+// CHECK: %[[OutArgCpyPtr:.+]] = bitcast %class.Parent* %[[OutArg]] to i8*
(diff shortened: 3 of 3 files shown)


llvm-beanz:
Should this code be valid? In C++ implicit cast-to-base is only valid for pointers not for value types.

python3kgae:
> Should this code be valid? In C++ implicit cast-to-base is only valid for pointers not for value types.

I thought 'this' is a "reference" for hlsl. :(

llvm-beanz:
> I thought 'this' is a "reference" for hlsl. :(

Sure, but the parameter is a value parameter. So it gets copied.

llvm-beanz:
FWIW, under HLSL 2018, the code in this test case is probably valid because of the old UDT casting rules that allowed structs of the same layout to be implicitly casted. It would not be valid under HLSL 2021+.

mstrgram:
If `this` is to be considered a value type, I assume all objects passed to methods are to have value semantics. If that is the case, shouldn't something like below also fail to compile then?

``` 
struct A
{
    uint _member;
    
    bool SomeFunc(A other)
    {
        return _member == other._member;
    }
};

struct B : A
{
    uint _extraMember;
};

A a = { 10 };
B b = { 11, 12 };
bool same = a.SomeFunc(b);
```

python3kgae:
> FWIW, under HLSL 2018, the code in this test case is probably valid because of the old UDT casting rules that allowed structs of the same layout to be implicitly casted. It would not be valid under HLSL 2021+.

I thought HLSL 2021 is closer to C++.

But C++ is fine for code like this where reference used for a value parameter.
```
struct A {
    float a;
};

struct B : public A {
   int b;
};


float foo(A a) {
    return a.a;
}

float bar(B &b) {
    return foo(b);
}
```

llvm-beanz:
> But C++ is fine for code like this where reference used for a value parameter.
> 
> ```
> struct A {
>     float a;
> };
> 
> struct B : public A {
>    int b;
> };
> 
> 
> float foo(A a) {
>     return a.a;
> }
> 
> float bar(B &b) {
>     return foo(b);
> }
> ```

The way this works in C++ is slightly unintuitive. Because `A` has implicit copy constructors what ends up happening is that a new `A` gets constructed from a `B&` casted up to an `A&`. This only works because the constructors are present.

If you add `A(const A&) = delete;` to `A`, the code fails.

Since we don't have constructors in HLSL, this code only works because we've defined a truncating cast 🤮 .

mstrgram:
> We should add some additional tests to verify parameter passing semantics are correctly observed.
>
> Specifically copy-in/copy-out for this objects implicitly casted to bases. We should make sure that the truncated base object is what is copied in all cases when an HLSLDerivedToBase cast is present.

Does this mean you agree the original `this` cast to base and the implicit `B` --> `A` param cast I showed should be valid?

llvm-beanz:
> Does this mean you agree the original `this` cast to base and the implicit `B` --> `A` param cast I showed should be valid?

Yes, but keep in mind this is a truncating value cast. When used in a parameter it amounts to copying the base class data out of the derived object.

When used with `inout` it should result in copying the base data out, then copying it back at the end of the function, and when used with an `out` parameter it would result in only initializing the base class portion of `this`.

llvm-beanz (commented):
Your test cases here are revealing that we're not eliding copy-in semantics in some places we could safely do so. That said, we can resolve that separately. This code generation is sub-optimal but correct.

A few minor notes inline.

tools/clang/test/HLSLFileCheck/hlsl/classes/this_cast_to_base_class.hlsl
```diff
@@ -1,5 +1,36 @@
...
+// CHECK: %[[OutArgCpyPtr:.+]] = bitcast %class.Parent* %[[OutArg]] to i8*
+// CHECK: call void @llvm.memcpy.p0i8.p0i8.i64(i8* %[[OutThisCpyPtr]], i8* %[[OutArgCpyPtr]], i64 8, i32 1, i1 false)
+
+// CHECK-LABLE: define linkonce_odr void @"\01?bar@Child@@QAAXXZ"(%class.Child* %this)
+// CHECK: %[[InOutArg:.+]] = alloca %class.Parent
+// CHECK: %[[InOutThisPtr:.+]] = getelementptr inbounds %class.Child, %class.Child* %this, i32 0, i32 0
+
+// Skip lifetime marker
+// CHECK: %[[InOutArgPtrForMarker:.+]] = bitcast %class.Parent* %[[InOutArg]] to i8*
+// CHECK: call void @llvm.lifetime.start(i64 8, i8* %[[InOutArgPtrForMarker]])
```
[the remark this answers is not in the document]
python3kgae: It will make an extra bitcast, one solution could be disable lifetime marker.

tools/clang/test/HLSLFileCheck/hlsl/classes/this_cast_to_base_class.hlsl
```diff
@@ -33,13 +64,21 @@ class Parent
 };
 
 int lib_func(Parent obj, int a, int b);
+void lib_func2(out Parent obj);
+void lib_func3(inout Parent obj);
 
 class Child : Parent
 {
     int foo(int a, int b)
```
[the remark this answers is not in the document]
python3kgae: Done.

tools/clang/test/HLSLFileCheck/hlsl/classes/this_cast_to_base_class.hlsl
```diff
@@ -0,0 +1,93 @@
+// RUN: %dxc -T lib_6_6 -HV 2021 -disable-lifetime-markers -fcgl %s | FileCheck %s
```
llvm-beanz, line 1: Lifetime markers shouldn’t impact the test. If it does, your test matches are too fragile. You should be able to use a `CHECK-DAG` to work around the `bitcast` instructions inserted with lifetime markers.

tools/clang/lib/Sema/SemaOverload.cpp
```diff
@@ -4967,8 +4967,7 @@ InitCallParamConversions(Sema &S, const FunctionProtoType *Proto,
     Expr *OutFrom = DeclRefExpr::Create(
         S.getASTContext(), NestedNameSpecifierLoc(), SourceLocation(), Param,
         true, Param->getLocation(), ParamType.getNonReferenceType(), VK_RValue, nullptr);
-    OutConversion = TryCopyInitialization(
-        S, OutFrom, Arg->getType(), SuppressUserConversions,
+    OutConversion = TryCopyInitialization(S, OutFrom, ParamType, SuppressUserConversions,
```
llvm-beanz, line 4970: I’m not sure this change is correct. This alters the AST so that instead of converting the return value from Parent->Child, it converts Parent->Parent.

llvm-beanz: It also looks like we don't support `HLSLDerivedToBase` truncation for `inout` or `out` on other parameters (https://godbolt.org/z/aqaKK1qTd), so we don't need to for `this` either.

python3kgae: Reverted.

mstrgram: Isn't it somewhat strange to have different cast behavior depending on whether the var is `in` only or `inout`/ `out`?

python3kgae: Yeah, it is strange.
Do you want to always require cast or always implicit cast for case like this?

llvm-beanz: This code is wonky, but I think correct.

For `in` and `inout` parameters we copy initialize a new object from the argument type to the parameter type.

Then for `out` and `inout` parameters we copy initialize from the parameter to the argument type.

The trick here is that `InConversion` and `OutConversion` are two different casts, and they can both be set.

This should correctly model the semantics that `in` and `inout` parameters need to be casted going in, and that `out` and `inout` need to be casted coming out.

python3kgae: What's the point to cast to bigger type when the copy has to be truncated?

If cast to parameter type, the copy will be on smaller type which is just normal copy.

llvm-beanz: It isn't really a cast. It is copy-initializing.

On `in` and `inout` parameters the argument (value being passed in) is copy-initialized to a temporary of the parameter type, which is passed into the function. For `out` parameters the temporary is uninitialized.
On `out` and `inout` parameters after the call, the temporary value of the parameter-type is copy-initialized back to the argument variable.

So the type conversion go from argument -> parameter on input and from parameter -> argument on output, but these are copy-initializations not casts.

tools/clang/lib/Sema/SemaOverload.cpp
```diff
@@ -4966,7 +4966,8 @@ InitCallParamConversions(Sema &S, const FunctionProtoType *Proto,
     }
     Expr *OutFrom = DeclRefExpr::Create(
         S.getASTContext(), NestedNameSpecifierLoc(), SourceLocation(), Param,
-        true, Param->getLocation(), ParamType.getNonReferenceType(), VK_RValue, nullptr);
+        true, Param->getLocation(), ParamType.getNonReferenceType(), VK_RValue,
+        nullptr);
```
llvm-beanz, line 4970: Seems odd to have a formatting change to unrelated code here.

tools/clang/lib/CodeGen/CGExprAgg.cpp
```diff
@@ -741,10 +741,13 @@ void AggExprEmitter::VisitCastExpr(CastExpr *E) {
         LValue LV;
         if (DeclRefExpr *SrcDecl = dyn_cast<DeclRefExpr>(Src))
           LV = CGF.EmitLValue(SrcDecl);
-        else if (ArraySubscriptExpr *ArraySubExpr = dyn_cast<ArraySubscriptExpr>(Src))
+        else if (ArraySubscriptExpr *ArraySubExpr =
+                     dyn_cast<ArraySubscriptExpr>(Src))
```
tex3d, line 745: Is this formatting-only change on purpose?

tools/clang/test/HLSLFileCheck/hlsl/classes/this_cast_to_base_class.hlsl
```diff
@@ -0,0 +1,93 @@
...
+void lib_func3(inout Parent obj);
+
+class Child : Parent
+{
+    int foo(int a, int b)
+    {
+        return lib_func(this, a, b);
+    }
+    void foo() {
+        lib_func2((Parent)this);
```
tex3d, line 69: Is explicit cast here required for `out` param?  Same for `inout` case in lib_func3 call?

tools/clang/lib/CodeGen/CGExprAgg.cpp
python3kgae: No. I'll revert it.

tools/clang/lib/Sema/SemaOverload.cpp
python3kgae: Will fix it.

tools/clang/test/HLSLFileCheck/hlsl/classes/this_cast_to_base_class.hlsl
python3kgae: This follows fxc which require explicit cast here.
We could change it in TryCopyInitialization if it is OK to allow implicit cast.

llvm-beanz: Ooof... These casts probably cause an mis-compile.

Unless there is some magic and unexplainable casting rule applying here, in these cases the value getting passed in _should_ be the rvalue produced by the cast. They shouldn't result in modifying the underlying value, and we _should_ error on passing rvalues into `out` and `inout` parameters.

llvm-beanz: Okay... another sad face from me. This does _seem_ to be the language rules:

https://godbolt.org/z/z47Mqh4fa

Basically, a c-style derived-to-base cast is defined to be some special intuited magic.

tools/clang/test/HLSLFileCheck/hlsl/classes/this_cast_to_base_class.hlsl
llvm-beanz: I looked at this with @python3kgae. Because the lifetime markers generate identical instructions to the instruction we need to match there isn't a good way to match around it. Disabling lifetime markers simplifies the test, so that's probably the best approach here.

llvm-beanz (commented):
I think once the formatting fixes are in, this is good.

tex3d (approved):

PR #4992, opened 2023-02-02
Fix assert when a function been called more than once by entry.

python3kgae:
For better support of noinline.

If a function is processed, just skip in DxilViewIdStateBuilder::ComputeReachableFunctionsRec.

Added ViewIdStatePrinter pass to test the change.


lib/HLSL/ComputeViewIdStateBuilder.cpp
```diff
@@ -925,3 +927,47 @@ ModulePass *createComputeViewIdStatePass() {
 }
 
 } // end of namespace llvm
+
+namespace {
+class ViewIdStatePrinter : public ModulePass {
```
pow2clk, line 932: I'm not very familiar with these printer passes, but it looks to me that we have very few of them explicitly specified. Instead, if you add a `print()` method to the `ComputeViewIdState` pass and invoke `opt` with `-viewid-state -analyze`, I think the `opt.cpp` code will invoke your print function using the `ModulePassPrinter` class

pow2clk (commented):
It looks good, but I fear we might be going against the "usual way" of doing things. I hate to trip you up, but I think it's just a matter of moving the code you've written, which looks fine into a different function of the ComputeViewIdState pass

lib/HLSL/ComputeViewIdStateBuilder.cpp
python3kgae: Never know that. New things learned.
Fixed.

pow2clk (approved):
Looks good!

llvm-beanz:
The tests in this PR only run under lit. We cannot add tests that only run under lit until we remove support for the non-LIT based test workflows.

@python3kgae, can you please convert these to non-lit tests?

PR #4993, opened 2023-02-03
Enable lit by default.

python3kgae:
Enable lit by default in hctbuild.cmd and hcttest.cmd.
 -enable-lit option is replaced by -disable-lit which switch back to batch.

### DIFF
--- a/utils/hct/hctbuild.cmd
+++ b/utils/hct/hctbuild.cmd
@@ -42,6 +42,7 @@ set SPV_TEST=OFF
 set DXILCONV=ON
 set DXC_CMAKE_SYSTEM_VERSION=
 set SHOW_CMAKE_LOG=0
+set ENABLE_LIT=ON
 set WINSDK_MIN_VERSION=10.0.17763.0
 
 :parse_args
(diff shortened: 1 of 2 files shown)


llvm-beanz:
Can you elaborate on what specifically is different in the TAEF output that breaks AppVeyor?

I don't necessarily think we should spend time fixing it rather than migrating off AppVeyor, but it would be good to understand the difference, and if the difference is trivial (or might impact other users) we may need to resolve it.

tex3d (approved):

python3kgae:
> Can you elaborate on what specifically is different in the TAEF output that breaks AppVeyor?
> 
> I don't necessarily think we should spend time fixing it rather than migrating off AppVeyor, but it would be good to understand the difference, and if the difference is trivial (or might impact other users) we may need to resolve it.

Tried to repro the issue locally but failed. :(
Don't know how to setup the environment same as appveyor.

python3kgae:
> Okay. We can't make lit the default behavior until we understand that issue. Please investigate.

https://github.com/microsoft/DirectXShaderCompiler/issues/4997 created to track the issue.

PR #4994, opened 2023-02-03
Update Azure pipelines to report results

llvm-beanz:
I didn't have this configured correctly to run on failing builds... obviously that is where this is more useful.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -128,4 +128,5 @@ stages:
       inputs:
         testResultsFormat: 'XUnit'
         testREsultsFiles: '**/testresults.xunit.xml'
+      condition: succeededOrFailed()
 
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -128,4 +128,5 @@ stages:
       inputs:
         testResultsFormat: 'XUnit'
         testREsultsFiles: '**/testresults.xunit.xml'
```
dmpots, line 130: Obviously not related to your change, but it looks like there is a case typo here. Might not matter if its case-insensitive.

dmpots (approved):

azure-pipelines.yml
llvm-beanz: lol... I'll fix that too

hekota (approved):
LGTM!

PR #5002, opened 2023-02-08
Fix two issues found in our internal build

s-perron:
1. The StringRef in the `stringLiterals` in the `SpirvBuilder` can have a
dangling reference causing undefined behaviour.
2. Two symbols can have the same name, so when we sort a vector of symbols by
name, we want to use std::stable_sort to make sure we get deterministic
behaviour.


cassiebeckley (approved):
LGTM

PR #5004, opened 2023-02-09
Allow libdxcompiler.so and dxc to find libdxil.so

pow2clk:
When present in the library search path, linux builds of libdxcompiler.so and the dxc executable should load libdxil.so and use it for validation and signing. However, the code to do that was still searching for dxil.dll even on these platforms. This adds code to search for the right name on the right platforms.

### DIFF
--- a/include/dxc/Support/dxcapi.use.h
+++ b/include/dxc/Support/dxcapi.use.h
@@ -16,6 +16,9 @@
 
 namespace dxc {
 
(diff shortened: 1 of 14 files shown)


llvm-beanz (approved):
Other than my dislike of Hungarian notation... this looks good to me.

pow2clk:
The k- prefix is included in all other such string constants. Just trying to be consistent! 😄

PR #5006, opened 2023-02-09
Just load dxcompiler on non-win platforms

pow2clk:
There is thought to be a benefit to both dynamically linking and loading the dxcompiler library on Windows. However no such benefit is expected for other platforms and it just makes us ship two forms of the library.

By only linking on Windows, we can ship just the unversioned named library for non-win platforms and still function correctly.

### DIFF
--- a/tools/clang/unittests/HLSL/CMakeLists.txt
+++ b/tools/clang/unittests/HLSL/CMakeLists.txt
@@ -86,16 +86,11 @@ if (WIN32)
 target_link_libraries(ClangHLSLTests PRIVATE
   dxcompiler
   HLSLTestLib
-  LLVMDxilContainer
   LLVMDxilDia
   ${TAEF_LIBRARIES}
-  ${DIASDK_LIBRARIES}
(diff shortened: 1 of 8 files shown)


tools/clang/unittests/HLSL/CMakeLists.txt
```diff
@@ -104,8 +99,6 @@ if(WIN32)
 # Add includes for platform helpers and dxc API.
 include_directories(${TAEF_INCLUDE_DIRS})
 include_directories(${DIASDK_INCLUDE_DIRS})
-include_directories(${D3D12_INCLUDE_DIRS})
-include_directories(${LLVM_MAIN_INCLUDE_DIR}/dxc/Test)
```
pow2clk, line 108: Lots of incidentals here. I noticed more differences than should exist between the platforms given the support on both currently. So I took them out.

pow2clk:
Not the right approach for the ultimate goal. turns out cmake operations are badly named. who knew?

PR #5007, opened 2023-02-09
Support dumping implicit top-level decls

llvm-beanz:
This change reworks how we were filtering top-level implicit decls from the `-ast-dump` output. The change has a couple of key differences, but the behavior of the `-ast-dump` flag remains unchanged.

The differences in this change are:
* By default Decl::dump _will_ dump implicit top-level decls which allows attached debuggers to dump top-level decls.
* Top-level implicit decls are determined by the decl's `Implicit` bit rather than an invalid source location.
* Disabling dumping implicit decls is now a language option, so it can be toggled.
* A new `-ast-dump-implicit` flag is added to dump the AST including implicit top-level decls.

The attached test case verifies that the `-ast-dump` flag does not dump implicit top-level decls, and it is intended to be strict enough that if we add new top-level decls in the future and fail to mark them as `implicit` the test will fail.

The attached test also trys to fill some gaps in testing around the lazy decl initialization by verifying that incomplete implicit decls are generated, and that a used decl gets completed as expected.

### DIFF
--- a/tools/clang/lib/AST/ASTDumper.cpp
+++ b/tools/clang/lib/AST/ASTDumper.cpp
@@ -118,6 +118,8 @@ namespace  {
 
     bool ShowColors;
 
+    bool DumpImplicitTopLevelDecls = true; // HLSL Change
+
     /// Dump a child of the current node.
     template<typename Fn> void dumpChild(Fn doDumpChild) {
       // If we're at the top level, there's nothing interesting to do; just
@@ -209,6 +211,10 @@ namespace  {
       : OS(OS), Traits(Traits), SM(SM), TopLevel(true), FirstChild(true),
         LastLocFilename(""), LastLocLine(~0U),
         ShowColors(ShowColors) { }
+
+    // HLSL Change Start - filter implicit decls.
+    void setDumpImplicitTopLevelDecls(bool B) { DumpImplicitTopLevelDecls = B; }
+    // HLSL Change End - filter implicit decls.
 
     void dumpDecl(const Decl *D);
     void dumpStmt(const Stmt *S);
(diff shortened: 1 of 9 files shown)


tools/clang/lib/AST/ASTDumper.cpp
```diff
@@ -118,6 +118,8 @@ namespace  {
 
     bool ShowColors;
 
+    bool DumpImplicitTopLevelDecls = true; // HLSL Change
```
python3kgae, line 121: Why initialize to true here?

dmpots (approved):
LGTM. Retains the existing flag behavior so the big difference in default behavior will be in the debugger. I don't know whether it makes it nicer to dump the implicit ones by default in the debugger, but it seems reasonable. Should be able to suppress the dumps by modifying the value in the debugger so we still have a way back to the old behavior.

tools/clang/lib/AST/ASTDumper.cpp
llvm-beanz: So that the default behavior of the dumper is to always dump top-level decls. Then it only ends up false if you explicitly specify false.

python3kgae: I see.
That sounds reasonable.

tools/clang/lib/AST/ASTDumper.cpp
```diff
@@ -1072,12 +1078,10 @@ void ASTDumper::dumpHLSLUnusualAnnotations(const ArrayRef<hlsl::UnusualAnnotatio
 
 void ASTDumper::dumpDecl(const Decl *D) {
   // HLSL Change Starts: Don't display decls with invalid SourceLocations.
-  if (D && D->getDeclContext() &&
+  if (!DumpImplicitTopLevelDecls && D->getDeclContext() &&
```
tex3d, line 1081: Can we move the top-level implicit decl skip up to `dumpDeclContext` instead?  Then I think it simplifies things, doesn't it?  We no longer have to default `DumpImplicitTopLevelDecl` to `true` in order to make `dumpDecl` work under the debugger, for instance.

tex3d (approved):
LGTM

tex3d:
>

PR #5009, opened 2023-02-09
Simplify header/link dependencies for unittests

pow2clk:
After execution tests moved into their own place, a lot of the libraries and headers that unittests required weren't necessary anymore. This pares those requirements down as far as possible


hekota (approved):
LGTM! Thanks for the cleanup!

PR #5010, opened 2023-02-09
Remove SO version from unix binaries

llvm-beanz:
This is a bit hacky. We don't want to have SO versions but we do want SO names. This makes it a bit wonky since CMake doesn't support this usage.


pow2clk (approved):
cmake is never as kind to you as you hope it will be 🤣

PR #5013, opened 2023-02-10
debugging appveyor's aversion to lit

pow2clk:


pow2clk:
Well it timed out, but the dialog indicated that it started running the TAEF tests without erroring immediately. The problem we have here is that the lit testing is ignoring the -spirv_only flag and testing everything, which is why we are timing out. It's arguable that the spirv_only and powershell based TAEF testing is redundant to lit at this point in favor of just going all in on lit

PR #5016, opened 2023-02-10
Don't add template record decl to declcontext

llvm-beanz:
CXXRecordDecls that are part of ClassTemplateDecls should not be added to the DeclContext that the ClassTemplateDecl gets added to.

I'm a little surprised this didn't cause problems somewhere, but it does result in the decl being dumped twice in the ast dumping. I think it should have also resulted in conflicting decls in name lookup, but there may be some other change somewhere that worked around that, or it might get ignored since the decl is invalid by itself.

Either way, we can skip adding CXXRecordDecls for template types to the DeclContext and everything works fine.


tex3d (approved):

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -850,7 +851,9 @@ CXXMethodDecl* hlsl::CreateObjectFunctionDeclarationWithParams(
                                            parmVarDecls.size());
   }
 
-  recordDecl->addDecl(functionDecl);
+  // If this is going to be part of a template decl, don't add it to the record.
+  if (!isTemplate)
+    recordDecl->addDecl(functionDecl);
```
tex3d, line 856: So `isTemplate` describes whether the object itself is a template?  Is there a way to tell this from the recordDecl?

llvm-beanz: I should clarify the naming. It is if the function is going to be part of a FunctionTemplateDecl. In that case we need to not add it to the record since the FunctionTemplateDecl will be added to the Record and the Method is a child of the FunctionTemplateDecl.

tex3d (approved):

PR #5017, opened 2023-02-10
[SPIR-V] Fix composites with empty struct fields.

LukasBanana:
`indexAST` must be in the half-open range [0, N) instead of [0, N]. This caused a crash while I was fixing other issues with empty structs. I will file separate PRs for these fixes.

### DIFF
--- a/tools/clang/lib/SPIRV/InitListHandler.cpp
+++ b/tools/clang/lib/SPIRV/InitListHandler.cpp
@@ -366,13 +366,15 @@ InitListHandler::createInitForStructType(QualType type, SourceLocation srcLoc,
     while (tryToSplitConstantArray())
       ;
 
-    auto init = initializers.back();
-    // We can only avoid decomposing and reconstructing when the type is
-    // exactly the same.
-    if (type.getCanonicalType() ==
-        init->getAstResultType().getCanonicalType()) {
-      initializers.pop_back();
-      return init;
+    if (!initializers.empty()) {
+      auto init = initializers.back();
+      // We can only avoid decomposing and reconstructing when the type is
+      // exactly the same.
+      if (type.getCanonicalType() ==
+          init->getAstResultType().getCanonicalType()) {
+        initializers.pop_back();
+        return init;
+      }
     }
 
     // Otherwise, if the next initializer is a struct, it is not of the same
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -556,7 +556,7 @@ uint32_t getFieldIndexInStruct(const StructType *spirvStructType,
       getNumBaseClasses(astStructType) + fieldDecl->getFieldIndex();
 
   const auto &fields = spirvStructType->getFields();
-  assert(indexAST <= fields.size());
+  assert(indexAST < fields.size());
   return fields[indexAST].fieldIndex;
 }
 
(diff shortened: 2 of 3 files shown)


llvm-beanz:
Hey @LukasBanana, thank you for the fix. Can you please provide a test case?

LukasBanana:
I understand that test cases are the preferred way for PRs and I try to be better about, but in this case, the bug can be identified right in the next line. Indexing into an array/vector within `[0, N]` is not the proper index range.

llvm-beanz:
Our contribution guidelines require test case for PRs. This PR cannot be merged without a test:
https://github.com/microsoft/DirectXShaderCompiler/blob/main/CONTRIBUTING.md

I understand that the test case may be trivial to derive, and the fix may be obvious. Neither of those factors justify merging a change without a test case.

LukasBanana:
I might just consolidate this change with my other changes to fix multiple crashes. I will follow up on this PR next week.

LukasBanana:
@llvm-beanz I added the remaining changes that fixes the SPIR-V backend for the test file I added. Empty structs were not handled correctly but they are becoming more common with the advent of templates.

tools/clang/lib/SPIRV/InitListHandler.cpp
```diff
@@ -385,6 +385,8 @@ InitListHandler::createInitForStructType(QualType type, SourceLocation srcLoc,
   llvm::SmallVector<SpirvInstruction *, 4> fields;
   const RecordDecl *structDecl = type->getAsStructureType()->getDecl();
   for (const auto *field : structDecl->fields()) {
+    if (field->getType()->isEmptyStructureType())
+      continue;
```
Keenuts, line 389: this line actually causes the test to pass, but generates invalid SPIR-V:
```
%23 = OpCompositeConstruct %B %22
```

If I comment this out, and change the lines above, it fixes it.
```diff
--- a/tools/clang/lib/SPIRV/InitListHandler.cpp
+++ b/tools/clang/lib/SPIRV/InitListHandler.cpp
@@ -366,13 +366,15 @@ InitListHandler::createInitForStructType(QualType type, SourceLocation srcLoc,
     while (tryToSplitConstantArray())
       ;

-    auto init = initializers.back();
-    // We can only avoid decomposing and reconstructing when the type is
-    // exactly the same.
-    if (type.getCanonicalType() ==
-        init->getAstResultType().getCanonicalType()) {
-      initializers.pop_back();
-      return init;
+    if (initializers.size() > 0) {
+      auto init = initializers.back();
+      // We can only avoid decomposing and reconstructing when the type is
+      // exactly the same.
+      if (type.getCanonicalType() ==
+          init->getAstResultType().getCanonicalType()) {
+        initializers.pop_back();
+        return init;
+      }
     }

     // Otherwise, if the next initializer is a struct, it is not of the same

```

Keenuts (commented):
Thanks for fixing! The off-by-one is clearly a bad oversight...
Just a few comments

tools/clang/lib/SPIRV/InitListHandler.cpp
LukasBanana: Thanks, that seems to work better and I updated the PR.

tools/clang/lib/SPIRV/LiteralTypeVisitor.cpp
```diff
@@ -395,6 +395,8 @@ bool LiteralTypeVisitor::updateTypeForCompositeMembers(
         // from this bitfield are not guaranteed, and we shouldn't touch them.
         if (field->isBitField())
           break;
+        if (field->getType()->isEmptyStructureType())
```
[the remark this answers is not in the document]
LukasBanana: The example I added boiled down the issue pretty good, but as mentioned in the other comment, this function doesn't seem to be necessary with the added condition for empty initializers.

LukasBanana:
@Keenuts is there anything else left to update the PR?

PR #5018, opened 2023-02-10
Introduce HLSLCXXOverload attribute

### LINKED ISSUE #4340: Overloaded array operator throwing error about constness
The following snippet illustrates an error I get when trying to overload the array operator:

`Texture2D Textures2D[] : register(t0, space10);

struct BindlessTexture2D
{
	uint Idx;

	float4 operator[](int2 pos)
	{
		return Textures2D[Idx][pos];
	}
};

struct MyConstantBuffer
{
	BindlessTexture2D MyTexture2D;
};

ConstantBuffer<MyConstantBuffer> CB : register(b1);

[numthreads(1, 1, 1)]
void CSMain()
{
	float4 value = CB.MyTexture2D[0];
}`

The error output is:

> error: no viable overloaded operator[] for type 'const BindlessTexture2D'
> note: candidate function not viable: 'this' argument has type 'const BindlessTexture2D', but method is not marked const

Regards,
Tom

llvm-beanz:
This change should have no functional impact on compilation. The goal is to remove cases where we rely on the subscript operator's name for special resolution rules and instead key the behavior off an attribute on the method declaration.

The added test cases are not exhaustive but try to cover at least one object type exercising each code path modified in this change.

Partially fixes #4340

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -2686,9 +2686,12 @@ public:
   }
 };
 
-static void AddHLSLSubscriptAttr(Decl *D, ASTContext &context, HLSubscriptOpcode opcode) {
+static void AddHLSLSubscriptAttr(Decl *D, ASTContext &context,
+                                 HLSubscriptOpcode opcode) {
   StringRef group = GetHLOpcodeGroupName(HLOpcodeGroup::HLSubscript);
-  D->addAttr(HLSLIntrinsicAttr::CreateImplicit(context, group, "", static_cast<unsigned>(opcode)));
+  D->addAttr(HLSLIntrinsicAttr::CreateImplicit(context, group, "",
+                                               static_cast<unsigned>(opcode)));
+  D->addAttr(HLSLCXXOverloadAttr::CreateImplicit(context));
 }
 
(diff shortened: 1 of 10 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12279,6 +12283,10 @@ void hlsl::HandleDeclAttributeForHLSL(Sema &S, Decl *D, const AttributeList &A,
     declAttr = ::new (S.Context) HLSLRayPayloadAttr(
         A.getRange(), S.Context, A.getAttributeSpellingListIndex());
     break;
+  /*case AttributeList::AT_HLSLCXXOverload:
+    declAttr = ::new (S.Context) HLSLCXXOverloadAttr(
+        A.getRange(), S.Context, A.getAttributeSpellingListIndex());
+    break;*/
```
tex3d, line 12289: Why is this commented case here?  Should it be removed (not accessible from HLSL directly)?

llvm-beanz: Removed. I was strugging to figure out why the attribute wasn't propagating as expected and this was my (incorrect) thought.

tex3d (approved):

PR #5020, opened 2023-02-10
[SPIR-V] Forward replacement expression for SubstNonTypeTemplateParmExpr

LukasBanana:
`SubstNonTypeTemplateParmExpr` was missing in `doExpr` of the SpirvEmitter. I boiled down the repro example to a simple function template with an input parameter that cannot be constant-folded.


cassiebeckley (approved):
Thanks for the fix!

PR #5021, opened 2023-02-10
Enable -Werror for C sources too

llvm-beanz:
This fixes some fallthrough warnings that were appearing in C source files and enables -Werror for both C and C++ sources.

This change does have the impact of opting people in to -Werror on local builds on Linux/Unix platforms with your next clean configuration. Since our bots do it, I think the default being on for users makes sense too.


Keenuts (commented):
Enforcing -Werror for submit? love it.
But not sure this should be enabled by default. Package managers and people will use the default behavior as it's the simplest, meaning we'll break them is compilers adds new warnings (like a few years ago when the implicit-fallthrough flag became live).
Example, https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=directx-shader-compiler-git doesn't specify the warning flag.

I'd be more in favor of having all bots hard-fail with warnings, which is quite easy to fix, and a note in the README to explain how to enable -Werror on a local build.

llvm-beanz:
> Enforcing -Werror for submit? love it. But not sure this should be enabled by default. Package managers and people will use the default behavior as it's the simplest, meaning we'll break them is compilers adds new warnings (like a few years ago when the implicit-fallthrough flag became live).

This is not the default for the build system, but the default for the predefined parameters config file we provide for our documented workflow. We don't have a recommended set of settings for package managers, and we don't officially support any, so I'm not sure this objection is entirely relevant.

The goal here is to make it the default for contributor local builds too so they don't have to rely on the bots to catch issues.

> Example, https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=directx-shader-compiler-git doesn't specify the warning flag.

My personal dislike for how Arch packages things comes full swing here. That configuration is already broken over a year ago by https://github.com/microsoft/DirectXShaderCompiler/pull/3918, as a result this change shouldn't impact their packaging.

If we want to support a build configuration for package managers, we should do that separately from the build configuration for contributors. There are options like enabling assertions and tests that are relevant to contributors but not relevant to package managers.

> I'd be more in favor of having all bots hard-fail with warnings, which is quite easy to fix, and a note in the README to explain how to enable -Werror on a local build.

The bots already hard fail (with the caveat that we were only enabling -Werror on c++ sources, which this change addresses). The in-tree LLVM documentation does document the option to enable/disable this setting:

https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/CMake.rst#L273

Keenuts:
> We don't have a recommended set of settings for package managers, and we don't officially support any,

> If we want to support a build configuration for package managers, we should do that separately from the build configuration for contributors. There are options like enabling assertions and tests that are relevant to contributors but not relevant to package managers.

Given those 2 points, I'm OK with this change. 😊

> The in-tree LLVM documentation does document the option to enable/disable this setting:

Good to know, thanks!

lib/Support/ConvertUTF.c
```diff
@@ -48,6 +48,7 @@
 
 
 #include "llvm/Support/ConvertUTF.h"
+#include "llvm/Support/Compiler.h" // HLSL Change - for LLVM_FALLTHROUGH
```
pow2clk, line 51: The warning disable for GCC added below is probably not needed now since the implicit passthrough warnings shouldn't be hit anymore

pow2clk (approved):
Far fewer places than I imagined this would be needed. I recommend removing the diagnostic disable I added to ConvertUTF.c

bob80905 (approved):

Keenuts (approved):
Thanks, just a small change 😊

PR #5024, opened 2023-02-11
Fix AppVeyor Linux tests failing loading dxil.so

llvm-beanz:
This patch resolves an issue where dxil.so was failing to load but the error status was not being propagated, which caused repeated load attempts and eventually incorrect error propagation in a test case.

A bunch of this code is fairly gnarly, but the general gist is that `dlopen` is not required to set `errno`, so we were returning back success on failure cases unless there happened to be a previous failure value in `errno`.

Resolving this issue should get AppVeyor back on the rails.

There's a related PR that tries to handle the errors correctly (#4713), and we should adopt something similar.

### DIFF
--- a/include/dxc/Support/dxcapi.use.h
+++ b/include/dxc/Support/dxcapi.use.h
@@ -32,14 +32,10 @@ protected:
 #ifdef _WIN32
     m_dll = LoadLibraryA(dllName);
     if (m_dll == nullptr) return HRESULT_FROM_WIN32(GetLastError());
+    m_createFn = (DxcCreateInstanceProc)GetProcAddress(m_dll, fnName);
 #else
     m_dll = ::dlopen(dllName, RTLD_LAZY);
     if (m_dll == nullptr) return E_FAIL;
-#endif
-
-#ifdef _WIN32
-    m_createFn = (DxcCreateInstanceProc)GetProcAddress(m_dll, fnName);
-#else
     m_createFn = (DxcCreateInstanceProc)::dlsym(m_dll, fnName);
 #endif
 
--- a/include/dxc/Support/dxcapi.use.h
+++ b/include/dxc/Support/dxcapi.use.h
@@ -33,22 +33,24 @@ protected:
     m_dll = LoadLibraryA(dllName);
     if (m_dll == nullptr) return HRESULT_FROM_WIN32(GetLastError());
     m_createFn = (DxcCreateInstanceProc)GetProcAddress(m_dll, fnName);
+    
+    if (m_createFn == nullptr) {
+      HRESULT hr = HRESULT_FROM_WIN32(GetLastError());
+      FreeLibrary(m_dll);
+      m_dll = nullptr;
+      return hr;
+    }
 #else
     m_dll = ::dlopen(dllName, RTLD_LAZY);
     if (m_dll == nullptr) return E_FAIL;
     m_createFn = (DxcCreateInstanceProc)::dlsym(m_dll, fnName);
-#endif
-
+    
     if (m_createFn == nullptr) {
-      HRESULT hr = HRESULT_FROM_WIN32(GetLastError());
-#ifdef _WIN32
-      FreeLibrary(m_dll);
-#else
       ::dlclose(m_dll);
-#endif
       m_dll = nullptr;
-      return hr;
+      return E_FAIL;
     }
+#endif
 
     // Only basic functions used to avoid requiring additional headers.
     m_createFn2 = nullptr;
--- a/include/dxc/Support/dxcapi.use.h
+++ b/include/dxc/Support/dxcapi.use.h
@@ -31,11 +31,11 @@ protected:
 
 #ifdef _WIN32
     m_dll = LoadLibraryA(dllName);
+    if (m_dll == nullptr) return HRESULT_FROM_WIN32(GetLastError());
 #else
     m_dll = ::dlopen(dllName, RTLD_LAZY);
+    if (m_dll == nullptr) return E_FAIL;
 #endif
-
-    if (m_dll == nullptr) return HRESULT_FROM_WIN32(GetLastError());
 
 #ifdef _WIN32
     m_createFn = (DxcCreateInstanceProc)GetProcAddress(m_dll, fnName);
--- a/tools/clang/unittests/HLSL/DxilContainerTest.cpp
+++ b/tools/clang/unittests/HLSL/DxilContainerTest.cpp
@@ -810,10 +810,10 @@ TEST_F(DxilContainerTest, CompileWhenOKThenIncludesSignatures) {
       "; COLOR                    0   xyzw        1     NONE   float   xyzw\n"; // should read '1' in register
     if (hlsl::DXIL::CompareVersions(m_ver.m_ValMajor, m_ver.m_ValMinor, 1, 5) < 0) {
       std::string start(s.c_str(), strlen(expected_1_4));
-      VERIFY_ARE_EQUAL_STR(expected_1_4, start.c_str());
+      EXPECT_STREQ(expected_1_4, start.c_str());
     } else {
       std::string start(s.c_str(), strlen(expected));
-      VERIFY_ARE_EQUAL_STR(expected, start.c_str());
+      EXPECT_STREQ(expected, start.c_str());
     }
   }
 
--- a/tools/clang/unittests/HLSL/DxilContainerTest.cpp
+++ b/tools/clang/unittests/HLSL/DxilContainerTest.cpp
@@ -810,10 +810,10 @@ TEST_F(DxilContainerTest, CompileWhenOKThenIncludesSignatures) {
       "; COLOR                    0   xyzw        1     NONE   float   xyzw\n"; // should read '1' in register
     if (hlsl::DXIL::CompareVersions(m_ver.m_ValMajor, m_ver.m_ValMinor, 1, 5) < 0) {
       std::string start(s.c_str(), strlen(expected_1_4));
-      EXPECT_STREQ(expected_1_4, start.c_str());
+      VERIFY_ARE_EQUAL_STR(expected_1_4, start.c_str());
     } else {
       std::string start(s.c_str(), strlen(expected));
-      EXPECT_STREQ(expected, start.c_str());
+      VERIFY_ARE_EQUAL_STR(expected, start.c_str());
     }
   }
 


llvm-beanz:
Just want to capture the error here:

```
/home/appveyor/projects/DirectXShaderCompiler/tools/clang/unittests/HLSL/DxilContainerTest.cpp:813: Failure
[902](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L902)      Expected: expected_1_4
[903](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L903)      Which is: ";\n; Input signature:\n;\n; Name                 Index   Mask Register SysValue  Format   Used\n; -------------------- ----- ------ -------- -------- ------- ------\n; POSITION                 0   xyzw        0     NONE   float       \n; COLOR                    0   xyzw        1     NONE   float       \n;\n;\n; Output signature:\n;\n; Name                 Index   Mask Register SysValue  Format   Used\n; -------------------- ----- ------ -------- -------- ------- ------\n; SV_Position              0   xyzw        0      POS   float   xyzw\n; COLOR                    0   xyzw        1     NONE   float   xyzw\n"
[904](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L904)To be equal to: start.c_str()
[905](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L905)      Which is: ";\n; Input signature:\n;\n; Name                 Index   Mask Register SysValue  Format   Used\n; -------------------- ----- ------ -------- -------- ------- ------\n; POSITION                 0   xyzw        0     NONE   float   xyzw\n; COLOR                    0   xyzw        1     NONE   float   xyzw\n;\n;\n; Output signature:\n;\n; Name                 Index   Mask Register SysValue  Format   Used\n; -------------------- ----- ------ -------- -------- ------- ------\n; SV_Position              0   xyzw        0      POS   float   xyzw\n; COLOR                    0   xyzw        1     NONE   float   xyzw\n"
[906](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L906)With diff:
[907](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L907)@@ -4,6 +4,6 @@
[908](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L908) ; Name                 Index   Mask Register SysValue  Format   Used
[909](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L909) ; -------------------- ----- ------ -------- -------- ------- ------
[910](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L910)-; POSITION                 0   xyzw        0     NONE   float       
[911](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L911)-; COLOR                    0   xyzw        1     NONE   float       
[912](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L912)+; POSITION                 0   xyzw        0     NONE   float   xyzw
[913](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L913)+; COLOR                    0   xyzw        1     NONE   float   xyzw
[914](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L914) ;
[915](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L915) ;
[916](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L916)
[917](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L917)/home/appveyor/projects/DirectXShaderCompiler/tools/clang/unittests/HLSL/DxilContainerTest.cpp:855: Failure
[918](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L918)      Expected: 0
[919](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L919)To be equal to: strcmp(pTmpA, pTmpB)
[920](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L920)      Which is: -88
[921](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L921)[  FAILED  ] DxilContainerTest.CompileWhenOKThenIncludesSignatures (7 ms)
[922](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/46207505#L922)
```

include/dxc/Support/dxcapi.use.h
```diff
@@ -31,15 +31,11 @@ class DxcDllSupport {
...
-#endif
-
     if (m_dll == nullptr) return HRESULT_FROM_WIN32(GetLastError());
-
-#ifdef _WIN32
     m_createFn = (DxcCreateInstanceProc)GetProcAddress(m_dll, fnName);
 #else
+    m_dll = ::dlopen(dllName, RTLD_LAZY);
+    if (m_dll == nullptr) return E_FAIL;
     m_createFn = (DxcCreateInstanceProc)::dlsym(m_dll, fnName);
```
tex3d, line 39: Still relying on `HRESULT hr = HRESULT_FROM_WIN32(GetLastError())` later to pick up error if `m_createFn` is not found.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -810,10 +810,10 @@ TEST_F(DxilContainerTest, CompileWhenOKThenIncludesSignatures) {
       "; COLOR                    0   xyzw        1     NONE   float   xyzw\n"; // should read '1' in register
     if (hlsl::DXIL::CompareVersions(m_ver.m_ValMajor, m_ver.m_ValMinor, 1, 5) < 0) {
       std::string start(s.c_str(), strlen(expected_1_4));
-      VERIFY_ARE_EQUAL_STR(expected_1_4, start.c_str());
+      EXPECT_STREQ(expected_1_4, start.c_str());
```
tex3d, line 813: Isn't this what you want `VERIFY_ARE_EQUAL_STR` to resolve to on non-win32?  Are we switching all the macros now?

include/dxc/Support/dxcapi.use.h
pow2clk, line 39: I think we should do similar for dlsym() if you feel like it, you can use the error codes I suggested here: https://github.com/microsoft/DirectXShaderCompiler/pull/4713#issuecomment-1275447597

pow2clk (commented):
I'm not going to insist on error codes, but I think we should handle dlsym() properly.

Shout out to @ruiling for catching this months ago. If you'd removed the printfs in #4713, we would have taken that change

pow2clk (approved):

tools/clang/unittests/HLSL/DxilContainerTest.cpp
llvm-beanz: I actually think we should invert this. We’ve talked about wanting to remove the dependency on TAEF for our unit tests on Windows, so we should probably replace all the TAEF-style macros with the native GoogleTest macros. Right now we have adapter macros going both ways. Someday we can hopefully get rid of all of those.

tex3d: We may wish to change all the code away from TAEF macros, but this doesn't seem like the right PR for that.

You didn't change other macros nearby and EXPECT_STREQ actually has a worse mapping for Win32 at the moment.  It would have been better to map VERIFY_ARE_EQUAL_STR to EXPECT_STREQ on linux for minimal change improvement in this scenario.

Mapping on Windows (HlslTestUtils.h):
```
#define EXPECT_STREQ(a, b) VERIFY_ARE_EQUAL(0, strcmp(a, b))
```

include/dxc/Support/dxcapi.use.h
```diff
@@ -31,28 +31,26 @@ class DxcDllSupport {
...
     if (m_createFn == nullptr) {
       HRESULT hr = HRESULT_FROM_WIN32(GetLastError());
-#ifdef _WIN32
       FreeLibrary(m_dll);
+      m_dll = nullptr;
+      return hr;
+    }
 #else
+    m_dll = ::dlopen(dllName, RTLD_LAZY);
+    if (m_dll == nullptr) return E_FAIL;
```
MarijnS95, line 45: Is `E_FAIL` equivalent to `Unspecified error`, as in "we don't know what happened"? And is that what you are referring to with:

> There's a related PR that tries to handle the errors correctly (https://github.com/microsoft/DirectXShaderCompiler/pull/4713), and we should adopt something similar.

... that a _followup_ change should adopt something that more clearly conveys the reason from `dlerror()` or at least make it clear that it is a `dlopen` error and that the caller _could_ check `dlerror()` for more info?

tools/clang/unittests/HLSL/DxilContainerTest.cpp
llvm-beanz: I'll revert these changes but I think you're splitting hairs here.

This test was failing on Linux. It provided output that I didn't find useful in the log, so I fixed that as part of debugging. If you're saying I can't fix one usability issue without fixing all occurrences of the usability issue that more or less equates to "we can't incrementally improve things".

include/dxc/Support/dxcapi.use.h
llvm-beanz: Yes. `E_FAIL` is an unspecified error.

Using `dlerror()` would allow us to translate that to a specific error, however `dlerror()` returns errors as strings, and our interface as it is now doesn't really have a good way to propagate the error to the library user (logging to stderr is a pretty terrible way for a library to behave).

I don't think we have clear consensus within the team about how to handle these kinds of errors. My personal opinion is to move away from the Windows-style HRESULT error propagating to a more robust error propagation interface that is cross-platform and more expressive (possibly an extension of `std::error_code` or an `llvm::Error`-like interface).

Right now the best guidance we can give users on Linux is that if library loading fails, they should call `dlerror()` themselves to get the error string.

MarijnS95: Thanks for making that explicit, it matches what I was assuming.

> Right now the best guidance we can give users on Linux is that if library loading fails, they should call `dlerror()` themselves to get the error string.

Is this documented somewhere, should it be documented somewhere?

llvm-beanz: It is not documented. To my knowledge currently have no documentation for the Linux-specific aspects of our API, so I'm not sure we even have a good place to document it.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
tex3d: That's a strawman, and not what I was saying at all.

The change will drop the string details that would have been present before when running on the Windows platform.

So it's an improvement only in this case for Linux (because on Linux we don't map the TAEF logging settings for the test), and a regression for this case on Windows - we would no longer log the string contents that didn't match, so now, we'd have the same problem on Windows that we ran into on Linux.

A slightly different change could have helped on Linux even more broadly, without regressing the functionality on Windows.

Therefore, I didn't see a good justification for this specific change, and the regression it would cause for the Windows path.

PR #5029, opened 2023-02-15
Enable Additional Security Checks (/sdl) and fix errors that surfaced

hekota:
Enables Additional Security Checks (/sdl compiler flag). These checks change security-relevant warnings into errors and set additional secure code-generation features.

Fix errors that surfaced by adding /sdl. The errors found were in these two categories:
- C4146: unary minus operator applied to unsigned type, result still unsigned
- C4703: potentially uninitialized local pointer variable


tex3d (approved):
LGTM.

PR #5031, opened 2023-02-15
Don't skip copy-in copy-out when the callee is noinline or user define need external function.

python3kgae:
This is for better noinline support.
Doing it early can avoid doing the copy-in copy-out later in pipeline
 for things like Input/Output Signature and static global variables.


python3kgae:
Close for perf impact.

PR #5033, opened 2023-02-16
remove uninitialized assign to an unused variable

bob80905:
HLK tests are failing for IsNormal and BarycentricsChecKOrdering.
This fix repairs the IsNormal test. 
The Barycentrics test seems to be a driver problem, since the output of the shader differs at the lowest level, when compared against a different IHV's driver.


llvm-beanz (approved):
Deleting dead code is always good...

I suspect there is some undefined behavior or memory safety error somewhere if this is causing issues.

PR #5037, opened 2023-02-16
[NFC] Remove custom StringCchCopyEx

llvm-beanz:
This patch got a bit away from me. It started as me trying to remove one of the Windows system functions we have a custom implementation of (StringCchCopyEx). It didn't really make sense to me that we needed a custom override of a function where we could have just used `strncpy` or other LLVM utilities to provide the functionality.

After I rewrote the code calling `StringCchCopyEx`, I wanted to make sure it was tested, which led me to discover that it was never called in our existing tests (see:
https://microsoft.github.io/DirectXShaderCompiler/coverage/home/runner/w ork/DirectXShaderCompiler/DirectXShaderCompiler/lib/DxcSupport/WinFuncti ons.cpp.html#L27).

I then looked into the behavior of
`clang_getCursorSpellingWithFormatting`, which it turns out _does_ return fully qualified names. As a result the parent traversal we were attempting always terminated immediately with one name segment found.

I wrote an additional test where the cursor name is _not_ fully qualified in source to verify that the API works as expected in all cases then started deleting code.

This patch deletes two functions (`StringCchCopyEx` & `ConcatPartsReversed`), and simplifies the code in `GetCursorQualifiedName`. It also adds the additional non-fully-qualified name test case.

### DIFF
--- a/include/dxc/Support/WinFunctions.h
+++ b/include/dxc/Support/WinFunctions.h
@@ -19,8 +19,6 @@
 
 #ifndef _WIN32
 
-HRESULT StringCchCopyEx(LPSTR pszDest, size_t cchDest, LPCSTR pszSrc,
-                        LPSTR *ppszDestEnd, size_t *pcchRemaining, DWORD dwFlags);
 HRESULT StringCchPrintfA(char *dst, size_t dstSize, const char *format, ...);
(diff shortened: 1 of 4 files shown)


tex3d (approved):
Yay, deleting unnecessary code!

PR #5041, opened 2023-02-17
Always create temp for output signature

python3kgae:
This is for case where output signature is used as argument for functions call.
 A tmp is needed because the signature will be removed in dxil.

Doing it in LegalizeDxilInputOutputs is to reused existing feature which create the tmp. 
Only add the temp for signature when there's real function call.

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -4038,7 +4038,8 @@ static void DeleteOutOfScopeDebugInfo(Function &F) {
 static void LegalizeDxilInputOutputs(Function *F,
                                      DxilFunctionAnnotation *EntryAnnotation,
                                      const DataLayout &DL,
-                                     DxilTypeSystem &typeSys);
+                                     DxilTypeSystem &typeSys,
+                                     bool UsesSignature);
 static void InjectReturnAfterNoReturnPreserveOutput(HLModule &HLM);
 
 namespace {
@@ -4096,7 +4097,8 @@ public:
           !m_pHLModule->IsEntryThatUsesSignatures(&F)) {
         if (!F.isDeclaration())
           LegalizeDxilInputOutputs(&F, m_pHLModule->GetFunctionAnnotation(&F),
-                                   DL, m_pHLModule->GetTypeSystem());
+                                   DL, m_pHLModule->GetTypeSystem(),
+                                   false);
         continue;
       }
 
@@ -5607,11 +5609,36 @@ static void InjectReturnAfterNoReturnPreserveOutput(HLModule &HLM) {
   }
 }
 
+// Used by noinline function or external user define function call.
+static bool UsedByRealFunctionCall(const Value *V) {
+  for (const User *U : V->users()) {
+    if (isa<GEPOperator>(U))
+      if (UsedByRealFunctionCall(U))
+        return true;
+    const CallInst *CI = dyn_cast<CallInst>(U);
+    if (!CI)
+      continue;
+    const Function *F = CI->getCalledFunction();
+    if (!F)
+      continue;
+    // Noinline function.
+    if (F->hasFnAttribute(llvm::Attribute::NoInline))
+      return true;
+    if (!F->hasExternalLinkage())
+      continue;
+    // User define external function.
+    if (hlsl::GetHLOpcodeGroupByName(F) == HLOpcodeGroup::NotHL)
+      return true;
+  }
+  return false;
+}
+
 // Support store to input and load from output.
 static void LegalizeDxilInputOutputs(Function *F,
                                      DxilFunctionAnnotation *EntryAnnotation,
                                      const DataLayout &DL,
-                                     DxilTypeSystem &typeSys) {
+                                     DxilTypeSystem &typeSys,
+                                     bool UsesSignature) {
   BasicBlock &EntryBlk = F->getEntryBlock();
   Module *M = F->getParent();
   // Map from output to the temp created for it.
@@ -5691,6 +5718,10 @@ static void LegalizeDxilInputOutputs(Function *F,
         bLoadOutputFromTemp = true;
       }
     }
+
+    // Always create temp for output signature so user function will not use
+    // output signature directly to avoid generate copy in signature lowering.
+    bLoadOutputFromTemp |= (UsesSignature && UsedByRealFunctionCall(&arg));
 
     if (bStoreInputToTemp || bLoadOutputFromTemp) {
       IRBuilder<> Builder(EntryBlk.getFirstInsertionPt());
@@ -5915,6 +5946,8 @@ void SROA_Parameter_HLSL::createFlattenedFunction(Function *F) {
   }
 
   FunctionType *flatFuncTy = FunctionType::get(retType, FinalTypeList, false);
+  bool UsesSignature = F == m_pHLModule->GetEntryFunction() ||
+                       m_pHLModule->IsEntryThatUsesSignatures(F);
   // Return if nothing changed.
   if (flatFuncTy == F->getFunctionType()) {
     // Copy semantic allocation.
@@ -5930,7 +5963,7 @@ void SROA_Parameter_HLSL::createFlattenedFunction(Function *F) {
     }
     if (!F->isDeclaration()) {
       // Support store to input and load from output.
-      LegalizeDxilInputOutputs(F, funcAnnotation, DL, typeSys);
+      LegalizeDxilInputOutputs(F, funcAnnotation, DL, typeSys, UsesSignature);
     }
     return;
   }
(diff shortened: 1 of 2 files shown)


lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -5692,6 +5719,10 @@ static void LegalizeDxilInputOutputs(Function *F,
       }
     }
 
+    // Always create temp for output signature so user function will not use
+    // output signature directly to avoid generate copy in signature lowering.
+    bLoadOutputFromTemp |= (UsesSignature && UsedByRealFunctionCall(&arg));
```
tex3d, line 5724: When `UsesSignature` is true, shouldn't this set `bStoreInputToTemp` if the arg is any `in` (`in`/`inout`), and set `bLoadOutputFromTemp` if the arg is any `out` (`out`/`inout`)?

tex3d: Plus, doesn't this also depend on the potential usage of that pointer argument in the called function (`in`/`inout`/`out` there) as well?  In other words, we would need to know if any called function reads from the pointer, separately from whether any called function writes to the pointer.

Perhaps that's supposed to be handled by `CheckArgUsage`, but it seems 

Also, our `CheckArgUsage` function is incomplete.  It doesn't expect any user function calls.  It presumes a function call that returns a pointer must be a subscript operation, so it recurses to its users to find out whether it loads/stores.  It doesn't consider pointer arguments to functions that do not return a pointer as reading from or writing to that pointer, other than specific HLMatrix operations.  This will impact the correctness of the copy-on-write-input and copy-on-read-output optimization.

Even if that's fixed, I think we would want to know specifics about how a function call user uses a pointer in relation to the `UsesSignature` flag.  It's one thing for some other code to load from a pointer, but if a function potentially loads from the pointer *and* the `UsesSignature` flag is set, then we need to copy-in.  If a function potentially writes to the pointer *and* the `UsesSignature` flag is set, then we need to copy-out.  This is more information than the simple `bLoad`/`bStore` provides.

This also indicates that we need more thorough testing of pointer arguments to noinline function calls as well, to ensure these scenarios are handled correctly.

python3kgae: > When `UsesSignature` is true, shouldn't this set `bStoreInputToTemp` if the arg is any `in` (`in`/`inout`), and set `bLoadOutputFromTemp` if the arg is any `out` (`out`/`inout`)?

Input will get temp generated in clang codeGen if is struct. For scalar Input, it will just replace with LdInput, don't need temp.

tex3d: > This is more information than the simple bLoad/bStore provides.

Also, `UsedByRealFunctionCall` doesn't provide this information either (plus you can't early exit from that until you know you do both).  It seems like the functionality of `UsedByRealFunctionCall` would be best integrated into `CheckArgUsage`, with a couple more bits of information returned: read by function/needs copy-in (`bStoreInputToTemp`), and written by function/needs copy-out (`bLoadOutputFromTemp`).

python3kgae: > Plus, doesn't this also depend on the potential usage of that pointer argument in the called function (`in`/`inout`/`out` there) as well? In other words, we would need to know if any called function reads from the pointer, separately from whether any called function writes to the pointer.
> 
> Perhaps that's supposed to be handled by `CheckArgUsage`, but it seems
> 
> Also, our `CheckArgUsage` function is incomplete. It doesn't expect any user function calls. It presumes a function call that returns a pointer must be a subscript operation, so it recurses to its users to find out whether it loads/stores. It doesn't consider pointer arguments to functions that do not return a pointer as reading from or writing to that pointer, other than specific HLMatrix operations. This will impact the correctness of the copy-on-write-input and copy-on-read-output optimization.
> 
> Even if that's fixed, I think we would want to know specifics about how a function call user uses a pointer in relation to the `UsesSignature` flag. It's one thing for some other code to load from a pointer, but if a function potentially loads from the pointer _and_ the `UsesSignature` flag is set, then we need to copy-in. If a function potentially writes to the pointer _and_ the `UsesSignature` flag is set, then we need to copy-out. This is more information than the simple `bLoad`/`bStore` provides.
> 
> This also indicates that we need more thorough testing of pointer arguments to noinline function calls as well, to ensure these scenarios are handled correctly.

How the function uses the pointer arguments doesn't affect the result that a temp is needed.
The temp will be created to replace all uses on the output, then store the val back to output before each return.
No copyin/copyout need to check at this point. It should be already done when set output as the argument.

python3kgae: > > This is more information than the simple bLoad/bStore provides.
> 
> Also, `UsedByRealFunctionCall` doesn't provide this information either (plus you can't early exit from that until you know you do both). It seems like the functionality of `UsedByRealFunctionCall` would be best integrated into `CheckArgUsage`, with a couple more bits of information returned: read by function/needs copy-in (`bStoreInputToTemp`), and written by function/needs copy-out (`bLoadOutputFromTemp`).

It is different from in clang codeGen where we generate copyin/copyout.
Here it just replaces all uses of the output with tmp. Then copy the val back to output before each return.

tex3d: > Input will get temp generated in clang codeGen if is struct.

For `in` and `inout`, but not for `out`?  Does this only apply to shader entry points?

If so, why was it done, and why don't we always generate a temp for `out` as well in codegen?  And won't the copy-in code here add an additional copy in that case if stored to?

In any case, locally, it's not clear that this code is handling functions correctly for what it appears to do.  You have to assume some additional quirks of our CodeGen, and even then, I'm still not sure it's doing things correctly.

It might be best if CodeGen generated an HL operation for loading/storing inputs/outputs to temps in entry points in the first place.  This would provide a more correct initial CodeGen and make additional things like this unnecessary.

> How the function uses the pointer arguments doesn't affect the result that a temp is needed

But if the original arg wasn't an output in the first place, or nobody actually writes to the output, we don't need to copy back, right?  That's why we don't add temps to `outputTempMap` when not stored in the first place.

python3kgae: > > Input will get temp generated in clang codeGen if is struct.
> 
> For `in` and `inout`, but not for `out`? Does this only apply to shader entry points?
> 
> If so, why was it done, and why don't we always generate a temp for `out` as well in codegen? And won't the copy-in code here add an additional copy in that case if stored to?
> 
> In any case, locally, it's not clear that this code is handling functions correctly for what it appears to do. You have to assume some additional quirks of our CodeGen, and even then, I'm still not sure it's doing things correctly.
> 
> It might be best if CodeGen generated an HL operation for loading/storing inputs/outputs to temps in entry points in the first place. This would provide a more correct initial CodeGen and make additional things like this unnecessary.
> 
> > How the function uses the pointer arguments doesn't affect the result that a temp is needed
> 
> But if the original arg wasn't an output in the first place, or nobody actually writes to the output, we don't need to copy back, right? That's why we don't add temps to `outputTempMap` when not stored in the first place.

It the original output used as readonly for a function call, LegalizeDxilInputOutputs should already set bLoadOutputFromTemp to be true. That's the reason we do LegalizeDxilInputOutputs.

For out/inout not generate temp in clang CodeGen, it is because we don't like ptr of ptr to be generated.

tex3d: For `in`, it appears we add the temp in this pass, not in CodeGen.

Looking at where we do this, it's a side effect of how we flatten and replace users of the function parameters.  We also go down convoluted paths here for in vs. out arguments.  We also have complex differences that depend on whether we are compiling a library.

This area is already very hard to follow, I think there are still potential bugs hidden in the complexity, and I think this addition only complicates the picture a bit more instead of simplifying things.  We could accept it because it fixes the issue you're seeing, but I think we really need a better approach at a fundamental level.

python3kgae: I guess we could add
bStoreInputToTemp |=  (UsesSignature && UsedByRealFunctionCall(&arg));
So the logical is easier to understand.

If UsedByRealFunctionCall is false, nothing will change ( which is always true now). 
Else a tmp will be required and bStoreInputToTemp will be true and do the job.

python3kgae: How about this?

```
bool NeedTmpForFunctionCall = (UsesSignature && UsedByRealFunctionCall(&arg));

    if (qual == DxilParamInputQual::In && (bStore || NeedTmpForFunctionCall )) {
      bStoreInputToTemp = true;
    } else if (qual == DxilParamInputQual::Out && (bLoad || NeedTmpForFunctionCall )) {
      bLoadOutputFromTemp = true;
} ...
```

python3kgae: PR updated.

python3kgae:
The issue will be fixed with new calling convention feature.

PR #5044, opened 2023-02-21
[Support] Add the 'Error' class for structured error handling.

llvm-beanz:
This patch introduces the Error classs for lightweight, structured, recoverable error handling. It includes utilities for creating, manipulating
and handling errors. The scheme is similar to exceptions, in that errors are
described with user-defined types. Unlike exceptions however, errors are represented as ordinary return types in the API (similar to the way std::error_code is used).

For usage notes see the LLVM programmer's manual, and the Error.h header.
Usage examples can be found in unittests/Support/ErrorTest.cpp.

Many thanks to David Blaikie, Mehdi Amini, Kevin Enderby and others on the
llvm-dev and llvm-commits lists for lots of discussion and review.

llvm-svn: 263609

Original commit information:
From f7f6d3e93f739077dc280532d906ca523489b01f Mon Sep 17 00:00:00 2001
From: Lang Hames <@lhames>
Date: Wed, 16 Mar 2016 01:02:46 +0000
Subject: [PATCH] [Support] Add the 'Error' class for structured error
 handling.

### DIFF
--- a/docs/ProgrammersManual.rst
+++ b/docs/ProgrammersManual.rst
@@ -263,7 +263,175 @@ almost never be stored or mentioned directly.  They are intended solely for use
 when defining a function which should be able to efficiently accept concatenated
 strings.
 
+.. _error_apis:
+
+Error handling
+--------------
+
+Proper error handling helps us identify bugs in our code, and helps end-users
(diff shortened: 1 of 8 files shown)


tex3d (approved):

llvm-beanz:

PR #5047, opened 2023-02-22
Add diagnostics for uninitialized `out` parameters

### LINKED ISSUE #4951: Improve diagnostics for `out` parameters
We should issue diagnostics when `out` parameters are not written to or may not be written to in a function.

llvm-beanz:
This change adds diagnostic support for uninitialized `out` parameters. With this change, `out` parameters are treated as uninitiailzed values until they are initialized and use of the values produdces a diagnostic.

This change is a slightly hacky extension of Clang's uninitiailzed values analysis which is achieved by adding ParamVarDecls for `out` parameters to the list of variables to analyize. It also adds dummy uses of the variables on `return` statements.

Fixes #4951

### DIFF
--- a/tools/clang/lib/Analysis/UninitializedValues.cpp
+++ b/tools/clang/lib/Analysis/UninitializedValues.cpp
@@ -1017,37 +1017,6 @@ void clang::runUninitializedVariablesAnalysis(
       ++stats.NumBlockVisits;
     }
   }
-
-  // HLSL Change Begin - Treat `out` parameters as uninitialized values.
-  // If this block has no successors and does not have a terminator stmt which
-  // we would have handled above.
-  auto ExitBlock = cfg.getExit();
-  bool AllHandled = true;
-  TransferFunctions tf(vals, cfg, &ExitBlock, ac, classification, handler);
-  for (auto P = ExitBlock.pred_begin() ; P != ExitBlock.pred_end(); ++P) {
-    if (auto Term = (*P)->getTerminator()) {
-      if (isa<ReturnStmt>(*Term))
-        continue;
-    }
-    AllHandled = false;
-    break;
-  }
-  if (!AllHandled) {
-    // Create a dummy use DeclRefExpr for all the `out` params.
-    for (auto *P : vals.getHLSLOutParams()) {
-      Value v = vals[P];
-      if (!isUninitialized(v))
-        continue;
-      auto *DRE = DeclRefExpr::Create(
-          P->getASTContext(), NestedNameSpecifierLoc(), SourceLocation(),
-          const_cast<VarDecl *>(P), false,
-          DeclarationNameInfo(P->getDeclName(),
-                              cast<Decl>(&dc)->getBody()->getLocEnd()),
-          P->getASTContext().VoidTy, ExprValueKind::VK_RValue);
-      tf.reportUse(DRE, P);
-    }
-  }
-  // HLSL Change End - Treat `out` parameters as uninitialized values.
 }
 
 UninitVariablesHandler::~UninitVariablesHandler() {}
--- a/tools/clang/lib/Analysis/UninitializedValues.cpp
+++ b/tools/clang/lib/Analysis/UninitializedValues.cpp
@@ -34,6 +34,10 @@ using namespace clang;
 #define DEBUG_LOGGING 0
 
 static bool isTrackedVar(const VarDecl *vd, const DeclContext *dc) {
+  // HLSL Change Begin - Treat `out` parameters as uninitialized values.
+  if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>())
+    return true;
+  // HLSL Change End - Treat `out` parameters as uninitialized values.
   if (vd->isLocalVarDecl() && !vd->hasGlobalStorage() &&
       !vd->isExceptionVariable() && !vd->isInitCapture() &&
       !vd->isImplicit() && vd->getDeclContext() == dc) {
@@ -50,6 +54,10 @@ static bool isTrackedVar(const VarDecl *vd, const DeclContext *dc) {
 namespace {
 class DeclToIndex {
   llvm::DenseMap<const VarDecl *, unsigned> map;
+  
+  // HLSL Change Begin - Treat `out` parameters as uninitialized values.
+  llvm::SmallVector<const VarDecl *, 4> hlslOutParams;
+  // HLSL Change End - Treat `out` parameters as uninitialized values.
 public:
   DeclToIndex() {}
   
@@ -61,6 +69,12 @@ public:
   
   /// Returns the bit vector index for a given declaration.
   Optional<unsigned> getValueIndex(const VarDecl *d) const;
+
+  // HLSL Change Begin - Treat `out` parameters as uninitialized values.
+  const llvm::SmallVector<const VarDecl *, 4> &getHLSLOutParams() {
+    return hlslOutParams;
+  }
+  // HLSL Change End - Treat `out` parameters as uninitialized values.
 };
 }
 
@@ -72,6 +86,11 @@ void DeclToIndex::computeMap(const DeclContext &dc) {
     const VarDecl *vd = *I;
     if (isTrackedVar(vd, &dc))
       map[vd] = count++;
+    // HLSL Change Begin - Treat `out` parameters as uninitialized values.
+    // Keep HLSL parameters in a separate index.
+    if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>())
+      hlslOutParams.push_back(vd);
+    // HLSL Change End - Treat `out` parameters as uninitialized values.
   }
 }
 
@@ -137,6 +156,12 @@ public:
     assert(idx.hasValue());
     return getValueVector(block)[idx.getValue()];
   }
+
+  // HLSL Change Begin - Treat `out` parameters as uninitialized values.
+  const llvm::SmallVector<const VarDecl *, 4> &getHLSLOutParams() {
+    return declToIndex.getHLSLOutParams();
+  }
+  // HLSL Change End - Treat `out` parameters as uninitialized values.
 };  
 } // end anonymous namespace
 
@@ -454,8 +479,23 @@ void ClassifyRefs::VisitCallExpr(CallExpr *CE) {
   // If a value is passed by const pointer or by const reference to a function,
   // we should not assume that it is initialized by the call, and we
   // conservatively do not assume that it is used.
+  unsigned ParamIdx = 0; // HLSL Change
   for (CallExpr::arg_iterator I = CE->arg_begin(), E = CE->arg_end();
-       I != E; ++I) {
+       I != E; ++I, ++ParamIdx) { // HLSL Change - add ParamIdx
+    // HLSL Change Begin - Treat `out` parameters as uninitialized values.
+    if (auto FD = CE->getDirectCallee()) {
+      if (FD->getNumParams() > ParamIdx) {
+        ParmVarDecl *PD = FD->getParamDecl(ParamIdx);
+        bool HasIn = PD->hasAttr<HLSLInAttr>();
+        bool HasOut = PD->hasAttr<HLSLOutAttr>();
+        bool HasInOut = PD->hasAttr<HLSLInOutAttr>();
+        // If we have an in annotation or no annotation (implcit in), this is a
+        // use not an initialization.
+        if(HasIn || HasInOut || (!HasIn && !HasOut && !HasInOut))
+          classify(*I, Use);
+      }
+    }
+    // HLSL Change End - Treat `out` parameters as uninitialized values.
     if ((*I)->isGLValue()) {
       if ((*I)->getType().isConstQualified())
         classify((*I), Ignore);
@@ -514,6 +554,7 @@ public:
   void VisitDeclStmt(DeclStmt *ds);
   void VisitObjCForCollectionStmt(ObjCForCollectionStmt *FS);
   void VisitObjCMessageExpr(ObjCMessageExpr *ME);
+  void VisitReturnStmt(ReturnStmt *RS); // HLSL Change
 
   bool isTrackedVar(const VarDecl *vd) {
     return ::isTrackedVar(vd, cast<DeclContext>(ac.getDecl()));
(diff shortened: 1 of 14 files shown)


tools/clang/lib/Analysis/UninitializedValues.cpp
```diff
@@ -794,6 +835,25 @@ void TransferFunctions::VisitObjCMessageExpr(ObjCMessageExpr *ME) {
...
+  for (auto *P : vals.getHLSLOutParams()) {
+    Value v = vals[P];
+    if (!isUninitialized(v))
+      continue;
+    auto *DRE = DeclRefExpr::Create(
+        P->getASTContext(), NestedNameSpecifierLoc(), SourceLocation(),
+        const_cast<VarDecl *>(P), false,
+        DeclarationNameInfo(P->getDeclName(), RS->getLocStart()),
+        P->getASTContext().VoidTy, ExprValueKind::VK_RValue);
+    reportUse(DRE, P);
```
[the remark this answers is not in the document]
llvm-beanz: I don't think we should suppress this for entry functions. If an `out` parameter in an entry function is not written to and the caller expects it to be written to, an error occurs in PSO creation. We can't know at compile time if the `out` is intended to be written or not, so we should err on the side of always warning.

I've added the `[maybe_unused]` attribute to allow users to annotate `out` parameters to omit them from the end of function implicit uses, but I think we should force users to explicitly opt-out on a per-parameter basis.

tools/clang/lib/Analysis/UninitializedValues.cpp
```diff
@@ -824,6 +885,37 @@ static bool runOnBlock(const CFGBlock *block, const CFG &cfg,
     if (Optional<CFGStmt> cs = I->getAs<CFGStmt>())
       tf.Visit(const_cast<Stmt*>(cs->getStmt()));
   }
+
+  // HLSL Change Begin - Treat `out` parameters as uninitialized values.
+  // If this block has no successors and does not have a terminator stmt which
+  // we would have handled above.
+  if (block->succ_size() == 0 && !block->getTerminator()) {
```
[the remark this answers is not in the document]
llvm-beanz: This case occurs when a function (usually `void`) doesn't have an explicit return statement at the end of the function. It isn't strictly related to `noreturn`. This is an extremely common case where out parameters are used, so omitting it would significantly limit the usefulness of the diagnostic.

tools/clang/lib/Analysis/UninitializedValues.cpp
```diff
@@ -916,10 +1010,44 @@ void clang::runUninitializedVariablesAnalysis(
...
+  for (auto P = ExitBlock.pred_begin() ; P != ExitBlock.pred_end(); ++P) {
+    if (auto Term = (*P)->getTerminator()) {
+      if (isa<ReturnStmt>(*Term))
+        continue;
+    }
+    AllHandled = false;
+    break;
+  }
+  if (!AllHandled) {
+    // Create a dummy use DeclRefExpr for all the `out` params.
```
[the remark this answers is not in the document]
llvm-beanz: One of those was left in by mistake. I've removed it. I'll investigate a further refactoring in the morning.

tools/clang/lib/Sema/SemaDecl.cpp
```diff
@@ -10347,7 +10347,7 @@ void Sema::DiagnoseUnusedParameters(ParmVarDecl * const *Param,
     if (!(*Param)->isReferenced() && (*Param)->getDeclName() &&
         !(*Param)->hasAttr<UnusedAttr>()) {
       Diag((*Param)->getLocation(), diag::warn_unused_parameter)
-        << (*Param)->getDeclName();
+          << (*Param)->getDeclName();
```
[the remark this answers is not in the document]
llvm-beanz: Doh! I initially had a different warning for completely unused `out` parameters which was logged here. I removed that in the latest code opting to keep the traditional unused parameter warning.

tools/clang/lib/Sema/AnalysisBasedWarnings.cpp
```diff
@@ -2006,9 +2017,13 @@ AnalysisBasedWarnings::IssueWarnings(sema::AnalysisBasedWarnings::Policy P,
...
   }
 
+  // HLSL Change Start - Generate warnings for uninitialized out params.
   if (!Diags.isIgnored(diag::warn_uninit_var, D->getLocStart()) ||
       !Diags.isIgnored(diag::warn_sometimes_uninit_var, D->getLocStart()) ||
-      !Diags.isIgnored(diag::warn_maybe_uninit_var, D->getLocStart())) {
+      !Diags.isIgnored(diag::warn_maybe_uninit_var, D->getLocStart()) ||
+      !Diags.isIgnored(diag::warn_hlsl_uninit_out_param, D->getLocStart()) ||
+      !Diags.isIgnored(diag::warn_hlsl_sometimes_uninit_out_param, D->getLocStart())) {
+    // HLSL Change End  - Generate warnings for uninitialized out params.
```
pow2clk, line 2026: This looks like if we enable one of these, we get them all. I'm not saying that's wrong, but I want to be sure we're consciously making that decision

tools/clang/test/HLSL/out-param-diagnostics.hlsl
```diff
@@ -0,0 +1,116 @@
...
+
+// Verify attribute annotation to opt out of uninitialized parameter analysis.
+void UnusedOutput([maybe_unused] out int Val) {}
+
+void UsedMaybeOutput([maybe_unused] out int Val) { // expected-note{{variable 'Val' is declared here}}
+  Val += Val; // expected-warning{{parameter 'Val' is uninitialized when used here}}
+}
+
+void MaybeUsedMaybeUnused([maybe_unused] out int Val, int Cnt) { // expected-note{{variable 'Val' is declared here}}
+  if (Cnt % 2) // expected-warning{{parameter 'Val' is used uninitialized whenever 'if' condition is fals}} expected-note{{remove the 'if' if its condition is always true}}
```
pow2clk, line 114: Is there an existing test we can modify that runs through all the sometimes possibilities involving loops and switches and such?

tools/clang/test/HLSLFileCheck/hlsl/functions/arguments/maybe_unused_out.hlsl
```diff
@@ -0,0 +1,10 @@
+// RUN: %dxc -T lib_6_4 %s -ast-dump | FileCheck %s
+// Verify attribute annotation to opt out of uninitialized parameter analysis.
+
+void UnusedOutput([maybe_unused] out int Val) {}
```
pow2clk, line 4: Do we need to test this for every possible way this error can be produced? Maybe a variant of the above that ensures this sidesteps any such warnings?

I think we need a test for the disabling flag

pow2clk (commented):
I'm still interested in @tex3d's opinion on this direction given the responses. Most of my suggestions are minor

tools/clang/lib/Sema/AnalysisBasedWarnings.cpp
llvm-beanz: If you enable one of those warnings you need to perform the analysis (which also generates them). Calling `Diag(...)` on an ignored diagnostic still suppresses the message, so you'll only get the output from the ones that are enabled.

This check exists to skip the AST traversal analysis if all the warnings generated by it are ignored.

tools/clang/test/HLSLFileCheck/hlsl/functions/arguments/maybe_unused_out.hlsl
llvm-beanz: Adding more robust testing around the attribute makes sense.

Generally clang doesn't add per-warning or per-warning group tests for disabling warnings because the general infrastructure is tested, and that adds a lot of potential tests with minimal added value.

Since DXC likely isn't running the warning disabling tests, it probably also makes sense to add tests for that case too.

tools/clang/test/HLSL/out-param-diagnostics.hlsl
llvm-beanz: There is no test that we can just grab from Clang that runs through all the permutations. I can port over some of the cases from the other uninitialized values tests to enhance this coverage, but since many of clang's tests use pointers, references and C++ object types none of the tests are likely to work out of the box.

tools/clang/lib/Analysis/UninitializedValues.cpp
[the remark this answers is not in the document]
llvm-beanz: I was trying simplify reporting for this case (but didn't add it to the tests):

```c++
void fn(out float F, bool B) {
  if(B)
    return;
  else
    return;
```

The idea being if all control flow blocks that fall off the end of the function have explicit return statements we shouldn't separately diagnose the end of the function.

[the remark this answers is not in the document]
llvm-beanz: Adding a test for this, it is actually important. Take this example:

```c++
void Fn(out Float F, bool B) {
  if (B) {
    F = 1.0;
    return;
  } else {
    F = 2.0;
    return;
}
```

We should not emit any diagnostic here.

tools/clang/lib/Sema/AnalysisBasedWarnings.cpp
pow2clk: TIL that! Very useful information for something else I'm working on!

pow2clk (approved):
Looks good to me!

llvm-beanz:
> I might have missed it, but I didn't notice any tests for `[maybe_unused]` with structs. Should we add a test that shows it works for structs and for struct members?
> 
> I think a lot of entry functions tend to work with structs.

I can add a test for structs. It shouldn’t make any difference because the type of the variable isn’t taken into account in the analysis, but it is a good additional test to have.

It doesn’t have member-level tracking, and that would be a significant enhancement to add so we’d need to scope it as separate work.

MarijnS95:
@llvm-beanz correct me if I'm wrong, but this PR seems to describe uninitialized `out` parameters _within_ functions. For me it is now (on the latest master at https://github.com/microsoft/DirectXShaderCompiler/commit/667fb773cc2fb40bc31dba6d53e5230a196838ff) failing to compile callers to `Interlocked*` functions:

```hlsl
        uint original;
        buffer.InterlockedAdd(offset, value, original);
        return original;
```

```
bindless.hlsl:110:46: error: variable 'original' is uninitialized when used here [-Werror,-Wuninitialized]
        buffer.InterlockedAdd(offset, value, original);
                                             ^~~~~~~~
bindless.hlsl:109:22: note: initialize the variable 'original' to silence this warning
        uint original;
                     ^
                      = 0
```

I thought the whole point of `out` parameters _and by extension this diagnostic_ is to know and force that the function will initialize `original` here, and it should thus not need to be pre-initialized by the caller.

MarijnS95:
Doing a dirty bisect by taking `libdxcompiler.so` from the parent of this merge commit https://github.com/microsoft/DirectXShaderCompiler/commit/22b392ab8f421ce34c8842f1e07382109067f7db shows that the issue is gone, so it is likely this PR (or a followup one).

llvm-beanz:
Hey @MarijnS95, you just found another wonderful place where our ASTs are wrong and we're doing crazy things with them 😭 .

The AST for the function decl for InterlockedAdd is:
```
 CXXMethodDecl 0x1210567e8 <<invalid sloc>> <invalid sloc> used InterlockedAdd 'void (unsigned int, unsigned int, unsigned int &)' extern
|-TemplateArgument type 'void'
|-TemplateArgument type 'unsigned int'
|-TemplateArgument type 'unsigned int'
|-TemplateArgument type 'unsigned int'
|-ParmVarDecl 0x121056670 <<invalid sloc>> <invalid sloc> InterlockedAdd 'unsigned int'
|-ParmVarDecl 0x1210566e0 <<invalid sloc>> <invalid sloc> byteOffset 'unsigned int'
|-ParmVarDecl 0x121056750 <<invalid sloc>> <invalid sloc> value 'unsigned int &&__restrict'
`-HLSLIntrinsicAttr 0x1210568a0 <<invalid sloc>> Implicit "op" "" 252
```

Note the r-value reference for the `original` parameter instead of an l-value reference... That signifies a use rather than a write so we emit the diagnostic 😞 .

I'm working on a bunch of changes to fix AST-generation to fix these kinds of mistakes in the ASTs. Can you file an issue for this specific case so that I can track it against a user-reported problem?

MarijnS95:
> Hey @MarijnS95, you just found another wonderful place where our ASTs are wrong and we're doing crazy things with them :sob: .

Hehe, my bad, sorry for turning the wrong stone :). Thanks for the blazing-fast reply!

> 
> The AST for the function decl for InterlockedAdd is:
> 
> ```
>  CXXMethodDecl 0x1210567e8 <<invalid sloc>> <invalid sloc> used InterlockedAdd 'void (unsigned int, unsigned int, unsigned int &)' extern
> |-TemplateArgument type 'void'
> |-TemplateArgument type 'unsigned int'
> |-TemplateArgument type 'unsigned int'
> |-TemplateArgument type 'unsigned int'
> |-ParmVarDecl 0x121056670 <<invalid sloc>> <invalid sloc> InterlockedAdd 'unsigned int'
> |-ParmVarDecl 0x1210566e0 <<invalid sloc>> <invalid sloc> byteOffset 'unsigned int'
> |-ParmVarDecl 0x121056750 <<invalid sloc>> <invalid sloc> value 'unsigned int &&__restrict'
> `-HLSLIntrinsicAttr 0x1210568a0 <<invalid sloc>> Implicit "op" "" 252
> ```
> 
> Note the r-value reference for the `original` parameter instead of an l-value reference... That signifies a use rather than a write so we emit the diagnostic :disappointed: .

How do you make `dxc` print this amazing useful AST? Would help if I could include such info from the get-go next time.

> I'm working on a bunch of changes to fix AST-generation to fix these kinds of mistakes in the ASTs. Can you file an issue for this specific case so that I can track it against a user-reported problem?

Sounds nice! Consider it done at https://github.com/microsoft/DirectXShaderCompiler/issues/5093.

llvm-beanz:
> Hehe, my bad, sorry for turning the wrong stone :). Thanks for the blazing-fast reply!

No, totally fair. The magic black box that is supposed to turn your code into a program just started yelling at you like the crazy person on the subway. I’m waist deep in the AST trying to make sense of it, so it is very much on my mind this week.

> How do you make `dxc` print this amazing useful AST? Would help if I could include such info from the get-go next time.

There are two flags, `-ast-dump` and `-ast-dump-implicit`. The former dumps the AST just for the code in the source files DXC is parsing, the later also dumps all the implicitly-generated AST structures from the compiler. The later is a massive dump of text, but it is what you need to spot things like this because they are errors in the compiler-generated ASTs.

> Sounds nice! Consider it done at #5093.

Thank you!

MarijnS95:
I wasn't about to just give up and commit the `= 0` suggestion, that will definitely raise eyebrows in review. But then we have more pressing issues now with the latest compiler repeatedly SEGFAULTing. Will have to tune some tools on our end to get sensible stacktraces out :sweat_smile:. Good luck on the AST rework meanwhile, especially knowing that you were also working on the HLSL frontend in upstream LLVM and must undoubtedly come across code and issues that were already solved/cleaned/refactored there :grin:

Thanks for the commands, those did not surface when running `dxc --help | rg -i ast` :cold_sweat:!

PR #5050, opened 2023-02-23
Added -opt-disable sink to disable instruction sinking in instcombine

adam-yang:
In some cases, sinking instructions can significantly extend register lifetime. Added option -opt-disable sink in the same vein as -opt-disable "gvn" so shader authors can have some degree of control over this behaviour.


lib/HLSL/DxilLoopDeletion.cpp
```diff
@@ -48,7 +53,7 @@ bool DxilLoopDeletion::runOnFunction(Function &F) {
   legacy::FunctionPassManager SimplifyPM(F.getParent());
   SimplifyPM.add(createCFGSimplificationPass());
   SimplifyPM.add(createDeadCodeEliminationPass());
-  SimplifyPM.add(createInstructionCombiningPass());
+  SimplifyPM.add(createInstructionCombiningPass(/*HLSL No sink*/m_HLSLNoSink));
```
dmpots, line 56: Do we need test coverage for this? Is there a case where instcombine can run in the dead loop pass and sink things when `-opt-disable sink` is given?

Or is this change just needed because instcombine takes an argument now?

tools/clang/test/HLSLFileCheck/passes/llvm/instcombine/sink.hlsl
```diff
@@ -0,0 +1,38 @@
...
+    bool br[6];
+    float a[6];
+}
+
+float main() : SV_Target {
+    float val = 0;
+    float ret = 0;
+    // CHECK: br
+    if (br[0]) {
+        // CHECK: select
```
dmpots, line 16: Just to confirm, the change to instcombine is not restricted to selects, right? It will stop sinking of all instruction types?

tools/clang/test/HLSLFileCheck/passes/llvm/instcombine/sink.hlsl
```diff
@@ -0,0 +1,38 @@
+// RUN: %dxc %s /T ps_6_0 -opt-disable sink | FileCheck %s
```
dmpots, line 1: I think you have the sink.hlsl and nosink.hlsl test cases reversed (i.e. sink should be enabled here based on the file name).

tools/clang/test/HLSLFileCheck/passes/llvm/instcombine/nosink.hlsl
```diff
@@ -0,0 +1,38 @@
+// RUN: %dxc %s /T ps_6_0 -opt-enable sink | FileCheck %s
```
dmpots, line 1: I think it would be good to have a RUN line without an explicit `-opt-` flag for sink to show what the expected default is.

dmpots (approved):
LGTM. Follows existing patterns for `-opt-enable` and adding pass-specific options.

tools/clang/test/HLSLFileCheck/passes/llvm/instcombine/sink.hlsl
adam-yang: That's correct. It stops sinking everything.

lib/HLSL/DxilLoopDeletion.cpp
adam-yang: Yes good point. It did sink things in here too. I'll add a test for it.

tools/clang/test/HLSLFileCheck/passes/llvm/instcombine/sink.hlsl
adam-yang: Fixed.

tools/clang/test/HLSLFileCheck/passes/llvm/instcombine/nosink.hlsl
adam-yang: Done.

lib/HLSL/DxilLoopDeletion.cpp
adam-yang: Added a test. Also realized the options are not properly hooked up to -Odump. Fixed it and added a test.

lib/Transforms/IPO/PassManagerBuilder.cpp
```diff
@@ -365,12 +365,7 @@ void PassManagerBuilder::populateModulePassManager(
     addExtensionsToPM(EP_EnabledOnOptLevel0, MPM);
 
     // HLSL Change Begins.
-    addHLSLPasses(HLSLHighLevel, OptLevel,
-      this->HLSLOnlyWarnOnUnrollFail,
-      this->StructurizeLoopExitsForUnroll,
-      this->HLSLEnableLifetimeMarkers,
-      this->HLSLExtensionsCodeGen,
-      MPM);
+    addHLSLPasses(MPM);
```
pow2clk, line 368: So much cleaner!

lib/Transforms/InstCombine/InstructionCombining.cpp
```diff
@@ -3028,7 +3030,7 @@ PreservedAnalyses InstCombinePass::run(Function &F,
   auto *LI = AM->getCachedResult<LoopAnalysis>(F);
 
   // FIXME: The AliasAnalysis is not yet supported in the new pass manager
-  if (!combineInstructionsOverFunction(F, Worklist, nullptr, AC, TLI, DT, LI))
+  if (!combineInstructionsOverFunction(F, Worklist, false /*HLSL Change*/, nullptr, AC, TLI, DT, LI))
```
pow2clk, line 3033: It's a nit and maybe not worth delaying further, but we often put something like /*NoSink*/ on literal arguments

pow2clk (approved):
Looks good! One very minor nit and a bit of kudos for cleanup

lib/Transforms/InstCombine/InstructionCombining.cpp
adam-yang: Fixed!

PR #5053, opened 2023-02-23
PIX: Preserve shader-access-tracking flag for descriptor-heap-indexed…

jeffnn:
PIX recently changed to stop running the instruction-numbering pass before running the pass that's being changed here. This meant that no instructions had an instruction ordinal, so we were always hitting the deleted code. The trouble with that is that the "InstructionOrdinalndicator" flag has a secondary task of denoting an access of a descriptor-heap-indexed resource, and the lack of this flag is breaking some PIX functionality. It's ok to delete this- it was only ever a "just in case" kind of check.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/pix/DynamicSamplerOutOfBounds.hlsl
+++ b/tools/clang/test/HLSLFileCheck/pix/DynamicSamplerOutOfBounds.hlsl
@@ -21,4 +21,4 @@ float4 Main() : SV_Target
 
 // Out of bounds sampler access should be at offset 512
 // CHECK: call void @dx.op.bufferStore.i32(
-// CHECK:i32 512, i32 undef, i32 16777216
+// CHECK:i32 512, i32 undef, i32 134217728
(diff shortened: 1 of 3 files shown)


tools/clang/test/HLSLFileCheck/pix/DynamicSamplerOutOfBounds.hlsl
```diff
@@ -21,4 +21,4 @@ float4 Main() : SV_Target
 
 // Out of bounds sampler access should be at offset 512
 // CHECK: call void @dx.op.bufferStore.i32(
-// CHECK:i32 512, i32 undef, i32 16777216
+// CHECK:i32 512, i32 undef, i32 134217728
```
tex3d, line 24: Note: you can use `CHECK-SAME:` to check something on the same line that the prior `CHECK:` matched.  This can help eliminate both accidental matches, and make it clear where the snippet is supposed to be to future editors of the test.

tex3d (approved):

tools/clang/test/HLSLFileCheck/pix/DynamicSamplerOutOfBounds.hlsl
jeffnn: Thanks for the tip!

PR #5054, opened 2023-02-23
Break test dependencies on mangling (Part 1)

llvm-beanz:
This is the simplest update, changing FileCheck-based tests to use a regular expression for matching symbol names rather than explicitly matching mangled symbols. The regex cases here are _extremely_ flexible skipping calling convention, parameter, and target-specific mangled information.

### DIFF
--- a/tools/clang/test/CodeGenHLSL/lib_no_alias.hlsl
+++ b/tools/clang/test/CodeGenHLSL/lib_no_alias.hlsl
@@ -1,7 +1,7 @@
 // RUN: %dxc -T lib_6_3 -auto-binding-space 11 -default-linkage external %s | FileCheck %s
 
 // Make sure out param has no-alias.
-// CHECK: float @"\01?test@@YAMMUT@@AIAV?$matrix@M$01$01@@M@Z"(float %a, %struct.T* noalias nocapture %t, %class.matrix.float.2.2* noalias nocapture dereferenceable(16) %m, float %b)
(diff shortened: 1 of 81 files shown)


tools/clang/test/CodeGenHLSL/lib_no_alias.hlsl
```diff
@@ -1,7 +1,7 @@
 // RUN: %dxc -T lib_6_3 -auto-binding-space 11 -default-linkage external %s | FileCheck %s
 
 // Make sure out param has no-alias.
-// CHECK: float @"\01?test@@YAMMUT@@AIAV?$matrix@M$01$01@@M@Z"(float %a, %struct.T* noalias nocapture %t, %class.matrix.float.2.2* noalias nocapture dereferenceable(16) %m, float %b)
+// CHECK: float @"\01?test{{[@$?.A-Za-z0-9_]+}}"(float %a, %struct.T* noalias nocapture %t, %class.matrix.float.2.2* noalias nocapture dereferenceable(16) %m, float %b)
```
python3kgae, line 4: Will the \01? at begging of function name change in future mangling?

llvm-beanz: No, that's part of the Microsoft ABI and we won't be getting rid of that just extending it.

python3kgae (approved):

tools/clang/test/CodeGenHLSL/lib_no_alias.hlsl
bfavela: Would it make more sense to use '*' (0 or more) instead of '+'? Or will name mangling always exist in some form or another?

bfavela (approved):

PR #5057, opened 2023-02-24
Add separate target for spirv unit test.

python3kgae:
This is to help enable lit by default.

Allow running spirv unit test separately.


llvm-beanz:
I don't think we should do this.

This would be at best a temporary solution, and we could instead merge #4801, which automatically adds finer grained test targets.

PR #5058, opened 2023-02-24
Generate SharedLibAffix.h into the include directory

hekota:
It is currently generated into lib\DxcSupport build output directory. Since it is a header, it should go into include/dxc/Support build output directory with other generated .h and .inc files.

### DIFF
--- a/lib/DxcSupport/CMakeLists.txt
+++ b/lib/DxcSupport/CMakeLists.txt
@@ -14,9 +14,9 @@ add_llvm_library(LLVMDxcSupport
 #generate header with platform-specific library name
 configure_file(
 ${LLVM_MAIN_SRC_DIR}/lib/DxcSupport/SharedLibAffix.inc
-SharedLibAffix.h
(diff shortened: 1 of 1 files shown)


dmpots:
I think typically we do expect generated files to go in the build directory. If they go in the source directory it would show up in `git status` and friends.

What is the problem with generating the header in the build directory?

hekota:
Sorry, my description was not clear. I am moving the generated file from lib\DxcSupport build output directory into the build output directory with dxc includes (include/dxc/Support), which already contains other generated files such as HLSLOptions.inc. I have updated the description to make it clear. It will not be generated into the source directory.

dmpots:
> Sorry, my description was not clear. I am moving the generated file from lib\DxcSupport build output directory into the build output directory with dxc includes (include/dxc/Support), which already contains other generated files such as HLSLOptions.inc. I have updated the description to make it clear. It will not be generated into the source directory.

Ah, ok. Yes that makes sense now. I misunderstood the change before. LGTM.

dmpots (approved):

PR #5060, opened 2023-02-24
Add separate test for hlsl cmds

python3kgae:
This is to help enable lit by default.
Allow running hlsl cmds test separately.


llvm-beanz:
Or... we could just merge this change:
https://github.com/microsoft/DirectXShaderCompiler/pull/4801

Which adds test suite targets for all the subdirectories.

PR #5061, opened 2023-02-24
[NFC] Fix typo in FileCheck line.

python3kgae:
FileCheck will skip typo lines like CHECK-LABLE.
Fix the typo so we got things tested.


pow2clk (approved):
Apart from a missing commit message, this looks good!

I really hope nothing fails as a result of this. Because these lines are unrecognized, filecheck would have just skipped merrily past them, possibly missing failures as a result 😟

llvm-beanz (approved):

PR #5062, opened 2023-02-24
Break test dependencies on mangling (Part 2)

llvm-beanz:
This is the second part of my changes to break test dependencies on symbol mangling.

This change rewrites a handful of the tests in ValidationTest.cpp that rely on rewriting IR from compiled shaders to be invalid. The new versions of the tests use simplified IR as the source instead of HLSL. Each test has the HLSL source it was derived from captured in code comments.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/validation/CallableParamIsStruct.ll
+++ b/tools/clang/test/HLSLFileCheck/validation/CallableParamIsStruct.ll
@@ -1,4 +1,4 @@
-; RUN: %dxv %s | FileCheck %s
+; RUN: %dxilver 1.3 | %dxv %s | FileCheck %s
 
 ; CHECK: Function: BadCallable: error: Argument 'f' must be a struct type for callable shader function 'BadCallable'.
 
@@ -25,7 +25,7 @@ attributes #0 = { nounwind }
 !dx.entryPoints = !{!3, !4}
 
 !0 = !{i32 1, i32 3}
-!1 = !{i32 1, i32 7}
+!1 = !{i32 1, i32 3}
 !2 = !{!"lib", i32 6, i32 3}
 !3 = !{null, !"", null, null, null}
 !4 = !{void (float*)* @BadCallable, !"BadCallable", null, null, !5}
--- a/tools/clang/test/HLSLFileCheck/validation/CallableParamIsStruct.ll
+++ b/tools/clang/test/HLSLFileCheck/validation/CallableParamIsStruct.ll
@@ -1,0 +1,33 @@
+; RUN: %dxv %s | FileCheck %s
+
+; CHECK: Function: BadCallable: error: Argument 'f' must be a struct type for callable shader function 'BadCallable'.
+
+; Test based on IR generated from the following HLSL:
+; struct Param { float f; };
+; [shader("callable")]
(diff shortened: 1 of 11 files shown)


tools/clang/test/HLSLFileCheck/validation/CallableParamIsStruct.ll
```diff
@@ -0,0 +1,33 @@
+; RUN: %dxv %s | FileCheck %s
```
python3kgae, line 1: Could we add validator version in the %dxv command line in case validation rule changed in the future?

llvm-beanz: I'd prefer not to because the rule is valid on future versions until we change that. I'd rather we add a version when we change it so that we get a failure on changing the rules.

tex3d: We can require a minimum version with `%dxilver`, but not a maximum.  We should be using `%dxilver` for validation rules added/changed at some version.  In ValidationTest.cpp, we use something like: `if (m_ver.SkipDxilVersion(1, 3)) return;` to enforce a minimum of 1.3.  For tests ported from there, we should retain this through the use of `%dxilver`.

Instead of having a maximum version, we should probably just remove the test, since it's only testing something we no longer want.

When adding/updating validation tests, it would be best to verify that they pass (and skip appropriately) when run against that last few external validators we released.

tools/clang/test/HLSLFileCheck/validation/CallableParamIsStruct.ll
```diff
@@ -0,0 +1,33 @@
...
+
+attributes #0 = { nounwind }
+
+!dx.version = !{!0}
+!dx.valver = !{!1}
+!dx.shaderModel = !{!2}
+!dx.entryPoints = !{!3, !4}
+
+!0 = !{i32 1, i32 3}
+!1 = !{i32 1, i32 7}
```
tex3d, line 28: This is the validator version.  This value (1.7) a problem, since it will restrict the test to working when run against validator version 1.7 (the most recently released version).

With tests in `ValidationTest.cpp` that use the compile->disassemble->modify->assemble->validate path, the validator version would be set to the currently attached one (including the version from prior releases of external validators, against which we run tests regularly).  This would appropriately impact IR (metadata) in certain key cases.

With tests written in IR, we would likely need to permute this with different validator versions for some tests.  For most tests, we can probably get away with setting this to the minimum validator version that supported the rule we are testing.  For instance, I think that would be 1.3 for this test instead of 1.7.

However, just setting this to a minimum version may not be sufficient for some tests, since there can be different decisions/allowances in the validator depending on whether the IR was originally targeting an older validator version or the current one being run.  We may also need to test different DXIL versions for some scenarios.

Permuting versions within a single test can't quite be done unless we have a way to replace certain IR input elements with variables from a RUN line.  If that was supported, we could use different RUN lines, different `%dxilver` requirements for each, and different `-D` variables for FileCheck.

python3kgae (approved):

llvm-beanz:
Closing this since I can re-create it if needed.

Issue #4918, opened 2023-01-07
DXCAPI.H makes hard-coded path assumptions on Linux


walbourn:
The DXCAPI.H header has:

```
#include "dxc/Support/WinAdapter.h"
```

while the DXCISENSE.H header has:

```
#include "Support/WinAdapter.h"
```

Which both make explicit (and mismatched) assumptions about the layout that doesn't match the contents of the ZIP. This can be worked around, but in a future update it would be best to make both of these:

```
#include "WinAdapter.h"
```

Keenuts:
HI,

Thanks for the report! We'll add that to the cleanup list we have in queue on DXC codebase. (If the change is as simple are changing an include path, PRs are always welcome 😊)

PR #5066, opened 2023-02-28
Move WinAdapter into interface directory

pow2clk:
Since it is shipped and relied upon, the WinAdapter header belongs in the release directory. This removes the need to include the Support directory for building purposes

revises fix to #4918


llvm-beanz (approved):
LTGM

PR #5069, opened 2023-02-28
Delete some dead/unreachable code in CGCall

llvm-beanz:
This code is all unreachable. Reverting this strips out some cascading changes adding an unused parameter to a few functions.


hekota (approved):
LGTM!

dmpots (approved):
Maybe add an `[NFC]` tag for this change?

pow2clk (approved):
I lament that the 80 column limit will require so many double comments to highlight a single line of HLSL-specific code 😒 , but the change is good!

PR #5074, opened 2023-03-02
Add #pragma region support

### LINKED ISSUE #676: Preprocessor pragmas
It would be nice if the HLSL preprocessor could recognize:

- pragma once
- pragma region
- pragma endregion

llvm-beanz:
Add the pragma region handler to the HLSL pragma handlers list.

Fixes #676

Sorry for the slow turn around time.


dmpots (approved):
The bug also mentions `pragma once`. Do we already support that?

hekota (approved):
LGTM!

llvm-beanz:
> The bug also mentions `pragma once`. Do we already support that?

Yea, `pragma once` support has always been in DXC. The GitHub diff view cuts it off, but it is 3 lines above the ones I added.

PR #5078, opened 2023-03-03
Only track scalar `out` params

llvm-beanz:
The UninitializedValues analysis doesn't correctly handle arrays or pointer objects, so we should not enable `out` parameter analysis for those types. Additionally, because the HLSL ASTs incorrectly represent structure `out` parameters as non-reference types, the analysis will generate incorrect diagnostics for uses of structs in call expressions. There isn't really an easy solution for that. This change takes a conservative approach to addressing those issues by disabling the analysis for values that we can't correctly identify.

This change also disables the "suggest initialization" fix-it for `out` parameters. Default initialization for `out` parameters doesn't work, and would be confusing.

### DIFF
--- a/tools/clang/lib/Analysis/UninitializedValues.cpp
+++ b/tools/clang/lib/Analysis/UninitializedValues.cpp
@@ -35,8 +35,10 @@ using namespace clang;
 
 static bool isTrackedVar(const VarDecl *vd, const DeclContext *dc) {
   // HLSL Change Begin - Treat `out` parameters as uninitialized values.
-  if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>())
-    return true;
+  if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>()) {
+    QualType ty = vd->getType().getNonReferenceType();
+    return ty->isScalarType() || ty->isVectorType() || ty->isRecordType();
+  }
   // HLSL Change End - Treat `out` parameters as uninitialized values.
   if (vd->isLocalVarDecl() && !vd->hasGlobalStorage() &&
       !vd->isExceptionVariable() && !vd->isInitCapture() &&
@@ -87,6 +89,8 @@ void DeclToIndex::computeMap(const DeclContext &dc) {
     if (isTrackedVar(vd, &dc))
       map[vd] = count++;
     // HLSL Change Begin - Treat `out` parameters as uninitialized values.
+    else
+      continue;
     // Keep HLSL parameters in a separate index.
     if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>())
       hlslOutParams.push_back(vd);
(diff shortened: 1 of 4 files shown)


dmpots:
Does this also handle the case of passing through multiple function calls?

This shader
```
// dxc /Tcs_6_0 unused.hlsl
RWStructuredBuffer<uint> obuf;
struct S { uint a; uint b;};
void baz(out S s) {
    s.a = 0;
    s.b = 1;
}
void bar(out S s) {
    baz(s);
}
[RootSignature("DescriptorTable(UAV(u0))")]
[numthreads(64,1,1)]
void main()
{
    S s;
    bar(s);
    obuf[1] = s.a;
}
```

Currently generates a warning

```
.\unused.hlsl:8:9: error: parameter 's' is uninitialized when used here [-Werror,-Wparameter-usage]
    baz(s);
```

Even though the struct is fully initialized.

python3kgae:
How about matrix type?
Does it treat as struct here?

tools/clang/lib/Analysis/UninitializedValues.cpp
```diff
@@ -35,8 +35,10 @@ using namespace clang;
 
 static bool isTrackedVar(const VarDecl *vd, const DeclContext *dc) {
   // HLSL Change Begin - Treat `out` parameters as uninitialized values.
-  if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>())
-    return true;
+  if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>()) {
+    QualType ty = vd->getType().getNonReferenceType();
+    return ty->isScalarType() || ty->isVectorType() || ty->isRecordType();
```
python3kgae, line 40: If only thing we want to skip is array, is it possible to change to return !ty->isArrayType()?

llvm-beanz: If you look a couple lines down, the analysis was only built for these types, so we shouldn't try to use it for more. In our case vectors and matrices are all record types anyways because of how we wrap them in class decls.

tools/clang/lib/Analysis/UninitializedValues.cpp
```diff
@@ -35,8 +35,15 @@ using namespace clang;
...
-  if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>())
-    return true;
+  if (vd->hasAttr<HLSLOutAttr>() && !vd->hasAttr<HLSLInAttr>()) {
+    QualType ty = vd->getType().getNonReferenceType();
+    // FIXME: HLSL doesn't model parameter passing of structs correctly in the
+    // AST. Struct types are passed by value in the AST regardless of whether
+    // they are out or inout, which results in LValueToRValue casts in the AST
+    // which look like uses (because they are). For now this analysis can
+    // generate too many false positives for structs.
+    return ty->isScalarType();
```
python3kgae, line 45: Is it possible to use hlsl::IsHLSLVecMatType(ty) to support vector/matrix type?

llvm-beanz: No. We represent vector and matrix types in the AST incorrectly so we can't analyze them accurately.

bogner (approved):

pow2clk (approved):
Looks good!

PR #5085, opened 2023-03-07
Change SharedLibAffix.h include path to eliminate the need to change he build setting

hekota:
The file dxcapi.use.cpp is used in other projects and would require build changes if the #include did not change.


tex3d (approved):

llvm-beanz (approved):
LGTM

PR #5087, opened 2023-03-08
Switch HLSL matrix major attribute to TypeAttr

python3kgae:
This remove extra information saved in AST.
Now major information only saved in AttributedType.


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12888,7 +12879,6 @@ void Sema::TransferUnusualAttributes(Declarator &D, NamedDecl *NewDecl) {
   if (!getLangOpts().HLSL) {
     return;
   }
-
```
tex3d, line 12891: Why one deleted blank line here?

python3kgae: Fixed. It is caused revert local change.

tex3d (approved):
Looks good.

PR #5094, opened 2023-03-10
No special case for unannotated params

### LINKED ISSUE #5093: `out` parameter diagnostics affect uninitialized variables at the _call site_
A recent PR to prevent leaving `out` parameters uninitialized in function calls (https://github.com/microsoft/DirectXShaderCompiler/pull/5047) seems to have escaped its cage and is now also disallowing uninitialized _arguments into_ functions with `out` paramameters. See for example this (IMO totally valid) `Interlocked*` invocation:

```hlsl
        uint original;
        buffer.InterlockedAdd(offset, value, original);
        return original;
```

```
bindless.hlsl:110:46: error: variable 'original' is uninitialized when used here [-Werror,-Wuninitialized]
        buffer.InterlockedAdd(offset, value, original);
                                             ^~~~~~~~
bindless.hlsl:109:22: note: initialize the variable 'original' to silence this warning
        uint original;
                     ^
                      = 0
```

(Compiler from https://github.com/microsoft/DirectXShaderCompiler/commit/667fb773cc2fb40bc31dba6d53e5230a196838ff)

I thought the whole point of `out` parameters _and by extension this diagnostic_ is to know and force that the function will initialize `original` here, and it should thus not need to be pre-initialized by the caller.

_Originally posted by @MarijnS95 in https://github.com/microsoft/DirectXShaderCompiler/issues/5047#issuecomment-1462949361_, cc @llvm-beanz

llvm-beanz:
Initially the change here was forcing unannotated parameters to be treated as uses (`in` parameters). Instead this changes to not handle them explicitly.

Most `in` parameter uses should produce `LValueToRValue` casts which will be identified as uses through other means, but by removing this special case we can support unannotated AST nodes from generated ASTs through this analysis based on their type and usage alone.

Fixes #5093

### DIFF
--- a/tools/clang/lib/Analysis/UninitializedValues.cpp
+++ b/tools/clang/lib/Analysis/UninitializedValues.cpp
@@ -497,8 +497,8 @@ void ClassifyRefs::VisitCallExpr(CallExpr *CE) {
         ParmVarDecl *PD = FD->getParamDecl(ParamIdx);
         bool HasIn = PD->hasAttr<HLSLInAttr>();
         bool HasInOut = PD->hasAttr<HLSLInOutAttr>();
-        // If we have an in annotation or no annotation (implcit in), this is a
-        // use not an initialization.
+        // If we have an explicit `in` or `inout` annotation we should treat
+        // this as a use, otherwise leave it up to C/C++ rules.
         if(HasIn || HasInOut)
           classify(*I, Use);
       }
--- a/tools/clang/lib/Analysis/UninitializedValues.cpp
+++ b/tools/clang/lib/Analysis/UninitializedValues.cpp
@@ -496,11 +496,10 @@ void ClassifyRefs::VisitCallExpr(CallExpr *CE) {
       if (FD->getNumParams() > ParamIdx) {
         ParmVarDecl *PD = FD->getParamDecl(ParamIdx);
         bool HasIn = PD->hasAttr<HLSLInAttr>();
-        bool HasOut = PD->hasAttr<HLSLOutAttr>();
         bool HasInOut = PD->hasAttr<HLSLInOutAttr>();
         // If we have an in annotation or no annotation (implcit in), this is a
         // use not an initialization.
-        if(!HasOut || HasIn || HasInOut)
+        if(HasIn || HasInOut)
           classify(*I, Use);
       }
     }
--- a/tools/clang/test/HLSL/out-param-diagnostics.hlsl
+++ b/tools/clang/test/HLSL/out-param-diagnostics.hlsl
@@ -179,7 +179,8 @@ void NoAnnotationIsUse(out int V) { // expected-note{{variable 'V' is declared h
 RWByteAddressBuffer buffer;
 
 // No expected diagnostic here. InterlockedAdd is not annotated with HLSL
-// parameter annotations.
+// parameter annotations, so we fall back to C/C++ rules, which don't treat
+// reference passed parameters as uses.
 void interlockWrapper(out uint original) {
   buffer.InterlockedAdd(16, 1, original);
 }
--- a/tools/clang/test/HLSL/out-param-diagnostics.hlsl
+++ b/tools/clang/test/HLSL/out-param-diagnostics.hlsl
@@ -170,6 +170,20 @@ void MaybeUsedMaybeUnused([maybe_unused] out int Val, int Cnt) { // expected-not
     Val = 1;
 } // expected-note{{uninitialized use occurs here}}
 
+void Use(int V) {}
+
+void NoAnnotationIsUse(out int V) { // expected-note{{variable 'V' is declared here}}
+  Use(V); // expected-warning{{parameter 'V' is uninitialized when used here}}
+}
+
+RWByteAddressBuffer buffer;
+
+// No expected diagnostic here. InterlockedAdd is not annotated with HLSL
+// parameter annotations.
+void interlockWrapper(out uint original) {
+  buffer.InterlockedAdd(16, 1, original);
+}
+
 // Neither of these will warn because we don't support element-based tracking.
 void UnusedSizedArray(out uint u[2]) { }
 void UnusedUnsizedArray(out uint u[]) { }


tools/clang/test/HLSL/out-param-diagnostics.hlsl
```diff
@@ -170,6 +170,20 @@ void MaybeUsedMaybeUnused([maybe_unused] out int Val, int Cnt) { // expected-not
...
+void Use(int V) {}
+
+void NoAnnotationIsUse(out int V) { // expected-note{{variable 'V' is declared here}}
+  Use(V); // expected-warning{{parameter 'V' is uninitialized when used here}}
+}
+
+RWByteAddressBuffer buffer;
+
+// No expected diagnostic here. InterlockedAdd is not annotated with HLSL
+// parameter annotations.
```
bogner, line 182: This comment is confusing. Isn't the reason there's no diagnostic because it *is* annotated with parameter annotations?

tools/clang/test/HLSL/out-param-diagnostics.hlsl
```diff
@@ -170,6 +170,20 @@ void MaybeUsedMaybeUnused([maybe_unused] out int Val, int Cnt) { // expected-not
     Val = 1;
 } // expected-note{{uninitialized use occurs here}}
 
+void Use(int V) {}
```
bogner, line 173: Let's also add a case where we call a function with a signature like `void Def(out int V)` to show that there's no warning there.

tools/clang/lib/Analysis/UninitializedValues.cpp
```diff
@@ -496,11 +496,10 @@ void ClassifyRefs::VisitCallExpr(CallExpr *CE) {
       if (FD->getNumParams() > ParamIdx) {
         ParmVarDecl *PD = FD->getParamDecl(ParamIdx);
         bool HasIn = PD->hasAttr<HLSLInAttr>();
-        bool HasOut = PD->hasAttr<HLSLOutAttr>();
         bool HasInOut = PD->hasAttr<HLSLInOutAttr>();
         // If we have an in annotation or no annotation (implcit in), this is a
         // use not an initialization.
```
bogner, line 501: Is this comment stale? I don't think it describes what's happening accurately even before your change.

bogner (approved):
lg pending some editing in the comments and maybe another test case

tools/clang/test/HLSL/out-param-diagnostics.hlsl
llvm-beanz: Confusingly and probably incorrectly it isn’t annotated. My comment lacks sufficient details here. This change ignores unannotated parameters and instead leaves it up to the rules for C/C++, which ignores values used as pass-by-reference parameters.

tools/clang/lib/Analysis/UninitializedValues.cpp
tex3d: Is `in` added elsewhere by default for user function params?

llvm-beanz: No, this only covers explicit annotations. Pass-by-value parameters (including explicitly annotated `in`) should automatically be treated as uses because they’ll pretty much always have an `LValueToRValue` cast which is a use.

Thinking on it we need the `in` check here is to cover the case of `in out` vs `inout`, but otherwise we wouldn’t need to check `in` at all.

tools/clang/test/HLSL/out-param-diagnostics.hlsl
tex3d: Yeah, that is confusing.  I would think that the reason there's no diagnostic is that the third parameter *is* annotated with `out`, like an ordinary case where no warning should be emitted (and where we should have a test for that as you pointed out above).  Here's the intrinsic definition:
```
void [[]] InterlockedAdd64(in uint byteOffset, in u64 value, out any_int64 original) : interlockedadd_immediate;
```
Or is it the case that even though this has `out` in the intrinsic definition, it doesn't add the `HLSLOutAttr` to the parameter decl like an ordinary HLSL function would, yet we still need to test to ensure that it doesn't think this is a use of the uninitialized value?

llvm-beanz: Yes, we don’t translate the annotations from the intrinsic definitions to the annotations in the AST, we just use them to generate the qual types for the arguments. `InterlockedAdd` ends up with a signature something like `InterlockedAdd(int, int, int && __restrict)`, with no HLSL annotations on the `ParamVarDecl`s

I don’t think many of these are generating correct AST representations, but since they’re completely missing the HLSL annotations the best thing we can do is rely on C/C++ rules.

MarijnS95:
Many thanks, this solves the issue on our end! However, it also surfaces a bunch of issues on the following pattern (that I discarded yesterday by writing in #5093 that the `out` parameter should/could be relied on to initialize the argument at the call site):

```hlsl
bool GetOptionalThing(out float result) {
    if (...) {
        result = ...;
        return true;
    }
    return false;
}
```

with DXC obviously complaining that `result` is left uninitialized at `return false;`.

How should this pattern be implemented in HLSL? I worked around the warning by adding `result = result;` but that seems wrong given that it's an `out`, not an `inout`. In most of these cases I think we expect `result` to already contain a sensible value that we _do not_ want to overwrite, perhaps it should just be an `inout`.

MarijnS95:
@llvm-beanz thanks for clarifying that in such a super-detailed manner :star_struck:! I don't want the warnings to go away at all but rather solve them correctly, less UB means less time spent debugging (we code in Rust for a reason, too). Sounds like I'll be determining on a case-by-case basis whether to use `inout` or rearchitect the APIs a bit.

tex3d (approved):
Looks good!

MarijnS95:
> Self-assigning the variable tells the analysis to ignore the variable (as you discovered).

In hindsight, is this intentional? I'd have a similar "uninitialized" error when using the `out` parameter on the right of `=`, and this seems like a poor/cheaty way to get out of the undefined-behaviour check without actually addressing undefined behaviour.

PR #5101, opened 2023-03-17
Add script to compile to intermediate IR for easier pass test authoring

tex3d:
This script automates some operations to make it easier to write IR tests:
  1. Gets the pass list for an HLSL compilation using -Odump
  2. Compiles HLSL with -fcgl and outputs to intermediate IR
  3. Collects list of passes before the desired pass and adds -hlsl-passes-pause to write correct metadata
  4. Invokes dxopt to run passes on -fcgl output and write bitcode result
  5. Disassembles bitcode to .ll file for use as a test
  6. Inserts RUN line with -hlsl-passes-resume and desired pass

### DIFF
--- a/utils/hct/ExtractIRForPassTest.py
+++ b/utils/hct/ExtractIRForPassTest.py
@@ -1,0 +1,138 @@
+"""ExtractIRForPassTest.py - extract IR just before selected pass would be run
+
+This script automates some operations to make it easier to write IR tests:
+  1. Gets the pass list for an HLSL compilation using -Odump
+  2. Compiles HLSL with -fcgl and outputs to intermediate IR
+  3. Collects list of passes before the desired pass and adds
+     -hlsl-passes-pause to write correct metadata
+  4. Invokes dxopt to run passes on -fcgl output and write bitcode result
+  5. Disassembles bitcode to .ll file for use as a test
+  6. Inserts RUN line with -hlsl-passes-resume and desired pass
+
+Usage:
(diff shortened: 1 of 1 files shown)


utils/hct/ExtractIRForPassTest.py
```diff
@@ -0,0 +1,142 @@
...
+
+def Usage():
+  print(__doc__)
+  exit(3)
+
+class Options(object):
+  def __init__(self):
+    self.instance = 1
+
+def ParseArgs(args):
```
adam-yang, line 39: Might be cleaner to use argparse.

adam-yang (approved):
Thanks for this change! Very convenient.

utils/hct/ExtractIRForPassTest.py
tex3d: I don't know argparse well enough to get it to do what I want here.

Most of the arguments could be switched to using named options, and you can use `--` to separate options for the script from the rest of the options for DXC, because I don't want argparse parsing those options.  But the only way I've found to show `--` in the help text is to prefix the metavar string with it.

Then I can't seem to get the description text to format the way I want.  It just deletes all newlines and puts everything in one line.

I feel like argparse is useful for more complex scripts with various options, but for a tiny one-off like this, it seems like more of a pain than it's worth.

tex3d: Ok, looks like you can pass `formatter_class=argparse.RawDescriptionHelpFormatter` to prevent it from messing up the formatting of the description.  I'll still have to change the code handling the desired pass if I want to use this though.

PR #5102, opened 2023-03-17
Fixed a crash due to an incorrect call to dropAllReferences

adam-yang:
Fixed a crash when there's addrspacecast going into a getelementptr, which goes into a dbg.decare. The HLSL SROA pass was incorrectly calling dropAllReferences on the addrspacecast ConstExpr, which invalidates the getelementptr ConstExpr (sets the operand to nullptr).

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -1674,9 +1674,6 @@ void RemoveUnusedInternalGlobalVariable(Module &M) {
         Value *User = *(UserIt++);
         if (Instruction *I = dyn_cast<Instruction>(User)) {
           I->eraseFromParent();
-        } else {
-          ConstantExpr *CE = cast<ConstantExpr>(User);
-          CE->dropAllReferences();
         }
       }
       GV->eraseFromParent();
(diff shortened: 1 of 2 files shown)


lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -1674,9 +1674,6 @@ void RemoveUnusedInternalGlobalVariable(Module &M) {
         Value *User = *(UserIt++);
         if (Instruction *I = dyn_cast<Instruction>(User)) {
           I->eraseFromParent();
-        } else {
-          ConstantExpr *CE = cast<ConstantExpr>(User);
-          CE->dropAllReferences();
```
tex3d, line 1679: Could there still be constant users which are used by instructions that would prevent `GV->eraseFromParent()` from succeeding?

tex3d: Oh maybe it's not possible because it's doing:
```
if (isa<ConstantExpr>(user) && user->user_empty())
  continue;
```
Then, if it's a constant user that's not `user_empty()`, it will fall through to `onlyStoreUse = false` and `break;`, preventing it from going down this code path.

dmpots (approved):
LGTM

tex3d (approved):

PR #5106, opened 2023-03-20
Release 1.7.2212

Murdock98:
Trying to make windows work. It says i have shader model 0.0


llvm-beanz (changes requested):
I’m not sure what you’re trying to do here, but we’re not going to merge main into our release branch. The release branch is tested and released on Windows so your description doesn’t help me understand what you’re trying to accomplish.

PR #5110, opened 2023-03-22
Fix ScoreCast for class types, fix more struct assumptions in SemaHLSL

tex3d:
SemaHLSL had various locations where getAsStructureType or isStructureType were used, when classes should have been handled in the same way.

This change uses getAs<RecordType> instead of getAsStructureType and replaces isStructureType with isStructureOrClassType in locations that should handle structs and classes.

Also, it deletes unused HLSL methods in clang::Sema found while adding unit tests.
New tests added and existing tests updated.

### DIFF
--- a/tools/clang/lib/AST/Type.cpp
+++ b/tools/clang/lib/AST/Type.cpp
@@ -446,27 +446,6 @@ const RecordType *Type::getAsStructureType() const {
   return nullptr;
 }
 
-const RecordType *Type::getAsStructureOrClassType() const {
-  // If this is directly a structure type, return it.
-  if (const RecordType *RT = dyn_cast<RecordType>(this)) {
-    RecordDecl *RD = RT->getDecl();
-    if (RD->isStruct() || RD->isClass() || RD->isInterface())
-      return RT;
-  }
-
-  // If the canonical form of this type isn't the right kind, reject it.
-  if (const RecordType *RT = dyn_cast<RecordType>(CanonicalType)) {
-    RecordDecl *RD = RT->getDecl();
-    if (!(RD->isStruct() || RD->isClass() || RD->isInterface()))
-      return nullptr;
-
-    // If this is a typedef for a structure type, strip the typedef off without
-    // losing all typedef information.
-    return cast<RecordType>(getUnqualifiedDesugaredType());
-  }
-  return nullptr;
-}
-
 const RecordType *Type::getAsUnionType() const {
   // If this is directly a union type, return it.
   if (const RecordType *RT = dyn_cast<RecordType>(this)) {
--- a/tools/clang/lib/AST/Type.cpp
+++ b/tools/clang/lib/AST/Type.cpp
@@ -446,6 +446,27 @@ const RecordType *Type::getAsStructureType() const {
   return nullptr;
 }
 
+const RecordType *Type::getAsStructureOrClassType() const {
+  // If this is directly a structure type, return it.
+  if (const RecordType *RT = dyn_cast<RecordType>(this)) {
+    RecordDecl *RD = RT->getDecl();
+    if (RD->isStruct() || RD->isClass() || RD->isInterface())
+      return RT;
+  }
+
+  // If the canonical form of this type isn't the right kind, reject it.
+  if (const RecordType *RT = dyn_cast<RecordType>(CanonicalType)) {
+    RecordDecl *RD = RT->getDecl();
+    if (!(RD->isStruct() || RD->isClass() || RD->isInterface()))
+      return nullptr;
+
+    // If this is a typedef for a structure type, strip the typedef off without
+    // losing all typedef information.
+    return cast<RecordType>(getUnqualifiedDesugaredType());
+  }
+  return nullptr;
+}
+
 const RecordType *Type::getAsUnionType() const {
   // If this is directly a union type, return it.
   if (const RecordType *RT = dyn_cast<RecordType>(this)) {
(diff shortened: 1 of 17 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12781,63 +12775,10 @@ Decl* Sema::getActiveHLSLBuffer() const
   return HLSLBuffers.empty() ? nullptr : HLSLBuffers.back();
 }
 
-Decl *Sema::ActOnHLSLBufferView(Scope *bufferScope, SourceLocation KwLoc,
```
python3kgae, line 12784: Is this just removing unused code which is not related to the struct change?

python3kgae (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5095,7 +5095,7 @@ class HLSLExternalSource : public ExternalSemaSource {
               llvm::ArrayRef<TemplateArgument>(TST->getArgs(),
                                                TST->getNumArgs()));
         }
-        if (const RecordType* recordType = argType->getAsStructureType()) {
+        if (const RecordType* recordType = argType->getAsStructureOrClassType()) {
```
llvm-beanz, line 5098: Does this need to be only struct or class types, or could it just be `getAsRecordType()`?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6963,7 +6963,7 @@ QualType HLSLExternalSource::GetNthElementType(QualType type, unsigned index) {
     return (index == 0) ? type : QualType();
   case AR_TOBJ_COMPOUND: {
     // TODO: consider caching this value for perf
-    const RecordType *recordType = type->getAsStructureType();
+    const RecordType *recordType = type->getAsStructureOrClassType();
```
llvm-beanz, line 6966: We don't acknowledge the possibility of failure here (which is fine since we've already are under `AR_TOBJ_COMPOUND`, so this could be `getAsRecordType()`.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -7744,13 +7744,8 @@ bool HLSLExternalSource::IsConversionToLessOrEqualElements(
...
   if (sourceTypeInfo.ShapeKind == AR_TOBJ_COMPOUND ||
       GetTypeObjectKind(sourceType) == AR_TOBJ_COMPOUND) {
-    const RecordType *targetRT = targetType->getAsStructureType();
-    if (!targetRT)
-      targetRT = dyn_cast<RecordType>(targetType);
-
-    const RecordType *sourceRT = sourceType->getAsStructureType();
-    if (!sourceRT)
-      sourceRT = dyn_cast<RecordType>(sourceType);
+    const RecordType *targetRT = targetType->getAsStructureOrClassType();
```
llvm-beanz, line 7747: This looks like it was doing the same as just calling `getAsRecordType()`

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11633,7 +11627,7 @@ bool FlattenedTypeIterator::pushTrackerForType(QualType type, MultiExprArg::iter
   case ArTypeObjectKind::AR_TOBJ_OBJECT: {
     if (m_source.IsSubobjectType(type)) {
       // subobjects are initialized with initialization lists
-      recordType = type->getAsStructureType();
+      recordType = type->getAsStructureOrClassType();
```
llvm-beanz, line 11630: This also could just be `getAsRecordType()`

tools/clang/lib/AST/Type.cpp
```diff
@@ -446,6 +446,27 @@ const RecordType *Type::getAsStructureType() const {
   return nullptr;
 }
 
+const RecordType *Type::getAsStructureOrClassType() const {
```
llvm-beanz, line 449: It looks to me like we don't actually need this, we could instead just use `getAsRecordType()` (and maybe `isStructureOrClassType()`) in all the places we're using this.

tex3d: There is no such method as `getAsRecordType()`.  If you mean `getAs<RecordType>()`, I suppose that would be fine.  I wasn't sure if it was truly equivalent, but I guess it only additionally includes unions, which haven't been enabled yet.  I'm not sure we want all these cases to allow unions when those are enabled though.  I guess I just haven't thought through each of these scenarios with the potential of unions.

It just seemed odd that there was a `getAsStructureType()` that would only work if it was declared `struct` but nothing that would work if it was `struct` or `class`, yet those two are almost semantically identical in C++ (and are semantically identical in HLSL), while unions are not.  I think just about everywhere we used `getAsStructureType()` we really wanted a RecordType for either `struct` or `class`, except for some weird built-in HLSL object detection that uses the class path to potentially drill into a template specialization.

tex3d: By the way, `getAs<RecordType>()` was my first thought as to how this should be fixed.  But then I thought maybe that would be too broad, and what we really want is either a `struct` or `class` type.

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d: Yes, unused code that had a change related to this struct change, which I tried and failed to add a test for because the code was dead.

tools/clang/lib/AST/Type.cpp
llvm-beanz: Makes sense. My reading of the code was that all the places where this was being used are pretty well already assured to be struct or class. I suppose there might be some complications with union types, but we can look at that when we introduce them.

llvm-beanz (approved):
LGTM

PR #5111, opened 2023-03-23
Replace deprecated std::iterator inheritance

Flakebi:
std::iterator is deprecated since C++17. E.g. when using the standard library of gcc 12, deprecation warnings are shown.

There is some background info here:
https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/


Flakebi:

tex3d (approved):

PR #5112, opened 2023-03-23
build: add exception header

Keenuts:
This file uses std::current_exception, so it should include <exception>

### DIFF
--- a/lib/IR/Value.cpp
+++ b/lib/IR/Value.cpp
@@ -33,6 +33,7 @@
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/raw_ostream.h"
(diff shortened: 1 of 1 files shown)


s-perron (approved):

lib/IR/Value.cpp
```diff
@@ -33,6 +33,7 @@
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/ManagedStatic.h"
 #include "llvm/Support/raw_ostream.h"
+#include <exception>
```
llvm-beanz, line 36: Please add `// HLSL Change` to mark this as changed by DXC.

Keenuts: Oh right, this. done thanks!

llvm-beanz (approved):

PR #5114, opened 2023-03-24
[template] Support empty template arg list on TypeAliasTemplateDecl.

python3kgae:
This is to help add sperate type for row major and column major matrix.

The issue is only support ClassTemplateSpecializationDecl in getHLSLDefaultSpecialization. It causes compile error on t22 in following shader.

  template<typename T=float, int i=0>
  struct test {
    T d;
  };

  template<typename T=float>
  using test2 = test<T, 1>;

  test2 t22;

The fix is using Sema::CheckTemplateIdType to generate the default specialization for all TemplateDecl.


tex3d (approved):
LGTM.

llvm-beanz (approved):

PR #5121, opened 2023-03-30
Enhance `globallycoherent` mismatch diagnostics

### LINKED ISSUE #4537: Returning a `globallycoherent` resource from a function results in a "global coherent mismatch" warning
## Premise

While other RHIs still do not have dynamic resource indexing, it's convenient to implement a common platform-neutral interface with getters for various resource types. However, an issue seems to arise specifically when trying to return a `globallycoherent` annotated UAV from a function. While the DXIL generated correctly annotates the resource as `globallycoherent`, the shader compilation result produces warnings. Please let me know if I missed something in the correct way to do this, thanks!

## Issue Description

Consider the following snippet:

```
// test.hlsl
RWTexture2D<float> get_texture() {
    globallycoherent RWTexture2D<float> output = ResourceDescriptorHeap[0];
    return output;
}
[numthreads(1, 1, 1)]
void CSMain() {
    get_texture()[uint2(0, 0)] = 1.0;
}
```

Compiling this as `dxc.exe -E CSMain -T cs_6_6 -HV 2021 test.hlsl > test.dxil` produces the following warning:

```
.\test.hlsl::4:5: warning: global coherent mismatch
    return output;
    ^
```

which is the result of [this diagnostic](https://github.com/microsoft/DirectXShaderCompiler/blob/50c0199c17b0b7530cbc0bdddadc3697eaece695/tools/clang/lib/CodeGen/CGHLSLMS.cpp#L2602).

However, there doesn't appear to be a clean way to propagate the annotation to the function's return type. For example:

```
globallycoherent RWTexture2D<float> get_texture() {
    globallycoherent RWTexture2D<float> output = ResourceDescriptorHeap[0];
    return output;
}
[numthreads(1, 1, 1)]
void CSMain() {
    get_texture()[uint2(0, 0)] = 1.0;
}
```

resolves the diagnostic above, but leaves you with a new warning:

```
.\test.hlsl:1:1: warning: 'globallycoherent' attribute only applies to variables [-Wignored-attributes]
globallycoherent RWTexture2D<float> get_texture()
^
```

This warning may be suppressed using `-Wno-ignored-attributes` but would globally turn off the warning.

llvm-beanz:
This change moves the `globallycoherent` mismatch diagnostics out of code generation and into semantic analysis. Additionally the diagnostic is enhanced to provide more information in the text about the mismatch, and to fire in source locations more directly attributable to the location of the mismatch.

This change also allows the `globallycoherent` annotation to apply to functions, where it applies to the return type of the function, and mismatches can be diagnosed on assignment of the return value and on the return statements.

Fixes #4537.

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7676,7 +7676,7 @@ def err_hlsl_incorrect_bind_semantic: Error< // Is an error in fxc in some cases
 def warn_hlsl_unary_negate_unsigned : Warning<
   "unary negate of unsigned value is still unsigned">,
   InGroup<Conversion>, DefaultWarn;
-def warn_hlsl_impcast_gl_mismatch : Warning<
+def warn_hlsl_impcast_glc_mismatch : Warning<
   "implicit conversion from %0 to %1 %select{loses|adds}2 globallycoherent annotation">,
   InGroup<Conversion>, DefaultWarn;
 def warn_hlsl_narrowing : Warning<
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7676,6 +7676,9 @@ def err_hlsl_incorrect_bind_semantic: Error< // Is an error in fxc in some cases
 def warn_hlsl_unary_negate_unsigned : Warning<
   "unary negate of unsigned value is still unsigned">,
   InGroup<Conversion>, DefaultWarn;
+def warn_hlsl_impcast_gl_mismatch : Warning<
+  "implicit conversion from %0 to %1 %select{loses|adds}2 globallycoherent annotation">,
+  InGroup<Conversion>, DefaultWarn;
 def warn_hlsl_narrowing : Warning<
   "conversion from larger type %0 to smaller type %1, possible loss of data">,
   InGroup<Conversion>, DefaultWarn;
--- a/tools/clang/lib/Sema/SemaChecking.cpp
+++ b/tools/clang/lib/Sema/SemaChecking.cpp
@@ -48,6 +48,19 @@ SourceLocation Sema::getLocationOfStringLiteralByte(const StringLiteral *SL,
                                Context.getTargetInfo());
 }
 
+// HLSL Change begin
+static void CheckGloballyCoherentMismatch(Sema &S, Expr *SrcExpr,
+                                          QualType DstTy, SourceLocation CC) {
+  if(!hlsl::IsHLSLResourceType(SrcExpr->getType()))
+    return;
+  bool SrcGL = hlsl::HasHLSLGloballyCoherent(SrcExpr->getType());
+  bool DstGL = hlsl::HasHLSLGloballyCoherent(DstTy);
+  if (SrcGL != DstGL)
+    S.Diag(CC, diag::warn_hlsl_impcast_gl_mismatch)
+        << SrcExpr->getType() << DstTy << /*loses|adds*/ DstGL;
+}
+// HLSL Change end
+
 /// Checks that a call expression's argument count is the desired number.
 /// This is useful when doing custom type-checking.  Returns true on error.
 static bool checkArgCount(Sema &S, CallExpr *call, unsigned desiredArgCount) {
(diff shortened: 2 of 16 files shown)


tools/clang/test/HLSL/globallycoherent-mismatch.hlsl
```diff
@@ -0,0 +1,47 @@
+// RUN: %clang_cc1 -fsyntax-only -ffreestanding -verify %s
+
+RWByteAddressBuffer NonGCBuf;
```
python3kgae, line 3: Could we a test case for resource array?

tools/clang/test/HLSL/globallycoherent-mismatch.hlsl
```diff
@@ -0,0 +1,72 @@
...
+globallycoherent RWByteAddressBuffer GCBuf;
+
+RWByteAddressBuffer NonGCBufArr[2];
+globallycoherent RWByteAddressBuffer GCBufArr[2];
+
```
python3kgae, line 8: Could we have something like this?
static globallycoherent  RWByteAddressBuffer SGCBufArr[2] = NonGCBufArr;
void getNonGCBufArrayArr(inout globallycoherent RWByteAddressBuffer
 TGCBufArr[2]) {
  TGCBufArr = NonGCBufArr;
}

tools/clang/lib/Sema/SemaChecking.cpp
```diff
@@ -48,6 +48,24 @@ SourceLocation Sema::getLocationOfStringLiteralByte(const StringLiteral *SL,
                                Context.getTargetInfo());
 }
 
+// HLSL Change begin
+static void CheckGloballyCoherentMismatch(Sema &S, Expr *SrcExpr,
+                                          QualType DstTy, SourceLocation CC) {
+  QualType SrcTy = SrcExpr->getType();
+  if (SrcTy->isArrayType() && DstTy->isArrayType()) {
+    SrcTy = SrcTy->getAsArrayTypeUnsafe()->getElementType();
```
python3kgae, line 56: Do we want to support multi-dim resource array ?

tools/clang/lib/Sema/SemaDecl.cpp
```diff
@@ -9153,6 +9153,28 @@ void Sema::AddInitializerToDecl(Decl *RealDecl, Expr *Init,
...
   QualType DclT = VDecl->getType(), SavT = DclT;
+
+  // HLSL Change begin
+  // When initializing an HLSL resource type we should diagnose mismatches in
+  // globally coherent annotations _unless_ the source is a dynamic resource
+  // placeholder type where we safely infer the globallycoherent annotaiton.
+  if (getLangOpts().HLSL) {
+    QualType SrcTy = Init->getType();
+    QualType DstTy = DclT;
+    if (SrcTy->isArrayType() && DstTy->isArrayType()) {
```
python3kgae, line 9164: The code here feels similar as CheckGloballyCoherentMismatch, is it possible to modify CheckGloballyCoherentMismatch and share the code?

python3kgae (approved):

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7676,6 +7676,9 @@ def err_hlsl_incorrect_bind_semantic: Error< // Is an error in fxc in some cases
 def warn_hlsl_unary_negate_unsigned : Warning<
   "unary negate of unsigned value is still unsigned">,
   InGroup<Conversion>, DefaultWarn;
+def warn_hlsl_impcast_gl_mismatch : Warning<
```
tex3d, line 7679: The precedent for the abbreviation of `globallycoherent` is `glc` which I think would be preferred over `gl` here.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12142,21 +12142,42 @@ void Sema::DiagnoseHLSLDeclAttr(const Decl *D, const Attr *A) {
...
 }
 
+void Sema::DiagnoseGloballyCoherentMismatch(const Expr *SrcExpr,
+                                            QualType TargetType,
+                                            SourceLocation Loc) {
+  QualType SrcTy = SrcExpr->getType();
+  QualType DstTy = TargetType;
+  if (SrcTy->isArrayType() && DstTy->isArrayType()) {
+    SrcTy = SrcTy->getAsArrayTypeUnsafe()->getElementType();
+    DstTy = DstTy->getAsArrayTypeUnsafe()->getElementType();
```
tex3d, line 12169: Does this work with multi-dim array type?  Wouldn't that have to unwrap the nested arrays?

tex3d (approved):
A nit regarding `gl` abbreviation - suggest `glc` instead, and question regarding multi-dim arrays.  Otherwise, looks good.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
llvm-beanz: Fixed in 64bf4c5d873ec447afecd2d6574398becbb425b6

tools/clang/lib/Sema/SemaDecl.cpp
llvm-beanz: Fixed in 26728cd488ce0c06b7828b10e6e814a6e28ff6bf

tools/clang/lib/Sema/SemaChecking.cpp
llvm-beanz: Fixed in b104eb4b7b15dc4c43eda3ecc8e128296cd4aef7

PR #5125, opened 2023-03-31
Make matrix with different orientation different types.

python3kgae:
Change hlsl matrix type into

namespace matrix.internal {
template
struct matrix {
vector<T, colCount> h[row_count];

  vector<T, colCount> operator[] (int i);
  vector<T, colCount> operator[] (int i) const;

};
}
namespace row.major { template using matrix = matrix.internal::matrix<T, row_count, col_count, 1>; } namespace column.major { template using matrix = matrix.internal::matrix<T, row_count, col_count, 0>; }

template using matrix = matrix.internal::matrix<T, row_count, col_count, 0>;

This is to allow orientation information saved in template.

TODO:
1. remove attributed type for matrix major.
2. generate different IR type for row major and column major matrix.

### DIFF
--- a/tools/clang/include/clang/AST/HlslTypes.h
+++ b/tools/clang/include/clang/AST/HlslTypes.h
@@ -32,6 +32,7 @@ namespace clang {
   class CXXMethodDecl;
   class CXXRecordDecl;
   class ClassTemplateDecl;
+  class TypeAliasTemplateDecl;
   class ExtVectorType;
   class FunctionDecl;
   class FunctionTemplateDecl;
@@ -78,6 +79,11 @@ enum HLSLScalarType {
   HLSLScalarType_float64,
   HLSLScalarType_int8_4packed,
   HLSLScalarType_uint8_4packed
+};
+
+enum HLSLMatrixOrientation {
+  ColumnMajor = 0,
+  RowMajor,
 };
 
 HLSLScalarType MakeUnsigned(HLSLScalarType T);
@@ -307,7 +313,10 @@ ParamModFromAttributeList(_In_opt_ clang::AttributeList *pAttributes);
 void AddHLSLMatrixTemplate(
   clang::ASTContext& context,
   _In_ clang::ClassTemplateDecl* vectorTemplateDecl,
-  _Outptr_ clang::ClassTemplateDecl** matrixTemplateDecl);
+  _Outptr_ clang::ClassTemplateDecl** matrixTemplateDecl,
+  _Outptr_ clang::TypeAliasTemplateDecl **defaultMatrixAliasDecl,
+  _Outptr_ clang::TypeAliasTemplateDecl **rowMajorMatrixAliasDecl,
+  _Outptr_ clang::TypeAliasTemplateDecl **columnMajorMatrixAliasDecl);
 
 void AddHLSLVectorTemplate(
   clang::ASTContext& context, 
@@ -414,6 +423,10 @@ clang::QualType GetHLSLInputPatchElementType(clang::QualType type);
 unsigned GetHLSLInputPatchCount(clang::QualType type);
 clang::QualType GetHLSLOutputPatchElementType(clang::QualType type);
 unsigned GetHLSLOutputPatchCount(clang::QualType type);
+clang::QualType GetHLSLMatrixTypeWithMajor(clang::QualType matType,
+                                           bool isRowMajor, clang::Sema &sema);
+bool IsRowMajorMatrixTemplate(clang::QualType matType);
+bool IsDefaultMajorMatrixTemplate(clang::QualType matType);
 
 bool IsHLSLSubobjectType(clang::QualType type);
 bool GetHLSLSubobjectKind(clang::QualType type, DXIL::SubobjectKind &subobjectKind, 
--- a/tools/clang/include/clang/AST/OperationKinds.h
+++ b/tools/clang/include/clang/AST/OperationKinds.h
@@ -314,6 +314,8 @@ enum CastKind {
   CK_HLSLVectorToMatrixCast,
   CK_HLSLMatrixToVectorCast,
   CK_HLSLDerivedToBase,
+  CK_HLSLRowMajorToColMajor,
+  CK_HLSLColMajorToRowMajor,
   // HLSL ComponentConversion (HLSLCC) Casts:
   CK_HLSLCC_IntegralCast,
   CK_HLSLCC_IntegralToBoolean,
--- a/tools/clang/include/clang/Sema/Overload.h
+++ b/tools/clang/include/clang/Sema/Overload.h
@@ -97,6 +97,8 @@ namespace clang {
     ICK_HLSLVector_Splat,      ///< HLSLVector/Matrix splat
     ICK_HLSLVector_Truncation, ///< HLSLVector/Matrix truncation
     ICK_HLSL_Derived_To_Base,  ///< HLSL Derived-to-base
+    ICK_HLSLRowMajorToColMajor,///< HLSL row major to column major
+    ICK_HLSLColMajorToRowMajor,///< HLSL column major to row major
     // HLSL Change Ends
 
     ICK_Num_Conversion_Kinds   ///< The number of conversion kinds
@@ -160,6 +162,9 @@ namespace clang {
     /// vector or matrix type.  If used, Second must be one of the 
     /// ICK_HLSLVector_* implicit conversion kinds.
     ImplicitConversionKind ComponentConversion : 8;
+    /// ComponentConversion - If this is not ICK_Identity, this describes
+    /// the type of conversion to apply to matrix major.
+    ImplicitConversionKind MatrixMajorConversion : 8;
     // HLSL Change Ends
 
     /// Third - The third conversion can be a qualification conversion.
--- a/tools/clang/lib/AST/ASTContextHLSL.cpp
+++ b/tools/clang/lib/AST/ASTContextHLSL.cpp
@@ -19,6 +19,7 @@
 #include "clang/AST/ExternalASTSource.h"
 #include "clang/AST/HlslBuiltinTypeDeclBuilder.h"
 #include "clang/AST/TypeLoc.h"
+#include "clang/AST/DeclTemplate.h"
 #include "clang/Sema/SemaDiagnostic.h"
 #include "clang/Sema/Sema.h"
 #include "clang/Sema/Overload.h"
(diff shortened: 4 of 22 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -4265,6 +4346,11 @@ class HLSLExternalSource : public ExternalSemaSource {
         DXASSERT(decl->isImplicit(),
                  "otherwise object template decl is not set to implicit");
         return AR_TOBJ_OBJECT;
+      } else if (auto *TAT = dyn_cast<TypeAliasTemplateDecl>(typeRecordDecl)) {
+        if (TAT->getCanonicalDecl() == +m_defaultMatDecl->getCanonicalDecl() ||
```
tex3d, line 4350: What's with the `+` in `+m_defaultMatDecl->getCanonicalDecl()` and on the next line?  Was this a copy-paste from a diff with clang-format on paste incorporating the plus into the expression?

python3kgae: Good catch. Fixed.

pow2clk (commented):
It's only a partial review I'm afraid, but I wanted to give you something to consider

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -347,7 +357,151 @@ void hlsl::AddHLSLMatrixTemplate(ASTContext& context, ClassTemplateDecl* vectorT
...
+                            /*Inline*/ false, SourceLocation(),
+                            SourceLocation(), &context.Idents.get(Name),
+                            /*PrevDecl*/ nullptr);
+  NSDecl->setImplicit();
+  context.getTranslationUnitDecl()->addDecl(NSDecl);
+  return NSDecl;
+}
+
+/// <summary>Adds up-front support for HLSL matrix types (just the template declaration).</summary>
+void hlsl::AddHLSLMatrixTemplate(ASTContext& context, ClassTemplateDecl* vectorTemplateDecl, ClassTemplateDecl** matrixTemplateDecl,
```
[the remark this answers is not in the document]
python3kgae: The function itself is not moved; some functions added before it with part of the content of this function which make the diff hard to follow :(

tools/clang/include/clang/Sema/Overload.h
```diff
@@ -160,6 +162,9 @@ namespace clang {
     /// vector or matrix type.  If used, Second must be one of the 
     /// ICK_HLSLVector_* implicit conversion kinds.
     ImplicitConversionKind ComponentConversion : 8;
+    /// ComponentConversion - If this is not ICK_Identity, this describes
+    /// the type of conversion to apply to matrix major.
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/lib/CodeGen/CodeGenTypes.cpp
```diff
@@ -41,14 +41,17 @@ CodeGenTypes::CodeGenTypes(CodeGenModule &cgm)
 }
 
 CodeGenTypes::~CodeGenTypes() {
+  for (auto *Key : dupMatTys) {
+    RecordDeclTypes.erase(Key);
+    CGRecordLayouts.erase(Key);
+  }
```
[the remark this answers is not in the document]
python3kgae: The hack is removed.

tools/clang/lib/CodeGen/CodeGenTypes.cpp
```diff
@@ -102,6 +105,7 @@ void CodeGenTypes::addRecordTypeName(const RecordDecl *RD,
   // HLSL Change Ends
 
   Ty->setName(OS.str());
+  return Ty->getName() != OS.str();
```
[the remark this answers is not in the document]
python3kgae: The hack is removed.

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -347,7 +357,151 @@ void hlsl::AddHLSLMatrixTemplate(ASTContext& context, ClassTemplateDecl* vectorT
...
+    TypeAliasTemplateDecl **defaultMatrixAliasDecl,
+    TypeAliasTemplateDecl **rowMajorMatrixAliasDecl,
+    TypeAliasTemplateDecl **columnMajorMatrixAliasDecl) {
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -298,21 +299,29 @@ void AddSubscriptOperator(
...
+  // matrix.internal::matrix { ... }
+  BuiltinTypeDeclBuilder typeDeclBuilder(DC, "matrix");
+  TemplateTypeParmDecl *elementTemplateParamDecl =
+      typeDeclBuilder.addTypeTemplateParam("element",
+                                           (QualType)context.FloatTy);
+  NonTypeTemplateParmDecl *rowCountTemplateParamDecl =
+      typeDeclBuilder.addIntegerTemplateParam("row_count", context.IntTy, 4);
+  NonTypeTemplateParmDecl *colCountTemplateParamDecl =
+      typeDeclBuilder.addIntegerTemplateParam("col_count", context.IntTy, 4);
+  // hlsl::HLSLMatrixOrientation, 0 is col major, 2 is row major.
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -3341,6 +3341,11 @@ SpirvInstruction *SpirvEmitter::doCastExpr(const CastExpr *expr,
       return 0;
     }
   }
+  case CastKind::CK_HLSLColMajorToRowMajor:
+  case CastKind::CK_HLSLRowMajorToColMajor: {
+    // FIXME: support major cast.
```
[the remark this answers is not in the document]
python3kgae: Does SPIR-V codeGen assume value are always row_major like dxc?
If the answer is true, then no change is required for this PR.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3845,7 +3845,8 @@ bool SROA_Helper::LowerMemcpy(Value *V, DxilFieldAnnotation *annotation,
           // Check Src only have 1 store now.
           hlutil::PointerStatus SrcPS(Src, size, /*bLdStOnly*/ false);
           SrcPS.analyze(typeSys, bStructElt);
-          if (SrcPS.storedType != hlutil::PointerStatus::StoredType::Stored) {
+          if (SrcPS.storedType != hlutil::PointerStatus::StoredType::Stored ||
```
python3kgae, line 3848: This is to work around the MacOS only fail.
Will create separate PR for it.

python3kgae: The workaround is merged.

tools/clang/include/clang/AST/HlslTypes.h
```diff
@@ -80,6 +82,12 @@ enum HLSLScalarType {
   HLSLScalarType_uint8_4packed
 };
 
+enum HLSLMatrixOrientation {
```
llvm-beanz, line 85: Why don't we re-use the `hlsl::MatrixOrientation` enum out of `DxilTypeSystem.h` instead of creating a new enum?

tools/clang/include/clang/AST/OperationKinds.h
```diff
@@ -314,6 +314,8 @@ enum CastKind {
   CK_HLSLVectorToMatrixCast,
   CK_HLSLMatrixToVectorCast,
   CK_HLSLDerivedToBase,
+  CK_HLSLRowMajorToColMajor,
+  CK_HLSLColMajorToRowMajor,
```
llvm-beanz, line 318: Can you explain why you need a cast kind for this instead of this just being an extension of the FlatConversion?

tools/clang/include/clang/Sema/Overload.h
```diff
@@ -97,6 +97,8 @@ namespace clang {
     ICK_HLSLVector_Splat,      ///< HLSLVector/Matrix splat
     ICK_HLSLVector_Truncation, ///< HLSLVector/Matrix truncation
     ICK_HLSL_Derived_To_Base,  ///< HLSL Derived-to-base
+    ICK_HLSLRowMajorToColMajor,///< HLSL row major to column major
+    ICK_HLSLColMajorToRowMajor,///< HLSL column major to row major
```
llvm-beanz, line 101: Same here. Why does this need an implicit conversion kind instead of just being a FlatConversion?

llvm-beanz (commented):
A few preliminary comments. I think we may need to break down the impact of this change on casting and implicit conversion sequences. I'm concerned if this requires adding an additional conversion step because that is actually a semantic change to the language.

tools/clang/include/clang/Sema/Overload.h
```diff
@@ -160,6 +162,10 @@ namespace clang {
     /// vector or matrix type.  If used, Second must be one of the 
     /// ICK_HLSLVector_* implicit conversion kinds.
     ImplicitConversionKind ComponentConversion : 8;
+
+    /// MatrixOrientationConversion - If this is not ICK_Identity, this
+    /// describes the type of conversion to apply to matrix orientation.
+    ImplicitConversionKind MatrixOrientationConversion : 8;
```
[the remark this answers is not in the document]
python3kgae: Removed.

tools/clang/include/clang/AST/HlslTypes.h
python3kgae: Orientation in HLSLMatrixOrientation is to match the is_row_major in matrix.internal::matrix.
Column is 0, Row is 1.
Default is only used to save shorthanded type and set to 2.

tools/clang/include/clang/AST/OperationKinds.h
python3kgae: It is clearer than reuse FlatConversion.
The data is changed for major change. For FlatConversion, the data is not changed, more like a bitcast.

tools/clang/include/clang/Sema/Overload.h
python3kgae: It is clearer than reuse FlatConversion.
The data is changed for major change. For FlatConversion, the data is not changed, more like a bitcast.

llvm-beanz: My understanding is that majority only impacts the orientation on load/store, not in local memory. If that is correct orientation shouldn't change layout.

(cc @tex3d)

python3kgae: It is changed with this PR. :(
The reason for doing this is hack clang type codeGen to generate same IR type is ugly.
With a llvm IR type named matrix.col, put the data in row matrix will make people confuse.

Maybe we could give a better name so it is not confusing?

lib/HLSL/HLMatrixLowerPass.cpp
llvm-beanz, line 1: The changes here definitely warrant some IR->IR unit tests.

tools/clang/include/clang/AST/HlslTypes.h
llvm-beanz: Why don't we re-use the hlsl::MatrixOrientation enum out of DxilTypeSystem.h instead of creating a new enum?

(you didn't actually explain why you need a new enum that has the same case values)

tools/clang/include/clang/Sema/Overload.h
llvm-beanz: I think that changing the data storage is a problem since the underlying dxil operations for matrices assume that orientation doesn't impact the layout.

tools/clang/lib/CodeGen/CGExpr.cpp
```diff
@@ -3333,6 +3333,12 @@ EmitConditionalOperatorLValue(const AbstractConditionalOperator *expr) {
 /// cast from scalar to union.
 LValue CodeGenFunction::EmitCastLValue(const CastExpr *E) {
   // HLSL Change Begins
+  if (E->getCastKind() == CK_HLSLRowMajorToColMajor ||
+      E->getCastKind() == CK_HLSLColMajorToRowMajor) {
+    // Orientation cast on LValue.
+    // Do the cast when copyin/out by create LValue with target type.
+    return EmitLValue(E->getSubExpr());
```
llvm-beanz, line 3340: This seems like a problem. If the orientation impacts the data layout, we can't create an l-value cast without transposing going in _and_ back out of the l-value because any writes need to be transposed.

llvm-beanz (commented):
I'm concerned a lot about the change in the in-memory orientation.

I'm also think this change needs more testing. Most of the tests here are just updates to the tests that this change caused to fail. This change introduces a lot of code with behavior changes up and down the compiler which should all be tested.

damyanp:

damyanp:

PR #5126, opened 2023-03-31
Fill in MDVals with push_back for maintainability.

bob80905:
To prevent potential out of bounds assignments to the vector in the future, MDVals now uses push_back instead of [] assignment.

### DIFF
--- a/lib/DXIL/DxilMetadataHelper.cpp
+++ b/lib/DXIL/DxilMetadataHelper.cpp
@@ -1667,7 +1667,8 @@ MDTuple *
 DxilMDHelper::EmitDxilFunctionProps(const hlsl::DxilFunctionProps *props,
                                     const Function *F) {
   bool bRayAttributes = false;
-  std::vector<Metadata *> MDVals;
+  SmallVector<Metadata *, 35> MDVals;
+  
 
   MDVals.push_back(ValueAsMetadata::get(const_cast<Function *>(F)));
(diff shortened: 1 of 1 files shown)


tex3d (approved):

lib/DXIL/DxilMetadataHelper.cpp
```diff
@@ -1665,49 +1665,46 @@ void DxilMDHelper::LoadDxilEntryProperties(const MDOperand &MDO,
...
 DxilMDHelper::EmitDxilFunctionProps(const hlsl::DxilFunctionProps *props,
-                                   const Function *F) {
+                                    const Function *F) {
   bool bRayAttributes = false;
-  Metadata *MDVals[30];
-  std::fill(MDVals, MDVals + _countof(MDVals), nullptr);
-  unsigned valIdx = 0;
-  MDVals[valIdx++] = ValueAsMetadata::get(const_cast<Function*>(F));
-  MDVals[valIdx++] = Uint32ToConstMD(static_cast<unsigned>(props->shaderKind));
+  std::vector<Metadata *> MDVals;
```
llvm-beanz, line 1670: Should we use a SmallVector here instead to avoid growing? If we know the maximum possible, or even just a reasonable common size a SmallVector should have a significant performance benefit.

llvm-beanz (approved):
LGTM

PR #5129, opened 2023-04-03
[Rewriter] Use clang::Rewriter for DoSimpleReWrite.

python3kgae:
Use clang::Rewriter to replace tu->print when rewrite.

This is prepared for support matrix orientation as different type.

Many rewriter tests are affected because clang::Rewriter will keep the original shape of the source.


tools/clang/test/HLSL/rewriter/correct_rewrites/effects-syntax_gold.hlsl
```diff
@@ -1,24 +1,132 @@
...
+
+Texture2D tex : register(t1), tex2 : register(t2)
+
+
+
+
+
+
+
+< int foo=1; >
```
tex3d, line 13: I think part of the point of regenerating the code before was to be able to get rid of deprecated effects constructs like the ones in this file, such as: `<effect annotations>`, `{ state blocks }`, `technique {}`, and `pass {}`.  These are allowed on `-HV 2015`, but should not be allowed on newer language versions.

Should we have a rewriter mode that can do what we did before, reconstructing the source from AST without the parts that got ignored/eliminated?

tools/clang/test/HLSL/rewriter/correct_rewrites/shared.hlsl
```diff
@@ -1,22 +1,22 @@
...
-Texture1D<double> g_tex1d;
-Texture1DArray<int> g_tex1dArray;
-Texture2D<float> g_tex2d;
-Texture2DArray<float> g_tex2dArray;
-Texture2DMS<half> g_texture2dms;
-Texture2DMSArray<float> g_texture2dmsArray;
-float main() : SV_Target {
-  return -X + Y;
-}
+shared cbuffer cb0 {
```
tex3d, line 2: See comment about deprecated effects syntax and whether we should have a mode to remove such syntax to help migrate HLSL to new language versions.

tools/clang/test/HLSL/rewriter/correct_rewrites/effects-syntax_gold.hlsl
tex3d: For this, do you think we'd need some sort of AST node that represents deprecated and ignored text so that it can be explicitly removed?

tools/clang/test/HLSL/rewriter/correct_rewrites/includes_gold.hlsl
tex3d, line 1: AppVeyor is failing on this file for some reason (as well as files for RunNoFunctionBodyInclude and RunCppErrors).  It concerns me that we would get different behavior between that and other pipelines.  Is there a difference in how these test the rewriter?

tex3d, line 1: AppVeyor is failing on this file for some reason (as well as files for RunNoFunctionBodyInclude and RunCppErrors).  It concerns me that we would get different behavior between that and other pipelines.  Is there a difference in how these test the rewriter?

python3kgae: Have no idea :(
We need to be able to run the same command used by the remote pipeline locally to debug the issue.

damyanp:

damyanp:

PR #5140, opened 2023-04-11
Support dependent template args in HLSL Object templates

### LINKED ISSUE #5139: DXC doesn't allow dependent template type argument in HLSL object templates
Special HLSL Object template argument checking code doesn't handle dependent types, leading to errors such as:
> error: 'vector<T, 4>' cannot be used as a type parameter

With code like this:
```hlsl
template<typename T>
T foo(Texture2D<vector<T, 4> > tex) {
  return tex[uint2(0,0)].x;
}
```

tex3d:
HLSL objects have a special hook for validating template arguments. Previously, this didn't handle dependent arguments. This change skips dependent arguments, which will be checked once they are concrete.

Fixes #5139


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5121,9 +5121,11 @@ class HLSLExternalSource : public ExternalSemaSource {
       const TemplateArgument &arg = argLoc.getArgument();
```
pow2clk, line 5121: Is the comment above still accurate that this only checks that types are primitive types? I'm not sure it was true when it was committed. :/

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5121,9 +5121,11 @@ class HLSLExternalSource : public ExternalSemaSource {
...
       if (arg.getKind() == TemplateArgument::ArgKind::Type) {
         QualType argType = arg.getAsType();
-        if (!IsValidTemplateArgumentType(argSrcLoc, argType, requireScalar)) {
-          // NOTE: IsValidTemplateArgumentType emits its own diagnostics
-          return true;
+        if (!argType->isDependentType()) {
```
pow2clk, line 5124: A short comment explaining that these will be checked when they become concrete might be helpful

tools/clang/test/HLSLFileCheck/hlsl/template/DependentTypeInObjectParam.hlsl
```diff
@@ -0,0 +1,25 @@
...
+
+// CHECK: call %dx.types.ResRet.f32 @dx.op.textureLoad.f32
+// CHECK: call %dx.types.ResRet.f16 @dx.op.textureLoad.f16
+
+template<typename T>
+T foo(Texture2D<vector<T, 4> > tex) {
+  return tex[uint2(0,0)].x;
+}
+
+Texture2D<float4> T;
```
pow2clk, line 20: Re-using T here confused me a litte

pow2clk (approved):
Nothing really functional. It looks like a valid change. Just suggesting non functional additions.

tools/clang/test/HLSLFileCheck/hlsl/template/DependentTypeInObjectParamInvalid.hlsl
```diff
@@ -0,0 +1,21 @@
+// RUN: %dxc -E main -T vs_6_6 -HV 2021 -enable-16bit-types -ast-dump %s | FileCheck %s
+
+// CHECK: error: 'MyStruct' cannot be used as a type parameter where a scalar is required
```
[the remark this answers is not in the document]
tex3d: I agree, but I could not supply the shader model for the verifier test at the time, so this test wouldn't compile because of the `ResourceDescriptorHeap` usage.

On that note, I had found what looked like a verifier test that it looked like this would fit well into, but it also used `ResourceDescriptorHeap`, and wasn't being used at all (`template-diag-deferred.hlsl`).  I tried adding that to VerifierTest.cpp, along with a RUN line and ran into this trouble.  I also noticed the VerifierHelper.py was out of date for the list of tests, and especially the set of tests that wouldn't be compatible with fxc.

Since then I found another way to formulate the test that doesn't depend on `ResourceDescriptorHeap`, so I'll update it with that.

As a side note, I think we likely need to keep these dynamic resource built-in global identifiers available regardless of shader model, and check shader model to validate legal usage instead.

tools/clang/test/HLSLFileCheck/hlsl/template/DependentTypeInObjectParamInvalid.hlsl
```diff
@@ -0,0 +1,21 @@
...
+// CHECK-NEXT: Texture2D<vector<T, 4> > tex = ResourceDescriptorHeap[0];
+
+// CHECK: note: in instantiation of function template specialization 'foo<MyStruct>' requested here
+// CHECK-NEXT: return foo<MyStruct>().f;
+
+template<typename T>
+T foo() {
+  Texture2D<vector<T, 4> > tex = ResourceDescriptorHeap[0];
```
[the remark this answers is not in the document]
tex3d: Because struct cannot be used in a vector that way.  This is to check that the skip for dependent type doesn't prevent this from being checked once we have a concrete type.

I couldn't do this check the same way with the previous version of the function that took a texture object parameter because you would get an error on constructing the texture object when attempting to supply the parameter.  Since then I found a better way to test this without the dynamic resource, that's closer to the positive case but with the invalid type.  I'll update the test with that.

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d: I will remove the inaccurate part.

tools/clang/test/HLSLFileCheck/hlsl/template/DependentTypeInObjectParamInvalid.hlsl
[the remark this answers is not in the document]
tex3d: I changed the test (moved to VerifierTest as well).  Hopefully that alleviates any concerns.

tools/clang/test/HLSLFileCheck/hlsl/template/DependentTypeInObjectParamInvalid.hlsl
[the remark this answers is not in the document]
tex3d: I changed the test and moved it to a verifier test.

tex3d:
Note: I put the VerifierHelper.py and existing test updates into a separate commit.  So, if we want to split this change into a separate PR, we can do that:

https://github.com/microsoft/DirectXShaderCompiler/pull/5140/commits/d876827d987322792644a11d58b3f68e65e3b556#diff-95d82eff019d6a62f4085bd2d9ad3e7f4e3f15ebb362bd89fc851bdd38920d07

tools/clang/test/HLSL/out-param-diagnostics.hlsl
```diff
@@ -1,22 +1,22 @@
 // RUN: %clang_cc1 -fsyntax-only -verify %s
 
-void UnusedEmpty(out int Val) {} // expected-warning{{parameter 'Val' is uninitialized when used here}} expected-note{{variable 'Val' is declared here}}
+void UnusedEmpty(out int Val) {} // expected-note {{variable 'Val' is declared here}} expected-warning {{parameter 'Val' is uninitialized when used here}} fxc-pass {{}}
```
llvm-beanz, line 3: The only FXC checks here are pretty nonsensical, so if we're going to have this test auto-updated, we should exclude this on FXC since this is testing a feature that doesn't exist in FXC.

tools/clang/test/HLSL/write-const-arrays.hlsl
```diff
@@ -13,38 +13,38 @@ StructuredBuffer<uint> g_robuf;
 Texture1D<uint> g_tex;
 uint g_cbuf[4];
 
-const groupshared uint gs_val[4];
+const groupshared uint gs_val[4];                           /* fxc-error {{X3012: 'gs_val': missing initial value}} */
```
llvm-beanz, line 16: I also question whether there is value in having this file tested by FXC since it is just a mess of warnings, and since we're not actively developing FXC any regression here is of limited utility.

llvm-beanz (commented):
In a general sense I don't particularly like the VerifierHelper... I think it lends itself to large unreviewable changes to diagnostics.

At a minimum I think we should pull VerifierHelper-driven changes that aren't dependent on this change out into a separate PR so that they can be reviewed independently of the bug fix.

tools/clang/test/HLSL/out-param-diagnostics.hlsl
tex3d: Changes moved to #5148.

tools/clang/test/HLSL/write-const-arrays.hlsl
tex3d: Changes moved to #5148.

Even this particular error highlights where DXC is lacking.  A `const groupshared` global *should* have an initializer, otherwise it has no way to be used without reading undef values.  I believe DXC's support for initializers on `groupshared` globals is also missing (unless that was fixed more recently).

tex3d: We should discuss this in the other PR #5148, but we have an FXC exclusion list for tests we don't want to try on FXC.  I think this file is still valuable to compare with FXC.  Some of this "mess" stems from this difference in DXC allowing uninitialized const variables to be declared without an initializer.  For FXC, because it's less resilient to prior errors, any line that causes an error to be emitted is removed before compiling again and looking for more errors later in the test.  That means if you use the decl that produced an error on FXC, that use will also produce an error.

llvm-beanz (approved):
LGTM

tools/clang/test/HLSL/write-const-arrays.hlsl
llvm-beanz: I wonder if there is a better way for us to contrast then diagnostic differences between FXC and DXC than having our test cases marked up with FXC error and warning text. I don't really have an idea here, but I'm curious if we can do something.

PR #5142, opened 2023-04-12
Fix reflection on bitfield.

python3kgae:
This is for case where bitfield member lost when reflection.

Save bitfield information in DxilFieldAnnotation.
Then use the information in CShaderReflectionType::Initialize.


include/dxc/DXIL/DxilTypeSystem.h
```diff
@@ -96,6 +104,8 @@ class DxilFieldAnnotation {
   InterpolationMode m_InterpMode;
   std::string m_FieldName;
   bool m_bCBufferVarUsed; // true if this field represents a top level variable in CB structure, and it is used.
+  std::vector<DxilFieldAnnotation> m_BitFields;
+  unsigned m_BitFieldWidth; // For bit field. 0 means not bitfiled.
```
tex3d, line 108: Comment typo: `bitfield` rather than `bitfiled`

python3kgae: Fixed.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
```diff
@@ -0,0 +1,77 @@
+// RUN: %dxc -E main -T ps_6_0 -HV 2021 -Vd -validator-version 0.0 %s | %D3DReflect %s | FileCheck %s
+
+// Make sure bitfiled info is saved.
+// TODO - make better dump. Now, Rows = 0 means bitfield. Columns is bit width. Elements is bit offset.
```
tex3d, line 4: Comment out of date

tools/clang/unittests/HLSLTestLib/D3DReflectionStrings.cpp
```diff
@@ -198,7 +198,18 @@ LPCSTR ToString(D3D_TESSELLATOR_DOMAIN TessellatorDomain) {
   default: return nullptr;
   }
 }
+
+// FIXME: remove this once D3D_SVC_BIT_FIELD added into
+// D3D_SHADER_VARIABLE_CLASS.
+const D3D_SHADER_VARIABLE_CLASS D3D_SVC_BIT_FIELD =
```
tex3d, line 204: See how values are added for 16 and 64 bit types under `ADD_16_64_BIT_TYPES`.  Maybe we should follow the same pattern.  It allows for compilation against the header after you add it as well as before.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
tex3d: To clarify:
- `Rows = 0 means bitfield` shouldn't be the indicator, since it uses `D3D_SVC_BIT_FIELD` now.  However, I still see that Rows are 0 for bitfields.  Should they be?  I'm not sure it matters, but it's not necessary anymore.
- `Elements is bit offset` - this is `Offset` now.

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -102,6 +103,14 @@ enum class PublicAPI { D3D12 = 0, D3D11_47 = 1, D3D11_43 = 2 };
 #define D3D_SVT_UINT64  ((D3D_SHADER_VARIABLE_TYPE)62)
 #endif // ADD_16_64_BIT_TYPES
 
+#ifdef ADD_SVC_BIT_FIELD
+// Disable warning about value not being valid in enum
+#pragma warning(disable : 4063)
+// FIXME: remove the define once D3D_SVC_BIT_FIELD added into
+// D3D_SHADER_VARIABLE_CLASS.
+#define D3D_SVC_BIT_FIELD (D3D_SVC_INTERFACE_POINTER + 1)
```
tex3d, line 111: I believe this should have a cast inside this expression.
```cpp
#define D3D_SVC_BIT_FIELD ((D3D_SHADER_VARIABLE_CLASS)(D3D_SVC_INTERFACE_POINTER + 1))
```

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
pow2clk: Xiang, are you going to address this comment?

python3kgae: Fixed. The outdated comment is removed. Rows will be 1 now.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
```diff
@@ -0,0 +1,76 @@
...
+// CHECK-NEXT:                 }
+// CHECK-NEXT:             }
+// CHECK-NEXT: CBuffer: B
+
+struct BF {
+   int i0 : 8;
+   int i1 : 8;
+   int i2 : 8;
+   int i3 : 8;
+};
```
pow2clk, line 69: I think we need more test cases. There are places where I was trying to think how it might mess up other elements of the struct. Can we get tests where the bitfield doesn't fully fill the integer? One where a zero width ends the run? one with a non-bitfield member at the start? Maybe one with a single member with a bitfield width?

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -985,6 +1001,12 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(DxilStructAnnotation *annota
           ConstructFieldAttributedAnnotation(fieldAnnotation, fieldTy,
                                              bDefaultRowMajor);
           fieldAnnotation.SetCBufferOffset(CBufferOffset);
+          if (StartBitOffset > 0) {
+            DxilFieldAnnotation &prevFieldAnnotation =
+                annotation->GetFieldAnnotation(fieldIdx - 1);
```
pow2clk, line 1006: Can `fieldIdx` be zero? What if the struct is just one member with a bitfield width?

pow2clk (commented):
Most of my comments are nits, but I have some concerns about subtracting one from the index. I would like to better understand the idea here. The first bitfielded member will have an array containing all the others that make up the full subdivided value? is that right?

tools/clang/lib/CodeGen/CGHLSLMS.cpp
python3kgae: If fieldIdx is zero then, StartBitOffset will be zero too.

python3kgae:
> Most of my comments are nits, but I have some concerns about subtracting one from the index. I would like to better understand the idea here. The first bitfielded member will have an array containing all the others that make up the full subdivided value? is that right?

There will be a fake field created for the whole integer. Then all fields are saved as members of that fake field.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -958,15 +958,31 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(DxilStructAnnotation *annota
...
+        DxilFieldAnnotation bitfieldAnnotation;
+
+        bitfieldAnnotation.SetBitFieldWidth(Field->getBitWidthValue(Context));
+        const BuiltinType *BTy = Field->getType()->getAs<BuiltinType>();
+        CompType::Kind kind =
+            BuiltinTyToCompTy(BTy, /*bSNorm*/ false, /*bUNorm*/ false);
+        bitfieldAnnotation.SetCompType(kind);
+        bitfieldAnnotation.SetFieldName(Field->getName());
+        bitfieldAnnotation.SetCBufferOffset((unsigned)(BitOffset - StartBitOffset));
+        // TODO: more information for bitfield.
```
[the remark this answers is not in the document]
python3kgae: Do we plan to support semantic on bit field?
It will be strange to have things like uint bf : SemanticA : 8 :)

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -1169,15 +1185,47 @@ HRESULT CShaderReflectionType::Initialize(
...
+          unsigned bitOffset = 0;
+          CShaderReflectionType *bitFieldReflectionType = nullptr;
+          for (auto &bitfieldAnnotation : fieldAnnotation.GetBitFields()) {
+            bitFieldReflectionType = new CShaderReflectionType();
+            allTypes.push_back(
+                std::unique_ptr<CShaderReflectionType>(bitFieldReflectionType));
+
+            bitFieldReflectionType->Initialize(M, fieldType, fieldAnnotation,
+                                            elementOffset, allTypes, isCBuffer);
+            // FIXME: remove the cast once D3D_SVC_BIT_FIELD added into D3D_SHADER_VARIABLE_CLASS.
```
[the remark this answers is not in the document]
python3kgae: Created https://github.com/microsoft/DirectXShaderCompiler/issues/5169 to track this issue

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
python3kgae: Added.

pow2clk (approved):
Just one comment nit. Looks good!

lib/DXIL/DxilMetadataHelper.cpp
```diff
@@ -1214,6 +1214,20 @@ Metadata *DxilMDHelper::EmitDxilFieldAnnotation(const DxilFieldAnnotation &FA) {
...
+    for (const DxilFieldAnnotation &BitField : BitFields) {
+      MDBitFieldVals.emplace_back(EmitDxilFieldAnnotation(BitField));
+    }
+    auto *BitFieldsMD = MDNode::get(m_Ctx, MDBitFieldVals);
+    MDVals.emplace_back(Uint32ToConstMD(kDxilFieldAnnotationBitFieldsTag));
+    MDVals.emplace_back(BitFieldsMD);
+  }
+  if (FA.HasBitFieldWidth()) {
+    MDVals.emplace_back(Uint32ToConstMD(kDxilFieldAnnotationBitFieldWidthTag));
+    MDVals.emplace_back(Uint32ToConstMD((unsigned)FA.GetBitFieldWidth()));
```
bob80905, line 1229: why is the (unsigned) cast necessary? 
Doesn't GetBitFieldWidth() already return an unsigned type?

hekota (approved):
LGTM!

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
tex3d: Oh, I think we need a test that demonstrates how bitfields spill over into additional primitive elements, and should have additional top-level type with sub-fields.  Also, ensuring that an unnamed padding field is handled properly, and the `:0` forces the next bit field to start at the next primitive.

For instance:
```hlsl
struct BF {
   int i0 : 8;
   int i1 : 8;
   int i2 : 8;
   int i3 : 8;
   int i4 : 8; // spill over, start of second primitive
   int : 4; // what about blank field?  do we support this?
   int i5 : 5;
   int : 0; // force next bit field to start at next primitive
   int i6 : 6; // start of third primitive
};
```

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
tex3d, line 1: More descriptive names would help.  For this, I think something like `bitfield-in-structured-buffer.hlsl` might be good.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection2.hlsl
tex3d, line 1: Name could be better, maybe `bitfield-combine-to-unsigned.hlsl`?

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection3.hlsl
tex3d, line 1: What does this test add?

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection4.hlsl
tex3d, line 1: I don't understand what this test adds.  Also, the FIXME comment doesn't seem relevant to bitfield testing, is it?

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
```diff
@@ -0,0 +1,76 @@
+// RUN: %dxc -E main -T ps_6_0 -HV 2021 -Vd -validator-version 0.0 %s | %D3DReflect %s | FileCheck %s
```
tex3d, line 1: I think we might be able to switch to using `-select-validator internal` instead of `-Vd -validator-version 0.0`

tools/clang/lib/CodeGen/CGHLSLMS.cpp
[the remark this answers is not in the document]
tex3d: I do not think we support (or should support) semantics on bitfields.  If you want bitfield in signature element (if we support that?), then you would need to wrap the bitfields in a struct and put the semantic on the field or parameter using that struct type.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
python3kgae: Done.

lib/DXIL/DxilMetadataHelper.cpp
python3kgae: Good catch. Removed.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection.hlsl
python3kgae: unknown argument :(

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection4.hlsl
python3kgae: Removed.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_reflection3.hlsl
python3kgae: test bit field as single member.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_spill-over.hlsl
```diff
@@ -0,0 +1,141 @@
...
+// CHECK-NEXT:                       Class: D3D_SVC_BIT_FIELD
+// CHECK-NEXT:                       Type: D3D_SVT_INT
+// CHECK-NEXT:                       Elements: 0
+// CHECK-NEXT:                       Rows: 1
+// CHECK-NEXT:                       Columns: 8
+// CHECK-NEXT:                       Members: 0
+// CHECK-NEXT:                       Offset: 0
+// CHECK-NEXT:                   ID3D12ShaderReflectionType:
+// CHECK-NEXT:                     D3D12_SHADER_TYPE_DESC: Name: int
+// CHECK-NEXT:                       Class: D3D_SVC_BIT_FIELD
```
tex3d, line 83: I'm torn.  Should padding be reflected here?  I don't suppose it hurts, but we also have offsets for all the named fields, so we don't technically need it, right?  When we add field names, I guess the name of this field would be nullptr, or "\0"?

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_spill-over.hlsl
```diff
@@ -0,0 +1,141 @@
...
+// CHECK-NEXT:             }
+// CHECK-NEXT: CBuffer: B
+
+struct BF {
+   int i0 : 8;
+   int i1 : 8;
+   int i2 : 8;
+   int i3 : 8;
+   int i4 : 8; // spill over, start of second primitive
+   int : 4; // what about blank field?  do we support this?
```
tex3d, line 129: Great, now we should drop the questions in the comment from my example, since they are answered ;)

python3kgae: :). I'll remove it.

tools/clang/test/HLSLFileCheck/d3dreflect/bitfield_spill-over.hlsl
python3kgae: Add padding filed seems redundant like you mentioned.
For the name, I'm not sure. Both nullptr and "\0" feels OK.

PR #5146, opened 2023-04-12
[LIT] Find te in PATH if not set when run lit

python3kgae:

### DIFF
--- a/utils/lit/lit/formats/taef.py
+++ b/utils/lit/lit/formats/taef.py
@@ -38,6 +38,10 @@ def executeCommandForTaef(command, cwd=None, env=None):
 
 class TaefTest(TestFormat):
     def __init__(self, te_path, test_dll, test_path, select_filter, extra_params):
+        # Try to find te in %PATH% if not exist.
+        if not os.path.exists(te_path):
+            te_path = lit.util.which('TE')
+
         self.te = te_path
         self.test_dll = test_dll
         self.test_path = test_path


adam-yang (approved):

utils/lit/lit/formats/taef.py
```diff
@@ -38,6 +38,10 @@ def executeCommandForTaef(command, cwd=None, env=None):
 
 class TaefTest(TestFormat):
     def __init__(self, te_path, test_dll, test_path, select_filter, extra_params):
+        # Try to find te in %PATH% if not exist.
+        if not os.path.exists(te_path):
```
dmpots, line 42: Is it an error to provide a `te_path` that does not exist? Can we push the path search to the caller instead?

utils/lit/lit/formats/taef.py
```diff
@@ -38,6 +38,10 @@ def executeCommandForTaef(command, cwd=None, env=None):
...
+            te_path = lit.util.which('TE')
+
         self.te = te_path
```
dmpots, line 45: Should we raise an exception here if the te_path does not exist?

llvm-beanz:
Can you please explain more about this change and why it is needed?

The lit suite configs all set the `te` executable based on the build configuration. Inferring the value from the path seems like it is probably covering up a different error and might not be addressing the real problem.

utils/lit/lit/formats/taef.py
python3kgae: It will throw exception later when use self.te and print error like this.
 error: unable to discover taef tests in 'D:/agent/_work/2/b/DXC\\Debug\\bin\\ClangHLSLTests.dll', using TAEF_EXECUTABLE-NOTFOUND

utils/lit/lit/formats/taef.py
python3kgae: That sounds a good idea.

python3kgae:
Issue resolved by https://github.com/microsoft/DirectXShaderCompiler/pull/5150

PR #5150, opened 2023-04-15
Allow find te.exe with %TAEF_PATH% directly.

python3kgae:
This is for pipeline test where te.exe is set with %TAEF_PATH%.


hekota (approved):

cmake/modules/FindTAEF.cmake
```diff
@@ -64,12 +64,13 @@ find_program(TAEF_EXECUTABLE te.exe PATHS
   ${TAEF_SDK_BIN}/${TAEF_BIN_ARCH}
   ${WINDOWS_KIT_10_PATH}
   ${WINDOWS_KIT_81_PATH}
+  $ENV{TAEF_PATH}
```
llvm-beanz, line 67: If a user is explicitly setting an environment variable to find TAEF, shouldn't that be the first path we check not the last?

python3kgae: Fixed

llvm-beanz (approved):
LGTM

PR #5154, opened 2023-04-17
Setup `install-distribution` target for DXC

llvm-beanz:
This cherry-picks a change I made in LLVM 3.8 (SVN revision 261681, https://github.com/llvm/llvm-project/commit/c90e12fe9fd7eb9fd7b342e7ac0cec971dd3e4db), and provides updates to DXC-specific build logic to enable the change.

This only works with single-configuration generators (Ninja, Makefiles), but could probably be extended to work with Visual Studio as well.

With this change the `install-distribution` target will build and install the DXC components and headers into the `CMAKE_INSTALL_PREFIX` directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -788,6 +788,23 @@ endif()
 
 include(CoverageReport)
 
+# This must be at the end of the LLVM root CMakeLists file because it must run
+# after all targets are created.
+if(LLVM_DISTRIBUTION_COMPONENTS)
+  if(CMAKE_CONFIGURATION_TYPES)
+    message(FATAL_ERROR "LLVM_DISTRIBUTION_COMPONENTS cannot be specified with multi-configuration generators (i.e. Xcode or Visual Studio)")
+  endif()
+  
+  add_custom_target(install-distribution)
+  foreach(target ${LLVM_DISTRIBUTION_COMPONENTS})
+    if(TARGET install-${target})
+      add_dependencies(install-distribution install-${target})
+    else()
+      message(FATAL_ERROR "Specified distribution component '${target}' doesn't have an install target")
+    endif()
+  endforeach()
+endif()
+
 # Disable regeneration target for official builds as they always fun a full clean build.
 # This should eliminate the race-condition issue with "Cannot restore timestamp".
 if (HLSL_OFFICIAL_BUILD)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -788,6 +788,12 @@ endif()
 
 include(CoverageReport)
 
+# HLSL Change begin - Set default distribution value.
+if (NOT CMAKE_CONFIGURATION_TYPES)
+  set(LLVM_DISTRIBUTION_COMPONENTS "dxc;dxcompiler;dxc-headers" CACHE STRING "")
+endif ()
+# HLSL Change end - Set default distribution value.
+
 # This must be at the end of the LLVM root CMakeLists file because it must run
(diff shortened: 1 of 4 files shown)


Keenuts (approved):
small nit, otherwise LGTM

include/dxc/CMakeLists.txt
```diff
@@ -15,3 +15,29 @@ add_subdirectory(DxilContainer)
...
+set(DXC_PUBLIC_HEADERS
+    ${CMAKE_CURRENT_SOURCE_DIR}/dxcapi.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/dxcerrors.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/dxcisense.h)
+
+if (NOT WIN32)
+  set(DXC_PLATFORM_PUBLIC_HEADERS ${CMAKE_CURRENT_SOURCE_DIR}/WinAdapter.h)
+endif()
+
+# This target just needs to exist for the distribuiton build, it doesn't need to
```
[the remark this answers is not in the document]
llvm-beanz: lol.... someday I'll learn to spell. That day is obviously not today.

bogner (approved):
lgtm

hekota (approved):
LGTM!

CMakeLists.txt
```diff
@@ -789,17 +789,14 @@ endif()
...
 # HLSL Change end - Set default distribution value.
 
 # This must be at the end of the LLVM root CMakeLists file because it must run
 # after all targets are created.
 if(LLVM_DISTRIBUTION_COMPONENTS)
-  if(CMAKE_CONFIGURATION_TYPES)
-    message(FATAL_ERROR "LLVM_DISTRIBUTION_COMPONENTS cannot be specified with multi-configuration generators (i.e. Xcode or Visual Studio)")
-  endif()
+  # HLSL Change - Deleted restriction for non-IDE builds. This works with Visual
+  # Studio for DXC because of the limited number of targets.
```
bogner, line 799: Is this because of the number of targets that are in `LLVM_DISTRIBUTION_COMPONENTS` in particular? That is to say, would this work in LLVM in general in cases where people set a very small number of distribution components?

If so, consider changing the check to be related to the size of the list.

CMakeLists.txt
```diff
@@ -789,17 +789,14 @@ endif()
 include(CoverageReport)
 
 # HLSL Change begin - Set default distribution value.
-if (NOT CMAKE_CONFIGURATION_TYPES)
-  set(LLVM_DISTRIBUTION_COMPONENTS "dxc;dxcompiler;dxc-headers" CACHE STRING "")
-endif ()
+set(LLVM_DISTRIBUTION_COMPONENTS "dxc;dxcompiler;dxc-headers" CACHE STRING "")
 # HLSL Change end - Set default distribution value.
```
bogner, line 793: It's a little awkward to set the default here rather than in some dxc-specific place, but I guess it's okay for now

CMakeLists.txt
llvm-beanz: There is really nothing about the `LLVM_DISTRIBUTION_COMPONENTS` and its targets that can't be configured always. The IDE generators _really_ struggle with large numbers of utility targets, so in upstream (and in DXC) we disable generating a lot of utility targets (like the `install-llc` target) when you're using an IDE generator.

Since DXC's only real supported distribution is three targets, I enabled the per-target install rules for them all the time, and just made them always grouped as a distribution.

PR #5155, opened 2023-04-17
[NFC] Delete duplicated tests

llvm-beanz:
These tests are just copies of HLSLFileCheck/shader_targets/library/lib_no_alias.hlsl testing the exact same output, so they provide no unique value.


llvm-beanz:
Worth noting, one of the versions I delete here is never run.

pow2clk (approved):
Looks good! I fear we have a few of these. It would be nice to do a pass to get them all

python3kgae (approved):

PR #5156, opened 2023-04-17
Fix unused local dwOffsetInBytes in DxilDiaSymbolManager.cpp

tex3d:
Additionally, cast uint64_t to unsigned for comparison and for expected var arg based on format string.


jeffnn (approved):

jeffnn:
Approved!

llvm-beanz (approved):

PR #5157, opened 2023-04-17
[NFC] Update tests that depend on 2018 behavior

llvm-beanz:
Under HLSL 2021 short circuiting ternary operators result in _very_ different control flow in these tests, so they aren't really portable. Changing these tests to being HV 2018 keeps them working with minimal changes.


adam-yang (approved):
LGTM!

tex3d:
Alternatively, the ternary operator uses could simply be replaced with `select()` intrinsic uses.

tex3d (approved):
I think this is fine, but would hope that replacing the ternary operators with select() for -HV 2021 would preserve the expected control flow as well.  If not, I'd want to know why.

llvm-beanz:
> I think this is fine, but would hope that replacing the ternary operators with select() for -HV 2021 would preserve the expected control flow as well. If not, I'd want to know why.

The tests run fine with the `select` intrinsic too. We don't have a lot of tests that run explicitly under HV 2018, so I think it is good to keep some running that way.

PR #5159, opened 2023-04-19
test: add static assert for the git count define

Keenuts:
The `target_version_vertex.hlsl` test looks at different defines values, and generate the good or the bad instruction depending on those defines.

This means an HLSL test will fail, because you are building DXC from a shallow clone.
This commit adds a static assert if you build tests on a shallow clone to let you fail earlier, with a more user-friendly explanation.


llvm-beanz (commented):
Sorry for taking so long to reply to this. I’ve been trying to think this through. I don’t think this is the right approach to the problem you’re trying to solve.

These variables are all specified by CMake. If instead we make these CMake configuration-time errors we can provide the diagnostic earlier and with direct messaging.

llvm-beanz (commented):
Sorry for taking so long to reply to this. I’ve been trying to think this through. I don’t think this is the right approach to the problem you’re trying to solve.

These variables are all specified by CMake. If instead we make these CMake configuration-time errors we can provide the diagnostic earlier and with direct messaging.

PR #5162, opened 2023-04-19
Fix build error for deprecated function in UCRT.

python3kgae:
This is for enable-lit by default.
gtest got link error for missing dup/dup2/close/creat.

### DIFF
--- a/utils/unittest/googletest/src/gtest-all.cc
+++ b/utils/unittest/googletest/src/gtest-all.cc
@@ -38,6 +38,14 @@
 // when it's fused.
 #include "gtest/gtest.h"
 
+// Alias deprecated functions for UCRT.
+#ifdef _UCRT
+#define close _close
+#define creat _creat
+#define dup2  _dup2
+#define dup   _dup
+#endif
+
 // The following lines pull in the real gtest *.cc files.
 #include "src/gtest.cc"
 #include "src/gtest-death-test.cc"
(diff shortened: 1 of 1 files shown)


hekota (approved):

hekota (commented):
Could you please add a comment explaining why this is done, that it is due to obsolete function that have been removed in the newer UCRT libraries?

python3kgae:
> 

Updated.

hekota (approved):

bob80905:
Shouldn't we avoid using the deprecated functions entirely, instead of making aliases for them? 
For example, this page:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/creat-wcreat?view=msvc-170
recommends using _sopen_s instead of _creat.
Or is creat a part of the gtest framework that we can't get around?

python3kgae:
> Shouldn't we avoid using the deprecated functions entirely, instead of making aliases for them? For example, this page: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/creat-wcreat?view=msvc-170 recommends using _sopen_s instead of _creat. Or is creat a part of the gtest framework that we can't get around?

I don't know if it is OK to modify google test code. :(

utils/unittest/googletest/src/gtest-all.cc
```diff
@@ -38,6 +38,15 @@
 // when it's fused.
 #include "gtest/gtest.h"
 
+// Alias deprecated functions for UCRT due to these obsolete function have been
```
llvm-beanz, line 41: Please wrap in an `HLSL Change` block.

python3kgae: Fixed.

pow2clk (approved):
Looks good!

PR #5166, opened 2023-04-20
Allow `ref` parameters for intrinsics

### LINKED ISSUE #5164: Add special reference passing for built-in intrinsics
# What

Several of the intrinsics in gen_intrin_main (specifically the atomic intrinsics) need reference parameters, but we currently don't support them. To address this we currently have hacks like:

* Implicitly making atomic result operands references - https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/lib/Sema/SemaHLSL.cpp#L1980
* Custom handling in overload matching for atomics - https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/lib/Sema/SemaHLSL.cpp#L6201

# Why

This change is needed to make our ASTs accurate and consistent. Differentiating between reference passed and out/inout is a key step to allow some reference passing behaviors where we need them.

# Acceptance Criteria

This change should be NFC, however in addition to the changes described we should enhance our testing around built-in AST formulations.

llvm-beanz:
We effectively have reference parameter passing for intrinsics, but we accomplish it by extremely careful interpretation of `out` and `inout` in specific contexts.

This change separates the two in our intrinsic definitions. Parameters marked `ref` become lvalue reference types in the intrinsic function declarations but do not get annotated with `AR_QUAL_OUT`. This allows us to differentiate them as true reference parameters.

For the atomic result intrinsics this allows us to clear up some hacks because we needed them to not be `out` parameters but we needed them to be passed by reference.

I've left a comment in SemaHLSL where we match argument types about a further cleanup we should do to correct reference binding errors, but that is a larger task that we should do separately.

Fixes #5164


python3kgae:
Will ref be exposed to user functions?

llvm-beanz:
> Will ref be exposed to user functions?

No, this is just for intrinsics. We effectively already hack around needing reference parameter semantics for some of the HLSL intrinsics.

Adding proper reference types to HLSL is in [this proposal](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0006-reference-types.md).

python3kgae (approved):

tools/clang/test/HLSL/atomic-float-errors.hlsl
```diff
@@ -0,0 +1,55 @@
...
+  int iv = b / c;
+  int iv2 = b * c;
+  int oiv = 0;
+  uint64_t bb = b;
+  uint64_t cc = c;
+  uint64_t lv = bb * cc;
+  uint64_t lv2 = bb / cc;
+  uint64_t olv = 0;
+
+  InterlockedCompareStoreFloatBitwise( resBI[a], iv, iv2 ); // expected-error{{no matching function for call to 'InterlockedCompareStoreFloatBitwise'}} expected-note{{candidate function not viable: no known conversion from 'int' to 'float &' for 1st argument}}
```
bfavela, line 35: Sorry for the basic question, but does this test atomics with and without a return value? Or is the 'ref' aspect only valid for atomics with return?

bfavela (approved):

tools/clang/test/HLSL/atomic-float-errors.hlsl
llvm-beanz: That’s a good question. This test case is non-exhaustive, but it just verifies that overload resolution is properly generating a mismatch on the first parameter. There are other test cases (`atomic_overload.hlsl`, `atomic_restypes.hlsl`, and `atomicop_float.hlsl`) which verify the same error-generation for other atomic functions.

This is showing up in the change as a new file, but it is just a rewritten test from being FileCheck-based to being Clang verifier based. Making it verifier based means the source location of the diagnostics also get verified. In general we shouldn’t be using FileCheck for verifying diagnostics because of the lack of verifying source locations.

For a little more context: This change is more or less NFC to the user. The code I deleted from SemaHLSL.cpp matched the intrinsic opcode then treated the first parameter of those intrinsics as a reference. This change effectively moves that logic out of special casing for overload resolution and function generation, and instead captures it where we specify the function signature so we just generate the signature correctly up front and everything else mostly just works as expected.

llvm-beanz: Yea. I didn't mark the change NFC because it does produce some slightly different diagnostics (hence the updated tests), and it produces some verifiable AST differences.

Thank you for the through review 😀

PR #5167, opened 2023-04-21
[enable lit] Keep same name for ClangHLSLTests when enable lit

python3kgae:
It will make sure with/without lit enabled, output dll for ClangHLSLTests  has same name.


llvm-beanz:
I made the change to the dll name explicitly and intentionally to match LLVM conventions. Is there a reason this is important?

hekota (approved):
LGTM!

bob80905 (approved):
LGTM

PR #5170, opened 2023-04-21
[enable-lit] use !TEST_CLANG! to get updated result.

python3kgae:
It fix issue when running single test instead of all test under lit. %TEST_CLANG% will not get update until out of the if.


hekota (approved):
LGTM!

bob80905 (approved):
LGTM

PR #3445, opened 2021-02-10
Fix memory corruption in DxilCounters.cpp

jeffnn:
Noticed a crash because of this while running PIX tests and then narrowed it down with appverif's full page-heap.
The previous "PI" variable was pointing into nowhere because the recursion could cause the map's contents to be reallocated.
"PI" could have been used safely in a few of these places, but rather than spot-fix it, I removed the use of this variable in order to preclude this class of bug from the other locations which I may or may not be exercising in my tests.

PR #5185, opened 2023-05-05
[NFC] fix typo in GetPointerInfo.

python3kgae:
Use CE instead AC which must be nullptr.

### DIFF
--- a/lib/DXIL/DxilCounters.cpp
+++ b/lib/DXIL/DxilCounters.cpp
@@ -70,7 +70,7 @@ PointerInfo GetPointerInfo(Value* V, PointerInfoMap &ptrInfoMap) {
     ptrInfoMap[V] = GetPointerInfo(AC->getOperand(0), ptrInfoMap);
   } else if (ConstantExpr *CE = dyn_cast<ConstantExpr>(V)) {
     if (CE->getOpcode() == LLVMAddrSpaceCast)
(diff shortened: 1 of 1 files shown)


llvm-beanz:
This doesn't seem NFC, it seems like a legitimate bug fix. Can we get a test case?

llvm-beanz:
I also just had to look... This code isn't even valid C++, but it compiles because we build with the MS-extensions enabled in GCC and Clang (and MSVC does wrong things with variable scopes). 😞

python3kgae:
> This doesn't seem NFC, it seems like a legitimate bug fix. Can we get a test case?



> This doesn't seem NFC, it seems like a legitimate bug fix. Can we get a test case?

Catches the issue by check the code, no test right now.
I'll try to create one.

jeffnn:
Xiang, did you mean to add me or someone else? I'm happy to review, but it's nowhere near where I usually hang out, so to speak.

python3kgae:
> Xiang, did you mean to add me or someone else? I'm happy to review, but it's nowhere near where I usually hang out, so to speak.

Never mind.
You're suggested by github :).

llvm-beanz:
@jeffnn, you were on the blame for this because of #3445, but the change you made there was pretty mechanical.

pow2clk (approved):
Please reword the commit message and title before submitting since this is no longer related to a typo at all

tex3d (approved):
LGTM.

PR #5187, opened 2023-05-06
CleanupPerThreadFileSystem after llvm shutdown to avoid assert.

python3kgae:
When remove timer in llvm shutdown, CreateInfoOutputFile will be called. It will hit assert for TLS not intialized if CleanupPerThreadFileSystem is already executed.

### DIFF
--- a/tools/clang/tools/dxcompiler/DXCompiler.cpp
+++ b/tools/clang/tools/dxcompiler/DXCompiler.cpp
@@ -106,8 +106,8 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD Reason, LPVOID reserved) {
     DxcEtw_DXCompilerShutdown_Start();
     DxcSetThreadMallocToDefault();
     ::hlsl::options::cleanupHlslOptTable();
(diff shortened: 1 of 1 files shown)


tools/clang/tools/dxcompiler/DXCompiler.cpp
```diff
@@ -106,8 +106,8 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD Reason, LPVOID reserved) {
     DxcEtw_DXCompilerShutdown_Start();
     DxcSetThreadMallocToDefault();
     ::hlsl::options::cleanupHlslOptTable();
-    ::llvm::sys::fs::CleanupPerThreadFileSystem();
     ::llvm::llvm_shutdown();
+    ::llvm::sys::fs::CleanupPerThreadFileSystem();
```
tex3d, line 110: This may fix the issue, but I'm still concerned that we are attempting to create a file in a destructor during `llvm_shutdown()` to write timer results.  Especially from the perspective of the dxcompiler.dll.

damyanp:

damyanp:

PR #5192, opened 2023-05-08
Set the layout rule in processTemplatedLoadFromBuffer

### LINKED ISSUE #5123: [SPIR-V] Raytracing anyhit shader generating invalid spir-v (expected operands to be scalar or vector int: INotEqual)
I am getting the following validation error from spirv-val with the attached raytracing shader:

```
fatal error: generated SPIR-V is invalid: Expected operands to be scalar or vector int: INotEqual
  %118 = OpINotEqual %bool %113 %uint_0
```

I compile the file with the following command line:

```
dxc.exe -spirv -T lib_6_5 -E AnyHit1 -fspv-target-env=vulkan1.3 -Wno-effects-syntax  test.hlsl
```

Interestingly, in the process of stripping this shader down into something I could share, I found a workaround to the bug.  There is an element in this struct that is bool:

```
struct __RTX_MATERIAL_DATA__ 
{
	SamplerStateIndex_t g_sAniso  ;
	Texture2DIndex_t g_tColor  ;
	bool g_bNoEnvMapOcclusion  ;	// <--- REMOVE THIS TO MAKE THE BUG GO AWAY!!!!
	float g_flAlphaTestReference  ;
};
```

If you remove that member variable (or change its type from bool -> int), the spirv-val error goes away.  Interestingly, that member variable is never referenced in the shader.

Shader code attached.

[test.zip](https://github.com/microsoft/DirectXShaderCompiler/files/11115895/test.zip)

s-perron:
They layout rules for value coming from processTemplatedLoadFromBuffer
are inconsistent, and this is causing problems. This PR changes the
function to return an instruction whose layout rule is Void.

This should remove the inconsistenty.

Fixes #5123

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.load.layout.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.load.layout.hlsl
@@ -1,0 +1,23 @@
+// RUN: %dxc -spirv -T lib_6_5 -E AnyHit1 -fspv-target-env=vulkan1.3 -Wno-effects-syntax -fcgl -Vd %s | FileCheck %s
+
+// CHECK-NOT: error
+struct DataType
+{
+	uint u;
+	bool b;
+};
+
+static DataType data;
+
+ByteAddressBuffer buffer : register ( t4 , space3 )  ; 
+
+struct PayloadShadow_t 
+{ 
+    float m_flVisibility ; 
+} ; 
+
+[ shader ( "anyhit" ) ] 
+void AnyHit1 ( inout PayloadShadow_t payload , in BuiltInTriangleIntersectionAttributes attrs ) 
+{
+	data = buffer.Load<DataType>( 0 );
+} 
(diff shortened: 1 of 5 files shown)


tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.load.layout.hlsl
```diff
@@ -0,0 +1,23 @@
+// RUN: %dxc -spirv -T lib_6_5 -E AnyHit1 -fspv-target-env=vulkan1.3 -Wno-effects-syntax -fcgl -Vd %s | FileCheck %s
```
Keenuts, line 1: `-Vd`, but wasn't the bug a validation error? Is the test failing without the fix?

Keenuts (changes requested):
Thanks for the fix, seems logical! Just something on the new test file

tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.load.layout.hlsl
s-perron: Good catch. I cut & paste the command I was using, and I turned off validatoin so I could see the code that was generated.

Keenuts (approved):
Thanks!

Keenuts:
Added @dnovillo for the second approval

dnovillo (approved):

PR #5197, opened 2023-05-09
Fix loss of buffer type info with libraries and linker

python3kgae:
## Fix loss of buffer type info.

When undef, there was no global symbol, thus type-preserving cast was lost, and undef type for resource would be dx.types.Handle instead.  This would particularly impact path through higher lib target then linked to a lower final shader target.

The fix is to use the HLSL type for undef when the GlobalSymbol is undef.

## Copy resource type annotation in linker.

Linker was dropping the resource type annotation for the target module. Fixed by copying the type annotation to the target after AddResource.

## Check resource type annotation with libs and linking and various targets

These tests check both module metadata and reflection dump. While IR may be unnecessary given reflection check, it does reveal some format inconsistencies that should be addressed.

### DIFF
--- a/lib/DXIL/DxilMetadataHelper.cpp
+++ b/lib/DXIL/DxilMetadataHelper.cpp
@@ -617,9 +617,11 @@ void DxilMDHelper::EmitDxilResourceBase(const DxilResourceBase &R, Metadata *ppM
   Constant *GlobalSymbol = R.GetGlobalSymbol();
   // For sm66+, global symbol will be mutated into handle type.
   // Save hlsl type by generate bitcast on global symbol.
-  if (m_pSM->IsSM66Plus()) {
-    Type *HLSLTy = R.GetHLSLType();
-    if (HLSLTy && HLSLTy != GlobalSymbol->getType())
+  Type *HLSLTy = R.GetHLSLType();
+  if (HLSLTy && HLSLTy != GlobalSymbol->getType()) {
+    if (isa<UndefValue>(GlobalSymbol))
+      GlobalSymbol = UndefValue::get(HLSLTy);
+    else if (m_pSM->IsSM66Plus())
       GlobalSymbol = cast<Constant>(
           ConstantExpr::getCast(Instruction::BitCast, GlobalSymbol, HLSLTy));
   }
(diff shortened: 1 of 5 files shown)


tex3d (approved):
LGTM.

lib/DXIL/DxilMetadataHelper.cpp
```diff
@@ -617,9 +617,11 @@ void DxilMDHelper::EmitDxilResourceBase(const DxilResourceBase &R, Metadata *ppM
   Constant *GlobalSymbol = R.GetGlobalSymbol();
   // For sm66+, global symbol will be mutated into handle type.
   // Save hlsl type by generate bitcast on global symbol.
```
pow2clk, line 619: adding a line to explain the creation of an HLSL undef for the global symbol would be helpful since it is mingled with what this comment describes.

tools/clang/test/HLSLFileCheck/hlsl/linker/resources/preserve_cb_types.hlsl
```diff
@@ -0,0 +1,164 @@
+// RUN: %dxilver 1.7 | %dxc -T lib_6_5 -Fo dxl_6_5 %s | FileCheck %s -check-prefixes=CHECK,LIBGV_ORIG
+
+// Why is GV not mutated to handle for lib_6_6+?
```
pow2clk, line 3: I feel like we should have an answer to that. Is it really the case?

tools/clang/test/HLSLFileCheck/hlsl/linker/resources/preserve_sb_types.hlsl
```diff
@@ -0,0 +1,159 @@
+// RUN: %dxilver 1.7 | %dxc -T lib_6_5 -Fo dxl_6_5 %s | FileCheck %s -check-prefixes=CHECK,LIBGV_ORIG
+
+// Why is GV not mutated to handle for lib_6_6+?
```
pow2clk, line 3: Here as above. If we really don't know, that seems like a problem

pow2clk (approved):
Comment suggestions only

tools/clang/test/HLSLFileCheck/hlsl/linker/resources/preserve_sb_types.hlsl
python3kgae: Created issue https://github.com/microsoft/DirectXShaderCompiler/issues/5202 to track it.

tools/clang/test/HLSLFileCheck/hlsl/linker/resources/preserve_cb_types.hlsl
python3kgae: Created issue https://github.com/microsoft/DirectXShaderCompiler/issues/5202 to track it.

PR #5200, opened 2023-05-10
Support case when memcpy src/dst are only use/def of the src/dst.

python3kgae:
When src of memcpy is the only read from src and dst of memcpy is the only write to dst, as long all writes to src are before the memcpy, it is safe to replace dst with src.

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -17,6 +17,7 @@
 #include "llvm/ADT/SetVector.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/PostOrderIterator.h"
 #include "llvm/ADT/Statistic.h"
 #include "llvm/Analysis/AssumptionCache.h"
 #include "llvm/Analysis/Loads.h"
@@ -1794,6 +1795,8 @@ bool SROAGlobalAndAllocas(HLModule &HLM, bool bHasDbgInfo) {
       DominatorTree &DT = domTreeMap[F];
       if (SROA_Helper::LowerMemcpy(AI, /*annotation*/ nullptr, typeSys, DL, &DT,
                                    bAllowReplace)) {
+        if (AI->use_empty())
+          AI->eraseFromParent();
         Changed = true;
         continue;
       }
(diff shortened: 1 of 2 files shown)


tex3d:
The description is confusing me.

> src of memcpy _is only use of the src_
Don't you mean "is the only read from src?

> dst of memcpy _is only def of the dst_
Don't you mean "is the only write to dst"

> as long as _all def of the src is before the memcpy_
Don't you mean "all writes to src are before the memcpy"?

> it is safe to _replace src with dst_
Don't you mean "replace dst with src"?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3769,6 +3772,110 @@ static bool isReadOnlyResSubscriptOrLoad(CallInst *PtrCI) {
...
+  llvm_unreachable("cannot find A and B in the block");
+  return false;
+}
+
+// Make sure all store on V are before I.
+static bool noUnusedStoreAfter(Value *V, Instruction *I) {
+  SmallVector<const Instruction *, 4> Stores;
+  collectAllStores(V, Stores);
+
+  ReversePostOrderTraversal<Function *> RPOT(I->getParent()->getParent());
```
tex3d, line 3866: Is this redundant/expensive when we should have a DominatorTree already?

python3kgae:
> The description is confusing me.
> 
> > src of memcpy _is only use of the src_
> > Don't you mean "is the only read from src?
> 
> > dst of memcpy _is only def of the dst_
> > Don't you mean "is the only write to dst"
> 
> > as long as _all def of the src is before the memcpy_
> > Don't you mean "all writes to src are before the memcpy"?
> 
> > it is safe to _replace src with dst_
> > Don't you mean "replace dst with src"?

Updated.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
tex3d: Can we at least cache this for the function like the dom tree when it's necessary?

python3kgae: I'm not sure :(
This is a rare case,

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/single_memcpy_use_with_multiple_defines.ll
```diff
@@ -0,0 +1,200 @@
+; RUN: %opt %s -scalarrepl-param-hlsl -S | FileCheck %s
```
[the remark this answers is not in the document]
python3kgae: Reduced.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
tex3d: I didn't mean to pre-build this, but to cache it for the functions that required it, so it won't be redundantly computed if called again on the same function.

python3kgae: I see. I'll add cache then.

We pre-build the dom tree for all functions. :)

python3kgae: Done.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3769,6 +3772,110 @@ static bool isReadOnlyResSubscriptOrLoad(CallInst *PtrCI) {
...
+  DenseMap<const BasicBlock *, unsigned> RPOTOrder;
+  unsigned Count = 0;
+  for (BasicBlock *BB : RPOT) {
```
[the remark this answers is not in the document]
bob80905: Talked to Xiang offline, there's already validation that a def and a use that don't dominate each other won't be allowed, so my concern is resolved.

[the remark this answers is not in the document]
python3kgae: You're right.
For case like
if () {
   def a
   memcpy b, a
   use b
} else {
   another def a
}
the code will fail.

We can do more transform to remove the unused def in else in the future this is an important user case we want to optimize.
For now, it should be OK to just miss the optimization in this case.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3769,6 +3772,110 @@ static bool isReadOnlyResSubscriptOrLoad(CallInst *PtrCI) {
...
+  }
+}
+
+// return true if A is before B.
+// When A and B in same basic block, before means A will be reach first when
+// iterate the basic block. When A and B is not in same basic block, before
+// means block A will be reach first when doing reverse post order traversal on
+// basic blocks.
+static bool isBefore(const Instruction *A, const Instruction *B,
+                     DenseMap<const BasicBlock *, unsigned> &RPOTOrder) {
```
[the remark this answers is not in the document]
python3kgae: Maybe RPOTIndex?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -79,34 +80,42 @@ class SROA_Helper {
...
-                                  bool hasPrecise, DxilTypeSystem &typeSys,
-                                  const DataLayout &DL,
-                                  SmallVector<Value *, 32> &DeadInsts,
-                                  DominatorTree *DT);
+  static bool DoScalarReplacement(
+      Value *V, std::vector<Value *> &Elts, Type *&BrokenUpTy,
+      uint64_t &NumInstances, IRBuilder<> &Builder, bool bFlatVector,
+      bool hasPrecise, DxilTypeSystem &typeSys, const DataLayout &DL,
+      SmallVector<Value *, 32> &DeadInsts, DominatorTree *DT,
+      DenseMap<const Function *, DenseMap<const BasicBlock *, unsigned>>
```
tex3d, line 88: We could really use a typedef for `DenseMap<const Function *, DenseMap<const BasicBlock *, unsigned>>`.  Perhaps `RPOTFunctionMap`?  And `RPOTIndexMap` for `DenseMap<const BasicBlock *, unsigned>`?  Open to better naming suggestions of course.

tex3d: Of course, this could be wrapped in a simple class with a getter that retrieves or creates the RPOTIndexMap for a function.  This would nicely encapsulate the block to index map construction.

python3kgae: Updated.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -75,6 +76,28 @@ STATISTIC(NumReplaced, "Number of allocas broken up");
...
+      for (BasicBlock *BB : RPOT) {
+        RPOTOrder[BB] = Count++;
+      }
+      RPOTCache[F] = RPOTOrder;
+    }
+    return RPOTCache[F];
+  }
+
+private:
+  using RPOTFunctionMap = DenseMap<const Function *, RPOTIndexMap>;
```
tex3d, line 97: Now that this type is totally private and no uses of it are required elsewhere, naming this isn't really necessary.  I'm not asking for a change, just pointing out that naming it is really not necessary for the original reason I asked anymore - the class serves that purpose now.

tex3d (approved):
I suppose my only objection is the introduction of yet more complex special-case optimizations to this already overly-complex and special-cased pass, rather than moving towards more standardized representations and transformations.  However, this makes an incremental improvement in the short term, so I suppose it's ok for now.

python3kgae:
> I suppose my only objection is the introduction of yet more complex special-case optimizations to this already overly-complex and special-cased pass, rather than moving towards more standardized representations and transformations. However, this makes an incremental improvement in the short term, so I suppose it's ok for now.

Yeah. memcpy replacement is too complicated now.
Hope Chris's inout change could simplify it.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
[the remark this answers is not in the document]
python3kgae: Changed to IndexMap with type as RPOTIndexMap.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3769,9 +3800,110 @@ static bool isReadOnlyResSubscriptOrLoad(CallInst *PtrCI) {
...
+          break;
+        default:
+          // Rest are resource ptr like buf[i].
+          // Only read of resource handle.
+          break;
+        }
+      } break;
+      default: {
+        // If not sure its out param or not. Take as out param.
+        Stores.emplace_back(CI);
```
pow2clk, line 3860: What would get us here? Is the fact that it is users of an alloca what lets us assume that the only possibilities are those already handled or is this a large number of potential operations we're just assuming might be stores?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3769,9 +3800,110 @@ static bool isReadOnlyResSubscriptOrLoad(CallInst *PtrCI) {
...
+      }
+    }
+  }
+}
+
+// return true if A is before B.
+// When A and B in same basic block, before means A will be reach first when
+// iterate the basic block. When A and B is not in same basic block, before
+// means block A will be reach first when doing reverse post order traversal on
+// basic blocks.
```
pow2clk, line 3871: Is a post order traversal really sufficient? Wouldn't fellow child blocks of the entry function (for example) that get ordered arbitrarily in a post order traversal seem to come one after the other and so stores within them would seem to be sequential when they are not? I fear I'm missing why this is needed over the dominator checks

pow2clk (commented):
I'm afraid I need a little more instruction to understand the approach better

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
python3kgae: Here we treat all function call argument use as a store.
Might be a user define function here which marked as noinline.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
python3kgae: Post dominator check should be precise for this case. But it is expensive comparing to reverse post order traversal.

We know src of memcpy is the only read from src and dst of memcpy is the only write to dst, and all writes on src are "before" memcpy.

In the case where the writes on src are not sequential, as long as the memcpy dominates all the reads on dst, it should be safe.
In case there're reads on dst not dominated by memcpy, there's read not dominated by write already, the result will be undef anyway.

So it should be safe to use reverse post order traversal.

python3kgae: Switched dominator tree since we already prebuild dominator tree for all functions.
It will not cover case where there all unused write of the src, but is good enough for common cases.

pow2clk (approved):
Looks good! I saw a lot of warnings about how expensive the post order traversal calculation was as well. It's nice we already have the tree to use

PR #5201, opened 2023-05-10
[enable-lit] Allow file-check when enable-lit

python3kgae:
Also run exec test separately.

This is help to keep existing workflow after switch to lit.


hekota (approved):
LGTM!

tex3d (approved):

PR #5208, opened 2023-05-12
PIX: Don't run PIX prep passes during variable evaluation (PIX already did that)

jeffnn:
These passes have already been run by PIX at this point, and running them twice is both a waste and a source of issues: the annotate-with-virtual-registers pass might thereby change virtual register ordinals, which will cause PIX's view of HLSL variables to degrade as the user steps further into a long function.


python3kgae (approved):

adam-yang (approved):
LGTM!

PR #5209, opened 2023-05-12
[GlobalOpt] Sort members of llvm.used deterministically

bfavela:
Patch by Jake VanAdrighem!

Summary:
Fix the way we sort the llvm.used and llvm.compiler.used members.

This bug seems to have been introduced in rL183756 through a set of improper casts to GlobalValue*. In subsequent patches this problem was missed and transformed into a getName call on a ConstantExpr.

Reviewers: silvas

Subscribers: silvas, llvm-commits

Differential Revision: http://reviews.llvm.org/D12851

llvm-svn: 248728

### DIFF
--- a/test/Transforms/GlobalOpt/alias-used.ll
+++ b/test/Transforms/GlobalOpt/alias-used.ll
@@ -7,7 +7,7 @@
 @ia = internal alias i8* @i
 
 @llvm.used = appending global [3 x i8*] [i8* bitcast (void ()* @fa to i8*), i8* bitcast (void ()* @f to i8*), i8* @ca], section "llvm.metadata"
-; CHECK-DAG: @llvm.used = appending global [3 x i8*] [i8* bitcast (void ()* @fa to i8*), i8* bitcast (void ()* @f to i8*), i8* @ca], section "llvm.metadata"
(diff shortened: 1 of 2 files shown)


bfavela:
This is a cherry-pick from a pretty old change to fix non-determinism with global variable handling that was causing codegen differences.

python3kgae (approved):

test/Transforms/GlobalOpt/alias-used.ll
```diff
@@ -7,7 +7,7 @@
 @ia = internal alias i8* @i
 
 @llvm.used = appending global [3 x i8*] [i8* bitcast (void ()* @fa to i8*), i8* bitcast (void ()* @f to i8*), i8* @ca], section "llvm.metadata"
-; CHECK-DAG: @llvm.used = appending global [3 x i8*] [i8* bitcast (void ()* @fa to i8*), i8* bitcast (void ()* @f to i8*), i8* @ca], section "llvm.metadata"
```
dmpots, line 10: Do we actually run the tests in this directory? If not we should copy the test somewhere that it will get run.

dmpots (approved):

test/Transforms/GlobalOpt/alias-used.ll
bfavela: I'll create a new PR with a copied test if necessary (based on internal chat).

PR #5210, opened 2023-05-12
[lit] Move Validation taef FileCheck to lit FileCheck.

python3kgae:
Add REQUIRES: dxilver_1_6 to replace %dxilver 1.6
Add not and 2>&1 for error message.

Require https://github.com/microsoft/DirectXShaderCompiler/pull/5133


adam-yang (dismissed):

llvm-beanz (changes requested):
We cannot move tests into running only under LIT until we have _removed_ the non-LIT workflows from our build and test tooling.

tools/clang/test/lit.cfg
```diff
@@ -368,6 +368,18 @@ for pattern in [r"\bFileCheck\b",
...
+dxil_ver_cmd = [lit.util.which('dxc', llvm_tools_dir),'--version']
+dxil_ver = subprocess.check_output(dxil_ver_cmd)
+dxil_ver_text = dxil_ver.decode('utf-8')
+dxil_ver_match = dxil_ver_pat.match(dxil_ver_text)
+if dxil_ver_match == None:
+    lit_config.fatal('Fail to find dxil version!')
+dxil_ver_major, dxil_ver_minor = dxil_ver_match.group('major','minor')
+for major in range(1, int(dxil_ver_major) + 1):
+    for minor in range(0, int(dxil_ver_minor) + 1):
+        config.available_features.add('dxilver_{}_{}'.format(major,minor))
```
llvm-beanz, line 381: These are just string matched, so why not `dxil{}.{}` so that the requirements are things like dxil1.7? That seems much more concise to read.

llvm-beanz:
I think we also need to come up with a strategy for swapping in and out validators, so this probably needs more design.

(cc @tex3d)

tools/clang/test/lit.cfg
python3kgae: I'll switch to dxil{}.{}.

Another issue I hit is that there're tests which has more than one %dxilver and check different validator version.
Like
// RUN: %dxilver 1.5 | ...
// RUN: %dxilver 1.6 | ... 
Is there way to support that using REQUIRES? Or it has to be split into multiple tests?

llvm-beanz: This is exactly why I cc'd @tex3d. We've had discussions about this before. We need to figure out a strategy for expressing those tests.

One option is we could extend LIT to support conditional run lines, which might be useful upstream too.

python3kgae: For conditional run lines, how about something like
  COND_RUN(require_list): cmd
The cmd will be executed only when all the features in require_list are supported.

python3kgae: Created https://github.com/microsoft/DirectXShaderCompiler/pull/5224

PR #5211, opened 2023-05-13
[SPIR-V] Add fspv-preserve-bindings option to preserve decoration bin…

rajivmg:
…dings even when unused

This change adds a new spirv backend specific option `fspv-preserve-bindings`. By default it will be disabled.

spirv-opt has a flag to stop dead-code elimination pass from removing the decoration bindings when those bindings are unused. There are cases where the user may need this information to process the generated spirv. 

In my case I need this information while converting the spirv to msl using spirv-cross.

### DIFF
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -392,6 +392,8 @@ def fspv_print_all: Flag<["-"], "fspv-print-all">, Group<spirv_Group>, Flags<[Co
   HelpText<"Print the SPIR-V module before each pass and after the last one. Useful for debugging SPIR-V legalization and optimization passes.">;
 def Oconfig : CommaJoined<["-"], "Oconfig=">, Group<spirv_Group>, Flags<[CoreOption]>,
   HelpText<"Specify a comma-separated list of SPIRV-Tools passes to customize optimization configuration (see http://khr.io/hlsl2spirv#optimization)">;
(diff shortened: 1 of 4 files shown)


rajivmg:

f32by:
Hi, I tried your patch and it does not work on my machine. You may also add `options.set_preserve_bindings(spirvOptions.preserveBindings);` to `SpirvEmitter::spirvToolsLegalize()` too.

rajivmg:
Hi @f32by 
The patch works fine for me. I don't understand why it wouldn't work for you. I'm using this command `dxc.exe -spirv -T cs_6_0 -E computeWithRWBuffer -fspv-preserve-bindings -Fo F:\SPIRV-Cross_Build\Debug\sample.spirv F:\SPIRV-Cross_Build\Debug\sample.hlsl`
If it still doesn't work for you, can you please share the command line that you're using.

rajivmg:
Any update on this?

rajivmg:
@f32by : I guess it wasn't working for you because the new option was not added in the legalization pass. Please try it again now, it should work. :)

rajivmg:
@llvm-beanz @tex3d @mtodd @sverrejoh @s-perron // Kindly review this change.. :)

llvm-beanz (approved):
Looks reasonable to me. Any thoughts @Keenuts, @s-perron, or @dnovillo?

s-perron (approved):
LGTM. I thought we already had an option for this.

PR #5214, opened 2023-05-15
Add Option to Force loading external dxil lib for validation test

### LINKED ISSUE #5394: [Feature Request] Add test for DXIL shared library Loadability

Past releases have included unloadble dxil libraries. These slipped through all testing due to lack of any test that would have caused that unloadability to result in a failure.

We need to be able to verify that this is possible to load the dxil library for all platforms. For this, we need a variant of the validation test that forces a loading of the external library and fails if it can't be done.

pow2clk:
Adds a parameter that will force loading of the external dxil validation library using the usual search name and locations for the given platform. When this parameter is provided, the dxil library is loaded independently of the compiler library, which would try to load it if there. Unlike the usual behavior, which will fallback to the builtin validator if the external library isn't found, when the parameter is provided, If the external library isn't found, the validation test will fail with an informative message.

Fixes #5394

### DIFF
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -187,6 +187,8 @@ if "%1"=="-clean" (
 ) else if "%1"=="-dxil-loc" (
   set DXIL_DLL_LOC=%~2
   shift /1
+) else if "%1"=="-load-external-dxil" (
+  set ADDITIONAL_OPTS=%ADDITIONAL_OPTS% /p:"LoadExternalDxil=1"
 ) else if "%1"=="--" (
   shift /1
   goto :done_opt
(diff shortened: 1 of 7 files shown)


pow2clk:
Good news, the PR description test works! I think I need to do a few things to get this working on non-win, then I'll undraft it

utils/hct/hcttest.cmd
```diff
@@ -187,6 +187,8 @@ if "%1"=="-clean" (
 ) else if "%1"=="-dxil-loc" (
   set DXIL_DLL_LOC=%~2
   shift /1
+) else if "%1"=="-load-external-dxil" (
+  set ADDITIONAL_OPTS=%ADDITIONAL_OPTS% /p:"LoadExternalDxil=1"
```
python3kgae, line 191: Does this work with  lit?

pow2clk:
Looking into a different approach for this, so I'm closing this PR

PR #5215, opened 2023-05-15
Enable generating per-directory lit test targets

llvm-beanz:
This enables generation of per-directory test targets in non-IDE generators, and preserves a limited set of targets for IDE generators.


python3kgae (approved):

tex3d (approved):

llvm-beanz:
> Maybe add document about how to list the projects in cmake?

Not sure what you mean here. You were separately asking about listing targets. CMake doesn’t support listing targets. It has a `—graphviz` option to dump a graph of the targets.

PR #5216, opened 2023-05-15
Adding test as a copy into currently supported testsuite (for PR #5209)

bfavela:
This is just a test addition to be picked up by our current test runners. Please note it should be deleted when the original location is supported.

Original PR: https://github.com/microsoft/DirectXShaderCompiler/pull/5209

### DIFF
--- a/tools/clang/test/HLSLFileCheck/passes/llvm/globalopt/alias-used.ll
+++ b/tools/clang/test/HLSLFileCheck/passes/llvm/globalopt/alias-used.ll
@@ -1,0 +1,49 @@
+; RUN: %opt %s -globalopt -S | FileCheck %s
+
+; Not in DXC? ; XXCHECK: @main
+
+@c = global i8 42
+
+@i = internal global i8 42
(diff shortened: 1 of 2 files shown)


tools/clang/test/HLSLFileCheck/passes/llvm/globalopt/alias-used.ll
```diff
@@ -0,0 +1,49 @@
+; RUN: %opt %s -globalopt -S | FileCheck %s
+
+; Not in DXC? ; XXCHECK: @main
```
dmpots, line 3: What is this mean? I didn't notice that check in the original test.

bfavela: Argh, mistake of mine when getting this test to run. One of the other .ll files in this area uses @main. I'll delete it

dmpots (approved):
LGTM

llvm-beanz (approved):

PR #5217, opened 2023-05-15
[opt] initialize HLSL passes with SetupRegistryPassForHLSL

python3kgae:
call hlsl::SetupRegistryPassForHLSL in opt to initialize all HLSL/DXIL passes.

This is for FileCheck test which use opt to test HLSL/DXIL passes.

### DIFF
--- a/tools/opt/CMakeLists.txt
+++ b/tools/opt/CMakeLists.txt
@@ -7,6 +7,8 @@ set(LLVM_LINK_COMPONENTS
   DXIL # HLSL Change
   DxcBindingTable # HLSL Change
   HLSL # HLSL Change
+  DxilContainer #  HLSL Change for DxcOptimizerPass
+  DxilRootSignature #  HLSL Change for DxcOptimizerPass
   IPA
   IPO
   IRReader
--- a/tools/opt/opt.cpp
+++ b/tools/opt/opt.cpp
@@ -345,7 +345,6 @@ int __cdecl main(int argc, char **argv) {
   PassRegistry &Registry = *PassRegistry::getPassRegistry();
   initializeCore(Registry);
   initializeScalarOpts(Registry);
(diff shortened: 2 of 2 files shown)


tools/opt/opt.cpp
```diff
@@ -361,10 +364,8 @@ int __cdecl main(int argc, char **argv) {
   //initializeDwarfEHPreparePass(Registry); // HLSL Change: remove EH passes
   //initializeSjLjEHPreparePass(Registry);  // HLSL Change: remove EH passes
   // HLSL Change Starts
-  initializeReducibilityAnalysisPass(Registry);
```
pow2clk, line 364: This pass is actually initialized a few lines above as well. It was redundant then and redundant now, but we might as well remove it.

tex3d (approved):

tools/opt/CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ set(LLVM_LINK_COMPONENTS
   DXIL # HLSL Change
   DxcBindingTable # HLSL Change
   HLSL # HLSL Change
+  DxilContainer #  HLSL Change for DxcOptimizerPass
+  DxilRootSignature #  HLSL Change for DxcOptimizerPass
```
pow2clk, line 11: I didn't look into exactly which passes require these, but I take your word for it 😄

pow2clk (approved):
I would prefer to remove the redundant pass, but it's no worse than before. It's definitely right to initialize all the passes instead of piecemeal as they are needed

tools/opt/opt.cpp
python3kgae: Good catch.
Fixed.

tools/opt/CMakeLists.txt
python3kgae: I'm not sure which pass too. But I'm sure without these two libs the linking will fail.

PR #5219, opened 2023-05-16
Update dependencies, and bump SPIRV-Tools to v2023.3.rc1

Keenuts:
Prepare the next SDK release. Pulling all dependencies, except DirectX headers.


dnovillo (approved):

Keenuts:
> Approving.
> 
> @Keenuts, does this introduce any additional dependencies (like CMake version updates) that we need to be aware of?

Yes it does bump the cmake version to 3.17.2.
 - effce: 3.1 -> 3.17.2
 - spirv-headers: 3.0 -> 3.14

Doesn't seem to add other dependencies.

Keenuts:
@llvm-beanz : is the cmake bump an FYI only for you, or is there some limit? (Like the ~3 years rule for MSVC/clang version) ?

llvm-beanz:
> > Approving.
> > @Keenuts, does this introduce any additional dependencies (like CMake version updates) that we need to be aware of?
> 
> Yes it does bump the cmake version to 3.17.2.
> 
> * effce: 3.1 -> 3.17.2
> * spirv-headers: 3.0 -> 3.14
> 
> Doesn't seem to add other dependencies.

SPIR-V-Tools is already 3.17.2, so I don’t think this meaningfully impacts our situation.

I do think we should investigate removing the dependency on effcee. It is slightly silly that we have a dependency on an external implementation of FileCheck when our project is a fork of the original FileCheck.

This will also be _much_ easier once we’ve adopted LIT as our only test runner, which will allow us to transition them to being true FileCheck tests.

PR #5220, opened 2023-05-16
Remove unused and failing script

bfavela:
This script was being referenced here: https://github.com/microsoft/DirectXShaderCompiler/wiki/Building-Sources

However, the script itself was generating an error as it was out of date (--parallel option to hctbuild was no longer valid as it's parallel by default)

I updated the wiki page and this PR deletes the script.


python3kgae (approved):

llvm-beanz (commented):
There are a few other places that hctcheckin is referenced. Can you remove those references too just to make it a clean removal?

pow2clk (approved):
It's not a very useful script and besides what hcttest already does, it forces something that most would probably do anyway and everyone clearly stopped doing since it wasn't running properly.

bfavela:
> There are a few other places that hctcheckin is referenced. Can you remove those references too just to make it a clean removal?

Good callout. I've removed/updated those indications in a new commit. Are you aware of any other wiki mentions? Google just tells me about hematocrit, whatever that is

llvm-beanz (approved):

PR #4894, opened 2022-12-20
Add RISC-V 64 bit

FantasqueX:
- Add riscv64 target to config.guess
- Update config-ix.cmake to update config-ix.cmake to support riscv64
- Tested on riscv64

I created a similar PR previously which was closed because we should use the latest GPLv2 version of config.guess. x264 project has the same issue and they solved it by only adding riscv64 support to current config.guess. I think same thing could be done in DirectXShaderCompiler.

Ref:
1. https://github.com/microsoft/DirectXShaderCompiler/pull/4690
2. https://code.videolan.org/videolan/x264/-/merge_requests/121

PR #5222, opened 2023-05-16
Added e2k support (MCST Elbrus 2000)

r-a-sattarov:
- Add e2k target to config.guess
- Update config-ix.cmake to support e2k

E2K - this is VLIW/EPIC architecture, like Intel Itanium (IA-64) architecture.

Ref:
1) https://en.wikipedia.org/wiki/Elbrus_2000
2) https://lists.gnu.org/archive/html/config-patches/2015-03/msg00000.html


r-a-sattarov:
> 

llvm-beanz (changes requested):
I'm very unclear about the utility of this change. Is there a modern OS that runs on e2k? While DXC is a fork of LLVM & Clang, DXC is not LLVM or Clang. It does not compile CPU programs, and it doesn't run on the wide array of architectures or operating systems that LLVM & Clang do.

r-a-sattarov:
> I'm very unclear about the utility of this change.

The initial motivation is to make DirectXShaderCompiler be built on e2k (how it was done for riscv64 - https://github.com/microsoft/DirectXShaderCompiler/pull/4894).

> Is there a modern OS that runs on e2k?

Various Linux distributions run on this processor (for example, AltLinux on the 5.4 kernel).

llvm-beanz:
I guess the question I'm really asking is do you have a real use case that involves porting our project to e2k?

RISC-V is different from e2k in that there are readily available devices running RISC-V Linux and we know there's a user base there due to Vulkan's support of RISC-V (which is admittedly limited).

Is there a DirectX12 or Vulkan port to e2k? Or does it have a datacenter use case for pre-compiling GPU workloads?

In general we don't want to accept patches into the project that add functionality which won't be used.

r-a-sattarov:
> I guess the question I'm really asking is do you have a real use case that involves porting our project to e2k?
> Is there a DirectX12 or Vulkan port to e2k? 

Yes, Vulkan was ported to e2k. Because of this, the launch of the Vulkan version of the game RBDOOM-3-BFG became possible. In version 1.5.0, this game switched to using NVRHI and now it became necessary to use DirectXShaderCompiler. That's why I decided to add support for the e2k architecture to DirectXShaderCompiler (under e2k there is also Wicked Engine which uses DirectXShaderCompiler.).

Toxblh (approved):

EntityFX:
Added E2K architecture support for autoconf and cmake. That's OK.

hekota (approved):

PR #5223, opened 2023-05-17
PIX: Synthesize dbg.value for global-backed arrays embedded in non-const global statics

jeffnn:
In some cases, arrays embedded within an HLSL global static will be backed by an llvm global variable, rather than an alloca or a set of discrete values.
In those cases, there will be no dbg.value or dbg.declare statements that will allow PIX to associate stores to those globals with the HLSL variable.
But it's possible to stitch the storage to the HLSL variable and emit the dbg.value (and "fake" alloca/stores that PIX's debug instrumentation uses for tracking changes to values during execution). 
Prior to this work, PIX had no visibility into non-const global statics at all. Now, PIX can see most things, inlcluding the above-described class of embedded arrays. However, more work remains:
-dynamic indexing of the arrays is not considered
-alloca-backed embedded arrays are not covered
-arrays that have been reduced to 1-dimensional equivalents are not covered
The first two will be tackled next. The last will probably be left until and unless requested by a PIX customer.

In brief, this change walks the global variables and compares types, offsets and sizes to determine where arrays might be embedded within other global variables. Maps are constructed to go from a global member's name to the actual member's offset/size, and from the HLSL DIGlobalVariable to the per-function DILocalVariable "alias" or "mirror" of that global (which are denoted by the "global." name prefix) . This allows the "fake" alloca/store and dbg.declares to be emitted for any store to a GEP operand that references those embedded arrays. (If no other members of the HLSL global are accessed, then there will be no DILocalVariable for the new dbg.declares to be associated with, so a new DILocalVariable must be emitted.)


lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -399,12 +419,212 @@ static OffsetInBits GetAlignedOffsetFromDIExpression(
...
+  GlobalVariableToLocalMirrorMap ret;
+  DenseMap<const Function *, DISubprogram *> FnMap;
+
+  for (llvm::Function const *fn : entryPoints) {
+    auto & LocalMirror = ret[fn];
+    auto &blocks = fn->getBasicBlockList();
+    for (auto &block : blocks) {
+      bool breakOut = false;
+      for (auto &instruction : block) {
+        if (auto const *DbgValue = llvm::dyn_cast<llvm::DbgValueInst>(&instruction)) {
```
adam-yang, line 445: Do we just need to check `DbgValueInst`? What about `DbgDeclareInst`?

adam-yang: If we alos need to check `DbgDeclareInst`, we could just dyn_cast to `DbgInfoIntrinsic`.

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -399,12 +419,212 @@ static OffsetInBits GetAlignedOffsetFromDIExpression(
...
+}
+
 bool DxilDbgValueToDbgDeclare::runOnModule(
     llvm::Module &M
 )
 {
   hlsl::DxilModule &DM = M.GetOrCreateDxilModule();
 
+   for (const NamedMDNode &NMD : M.named_metadata()) {
+    NMD.dump();
```
adam-yang, line 624: Debug code?

tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -1837,43 +1864,80 @@ static void CompileAndLogErrors(dxc::DxcDllSupport &dllSupport, LPCSTR pText,
...
   args.push_back(L"/Zi");
   args.push_back(L"/Qembed_debug");
+  args.insert(args.end(), extraArgs.begin(), extraArgs.end());
 
   CompileAndLogErrors(dllSupport, source, profile, args, &pContainer);
 
   auto annotated = RunAnnotationPasses(pContainer);
 
+  auto disText = Disassemble(annotated.blob);
+  (void)disText.c_str();
```
adam-yang, line 1922: Is this just to silence the warning?

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
jeffnn: Good point. I added a check for DbgDeclare too. (I need the variable, so DbgInfoIntrinsic isn't for enough down the class hierarchy.)

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
jeffnn: Darn. Thanks.

tools/clang/unittests/HLSL/PixTest.cpp
jeffnn: Yes.

jeffnn: Actually I just put that .c_str() there for a convenient breakpoint. I'll delete it.

adam-yang (approved):
LGTM!

lib/DxilDia/DxcPixLiveVariables.cpp
```diff
@@ -148,6 +152,13 @@ void dxil_debug_info::LiveVariables::Impl::Init_DbgDeclare(
     return;
   }
 
+  VariableInfoMap *LiveVarInfoMap;
+  if (Variable->getName().startswith("global.")) {
```
python3kgae, line 156: Why is the name start with "global."?
Any pass did that?

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -399,12 +419,217 @@ static OffsetInBits GetAlignedOffsetFromDIExpression(
...
+llvm::DISubprogram* GetFunctionDebugInfo(llvm::Module& M, llvm::Function* fn) {
+  auto FnMap = makeSubprogramMap(M);
+  return FnMap[fn];
+}
+
+GlobalVariableToLocalMirrorMap
+GenerateGlobalToLocalMirrorMap(llvm::Module &M,
+                                    llvm::DIGlobalVariable *DIGV) {
+  hlsl::DxilModule &DM = M.GetOrCreateDxilModule();
+  auto entryPoints = DM.GetExportedFunctions();
```
python3kgae, line 431: So internal functions don't need to add the global mirror?
We might need it if noinline is marked.

lib/DxilDia/DxcPixLiveVariables.cpp
python3kgae: Found it.

adam-yang: ScalarReplAggregateHLSL's `PatchDebugInfo` did it.

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
jeffnn: Good point. I made changes to operate on all functions.

PR #5224, opened 2023-05-17
[lit] Add COND_RUN to allow each RUN line has different requires

python3kgae:
A new command type COND_RUN is added to lit.
The format is COND_RUN(feature_list): cmd
The cmd will only be executed when all the features in feature_list are supported.

This is to support case like
%dxilver 1.5 and %dxilver 1.6 in same test file.

### DIFF
--- a/utils/lit/lit/TestRunner.py
+++ b/utils/lit/lit/TestRunner.py
@@ -479,12 +479,36 @@ def parseIntegratedTestScript(test, normalize_slashes=False,
             # Trim spaces.
             cond_run_requires = [require.strip()
                       for require in cond_run_requires]
+
+            # Trim trailing whitespace.
+            ln = ln.rstrip()
+
             # Collapse lines with trailing '\\'.
             if cond_script and cond_script[-1][1][-1] == '\\':
-                cond_script[-1] = cond_script[-1][1][:-1] + ln
+                prev = cond_script[-1]
+                cond_run = (prev[0], prev[1][:-1] + ln)
+                cond_script[-1] = cond_run
             else:
                 cond_run = (cond_run_requires, ln)
                 cond_script.append(cond_run)
+
+            # No collapse anymore.
+            if ln[-1] != '\\':
+                prev = cond_script[-1]
+                cond_run_requires = prev[0]
+                # Check that we have the required features:
+                missing_required_features = [f for f in cond_run_requires
+                                    if f not in test.config.available_features]
+                ln = prev[1]
+                if missing_required_features:
+                    msg = ', '.join(missing_required_features)
+                    print('COND_RUN {} skipped for unsupport {}'.format(ln, msg))
+                    cond_run = (False, ln)
+                    cond_script.append(cond_run)
+                    continue
+
+                # Feature supported, append to script.
+                script.append(ln)
         else:
             raise ValueError("unknown script command type: %r" % (
                     command_type,))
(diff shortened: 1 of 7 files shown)


utils/lit/lit/TestRunner.py
```diff
@@ -471,6 +473,18 @@ def replace_line_number(match):
...
+            cond_run_requires = cond_run_requires_str.split(',')
+            # Trim spaces.
+            cond_run_requires = [require.strip()
+                      for require in cond_run_requires]
+            # Collapse lines with trailing '\\'.
+            if cond_script and cond_script[-1][1][-1] == '\\':
+                cond_script[-1] = cond_script[-1][1][:-1] + ln
+            else:
+                cond_run = (cond_run_requires, ln)
+                cond_script.append(cond_run)
```
tex3d, line 487: Shouldn't conditional run lines be ordered in the same list as normal run lines, rather than gathered separately and tacked onto the end?  Can these conditions be tested within the code gathering the lines, then added to the same script if they pass?  Then the `if command_type == 'RUN':` block could be modified to handle `COND_RUN(` at the same time, only checking the condition and potentially skipping if it's `command_type.startswith('COND_RUN(')`.

utils/lit/tests/Inputs/shtest-format/cond_run_requires-missing.txt
```diff
@@ -0,0 +1,2 @@
+COND_RUN(a-missing-feature): true
+COND_RUN(a-missing-feature, a-present-feature): true
```
tex3d, line 2: Does `true` simply return a non-failing exit code?  If so, don't we want a failing exit code to ensure it was skipped?

tex3d (changes requested):

utils/lit/lit/TestRunner.py
python3kgae: Good catch.
Will fix it.

python3kgae: Fixed.

utils/lit/tests/Inputs/shtest-format/cond_run_requires-present.txt
```diff
@@ -0,0 +1,3 @@
+COND_RUN(a-present-feature): true
```
llvm-beanz, line 1: These test files are _inputs_ to tests, you don't have any updates to the FileCheck test to verify that they actually do what you're intending.

llvm-beanz (commented):
I feel like the parsing and language for this is a bit clunky. What about something more like `RUN-IF(...):`. This keeps a `RUN` prefix, makes it brief without shortening words, and uses `-` instead of `_` which matches FileCheck directives.

bogner:
If we want to add something like this to lit we should probably bring it up in upstream for discussion too. There are a lot of users of lit there that might have helpful ideas on the design of this feature.

python3kgae:
> If we want to add something like this to lit we should probably bring it up in upstream for discussion too. There are a lot of users of lit there that might have helpful ideas on the design of this feature.

Should I create a RFC somewhere?

bogner:
> > If we want to add something like this to lit we should probably bring it up in upstream for discussion too. There are a lot of users of lit there that might have helpful ideas on the design of this feature.
> 
> Should I create a RFC somewhere?

Yeah, writing up an RFC on https://discourse.llvm.org/ is probably the right course of action.

python3kgae:
> > > If we want to add something like this to lit we should probably bring it up in upstream for discussion too. There are a lot of users of lit there that might have helpful ideas on the design of this feature.
> > 
> > 
> > Should I create a RFC somewhere?
> 
> Yeah, writing up an RFC on https://discourse.llvm.org/ is probably the right course of action.

Found https://github.com/llvm/llvm-project/commit/1041a9642ba035fd2685f925911d705e8edf5bb0
Might works for our use case.

tex3d:
> > > > If we want to add something like this to lit we should probably bring it up in upstream for discussion too. There are a lot of users of lit there that might have helpful ideas on the design of this feature.
> > > 
> > > 
> > > Should I create a RFC somewhere?
> > 
> > 
> > Yeah, writing up an RFC on https://discourse.llvm.org/ is probably the right course of action.
> 
> Found [llvm/llvm-project@1041a96](https://github.com/llvm/llvm-project/commit/1041a9642ba035fd2685f925911d705e8edf5bb0) Might works for our use case.

Looks like it might.  I suppose the only disadvantage compared with this one is that there's no feedback like UNSUPPORTED for you to tell how much was skipped entirely.  Of course the existing mechanism doesn't have that feedback under FileCheck either.

python3kgae:
Ported https://github.com/llvm/llvm-project/commit/1041a9642ba035fd2685f925911d705e8edf5bb0 in
https://github.com/microsoft/DirectXShaderCompiler/pull/5227

PR #5225, opened 2023-05-18
[FileCheck] Add -dump-input to dump full input.

python3kgae:
New option -dump-input is added to help debug.
The default behavior is to dump when fail.
-dump-input=never will disable this.


python3kgae:
-dump-input in upstream is more powerful but it is not an easy single PR merge.
This simple dump should be good enough for now.

llvm-beanz (commented):
I would greatly prefer we not change the default behavior for this.

Our test logs are already extremely long and verbose when things go wrong and spewing lots and lots of extra output for each failure will only make that worse.

test/FileCheck/dump-input/fail.test
```diff
@@ -0,0 +1,5 @@
+RUN: not FileCheck -dump-input=fail --input-file=%S/Input/input.input %s  2>&1 | FileCheck  %S/Input/check-input.input
```
llvm-beanz, line 1: I think you have this test written backwards. The files in the `Inputs` should be the full test case, so they should be the input and check lines for the first `FileCheck` call. This file's check lines should be the ones verifying the result (the second FileCheck call).

python3kgae: Fixed.

tex3d (approved):
I think this is a useful capability, which we can experiment with turning on for pipeline builds.  It doesn't add to the large log when there are no failures, but could help investigations significantly in certain failure cases, especially when those failures prove difficult to reproduce outside the pipeline configuration.

llvm-beanz:
This change doesn’t give us a way to opt-in globally, so I’m unclear on how it would be used. Opting in on pipeline builds is _exactly_ what I’m concerned about. Our pipeline builds are already so overly verbose that it is really difficult to figure out which tests failed, much less identify the failure.

python3kgae:
> This change doesn’t give us a way to opt-in globally, so I’m unclear on how it would be used. Opting in on pipeline builds is _exactly_ what I’m concerned about. Our pipeline builds are already so overly verbose that it is really difficult to figure out which tests failed, much less identify the failure.

How about something like
#ifdef FORCE_DUMP
        cl::init(DumpInputFail),
#else
        cl::init(DumpInputNever),
#endif

damyanp:

damyanp:

PR #5228, opened 2023-05-19
[WIP] Fixed entry point omitted from the PDB when it's the same as input filename

adam-yang:


tools/clang/tools/dxcompiler/dxcshadersourceinfo.h
```diff
@@ -56,7 +56,7 @@ struct SourceInfoWriter {
   Buffer m_Buffer;
 
   const hlsl::DxilSourceInfo *GetPart() const;
-  void Write(llvm::StringRef targetProfile, llvm::StringRef entryPoint, clang::CodeGenOptions &cgOpts, clang::SourceManager &srcMgr);
+  void Write(clang::CodeGenOptions &cgOpts, llvm::ArrayRef<hlsl::options::ArgPair> ArgPairs, clang::SourceManager &srcMgr);
```
damyanp, line 59: Any reason this uses the full `hlsl::optons::ArgPair` while other instances in this file use just `options::ArgPair`?

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -69,6 +69,116 @@ const OptTable * hlsl::options::getHlslOptTable() {
...
+    //   { "",    "Value0" }
+    //   { "",    "Value1" }
+    //   { "",    "Value2" }
+    // 
+    // ..Which is not ideal, since the whole point of arg pair is to easily break down what arg is part of what.
+    // 
+    // Alternative is to try this:
+    //   { "arg", "Value0 Value1 Value2" }
+    // 
+    // But this causes problem when recompiling
```
damyanp, line 140: > "But this causes problem when recompiling"

What does this mean?  Is the problem that it doesn't round-trip properly - ie you can't trivially reconstruct a command-line when it's stored in this alternate form?

tools/clang/tools/dxcompiler/dxcshadersourceinfo.h
damyanp, line 59: Params to this function look like they have inconsistent naming conventions to me:
* `cgOpts`
* `ArgPairs`
* `srcMgr`

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -69,6 +69,116 @@ const OptTable * hlsl::options::getHlslOptTable() {
   return g_HlslOptTable;
 }
 
+static std::vector<ArgPair> ComputeArgPairsFromArgList(const llvm::opt::InputArgList &argList) {
```
damyanp, line 72: FWIW, it's a shame there's not a single set of unit tests I can look at that directly exercise all the various cases and edge cases that `ComputeArgPairsFromArgList` handles.  Presumably there's good coverage for all 5 or so types of arguments this supports through the rest of the test suite?

damyanp (approved):

lib/DxcSupport/HLSLOptions.cpp
adam-yang: Actually we do in the new test case. -Zi and -Od are ::FlagClass, -T and -E are `arg->getNumValues() == 1`. The input file is `InputClass`.

adam-yang: This made me revise this area. Before the latest revision, the OP_INPUT check would skip the iteration and we would never hit the ::InputClass case. Now the skip is done by the InputClass.

lib/DxcSupport/HLSLOptions.cpp
adam-yang: Added an explanation for how it actually failed.

tools/clang/tools/dxcompiler/dxcshadersourceinfo.h
adam-yang: Removed it. That might have been from copypasting from other places.

adam-yang: Fixed it.

damyanp (approved):

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1571,15 +1571,14 @@ class DxcCompiler : public IDxcCompiler3,
     else
       compiler.getCodeGenOpts().HLSLSignaturePackingStrategy = (unsigned)DXIL::PackingStrategy::Default;
 
-    // Constructing vector of wide strings to pass in to codegen. Just passing
-    // in pArguments will expose ownership of memory to both CodeGenOptions and
-    // this caller, which can lead to unexpected behavior.
-    for (UINT32 i = 0; i != Opts.Args.getNumInputArgStrings(); ++i) {
```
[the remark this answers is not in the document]
adam-yang: I think that's the right short term fix to minimize risk for release, so I'll make a separate PR with just that fix, but I still think this is the right change to do long term. I'll rebase later.

include/dxc/Support/HLSLOptions.td
```diff
@@ -170,6 +170,11 @@ def opt_enable : Separate<["-", "/"], "opt-enable">, Group<hlsloptz_Group>, Flag
 def opt_select : MultiArg<["-", "/"], "opt-select", 2>, MetaVarName<"<opt> <variant>">, Group<hlsloptz_Group>, Flags<[CoreOption, DriverOption, HelpHidden]>,
   HelpText<"Select this optimization variant.">;
 
+def CommaJoinTest : CommaJoined<["-"], "CommaJoinTest=">, Group<spirv_Group>, Flags<[CoreOption, HelpHidden]>,
```
llvm-beanz, line 173: I assume these options aren’t actually SPIR-V specific right? They should probably be in a different Group,

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -69,6 +69,119 @@ const OptTable * hlsl::options::getHlslOptTable() {
...
+  std::vector<ArgPair> ret;
```
llvm-beanz, line 73: nit: `SmallVector<>` is probably better here to reduce resize occurrences.

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -69,6 +69,119 @@ const OptTable * hlsl::options::getHlslOptTable() {
...
+
+    llvm::SmallString<64> argumentStorage; // Buffer for when we need to concatenate things.
+
+    // The input argument should be excluded from this list. All the uses for
+    // the ArgPair list used to manually filter out the input file. Do it here
+    // in the source instead.
+    if (arg->getOption().getKind() == llvm::opt::Option::InputClass) {
+      continue;
+    }
+    // This is a flag without value, like -Zi, or -Od
```
llvm-beanz, line 87: nit: can you clang-format the new code?

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -69,6 +69,119 @@ const OptTable * hlsl::options::getHlslOptTable() {
   return g_HlslOptTable;
 }
 
+static std::vector<ArgPair> ComputeArgPairsFromArgList(const llvm::opt::InputArgList &argList) {
```
llvm-beanz, line 72: Rather than returning a std::vector which guarantees a copy, can we pass a `SmallVectorImpl<>&` to avoid copying the array on return? That also allows us to avoid resizes.

lib/DxcSupport/HLSLOptions.cpp
llvm-beanz: Since this function is pretty isolated and static, can we write unit tests for this function?

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -69,6 +69,119 @@ const OptTable * hlsl::options::getHlslOptTable() {
...
+
+    // The input argument should be excluded from this list. All the uses for
+    // the ArgPair list used to manually filter out the input file. Do it here
+    // in the source instead.
+    if (arg->getOption().getKind() == llvm::opt::Option::InputClass) {
+      continue;
+    }
+    // This is a flag without value, like -Zi, or -Od
+    else if (arg->getOption().getKind() == llvm::opt::Option::FlagClass) {
+      assert(arg->getNumValues() == 0);
```
llvm-beanz, line 89: nit: since the body of this condition is just an assert, we could hoist the assert out before the first `if` and have the assert contain the condition and the assert check, then this `else if` block could disappear.

damyanp:

damyanp:

PR #5179, opened 2023-04-27
Compilation error fixes

neurotok:
Bugs found during Vulkan SDK 1.3.243 compilation under Fedora 38

- changes meaning of ‘ValueMap’ error
- missing cstdint include

PR #5229, opened 2023-05-19
[NFC] Fix GCC 13 warnings

ncesario-lunarg:
Fixes some of the warnings found when compiling on GCC 13.

### DIFF
--- a/include/llvm/Analysis/DxilValueCache.h
+++ b/include/llvm/Analysis/DxilValueCache.h
@@ -51,8 +51,7 @@ struct DxilValueCache : public ImmutablePass {
   };
 
 private:
-
-  WeakValueMap ValueMap;
+  WeakValueMap DxilValueMap;
   bool (*ShouldSkipCallback)(Value *V) = nullptr;
 
   void MarkUnreachable(BasicBlock *BB);
@@ -77,8 +76,8 @@ public:
   Value *GetValue(Value *V, DominatorTree *DT=nullptr);
   Constant *GetConstValue(Value *V, DominatorTree *DT = nullptr);
   ConstantInt *GetConstInt(Value *V, DominatorTree *DT = nullptr);
-  void ResetUnknowns() { ValueMap.ResetUnknowns(); }
-  void ResetAll() { ValueMap.ResetAll(); }
+  void ResetUnknowns() { DxilValueMap.ResetUnknowns(); }
+  void ResetAll() { DxilValueMap.ResetAll(); }
   bool IsUnreachable(BasicBlock *BB, DominatorTree *DT=nullptr);
   void SetShouldSkipCallback(bool (*Callback)(Value *V)) { ShouldSkipCallback = Callback; };
 };
--- a/lib/ProfileData/SampleProfWriter.cpp
+++ b/lib/ProfileData/SampleProfWriter.cpp
@@ -122,5 +122,5 @@ SampleProfileWriter::create(StringRef Filename, SampleProfileFormat Format) {
   if (EC)
     return EC;
 
-  return std::move(Writer);
+  return Writer;
 }
--- a/tools/clang/lib/AST/DeclGroup.cpp
+++ b/tools/clang/lib/AST/DeclGroup.cpp
@@ -30,5 +30,6 @@ DeclGroup* DeclGroup::Create(ASTContext &C, Decl **Decls, unsigned NumDecls) {
 DeclGroup::DeclGroup(unsigned numdecls, Decl** decls) : NumDecls(numdecls) {
   assert(numdecls > 0);
   assert(decls);
-  memcpy(this+1, decls, numdecls * sizeof(*decls));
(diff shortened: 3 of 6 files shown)


include/llvm/Analysis/DxilValueCache.h
```diff
@@ -51,8 +51,7 @@ struct DxilValueCache : public ImmutablePass {
   };
 
 private:
-
-  WeakValueMap ValueMap;
```
ncesario-lunarg, line 55: Hides the `ValueMap` above.

tools/clang/lib/Parse/ParseExpr.cpp
```diff
@@ -446,24 +446,28 @@ Parser::ParseRHSOfBinaryExpression(ExprResult LHS, prec::Level MinPrec) {
...
+    } else {
       // Ensure potential typos in the RHS aren't left undiagnosed.
       Actions.CorrectDelayedTyposInExpr(RHS);
-      // HLSL Change Begin - Take care TernaryMiddle.
-      Actions.CorrectDelayedTyposInExpr(TernaryMiddle);
-      // HLSL Change End.
+    }
+
+    // HLSL Change Begin - Take care TernaryMiddle.
+    Actions.CorrectDelayedTyposInExpr(TernaryMiddle);
```
ncesario-lunarg, line 469: Was this meant to be part of the else clause?

llvm-beanz:
Some of these changes were already proposed in #5179.

lib/ProfileData/SampleProfWriter.cpp
```diff
@@ -122,5 +122,5 @@ SampleProfileWriter::create(StringRef Filename, SampleProfileFormat Format) {
   if (EC)
     return EC;
 
-  return std::move(Writer);
```
llvm-beanz, line 125: This likely cuts both ways. Clang and GCC disagree about when std::move is valid, and that behavior has changed based on which C++ language version you're targeting.

tools/clang/lib/AST/DeclGroup.cpp
```diff
@@ -30,5 +30,6 @@ DeclGroup* DeclGroup::Create(ASTContext &C, Decl **Decls, unsigned NumDecls) {
 DeclGroup::DeclGroup(unsigned numdecls, Decl** decls) : NumDecls(numdecls) {
   assert(numdecls > 0);
   assert(decls);
-  memcpy(this+1, decls, numdecls * sizeof(*decls));
+  memcpy(reinterpret_cast<void *>(this + 1),
```
llvm-beanz, line 33: `static_cast` is preferred when converting to `void*` as it is legal for the narrowest valid cast.

tools/clang/lib/Parse/ParseExpr.cpp
```diff
@@ -446,24 +446,28 @@ Parser::ParseRHSOfBinaryExpression(ExprResult LHS, prec::Level MinPrec) {
         // If we're using '>>' as an operator within a template
         // argument list (in C++98), suggest the addition of
         // parentheses so that the code remains well-formed in C++0x.
-        if (!GreaterThanIsOperator && OpToken.is(tok::greatergreater))
+        if (!GreaterThanIsOperator && OpToken.is(tok::greatergreater)) {
```
llvm-beanz, line 449: This change looks like it changes behavior, and not in a correct way. I assume the GCC warning you're fixing here is something about alignment and conditional nesting... I suspect that warning should be disabled.

tools/clang/lib/Parse/ParseExpr.cpp
llvm-beanz: This line probably does nothing ever and should probably just be removed. Typo correction for the middle expression occurs earlier, and HLSL doesn't have anything special here from C/C++ behaviors.

Locally if I remove this block of HLSL-specific code nothing fails, so it should be fine.

llvm-beanz (changes requested):

include/llvm/Analysis/DxilValueCache.h
ncesario-lunarg: Looks like this was taken care of with a98b242cacfb42c0277c5eabe876fbb47a885f60.

lib/ProfileData/SampleProfWriter.cpp
ncesario-lunarg: :+1:  FWIW, clang 16.0.3 appears to be "happy" with and without `std::move`.

tools/clang/lib/Parse/ParseExpr.cpp
ncesario-lunarg: Sorry, I'm not seeing how this changes functionality? In any case I will revert as I do not want to incur undue confusion. I found this code rather difficult to read, but that could just be me.

ncesario-lunarg:
Closing this as it looks like a lot of these proposed changes have already merged.

Issue #4911, opened 2023-01-04
[SPIR-V] Shader failes to compile when activating debug infos


Clothoid1:
The following reproducer does not compile, leading to the error message:
```
fatal error : generated SPIR-V is invalid: ID '52[%52]' has not been defined
1>  %51 = OpExtInst %void %1 DebugTypeFunction %uint_3 %52 %48 %49
1>
1>note: please file a bug report on https://github.com/Microsoft/DirectXShaderCompiler/issues with source code if possible
```

Reproducer:
```hlsl
StructuredBuffer<uint16_t> myIntersectionMaskBuffer_BitCompressed : register(t0);

class CompressedIntersectionBits
{
	uint16_t myCompressedData;

	static CompressedIntersectionBits read_const(StructuredBuffer<uint16_t> aIntersectionMaskBuffer_BitCompressed_in, const uint aRayIndex_in)
	{
		CompressedIntersectionBits aCompressedIntersectionBits;
		aCompressedIntersectionBits.myCompressedData = aIntersectionMaskBuffer_BitCompressed_in[aRayIndex_in];
		return aCompressedIntersectionBits;
	}
};

class CompressedIntersectionBits_Test
{
	static CompressedIntersectionBits read_const(StructuredBuffer<uint16_t> aIntersectionMaskBuffer_BitCompressed_in, const uint aRayIndex_in)
	{
		CompressedIntersectionBits aCompressedIntersectionBits;
		aCompressedIntersectionBits.myCompressedData = aIntersectionMaskBuffer_BitCompressed_in[aRayIndex_in];
		return aCompressedIntersectionBits;
	}
};

void runKernel(const uint aIndexX_in, const uint aIndexY_in)
{
	uint aRayIndex = aIndexY_in * 512 + aIndexX_in;

	CompressedIntersectionBits aCompressedIntersectionBits = CompressedIntersectionBits::read_const(myIntersectionMaskBuffer_BitCompressed, aRayIndex);
	//CompressedIntersectionBits aCompressedIntersectionBits = CompressedIntersectionBits_Test::read_const(myIntersectionMaskBuffer_BitCompressed, aRayIndex);
}

[numthreads(8, 8, 1)]
void main(uint3 GlobalInvocationID : SV_DispatchThreadID, uint3 LocalInvocationID : SV_GroupThreadID)
{
	runKernel(GlobalInvocationID[0], GlobalInvocationID[1]);
}
```

Command Line used:
[...]

PR #5233, opened 2023-05-23
spirv: fix forward reference order for composite

Keenuts:
The blanket rule for NonSemantic.Shader.DebugInfo.100 is "no forward references". But there is a type-specific rule: DebugTypeComposite members can be forward references:

```
Members must be the <id>s of DebugTypeMember, DebugFunction, or
DebugTypeInheritance. This could be a forward reference.
```

This is mandatory to correctly support HLSL class methods:
 - the composite has a function as member.
 - the function as the composite as scope. This cycle makes the forward reference mandatory, hence the exception in the spec.

The removed code wrongly implemented the general rule, ignoring this exception.

This change doesn't just solves the cycle issue, but might add forward references that were avoided (since we never visit members). But as far as the spec goes, this is a valid behavior. So I believe there is no need to add any additional logic to avoid forward references.

This PR depends on the validator to correctly validate those forward references (KhronosGroup/SPIRV-Tools#5230).

Fixes: #4911.

### DIFF
--- a/tools/clang/lib/SPIRV/SortDebugInfoVisitor.cpp
+++ b/tools/clang/lib/SPIRV/SortDebugInfoVisitor.cpp
@@ -120,20 +120,6 @@ void SortDebugInfoVisitor::whileEachOperandOfDebugInstruction(
       break;
     if (!visitor(inst->getDebugInfoNone()))
       break;
-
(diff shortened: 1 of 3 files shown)


Keenuts:
This PR is ready for review. Marking as draft only because it depends on https://github.com/KhronosGroup/SPIRV-Tools/pull/5230

PR #5234, opened 2023-05-23
Allow function template default parameters in HLSL

### LINKED ISSUE #5221: default template arguments for a function template are a C++11 extension
When trying to compile HLSL like this:
```
template<typename TextureType, typename MixerType = TrivialMixer>
vec4	textureMaterial( TextureType tex, UVProjector p, MixerType m = MixerType() )
{
    return textureWithSampler( tex, sMaterialSampler, p.uv.xy );
}
```
With arguments:
```
-T vs_6_5 -H 2021 -O shader_0.hls
```

I get the following warning:
```
warning: default template arguments for a function template are a C++11 extension [-Wc++11-extensions]
```

It seems I can silence it with either `-no-warnings` or `-Wno-c++11-extensions`.

Additionally it seems specifying the C++ version with `-std=c++11` does work:
```
Unknown argument: '-std=c++11'
```

It feels like this should either:
- be not be allowed
- possible to disable a warning for this specific extension not all of them
- possible to be able to specify c++ version

In addition - is HLSL 2021 compliant with any actual version of C++? If not this is a strange warning to get altogether,

llvm-beanz:
Default parameters for function templates is a C++11 feature. Since HLSL is defined as C++03 using default parameters for function templates produces a language extension warning. There is really no need for the warning since the template syntax is disallowed in HLSL 2018 and earlier and we effectively always support this syntax in 2021.

This change suppresses the warning by treating HLSL as C++11 when diagnosing default template arguments. This changes the warning to a C++98 compatability warning, which is only enabled if explicitly requested.

In the future we should consider making future HLSL language versions based on a more modern C++ to avoid issues like this.

Fixes #5221

### DIFF
--- a/tools/clang/unittests/HLSL/VerifierTest.cpp
+++ b/tools/clang/unittests/HLSL/VerifierTest.cpp
@@ -50,6 +50,7 @@ public:
   TEST_METHOD(RunOperatorOverloadingNotDefinedBinaryOp)
   TEST_METHOD(RunCXX11Attributes)
   TEST_METHOD(RunEnums)
+  TEST_METHOD(RunFunctionTemplateDefault)
   TEST_METHOD(RunFunctions)
   TEST_METHOD(RunIncompleteType)
   TEST_METHOD(RunIndexingOperator)
@@ -241,6 +242,10 @@ TEST_F(VerifierTest, RunCXX11Attributes) {
 
 TEST_F(VerifierTest, RunEnums) {
   CheckVerifiesHLSL(L"enums.hlsl");
+}
+
+TEST_F(VerifierTest, RunFunctionTemplateDefault) {
+  CheckVerifiesHLSL(L"function-template-default.hlsl");
 }
 
 TEST_F(VerifierTest, RunFunctions) {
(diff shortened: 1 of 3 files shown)


tools/clang/unittests/HLSL/VerifierTest.cpp
```diff
@@ -243,6 +244,10 @@ TEST_F(VerifierTest, RunEnums) {
   CheckVerifiesHLSL(L"enums.hlsl");
 }
 
+TEST_F(VerifierTest, RunFunctionTemplateDefault) {
+  CheckVerifiesHLSL(L"function-template-default.hlsl");
```
bob80905, line 248: where is this hlsl file located?

llvm-beanz: Doh!

I didn't add it to the commit. Fixed in d46e73e3d60b33c9548b1f98ca77ded985f8ee38

bob80905 (approved):

dmpots (approved):
LGTM

PR #5235, opened 2023-05-23
[lit] Use ls -l instead of file -E

python3kgae:
Avoid use file -E because ls is more available on Windows.
Use ls -1 %T | not grep "\\.pdb$" instead of not ls -1 %T because not ls *.pdb not work on windows.

### DIFF
--- a/tools/clang/test/DXC/embed_dbg.test
+++ b/tools/clang/test/DXC/embed_dbg.test
@@ -1,7 +1,7 @@
 // Embed debug info
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
-// RUN: not ls -l *.pdb
+// RUN: not ls *.pdb && count 0
 
 // Auto-embed debug info when no debug output, and expect warning signifying that this is the case.
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl  /Zi /Fo %t.embedpdb2 /Fe %t.err.embedpdb
--- a/tools/clang/test/DXC/embed_dbg.test
+++ b/tools/clang/test/DXC/embed_dbg.test
@@ -2,8 +2,7 @@
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
 
-// RUN: ls  %T | not grep "\\.pdb$"
-
+// RUN: ls -1 %T | not grep "\\.pdb$"
 
 // Auto-embed debug info when no debug output, and expect warning signifying that this is the case.
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl  /Zi /Fo %t.embedpdb2 /Fe %t.err.embedpdb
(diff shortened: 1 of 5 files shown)


dmpots (approved):
LGTM

pow2clk (approved):
I'm still slightly suspicious that it can't be this simple 🤔. I can't see why it shouldn't be though!

tools/clang/test/DXC/root_sig.test
```diff
@@ -27,7 +27,7 @@
 // Set root signature when already has one should succeed
 // RUN: %dxc %t.define.cso /dumpbin /setrootsignature %t.rootsig.cso /Fo %t.smoke.rsadded.cso
 // Make sure %t.smoke.rsadded.cso exist.
-// RUN: file -E %t.smoke.rsadded.cso
+// RUN: ls -l %t.smoke.rsadded.cso
```
bogner, line 30: Why `-l`? I think `ls` by itself (or `ls -1` if we're checking the output maybe) is sufficient.

tools/clang/test/DXC/embed_dbg.test
```diff
@@ -1,7 +1,7 @@
 // Embed debug info
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
-// RUN: not file -E *.pdb
+// RUN: not ls -l *.pdb
```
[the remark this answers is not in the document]
python3kgae: Updated.

tools/clang/test/DXC/root_sig.test
python3kgae: Change to only ls.

tools/clang/test/DXC/embed_dbg.test
```diff
@@ -1,7 +1,7 @@
 // Embed debug info
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
-// RUN: not file -E *.pdb
+// RUN: not ls *.pdb && count 0
```
bogner, line 4: This should be `not ls *.pdb | count 0` rather than `&&` - count reads its standard input

python3kgae: Fixed.

tools/clang/test/DXC/embed_dbg.test
```diff
@@ -1,7 +1,7 @@
 // Embed debug info
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
-// RUN: not file -E *.pdb
+// RUN: not ls *.pdb | count 0
```
tex3d, line 4: Why isn't this `ls *.pdb | not count 0`?

hekota (approved):

tools/clang/test/DXC/strip_dbg.test
```diff
@@ -1,7 +1,7 @@
 // /Fd implies /Qstrip_debug ; path with / produces auto hash-named .pdb file
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Fd ./ /Fo %t.strip
 // RUN: %dxc -dumpbin %t.strip | FileCheck  %s --check-prefix=STRIP_DBG
-// RUN: file -E *.pdb && rm *.pdb
+// RUN: ls *.pdb && rm *.pdb
```
tex3d, line 4: I still feel like we should have changed the option above where it says `/Fd ./` to `/Fd %t.pdbdir/` after making that directory.  Then we don't have to remove the pdb here.

tex3d (approved):
Tricky getting these shell commands all working the way you want!

PR #5236, opened 2023-05-23
Fixed entry point omitted from the PDB when it's the same as input filename

adam-yang:
The args in the PDB and debug DXIL are based `CodeGenOptions::HLSLArgs`. This list was created by copying the raw unprocessed arg list and excluding any args equal to the input file name. When the entry point happens to be the same as the input filename, it would get excluded from this list and the PDB and debug module end up with corrupted arg list where -E has no value or wrong value.

Fixed by only excluding args that are InputClass.

### DIFF
--- a/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
+++ b/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
@@ -1575,14 +1575,18 @@ public:
     // in pArguments will expose ownership of memory to both CodeGenOptions and
     // this caller, which can lead to unexpected behavior.
     {
+      // Find all args that are of Option::InputClass and record their indices
+      // in a set. If there are multiple Option::InputClass arguments, exclude
+      // all of them. We only use the last one and there's no point recording
+      // the rest of them.
       llvm::DenseSet<unsigned> InputArgIndices;
       for (llvm::opt::Arg *arg : Opts.Args.getArgs()) {
         if (arg->getOption().getKind() == llvm::opt::Option::InputClass)
           InputArgIndices.insert(arg->getIndex());
       }
       for (unsigned i = 0; i < Opts.Args.getNumInputArgStrings(); ++i) {
-        StringRef argStr = Opts.Args.getArgString(i);
-        if (InputArgIndices.count(i) == 0) {
+        if (InputArgIndices.count(i) == 0) { // Only include this arg if it's not in the set of Option::InputClass args.
+          StringRef argStr = Opts.Args.getArgString(i);
           compiler.getCodeGenOpts().HLSLArguments.emplace_back(argStr);
         }
       }
--- a/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
+++ b/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
@@ -1579,6 +1579,10 @@ public:
       // in a set. If there are multiple Option::InputClass arguments, exclude
       // all of them. We only use the last one and there's no point recording
       // the rest of them.
+      //
+      // This list is used to populate the argument list in debug module and
+      // PDB, which are for recompiling. The input filenames are not needed for
+      // it and should be excluded.
       llvm::DenseSet<unsigned> InputArgIndices;
       for (llvm::opt::Arg *arg : Opts.Args.getArgs()) {
         if (arg->getOption().getKind() == llvm::opt::Option::InputClass)
--- a/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
+++ b/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
@@ -1571,15 +1571,23 @@ public:
     else
       compiler.getCodeGenOpts().HLSLSignaturePackingStrategy = (unsigned)DXIL::PackingStrategy::Default;
 
-    // Constructing vector of wide strings to pass in to codegen. Just passing
+    // Constructing vector of strings to pass in to codegen. Just passing
     // in pArguments will expose ownership of memory to both CodeGenOptions and
     // this caller, which can lead to unexpected behavior.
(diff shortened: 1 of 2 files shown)


tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1571,15 +1571,23 @@ class DxcCompiler : public IDxcCompiler3,
...
 
-    // Constructing vector of wide strings to pass in to codegen. Just passing
+    // Constructing vector of strings to pass in to codegen. Just passing
     // in pArguments will expose ownership of memory to both CodeGenOptions and
     // this caller, which can lead to unexpected behavior.
-    for (UINT32 i = 0; i != Opts.Args.getNumInputArgStrings(); ++i) {
-      auto arg = Opts.Args.getArgString(i);
-      if (Opts.InputFile.compare(arg) != 0) {
-        compiler.getCodeGenOpts().HLSLArguments.emplace_back(arg);
+    {
```
damyanp, line 1577: Could probably do with a comment here explaining what's going on.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1571,15 +1571,23 @@ class DxcCompiler : public IDxcCompiler3,
...
+      llvm::DenseSet<unsigned> InputArgIndices;
+      for (llvm::opt::Arg *arg : Opts.Args.getArgs()) {
+        if (arg->getOption().getKind() == llvm::opt::Option::InputClass)
+          InputArgIndices.insert(arg->getIndex());
+      }
+      for (unsigned i = 0; i < Opts.Args.getNumInputArgStrings(); ++i) {
+        StringRef argStr = Opts.Args.getArgString(i);
```
damyanp, line 1584: Is there a reason this is outside the `if`?  (Maybe it's important that `getArgString` gets called for each argument?

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1571,15 +1571,23 @@ class DxcCompiler : public IDxcCompiler3,
...
     // in pArguments will expose ownership of memory to both CodeGenOptions and
     // this caller, which can lead to unexpected behavior.
-    for (UINT32 i = 0; i != Opts.Args.getNumInputArgStrings(); ++i) {
-      auto arg = Opts.Args.getArgString(i);
-      if (Opts.InputFile.compare(arg) != 0) {
-        compiler.getCodeGenOpts().HLSLArguments.emplace_back(arg);
+    {
+      llvm::DenseSet<unsigned> InputArgIndices;
+      for (llvm::opt::Arg *arg : Opts.Args.getArgs()) {
+        if (arg->getOption().getKind() == llvm::opt::Option::InputClass)
```
dmpots, line 1580: It would be good to add a comment here explaining why we exclude the InputClass arguments.

damyanp (approved):

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2312,6 +2313,80 @@ TEST_F(CompilerTest, CompileWithResourceBindingFileThenOK) {
...
+
+  const WCHAR *OtherInputs[] = {
+    L"AnotherEntry",
+    L"AnotherEntry2",
+  };
+
+  const WCHAR *Args[] = {
+    OtherInputs[0], OtherInputs[1],
+    L"-Od",
+    L"-Zi",
```
dmpots, line 2346: Maybe add an `OtherInputs` after the real args as well?

dmpots (approved):

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2312,6 +2313,80 @@ TEST_F(CompilerTest, CompileWithResourceBindingFileThenOK) {
...
+  };
+  VERIFY_SUCCEEDED(pCompiler->Compile(pShaderBlob, EntryPoint.c_str(), EntryPoint.c_str(), L"ps_6_0", Args, _countof(Args), nullptr, 0, nullptr, &pOpResult));
```
dmpots, line 2348: Should we also have a test with the `IDxcCompiler3` api?

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
adam-yang: Done.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
adam-yang: Nope! That was left over from an earlier iteration. Fixed now.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
adam-yang: Done!

tools/clang/unittests/HLSL/CompilerTest.cpp
adam-yang: Done. Also added one inbetween them, as well as checking these flags show up in the final arg list.

tools/clang/unittests/HLSL/CompilerTest.cpp
adam-yang: I don't think there's too much point to that. IDxcCompiler3 gets called internally anyway, and the -T, -E, and input file name get added in predictable orders.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
damyanp: Reading this, I don't understand _why_ we're doing that.

damyanp (approved):

PR #5237, opened 2023-05-25
PIX Shader debugger: Fixes for structs that contain resources or member functions

jeffnn:
A certain ISV makes heavy use of both of these language features. PIX was stumbling over both.


pow2clk (approved):
Looks reasonable!

tex3d (approved):
LGTM.

PR #5239, opened 2023-05-25
[lit] avoid rm *.pdb

python3kgae:
Rename .pdb to .dbg and put tmp pdb to tmp directory.

This is to avoid random fail caused by check *.pdb and rm *.pdb in different test.

### DIFF
--- a/tools/clang/test/DXC/embed_dbg.test
+++ b/tools/clang/test/DXC/embed_dbg.test
@@ -2,7 +2,7 @@
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
 
-// RUN: ls -1 %T | FileCheck %s --check-prefix=PDB_NAME
+// RUN: ls -1 | FileCheck %s --check-prefix=PDB_NAME
 // PDB_NAME-NOT:.pdb{{$}}
 
 // Auto-embed debug info when no debug output, and expect warning signifying that this is the case.
--- a/tools/clang/test/DXC/embed_dbg.test
+++ b/tools/clang/test/DXC/embed_dbg.test
@@ -2,7 +2,8 @@
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
 
-// RUN: ls -1 %T | not grep "\\.pdb$"
+// RUN: ls -1 %T | FileCheck %s --check-prefix=PDB_NAME
+// PDB_NAME-NOT:.pdb{{$}}
 
 // Auto-embed debug info when no debug output, and expect warning signifying that this is the case.
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl  /Zi /Fo %t.embedpdb2 /Fe %t.err.embedpdb
--- a/tools/clang/test/DXC/fd_zs.test
+++ b/tools/clang/test/DXC/fd_zs.test
@@ -1,7 +1,7 @@
 // /Fd plus /Zs
-// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zs /Fd %t.pdb /Fo %t.dxo_fd
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zs /Fd %t.dbg /Fo %t.dxo_fd
 
-// RUN:FileCheck --input-file=%t.pdb %s --check-prefix=PDB && rm %t.pdb
+// RUN:FileCheck --input-file=%t.dbg %s --check-prefix=PDB
 // Make sure PDB not empty.
 // PDB:{{.+}}
 // RUN:FileCheck --input-file=%t.dxo_fd %s --check-prefix=FD_FO
--- a/tools/clang/test/DXC/ftime-trace.hlsl
+++ b/tools/clang/test/DXC/ftime-trace.hlsl
@@ -1,6 +1,5 @@
 // RUN: %dxc -E main -T vs_6_0 %s -ftime-trace | FileCheck %s
-// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace=%t.json
-// RUN: cat %t.json | FileCheck %s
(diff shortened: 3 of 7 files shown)


tools/clang/test/DXC/strip_dbg.test
```diff
@@ -1,7 +1,11 @@
 // /Fd implies /Qstrip_debug ; path with / produces auto hash-named .pdb file
-// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Fd ./ /Fo %t.strip
+
+// RUN: mkdir -p %T/pdbdir
```
tex3d, line 3: The issue with using `%T` instead of basing it off of `%t` is that it's not unique for this test.

tools/clang/test/DXC/embed_dbg.test
```diff
@@ -2,7 +2,8 @@
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
 
-// RUN: ls -1 %T | not grep "\\.pdb$"
+// RUN: ls -1 %T | FileCheck %s --check-prefix=PDB_NAME
```
tex3d, line 5: How is there any chance of it randomly writing a pdb into this particular folder?  Originally, the check was for the working directory, not the folder containing the test, wasn't it?

tools/clang/test/DXC/strip_dbg.test
python3kgae: Fixed.

tex3d (approved):
LGTM.

tools/clang/test/DXC/embed_dbg.test
```diff
@@ -2,7 +2,8 @@
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
 
-// RUN: ls -1 %T | not grep "\\.pdb$"
+// RUN: ls -1 | FileCheck %s --check-prefix=PDB_NAME
+// PDB_NAME-NOT:.pdb{{$}}
```
llvm-beanz, line 6: Is the point of this to get rid of the `grep` call? How are we going to handle tests from upstream that rely on `grep`?

tools/clang/test/DXC/ftime-trace.hlsl
```diff
@@ -1,6 +1,5 @@
 // RUN: %dxc -E main -T vs_6_0 %s -ftime-trace | FileCheck %s
-// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace=%t.json
-// RUN: cat %t.json | FileCheck %s
+// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace=%t.json && FileCheck %s --input-file=%t.json
```
llvm-beanz, line 2: This change is fine, but it also seems like you’re trying to remove a dependency on `cat`. I worry about this with upstream LLVM where tests are allowed to use `cat`

tools/clang/test/DXC/root_sig.test
```diff
@@ -26,8 +26,9 @@
 
 // Set root signature when already has one should succeed
 // RUN: %dxc %t.define.cso /dumpbin /setrootsignature %t.rootsig.cso /Fo %t.smoke.rsadded.cso
-// Make sure %t.smoke.rsadded.cso exist.
-// RUN: ls %t.smoke.rsadded.cso
+// Make sure %t.smoke.rsadded.cso not empty.
+// RUN:FileCheck --input-file=%t.define.cso %s --check-prefix=CSO
+// CSO:{{.+}}
```
llvm-beanz, line 31: This is probably fine for this change, but we should have a tool for inspecting the contents of the `cso` file so that we can ensure it is generated correctly.

tools/clang/test/DXC/fd_zs.test
```diff
@@ -1,7 +1,7 @@
 // /Fd plus /Zs
-// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zs /Fd %t.pdb /Fo %t.dxo_fd
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zs /Fd %t.dbg /Fo %t.dxo_fd
 
-// RUN:FileCheck --input-file=%t.pdb %s --check-prefix=PDB && rm %t.pdb
+// RUN:FileCheck --input-file=%t.dbg %s --check-prefix=PDB
```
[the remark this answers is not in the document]
python3kgae: This is to make sure no *.pdb in the directory so the test to test that can work.
If not do this, we have to rm the generated %t.pdb and still hit random fails when tests running at same time.

tools/clang/test/DXC/ftime-trace.hlsl
python3kgae: This is trying to unblock one branch pipeline temporarily.
Once we get updated VM which can run git usr bin, we can switch it back.

tools/clang/test/DXC/fd_zs.test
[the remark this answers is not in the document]
llvm-beanz: Ah, you're right. `%T` is shared across the directory.

I'm worried about the extent of issues we're having to work around in our tests to avoid problems with the internal build environment. I think we need to figure out a better solution.

tools/clang/test/DXC/embed_dbg.test
pow2clk: I think it bears adding a comment explaining that

tools/clang/test/DXC/fd_zs.test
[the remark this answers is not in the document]
pow2clk: The name isn't exactly random though is it? Can we duplicate the hash that determines what the name is and check the expected name?

Or perhaps using a custom target like below that uses `%t_pdbdir`? Or did I misunderstand that?

tools/clang/test/DXC/ftime-trace.hlsl
pow2clk: I think a comment should explain that then and these changes should be mentioned in the description if not the title since it only refers to the pdb naming changes.

pow2clk (commented):
I think we should mention that the main thrust of these changes is to avoid dependencies that are currently unavailable in certain testing environments in addition to the pdb conflict issue.

tools/clang/test/DXC/fd_zs.test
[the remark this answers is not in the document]
python3kgae: Created #5254 to check it by filename.

PR #5246, opened 2023-05-29
Add new script to update external dependencies for the SPIRV backend.

dnovillo:
This simplifies the updating of the external dependencies for the SPIRV backend.

Tested on Linux.

### DIFF
--- a/utils/update_spirv_deps.sh
+++ b/utils/update_spirv_deps.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+if [ ! -d external -o ! -d .git -o ! -d azure-pipelines ] ; then
+  echo "Run this script on the top-level directory of the DXC repository."
+  exit 1
+fi
+
+set -ex
+git submodule foreach 'set -x && git switch main && git pull --ff-only'
+git add external
+exit 0


Keenuts (approved):

llvm-beanz (approved):

s-perron (changes requested):
Let's wait until we know what we want to do with the deps. RE2 is now using abseil. We need to decide what we want to do about that.  My feeling is to no longer update effcee and RE2.

dnovillo:
> Let's wait until we know what we want to do with the deps. RE2 is now using abseil. We need to decide what we want to do about that. My feeling is to no longer update effcee and RE2.

Oh, sorry!  You had mentioned that and I completely spaced out.  I'll update the script to ignore effcee/re2.

llvm-beanz:
I think we're also getting close to being able to remove the effcee & re2 dependencies by moving the SPIR-V tests to FileCheck.

dnovillo:
> I think we're also getting close to being able to remove the effcee & re2 dependencies by moving the SPIR-V tests to FileCheck.

That would be fantastic.  I recently had to debug a failing test case.  I do not get along well with the current setup 😉

Issue #2419, opened 2019-08-20
DXC exhibits different behavior than FXC when encountering uninitialized variables


vcsharma:
**Scenario 1**: Uninitialized variable
DXC: Below shader compiles fine without any warning.
FXC: Compilation fails with an error message `error X4000: variable 'sum' used without having been completely initialized`

```
uint main() : OUT
{
	uint sum;
        sum++;
	return sum;
}
```

**Scenario 2**: Uninitialized variable passed to another function.
DXC: Below shader compiles fine without any warning.
FXC: Compilation fails with an error message `error X4000: variable 'sum' used without having been completely initialized`

```
uint foo(uint s)
{
	s = s << 4;
	return s;
}

uint main(uint i : IN) : OUT
{
	uint sum;
	return foo(sum);
}
```

**Scenario 3**: Uninitialized variable used in its own initialization
DXC: Below shader compiles fine without any warning.
FXC: Compiles fine with a warning message `warning X3578: Output value 'main' is not completely initialized`

```
uint main() : OUT
{
	uint sum = sum;
	return sum;
}
```

**Scenario 4**: Returning struct with uninitialized fields

DXC: Below shader compiles fine without any warning.
FXC: Compiles fine with a warning message `warning X3578: Output value 'main' is not completely initialized`

```
struct MyStruct {
 float a;
 int b;
};


MyStruct main() : OUT
{
	MyStruct val;
	return val;
}
```

**Scenario 5**: Conditional initialization
Both DXC and FXC compiles below shader without any warning. 

```
uint main(uint i : IN) : OUT
{
	uint sum;
	if(i > 10) {
		sum = 10;
	}
	return sum;
} 
```

vcsharma:
Reopening this issue as the fix for this would be needed along with fix for #2494.

Dredhog:
@vcsharma  We've also noticed this at Unity, do you plan on fixing this anytime soon? Having warnings for uninitialized outputs would be really nice to have.

damyanp:
No plans to address this in DXC, but this is something we're aware of in clang.

PR #5249, opened 2023-05-30
[Draft] Rewrite output parameters

### LINKED ISSUE #5377: `out` and `inout` should always be references
**Description**
All `out` and `inout` parameters should use reference types as their qual types in the AST.

DXC's ASTs currently model some types as pass-by-value even when they are `out` or `inout`, which makes the ASTs inaccurately reflect the program. In order to address issues with AST-based error and warning analysis and implement a correct calling convention for DXIL libraries we need to produce accurate ASTs.

**Steps to Reproduce**
[Example shader](https://godbolt.org/z/sasd86d5P)

See the produced AST:

```
TranslationUnitDecl 0x55e2317384a0 <<invalid sloc>> <invalid sloc>
|-CXXRecordDecl 0x55e2317668c0 <<source>:1:1, line:3:1> line:1:8 referenced struct SomeStruct definition
| |-CXXRecordDecl 0x55e231766a00 <col:1, col:8> col:8 implicit struct SomeStruct
| `-FieldDecl 0x55e231766ac8 <line:2:3, col:7> col:7 SomeVal 'int'
|-FunctionDecl 0x55e2317741b0 <line:5:1, line:6:1> line:5:6 used Fn 'void (__restrict SomeStruct)'
| |-ParmVarDecl 0x55e231766b30 <col:9, col:26> col:26 S '__restrict SomeStruct'
| | `-HLSLInOutAttr 0x55e231766ba0 <col:9>
| `-CompoundStmt 0x55e2317742c8 <col:29, line:6:1>
`-FunctionDecl 0x55e231774300 <line:8:1, line:11:1> line:8:6 Fn2 'void ()'
  `-CompoundStmt 0x55e231774620 <col:12, line:11:1>
    |-DeclStmt 0x55e231774450 <line:9:3, col:15>
    | `-VarDecl 0x55e2317743e0 <col:3, col:14> col:14 used S 'SomeStruct'
    `-CallExpr 0x55e2317745c0 <line:10:3, col:7> 'void'
      |-ImplicitCastExpr 0x55e2317745a8 <col:3> 'void (*)(__restrict SomeStruct)' <FunctionToPointerDecay>
      | `-DeclRefExpr 0x55e231774528 <col:3> 'void (__restrict SomeStruct)' lvalue Function 0x55e2317741b0 'Fn' 'void (__restrict SomeStruct)'
      `-ImplicitCastExpr 0x55e231774608 <col:6> 'SomeStruct' <LValueToRValue>
        `-DeclRefExpr 0x55e2317744c0 <col:6> 'SomeStruct' lvalue Var 0x55e2317743e0 'S' 'SomeStruct'
```

`FunctionDecl 0x55e2317741b0 <line:5:1, line:6:1> line:5:6 used Fn 'void (__restrict SomeStruct)'` should have a reference parameter.

The argument copy semantics are also not represented in the AST, which will need to be fixed to address this.

**Environment**
- DXC version: All versions of DXC
- Host Operating System: All OS environments

llvm-beanz:
This change is a huge swath of changes rewriting how HLSL represents pointers, references and arrays as well as the handling of output parameters in function signatures.

The main changes included in this PR are:
* Restoring Array->Pointer Decay in the HLSL ASTs.
* Representing all output parameters using reference types in the AST.
* Adding a new HLSLOutParamExpr AST node for output parameters. *Adding a new HLSLArrayTemporaryExpr to represent temporary array values.

There are also a large number of associated fixes and patches to make the code generation and optimization code resilient to the new AST and IR structures that are generated.

Disclaimer: This is a _massive_ PR with no good way to break it up. In the current state there are still some test failures that I need to resolve, but I want to get this in front of people to review before it gets any bigger.

Fixes #5377


llvm-beanz:
For more information on the changes contained here see: [Revising HLSL `out` Parameters.pdf](https://github.com/microsoft/DirectXShaderCompiler/files/11602657/Revising.HLSL.out.Parameters.pdf)

s-perron (commented):
I took a look at the SPIR-V changes, and nothing looks like it should be a problem.

s-perron:
Some of the failing tests are real problems that need to be fixed:

- [ ] FileTest.BinaryOpAssignOpaqueArray - Some optimizations are not happening. They are probably unaware of OpCopyMemory. Spirv-opt will have to be improved. This has to be fixed before merging.
- [ ] FileTest.FunctionInOutParamLocalResource - The is an error related to the counter variables. I don't know if this is intended, but I expect it is not. I'm not sure what needs to be done in this case.
- [ ] FileTest.FunctionInOutParamIsomorphism - The optimization is not as good as it could be. Not necessary to fix, but would be nice.

I still have to look at ~30 more failing test cases.

s-perron:
- [ ] FileTest.TextureGetDimensions - I cannot tell from the test if this is a problem or not. In some cases where the output variable is not used, the temp does not get copied out to the original variable. This is one example of many. We should check what happens when the value it used to make sure the code is okay.
- [ ] FileTest.ByteAddressBufferTemplatedLoadMatrix - The spirv is incorrect. The OpCopyMemory is trying to copy a value that has already been loaded. Affects other test cases as well.
- [ ] FileTest.SpirvInterpolationVS - I think the test case is invalid, and needs to be updated. This is more than just updating the expected outputs.
- [ ] FileTest.MeshShadingEXTAmplification - There is a new error message, but the error message seems wrong. It says that `expected groupshared object as argument to DispatchMesh()`, but the argument that it points to is groupshared.
- [ ] FileTest.ReduceLoadSize - OpCopyMemory is trying to copy between the objects of different type. The types are different because of their decorations. Before spir-v 1.4 you have to do what the old code does: extract individual part, and reconstruct them. In spir-v 1.4, we can load, [OpCopyLogical](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpCopyLogical), and store.

llvm-beanz:
Related to #2419

dnovillo:
@llvm-beanz are you still interested in this? It's quite old and still marked as draft.

PR #5251, opened 2023-05-30
Another case of non-determinism found in SROA

bfavela:
Internal test (much larger shader) when doing a large number of iterations of compiles. This compare in SROA appears to be another source of the differences which matches the previous non-determinism fix that was cherry-picked from upstream (https://github.com/microsoft/DirectXShaderCompiler/commit/8330655de7250fb0f8f479c6e208b3f262f7eff5)

There seems to be a fundamental issue with comparing names of globals in that the pointer casts need to be stripped from them in order to get the comparisons to evaluate exactly the same way each time. There may be a way of finding other instances of this issue in the rest of the code, but I'm unsure what that is.


bfavela:
Note: the test doesn't actually work. And I'm not sure how to fix it to exacerbate this issue?

Issue #4704, opened 2022-10-04
[SPIR-V] SubpassInput error when compiling non-pixel shader function in same source file that doesn't reference it at all


jmccarthy634:
As per title, if I have a HLSL file with a SubpassInput which declares a number of different functions, some of which are not pixel shaders, and I try and compile one of the non-pixel shader functions, DXC gives me an error, eg compiling the following with the command `dxc.exe -T cs_6_0 -E CsTest -spirv SubpassTest.hlsl`:

```
[[vk::input_attachment_index (0)]] SubpassInput<float4> subInput;

float4 PsSubpassTest() : SV_TARGET
{
	return subInput.SubpassLoad();
}

Buffer<uint> src : register(t0);
RWBuffer<uint> dst : register(u0);

[numthreads (8,1,1)]
void CsTest(uint id : SV_DispatchThreadID)
{
    dst[id] = src[id];
}
```
give this error:
```
SubpassTest.hlsl:1:57: error: SubpassInput(MS) only allowed in pixel shader
[[vk::input_attachment_index (0)]] SubpassInput<float4> subInput;
```
Ideally this would only happen if the SubpassInput is referenced in the function that is getting compiled, which it clearly isn't here.

PR #5252, opened 2023-05-30
Move the check for Subpass into to uses 

s-perron:
We currently check if the shader is a pixel shader when we see a
declaration of a SubpassInput type. However, it is possible that there
are multiple shader in the same hlsl file, and that the SubpassInput is
only used in the pixel shader.

So that we do not get an error in this case, I moved the check for the
shader type to the use of the variable. I only has a single member
function, so identifying uses of that member function is a good proxy.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/4704

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/vk.subpass-input.unused.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/vk.subpass-input.unused.hlsl
@@ -1,0 +1,23 @@
+// Test that the we can correctly compile the compute shader when the
+// subpass input is used in  PsSubpassTest, but not in the compute shader.
+
+// RUN: %dxc -T cs_6_0 -E CsTest
+
+// The variable is declared, but should not be used.
+// CHECK: %subInput = OpVariable
+// CHECK-NOT: %subInput
+
+[[vk::input_attachment_index (0)]] SubpassInput<float4> subInput;
+
+float4 PsSubpassTest() : SV_TARGET
+{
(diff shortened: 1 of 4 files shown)


tools/clang/test/CodeGenSPIRV/vk.subpass-input.unused.hlsl
```diff
@@ -0,0 +1,23 @@
+// Test that the we can correctly compile the compute shader when the
+// subpass input is used in  PsSubpassTest, but not in the compute shader.
+
+// RUN: %dxc -T cs_6_0 -E CsTest
+
+// The variable is declared, but should not be used.
+// CHECK: %subInput = OpVariable
```
Keenuts, line 7: Are we sure this wouldn't break some drivers?
Seems like this would go against the SPIR-V spec, as it states of OpTypeImage has SubpassData, the execution type must be fragment.
This should get DCOd, so we must request the legalization pass no? (And have a test making sure we don't emit the variable)

Keenuts (changes requested):

tools/clang/test/CodeGenSPIRV/vk.subpass-input.unused.hlsl
s-perron: Fixed. I just avoided added the declaration.

Keenuts (approved):

dnovillo (approved):

PR #5253, opened 2023-05-30
Revert "[DxbcConverter] Fix corruption of ICB integer values (#4790)"

hekota:
This reverts commit 0a1f7a19f4324642bde0a92bf49ecd41011e60a7:

Use an array of 32-bit integer values instead of 32-bit float values for the `"dx.icb"` immediate constant buffer data.
This prevents the FPU flipping the 22nd bit of integer constants, when their bit pattern represents a float32 signalling NaN on x86 architecture.

This change is causing rendering issues on PCs with AMD GPU and is blocking an upcoming Windows release.

The target branch of this revert is a release branch based on `release-1.7.2212` at commit `83f8c6c5` and it will be used for the upcoming Windows release. The dxilconv change stays in `main`, at least for now.


hekota:
FYI @ben-clayton

jenatali (approved):
Perhaps the title of the PR should call out that this is targeting a branch that isn't `main`?

llvm-beanz (approved):

tex3d (approved):

ben-clayton:
> The dxilconv change stays in `main`, at least for now.

I'd be concerned if this got pulled from `main` as the silent corruption of integer values was a nightmare to track down and with no simple workaround. I'd be happy to discuss ideas that would allow backwards compatibility with these drivers while keeping the logic in `main` corruption-free (obviously not before you unblock your release).

PR #5254, opened 2023-05-31
[lit] Check no output pdb by fixed file name.

python3kgae:
Avoid check *.pdb by check fixed output pdb name.
This will fix the random failure caused by entangled with pdb output of other tests.

The output filename should be stable because the shader is simple.

### DIFF
--- a/tools/clang/test/DXC/embed_dbg.test
+++ b/tools/clang/test/DXC/embed_dbg.test
@@ -2,7 +2,13 @@
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
 
-// RUN: ls -1 %T | not grep "\\.pdb$"
+// NOTE: dae4b0d5020646ac43cd4e4924f67eca.pdb is the output filename if exist.
+// Make sure the pdb file not exist with /Zi /Qembed_debug.
+// RUN: not ls dae4b0d5020646ac43cd4e4924f67eca.pdb
+
+// Make sure dae4b0d5020646ac43cd4e4924f67eca.pdb is the pdb filename.
+// RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fd %T/ /Fo %t.embedpdb
+// RUN: ls %T/dae4b0d5020646ac43cd4e4924f67eca.pdb && rm %T/dae4b0d5020646ac43cd4e4924f67eca.pdb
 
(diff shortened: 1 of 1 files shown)


tools/clang/test/DXC/embed_dbg.test
```diff
@@ -2,7 +2,14 @@
 // RUN: %dxc /T ps_6_0 %S/Inputs/smoke.hlsl /Zi /Qembed_debug /Fo %t.embedpdb
 // RUN: %dxc -dumpbin %t.embedpdb | FileCheck  %s --check-prefix=EMBED_PDB
 
-// RUN: ls -1 %T | not grep "\\.pdb$"
+// NOTE: dae4b0d5020646ac43cd4e4924f67eca.pdb is the output filename if exist.
```
llvm-beanz, line 5: What is this a hash of?

python3kgae: It is based on output bc.
If we change the order of the instructions, the hash will change.

llvm-beanz (commented):
IIUC, the issue with this test is that if LIT parallelizes this test with another test, and that test happens to write a PDB to the same directory... things fail. Correct?

If that's the case, why not just `mkdir %T-embed_dbg-unique && cd %T-embed_dbg-unique` then all the output goes into a unique subdirectory.

That removes any hard dependency on the hash value.

python3kgae:
> IIUC, the issue with this test is that if LIT parallelizes this test with another test, and that test happens to write a PDB to the same directory... things fail. Correct?
> 
> If that's the case, why not just `mkdir %T-embed_dbg-unique && cd %T-embed_dbg-unique` then all the output goes into a unique subdirectory.
> 
> That removes any hard dependency on the hash value.

Good idea.
Updated.

hekota:
Is this going to leave the repo dirty with untracked files and folders?

tools/clang/test/DXC/embed_dbg.test
```diff
@@ -1,8 +1,12 @@
 // Embed debug info
+
+// NOTE: cd into unique directory to avoid ls *.pdb affected by pdb files generated by other tests.
+// RUN: mkdir -p %T/embed_dbg && cd %T/embed_dbg
```
tex3d, line 4: Are we sure the `cd %T/embed_dbg` works for subsequent commands in this test and doesn't change CWD for any other tests?  And that this works correctly across platforms?

python3kgae:
> Is this going to leave the repo dirty with untracked files and folders?

No. %T is in temp directory.

tools/clang/test/DXC/embed_dbg.test
python3kgae: Each test has separate ShellEnvironment.
cd will not execute the shell command, just update the ShellEnvironment for current test.
So it should work for all platforms.

hekota (approved):
LGTM!

PR #5259, opened 2023-06-01
Fix debug info nondeterminism in SROAGlobalsAndAllocas from staticGVs

### LINKED ISSUE #5196: Stale pointer issue in SROAGlobalAndAllocas leading to indeterministic debug info
### Functional impact
The debug info (/Zi) output of compiling certain shaders is not deterministic (i.e. given identical inputs, the outputs are not always identical). Our content build pipeline relies on determinism for various things (caching etc.) so this is a pretty big problem for us. Having debugged the issue it is clear that certain processing is erroneously skipped for certain global variables, so this may be the source of correctness issues as well, but I have not been able to confirm that.

### Analysis

I have debugged this issue at length, and I believe it is caused by use of stale pointers inside `SROAGlobalAndAllocas`.

* At the start of the function pointers to all of the module's static globals (`M.globals()`) are added to to the `staticGVs` set.
* During the course of the function, globals may be removed from the module and freed (`eraseFromParent`), leaving stale pointers in `staticGVs`.
* The function can also allocate new globals. If it is "unlucky", the allocator will return the exact same pointer as one of the globals that was freed.
* When it comes to process this new global, it first checks if it's in `staticGVs`. It should not be, because the set is only filled at the start with the initial set, so should not contain the newly-allocated variable. But if you were "unlucky" it will, since it happens to have the same pointer as one of the stale entries still in the set. When this happens, the new variable is erroneously skipped from further processing.
* The end result is that this global does not have debug data added for it (like I said above, there may be other effects as well). This affects the final pdb data, and since it depends on the allocator reusing an address, it doesn't happen consistently, hence the indeterminism.

### Proposed Fix

The issue can be resolved by removing entries from the `staticGVs` set before freeing them. From what I can tell this is done in just two places inside the function. So I propose adding a `staticGVs.remove(GV);` call before the 2 `GV->eraseFromParent();` calls inside `SROAGlobalAndAllocas`.

### Repro

1. Download the attached HLSL file and powershell script.
2. Remove .txt extension (github didn't like hlsl/ps1 files).
3. Update the path to dxc.exe in the script.
4. Run the script from the same directory as the HLSL file.

Files:
[NonDetermisticDebugData.hlsl.txt](https://github.com/microsoft/DirectXShaderCompiler/files/11431544/NonDetermisticDebugData.hlsl.txt)
[Repro.ps1.txt](https://github.com/microsoft/DirectXShaderCompiler/files/11431546/Repro.ps1.txt)

The script is very basic/unsophisticated, but was useful for me when debugging this. It compiles the shader 100 times, and compares the hash of the pdb to the that from the first run. If it's different it prints out a message.

Note that due to the nature of this bug I found it to be extremely sensitive to seemingly unrelated factors -- even the path of the directory I put the files in seemed to affect it. I found that generally it would be "different" about 5-10% of the time, but sometimes it would be as low as 1-2% or as high as 50%, seemingly for no reason. Your millage may vary :).

I tried this with various version of DXC, including the latest release (v1.7.2212.1) and the current head (2dd8344), and all seemed to demonstrate the same problem.

tex3d:
staticGVs is used to track which globals are static and don't need new debug info added at a certain point. Some global values are erased during iteration of the work list, including globals that could be in this set. This could lead to stale pointers in the set, which could alias new global values from SROA, nondeterministically causing missing debug info.

This change removes globals from the set (if present) in each path where they are deleted. It also skips adding globals to the set (and MergeGepUse) that have no users in the first place, since these will just be deleted and have to be removed from the set later.

Fixes #5196.


llvm-beanz (commented):
[Testcase?](https://reviews.llvm.org/file/data/4hypousprhy32dtdm4p3/PHID-FILE-jz27ysqevr3no2mavhs3/test_case)
image

bfavela:
> [Testcase?](https://reviews.llvm.org/file/data/4hypousprhy32dtdm4p3/PHID-FILE-jz27ysqevr3no2mavhs3/test_case) image

I've been trying to create one, but am having difficulty getting the right static GVs into SROA. They seem to get optimized away earlier. Any advice? I tried writing a direct pass testcase, but the GVs didn't seem to get reordered (targeting my other change, but should be able to extend to this one as well)

simontaylor81:
Is there any chance we could get this change in without a test case, since it's proving problematic? We're still waiting on the fix to resolve our issue, and people are starting to pester me about it :). Seems unfortunate that a fix is available but we can't use it due to a missing test.

bfavela:
The concern is moreover that we are band-aiding the issue and we don't have a good understanding of what the actual root cause is if we can't actually reliably test it. We're attempting to tackle the problem with other techniques.
I'll leave it up to @llvm-beanz and @tex3d if they want to check it in without the testcase

simontaylor81:
I provided a reproduction case with the original issue, along with a detailed explanation of the root cause. I'm not surprised you can't reproduce it in a unit test, given how reliant it is on a specific result from the allocator. That's not really a mystery, nor does it add any mystery to the underlying issue -- sometimes an issue is just very hard to replicate :).

llvm-beanz:
At a minimum we need to introduce a test case that experiences the non-deterministic behavior (even if it doesn't deterministically reproduce it), and we need to have that test verify the expected correct output. Having a test that does that will at a minimum introduce a test case that fails non-deterministically, if this fix doesn't address the root cause of the non-determinism.

I'm suspect of the theory that the memory allocator itself is non-deterministic. Generally OS page allocators are non-deterministic in the virtual memory address layout of pages, but usually sub-page allocation patterns are deterministic, so if a memory block is getting re-used, it should always get re-used if the program's allocation patterns are deterministic.

The big catch there is that last bit. DXC's allocation patterns _should_ be deterministic, but if they're not (and they might not be) that would be the root cause of the issue that we need to address, and this change potentially masks it.

simontaylor81:
> The big catch there is that last bit. DXC's allocation patterns should be deterministic, but if they're not (and they might not be) that would be the root cause of the issue that we need to address, and this change potentially masks it.

Apologies if I'm re-stating something already known, but I think there's a misunderstanding of the core issue here -- the _indeterminism_ is not the root issue, it is the symptom. The root issue is that there are stale pointers to freed objects in the set. These pointers could alias to newly-allocated object regardless of the determinism or lack thereof of the allocator. The goal is to ensure the behaviour of the code is correct, not ensure that it deterministically wrong :). This change does the former, it does not mask an underlying issue. Any change that made the allocator deterministic, on the other hand, _would_ be masking the underlying bug, i.e. the dangling pointers.

llvm-beanz:
> Apologies if I'm re-stating something already known, but I think there's a misunderstanding of the core issue here -- the _indeterminism_ is not the root issue, it is the symptom. The root issue is that there are stale pointers to freed objects in the set.

I think your perspective on this issue is not in line with mine. In general, any non-determinism in the output of a compiler is a bug. We know the symptom of this non-determinism (the output being non-deterministic due to stale pointers). We know a way to treat the symptom (removing stale pointers). We don't know the root cause of the non-determinism.

> These pointers could alias to newly-allocated object regardless of the determinism or lack thereof of the allocator. The goal is to ensure the behaviour of the code is correct, not ensure that it deterministically wrong :). This change does the former, it does not mask an underlying issue. Any change that made the allocator deterministic, on the other hand, _would_ be masking the underlying bug, i.e. the dangling pointers.

All of this is irrelevant at the moment. I'm not suggesting we not take this change. All I'm saying in this review is that this change as written does not meet our [contribution standards](https://github.com/microsoft/DirectXShaderCompiler/blob/main/CONTRIBUTING.md#contributing-code-and-content).

As soon as this change is updated to include tests, we can review it fully and proceed with getting it merged.

_We do not accept functional changes to the codebase that are not accompanied with tests_. This is even more crucial for a code change like this that is modifying a poorly tested part of our codebase.

bfavela (approved):
@llvm-beanz @tex3d Thanks for the tests, Tex! LGTM...

 I'll still work on the ReverseIterator stuff, but this should unblock @simontaylor81 hopefully?

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/nondet_debug_data_static_gv.ll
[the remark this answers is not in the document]
tex3d: There is a CHECK for each global and scalarized static global.  If the nondeterministic behavior occurs, one or more of these would be missing and the test should fail.  I'm not sure what additional checks you're asking for.

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/nondet_debug_data_static_gv.hlsl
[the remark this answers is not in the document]
tex3d: It does ensure the debug info doesn't get omitted or re-arranged in non-deterministic way end-to-end.  I guess you don't see any value in that, but I think that's a good check for the original scenario which could catch other potential regressions in an area with complexity beyond the pass tested in the IR test.  For instance, if some other change modified the ordering of debug metadata (unexpectedly - potential nondeterminism), or dropped some for any other reason, before or after that pass, this test would catch it, while the IR test wouldn't.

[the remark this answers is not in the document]
tex3d: Also, having the HLSL in a big comment block inside the IR test makes it much more of a pain to use it to update the IR test when necessary.

llvm-beanz (approved):

simontaylor81:
What is the status of this? Was it ever submitted?

llvm-beanz:
> What is the status of this? Was it ever submitted?

No, the provided test doesn't pass on all our supported configurations. I looked at it a little bit this morning, and the issue seems to be that we are relying on the order of unordered `std` containers (likely the source of some of our underlying non-determinism).

This presents itself as the tests failing on macOS, but they also fail on Linux or Windows if you use an alternate implementation of the C++ standard library (like libc++ which is default on macOS).

@tex3d is out of the office, so I'm not sure what the timeline for further investigation to resolve the issue is.

pow2clk:
Pushed a commit that should fix the MacOS errors. It is a refinement of the priority queue nondeterminism when sizes are the same that I started in #4476

pow2clk:
Sorry if I'm overstepping, but I think my addition should be reviewed so I disabled auto-merge

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -1714,10 +1714,12 @@ bool SROAGlobalAndAllocas(HLModule &HLM, bool bHasDbgInfo) {
...
       sz1 = getNestedLevelInStruct(a1ty);
     }
-    // If sizes are equal, tiebreak with alphabetically lesser at higher priority
-    return sz0 < sz1 || (sz0 == sz1 && isa<GlobalVariable>(a0) &&
-                         isa<GlobalVariable>(a1) &&
-                         a0->getName() > a1->getName());
+    // If sizes are equal, and the new value is a GV,
+    // replace the existing node if it isn't GV or comes later alphabetically
+    // Thus, entries are sorted by size, global variableness, and then name
+    return sz0 < sz1 || (sz0 == sz1 && isa<GlobalVariable>(a1) &&
```
llvm-beanz, line 1720: This is pretty hacky. Can we file a bug tagged as `tech-debt` to revisit this and have a better way to maintain stable ordering? Respecting source ordering would be much better because it is at least intuitive.

llvm-beanz:
Looking at the last change more closely, it is re-normalizing underlying non-determinism by adding additional criteria for tie-breaking the priority queue sorting. We wouldn't need this tie-breaking at all if the order elements are inserted into the priority queue is deterministic in the first place (and it should be since that's just based on the module globals list). That means somewhere inside this pass we're creating globals in a nondeterministic order, which is bad.

I'm okay with this change going in, but we need to file an issue to track identifying the root cause of the nondeterminism.

pow2clk:
I've spoken to @llvm-beanz about this offline. I've tested the sequence of GVs and allocas as they go into the priority queue on both macos and PC and the order is the same. Without this change to the comparison function, they come out in different orders when the sizes match. I don't think there's an underlying nondeterminism apart from how the ties are broken.

I won't pretend its a pretty conditional, but I don't know how else to have a sequential ordering using a single comparison function.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
bogner: It's also pretty hard to read. Can we split up the logic or use `std::tie` to simplify? If I'm reading this correctly it should be equivalent to something like the following:
```c++
    return std::tie(sz0, isa<GlobalVariable>(a0), a0->getName()) <
           std::tie(sz1, isa<GlobalVariable>(a1), a1->getName());
```
That is, first compare on size, then sort non-GVs before GVs, then compare the names.

bogner: Oh, we'd probably need a couple of variables to do this since `std::tie` needs lvalues, so more like this I guess, but you get the idea.
```c++
    bool A0IsGV = isa<GlobalVariable>(a0);
    bool A1IsGV = isa<GlobalVariable>(a1);
    StringRef A0Name = a0->getName();
    StringRef A1Name = a1->getName();
    return std::tie(sz0, A0IsGV, A0Name) < std::tie(sz1, A1IsGV, A1Name);
```

llvm-beanz:
Thank you @pow2clk for following up and pointing out that this is a very special behavior of `std::priority_queue`. I wrote up a [minimal test on compiler explorer](https://godbolt.org/z/TqxcM6GGP) that shows that the behavior of sorting values with the same priority is not consistent.

From my minimal poking it looks like libc++ is consistent on adding elements with the same priority after other elements with equal priority, but libstdc++ seems to be flipping the order of things if they're added in the second half of the set, which makes me think it is a binary insertion sort.

Since it seems that `std::priority_queue` implementations vary, we probably shouldn't be using it in a place where ordering matters. An alternate approach to provide consistent ordering would be to use an `llvm::SmallVector<std::pair<uint64_t,Value*> >` for the work list where the `uint64_t` is the size. After all the elements are added to the work list a `std::stable_sort` should give a consistent and implementation-independent ordering.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
tex3d: Couldn't you use `std::make_tuple()` instead?

tex3d:
I filed tech-dept issue #5701 for follow-up on `priority_queue` usage here.

Issue #5257, opened 2023-05-31
Bit field of enum type causes crash when generating DxilFieldAnnotation


tex3d:
The following test produces a crash during clang codegen.

```hlsl
// RUN: %dxc -T lib_6_6 -HV 2021 %s

enum SomeEnum { Val1 };

struct SomeStruct
{
    // For some reason a uint32_t starting element is needed to allow for conversion from literal 0 in Sema
    uint32_t m1 : 16;
    SomeEnum m3 : 3;
};

export int SomeFunc()
{
    SomeStruct s = (SomeStruct)0;
    s.m3 = Val1;
    return (int)s.m3;
}
```

This code in `CGHLSLMS.cpp` assumes the bitfield type is a `clang::BuiltinType`, which is not true when it's an enum.
https://github.com/microsoft/DirectXShaderCompiler/blob/19c32587bee40189e8805f6dc12d96adb3550e2e/tools/clang/lib/CodeGen/CGHLSLMS.cpp#L1012

tex3d:
@python3kgae note, this crash would have been introduced from this PR: #5142

Before it wouldn't have had reflection for the bitfields, but it also may not have crashed when using an enum in a bitfield.

pow2clk:
Fixed by #5260

PR #5260, opened 2023-06-01
[BitField] Fix crash on enum bit field.

python3kgae:
Support bit field on enum type.
Fixes 5257 https://github.com/microsoft/DirectXShaderCompiler/issues/5257

### DIFF
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -1009,7 +1009,13 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(DxilStructAnnotation *annota
         DxilFieldAnnotation bitfieldAnnotation;
 
         bitfieldAnnotation.SetBitFieldWidth(Field->getBitWidthValue(Context));
-        const BuiltinType *BTy = Field->getType()->getAs<BuiltinType>();
+        QualType FieldTy = Field->getType();
+        const BuiltinType *BTy = FieldTy->getAs<BuiltinType>();
+        if (!BTy) {
+          // Should be enum type.
+          EnumDecl *Decl = FieldTy->getAs<EnumType>()->getDecl();
+          BTy = Decl->getPromotionType()->getAs<BuiltinType>();
+        }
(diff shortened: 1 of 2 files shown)


tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1009,7 +1009,13 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(DxilStructAnnotation *annota
         DxilFieldAnnotation bitfieldAnnotation;
 
         bitfieldAnnotation.SetBitFieldWidth(Field->getBitWidthValue(Context));
-        const BuiltinType *BTy = Field->getType()->getAs<BuiltinType>();
+        QualType FieldTy = Field->getType();
+        const BuiltinType *BTy = FieldTy->getAs<BuiltinType>();
```
llvm-beanz, line 1013: Other than a `BuiltinType` or an `EnumType` are there other types that could slip in here? Is it possible to get an `AttributedType` here?

llvm-beanz, line 1013: Other than a `BuiltinType` or an `EnumType` are there other types that could slip in here? Is it possible to get an `AttributedType` here?

python3kgae: Cannot think of an example to make an AttributedType here.
Added getCanonicalType to take care AttributedType for the future.

tex3d (approved):
Looks ok to me.  It would be nice if there was a standard way/function to get the built-in integer type from anything legally usable for bit fields.

I also noticed that the type name (`D3D12_SHADER_TYPE_DESC::Name`) comes out as `dword` in reflection.  I wonder why that is.

llvm-beanz (approved):
LGTM

python3kgae:
> Looks ok to me. It would be nice if there was a standard way/function to get the built-in integer type from anything legally usable for bit fields.
> 
> I also noticed that the type name (`D3D12_SHADER_TYPE_DESC::Name`) comes out as `dword` in reflection. I wonder why that is.

Will create another PR to fix the 'dword'.

PR #5262, opened 2023-06-01
Remove some warnings from GCC when building DXC

bob80905:
GCC generates some warnings when building DXC in a WSL environment. This PR is meant to reduce some of the warnings, mainly targeting unused variables / functions, and fallthroughs in switch statements.

### DIFF
--- a/lib/Bitcode/Reader/BitcodeReader.cpp
+++ b/lib/Bitcode/Reader/BitcodeReader.cpp
@@ -420,6 +420,11 @@ std::error_code BitcodeReader::error(const Twine &Message) {
 }
 
 // HLSL Change: remove unused function
+#if 0
+std::error_code BitcodeReader::error(BitcodeError E) {
+  return ::error(DiagnosticHandler, make_error_code(E));
+}
+#endif
 
 static DiagnosticHandlerFunction getDiagHandler(DiagnosticHandlerFunction F,
                                                 LLVMContext &C) {
@@ -4816,7 +4821,8 @@ std::error_code BitcodeReader::findFunctionInStream(
 // GVMaterializer implementation
 //===----------------------------------------------------------------------===//
 
-// HLSL Change: remove unused function
+// HLSL Change: remove unused function 
+// void BitcodeReader::releaseBuffer() { Buffer.release(); }
 
 std::error_code BitcodeReader::materialize(GlobalValue *GV) {
   if (std::error_code EC = materializeMetadata())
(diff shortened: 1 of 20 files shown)


lib/Bitcode/Reader/BitcodeReader.cpp
```diff
@@ -4818,7 +4816,7 @@ std::error_code BitcodeReader::findFunctionInStream(
 // GVMaterializer implementation
 //===----------------------------------------------------------------------===//
 
-void BitcodeReader::releaseBuffer() { Buffer.release(); }
+// HLSL Change: remove unused function
```
pow2clk, line 4819: As above, just as easy to leave the line and comment it out

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -1456,15 +1455,7 @@ void VariableRegisters::PopulateAllocaMap_StructType(
...
 
-llvm::DILocation *VariableRegisters::GetVariableLocation() const
-{
-  const unsigned DefaultColumn = 1;
-  return llvm::DILocation::get(
-      m_B.getContext(),
-      m_Variable->getLine(),
-      DefaultColumn,
-      m_Variable->getScope());
-}
```
pow2clk, line 1467: Mild preference for if 0

tools/clang/lib/Parse/ParseExpr.cpp
```diff
@@ -454,16 +454,18 @@ Parser::ParseRHSOfBinaryExpression(ExprResult LHS, prec::Level MinPrec) {
...
                                  OpToken.getKind(), LHS.get(), RHS.get());
-      } else
+      } else {
         LHS = Actions.ActOnConditionalOp(OpToken.getLocation(), ColonLoc,
                                          LHS.get(), TernaryMiddle.get(),
                                          RHS.get());
-    } else
+      }
+    } else {
+      // HLSL Change Begin - Take care TernaryMiddle.
```
tex3d, line 463: I believe this comment belongs after the else block, before:
```cpp
    // HLSL Change Begin - Take care TernaryMiddle.
    Actions.CorrectDelayedTyposInExpr(TernaryMiddle);
```

Note: The original bracing style was trying to preserve the existing code, unchanged, outside the `// HLSL Change Begin`/`End` marks.  Since we are unlikely to ever integrate upstream changes back into this codebase on mass, this matters less now.

tex3d (approved):
Just minor comment issues.

pow2clk (approved):
Thanks for the responses!

PR #5264, opened 2023-06-02
[lit] Remove temp directory for strip_dbg test.

python3kgae:
This is for case the temp directory not remove and pdb hash changed. If that happens, there will be 2 pdb files in the temp dir which cause strip_dbg.test fail.


hekota (approved):

pow2clk (approved):
Looks good!

Issue #5255, opened 2023-05-31
Rewriter removed struct declaration which used in constant buffer.


MioMioReimu:
Hello,
the following shader rewrite with `dxr.exe -remove-unused-functions -remove-unused-globals -E vs_main`
```hlsl
struct InstanceDataStructType
{
	float4 data;
};

cbuffer InstanceData
{
	InstanceDataStructType mData[2];
};

struct InstanceDataStructTypeNotUsed
{
	float4 data;
};

cbuffer InstanceDataNotUsed
{
	InstanceDataStructType mDataNotUsed[2];
};

struct VS_INPUT
{
	float3 position : POSITION;    
    float4 texcoord0 : TEXCOORD0;
    float4 diffuse: COLOR0;
};

struct VS_OUTPUT
{
	float4 position : SV_Position;
	float4 uv : TEXCOORD0;
	float4 color : VC;
};   
 
VS_OUTPUT vs_main(VS_INPUT input, uint instanceID : SV_InstanceID)
{
	VS_OUTPUT output = (VS_OUTPUT)0;
	output.position += mData[0].data;

	return output;
}

float4 ps_main(VS_OUTPUT input) : SV_Target0 
{
    return input.color;
}
```
Output's all cbuffer declaration are still remaining, but its referenced struct definitions are removed.
```hlsl
cbuffer InstanceData {
  const InstanceDataStructType mData[2];
}
;
cbuffer InstanceDataNotUsed {
  const InstanceDataStructType mDataNotUsed[2];
}
;
struct VS_OUTPUT {
  float4 position : SV_Position;
  float4 uv : TEXCOORD0;
  float4 color : VC;
};
VS_OUTPUT vs_main(VS_INPUT input, uint instanceID : SV_InstanceID) {
  VS_OUTPUT output = (VS_OUTPUT)0;
  output.position += mData[0].data;
  return output;
}
```

Snowapril:
I also have this problem :(

MioMioReimu:
Thanks very much, it works.
Is it possible to remove cbuffer definition which all members are not used? if it possible, can you give me some advices to implement it or any schedule plan. thanks very much.

PR #5265, opened 2023-06-02
[Rewriter] Support struct type used by array.

python3kgae:
Save element type of array type when collect visited type.

Fixes 5255 https://github.com/microsoft/DirectXShaderCompiler/issues/5255


tools/clang/test/HLSLFileCheck/rewriter/struct_array.hlsl
```diff
@@ -0,0 +1,53 @@
+// RUN: %dxr -remove-unused-functions -remove-unused-globals -E vs_main %s | FileCheck %s
+
+// Make sure InstanceDataStructType not removed.
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/tools/libclang/dxcrewriteunused.cpp
```diff
@@ -301,6 +305,10 @@ class TypeVisitor : public RecursiveASTVisitor<TypeVisitor> {
     }
     return true;
   }
+  bool VisitArrayType(const ArrayType *AT) {
+    TraverseType(AT->getElementType());
+    return true;
+  }
```
pow2clk, line 311: The corresponding DataRecursiveASTVistor code is a bit hard to follow due to macro skullduggery, but I don't think this is ever called. The only use of TypeVisitor is https://github.com/microsoft/DirectXShaderCompiler/blob/c2233c6c0a528ecaf5d2d008684e08daf722d845/tools/clang/tools/libclang/dxcrewriteunused.cpp#L1462
The implementation of which is here
https://github.com/microsoft/DirectXShaderCompiler/blob/c2233c6c0a528ecaf5d2d008684e08daf722d845/tools/clang/include/clang/AST/DataRecursiveASTVisitor.h#L548
Which defines ABSTRACT_TYPE as nothing and for Arrays, this is the macro that Array uses:
https://github.com/microsoft/DirectXShaderCompiler/blob/c2233c6c0a528ecaf5d2d008684e08daf722d845/tools/clang/include/clang/AST/TypeNodes.def#L70

So I really don't see how this could be called. Am I missing something?

pow2clk (commented):
I have one major uncertainty about VisitArrayType, but the description could really be more informative. You might mention how the SaveTypeDecl is used for appropriate tag types, but how it previously ignored array types that contained tag types and how this passes through that information.

If the VisitArrayType function is not dead as I have concluded that it is, some description of how that works would be helpful too.

tools/clang/tools/libclang/dxcrewriteunused.cpp
python3kgae: VisitArrayType will be called here https://github.com/microsoft/DirectXShaderCompiler/blob/c2233c6c0a528ecaf5d2d008684e08daf722d845/tools/clang/include/clang/AST/DataRecursiveASTVisitor.h#L350

But the change is still useless because VisitRecordType will be invoked anyway with/without the change. :)

Removed the VisitArrayType.

damyanp:

damyanp:

PR #5266, opened 2023-06-02
[dndxc] Add option when dump ast.

python3kgae:
This will allow dump ast to add option like -HV 2021.

### DIFF
--- a/tools/clang/tools/dotnetc/EditorForm.cs
+++ b/tools/clang/tools/dotnetc/EditorForm.cs
@@ -624,7 +624,10 @@ namespace MainNs
                 {
                     try
                     {
-                        var result = compiler.Compile(source, fileName, fileVars.Entry, fileVars.Target, new string[] { "-ast-dump" }, 1, null, 0, library.CreateIncludeHandler());
+                        List<string> args = new List<string>();
+                        args.Add("-ast-dump");
+                        args.AddRange(tbOptions.Text.Split());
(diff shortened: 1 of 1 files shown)


hekota (approved):

tools/clang/tools/dotnetc/EditorForm.cs
```diff
@@ -624,7 +624,10 @@ private void CompileDocument()
                 {
                     try
                     {
-                        var result = compiler.Compile(source, fileName, fileVars.Entry, fileVars.Target, new string[] { "-ast-dump" }, 1, null, 0, library.CreateIncludeHandler());
+                        List<string> args = new List<string>();
+                        args.Add("-ast-dump");
+                        args.AddRange(tbOptions.Text.Split());
+                        var result = compiler.Compile(source, fileName, fileVars.Entry, fileVars.Target, args.ToArray(), args.Count, null, 0, library.CreateIncludeHandler());
```
llvm-beanz, line 630: nit: line wrapping? This seems like an excessively long line. I know it is C# not C++, but having more consistent formatting would be nice.

llvm-beanz (approved):

tools/clang/tools/dotnetc/EditorForm.cs
python3kgae: Did it manually.
Ctrl + K, Ctrl + F not work :(

PR #5277, opened 2023-06-08
Fix variable definition checks in hcttest.cmd

tex3d:
Replaced all the `if "%var%"==""` with `if [not] defined var`, which is the proper way to determine whether the variable is defined or empty. The other way breaks if statement parsing under various conditions.

This fixes handling of TAEF's `/p:"var=val"` parameter syntax as long as you specify your additional TAEF parameters after a `--` argument break.


python3kgae (approved):

dmpots (approved):
LGTM

PR #5282, opened 2023-06-12
[Metadata] Limit bitfield annotation to validator version >= 1.7

python3kgae:
Only emit bitfield annotation for validator version >= 1.7
This is to avoid validation error on old validator.


hekota (approved):
LGTM!

coopp (approved):

PR #5283, opened 2023-06-12
[flag] Update the lifetime markers flag to correctly handle multiple …

dmpots:
…instances

The current behavior of the lifetime markers flags is to allow the `-disable-lifetime-markers` flag to override the enable flag no matter where it occurs. This means that if you write

    dxc -disable-lifetime-markers ... -enable-lifetime-markers

the lifetime markers will be disabled. This works differently from most clang flags which allow the last version of the enable/disable flag to win.

This PR modifies these flags so that when there are multiple occurences of enable/disable the last one will win. It is not clear if the original behavior was intentional or just a  mis-understanding of the flag API. There were no tests enforcing the original behavior.


adam-yang (approved):

llvm-beanz (approved):
LGTM

PR #5296, opened 2023-06-15
Add macro to disable usage of allocator overrides

amaiorano:
Add macro DXC_DISABLE_ALLOCATOR_OVERRIDES which, if defined, will use
the default CRT allocator for new/delete and malloc/realloc/free.

We need this on Chrome for our GN build of DXC, because our component
builds use libc++as a shared library, and we end up with mismatched
new/delete calls between libc++ and these custom overrides.

### DIFF
--- a/include/dxc/CMakeLists.txt
+++ b/include/dxc/CMakeLists.txt
@@ -10,6 +10,11 @@ configure_file(
   ${LLVM_INCLUDE_DIR}/dxc/Test/TestConfig.h
   )
 
+configure_file(
+  ${CMAKE_CURRENT_SOURCE_DIR}/config.h.cmake
+  ${LLVM_INCLUDE_DIR}/dxc/config.h
+  )
+
 add_subdirectory(DXIL)
 add_subdirectory(DxilContainer)
 add_subdirectory(HLSL)
@@ -17,6 +22,7 @@ add_subdirectory(Support)
 add_subdirectory(Tracing)
 
 set(DXC_PUBLIC_HEADERS
+    ${CMAKE_CURRENT_BINARY_DIR}/config.h
     ${CMAKE_CURRENT_SOURCE_DIR}/dxcapi.h
     ${CMAKE_CURRENT_SOURCE_DIR}/dxcerrors.h
     ${CMAKE_CURRENT_SOURCE_DIR}/dxcisense.h)
--- a/include/dxc/Support/WinIncludes.h
+++ b/include/dxc/Support/WinIncludes.h
@@ -47,6 +47,8 @@
 #include <intsafe.h>
 #include <ObjIdl.h>
 
+#include "dxc/config.h"
+
 // Support older atlbase.h if needed
 #ifndef _ATL_DECLSPEC_ALLOCATOR
 #define _ATL_DECLSPEC_ALLOCATOR
@@ -112,3 +114,26 @@ inline ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) { return (ENUMTYPE &)(((_
 #endif
 
 #endif // _MSC_VER
+
+/// DxcCoGetMalloc
+#if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
+
+#define DxcCoGetMalloc CoGetMalloc
+
+#else // defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
+
+#ifndef _WIN32
+CROSS_PLATFORM_UUIDOF(IMalloc, "00000002-0000-0000-C000-000000000046")
+struct IMalloc : public IUnknown {
+  virtual void *Alloc(SIZE_T size) = 0;
+  virtual void *Realloc(void *ptr, SIZE_T size) = 0;
+  virtual void Free(void *ptr) = 0;
+  virtual SIZE_T GetSize(void *pv) = 0;
+  virtual int DidAlloc(void *pv) = 0;
+  virtual void HeapMinimize(void) = 0;
+};
+#endif
+
+HRESULT DxcCoGetMalloc(DWORD dwMemContext, IMalloc **ppMalloc);
+
+#endif // defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
--- a/lib/DxcSupport/CMakeLists.txt
+++ b/lib/DxcSupport/CMakeLists.txt
@@ -8,6 +8,7 @@ add_llvm_library(LLVMDxcSupport
   HLSLOptions.cpp
   Unicode.cpp
   WinAdapter.cpp
+  WinIncludes.cpp
   WinFunctions.cpp
   )
 
@@ -17,4 +18,6 @@ ${LLVM_MAIN_SRC_DIR}/lib/DxcSupport/SharedLibAffix.inc
 ${LLVM_INCLUDE_DIR}/dxc/Support/SharedLibAffix.h
 )
 
-add_dependencies(LLVMDxcSupport TablegenHLSLOptions)
+target_link_libraries(LLVMDxcSupport PUBLIC LLVMSupport)
+
+add_dependencies(LLVMDxcSupport LLVMSupport TablegenHLSLOptions)
--- a/lib/DxcSupport/WinIncludes.cpp
+++ b/lib/DxcSupport/WinIncludes.cpp
@@ -10,14 +10,14 @@
 
 #include "dxc/Support/WinIncludes.h"
 
-#include "dxc/Support/microcom.h"
 #include "assert.h"
+#include "dxc/Support/microcom.h"
 
 #if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
 // CoGetMalloc from combaseapi.h is used
 #else
 struct DxcCoMalloc : public IMalloc {
-  DxcCoMalloc() : m_dwRef(0) {};
+  DxcCoMalloc() : m_dwRef(0){};
 
   DXC_MICROCOM_ADDREF_RELEASE_IMPL(m_dwRef)
   STDMETHODIMP QueryInterface(REFIID riid, void **ppvObject) override {
@@ -26,7 +26,9 @@ struct DxcCoMalloc : public IMalloc {
   }
 
   void *STDMETHODCALLTYPE Alloc(SIZE_T size) override { return malloc(size); }
-  void *STDMETHODCALLTYPE Realloc(void *ptr, SIZE_T size) override { return realloc(ptr, size); }
+  void *STDMETHODCALLTYPE Realloc(void *ptr, SIZE_T size) override {
+    return realloc(ptr, size);
+  }
   void STDMETHODCALLTYPE Free(void *ptr) override { free(ptr); }
   SIZE_T STDMETHODCALLTYPE GetSize(void *pv) override { return -1; }
   int STDMETHODCALLTYPE DidAlloc(void *pv) override { return -1; }
--- a/lib/DxcSupport/WinIncludes.cpp
+++ b/lib/DxcSupport/WinIncludes.cpp
@@ -1,0 +1,44 @@
+//===- WinIncludes.cpp -----------------------------------------*- C++ -*-===//
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// WinIncludes.cpp                                                           //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#include "dxc/Support/WinIncludes.h"
+
+#include "dxc/Support/microcom.h"
+#include "assert.h"
+
+#if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
+// CoGetMalloc from combaseapi.h is used
+#else
+struct DxcCoMalloc : public IMalloc {
+  DxcCoMalloc() : m_dwRef(0) {};
+
+  DXC_MICROCOM_ADDREF_RELEASE_IMPL(m_dwRef)
+  STDMETHODIMP QueryInterface(REFIID riid, void **ppvObject) override {
+    assert(false && "QueryInterface not implemented for DxcCoMalloc.");
+    return E_NOINTERFACE;
+  }
+
+  void *STDMETHODCALLTYPE Alloc(SIZE_T size) override { return malloc(size); }
+  void *STDMETHODCALLTYPE Realloc(void *ptr, SIZE_T size) override { return realloc(ptr, size); }
+  void STDMETHODCALLTYPE Free(void *ptr) override { free(ptr); }
+  SIZE_T STDMETHODCALLTYPE GetSize(void *pv) override { return -1; }
+  int STDMETHODCALLTYPE DidAlloc(void *pv) override { return -1; }
+  void STDMETHODCALLTYPE HeapMinimize(void) override {}
+
+private:
+  DXC_MICROCOM_REF_FIELD(m_dwRef)
+};
+
+HRESULT DxcCoGetMalloc(DWORD dwMemContext, IMalloc **ppMalloc) {
+  *ppMalloc = new DxcCoMalloc;
+  (*ppMalloc)->AddRef();
+  return S_OK;
+}
+#endif
--- a/tools/clang/tools/dxclib/dxc.cpp
+++ b/tools/clang/tools/dxclib/dxc.cpp
@@ -555,10 +555,8 @@ void DxcContext::ExtractRootSignature(IDxcBlob *pBlob, IDxcBlob **ppResult) {
   hlsl::DxilContainerHeader newHeader;
   uint32_t containerSize = hlsl::GetDxilContainerSizeFromParts(1, pPartHeader->PartSize);
   hlsl::InitDxilContainer(&newHeader, 1, containerSize); 
-  CComPtr<IMalloc> pMalloc;
   CComPtr<hlsl::AbstractMemoryStream> pMemoryStream;
-  IFT(CoGetMalloc(1, &pMalloc));
-  IFT(hlsl::CreateMemoryStream(pMalloc, &pMemoryStream));
+  IFT(hlsl::CreateMemoryStream(DxcGetThreadMallocNoRef(), &pMemoryStream));
   ULONG cbWritten;
 
   // Write Container Header
--- a/tools/clang/tools/dxcompiler/dxclinker.cpp
+++ b/tools/clang/tools/dxcompiler/dxclinker.cpp
@@ -128,11 +128,8 @@ DxcLinker::RegisterLibrary(_In_opt_ LPCWSTR pLibName, // Name of the library.
   try {
     std::unique_ptr<llvm::Module> pModule, pDebugModule;
 
-    CComPtr<IMalloc> pMalloc;
     CComPtr<AbstractMemoryStream> pDiagStream;
-
-    IFT(CoGetMalloc(1, &pMalloc));
-    IFT(CreateMemoryStream(pMalloc, &pDiagStream));
+    IFT(CreateMemoryStream(DxcGetThreadMallocNoRef(), &pDiagStream));
 
     raw_stream_ostream DiagStream(pDiagStream);
 
@@ -180,12 +177,10 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
 
   HRESULT hr = S_OK;
   try {
-    CComPtr<IMalloc> pMalloc;
     CComPtr<IDxcBlob> pOutputBlob;
     CComPtr<AbstractMemoryStream> pDiagStream;
 
-    IFT(CoGetMalloc(1, &pMalloc));
-    IFT(CreateMemoryStream(pMalloc, &pOutputStream));
+    IFT(CreateMemoryStream(DxcGetThreadMallocNoRef(), &pOutputStream));
 
     // Read and validate options.
     int argCountInt;
@@ -207,7 +202,7 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
 
     std::string warnings;
     //llvm::raw_string_ostream w(warnings);
-    IFT(CreateMemoryStream(pMalloc, &pDiagStream));
+    IFT(CreateMemoryStream(DxcGetThreadMallocNoRef(), &pDiagStream));
     raw_stream_ostream DiagStream(pDiagStream);
     llvm::DiagnosticPrinterRawOStream DiagPrinter(DiagStream);
     PrintDiagnosticContext DiagContext(DiagPrinter);
@@ -273,7 +268,7 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
         // Validation.
         HRESULT valHR = S_OK;
         dxcutil::AssembleInputs inputs(
-          std::move(pM), pOutputBlob, pMalloc, SerializeFlags,
+          std::move(pM), pOutputBlob, DxcGetThreadMallocNoRef(), SerializeFlags,
           pOutputStream,
           opts.DebugFile, &Diag);
         if (needsValidation) {
--- a/tools/clang/tools/dxcompiler/dxclinker.cpp
+++ b/tools/clang/tools/dxcompiler/dxclinker.cpp
@@ -268,9 +268,8 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
         // Validation.
         HRESULT valHR = S_OK;
         dxcutil::AssembleInputs inputs(
-          std::move(pM), pOutputBlob, DxcGetThreadMallocNoRef(), SerializeFlags,
-          pOutputStream,
-          opts.DebugFile, &Diag);
+            std::move(pM), pOutputBlob, DxcGetThreadMallocNoRef(),
+            SerializeFlags, pOutputStream, opts.DebugFile, &Diag);
         if (needsValidation) {
           valHR = dxcutil::ValidateAndAssembleToContainer(inputs);
         } else {
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -165,7 +165,6 @@ static void SavePixelsToFile(LPCVOID pPixels, DXGI_FORMAT format, UINT32 m_width
   CComPtr<IWICBitmapEncoder> pEncoder;
   CComPtr<IWICBitmapFrameEncode> pFrameEncode;
   CComPtr<IStream> pStream;
-  CComPtr<IMalloc> pMalloc;
 
   struct PF {
     DXGI_FORMAT Format;
@@ -182,7 +181,6 @@ static void SavePixelsToFile(LPCVOID pPixels, DXGI_FORMAT format, UINT32 m_width
 
   VERIFY_SUCCEEDED(ctx.Init());
   VERIFY_SUCCEEDED(CoCreateInstance(CLSID_WICImagingFactory, NULL, CLSCTX_INPROC_SERVER, IID_IWICImagingFactory, (LPVOID*)&pFactory));
-  VERIFY_SUCCEEDED(CoGetMalloc(1, &pMalloc));
   VERIFY_ARE_NOT_EQUAL(pFormat, Vals + _countof(Vals));
   VERIFY_SUCCEEDED(pFactory->CreateBitmapFromMemory(m_width, m_height, pFormat->PixelFormat, m_width * pFormat->PixelSize, m_width * m_height * pFormat->PixelSize, (BYTE *)pPixels, &pBitmap));
   VERIFY_SUCCEEDED(pFactory->CreateEncoder(GUID_ContainerFormatBmp, nullptr, &pEncoder));
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,9 @@ if ( HLSL_SUPPORT_QUERY_GIT_COMMIT_INFO )
   add_definitions(-DSUPPORT_QUERY_GIT_COMMIT_INFO)
 endif()
 
+option(DXC_DISABLE_ALLOCATOR_OVERRIDES "Disable usage of allocator overrides" OFF)
+mark_as_advanced(DXC_DISABLE_ALLOCATOR_OVERRIDES)
+
 # adjust link option to enable debugging from kernel mode; not compatible with incremental linking
 if(NOT CMAKE_VERSION VERSION_LESS "3.13" AND WIN32 AND NOT CMAKE_C_COMPILER_ARCHITECTURE_ID STREQUAL "ARM64EC")
   add_link_options(/DEBUGTYPE:CV,FIXUP,PDATA /INCREMENTAL:NO)
--- a/include/dxc/Support/WinFunctions.h
+++ b/include/dxc/Support/WinFunctions.h
@@ -31,7 +31,6 @@ int _wcsicmp(const wchar_t *str1, const wchar_t *str2);
 int _wcsnicmp(const wchar_t *str1, const wchar_t *str2, size_t n);
 int wsprintf(wchar_t *wcs, const wchar_t *fmt, ...);
 unsigned char _BitScanForward(unsigned long * Index, unsigned long Mask);
-HRESULT CoGetMalloc(DWORD dwMemContext, IMalloc **ppMalloc);
 
 HANDLE CreateFile2(_In_ LPCWSTR lpFileName, _In_ DWORD dwDesiredAccess,
                    _In_ DWORD dwShareMode, _In_ DWORD dwCreationDisposition,
--- a/include/dxc/Support/microcom.h
+++ b/include/dxc/Support/microcom.h
@@ -13,6 +13,7 @@
 #define __DXC_MICROCOM__
 
 #include <atomic>
+#include "dxc/Support/WinIncludes.h"
 #include "llvm/Support/Atomic.h"
 
 template <typename TIface>
--- a/include/dxc/Support/microcom.h
+++ b/include/dxc/Support/microcom.h
@@ -12,9 +12,9 @@
 #ifndef __DXC_MICROCOM__
 #define __DXC_MICROCOM__
 
-#include <atomic>
 #include "dxc/Support/WinIncludes.h"
 #include "llvm/Support/Atomic.h"
+#include <atomic>
 
 template <typename TIface>
 class CComInterfaceArray {
--- a/include/dxc/WinAdapter.h
+++ b/include/dxc/WinAdapter.h
@@ -642,16 +642,6 @@ struct IUnknown {
 CROSS_PLATFORM_UUIDOF(INoMarshal, "ECC8691B-C1DB-4DC0-855E-65F6C551AF49")
 struct INoMarshal : public IUnknown {};
 
-CROSS_PLATFORM_UUIDOF(IMalloc, "00000002-0000-0000-C000-000000000046")
-struct IMalloc : public IUnknown {
-  virtual void *Alloc(size_t size) = 0;
-  virtual void *Realloc(void *ptr, size_t size) = 0;
-  virtual void Free(void *ptr) = 0;
-  virtual size_t GetSize(void *pv) = 0;
-  virtual int DidAlloc(void *pv) = 0;
-  virtual void HeapMinimize(void) = 0;
-};
-
 CROSS_PLATFORM_UUIDOF(ISequentialStream, "0C733A30-2A1C-11CE-ADE5-00AA0044773D")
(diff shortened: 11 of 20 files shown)


llvm-beanz (commented):
I think this is probably generally useful, so we should expose it through CMake as well. It should be pretty simple to add a CMake option, mark it as advanced (`mark_as_advanced`), and have it pass the define to the two source files that need it using `set_source_file_properties`.

amaiorano:
The change I just pushed will fail the build. I need to add the new cpp file to the CMake build. I'm iterating on this based on a discussion with Tex.

amaiorano:
> I think this is probably generally useful, so we should expose it through CMake as well. It should be pretty simple to add a CMake option, mark it as advanced (`mark_as_advanced`), and have it pass the define to the two source files that need it using `set_source_file_properties`.

After talking with Tex, he made the valid point that disabling the custom new/delete operators alone was a bit of a hack, given that there are places where the current thread (IMalloc) or an explicit allocator is used to allocate memory. In other words, some allocations would be using the CRT, while others would make use of the IMalloc allocator. So I've updated the PR to make it so that when HLSL_USE_DEFAULT_NEW_DELETE is defined, all allocations will go through the CRT. As per Tex, we also fail if you attempt to use DxcCreateInstance2 with an IMalloc* that doesn't match the one returned by DxcGetThreadMallocNoRef().

I think this latest commit I pushed should build fine. It's a bit complex for me as I need to iterate in our GN-based project as well.

amaiorano:
Oh, I forgot to actually address your comment about exposing this in CMake. We could do, though now this macro is used in many more places, so it would need to be more globally defined. Would it be alright if exposing this flag were done in a follow-up change?

llvm-beanz:
> Oh, I forgot to actually address your comment about exposing this in CMake. We could do, though now this macro is used in many more places, so it would need to be more globally defined. Would it be alright if exposing this flag were done in a follow-up change?

I agree that globally defining it is likely the best approach given the expanded scope of the change. Because of the expanded scope, I think that adding it as a config.h.in define is the right way to do the change generally rather than relying on setting the flag on individual (or global) compiles.

I don't think that should be a separate change.

amaiorano:
> I agree that globally defining it is likely the best approach given the expanded scope of the change. Because of the expanded scope, I think that adding it as a config.h.in define is the right way to do the change generally rather than relying on setting the flag on individual (or global) compiles.

Can you walk me through the changes I would need to make? There are two `config.h.in` files, one for llvm and one for clang. Also, I'm not seeing any HLSL changes to these files, so am not sure what to model this after.

llvm-beanz:
> Can you walk me through the changes I would need to make? There are two `config.h.in` files, one for llvm and one for clang. Also, I'm not seeing any HLSL changes to these files, so am not sure what to model this after.

With your updated changes the scope of what we need to do here expands significantly. Your new change uses the macro to impact our publicly exposed interface, which means we need to ship a definition of the macro in our headers with our prebuilt releases.

We don't currently ship LLVM's config.h, and probably shouldn't because it has a lot of things that don't matter, so we'll need to setup a generator for a new config.h.

You'll need to add a CMake option as described in my previous comment, marked as advanced. Instead of setting the flag per-file you create a new fine "include/dxc/config.h.cmake" with something like:

```c++
/* Disable overriding memory allocators. */
#cmakedefine DXC_DISABLE_ALLOCATOR_OVERRIDES ${DXC_DISABLE_ALLOCATOR_OVERRIDES}
```

> Note: I changed the define here from what is in your patch. We should prefix with `DXC` rather than `HLSL`, and I think "disabling allocator overriding" is a more accurate description than using default new/delete based on the extended scope of the changes.

Then in `include/dxc/CMakeLists.txt` you'll add something like:

```cmake
configure_file(
  ${CMAKE_CURRENT_SOURCE_DIR}config.h.cmake
  ${LLVM_INCLUDE_DIR}/dxc/config.h)
```

Which will cause CMake to generate the configuration file.

You'll also need to add `${CMAKE_CURRENT_BINARY_DIR}/config.h` to the `DXC_PUBLIC_HEADERS` variable also in `include/dxc/CMakeLists.txt`.

(cc @hekota, we'll need to make sure our future releases also contain config.h in the built packages)

amaiorano:
Thanks, @llvm-beanz for the detailed instructions. I've applied the changes (with some required tweaks to get it working). Please take a look.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -182,7 +182,7 @@ static void SavePixelsToFile(LPCVOID pPixels, DXGI_FORMAT format, UINT32 m_width
 
   VERIFY_SUCCEEDED(ctx.Init());
   VERIFY_SUCCEEDED(CoCreateInstance(CLSID_WICImagingFactory, NULL, CLSCTX_INPROC_SERVER, IID_IWICImagingFactory, (LPVOID*)&pFactory));
-  VERIFY_SUCCEEDED(CoGetMalloc(1, &pMalloc));
+  VERIFY_SUCCEEDED(DxcCoGetMalloc(1, &pMalloc));
```
tex3d, line 185: It turns out this is dead code (pMalloc is unused), so you can get rid of it, and get rid of the dependency too!

tools/clang/tools/dxclib/dxc.cpp
```diff
@@ -557,7 +557,7 @@ void DxcContext::ExtractRootSignature(IDxcBlob *pBlob, IDxcBlob **ppResult) {
   hlsl::InitDxilContainer(&newHeader, 1, containerSize); 
   CComPtr<IMalloc> pMalloc;
   CComPtr<hlsl::AbstractMemoryStream> pMemoryStream;
-  IFT(CoGetMalloc(1, &pMalloc));
+  IFT(DxcCoGetMalloc(1, &pMalloc));
```
tex3d, line 560: I think we can get rid of this, and just use `DxcGetThreadMallocNoRef()` in place of pMalloc below.

tools/clang/tools/dxcompiler/dxclinker.cpp
```diff
@@ -131,7 +131,7 @@ DxcLinker::RegisterLibrary(_In_opt_ LPCWSTR pLibName, // Name of the library.
     CComPtr<IMalloc> pMalloc;
     CComPtr<AbstractMemoryStream> pDiagStream;
 
-    IFT(CoGetMalloc(1, &pMalloc));
+    IFT(DxcCoGetMalloc(1, &pMalloc));
```
tex3d, line 134: I think we should get rid of this, and just use `DxcGetThreadMallocNoRef()` in place of pMalloc below.

tools/clang/tools/dxcompiler/dxclinker.cpp
```diff
@@ -184,7 +184,7 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
     CComPtr<IDxcBlob> pOutputBlob;
     CComPtr<AbstractMemoryStream> pDiagStream;
 
-    IFT(CoGetMalloc(1, &pMalloc));
+    IFT(DxcCoGetMalloc(1, &pMalloc));
```
tex3d, line 187: I think we can get rid of this, and just use `DxcGetThreadMallocNoRef()` in place of pMalloc below.

include/dxc/Support/WinIncludes.h
```diff
@@ -112,3 +114,26 @@ inline ENUMTYPE &operator ^= (ENUMTYPE &a, ENUMTYPE b) { return (ENUMTYPE &)(((_
 #endif
 
 #endif // _MSC_VER
+
+/// DxcCoGetMalloc
+#if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
+
+#define DxcCoGetMalloc CoGetMalloc
```
tex3d, line 121: Alternatively, we could always have a DxcCoGetMalloc function defined, which forwards to CoGetMalloc conditionally.  But this works too.

lib/DxcSupport/CMakeLists.txt
```diff
@@ -17,4 +18,6 @@ ${LLVM_MAIN_SRC_DIR}/lib/DxcSupport/SharedLibAffix.inc
 ${LLVM_INCLUDE_DIR}/dxc/Support/SharedLibAffix.h
 )
 
-add_dependencies(LLVMDxcSupport TablegenHLSLOptions)
+target_link_libraries(LLVMDxcSupport PUBLIC LLVMSupport)
+
+add_dependencies(LLVMDxcSupport LLVMSupport TablegenHLSLOptions)
```
tex3d, line 23: Were these changes to fix something that was already missing/incorrect, or newly needed by changes introduced in this PR?

tools/clang/unittests/HLSLExec/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set_target_properties(ExecHLSLTests PROPERTIES FOLDER "Execution tests")
 
 target_link_libraries(ExecHLSLTests PRIVATE
   dxcompiler
+  LLVMDxcSupport
```
tex3d, line 18: If we can get rid of the DxcCoGetMalloc dependency (see other comment), can we get rid of this dependency as well?

tex3d (changes requested):
I pointed out the places where I believe we can get rid of `DxcCoGetMalloc()` uses.  Other than that, I think this is looking good!

tools/clang/tools/dxclib/dxc.cpp
amaiorano: Done.

tools/clang/tools/dxcompiler/dxclinker.cpp
amaiorano: Done.

tools/clang/tools/dxcompiler/dxclinker.cpp
amaiorano: Done.

tools/clang/unittests/HLSLExec/CMakeLists.txt
amaiorano: Yes, and done.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
amaiorano: Nice :)

amaiorano:
I just [pushed a change](https://github.com/microsoft/DirectXShaderCompiler/compare/2f3a89673f8df88b179c82694527aad0007c5392..7dccd21fa0785136589ebc982dd5f4c19595ad35) that was necessary for our GN build when targeting Windows x86 32-bit. The compiler was failing because the overridden member functions of `IMalloc` did not match the ones declared in IUnknown because of the mismatch between `size_t` vs `SIZE_T`. This was not a problem when `IMalloc` was only defined for non-Windows targets, but now that it's also possibly defined for Windows, it must use `SIZE_T`. This wasn't caught by the CI here because no build defines `DXC_DISABLE_ALLOCATOR_OVERRIDES`. Might be a good idea to add at least one such CI target.

Anyway, with this, we are able to build both for x64 and x86 Windows.

amaiorano:
> ❌ [Build DirectXShaderCompiler 1.0.3384 failed](https://ci.appveyor.com/project/dnovillo/directxshadercompiler/builds/47349462) (commit [4129dd2af2](https://github.com/microsoft/DirectXShaderCompiler/commit/4129dd2af2) by @amaiorano)

This build failure seems to be a timeout? `Build execution time has reached the maximum allowed time for your plan (90 minutes).`

tools/clang/unittests/HLSLExec/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ include_directories(${DIASDK_INCLUDE_DIRS})
 include_directories(${D3D12_INCLUDE_DIRS})
 include_directories(${LLVM_MAIN_INCLUDE_DIR}/dxc/Test)
 
-add_dependencies(ExecHLSLTests dxcompiler)
+add_dependencies(ExecHLSLTests dxcompiler LLVMDxcSupport)
```
tex3d, line 28: Is this change still necessary, now that CoGetMalloc has been removed?

include/dxc/Support/WinIncludes.h
tex3d: I think this is fine, unless changing it would reduce dependencies or complexity somewhere.

tex3d (commented):
Just one remaining question about an added dependency in `HLSLExec/CMakeLists.txt`.

tools/clang/unittests/HLSLExec/CMakeLists.txt
amaiorano: It is not. Good catch! Changes reverted.

tex3d (approved):
LGTM!

amaiorano:
Thank you @tex3d ! Looks like I still need another review. @llvm-beanz, would you mind taking another look?

amaiorano:
I've rebased my change on ToT. Please take a look @llvm-beanz.

amaiorano:
Friendly ping for another review?

include/dxc/CMakeLists.txt
```diff
@@ -10,13 +10,19 @@ configure_file(
...
+  )
+
 add_subdirectory(DXIL)
 add_subdirectory(DxilContainer)
 add_subdirectory(HLSL)
 add_subdirectory(Support)
 add_subdirectory(Tracing)
 
 set(DXC_PUBLIC_HEADERS
+    ${CMAKE_CURRENT_BINARY_DIR}/config.h
```
llvm-beanz, line 25: @hekota & @pow2clk, we need to make sure this header gets picked up in the DXC packages.

lib/DxcSupport/WinIncludes.cpp
```diff
@@ -0,0 +1,44 @@
+//===- WinIncludes.cpp -----------------------------------------*- C++ -*-===//
```
llvm-beanz, line 1: I know you just pulled this code out from another file, but we should clang-format it, and ideally the full patch.

(see: https://clang.llvm.org/docs/ClangFormat.html#script-for-patch-reformatting).

llvm-beanz (approved):
LGTM. Can you please clang-format the patch before we merge it?

include/dxc/CMakeLists.txt
llvm-beanz: Filed #5414 to track making sure the header gets into our releases.

lib/DxcSupport/WinIncludes.cpp
amaiorano: I tried the latest LLVM 15 version of clang-format, and unforunately, when I use clang-format-diff.py on the patch, I get unexpected results. For instance:

```
C:\src\external\DirectXShaderCompiler>where clang-format.exe
c:\Program Files\LLVM\bin\clang-format.exe

C:\src\external\DirectXShaderCompiler>git diff -U0 --no-color --relative HEAD~1 | py -3 "c:\Program Files\LLVM\share\clang\clang-format-diff.py" -p1
--- include/dxc/Support/WinIncludes.h   (before formatting)
+++ include/dxc/Support/WinIncludes.h   (after formatting)
@@ -125,12 +125,12 @@
 #ifndef _WIN32
 CROSS_PLATFORM_UUIDOF(IMalloc, "00000002-0000-0000-C000-000000000046")
 struct IMalloc : public IUnknown {
-  virtual void *Alloc(SIZE_T size) = 0;
-  virtual void *Realloc(void *ptr, SIZE_T size) = 0;
-  virtual void Free(void *ptr) = 0;
-  virtual SIZE_T GetSize(void *pv) = 0;
-  virtual int DidAlloc(void *pv) = 0;
-  virtual void HeapMinimize(void) = 0;
+        virtual void *Alloc(SIZE_T size) = 0;
+        virtual void *Realloc(void *ptr, SIZE_T size) = 0;
+        virtual void Free(void *ptr) = 0;
+        virtual SIZE_T GetSize(void *pv) = 0;
+        virtual int DidAlloc(void *pv) = 0;
+        virtual void HeapMinimize(void) = 0;
 };
 #endif
```

Note how the diff indented the member functions in `IMalloc` with 8 spaces. The `.clang-format` file sets the style to LLVM, which uses 2 space indentation. If I run clang-format directly against this file, it indeed spits out this class with 2-space indented functions as expected:

```
C:\src\external\DirectXShaderCompiler>clang-format C:\src\external\DirectXShaderCompiler\include\dxc\Support\WinIncludes.h

<SNIP>

#ifndef _WIN32
CROSS_PLATFORM_UUIDOF(IMalloc, "00000002-0000-0000-C000-000000000046")
struct IMalloc : public IUnknown {
  virtual void *Alloc(SIZE_T size) = 0;
  virtual void *Realloc(void *ptr, SIZE_T size) = 0;
  virtual void Free(void *ptr) = 0;
  virtual SIZE_T GetSize(void *pv) = 0;
  virtual int DidAlloc(void *pv) = 0;
  virtual void HeapMinimize(void) = 0;
};
#endif

HRESULT DxcCoGetMalloc(DWORD dwMemContext, IMalloc **ppMalloc);

#endif // defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
```

I'm not sure what's going on. Maybe clang-format doesn't work properly without the full context fo the file? Should I bother with this? Other files also now have 8 space indentation.

amaiorano: To add to what I wrote above, attempting to fully format almost any file in the codebase seems to produce a significant difference. I think if the goal is to clang-format the entire codebase, this should perhaps be done first, and then enforced with pre-push hooks and CI. If it's alright, I'd like to land this as-is.

There seems to be one more "workflow awaiting approval", but I'm nto sure what that means:

image

llvm-beanz: Our codebase is not clang-format clean, but new code introduced should be clang-formatted. If `clang-format-diff` isn't finding the format settings in the root of the repo that seems like a configuration issue. It has worked for me without issue. `clang-format-diff` does work without the full context, so this seems like a configuration problem. Beyond that I don't really have guidance.

amaiorano:
I don't believe it's a configuration issue. Maybe you can try it with my
patch?

llvm-beanz:
> I don't believe it's a configuration issue. Maybe you can try it with my patch?

On your PR, running this command ` git diff -U0 --no-color HEAD~1...HEAD | python3 ~/dev/llvm-project/clang/tools/clang-format/clang-format-diff.py -p1 | git apply -p0` produces the following diff for me:

```diff
diff --git a/include/dxc/Support/microcom.h b/include/dxc/Support/microcom.h
index 9ac4f13b299..b47f1cb5afd 100644
--- a/include/dxc/Support/microcom.h
+++ b/include/dxc/Support/microcom.h
@@ -12,9 +12,9 @@
 #ifndef __DXC_MICROCOM__
 #define __DXC_MICROCOM__
 
-#include <atomic>
 #include "dxc/Support/WinIncludes.h"
 #include "llvm/Support/Atomic.h"
+#include <atomic>
 
 template <typename TIface>
 class CComInterfaceArray {
diff --git a/lib/DxcSupport/WinIncludes.cpp b/lib/DxcSupport/WinIncludes.cpp
index 9052163869d..878830c35c2 100644
--- a/lib/DxcSupport/WinIncludes.cpp
+++ b/lib/DxcSupport/WinIncludes.cpp
@@ -10,14 +10,14 @@
 
 #include "dxc/Support/WinIncludes.h"
 
-#include "dxc/Support/microcom.h"
 #include "assert.h"
+#include "dxc/Support/microcom.h"
 
 #if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
 // CoGetMalloc from combaseapi.h is used
 #else
 struct DxcCoMalloc : public IMalloc {
-  DxcCoMalloc() : m_dwRef(0) {};
+  DxcCoMalloc() : m_dwRef(0){};
 
   DXC_MICROCOM_ADDREF_RELEASE_IMPL(m_dwRef)
   STDMETHODIMP QueryInterface(REFIID riid, void **ppvObject) override {
@@ -26,7 +26,9 @@ struct DxcCoMalloc : public IMalloc {
   }
 
   void *STDMETHODCALLTYPE Alloc(SIZE_T size) override { return malloc(size); }
-  void *STDMETHODCALLTYPE Realloc(void *ptr, SIZE_T size) override { return realloc(ptr, size); }
+  void *STDMETHODCALLTYPE Realloc(void *ptr, SIZE_T size) override {
+    return realloc(ptr, size);
+  }
   void STDMETHODCALLTYPE Free(void *ptr) override { free(ptr); }
   SIZE_T STDMETHODCALLTYPE GetSize(void *pv) override { return -1; }
   int STDMETHODCALLTYPE DidAlloc(void *pv) override { return -1; }
diff --git a/projects/dxilconv/tools/dxilconv/dxilconv.cpp b/projects/dxilconv/tools/dxilconv/dxilconv.cpp
index 0e3e53b062d..8cbb2239021 100644
--- a/projects/dxilconv/tools/dxilconv/dxilconv.cpp
+++ b/projects/dxilconv/tools/dxilconv/dxilconv.cpp
@@ -18,12 +18,12 @@
 
 #define DXC_API_IMPORT
 
+#include "DxbcConverter.h"
+#include "Tracing/DxcRuntimeEtw.h"
 #include "dxc/config.h"
 #include "dxc/dxcisense.h"
 #include "dxc/dxctools.h"
 #include "dxcetw.h"
-#include "Tracing/DxcRuntimeEtw.h"
-#include "DxbcConverter.h"
 
 // Defined in DxbcConverter.lib (projects/dxilconv/lib/DxbcConverter/DxbcConverter.cpp)
 HRESULT CreateDxbcConverter(_In_ REFIID riid, _Out_ LPVOID *ppv);
diff --git a/tools/clang/tools/dxcompiler/DXCompiler.cpp b/tools/clang/tools/dxcompiler/DXCompiler.cpp
index d035cb8dc2a..4ce73b2ba61 100644
--- a/tools/clang/tools/dxcompiler/DXCompiler.cpp
+++ b/tools/clang/tools/dxcompiler/DXCompiler.cpp
@@ -9,12 +9,12 @@
 //                                                                           //
 ///////////////////////////////////////////////////////////////////////////////
 
-#include "llvm/Support/ManagedStatic.h"
-#include "llvm/Support/FileSystem.h"
-#include "dxc/config.h"
 #include "dxc/Support/Global.h"
-#include "dxc/Support/WinIncludes.h"
 #include "dxc/Support/HLSLOptions.h"
+#include "dxc/Support/WinIncludes.h"
+#include "dxc/config.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/ManagedStatic.h"
 #ifdef LLVM_ON_WIN32
 #include "dxcetw.h"
 #endif
diff --git a/tools/clang/tools/dxcompiler/dxcapi.cpp b/tools/clang/tools/dxcompiler/dxcapi.cpp
index 77ec0820371..53596beac9e 100644
--- a/tools/clang/tools/dxcompiler/dxcapi.cpp
+++ b/tools/clang/tools/dxcompiler/dxcapi.cpp
@@ -17,10 +17,10 @@
 #define DXC_API_IMPORT __attribute__ ((visibility ("default")))
 #endif
 
+#include "dxc/Support/Global.h"
 #include "dxc/config.h"
 #include "dxc/dxcisense.h"
 #include "dxc/dxctools.h"
-#include "dxc/Support/Global.h"
 #ifdef _WIN32
 #include "dxcetw.h"
 #endif
@@ -160,9 +160,9 @@ DxcCreateInstance2(
     return E_POINTER;
   }
 #ifdef DXC_DISABLE_ALLOCATOR_OVERRIDES
-    if (pMalloc != DxcGetThreadMallocNoRef()) {
-      return E_INVALIDARG;
-    }
+  if (pMalloc != DxcGetThreadMallocNoRef()) {
+    return E_INVALIDARG;
+  }
 #endif // DXC_DISABLE_ALLOCATOR_OVERRIDES
 
   HRESULT hr = S_OK;
diff --git a/tools/clang/tools/dxcompiler/dxclinker.cpp b/tools/clang/tools/dxcompiler/dxclinker.cpp
index a376bbd621d..735979b4127 100644
--- a/tools/clang/tools/dxcompiler/dxclinker.cpp
+++ b/tools/clang/tools/dxcompiler/dxclinker.cpp
@@ -268,9 +268,8 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
         // Validation.
         HRESULT valHR = S_OK;
         dxcutil::AssembleInputs inputs(
-          std::move(pM), pOutputBlob, DxcGetThreadMallocNoRef(), SerializeFlags,
-          pOutputStream,
-          opts.DebugFile, &Diag);
+            std::move(pM), pOutputBlob, DxcGetThreadMallocNoRef(),
+            SerializeFlags, pOutputStream, opts.DebugFile, &Diag);
         if (needsValidation) {
           valHR = dxcutil::ValidateAndAssembleToContainer(inputs);
         } else {
diff --git a/tools/clang/tools/dxrfallbackcompiler/DXCompiler.cpp b/tools/clang/tools/dxrfallbackcompiler/DXCompiler.cpp
index b2e1a0929b3..f3d0bfcfd7f 100644
--- a/tools/clang/tools/dxrfallbackcompiler/DXCompiler.cpp
+++ b/tools/clang/tools/dxrfallbackcompiler/DXCompiler.cpp
@@ -9,14 +9,14 @@
 //                                                                           //
 ///////////////////////////////////////////////////////////////////////////////
 
-#include "llvm/Support/ManagedStatic.h"
-#include "llvm/Support/FileSystem.h"
-#include "dxc/config.h"
 #include "dxc/Support/Global.h"
-#include "dxc/Support/WinIncludes.h"
 #include "dxc/Support/HLSLOptions.h"
+#include "dxc/Support/WinIncludes.h"
+#include "dxc/config.h"
 #include "dxcetw.h"
 #include "dxillib.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/ManagedStatic.h"
 
 namespace hlsl { HRESULT SetupRegistryPassForHLSL(); }
 
diff --git a/tools/clang/unittests/HLSL/ValidationTest.cpp b/tools/clang/unittests/HLSL/ValidationTest.cpp
index 1ceae90bed1..9d634bc9efc 100644
--- a/tools/clang/unittests/HLSL/ValidationTest.cpp
+++ b/tools/clang/unittests/HLSL/ValidationTest.cpp
@@ -15,12 +15,12 @@
 #include <string>
 #include <algorithm>
 
-#include "llvm/ADT/StringRef.h"
-#include "llvm/Support/Regex.h"
-#include "llvm/ADT/ArrayRef.h"
 #include "dxc/DxilContainer/DxilContainer.h"
 #include "dxc/DxilContainer/DxilContainerAssembler.h"
 #include "dxc/Support/WinIncludes.h"
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/Regex.h"
 
 #ifdef _WIN32
 #include <atlbase.h>
```

lib/DxcSupport/WinIncludes.cpp
amaiorano: Oops, pasted my comment in the wrong place:

I don't believe it's a configuration issue. Maybe you can try it with my
patch?

amaiorano:
> > I don't believe it's a configuration issue. Maybe you can try it with my patch?
> 
> On your PR, running this command ` git diff -U0 --no-color HEAD~1...HEAD | python3 ~/dev/llvm-project/clang/tools/clang-format/clang-format-diff.py -p1 | git apply -p0` produces the following diff for me:
> 
> ```diff
> diff --git a/include/dxc/Support/microcom.h b/include/dxc/Support/microcom.h
> index 9ac4f13b299..b47f1cb5afd 100644
> --- a/include/dxc/Support/microcom.h
> +++ b/include/dxc/Support/microcom.h
> @@ -12,9 +12,9 @@
>  #ifndef __DXC_MICROCOM__
>  #define __DXC_MICROCOM__
>  
> -#include <atomic>
>  #include "dxc/Support/WinIncludes.h"
>  #include "llvm/Support/Atomic.h"
> +#include <atomic>
>  
>  template <typename TIface>
>  class CComInterfaceArray {
> diff --git a/lib/DxcSupport/WinIncludes.cpp b/lib/DxcSupport/WinIncludes.cpp
> index 9052163869d..878830c35c2 100644
> --- a/lib/DxcSupport/WinIncludes.cpp
> +++ b/lib/DxcSupport/WinIncludes.cpp
> @@ -10,14 +10,14 @@
>  
>  #include "dxc/Support/WinIncludes.h"
>  
> -#include "dxc/Support/microcom.h"
>  #include "assert.h"
> +#include "dxc/Support/microcom.h"
>  
>  #if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
>  // CoGetMalloc from combaseapi.h is used
>  #else
>  struct DxcCoMalloc : public IMalloc {
> -  DxcCoMalloc() : m_dwRef(0) {};
> +  DxcCoMalloc() : m_dwRef(0){};
>  
>    DXC_MICROCOM_ADDREF_RELEASE_IMPL(m_dwRef)
>    STDMETHODIMP QueryInterface(REFIID riid, void **ppvObject) override {
> @@ -26,7 +26,9 @@ struct DxcCoMalloc : public IMalloc {
>    }
>  
>    void *STDMETHODCALLTYPE Alloc(SIZE_T size) override { return malloc(size); }
> -  void *STDMETHODCALLTYPE Realloc(void *ptr, SIZE_T size) override { return realloc(ptr, size); }
> +  void *STDMETHODCALLTYPE Realloc(void *ptr, SIZE_T size) override {
> +    return realloc(ptr, size);
> +  }
>    void STDMETHODCALLTYPE Free(void *ptr) override { free(ptr); }
>    SIZE_T STDMETHODCALLTYPE GetSize(void *pv) override { return -1; }
>    int STDMETHODCALLTYPE DidAlloc(void *pv) override { return -1; }
> diff --git a/projects/dxilconv/tools/dxilconv/dxilconv.cpp b/projects/dxilconv/tools/dxilconv/dxilconv.cpp
> index 0e3e53b062d..8cbb2239021 100644
> --- a/projects/dxilconv/tools/dxilconv/dxilconv.cpp
> +++ b/projects/dxilconv/tools/dxilconv/dxilconv.cpp
> @@ -18,12 +18,12 @@
>  
>  #define DXC_API_IMPORT
>  
> +#include "DxbcConverter.h"
> +#include "Tracing/DxcRuntimeEtw.h"
>  #include "dxc/config.h"
>  #include "dxc/dxcisense.h"
>  #include "dxc/dxctools.h"
>  #include "dxcetw.h"
> -#include "Tracing/DxcRuntimeEtw.h"
> -#include "DxbcConverter.h"
>  
>  // Defined in DxbcConverter.lib (projects/dxilconv/lib/DxbcConverter/DxbcConverter.cpp)
>  HRESULT CreateDxbcConverter(_In_ REFIID riid, _Out_ LPVOID *ppv);
> diff --git a/tools/clang/tools/dxcompiler/DXCompiler.cpp b/tools/clang/tools/dxcompiler/DXCompiler.cpp
> index d035cb8dc2a..4ce73b2ba61 100644
> --- a/tools/clang/tools/dxcompiler/DXCompiler.cpp
> +++ b/tools/clang/tools/dxcompiler/DXCompiler.cpp
> @@ -9,12 +9,12 @@
>  //                                                                           //
>  ///////////////////////////////////////////////////////////////////////////////
>  
> -#include "llvm/Support/ManagedStatic.h"
> -#include "llvm/Support/FileSystem.h"
> -#include "dxc/config.h"
>  #include "dxc/Support/Global.h"
> -#include "dxc/Support/WinIncludes.h"
>  #include "dxc/Support/HLSLOptions.h"
> +#include "dxc/Support/WinIncludes.h"
> +#include "dxc/config.h"
> +#include "llvm/Support/FileSystem.h"
> +#include "llvm/Support/ManagedStatic.h"
>  #ifdef LLVM_ON_WIN32
>  #include "dxcetw.h"
>  #endif
> diff --git a/tools/clang/tools/dxcompiler/dxcapi.cpp b/tools/clang/tools/dxcompiler/dxcapi.cpp
> index 77ec0820371..53596beac9e 100644
> --- a/tools/clang/tools/dxcompiler/dxcapi.cpp
> +++ b/tools/clang/tools/dxcompiler/dxcapi.cpp
> @@ -17,10 +17,10 @@
>  #define DXC_API_IMPORT __attribute__ ((visibility ("default")))
>  #endif
>  
> +#include "dxc/Support/Global.h"
>  #include "dxc/config.h"
>  #include "dxc/dxcisense.h"
>  #include "dxc/dxctools.h"
> -#include "dxc/Support/Global.h"
>  #ifdef _WIN32
>  #include "dxcetw.h"
>  #endif
> @@ -160,9 +160,9 @@ DxcCreateInstance2(
>      return E_POINTER;
>    }
>  #ifdef DXC_DISABLE_ALLOCATOR_OVERRIDES
> -    if (pMalloc != DxcGetThreadMallocNoRef()) {
> -      return E_INVALIDARG;
> -    }
> +  if (pMalloc != DxcGetThreadMallocNoRef()) {
> +    return E_INVALIDARG;
> +  }
>  #endif // DXC_DISABLE_ALLOCATOR_OVERRIDES
>  
>    HRESULT hr = S_OK;
> diff --git a/tools/clang/tools/dxcompiler/dxclinker.cpp b/tools/clang/tools/dxcompiler/dxclinker.cpp
> index a376bbd621d..735979b4127 100644
> --- a/tools/clang/tools/dxcompiler/dxclinker.cpp
> +++ b/tools/clang/tools/dxcompiler/dxclinker.cpp
> @@ -268,9 +268,8 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
>          // Validation.
>          HRESULT valHR = S_OK;
>          dxcutil::AssembleInputs inputs(
> -          std::move(pM), pOutputBlob, DxcGetThreadMallocNoRef(), SerializeFlags,
> -          pOutputStream,
> -          opts.DebugFile, &Diag);
> +            std::move(pM), pOutputBlob, DxcGetThreadMallocNoRef(),
> +            SerializeFlags, pOutputStream, opts.DebugFile, &Diag);
>          if (needsValidation) {
>            valHR = dxcutil::ValidateAndAssembleToContainer(inputs);
>          } else {
> diff --git a/tools/clang/tools/dxrfallbackcompiler/DXCompiler.cpp b/tools/clang/tools/dxrfallbackcompiler/DXCompiler.cpp
> index b2e1a0929b3..f3d0bfcfd7f 100644
> --- a/tools/clang/tools/dxrfallbackcompiler/DXCompiler.cpp
> +++ b/tools/clang/tools/dxrfallbackcompiler/DXCompiler.cpp
> @@ -9,14 +9,14 @@
>  //                                                                           //
>  ///////////////////////////////////////////////////////////////////////////////
>  
> -#include "llvm/Support/ManagedStatic.h"
> -#include "llvm/Support/FileSystem.h"
> -#include "dxc/config.h"
>  #include "dxc/Support/Global.h"
> -#include "dxc/Support/WinIncludes.h"
>  #include "dxc/Support/HLSLOptions.h"
> +#include "dxc/Support/WinIncludes.h"
> +#include "dxc/config.h"
>  #include "dxcetw.h"
>  #include "dxillib.h"
> +#include "llvm/Support/FileSystem.h"
> +#include "llvm/Support/ManagedStatic.h"
>  
>  namespace hlsl { HRESULT SetupRegistryPassForHLSL(); }
>  
> diff --git a/tools/clang/unittests/HLSL/ValidationTest.cpp b/tools/clang/unittests/HLSL/ValidationTest.cpp
> index 1ceae90bed1..9d634bc9efc 100644
> --- a/tools/clang/unittests/HLSL/ValidationTest.cpp
> +++ b/tools/clang/unittests/HLSL/ValidationTest.cpp
> @@ -15,12 +15,12 @@
>  #include <string>
>  #include <algorithm>
>  
> -#include "llvm/ADT/StringRef.h"
> -#include "llvm/Support/Regex.h"
> -#include "llvm/ADT/ArrayRef.h"
>  #include "dxc/DxilContainer/DxilContainer.h"
>  #include "dxc/DxilContainer/DxilContainerAssembler.h"
>  #include "dxc/Support/WinIncludes.h"
> +#include "llvm/ADT/ArrayRef.h"
> +#include "llvm/ADT/StringRef.h"
> +#include "llvm/Support/Regex.h"
>  
>  #ifdef _WIN32
>  #include <atlbase.h>
> ```

Hmm yeah that's different. Which version of clang-format are you using?

amaiorano:
Actually, would you mind uploading your formatted version and landing this?

llvm-beanz:
> Hmm yeah that's different. Which version of clang-format are you using?

I have clang 14, 15, and 16 locally, they all produce that. The shell alias I use I've had for at least 5 or 6 years, so this shouldn't be new support.

I can try to push the changes to your branch.

amaiorano:
> > Hmm yeah that's different. Which version of clang-format are you using?
> 
> I have clang 14, 15, and 16 locally, they all produce that. The shell alias I use I've had for at least 5 or 6 years, so this shouldn't be new support.

I'll test my setup again on Monday for future changes. 
 
> I can try to push the changes to your branch.

Thanks!

amaiorano:
Thanks for the commit, @llvm-beanz. The header re-ordering broke the build, so I fixed it by fixing the order and introducing new lines appropriately so that clang-format doesn't change it. Btw, I tried again to clang-format the patch, this time in WSL, using both Clang 16 and 15, and I still get the same patch that I was getting before (e.g. 8 spaces instead of 4, etc.). It's very strange indeed.

PR #5301, opened 2023-06-15
[PIX] Add pass to log DXR 1.0 shader invocation info into a UAV

austinkinross:
This PR adds a new PIX pass that instruments DXR 1.0 shaders (intersection, any/closest hit and miss). The new pass emits each shader invocation's [ray system values](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#system-value-intrinsics) and DispatchRaysIndex into a UAV that can be read back later. This will enable new DXR debugging features in PIX.

### DIFF
--- a/lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
+++ b/lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
@@ -1,0 +1,234 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DxilPIXDXRInvocationsLog.cpp                                              //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#include "dxc/DXIL/DxilFunctionProps.h"
+#include "dxc/DXIL/DxilOperations.h"
+#include "dxc/DXIL/DxilUtil.h"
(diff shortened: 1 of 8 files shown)


lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
```diff
@@ -0,0 +1,234 @@
...
+    
+    switch (ShaderKind) {
+    case DXIL::ShaderKind::Intersection:
+    case DXIL::ShaderKind::AnyHit:
+    case DXIL::ShaderKind::ClosestHit:
+    case DXIL::ShaderKind::Miss:
+      break;
+    
+    default:
+      continue;
```
jeffnn, line 79: This loop doesn't seem to accomplish much?

jeffnn (approved):

lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    IRBuilder<> Builder(dxilutil::FirstNonAllocaInsertionPt(entryFunction));
+
+    // Add the UAVs that we're going to write to
+    CallInst* HandleForCountUAV = PIXPassHelpers::CreateUAV(DM, Builder, 0, "PIX_CountUAV_Handle");
```
adam-yang, line 85: Should the 0 and the 1 on the next line be put in a constant somewhere? I see it repeated in a few places.

adam-yang (approved):
Just one suggestion. Otherwise, LGTM!

PR #5303, opened 2023-06-17
[WIP] Fixed semantic defines not taking effect until after codegen.

adam-yang:
This change does the following things:
- Refactors the -opt-enable and -opt-disable toggles into its own type.
- Moved reading the semantic define to earlier in the compilation.
- Removed all options on `clang::CodeGenOptions` that can be influenced by semantic define, in favour of using `HLSLExtensionsCodegenHelper` instead.
- Added a bunch more tests to verify that lifetime-markers works with all the different ways to turn it on/off.

### Refactors the -opt-enable and -opt-disable toggles into its own type.
We used to have a `std::map` on `hlsl::HLSLOptions` to represent these options. The result is messy, because different options want to be on/off by default, and the logic for that is hard to read at a glance. This change moves all the toggles into static constant values in the header, and whether they're default on/off is contained within the value.

### Moved reading the semantic define to earlier in the compilation.
Semantic defines used to be initialized in `CodeGeneratorImpl::HandleTranslationUnit` by calling `HLSLExtensionsCodegenHelper::WriteSemanticDefines`. This happens too late for some of the semantic defines to take effect. For example: lifetime-markers need to be enabled/disabled before parsing any top-level entities, because

### DIFF
--- a/include/dxc/HLSL/HLSLExtensionsCodegenHelper.h
+++ b/include/dxc/HLSL/HLSLExtensionsCodegenHelper.h
@@ -67,7 +67,9 @@ public:
 
   // Write semantic defines as metadata in the module.
   virtual void WriteSemanticDefines(llvm::Module *M) = 0;
-  virtual void UpdateCodeGenOptions(clang::CodeGenOptions &CGO) = 0;
+  // Reads semantic defines and translates them into opt-enable/opt-disable toggles. This only
+  // works after parser is initialized.
+  virtual void UpdateSemanticDefinesAndOptToggles() = 0;
   // Query the named option enable
   // Needed because semantic defines may have set it since options were copied
   virtual bool IsOptionEnabled(hlsl::options::Toggle toggle) = 0;
--- a/include/dxc/HLSL/HLSLExtensionsCodegenHelper.h
+++ b/include/dxc/HLSL/HLSLExtensionsCodegenHelper.h
@@ -11,6 +11,7 @@
 
 #pragma once
 #include "dxc/DXIL/DxilOperations.h"
+#include "dxc/Support/DxcOptToggles.h"
 #include <vector>
 #include <string>
 
@@ -69,7 +70,8 @@ public:
   virtual void UpdateCodeGenOptions(clang::CodeGenOptions &CGO) = 0;
   // Query the named option enable
   // Needed because semantic defines may have set it since options were copied
-  virtual bool IsOptionEnabled(std::string option) = 0;
+  virtual bool IsOptionEnabled(hlsl::options::Toggle toggle) = 0;
+  virtual bool IsLifetimeMarkersEnabled() = 0;
 
   // Get the name to use for the dxil intrinsic function.
   virtual std::string GetIntrinsicName(unsigned opcode) = 0;
--- a/include/dxc/Support/DxcOptToggles.h
+++ b/include/dxc/Support/DxcOptToggles.h
@@ -1,0 +1,73 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DxcOptToggles.h                                                           //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Helper code for representing -opt-disable, -opt-enable, -opt-select       //
+// options                                                                   //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#ifndef LLVM_HLSL_DXC_OPT_TOGGLES_H
+#define LLVM_HLSL_DXC_OPT_TOGGLES_H
+
+#include "llvm/ADT/StringRef.h"
+#include <map>
+#include <set>
+#include <string>
+
+namespace hlsl {
+
+namespace options {
+
+static const llvm::StringRef TOGGLE_GVN  = "gvn";
+static const llvm::StringRef TOGGLE_LICM = "licm";
+static const llvm::StringRef TOGGLE_SINK  = "sink";
+static const llvm::StringRef TOGGLE_LIFETIME_MARKERS = "lifetime-markers";
+static const llvm::StringRef TOGGLE_PARTIAL_LIFETIME_MARKERS = "partial-lifetime-markers";
+static const llvm::StringRef TOGGLE_STRUCTURIZE_LOOP_EXITS_FOR_UNROLL = "structurize-loop-exits-for-unroll";
+static const llvm::StringRef TOGGLE_DEBUG_NOPS = "debug-nops";
+static const llvm::StringRef TOGGLE_STRUCTURIZE_RETURNS = "structurize-returns";
+
+struct OptimizationToggles {
+  // Optimization pass enables, disables and selects
+  std::map<std::string, bool>        Toggles; // OPT_opt_enable & OPT_opt_disable
+  std::map<std::string, std::string> Selects; // OPT_opt_select
+
+  inline void Set(llvm::StringRef Opt, bool Value) {
+    Toggles[Opt] = Value;
+  }
+  inline bool SetAndTrue(llvm::StringRef Opt) const {
+    auto It = Toggles.find(Opt);
+    return It != Toggles.end() && It->second;
+  }
+  inline bool SetAndFalse(llvm::StringRef Opt) const {
+    auto It = Toggles.find(Opt);
+    return It != Toggles.end() && !It->second;
+  }
+  inline bool Get(llvm::StringRef Opt, bool DefaultOn) const {
+    auto It = Toggles.find(Opt);
+    const bool Found = It != Toggles.end();
+    if (DefaultOn) {
+      return !Found || It->second;
+    }
+    else {
+      return Found && It->second;
+    }
+  }
+  inline bool GetDefaultOn(llvm::StringRef Opt) const {
+    return Get(Opt, /*DefaultOn*/true);
+  }
+  inline bool GetDefaultOff(llvm::StringRef Opt) const {
+    return Get(Opt, /*DefaultOn*/false);
+  }
+};
+
+} // namespace options
+} // namespace hlsl
+
+#endif
--- a/include/dxc/Support/DxcOptToggles.h
+++ b/include/dxc/Support/DxcOptToggles.h
@@ -49,6 +49,9 @@ struct OptimizationToggles {
     auto It = Toggles.find(Opt);
     return It != Toggles.end() && !It->second;
   }
+  inline bool Has(llvm::StringRef Opt) const {
+    return Toggles.find(Opt) != Toggles.end();
+  }
   inline bool Get(llvm::StringRef Opt, bool DefaultOn) const {
     auto It = Toggles.find(Opt);
     const bool Found = It != Toggles.end();
--- a/include/dxc/Support/DxcOptToggles.h
+++ b/include/dxc/Support/DxcOptToggles.h
@@ -24,49 +24,39 @@ namespace hlsl {
 
 namespace options {
 
-static const llvm::StringRef TOGGLE_GVN  = "gvn";
-static const llvm::StringRef TOGGLE_LICM = "licm";
-static const llvm::StringRef TOGGLE_SINK  = "sink";
-static const llvm::StringRef TOGGLE_LIFETIME_MARKERS = "lifetime-markers";
-static const llvm::StringRef TOGGLE_PARTIAL_LIFETIME_MARKERS = "partial-lifetime-markers";
-static const llvm::StringRef TOGGLE_STRUCTURIZE_LOOP_EXITS_FOR_UNROLL = "structurize-loop-exits-for-unroll";
-static const llvm::StringRef TOGGLE_DEBUG_NOPS = "debug-nops";
-static const llvm::StringRef TOGGLE_STRUCTURIZE_RETURNS = "structurize-returns";
+struct Toggle {
+  llvm::StringRef Name;
+  bool Default = false;
+};
+
+static const Toggle TOGGLE_GVN = {"gvn", true};
+static const Toggle TOGGLE_LICM = {"licm", true};
+static const Toggle TOGGLE_SINK = {"sink", true};
+static const Toggle TOGGLE_LIFETIME_MARKERS = {"lifetime-markers", false};
+static const Toggle TOGGLE_PARTIAL_LIFETIME_MARKERS = {
+    "partial-lifetime-markers", false};
+static const Toggle TOGGLE_STRUCTURIZE_LOOP_EXITS_FOR_UNROLL = {
+    "structurize-loop-exits-for-unroll", true};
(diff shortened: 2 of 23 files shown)


include/dxc/Support/DxcOptToggles.h
```diff
@@ -0,0 +1,66 @@
...
+    "structurize-loop-exits-for-unroll", true};
+static const Toggle TOGGLE_DEBUG_NOPS = {"debug-nops", true};
+static const Toggle TOGGLE_STRUCTURIZE_RETURNS = {"structurize-returns", false};
+
+struct OptimizationToggles {
+  // Optimization pass enables, disables and selects
+  std::map<std::string, bool>        Toggles; // OPT_opt_enable & OPT_opt_disable
+  std::map<std::string, std::string> Selects; // OPT_opt_select
+
+  inline void Set(Toggle Opt, bool Value) {
```
dmpots, line 48: Is there a reason to add `inline` here? Isn't a function defined in a class/struct body already implicitly inline?

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -604,17 +604,17 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
     opts.ScanLimit = std::stoul(std::string(limit));
 
   for (std::string opt : Args.getAllArgValues(OPT_opt_disable))
-    opts.DxcOptimizationToggles[llvm::StringRef(opt).lower()] = false;
+    opts.OptToggles.Toggles[llvm::StringRef(opt).lower()] = false;
```
dmpots, line 607: Should we validate that these options should map to a known toggle and issue a warning if they don't? Do we expect there is a valid use case for passing a string that does not map to a known toggle?

include/dxc/HLSL/HLSLExtensionsCodegenHelper.h
```diff
@@ -66,10 +67,13 @@ class HLSLExtensionsCodegenHelper {
...
   virtual void WriteSemanticDefines(llvm::Module *M) = 0;
-  virtual void UpdateCodeGenOptions(clang::CodeGenOptions &CGO) = 0;
+  // Reads semantic defines and translates them into opt-enable/opt-disable toggles. This only
+  // works after parser is initialized.
+  virtual void UpdateSemanticDefinesAndOptToggles() = 0;
   // Query the named option enable
   // Needed because semantic defines may have set it since options were copied
-  virtual bool IsOptionEnabled(std::string option) = 0;
+  virtual bool IsOptionEnabled(hlsl::options::Toggle toggle) = 0;
+  virtual bool IsLifetimeMarkersEnabled() = 0;
```
dmpots, line 76: Why is lifetime markers the only one that gets an explicit method (`IsLIfetimeMarkersEnabled`)? All the others use `IsOptionEnabled`.

tools/clang/lib/CodeGen/CGDecl.cpp
```diff
@@ -891,7 +891,7 @@ llvm::Value *CodeGenFunction::EmitLifetimeStart(uint64_t Size,
   llvm::CallInst *C =
       Builder.CreateCall(CGM.getLLVMLifetimeStartFn(), {SizeV, Addr});
   C->setDoesNotThrow();
-  if (CGM.getCodeGenOpts().HLSLEnablePartialLifetimeMarkers) return nullptr; // HLSL Change - Returning nullptr prevents generating lifetime.end
+  if (CGM.getCodeGenOpts().HLSLExtensionsCodegen && !CGM.getCodeGenOpts().HLSLExtensionsCodegen->IsOptionEnabled(hlsl::options::TOGGLE_PARTIAL_LIFETIME_MARKERS)) return nullptr; // HLSL Change - Returning nullptr prevents generating lifetime.end
```
dmpots, line 894: I think it would be nice to have a helper function that can do the check in a single call instead of having to always check for nullptr.

So we could write something like 

`if (CGM.getCodeenOpts().IsFooEnabled()) {...}` 

instead of 

`if (GM.getCodeGenOpts().HLSLExtensionsCodegen && !CGM.getCodeGenOpts().HLSLExtensionsCodegen->IsOptionEnabled(hlsl::options::TOGGLE_FOO)`

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -476,48 +450,107 @@ class HLSLExtensionsCodegenHelperImpl : public HLSLExtensionsCodegenHelper {
...
+  , m_HLSLEnableLifetimeMarkers(HLSLEnableLifetimeMarkers)
   {}
 
   // Write semantic defines as metadata in the module.
   virtual void WriteSemanticDefines(llvm::Module *M) override {
+    WriteSemanticDefines(M, m_semanticDefines);
+  }
+  // Update CodeGenOption based on HLSLOptimizationToggles.
+  void UpdateSemanticDefinesAndOptToggles() override {
+    // This shouldn't really be called more than once, but just in case it is,
```
dmpots, line 466: Should we add an assert here to catch cases where it is accidentally getting called twice?

tools/clang/tools/libclang/dxcrewriteunused.cpp
```diff
@@ -502,7 +502,21 @@ ParsedSemanticDefineList hlsl::CollectSemanticDefinesParsedByCompiler(
...
+    // state of the preprocessor will cause crashes later on, as more things
+    // are being parsed.
+    std::unique_ptr<PreprocessorOptions> Opts(
+        new PreprocessorOptions(pp.getPreprocessorOpts()));
+    clang::LangOptions langOptionsCopy = pp.getLangOpts();
+    std::unique_ptr<Preprocessor> ppCopy(new Preprocessor(
+        Opts.get(), compiler.getDiagnostics(), langOptionsCopy, pp.getSourceManager(),
+        pp.getHeaderSearchInfo(), pp.getModuleLoader(),
+        pp.getIdentifierTable().getExternalIdentifierLookup()));
+    Opts.release(); // Opts are owned by the preprocessor copy now.
```
dmpots, line 517: Should we use `Opts.release()` in the call above instead of `Opts.get()`?

adam-yang: This is a pattern that I see elsewhere (like in `CompilerInstance::createPreprocessor`). The more I look at the called code though, the more I feel wrapping it in std::unique_ptr is not the right thing to do. I think the right way to do this is to just directly pass `new PreprocessorOptions`, or if the `new PreprocessorOptions` needs to be far away, use a `IntrusiveRefCntPtr` instead of a `std::unique_ptr` and calling release.

lib/DxcSupport/HLSLOptions.cpp
adam-yang: I think with this change, the most we could do here is to add a warning, and even that might be too much of a functional change. We should have some more discussion about that.

include/dxc/Support/DxcOptToggles.h
adam-yang: There's no particular reason. I'll remove them.

tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -337,26 +338,14 @@ void EmitAssemblyHelper::CreatePasses() {
...
-                        !CodeGenOpts.HLSLOptimizationToggles.count("debug-nops") ||
-                        CodeGenOpts.HLSLOptimizationToggles.find("debug-nops")->second;
-
-  PMBuilder.HLSLEnableLifetimeMarkers =
-      CodeGenOpts.HLSLEnableLifetimeMarkers &&
-      (!CodeGenOpts.HLSLOptimizationToggles.count("lifetime-markers") ||
-       CodeGenOpts.HLSLOptimizationToggles.find("lifetime-markers")->second);
-
-  PMBuilder.HLSLEnablePartialLifetimeMarkers = CodeGenOpts.HLSLEnablePartialLifetimeMarkers;
+  if (const std::shared_ptr<hlsl::HLSLExtensionsCodegenHelper> &extHelper = CodeGenOpts.HLSLExtensionsCodegen) {
```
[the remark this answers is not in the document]
adam-yang: If `HLSLExtensionsCodegenHelper` is not setup, then the opt-enable stuff also didn't get setup. So we do require it here. I'll add an assert.

include/dxc/HLSL/HLSLExtensionsCodegenHelper.h
adam-yang: Because it checks both the flag (based on the -enable/disable-lifetime-markers flag and the -opt-enable/disable toggle).

tools/clang/tools/libclang/dxcrewriteunused.cpp
dmpots: The only reason I can think of to have the `release()` after the constructor would be if the constructor throws an exception before taking ownership of the pointer. I would consider that a bug in the called code. I don't think we have to change it, but it just looked a little weird to me.

tools/clang/lib/CodeGen/BackendUtil.cpp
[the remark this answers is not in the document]
dmpots: The one thing that's standing out to me with this change is that it seems to put a lot more emphasis on having the extensions helper there to generate code correctly. I'm not sure if that is a big change or not (i.e. did we always have it available anyway).

In some ways it feels like we are tying two independent things together by placing the optimization toggles there. It feels like there should be some way to setup and query the optimization toggles. Then if we have extensions present, we use those to modify the current optimization toggles.

Would it make sense to separate these things more do you think?

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1612,7 +1639,10 @@ class DxcCompiler : public IDxcCompiler3,
       Inlining = clang::CodeGenOptions::NormalInlining;
     compiler.getCodeGenOpts().setInlining(Inlining);
 
-    compiler.getCodeGenOpts().HLSLExtensionsCodegen = std::make_shared<HLSLExtensionsCodegenHelperImpl>(compiler, m_langExtensionsHelper, Opts.RootSignatureDefine);
+    compiler.getCodeGenOpts().HLSLExtensionsCodegen =
+        std::make_shared<HLSLExtensionsCodegenHelperImpl>(
+            compiler, m_langExtensionsHelper, Opts.RootSignatureDefine,
```
tex3d, line 1644: An existing issue I noticed here is that we use `m_langExtensionHelper` here, but we are in `SetupCompilerForCompile` which takes a pointer to the helper as a parameter `helper`.  I think this code should use `*helper` instead.  I don't know if it makes a material difference with the current state of the code, but I think it's more _correct_ given the definition of this method.

tools/clang/lib/CodeGen/BackendUtil.cpp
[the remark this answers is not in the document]
adam-yang: The reason everything is in `HLSLExtensionsCodegenHelper` is a single instance shared by the many copies of CodeGenOpts. There are many different copies of CodeGenOpts throughout the frontend; in the old code we only modified the original copy in `CompilerInstance` and the copy in `CodeGen::CodeGenModule` in `CodeGeneratorImpl::HandleTranslationUnit`, which successfully affected the `PassManagerBuilder`, but this was both too late and not the right copy to affect the earlier codegen.

If we stick to modifying the CodeGenOpts, but just doing it earlier, we will need to find the right copy to modify (which is not easy to find the right one (because there's many), and inconvenient because they're hidden inside abstract 'handler', 'action' classes, and kind hacky (becaues they're all supposed to be const)).

To sum up: I think to make 'reading semantic define earlier` work, relying on `HLSLExtensionsCodegenHelper` is the cleanest way to do it.

(Actually, an even cleaner way to do it would be to preprocess the input first before we even setup the original first copy of CodeGenOpts. However, I think that's a much much more drastic, riskier change.)

tools/clang/tools/libclang/dxcrewriteunused.cpp
adam-yang: This whole section is gone. We're no longer making a copy of Preprocessor, because I realized it didn't actually work.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
adam-yang: Probably no functional difference right now, but is cleaner. Fixed.

tools/clang/lib/CodeGen/BackendUtil.cpp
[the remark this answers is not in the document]
dmpots: What if we separated out the optimization toggles into its own shared_ptr that is shared by all the `CodeGenOpts` instances? It sounds a bit like we are putting them in the `HLSLExtensionsCodegenHelper` because it is a single place that is common to all the `CodeGenOpts` copies, but otherwise it doesn't really belong there.

Could we add the `IsOptionEnabled()` function to the new `OptToggles` class instead of putting it in `HLSLExtensionsCodegenHelper`? Then we can check whether the options are enabled without requiring the extensions. After parsing semantic defines the helper could update the shared `OptToggles` state with the values coming from the defines.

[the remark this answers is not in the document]
adam-yang: I like this suggestion. Conceptually I didn't register that opt-enable/disable toggles is a separate concept from the extension stuff. Updated in a new push.

tools/clang/include/clang/Frontend/CodeGenOptions.h
```diff
@@ -212,6 +213,21 @@ class CodeGenOptions : public CodeGenOptionsBase {
   std::vector<std::string> HLSLArguments;
   /// Helper for generating llvm bitcode for hlsl extensions.
   std::shared_ptr<hlsl::HLSLExtensionsCodegenHelper> HLSLExtensionsCodegen;
+  /// Helper for generating llvm bitcode for hlsl extensions.
```
dmpots, line 216: Copy-pasteo on the comment.

dmpots (approved):
LGTM. Looks much better after the refactor, I think. Thanks for the change!

tools/clang/include/clang/Frontend/CodeGenOptions.h
```diff
@@ -212,6 +213,21 @@ class CodeGenOptions : public CodeGenOptionsBase {
...
+  std::shared_ptr<hlsl::options::OptimizationToggles> HLSLOptToggles;
```
bogner, line 217: Why `std::shared_ptr`? This is the only owner so I think `std::unique_ptr` would be more appropriate

bogner (approved):

python3kgae (approved):

tools/clang/include/clang/Frontend/CodeGenOptions.h
adam-yang: `CodeGenOptions` gets many copies created, and we want them all to share the same instance of `HLSLOptToggles`, just like `HLSLExtensionsCodegen` above.

adam-yang: Actually I'll add a comment about that. I used to wonder about the same thing.

damyanp:

damyanp:

Issue #5306, opened 2023-06-20
[SPIR-V] DXC emits OpCapability RayQueryKHR even when not requested


natevm:
I'm running into an unfortunate bug where DXC requests RayQueryKHR even when ray queries are never used. 

This is an issue, because NVIDIA cards without RT cores using their software fallback _do not support_ ray queries. 

As a consequence, no ray tracing shaders compiled by DXC are compatible with GTX cards or NVIDIA cards like the A100 without RT cores. 

Further, we get incorrect compilation errors stating `generated SPIR-V is invalid: 1st operand of Capability: operand RayQueryKHR(4472) requires one of these extensions: SPV_KHR_ray_query`

natevm:
I believe it was this pull request that caused this regression.

https://github.com/microsoft/DirectXShaderCompiler/pull/4151

natevm:
Here's the A100 GPU in question. We want to use these A100s with Vulkan at ANL on our Polaris supercomputer
https://vulkan.gpuinfo.org/displayreport.php?id=21882#extensions

If you look at this extension survey, VK_KHR_ray_tracing has more coverage than VK_KHR_ray_query, so the assumption made by @sudonatalie in PR #4151 is not true. 

https://vulkan.gpuinfo.org/listextensions.php

We also have some issues reported by end users of DXC, eg @ashpil, that this is causing incompatibilities with Pascal cards like their GTX 1080.

natevm:
This is the logic that I'd propose we revert. The original author of this PR suggested a spirv-opt pass

https://github.com/microsoft/DirectXShaderCompiler/pull/4151/commits/ce31e10902732c8cd8f6f3b5b78699110afddb2b#diff-44e37c9720575ff94b7842b9ceb70a87fe72486d2b5da2e3828512dc64a352e6R217-R222

natevm:
@ashpil In retrospect, it seems that the ray query extension is only really added if you have both a non-RT-pipeline entrypoint in the same file that a `RaytracingAccelerationStructure` handle is used. So, the current behavior in DXC might be right actually...

Keenuts:
Hello!

Just FYI, I think this new spirv-opt pass will solve this issue (https://github.com/KhronosGroup/SPIRV-Tools/pull/5278).

>  The original author of this PR suggested a spirv-opt pass. I'm not sure that's such a good idea, since then we're dependent on spirv-opt to add more tooling rather than addressing the issue here.

[...]

PR #5307, opened 2023-06-20
[SPIR-V] Correctly request RayQueryKHR only when requested

natevm:
This PR fixes a bug where ray query compatibility was required whenever acceleration structure compatibility is requested. 

More specifically, this fixes issue #5306 

In practice, not all GPUs capable of ray tracing supporting support ray queries, and not all support the ray tracing pipeline. Many AMD GPUs support only khr_ray_query, while many NVIDIA cards support only khr_ray_tracing_pipeline. 

Currently, if you own a Pascal card, a Volta, Turing or Ampere cards like the 1660 ti, then the current logic in DXC prevents RT from working on these platforms. At ANL, we have A100s in server systems like Polaris where we want to use Vulkan ray tracing + HLSL, but these also only support khr_ray_tracing_pipeline.

Macs through MoltenVK where we're working to add a translation might also only support the ray tracing pipeline due to Metal RT constraints. I'm not entirely sure here, but it's possible.

On the AMD side of things, its often the case that only khr_ray_query is supported.

This change removes the false assumption that the presence of an acceleration structure type implies ray queries or ray tracing pipeline. Instead, we assume that the user will request either the ray query and/or the ray tracing pipeline extensions during compilation.

Down the road, we could request that spirv-opt add a pass to remove any requested ray query extension if no queries are used, or remove the ray pipeline extension if that's not used, but until that's added to spirv-opt, I'd argue we should accept this PR in favor of adding support for all these NVIDIA cards where RT core acceleration isn't available but software fallbacks in RT core are.

### DIFF
--- a/tools/clang/lib/SPIRV/CapabilityVisitor.cpp
+++ b/tools/clang/lib/SPIRV/CapabilityVisitor.cpp
@@ -214,16 +214,6 @@ void CapabilityVisitor::addCapabilityForType(const SpirvType *type,
     for (auto field : structType->getFields())
       addCapabilityForType(field.type, loc, sc);
   }
-  // AccelerationStructureTypeNV 
-  else if (isa<AccelerationStructureTypeNV>(type)) {
-    if (featureManager.isExtensionEnabled(Extension::NV_ray_tracing)) {
-      addCapability(spv::Capability::RayTracingNV);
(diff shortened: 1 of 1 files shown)


natevm:
In retrospect, it seems this is a special issue with the way that I handle multiple entrypoints in the same file with my particular project. The Ray Query extension seems to only really be added when there is the presence of an Acceleration Structure handle _with_ entry points like vertex, pixel, etc.

PR #5312, opened 2023-06-22
[lit] Support things like rm/echo as lit built-in commands for windows.

python3kgae:
Support things like rm/echo in python for windows like upstream lit built-in commands.

Also replace find/cmp/grep which not in upstream lit built-in commands.

Use rm to replace find -delete
diff to replace cmp
FileCheck and count to replace grep

NOTE: ls is not lit builtin in upstream. Need to better solution for it.


python3kgae:
Fixed by https://github.com/microsoft/DirectXShaderCompiler/pull/5331

PR #5314, opened 2023-06-22
Disambiguate RecordDispatchGrid type name for GCC

pow2clk:
GCC failed when faced with a variable of both type and name RecordDispatchGrid. Adding struct to the type to make it clear eliminates the error


jenatali (approved):
Assuming all the compilers are happy with this, it looks good to me.

python3kgae (approved):

include/dxc/DxilContainer/RDAT_LibraryTypes.inl
```diff
@@ -470,7 +470,7 @@ RDAT_STRUCT_TABLE(NodeShaderIOAttrib, NodeShaderIOAttribTable)
     RDAT_UNION_ELIF(RecordSizeInBytes, getAttribKind() == hlsl::RDAT::NodeAttribKind::RecordSizeInBytes)
       RDAT_VALUE(uint32_t, RecordSizeInBytes)
     RDAT_UNION_ELIF(RecordDispatchGrid, getAttribKind() == hlsl::RDAT::NodeAttribKind::RecordDispatchGrid)
-      RDAT_RECORD_VALUE(RecordDispatchGrid, RecordDispatchGrid)
+      RDAT_RECORD_VALUE(struct RecordDispatchGrid, RecordDispatchGrid)
```
tex3d, line 473: Could we change the macro instead?

pow2clk: done

tex3d (approved):
Looks good!

include/dxc/DxilContainer/RDAT_Macros.inl
```diff
@@ -72,7 +72,7 @@
   #define RDAT_UNION_END()                    };
   #define RDAT_RECORD_REF(type, name)         RecordRef<type> name;
   #define RDAT_RECORD_ARRAY_REF(type, name)   RecordArrayRef<type> name;
-  #define RDAT_RECORD_VALUE(type, name)       type name;
+  #define RDAT_RECORD_VALUE(type, name)       struct type name;
```
tex3d, line 75: Oh, I think there may be another place this should go, but I guess it's not hit by anything right now.  Same change under `DEF_RDAT_TYPES_BASIC_STRUCT` block.

PR #5317, opened 2023-06-23
Made linker preserve nops for Od.

adam-yang:
For `-Od`/`-O0` compilations, the compiler generates nop instructions in the form of:

```
@dx.nothing.a = internal constant [1 x i32] zeroinitializer
...
%0 = load i32, i32* getelementptr inbounds ([1 x i32], [1 x i32]* @dx.nothing.a, i32 0, i32 0)
```
This instruction is intended to be a complete no-op, and backends that recognizes this pattern can turn it into proper hardware specific no-op instructions for debug purposes.

When linking DXIL libraries, the passes that run end up removing these nops. This change adds a new pass `DxilReinsertNops` to run at the beginning of linking to read the `LoadInst`s representing nops and turn them back into calls of `dx.noop`, before turning back into loads again at the end.

### DIFF
--- a/lib/HLSL/DxilNoops.cpp
+++ b/lib/HLSL/DxilNoops.cpp
@@ -95,6 +95,7 @@
 #include "llvm/Transforms/Utils/Local.h"
 #include "llvm/IR/DIBuilder.h"
 #include "llvm/Support/raw_os_ostream.h"
+#include "llvm/ADT/StringRef.h"
 #include "dxc/DXIL/DxilMetadataHelper.h"
 #include "dxc/DXIL/DxilConstants.h"
 #include "dxc/HLSL/DxilNoops.h"
@@ -581,6 +582,78 @@ Pass *llvm::createDxilRewriteOutputArgDebugInfoPass() {
 }
 
 INITIALIZE_PASS(DxilRewriteOutputArgDebugInfo, "dxil-rewrite-output-arg-debug-info", "Dxil Rewrite Output Arg Debug Info", false, false)
+
+//==========================================================
+// Reader pass
+//
+
+namespace {
+
+class DxilReinsertNops : public ModulePass {
+public:
+  static char ID;
+
+  DxilReinsertNops() : ModulePass(ID) {
+    initializeDxilReinsertNopsPass(*PassRegistry::getPassRegistry());
+  }
+
+  static bool IsLegitimateNothingGlobalVar(StringRef Name) {
+    if (1 != Name.count(hlsl::kNothingName))
+      return false;
+    size_t Loc = Name.find(hlsl::kNothingName);
+    StringRef Prefix = Name.substr(0, Loc);
+    StringRef Suffix = Name.substr(Loc+Name.size());
+    if (!Prefix.empty() && !Prefix.endswith(".")) {
+      return false;
+    }
+    if (!Suffix.empty() && !Suffix.startswith(".")) {
+      return false;
+    }
+    return true;
+  }
+
+  bool runOnModule(Module& M) override {
+    bool Changed = false;
+    for (GlobalVariable &GV : M.globals()) {
+      if (!IsLegitimateNothingGlobalVar(GV.getName()))
+        continue;
+
+      const bool IsValidType = GV.getValueType()->isArrayTy() &&
+                               GV.getValueType()->getArrayElementType() ==
+                                   Type::getInt32Ty(M.getContext()) &&
+                               GV.getValueType()->getArrayNumElements() == 1;
+      if (!IsValidType)
+        return false;
+
+      for (User *GVU : GV.users()) {
+        ConstantExpr *CE = dyn_cast<ConstantExpr>(GVU);
+        if (!CE || CE->getOpcode() != Instruction::GetElementPtr)
+          continue;
+
+        for (auto it = CE->user_begin(), end = CE->user_end(); it != end;) {
+          User *U = *(it++);
+          LoadInst *LI = dyn_cast<LoadInst>(U);
+          if (!LI)
+            continue;
+          InsertNoopAt(LI);
+          LI->eraseFromParent();
+          Changed = true;
+        }
+      }
+    }
+
+    return Changed;
+  }
+  StringRef getPassName() const override { return "Dxil Reinsert Nops"; }
+};
+
+char DxilReinsertNops::ID;
+}
+Pass *llvm::createDxilReinsertNopsPass() {
+  return new DxilReinsertNops();
+}
+
+INITIALIZE_PASS(DxilReinsertNops, "dxil-reinsert-nops", "Dxil Reinsert Nops", false, false)
 
 
 //==========================================================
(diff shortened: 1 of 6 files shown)


python3kgae:
Could we just not add DCE if link with Od?

adam-yang:
> Could we just not add DCE if link with Od?

Just tried it, but a bunch of other passes also do DCE or simplification that get rid of the nops (inlining, and SROA). Disabling even just SROA causes a bunch of test failures. The `dx.noop` calls are already tolerated by all these passes in normal compilation passes, so I think it's a much less risky change than disabling passes in the linker.

lib/HLSL/DxilNoops.cpp
```diff
@@ -582,6 +583,93 @@ Pass *llvm::createDxilRewriteOutputArgDebugInfoPass() {
...
+  //   <library_name>.dx.nothing.a.<another_thing>
+  //
+  // This routine looks for the "dx.nothing.a" string inside of it, and as long
+  // as it's used in the expected way:
+  // 
+  // %0 = load i32, i32* getelementptr inbounds ([1 x i32], [1 x i32]* @dx.nothing.a, i32 0, i32 0)
+  // 
+  // ...it is deemed a valid nop.
+  //
+  static bool IsLegalNothingVarName(StringRef Name) {
```
dmpots, line 612: I think we have a couple of other special globals like this. 

1. Lifetime markers prior to 6.6
2. The dx.break global

Do they need this special handling for linking as well?

Pretty sure there is one for lifetime markers prior to 6.6,

bogner:
Do the nops need to be represented by loads from a global? I feel like this would all be much easier if they were treated as a call to an opaque function that claims it has side-effects. This would naturally limit code motion and avoid the issue of DCE removing it, and then the backends could just provide a body for the function that does the NOP, no?

adam-yang:
> Do the nops need to be represented by loads from a global? I feel like this would all be much easier if they were treated as a call to an opaque function that claims it has side-effects. This would naturally limit code motion and avoid the issue of DCE removing it, and then the backends could just provide a body for the function that does the NOP, no?

When this was added, we wanted to do so without having to rev the DXIL version. I don't think we ever seriously considered adding a new dx intrinsic just for this.

lib/HLSL/DxilNoops.cpp
adam-yang: Yeah we might need similar handling for `@dx.break`, but I'm not sure what lifetime-marker needs. I think it's just a `store undef`.

lib/HLSL/DxilNoops.cpp
```diff
@@ -582,6 +583,93 @@ Pass *llvm::createDxilRewriteOutputArgDebugInfoPass() {
...
+  static char ID;
+
+  DxilReinsertNops() : ModulePass(ID) {
+    initializeDxilReinsertNopsPass(*PassRegistry::getPassRegistry());
+  }
+
+  // In various linking scenarios, the dx.nothing.a variable might be prefixed
+  // and/or suffixed with something:
+  //
+  //   <library_name>.dx.nothing.a.<another_thing>
```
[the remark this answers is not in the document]
adam-yang: We don't :(  I actually didn't anticipate linking to add prefixes and suffixes like this. Another option is to have the linker not duplicate these variables (and therefore no hacky string searching, just one strcmp). I think using global variable name is fine, since it's not quite as transient and unreliable as local value names.

tools/clang/unittests/HLSL/LinkerTest.cpp
```diff
@@ -484,6 +485,29 @@ TEST_F(LinkerTest, RunLinkToLib) {
   Link(L"", L"lib_6_3", pLinker, {libName, libName2}, {"!llvm.dbg.cu"}, {}, option);
 }
 
+TEST_F(LinkerTest, RunLinkToLibOdNops) {
```
python3kgae, line 488: Could this be a FileCheck test instead of unit test?

python3kgae:
> > Could we just not add DCE if link with Od?
> 
> Just tried it, but a bunch of other passes also do DCE or simplification that get rid of the nops (inlining, and SROA). Disabling even just SROA causes a bunch of test failures. The `dx.noop` calls are already tolerated by all these passes in normal compilation passes, so I think it's a much less risky change than disabling passes in the linker.

Are these test fails pattern match fail or other kind of failure?

adam-yang:
> > > Could we just not add DCE if link with Od?
> > 
> > 
> > Just tried it, but a bunch of other passes also do DCE or simplification that get rid of the nops (inlining, and SROA). Disabling even just SROA causes a bunch of test failures. The `dx.noop` calls are already tolerated by all these passes in normal compilation passes, so I think it's a much less risky change than disabling passes in the linker.
> 
> Are these test fails pattern match fail or other kind of failure?

I think they're all compilation failures from `MatrixBitcastLowerPass`, I didn't look too closely. But in any case, I don't think we should disable DCE and SimplifyInst for Od. We run them for normal Od compiles too.

python3kgae:
> > > > Could we just not add DCE if link with Od?
> > > 
> > > 
> > > Just tried it, but a bunch of other passes also do DCE or simplification that get rid of the nops (inlining, and SROA). Disabling even just SROA causes a bunch of test failures. The `dx.noop` calls are already tolerated by all these passes in normal compilation passes, so I think it's a much less risky change than disabling passes in the linker.
> > 
> > 
> > Are these test fails pattern match fail or other kind of failure?
> 
> I think they're all compilation failures from `MatrixBitcastLowerPass`, I didn't look too closely. But in any case, I don't think we should disable DCE and SimplifyInst for Od. We run them for normal Od compiles too.

I see.
Finally got why you're doing this.
We should add dx.noop to future DXIL spec.

lib/HLSL/DxilNoops.cpp
dmpots: I just noticed another global `dx.preserve.value` that looks related to debugging

https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/HLSL/DxilNoops.cpp#L74

Should we handle that here as well?

lib/HLSL/DxilNoops.cpp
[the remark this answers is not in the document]
dmpots: Ok, so it looks like the expectation is that we can find all the `dx.noop` values by looking for a global with the `hlsl::kNothingName`. In that case I think this pass seems fine.

It also seems like the global variable is more of a "fallback" path for lower shader models. Seems like it would be good to add official support for `dx.noop` to dxil.

dmpots (approved):

lib/HLSL/DxilNoops.cpp
adam-yang: Those were for an experimental way to do Od and also preserve dbg.value order. It turned out to produce way too many values and we have better solutions now anyway, so they should be removed. I just haven't had the chance to yet.

python3kgae (approved):
LGTM.
We need to create a task about add dx.noop to future DXIL spec.

PR #5320, opened 2023-06-23
[lit] [SPIRV} Copy legal-examples to lit

python3kgae:
Copy SPIRV tests under legal-examples and modify to work with lit.

This is first PR for moving SPIRV test to lit.
The reason to copy instead of moving the tests is to make sure not lost coverage before all tests switch to lit. And update all tests in one PR will be a PR with 1000 files.

Once all tests under CodeGenSPIRV is copied, we can replace the old tests with updated lit tests.


Keenuts (approved):

s-perron (approved):

python3kgae:
> 

You can continue add new test to CodegenSPIRV.
Once everything is updated, CodeGenSPIRV_lit will be rename to CodeGenSPIRV. Old CodeGenSPIRV will be deleted.
That might be a big PR which has more than 1000 files been updated :(

Maybe the best way is to allow test to be part for lit, part as is. Then we could switch one by one.

s-perron:
Okay. I'll keep adding tests the same way. Let us know when to switch.

python3kgae:
> Okay. I'll keep adding tests the same way. Let us know when to switch.

Finally taef is hide behind lit.
And lit FileCheck shell test could replace other FileCheck tests.
You could start to add new tests as lit FileCheck test now.

PR #5321, opened 2023-06-23
Cherry-Pick: WinPIX: Pass to log shader invocations + regression fix (#5301, #5315)

jeffnn:
Recent small regression: the WinPIX shader debug pass should always write to UAV 0 (that's the only one that's correctly setup by WinPIX).

### DIFF
--- a/lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
+++ b/lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
@@ -1,0 +1,234 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DxilPIXDXRInvocationsLog.cpp                                              //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#include "dxc/DXIL/DxilFunctionProps.h"
+#include "dxc/DXIL/DxilOperations.h"
+#include "dxc/DXIL/DxilUtil.h"
+
+#include "dxc/DXIL/DxilInstructions.h"
+#include "dxc/DXIL/DxilModule.h"
+#include "dxc/DxilPIXPasses/DxilPIXPasses.h"
+
+#include "llvm/IR/InstIterator.h"
+#include "llvm/IR/PassManager.h"
+#include "llvm/Support/FormattedStream.h"
+#include "llvm/Transforms/Utils/Local.h"
+
+#include "PixPassHelpers.h"
+
+using namespace llvm;
+using namespace hlsl;
+using namespace PIXPassHelpers;
+
+class DxilPIXDXRInvocationsLog : public ModulePass {
+    uint64_t m_MaxNumEntriesInLog = 1;
+public:
+  static char ID;
+  DxilPIXDXRInvocationsLog() : ModulePass(ID) {}
+  StringRef getPassName() const override {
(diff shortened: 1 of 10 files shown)


tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -5197,4 +5221,38 @@ void ASMain()
   ValidateStructMember(2, L"bigOne", 24*8);
 }
 
+TEST_F(PixTest, DxilPIXDXRInvocationsLog_SanityTest) {
+
+    const char* source = R"x(
+struct MyPayload
+{
```
python3kgae, line 5228: Is it possible to create FileCheck test instead of unit test?

lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+void DxilPIXDXRInvocationsLog::applyOptions(PassOptions O) {
+    GetPassOptionUInt64(O, "maxNumEntriesInLog", &m_MaxNumEntriesInLog, 1); // Use a silly default value. PIX should set a better value here.
+}
+
+bool DxilPIXDXRInvocationsLog::runOnModule(Module &M) {
+
+  DxilModule &DM = M.GetOrCreateDxilModule();
+  LLVMContext &Ctx = M.getContext();
+  OP *HlslOP = DM.GetOP();
```
python3kgae, line 65: Shall we ignore non lib profiles for DxilPIXDXRInvocationsLog pass?

jeffnn: Since DXR requries libs, that seems ok?

tools/clang/unittests/HLSL/PixTest.cpp
jeffnn: Yes. Are we concerned about the running time for unit tests or something?

jeffnn: (Meaning, is it OK to have FileCheck tests AND these unit tests? Or do you mean it's better to have just filecheck tests?)

jeffnn: See:

https://github.com/microsoft/DirectXShaderCompiler/pull/5322

python3kgae: If the FileCheck and unit test are testing same thing, it is better to have just filecheck tests.

lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
python3kgae: We can return false for non lib profiles.

tools/clang/unittests/HLSL/PixTest.cpp
jeffnn: I've added a file-check test

lib/DxilPIXPasses/DxilPIXDXRInvocationsLog.cpp
jeffnn: Fair enough. I can make that fix in main, but since PIX will never invoke this pass on non-lib modules, it's little more than documentation.

jeffnn: https://github.com/microsoft/DirectXShaderCompiler/pull/5323

jeffnn: Fix has been cherry-picked here

adam-yang (approved):

PR #5323, opened 2023-06-25
PIX: don't report modification from DXRInvocationsLog pass for non-DXR-hit libs

jeffnn:
PIX may well have libs in a state object that have no DXR hit groups.


python3kgae (approved):

tex3d (approved):
Looks good, but I'm reminded that the DxilModue method name `GetExportedFunctions` is a misleading name.

PR #5324, opened 2023-06-26
Propagate bitwidth for BitCast in LiteralTypeVisitor

### LINKED ISSUE #5319: fatal error: generated SPIR-V is invalid: Invalid word count: OpConstant starting at word 108 says it has 4 words, but found 5 words instead.
```
$ cat repro.hlsl 

RWByteAddressBuffer outp : register(u1);
[numthreads(1, 1, 1)]
void entrypoint()
{ 
  uint val = 0;
  uint64_t res = val ? 1 : 0;    
  outp.Store(0, res);
}
```

```
$ dxc   -fspv-target-env=vulkan1.1 -spirv -E entrypoint -T cs_6_2 -no-warnings  repro.hlsl
fatal error: generated SPIR-V is invalid: Invalid word count: OpConstant starting at word 108 says it has 4 words, but found 5 words instead.
```

s-perron:
For a BitCast instruction, the bitwidth of the operand must be the same
as the bitwidth of the result type. That requirement is coded into the
LiteralTypeVisitor, so that we do not get validation errors.

Fixes #5319.

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/select.long.lit.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/select.long.lit.hlsl
@@ -1,6 +1,7 @@
 // RUN: %dxc -T ps_6_0 -HV 2018 -E main
 
-// the literal type.
+// Check that the literals get a 64-bit type, and the result of the select is
+// then cast to an unsigned 64-bit value.
 void foo(uint x) {
   // CHECK: %foo = OpFunction
   // CHECK: [[cond:%\d+]] = OpULessThan %bool {{%\d+}} %uint_64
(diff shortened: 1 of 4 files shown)


dnovillo (approved):
For my own edifications: the two new conditionals at the top are for short-circuting non-lit types, right?

s-perron:
> For my own edifications: the two new conditionals at the top are for short-circuting non-lit types, right?

Correct. If either if those checks fail, `canDeduceTypeFromLitType` will return false, and causing `tryToUpdateInstLitType` to do nothing.

tools/clang/test/CodeGenSPIRV/select.long.lit.hlsl
```diff
@@ -0,0 +1,29 @@
+// RUN: %dxc -T ps_6_0 -HV 2018 -E main
+
+// Check that the literals get a 64-bit type, and the result of the select is
+// then cast to an unsigned 64-bit value.
+void foo(uint x) {
```
Keenuts, line 5: nit: maybe add the same tests for the enable-16bit-types flag/types?

Keenuts (approved):
LGTM. Just test coverage?

tools/clang/test/CodeGenSPIRV/select.long.lit.hlsl
s-perron: done.

PR #5329, opened 2023-06-27
[NFC] Cherry-picking some changes to prepare for ReverseIterator support

bfavela:
There are a number of changes required to get ReverseIterator support into DXC. Rather than attempting to do it all as one large PR, it makes more sense to break it up into smaller PRs of 2-3 changes.

These first two are just some small refactoring in the SmallPtrSet classes and should not change any behavior.

There was one small change in the DXC codebase that was necessary to bring forward which deviates from the original cherry-pick slightly (will mark it in a comment).

### DIFF
--- a/lib/Support/SmallPtrSet.cpp
+++ b/lib/Support/SmallPtrSet.cpp
@@ -35,22 +35,7 @@ void SmallPtrSetImplBase::shrink_and_clear() {
 }
 
 std::pair<const void *const *, bool>
-SmallPtrSetImplBase::insert_imp(const void *Ptr) {
-  if (isSmall()) {
-    // Check to see if it is already in the set.
-    for (const void **APtr = SmallArray, **E = SmallArray+NumElements;
-         APtr != E; ++APtr)
-      if (*APtr == Ptr)
-        return std::make_pair(APtr, false);
-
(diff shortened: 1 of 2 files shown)


lib/Support/SmallPtrSet.cpp
```diff
@@ -257,10 +218,14 @@ void SmallPtrSetImplBase::CopyFrom(const SmallPtrSetImplBase &RHS) {
 
 void SmallPtrSetImplBase::MoveFrom(unsigned SmallSize,
                                    SmallPtrSetImplBase &&RHS) {
-  assert(&RHS != this && "Self-move should be handled by the caller.");
-
   if (!isSmall())
     delete[] CurArray; // HLSL Change: Use overridable operator delete
```
bfavela, line 222: This line is the only change from upstream.

bogner (approved):
lgtm

adam-yang (approved):

llvm-beanz (approved):
Just as an administrative piece here: these changes are not functional refactoring. The SmappPtrSet type is tested extensively here:

https://github.com/microsoft/DirectXShaderCompiler/blob/main/unittests/ADT/SmallPtrSetTest.cpp

No additional tests are required for this change because it is a non-functional change and the changed code is sufficiently tested in the codebase.

PR #5332, opened 2023-06-27
Fixes to build with C++20

ben-clayton:
Add another template typename for pointer <-> ilist_iterator comparison operators. C++20 considers these ambiguous for the same reasons outlined in https://github.com/godotengine/godot/issues/54058

The rest of the changes are general bug fixes which are now caught by the compiler.

### DIFF
--- a/include/llvm/ADT/ilist.h
+++ b/include/llvm/ADT/ilist.h
@@ -248,20 +248,20 @@ void operator+(ilist_iterator<T>,int) = delete;
 
 // operator!=/operator== - Allow mixed comparisons without dereferencing
 // the iterator, which could very likely be pointing to end().
-template<typename T>
-bool operator!=(const T* LHS, const ilist_iterator<const T> &RHS) {
+template<typename T, typename U>
+bool operator!=(const T* LHS, const ilist_iterator<const U> &RHS) {
   return LHS != RHS.getNodePtrUnchecked();
 }
-template<typename T>
-bool operator==(const T* LHS, const ilist_iterator<const T> &RHS) {
+template<typename T, typename U>
+bool operator==(const T* LHS, const ilist_iterator<const U> &RHS) {
   return LHS == RHS.getNodePtrUnchecked();
 }
-template<typename T>
-bool operator!=(T* LHS, const ilist_iterator<T> &RHS) {
+template<typename T, typename U>
+bool operator!=(T* LHS, const ilist_iterator<U> &RHS) {
   return LHS != RHS.getNodePtrUnchecked();
 }
-template<typename T>
-bool operator==(T* LHS, const ilist_iterator<T> &RHS) {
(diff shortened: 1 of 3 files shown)


ben-clayton:
Friendly ping!

ben-clayton:
Another gentle ping.

amaiorano:
@llvm-beanz Can you please take a look? @ben-clayton is our Tech Lead of my team (Tint, the compiler used by Dawn). This fixes builds when C++20 is enabled. ATM, we're able to specify building modules with C++17, but eventually all of Chrome expects to be built with C++20.

amaiorano (approved):

s-perron (commented):
LGTM but all of the changes are in the DXIL code some an owner of that code should be the second reviewer.

s-perron:
@tex3d @llvm-beanz Could one of you review this? Thanks.

include/llvm/ADT/ilist.h
```diff
@@ -248,20 +248,20 @@ void operator+(ilist_iterator<T>,int) = delete;
 
 // operator!=/operator== - Allow mixed comparisons without dereferencing
 // the iterator, which could very likely be pointing to end().
-template<typename T>
-bool operator!=(const T* LHS, const ilist_iterator<const T> &RHS) {
+template<typename T, typename U>
```
llvm-beanz, line 251: Can you wrap your changes here in `// HLSL Change` comments. We try to mark out the core parts of LLVM that we've modified.

llvm-beanz (commented):
One minor nit. Otherwise good.

include/llvm/ADT/ilist.h
ben-clayton: Will do. Out of curiosity, what's the purpose of the `// HLSL Change` comments? I assumed they were to aid in updating DXC to newer versions of LLVM, but that seems unlikely now. Do these comments help you with the merge into mainline LLVM?

ben-clayton: Done

llvm-beanz: There are two reasons these comments help. One is that we do cherry-pick back selective fixes from newer LLVM versions, so marking out the comments allow us to identify where we changed something in a merge. The other is that I frequently diff DXC's code against LLVM 3.7 to help narrow down the causes of bugs. More often than not, the bugs in some parts of DXC are caused by the way changes were introduced into the LLVM-inherited code, so having the markers to denote those areas is extremely usefu.

llvm-beanz (approved):

include/llvm/ADT/ilist.h
ben-clayton: Good to know. Thanks for the additional info!

ben-clayton:
Many thanks!

PR #5334, opened 2023-06-27
[NFC] Add dxcutil::ComputeSerializeDxilFlags to share code with linker

python3kgae:
Add dxcutil::ComputeSerializeDxilFlags to share code with linker

Use refactored ComputeSerializeDxilFlags in DxcCompiler::Compile()


bob80905 (approved):
I think the removed comment was helpful, but otherwise LGTM.

llvm-beanz (commented):
Can we add unit tests for this? It is a standalone function, seems like we should be able to.

llvm-beanz (changes requested):
This refactoring enables writing unit tests for code that is not previously tested. Please add appropriate testing:
https://github.com/microsoft/DirectXShaderCompiler/blob/main/CONTRIBUTING.md#contributing-code-and-content

python3kgae:
> Can we add unit tests for this? It is a standalone function, seems like we should be able to.

We have 18 tests will fail if the SerialeDxilFlags is wrong. Is unit test necessary in this case?

llvm-beanz:
> We have 18 tests will fail if the SerialeDxilFlags is wrong. Is unit test necessary in this case?

Yes. We should have a unit test for it.

llvm-beanz:
If instead of putting this function into DxcUtil, you move it into a source under DXC/Support, we could have a unit test binary that directly links the DXCSupport static archive. That would allow us to test this without altering the exposed symbol interface for dxcompiler.

tools/clang/unittests/HLSL/OptionsTest.cpp
```diff
@@ -380,3 +384,436 @@ TEST_F(OptionsTest, TestPreprocessOption) {
...
+  EXPECT_EQ(
+      static_cast<uint32_t>(hlsl::options::ComputeSerializeDxilFlags(dxcOpts)),
+      ExpectFlags);
+}
+
+TEST_F(OptionsTest, SerializeDxilFlags) {
+  VerifySerializeDxilFalgs(
+      "-T lib_6_3 input.hlsl -Qstrip_rootsignature -Qkeep_reflect_in_dxil "
+      "-Qstrip_reflect -Zss -Zs",
+      static_cast<uint32_t>(SerializeDxilFlags::StripRootSignature) |
```
dmpots, line 413: I'm finding it hard to understand the intent of each of these tests and determining the overall coverage.

One thing that confuses me is the repeated 

```
    static_cast<uint32_t>(SerializeDxilFlags::None) |
    static_cast<uint32_t>(SerializeDxilFlags::None) |
```
in some of the tests.

Is this trying to test all of the combinations? Its not clear to me that is the case and if not why were these specific examples chosen?

Maybe we could use a table driven approach here to make it a bit more concise about what combinations we are testing and why?

```
CombineFlags(ArrayRef<SerializeDxilFlags> flags)
{
   uint32_t combined = 0;
   for (auto f : flags) combined |= static_cast<uint32_t>(f);
  return combined;
}

using F=SerializeDxilFlags;
struct Test {char *flags, uint32_t expected};
Test Tests[] = {
  {"",
    CombineFlags({
        F::StripReflectionFromDxilPart,
        F::IncludeReflectionPart}) },
  {"-Qstrip_rootsignature -Qkeep_reflect_in_dxil -Qstrip_reflect -Zss -Zs",
    CombineFlags({
        F::StripRootSignature,
        F::IncludeDebugNamePart,
        F::DebugNameDependOnSource}) },
}

for (const Test &T : Tests) {
    string cmd = "-T lib_6_3 input.hlsl" + T.flags;
    VerifySerializeDxilFlags(cmd, T.expected)
}

Comment (python3kgae) [reply tools/clang/unittests/HLSL/OptionsTest.cpp]: Fixed.

File: tools/clang/unittests/HLSL/OptionsTest.cpp [hunk 2]
```diff
@@ -380,3 +384,436 @@ TEST_F(OptionsTest, TestPreprocessOption) {
   VerifyPreprocessOption("/T ps_6_0 -P out.pp input.hlsl", "out.pp", Warning);
   VerifyPreprocessOption("/T ps_6_0 input.hlsl -P out.pp ", "out.pp", Warning);
 }
+
+static void VerifySerializeDxilFalgs(llvm::StringRef command,
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/unittests/HLSL/OptionsTest.cpp
```diff
@@ -380,3 +384,219 @@ TEST_F(OptionsTest, TestPreprocessOption) {
...
+
+struct SerializeDxilFlagsTest {
+  const char *command;
+  uint32_t flags;
+};
+
+using F = SerializeDxilFlags;
+
+TEST_F(OptionsTest, SerializeDxilFlags) {
+  // Test cases for SerializeDxilFlags
```
dmpots, line 426: I think it would be good to add a comment here about why these flag combinations were chosen. There is a lot of tests here, which is good, but its not clear how/why these ones in particular are chosen. I'm failing to see a comprehensive pattern just by looking at them.

dmpots (approved):

tools/clang/unittests/HLSL/OptionsTest.cpp
```diff
@@ -380,3 +384,233 @@ TEST_F(OptionsTest, TestPreprocessOption) {
...
+    {"-Qstrip_rootsignature -Qkeep_reflect_in_dxil -Qstrip_reflect -FdDbgName.pdb -Qembed_debug -Zi",
+    CombineFlags({F::IncludeDebugNamePart, F::IncludeDebugInfoPart, F::StripRootSignature})
+    },
+    {"-Qstrip_rootsignature -Qkeep_reflect_in_dxil -Qstrip_reflect ",
+    CombineFlags({F::StripRootSignature})
+    },
+    {"-Qstrip_rootsignature -Qkeep_reflect_in_dxil  -Zss -Zs",
+    CombineFlags({F::IncludeDebugNamePart, F::IncludeReflectionPart, F::DebugNameDependOnSource, F::StripRootSignature})
+    },
+    {"-Qstrip_rootsignature -Qkeep_reflect_in_dxil  -Zsb",
```
llvm-beanz, line 466: Since we are imminently going to be clang-formatting things, it would be nice to clang-format the new code coming in.

llvm-beanz (approved):
Other than a request to clang-format the new code. This looks good.

PR #5335, opened 2023-06-27
[NFC] Add more information on testing MDVals insertion in EmitDxilFunctionProps

bob80905:
A recent change from a recent PR includes changes to how the MDVals data structure inside of EmitDxilFunctionProps gets elements (push_back instead of index assignment).
However, there was no accompanying test for these changes, because HLSLFileCheck\hlsl\workgraph\called_function_arg_nodeoutput.hlsl exercised this code path and failed without the MDVals changes.
This PR adds some extra context to the above test so that the extra purpose the above test serves isn't lost with time.

The test passes with the MDVals pushback changes, and fails with the original valIdx index assignments.


tools/clang/test/HLSLFileCheck/hlsl/workgraph/test_EmitDxilFunctionProps_MDVals.hlsl
```diff
@@ -0,0 +1,30 @@
+// RUN: %dxc -fcgl -T lib_6_8 %s | FileCheck -check-prefix=CHECK_FCGL %s
+//
+
+// This test will fail if MDVals in DxilMetadataHelper.cpp gets assigned out of bounds.
+// The test is primarily from called_function_arg_nodeoutput.hlsl
+
+// Verify that the correct parameter, the NodeOutputRecord parameter, gets the cast
```
python3kgae, line 7: Do we still verify parameters?

bob80905: tools\clang\test\HLSLFileCheck\hlsl\workgraph\called_function_arg_nodeoutput.hlsl does still verify parameters, and is the test file where this new test file comes from. Although this new test doesn't necessarily need to verify that the parameter gets the cast, there is an implication that if the cast is successful, then the buffer didn't get assigned out of bounds. 
So, there might be a better way to write a check for this test, but essentially any test that checks for successful compilation here is sufficient.

python3kgae (approved):

tools/clang/test/HLSLFileCheck/hlsl/workgraph/test_EmitDxilFunctionProps_MDVals.hlsl
```diff
@@ -0,0 +1,29 @@
+// RUN: %dxc -fcgl -T lib_6_8 %s | FileCheck -check-prefix=CHECK_FCGL %s
```
tex3d, line 1: Isn't this just a partial/equivalent copy of the `called_function_arg_nodeoutput.hlsl` test?  This doesn't change verification to check the metadata values, which the test name implies either.  Is there any added value of this new test over just running the other test?

bob80905: This test has no added value when compared to the other test. It was originally intended to just check for successful compilation, since this specific compilation path would have caused an out-of-bounds assignment in the older code. 
One improvement I could make is to remove the checks and add a CHECK: !36 to prove that there are more than 35 metadata values in the MDVals data structure and no out of bounds behavior is happening.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/test_EmitDxilFunctionProps_MDVals.hlsl
```diff
@@ -0,0 +1,28 @@
+// RUN: %dxc -fcgl -T lib_6_8 %s | FileCheck %s
+//
+// Verify that the MDVals data structure inside EmitDxilFunctionProps in 
+// DxilMetadataHelper.cpp doesn't get any out-of-bounds assignments.
+
+// CHECK: !36
```
tex3d, line 6: What is this verifying?  I don't think this is sufficient, and numeric metadata nodes are just unnamed nodes with automatically assigned indices.  It could be anything.  In the future, metadata could be written and this would probably not be checking what you think it is.

bob80905: Originally, when the MDVals data structure was capped at 35 entries, and entries were entered via index assignment, an out-of-bounds assignment would occur because this specific path would require more than 35 values to be entered into MDVals. When the data structure was changed (as well as the insertion method to push_back), there was no accompanying test. Ideally, we would want a regression test that passes when push_back is used, and fails in the old code where the size was capped and index assignment was used. 
I thought this test would be sufficient because it verifies that the data structure has at least 36 entries, which was impossible with the old code, since an out of bounds assignment would have prevented successful compilation.

tex3d: The indices of metadata nodes aren't 1:1 with the indices of that particular array.  These indices are automatically assigned for each unnamed metadata node, of which there are others besides ones used in the array, and not every array element refers to another unique metadata node either.

bob80905: Given that the old test, "D:\DXC\tools\clang\test\HLSLFileCheck\hlsl\workgraph\test_EmitDxilFunctionProps_MDVals.hlsl", already serves as a regression test for the change of the insertion method, I've removed the redundant test. There is no way to trigger the out-of-bounds behavior without the definition of a node shader, and so an explanation has been added to the old test on how it serves its extra purpose.

PR #5336, opened 2023-06-27
WaveMatrix test data refactor: remove ShaderOpArithTable.xml dependency

tex3d:
This test used the ShaderOpArithTable.xml in a weird way that breaks the way we map to the HLK test. Use of ShaderOpArithTable.xml in this test was unnecessary, since it wasn't using more than one row with different parameter sets to define data-driven test cases.

This change simplifies things and gets rid of this dependency. The impact is that the shaders are now defined in ShaderOpArith.xml in place of the dummy shaders that used to be there, instead of ShaderOpArithTable.xml. The shader text and target are used as defined from ShaderOpArith.xml, instead of overriding those values with ones from ShaderOpArithTable.xml.

The only changes to the shader content during the move is in whitespace: indentation changed to be consistent with target file, trailing whitespace removed.


python3kgae (approved):

hekota (approved):
LGTM! Thank for moving this out of ShaderOpArithTable.xml.

PR #5339, opened 2023-06-28
Propagate literal types through loads.

### LINKED ISSUE #5319: fatal error: generated SPIR-V is invalid: Invalid word count: OpConstant starting at word 108 says it has 4 words, but found 5 words instead.

s-perron:
We currently do not do anything to determine the correct type for
variables with a literal type. This leads to type mismatches. In this
commit, I propagate the type the result of the load to the variable that
is loaded. This makes sure that the load is correct.

Fixes #5319

### DIFF
--- a/tools/clang/lib/SPIRV/LiteralTypeVisitor.cpp
+++ b/tools/clang/lib/SPIRV/LiteralTypeVisitor.cpp
@@ -298,6 +298,27 @@ bool LiteralTypeVisitor::visit(SpirvNonUniformBinaryOp *inst) {
   return true;
 }
 
+bool LiteralTypeVisitor::visit(SpirvLoad *inst) {
+  auto *pointer = inst->getPointer();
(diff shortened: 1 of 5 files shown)


cassiebeckley (approved):
LGTM

Keenuts (approved):

Issue #5310, opened 2023-06-21
Prevent two blobs compiled with different compile versions from linking


bob80905:
The compiler should prevent two blobs compiled with different compiler versions from being able to link.

A diagnostic should be emitted when two blobs that should be linked are detected to have been compiled with different compiler versions. The diagnostic should include information on what compiler version was used to compile each blob, to see how the versions differ.

bob80905:
Completed with https://github.com/microsoft/DirectXShaderCompiler/pull/5361

PR #5361, opened 2023-06-28
Error out if two blobs compiled with different compiler versions are linked.

bob80905:
This PR aims to accomplish the final step of the task to prevent blobs compiled with different compiler versions from being linked together.
The same compiler is used to compile both blobs in the test, but the compilerversion part in the first blob is forcefully changed to conflict with the compilerversion part within the second blob.
The test detects the correct error message, which mentions the names of the two libraries that differ, and what their differing versions are exactly.
The PR aims to resolve this issue:
https://github.com/microsoft/DirectXShaderCompiler/issues/5310

### DIFF
--- a/tools/clang/tools/dxcompiler/dxclinker.cpp
+++ b/tools/clang/tools/dxcompiler/dxclinker.cpp
@@ -219,7 +219,7 @@ DxcLinker::RegisterLibrary(_In_opt_ LPCWSTR pLibName, // Name of the library.
         pHeader, hlsl::DxilFourCC::DFCC_CompilerVersion);
     if (pDPH) {
         const hlsl::DxilCompilerVersion *pDCV =
-            (hlsl::DxilCompilerVersion *)(pDPH + 1);
+            (const hlsl::DxilCompilerVersion *)(pDPH + 1);
         // If the compiler version string is non-empty, add the struct to the
         // map
         bool success =
@@ -336,7 +336,26 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
       if (result != m_libNameToCompilerVersionPart.end()) {
         cur_version = ((*result).second);
       } else {
-        cur_version = nullptr;
+        UINT32 valMajor, valMinor;
+        dxcutil::GetValidatorVersion(&valMajor, &valMinor);
+        bool bValidatorAtLeast_1_8 =
+            DXIL::CompareVersions(valMajor, valMinor, 1, 8) >= 0;
+        if (bValidatorAtLeast_1_8) {
+          std::string errorMsg =
+              "error: Cannot link library with no compiler version part.\n";
+
+          std::string noteStr =
+              "note: library name is \"" + first_lib_name + "\"";
+
+          errorMsg += noteStr;
+          DiagStream << errorMsg;
+          bSuccess = false;
+
+          if (i != 0) {
+            // only one error per library
+            continue;
+          }
+        }
       }
 
       if (i == 0) {
--- a/tools/clang/tools/dxcompiler/dxclinker.cpp
+++ b/tools/clang/tools/dxcompiler/dxclinker.cpp
@@ -158,11 +158,8 @@ public:
 
     CW2A pUtf8LibName(libName, CP_UTF8);
     std::string libNameStr = std::string(pUtf8LibName);
-    m_uniqueCompilerVersions.insert(DeserializedDxilCompilerVersion(pDCV));
-    std::set<DeserializedDxilCompilerVersion>::iterator it =
-        m_uniqueCompilerVersions.find(DeserializedDxilCompilerVersion(pDCV));
-
-    m_libNameToCompilerVersionPart[libNameStr] = &(*it);
+    auto result = m_uniqueCompilerVersions.insert(DeserializedDxilCompilerVersion(pDCV));    
+    m_libNameToCompilerVersionPart[libNameStr] = &(*result.first);
 
     return true;
   }
@@ -222,10 +219,8 @@ DxcLinker::RegisterLibrary(_In_opt_ LPCWSTR pLibName, // Name of the library.
             (const hlsl::DxilCompilerVersion *)(pDPH + 1);
         // If the compiler version string is non-empty, add the struct to the
         // map
-        bool success =
-            AddCompilerVersionMapEntry(pLibName, pDCV, pDPH->PartSize);
-        if (!success) {
-        return E_INVALIDARG;
+        if (!AddCompilerVersionMapEntry(pLibName, pDCV, pDPH->PartSize)) {
+          return E_INVALIDARG;
         }
     }
     
@@ -334,7 +329,7 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
       auto result = m_libNameToCompilerVersionPart.find(cur_lib_name);
 
       if (result != m_libNameToCompilerVersionPart.end()) {
-        cur_version = ((*result).second);
+        cur_version = result->second;
       } else {
         UINT32 valMajor, valMinor;
         dxcutil::GetValidatorVersion(&valMajor, &valMinor);
--- a/tools/clang/tools/dxcompiler/dxclinker.cpp
+++ b/tools/clang/tools/dxcompiler/dxclinker.cpp
@@ -323,18 +323,19 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
       CW2A pUtf8LibName(pLibNames[i], CP_UTF8);
       bSuccess &= m_pLinker->AttachLib(pUtf8LibName.m_psz);
 
-      cur_lib_name = std::string(pUtf8LibName);
-      if (i == 0) {
-        first_lib_name = cur_lib_name;
-        first_version = cur_version;
-      }
+      cur_lib_name = std::string(pUtf8LibName);      
 
       // only libraries with compiler version parts are in the map
       auto result = m_libNameToCompilerVersionPart.find(cur_lib_name);
 
       if (result != m_libNameToCompilerVersionPart.end()) {
         cur_version = result->second;
-      } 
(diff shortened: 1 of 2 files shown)


tools/clang/tools/dxcompiler/dxclinker.cpp
```diff
@@ -109,6 +178,8 @@ class DxcLinker : public IDxcLinker, public IDxcContainerEvent {
   std::unique_ptr<DxilLinker> m_pLinker;
   CComPtr<IDxcContainerEventsHandler> m_pDxcContainerEventsHandler;
   std::vector<CComPtr<IDxcBlob>> m_blobs; // Keep blobs live for lazy load.
+  std::map<std::string, const DeserializedDxilCompilerVersion*> m_libNameToCompilerVersionPart;
```
python3kgae, line 181: Why save ptr here?
If std::set hit resize, the ptr will be invalid.

tex3d: IIRC, it is safe to take pointers to std::set keys, as long as they aren't removed from the set - they will never be reallocated upon resize.  I can't seem to find a reference that explicitly states this right now, however.

python3kgae (approved):

tools/clang/tools/dxcompiler/dxclinker.cpp
```diff
@@ -98,6 +140,33 @@ class DxcLinker : public IDxcLinker, public IDxcContainerEvent {
...
+        valid = vStr[pDCV->VersionStringListSizeInBytes - 1] == 0;
+    }
+
+    DXASSERT(valid, "DxilCompilerVersion part malformed");
+    if (!valid)
+        return false;
+
+    CW2A pUtf8LibName(libName, CP_UTF8);
+    std::string libNameStr = std::string(pUtf8LibName);
+    m_uniqueCompilerVersions.insert(DeserializedDxilCompilerVersion(pDCV));
```
pow2clk, line 161: This returns a pair containing the iterator and a bool indicating if it was inserted. That would eliminate the need to find it in the next line.

tools/clang/tools/dxcompiler/dxclinker.cpp
```diff
@@ -140,6 +211,24 @@ DxcLinker::RegisterLibrary(_In_opt_ LPCWSTR pLibName, // Name of the library.
...
+        pHeader, hlsl::DxilFourCC::DFCC_CompilerVersion);
+    if (pDPH) {
+        const hlsl::DxilCompilerVersion *pDCV =
+            (const hlsl::DxilCompilerVersion *)(pDPH + 1);
+        // If the compiler version string is non-empty, add the struct to the
+        // map
+        bool success =
+            AddCompilerVersionMapEntry(pLibName, pDCV, pDPH->PartSize);
+        if (!success) {
+        return E_INVALIDARG;
```
pow2clk, line 228: minor indent issue. 

Additionally, the `success` variable is only used for this check and might as well be `if(!AddCompilerVersionMapEntry...`

tools/clang/tools/dxcompiler/dxclinker.cpp
```diff
@@ -229,13 +318,74 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
...
       CW2A pUtf8LibName(pLibNames[i], CP_UTF8);
       bSuccess &= m_pLinker->AttachLib(pUtf8LibName.m_psz);
+
+      cur_lib_name = std::string(pUtf8LibName);
+
+      // only libraries with compiler version parts are in the map
+      auto result = m_libNameToCompilerVersionPart.find(cur_lib_name);
+
+      if (result != m_libNameToCompilerVersionPart.end()) {
+        cur_version = ((*result).second);
```
pow2clk, line 337: `result->second` seems more natural to me. Does it not work?

tools/clang/tools/dxcompiler/dxclinker.cpp
```diff
@@ -229,13 +318,74 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
...
+      } else {
+        UINT32 valMajor, valMinor;
+        dxcutil::GetValidatorVersion(&valMajor, &valMinor);
+        bool bValidatorAtLeast_1_8 =
+            DXIL::CompareVersions(valMajor, valMinor, 1, 8) >= 0;
+        if (bValidatorAtLeast_1_8) {
+          std::string errorMsg =
+              "error: Cannot link library with no compiler version part.\n";
```
pow2clk, line 345: From our discussions, it sounded like perhaps the version was coming from the compiler not the validator. If that's the case, checking the validator version doesn't seem right. Is it?

Regardless, this is going to fall through for the first iteration anyway and produce additional errors anyway. Why is this special error here?

pow2clk (commented):
Just questions really

tools/clang/unittests/HLSL/LinkerTest.cpp
```diff
@@ -158,14 +160,14 @@ class LinkerTest
...
-            llvm::ArrayRef<LPCWSTR> pArguments = {}) {
+            llvm::ArrayRef<LPCWSTR> pArguments = {}, bool bRegex=false) {
     CComPtr<IDxcOperationResult> pResult;
     VERIFY_SUCCEEDED(pLinker->Link(pEntryName, pShaderModel,
                                    libNames.data(), libNames.size(),
                                    pArguments.data(), pArguments.size(),
                                    &pResult));
     CheckOperationResultMsgs(pResult, pErrorMsgs.data(), pErrorMsgs.size(),
-                             false, false);
+                             false, bRegex);
```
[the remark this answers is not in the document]
bob80905: Correct. I think this is an irrelevant, small bug that I thought to fix while I was here.

tools/clang/tools/dxcompiler/dxclinker.cpp
bob80905: There are 3 cases that can happen:

1. The compiler must link a set of libraries that each have dxil containers that contain compiler version parts.
2. The compiler must link a set of libraries, where not every library has a compiler version part in the dxil container.
3. The compiler must link a set of libraries, where no dxil container has a compiler version part.

There should be no error for case 1, assuming all the parts are the same. There should be an error for case 2, because a compiler version part existing for one library implies the library with the non-existent part is different. However, it isn't obvious how to move forward in case 3, because we may want to treat a set of libraries with no compiler version parts as equal, and we may later want to treat them as unequal. I introduced the validator check because I thought it was logical that linking a set of libraries without compiler version parts depends on the version of the validator, and the check may set a framework for adjusting the version later. (Adjusting the version threshold would prevent users with older validator versions from seeing errors). But, we may also consider disallowing case 3 and emitting an error unconditionally.

tex3d: I think basing the expectation of a version part on the validator version of each library in `RegisterLibrary()` could make sense, since the version part is not legal in current released validators.  Let's call this the "version part expected depends on lib validator version" approach

Given that, we could say that adding libraries without an *expected* version part is an error.  Then we can go ahead and link libraries where none of them have a version part, because that's what's required for the linker to keep working for existing users and an updated compiler for retail scenarios.

The downside to failing in `RegisterLibrary()` on a missing *expected* version part is that it has no error message output, only an HRESULT.  An unexpected new `E_INVALIDARG` is not likely to go over well, especially without any clear indication as to what went wrong.  This means the checking and error reporting would likely need to remain in `Link()`.

Another downside to this ("version part expected depends on lib validator version") approach would be blocking attempted use of older compiler versions to compile a library, when working around a bug for instance.  If that were to occur after a validator that supports the version part, you would be unable to link that library.  

The current approach in this PR of using the current validator version to determine whether it's legal to link libraries without version parts (let's call it "version parts expected depends on linker validator version") will prevent linking of libraries compiled (or validated) with a compiler (or validator) version that did not support the version part, if a validator version that supports the version part is available at link time.  I'm not sure that's the right approach either, but there is a setting for validator version, and if we respect that, we should be able to offer a way for scenarios to continue to work in this intermediate time frame.

I'm also concerned about the scenario where pre-built libraries are shipped to link with user code, which could be the case for DXR scenarios.

Ultimately, this compiler version usage is the lazy, brittle way to avoid having to carefully and explicitly version the lib ABI/conventions.  I still fear turning it on without an opt-out, breaking users, and expecting to get that feedback and respond by adding an opt-out mechanism if necessary.  It's likely users will find and stick with a workaround of their own if they can find one, rather than give us the feedback and wait for us to do something to unblock them.  We don't want them doing that, because it may block users from adopting newer compilers, features, and validators, or at least add friction here.

tex3d: I think the best approach may be to simply allow the case where no libraries being linked have a compiler version part.

That will probably be the least problematic and blocking.  You would still get errors when compiling new libraries with the part and trying to link them with other libraries without the part, or with a different version.  That would be the new default behavior for compilation going forward, so it's only in legacy/corner cases that you would pass through this no version part path, and I think it's acceptable to allow it.

tools/clang/tools/dxcompiler/dxclinker.cpp
```diff
@@ -324,39 +324,17 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
       bSuccess &= m_pLinker->AttachLib(pUtf8LibName.m_psz);
 
       cur_lib_name = std::string(pUtf8LibName);
+      if (i == 0) {
+        first_lib_name = cur_lib_name;
+        first_version = cur_version;
```
tex3d, line 329: `cur_version` is not initialized yet, so won't it always be `nullptr`?

tex3d: I think this is likely the reason for all the test failures, BTW.

tex3d (approved):
Looks good!

PR #5378, opened 2023-06-29
Error on diff blob versions part 1

bob80905:
This PR is the first of 2 PRs to prevent libraries of different compiler versions from linking together. This first PR implements adding the compiler version information into the Dxil Container itself, to allow for comparison later. It also adds some basic validation to the new part in the Dxil Container.

### DIFF
--- a/lib/DxilContainer/DxilContainerAssembler.cpp
+++ b/lib/DxilContainer/DxilContainerAssembler.cpp
@@ -1682,6 +1682,7 @@ void hlsl::SerializeDxilContainerForModule(
 
   unsigned ValMajor, ValMinor;
   pModule->GetValidatorVersion(ValMajor, ValMinor);
+  bool bValidatorAtLeast_1_8 = DXIL::CompareVersions(ValMajor, ValMinor, 1, 8) >= 0;
   if (DXIL::CompareVersions(ValMajor, ValMinor, 1, 1) < 0)
     Flags &= ~SerializeDxilFlags::IncludeDebugNamePart;
   bool bSupportsShaderHash = DXIL::CompareVersions(ValMajor, ValMinor, 1, 5) >= 0;
@@ -1753,9 +1754,7 @@ void hlsl::SerializeDxilContainerForModule(
     DXASSERT(pModule->GetSerializedRootSignature().empty(),
              "otherwise, library has root signature outside subobject definitions");
     // Write the DxilCompilerVersion (VERS) part.        
-    unsigned ValMajor, ValMinor;
-    pModule->GetValidatorVersion(ValMajor, ValMinor);
-    bool bValidatorAtLeast_1_8 = DXIL::CompareVersions(ValMajor, ValMinor, 1, 8) >= 0;
+
     if (DXCVersionInfo && bValidatorAtLeast_1_8) {
 
       pVERSWriter = llvm::make_unique<DxilVersionWriter>(DXCVersionInfo);
--- a/lib/DxilContainer/DxilContainerAssembler.cpp
+++ b/lib/DxilContainer/DxilContainerAssembler.cpp
@@ -1749,27 +1749,25 @@ void hlsl::SerializeDxilContainerForModule(
   DXASSERT_NOMSG(pModule->GetSerializedRootSignature().empty());
 
   bool bMetadataStripped = false;
-  const hlsl::ShaderModel *pSM = pModule->GetShaderModel();
   if (pModule->GetShaderModel()->IsLib()) {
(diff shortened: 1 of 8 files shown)


tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -2084,6 +2085,115 @@ TEST_F(DxilContainerTest, ValidateFromLL_Abs2) {
   CodeGenTestCheck(L"..\\CodeGenHLSL\\container\\abs2_m.ll");
 }
 
+// Test to see if the Compiler Version (VERS) part gets added to library shaders
+// with SM >= 6.8
```
tex3d, line 2089: Should change comment that says "`with SM >= 6.8`" to "`with validator version >= 1.8`"
The `m_ver.SkipDxilVersion(1, 8)` below is still correct to constrain for validator version.  However, the library target can be lower.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -2084,6 +2085,115 @@ TEST_F(DxilContainerTest, ValidateFromLL_Abs2) {
...
+  CComPtr<IDxcBlob> pProgram;
+  CComPtr<IDxcBlobEncoding> pDisassembly;
+  CComPtr<IDxcOperationResult> pResult;
+
+  VERIFY_SUCCEEDED(CreateCompiler(&pCompiler));
+  CreateBlobFromText("export float4 main() : SV_Target { return 0; }",
+                     &pSource);
+  // Test DxilContainer with ShaderDebugInfoDXIL
+  VERIFY_SUCCEEDED(pCompiler->Compile(pSource, L"hlsl.hlsl", L"main",
+                                      L"lib_6_7", nullptr, 0, nullptr, 0,
```
tex3d, line 2104: This `lib_6_7` could be a lower library target, such as `lib_6_3` to ensure that we add the part on all down-level library targets, as long as validator version is high enough.

tex3d (changes requested):

lib/HLSL/DxilValidation.cpp
```diff
@@ -5732,7 +5732,7 @@ HRESULT ValidateDxilContainerParts(llvm::Module *pModule,
       break;
     case DFCC_CompilerVersion:
       // Either this blob is a PDB, or it is a library with shader model at least 6.8
-      if (pDxilModule->GetShaderModel()->IsSM68Plus() && ValCtx.isLibProfile) {
+      if (ValCtx.isLibProfile) {
```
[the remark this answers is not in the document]
bob80905: I've now changed the behavior so that an error is emitted if the part is detected in a non library blob.

tex3d (approved):
LGTM!

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1657,6 +1752,25 @@ void hlsl::SerializeDxilContainerForModule(
   if (pModule->GetShaderModel()->IsLib()) {
     DXASSERT(pModule->GetSerializedRootSignature().empty(),
              "otherwise, library has root signature outside subobject definitions");
+    // Write the DxilCompilerVersion (VERS) part.        
+    unsigned ValMajor, ValMinor;
+    pModule->GetValidatorVersion(ValMajor, ValMinor);
+    bool bValidatorAtLeast_1_8 = DXIL::CompareVersions(ValMajor, ValMinor, 1, 8) >= 0;
```
tex3d, line 1758: We already have `ValMajor` and `ValMinor` retrieved earlier, and used to initialize some bool variables there.  Can we just move the `bValidatorAtLeast_1_8` up with the others and remove the extra validator version retrieval here?

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1753,9 +1754,7 @@ void hlsl::SerializeDxilContainerForModule(
     DXASSERT(pModule->GetSerializedRootSignature().empty(),
              "otherwise, library has root signature outside subobject definitions");
     // Write the DxilCompilerVersion (VERS) part.        
-    unsigned ValMajor, ValMinor;
-    pModule->GetValidatorVersion(ValMajor, ValMinor);
-    bool bValidatorAtLeast_1_8 = DXIL::CompareVersions(ValMajor, ValMinor, 1, 8) >= 0;
+
```
tex3d, line 1757: nit: there's some trailing whitespace on the prior line and an extra newline here.

tex3d (approved):
LGTM.

python3kgae (approved):

PR #5382, opened 2023-06-30
Make assert consistent with the code

### LINKED ISSUE #5273: Assert on `static const`  in `template` functions
Compiling some HLSL code with `static const` in templates seems to be fine in the release builds, but when asserts are turned on in a debug build the following appears:

```
DirectXShaderCompiler/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp:4104: clang::QualType clang::spirv::DeclResultIdMapper::getTypeAndCreateCounterForPotentialAliasVar(const clang::DeclaratorDecl *, bool *): Assertion `!varDecl->isExternallyVisible() || varDecl->isStaticDataMember()' failed.
```

Repro with:

```hlsl
template <typename R> R test(R x) {
    static const R v = 20;
    return v * x;
}

[numthreads(32, 32, 1)] void main(uint2 threadId: SV_DispatchThreadID) {
    float x = test(10);
}
```

Should this be turned into an `emitError()` to disallow the case globally, with a fitting error message? I haven't checked if the code generates and runs as expected, but it seems so.

s-perron:
There is an assert that checks if a variable will be placed in the
Function or Private storage class in the spir-v. However, the condition
checked in that assert is inconsistent with the part of the code that
determines which storage class a variable belongs to.

The fix is to make the function `isExternalVar` available outside of
SprivEmitter.cpp, and use that in the assert. Now the same check is used
everywhere.

Fixes #5273

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -4135,7 +4135,7 @@ QualType DeclResultIdMapper::getTypeAndCreateCounterForPotentialAliasVar(
   if (const auto *varDecl = dyn_cast<VarDecl>(decl)) {
     // This method is only intended to be used to create SPIR-V variables in the
     // Function or Private storage class.
-    assert(!varDecl->isExternallyVisible() || varDecl->isStaticDataMember());
+    assert(!SpirvEmitter::isExternalVar(varDecl));
   }
 
   const QualType type = getTypeOrFnRetType(decl);
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -119,18 +119,6 @@ const Expr *isStructuredBufferLoad(const Expr *expr, const Expr **index) {
   return nullptr;
 }
 
-/// Returns true if the given VarDecl will be translated into a SPIR-V variable
-/// not in the Private or Function storage class.
-inline bool isExternalVar(const VarDecl *var) {
-  // Class static variables should be put in the Private storage class.
-  // groupshared variables are allowed to be declared as "static". But we still
-  // need to put them in the Workgroup storage class. That is, when seeing
-  // "static groupshared", ignore "static".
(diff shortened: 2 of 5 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -4135,7 +4135,7 @@ QualType DeclResultIdMapper::getTypeAndCreateCounterForPotentialAliasVar(
   if (const auto *varDecl = dyn_cast<VarDecl>(decl)) {
     // This method is only intended to be used to create SPIR-V variables in the
     // Function or Private storage class.
-    assert(!varDecl->isExternallyVisible() || varDecl->isStaticDataMember());
+    assert(!SpirvEmitter::isExternalVar(varDecl));
```
Keenuts, line 4138: I'm not sure but it seems those asserts are not equivalent?

passes if varDecl->isStaticDataMember is true, same.
But the isExternallyVisible() is not replicated in the new code.

isExternallyVisible returns true if the linkage is >= VisibleNoLinkage while the current checks only allows ExternalLinkage no?
Is that desirable?

s-perron: The old and new assert are not suppose to be equivalent. The existing assert is triggering in cases when it should not be.

The new assert is using the condition in that DXC uses in other places to determine which storage class to place the variable in. This will make all places that check if a variable should be in a storage class other than function or private consistent. That is desirable.

For the other issue: is `isExternalVar` wrong? That function is used to determine if an external variable should be created, so it will never be incorrect for the assert. We also have no reason to believe that a variable is being created in the wrong storage class.

https://github.com/microsoft/DirectXShaderCompiler/blob/2096d1b8054d1de864924ef0c5009f533597b9d4/tools/clang/lib/SPIRV/SpirvEmitter.cpp#L1771-L1773

Keenuts: Ok, thanks for the details!

Keenuts (approved):

cassiebeckley (approved):

PR #5387, opened 2023-07-03
Update build artifact permissions

### LINKED ISSUE #5376: Code coverage testing is failing with 'deployment_perms_error'
Started happening with this commit: https://github.com/microsoft/DirectXShaderCompiler/commit/6a3e2485ded47dd6503da97d64e2a75fe0fa3b29

Created deployment for 6a3e2485ded47dd6503da97d64e2a75fe0fa3b29, ID: 6a3e2485ded47dd6503da97d64e2a75fe0fa3b29
Getting Pages deployment status...
Current status: deployment_in_progress
Getting Pages deployment status...
Current status: deployment_perms_error

llvm-beanz:
I found other users reporting a similar issue:

https://github.com/actions/deploy-pages/issues/188

The root cause seems to be that the deploy-pages action changed to stop correcting permissions problems in the upload since it was slowing down deployment. The recommended fix is to set the permissions before uploading the artifact.

Let's see if this works.

Fixes #5376

### DIFF
--- a/.github/workflows/coverage-gh-pages.yml
+++ b/.github/workflows/coverage-gh-pages.yml
@@ -41,7 +41,7 @@ jobs:
         run: ninja -C build check-all
       - name: Generate Report
         run: ninja -C build generate-coverage-report
-      - name: Force artificat permissions
+      - name: Force artifact permissions
         run: chmod -c -R +rX ${{github.workspace}}/build/report
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v1
--- a/.github/workflows/coverage-gh-pages.yml
+++ b/.github/workflows/coverage-gh-pages.yml
@@ -41,6 +41,8 @@ jobs:
         run: ninja -C build check-all
       - name: Generate Report
         run: ninja -C build generate-coverage-report
+      - name: Force artificat permissions
+        run: chmod -c -R +rX ${{github.workspace}}/build/report
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage-gh-pages.yml
```diff
@@ -41,6 +41,8 @@ jobs:
         run: ninja -C build check-all
       - name: Generate Report
         run: ninja -C build generate-coverage-report
+      - name: Force artificat permissions
```
bfavela, line 44: *Artifact

.github/workflows/coverage-gh-pages.yml
```diff
@@ -41,6 +41,8 @@ jobs:
...
+        run: chmod -c -R +rX ${{github.workspace}}/build/report
```
bfavela, line 45: Is the -c for diagnostics? I don't know how many files are typically in this directory for how chatty that might be.

bfavela (commented):
LGTM

bfavela (approved):

.github/workflows/coverage-gh-pages.yml
llvm-beanz: Fixed

.github/workflows/coverage-gh-pages.yml
llvm-beanz: It only reports when a change is made to the permissions. I'd like to keep that in to see how wide the impact is, but will silence in a follow-up if the log is too big.

PR #5388, opened 2023-07-04
Emit Offset decoration for PhysicalStorage structs

### LINKED ISSUE #5327: [SPIR-V] DXC generates invalid SPIR-V when using RawBufferLoad with a struct.
From the SPIR-V spec:

> [Composite](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#CompositeType) objects in the StorageBuffer, PhysicalStorageBuffer, Uniform, and PushConstant [Storage Classes](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Storage_Class) must be explicitly laid out. The following apply to all the aggregate and matrix types describing such an object, recursively through their nested types:
>
> - Each structure-type member must have an Offset [decoration](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Decoration).

SPIRV-Tools started to validate this rule, meaning DXC now fails hard.
Seems like we simply don't generate the `Offset` member decoration when the offset is not explicit.

Looking at the code, I see 2 places to handle that:
 - LowerTypeVisitor.cpp: only add the offset decoration when required.
 - EmitVisitor.cpp: always add the offset decoration.
 

Method 1:
To fix this in the LowerTypeVisitor.cpp, we should not only pass the layout, but also the storage class where the type is used (which here is what's changing the offsets/decorations).
Then, depending on the storage class, the offset will be set in the type or not, the the rest of the code should handle it.
This seems to be the clean method, but requires some larger changes to the lowerTypeVisitor.

Method 2:
The second method is a bit more basic: always add the decoration.
I don't think the spec is against always specifying the Offset member decoration.
This would allow us to have a simple fix.
 - Bitfields should be handled correctly since the fields are already squashed.
 - Specific offsets (layout rule by ex), the offset is pre-computed, so we don't have to compute it. Otherwise we have the full-type, so we should always have the correct size.
 
 Repro steps:
 
 
 ```hlsl
 struct S {
  uint f1;
};

uint64_t Address;

[numthreads(64, 1, 1)]
void main(uint3 id : SV_DispatchThreadID) {
  uint tmp = vk::RawBufferLoad<S>(Address).f1;
}
 ```
 
 ```bash
 dxc -spirv -T cs_6_0 -E main -HV 2021 repro.hlsl
 ```

Keenuts:
Depending on the struct layout, the `Offset` decoration is not always required.
When this storage class is setup, the decoration needs to be added.

The tricky part if the mix of AST types and SPIRV types with bitfields:
 - in the AST, bitfields are represented as their own fields.
 - in SPIRV, bitfields are meld into 1 field when possible.

This means we need to lower the type to SPIR-V to compute the correct offset.
This would be fine if we could always generate the Offset decoration, but some types like image samplers are structs, and computing an offset for them is not clean (maybe no legal?).
So we need to do it properly, and only generate offsets when the storage
type is physical.

Fixes #5327

### DIFF
--- a/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
+++ b/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
@@ -264,13 +264,8 @@ const SpirvType *LowerTypeVisitor::lowerType(const SpirvType *type,
                                              SourceLocation loc) {
   if (const auto *hybridPointer = dyn_cast<HybridPointerType>(type)) {
     const QualType pointeeType = hybridPointer->getPointeeType();
-    SpirvLayoutRule layoutRule =
-        hybridPointer->getStorageClass() ==
-                spv::StorageClass::PhysicalStorageBuffer
-            ? SpirvLayoutRule::Physical
-            : rule;
     const SpirvType *pointeeSpirvType =
-        lowerType(pointeeType, layoutRule, /*isRowMajor*/ llvm::None, loc);
+        lowerType(pointeeType, rule, /*isRowMajor*/ llvm::None, loc);
     return spvContext.getPointerType(pointeeSpirvType,
                                      hybridPointer->getStorageClass());
   } else if (const auto *hybridSampledImage =
--- a/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
+++ b/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
@@ -264,8 +264,13 @@ const SpirvType *LowerTypeVisitor::lowerType(const SpirvType *type,
                                              SourceLocation loc) {
   if (const auto *hybridPointer = dyn_cast<HybridPointerType>(type)) {
     const QualType pointeeType = hybridPointer->getPointeeType();
+    SpirvLayoutRule layoutRule =
+        hybridPointer->getStorageClass() ==
+                spv::StorageClass::PhysicalStorageBuffer
+            ? SpirvLayoutRule::Physical
+            : rule;
     const SpirvType *pointeeSpirvType =
-        lowerType(pointeeType, rule, /*isRowMajor*/ llvm::None, loc);
+        lowerType(pointeeType, layoutRule, /*isRowMajor*/ llvm::None, loc);
     return spvContext.getPointerType(pointeeSpirvType,
                                      hybridPointer->getStorageClass());
   } else if (const auto *hybridSampledImage =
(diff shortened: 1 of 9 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -13634,12 +13682,15 @@ SpirvEmitter::loadDataFromRawAddress(SpirvInstruction *addressInUInt64,
   assert(loadInst);
   loadInst->setAlignment(alignment);
   loadInst->setRValue();
+  loadInst->setLayoutRule(SpirvLayoutRule::Physical);
```
s-perron, line 13685: Should `address` get the layout rule set as well? I wonder if this will remove the need for the change in `lowerType`, because it will now get the correct layout rule.

s-perron (changes requested):
My main concern is that you are creating a new layout rule, and I do not understand why. Is there one that matches the existing layout?

Not for this PR, but we should decide how the layout rules should change with the options:   

```
  -fvk-use-dx-layout      Use DirectX memory layout for Vulkan resources
  -fvk-use-gl-layout      Use strict OpenGL std140/std430 memory layout for Vulkan resources
  -fvk-use-scalar-layout  Use scalar memory layout for Vulkan resources
  ```

Technically the physical storage buffers are not Vulkan resources (not set or binding), but it would make sense that they match. We should wait to see if anyone wants it. Until then, we can always use the same layout.

File: tools/clang/lib/SPIRV/SpirvEmitter.cpp (hunk 1 above)
Comment (Keenuts) [reply tools/clang/lib/SPIRV/SpirvEmitter.cpp:13685]: Yes, thought changing the incoming instruction was weird, but it is done for the storage class, so makes sense.
And yes, it allows me to remove the weird thing in lowerTypeVisitor.

File: tools/clang/lib/SPIRV/LowerTypeVisitor.cpp [hunk 2]
```diff
@@ -264,8 +264,13 @@ const SpirvType *LowerTypeVisitor::lowerType(const SpirvType *type,
                                              SourceLocation loc) {
   if (const auto *hybridPointer = dyn_cast<HybridPointerType>(type)) {
     const QualType pointeeType = hybridPointer->getPointeeType();
+    SpirvLayoutRule layoutRule =
+        hybridPointer->getStorageClass() ==
+                spv::StorageClass::PhysicalStorageBuffer
+            ? SpirvLayoutRule::Physical
+            : rule;
```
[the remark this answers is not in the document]
Keenuts: removed

tools/clang/lib/SPIRV/SpirvEmitter.cpp
s-perron: This does not change the storage class for the incoming instruction. It sets it for the BitCast that was just created to make sure you are casting to the correct type.

PR #5390, opened 2023-07-05
Fix SPIRV struct reconstruction with bitfields

Keenuts:
HLSL/SPIR-V structs have some layout differences due to bitfields being squashed.
The reconstruction logic was using the AST layout, and not the new SPIR-V layout, meaning we could generate invalid indices during extraction/construction.

This PR fixes a potential bug with bitfields, while making the code reusable.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -6470,7 +6470,7 @@ SpirvInstruction *SpirvEmitter::reconstructValue(SpirvInstruction *srcVal,
         [&](size_t spirvFieldIndex, const QualType &fieldType,
             const auto &field) {
           SpirvInstruction *subSrcVal = spvBuilder.createCompositeExtract(
-              fieldType, srcVal, {spirvFieldIndex}, loc, range);
+              fieldType, srcVal, {static_cast<uint32_t>(spirvFieldIndex)}, loc, range);
           subSrcVal->setLayoutRule(srcVal->getLayoutRule());
           elements.push_back(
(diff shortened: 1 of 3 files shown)


Keenuts:
This PR is a split from #5388.
This is a different bug I needed to fix before. This one is can be submitted first, to make sure we focus on physical storage issue on the other PR.

s-perron (approved):

s-perron (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -584,6 +584,54 @@ const StructType *lowerStructType(const SpirvCodeGenOptions &spirvOptions,
   return output;
 }
 
+// Calls `operation` on for each field in the base and derives class defined by
```
dnovillo, line 587: nit: s/on for/on/

dnovillo: nit: s/derives/derived/

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -584,6 +584,54 @@ const StructType *lowerStructType(const SpirvCodeGenOptions &spirvOptions,
...
+// `recordType`. The `operation` will receive the AST type linked to the field,
+// the SPIRV type linked to the field, and the index of the field in the final
+// SPIR-V representation. This index of the field can vary from the AST
```
dnovillo, line 590: The AST type is used to determine derived types exclusively, right?

dnovillo (approved):
LGTM.  Just a couple of nits and a question.

PR #5392, opened 2023-07-06
[SPIRV] Emit Offset decoration for PhysicalStorage structs

### LINKED ISSUE #5327: [SPIR-V] DXC generates invalid SPIR-V when using RawBufferLoad with a struct.

Keenuts:
When the SPV_EXT_physical_storage_buffer extensions is used, vulkan can load raw addresses. For structs loaded through this mecanism, the offsets must be explicit.

This commit fixes decoration emission by attaching the correct layout when a struct is loaded using this extensions.
(Specifying a layout different than void forces explicit offsets).

Fixes #5327

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvBuilder.cpp
+++ b/tools/clang/lib/SPIRV/SpirvBuilder.cpp
@@ -369,10 +369,13 @@ SpirvUnaryOp *SpirvBuilder::createUnaryOp(spv::Op op, QualType resultType,
                                           SpirvInstruction *operand,
                                           SourceLocation loc,
                                           SourceRange range) {
+  if (!operand)
+    return nullptr;
   assert(insertPoint && "null insert point");
   auto *instruction =
       new (context) SpirvUnaryOp(op, resultType, loc, operand, range);
   insertPoint->addInstruction(instruction);
+  instruction->setLayoutRule(operand->getLayoutRule());
   return instruction;
 }
 
@@ -380,8 +383,11 @@ SpirvUnaryOp *SpirvBuilder::createUnaryOp(spv::Op op,
                                           const SpirvType *resultType,
                                           SpirvInstruction *operand,
                                           SourceLocation loc) {
+  if (!operand)
+    return nullptr;
   assert(insertPoint && "null insert point");
   auto *instruction = new (context) SpirvUnaryOp(op, resultType, loc, operand);
+  instruction->setLayoutRule(operand->getLayoutRule());
   insertPoint->addInstruction(instruction);
   return instruction;
 }
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -3067,6 +3067,9 @@ SpirvInstruction *SpirvEmitter::doCastExpr(const CastExpr *expr,
 
     auto *value = castToInt(loadIfGLValue(subExpr), subExprType, toType,
                             subExpr->getLocStart(), range);
+    if (!value)
+      return nullptr;
+
     value->setRValue();
     return value;
   }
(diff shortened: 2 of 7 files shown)


Keenuts:
This PR fails tests right now (requires fixes in #5390)

tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -369,19 +369,25 @@ SpirvUnaryOp *SpirvBuilder::createUnaryOp(spv::Op op, QualType resultType,
                                           SpirvInstruction *operand,
                                           SourceLocation loc,
                                           SourceRange range) {
+  if (!operand)
+    return nullptr;
```
s-perron, line 373: How does this happen? This is a change in behaviour, so I want to understand why.

s-perron (changes requested):
I think this looks good. I just want to understand how we get a null operand. That seems like an error on the part of the caller.

tools/clang/lib/SPIRV/SpirvBuilder.cpp
Keenuts: So when we have an error, we call emittError and then return null.
Example: when handling an non-implemented intrinsic function, if the function is not implemented, we emit, and return null.

But the expression is still getting parsed. The code usually is fine with that, the null moves around, and in the end, we return nullptr and show the error.

But this means functions like that can get called with nullptr.
The exact chain for this bug is:
 - castToInt(loadIfGL(subExpr)).

The loadIfGL is done on a subExpr using a non-implemented intrinsic (dot4), and returns nullptr. castToInt calls doUnaryOp, because it's a cast, but with a null operand.

Initially, I thought I should fix this call-site, but the pattern is a bit everywhere.

What I think should be good is:
 - in the SpirvBuilder, assert for null parameters.
 - in the emitter, allow null params, but check for validity before doing ptr->toRValue() but also when calling spirvBuilder functions like this one.

s-perron: The ideal solution is to emit and error before we start translating the AST into spir-v, but making the code more robust is good too.

s-perron (approved):

cassiebeckley (approved):

PR #5393, opened 2023-07-06
[lit] allow hcttest file-check to run on lit.

python3kgae:
This is prepare for allow both taef FileCheck and lit shell FileCheck exist.
For tests under HLSLFileCheck directory, taef FileCheck will run. Else py llvm-lit.py will run.

### DIFF
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -265,7 +265,20 @@ if "%TEST_CLEAN%"=="1" (
 )
 
 if "%TEST_MANUAL_FILE_CHECK%"=="1" (
-  set TEST_USE_LIT=0
+  echo %MANUAL_FILE_CHECK_PATH%|find "\HLSLFileCheck\" >nul
+  if errorlevel 1 (
+        echo "run lit file-check"
+        set TEST_MANUAL_FILE_CHECK=0
+        set TEST_CLANG=0
+        set TEST_DXILCONV=0
+        set TEST_SPIRV=0
+        set TEST_EXEC=0
+        set TEST_CMD=0
+        py %BIN_DIR%\llvm-lit.py %MANUAL_FILE_CHECK_PATH% -v
+  ) else (
+        set TEST_USE_LIT=0
+        echo "run taef file-check"
+  )
 )
 
 if "%TEST_USE_LIT%"=="1" (


bob80905 (commented):
Shouldn't line 269 be:
if "%errorlevel%"=="1" 
?

utils/hct/hcttest.cmd
```diff
@@ -265,7 +265,20 @@ if "%TEST_CLEAN%"=="1" (
 )
 
 if "%TEST_MANUAL_FILE_CHECK%"=="1" (
-  set TEST_USE_LIT=0
+  echo %MANUAL_FILE_CHECK_PATH%|find "\HLSLFileCheck\" >nul
```
tex3d, line 268: This is case-sensitive, but the file system on Windows is not.

utils/hct/hcttest.cmd
```diff
@@ -265,7 +265,20 @@ if "%TEST_CLEAN%"=="1" (
...
+  if errorlevel 1 (
```
tex3d, line 269: What's the intention here?  If `"\HLSLFileCheck\"` is *not* in the path given, it will use lit.py, otherwise it runs the legacy manual file-check path.  I don't understand why this is the intended behavior.

python3kgae:
> Shouldn't line 269 be:
> if "%errorlevel%"=="1" 
> ?

Not work after the change

utils/hct/hcttest.cmd
python3kgae: It is to support hcttest file-check for both taef File check test and lit File check shell test.

Once we convert all taef File check tests to lit File check shell test, we can remove this just use llvm-lit directly.

tex3d: Ok, got it.  Maybe we need a comment explaining this.

Also, how are relative paths handled?  Currently, it would pass the relative path to TAEF, and the ManualFileCheck test would interpret a relative path as relative to the HLSLFileCheck directory.  This would change that behavior to use lit on a relative path from the current directory instead.  That sounds ok, unless the relative path is under HLSLFileCheck.  We could check the absolute version of the input path for HLSLFileCheck instead, which could catch this case, and we would pass that absolute version into ManualFileCheck.  This would change the relative path behavior to always be relative to the current working directory (which I think is better - I never used the other relative path behavior, and the previous behavior couldn't be used for lit otherwise).

python3kgae: relative paths handled by use %~f2 instead of %~2.

utils/hct/hcttest.cmd
python3kgae: Fixed with -i.

tex3d (approved):
LGTM.

pow2clk (approved):
Looks good!

PR #5397, opened 2023-07-07
SPIRV: don't generate invalid debug instructions

Keenuts:
When generating debug instructions for a function, each one is linked to its scope.
In case of member functions, this scope is the class.

When declaring a class, all its member, including functions must be declared.

This cycle requires a forward reference, which is allowed by the debug instruction spec, but not by parents: NonSemantic & SPIR-V specs. This is a spec issue we have to fix. In the meantime, we decided to emit a warning, and generate slightly worse debug instructions.

Context: https://github.com/KhronosGroup/SPIRV-Registry/issues/203

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -782,7 +782,7 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
     return;
 
   if (spirvOptions.debugInfoRich) {
-    emitWarning("Debug information for methods won't be linked to their class. "
+    emitWarning("Member functions will not be linked to their class in the debug information. "
                 "See https://github.com/KhronosGroup/SPIRV-Registry/issues/203",
                 {});
   }
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -781,6 +781,12 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
   if (context.getDiagnostics().hasErrorOccurred())
     return;
 
+  if (spirvOptions.debugInfoRich) {
+    emitWarning("Debug information for methods won't be linked to their class. "
+                "See https://github.com/KhronosGroup/SPIRV-Registry/issues/203",
+                {});
+  }
+
   TranslationUnitDecl *tu = context.getTranslationUnitDecl();
   uint32_t numEntryPoints = 0;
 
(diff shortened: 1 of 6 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -745,6 +745,12 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
   if (context.getDiagnostics().hasErrorOccurred())
     return;
 
+  if (spirvOptions.debugInfoRich) {
+    emitWarning("Debug information for methods won't be linked to their class. "
+                "See https://github.com/KhronosGroup/SPIRV-Registry/issues/203",
+                {});
+  }
```
s-perron, line 752: This gets output even if there are no member functions. Is that what the SPIR WG agreed to? Should this be moved to DebugTypeVisitor.cpp, and only emit the warning if there was a function that should be skipped?  However, we do not want to issue the warning multiple times.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -745,6 +745,12 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
   if (context.getDiagnostics().hasErrorOccurred())
     return;
 
+  if (spirvOptions.debugInfoRich) {
+    emitWarning("Debug information for methods won't be linked to their class. "
```
s-perron, line 749:
```suggestion
    emitWarning("Member functions will not be linked to their class in the debug information. "
```

s-perron (changes requested):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts: Yeah, originally I added the warning there, but the visitor is called multiple times, even with 1 single method. (At least if I put the warning where I know we have the debug instruction "badly" generated)

Keenuts: What could be done is storing some bit in the visitor, like "have you encountered a member function", then read this bit once in the SpirvEmitter, but it looked a bit ugly

Keenuts: @s-perron anything blocking on your end? Would you be Ok with this warning being shown once, but even when no class is compiled or do you want me to add some state to minimize printing?

s-perron: Thanks for the reminder. I'm fine with the warning as is. We'll see if it bothers people.  

I just thought of something. Can dxc turn warnings into errors? If so, what happens when this warning is issued?

Keenuts: So there is `-WX`, and `-Werror` is accepted. Both looks to keep this warning as a warning. And there is also `-no-warnings` which disables this warning.

s-perron (approved):

cassiebeckley (approved):

Issue #5356, opened 2023-06-28
Work Graphs: Validate null or undef handle usage


llvm-beanz:
The verify_output_complete.hlsl has a TBD regarding preventing the null handle usage that this test case introduces.  This test can be evolved/adapted as pieces are implemented.

There are several things we can do in priority order:
1. Update validation to prevent null (or undef) handle usage in any intrinsic other than OutputComplete, where it's defined to be a no-op.
2. Fail earlier in the compilation pipeline when null (or undef) handle usage is detected.
3. If possible detect these early through AST analysis for the best diagnostic quality.

bob80905:
Fixed by #5399

PR #5399, opened 2023-07-07
[Validation] Prevent instructions that accept handle arguments from accepting malformed handle arguments

bob80905:
Except for Output Complete, instructions should not accept handle arguments that are undefs / zeroinitializers, and should emit a diagnostic when this happens. This PR adds tests to specifically exercise the cases where a call instruction :

1.  receives an undef argument
2.  receives a zeroinitializer argument
3.  receives a handle argument that is invalid

And combines these cases with all different types of handle arguments (resource handle, node handle, and node record handle)
This PR intends to at least implement the first work item in [#5356](https://github.com/microsoft/DirectXShaderCompiler/issues/5356)

### DIFF
--- a/docs/DXIL.rst
+++ b/docs/DXIL.rst
@@ -3089,6 +3089,7 @@ INSTR.RESOURCECLASSFORSAMPLERGATHER       sample, lod and gather should be on sr
 INSTR.RESOURCECLASSFORUAVSTORE            store should be on uav resource.
 INSTR.RESOURCECOORDINATEMISS              coord uninitialized.
 INSTR.RESOURCECOORDINATETOOMANY           out of bound coord must be undef.
+INSTR.RESOURCEINVALID                     Resource is invalid.
 INSTR.RESOURCEKINDFORBUFFERLOADSTORE      buffer load/store only works on Raw/Typed/StructuredBuffer.
 INSTR.RESOURCEKINDFORCALCLOD              lod requires resource declared as texture1D/2D/3D/Cube/CubeArray/1DArray/2DArray.
 INSTR.RESOURCEKINDFORGATHER               gather requires resource declared as texture/2D/Cube/2DArray/CubeArray.
--- a/docs/DXIL.rst
+++ b/docs/DXIL.rst
@@ -3089,7 +3089,6 @@ INSTR.RESOURCECLASSFORSAMPLERGATHER       sample, lod and gather should be on sr
 INSTR.RESOURCECLASSFORUAVSTORE            store should be on uav resource.
 INSTR.RESOURCECOORDINATEMISS              coord uninitialized.
 INSTR.RESOURCECOORDINATETOOMANY           out of bound coord must be undef.
-INSTR.RESOURCEINVALID                     Resource is invalid.
 INSTR.RESOURCEKINDFORBUFFERLOADSTORE      buffer load/store only works on Raw/Typed/StructuredBuffer.
 INSTR.RESOURCEKINDFORCALCLOD              lod requires resource declared as texture1D/2D/3D/Cube/CubeArray/1DArray/2DArray.
 INSTR.RESOURCEKINDFORGATHER               gather requires resource declared as texture/2D/Cube/2DArray/CubeArray.
--- a/lib/HLSL/DxilValidation.cpp
+++ b/lib/HLSL/DxilValidation.cpp
@@ -1152,7 +1152,7 @@ void ValidateHandleArgsForInstruction(CallInst *CI, DXIL::OpCode opcode,
       }
       else if (argTy == pHandleTy) {
         // GetResourceFromHandle will emit an error on an invalid handle
-        hlsl::DxilResourceProperties DRP = GetResourceFromHandle(op, ValCtx);    
+        GetResourceFromHandle(op, ValCtx);    
       }
     }
   }
--- a/lib/HLSL/DxilValidation.cpp
+++ b/lib/HLSL/DxilValidation.cpp
@@ -1131,33 +1131,34 @@ static int GetCBufSize(Value *cbHandle, ValidationContext &ValCtx) {
   return RP.CBufferSizeInBytes;
 }
 
-void ValidateHandleArgsForInstruction(CallInst* CI, DXIL::OpCode opcode, ValidationContext &ValCtx) {
-  // make sure none of the handle arguments are undef / zero-initializer,
-  // The only exception is that a zero-initializer is allowed
-  // as an argument to Output Complete (CAZ == zero-initializer)
-  // Also, do not accept any resource handles with invalid dxil resource properties
+// make sure none of the handle arguments are undef / zero-initializer,
+// The only exception is that a zero-initializer is allowed
+// as an argument to Output Complete (CAZ == zero-initializer)
+// Also, do not accept any resource handles with invalid dxil resource
+// properties
+void ValidateHandleArgsForInstruction(CallInst *CI, DXIL::OpCode opcode,
+                                      ValidationContext &ValCtx) {
+
   for (Value *op : CI->operands()) {
     const Type *pHandleTy = ValCtx.HandleTy; // This is a resource handle
     const Type *pNodeHandleTy = ValCtx.DxilMod.GetOP()->GetNodeHandleType();
-    const Type *pNodeRecordHandleTy = ValCtx.DxilMod.GetOP()->GetNodeRecordHandleType();
+    const Type *pNodeRecordHandleTy =
+        ValCtx.DxilMod.GetOP()->GetNodeRecordHandleType();
 
     const Type *argTy = op->getType();
-    if (argTy == pNodeHandleTy || argTy == pNodeRecordHandleTy || argTy == pHandleTy) {
-      bool bIsUninitialized = isa<UndefValue>(op) || isa<ConstantAggregateZero>(op);
+    if (argTy == pNodeHandleTy || argTy == pNodeRecordHandleTy ||
+        argTy == pHandleTy) {
+      bool bIsUninitialized =
+          isa<UndefValue>(op) || isa<ConstantAggregateZero>(op);
       if (argTy == pHandleTy) {
-        hlsl::DxilResourceProperties pDRP = ValCtx.GetResourceFromVal(op);
-        if (!pDRP.isValid()) {
-          ValCtx.EmitInstrError(
-            CI, ValidationRule::InstrResourceInvalid);
-        }       
+        hlsl::DxilResourceProperties DRP = GetResourceFromHandle(op, ValCtx);
+        if (!DRP.isValid()) {
+          ValCtx.EmitInstrError(CI, ValidationRule::InstrResourceInvalid);
+        }
       }
       if (bIsUninitialized) {
-        if (opcode == DXIL::OpCode::OutputComplete && isa<ConstantAggregateZero>(op)) {
-          continue;
-        }        
-        ValCtx.EmitInstrError(
-          CI, ValidationRule::InstrNoReadingUninitialized);        
-      }      
+        ValCtx.EmitInstrError(CI, ValidationRule::InstrNoReadingUninitialized);
+      }
     }
   }
 }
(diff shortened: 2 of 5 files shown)


lib/HLSL/DxilValidation.cpp
```diff
@@ -3100,6 +3100,15 @@ static void ValidateFunctionBody(Function *F, ValidationContext &ValCtx) {
             }
             dispatchMesh = CI;
           }
+
+          for (Value *op : CI->operands()) {
+            if (isa<UndefValue>(op)) {
+              if (dxilOpcode != DXIL::OpCode::OutputComplete) {
+                ValCtx.EmitInstrError(
```
python3kgae, line 3107: What about dxil op like sample which use undef for unused channel?

tex3d: Right, this is supposed to catch use of undef or zeroinitializer handle values only.  The only case that's allowed is zeroinitializer (not undef) for OutputComplete which is a no-op.  Also, we could eliminate the zeroinitializer OutputComplete when finalizing IR and still say that it is invalid during validation.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3100,6 +3100,15 @@ static void ValidateFunctionBody(Function *F, ValidationContext &ValCtx) {
             }
             dispatchMesh = CI;
           }
+
+          for (Value *op : CI->operands()) {
```
tex3d, line 3104: I don't think this is the right place for this code.

This should be implemented in `ValidateDxilOperationCallInProfile`, in the switch statement under the appropriate OpCodes.  See for instance how resource ops will use `ValidateResourceDxilOp(...)`.

lib/HLSL/DxilValidation.cpp
```diff
@@ -1131,6 +1131,111 @@ static int GetCBufSize(Value *cbHandle, ValidationContext &ValCtx) {
...
+  for (Value *op : CI->operands()) {
+    const Type *pHandleTy = ValCtx.HandleTy; // This is a resource handle
+    const Type *pNodeHandleTy = ValCtx.DxilMod.GetOP()->GetNodeHandleType();
+    const Type *pNodeRecordHandleTy = ValCtx.DxilMod.GetOP()->GetNodeRecordHandleType();
+
+    const Type *argTy = op->getType();
+    if (argTy == pNodeHandleTy || argTy == pNodeRecordHandleTy || argTy == pHandleTy) {
+      bool bIsUninitialized = isa<UndefValue>(op) || isa<ConstantAggregateZero>(op);
+      if (argTy == pHandleTy) {
+        hlsl::DxilResourceProperties pDRP = ValCtx.GetResourceFromVal(op);
```
tex3d, line 1148: - This should use `GetResourceFromHandle` instead.
- The name `pDRP` implies a pointer, but this isn't a pointer.
- There's no need to emit another error when `pDRP` is invalid, since `GetResourceFromHandle` will emit `InstrHandleNotFromCreateHandle`
- We should update the error emitted from `InstrHandleNotFromCreateHandle`, which is currently: `"Resource handle should returned by createHandle."`.  Grammar issue aside, the error is wrong on SM 6.6+ where the handle should come from annotateHandle instead of createHandle (and createHandleForLib for lib targets < 6.6).

tex3d (changes requested):

lib/HLSL/DxilValidation.cpp
bob80905: Thanks, good comments. I only have one problem:
There is a difference when switching from GetResourceFromHandle and GetResourceFromVal.
The ResourceClass is tested for validity in the GetResourceFromHandle case, while the ResourceKind is tested for validity in the GetResourceFromVal case. I've switched over to GetResourceFromHandle, but left the .isValid check in since it checks for a different validity than what's vaidated within the function itself.
Unless there's some assumption we can make that ResourceKind will be invalid if and only if ResourceClass is, I think the check should be left.

lib/HLSL/DxilValidation.cpp
```diff
@@ -1131,6 +1131,111 @@ static int GetCBufSize(Value *cbHandle, ValidationContext &ValCtx) {
...
+  }
+}
+
+void ValidateHandleArgs(CallInst* CI, DXIL::OpCode opcode, ValidationContext &ValCtx) {
+
+  switch (opcode) {
+  // Imm input value validation.
+  case DXIL::OpCode::Asin:
+  case DXIL::OpCode::Acos:
+  case DXIL::OpCode::Log:
```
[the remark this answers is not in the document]
bob80905: I've added a case for all dxil ops with a handle, nodehandle, or noderecordhandle arg, except for one which doesn't seem to have a dxil opcode, IndexNodeRecordHandle. I can address that op in a separate PR. 
Additionally, I think the extra custom validation for the annotation ops can be addressed in a separate PR.

tools/clang/test/HLSLFileCheck/validation/verify_output_complete.hlsl
```diff
@@ -1,11 +1,9 @@
-// RUN: %dxc -T lib_6_8 %s | FileCheck %s
-
+// RUN: %dxc -Vd -T lib_6_8 %s | FileCheck %s
+// validation disabled to prevent the validator failing when (%dx.types.NodeRecordHandle zeroinitializer) gets generated
```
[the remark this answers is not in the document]
bob80905: The problem was that the validator prevented final dxil from being piped into filecheck, and so the Check Line would fail. 
The original purpose of this end-to-end test is to validate that re-using an output-completed handle should fail validation. With the new validation addition in this PR, it reveals an extra bug in the compiler (where zeroinitializer should be prevented from ever being a handle argument in the first place). I think this test has no purpose until this deeper issue is resolved, and then I can add a check line that shows the validator prevents reusing output-completed handles.

docs/DXIL.rst
```diff
@@ -3089,6 +3089,7 @@ INSTR.RESOURCECLASSFORSAMPLERGATHER       sample, lod and gather should be on sr
 INSTR.RESOURCECLASSFORUAVSTORE            store should be on uav resource.
 INSTR.RESOURCECOORDINATEMISS              coord uninitialized.
 INSTR.RESOURCECOORDINATETOOMANY           out of bound coord must be undef.
+INSTR.RESOURCEINVALID                     Resource is invalid.
```
pow2clk, line 3092: I don't see a test that actually checks for this error?

lib/HLSL/DxilValidation.cpp
```diff
@@ -1131,6 +1131,112 @@ static int GetCBufSize(Value *cbHandle, ValidationContext &ValCtx) {
...
+  case DXIL::OpCode::IndexNodeHandle:
+  case DXIL::OpCode::NodeOutputIsValid:
+  // node record handle ops
+  case DXIL::OpCode::GetNodeRecordPtr:
+  case DXIL::OpCode::OutputComplete:
+  case DXIL::OpCode::GetInputRecordCount:
+  case DXIL::OpCode::FinishedCrossGroupSharing:
+  case DXIL::OpCode::BarrierByNodeRecordHandle:
+  case DXIL::OpCode::CreateNodeInputRecordHandle:
+  // TODO: add case DXIL::OpCode::IndexNodeRecordHandle:
```
pow2clk, line 1217: I see that this and other TODOs were to be addressed in a followup PR. Could you create an issue and reference it here to make sure that isn't lost and all the information is stored?

pow2clk (approved):
I'd like to see a followup issue filed. Otherwise fine!

lib/HLSL/DxilValidation.cpp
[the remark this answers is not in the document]
tex3d: My point was that we shouldn't be listing the handle ops we want to check in the switch statement.  We should only detect and skip the generic check on the exceptions.  The list of ops that have handles is an easy way to miss one or to forget to add one in the future.  All ops should run the generic handle validation, except several specific handle initialization operations.

lib/HLSL/DxilValidation.cpp
```diff
@@ -1131,6 +1131,112 @@ static int GetCBufSize(Value *cbHandle, ValidationContext &ValCtx) {
...
+      if (bIsUninitialized) {
+        ValCtx.EmitInstrError(CI, ValidationRule::InstrNoReadingUninitialized);
+      }
+    }
+  }
+}
+
+void ValidateHandleArgs(CallInst* CI, DXIL::OpCode opcode, ValidationContext &ValCtx) {
+
+  switch (opcode) {
```
tex3d, line 1166: My previous comment regarding the switch isn't visible in the file after changes.  So I'm putting this comment here.

My point was that we shouldn't be listing all the ops with handle args we want to check in the switch statement. We should only detect and skip the generic check on the specific known exceptions.  All other ops should just get the generic check. The list of ops that have handle args is an easy way to miss one or to forget to add one in the future. All ops should run the generic handle validation, except several specific handle initialization operations.

python3kgae (approved):

lib/HLSL/DxilValidation.cpp
```diff
@@ -1131,6 +1131,54 @@ static int GetCBufSize(Value *cbHandle, ValidationContext &ValCtx) {
...
+
+    const Type *argTy = op->getType();
+    if (argTy == pNodeHandleTy || argTy == pNodeRecordHandleTy ||
+        argTy == pHandleTy) {
+      bool bIsUninitialized =
+          isa<UndefValue>(op) || isa<ConstantAggregateZero>(op);
+      if (argTy == pHandleTy) {
+        hlsl::DxilResourceProperties DRP = GetResourceFromHandle(op, ValCtx);
+        if (!DRP.isValid()) {
+          ValCtx.EmitInstrError(CI, ValidationRule::InstrResourceInvalid);
```
tex3d, line 1154: My point from the previous comment here was that using `GetResourceFromHandle` will emit an error on an invalid handle, including the case you're checking here, without the need for an additional check.  You commented that `GetResourceFromHandle` checks ResourceClass, but the `getResourceClass` method it uses actually checks the `ResourceKind` and will return `ResourceClass::Invalid` when `ResourceKind::Invalid`, so there is no gap.  This means that the new error will always be emitted after another error has already been emitted from `GetResourceFromHandle`.  I don't think adding the new error is necessary.  A comment above the call to `GetResourceFromHandle` could indicate that the code expects that call to emit an error when the handle is not valid (like: "GetResourceFromHandle will emit an error on an invalid handle").

I also think that for the zeroinit/undef case, we could universally emit the error, before checking which handle type we have.  That way, this error will be consistent for all zeroinit or undef handle usage.  You can just change the initialization of `bIsUninitialized` to an if that emits this error, with an `else if` for the additional resource handle checks.

This may require test updates, if there are any tests that exercise the zeroinit or undef handle case for resources.

Another issue I just thought of:
For invalid resource handles, if/when there is validation of the operation itself, usually this will require retrieval of the handle, often using `GetResourceFromHandle` again.  This means any errors emitted there will now be emitted twice for these operations.

So, while this change covers potential gaps in invalid handle usage in DXIL ops, it also duplicates the coverage where this was already detected later.  The current way that some things flow through to `GetResourceFromHandle` and emit errors, while some others use `GetResourceFromVal` which emits no errors, is currently a mess.

To resolve this, I think the basic handle validation can be up-front, like this, to guarantee that it always emits a consistent error for invalid handles.  Then, when looking up handle properties in the individual operation validation cases, those should be using `GetResourceFromVal()` and `IsValid()` to skip the ResourceProperties-based checks on invalid handles, to avoid spurious additional errors when an error will have already been emitted for an invalid handle.

I also don't think the distinction between `GetResourceFromHandle` and `GetResourceFromVal` name-wise and functionality-wise makes sense.  The first should be changed into a validation-only check with a new name, such as `ValidateResourceHandle()`.  Then this would be called under this basic handle validation pass only.  Other uses of `GetResourceFromHandle` would be changed to use `GetResourceFromVal` instead.  Then all uses of `GetResourceFromVal` should be checked to make sure any uses of those resource properties for additional validation are placed under control flow guarded by `IsValid()` to avoid the additional spurious errors.

Note that `GetResourceFromHandle` is used in functions for getting specific properties (`GetSamplerKind`, `GetResourceKindAndCompTy`, `GetCBufSize`).  Those functions should probably DXASSERT if the handle is invalid, since callers should guard against that before using them.  Then the callers need new code to call `GetResourceFromVal` and `IsValid()` to guard those code paths.  Note that `ValidateSampleInst` and `ValidateGather` could share the code that calls `GetSamplerKind` (and checks for valid handle) with a tiny refactor.

The additional changes I suggest could be done as part of this change, or we could file an additional issue to track that work and follow up with a later change instead.

lib/HLSL/DxilValidation.cpp
tex3d: `GetResourceFromVal()` doesn't check for validity at all, which is why you needed the .IsValid() check.
`GetResourceFromHandle()` calls `DxilResourceProperties::getResourceClass()`, which checks the `ResourceKind`, and will return `ResourceClass::Invalid` when `ResourceKind` is `Invalid` (same condition as for `IsValid()`).  See my new comment which points this out and more.

Earlier, I missed following up on this because you marked the issue as resolved, and it disappeared from the code due to changes.  If there's a response like this and it may not have fully resolved all concerns, please don't resolve the comment, so we don't lose track of it.

tools/clang/test/HLSLFileCheck/validation/verify_output_complete.hlsl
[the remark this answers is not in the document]
tex3d: > where zeroinitializer should be prevented from ever being a handle argument in the first place

This could be generated/discovered late in compilation, so I don't think we can always detect it through AST analysis.  An error later in the pipeline could be introduced to cause a failure before we complete compilation and pass it to the validator, but that may have diminishing returns.  Detection of clear cases at the AST level should be performed for better diagnostics, which should easily be able to catch this particular case.

Since the IR test that's part of this PR already checks the validation for zeroinit/undef handle usage, we can probably remove this test from this PR, as long as we have an issue tracking the work item to detect the case in the AST.  Then we can attach this example to that work item, since this test could be converted to a verifier test at that point.

tex3d (approved):
LGTM.

PR #5402, opened 2023-07-10
[lit] Make lit the only test runner

python3kgae:
Remove disable-lit and enable-lit from hctbuild/hcttest.

This is to allow write lit FileCheck shell test.

Will need https://github.com/microsoft/DirectXShaderCompiler/pull/5393 to make file-check option work for both taef FileCheck and lit FileCheck.
Also need https://github.com/microsoft/DirectXShaderCompiler/pull/5404 to fallback to taef for taef only options.


bob80905 (approved):
LGTM

llvm-beanz (commented):
I think this change needs sign-off by @tex3d and @hekota.

tex3d (approved):

hekota (approved):
LGTM!

utils/hct/hctbuild.cmd
```diff
@@ -188,16 +188,6 @@ if "%1"=="-dxc-cmake-system-version" (
 if "%1"=="-show-cmake-log" (
   set SHOW_CMAKE_LOG=1
   shift /1 & goto :parse_args
-)  
-if "%1"=="-disable-lit" (
-  echo Disable LIT testing
-  set ENABLE_LIT=OFF
```
llvm-beanz, line 194: It seems like there should be more cleanup than this...

python3kgae: Updated.

cmake/caches/PredefinedParams.cmake
```diff
@@ -12,12 +12,8 @@ if (DXC_COVERAGE)
...
 endif()
 
-# TODO: remove this option after switch to support lit only.
-if (DXC_DISABLE_LIT)
-  set(LLVM_TEST_VALUE OFF)
-else()
-  set(LLVM_TEST_VALUE ON)
-endif()
+# Enable testing by default.
+set(LLVM_TEST_VALUE ON)
```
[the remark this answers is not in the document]
python3kgae: Updated.
Also added option for HLSL_INCLUDE_TESTS.

LLVM/CLANG_INCLUDE_TESTS are already default to ON on main CMake files.

cmake/caches/PredefinedParams.cmake
```diff
@@ -12,15 +12,9 @@ if (DXC_COVERAGE)
...
-  set(LLVM_TEST_VALUE OFF)
-else()
-  set(LLVM_TEST_VALUE ON)
-endif()
-
-set(LLVM_INCLUDE_TESTS ${LLVM_TEST_VALUE} CACHE BOOL "")
-set(CLANG_INCLUDE_TESTS ${LLVM_TEST_VALUE} CACHE BOOL "")
+# Enable testing by default.
+set(LLVM_INCLUDE_TESTS ON CACHE BOOL "")
+set(CLANG_INCLUDE_TESTS ON CACHE BOOL "")
```
llvm-beanz, line 17: If these default to `On`, you don't need to set them here.

utils/hct/hctbuild.cmd
```diff
@@ -319,7 +308,7 @@ set CMAKE_OPTS=%CMAKE_OPTS% -DENABLE_SPIRV_CODEGEN:BOOL=%SPIRV%
 set CMAKE_OPTS=%CMAKE_OPTS% -DSPIRV_BUILD_TESTS:BOOL=%SPV_TEST%
 set CMAKE_OPTS=%CMAKE_OPTS% -DCLANG_ENABLE_ARCMT:BOOL=OFF
 set CMAKE_OPTS=%CMAKE_OPTS% -DCLANG_ENABLE_STATIC_ANALYZER:BOOL=OFF
-set CMAKE_OPTS=%CMAKE_OPTS% -DCLANG_INCLUDE_TESTS:BOOL=%ENABLE_LIT% -DLLVM_INCLUDE_TESTS:BOOL=%ENABLE_LIT%
+set CMAKE_OPTS=%CMAKE_OPTS% -DCLANG_INCLUDE_TESTS:BOOL=ON -DLLVM_INCLUDE_TESTS:BOOL=ON
```
llvm-beanz, line 311: If these are the default values, they don't need to be set here.

cmake/caches/PredefinedParams.cmake
python3kgae: Removed.

utils/hct/hctbuild.cmd
python3kgae: Removed.

llvm-beanz (approved):

utils/hct/hcttest.cmd
llvm-beanz, line 1: Shouldn't there be more cleanup in this file too?

At a minimum I think hcttestcmds needs to get deleted and removed somewhere right?

python3kgae: Removed hcttestcmds.
Since we'll fallback to taef for filter so other parts are left to make things easier.

PR #5404, opened 2023-07-10
[lit] fall back to taef for taef only test options

python3kgae:
It is not easy to pass through the quoted /p options thru python. And these options are mainly for local debug filter, not for taef FileCheck tests. Just fall back to taef for these options and treat them like taef helper.

This is help to allow lit FileCheck shell test.

### DIFF
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -94,6 +94,7 @@ if "%1"=="-clean" (
   set TEST_ALL=0
   set TEST_CLANG=1
   set TEST_USE_LIT=0
+  echo Fallback to taef when use taef only options.
   set TEST_CLANG_FILTER=%2
   shift /1
 ) else if "%1"=="file-check" (
@@ -105,6 +106,7 @@ if "%1"=="-clean" (
   set TEST_ALL=0
   set TEST_CLANG=1
   set TEST_USE_LIT=0
+  echo Fallback to taef when use taef only options.
   set TEST_CLANG_FILTER=VerifierTest::*
 ) else if "%1"=="cmd" (
   set TEST_ALL=0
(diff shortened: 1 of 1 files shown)


tex3d:
I think this also needs to disable lit if there are `ADDITIONAL_OPTS` set:

https://github.com/microsoft/DirectXShaderCompiler/pull/5404/files#diff-3e95beec922c9f8ca195fa77527bc226b01b588cd853a65ca9fa88dcf00b79e4R193

https://github.com/microsoft/DirectXShaderCompiler/pull/5404/files#diff-3e95beec922c9f8ca195fa77527bc226b01b588cd853a65ca9fa88dcf00b79e4R213

Also, `-file-check-dump` isn't going to be supported on the lit runner, unless we do some special thing, right?  This is an option that dumps all the input to FileCheck for each test to a folder in a tree matching the test layout under HLSLFileCheck.  This allows for convenient comparisons for changes that you think might be disruptive to the IR shape, or metadata, or something like that.

pow2clk (approved):
I'm still a bit uncomfortable with this. I'd feel slightly better if there were a message that said the fallback was taking place, but I won't insist

tex3d (approved):
LGTM.

utils/hct/hcttest.cmd
```diff
@@ -93,6 +93,8 @@ if "%1"=="-clean" (
 ) else if "%1"=="clang-filter" (
   set TEST_ALL=0
   set TEST_CLANG=1
+  set TEST_USE_LIT=0
+  echo Fallback to taef when use taef only options.
```
tex3d, line 97: `echo` could happen in one place when you detect that TEST_USE_LIT==0 after option parsing.

tex3d (commented):
Just one optional suggestion regarding the fallback message.

PR #5405, opened 2023-07-11
Allow arrays of RW, consume, and append structured buffers.

s-perron:
This is the first commit in a series that will enable array of
RW, consume, and append Structured buffers. This commit removes the
error when one is used on its own.

There will be a failure if a counter variable is needed.

This is the first PR is a series that will slowly enable different features for arrays of RWStructuredBuffers. To see the general direction the changes will go see https://github.com/microsoft/DirectXShaderCompiler/pull/5407.


Keenuts (approved):
Seems like there are some conflicts, etc, but as is, I this PR LGTM!
Thanks for the extra work, I do prefer this vs adding a bug at HEAD until the feature is complete.

cassiebeckley (approved):

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -1116,6 +1119,8 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
       Args.hasFlag(OPT_Wno_vk_ignored_features, OPT_INVALID, false) ||
       Args.hasFlag(OPT_Wno_vk_emulated_features, OPT_INVALID, false) ||
       Args.hasFlag(OPT_fvk_auto_shift_bindings, OPT_INVALID, false) ||
+      Args.hasFlag(OPT_fvk_allow_rwstructuredbuffer_arrays, OPT_INVALID,
+                   true) ||
```
tex3d, line 1123: @s-perron - This `true` breaks testing on non-spirv builds.

tex3d: I put up #5422 for the fix.

PR #5406, opened 2023-07-11
Remove 'using namespace std' to fix __noop ambiguity with latest libc++

amaiorano:
A recent change to libc++ added a 'struct __noop' to the 'std' namespace. When building with clang-cl using libc++ on Windows, this now introduces an ambiguity between this std::__noop struct, and the MSVC __noop intrinsic, in translation units where 'using namespace std;' is used. This commit replaces these with using clauses for the exact symbols required in the non-test cpp files.

### DIFF
--- a/lib/HLSL/DxilValidation.cpp
+++ b/lib/HLSL/DxilValidation.cpp
@@ -60,7 +60,9 @@
 #include <deque>
 
 using namespace llvm;
-using namespace std;
+using std::unique_ptr;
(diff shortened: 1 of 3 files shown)


amaiorano:
Friendly ping for reviews. We need this fix to land for building DXC in Dawn/Chrome. Thanks!

llvm-beanz (approved):
LGTM

s-perron (approved):

PR #5391, opened 2023-07-05
[Matrix type] Add AST matrix type.

python3kgae:
An AST matrix type is added based on upstream matrix type. This is prepare for support matrix orientation in template.

matrix<float, 2, 3> will generate a AST matrix type instead of a TemplateSpecialization. ArraySubscriptExpr will be created for matrix subscript instead of a CallExpr. Rewriter will keep const and default parameter for matrix. Error message will be different for matrix is not a class in AST anymore.

The rest will keep the same as before.

Next PR will add orientation to the AST matrix type. Many matrix related code will be useless after this change. Will remove the useless code in a separate PR.

PR #5412, opened 2023-07-12
Add Orientation to AST MatrixType

python3kgae:
Based on https://github.com/microsoft/DirectXShaderCompiler/pull/5391

Only the last 2 commits are for orientation.

1 commit save the orientation information to MatrixType
The last commit apply explicit matrix orientation to MatrixType instead of Attributed Type.

Explicit orientation only apply on Desugared type.
unorm/snorm has to apply on the element type of matrix instead of typedef after desugtared.
This will affect rewriter and ast dump for things like unorm row_major float4x4.

float2x3 and column_major float2x3 is different type now because the isExplicit is saved into MatrixType.

No cast is generated for orientation mismatch at AST level.
This could be changed later.


damyanp:

damyanp:

PR #5415, opened 2023-07-13
Generate counter for arrays of RW, consume, and append structured buffers.

s-perron:
This commit adds code to generate counters for single dimensional arrays
of RW,consume, and append structured buffers.

After this commit, the counter can be used if the global resource is
used directly.

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1687,6 +1687,20 @@ void DeclResultIdMapper::createCounterVar(
   }
 
   const SpirvType *counterType = spvContext.getACSBufferCounterType();
+  QualType declType = decl->getType();
+  if (declType->isArrayType()) {
+    // TODO: Handle multi-dimentional arrays.
+    uint32_t arrayStride = 4;
+    if (const auto *constArrayType =
+            astContext.getAsConstantArrayType(declType)) {
+      counterType = spvContext.getArrayType(
+          counterType, constArrayType->getSize().getZExtValue(), arrayStride);
+    } else {
+      assert(declType->isIncompleteArrayType());
+      counterType = spvContext.getRuntimeArrayType(counterType, arrayStride);
+    }
+  }
+
   // {RW|Append|Consume}StructuredBuffer are all in Uniform storage class.
   // Alias counter variables should be created into the Private storage class.
   const spv::StorageClass sc =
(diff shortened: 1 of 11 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1687,6 +1687,20 @@ void DeclResultIdMapper::createCounterVar(
   }
 
   const SpirvType *counterType = spvContext.getACSBufferCounterType();
+  QualType declType = decl->getType();
+  if (declType->isArrayType()) {
+    // TODO: Handle multi-dimentional arrays.
```
Keenuts, line 1692: Link this TODO with an issue?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -4536,8 +4540,15 @@ SpirvEmitter::incDecRWACSBufferCounter(const CXXMemberCallExpr *expr,
     return nullptr;
   }
 
+  llvm::SmallVector<SpirvInstruction *, 2> indexes;
+  if(const auto *arraySubscriptExpr = dyn_cast<ArraySubscriptExpr>(object)) {
+    //TODO: Handle multi-dimensional arrays.
```
Keenuts, line 4545: Link this TODO with an issue?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -4536,8 +4540,15 @@ SpirvEmitter::incDecRWACSBufferCounter(const CXXMemberCallExpr *expr,
...
+    indexes.push_back(doExpr(arraySubscriptExpr->getIdx()));
+  }
+  indexes.push_back(zero);
```
Keenuts, line 4548: Can you add a comment to explain why you push zero? I'd assume it's because the counter is at index zero, but might be unclear.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts, line 4545: Is handling multi-dimensional arrays more complex than doing:

```cpp
auto head = object;
while (const auto *arraySubscriptExpr = dyn_cast<ArraySubscriptExpr>(head)) {
   indexes.push_back(doExpr(arraySubscriptExpr->getIdx());
   head = arraySubscriptExpr;
}
```

?

s-perron: I don't know how complex it is. I have not looked at an example yet. We may need to reverse the subscripts. I don't want to write the code until I have a test, and I want to keep the PRs focused.

Keenuts: Fine by me then. Would be nice to have an issue open referencing this issue in case someones hits this issue.

s-perron: Sorry, I had updated the comments, but the changes did not get pushed. Today I created an over arching issue, and I link to it now.

cassiebeckley (approved):

Issue #5410, opened 2023-07-12
WaveActiveAllEqual and vector parameters


s-perron:
**Description**

The behaviour of `WaveActiveAllEqual` does not seem to match the documentation when the parameter is a vector.

The [documentation](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/waveactiveallequal) says that `WaveActiveAllEqual` always returns a bool.  However, DXC will treats the function as returning a vector.

**Steps to Reproduce**

```
// RUN: dxc -T cs_6_0 -HV 2018 -E main -fspv-target-env=vulkan1.1 t.hlsl -ast-dump

struct S {
    float4 val1;
    uint val2;
    bool res;
};

RWStructuredBuffer<S> values;

[numthreads(32, 1, 1)]
void main(uint3 id: SV_DispatchThreadID) {
    values[id.x].res = WaveActiveAllEqual(values[id.x].val1);
}
```

**Actual Behavior**

When run the compiler emits the following warning:

```
t.hlsl:13:22: warning: implicit truncation of vector type [-Wconversion]
    values[id.x].res = WaveActiveAllEqual(values[id.x].val1);
```

The ast that is dumped shows:

```
  |       `-CallExpr 0x55e01375b760 <col:24, col:60> 'vector<bool, 4>':'vector<bool, 4>'
  |         |-ImplicitCastExpr 0x55e01375b748 <col:24> 'vector<bool, 4> (*)(vector<float, 4>)' <FunctionToPointerDecay>
  |         | `-DeclRefExpr 0x55e01375b6b0 <col:24> 'vector<bool, 4> (vector<float, 4>)' lvalue Function 0x55e01375b510 'WaveActiveAllEqual' 'vector<bool, 4> (vector<float, 4>)'
  |         `-ImplicitCastExpr 0x55e01375b7a8 <col:43, col:56> 'float4':'vector<float, 4>' <LValueToRValue>
```

**Environment**
- DXC version  Using Top of tree
- Host Operating System  Linux, but but should be reproducible everywhere.

python3kgae:
Hi Steven,
This might be a bug in the doc.
In github wiki WaveActiveAllEqual is like this
https://github.com/microsoft/DirectXShaderCompiler/wiki/Wave-Intrinsics#booln-waveactiveallequaltype-expr-

[...]

PR #5419, opened 2023-07-14
Update spirv-headers and spirv-tools

s-perron:
Updates the submodules, and disables validation on two tests. The SPIR-V
is invalid, and spir-val just started checking it. Since we have been
generating this code for a while, I do not see the need to modify the
generated code immediatly.


s-perron:
I have disabled validation on two tests where we generate invalid code. The code generation will be fixed onece we determine the correct way to fix it.  Once #5410 is resolved, we will know if we need to change the AST or SpirvEmitter.

dnovillo (approved):
Thanks.  Will this affect the upcoming SDK release?  (because of #5410 )

s-perron:
> Thanks. Will this affect the upcoming SDK release? (because of #5410 )

It might. We should bring it up with LunarG now.

PR #5421, opened 2023-07-14
[lit] add clang_cc1 for lit to allow write verify test for lit

python3kgae:
We have some matrix ast test mixed with verify test. But withtout clang_cc1, we cannot add verify test for lit.

This change will enable write verify test for lit.

Only clang_cc1 is enabled, clang and clang_cc1as are still disabled.


llvm-beanz:
See #5032

python3kgae:
Close for -verify is already exist.

Add another PR for the dependency issue exposed by this.
https://github.com/microsoft/DirectXShaderCompiler/pull/5432

PR #5422, opened 2023-07-14
Fix test break on non-spirv build

tex3d:
A previous PR added this spriv option.  The default should have been false, but in the block that checks for spirv options on a non-spirv build, the default was set to true.  This broke testing on non-spirv builds.


bob80905 (approved):

bfavela (approved):

s-perron (commented):
LGTM

Issue #4357, opened 2022-03-27
Missing preprocessor defines causes rewriter to incorrectly remove types


tomjohnstone:
In the following example, TestStruct is stripped and the variable declaration causes the shader to fail to compile.

```
// dxr -E CSMain -remove-unused-globals test.hlsl
struct TestStruct
{
	uint Value;
};

TestStruct DoSomething()
{
	TestStruct value = (TestStruct)0;
	return value;
}

[numthreads(1,1,1)]
void CSMain()
{
	TestStruct orNot;
#if ENABLED
	orNot = DoSomething();
#endif
}

llvm-beanz:
The rewriter can't see permutations of preprocess or defines. That is a limitation of the AST design that isn't solvable without a huge architectural revamp. We're not going to dedicate resources to that scale of project in the DXC codebase as we are working to move to Clang for HLSL.

PR #5424, opened 2023-07-16
[rewriter] send define to rewriter functions.

### LINKED ISSUE #5423: `dxr.exe` doesn't support macro definitions via its CLI
**Description**
It would be extremely useful if the Rewriter (`dxr.exe`) supported the `-D` command-line parameter similar to `dxc.exe`. After all, it's similar to the preprocessor.

In the absence of this functionality, I had to generate source files with `#ifdef`s in order to rewrite the same source with different macros.

What's worse, `dxr.exe` accepts `-D` arguments without errors or warnings and rewrites the source ignoring the macros, which may take some experimentation to realize.

**Steps to Reproduce**

- Compile this trivial shader with `-D float4=0`: https://godbolt.org/z/5q5WKvq6e The compilation fails, which proves that the macro was expanded.
- Now, rewrite the same shader with `dxr.exe -D float4=0 -E PSMain`. The file is rewritten verbatim, which proves that the macro wasn't expanded.

**Environment**
- DXC version: 1.7.2212
- Host Operating System: Win 10

python3kgae:
Send opts.Defines to rewriter functions so macro definitions can work.

Fixes #5423

### DIFF
--- a/tools/clang/test/HLSLFileCheck/rewriter/dxr_macro.hlsl
+++ b/tools/clang/test/HLSLFileCheck/rewriter/dxr_macro.hlsl
@@ -1,0 +1,23 @@
+// RUN: %dxr -Dfloat4=0  %s 2>&1 | FileCheck %s
+// RUN: %dxr -line-directive -Dfloat4=0  %s 2>&1 | FileCheck %s
+// RUN: %dxr -decl-global-cb -Dfloat4=0  %s 2>&1 | FileCheck %s
+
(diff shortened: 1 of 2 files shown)


llvm-beanz:
There's a big downside to supporting this. If users try to use this in patterns like the one described in #4357, this is an unsolvable problem in Clang.

Are we really sure we want to open this door?

damyanp:

damyanp:

PR #5432, opened 2023-07-17
[Build] cleanup build dependency

python3kgae:
Add LLVMHLSL dependency to LLVMIPO and LLVMScalarOpts. Remove passprinters from LLVMHLSL to avoid cycle.

This is to fix link error found when trying to enable clang_cc1.


pow2clk (approved):
For reference, this is the more correct fix for #4326

adam-yang (approved):

PR #5433, opened 2023-07-17
Fix include in DxilNodeProps.h

jenatali:
Globals.h defines things like `IFC` macros that leak into callers. All that was really needed here was a definition of `std::string`.


tex3d (approved):
Looks good, thanks!

llvm-beanz (approved):
LGTM

PR #5437, opened 2023-07-18
Fixed a crash when linker erases functions that still have users.

adam-yang:
When linking to library with a list of `-exports`, the linker erases functions that are not exported assuming they are no longer used. This assumption is not correct when the function is a constructor for a global variable. This change fixes the crash by first checking whether the function still has uses before erasing.

This change adds a regression test with a non-trivially removable global variable constructor that would crash without this change.


python3kgae (approved):
If allow noinline, there might be case where a function is used by another unused function.
But for now this is good enough.

damyanp (approved):

PR #5438, opened 2023-07-19
[lit] use lit.py directly to allow send parameter to lit from hcttest

python3kgae:
This is for support adapter option for exec test when using lit. no_priority is enabled for clang test too.

CMake behavior will be the same except not support adapter option.
Also hcttest cmd will only run cmd test for lit.

### DIFF
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -185,6 +185,7 @@ if "%1"=="-clean" (
 ) else if "%1"=="-adapter" (
   set TEST_USE_LIT=0
   set TEST_ADAPTER= /p:"Adapter=%~2"
+  set EXEC_ADAPTER=--param adapter=%~2
   shift /1
 ) else if "%1"=="-verbose" (
   set LOG_FILTER=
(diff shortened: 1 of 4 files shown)


utils/hct/hcttest.cmd
```diff
@@ -277,24 +278,25 @@ if "%TEST_USE_LIT%"=="1" (
     set TEST_CLANG=1
   )
   if "%TEST_ALL%"=="1" (
-    rem check all includes clang, dxilconv, and exec
-    cmake --build %HLSL_BLD_DIR% --config %BUILD_CONFIG% --target check-all
+    rem check all includes clang, dxilconv
+	py %HLSL_SRC_DIR%/utils/lit/lit.py -sv --no-progress-bar --param build_mode=%BUILD_CONFIG% --param dxilconv_site_config=%HLSL_BLD_DIR%/projects/dxilconv/test/taef/lit.site.cfg --param clang_site_config=%HLSL_BLD_DIR%/tools/clang/test/lit.site.cfg --param skip_taef_exec=True  --param no_priority=True --param llvm_site_config=%HLSL_BLD_DIR%/test/lit.site.cfg --param llvm_unit_site_config=G:/repos/DXC/hlsl.bin/test/Unit/lit.site.cfg %HLSL_SRC_DIR%/projects/dxilconv/test/taef %HLSL_BLD_DIR%/tools/clang/test %HLSL_BLD_DIR%/test
```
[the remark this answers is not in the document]
python3kgae: So only exec test use lit.py and set no_priority=True as default for other tests?

tex3d (approved):
I think it's ok by me, assuming you resolve @llvm-beanz issues.

utils/hct/hcttest.cmd
```diff
@@ -323,13 +320,20 @@ if "%TEST_USE_LIT%"=="1" (
       cmake --build %HLSL_BLD_DIR% --config %BUILD_CONFIG% --target check-dxilconv
       set RES_DXILCONV=!ERRORLEVEL!
     )
+    if "%TEST_CMD%"=="1" (
+      py %BIN_DIR%\llvm-lit.py %HLSL_SRC_DIR%/tools/clang/test/DXC -v
+      set RES_CMD=!ERRORLEVEL!
+    )
```
pow2clk, line 326: This is another change in behavior that isn't mentioned in the description. I think it's worth mentioning therre

pow2clk (approved):
Looks good! I'd just like the cmd change mentioned

utils/hct/hcttest.cmd
python3kgae: Updated.

PR #5442, opened 2023-07-20
[lit] Merge git usr bin changes to release branch

python3kgae:
This will avoid test fail when forget to add git usr bin to PATH.
If git usr bin already on PATH, it will just use it.

Port from
https://github.com/llvm/llvm-project/commit/0f1f13fcb17fbc8c93d505da989a04ab5cbd9ed3


hekota (approved):
LGTM, but it should go to the main branch,

tex3d (approved):
It would be useful to update the description to indicate that this is cherry-picked and from what commit.

PR #5443, opened 2023-07-20
Instantiate argument default parameters

### LINKED ISSUE #5145: llvm assert in CodeGen with templated resource argument defaulted to ResourceDescriptorHeap element
llvm assert in CodeGen with templated resource argument defaulted to ResourceDescriptorHeap element.

Requires that the resource type has not yet been instantiated.  For instance, if you use `Texture2D<float4>` before the instantiation, it succeeds.

Found when testing fix for issue #5139, but this is a separate issue.

```hlsl
// RUN: %dxc -T vs_6_6 -HV 2021
template<typename T>
void foo(Texture2D<vector<T, 4> > tex = ResourceDescriptorHeap[0]) {}
// Texture2D<float4> Tex; // If you uncomment this line, it will no longer assert.  But it must precede the instantiation in main()
void main() {
  foo<float>();
}
```

The assert:
```cpp
  assert(D && "Cannot get layout of forward declarations!");
```

When retrieving layout of this RecordDecl:
```cpp
ClassTemplateSpecializationDecl 0x2c4866bc3f8 <<invalid sloc>> <invalid sloc> class Texture2D
`-TemplateArgument type 'vector<float, 4>'
```

In `ASTContext::getASTRecordLayout`, `D = D->getDefinition();` results in nullptr because the type has not yet been defined.  Preceding this is:
```cpp
  if (D->hasExternalLexicalStorage() && !D->getDefinition())
    getExternalSource()->CompleteType(const_cast<RecordDecl*>(D));
```
But this type does not return true for `hasExternalLexicalStorage()`.  Should we be using this mechanism to instantiate built-in template specializations?

llvm-beanz:
This addresses a latent bug in Clang that was fixed _way_ after DXC was forked. We encounter it in DXC because of the way that we inject AST bits through the external sema source, but it likely wasn't exposable through C++.

This change is mostly a back porting of the following upstream LLVM commits:

llvm-project/73c6a2448f24
llvm-project/f721e0582b15
llvm-project/c601377b2376
llvm-project/4409a83c2935

The end result of this change is that we still lazily instantiate default arguments, but we do force them to be instantiated during sema when they are used.

Fixes #5145

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/template/InstantiateDefaultArg.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/template/InstantiateDefaultArg.hlsl
@@ -1,0 +1,74 @@
+// RUN: %dxc -T vs_6_6 -HV 2021 %s -ast-dump | FileCheck %s -check-prefix=AST
+// RUN: %dxc -T vs_6_6 -HV 2021 %s -fcgl | FileCheck %s
+
+template<typename T>
+T test(const T a, const T b = 0)
+{
+  return a + b;
+}
+
+// This test verifies a few different modes if instantiation for default
+// arguments.
+
+// The first chunk of AST being verified is the FunctionTemplateDecl. In the
+// template we have boh parameters defined and the second has a template pattern
+// for the `b` parameter.
+
+// AST: FunctionTemplateDecl {{.*}} test
+// AST-NEXT: TemplateTypeParmDecl {{.*}} referenced typename T
+// AST-NEXT: FunctionDecl {{.*}} test 'T (const T, const T)'
+// AST-NEXT: ParmVarDecl {{.*}} referenced a 'const T'
+// AST-NEXT: ParmVarDecl {{.*}} referenced b 'const T' cinit
+// AST-NEXT: IntegerLiteral {{.*}} 'literal int' 0
+// AST-NEXT: CompoundStmt
+// AST-NEXT: ReturnStmt
+// AST-NEXT: BinaryOperator {{.*}} 'const T' '+'
+// AST-NEXT: DeclRefExpr {{.*}} 'const T' lvalue ParmVar {{.*}} 'a' 'const T'
+// AST-NEXT: DeclRefExpr {{.*}} 'const T' lvalue ParmVar {{.*}} 'b' 'const T'
+
+// The second AST block is the test<float> instantiation. It has a full
+// instantiation including instantiation of the default template parameter. This
+// instantiation is called both with and without the default argument.
+
+// AST: FunctionDecl {{.*}} used test 'float (const float, const float)'
+// AST-NEXT: TemplateArgument type 'float'
+// AST-NEXT: ParmVarDecl {{.*}} used a 'const float':'const float'
+// AST-NEXT: ParmVarDecl {{.*}} used b 'const float':'const float' cinit
+// AST-NEXT: ImplicitCastExpr {{.*}} 'float':'float' <IntegralToFloating>
+// AST-NEXT: IntegerLiteral {{.*}} 'literal int' 0
+// AST-NEXT: CompoundStmt
+
+// The third AST block is the test<int> instantiation. This instantiation is
+// is also a full instantiation including the initializer for the `b` parameter.
+// This function is only called with the default argument used.
+
+// AST: FunctionDecl {{.*}} used test 'int (const int, const int)'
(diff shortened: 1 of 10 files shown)


llvm-beanz:
@cassiebeckley or @s-perron, can one of you take a look at the SPIR-V test I updated here. I think the update is valid. Since the AST instantiates the default parameter the codegen comes out a bit more clean now.

tools/clang/test/HLSLFileCheck/hlsl/template/InstantiateDefaultArg.hlsl
```diff
@@ -0,0 +1,74 @@
...
+// both arguments.
+
+// AST: FunctionDecl {{.*}} used test 'double (const double, const double)'
+// AST-NEXT: TemplateArgument type 'double'
+// AST-NEXT: ParmVarDecl {{.*}}  used a 'const double':'const double'
+// AST-NEXT: ParmVarDecl {{.*}} used b 'const double':'const double'
+// AST-NEXT: CompoundStmt
+
+
+float4 main(uint vertex_id : SV_VertexID) : SV_Position
```
tex3d, line 65: Just a note in case it helps:  For a minimal test shader, vertex shaders don't require any inputs, and they can have any arbitrary outputs, without requiring `SV_Position`, allowing greater flexibility without the `float4` type requirement for `SV_Position`.

tools/clang/test/HLSLFileCheck/hlsl/template/InstantiateDefaultArgMember.hlsl
```diff
@@ -0,0 +1,86 @@
...
+{
+  Pupper<float> PF;
+  Pupper<int> PI;
+  Pupper<double> PD;
+  return float4(PF.test(1, 2), PF.test(4), (float)PI.test(5),
+                (float)PD.test(6, 7));
+}
+
+// CHECK: call float @"\01?test@?$Pupper@{{.*}}"(%"struct.Pupper<float>"* {{.*}}, float 1.000000e+00, float 2.000000e+00)
+// CHECK: call float @"\01?test@?$Pupper@{{.*}}"(%"struct.Pupper<float>"* {{.*}}, float 4.000000e+00, float 0.000000e+00)
```
tex3d, line 84: Would this check have failed without the change?  Or would the conversion have been constant-folded during CodeGen, making the fix undetectable in the IR here?

tools/clang/test/HLSLFileCheck/hlsl/template/InstantiateDefaultArgResource.hlsl
```diff
@@ -0,0 +1,36 @@
...
+// AST-NEXT: IntegerLiteral {{.*}} 'literal int' 0
+
+template<typename T>
+void foo(Texture2D<vector<T, 4> > tex = ResourceDescriptorHeap[0]) {}
+
+void main() {
+  foo<float>();
+}
+
+// CHECK: call void @"\01??$foo{{.*}}"(%"class.Texture2D<vector<float, 4> >"* %2)
```
tex3d, line 36: Shouldn't `%2` be changed to avoid errors between release/debug builds, since that changes whether named values are used (including for `%entry` block label)?

tools/clang/test/HLSLFileCheck/hlsl/template/InstantiateDefaultArgMember.hlsl
llvm-beanz: In DXIL, the conversion was always constant folded so this should have succeeded before the change. Since we didn't have any tests that verified this behavior previously I felt it prudent to add one. The similar SPIR-V test case did change codegen.

The AST is different for this case though, so the AST test would have failed for this test case without the change.

tools/clang/test/HLSLFileCheck/hlsl/template/InstantiateDefaultArgResource.hlsl
llvm-beanz: Updated.

tools/clang/test/HLSLFileCheck/hlsl/template/InstantiateDefaultArgMember.hlsl
tex3d: Yup, thought so, sounds good.

tex3d (approved):
LGTM!

s-perron (approved):
The spir-v code change is still correct, and better.

PR #5445, opened 2023-07-21
Fixed QueryInterface for ID3D12ShaderReflection interface

### LINKED ISSUE #3887: Unintentional behavior of ID3D12ShaderReflection::QueryInterface
```
ComPtr<ID3D12ShaderReflection> p_shader_reflection;
ComPtr<ID3D12LibraryReflection> p_library_reflection;
p_shader_reflection.As(&p_library_reflection); // Success

D3D12_LIBRARY_DESC library_desc;
p_library_reflection->GetDesc(&library_desc);
// Cause Run-Time Check Failure #2 - Stack around the variable 'library_desc' was corrupted.
```

DxilShaderReflection doesn't inherit from ID3D12LibraryReflection, but calling DxilShaderReflection::QueryInterface with __uuidof(ID3D12LibraryReflection) wil be succeed.

tex3d:
QueryInterface would accept any interface IID, instead of only the ones it recognizes. Fixed by adding PublicAPI::Invalid, returned from IIDToAPI when not one of the recognized APIs, and only accepting IUnknown or the original interface IID provided at create time.

CreateDxil{Shader|Library}Reflection would also completely ignore IID. Fixed by only accepting supported IIDs.

Get*ParameterDesc for signature elements also incorrectly computed structure size (due to alignment) to memcpy for D3D11_43 version. Fixed by copying only up to the end of the last field in the old structure version.

CreateReflectionObjectsForSignature had a special case avoiding filling in the MinPrecision member for D3D11_43, though this is unnecessary, since for that API version, the struct will be truncated to exclude the MinPrecision member when copying out the desc. Removed unnecessary special case.

Added DxilContainerTest::CheckReflectionQueryInterface test for Create and QI scenarios for supported and unsupported IIDs.  This also tests that the Get*ParameterDesc methods don't write beyond the structure size for the interface version.

Fixes #3887


lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -180,16 +182,14 @@ class DxilShaderReflection : public DxilModuleReflection, public ID3D12ShaderRef
   DXC_MICROCOM_TM_ADDREF_RELEASE_IMPL()
   DXC_MICROCOM_TM_CTOR(DxilShaderReflection)
   HRESULT STDMETHODCALLTYPE QueryInterface(REFIID iid, void **ppvObject) override {
-    HRESULT hr = DoBasicQueryInterface<ID3D12ShaderReflection>(this, iid, ppvObject);
```
bob80905, line 183: Why is it ok to completely drop the DoBasicQueryInterface call here?

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -272,9 +272,11 @@ namespace hlsl {
 HRESULT CreateDxilShaderReflection(const DxilProgramHeader *pProgramHeader, const DxilPartHeader *pRDATPart, REFIID iid, void **ppvObject) {
   if (!ppvObject)
     return E_INVALIDARG;
+  PublicAPI api = DxilShaderReflection::IIDToAPI(iid);
```
bob80905, line 275: Nit, shouldn't this be moved down lower so that api is closer to where it's used, sort of like how the code was originally?

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -2043,9 +2047,7 @@ void DxilShaderReflection::CreateReflectionObjectsForSignature(
     D3D12_SIGNATURE_PARAMETER_DESC Desc;
     Desc.ComponentType = CompTypeToRegisterComponentType(SigElem->GetCompType());
     Desc.Mask = SigElem->GetColsAsMask();
-    // D3D11_43 does not have MinPrecison.
-    if (m_PublicAPI != PublicAPI::D3D11_43)
-      Desc.MinPrecision = CompTypeToMinPrecision(SigElem->GetCompType());
+    Desc.MinPrecision = CompTypeToMinPrecision(SigElem->GetCompType());
```
bob80905, line 2050: The comment says D3D11_43 doesn't have min precision. So then, what will this be assigned to in the case that D3D11_43 is present? Or, will this code never be reached in that case?

bob80905 (commented):
Just a couple questions.

lib/HLSL/DxilContainerReflection.cpp
tex3d: Because we will manually check against `IUnknown` and the IID of the interface we are supposed to be emulating only.

Before, it would say if you QI for `ID3D12ShaderReflection`, it would be ok, even though our `m_PublicAPI` might be a different version.  When you call a method on the interface, you could get behavior that's incorrect for `ID3D12ShaderReflection`, because we use `m_PublicAPI` to determine certain behaviors later.

lib/HLSL/DxilContainerReflection.cpp
tex3d: We want to recognize an invalid IID early and `return E_INVALIDARG` before creating the object.  Otherwise, it will happily accept any IID and treat unrecognized ones as if they match `ID3D12ShaderReflection` (as it did before).

lib/HLSL/DxilContainerReflection.cpp
tex3d: Our internal Desc is the full `D3D12_SIGNATURE_PARAMETER_DESC` that has min precision, so there's no harm in capturing it in the first place.  The code before was simply leaving that field uninitialized.  Combined with the interface bugs or the later bugs in `Get*ParameterDesc` functions, You could get uninitialized data from this stack copied out to your user struct in a later API call.

The code that copies the struct out to the user's struct is the code that needs to be careful about what it copies, hence the fixes in that code later.  Subtracting the `sizeof(D3D_MIN_PRECISION)` wasn't correct there because of the alignment of the struct size.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -1762,6 +1763,162 @@ TEST_F(DxilContainerTest, DxcUtils_CreateReflection) {
...
+    DxcBuffer buffer = {pReflectionPart->GetBufferPointer(),
+                        pReflectionPart->GetBufferSize(), 0};
+
+    {
+      CComPtr<ID3D12LibraryReflection> pLibraryReflection;
+      VERIFY_SUCCEEDED(
+          pUtils->CreateReflection(&buffer, IID_PPV_ARGS(&pLibraryReflection)));
+
+      CComPtr<ID3D12LibraryReflection> pLibraryReflection2;
+      VERIFY_SUCCEEDED(pLibraryReflection->QueryInterface(IID_PPV_ARGS(&pLibraryReflection2)));
```
bob80905, line 1801: Was this really failing before your change? I think if not, it can be removed, this seems like very basic functionality.

bob80905 (approved):
Other than the removal of QI'ing the same interface, which I imagine has been tested in other places, LGTM

tools/clang/unittests/HLSL/DxilContainerTest.cpp
tex3d: Made a change to `CreateDxilLibraryReflection`, might as well test some basic QueryInterface while we're at it for completeness.

bob80905 (approved):

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -285,6 +287,8 @@ HRESULT CreateDxilShaderReflection(const DxilProgramHeader *pProgramHeader, cons
 HRESULT CreateDxilLibraryReflection(const DxilProgramHeader *pProgramHeader, const DxilPartHeader *pRDATPart, REFIID iid, void **ppvObject) {
   if (!ppvObject)
     return E_INVALIDARG;
+  if (!IsEqualIID(__uuidof(ID3D12LibraryReflection), iid))
+    return E_INVALIDARG;
```
damyanp, line 291: E_NOINTERFACE would be the usual error to return from factory functions like this where the object being created doesn't support the requested interface.

damyanp (approved):

lib/HLSL/DxilContainerReflection.cpp
tex3d: Thanks for this.  As it turns out, I realized that `IUnknown` should be allowed for these create functions as well.  There's regression risk for disallowing them, and we know whether it's a shader or library from the blob, so we don't need it for that.  For shader reflection, if `IUnknown` is used on create, the version defaults to the latest (D3D12).

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -2428,7 +2430,8 @@ HRESULT DxilShaderReflection::GetInputParameterDesc(UINT ParameterIndex,
   else
     memcpy(pDesc, &m_InputSignature[ParameterIndex],
            // D3D11_43 does not have MinPrecison.
-           sizeof(D3D12_SIGNATURE_PARAMETER_DESC) - sizeof(D3D_MIN_PRECISION));
+           offsetof(D3D12_SIGNATURE_PARAMETER_DESC, Stream) +
+               sizeof(D3D12_SIGNATURE_PARAMETER_DESC::Stream));
```
pow2clk, line 2439: I'm guessing the simpler approach of using `offsetof(D3D12_SIGNATURE_PARAMETER_DESC, MinPrecision)` for the size fails because of alignment issues?

lib/HLSL/DxilContainerReflection.cpp
damyanp: The way I usually implement this sort of thing is to create the object and then use QueryInterface to return the appropriate interface from it.  That way we don't need to update it if the set of interfaces supported by the object changes.

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -180,16 +182,14 @@ class DxilShaderReflection : public DxilModuleReflection, public ID3D12ShaderRef
...
-      if (api == m_PublicAPI) {
-        *ppvObject = (ID3D12ShaderReflection *)this;
-        this->AddRef();
-        hr = S_OK;
-      }
+    HRESULT hr = E_NOINTERFACE;
+    PublicAPI api = IIDToAPI(iid);
+    // ID3D11ShaderReflection is identical to ID3D12ShaderReflection, except
+    // for some shorter data structures in some out parameters.
+    if (api == m_PublicAPI || IsEqualIID(__uuidof(IUnknown), iid)) {
```
pow2clk, line 189: This is my ignorance of what QI is meant to do. Given IUnknown, it should just return whatever it has because everything matches. Is that right?

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -1762,6 +1763,274 @@ TEST_F(DxilContainerTest, DxcUtils_CreateReflection) {
   }
 }
 
+TEST_F(DxilContainerTest, CheckReflectionQueryInterface) {
+  if (m_ver.SkipDxilVersion(1, 5))
```
pow2clk, line 1767: Could we get a comment explaining why this version cutoff is needed?

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -1762,6 +1763,274 @@ TEST_F(DxilContainerTest, DxcUtils_CreateReflection) {
...
+    return;
+
+  // Check that QueryInterface for shader and library reflection accepts/rejects
+  // interfaces properly
+
+  // Also check that DxilShaderReflection::Get*ParameterDesc methods do not
+  // write out-of-bounds for earlier interface version.
+
+  // Use domain shader to test DxilShaderReflection::Get*ParameterDesc because
+  // it can use all three signatures.
```
pow2clk, line 1777: Patch constants are part of hull shaders. I see the query is successful, but is it returning anything? There's no patch constant function designated in the HLSL source.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -1762,6 +1763,274 @@ TEST_F(DxilContainerTest, DxcUtils_CreateReflection) {
...
+
+  CComPtr<IDxcUtils> pUtils;
+  VERIFY_SUCCEEDED(m_dllSupport.CreateInstance(CLSID_DxcUtils, &pUtils));
+
+  CComPtr<IDxcCompiler> pCompiler;
+  VERIFY_SUCCEEDED(CreateCompiler(&pCompiler));
+
+  struct TestDescStruct {
+    D3D12_SIGNATURE_PARAMETER_DESC paramDesc;
+    uint32_t pad;
```
pow2clk, line 1812: Understanding the code below makes it clear why this pad is needed, but I think it would be helpful to add a comment here.

pow2clk (commented):
Questions in the code. Concerns in the test. 😟

lib/HLSL/DxilContainerReflection.cpp
tex3d: Right, happens to be equivalent in this case, but generally, if there was any padding between these fields, we wouldn't want to write into that.  So, I feel like the structure of this statement more accurately reflects the intent (only write to the end of the Stream field).

lib/HLSL/DxilContainerReflection.cpp
tex3d: You should be able to QI `IUnknown` on any interface derived from such.  I'm not sure whether that answers your question, but then maybe I don't know what you're getting at.  In order to use the `IUnknown` result beyond the interface methods there, you'll have to QI for another interface, at which point we check that the interface matches our state.

damyanp: If it is IUnknown then we should be doing `*ppvObject = (IUnknown*)this`.

damyanp (changes requested):

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -1762,6 +1763,296 @@ TEST_F(DxilContainerTest, DxcUtils_CreateReflection) {
...
+          pUtils->CreateReflection(&buffer, IID_PPV_ARGS(&pShaderReflection)));
+    }
+  }
+
+  // Supported shader reflection interfaces defined by legacy APIs
+  const GUID IID_ID3D11ShaderReflection_43 = {
+      0x0a233719,
+      0x3960,
+      0x4578,
+      {0x9d, 0x7c, 0x20, 0x3b, 0x8b, 0x1d, 0x9c, 0xc1}};
```
coopp, line 1893: Should these come in from a header, instead of defining them here?

lib/HLSL/DxilContainerReflection.cpp
tex3d: That's what we normally do, but for shader reflection, we have some code to capture the initial interface version you use to initialize the reflection object versioning.  This is so it behaves as expected for the interface version.  You're then not allowed to QI to a different interface version, since that interface would misbehave, since we don't support multiple interfaces of different versions on the same underlying reflection object.  There would be another way to implement this, like with different interface implementations for each version, delegating most things to a central implementation and adapting for interface differences, but that's more complicated, and not the way this was implemented.  Nor is it likely that someone using one version of the reflection interface is going to turn around and use another version of the reflection interface through QI.  These normally would have been defined based on the SDK you linked with.  So, there is a check for the interface used before even creating the reflection, skipping it altogether if it's not valid.

This pattern is carried over to library reflection creation, where there's only one interface version.  It could be ripped out, and that would change it to creating the reflection, then releasing it if you provided the wrong interface.  However, if we version the library reflection interface to add some needed properties (not unlikely given current requests for work graphs), we might need a similar mechanism to the shader reflection versioning, and thus a similar sequence of checking the version before creating the reflection object.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -1762,6 +1763,296 @@ TEST_F(DxilContainerTest, DxcUtils_CreateReflection) {
...
+      VERIFY_IS_TRUE(testParamDesc.CheckBytesAfterParamDesc());
+      VERIFY_SUCCEEDED(pShaderReflection->GetOutputParameterDesc(
+          0, &testParamDesc.paramDesc));
+      VERIFY_IS_TRUE(testParamDesc.CheckBytesAfterParamDesc());
+      VERIFY_SUCCEEDED(pShaderReflection->GetPatchConstantParameterDesc(
+          0, &testParamDesc.paramDesc));
+      VERIFY_IS_TRUE(testParamDesc.CheckBytesAfterParamDesc());
+    }
+
+    { // Verify with initial interface IID_ID3D11ShaderReflection_47
```
coopp, line 1958: Was this intended?  The code below defines a CComPtr for a ID3D12ShaderReflection interface.  Then calls CreateReflection( ), asking for IID_ID3D11ShaderReflection_47.  That seems odd to me.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
tex3d: This was copied from another reflection test, but this one can probably get away with a min version of 1.3 (for library support).  I'm updating it, along with a comment.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
tex3d: Patch constants can be read (as input) in the domain shader, and that's all we are compiling.  Since we aren't compiling the hull shader, no patch constant function is used.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
tex3d: Will do, thanks.

lib/HLSL/DxilContainerReflection.cpp
damyanp: I've found the "ID3D11ShaderReflection is identical to ID3D12ShaderReflection, except..." comment extremely confusing while reading this.  This implementation of QueryInterface is pretty complicated and unusual compared to most others you'd encounter implementing COM objects.

I think some key points here:
* although everything uses the same vtable, as from `ID3D12ShaderReflection`, there are differences in behavior depending on the API version, and there are 3 of these - it's not just d3d11 vs d3d12.
* when the object is created the API version is fixed
* from that point on, this object can only be QI'd for the matching API version.

lib/HLSL/DxilContainerReflection.cpp
damyanp: It'd be great to capture some of this in the code because it is extremely unusual and not at all obvious from reading it.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
tex3d: Yes, this is intentional.  I don't have any other version of the interface defined in any header I would want to include, and it shouldn't matter.  I only use a common part of the interface and verify the versioned behavior of those methods.  The difference between interface versions comes down to an additional signature element field in the structure pointer passed to the `Get*ParameterDesc` methods.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
tex3d: They aren't defined in any header I could include, and they aren't versioned the way I need.  Normally, you'd link to one SDK version and the GUID for ID3D11ShaderReflection would be one of these, depending on your SDK version.  I don't think there's any accessible standard way of using that here.  This is the simplest way to test behavior differences that depend on which interface GUID you used to create the reflection.  These GUIDs are similarly defined in DxilContainerReflection.cpp for the same reason.

lib/HLSL/DxilContainerReflection.cpp
tex3d: > If it is IUnknown then we should be doing *ppvObject = (IUnknown*)this.

What difference could this make without weird multiple inheritance situations?

> I've found the "ID3D11ShaderReflection is identical to ID3D12ShaderReflection, except..." comment extremely confusing while reading this.

Would you be satisfied with the point-form explanation you just wrote instead?

damyanp: IMO, too much of this code is based on (currently valid) assumptions about equivalent vtable layouts of various different interfaces that have appeared in various different iterations of the SDK.  I think there's value in idiomatic, non-surprising code.

I think the explanation you wrote in another thread helped capture it as well, but some acknowledge in the code that this is not your normal implementation of a COM object would be helpful.

damyanp (approved):

pow2clk (approved):
Thanks for the updates!

PR #5446, opened 2023-07-21
[SPIR-V] Handle non-global ConstantBuffer

### LINKED ISSUE #5401: [SPIR-V] Accessing a ConstantBuffer returned from a function throws an exception
**Description**
Ideally, I'd like to wrap my bindless indices using a Handle<> struct that exposes a function to return the resource it's meant to represent, but that is currently not possible, as returning a ConstantBuffer from a function throws an exception.
Accessing the array element manually or changing the buffer type to a StructuredBuffer works as a workaround, but is suboptimal.

**Steps to Reproduce**
``` hlsl
struct Indices
{
    uint texIndex;
    uint bufIndex;
};             
[[vk::push_constant]]
Indices indices;

struct S
{
    float someFloat;
    int someInt;
};

[[vk::binding(0, 0)]]
Texture2D<float4> textures[];
[[vk::binding(0, 1)]]
SamplerState samplers[];
[[vk::binding(0, 2)]]
ConstantBuffer<S> buffers[];

Texture2D<float4> getTex(uint indx)
{
    return textures[indx];
};
ConstantBuffer<S> getBuf(uint indx)
{
    return buffers[indx];
};

struct VSOutput
{
    [[vk::location(0)]] float2 uv : TEXCOORD0;
};
float4 main(VSOutput input) : SV_TARGET
{
    float4 color = 1.0;

    color += textures[indices.texIndex].SampleLevel(samplers[0], input.uv, 0); // Works
    // color += getTex(indices.texIndex).SampleLevel(samplers[0], input.uv, 0); // Works
    color += buffers[indices.bufIndex].someFloat; // Works
    // color += getBuf(indices.bufIndex).someFloat; // crashes: Internal compiler error: access violation. Attempted to read from address 0xFFFFFFFFFFFFFFFF
    
    return color;
}
```
Compiled using
```
dxc BufReturn.hlsl -E main -HV 2021 -T ps_6_6 -spirv -fspv-target-env=vulkan1.3
```

**Environment**
- dxcompiler.dll: 1.7 - 1.7.0.3966 (6289ba317)
- Windows 10

cassiebeckley:
Previously, `ConstantBuffer<T>` in variable declarations were being lowered in `SpirvEmitter`, and in all other cases in `LowerTypeVisitor`, which was improperly lowering the type. This pull request changes the existing logic so only `ConstantBuffer<T>` declarations without an initializer (ie, global declarations) are lowered in `SpirvEmitter`, and adds logic to lower `ConstantBuffer<T>` as `T` in `LowerTypeVisitor`.

This does result in two separate type declarations for `ConstantBuffer<T>`, but after legalization the output is correct in the cases I've checked.

Fixes #5401

### DIFF
--- a/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
+++ b/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
@@ -761,6 +761,26 @@ const SpirvType *LowerTypeVisitor::lowerResourceType(QualType type,
     return valType;
   }
 
+  if (name == "ConstantBuffer") {
+    // ConstantBuffer<T> is lowered as T
+    if (rule == SpirvLayoutRule::Void) {
+      rule = getCodeGenOptions().sBufferLayoutRule;
+    }
+
+    // Get the underlying resource type.
+    const auto s = hlsl::GetHLSLResourceResultType(type);
+
+    // If the underlying type is a matrix, check majorness.
+    llvm::Optional<bool> isRowMajor = llvm::None;
+    if (isMxNMatrix(s))
+      isRowMajor = isRowMajorMatrix(spvOptions, type);
+
+    // Lower the underlying type.
+    const auto *structType = lowerType(s, rule, isRowMajor, srcLoc);
+
+    return structType;
+  }
+
   // ByteAddressBuffer types.
   if (name == "ByteAddressBuffer") {
     const auto *bufferType =
--- a/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
+++ b/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
@@ -519,36 +519,7 @@ const SpirvType *LowerTypeVisitor::lowerType(QualType type,
       return spvType;
     }
 
-    // Collect all fields' information.
-    llvm::SmallVector<HybridStructType::FieldInfo, 8> fields;
-
-    // If this struct is derived from some other struct, place an implicit
-    // field at the very beginning for the base struct.
-    if (const auto *cxxDecl = dyn_cast<CXXRecordDecl>(decl)) {
-      for (const auto &base : cxxDecl->bases()) {
-        fields.push_back(HybridStructType::FieldInfo(base.getType()));
-      }
-    }
-
-    // Create fields for all members of this struct
-    for (const auto *field : decl->fields()) {
-      llvm::Optional<BitfieldInfo> bitfieldInfo;
-      if (field->isBitField()) {
-        bitfieldInfo = BitfieldInfo();
-        bitfieldInfo->sizeInBits =
-            field->getBitWidthValue(field->getASTContext());
-      }
-
-      fields.push_back(HybridStructType::FieldInfo(
-          field->getType(), field->getName(),
-          /*vkoffset*/ field->getAttr<VKOffsetAttr>(),
-          /*packoffset*/ getPackOffset(field),
-          /*RegisterAssignment*/ nullptr,
-          /*isPrecise*/ field->hasAttr<HLSLPreciseAttr>(),
-          /*bitfield*/ bitfieldInfo));
-    }
-
-    auto loweredFields = populateLayoutInformation(fields, rule);
+    auto loweredFields = lowerStructFields(decl, rule);
 
     const auto *spvStructType =
         spvContext.getStructType(loweredFields, decl->getName());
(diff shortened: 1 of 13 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -1789,7 +1789,7 @@ void SpirvEmitter::doVarDecl(const VarDecl *decl) {
     return;
   }
 
-  if (isConstantTextureBuffer(decl->getType())) {
+  if (isConstantTextureBuffer(decl->getType()) && !decl->hasInit()) {
```
s-perron, line 1792: Why is `hasInit` the correct condition? Your code seems to fail with this example https://godbolt.org/z/WGjTzjc61.

tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
```diff
@@ -761,6 +761,26 @@ const SpirvType *LowerTypeVisitor::lowerResourceType(QualType type,
     return valType;
   }
 
+  if (name == "ConstantBuffer") {
+    // ConstantBuffer<T> is lowered as T
+    if (rule == SpirvLayoutRule::Void) {
+      rule = getCodeGenOptions().sBufferLayoutRule;
+    }
```
s-perron, line 768: I don't think you need this. This will cause the type to get the offset and other decorations. They are not needed for the local type. This is done for some other resource types because their "local" version are actually pointers to the global type, so the rule is change to get the correct type for the pointee.

tools/clang/test/CodeGenSPIRV/type.constant-buffer.assign.hlsl
```diff
@@ -0,0 +1,32 @@
...
+
+// CHECK: %buffer = OpVariable %_ptr_Uniform_type_ConstantBuffer_S Uniform
+ConstantBuffer<S> buffer;
+
+void main()
+{
+// CHECK:           %local = OpVariable %_ptr_Function_S Function
+
+// CHECK: [[val:%\d+]] = OpLoad %type_ConstantBuffer_S %buffer
+// CHECK:                OpStore %local [[val]]
```
s-perron, line 30: We are very selective about the types of invalid code that we are willing to generate. I'm wondering if in this case we are missing a call to 

https://github.com/microsoft/DirectXShaderCompiler/blob/15c87234584676574ca4a4ebebde03aab9822994/tools/clang/lib/SPIRV/SpirvEmitter.h#L203-L207

Can you see if you can modify the generated code to have it store a value of the correct type?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
s-perron: This might be the better condition to check:

https://github.com/microsoft/DirectXShaderCompiler/blob/15c87234584676574ca4a4ebebde03aab9822994/tools/clang/lib/SPIRV/SpirvEmitter.h#L97-L105

s-perron (changes requested):
Thanks for looking into this. This is a good starts, but I think there are a couple small change that will make it more robust.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: Yeah, that's much better, thanks

tools/clang/test/CodeGenSPIRV/type.constant-buffer.assign.hlsl
cassiebeckley: It turns out that this kind of invalid code was already being generated for `ConstantBuffer` and `TextureBuffer` in some cases; I believe these cases no longer require validation.

tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
cassiebeckley: I believe now that this is being used to generate both the extern variable and the local ones that it's best to set this as `cBufferLayoutRule` or `tBufferLayoutRule` according to the buffer type, so that only one type is generated.

s-perron (approved):

s-perron:
@sudonatalie can you do the second review? I did a fairly quick one.

sudonatalie (approved):
LGTM!

PR #5451, opened 2023-07-25
Fix attribute collision for HL intrinsics

### LINKED ISSUE #3505: Inline Raytrace RayQuery::CommittedTriangleFrontFace Not Working
When using inline raytrace, after calling proceed on a ray query, using the call to RayQuery::CommittedTriangleFrontFace makes all the other calls to RayQuery::Committed*** return zeros or invalid. Not using the call to RayQuery::CommittedTriangleFrontFace makes the other calls return meaningful values.

tex3d:
HL Intrinsic functions share declarations with those that match group and function signature, regardless of the original intrinsic name. This means that intrinsics with differing attributes can be collapsed into the same HL functions, leading to incorrect attributes for some HL intrinsics.

This fixes this issue by adding the attributes to the HL operation mangling, the same way this issue was fixed for the HLWaveSensitive attribute before.

Fixes #3505


python3kgae (approved):

lib/HLSL/HLOperations.cpp
```diff
@@ -509,9 +509,15 @@ Function *GetOrCreateHLFunction(Module &M, FunctionType *funcTy,
...
-        mangledNameStr << "wave";
+    // Need to add attributes to name to prevent clashes with intrinsics with
+    // different attributes
+    if (attribs.hasAttribute(AttributeSet::FunctionIndex, Attribute::ReadNone))
+      mangledNameStr << "rn";
+    else if (attribs.hasAttribute(AttributeSet::FunctionIndex,
+                                  Attribute::ReadOnly))
+      mangledNameStr << "ro";
+    if (attribs.hasAttribute(AttributeSet::FunctionIndex, HLWaveSensitive))
+      mangledNameStr << "wave";
```
bogner, line 520: What about other attributes? I see we ignore `nounwind` based on the tests, but maybe we should do that explicitly and warn or error or something if we see something odd, otherwise this is just going to pop up again.

lib/HLSL/HLOperations.cpp
```diff
@@ -509,9 +509,15 @@ Function *GetOrCreateHLFunction(Module &M, FunctionType *funcTy,
   }
   else {
     mangledNameStr << GetHLFullName(group, opcode);
-    // Need to add wave sensitivity to name to prevent clashes with non-wave intrinsic
-    if(attribs.hasAttribute(AttributeSet::FunctionIndex, HLWaveSensitive))
-        mangledNameStr << "wave";
+    // Need to add attributes to name to prevent clashes with intrinsics with
+    // different attributes
+    if (attribs.hasAttribute(AttributeSet::FunctionIndex, Attribute::ReadNone))
+      mangledNameStr << "rn";
```
bogner, line 515: I'm surprised there are no test updates with `rn` - do we not have any ops that have this attribute or do we just not have any tests for them? Similarly, it'd be nice to have tests for cases with multiple attributes if they exist.

pow2clk: The issue here is mistakenly temporarily treating HL intrinsics that have the same signature as the same intrinsic, which causes attributes as defined in gen_intrin_main to bleed into those that match. The only attributes there are `ro`, `rn`, and `wv`. There are none that have more than one.

lib/HLSL/HLOperations.cpp
pow2clk: The only relevant attributes for this issue are those defined in gen_intrin_main

tex3d: `nounwind` is always added to these intrinsic functions, so isn't presumed to be a dimension for overloading.  Since these are just for mangling the built-in intrinsics, there is an assumption that the attributes used are limited to a certain set.

However, I do see that we have special cases where we add attributes under `SetHLFunctionAttribute()`, including potentially `NoDuplicate`.  That isn't accounted for in this code.

Ideally, one place would define the correct attributes on the original functions, and the mangling would follow, but this code isn't written to be able to do that, since special cases for attributes are applied after the mangling was defined, and the function looked up.  I think combining the attribute adding and attribute mangling code into the same function would prevent them from getting out of sync so easily, at least.

Some of these attributes are definable in the intrinsic table input generated from `gen_intrin_main.txt` (ro/rn/wave), and I think they all should be (at least the ones that could be different).  There also should be use of `ArgMemOnly` in SM 6.8 branch, and another old PR I had up here changed the way the attributes are defined in the intrinsic tables to allow for easier expansion of attributes as needed.  But some of these functions don't come from the intrinsic tables, and are generated elsewhere as built-in operators or intermediate operations, and some of those are special-cased in `SetHLFunctionAttribute()`.

tools/clang/test/HLSLFileCheck/hlsl/objects/RayQuery/readnone_intrinsicAttributeCollision.hlsl
```diff
@@ -0,0 +1,41 @@
...
+
+// Ensure EvaluateAttributeCentroid not eliminated in final DXIL:
+// CHECKDXIL: call float @dx.op.evalCentroid.f32(
+// CHECKDXIL: call float @dx.op.quadOp.f32(
+
+StructuredBuffer<int> buf[]: register(t3);
+RWStructuredBuffer<int> uav;
+
+// test read-none attr
+int PS(float4x4 a : A, int b : B) : SV_Target
```
pow2clk, line 27: Mildly curious why this became a matrix, but it's fine either way.

tools/clang/test/HLSLFileCheck/hlsl/objects/RayQuery/wave_intrinsicAttributeCollision.hlsl
```diff
@@ -0,0 +1,46 @@
+// RUN: %dxc -T ps_6_5 -E PS -fcgl %s | FileCheck %s
+// RUN: %dxc -T ps_6_5 -E PS %s | FileCheck %s -check-prefix=CHECKDXIL
```
pow2clk, line 2: It seems we allow QuadAny with shader models earlier than 6.7 😣

We shouldn't though. One of us should file a bug for that and this test should be updated to require 6.7

pow2clk (approved):
Please update the QuadAny test to require 6.7 and then this is good to go!

tools/clang/test/HLSLFileCheck/hlsl/objects/RayQuery/wave_intrinsicAttributeCollision.hlsl
hekota: Don't we have a fallback for QuadAny on shader models earlier than 6.7?

tools/clang/test/HLSLFileCheck/hlsl/objects/RayQuery/readnone_intrinsicAttributeCollision.hlsl
bob80905: I think I was pulling from some other tests to get an idea of how to properly get inputs into a pixel shader. I simplified to float regardless.

tools/clang/test/HLSLFileCheck/hlsl/objects/RayQuery/wave_intrinsicAttributeCollision.hlsl
bob80905: Made an issue here:
https://github.com/microsoft/DirectXShaderCompiler/issues/5521

PR #5452, opened 2023-07-25
Fix TAEF lookup for nuget and SDK paths

llvm-beanz:
The standard TAEF install paths use the x64 and x86 paths just like the TAEF_ARCH. I'm not entirely sure where the alternative TAEF_BIN_ARCH naming is used, but it seems to be used in the custom tooling for fetching and building TAEF.

This only alters the parts of the code that I've been able to verify locally.


hekota (approved):
LGTM!

Issue #5368, opened 2023-06-28
Move more diagnostics earlier, changing tests to verifier tests


llvm-beanz:
Work graphs introduces new diagnostics emitted during code generation. These should be performed in Sema and the associated tests should be verifier tests.

See internal work item 131 for more information.

bob80905:
Fixed by #5484

PR #5453, opened 2023-07-25
[Verifier] Move shader attribute diagnostics earlier.

bob80905:
Diagnostics for invalid shader attributes to shader functions were appearing too late. This PR moves the emission of these diagnostics earlier so the verifier can catch them, and adds some tests to accompany the changes.
This PR is one of a few that will help implement this task:
https://github.com/microsoft/DirectXShaderCompiler/issues/5368


tools/clang/test/HLSL/shader_cvc_mismatch.hlsl
```diff
@@ -0,0 +1,9 @@
+// RUN: %clang_cc1 -fsyntax-only -ffreestanding -verify %s
+
+
+[shader("compute")]
+[shader("vertex")]
+[shader("compute")]
```
pow2clk, line 6: Should we allow two `shader("compute")` attributes and nothing else? That seems weird, but valid.

python3kgae (approved):

tools/clang/include/clang/Sema/Sema.h
```diff
@@ -2978,6 +2978,8 @@ class Sema {
                                    ObjCMethodDecl *Overridden,
                                    bool IsProtocolMethodDecl);
 
+  void ValidateShaderAttribtues(Scope *S, Decl *D, const AttributeList *A);
```
bogner, line 2981: typo. `s/Attribtues/Attributes/`

tools/clang/lib/Sema/SemaDeclAttr.cpp
```diff
@@ -5097,14 +5097,65 @@ static void ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D,
...
+        hasComputeAttr = true;
+        attrNum++;
+      } else if (str == "node") {
+        hasNodeAttr = true;
+        attrNum++;
+      } else {
+        for (auto shaderAttrStr :
+             {"vertex", "pixel", "hull", "domain", "geometry", "raygeneration",
+              "intersection", "anyhit", "closesthit", "miss", "callable",
+              "mesh", "amplification"}) {
```
pow2clk, line 5125: InB4: Is there a way to compare these without using strings that wouldn't involve a major reworking of how the attribs are stored? I suspect the answer is no, but I think it needs to be stated.

tools/clang/lib/Sema/SemaDeclAttr.cpp
```diff
@@ -5097,14 +5097,65 @@ static void ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D,
...
+          if (str == shaderAttrStr) {
+            attrNum++;
+            break;
```
pow2clk, line 5128: I think it would be better to keep a note of the unique attribute that you've already seen and emit an error if you find another one here that conflicts. This woudl avoid having to perform any additional checks. 

So if you had a `lastAttr` string initialized to empty, then when you found one of these, you could:
* assign it if the string is empty
* skip if it's identical
* skip if one is node and the other is compute
* error in any other case.

tools/clang/lib/Sema/SemaDeclAttr.cpp
```diff
@@ -5097,14 +5097,65 @@ static void ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D,
...
+  for (const AttributeList *l = A; l; l = l->getNext()) {
+    if (l->getKind() == AttributeList::AT_HLSLShader) {
+      Expr *E = l->getArgAsExpr(0);
+      if (E->isTypeDependent() || E->isValueDependent() ||
+          E->getStmtClass() != Stmt::StringLiteralClass) {
+        break;
+      }
+
+      StringLiteral *sl = cast<StringLiteral>(E);
+      std::string str = sl->getString();
```
bogner, line 5114: Why use a `std::string` here? `st->getString()` returns a `StringRef` which should be sufficient here and is more efficient.

tools/clang/lib/Sema/SemaDeclAttr.cpp
```diff
@@ -5097,14 +5097,65 @@ static void ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D,
   }
 }
 
+void Sema::ValidateShaderAttribtues(Scope *S, Decl *D, const AttributeList *A) {
```
llvm-beanz, line 5100: I don't think this is the right place to do this validation. We should probably do it off `HandleDeclAttributeForHLSL`

pow2clk (commented):
Good tests!

I'd like to see one with duplicates of the same shader types and maybe one with duplicates of compute and node types similar to cvc, but cnc say and it should pass

tools/clang/lib/Sema/SemaDeclAttr.cpp
```diff
@@ -5097,14 +5097,65 @@ static void ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D,
...
+  bool hasComputeAttr = false;
+  bool hasNodeAttr = false;
+  int attrNum = 0;
+
+  for (const AttributeList *l = A; l; l = l->getNext()) {
```
llvm-beanz, line 5105: This pattern is also not great. Rather than validating this after we've constructed an attribute list, we can do it during construction.

If you look at how `handleWorkGroupSize` is implemented. Basically it checks the decl to see if an attribute specifying the work group has already been attached, if it has then it verifies the work group size matches the new specification.

In this case you could see if there is already an attribute, and if there is verify that it is the `node`/`compute` combination that is valid, otherwise issue a diagnostic.

tools/clang/lib/Sema/SemaDeclAttr.cpp
```diff
@@ -5097,14 +5097,65 @@ static void ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D,
...
+      } else {
+        for (auto shaderAttrStr :
+             {"vertex", "pixel", "hull", "domain", "geometry", "raygeneration",
+              "intersection", "anyhit", "closesthit", "miss", "callable",
+              "mesh", "amplification"}) {
+          if (str == shaderAttrStr) {
+            attrNum++;
+            break;
+          }
+        }
```
bogner, line 5130: I think you can structure this slightly differently in a way that will both simplify it and make it more maintainable if we add other attributes. Instead of tracking `hasComputeAttr` and `hasNodeAttr`, we could keep a `shaderStage` (which is either empty or one of "compute", "vertex", etc), and `hasNodeAttr`. Then when we see a shader stage we emit the error if 1) there already is one, or 2) hasNodeAttr is true and the stage is not "compute". When we see "node" we emit an error if the shader stage is set and not "compute".

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7471,6 +7471,8 @@ def err_hlsl_attribute_expects_string_literal_from_list: Error<
   "attribute %0 must have one of these values: %1">;
 def err_hlsl_attribute_valid_on_function_only: Error<
   "attribute is valid only on functions">;
+def err_hlsl_attribute_mismatch: Error<
+  "Invalid shader stage attribute combination">;
```
llvm-beanz, line 7475: nit: diagnostics should start with lowercase letters

Also, I wonder if there isn't a better way to say this. The existing `warn_duplicate_attribute` warning is similar. We also might want to add a `note` pointing at previous attributes.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12930,6 +12930,40 @@ bool ValidateAttributeTargetIsFunction(Sema& S, Decl* D, const AttributeList &A)
   return false;
 }
 
+void Sema::ValidateShaderAttributes(Decl *D, const AttributeList *A) {
```
llvm-beanz, line 12933: I think you can refactor this code to be much simpler.

The code just above here is actually creating the AST node for the attribute. This block should get hit once for each time the `[shader(...)]` block shows up in the source. Instead of iterating over the list of attributes, you can rely on this being hit over and over again.

If you validate the attribute before creating the AST node (so you don't create the node if it is invalid), you can use the existing AST node as the "lastAttr" value.

That could simplify this code to something more like:

```c++
HLSLShaderAttr *ValidateShaderAttributes(Sema &S, Decl *D, const AttributeList *A) {
  Expr *ArgExpr = A->getArgAsExpr(0);
  StringLiteral *Literal = dyn_cast<StringLiteral>(ArgExpr->IgnoreParenCasts());
  DXIL::ShaderKind Stage = ShaderModel::KindFromFullName(Literal->getString());
  if (Stage == DXIL::ShaderKind::Invalid) {
    S.Diag(...); // Diagnose invalid stage specifier.
    return nullptr; // don't create the attribute
  }
  HLSLShaderAttr *Existing = D->getAttr<HLSLShaderAttr>();
  if (Existing) {
    DXIL::ShaderKind NewStage = ShaderModel::KindFromFullName(Existing->getStage());
   if (Stage == NewStage)
    return nullptr; // don't create, but no error.
    if ((Stage != DXIL::ShaderKind::Compute && Stage != DXIL::ShaderKind::Node) ||
        (NewStage != DXIL::ShaderKind::Compute && NewStage != DXIL::ShaderKind::Node)) {
      S.Diag(A->getLoc(), diag::err_hlsl_attribute_mismatch);
      S.Diag(A->getLoc(), diag::note_conflicting_attribute);
      S.Diag(Existing->getLoc(), diag::note_conflicting_attribute);
      return nullptr;
    }
  }
  return ::new (S.Context) HLSLShaderAttr(
        A.getRange(), S.Context,Literal->getString())
}
```

The idea here is that the function gets called each time there is a parsed `shader` attribute, but it only creates the AST node if they are valid. This would allow something like:

```hlsl
[shader("compute")]
[shader("node")]
[shader("compute")]
[shader("node")]
```

But it should fail for combinations of other shaders. We could additionally de-dupe compute/node shader attributes, but we don't do that today so I don't know that we need to.

tools/clang/test/HLSL/shader_cc_mismatch.hlsl
```diff
@@ -0,0 +1,8 @@
+// RUN: %clang_cc1 -fsyntax-only -ffreestanding -verify %s
+
+
+[shader("compute")] 
+[shader("compute")] 
+[ numthreads( 64, 2, 2 ) ]  /* expected-no-diagnostics */
```
tex3d, line 6: Can we avoid adding so many separate test files?  Part of the point of the verifier tests is that you can verify a lot of related scenarios in one file, rather than having to have a whole bunch of files, one testing each failure case.  You should be able to put all of these into one `shader_attribute.hlsl` verifier test for testing all the shader attribute diagnostics.

tools/clang/unittests/HLSL/VerifierTest.cpp
```diff
@@ -193,6 +194,21 @@ TEST_F(VerifierTest, RunArrayIndexOutOfBounds) {
   CheckVerifiesHLSL(L"array-index-out-of-bounds-HV-2016.hlsl");
 }
 
+TEST_F(VerifierTest, RunShaderMismatch) {
```
tex3d, line 197: If we can group the shader attribute tests into one test called `shader_attribute.hlsl`, then we can call the group a matching name `RunShaderAttribute`, which fits the pattern for the rest of the verifier tests.  Plus, one test group for one file keeps with the existing pattern, and won't break VerifierHelper.py (even though some don't want to keep that alive anyway).

tools/clang/test/HLSL/shader_cc_mismatch.hlsl
llvm-beanz: Alternatively since @python3kgae just deleted the old test running logic, we could make these lit tests using the DXC -verify flag. That eliminates the need for any unit test cases.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: - I think my implementation is sufficiently simple.
- With the way the code is structured, there is no way to gain access to the source location of the previous attribute. My implementation allows access to an accurate source location for the previous attribute because of the usage of a loop. The Decl variable D has no relevant source location to the new attribute being processed, and "Existing" has no source location available either.
- This implementation differs from every other validation function in that it takes responsibility for creating the new attribute object. Because of this, I need to personally set Handled to true if nullptr is returned, and it seems to defeat the purpose of the switch statement that calls this validation function. To match the convention with the rest of the validation in the switch statement I don't think this function should create/return the Attr object.
- My implementation is correct when testing the CNCN case you pasted above.

Unless there's further reason, I think I should leave the code as is.

tools/clang/test/HLSL/shader_cc_mismatch.hlsl
bob80905: I merged the files into two new files. One for expected errors/notes, and another that expects no errors/notes.

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: The implementation I proposed is how clang does attribute verification. The HLSL pattern is not really better.
I'm unsure why the `Existing` attribute in my example wouldn't have a valid source location. It should. If it doesn't that would be a bug.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12930,6 +12930,40 @@ bool ValidateAttributeTargetIsFunction(Sema& S, Decl* D, const AttributeList &A)
...
+  std::string lastAttrStr = "";
+  const AttributeList *lastAttr = nullptr;
+
+  for (const AttributeList *l = A; l; l = l->getNext()) {
```
llvm-beanz, line 12937: A key flaw of this implementation is this loop. You're looping over the attribute list _inside_ a loop over the attribute list. So this is `n^2`, _and_ you risk duplicating checks and diagnostics.

The intent of `HandleDeclAttributeForHLSL` is to handle a single attribute, not a list of them.

tools/clang/lib/Sema/SemaDeclAttr.cpp
```diff
@@ -5039,7 +5039,7 @@ static void ProcessDeclAttribute(Sema &S, Scope *scope, Decl *D,
     handleAcquiredAfterAttr(S, D, Attr);
     break;
 
-  // Capability analysis attributes.
+    // Capability analysis attributes.
```
llvm-beanz, line 5042: There are still some unexpected formatting changes here.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -13288,6 +13322,8 @@ void hlsl::HandleDeclAttributeForHLSL(Sema &S, Decl *D, const AttributeList &A,
             "intersection,anyhit,closesthit,miss,callable,mesh,amplification,"
```
llvm-beanz, line 13322: Worth noting, the creation of the attribute is right here.

The distinction of decoupling validating and creating the attributes is arbitrary. DXC also does something _really_ silly here where it creates attributes regardless of whether or not the attribute being created is valid. You can see that in [this example](https://godbolt.org/z/7Pa6Pe96s) where the shader attribute is created with an empty string as its shader type.

The correct solution here should be to not create the attribute at all.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12930,6 +12930,41 @@ bool ValidateAttributeTargetIsFunction(Sema& S, Decl* D, const AttributeList &A)
   return false;
 }
 
+HLSLShaderAttr* ValidateShaderAttributes(Sema &S, Decl *D,
+                                         const AttributeList &A) {
+  StringRef sRef = ValidateAttributeStringArg(
```
llvm-beanz, line 12935: Is there a reason to use this old validation code instead of just handling this below when we call `KindFromFullName`?

tools/clang/test/HLSL/shader_attribute.hlsl
```diff
@@ -44,3 +44,7 @@ void CVCMain() {
 [shader("pixel")]  /* expected-note {{conflicting attribute is here}} */  /* expected-note {{conflicting attribute is here}} */ 
 void NVPMain() {
 }
+
+[shader("I'm invalid")]   /* expected-error {{attribute 'shader' must have one of these values: compute,vertex,pixel,hull,domain,geometry,raygeneration,intersection,anyhit,closesthit,miss,callable,mesh,amplification,node}} */
```
llvm-beanz, line 48: ❤️

llvm-beanz (approved):
LGTM. Thank you for all your work on this!

Issue #5178, opened 2023-04-26
Dynamically linked CRT breaks deployment


sherief:
It's generally expected to be able to ship non-C++ applications on Windows, and the dxcompiler.dll bundled in the Windows SDK kit doesn't rely on third-party components or C++ runtime DLLs: 
![image](https://user-images.githubusercontent.com/1711646/234519177-ffd95452-a5cd-49f0-b73b-d4181d71e825.png)

At a certain point, this contract has been broken, and if you want to ship an app using a newer dxcompiler.dll or dxil.dll from this repository's releases you are required to also redistribute VC++ runtime libraries: 
![image](https://user-images.githubusercontent.com/1711646/234519757-a5027269-79b8-40a4-9cbc-bf0172cecdff.png)

This poses a few issues and has broken some deployment scenarios:
- Non-C++ apps are now required to distribute VC++ runtime DLLs. The COM interfaces exported by the dxcompiler and dxil DLLs are perfectly accessible from other languages, and apps using these languages run against the Windows SDK versions of these DLLs just fine.
- Standalone apps are now impossible. I discovered this issue while updating an application used to visualize data where the end-user writes simple visualization code that gets compiled to shaders and applied to images at runtime. The app is shipped as a single executable, and loads the dxil and dxcompiler DLLs from embedded resources. Suddenly it started failing on some machines, and it turned out that these machines never had the redist installed. In some deployment environments, it's not easy to just ship and install these due to legal or ops limitations. 
- On deployments where the VC++ redist was installed by some other app, we cannot guarantee which version of the DLL we'll end with. In certain deployment scenario, we require certain integrity checking of the runtime environment and all loaded DLLs, and we even manage loading everything involved with D3D Agility SDK and DX shader compiler manually from known versions that carry their own out-of-band signature for whitelisting. 

I'm not sure what the wins are from moving to a dynamically linked CRT, but I can see some issues and contract breaks. The top support issue by a ratio of 4:1 has now become related to the new DLL requirements. Can we have the artifacts from the latest shader compiler here match the Windows SDK versions when it comes to external requirements?

sherief:
Looks like #5163 is also running into this.

hekota:
Fixed in internal build system. Next DXC release will ship with hybrid CRT linking.

Issue #5163, opened 2023-04-20
dxil.dll: Use Hybrid CRT linkage or static CRT in releases


driver1998:
Currently the dxc releases links to the MSVC runtime dynamically, which requires MSVC Redist packages installed in order to work.

We (MSYS2) have been building and publishing Mesa3D with GLon12 for a while and found out that dxil.dll is required for it to function. 

We plan to package and redistribute dxil.dll ourselves (https://github.com/msys2/MINGW-packages/pull/16643). But the releases here requires MSVC redistributable DLLs and that would be problematic.

The one provided Windows SDK is static linked so it may be used, but that requires us to manually copy the DLLs from somebody's computer with the SDK installed, and manually manage updates. Not an ideal solution either.

So I would want to have releases here to link statically to CRT as well. 

If binary sizes are a concern, you may try [Hybrid CRT linkage](https://github.com/microsoft/WindowsAppSDK/blob/main/docs/Coding-Guidelines/HybridCRT.md) introduced by Windows App SDK, that way only the C++ runtime is linked statically, and UCRT is still linked dynamically, and uses the one built-in to the OS.

hekota:
Fixed in internal build system.

PR #5454, opened 2023-07-25
Allow new/delete operators to be called before DllMain

hekota:
Allow new/delete operators to be called before DllMain and allocator initialization in DxcInitThreadMalloc().

The operators can be called before DllMain from CRT libraries when static linking is enabled. If that happens, the new/delete operators will fallback to the standard allocator and use CoTaskMemAlloc/Free directly instead of CoGetMalloc, Alloc/Free & Release for better perf.

Enables fixing of [#5163](https://github.com/microsoft/DirectXShaderCompiler/issues/5163) and [#5178](https://github.com/microsoft/DirectXShaderCompiler/issues/5178).

### DIFF
--- a/lib/DxcSupport/dxcmem.cpp
+++ b/lib/DxcSupport/dxcmem.cpp
@@ -130,7 +130,7 @@ void DxcDelete(void *ptr) throw() {
 #if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
     CoTaskMemFree(ptr);
 #else
-    free(size);
+    free(ptr);
 #endif
   }
 }
--- a/lib/DxcSupport/dxcmem.cpp
+++ b/lib/DxcSupport/dxcmem.cpp
@@ -109,11 +109,7 @@ void* DxcNew(std::size_t size) throw() {
     // where the g_pDefaultMalloc is initialized, for example from CRT libraries when
     // static linking is enabled. In that case fallback to the standard allocator
     // and use CoTaskMemAlloc directly instead of CoGetMalloc, Alloc & Release for better perf.
-#if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
     ptr = CoTaskMemAlloc(size);
-#else
-    ptr = malloc(size);
-#endif
   }
   return ptr;
 }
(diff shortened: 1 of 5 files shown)


lib/DxcSupport/dxcmem.cpp
```diff
@@ -98,3 +98,39 @@ DxcThreadMalloc::DxcThreadMalloc(IMalloc *pMallocOrNull) throw() {
...
+#if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
+    ptr = CoTaskMemAlloc(size);
+#else
+    ptr = malloc(size);
+#endif
+  }
+  return ptr;
+}
+
+void DxcDelete(void *ptr) {
```
llvm-beanz, line 121: Is this safe? If an allocation gets allocated before `DllMain` and de-allocated after a custom allocator is enabled, what happens? Does that route the free to the custom allocator?

hekota: Alloc/free should always happen in the same context. The custom allocator is only enabled inside of compiler API calls. Every public API implementation uses `DxcThreadMalloc` local var to make sure the custom allocator is set to the thread-local storage for use and then re-set back to the previous one on exit.
@tex3d  - please correct me if I am wrong: The situation where something gets allocated by the default allocator (before or after `DllMain`) and de-allocated with a custom allocator should not happen.

llvm-beanz (approved):

lib/DxcSupport/dxcmem.cpp
```diff
@@ -98,3 +98,39 @@ DxcThreadMalloc::DxcThreadMalloc(IMalloc *pMallocOrNull) throw() {
...
+  void *ptr;
+  IMalloc* iMalloc = DxcGetThreadMallocNoRef();
+  if (iMalloc != nullptr) {
+    ptr = iMalloc->Alloc(size);
+  } else {
+    // DxcGetThreadMallocNoRef() returning null means the operator is called before DllMain
+    // where the g_pDefaultMalloc is initialized, for example from CRT libraries when
+    // static linking is enabled. In that case fallback to the standard allocator
+    // and use CoTaskMemAlloc directly instead of CoGetMalloc, Alloc & Release for better perf.
+#if defined(_WIN32) && !defined(DXC_DISABLE_ALLOCATOR_OVERRIDES)
```
pow2clk, line 112: Not strictly necessary for non-win platforms as they already have `CoTaskMemAlloc` and `CoTaskMemFree` defined to malloc and free in WinAdapter.h:
https://github.com/microsoft/DirectXShaderCompiler/blob/93555aa911171097d72036df15b2e3c5c0c3d893/include/dxc/WinAdapter.h#L46

I'm unsure if this is needed for `DXC_DISABLE_ALLOCATOR_OVERRIDES`. #5296 didn't change the usages of `CoTaskMem*` in `dxcompilerobj.cpp`: 
https://github.com/microsoft/DirectXShaderCompiler/blob/93555aa911171097d72036df15b2e3c5c0c3d893/tools/clang/tools/dxcompiler/dxcompilerobj.cpp#L1510C40-L1510C40

pow2clk (approved):
I think it's harmless, but I'm unsure if the explicit free/malloc usage is actually needed

PR #5467, opened 2023-07-27
Move work graph CodeGen diagnostics to Sema 

### LINKED ISSUE #5352: Make sure node shader inputs are checked in Sema and move related tests to VerifierTest
Node shader inputs should be checked in Sema and related tests can be moved to Verifier tests.

Applied to these tests:
tools/clang/test/HLSLFileCheck/hlsl/workgraph/case071a_broadcasting_groupnodeinputrecords.hlsl
tools/clang/test/HLSLFileCheck/hlsl/workgraph/case067_maxrecursiondepth_toolarge.hlsl
tools/clang/test/HLSLFileCheck/hlsl/workgraph/case052_nodelaunch_invalid.hlsl
tools/clang/test/HLSLFileCheck/hlsl/workgraph/case038_trackrwinputsharing_missing.hlsl
tools/clang/test/HLSLFileCheck/hlsl/workgraph/case035_finishedcrossgroupsharing_thread.hlsl
tools/clang/test/HLSLFileCheck/hlsl/workgraph/case034_finishedcrossgroupsharing_coalescing.hlsl

tcorringham:
Move work graph related diagnostics from CodeGen to Sema in order to provide earlier detection of invalid code, which will help ensure that the generated AST is valid.

Also replace tests for these with verifier tests.

Fixes #5352


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11457,6 +11499,34 @@ class WorkGraphVisitor : public RecursiveASTVisitor<WorkGraphVisitor> {
...
+      }
+      break;
+    case attr::HLSLNodeMaxRecursionDepth:
+      if (cast<HLSLNodeMaxRecursionDepthAttr>(A)->getCount() > 32)
+        S.Diags.Report(A->getLocation(), diag::err_hlsl_maxrecursiondepth_exceeded)
+          << A->getRange();
+      break;
+    case attr::HLSLNumThreads:
+      auto *numThreads = cast<HLSLNumThreadsAttr>(A);
+      if (numThreads->getX() * numThreads->getY() * numThreads->getZ() > 1024)
```
llvm-beanz, line 11523: We should be able to do this verification when we create the AST node rather than in a visitor.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11457,6 +11499,34 @@ class WorkGraphVisitor : public RecursiveASTVisitor<WorkGraphVisitor> {
     }
     return true;
   }
+
+  bool VisitAttr(Attr *A) {
+    switch (A->getKind()) {
+    case attr::HLSLNodeDispatchGrid:
+    case attr::HLSLNodeMaxDispatchGrid:
+      // These attributes are only valid for Broadcasting launch nodes
```
llvm-beanz, line 11507: In general I don’t think we should use an AST visitor to validate these. We can instead just look at the attributes on the Decl after we complete attribute parsing.

llvm-beanz (commented):
We have a (poorly named) `DiagnoseRaytracingEntry` function that gets called on any FunctionDecl that has an `HLSLShaderAttr` attached. We could rename and extend that to more holistically verify shader entry attributes without needing to rely on visitors walking the AST.

tools/clang/lib/Sema/SemaHLSL.cpp
tcorringham: Done.

tools/clang/lib/Sema/SemaHLSL.cpp
tcorringham: For now I have moved these checks up to the FunctionDecl visitor - as a (small) step toward consolidating all function attribute checks into a single verification function.

llvm-beanz:
> I agree that consolidating the function attribute checks to replace the visitors would be a good idea.
> Having started looking at that I think it would be a good idea to have a design for the intended approach before refactoring too much code - could we spin the design & refactor out to a separate issue?

There's a bit of a related change going in with #5484. I'm not sure what your design thoughts are, but the goal there is to hook in after the Decl is completely generated to verify the attributes. It would be nice if we can find a cohesive way to do this all in one way rather than a bunch of different approaches for each diagnostic.

tcorringham:
I think we should try to categorize the diagnostics to determine at what stage they could be generated in dxc - and then try to be consistent in where they are raised. In many cases a diagnostic could be raised in any of several places, but if we group similar ones together it would undoubtedly be neater (at present it is rather ad-hoc).
As a first pass at categories:

- syntax errors (raised as parsed)
- stand-alone attribute values - such as NodeMaxRecursionDepth max value. These could be treated as syntax errors at the point they are parsed, or as semantic errors later. I think we are moving them early?
- combinations - where combinations of attributes are required or illegal, or valid only for certain types of parameter, etc. These have to be done once a function is completely parsed - I think the DiagnoseEntry() from #5484 is intended to fill this role (although currently wouldn't cope with shaders that are both compute and node).
- code dependent - cases such as FinishedCrossGroupSharing() requiring NodeTrackRWInputSharing to be specified for the input record. This requires some traversal of the AST - currently done in an AST visitor (which is easy to code and understand), but may be more efficiently done as part of call graph analysis.
- other - I'm not sure , but I have a feeling there may be some corner cases...
Maybe some stylistic guidelines for diagnostics would aid consistency too.

llvm-beanz:
(cc @pow2clk)

> I think we should try to categorize the diagnostics to determine at what stage they could be generated in dxc - and then try to be consistent in where they are raised. In many cases a diagnostic could be raised in any of several places, but if we group similar ones together it would undoubtedly be neater (at present it is rather ad-hoc).

DXC's a mess in this regard. In general diagnostics should be raised as early as possible at the point where you have enough information to accurately generate the diagnostic. We shouldn't try to say "all diagnostics of a given kind go <somewhere>" because that would lead to something similar to the mess we have.

> As a first pass at categories:
> 
> * syntax errors (raised as parsed)

+1

> * stand-alone attribute values - such as NodeMaxRecursionDepth max value. These could be treated as syntax errors at the point they are parsed, or as semantic errors later. I think we are moving them early?

They are are diagnosed in Sema after attribute argument parsing (unless parsing itself fails), which is how clang handles other attributes so I think that's likely the best approach. DXC breaks some of the generic attribute parsing logic. Fixing that would be great, but in the meantime what we have here isn't awful.

> * combinations - where combinations of attributes are required or illegal, or valid only for certain types of parameter, etc. These have to be done once a function is completely parsed - I think the DiagnoseEntry() from [[Verifier] Add verifier tests to catch MaxRecord errors earlier #5484](https://github.com/microsoft/DirectXShaderCompiler/pull/5484) is intended to fill this role (although currently wouldn't cope with shaders that are both compute and node).

Yea... I think allowing shader entries to be both compute and node entries was a mistake. That introduces a lot of complexity for seemingly little value. I'd really like to discuss changing that when @tex3d is back from vacation.

> * code dependent - cases such as FinishedCrossGroupSharing() requiring NodeTrackRWInputSharing to be specified for the input record. This requires some traversal of the AST - currently done in an AST visitor (which is easy to code and understand), but may be more efficiently done as part of call graph analysis.

I'm a little confused by this. You should be able to verify that the argument passed to `FinishedCrossGroupSharing` has `NodeTrackRWInputSharing` in Sema without an AST traversal. It should be just looking at the call argument's template parameter. If we wanted to diagnose the undefined behavior around use of the argument after the call, that would be trickier.

> * other - I'm not sure , but I have a feeling there may be some corner cases...

We currently have some diagnostics that are emitted during and post optimization (resource resolution). I think the existence of those diagnostics are design flaws in the language. We shouldn't allow invalid code to be written in the hopes that it gets optimized away. I don't want us expanding those diagnostics because I'd like them to go away.

We also have verifier diagnostics. I think those should always be a last line of defense. We should strive to make the compiler diagnose anything and everything that could become a verifier error so that we can provide more meaningful and useful diagnostics.

> Maybe some stylistic guidelines for diagnostics would aid consistency too.

+1 on this too. I try to match clang-style since we inherit a fair number of those diagnostics, but it would be good to have a proper style guide.

tcorringham:
I have now moved most of the checks from the WorkGraphVisitor to DiagnoseNodeEntry().
I have modified the code so that DiagnoseNodeEntry() is also called for Compute shaders - it bails out quite quickly if it is not also a Node.
The one diagnostic remaining in the WorkGraphVisitor is the NodeTrackRWInputSharing check - I'll look at relocating that next.

tools/clang/lib/Sema/SemaHLSL.cpp
tcorringham: The WorkGraphVisitor has now been removed, and the diagnostics raised there have all been relocated.

tcorringham:
I believe all the diagnostics affected by this PR are now raised in appropriate place, building on the changes made by#5484.
My earlier comments regarding NodeTrackRWInputSharing were wrong - in my head I was conflating this with another issue which is no longer relevant with the current spec.

llvm-beanz (commented):
This mostly looks good. It is a huge improvement over what we had. Thank you!

tools/clang/lib/Sema/SemaDXR.cpp
```diff
@@ -1254,12 +1254,151 @@ void DiagnoseClosestHitEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+    nodeLaunchLoc = SourceLocation();
+  }
+
+  // If this is both a compute shader and work-graph node, it may only have
+  // broadcasting launch mode
+  if (computeLoc.isValid() && !nodeLaunchType.equals_lower("broadcasting")) {
+    S.Diags.Report(nodeLaunchLoc, diag::err_hlsl_compute_compatibility)
+      << FD->getName() << nodeLaunchType.lower() + " launch type";
+    S.Diags.Report(computeLoc, diag::note_defined_here) << "compute";
+    // ignore other compute incompatibilities (i.e. input/output records)
```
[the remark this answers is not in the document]
tcorringham: It is a matter of opinion - personally I dislike cascades of secondary error messages as they just confuse matters. In this case if the launch type is incompatible with compute then the shader is invalid - so whether it has input/outputs is academic. I also think that if there are input/outputs it is more likely that the programmer accidentally left compute there rather than accidentally adding the input/outputs.
However, I'm happy to change this if you think it would be more consistent with Clang behavior.

tools/clang/lib/Sema/SemaDXR.cpp
```diff
@@ -1254,12 +1254,151 @@ void DiagnoseClosestHitEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+                     diag::err_hlsl_incompatible_node_attr)
+        << FD->getName() << nodeMDG->getSpelling() << nodeDG->getSpelling()
+        << nodeMDG->getRange();
+      S.Diags.Report(nodeDG->getLocation(), diag::note_defined_here)
+        << nodeDG->getSpelling();
+    }
+  }
+
   if (!FD->getReturnType()->isVoidType())
     S.Diag(FD->getLocation(), diag::err_raytracing_must_return_void);
```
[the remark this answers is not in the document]
tcorringham: That was there before my changes - I agree it looks wrong so I'll revisit that.

tools/clang/lib/Sema/SemaDXR.cpp
```diff
@@ -1254,12 +1254,151 @@ void DiagnoseClosestHitEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+    nodeLaunchType = NodeLaunch->getLaunchType();
+    nodeLaunchLoc = NodeLaunch->getLocation();
+  } else {
+    nodeLaunchType = "broadcasting";
+    nodeLaunchLoc = SourceLocation();
+  }
+
+  // If this is both a compute shader and work-graph node, it may only have
+  // broadcasting launch mode
+  if (computeLoc.isValid() && !nodeLaunchType.equals_lower("broadcasting")) {
```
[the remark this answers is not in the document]
tcorringham: I've been thinking that for a while, but not done anything about it yet - I think now's the time...

tools/clang/lib/Sema/SemaDXR.cpp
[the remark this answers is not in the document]
llvm-beanz: The advantage of being more compatible with clang behavior is that the upstream compiler will model clang behavior, so being consistent here will give the same user experience in both places.

Additionally, the Clang behavior is the way it is so that IDE tools can present a comprehensive list of diagnostics rather than stopping at the first error. That is useful in situations where resolving the first error isn't actually enough to get the code to be correct. Getting users stuck in a compile->fail->compile->fail cycle is not a great workflow.

tools/clang/lib/Sema/SemaDXR.cpp
```diff
@@ -1254,12 +1254,151 @@ void DiagnoseClosestHitEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+  // Check parameter constraints
   for (unsigned Idx = 0; Idx < FD->getNumParams(); ++Idx) {
     ParmVarDecl *Param = FD->getParamDecl(Idx);
 
+    // compute is incompatible with node input/output
+    if (computeLoc.isValid() && hlsl::IsHLSLNodeType(Param->getType())) {
+      S.Diags.Report(Param->getLocation(), diag::err_hlsl_compute_compatibility)
+        << FD->getName() << "node input/output" << Param->getSourceRange();
+      S.Diags.Report(computeLoc, diag::note_defined_here) << "compute";
+      // ignore any other errors
```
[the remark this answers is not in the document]
tcorringham: I have removed the short-circuit here and earlier at line 1305.

tools/clang/lib/Sema/SemaDXR.cpp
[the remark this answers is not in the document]
tcorringham: I have amended the code to allow further errors here and later at line 1380.
There is also a bail-out in the case of errors in hlsl::DiagnoseTranslationUnit() in SemaDiag.cpp - should that be removed too?

tools/clang/test/HLSL/workgraph/attribute_diags.hlsl
```diff
@@ -0,0 +1,130 @@
...
+[NodeMaxDispatchGrid(256, 8, 8)]     // expected-error {{'nodemaxdispatchgrid' may only be used with broadcasting nodes}}
+void node04()
+{ }
+
+// NodeDispatchGrid and NodeMaxDispatchGrid may not both be used
+
+[Shader("node")]
+[NodeLaunch("Broadcasting")]
+[NodeMaxDispatchGrid(256, 64, 32)]   // expected-error {{Node shader 'node05' may not use both 'nodemaxdispatchgrid' and 'nodedispatchgrid'}}
+[NodeDispatchGrid(32, 1, 1)]         // expected-note  {{nodedispatchgrid defined here}}
```
pow2clk, line 36: Seems a bit odd that the second attribute is the one that gets the note. Any idea why that is? Maybe we need a deterministic container somewhere?

tools/clang/lib/Sema/SemaDXR.cpp
```diff
@@ -1254,12 +1258,165 @@ void DiagnoseClosestHitEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+    if (auto NumThreads = FD->getAttr<HLSLNumThreadsAttr>()) {
+      if (NumThreads->getX() != 1 || NumThreads->getY() != 1 ||
+          NumThreads->getZ() != 1) {
+        S.Diags.Report(NumThreads->getLocation(),
+                       diag::err_hlsl_wg_thread_launch_group_size)
+          << NumThreads->getRange();
+        // Only output the note if the source location is valid
+        if (nodeLaunchLoc.isValid())
+          S.Diags.Report(nodeLaunchLoc, diag::note_defined_here)
+              << "Launch type";
```
pow2clk, line 1333: I'm not sure a note indicating where the launch type is defined is necessary. There are several other errors here that reference the launch type, but don't include this note. I think it's clear enough where to find it without the note.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/ast-rwnodeinput.hlsl
```diff
@@ -46,7 +47,7 @@ void node05(RWDispatchNodeInputRecord<RECORD> input) {}
 [NumThreads(1024,1,1)]
 void node06(RWGroupNodeInputRecords<RECORD> input) {}
 
-//CHECK: ClssTemplateDecl {{0x[0-9a-fA-F]+}} <<invalid sloc>> <invalid sloc> implicit RWGroupNodeInputRecords
+//CHECK: ClassTemplateDecl {{0x[0-9a-fA-F]+}} <<invalid sloc>> <invalid sloc> implicit RWGroupNodeInputRecords
```
pow2clk, line 50: I wondered how this passed before and I think I know. There should be a bar between `%s` and `FileCheck` in the run line above.

pow2clk (commented):
The suggestions about how to format and present error text can be ignored if you disagree. I'm concerned about ast-rwnodeinput.hlsl that I still don't think is filechecked properly.

tools/clang/test/HLSL/workgraph/attribute_diags.hlsl
tcorringham: It appears that the attributes are processed in reverse order. I don't know whether there is a Clang reason to do it that way, or what the implications of changing it would be - I'll investigate.

pow2clk: Except that's not true of the NodeLaunch attrib above 🤔

tools/clang/lib/Sema/SemaDXR.cpp
pow2clk: Then I'm afraid there are a fair number of other errors that might similarly benefit

tools/clang/test/HLSL/workgraph/attribute_diags.hlsl
pow2clk: Scratch that, the above is a different situation. This may well be alphabetical. Anyway, I'm not too concerned about it just now

tools/clang/test/HLSLFileCheck/hlsl/workgraph/ast-rwnodeinput.hlsl
tcorringham: I hadn't spotted that! - and as a result the test didn't actually test anything. I have updated this test and related ones to make them more consistent.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7810,8 +7814,22 @@ def err_hlsl_dispatchgrid_component : Error<
    "%0 %1 component value must be between 1 and 65,535 (2^16-1) inclusive">;
 def err_hlsl_dispatchgrid_product : Error<
    "%0 X * Y * Z product may not exceed 16,777,215 (2^24-1)">;
-def err_hlsl_compute_compatibility : Error<
-   "Node shader '%0' with %1 is not compatible with compute">;
+def err_hlsl_compute_launch_compatibility : Error<
+   "Node shader '%0' with %1 launch type is not compatible with compute">;
```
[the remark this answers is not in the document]
tcorringham: Done

tools/clang/lib/Sema/SemaDXR.cpp
```diff
@@ -1254,12 +1258,165 @@ void DiagnoseClosestHitEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+      S.Diags.Report(computeLoc, diag::note_defined_here) << "compute";
+    }
+
+    // Check any node input is compatible with the node launch type
+    if (hlsl::IsHLSLNodeInputType(Param->getType())) {
+      inputCount++;
+      const RecordType* RT = Param->getType()->getAs<RecordType>();
+      StringRef typeName = RT->getDecl()->getName();
+      if (!NodeInputIsCompatible(typeName, nodeLaunchType)) {
+        S.Diags.Report(Param->getLocation(), diag::err_hlsl_wg_input_kind)
```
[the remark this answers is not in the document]
tcorringham: I considered whether we should specify the launch type that is valid with the input type, or the input types that are valid with the launch type. I decided on the latter so I've added a select based on the NodeLaunchType (-1) to indicate which input record types are valid for the specified launch node type.

pow2clk (approved):
Thanks for the responses!

tools/clang/lib/Sema/SemaExpr.cpp
```diff
@@ -5026,7 +5026,13 @@ Sema::ActOnCallExpr(Scope *S, Expr *Fn, SourceLocation LParenLoc,
     }
 
     if (Fn->getType() == Context.BoundMemberTy) {
-      return BuildCallToMemberFunction(S, Fn, LParenLoc, ArgExprs, RParenLoc);
+      // HLSL Change Begin - Diagnose HLSL member calls
+      auto memberCall = BuildCallToMemberFunction(S, Fn, LParenLoc, ArgExprs, RParenLoc);
+      if (getLangOpts().HLSL && !memberCall.isInvalid())
+        if (CXXMemberCallExpr *call = dyn_cast<CXXMemberCallExpr>(memberCall.get()))
+          DiagnoseHLSLMemberCallExpr(call);
```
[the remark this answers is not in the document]
llvm-beanz: It feels awkward (and not the normal clang pattern) to diagnose errors after generating the AST node. Normally the diagnosis occurs and on failure an ExprError (or nullptr) is returned without ever building the AST node.

Whether or not we continue to further errors is usually a decision higher up whether or not the parser can continue based on available context when the error is handled.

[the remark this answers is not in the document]
tcorringham: I've moved this to within BuildCallToMemberFunction() where some other checks are performed, and it returns ExprError() on failure.

llvm-beanz (approved):
LGTM. Thank you for the great cleanup!

PR #5468, opened 2023-07-27
Add TransformUtils dependency to DxilContainer for Linux builds

hekota:
The dependency is necessary because DxilContainerAssembler.cpp uses llvm::CloneModule.


llvm-beanz (approved):
Looks good. We should add a PR description explaining why this was needed.

python3kgae (approved):

PR #5472, opened 2023-07-28
[matrix test] Move matrix test as lit FileCheck test.

python3kgae:
This is to help check our existing tests on matrix type for matrix orientation change.


tools/clang/test/CodeGenDXIL/hlsl/types/matrix/default-matrix-in-template.hlsl
```diff
@@ -9,7 +9,6 @@ StructuredBuffer<matrix> buf1;
 
 RWBuffer<float4> buf2;
 
-[RootSignature("DescriptorTable(SRV(t0), UAV(u0))")]
```
pow2clk, line 12: Just to document this. This is a workaround for #5476. Since the root signature isn't really part of this test, we decided it was okay to remove this line to keep the test running

tools/clang/test/CodeGenDXIL/hlsl/types/matrix/selMat.hlsl
```diff
@@ -0,0 +1,25 @@
+// RUN: %dxc -E main -T ps_6_0 -HV 2018 %s | FileCheck %s
+// RUN: %dxc -E main -T ps_6_0 -HV 2018 -Zpr %s | FileCheck %s
+
+// CHECK-COUNT-16: fcmp fast olt float %{{.+}}, 3.000000e+00
+
+// CHECK-COUNT-16: select i1
+
+// CHECK: icmp ne i32
+// CHECK: 0
```
pow2clk, line 9: This line is weird. What is it checking for? Similar lines in matElt were changed to be clearer.

pow2clk (approved):
The changed tests look good! I'd like to understand the `CHECK 0` line.

As a tip that I find helps reviewers when moving and altering files, if you do it as two separate commits, one that moves the files, another that alters them, then the altering commit can be viewed to see what actually changed more easily than comparing the two files.

llvm-beanz (changes requested):
We should not move anymore tests out of the HLSLFileCheck directory until we resolve #5552.

PR #5473, opened 2023-07-28
This is really a reference

### LINKED ISSUE #4709: Accessing templated base class member var results in compiler crash
You can also check it out in shader playground too
https://shader-playground.timjones.io/8518ac9fa48d1f21c3c92a62dccde208

Using newest version dxc with -HV 2021

```c++
template<typename T, uint32_t N>
class array
{
    T mArr[N];
};

template<typename T, uint32_t N>
class array_ext : array<T,N>
{
    float test() {
       return array<T,N>::mArr[0]; 
    }
};

[numthreads(1, 1, 1)]
void main( uint3 DTid : SV_DispatchThreadID )
{
    array_ext<float, 3> arr1;
    arr1.test();
}
```

llvm-beanz:
This fixes a few cases that were missed when I updated `this` objects to be references. Specifically we were still generating some implicit `this` cases as pointers, and we were not always correctly looking through the reference types.

Fixes #4709

### DIFF
--- a/tools/clang/include/clang/AST/DeclCXX.h
+++ b/tools/clang/include/clang/AST/DeclCXX.h
@@ -1830,6 +1830,13 @@ public:
   /// Should only be called for instance (i.e., non-static) methods.
   QualType getThisType(ASTContext &C) const;
 
+  // HLSL Change Begin - This is a reference.
+  /// \brief Returns the type of the \c this object looking through the pointer.
+  ///
+  /// Should only be called for instance (i.e., non-static) methods.
+  QualType getThisObjectType(ASTContext &C) const;
+  // HLSL Change End - This is a reference.
+
   unsigned getTypeQualifiers() const {
     return getType()->getAs<FunctionProtoType>()->getTypeQuals();
   }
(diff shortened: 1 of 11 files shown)


pow2clk (approved):
Looks good though I'd still like the comments in the tests. 

I'd like to have a discussion about the nonsensical LValueToRvalue  casts. I've heard you mention it before and I'd like to understand why they exist and what to do about them.

llvm-beanz:
> I'd like to have a discussion about the nonsensical LValueToRvalue casts. I've heard you mention it before and I'd like to understand why they exist and what to do about them.

Here's a simple example of the kinds of nonsense casts we get in the AST. Take this code:
```hlsl
struct Doggo {
  int4 Legs;
};

void Walk(inout Doggo D);

void fn() {
  Doggo D = {1.xxxx};
  Walk(D);
}
```
[Godbolt Link](https://godbolt.org/z/TKx6WK7q7).

When `D` is passed to `Walk`, it is an l-value. It gets modified by `Walk` so it must be an l-value. But the AST is this:

```ast
CallExpr <line:10:3, col:9> 'void'
|-ImplicitCastExpr <col:3> 'void (*)(__restrict Doggo)' <FunctionToPointerDecay>
| `-DeclRefExpr <col:3> 'void (__restrict Doggo)' lvalue Function 'Walk' 'void (__restrict Doggo)'
`-ImplicitCastExpr <col:8> 'Doggo' <LValueToRValue>
  `-DeclRefExpr <col:8> 'Doggo' lvalue Var 'D' 'Doggo'
```

According to the AST, `D` gets l-value to r-value casted and the r-value of `D` gets passed into the AST. In this case it is because the AST doesn't represent UDTs being passed `inout`.

There are other cases where we do this wrong too. Fixing the ASTs to be accurate is the intent of #5249.

bogner (approved):

PR #5483, opened 2023-08-01
[C++17] Merge upsteam change for Don't use std::iterator in PATH.h

python3kgae:
Merge upstream change
https://github.com/llvm/llvm-project/commit/91c8daf3009209a63534b4312d74322e1399bba1

In converting this over to iterator_facade_base, some member
operators and methods are no longer needed since iterator_facade
implements them in the base class using CRTP.

This is part of enable C++17.


pow2clk (approved):
The code looks fine. I would like the description from the original change to be included.

bob80905 (approved):

Issue #5478, opened 2023-07-31
Create staging branch for merging SM 6.8 features to main


pow2clk:
**Problem description.**
We need a branch that we can deliver main deliveries to while we work to polish the shader model 6.8 features that are currently in the preview release branch 1.8.2306

**Solution**
Create a branch called something like "sm68-staging" where we can work on the features in such a way that they can be delivered to main

**alternatives**
We could use the existing release branch for this purpose, but we would either be working on an ever aging code base or have to make regular deliveries to a release branch, which isn't how release branches usually work.

pow2clk:
Created! https://github.com/microsoft/DirectXShaderCompiler/tree/staging-sm-6.8 Rules are set up as well

PR #5484, opened 2023-08-01
[Verifier] Add verifier tests to catch MaxRecord errors earlier

bob80905:
This PR is one of several more PR's that move diagnostic emission earlier for specific errors, so that the verifier can catch these errors. This PR focuses on the MaxRecord* attributes applied on parameters. Specifically, there are 2 behaviors that this PR will allow DXC to detect earlier:

1. Prevent more than one of {MaxRecord, MaxRecordsSharedWith} from being applied to a single parameter
2. Prevent MaxRecordsSharedWith from referencing the same parameter name that the attribute is being applied to.

Some tests have been merged or deleted to consolidate these verifier tests. There is still some behavior that cannot be tested at this time (verifying that MaxRecordsSharedWith doesn't reference an invalid parameter name), which will be addressed in the future. It's likely that this location is too early to properly validate this behavior, and the tests for this behavior may need to be left at the same location.

This PR is one of a few that will help implement this task:
https://github.com/microsoft/DirectXShaderCompiler/issues/5368


tools/clang/test/HLSL/max_output_records_duplicate.hlsl
```diff
@@ -0,0 +1,73 @@
...
+  [MaxRecords(5)] [MaxRecordsSharedWith(Output6)] NodeOutput<rec1> Output5) /* expected-error {{only one of MaxRecords or MaxRecordsSharedWith may be specified to the same parameter.}} */ /* expected-note {{conflicting attribute is here}} */ /* expected-note {{conflicting attribute is here}} */  
+{
+}
+
+
+void DuplicateMaxSameAttr(
+[MaxRecords(6)] [MaxRecords(3)] NodeOutput<rec1> Output1, /* expected-error {{only one of MaxRecords or MaxRecordsSharedWith may be specified to the same parameter.}} */ /* expected-note {{conflicting attribute is here}} */ /* expected-note {{conflicting attribute is here}} */
+  [MaxRecordsSharedWith(Output1)] [MaxRecordsSharedWith(Output1)] NodeOutput<rec1> Output2) /* expected-error {{only one of MaxRecords or MaxRecordsSharedWith may be specified to the same parameter.}} */ /* expected-note {{conflicting attribute is here}} */ /* expected-note {{conflicting attribute is here}} */
```
llvm-beanz, line 45: This seems like it shouldn’t produce a diagnostic. If the exact same attribute (and value) is specified twice we should accept that.

tools/clang/test/HLSL/max_output_records_duplicate.hlsl
```diff
@@ -0,0 +1,73 @@
...
+  [MaxRecordsSharedWith(Output6)] [MaxRecords(4)] NodeOutput<rec1> Output4, /* expected-error {{only one of MaxRecords or MaxRecordsSharedWith may be specified to the same parameter.}} */ /* expected-note {{conflicting attribute is here}} */ /* expected-note {{conflicting attribute is here}} */
+  [MaxRecords(5)] [MaxRecordsSharedWith(Output6)] NodeOutput<rec1> Output5) /* expected-error {{only one of MaxRecords or MaxRecordsSharedWith may be specified to the same parameter.}} */ /* expected-note {{conflicting attribute is here}} */ /* expected-note {{conflicting attribute is here}} */  
+{
+}
+
+
+void DuplicateMaxSameAttr(
+[MaxRecords(6)] [MaxRecords(3)] NodeOutput<rec1> Output1, /* expected-error {{only one of MaxRecords or MaxRecordsSharedWith may be specified to the same parameter.}} */ /* expected-note {{conflicting attribute is here}} */ /* expected-note {{conflicting attribute is here}} */
```
llvm-beanz, line 44: What if `MaxRecords` is specified twice with the same value? That should be okay right?

tools/clang/lib/Sema/SemaDXR.cpp
```diff
@@ -1138,171 +1138,198 @@ void DiagnoseRaytracingPayloadAccess(clang::Sema &S,
...
-        S.Diag(Param->getLocation(), diag::err_payload_requires_inout)
-            << /*payload|callable*/ 1 << Param;
-      QualType Ty = Param->getType().getNonReferenceType();
-
-      if (!(hlsl::IsHLSLCopyableAnnotatableRecord(Ty)))
-        S.Diag(Param->getLocation(), diag::err_payload_attrs_must_be_udt)
-            << /*payload|attributes|callable*/ 2 << Param;
-    }
+void DiagnoseMissOrAnyHitEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr,
+                               bool isMiss) {
```
llvm-beanz, line 1165: Rather than a bool here, why don't we just pass in the enum value? I feel like that will make the code easier to understand.

llvm-beanz (approved):
One small change inline

tools/clang/unittests/HLSL/VerifierTest.cpp
```diff
@@ -199,6 +200,13 @@ TEST_F(VerifierTest, RunShaderMismatch) {
   CheckVerifiesHLSL(L"shader_attribute_no_mismatch.hlsl");
 }
 
+TEST_F(VerifierTest, RunMaxRecordsAttribute) {
+  CheckVerifiesHLSL(L"max_output_records_duplicate.hlsl");
```
python3kgae, line 204: Is it possible to make these tests lit FileCheck test?

llvm-beanz: No, this is targeting the release branch.

python3kgae: Why release branch cannot have lit FileCheck test?

python3kgae: Because we won't hide taef for release branch?

python3kgae: Do we plan to merge the LIT change to preview branch?

python3kgae (approved):

tools/clang/unittests/HLSL/VerifierTest.cpp
llvm-beanz: I think that's a question for @pow2clk (#5478)

Issue #5489, opened 2023-08-02
ARM64 clang tests are intermittently failing when ARM64X is enabled


hekota:
**Description**
Internal ARM64 clang tests are intermittently failing when ARM64X is enabled. 

**Steps to Reproduce**
In internal ARM64 testing pipeline some of the clang tests are randomly failing in the nightly build, such as:
    clang-taef :: ClangHLSLTests.dll/CompilerTest::BatchSamples
    clang-taef :: ClangHLSLTests.dll/RewriterTest::RunEffect
    clang-taef :: ClangHLSLTests.dll/RewriterTest::RunIncludes

The list of tests is not complete. 

The Rewriter tests are failing on file I/O. My guess is the tests are running both ARM64 and ARM64EC versions in parallel and it is clashing with the file I/O. Tests for ARM64 and ARM64EC must be run separately.

**Actual Behavior**
The tests should be passing.

**Environment**
Internal ARM64 testing pipeline invoked by nightly build.

PR #5494, opened 2023-08-03
Lit TAEF fixes: Use filter when listing tests and fix verbose parameter

hekota:
Lit TAEF files fixes:
- Use filter parameter when enumerating tests. This enables listing of just the tests that should run, and it also filtering by architecture. This is crucial for isolating ARM64 and ARM64EC versions of tests that are in the same test dll.
- Fix verbose parameter - it was not getting applied

Related to #5489.

### DIFF
--- a/utils/lit/lit/formats/taef.py
+++ b/utils/lit/lit/formats/taef.py
@@ -64,10 +64,12 @@ class TaefTest(TestFormat):
 
         if litConfig.debug:
             litConfig.note('searching taef test in %r' % dll_path)
+        
+        select_arg = str.format('/select:"{}"', self.select_filter)
 
+        cmd = [self.te, dll_path, "/list", select_arg]
         try:
-            lines = lit.util.capture([self.te, dll_path, '/list'],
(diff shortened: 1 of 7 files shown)


utils/lit/lit/formats/taef.py
```diff
@@ -64,10 +64,12 @@ def getTaefTests(self, dll_path, litConfig, localConfig):
...
         if litConfig.debug:
             litConfig.note('searching taef test in %r' % dll_path)
+        
+        select_arg = str.format('/select:"{}"', self.select_filter)
 
+        cmd = [self.te, dll_path, "/list", select_arg]
         try:
-            lines = lit.util.capture([self.te, dll_path, '/list'],
-                                     env=localConfig.environment)
+            lines,_,_ = executeCommandForTaef(cmd)
```
hekota, line 72: Using executeCommandForTaef here because lit.util.capture has issues with quotes in the select parameter.

python3kgae (approved):

bob80905 (approved):

Issue #5366, opened 2023-06-28
Some D3DReflect RDAT CHECK lines need the sizes to be regex'd


llvm-beanz:
Some of the new RDAT tests have explicit sizes in the CHECK lines, these should be replaced with regex matches to make the tests less fragile.

bob80905:
Fixed by #5495

PR #5495, opened 2023-08-04
[Test] Replace size integer literal with regex

bob80905:
Some D3DReflect tests compare size values literally, when the size will change depending on the validator version.
This PR changes the literal integer in the checkline to a regex, so that the size value changing between validators won't break the test. 
Fixes  #5386


pow2clk (approved):
Looks good. The way you've referenced the issue in the comment will not close the bug. You need something on the same line such as `Fixes #5366`

llvm-beanz (approved):

Issue #5480, opened 2023-07-31
[Build] Compile with clang-cl on windows


python3kgae:
To be able to use clang Source Based Code Coverage tool https://clang.llvm.org/docs/SourceBasedCodeCoverage.html on windows, need to compile with clang-cl first.

Now hit a lot of error/warning when compile with clang-cl.

Goal:
Build and pass all test with clang-cl on windows.

pow2clk:
Fixed by #6084

PR #5496, opened 2023-08-04
[ClangCL] exception related fix for ClangCL

python3kgae:
Enable Exception for ClangCL.
Disable the DivByZero exception test for ClangCL.

This is for turn on ClangCL build.

### DIFF
--- a/cmake/modules/AddLLVM.cmake
+++ b/cmake/modules/AddLLVM.cmake
@@ -9,7 +9,7 @@ function(llvm_update_compile_flags name)
 
   # LLVM_REQUIRES_EH is an internal flag that individual
   # targets can use to force EH
-  if((LLVM_REQUIRES_EH OR LLVM_ENABLE_EH) AND NOT CLANG_CL)
+  if(LLVM_REQUIRES_EH OR LLVM_ENABLE_EH) # HLSL Change - allow CLANG_CL to use EH.
     if(NOT (LLVM_REQUIRES_RTTI OR LLVM_ENABLE_RTTI))
       message(AUTHOR_WARNING "Exception handling requires RTTI. Enabling RTTI for ${name}")
       set(LLVM_REQUIRES_RTTI ON)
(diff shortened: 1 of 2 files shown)


hekota (approved):

python3kgae:
Link to the issue https://github.com/microsoft/DirectXShaderCompiler/issues/5480

cmake/modules/AddLLVM.cmake
```diff
@@ -9,7 +9,7 @@ function(llvm_update_compile_flags name)
 
   # LLVM_REQUIRES_EH is an internal flag that individual
   # targets can use to force EH
-  if((LLVM_REQUIRES_EH OR LLVM_ENABLE_EH) AND NOT CLANG_CL)
+  if(LLVM_REQUIRES_EH OR LLVM_ENABLE_EH) # HLSL Change - allow CLANG_CL to use EH.
```
pow2clk, line 12: I worry that this previously refused to enable this flag with clang-cl. Do you know anything about the history?

pow2clk: Hmm. I found this, but it's not terribly illustrative: https://github.com/llvm/llvm-project/commit/8c18019d50127a79c92d6f6017bad0c0c83441d4

pow2clk: It was shortly after llvm 3.7 though. So maybe it's just a relic. I'd still like to know the history before enabling it and hoping for the best

python3kgae: It is disabled here https://github.com/llvm/llvm-project/commit/66bc9080d6d5ef4904f25d70c374a3dc2c218c9f

pow2clk: https://github.com/llvm/llvm-project/commit/66bc9080d6d5ef4904f25d70c374a3dc2c218c9f shows where it was introduced seemingly because clang-cl just didn't support it completely at the time: https://github.com/llvm/llvm-project/commit/c542d379958d1e44744aa9474f79c6732264121d There were a couple fixes that went in between those changes to add some support: https://github.com/llvm/llvm-project/commit/387fccd8dad1ab16f4d7332e8462b88bcbb2b26e that came after 3.7. 

In short, I spent too much time looking into this! It seems there was some support then, but not everything. So maybe it's enough? How is it used?

pow2clk (approved):
Looks good!

PR #5497, opened 2023-08-04
[lit][SPIRV] Convert SPIRV matrix tests to lit

python3kgae:
Convert matrix tests to lit FileCheck test.

This is only the first batch of matrix tests, more to be converted.


tools/clang/test/CodeGenSPIRV/binary-op.mul-assign.type-mismatch.hlsl
```diff
@@ -1,53 +0,0 @@
-// RUN: %dxc -T ps_6_2 -E main -fspv-target-env=vulkan1.1 -enable-16bit-types
```
cassiebeckley, line 1: Can you use `git mv` for these so that the history is preserved correctly?

tools/clang/unittests/SPIRV/CodeGenSpirvTest.cpp
```diff
@@ -2385,23 +2366,9 @@ TEST_F(FileTest, VulkanLayoutTextureBufferStd430) {
   setGlLayout();
   runFileTest("vk.layout.texture-buffer.std430.hlsl");
 }
-TEST_F(FileTest, VulkanLayout64BitTypesStd430) {
-  setGlLayout();
```
cassiebeckley, line 2389: When are these tests run now that they're LIT tests? Will we be able to preserve things like this call to `setGlLayout`?

cassiebeckley (changes requested):

tools/clang/unittests/SPIRV/CodeGenSpirvTest.cpp
python3kgae: These LIT test will run when ninja check-all or cmake --build %HLSL_BLD_DIR% --target check-all.
The setGlLayout should be translated into -fvk-use-gl-layout in the RUN line.

tools/clang/test/CodeGenSPIRV/binary-op.mul-assign.type-mismatch.hlsl
python3kgae: Tried modify the file in one commit first, then git mv in another commit.

Will this preserve history after squash?

cassiebeckley: Hopefully!

cassiebeckley (approved):

Keenuts (approved):

PR #5499, opened 2023-08-04
PIX: Updates to shader debugging pass to support DXR hit groups/miss shaders

jeffnn:
Two main themes:
-Add instrumentation to the DXR hit group and miss shader types
-Return a bit more info to the caller (PIX) via output text


lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -1042,8 +1035,18 @@ bool DxilDebugInstrumentation::RunOnFunction(
 
   auto &values = m_FunctionToValues[BC.Builder.GetInsertBlock()->getParent()];
 
+  unsigned int UAVRegisterId = 0;
+    switch (shaderKind) {
+  case DXIL::ShaderKind::ClosestHit:
+  case DXIL::ShaderKind::Intersection:
+  case DXIL::ShaderKind::AnyHit:
+  case DXIL::ShaderKind::Miss:
+    UAVRegisterId = 1;
```
jeffnn, line 1044: I'll add a comment about these distasteful hard-coded numbers.

tools/clang/test/HLSLFileCheck/pix/DebugHitGroupAndMiss.hlsl
```diff
@@ -0,0 +1,110 @@
+// RUN: %dxc -T lib_6_3 %s | %opt -S -hlsl-dxil-debug-instrumentation,parameter0=10,parameter1=20,parameter2=30 | %FileCheck %s
```
python3kgae, line 1: Could we move this test to somewhere like tools/clang/test/PIX  to make it a lit FileCheck test?

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -107,6 +107,29 @@ void DxilAnnotateWithVirtualRegister::applyOptions(llvm::PassOptions O) {
 
 char DxilAnnotateWithVirtualRegister::ID = 0;
 
+static const char* ShaderKindName(hlsl::DXIL::ShaderKind kind) {
```
python3kgae, line 110: Other passes may need this, could we move it to DxilShaderModel.cpp?

llvm-beanz: +1 to @python3kgae's suggestion. We have (at least) one other implementation of this logic in the `D3DReflectionDumper`. I think this implementation is cleaner (although clang-format would be nice).

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -120,12 +143,15 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
   }
 
   std::uint32_t InstNum = m_StartInstruction;
-  std::map<llvm::StringRef, std::pair<int, int>> InstructionRangeByFunctionName;
+  std::map<std::string, std::pair<int, int>> InstructionRangeByFunctionName;
```
llvm-beanz, line 146: What is the motivation for this? Do you need the map to own the string? If so, `llvm::StringMap` might be a better approach, it is certainly more optimal unless you need ordering.

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -147,12 +175,10 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
         skipOverLeadingUnprintableCharacters = 2;
       }
       *OSOverride << fn.second.first << " " << fn.second.second << " "
-                  << (fn.first.str().c_str() +
+                  << (fn.first.c_str() +
```
llvm-beanz, line 178: Rather than `c_str()` here, if this were still a StringRef, you could use `StringRef::slice`. Keeping the bounds with the string avoids needing to walk to the null character.

tools/clang/test/HLSLFileCheck/pix/DebugHitGroupAndMiss.hlsl
jeffnn: Does it make sense for me to move all the PIX tests at once?

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
jeffnn: I have now discovered that there already is such a function! Deleted mine, and used the existing one.

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
jeffnn: Switched to StringMap

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
jeffnn: Switched to a more string-ref-y approach. Much cleaner. Thanks!

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -120,12 +130,15 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
   }
 
   std::uint32_t InstNum = m_StartInstruction;
-  std::map<llvm::StringRef, std::pair<int, int>> InstructionRangeByFunctionName;
+  llvm::StringMap<std::pair<int, int>> InstructionRangeByFunctionName;
```
adam-yang, line 133: I don't think llvm::StringMap is guaranteed to be determinstic in order when iterating.

llvm-beanz: Yea, but the problem that Adam is pointing out here is real. PIX may not care about the order, but we don't want the compiler generating output non-deterministically, and since you're iterating over the map and printing output.

StringMap technically is deterministic, but it is unordered, so the ordering may not make sense when observed. Additionally, the ordering is not guaranteed to be stable across LLVM versions so relying on it will impact our move to modern LLVM.

One possible solution I see would be to merge the loops by adding the `OSOverride` printing happen inside the loop over the `instrumentableFunctions`.

jeffnn: Fixed by removing map, per Chris' suggestion.

adam-yang (approved):

PR #5508, opened 2023-08-07
Allow select() to work with samplers

### LINKED ISSUE #4673: Cannot use `select` intrinsic with SamplerState

pow2clk:
Add any_sampler builtin parameter type. This will match any sampler type as the select() built-in added with HLSL 2021 requires.

Adds a new built-in overload that takes the any_sampler type and ensures that the last two arguments and return type match. Many of the changes were required because we never had to ensure object parameter types matched nor transmit that type to the return type before. Most of the changes to MatchArguments in SemaHLSL are addressing new issues this revealed. Non-basic types can pass the conversion check if they outright match.

Adds testing to select() and the ternary operator that it supplements for these usages

Fixes #4673


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -2103,16 +2112,16 @@ static bool CombineBasicTypes(ArBasicKind LeftKind,
...
-    (RightKind < 0 || RightKind >= AR_BASIC_COUNT)) {
-    return false;
-  }
-
   if (LeftKind == RightKind) {
     *pOutKind = LeftKind;
     return true;
   }
 
+  if ((LeftKind < 0 || LeftKind >= AR_BASIC_COUNT) ||
```
llvm-beanz, line 2120: Moving this block down seems a little odd. Should we allow void or out of range values if they match?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6449,6 +6456,11 @@ bool HLSLExternalSource::MatchArguments(
              "have as many arguments and types as the intrinsic template");
   }
 
+  // For object return types that need to match arguments, we need to slot in the full type here
+  // Can't do it sooner because when return is encountered above, the other arg types haven't been set
```
llvm-beanz, line 6460: nit: can you clang format your changes?

utils/hct/gen_intrin_main.txt
```diff
@@ -372,6 +372,7 @@ resource [[hidden]] CreateResourceFromHeap(in uint index);
 $match<1, 0> bool<> [[rn]] and(in any<> x, in $type1 y);
 $match<1, 0> bool<> [[rn]] or(in any<> x, in $type1 y);
 $type2 [[rn]] select(in bool<> cond, in $match<1, 2> any<> t, in $type2 f);
+$type2 [[rn]] select(in bool cond, in any_sampler t, in $type2 f);
```
llvm-beanz, line 375: Am I correct in assuming this is required because `any` isn't really _anything_ but rather any scalar or vector?
Do we need any other overloads for this? Have we tested if it works with textures?

include/dxc/dxcapi.internal.h
```diff
@@ -99,7 +99,8 @@ enum LEGAL_INTRINSIC_COMPTYPES {
   LICOMPTYPE_UINT8_4PACKED = 41,
   LICOMPTYPE_ANY_INT16_OR_32 = 42,
   LICOMPTYPE_SINT16_OR_32_ONLY = 43,
-  LICOMPTYPE_COUNT = 44
+  LICOMPTYPE_ANY_SAMPLER = 44,
+  LICOMPTYPE_COUNT = 45
```
bob80905, line 103: Is there a specific reason why LICOMPTYPE_COUNT is now assigned to 45 instead of leaving it at 44 and letting LICOMPTYPE_ANY_SAMPLER taking the 45 spot?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -2103,16 +2112,16 @@ static bool CombineBasicTypes(ArBasicKind LeftKind,
...
   if (LeftKind == RightKind) {
     *pOutKind = LeftKind;
     return true;
   }
 
+  if ((LeftKind < 0 || LeftKind >= AR_BASIC_COUNT) ||
+    (RightKind < 0 || RightKind >= AR_BASIC_COUNT)) {
+    return false;
+  }
+
```
bob80905, line 2124: I would think validation takes higher priority than the equal type behavior above. Why was this if block moved under?

pow2clk: The reason is because the objects are not "BASIC". So even though they matched, they were failing due to this. I could maintain some of the validation though.

utils/hct/gen_intrin_main.txt
pow2clk: You are correct that `any` isn't really "any". It's just scalars and vectors. I had an earlier change that created a broader type. However, since all that's needed here are samplers since textures didn't work with ternary operators before so they aren't expected to work with `sample()`, I chose to simplify here. We might consider allowing textures for the sake of consistency, but that seems like a separate change rather than something that is needed to make HLSL 2021 consistent with what came before.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Greater than BASIC isn't out of range, it just means that it's not a basic type.  This allows a non-basic type. I could maintain the void and add a true out of range check here though.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: I did it. It changed more than I expected. It's all okay except for the way they shifted the AnySampler and ObjectTT initialization, which I would argue is harder to read and inconsistent with all the others, so I reverted that part.

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: Ah, makes sense.

llvm-beanz (approved):

bob80905 (approved):

PR #5515, opened 2023-08-08
[lit] support dxopt in lit

python3kgae:
When input is from stdin, need to add '\n' after each line.

This is for use dxopt to test pix passes.
opt doesn't support passes which has parameters right now. We could switch it to opt once pass parameter is supported.


jeffnn (approved):
LGTM. I can move the rest of the pix tests later.

hekota (approved):

PR #5516, opened 2023-08-08
[NFC] Refactor -opt-enable and -opt-disable

adam-yang:
Refactored -opt-enable and -opt-disable.

- Added `hlsl::option::OptimizationToggles` to manage the toggles instead of checking a `std::map`.
- All the options are moved into `hlsl::options::TOGGLE_*` constants, where each constant contains both the option's name and whether it's default on or off.
  + Previously, every check had to be either `toggles.count("my-flag") && toggles.find("my-flag")->second` for default off, and `!toggles.count("my-flag") || toggles.map.find("my-flag")->second` for default on.
  + Now, a check is simply `toggles.IsEnabled(TOGGLE_MY_FLAG)`.

### DIFF
--- a/include/dxc/Support/DxcOptToggles.h
+++ b/include/dxc/Support/DxcOptToggles.h
@@ -1,0 +1,72 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DxcOptToggles.h                                                           //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Helper code for representing -opt-disable, -opt-enable, -opt-select       //
+// options                                                                   //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#pragma once
+
+#ifndef LLVM_HLSL_DXC_OPT_TOGGLES_H
+#define LLVM_HLSL_DXC_OPT_TOGGLES_H
+
+#include "llvm/ADT/StringRef.h"
+#include <map>
+#include <set>
+#include <string>
+
+namespace hlsl {
+
+namespace options {
+
+struct Toggle {
+  llvm::StringRef Name;
+  bool Default = false;
+  Toggle(llvm::StringRef Name, bool Default) : Name(Name), Default(Default) {}
+};
+
+enum {
+  DEFAULT_ON  = 1,
+  DEFAULT_OFF = 0,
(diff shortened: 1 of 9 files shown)


tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1362,9 +1361,6 @@ class DxcCompiler : public IDxcCompiler3,
     compiler.getLangOpts().HLSLProfile =
           compiler.getCodeGenOpts().HLSLProfile = Opts.TargetProfile;
 
-    compiler.getCodeGenOpts().HLSLEnablePartialLifetimeMarkers =
```
adam-yang, line 1365: This was here because this expression encodes the fact that "partial-lifetime-markers" is default off, and we did not want to repeat that logic in many places, so store the result in a flag on CodeGenOpts. Now that default-off is stored in `hlsl::options::TOGGLE_PARTIAL_LIFETIME_MARKERS`, we can just check it every time.

include/dxc/Support/DxcOptToggles.h
```diff
@@ -0,0 +1,72 @@
...
+#include <string>
+
+namespace hlsl {
+
+namespace options {
+
+struct Toggle {
+  llvm::StringRef Name;
+  bool Default = false;
+  Toggle(llvm::StringRef Name, bool Default) : Name(Name), Default(Default) {}
```
dmpots, line 30: Using `StringRef` means that the memory is not copied and must be valid for the lifetime of the Toggle value. Since it is always passed a literal sting, this should be ok, but would potentially be a problem if it's using a dynamic string name.

dmpots (approved):

tools/clang/include/clang/Frontend/CodeGenOptions.h
```diff
@@ -24,6 +24,7 @@
 #include "dxc/HLSL/HLSLExtensionsCodegenHelper.h" // HLSL change
 #include "dxc/Support/SPIRVOptions.h" // SPIR-V Change
 #include "dxc/DxcBindingTable/DxcBindingTable.h" // HLSL chanhge
+#include "dxc/Support/DxcOptToggles.h" // HLSL chanhge
```
damyanp, line 27: typo: `chanhge`

damyanp (approved):

tools/clang/include/clang/Frontend/CodeGenOptions.h
adam-yang: Doh

PR #5517, opened 2023-08-09
[lit] add dxil passes to opt

python3kgae:
This will allow use opt in lit test.

### DIFF
--- a/test/tools/opt/pix_pass.test
+++ b/test/tools/opt/pix_pass.test
@@ -1,0 +1,7 @@
+RUN: opt -help | FileCheck %s
+
+This test check pix passes available in opt.
+
+================================= Simple link ================================
+CHECK: OPTIONS:
+CHECK: -hlsl-dxil-PIX-add-tid-to-as-payload
--- a/tools/opt/CMakeLists.txt
+++ b/tools/opt/CMakeLists.txt
@@ -9,6 +9,7 @@ set(LLVM_LINK_COMPONENTS
   HLSL # HLSL Change
   DxilContainer #  HLSL Change for DxcOptimizerPass
   DxilRootSignature #  HLSL Change for DxcOptimizerPass
+  DxilPIXPasses # HLSL Change for PIX passes.
   IPA
   IPO
   IRReader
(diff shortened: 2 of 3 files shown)


jeffnn (dismissed):

llvm-beanz (commented):
I’m sure the `-help` option is broken on all the LLVM-based tools because of how the cl::opt values are hacked off. Maybe we should come up with a more general solution.

We could use `cl::opt` external storage to allow the flag values to be trivial-initialized constants in the source files they are originally declared in (I think CommandLine.cpp), and we could implement a single source file that has a function declaring function-local static cl::opt variables that point at that storage, and we include that source file in the tools that need option parsing.

I think we have a similar issue with the debug options from `Debug.cpp`.

python3kgae:
> I’m sure the `-help` option is broken on all the LLVM-based tools because of how the cl::opt values are hacked off. Maybe we should come up with a more general solution.
> 
> We could use `cl::opt` external storage to allow the flag values to be trivial-initialized constants in the source files they are originally declared in (I think CommandLine.cpp), and we could implement a single source file that has a function declaring function-local static cl::opt variables that point at that storage, and we include that source file in the tools that need option parsing.
> 
> I think we have a similar issue with the debug options from `Debug.cpp`.

If we have to include a source file, why not just put static cl::opt in that source file?

llvm-beanz:
> > I’m sure the `-help` option is broken on all the LLVM-based tools because of how the cl::opt values are hacked off. Maybe we should come up with a more general solution.
> > We could use `cl::opt` external storage to allow the flag values to be trivial-initialized constants in the source files they are originally declared in (I think CommandLine.cpp), and we could implement a single source file that has a function declaring function-local static cl::opt variables that point at that storage, and we include that source file in the tools that need option parsing.
> > I think we have a similar issue with the debug options from `Debug.cpp`.
> 
> If we have to include a source file, why not just put static cl::opt in that source file?

We don't necessarily need to include a source file. We could have the cl::opt variables be function-scope globals where they're currently defined and we surface initialization function(s) to initialize them during tool initialization.

PR #5518, opened 2023-08-09
[SPIR-V] Add capability trimming pass

Keenuts:
This new optimization passes determines the required capabilities for a module, and strips unneeded capabilities.
This means if some dead-code uses a capability, it won't be added as a requirement for the module anymore.

This interacts with `[[vk::ext_capabilities]]` and [[vk::ext_extension]]` attributes, as the extension/capability they declare could be stripped if not required.

Still markes as draft for now as there are inconsistencies with debug instruction I need to figure out:
 - if 2 DebugScope are generated, SPIRV-Tools squash them into 1.
 - seems like we started to generated the wrong ones, but test didn't saw the 2 conflicting scopes, only checked the first one.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -900,20 +900,6 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
     }
   }
 
-  {
-    std::string messages;
-    if (!spirvToolsTrimCapabilities(&m, &messages)) {
-      emitFatalError("failed to trim capabilities: %0", {}) << messages;
-      emitNote("please file a bug report on "
-               "https://github.com/Microsoft/DirectXShaderCompiler/issues "
-               "with source code if possible",
-               {});
-      return;
-    } else if (!messages.empty()) {
-      emitWarning("SPIR-V capability trimming: %0", {}) << messages;
-    }
-  }
-
   if (!spirvOptions.codeGenHighLevel &&
       theCompilerInstance.getCodeGenOpts().OptimizationLevel > 0) {
     // Run optimization passes
@@ -925,6 +911,25 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
                "with source code if possible",
                {});
       return;
+    }
+  }
+
+  // Trim unused capabilities.
+  // When optimizations are enabled, some optimization passes like DCE could
+  // make some capabilities useless. To avoid logic duplication between this
+  // pass, and DXC, DXC generates some capabilities unconditionally. This means
+  // we should run this pass, even when optimizations are disabled.
+  {
+    std::string messages;
+    if (!spirvToolsTrimCapabilities(&m, &messages)) {
+      emitFatalError("failed to trim capabilities: %0", {}) << messages;
+      emitNote("please file a bug report on "
+               "https://github.com/Microsoft/DirectXShaderCompiler/issues "
+               "with source code if possible",
+               {});
+      return;
+    } else if (!messages.empty()) {
+      emitWarning("SPIR-V capability trimming: %0", {}) << messages;
     }
   }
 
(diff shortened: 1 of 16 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -874,6 +874,20 @@ void SpirvEmitter::HandleTranslationUnit(ASTContext &context) {
   auto dsetbindingsToCombineImageSampler =
       collectDSetBindingsToCombineSampledImage(resourceInfoForSampledImages);
 
+  {
+    std::string messages;
+    if (!spirvToolsTrimCapabilities(&m, &messages)) {
```
sudonatalie, line 879: I would have expected this to be added in the same place as other optimization passes, and be overridden by the command line options to change spirv opt passes. Are there reasons to treat this pass specially or that it needs to happen before legalization? If so, we should probably update some documentation at the same time.

sudonatalie: Down here: https://github.com/microsoft/DirectXShaderCompiler/pull/5518/files#diff-6b86e3fcb3f49e0a10dd39cc10c7ec49d6c00bf11422c86743ef63a34eac6303R14057

(Weird that GH won't let you comment on a line outside a diff but you can link to it?)

Keenuts: Good point.
I don't have a strong opinion, but thought about my usage to determine where to put it:

When I want to look at the non-optimized code, it's mostly to avoid DCE, unrolling, etc. I want to have a code to read that is "close" to my HLSL code, so I can understand what's happening quickly. But if I look at my empty compute shader, I might be wondering "why is DXC generating all those capabilities? like raytracing etc?".
To avoid this confusion, I decided to include this pass, even if optimization is turned off (considered this was a bit like a legalization pass).

Now, maybe this should be done after legalization, as the pass expects a valid module to work correctly (not important for now, but might later-on for variable usage detection, etc).

Changed the code to run after legalization, but still kept this pass with `-O0`. Let me know if you believe we should avoid running this if optimization are disabled.

sudonatalie: > When I want to look at the non-optimized code, it's mostly to avoid DCE, unrolling, etc. I want to have a code to read that is "close" to my HLSL code

I thought it was the opposite. Don't we specifically want this pass to run after all DCE passes so we can remove capabilities that were necessary pre-optimization but that are no longer needed post-optimization?

Keenuts: Oh yes, you are right, I'm dumb.

sudonatalie: We should add a test case that would have caught that, some HLSL using a capability which would be removed by DCE. Then you could `CHECK` that the capability is there when you `RUN` without this pass and `CHECK-NOT` when you `RUN` with this pass after DCE.

Maybe determining the API necessary to test this would help inform how we want it to interact with `-O` options. I think my inclination is that it should not run with `-O0` but I'm open to it if there's a good reason and it doesn't interfere with our ability to test.

sudonatalie: Sorry, never mind, I do see that it will end up in too weird of an `-O0` state when we add a lot more capabilities by default and rely on this pass for their removal, so maybe it does unfortunately need to run on `-O0`. I'm not sure what that means for our testing strategy though.

Keenuts: You are right we should have a test that validates that. I'll check that.
For the tests, it seems we default by running with `-fcgl`, so we should be able to test `-O0` and get the capability trimming. Also with `-O3` and some DCE. I'll try that.

Keenuts: All right, changed the behavior, and added tests.
This allowed me to revert some changed tests, but needed to put `-O0` to WholeFile test, otherwise they would run with `-fcgl`.

So behavior is:
`-fcgl`: keep all the capabilities, we are debugging compilation/opt issue.
So with time, those shaders will have 40 superfluous capabilities, but that's expected.
`-O0`: trim capabilities, but don't optimize.

We run after legalization, and optimization.

sudonatalie (approved):

cassiebeckley (approved):

PR #5520, opened 2023-08-09
[C++17] Force C++17 when build DXC

python3kgae:
This is for enable clang-cl build on windows.

The gtest fix is merging from
 https://github.com/google/googletest/pull/1620/commits/f66ab00704cd47e4e63ef6d425ca14b9192aaebb
to work around 'operator <<' is ambiguous on DefaultPrintNonContainerTo.

### DIFF
--- a/lib/Support/Windows/MSFileSystem.inc.cpp
+++ b/lib/Support/Windows/MSFileSystem.inc.cpp
@@ -236,25 +236,6 @@ static bool is_separator(const wchar_t value) {
   }
 }
 
-// TODO: consider erasing this
-namespace {
-  [[maybe_unused]]
-  error_code TempDir(_In_ MSFileSystemRef fsr, SmallVectorImpl<wchar_t> &result) {
-  retry_temp_dir:
-    DWORD len = fsr->GetTempPathW(result.capacity(), result.begin());
-
-    if (len == 0)
-      return mapWindowsError(::GetLastError());
-
-    if (len > result.capacity()) {
-      result.reserve(len);
-      goto retry_temp_dir;
-    }
-
-    result.set_size(len);
-    return error_code();
-  }
-}
 
 namespace llvm {
 namespace sys  {
--- a/lib/Support/Windows/MSFileSystem.inc.cpp
+++ b/lib/Support/Windows/MSFileSystem.inc.cpp
@@ -238,6 +238,7 @@ static bool is_separator(const wchar_t value) {
 
 // TODO: consider erasing this
 namespace {
+  [[maybe_unused]]
   error_code TempDir(_In_ MSFileSystemRef fsr, SmallVectorImpl<wchar_t> &result) {
   retry_temp_dir:
     DWORD len = fsr->GetTempPathW(result.capacity(), result.begin());
--- a/lib/Support/Windows/Mutex.inc
+++ b/lib/Support/Windows/Mutex.inc
@@ -24,7 +24,8 @@ using namespace sys;
 
 MutexImpl::MutexImpl(bool /*recursive*/)
 {
-  static_assert(sizeof(data_) == sizeof(CRITICAL_SECTION), "");
+  static_assert(sizeof(data_) == sizeof(CRITICAL_SECTION),
+    "CRITICAL_SECTION size mismatch");
   // data_ = new CRITICAL_SECTION; // HLSL Change
   InitializeCriticalSection((LPCRITICAL_SECTION)data_);
 }
--- a/lib/Support/Windows/Mutex.inc
+++ b/lib/Support/Windows/Mutex.inc
@@ -24,7 +24,7 @@ using namespace sys;
 
 MutexImpl::MutexImpl(bool /*recursive*/)
 {
-  C_ASSERT(sizeof(data_) == sizeof(CRITICAL_SECTION));
+  static_assert(sizeof(data_) == sizeof(CRITICAL_SECTION), "");
   // data_ = new CRITICAL_SECTION; // HLSL Change
   InitializeCriticalSection((LPCRITICAL_SECTION)data_);
 }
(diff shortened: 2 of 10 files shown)


projects/dxilconv/lib/DxilConvPasses/DxilCleanup.cpp
```diff
@@ -1307,6 +1309,8 @@ bool DxilCleanup::IsDxilBitcast(Value *pValue) {
       case OP::OpCode::BitcastI32toF32:
       case OP::OpCode::BitcastI64toF64:
         return true;
+      default:
+        break;
```
bob80905, line 1313: Nit: Would replacing this break; with return false; improve efficiency?

lib/Support/Windows/MSFileSystem.inc.cpp
```diff
@@ -238,6 +238,7 @@ static bool is_separator(const wchar_t value) {
 
 // TODO: consider erasing this
 namespace {
+  [[maybe_unused]]
```
pow2clk, line 241: I don't think this is used. The comment says we should consider erasing it. I think we should. Unlike the above that depends on a define and the platform in question

projects/dxilconv/lib/DxilConvPasses/DxilCleanup.cpp
pow2clk: I would trust the compiler to do the right thing and hesitate to make that change since we could theoretically add additional code afterward that we might want to run in this case. 

However, since this is a case that we don't expect to happen, an assert or unreachable might be appropriate as it might help us catch a bug if ever we add a bitcast

projects/dxilconv/lib/DxilConvPasses/DxilCleanup.cpp
```diff
@@ -1265,6 +1266,7 @@ Value *DxilCleanup::CastValue(Value *pValue, Type *pToType, Instruction *pOrigIn
     ArgTypes[2] = Type::getInt64Ty(*m_pCtx);
   } else {
     IFT(DXC_E_OPTIMIZATION_FAILED);
+    OpCode = DXIL::OpCode::NumOpCodes;
```
pow2clk, line 1269: Nit: Since the line above will unconditionally throw, this will never be reached. I know the error you're trying to avoid, but you could also avoid it by initializing the opcode on the first line it's declared.

tools/clang/lib/Parse/ParseStmt.cpp
```diff
@@ -189,11 +189,11 @@ Parser::ParseStatementOrDeclarationAfterAttributes(StmtVector &Stmts,
     // punctuator, then we are using them as identifers. Need to change
     // the token type to tok::identifier and fall through to the next case.
     // E.g., center = <RHS>.
-    if (tok::isPunctuator(NextToken().getKind())) {
+    if (!tok::isPunctuator(NextToken().getKind())) {
+      goto tok_default_case;
+    } else {
```
pow2clk, line 194: The goto makes this so weird. I'm guessing the problem here was that the fallthrough wasn't at the end?

tools/clang/unittests/HLSLHost/CMakeLists.txt
```diff
@@ -11,16 +11,10 @@ add_clang_executable(HLSLHost
...
-target_link_libraries(HLSLHost
-  ClangHLSLTests
-  )
-
 set_target_properties(HLSLHost PROPERTIES VERSION ${CLANG_EXECUTABLE_VERSION})
 
 include_directories(AFTER ${DIASDK_INCLUDE_DIRS})
 
-add_dependencies(HLSLHost ClangHLSLTests)
-
```
pow2clk, line 23: I presume this fixed the odd failure in automated builds. Why was this an invalid dependency?

utils/unittest/googletest/include/gtest/gtest-printers.h
```diff
@@ -683,6 +683,10 @@ void PrintTo(const ::std::pair<T1, T2>& value, ::std::ostream* os) {
   *os << ')';
 }
 
+#if GTEST_LANG_CXX11
+inline void PrintTo(std::nullptr_t, ::std::ostream* os) { *os << "(nullptr)"; }
+#endif  // GTEST_LANG_CXX11
+
```
pow2clk, line 689: Can you explain why this is necessary?

lib/Support/Windows/Mutex.inc
```diff
@@ -24,7 +24,7 @@ using namespace sys;
 
 MutexImpl::MutexImpl(bool /*recursive*/)
 {
-  C_ASSERT(sizeof(data_) == sizeof(CRITICAL_SECTION));
+  static_assert(sizeof(data_) == sizeof(CRITICAL_SECTION), "");
```
pow2clk, line 27: could we get a comment in ""?

utils/unittest/googletest/include/gtest/gtest-printers.h
python3kgae: More detail about this issue https://github.com/google/googletest/issues/1788

tools/clang/lib/Parse/ParseStmt.cpp
python3kgae: Yes.

lib/Support/Windows/Process.inc
```diff
@@ -178,6 +178,7 @@ ConvertAndPushArg(const wchar_t *Arg, SmallVectorImpl<const char *> &Args,
 
 /// \brief Perform wildcard expansion of Arg, or just push it into Args if it
 /// doesn't have wildcards or doesn't match any files.
+[[maybe_unused]]
```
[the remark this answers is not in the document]
python3kgae: Fixed.

lib/Support/Windows/MSFileSystem.inc.cpp
python3kgae: Fixed.

lib/Support/Windows/Mutex.inc
python3kgae: Fixed.

pow2clk (approved):
Looks good! Thanks for the responses

projects/dxilconv/lib/DxilConvPasses/DxilCleanup.cpp
python3kgae: Changed to return false to improve code reading efficiency.

tex3d (approved):

PR #5522, opened 2023-08-09
This is really a reference (#5473)

### LINKED ISSUE #4709: Accessing templated base class member var results in compiler crash

llvm-beanz:
This fixes a few cases that were missed when I updated `this` objects to be references. Specifically we were still generating some implicit `this` cases as pointers, and we were not always correctly looking through the reference types.

Fixes #4709

### DIFF
--- a/tools/clang/include/clang/AST/DeclCXX.h
+++ b/tools/clang/include/clang/AST/DeclCXX.h
@@ -1830,6 +1830,13 @@ public:
   /// Should only be called for instance (i.e., non-static) methods.
   QualType getThisType(ASTContext &C) const;
 
+  // HLSL Change Begin - This is a reference.
+  /// \brief Returns the type of the \c this object looking through the pointer.
+  ///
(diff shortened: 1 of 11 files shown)


hekota (approved):
LGTM! Please add the original commit sha to the description so it makes it into the commit message. Or next time do git cherry-pick -x :)

PR #5524, opened 2023-08-09
Allow select() to work with samplers (#5508)

### LINKED ISSUE #4673: Cannot use `select` intrinsic with SamplerState

pow2clk:
Add any_sampler builtin parameter type. This will match any sampler type as the select() built-in added with HLSL 2021 requires.

Adds a new built-in overload that takes the any_sampler type and ensures that the last two arguments and return type match. Many of the changes were required because we never had to ensure object parameter types matched nor transmit that type to the return type before. Most of the changes to MatchArguments in SemaHLSL are addressing new issues this revealed. Non-basic types can pass the conversion check if they outright match.

Adds testing to select() and the ternary operator that it supplements for these usages

Fixes #4673

(cherry picked from commit a6497c389403adf3ab76eef8f380cc749b93d818)

### DIFF
--- a/tools/clang/test/DXC/select_samplers.hlsl
+++ b/tools/clang/test/DXC/select_samplers.hlsl
@@ -1,0 +1,124 @@
+// RUN: %dxc -fcgl -HV 2018 -T ps_6_0 %s | FileCheck %s
+// RUN: %dxc -fcgl -HV 2021 -T ps_6_0 %s | FileCheck %s
+
+// Make sure the select() built-in works for sampler objects
+
+
+
+// SamplerState objects to use in selects
+SamplerState gSS1;
+SamplerState gSS2;
+SamplerComparisonState gSCS1;
+SamplerComparisonState gSCS2;
+
+Texture1D<float4> gTX1D;
+
+// A very slightly convoluted way to get a certain true result no matter the parameter
+bool getCond(int i) {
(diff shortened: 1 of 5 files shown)


bob80905 (approved):

tools/clang/test/DXC/select_samplers.hlsl
```diff
@@ -0,0 +1,124 @@
+// RUN: %dxc -fcgl -HV 2018 -T ps_6_0 %s | FileCheck %s
```
python3kgae, line 1: This should not be in test/DXC, test/DXC is for cmd tests on dxc.
Should be somewhere like CodeGenDXIL/operators/select.

pow2clk: Woops! I'll fix it in main in a follow up

llvm-beanz (approved):

PR #5529, opened 2023-08-10
CP: Pix changes for hitgroup/miss debugging

jeffnn:
…shaders (#5499)

Two main themes:
-Add instrumentation to the DXR hit group and miss shader types -Return a bit more info to the caller (PIX) via output text

(cherry picked from commit 3693755da61cfa419187dac56fe009e5c838a69c)


pow2clk (approved):
Still good!

llvm-beanz (approved):

Issue #5498, opened 2023-08-04
UINT underflow issue in ExecutionTests.


bob80905:
**Description**
In ExecutionTest.cpp, Under "VerifyOutputWithExpectedValueUInt", there is no check to prevent underflow. 
For example, with output = 2, ref = 3, and tolerance = 4, the expected behavior would be true, because output and ref are closer together than the tolerance value. However, the actual behavior would be that 2-3 would underflow to a very large number, and that number would be larger than 5, which would fail the check and return false.

bob80905:
Fixed by #5533

Issue #5369, opened 2023-06-28
Remove explicit __debugbreak() calls from WaveMatrix execution test


llvm-beanz:
This should not be there.

bob80905:
Fixed by #5533

PR #5533, opened 2023-08-11
[Wave Matrix] Improve stability of wave matrix tests

bob80905:
The wave matrix tests had a remaining __debug_break() call left in the code.
There was also a function that risked returning the wrong comparison value due to underflow behavior.
This PR improves the stability of the wave matrix tests by removing the __debug_break() call and preventing underflow from happening.
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5498
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5369

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -6433,20 +6433,29 @@ HRESULT TableParameterHandler::ParseTableRow() {
   return S_OK;
 }
 
-static bool CompareOutputWithExpectedValueInt(int output, int ref, int tolerance) {
-    return output - ref <= tolerance && ref - output <= tolerance;
+static bool CompareOutputWithExpectedValueInt(int output, int ref,
+                                              int tolerance) {
+  return output - ref <= tolerance && ref - output <= tolerance;
 }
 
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -6433,20 +6433,29 @@ HRESULT TableParameterHandler::ParseTableRow() {
   return S_OK;
 }
 
-static bool CompareOutputWithExpectedValueInt(int output, int ref, int tolerance) {
-    return output - ref <= tolerance && ref - output <= tolerance;
+static bool CompareOutputWithExpectedValueInt(int output, int ref,
+                                              int tolerance) {
+  return output - ref <= tolerance && ref - output <= tolerance;
```
python3kgae, line 6438: Could we change to ((output - ref) <= tolerance) && ((ref - output) <= tolerance); ?
It is hard to follow now.

python3kgae (approved):

pow2clk (approved):
The changes are fine. We are trying to have a 1:1 mapping of PRs to bugs where possible. I realize that these were both pretty small in terms of lines of code, but it would also have been easy to separate them that way. I wouldn't suggest doing it now though. Just a note for the future

PR #5537, opened 2023-08-11
DXC Tests: Filter out unsupported test, add verbose option and improve output

hekota:
Improve output of DXC test runs:
- Add `-lit-verbose` option to `hctbuild` to setup lit test runs with verbose output (`hcttest` will list test names as they run)
- Hide tests that are never supported (there was over 10K of unsupported tests found by LIT discovery)
- Show warnings about non-essential tools only in LIT debug mode
- Skip over file copying and non-LIT part of `hcttest.cmd` if all requested tests have already been run
- CMD tests have been move under clang's lit tests; keep the `hcttest cmd` option for running cmd tests separately

Number of unsupported tests was reduced from 10452 to 1 if spir-v is enabled, 39 if not. The number of tests that run did not change.


hekota:
Still work in progres....it is hiding some tests that ran before.

CMakeLists.txt
```diff
@@ -364,7 +364,9 @@ set(LIT_ARGS_DEFAULT "-sv")
 if (MSVC OR XCODE)
   set(LIT_ARGS_DEFAULT "${LIT_ARGS_DEFAULT} --no-progress-bar")
 endif()
-set(LLVM_LIT_ARGS "${LIT_ARGS_DEFAULT}" CACHE STRING "Default options for lit")
+
+# HLSL Change - disable short output mode for tests (will list all tests)
+set(LLVM_LIT_ARGS "-v --no-progress-bar" CACHE STRING "Options for lit")
```
llvm-beanz, line 369: Please don't make this the default. The default output is _much_ nicer in the terminal if you have an interactive session.

CMakeLists.txt
```diff
@@ -365,8 +368,22 @@ if (MSVC OR XCODE)
   set(LIT_ARGS_DEFAULT "${LIT_ARGS_DEFAULT} --no-progress-bar")
 endif()
 
-# HLSL Change - disable short output mode for tests (will list all tests)
-set(LLVM_LIT_ARGS "-v --no-progress-bar" CACHE STRING "Options for lit")
+# HLSL Change start - set arguments for the lit runner
+option(HLSL_TESTS_VERBOSE_OUTPUT "Enable verbose test output" OFF) # HLSL Change
```
llvm-beanz, line 372: Is there a reason for doing these as CMake options instead of just doing them as `hctbuild` options? If so, could we alternatively make them options implemented in a cache file like PredefinedParams.cmake?

The concern that I have here is that you’re adding new options that have values which are cached, and those in turn contribute to the default cached value of another option. CMake has no strategy for cache invalidation, so cached options defaulting based on other cached options results in really odd behavior usually forcing users to throw up their arms in disgust, delete their build directories and start all over again.

CMakeLists.txt
```diff
@@ -364,7 +364,6 @@ set(LIT_ARGS_DEFAULT "-sv")
 if (MSVC OR XCODE)
   set(LIT_ARGS_DEFAULT "${LIT_ARGS_DEFAULT} --no-progress-bar")
 endif()
-set(LLVM_LIT_ARGS "${LIT_ARGS_DEFAULT}" CACHE STRING "Default options for lit")
```
llvm-beanz, line 367: I assume this removal wasn't intentional.

hekota: Good catch, thank you!

llvm-beanz (approved):

tools/clang/test/CodeGenSPIRV/lit.local.cfg
```diff
@@ -1 +1,2 @@
-config.unsupported = True # HLSL Change - Disable lit suites.
+config.unsupported = True # HLSL Change - Disable and hide these lit suites.
+config.suffixes = []
```
pow2clk, line 2: It may not matter, but I wonder why some disable and hide as here, while others such as CodeGenHLSL above just hide?

pow2clk (approved):
Looks good! Just a question about consistency

tools/clang/test/CodeGenSPIRV/lit.local.cfg
hekota: The TAEF batch tests like CodeGenHLSL are running through the clang-taef suite, so they are not disabled, I am just hiding them from the LIT discovery scanner so they don't show up as unsupported tests. I don't think CodeGenSPIRV tests are included in our TAEF runs.

PR #5538, opened 2023-08-11
[Opt.exe] Add -help to opt.exe

### LINKED ISSUE #5514: opt.exe -help doesn't do anything except recommend you use opt.exe -help
**Description**
opt.exe -help doesn't do anything except recommend you use opt.exe -help. 

**Steps to Reproduce**
 opt.exe -help 

**Actual Behavior**
 opt.EXE: Unknown command line argument '-help'.  Try: 'D:/hlsl.bin/Debug/bin\opt.EXE -help'. 

**Environment**
- Host Operating System any?

bob80905:
opt.exe gave a contradictory message when given the -help argument on the command line.
opt.exe said that "-help" is an invalid command line argument, and recommends using -help.
The problem was that there was no clang opt that explicitly stored any information about the help command line arguments (-help/-?), and the default response that commandline.cpp outputted was to try -help. 
This PR addresses the problem by creating clang opts to store true or false depending on whether these help arguments are parsed from the command line, and adds a step to the main function to check these arguments, and if true, to display a help message.

Fixes #5514

### DIFF
--- a/tools/opt/opt.cpp
+++ b/tools/opt/opt.cpp
@@ -125,7 +125,14 @@ DisableOptimizations("disable-opt",
 static cl::opt<bool>
 StandardLinkOpts("std-link-opts",
                  cl::desc("Include the standard link time optimizations"));
-#endif // HLSL Change Ends
+#endif
+
+static cl::opt<bool> ShowHelp("?", cl::desc("Show the help message"));
+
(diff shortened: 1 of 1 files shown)


jeffnn (approved):
LGTM, but Xiang and Chris are the ones with stronger opinions about how this should be handled. Thanks for fixing it!

bob80905:
This apparently isn't the right direction, closing for now.

PR #5545, opened 2023-08-14
Prevent error from being emitted when WaveSize attribute used on entry point CS in lib target

### LINKED ISSUE #5374: WaveSize attribute should be allowed on a compute shader in a library
Currently we produce an error 'if (!isEntry && !isNode)'. This will produce an error when used on a compute shader entry when compiling to a library target.  That's not right.  It should be legal on a compute shader entry in a library target.

bob80905:
Previously, an error would be emitted when the WaveSize attribute was used on a compute shader contained in a lib target. This PR prevents the error from being emitted (that "attribute WaveSize only valid on entry point function") by incorporating whether or not the compiler is dealing with a lib target. 

Fixes #5374


hekota (approved):
LGTM!

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrAllowed.hlsl
```diff
@@ -0,0 +1,15 @@
+// RUN: %dxc -E main -T lib_6_7 %s | FileCheck %s
+
+// This test used to fail because it didn't take into account
+// the possibility of the compute shader being a library target.
+// Now, this test is an example of what should be allowed
+// and no errors are expected.
+
+// CHECK: define void @main()
```
pow2clk, line 8: I think we're banning tests that only check for main I'm afraid. Could you check that the attrib is applied?

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrAllowed.hlsl
```diff
@@ -0,0 +1,15 @@
+// RUN: %dxc -E main -T lib_6_7 %s | FileCheck %s
```
pow2clk, line 1: library compiles don't require entry points. To ensure that this is succeeding not due to it being an entry point, it would be best to eliminate the -E and rename the function too since "main" is the default entry point .

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1839,7 +1839,7 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
       Diags.Report(Attr->getLocation(), DiagID);
       return;
     }
-    if (!isEntry && !isNode) {
+    if ((!SM->IsLib() && !isEntry) && !isNode) {
```
pow2clk, line 1842: I think we need a little more here. If it's a library, we should still verify that it is annotated with `shader("compute")` I'm not sure that's done by the above. The assignment of isCS based on whether its numthreads is pretty suspicious

pow2clk (commented):
I think we need a few additional check and tests

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrAllowed.hlsl
```diff
@@ -0,0 +1,15 @@
...
+// This test used to fail because it didn't take into account
+// the possibility of the compute shader being a library target.
+// Now, this test is an example of what should be allowed
+// and no errors are expected.
+
+// CHECK: define void @main()
+
+[Shader("compute")]
+[WaveSize(8)]
+[numthreads(2,2,2)]
```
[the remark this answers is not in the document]
bob80905: FYI, by just removing the numthreads line on the commit that you reviewed, this is emitted:
```
error: validation errors
D:\DXC\tools\clang\test\HLSLFileCheck\validation\WaveSizeAttrAllowed.hlsl:12: error: Declared Thread Group X size 3575194952 outside valid range [1..1024].
D:\DXC\tools\clang\test\HLSLFileCheck\validation\WaveSizeAttrAllowed.hlsl:12: error: Declared Thread Group Z size 3575196168 outside valid range [1..64].
D:\DXC\tools\clang\test\HLSLFileCheck\validation\WaveSizeAttrAllowed.hlsl:12: error: Declared Thread Group Count 3743912256 (X*Y*Z) is beyond the valid maximum of 1024.
Validation failed.
```
That is why I had to add numthreads.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
bob80905: I don't think it's the responsibility of this code in this area to verify that the compute shader attribute exists. For example, I tried changing the shader attribute type to pixel, and later on in the compilation there was an error emitted that there's been an invalid shader attribute combination. So, validation happens later. If you're concerned about there not being validation for non-compute shaders in this specific attribute combination, that validation is covered later on.

Separately, however, I could remove isCS from being set if numthreads is detected, and instead set it if shader("compute") is detected.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrAllowed.hlsl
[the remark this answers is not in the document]
bob80905: As for the case where a shader just has the WaveSize and shader("node") attributes, should the failure reason only be because of an invalid attribute combination?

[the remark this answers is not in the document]
pow2clk: That's not a great error, but it's a separate issue. I think wavesize and numthreads should fail with this error for the sake of including numthreads:
`          DiagnosticsEngine::Error, "attribute numthreads only valid for CS/MS/AS.");`

There should be a similar error for wavesize alone in a lib shader:
       ` DiagnosticsEngine::Error,  "attribute WaveSize only valid for CS.");`

tools/clang/lib/CodeGen/CGHLSLMS.cpp
pow2clk: It's an oddity of how the code is structured that isCS ends up true here since when it finds numthreads and not AS nor MS, it sets isCS to true and only produces an error if there is an entry point. I think isCS shouldn't be true unless that atrrib is there. That could solve a lot of these problems. I'm not sure isNode should determine if this error is displayed at all, but we may have some conversations yet to have over combining those two things

The original bug references "a compute shader entry when compiling to a library target". The `shader("compute")` attrib is what makes it a compute shader entry. As is I think this change will allow WaveSize on any function when compiling to a library. That isn't right either.

pow2clk: How much later does the invalid attribute combination message take? We're trying to move a lot of errors sooner, and this is pretty much the last stop of codegen.

bob80905: Line 1989, where const unsigned profileAttributes is defined.

llvm-beanz: This diagnostic shouldn't be generated in CodeGen at all. We should move this into Sema.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1656,14 +1656,24 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+    if (Attr->getStage() == "compute") {
+      funcProps->shaderKind = DXIL::ShaderKind::Compute;
       isCS = true;
+    } 
+    else if (Attr->getStage() == "node") {
+      const HLSLShaderAttr *Attr2 = FD->getAttr<HLSLShaderAttr>();
+      if (Attr2 && Attr2->getStage() == "compute")  {
+        funcProps->shaderKind = DXIL::ShaderKind::Compute;
+        isCS = true;
+      }
```
pow2clk, line 1670: Just as an FYI, this isn't going to work. `getAttr` will return only the first atrribute of that type. Calling it twice won't give you the next attr. You'll need to do something like on https://github.com/microsoft/DirectXShaderCompiler/blob/040e4078b3525bccef0b53464b17e0b49d518c1b/tools/clang/lib/CodeGen/CGHLSLMS.cpp#L1574  

In fact, if you follow the lambda functions down, that's the place that isCS and friends gets set in the first place. So all that is needed here for isCS to be correct is to not overwrite it just because we bumped into a numthreads attribute. I think Tex cleaned this up a while ago, but that stray assignment got left behind.

pow2clk (commented):
I think if we trust the previous assignment of isCS, we should be good.

pow2clk:
I think the two of you have miraculously been able to dance between the rain drops in terms of changing the exact same lines of code, nevertheless, whichever of this change and #5467 gets merged first is probably going to create hopefully trivial conflicts for the other

tools/clang/lib/CodeGen/CGHLSLMS.cpp
pow2clk: We should likely land #5467 first then as it provides something of a framework or at least a model by which that could be done

bob80905: Agreed!

llvm-beanz:
Can we move this to Sema? We should not generate diagnostics in CodeGen.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1656,14 +1656,16 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
   }
 
+  // set isCs
+  for (auto *Attr : FD->specific_attrs<HLSLShaderAttr>()) {
+    if (Attr->getStage() == "compute") {
+      funcProps->shaderKind = DXIL::ShaderKind::Compute;
+      isCS = true;
+    }   
+  }
+
```
pow2clk, line 1666: This is correct now, but I think it's unnecessary. The isCS assignment present before was setting it inappropriately, but the `SetStageFlag` and `SetShaderKind` lambdas that get called by the `ParseShaderStage` lambda above will set both the shaderKind and isCS appropriately and I don't know of anything between there and here that would change that

pow2clk: Prior to the addition of those lambdas, the assignment of isCS and other flags was strewn throughout this code. I'm not sure of the exact steps, but I think the assignment this code is trying to replace got left in somehow needlessly

PR #5548, opened 2023-08-15
Enable CodeGenHashStability tests by default on all platforms

llvm-beanz:
On my local machine the CodeGenHashStability test takes 15 seconds to run. While that is longer than the entire rest of the test suite, it makes the full test suite run only take 17 seconds, which is still reasonable.

This change then goes a step further and breaks up the CodeGenHashStability test into multiple smaller tests. With that refactoring the full test suite drops from 17s to 10s. The end result is that enabling CodeGenHashStability costs only 1s on Release+Aserts builds.

Here is a rough breakdown of the performance impact of this change on test time for the full test suite:

```
┌───────────────┬───────────────┬───────────────┐
│               │Release+Asserts│ Debug+Asserts │
├───────────────┼───────────────┼───────────────┤
│ Before        │      9s       │      57s      │
├───────────────┼───────────────┼───────────────┤
│ After         │     10s       │      88s      │
└───────────────┴───────────────┴───────────────┘
```

We should enable this test case by default and evaluate ways to better parallelize this test so that so that we can run it more efficiently.

### DIFF
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -254,14 +254,20 @@ public:
   TEST_METHOD(BatchValidation)
   TEST_METHOD(BatchPIX)
 
+  TEST_METHOD(CodeGenHashStabilityD3DReflect)
+  TEST_METHOD(CodeGenHashStabilityDisassembler)
+  TEST_METHOD(CodeGenHashStabilityDXIL)
+  TEST_METHOD(CodeGenHashStabilityHLSL)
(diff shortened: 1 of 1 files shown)


bob80905 (approved):
Why remove the priority property? Otherwise LGTM

llvm-beanz:
> Why remove the priority property? Otherwise LGTM

The priority property prevents the test from running in the default case. This change’s intent is to make the test run with all the other tests (default priority 0), and to split it into multiple test cases that can run in parallel so that the impact on overall testing time is reduced.

PR #5549, opened 2023-08-15
Enable some tests on non-Windows

llvm-beanz:
These tests all build and pass on non-Windows platforms, so we should run them there too.


pow2clk (approved):
Looks good. Really not sure how I missed adding those refelection-dependent tests when I added support for it

bob80905 (approved):

PR #5555, opened 2023-08-16
[HCT] Add user defined query functionality

### LINKED ISSUE #5556: [Feature Request] hctdb.py should have some functionality to search through dxil instructions
There isn't a programmatic way to search through the dxil instructions defined in gen_intrin_main.txt. 

hctdb.py should have some functionality to search through the dxil instructions as if it were a database. A user should be able to define a query and obtain a set of instructions that satisfy that query from the total set of dxil instructions.

I considered creating a csv that represents the data in gen_intrin_main.txt and treating that as a formal database so that a query language could be used on it, but that felt like over kill.

**Additional context**
This issue was raised because searching for instructions that allowed for proper testing of https://github.com/microsoft/DirectXShaderCompiler/pull/5451
was difficult.

bob80905:
This PR aims to add some functionality to hctdb_instrhelp.py to efficiently search and find specific instructions.
For example, in #5451, a parser needed to be written to parse through gen_intrin_main.txt to search for any set of instructions that:

1. had at least one function attribute that was "readnone"
2. All had the same return type
3. All had the same argument list (types and length)

This search could've been made much easier if hctdb_instrhelp.py had a method to search through instructions in a database-like manner. 
This PR allows the user to define a query function with which to filter out dxil instructions in hctdb.py. The instructions can be filtered out based on function attributes, return type, function name, and the function argument list. It should allow the user to quickly find instructions that satisfy any arbitrary requirements.

On the command line, the user may run: `python3 hctdb_instrhelp.py --query-dxil=<filename-no-extension>`
This command will instruct hctdb_instrhelp.py to run a query using the inst_query_dxil function defined by the filename argument in the command line. Alternatively, for hl operations, one can instead use the --query-hlsl argument, and define a function called inst_query_hlsl in that file instead.
And then the query functionality of the script will be run, which applies the user-defined query function to all instructions and outputs the query results.

Fixes #5556


utils/hct/hctdb.py
```diff
@@ -3245,8 +3257,77 @@ def add_attr_arg(title_name, scope, args, doc):
...
+# return false on instructions that should fail the query
+def query(insts, inst):
+    #ex 
+    """
+    if inst.ret_type == "v":
+        return true
+    return false
+    """
+
+    #ex
```
pow2clk, line 3280: ex = example?

utils/hct/hctdb.py
```diff
@@ -3245,8 +3257,77 @@ def add_attr_arg(title_name, scope, args, doc):
...
+
+def parse_query(db, options):
+    if not options.query:
+        return
+
+    instructions = []
+
+    for dxil_inst in db.instr:
+        ops = []
+        #pdb.set_trace()
```
pow2clk, line 3305: can be removed?

utils/hct/hctdb.py
```diff
@@ -3245,8 +3257,77 @@ def add_attr_arg(title_name, scope, args, doc):
...
+
+    #ex 
+    """
+    if inst.fn_attr == "" and "Quad" in inst.name:
+        found = false
+        for other_inst in insts:
+            if not (other_inst == inst) and "Quad" in other_inst.name and other_inst.fn_attr != ""
+                found = true
+                break
+        return found
```
pow2clk, line 3288: The way this is called now, I don't know why you would check all the other instructions if the one you are on failed. The calling function for this iterates through all the functions so each should get a try? 

In which case, I don't know if the insts parameter is needed

pow2clk (commented):
I have a few cleanups and questions.

I'm feeling torn about the overall approach. This is obviously a developer feature and anyone who uses it can write a python query function, but it feels wrong somehow even though the only alternative I can think of is some unwieldy command line interface trying to represent badly on the command line what can easily be done in code. It might be slightly elegant to allow identifying a user-defined python script that implements query that can be loaded if only because it prevents intermingling of temporary code with a file that is checked in

utils/hct/hctdb.py
bob80905: The insts parameter will be useful if we want to query for instructions that should only pass based on the existence of other instructions. For example, querying for any instruction where there exists another instruction with a different function attribute but with all other properties the same.

utils/hct/hctdb.py
```diff
@@ -3245,8 +3260,62 @@ def add_attr_arg(title_name, scope, args, doc):
...
+    def __init__(self, fn_attr, ret_type, name, args):
+        self.fn_attr = fn_attr
+        self.ret_type = ret_type
+        self.name = name
+        self.args = args
+    def __str__(self):
+        str_args = ",".join(self.args)
+        return "[{}] {} {}({})".format(self.fn_attr, self.ret_type, self.name, str_args)
+
+def execfile(filepath, globals=None, locals=None):
```
pow2clk, line 3273: As of the latest commit, this is no longer used. I think I'd prefer if it is though.

pow2clk (commented):
Was there something that made it harder to allow a specific file to be used to find the query function?

utils/hct/query.py
```diff
@@ -0,0 +1,27 @@
+# user defined function
+def query(insts, inst):
```
python3kgae, line 2: Change the function name to inst_query?
We may want to add things like pass_query in the future.

utils/hct/hctdb.py
```diff
@@ -5,6 +5,21 @@
 ###############################################################################
 import os
 
+def parse_command_line_options():
+    import argparse
+    parser = argparse.ArgumentParser(description="A script for outputting and searching through all Dxil Instructions")
+    
+    parser.add_argument(
+        '--query',
+        help='If given, stores a path to a query file. The option requires that the user create the specified file \
```
tex3d, line 14: I don't think this is the right approach.  Writing a script that imports hctdb.py and groks the instructions would be better.  If shared functionality is desired, add that to hctdb_instrhelp.py and import that from your query script instead.

utils/hct/hctdb.py
```diff
@@ -3245,8 +3260,52 @@ def add_attr_arg(title_name, scope, args, doc):
         add_attr_arg("Unroll", "l", "Unroll the loop until it stops executing or a max count", [{"name":"Count", "type":"int"}])
         self.attributes = attributes
 
+class InstructionSignature:
```
tex3d, line 3263: The general utility of this class seems pretty limited.  It only captures a processed subset of properties from a dxil operation.  Instead, you could write a wrapper class that captures the dxil operation upon construction and provides helper methods to retrieve/compute the properties that aren't as easy to retrieve directly.  Then you can iterate through the normal operation list, constructing the wrapper, filtering, and printing matches on the fly.  And filtering would still have access to all the properties of the original operation, instead of only what you deemed useful when writing this code.

I also don't think such a utility class belongs in hctdb.py.  If needed, it can go into hctdb_instrhelp.py.

tex3d (changes requested):
Note: The functionality here searches DXIL ops, not HL operations from gen_intrin_main.txt.  For those, you would use `db_hlsl.intrinsics` - see the callers of `get_db_hlsl` function in `hctdb_instrhelp.py` for examples.

I think this is structured backwards.  We shouldn't have to add a command line argument to hctdb.py to load your custom script and process instructions.  Instead, you can write a small script that imports hctdb and searches through operations easily.  If there's a specific common/useful query pattern for operations, it can be implemented in hctdb_instrhelp.py, which is where similar functionality is implemented now.

utils/hct/hctdb_instrhelp.py
```diff
@@ -1435,10 +1435,112 @@ def RunCodeTagUpdate(file_path):
...
+def parse_query_hlsl(db, options):
+    if not options.query_hlsl:
+        return
+
+    query_module = __import__(options.query_hlsl)
+
+    instructions = []
+
+    # The query function should be using the db_hlsl_intrinsic interface
+    # defined in hctdb.py because that is what's being loaded into
```
pow2clk, line 1498: I don't think this comment is accurate anymore?

utils/hct/hctdb_instrhelp.py
```diff
@@ -1502,3 +1604,12 @@ def RunCodeTagUpdate(file_path):
             ]
         for relative_file_path in files:
             RunCodeTagUpdate(pj(hlsl_src_dir, relative_file_path))
+
+    if args.query_dxil or args.query_hlsl:
```
pow2clk, line 1608: Not sure why these have to be in the same conditional block. DXIL will get printed before HLSL regardless and it requires an extra check when calling parse*

pow2clk (approved):
Looks good to me. I just have a few nits and wonder if a sample module might be helpful.

tex3d (changes requested):
My suggestion was to leave top-level code to your query script, and with that, I'm not really sure how much we need any of this code, given the ease at which you can drill into operation properties in python.

Perhaps instead of modifying these with this specific query approach, you could add an example query script you find useful which imports `hctdb_instrhelp` instead.  That one could even be expanded to have command line parameters for different kinds of queries that could be useful.

For example, this script works without any changes to `hctdb.py` or `hctdb_instrhelp.py`:
```py
from hctdb_instrhelp import get_db_dxil

collections = {}
for op in get_db_dxil().instr:
    if not op.is_dxil_op:
        continue
    # Skip return and opcode arg:
    key = ', '.join(map(lambda arg: arg.llvm_type, op.ops[2:]))
    collections.setdefault(key, []).append(op)

for key in sorted(collections.keys()):
    ops = collections[key]
    if len(ops) < 2:
        continue
    print("Intrinsics with same input arg types: {}".format(key))
    for op in sorted(ops, key=lambda op: op.name):
        print("    [{op.fn_attr}] {op.ops[0].llvm_type} {op.name}({args})"
            .format(op=op,
                args=', '.join(map(lambda arg: "{} {}"
                        .format(arg.llvm_type, arg.name), op.ops[2:]))))
```

tex3d (approved):

PR #5557, opened 2023-08-16
[spirv] Prevent overlapping explicit locations

### LINKED ISSUE #5540: Invalid SPIR-V: Entry-point has conflicting input location assignment
**Description**
Using a mix of **BDA/RawBufferLoad** and **double4** as pixel shader input location generates invalid spir-v.
If you remove usage of either one the compilation succeeds. (change double4 to double2 or instead of RawBufferLoad to normal SRV load)

**Steps to Reproduce**

Compiler Options:
```
-spirv -T ps_6_0 -E PSMain
```

Code:
```
struct Test {
    float4 mem1;
    float mem2;
    int mem3;
};

StructuredBuffer<Test> mytests : register(t0, space0);

uint64_t address;

struct PSInput
{
    float4 position : SV_Position;
    [[vk::location(0)]] float4 data0 : COLOR;
    [[vk::location(1)]] nointerpolation uint4 data1 : COLOR1;
    [[vk::location(2)]] nointerpolation double4 data2 : COLOR2;
    [[vk::location(3)]] nointerpolation double4 data3 : COLOR3;
};

float4 PSMain(PSInput input) : SV_TARGET
{
	return input.data0 + vk::RawBufferLoad<Test>(address).mem1;
}
```

You can also see this in Compiler Explorer: https://godbolt.org/z/eYzrbfG4z

**Actual Behavior**

```
fatal error: generated SPIR-V is invalid: [VUID-StandaloneSpirv-OpEntryPoint-08721] Entry-point has conflicting input location assignment at location 3, component 0
  OpEntryPoint Fragment %PSMain "PSMain" %gl_FragCoord %in_var_COLOR %in_var_COLOR1 %in_var_COLOR2 %in_var_COLOR3 %out_var_SV_TARGET

Compiler returned: 5
```

**Environment**
- DXC (trunk)

I can't provide any further help as I'm not SPIR-V expert.

sudonatalie:
When explicit location number assignments are used, ensure that if a variable consumes more than one Location (for example, in the case of double4) that all Locations are marked as used so that error messages are correctly emitted.

Small improvements to error messaging and fixed up existing affected tests while migrating them to LIT at the same time.

Fixes #5540

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1932,6 +1932,7 @@ bool DeclResultIdMapper::finalizeStageIOLocations(bool forInput) {
 
       const auto *attr = var.getLocationAttr();
       const auto loc = attr->getNumber();
+      const auto locCount = var.getLocationCount();
       const auto attrLoc = attr->getLocation(); // Attr source code location
       const auto idx = var.getIndexAttr() ? var.getIndexAttr()->getNumber() : 0;
 
@@ -1943,13 +1944,17 @@ bool DeclResultIdMapper::finalizeStageIOLocations(bool forInput) {
       }
 
       // Make sure the same location is not assigned more than once
-      if (locSet.isLocUsed(loc, idx)) {
-        emitError("stage %select{output|input}0 location #%1 already assigned",
-                  attrLoc)
-            << forInput << loc;
-        noError = false;
+      for (uint32_t l = loc; l < loc + locCount; ++l) {
+        if (locSet.isLocUsed(l, idx)) {
+          emitError(
+              "stage %select{output|input}0 location #%1 already assigned",
+              attrLoc)
+              << forInput << l;
+          noError = false;
+        }
+
+        locSet.useLoc(l, idx);
       }
-      locSet.useLoc(loc, idx);
 
       spvBuilder.decorateLocation(var.getSpirvInstr(), loc);
       if (var.getIndexAttr())
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1946,10 +1946,10 @@ bool DeclResultIdMapper::finalizeStageIOLocations(bool forInput) {
       // Make sure the same location is not assigned more than once
       for (uint32_t l = loc; l < loc + locCount; ++l) {
         if (locSet.isLocUsed(l, idx)) {
-          emitError(
-              "stage %select{output|input}0 location #%1 already assigned",
-              attrLoc)
-              << forInput << l;
+          emitError("stage %select{output|input}0 location #%1 already "
+                    "consumed by semantic '%2'",
+                    attrLoc)
+              << forInput << l << stageVars[idx].getSemanticStr();
           noError = false;
         }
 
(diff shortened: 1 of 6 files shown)


tools/clang/test/CodeGenSPIRV_Lit/vk.location.double4.error.hlsl
```diff
@@ -0,0 +1,12 @@
+// RUN: not %dxc -T ps_6_0 -E main -fcgl  %s -spirv 2>&1 | FileCheck %s
```
sudonatalie, line 1: @python3kgae Added you here for review since this is the first lit test I've added and it's an expected error. I noticed that `not .... 2>&1 |` was how you handled it for some of the SPIR-V tests but it doesn't look like the DXIL Validation lit tests have these extra steps to redirect stderr. Is there a LIT configuration we can take advantage of to do the same thing for SPIR-V or is this the "right way"?

sudonatalie: Example: https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/CodeGenHLSL/check_warning.hlsl#L1

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1942,13 +1943,17 @@ bool DeclResultIdMapper::finalizeStageIOLocations(bool forInput) {
...
       // Make sure the same location is not assigned more than once
-      if (locSet.isLocUsed(loc, idx)) {
-        emitError("stage %select{output|input}0 location #%1 already assigned",
-                  attrLoc)
-            << forInput << loc;
-        noError = false;
+      for (uint32_t l = loc; l < loc + locCount; ++l) {
+        if (locSet.isLocUsed(l, idx)) {
+          emitError(
+              "stage %select{output|input}0 location #%1 already assigned",
```
Keenuts, line 1949: slight nit?:
Would it be more useful to say "location X overlaps with another variable location?"
If I don't know much about locations, and I see "already assigned", I would look at the actual value X being used in my code. Overlaps, I would think there is some kind of range, and search about location ranges?
Not sure.

Keenuts (approved):
LGTM, just a warning phrasing suggestion (not sure if it's a good one tho)

tools/clang/test/CodeGenSPIRV_Lit/vk.location.double4.error.hlsl
llvm-beanz: If the error can be generated in Sema rather than CodeGen (which is generally preferrable), you can write the test using the DXC `-verify` flag instead of FileCheck.

The clang verifier has the advantage of verifying the source locations of the error. Also performing validations in Sema allows them to be identified by the libClang-based intellisense plugin so that the errors can be surfaced in editor environments.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
llvm-beanz: It also could be handy to have a note that refers to the previous declaration.

sudonatalie: @Keenuts "consumed" is the language used in the spec so I've switched to that -- I think that it reads well now for both cases where it's 1-1 (the norm) and a range (this edge case).
@llvm-beanz Done, thanks!

cassiebeckley (approved):

PR #5558, opened 2023-08-16
PIX: Educate debug data interface about nested namespaces

jeffnn:
The comment from the checkin kinda says it all:
                // DINamespace has a getScope member (that hides DIScope's)
                // that returns a DIScope directly, but if that namespace
                // is at file-level scope, it will return nullptr.
(The upshot of this is that you couldn't step into functions within a namespace in WinPIX's shader debugger)

### DIFF
--- a/lib/DxilDia/DxcPixDxilDebugInfo.cpp
+++ b/lib/DxilDia/DxcPixDxilDebugInfo.cpp
@@ -229,7 +229,24 @@ dxil_debug_info::DxcPixDxilSourceLocations::DxcPixDxilSourceLocations(
         auto* S = llvm::dyn_cast<llvm::DIScope>(DL.getScope());
         while (S != nullptr && !llvm::isa<llvm::DIFile>(S))
         {
-            S = S->getScope().resolve(EmptyMap);
+            if(auto Namespace = llvm::dyn_cast<llvm::DINamespace>(S))
+            {
+                // DINamespace has a getScope member (that hides DIScope's)
+                // that returns a DIScope directly, but if that namespace
+                // is at file-level scope, it will return nullptr.
+                if (auto * ContainingScope = Namespace->getScope())
+                {
+                    S = ContainingScope;
+                }
+                else 
+                {
+                    S = S->getFile();
+                }
+            } 
+            else 
+            {
+                 S = S->getScope().resolve(EmptyMap);
+            }
         }
 
         if (S != nullptr)
--- a/lib/DxilDia/DxcPixDxilDebugInfo.cpp
+++ b/lib/DxilDia/DxcPixDxilDebugInfo.cpp
@@ -235,18 +235,12 @@ dxil_debug_info::DxcPixDxilSourceLocations::DxcPixDxilSourceLocations(
                 // that returns a DIScope directly, but if that namespace
                 // is at file-level scope, it will return nullptr.
                 if (auto * ContainingScope = Namespace->getScope())
-                {
                     S = ContainingScope;
-                }
                 else 
-                {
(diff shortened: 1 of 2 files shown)


lib/DxilDia/DxcPixDxilDebugInfo.cpp
```diff
@@ -229,7 +229,24 @@ dxil_debug_info::DxcPixDxilSourceLocations::DxcPixDxilSourceLocations(
...
         while (S != nullptr && !llvm::isa<llvm::DIFile>(S))
         {
-            S = S->getScope().resolve(EmptyMap);
+            if(auto Namespace = llvm::dyn_cast<llvm::DINamespace>(S))
+            {
+                // DINamespace has a getScope member (that hides DIScope's)
+                // that returns a DIScope directly, but if that namespace
+                // is at file-level scope, it will return nullptr.
+                if (auto * ContainingScope = Namespace->getScope())
+                {
```
llvm-beanz, line 238: nit: LLVM style guide puts `{` on the line above with the `if`/`while`/`for`/etc statement, and omits braces for single line bodies unless the braces meaningfully add to readability.

(see: https://llvm.org/docs/CodingStandards.html)

jeffnn: I think we've had this conversation before :-). I'm keeping the formatting of the surrounding code, with the thought that I should clang-format all of the PIX-related code in one go, but I admit I keep putting that part off. Do you prefer that I follow llvm guidelines just for these changed lines, or keep the current formatting (with the expectation that I'll fix everything later), or clang-format the entire file as part of this checkin?

lib/DxilDia/DxcPixDxilDebugInfo.cpp
```diff
@@ -229,7 +229,24 @@ dxil_debug_info::DxcPixDxilSourceLocations::DxcPixDxilSourceLocations(
...
+                    S = ContainingScope;
+                }
+                else 
+                {
+                    S = S->getFile();
+                }
+            } 
+            else 
+            {
+                 S = S->getScope().resolve(EmptyMap);
```
pow2clk, line 248: Initially, I wondered if the abover retrieved scopes required a resolve(), but now I'm wondering why this one is necessary. With an empty map, it looks like it will just cast to a pointer of DIScope or else it will assert

pow2clk (approved):
FWIW, I'm inclined to keep the file's formatting for now though I find it icky too.

I'm unsure if the resolve is needed here, but I think it's harmless

lib/DxilDia/DxcPixDxilDebugInfo.cpp
llvm-beanz: I think at a minimum we should remove the braces around the single-line bodies. clang-format doesn’t generally remove those because the readability criteria is subjective.

adam-yang (approved):

lib/DxilDia/DxcPixDxilDebugInfo.cpp
jeffnn: There appears to be no syntactically-allowed way to go from a DIScopeRef to a DIScope? (It's really weird, parenthetically, that DIScope::getScope returns a DIScopeRef but DINamespace::getScope and DIModule::getScope return a DIScope* directly. Seems like that was just a mistake in the original design/implementation.)

lib/DxilDia/DxcPixDxilDebugInfo.cpp
jeffnn: Roger, wilco

PR #5559, opened 2023-08-17
Workaround for wave loop getting deleted

pow2clk:
In a loop whose only substance consists of a wave op, the loop was getting deleted becuase of a check that presumed that the wave op is read only. That's not quite right. This might be the full fix, but requires more investigation


damyanp:

damyanp:

Issue #5560, opened 2023-08-17
[Feature Request] Support full set of comparison sampler variants


pow2clk:
**Is your feature request related to a problem? Please describe.**
HLSL is missing the ability to specify comparison samples using mip bias and explicit gradients. Additionally, calculating the level of detail using a comparison sampler is not possible. 

**Describe the solution you'd like**
Add the following build-ins as part of Shader Model 6.8:

* SampleCmpBias(SamplerComparisonState, ...
* SampleCmpGrad(SamplerComparisonState, ...
* CalculateLevelOfDetail(SamplerComparisonState, ...
* CalculateLevelOfDetailUnclamped(SamplerComparisonState, ...

**Describe alternatives you've considered**
Plagiarized from Jesse's request below: 
The only alternative I can see would be to add a non-comparison sampler to the shader for every comparison sampler, and to rewrite the shader for bias to explicitly compute derivatives, and then for bias/gradient to explicitly compute a mip level (using the non-comparison sampler), and then do an explicit-LOD sample. This is... ugly.

**Additional context**
See the spec proposal here for details https://github.com/microsoft/hlsl-specs/issues/30

pow2clk:
Fixed by #5716

pow2clk:
CalculateLevelOfDetail(SamplerComparisonState, ...
CalculateLevelOfDetailUnclamped(SamplerComparisonState, ...

These aren't in the PR that closed this. Reopening.

Issue #5345, opened 2023-06-28
[SM 6.8] Review intrinsic op availability for node shaders


llvm-beanz:
hctdb.py checks which ops can be in which stages. We need to make sure nothing that shouldn't be in node shaders can be and those that can't are disallowed.

pow2clk:
Investigation revealed missing wave matrix ops in node shaders that is to be fixed as #6400. As this task was to perform the investigation, it is complete.

PR #5562, opened 2023-08-17
Reserve 6.8 samplecmp variant opcodes

pow2clk:
Shader Model 6.8 will include SampleCmpGrad and SampleCmpBias intrinsics. This reserves the opcodes without adding the ability to generate the intrinsics (yet) for collabortive development purposes

contributes to #5560

### DIFF
--- a/docs/DXIL.rst
+++ b/docs/DXIL.rst
@@ -2350,6 +2350,8 @@ ID  Name                                                  Description
 251 AnnotateNodeRecordHandle                              annotate handle with node record properties
 252 NodeOutputIsValid                                     returns true if the specified output node is present in the work graph
 253 GetRemainingRecursionLevels                           returns how many levels of recursion remain
+254 SampleCmpGrad                                         samples a texture using a gradient and compares a single component against the specified comparison value
+255 SampleCmpBias                                         samples a texture after applying the input bias to the mipmap level and compares a single component against the specified comparison value
 === ===================================================== =======================================================================================================================================================================================================================
 
 
--- a/utils/hct/hctdb.py
+++ b/utils/hct/hctdb.py
@@ -289,8 +289,8 @@ class db_dxil(object):
             self.name_idx[i].category = "Resources"
         for i in "Sample,SampleBias,SampleLevel,SampleGrad,SampleCmp,SampleCmpLevelZero,SampleCmpLevel,SampleCmpBias,SampleCmpGrad,Texture2DMSGetSamplePosition,RenderTargetGetSamplePosition,RenderTargetGetSampleCount".split(","):
             self.name_idx[i].category = "Resources - sample"
-        for i in "Sample,SampleBias,SampleCmp".split(","):
-            self.name_idx[i].shader_stages = ("library", "pixel", "compute", "amplification", "mesh")
+        for i in "Sample,SampleBias,SampleCmp,SampleCmpBias".split(","):
+            self.name_idx[i].shader_stages = ("library", "pixel", "compute", "amplification", "mesh", "node")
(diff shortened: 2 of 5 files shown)


utils/hct/hctdb.py
```diff
@@ -287,7 +287,7 @@ def populate_categories_and_models(self):
             self.name_idx[i].category = "Dot"
         for i in "CreateHandle,CBufferLoad,CBufferLoadLegacy,TextureLoad,TextureStore,TextureStoreSample,BufferLoad,BufferStore,BufferUpdateCounter,CheckAccessFullyMapped,GetDimensions,RawBufferLoad,RawBufferStore".split(","):
             self.name_idx[i].category = "Resources"
-        for i in "Sample,SampleBias,SampleLevel,SampleGrad,SampleCmp,SampleCmpLevelZero,SampleCmpLevel,Texture2DMSGetSamplePosition,RenderTargetGetSamplePosition,RenderTargetGetSampleCount".split(","):
+        for i in "Sample,SampleBias,SampleLevel,SampleGrad,SampleCmp,SampleCmpLevelZero,SampleCmpLevel,SampleCmpBias,SampleCmpGrad,Texture2DMSGetSamplePosition,RenderTargetGetSamplePosition,RenderTargetGetSampleCount".split(","):
             self.name_idx[i].category = "Resources - sample"
         for i in "Sample,SampleBias,SampleCmp".split(","):
```
tex3d, line 292: I believe `SampleCmpBias` needs to be limited by shader stage like SampleBias here.

Side note: missing `"node"` on these, though I believe they are meant to be supported just like for compute.

utils/hct/hctdb.py
```diff
@@ -2138,6 +2141,47 @@ def UFI(name, **mappings):
             db_dxil_param(0, "i32", "", "number of levels of recursion remaining")])
         next_op_idx += 1
 
+        # Comparison Sampling
+        self.add_dxil_op("SampleCmpGrad", next_op_idx, "SampleCmpGrad", "samples a texture using a gradient and compares a single component against the specified comparison value", "hf", "ro", [
```
tex3d, line 2145: Overloads for other `SampleCmp` ops are `"hfwi"`, should these be the same, or is this a mistake for the others?

pow2clk: After some offline discussion. Tex and I have agreed that these were a mistake in the first place and I filed #5564

utils/hct/hctdb.py
pow2clk: I suppose adding "node" to a number of these is part of my pending task #5345. Might as well get these right now though

docs/DXIL.rst
```diff
@@ -2350,6 +2350,8 @@ ID  Name                                                  Description
 251 AnnotateNodeRecordHandle                              annotate handle with node record properties
 252 NodeOutputIsValid                                     returns true if the specified output node is present in the work graph
 253 GetRemainingRecursionLevels                           returns how many levels of recursion remain
+254 SampleCmpGrad                                         samples a texture using a gradient and compares a single component against the specified comparison value
+255 SampleCmpBias                                         samples a texture after applying the input bias to the mipmap level and compares a single component against the specified comparison value
 === ===================================================== =======================================================================================================================================================================================================================
```
llvm-beanz, line 2355: Can you comment over [here](https://github.com/microsoft/hlsl-specs/pull/77/files#diff-33b9ed472d740f1163951807242915ff0e6e2b0f02d22753944e82a18bbaefbcR118) with the table entries for these operations?

pow2clk:
Made unnecessary by #5716

PR #5565, opened 2023-08-17
update dndxc to support latest shader models

pow2clk:
dndxc has a dropdown to indicate targets. It has to be updated to latest shading models to be able to use them. This adds 6.7 and 6.8 variants for all stages

Included 6.8 even though it's not in main yet. Best to be prepared


hekota (approved):

tools/clang/tools/dotnetc/EditorForm.Designer.cs
```diff
@@ -631,47 +631,61 @@ private void InitializeComponent()
...
+            "hs_6_8",
             "ds_6_0",
             "ds_6_1",
             "ds_6_2",
             "ds_6_3",
             "ds_6_4",
             "ds_6_5",
             "ds_6_6",
+            "ds_6_7",
+            "ds_6_8",
```
tex3d, line 680: Perhaps we should add `ms` and `as`, and remove `lib_6_1` and `lib_6_2`, since these are not valid targets.

tex3d: Can we add `lib_6_x` as well?

tex3d (approved):
LGTM!

PR #5568, opened 2023-08-18
[ExecutionTests] Add validation and signing step to isNormal Test

### LINKED ISSUE #5337: ExecutionTest::IsNormalTest needs to validate/sign assembled shader
The `ExecutionTest::IsNormalTest` test compiles, modifies, and assembles a shader for testing.  Unfortunately, when assembling with IDxcAssembler, the validator is not invoked automatically.  The test needs to manually run a validation step on the shader to ensure it gets signed.  It can operate without signing, but only when ExperimentalShaderModels is enabled.

bob80905:
The IsNormal tests required an addition to the testing framework provided by ShaderOpTest.cpp, known as the MakeShaderReplacementCallback function. This function takes a shader, compiles it to DXIL, disassembles it into a string, replaces anything specified in the function parameters, and reassembles the shader back into a dxil container. The problem with the initial implementation was that after disassembling the shader, the signature applied to the shader on the compile step would be lost.

This PR adds a validation / signing step right after assembling the shader to ensure the new DXIL container has a signature. This will prevent problems with calling CreateComputePipelineState due to a missing signature.
Verified that this fails when disabling experimental mode, and passes only when dxil.dll is available.
Fixes #5337

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -11510,12 +11510,28 @@ st::ShaderOpTest::TShaderCallbackFn MakeShaderReplacementCallback(
     {
       CComPtr<IDxcAssembler> pAssembler;
       CComPtr<IDxcOperationResult> pResult;
-      VERIFY_SUCCEEDED(dllSupport.CreateInstance(CLSID_DxcAssembler, &pAssembler));
-      VERIFY_SUCCEEDED(pAssembler->AssembleToContainer(rewrittenDisassembly, &pResult));
+      CComPtr<IDxcOperationResult> pValidationResult;
+      CComPtr<IDxcValidator> pValidator;
+
       HRESULT status;
+      HRESULT validationStatus;
+      VERIFY_SUCCEEDED(
+          dllSupport.CreateInstance(CLSID_DxcAssembler, &pAssembler));
+      VERIFY_SUCCEEDED(pAssembler->AssembleToContainer(rewrittenDisassembly, &pResult));
       VERIFY_SUCCEEDED(pResult->GetStatus(&status));
       VERIFY_SUCCEEDED(status);
       VERIFY_SUCCEEDED(pResult->GetResult(&assembledShader));
+
+      // now validate the rewritten disassembly and sign the shader
+      VERIFY_SUCCEEDED(
+          dllSupport.CreateInstance(CLSID_DxcValidator, &pValidator));
+
+      VERIFY_SUCCEEDED(pValidator->Validate(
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11572,7 +11588,7 @@ TEST_F(ExecutionTest, IsNormalTest) {
 
   // The input is -Zero, Zero, -Denormal, Denormal, -Infinity, Infinity, -NaN, Nan, and then 4 normal float numbers.
   // Only the last 4 floats are normal, so we expect the first 8 results to be 0, and the last 4 to be 1, as defined by IsNormal.
-  std::vector<float> Validation_Input_Vec = {-0.0, 0.0, -(FLT_MIN / 2), FLT_MIN / 2, -(INFINITY), INFINITY, -(NAN), NAN, 530.99f, -530.99f, 122.101f, -.122101f};
+  std::vector<float> Validation_Input_Vec = {-0.0, 0.0, -(FLT_MIN / 2), FLT_MIN / 2, -(INFINITY), INFINITY, -(NAN), NAN, 530.99f, -530.99f, 122.900f, -.122900f};
```
pow2clk, line 11591: What made this necessary?

pow2clk (approved):
Great commit description!

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11510,12 +11510,28 @@ st::ShaderOpTest::TShaderCallbackFn MakeShaderReplacementCallback(
...
+      // now validate the rewritten disassembly and sign the shader
+      VERIFY_SUCCEEDED(
+          dllSupport.CreateInstance(CLSID_DxcValidator, &pValidator));
+
+      VERIFY_SUCCEEDED(pValidator->Validate(
+          assembledShader, DxcValidatorFlags_InPlaceEdit, &pValidationResult));
+      VERIFY_SUCCEEDED(pValidationResult->GetStatus(&validationStatus));
+      VERIFY_SUCCEEDED(validationStatus);
+
+      pValidator.Release();
```
tex3d, line 11534: Note: While there is no harm in doing this, none of the other `CComPtr`s here are explicitly released, and `pValidator` could be released automatically too.

tex3d (approved):

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
bob80905: This change isn't necessary, but the numbers can be arbitrary for this test, as long as they're normal.

PR #5569, opened 2023-08-18
Inline sub-licenses into root license file

amaiorano:
The root license file made reference to licenses in subdirectories. This change brings those licenses into the root file.

This changes nothing about licensing legally, but simply makes the license more clear. This is required for Google to be able to ship DXC as part of Chrome.


llvm-beanz (approved):
LGTM

dnovillo (approved):

PR #5570, opened 2023-08-18
[bug] Fix crash in dxilgen when buffer subscript gep feeds an lcssa phi

dmpots:
[bug] Fix crash in dxilgen when buffer subscript gep feeds an lcssa phi

    This change fixes a crash in the dxilgen pass when lowering buffer
    subscripts. These subscripts feed into a gep instruction that needs to
    be replaced with the computed index and offset for the dxil buffer load
    intrinsics.

    When processing all the users of the subscripts the code would crash on
    any phi node user. We can handle single-value phis by propagating
    through to the users of the phi. The single-value phis can be inserted
    by the lcssa pass.

    After fixing this initial crash it triggered another error later in the
    compiler for a similar issue. The lowered handle used by the buffer load
    was also being fed through an lcssa phi by a later run of that pass.
    This would trigger a dxil validation error because handles are not
    allowed in phis. Since it is a single-value phi we can again propagate
    the value to the phi users to avoid this error.

    The test case was reduced from a real-world shader and only occurred
    when compiling with optimization disabled (Od).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,8 @@ WexLogFileOutput/*
 #==============================================================================#
 bindings/go/llvm/llvm_config.go
 bindings/go/llvm/workdir
+
+#==============================================================================#
+# Default output path use by Visual Studio cmake project builds.
+#==============================================================================#
+out/
--- a/.gitignore
+++ b/.gitignore
@@ -76,8 +76,3 @@ WexLogFileOutput/*
 #==============================================================================#
 bindings/go/llvm/llvm_config.go
 bindings/go/llvm/workdir
-
-#==============================================================================#
-# Default output path use by Visual Studio cmake project builds.
-#==============================================================================#
-out/
--- a/lib/HLSL/DxilNoOptLegalize.cpp
+++ b/lib/HLSL/DxilNoOptLegalize.cpp
@@ -10,6 +10,8 @@
 #include "llvm/Pass.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Instructions.h"
+#include "dxc/DXIL/DxilModule.h"
+#include "dxc/DXIL/DxilOperations.h"
 #include "dxc/HLSL/DxilGenerationPass.h"
 #include "dxc/HLSL/DxilNoops.h"
 #include "llvm/IR/Operator.h"
@@ -105,6 +107,8 @@ public:
 
   bool runOnModule(Module &M) override {
     bool Changed = false;
+    SmallVector<PHINode*, 4> LcssaHandlePhis;
+    Type *HandleType = M.GetOrCreateDxilModule().GetOP()->GetHandleType();
     DxilValueCache *DVC = &getAnalysis<DxilValueCache>();
     for (Function &F : M) {
       for (BasicBlock &BB : F) {
@@ -120,10 +124,27 @@ public:
               I->eraseFromParent();
               Changed = true;
             }
+          } else if (PHINode* Phi = dyn_cast<PHINode>(I)) {
+            if (Phi->getNumIncomingValues() == 1
+             && Phi->getType() == HandleType) {
+                Changed = true;
+                LcssaHandlePhis.push_back(Phi);
+            }
           }
         }
       }
     }
+
+    // Replace all single value phi nodes (these are inserted by lcssa) of
+    // resource handles with the resource handle itself. This avoids
+    // validation errors since we do not allow handles in phis.
+    for (PHINode* Phi : LcssaHandlePhis) {
+      assert(Phi->getNumIncomingValues() == 1);
+      Value* Handle = Phi->getIncomingValue(0);
+      Phi->replaceAllUsesWith(Handle);
+      Phi->eraseFromParent();
+    }
+
     return Changed;
   }
 };
--- a/lib/HLSL/DxilNoOptLegalize.cpp
+++ b/lib/HLSL/DxilNoOptLegalize.cpp
@@ -15,6 +15,7 @@
 #include "dxc/Support/Global.h"
 #include "llvm/IR/Operator.h"
 #include "llvm/Analysis/DxilValueCache.h"
+#include "llvm/Analysis/InstructionSimplify.h"
 
 using namespace llvm;
 
@@ -106,7 +107,6 @@ public:
 
   bool runOnModule(Module &M) override {
     bool Changed = false;
-    SmallVector<PHINode*, 4> LcssaPhis;
     DxilValueCache *DVC = &getAnalysis<DxilValueCache>();
     for (Function &F : M) {
       for (BasicBlock &BB : F) {
@@ -123,25 +123,18 @@ public:
               Changed = true;
             }
           } else if (PHINode* Phi = dyn_cast<PHINode>(I)) {
-            if (Phi->getNumIncomingValues() == 1) {
-                Changed = true;
-                LcssaPhis.push_back(Phi);
+            // Replace all simple phi values (such as those inserted by lcssa) with the
+            // value itself. This avoids phis in places they are not expected because
+            // the normal simplify passes will clean them up.
+            if (Value *NewPhi = llvm::SimplifyInstruction(Phi, M.getDataLayout())) {
+              Phi->replaceAllUsesWith(NewPhi);
+              Phi->eraseFromParent();
+              Changed = true;
             }
           }
         }
       }
     }
-
-    // Replace all single value phi nodes (these are inserted by lcssa) with the
-    // value itself. This avoids phis in places they are not expected because
-    // the normal simplify passes will clean them up.
-    for (PHINode* Phi : LcssaPhis) {
-      DXASSERT(Phi->getNumIncomingValues() == 1, "unexpected number of phi operands");
-      Value* V = Phi->getIncomingValue(0);
-      Phi->replaceAllUsesWith(V);
-      Phi->eraseFromParent();
-    }
-
     return Changed;
   }
 };
--- a/lib/HLSL/DxilNoOptLegalize.cpp
+++ b/lib/HLSL/DxilNoOptLegalize.cpp
@@ -10,8 +10,6 @@
 #include "llvm/Pass.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Instructions.h"
-#include "dxc/DXIL/DxilModule.h"
-#include "dxc/DXIL/DxilOperations.h"
 #include "dxc/HLSL/DxilGenerationPass.h"
 #include "dxc/HLSL/DxilNoops.h"
 #include "dxc/Support/Global.h"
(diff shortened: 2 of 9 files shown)


python3kgae (approved):

.gitignore
```diff
@@ -76,3 +76,8 @@ WexLogFileOutput/*
 #==============================================================================#
 bindings/go/llvm/llvm_config.go
 bindings/go/llvm/workdir
+
+#==============================================================================#
+# Default output path use by Visual Studio cmake project builds.
+#==============================================================================#
+out/
```
pow2clk, line 83: It's an incidental change, but I prefer to mention those in the description lest it be construed as being related to the main thrust of the change

tex3d, line 83: Fine, but seems like it should be an independent change.  Plus no newline at end (don't know if we care about that though).

lib/HLSL/DxilNoOptLegalize.cpp
```diff
@@ -120,10 +124,27 @@ class DxilNoOptSimplifyInstructions : public ModulePass {
...
           }
         }
       }
     }
+
+    // Replace all single value phi nodes (these are inserted by lcssa) of
+    // resource handles with the resource handle itself. This avoids
+    // validation errors since we do not allow handles in phis.
+    for (PHINode* Phi : LcssaHandlePhis) {
+      assert(Phi->getNumIncomingValues() == 1);
```
pow2clk, line 142: We use DXASSERT in other parts of this directory. We should probably be consistent here

pow2clk (approved):
I appreciate the good comments, tests, and comments in tests!

I just have a couple nitpicks

lib/HLSL/HLOperationLower.cpp
```diff
@@ -7602,6 +7602,20 @@ void TranslateStructBufSubscriptUser(
         handle, ResKind, bufIdx, baseOffset, status, OP, DL);
     }
     BCI->eraseFromParent();
+  } else if (PHINode* Phi = dyn_cast<PHINode>(user)) {
```
tex3d, line 7605: I'm a bit afraid this might not be the only place impacted by phi on handle leftover from lcssa.  I wonder if it wouldn't be better to explicitly clean these up with a pass before they will pose a problem instead?

Since these get introduced by lcssa dependency of loop-rotate, loop-unroll, dxil-loop-unroll (I think, and maybe others), should we clean these up explicitly after these passes, when lcssa form is no longer necessary?  Perhaps we could clean up all single predecessor phis (not just handles)?

lib/HLSL/DxilNoOptLegalize.cpp
```diff
@@ -120,10 +124,27 @@ class DxilNoOptSimplifyInstructions : public ModulePass {
...
+             && Phi->getType() == HandleType) {
+                Changed = true;
+                LcssaHandlePhis.push_back(Phi);
+            }
           }
         }
       }
     }
+
+    // Replace all single value phi nodes (these are inserted by lcssa) of
```
tex3d, line 138: I would also expect these should be cleaned up by resource legalization (part of `hlsl-dxil-lower-handle-for-lib`).  Maybe we are missing some cleanup for this case.

lib/HLSL/HLOperationLower.cpp
dmpots: Is there a pre-dxilgen pass where we could clean up these phis?

.gitignore
dmpots: Removed from this change.

lib/HLSL/HLOperationLower.cpp
tex3d: There's no pass that I know of specifically for this; normally they would be cleaned up by simplify-inst, but we don't want to run that on `/Od`.

Is there any harm to debuggability if we make a new pass to remove all single-pred (or maybe even single-incoming-value) phis?  This pass doesn't have to be HL or DXIL specific.

I also wonder if it's possible to have multi-incoming-value phis that would break this (or other) code, which we simply haven't hit because we don't generate the code that way normally (due to HLSL language limitations mostly).  Code like this seems fragile, considering language features we may want to add.

lib/HLSL/DxilNoOptLegalize.cpp
dmpots: That pass does run, but leaves the phi in place. Let me see if we can move the cleanup there.

lib/HLSL/HLOperationLower.cpp
tex3d: > Should we do a separate pass?

I think I'd prefer a pass, but second (third?) opinions are welcome.

The way I see it, the phi here was breaking an assumption which could likely be made elsewhere as well.  Without /Od, we are relying on simplify-inst to clean these up.  A pass to clean these up when we don't want full simplify-inst would help keep that assumption from being broken unless it really is the case that requires much more work to fix (multiple incoming values).  A targeted fix avoids the assumption for this special case in this code, but wouldn't help other places that make the same assumption.

One advantage of the pass is being able to make the pass handle slightly more complex cases that can still be simplified this way, such as a chain of single-value phis, or phis with multiple preds using the same incoming value, and recursive simplification of each.  Maybe we can even use it to sink GEPs or handle create/annotate calls in order to clean up multi-value phis, preserving downstream assumptions for these cases as well.

lib/HLSL/DxilNoOptLegalize.cpp
dmpots: From what I can tell it looks like that pass does not try to remove phis for Handles, but just for the global variables that get fed into createHandleForLib.

At the time the pass runs the IR looks like this
```
@"\01?buf@@3V?$StructuredBuffer@US@@@@A" = external global %"class.StructuredBuffer<S>", align 4
@llvm.used = appending global [1 x i8*] [i8* bitcast (%"class.StructuredBuffer<S>"* @"\01?buf@@3V?$StructuredBuffer@US@@@@A" to i8*)], section "llvm.metadata"

; Function Attrs: convergent
declare void @dx.noop() #0

; Function Attrs: nounwind
define void @main(i32* noalias, i32) #1 {
entry:
  %2 = load %"class.StructuredBuffer<S>", %"class.StructuredBuffer<S>"* @"\01?buf@@3V?$StructuredBuffer@US@@@@A", !dbg !3, !noalias !10 ; line:6 col:12
  %3 = call i32 @dx.op.loadInput.i32(i32 4, i32 0, i32 0, i8 0, i32 undef), !dbg !13 ; line:12 col:9
  call void @dx.noop(), !dbg !13 ; line:12 col:9
  %cmp = icmp ult i32 0, %3, !dbg !14 ; line:16 col:16
  %tobool = icmp ne i1 %cmp, false, !dbg !14 ; line:16 col:16
  br label %do.body, !dbg !15 ; line:14 col:5

do.body:                                          ; preds = %do.cond, %entry
  call void @dx.noop(), !dbg !16 ; line:15 col:13
  %4 = call %dx.types.Handle @"dx.op.createHandleForLib.class.StructuredBuffer<S>"(i32 160, %"class.StructuredBuffer<S>" %2), !dbg !3 ; line:6 col:12
  %5 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %4, %dx.types.ResourceProperties { i32 524, i32 4 }), !dbg !3 ; line:6 col:12
  call void @dx.noop() #1, !dbg !3, !noalias !10 ; line:6 col:12
  call void @dx.noop() #1, !dbg !17, !noalias !10 ; line:6 col:5
  call void @dx.noop(), !dbg !16 ; line:15 col:13
  br label %do.cond, !dbg !18 ; line:16 col:5

do.cond:                                          ; preds = %do.body
  %tobool1 = icmp ne i1 %tobool, false, !dbg !18 ; line:16 col:5
  br i1 %tobool1, label %do.body, label %do.end, !dbg !18 ; line:16 col:5

do.end:                                           ; preds = %do.cond
  %.lcssa = phi %dx.types.Handle [ %5, %do.cond ]
  %RawBufferLoad = call %dx.types.ResRet.i32 @dx.op.rawBufferLoad.i32(i32 139, %dx.types.Handle %.lcssa, i32 0, i32 0, i8 1, i32 4), !dbg !19 ; line:18 col:14
  %6 = extractvalue %dx.types.ResRet.i32 %RawBufferLoad, 0, !dbg !19 ; line:18 col:14
  call void @dx.noop(), !dbg !20 ; line:18 col:5
  call void @dx.op.storeOutput.i32(i32 5, i32 0, i32 0, i8 0, i32 %6), !dbg !20 ; line:18 col:5
  ret void, !dbg !20 ; line:18 col:5
}
```
The problem is the `%5` that gets fed to the phi, but it looks like the pass is trying to fix up the uses of `@"\01?buf@@3V?$StructuredBuffer@US@@@@A"`


The `DxilLowerCreateHandleForLib` pass calls `LegalizeResourceUseHelper::runOnModule` which eventually calls `ColectResourceGVUsers` on the global variable for the resource. Inside `CollectResourceGVUsers` it stops recursing when it hits a use of the `createHandleForLib` function

https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/HLSL/DxilCondenseResources.cpp#L1093C9-L1093C9

lib/HLSL/HLOperationLower.cpp
dmpots: Is there really a broad assumption that phi's won't occur? I would have expected more bugs around this area if that were the case.

I would prefer to make a smaller fix first to fix the issue we have in hand. That will be less risk overall. If we want to create a new normalizing pass we could do that as a follow up, but I'm not convinced this is a widespread issue.

dmpots: Discussed this offline some more. We already have the `dxil-o0-simplify-inst` pass as a replacement for simplifyinst when compiling with O0. We can run that pass in place of simplifyinst before dxilgen to fix this bug.

Updated the PR with that new strategy.

dmpots: The new version is failing in some internal tests because it generates invalid dxil (a phi with an undef input). It seems like the DxilValueCache is maybe getting populated with more information that it was not previously, and a later pass (dxil-remove-dead-blocks) is taking advantage of it to simplify the code. 

```
error: Instructions should not read uninitialized value.
note: at '%472 = phi float [ fdiv (float undef, float undef), %469 ], [ %432, %470 ]' in block '#10'
```

Have not had time to debug to see what exactly is going wrong.

lib/HLSL/DxilNoOptLegalize.cpp
```diff
@@ -120,10 +122,26 @@ class DxilNoOptSimplifyInstructions : public ModulePass {
               I->eraseFromParent();
               Changed = true;
             }
+          } else if (PHINode* Phi = dyn_cast<PHINode>(I)) {
+            if (Phi->getNumIncomingValues() == 1) {
```
dmpots, line 126: We should probably just call `llvm::SimplifyInst` on the Phi. That will handle both the one incoming value case and the multiple incoming values when they are all the same.

dmpots: Done.

lib/HLSL/HLOperationLower.cpp
dmpots: I'm reverting this PR back to the previous targeted fix strategy. I did not have time to investigate the failures from running the `dxil-o0-simplify-inst` more widely. Going to get this fix in since it at least leaves us in a better position than before.

PR #5571, opened 2023-08-18
[NFC] Prepare DXC sources for clang-format

llvm-beanz:
This change prepares DXC sources to be ready to clang-format. This change adds annotations to skip formatting parts of the code that will break if clang-format modifies them due to include ordering or generated C++. It also adds missing include and forward declarations where clang-formatting resulted in moving things around that caused code breakages.

### DIFF
--- a/include/dxc/DXIL/DxilConstants.h
+++ b/include/dxc/DXIL/DxilConstants.h
@@ -172,7 +172,10 @@ namespace DXIL {
     Invalid,
   };
 
+  // clang-format off
+  // Python lines need to be not formatted.
   /* <py::lines('SemanticKind-ENUM')>hctdb_instrhelp.get_enum_decl("SemanticKind", hide_val=True, sort_val=False)</py>*/
+  // clang-format on
   // SemanticKind-ENUM:BEGIN
   // Semantic kind; Arbitrary or specific system value.
   enum class SemanticKind : unsigned {
@@ -211,7 +214,10 @@ namespace DXIL {
   };
   // SemanticKind-ENUM:END
 
+  // clang-format off
+  // Python lines need to be not formatted.
   /* <py::lines('SigPointKind-ENUM')>hctdb_instrhelp.get_enum_decl("SigPointKind", hide_val=True, sort_val=False)</py>*/
+  // clang-format on
   // SigPointKind-ENUM:BEGIN
   // Signature Point is more specific than shader stage or signature as it is unique in both stage and item dimensionality or frequency.
   enum class SigPointKind : unsigned {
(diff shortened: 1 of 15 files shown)


dmpots (approved):

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -18,6 +18,9 @@
 #define _WCHAR_H_CPLUSPLUS_98_CONFORMANCE_
 #endif
 
+// clang-format off
+// Windows include ordering is fragile.
```
damyanp, line 22: The other places the comment is "Includes on Windows are highly order dependent."

(I have a nit here in that it's not because the code is being compiled on Windows that makes it fragile, it's because the Windows SDK header files are order dependent.)

damyanp (approved):

tools/clang/unittests/HLSL/DxilContainerTest.cpp
llvm-beanz: It's more than just the Windows SDK headers. It is also the Windows-specific DXC code that is order dependent and extremely fragile. DXILConv in particular is a crazy mess.

include/dxc/DXIL/DxilConstants.h
```diff
@@ -172,7 +172,10 @@ namespace DXIL {
     Invalid,
   };
 
+  // clang-format off
+  // Python lines need to be not formatted.
   /* <py::lines('SemanticKind-ENUM')>hctdb_instrhelp.get_enum_decl("SemanticKind", hide_val=True, sort_val=False)</py>*/
+  // clang-format on
```
tex3d, line 178: I don't think we want to format the generated code block either, unless we run clang-format as part of code generation.
Then, if we aren't clang-formatting code as part of code generation yet want the generated code to be formatted differently, we should update the code generator output formatting instead, right?

llvm-beanz: I made the change minimal because I do want to enable formatting as part of the generation, but that requires people to all have clang-format available in the build environment, so I want to roll that out separately.

tex3d: Without excluding the generated portion, I imagine you would have to roll that out before doing the big reformatting change, otherwise you will get a build failure due to generated code formatting changes.  I'm not sure I feel good about including generated code for clang formatting and requiring clang-format to build the project.

llvm-beanz: I don’t feel good about us checking generated code into the repo at all, but here we are.

My current process has been to clang-format everything then re-generate the files. That works fine, but does leave us with files that can’t be clang-formatted in their entirety.

Since the whole point of this is to require consistent code formatting for all incoming PRs, I don’t think it is inappropriate to require clang-format for building DXC. Given that Visual Studio and Xcode both ship clang format and it is available on every *nix platform I can think of I don’t think it is much of a burden to contribution.

PR #5572, opened 2023-08-19
WArgV is needed only on non-Windows platforms.

hekota:
Fixes internal build break.


hekota:
Incorrect merge, I will create a new PR.

PR #5574, opened 2023-08-21
[lit] [SPIRV] move debug related tests as lit test

python3kgae:
Separated change in 2 commits.
1 move the files to lit folder.
1 modify to match lit FileCheck.


tools/clang/test/CodeGenSPIRV_Lit/rich.debug.scope.after.compound.statement.hlsl
```diff
@@ -20,24 +20,24 @@ VS_OUTPUT main(float4 pos : POSITION,
...
-//CHECK-NEXT: OpLine [[file:%\d+]] 24
+//CHECK-NEXT: OpLine [[file:%[0-9]+]] 24
     float a = 3.0;
     float b = 4.0;
     x += a + b + c;
     {
 //CHECK:      DebugScope [[bb2]]
-//CHECK-NEXT: OpLine [[file:%\d+]] 32
-//CHECK-NEXT: OpStore [[var_a:%\w+]] %float_6
+//CHECK-NEXT: OpLine [[file_0:%[0-9]+]] 32
```
Keenuts, line 29: Labels cannot be redefined in real LIT tests?
(Or is that frowned upon?)

tools/clang/test/CodeGenSPIRV_Lit/shader.debug.function.hlsl
```diff
@@ -1,33 +1,33 @@
-// RUN: %dxc -T ps_6_0 -E main -fspv-debug=vulkan
+// RUN: %dxc -E main -T ps_6_0 -spirv -fcgl -Vd -fspv-debug=vulkan -Qembed_debug  %s | FileCheck %s
```
Keenuts, line 1: validation wasn't disabled before, is that expected?
Is there an issue pending for this? Or is it a bad test?

Keenuts (approved):
LGTM. Just 2 questions.
note: a bit sad github doesn't show the files as moved in the review panel (but the commit view does marks them as moved..)

tools/clang/test/CodeGenSPIRV_Lit/rich.debug.scope.after.compound.statement.hlsl
python3kgae: llvm FileCheck will report error on case like this.
https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/CodeGenSPIRV/fn.param.isomorphism.hlsl#L103

// CHECK-NEXT: [[arr_0:%\d+]] = OpLoad %int [[arr]]
// CHECK-NEXT: [[arr_0:%\d+]] = OpIAdd %int [[arr_0]] %int_1
arr_0 is redefined but here we expect the [[arr_0]] still match the old define.

tools/clang/test/CodeGenSPIRV_Lit/shader.debug.function.hlsl
python3kgae: Good catch.

The reason I added -Vd here is to match 
// CHECK:         [[clOpts:%\d+]] = OpString " -E main -T ps_6_0 -spirv -fcgl -Vd -fspv-debug=vulkan -Qembed_debug" 

If it is OK to remove the -Vd in this CHECK line, we can remove the -Vd in the RUN line.

Do you prefer remove -Vd?

Keenuts: I believe removing `-Vd` from both would be best. Same for `Qembed_debug`, don't think that's required for SPIR-V debug informations since we don't emit PDB.

tools/clang/test/CodeGenSPIRV_Lit/rich.debug.scope.after.compound.statement.hlsl
Keenuts: Right, but here redefining it would be fine no?
(I'm fine with not redefining, just curious)

tools/clang/test/CodeGenSPIRV_Lit/shader.debug.function.hlsl
python3kgae: Cool.
I'll remove -Vd and Qembed_debug.

tools/clang/test/CodeGenSPIRV_Lit/rich.debug.scope.after.compound.statement.hlsl
Keenuts: Yes, noted, thanks 😊

tools/clang/test/CodeGenSPIRV_Lit/shader.debug.function.hlsl
python3kgae: The output still have -Qembed_debug,
Changed to
OpString " -E main -T ps_6_0 -spirv -fcgl -fspv-debug=vulkan

cassiebeckley (approved):

python3kgae:
> LGTM. Just 2 questions.
> note: a bit sad github doesn't show the files as moved in the review panel (but the commit view does marks them as moved..)

We're checking if it is possible to not squash for this PR.
Please don't finish the PR before we have a conclusion.

PR #5576, opened 2023-08-21
WArgV is needed only on non-Windows platforms

hekota:
Fixes internal build break.

### DIFF
--- a/tools/clang/tools/dxr/dxr.cpp
+++ b/tools/clang/tools/dxr/dxr.cpp
@@ -35,7 +35,10 @@ using namespace hlsl::options;
 #ifdef _WIN32
 int __cdecl wmain(int argc, const wchar_t **argv_) {
 #else
-int main(int argc, const char **argv_) {
+int main(int argc, const char **argv) {
+  // Convert argv to wchar.
+  WArgV ArgV(argc, argv);
+  const wchar_t **argv_ = ArgV.argv();
 #endif
   if (FAILED(DxcInitThreadMalloc())) return 1;
   DxcSetThreadMallocToDefault();
@@ -130,10 +133,8 @@ int main(int argc, const char **argv_) {
     IFT(pLibrary->CreateIncludeHandler(&pIncludeHandler));
     IFT(dxcSupport.CreateInstance(CLSID_DxcRewriter, &pRewriter));
 
-    // Convert argv to wchar.
-    WArgV ArgV(argc, argv_);
     IFT(pRewriter->RewriteWithOptions(pSource, wName.c_str(),
-                                      ArgV.argv(), argc, nullptr, 0,
+                                      argv_, argc, nullptr, 0,
                                       pIncludeHandler, &pRewriteResult));
                         
     if (dxcOpts.OutputObject.empty()) {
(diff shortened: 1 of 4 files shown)


unittests/DxcSupport/WinAdapterTest.cpp
```diff
@@ -23,16 +23,7 @@ TEST(WArgVTest, suppressAndTrap) {
   std::wstring data [] = {L"a", L"b"};
   const wchar_t *ref_argv[] = {data[0].c_str(), data[1].c_str()};
   {
-    WArgV ArgV(argc, ref_argv);
-    const wchar_t **wargv = ArgV.argv();
-    for (int i=0;i<argc;++i) {
-      EXPECT_EQ(0, std::wcscmp(ref_argv[i], wargv[i]));
-    }
-  }
-  {
```
pow2clk, line 32: I think we need to keep this around even if it's only run on Linux. This test was expressly added to test WArgV

tools/clang/tools/dxr/dxr.cpp
```diff
@@ -35,7 +35,10 @@ using namespace hlsl::options;
 #ifdef _WIN32
 int __cdecl wmain(int argc, const wchar_t **argv_) {
 #else
-int main(int argc, const char **argv_) {
+int main(int argc, const char **argv) {
+  // Convert argv to wchar.
+  WArgV ArgV(argc, argv);
+  const wchar_t **argv_ = ArgV.argv();
```
pow2clk, line 41: This is a little icky where the previous code followed the same flow for all platforms. If it's needed or even if it's just a temporary fix, that's fine, but https://github.com/microsoft/DirectXShaderCompiler/pull/4895/files#r1055802917 indicates it was done this way initially to avoid these very ifdefs

pow2clk (commented):
I think @python3kgae and @llvm-beanz should take a look here

python3kgae (approved):

tools/clang/tools/dxr/dxr.cpp
tex3d: This ickiness is amplified by the fact that this is the only place that WargV is used.

`Dxclib.cpp` parses args, then uses `CopyArgsToWStrings` to copy parsed args to wchar_t strings for use in the API.  Could this code follow the same pattern using `RewriteOption` in place of `CoreOption`?  Then we wouldn't even need this `WargV` adapter, right?

Or perhaps rewriter functionality should just be integrated into dxc so we can just delete this code.

unittests/DxcSupport/WinAdapterTest.cpp
hekota: This removed part is testing the `wchar_t` overload of `WArgV` constructor, which is getting removed. The test code below tests the remaining `char` overload of the constructor.

tools/clang/tools/dxr/dxr.cpp
tex3d: I guess I don't feel too strongly about it.  For now, I'm ok with this if @pow2clk is.

tex3d (approved):

tools/clang/tools/dxr/dxr.cpp
pow2clk: I don't personally object to the code. I wanted to ensure we had the discussion and I feel we have

pow2clk (approved):
LGTM!

PR #5579, opened 2023-08-21
Custom extensions to enable lowering for non-resource based functions

bfavela:
A downstream consumer requires the use of custom function lowering using a json string as well as methods. For instance, a direct memory load not based on a resource.

### DIFF
--- a/lib/HLSL/HLOperationLowerExtension.cpp
+++ b/lib/HLSL/HLOperationLowerExtension.cpp
@@ -396,6 +396,30 @@ static Value* PackStructIntoVector(IRBuilder<>& builder, Value* strukt) {
   return packed;
 }
 
+static StructType* ConvertVectorTypeToStructType(Type* vecTy) {
+  assert(vecTy->isVectorTy());
+  Type* elementTy = vecTy->getVectorElementType();
+  unsigned numElements = vecTy->getVectorNumElements();
+  SmallVector<Type*, 4> elements;
+  for (unsigned i = 0; i < numElements; ++i)
+    elements.push_back(elementTy);
+
+  return StructType::get(vecTy->getContext(), elements);
+}
+
+
+static Value* PackVectorIntoStruct(IRBuilder<>& builder, Value* vec) {
+  StructType* structTy = ConvertVectorTypeToStructType(vec->getType());
+  Value* packed = UndefValue::get(structTy);
+
+  unsigned numElements = structTy->getStructNumElements();
+  for (unsigned i = 0; i < numElements; ++i) {
+    Value* element = builder.CreateExtractElement(vec, i);
+    packed = builder.CreateInsertValue(packed, element, { i });
+  }
+
+  return packed;
+}
 
 ///////////////////////////////////////////////////////////////////////////////
 // Packed Lowering.
@@ -412,17 +436,6 @@ public:
     PackArgs(args);
     Value *result = CreateCall(args);
     return UnpackResult(result);
-  }
-  
-  static StructType *ConvertVectorTypeToStructType(Type *vecTy) {
-    assert(vecTy->isVectorTy());
-    Type *elementTy = vecTy->getVectorElementType();
-    unsigned numElements = vecTy->getVectorNumElements();
-    SmallVector<Type *, 4> elements;
-    for (unsigned i = 0; i < numElements; ++i)
-      elements.push_back(elementTy);
-
-    return StructType::get(vecTy->getContext(), elements);
   }
 
 private:
@@ -449,19 +462,6 @@ private:
     }
     return result;
   }
-
-  static Value *PackVectorIntoStruct(IRBuilder<> &builder, Value *vec) {
-    StructType *structTy = ConvertVectorTypeToStructType(vec->getType());
-    Value *packed = UndefValue::get(structTy);
-
-    unsigned numElements = structTy->getStructNumElements();
-    for (unsigned i = 0; i < numElements; ++i) {
-      Value *element = builder.CreateExtractElement(vec, i);
-      packed = builder.CreateInsertValue(packed, element, { i });
-    }
-
-    return packed;
-  }
 };
 
 class PackedFunctionTypeTranslator : public FunctionTypeTranslator {
@@ -474,7 +474,7 @@ class PackedFunctionTypeTranslator : public FunctionTypeTranslator {
 
   Type *TranslateIfVector(Type *ty) {
     if (ty->isVectorTy())
-      ty = PackCall::ConvertVectorTypeToStructType(ty);
+      ty = ConvertVectorTypeToStructType(ty);
     return ty;
   }
 };
@@ -728,18 +728,18 @@ public:
         std::map<ResourceKindName, std::vector<DxilArgInfo>> LoweringInfoMap =
             ParseLoweringInfo(LoweringInfo, CI->getContext());
 
-        // Select lowering info to use based on resource kind.
-        const char *DefaultInfoName = "default";
+        // Find the default lowering kind
         std::vector<DxilArgInfo> *pArgInfo = nullptr;
-        if (LoweringInfoMap.count(DefaultInfoName))
+        if (LoweringInfoMap.count(m_DefaultInfoName))
         {
-            pArgInfo = &LoweringInfoMap.at(DefaultInfoName);
+            pArgInfo = &LoweringInfoMap.at(m_DefaultInfoName);
         }
         else
         {
-            ThrowExtensionError("Unable to find lowering info for resource");
+            ThrowExtensionError("Unable to find lowering info for custom function");
         }
-        GenerateLoweredArgs(CI, *pArgInfo);
+        // Don't explode vectors for custom functions
+        GenerateLoweredArgs(CI, *pArgInfo, true);
     }
 
    CustomLowering(StringRef LoweringInfo, CallInst *CI, HLResourceLookup &ResourceLookup)
@@ -758,15 +758,14 @@ public:
         std::string Name(pName);
 
         // Select lowering info to use based on resource kind.
-        const char *DefaultInfoName = "default";
         std::vector<DxilArgInfo> *pArgInfo = nullptr;
         if (LoweringInfoMap.count(Name))
         {
             pArgInfo = &LoweringInfoMap.at(Name);
         }
-        else if (LoweringInfoMap.count(DefaultInfoName))
+        else if (LoweringInfoMap.count(m_DefaultInfoName))
         {
-            pArgInfo = &LoweringInfoMap.at(DefaultInfoName);
+            pArgInfo = &LoweringInfoMap.at(m_DefaultInfoName);
         }
         else
         {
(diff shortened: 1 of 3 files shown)


tools/clang/unittests/HLSL/ExtensionTest.cpp
```diff
@@ -183,14 +189,15 @@ Intrinsic Intrinsics[] = {
   // Make this intrinsic have the same opcode as an hlsl intrinsic with an unsigned
   // counterpart for testing purposes.
   {L"test_unsigned","test_unsigned",   "n", { static_cast<unsigned>(hlsl::IntrinsicOp::IOP_min), false, true, false, -1, countof(TestUnsigned), TestUnsigned}},
-  {L"wave_proc",    DEFAULT_NAME,      "r", { 16, false, true, true, -1, countof(WaveProcArgs), WaveProcArgs }},
-  {L"test_o_1",     "test_o_1.$o:1",   "r", { 18, false, true, true, -1, countof(TestOverloadArgs), TestOverloadArgs }},
-  {L"test_o_2",     "test_o_2.$o:2",   "r", { 19, false, true, true, -1, countof(TestOverloadArgs), TestOverloadArgs }},
-  {L"test_o_3",     "test_o_3.$o:3",   "r", { 20, false, true, true, -1, countof(TestOverloadArgs), TestOverloadArgs }},
+  {L"wave_proc",    DEFAULT_NAME,      "r", { 16, false, true, true, -1,  countof(WaveProcArgs), WaveProcArgs }},
```
dmpots, line 192: Do we need to change these lines? Having trouble seeing the actual diff.

lib/HLSL/HLOperationLowerExtension.cpp
```diff
@@ -713,10 +719,30 @@ Value *ExtensionLowering::Resource(CallInst *CI) {
...
 {
 public:
-    CustomResourceLowering(StringRef LoweringInfo, CallInst *CI, HLResourceLookup &ResourceLookup)
+    CustomLowering(StringRef LoweringInfo, CallInst* CI)
+    {
+        // Parse lowering info json format.
+        std::map<ResourceKindName, std::vector<DxilArgInfo>> LoweringInfoMap =
+            ParseLoweringInfo(LoweringInfo, CI->getContext());
+
+        // Select lowering info to use based on resource kind.
```
dmpots, line 731: Looks like a copy/paste of a comment that doesn't make sense in this context.

lib/HLSL/HLOperationLowerExtension.cpp
```diff
@@ -713,10 +719,30 @@ Value *ExtensionLowering::Resource(CallInst *CI) {
...
+        const char *DefaultInfoName = "default";
```
dmpots, line 732: We should share this constant name with the code below.

tools/clang/unittests/HLSL/ExtensionTest.cpp
```diff
@@ -1281,6 +1289,31 @@ TEST_F(ExtensionTest, ResourceExtensionIntrinsic) {
   VERIFY_IS_TRUE(regex.match(disassembly));
 }
 
+ TEST_F(ExtensionTest, CustomLoadOp) {
+  Compiler c(m_dllSupport);
+  c.RegisterIntrinsicTable(new TestIntrinsicTable());
+  c.Compile(
+    "float2 main(uint2 v1 : V1) : SV_Target {\n"
+    "  return CustomLoadOp(uint2(1, 2));\n"
```
dmpots, line 1297: I'm having trouble seeing how this code is exercising the new feature we are adding here, which is that we can now have functions that can have default values for parameters that are not given. So I would expect a test with two calls like

```
CustomLoadOp(uint2(1,2))
CustomLoadOp()
```

Where the first one gets the argument we pass and the second on gets the default (undef) value.

tools/clang/unittests/HLSL/ExtensionTest.cpp
```diff
@@ -1281,6 +1289,31 @@ TEST_F(ExtensionTest, ResourceExtensionIntrinsic) {
...
+    "}\n",
+    { L"/Vd" }, {}
+  );
+  std::string disassembly = c.Disassemble();
+
+  // Things to check
+  // - return type is 2xfloat
+  VERIFY_IS_TRUE(
+    disassembly.npos !=
+    disassembly.find(std::string("%1 = call { float, float } @CustomLoadOp(i32 21, <2 x i32> <i32 1, i32 2>)")));
```
dmpots, line 1307: Why do we convert the output to a struct, but leave the argument as a vector? Seems like it should handle them consistently.

dmpots:
About the commit message

> A downstream consumer requires the use of custom lowering to enable the use of direct memory loads. This change adds the support to the existing extension class.

I don't think this change has anything to do with direct memory loads. It is enabling the custom lowering via json string to now work for non-resource functions as well. Previously it would only work for resource methods.

tools/clang/unittests/HLSL/ExtensionTest.cpp
bfavela: This is whitespace to make things line up nicer. I can undo if it's distracting

dmpots: Fine with me as long as it's serving a purpose. Was just a bit hard to tell from the diff if it was intentional.

tools/clang/unittests/HLSL/ExtensionTest.cpp
bfavela: Input arguments for these custom intrinsics are now converted to structs (and not exploded)

tools/clang/unittests/HLSL/ExtensionTest.cpp
bfavela: Added optional argument test

lib/HLSL/HLOperationLowerExtension.cpp
```diff
@@ -935,7 +961,15 @@ class CustomResourceLowering
             if (ArgInfo.HighLevelArgIndex < CI->getNumArgOperands())
             {
                 Value *Arg = CI->getArgOperand(ArgInfo.HighLevelArgIndex);
-                if (ArgInfo.HasVectorIndex)
+                if (ConvertArgToStruct)
```
dmpots, line 964: I think we should check this after checking for the `ArgInfo.HasVectorIndex`. It would be useful to be able to explicitly explode the vector with the custom overload string and then pack into a struct by default.

Should also add a test to include coverage of this case. Should be fairly easy to add another vector argument that is explicitly exploded by the custom lowering string in the test.

dmpots: Thinking more about this I actually don't think we should even have the `ConvertArgToStruct` bool. We should always convert vector to structs if there is no vector index.

bfavela: That changes custom resource lowering as well, though, since it uses the same path

dmpots: Yeah I think that's ok. We should not be generating vectors in this lowering.

lib/HLSL/HLOperationLowerExtension.cpp
```diff
@@ -999,6 +991,14 @@ class CustomLowering
                         }
                     }
                 }
+                else
+                {
+                    // This also disables 'exploding' of vectors
```
dmpots, line 996: I think the comment here should be something like

Do this after checking for the vector index to allow the vector to be manually exploded.

dmpots (approved):

lib/HLSL/HLOperationLowerExtension.cpp
bfavela: Yep, saw that on a re-review just after I pushed.

tools/clang/unittests/HLSL/ExtensionTest.cpp
```diff
@@ -97,6 +97,27 @@ static const HLSL_INTRINSIC_ARGUMENT TestMyBufferOp[] = {
...
+};
+
+// float2 = CustomLoadOp(uint2 addr, bool val)
+static const HLSL_INTRINSIC_ARGUMENT TestCustomLoadOpBool[] = {
+  { "CustomLoadOp", AR_QUAL_OUT, 0, LITEMPLATE_VECTOR, 0, LICOMPTYPE_FLOAT, 1, 2 },
+  { "addr", AR_QUAL_IN, 1, LITEMPLATE_VECTOR, 1, LICOMPTYPE_UINT, 1, 2},
+  { "val",  AR_QUAL_IN, 2, LITEMPLATE_SCALAR, 2, LICOMPTYPE_BOOL, 1, 1},
+};
+
+// float2 = CustomLoadOp(uint2 addr, bool val, uint lo, uint hi)
```
python3kgae, line 113: Should be float2 = CustomLoadOp(uint2 addr, bool val, uint2 subscript) ?

bfavela: Yep! Thanks, that's the DXIL version, not HLSL.

tools/clang/unittests/HLSL/ExtensionTest.cpp
```diff
@@ -183,14 +204,17 @@ Intrinsic Intrinsics[] = {
...
-  {L"wave_proc",    DEFAULT_NAME,      "r", { 16, false, true, true, -1, countof(WaveProcArgs), WaveProcArgs }},
-  {L"test_o_1",     "test_o_1.$o:1",   "r", { 18, false, true, true, -1, countof(TestOverloadArgs), TestOverloadArgs }},
-  {L"test_o_2",     "test_o_2.$o:2",   "r", { 19, false, true, true, -1, countof(TestOverloadArgs), TestOverloadArgs }},
-  {L"test_o_3",     "test_o_3.$o:3",   "r", { 20, false, true, true, -1, countof(TestOverloadArgs), TestOverloadArgs }},
+  {L"wave_proc",    DEFAULT_NAME,      "r", { 16, false, true, true, -1,  countof(WaveProcArgs), WaveProcArgs }},
+  {L"test_o_1",     "test_o_1.$o:1",   "r", { 18, false, true, true, -1,  countof(TestOverloadArgs), TestOverloadArgs }},
+  {L"test_o_2",     "test_o_2.$o:2",   "r", { 19, false, true, true, -1,  countof(TestOverloadArgs), TestOverloadArgs }},
+  {L"test_o_3",     "test_o_3.$o:3",   "r", { 20, false, true, true, -1,  countof(TestOverloadArgs), TestOverloadArgs }},
+  {L"CustomLoadOp", "CustomLoadOp",    "c:{\"default\" : \"0,1,2:?i1,3.0:?i32,3.1:?i32\"}", { 21, true,  false, false, -1, countof(TestCustomLoadOp), TestCustomLoadOp}},
```
python3kgae, line 211: What does the "c:{\"default\" : \"0,1,2:?i1,3.0:?i32,3.1:?i32\"}" mean here?

bfavela: Similar to the MyTextureOp, it's a custom lowering with both optional arguments and vector exploding. I added this comment:
 // custom lowering with both optional arguments and vector exploding.
 // Arg 0 = Opcode
 // Arg 1 = Pass as is
 // Arg 2:?i1 = Optional boolean argument
 // Arg 3.0:?i32 = Optional x component (in i32) of 3rd HLSL arg
 // Arg 3.1:?i32 = Optional y component (in i32) of 3rd HLSL arg

python3kgae (approved):

Issue #5367, opened 2023-06-28
Need ability to specify profile and entry point args in verifier test


llvm-beanz:
Work graph constraints are checked no matter the target, even though they should only be checked if lib target.  But checking only if lib fails for verifier tests because there's no way to provide the profile.  We should be able to specify profile (and entry point) for verifier tests.

Note: this could be resolved by adopting LIT completely since the `-verify` option already supports this flow.

PR #5581, opened 2023-08-21
Improve diagnostics for wavesize/numthreads and move to sema

bob80905:
Previously, an error would be emitted when the WaveSize attribute was used on a compute shader contained in a lib target. This PR prevents the error from being emitted (that "attribute WaveSize only valid on entry point function") by incorporating whether or not the compiler is dealing with a lib target.

Additionally, for library shaders that only have the numthreads attribute, an error will now be emitted if there is no accompanying shader attribute e.g. shader("compute")
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5374


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14212,6 +14250,86 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
...
+              dyn_cast<StringLiteral>(ArgExpr->IgnoreParenCasts());
+          DXIL::ShaderKind Stage =
+              ShaderModel::KindFromFullName(Literal->getString());
+          isNode |= Stage == DXIL::ShaderKind::Node;
+          isCS |= Stage == DXIL::ShaderKind::Compute;          
+        }
+        pAL = pAL->getNext();
+      }
+      
+      if (!SM->IsLib() && !isEntry && !isNode) {
```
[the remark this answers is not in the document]
tex3d: In fact, without `!isCS`, this even fails if you put `[shader("compute")]` on that function.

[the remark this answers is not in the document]
tex3d: We need more tests for entry attributes where there is more than one entry function in the file, including when they don't have shader attributes, and when they are not the current entry for a non-lib-target compilation.

tex3d (changes requested):
Currently, this change regresses supported scenarios for other functions with attributes.  Needs more testing for these.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
bob80905: I think the error is expected, the error that is emitted is:
"error: WaveSize attribute only valid on entry point functions"
The error is correct, because foo_wavesize as I understand is NOT an entry point function. If the error shouldn't be emitted, then the error message should be changed. Isn't it true that wave size attributes are only valid on entry point functions?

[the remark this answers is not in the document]
bob80905: I should also note that changing your example from `void main() {}` to `void main() {foo_wavesize();}` emits the same error even without the changes in this PR:
"error: WaveSize attribute only valid on entry point functions"
Your example only compiles successfully in the current version of the compiler (without the changes in my PR) because foo_wavesize() is unreachable from the entry point function, and so there is no codegen pass executed for the foo_wavesize function. 
Given this behavior and the fact that the diagnostic is moved to Sema, I think we can expect the error to be emitted given the example you gave above.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrNotAllowed.hlsl
```diff
@@ -0,0 +1,9 @@
+// RUN: %dxc -T lib_6_7 %s | FileCheck %s
+
+// CHECK: error: attribute numthreads only valid for CS/MS/AS.
```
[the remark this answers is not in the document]
bob80905: That would be ideal, but #5367 prevents this. There will likely be a large migration of tests over to verifier tests, so I wonder if there's a way to mark these tests so that they're not forgotten.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
tex3d: This error isn't expected, unless we make a breaking change to HLSL language.  This is a commonly used pattern for HLSL to have multiple entry points in the same file and compile each separately.

In order to avoid that breaking change, we will need to follow control flow graph from the entry (or each entry and export for lib target), and collect all the functions that are used, then validate these.  This means this validation would need to happen under `DiagnoseTranslationUnit` instead of when each function is constructed.

If you call the function, however, we do need to validate the attributes as a called function, and in this case, they are not valid for a non-entry point function, so we would emit an error in that scenario.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11352,6 +11352,111 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+        ShaderModel::KindFromFullName(Literal);
+    isNode |= Stage == DXIL::ShaderKind::Node;
+    isCS |= Stage == DXIL::ShaderKind::Compute;
+  }
+
+  HLSLWaveSizeAttr *attr = FD->getAttr<HLSLWaveSizeAttr>();
+  
+  if (!SM->IsLib() && !isEntry && !isNode) {
+    S->Diag(attr->getRange().getBegin(),
+            diag::err_hlsl_attribute_valid_on_entry_function_only)
```
llvm-beanz, line 11384: This is going to be a problematic regression. We currently allow inactive entry points in shaders like:

```hlsl
[WaveSize(32)]
[numthreads(32, 1, 1)]
void csmain1(){ }

[WaveSize(32)]
[numthreads(32, 1, 1)]
void csmain2(){ }
```
[Compiler Explorer Link](https://godbolt.org/z/3qMMssM39)

And entry points can be called as regular functions, so the call graph walking won't help.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7473,6 +7473,14 @@ def err_hlsl_attribute_mismatch: Error<
   "invalid shader stage attribute combination">;
 def err_hlsl_attribute_valid_on_function_only: Error<
   "attribute is valid only on functions">;
+def err_hlsl_attribute_valid_on_entry_function_only: Error<
+  "%0 attribute only valid on entry point functions">;
+def err_hlsl_attr_for_wrong_shader: Error<
+  "attribute %0 only valid for %1.">;
```
llvm-beanz, line 7479: Maybe:

```
def err_hlsl_attribute_unsupported_stage : Error<
  "attribute %0 requires shader stage %1">;
```
Note no period at the end of the diagnostic message which is consistent with clang diagnostic styles.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14211,7 +14357,82 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
...
+          isCS |= Stage == DXIL::ShaderKind::Compute;
+          isMS |= Stage == DXIL::ShaderKind::Mesh;
+          isAS |= Stage == DXIL::ShaderKind::Amplification;
+        }
+        pAL = pAL->getNext();
+      }
+
+      if (!isNode && SM->IsLib() &&
+          ((!isCS && !isMS && !isAS) &&
+           (!SM->IsCS() && !SM->IsMS() && !SM->IsAS()))) {
```
bogner, line 14394: This looks similar to the other case, except for here we're checking both the local variable and the values from `SM` - should they be doing the same thing as each other? Furthermore this is a pretty difficult condition to read, it'd be nice to simplify it.

Can we consolidate the logic to work out which shader stage it is into a function and simplify all of this?

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7473,6 +7473,14 @@ def err_hlsl_attribute_mismatch: Error<
...
+def err_hlsl_numthreads_attr: Error<
+  "attribute numthreads only valid for CS/MS/AS.">;
```
llvm-beanz, line 7481: This error could just reuse the error above with a constructed string. It is also probably better to put the full stage names in a sentence like structure: "compute, mesh, or amplification".

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11352,6 +11352,111 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+    functionName = FD->getIdentifier()->getName();
+  bool isEntry = false;
+  bool isNode = false;
+  bool isCS = false;
+
+  isEntry = !SM->IsLib() &&
+            FD->getDeclContext()->getDeclKind() == Decl::Kind::TranslationUnit &&
+            functionName == entryName;
+
+  for (auto *pAL : FD->specific_attrs<HLSLShaderAttr>()) {
```
llvm-beanz, line 11372: This logic will also simplify with #5601

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrNotAllowed.hlsl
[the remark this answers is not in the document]
llvm-beanz: If you put this test into `tools/clang/test/SemaHLSL/...` instead of under `HLSLFileCheck` it can run as a lit test using the `dxc -verify` flag which respects the `-T` flag.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11352,6 +11352,111 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+    StringRef Literal = pAL->getStage();
+    DXIL::ShaderKind Stage =
+        ShaderModel::KindFromFullName(Literal);
+    isNode |= Stage == DXIL::ShaderKind::Node;
+    isCS |= Stage == DXIL::ShaderKind::Compute;
+    isMS |= Stage == DXIL::ShaderKind::Mesh;
+    isAS |= Stage == DXIL::ShaderKind::Amplification;    
+  }
+
+  if (!isNode && isEntry && !SM->IsCS() && !SM->IsMS() && !SM->IsAS()) {
```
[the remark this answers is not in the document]
bob80905: Nice catch, fixed.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: They should not be doing the same thing. For example, isCS means that the function has a [shader("compute")] shader stage attribute. SM->isCS() means that the command line has information that the compilation is targeted to a compute profile, like -T cs_6_6. Glad you pointed this out though, because SM->isLib() actually implies the last 3 conditions, so I can just remove them!

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
bob80905: Excellent simplification, fixed.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: I've added a test with this code (WaveSizeAttrAllowed4.hlsl), my changes don't regress successful compilation of the code you've posted above.
For extra clarity, I should note that the code inside DiagnoseTranslationUnit() is specifically for validation of entry point functions. There is separate code in DiagnoseHLSLDecl that is responsible for diagnosing lib shaders that are *not* entry points. The code locations are necessary because those are the only locations that validate the declarations that are of interest at that point. By that I mean, DiagnoseTranslationUnit() is the only convenient location that has a natural way to validate all functions reachable from the entry point. This location is worthless for shaders that are targeting libraries on compilation. Conversely, the DiagnoseHLSLDecl() location will be executed on all function declarations if the shader is targeting a library profile, which changes the rules of validation.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrAllowed.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: %dxc -T lib_6_7 %s | FileCheck %s
```
pow2clk, line 1: /validation/ isn't the right place for any of these. I suspect they should all be verification tests even the ones that just verify that no error is produced

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrNotAllowedNode.hlsl
```diff
@@ -0,0 +1,12 @@
+// RUN: %dxc -T lib_6_7 %s | FileCheck %s
```
pow2clk, line 1: This should be lib_6_8. I suspect once you get past the initial error, you'll get one about that (or should anyway)

pow2clk (commented):
I'm having some second thoughts about the behavior I recommended before. Sorry to have led you down a dead end.

I have some other comments about the implementation and testing specifically.

Issue #5479, opened 2023-07-31
[Build] Compile with C++17


python3kgae:
To build with clang-cl on windows need to use C++17.

Need to fix the error/warning when switch to C++17.

Goal:
Build and pass all test with C++17 enabled.

PR #5582, opened 2023-08-21
[C++17] Remove std option for linux build

python3kgae:
Use CMAKE_CXX_STANDARD instead of -std=c++14 for linux.

Dependent on https://github.com/microsoft/DirectXShaderCompiler/pull/5520

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5479


pow2clk:
I think if you were to include the CMakeLists.txt change from #5520, this should compile fine on non-win platforms. That way this could really go in before the clang-cl change if needed and if it's exactly the same, there shouldn't even be any merge conflicts

python3kgae:
> I think if you were to include the CMakeLists.txt change from #5520, this should compile fine on non-win platforms. That way this could really go in before the clang-cl change if needed and if it's exactly the same, there shouldn't even be any merge conflicts

Only include the CMakeList.txt change will get build error for msvc build :(
Maybe the best solution is put this change into #5520 or wait #5520 to be done first?

pow2clk (approved):
Looks good!

tex3d (approved):

PR #5588, opened 2023-08-23
Disallow NodeOutput or EmptyNodeOutput array node parameters

### LINKED ISSUE #5359: Diagnostics for native NodeOutput or EmptyNodeOutput array declaration
The Work Graphs spec added NodeOutputArray and EmptyNodeOutputArray in place of native NodeOutput and EmptyNodeOutput array declarations.
However, it is still possible to declare native NodeOutput or EmptyNodeOutput arrays, which can lead to one of several compiler asserts.
In order to avoid the inadvertent use of native arrays (and improve the user experience) error diagnostics should be generated for their use.

tcorringham:
The NodeOutputArray and EmptyNodeOutputArray types are provided for use as node shader parameter types to be used instead of native array types.
Add error diagnostics for the use of native NodeOutput or EmptyNodeOutput arrays, and add a corresponding verifier test.

Fixes #5359


llvm-beanz (commented):
I think this is good with two caveats:
1) It will need to be rebased on top of the clang-formatted branch.
2) If we can also rebase it on top of #5763, we should remove the string matching of type names.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15243,17 +15243,14 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
     // arrays of NodeOutput or EmptyNodeOutput are not supported as node
     // parameters
     if (ParamTy->isArrayType()) {
```
pow2clk, line 15245: In fact, I don't think that the array variants are meant to be used as a native array type either. The way this error is written, it might sound like you just swap one type for the other. I'm not sure that will produce an error. I believe it should. I don't see any tests that try adding array braces to *Array types and my simple experiment ended with asserts.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15341,6 +15357,7 @@ void DiagnoseEntry(Sema &S, FunctionDecl *FD) {
   case DXIL::ShaderKind::Mesh:
   case DXIL::ShaderKind::Amplification:
   case DXIL::ShaderKind::Invalid:
+  case DXIL::ShaderKind::Compute:
```
tex3d, line 15360: I think this change will conflict with a change that has already addressed this in the target branch, so you could revert it.

tools/clang/test/HLSL/workgraph/nodeoutput_array.hlsl
```diff
@@ -8,16 +8,37 @@ struct Record {
...
 { }
 
 [Shader("node")]
-[NodeLaunch("Broadcasting")]
+[NodeLaunch("broadcasting")]
 [NodeMaxDispatchGrid(65535, 1, 1)]
 [NumThreads(32, 1, 1)]
-void node02(EmptyNodeOutput output[4]) // expected-error {{arrays of 'EmptyNodeOutput' are not supported as node parameters - use 'EmptyNodeOutputArray' instead}}
+// expected-error@+2 {{entry parameter of type 'EmptyNodeOutput [4]' may not be an array}}
+// expected-note@+1 {{'EmptyNodeOutput' cannot be used as an array; did you mean 'EmptyNodeOutputArray'?}}
```
pow2clk, line 25: huh. Didn't know you could do that.

pow2clk (approved):
Looks good! Thanks for expanding the errors.

tools/clang/test/HLSL/workgraph/nodeoutput_array.hlsl
llvm-beanz: The even better trick that you can do in modern clang is put a comment like `// #match-here` on a line then do `//expected-error@match-here {{..}}`. That doesn't work in DXC, but does un upstream and it is super nice.

tools/clang/test/HLSLFileCheck/d3dreflect/empty_broadcasting_nodes.hlsl
```diff
@@ -302,6 +302,6 @@ struct rec1
 [NodeDispatchGrid(2, 8, 10)]
 export void depth18part0_wg_63_nodes_seed_255(
   DispatchNodeInputRecord<rec0> InputyMcInputFace,
-  [MaxRecords(47)] [AllowSparseNodes] [NodeArraySize(2)] NodeOutput<rec1> OutputyMcOutputFace[2])
+  [MaxRecords(47)] [AllowSparseNodes] [NodeArraySize(2)] NodeOutput<rec1> OutputyMcOutputFace)
```
tex3d, line 305: Shouldn't this have changed `NodeOutput<rec1>` to `NodeOutputArray<rec1>`?  This highlights an undiagnosed problem: If you use `[NodeArraySize()]` attribute or the `[AllowSparseNodes]` on a non-array type.

llvm-beanz: I've filed https://github.com/microsoft/DirectXShaderCompiler/issues/5810 to track the additional diagnostics.

llvm-beanz: @tex3d, I've posted a change to update the test as you requested and filed an issue for the missing diagnostics.

tex3d: Great, but shouldn't we change this one to be well-formed, since the purpose isn't to test the upcoming diagnostics?

tex3d: Note that this `IOFlagsAndKind` is `NodeOutput` (6: ReadWrite|Output), instead of `NodeOutputArray` (22: ReadWrite|Output|NodeArray) as it should be:
```
// CHECK:      Outputs: <21:RecordArrayRef<IONode>[1]>  = {
// CHECK:        [0]: <1:IONode> = {
// CHECK:          IOFlagsAndKind: 6
```
And yet, some of the other properties included only apply to Array type (we should validate this in DxilValidation as well).

tex3d (approved):
LGTM!

PR #5589, opened 2023-08-23
Amend compute node compatibility DXIL validation

### LINKED ISSUE #5348: DXIL Validation for input record type compatibility with launch mode
DXIL validation needs to produce errors for invalid combinations of input records and launch modes.

* Thread launch mode entries must use ThreadNodeInputRecord inputs.
* Broadcasting launch mode entries must use DispatchNodeInputRecord inputs.
* Coalescing launch mode entries must use GroupNodeInputRecord inputs.

This should be verified both at the source (AST) level and in the resulting DXIL through the validator.

tcorringham:
Update the DXIL validation for compute node compatibility, and for node input vs launch type compatibility, to make the diagnostics more informative, and update the validation tests accordingly.

Fixes #5348

### DIFF
--- a/tools/clang/test/DXILValidation/compute_node_compatibility.hlsl
+++ b/tools/clang/test/DXILValidation/compute_node_compatibility.hlsl
@@ -14,24 +14,24 @@ struct RECORD {
 };
 
 [Shader("node")]
-[NodeLaunch("Broadcasting")]
+[NodeLaunch("broadcasting")]
 [NodeDispatchGrid(1, 1, 1)]
 [NumThreads(1,1,1)]
 void node01(DispatchNodeInputRecord<RECORD> input) { }
 
 [Shader("node")]
-[NodeLaunch("Broadcasting")]
+[NodeLaunch("broadcasting")]
 [NodeDispatchGrid(2, 1, 1)]
 [NumThreads(1,1,1)]
 void node02(RWDispatchNodeInputRecord<RECORD> input) { }
 
 [Shader("node")]
-[NodeLaunch("Broadcasting")]
+[NodeLaunch("broadcasting")]
 [NodeDispatchGrid(3, 1, 1)]
 [NumThreads(1,1,1)]
 void node03(NodeOutput<RECORD> output) { }
 
 [Shader("node")]
-[NodeLaunch("Coalescing")]
+[NodeLaunch("coalescing")]
 [NumThreads(1,1,1)]
 void node04() { }
--- a/tools/clang/test/DXILValidation/compute_node_compatibility.hlsl
+++ b/tools/clang/test/DXILValidation/compute_node_compatibility.hlsl
@@ -1,12 +1,12 @@
 // ==================================================================
-// Errors are expected for shaders with both "node" and "compute"
-// specified when:
+// Errors are expected for compute shaders when:
 // - a broadcasting node has an input record and/or output records
 // - the launch type is not broadcasting
-// This test operates by changing the [Shader(node)] metadata entry
-// to [Shader(compute)] as if both had been specified in the HLSL,
-// for each shader in turn.
-// It also changes the coalescing to thread in the final test case.
+// This test operates by changing the [Shader("node")] metadata entry
+// to [Shader("compute")] for each shader in turn.
+// It also changes the coalescing to thread in the final test case,
+// after swapping out the shader kind to compute.
+// The shader is compiled with "-T lib_6_8 -HV 2021"
 // ==================================================================
 
 struct RECORD {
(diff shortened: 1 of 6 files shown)


tools/clang/test/DXILValidation/compute_node_compatibility.hlsl
```diff
@@ -1,83 +1,37 @@
 // ==================================================================
 // Errors are expected for shaders with both "node" and "compute"
 // specified when:
-// - the launch type is not broadcasting
 // - a broadcasting node has an input record and/or output records
+// - the launch type is not broadcasting
```
tex3d, line 5: I don't think this comment is correct anymore.  This test is not about checking errors when both node and compute are specified, since we no longer support that scenario, right?  This is now a template used for validation of launch types with various input record types.  As such, the comment should probably be updated.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4281,161 +4281,140 @@ TEST_F(ValidationTest, AtomicsInvalidDests) {
...
-    L"..\\DXILValidation\\compute_node_compatibility.hlsl", "lib_6_8",
-    pArguments.data(), 2, nullptr, 0,
-    {"!{i32 8, i32 15"},  // original: node shader
-    {"!{i32 8, i32 5"},   // changed to: compute shader
-    "Node node10 with input/outputs is not compatible with Compute");
+      L"..\\DXILValidation\\compute_node_compatibility.hlsl", "lib_6_8",
+      pArguments.data(), 2, nullptr, 0,
+      {"!{i32 8, i32 15"}, // original: node shader
+      {"!{i32 8, i32 5"},  // changed to: compute shader
+      "Node 'node04' coalescing launch type is not compatible with compute",
```
tex3d, line 4310: When changing this to compute shader, shouldn't the error be different?  It's no longer identified as a node shader at all, so the error should really be that the metadata defines NodeLaunch metadata on a compute entry, right?

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4281,161 +4281,140 @@ TEST_F(ValidationTest, AtomicsInvalidDests) {
...
-    {"!{i32 8, i32 15"},  // original: node shader
-    {"!{i32 8, i32 5"},   // changed to: compute shader
-    "Node node11 with input/outputs is not compatible with Compute");
+      L"..\\DXILValidation\\compute_node_compatibility.hlsl", "lib_6_8",
+      pArguments.data(), 2, nullptr, 0,
+      {"!{i32 8, i32 15, i32 13, i32 2"}, // original: node shader, coalesing
+                                          // launch type
+      {"!{i32 8, i32 5, i32 13, i32 3"},  // changed to: compute shader, thread
+                                          // launch type
+      "Node 'node04' thread launch type is not compatible with compute", false);
```
tex3d, line 4319: Same here: I think the error should be different.

tex3d: Also, a test with broadcasting launch should be thrown in for good measure, since it shouldn't be defined on a compute shader either, right?

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2486,6 +2486,26 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+        break;
+      }
+    }
+
+    if (!found) {
+      unsigned DiagID = Diags.getCustomDiagID(
+          DiagnosticsEngine::Error,
+          "node shader '%0' with NodeMaxDispatchGrid attribute must declare an"
+          " input record containing a field with the SV_DispatchGrid semantic");
+      Diags.Report(FD->getLocation(), DiagID) << FD->getIdentifier()->getName();
```
tex3d, line 2505: Where is the test for this one?

tex3d: I realize now that this code was moved in the base branch, and does have tests there now.  So why would it show up as added here?

bob80905: There were merge conflicts. I remember I had to add this code in to resolve the conflict.

python3kgae (approved):

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tex3d: I don't think the merge conflict was resolved correctly, otherwise this shouldn't be showing up in the diff anymore.  It might be simplest at this point to rebase and squash your changes on top of the target branch, then remove this from the final squashed change if your change adds it.

tex3d (changes requested):
I believe there is an erroneous merge conflict resolution which adds code to CGHLSLMS.cpp when it shouldn't.  Other than that, I think this change looks good.

tex3d (approved):
LGTM!

PR #5590, opened 2023-08-23
Add `-print-before` and `-print-before-all` flags

bogner:
These match `-print-after` and `-print-after-all`, and are needed if we want the opt pipeline view to work in compiler explorer.

### DIFF
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -307,6 +307,10 @@ def encoding : Separate<["-", "/"], "encoding">, Group<hlslcomp_Group>, Flags<[C
   HelpText<"Set default encoding for source inputs and text outputs (utf8|utf16(win)|utf32(*nix)|wide) default=utf8">;
 def validator_version : Separate<["-", "/"], "validator-version">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Override validator version for module.  Format: <major.minor> ; Default: DXIL.dll version or current internal version.">;
+def print_before_all : Flag<["-", "/"], "print-before-all">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
+  HelpText<"Print LLVM IR before each pass.">;
+def print_before : Separate<["-", "/"], "print-before">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
+  HelpText<"Print LLVM IR before a specific pass.">;
 def print_after_all : Flag<["-", "/"], "print-after-all">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Print LLVM IR after each pass.">;
 def print_after : Separate<["-", "/"], "print-after">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -310,11 +310,11 @@ def validator_version : Separate<["-", "/"], "validator-version">, Group<hlslcom
 def print_before_all : Flag<["-", "/"], "print-before-all">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Print LLVM IR before each pass.">;
 def print_before : Separate<["-", "/"], "print-before">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
-  HelpText<"Print LLVM IR before a specific pass.">;
+  HelpText<"Print LLVM IR before a specific pass. May be specificied multiple times.">;
 def print_after_all : Flag<["-", "/"], "print-after-all">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Print LLVM IR after each pass.">;
 def print_after : Separate<["-", "/"], "print-after">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
-  HelpText<"Print LLVM IR after a specific pass.">;
+  HelpText<"Print LLVM IR after a specific pass. May be specificied multiple times.">;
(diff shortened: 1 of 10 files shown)


bogner:
@microsoft-github-policy-service agree company="Microsoft"

lib/IR/LegacyPassManager.cpp
```diff
@@ -685,6 +685,24 @@ void PMTopLevelManager::schedulePass(Pass *P) {
     PP->assignPassManager(activeStack, getTopLevelPassManagerType());
   }
 
+  // HLSL Change - begin
+  if (PI && !PI->isAnalysis() && (this->HLSLPrintBeforeAll || (this->HLSLPrintBefore.size() && this->HLSLPrintBefore.count(PI->getPassArgument())))) {
+    class direct_stderr_stream : public raw_ostream {
```
tex3d, line 690: Can `direct_stderr_stream` be shared now between this and print after, rather than copied?

bogner: Yes it can. Fixed in the latest update (where I also clang-formatted this and the PrintAfter block it was copied from)

tex3d (approved):
LGTM!

include/dxc/Support/HLSLOptions.td
```diff
@@ -307,6 +307,10 @@ def encoding : Separate<["-", "/"], "encoding">, Group<hlslcomp_Group>, Flags<[C
   HelpText<"Set default encoding for source inputs and text outputs (utf8|utf16(win)|utf32(*nix)|wide) default=utf8">;
 def validator_version : Separate<["-", "/"], "validator-version">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
   HelpText<"Override validator version for module.  Format: <major.minor> ; Default: DXIL.dll version or current internal version.">;
+def print_before_all : Flag<["-", "/"], "print-before-all">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
+  HelpText<"Print LLVM IR before each pass.">;
+def print_before : Separate<["-", "/"], "print-before">, Group<hlslcomp_Group>, Flags<[CoreOption, HelpHidden]>,
+  HelpText<"Print LLVM IR before a specific pass.">;
```
pow2clk, line 313: I don't know if there's a succinct way to indicate here that multple passes can be specified with multiple flags, but I don't get that information from this description. Granted, it's vague in the print-after counterpart. Maybe just a parenthetical (multiples allowed)?

lib/IR/LegacyPassManager.cpp
```diff
@@ -679,12 +679,37 @@ void PMTopLevelManager::schedulePass(Pass *P) {
...
     Pass *PP = P->createPrinterPass(dbgs(), std::string("*** IR Dump Before ") +
                                          P->getPassName().str() + " ***");
     PP->assignPassManager(activeStack, getTopLevelPassManagerType());
   }
 
+  // HLSL Change - begin
+  if (PI && !PI->isAnalysis() &&
+      (this->HLSLPrintBeforeAll ||
+       (this->HLSLPrintBefore.size() &&
+        this->HLSLPrintBefore.count(PI->getPassArgument())))) {
```
pow2clk, line 704: It's a nit, but this conditional and the one below that it was modeled after are getting pretty hairy. The previous conditional also checks for `PI && !PI->isAnalysis()`. If we could couch both of these within two corresponding `if(PI && !PI->isAnalysis()` blocks, there would be that much more joy in my life 😄

tools/clang/test/DXC/print-passes.hlsl
```diff
@@ -0,0 +1,16 @@
+// RUN: %dxc -E main -T vs_6_0 %s -print-before-all 2>&1 | FileCheck -check-prefix=BEFORE1 --check-prefix=BEFORE2 %s
+// RUN: %dxc -E main -T vs_6_0 %s -print-after-all 2>&1 | FileCheck -check-prefix=AFTER1 --check-prefix=AFTER2 %s
+// RUN: %dxc -E main -T vs_6_0 %s -print-before-all -print-after-all 2>&1 | FileCheck -check-prefix=BEFORE1 -check-prefix=BEFORE2 -check-prefix=AFTER1 --check-prefix=AFTER2 %s
+// RUN: %dxc -E main -T vs_6_0 %s -print-before verify 2>&1 | FileCheck -check-prefix=BEFORE1 %s
+// RUN: %dxc -E main -T vs_6_0 %s -print-after hlsl-dxilemit 2>&1 | FileCheck -check-prefix=AFTER2 %s
```
pow2clk, line 5: Can we have a check for multiple `-print-after`s?

pow2clk (commented):
I have some notes. Minor ones, but notes just the same

tools/clang/test/DXC/print-passes.hlsl
pow2clk: and `print-before`s . . .

bogner:
> I am on the record disfavoring force pushes after a review has come in. I would have liked to see the form the code had when Tex commented and how it was addressed. You can edit the final commit message so multiple commits needn't make it ugly.

Is the "compare" button that github gives you not sufficient?
screenshot showing compare link

lib/IR/LegacyPassManager.cpp
bogner: Given that the other conditional is really only there as a way to avoid conflicts with upstream (ShouldPrintXPass is stubbed out elsewhere) I'm not sure that it makes sense to keep it if we're going to modify it. That said, it's probably not really doing it's job so maybe it's best to just delete it either way.

include/dxc/Support/HLSLOptions.td
bogner: Added "May be specificied multiple times."

lib/IR/LegacyPassManager.cpp
bogner: Refactored a bit and used a helper function to simplify this

tools/clang/test/DXC/print-passes.hlsl
bogner: done

pow2clk (approved):
Thanks!

PR #5592, opened 2023-08-23
WArgV is needed only on non-Windows platforms (#5576)

hekota:
Fixes internal build break.

(cherry picked from commit f206b3d19d487d974e16d7bb41fa0cdcf79b4836)


pow2clk (approved):
Just as good as before!

tex3d (approved):

PR #5593, opened 2023-08-23
[NFC] Update DXC sources for clang-format

llvm-beanz:
One more batch of NFC changes to get DXC ready to clang-format. This change disables clang-format of some include blocks and breaks other include blocks into groups. With this change all of the sources in DXC can be clang-formatted and will continue to build on all supported platforms.


include/dxc/DxilContainer/DxilRuntimeReflection.h
```diff
@@ -12,7 +12,7 @@
 #pragma once
 
 #include "dxc/DXIL/DxilConstants.h"
-#include "dxc/WinAdapter.h"
+#include "dxc/Support/WinIncludes.h"
```
pow2clk, line 15: Looks like this was a way to allow the include to remain after DxiLConstants?

include/dxc/Support/HLSLOptions.h
```diff
@@ -17,10 +17,14 @@
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Option/ArgList.h"
-#include "dxc/dxcapi.h"
 #include "dxc/Support/HLSLVersion.h"
 #include "dxc/Support/SPIRVOptions.h"
 #include "dxc/Support/DxcOptToggles.h"
+
+#include "dxc/Support/WinIncludes.h"
```
pow2clk, line 24: There are a few places that add new includes and most common is the addition of this include before dxcapi and friends. Can you explain why that's needed?

pow2clk (commented):
I'm fairly ignorant of clang-format. Mostly asking questions for my own edification. I'm particularly confused by the additional includes. Is it because clang-format shuffles the order such that a dependency that was incidentally satisfied by what was an earlier include no longer is?

include/dxc/DxilContainer/DxilRuntimeReflection.h
llvm-beanz: The ordering here doesn't matter. DxilRuntimeReflection.h depends on the windows headers which are from WinIncludes not WinAdapter. We got lucky by include ordering previously.

include/dxc/Support/HLSLOptions.h
llvm-beanz: dxcapi depends on WinIncludes, and we got lucky that it compiled based on the ordering.

I didn't want to modify dxcapi to include WinIncludes directly since dxcapi is part of our public shipped API, and WinIncludes sets macros that alter the windows header definitions in ways that our users might not want.

llvm-beanz:
> I'm fairly ignorant of clang-format. Mostly asking questions for my own edification. I'm particularly confused by the additional includes. Is it because clang-format shuffles the order such that a dependency that was incidentally satisfied by what was an earlier include no longer is?

Yes. LLVM coding standards for include style are listed here:
https://llvm.org/docs/CodingStandards.html#include-style

The unwritten rule here is that in order for this to work, headers can't have include-order-dependent behavior. It is expected that any header that depends on functionality from another header includes that header explicitly.

This allows each source file to include only the minimal list of headers it requires (see: https://llvm.org/docs/CodingStandards.html#minimal-list-of-includes).

DXC's code (especially the Windows-specific bits) don't conform to this, which is why additional includes were required in some cases, and why I disabled clang-format on include blocks in other cases.

Clang-format only re-orders includes within a block of include statements, so in cases where it was easy to do so I grouped the includes to force the ordering requirement.

python3kgae (approved):

damyanp (approved):
Is there any interest in cleaning up the tech debt of the order-dependent includes we have?

I enjoy doing that sort of thing and could imagine myself wanting taking a stab at it if I get snowed in this winter perhaps...

pow2clk (approved):
Looks good!

PR #5594, opened 2023-08-23
[SPIR-V] Correctly use halfword-aligned byte address

### LINKED ISSUE #5475: [SPIR-V] Generation of 16-bit load from ByteAddressBuffer incorrect
**Description**
The [ByteAddressBuffer typed load/store functionality](https://github.com/microsoft/DirectXShaderCompiler/wiki/ByteAddressBuffer-Load-Store-Additions) wiki page states `The byteOffset should be aligned by size of the largest component that the type contains`. Does that mean loading 16-bit floats from a ByteAddressBuffer only requires an alignment of 2? The generated code looks good for DXIL but is incorrect for SPIR-V.

**Steps to Reproduce**
[Shader Playground](https://shader-playground.timjones.io/c04f1a1bae3f0c71bd5a81aca1233cdd), note how we pass the `-enable-16bit-types` flag.
```hlsl
RWStructuredBuffer<float> DstBuf;
ByteAddressBuffer SrcBuf;

struct HalfStruct
{
    half a;
    half b;
};

[numthreads(8, 8, 1)]
void CSMain()
{
    // This doesn't work with SPIR-V, will load the same values in both instsructions
    DstBuf[0] = SrcBuf.Load<half>(0u);
    DstBuf[1] = SrcBuf.Load<half>(2u);
    
    // This works in SPIR-V generation
    HalfStruct hs = SrcBuf.Load<HalfStruct>(4u);
    DstBuf[2] = hs.a;
    DstBuf[3] = hs.b;
}
```

 I would expect this to correctly load the halfs from the respective offsets, given the documentation mentioned above

**Actual Behavior**
The generated SPIR-V for the first two lines in `CSMain` looks like this:
```spir-v
         %26 = OpAccessChain %_ptr_StorageBuffer_uint %SrcBuf %uint_0 %uint_0
         %27 = OpLoad %uint %26
         %28 = OpUConvert %ushort %27
         %29 = OpBitcast %half %28
         %30 = OpFConvert %float %29
         %31 = OpAccessChain %_ptr_StorageBuffer_float %DstBuf %int_0 %uint_0
               OpStore %31 %30
         %32 = OpLoad %uint %26
         %33 = OpUConvert %ushort %32
         %34 = OpBitcast %half %33
         %35 = OpFConvert %float %34
         %36 = OpAccessChain %_ptr_StorageBuffer_float %DstBuf %int_0 %uint_1
               OpStore %36 %35
```
In both cases, the same value is loaded, looks like the generation just assumes all loads are 4-byte aligned. The second case (loading a struct with halfs) looks correct (includes a `OpShiftRightLogical` of the loaded uint). The DXIL output also looks correct.

**Environment**
- dxcompiler.dll: 1.7 - 1.7.0.3939 (5e080a772)

cassiebeckley:
All loads and stores for ByteAddressBuffer were being aligned to the nearest word. This is incorrect behavior for 16-bit types. In order to implement this, I had to move some of the logic for calculating offsets from compile time to runtime. For loads and stores of 32 or 64-bit types with a literal byte address, the optimizer will be able to reduce this down to nearly the same code as was being output before.

For 16-bit types, stores are slightly more convoluted (we need to load the old value, shift and mask it, and then combine it with the new value, in order to avoid clobbering any existing value). The optimizer isn't currently able to optimize away unnecessary shifts and masks, but my intuition is that it should be possible, in case we want to optimize this more.

Fixes #5475


tools/clang/lib/SPIRV/RawBufferMethods.h
```diff
@@ -58,65 +53,38 @@ class RawBufferHandler {
   ///                 --> Store to the second 16-bit uint starting at address 0.
   void processTemplatedStoreToBuffer(SpirvInstruction *value,
```
sudonatalie, line 54: Comment above needs updating

tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.templated-load.scalar.hlsl
```diff
@@ -6,22 +6,27 @@ ByteAddressBuffer buf;
...
 
+  // CHECK:   [[byte:%\d+]] = OpUMod %uint {{%\d+}} %uint_4
+  // CHECK:   [[bits:%\d+]] = OpIMul %uint [[byte]] %uint_8
   // CHECK:    [[ptr:%\d+]] = OpAccessChain %_ptr_Uniform_uint %buf %uint_0 {{%\d+}}
   // CHECK:   [[uint:%\d+]] = OpLoad %uint [[ptr]]
-  // CHECK: [[ushort:%\d+]] = OpUConvert %ushort [[uint]]
+  // CHECK:  [[shift:%\d+]] = OpShiftRightLogical %uint [[uint]] [[bits]]
+  // CHECK: [[ushort:%\d+]] = OpUConvert %ushort [[shift]]
   // CHECK:                   OpStore %u16 [[ushort]]
   uint16_t u16 = buf.Load<uint16_t>(tid.x);
```
sudonatalie, line 16: Can you add a test similar to the one in the bug, explicitly indexing into the buffer at byte offset 0 and 2? It would be helpful to have a test where we can CHECK that the actual numeric values in the shift and mask operations are what we expect. It's harder to see with these ones what was incorrect before and correct now.

tools/clang/test/CodeGenSPIRV/spv.use-legacy-buffer-matrix-order.hlsl
```diff
@@ -10,26 +10,32 @@ RWByteAddressBuffer buf2;
 [numthreads(64, 1, 1)]
 void main(uint3 tid : SV_DispatchThreadId)
 {
-// CHECK:    [[ptr:%\d+]] = OpAccessChain %_ptr_Uniform_uint %buf %uint_0 [[index_0:%\d+]]
```
sudonatalie, line 13: I do want us to look a little closer at what the optimized code would be here. I don't think we will want to sacrifice performance on 32 and 64 bit loads/stores to fix 16 bit, which might mean having an extra branch in our logic when we're word-aligned. My guess is that if you have enough stuff to load/store, then this:

```
i = byteaddress >> 2
load(i)
i++
load(i)
i++
load(i)
i++
load(i)
...
```

is a lot faster than this:

```
load(byteaddress >> 2)
byteaddress += 4
load(byteaddress >> 2)
byteaddress += 4
load(byteaddress >> 2)
byteaddress += 4
load(byteaddress >> 2)
byteaddress += 4
...
```

and I don't think the optimizer can turn the latter into the former. Even for literal addresses I wouldn't expect it too, but I haven't validated either yet.

sudonatalie (changes requested):

tools/clang/lib/SPIRV/RawBufferMethods.cpp
```diff
@@ -42,73 +42,113 @@ SpirvInstruction *RawBufferHandler::bitCastToNumericalOrBool(
...
+  auto *index = spvBuilder.createBinaryOp(spv::Op::OpShiftRightLogical,
+                                          astContext.UnsignedIntTy, byteAddress,
+                                          constUint2, loc, range);
+
+  // take the remainder and multiply by 8 to get the bit offset within the word
+  auto *bitOffset =
+      spvBuilder.createBinaryOp(spv::Op::OpUMod, astContext.UnsignedIntTy,
+                                byteAddress, constUint4, loc, range);
+  bitOffset =
+      spvBuilder.createBinaryOp(spv::Op::OpIMul, astContext.UnsignedIntTy,
```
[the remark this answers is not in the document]
cassiebeckley: Ah, great suggestion, thank you

tools/clang/lib/SPIRV/RawBufferMethods.cpp
```diff
@@ -42,73 +42,113 @@ SpirvInstruction *RawBufferHandler::bitCastToNumericalOrBool(
...
   auto *constUint0 =
       spvBuilder.getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, 0));
+  auto *constUint2 =
+      spvBuilder.getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, 2));
+  auto *constUint4 =
+      spvBuilder.getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, 4));
+  auto *constUint8 =
+      spvBuilder.getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, 8));
+
+  // divide the byte index by 4 (shift right by 2) to get the index in the
```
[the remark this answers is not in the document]
cassiebeckley: Done

tools/clang/lib/SPIRV/RawBufferMethods.cpp
```diff
@@ -419,93 +369,131 @@ SpirvInstruction *RawBufferHandler::processTemplatedLoadFromBuffer(
...
+  auto *index = spvBuilder.createBinaryOp(spv::Op::OpShiftRightLogical,
+                                          astContext.UnsignedIntTy, byteAddress,
+                                          constUint2, loc, range);
+
+  // take the remainder and multiply by 8 to get the bit offset within the word
+  auto *bitOffset =
+      spvBuilder.createBinaryOp(spv::Op::OpUMod, astContext.UnsignedIntTy,
+                                byteAddress, constUint4, loc, range);
+  bitOffset =
+      spvBuilder.createBinaryOp(spv::Op::OpIMul, astContext.UnsignedIntTy,
```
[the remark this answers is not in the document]
cassiebeckley: Done

tools/clang/lib/SPIRV/RawBufferMethods.cpp
```diff
@@ -419,93 +369,131 @@ SpirvInstruction *RawBufferHandler::processTemplatedLoadFromBuffer(
...
+      spvBuilder.getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, 2));
+  auto *constUint4 =
+      spvBuilder.getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, 4));
+  auto *constUint8 =
+      spvBuilder.getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, 8));
   auto *constUint16 =
       spvBuilder.getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, 16));
+  // 0xffff
+  auto *constUint65535 = spvBuilder.getConstantInt(astContext.UnsignedIntTy,
+                                                   llvm::APInt(32, 65535));
```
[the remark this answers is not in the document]
cassiebeckley: Yeah, that's a good point

tools/clang/lib/SPIRV/RawBufferMethods.cpp
```diff
@@ -534,74 +522,9 @@ void RawBufferHandler::store64BitsAtBitOffset0(SpirvInstruction *value,
   ptr = spvBuilder.createAccessChain(astContext.UnsignedIntTy, buffer,
                                      {constUint0, index}, loc, range);
   spvBuilder.createStore(ptr, msb, loc, range);
-  index = spvBuilder.createBinaryOp(spv::Op::OpIAdd, astContext.UnsignedIntTy,
-                                    index, constUint1, loc, range);
-}
-
-void RawBufferHandler::storeArrayOfScalars(
```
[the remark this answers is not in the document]
cassiebeckley: Thanks!

tools/clang/lib/SPIRV/RawBufferMethods.h
cassiebeckley: Done

tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.templated-load.scalar.hlsl
cassiebeckley: Sure, I've added it to the top of `method.byte-address-buffer.templated-load.scalar.hlsl`

tools/clang/test/CodeGenSPIRV/spv.use-legacy-buffer-matrix-order.hlsl
cassiebeckley: Okay, so, for literal addresses, for example in this code

```
ByteAddressBuffer buf;

float4 MainPS(float4 pos: SV_Position): SV_Target {
    uint arr[10] = buf.Load<uint[10]>(0);
    return float4(arr[pos.y], 0, 0, 0);
}
```

the code emitted by the optimizer is identical for both before and after this PR. However, you are correct that if you instead `buf.Load<uint[10]>(pos.x)`, the optimizer is unable to remove the shift right instruction.

I've thought of an idea to fix this for word-aligned types, basically just keeping track of whether the latest load/store was from/to a word-aligned type or not, and then only converting from byte address to index and back if it's required. I'll test it out and see how well it works.

cassiebeckley: Okay, check out `BufferAddress` to see my idea for how to solve this. The optimized code for arrays, vectors, and matrices of 32 and 64-bit types is now identical to what it was before (as you can see in this test, for example).

Keenuts (approved):

sudonatalie (approved):
Great work on this! Just a couple nits, but LGTM!

tools/clang/lib/SPIRV/RawBufferMethods.h
```diff
@@ -54,18 +54,44 @@ class RawBufferHandler {
...
 private:
-  SpirvInstruction *load16Bits(SpirvInstruction *buffer,
-                               SpirvInstruction *&byteAddress,
+  class BufferAddress {
+  public:
+    BufferAddress(SpirvInstruction *&byteAddress, SpirvEmitter &emitter)
+        : byteAddress(byteAddress), wordIndex(byteAddress),
+          wordIndexOutdated(true), spvBuilder(emitter.getSpirvBuilder()),
+          astContext(emitter.getASTContext()) {}
+    SpirvInstruction *getByte();
```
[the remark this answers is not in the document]
cassiebeckley: Done

PR #5596, opened 2023-08-24
[BUILD] add clang-cl to CMakeSettings

python3kgae:
This is for enable clang-cl build.
The change will fix build error when use clang_cl_x64_x64 with CMakeSettings.

Still many warnings and other issue to fix.


llvm-beanz (approved):

projects/dxilconv/lib/DxbcConverter/DxbcConverter.cpp
```diff
@@ -7083,7 +7090,7 @@ CompType DxbcConverter::InferOperandType(const D3D10ShaderBinary::CInstruction &
       }
     }
 
-    default: LLVM_FALLTHROUGH;
+    default: break;
```
pow2clk, line 7093: I suppose clang-format will take care of this when the time comes, but it is a nitpicky inconsistency that this is on the same line where the others all created a newline.

pow2clk (approved):
A nit if you want to address it

projects/dxilconv/lib/DxbcConverter/DxbcConverter.cpp
python3kgae: Fixed.

PR #5598, opened 2023-08-24
Fix asserts when compiling node objects with -Od

### LINKED ISSUE #5597: Assert/internal compiler error when compiling various node shaders with -Od
**Description**
The compiler asserts on node shaders using output record objects or NodeOutputArrays when using `-Od`.

**Steps to Reproduce**
```hlsl
// dxc -T lib_6_8
[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeDispatchGrid(1, 1, 1)]
[NumThreads(128, 1, 1)]
void node_1_1(
    [NodeArraySize(128)] [MaxRecords(64)] NodeOutputArray<RECORD1> OutputArray
)
{
    ThreadNodeOutputRecords<RECORD1> outRec = OutputArray[1].GetThreadNodeOutputRecords(2);
    outRec.OutputComplete();
}
```

**Actual Behavior**
Internal compiler error/assert due to mismatched type for `replaceAllUsesWith`.

**Environment**
- DXC version: v1.8.2306-preview

tex3d:
IsHLSLNodeIOType is supposed to capture all node object types, but was missing output record and output array types.

TranslateHLCastHandleToRes had LLVM_FALLTHROUGH in for two cases where it made no sense to fall through to the next case.

MergeGepUse is needed after lowering GetRecordPtr to collapse GEP chains that could result in intermediate multi-dimension array pointers from GEPs, which is considered an invalid DXIL type.

Updated some tests for -Od compilations.  Added a test for the multi-dim GEP case.

Some CHECKs needed specialization for -Od:
- align 4 added to load/store only with optimizations enabled
- no CSE of index/annotate node handle
- also, use CHECK-LABEL to make failure investigations easier

Fixes #5597

### DIFF
--- a/lib/HLSL/HLLowerUDT.cpp
+++ b/lib/HLSL/HLLowerUDT.cpp
@@ -274,6 +274,7 @@ void hlsl::ReplaceUsesForLoweredUDT(Value *V, Value *NewV) {
       SmallVector<Value*, 4> idxList(GEP->idx_begin(), GEP->idx_end());
       Value *NewGEP = Builder.CreateGEP(NewV, idxList);
       ReplaceUsesForLoweredUDT(GEP, NewGEP);
+      dxilutil::MergeGepUse(NewGEP);
       GEP->eraseFromParent();
 
     } else if (GEPOperator *GEP = dyn_cast<GEPOperator>(user)) {
--- a/lib/DXIL/DxilUtil.cpp
+++ b/lib/DXIL/DxilUtil.cpp
@@ -677,10 +677,6 @@ bool IsHLSLNodeInputRecordType(llvm::Type *Ty) {
   return false;
 }
 
-bool IsHLSLNodeIOType(llvm::Type* Ty) {
-  return IsHLSLNodeInputRecordType(Ty) || IsHLSLNodeOutputType(Ty);
-}
-
 bool IsHLSLNodeEmptyInputRecordType(llvm::Type* Ty) {
   if (llvm::StructType* ST = dyn_cast<llvm::StructType>(Ty)) {
     if (!ST->hasName())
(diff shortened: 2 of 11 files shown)


lib/HLSL/HLLowerUDT.cpp
```diff
@@ -274,6 +274,7 @@ void hlsl::ReplaceUsesForLoweredUDT(Value *V, Value *NewV) {
       SmallVector<Value*, 4> idxList(GEP->idx_begin(), GEP->idx_end());
       Value *NewGEP = Builder.CreateGEP(NewV, idxList);
       ReplaceUsesForLoweredUDT(GEP, NewGEP);
+      dxilutil::MergeGepUse(NewGEP);
```
dmpots, line 277: Do we need to call `MergeGepUse` for the `GEPOperator` below as well? Seems like it would have the same problem (and points to a mising test case?).

dmpots (approved):

lib/HLSL/HLLowerUDT.cpp
tex3d: That case handles constant GEP.  Wouldn't constant GEP already be in a merged state?

tex3d: The purpose here was to avoid DXIL that had a GEP instruction returning an intermediate type which was illegal on its own in DXIL (DxilValidation).  Multi-dim array is legal within a struct, but not outside the struct, so if multiple GEPs are used to drill into a struct, an intermediate type could appear illegal to DxilValidation.  These have to be GEP instructions, otherwise the GEP is a constant expression passed to a load or store instruction, and DxilValidation only cares about the type loaded or stored, which should be scalar from SROA.

dmpots: Constant GEPs can have multiple levels of gep as well, its not necessarily a single level gep.

tex3d: Ok, so I would be able to add the MergeGepUse to the constant case as well, but I don't think it would cause any failure to have a multi-level constant GEP, and I'm not confident I could create a test case for it.  For this node record case, it'll never be a constant.  It might require a carefully crafted AS or MS test case to trigger it, using a groupshared mesh shader payload.

In any case, I don't think multi-level constant GEP will cause any problem for DxilValidation, so the test would only be making sure the output IR looks a particular way.  I don't think it would be fixing an actual bug.

Should I try to do it anyway?

adam-yang (approved):

PR #5600, opened 2023-08-24
[lit] Add hash stability test for lit.

python3kgae:
A new lit format, DxcHashTest, has been added. This test will scan all HLSL files under tools\clang\test and run a hash stability test on each file. The hash stability test will compile the shader twice, once with the -Zi flag and once without. The hashes of the output containers will then be compared.

The --param hash_stability_path=test_path flag can be used to debug a single file or folder.

The hash stability test is disabled by default for pipeline runs, due to time constraints. To enable it, set the skip_hash_stability flag to False.

### DIFF
--- a/tools/clang/test/CMakeLists.txt
+++ b/tools/clang/test/CMakeLists.txt
@@ -15,6 +15,11 @@ configure_lit_site_cfg(
   )
 
 configure_lit_site_cfg(
+  ${CMAKE_CURRENT_SOURCE_DIR}/hash_stability/lit.site.cfg.in
+  ${CMAKE_CURRENT_BINARY_DIR}/hash_stability/lit.site.cfg
+  )
+
+configure_lit_site_cfg(
   ${CMAKE_CURRENT_SOURCE_DIR}/Unit/lit.site.cfg.in
   ${CMAKE_CURRENT_BINARY_DIR}/Unit/lit.site.cfg
   )
@@ -90,6 +95,7 @@ add_lit_testsuite(check-clang "Running the Clang regression tests"
   #LIT ${LLVM_LIT}
   PARAMS ${CLANG_TEST_PARAMS}
     skip_taef_exec=True
+    skip_hash_stability=True
   DEPENDS ${CLANG_TEST_DEPS}
   ARGS ${CLANG_TEST_EXTRA_ARGS}
   )
@@ -108,6 +114,7 @@ set(CLANG_TEST_PARAMS
   clang_taef_site_config=${CMAKE_CURRENT_BINARY_DIR}/taef/lit.site.cfg
   no_priority=True
   clang_taef_exec_site_config=${CMAKE_CURRENT_BINARY_DIR}/taef_exec/lit.site.cfg
+  clang_hash_stability_site_config=${CMAKE_CURRENT_BINARY_DIR}/hash_stability/lit.site.cfg
   )
 
 add_lit_testsuites(CLANG ${CMAKE_CURRENT_SOURCE_DIR}
@@ -146,6 +153,13 @@ if (WIN32)
             PARAMS ${CLANG_TEST_PARAMS}
                 adapter=${TAEF_EXEC_ADAPTER}
             DEPENDS ExecHLSLTests dxexp
+            ARGS ${CLANG_TEST_EXTRA_ARGS}
+          )
+
+  add_lit_target("check-clang-hash-stability" "Running lit suite dxc hash stability test"
+            ${CMAKE_CURRENT_SOURCE_DIR}/hash_stability
+            PARAMS ${CLANG_TEST_PARAMS}
+            DEPENDS ${CLANG_TEST_DEPS}
             ARGS ${CLANG_TEST_EXTRA_ARGS}
           )
 endif()
--- a/tools/clang/test/hash_stability/lit.cfg
+++ b/tools/clang/test/hash_stability/lit.cfg
@@ -1,0 +1,47 @@
+# -*- Python -*-
+
+# Configuration file for the 'lit' test runner.
+
+import os
+import platform
+import subprocess
+
+import lit.formats
+import lit.util
+
+# name: The name of this test suite.
+config.name = 'hash-stability'
+
+skip_hash_stability = lit_config.params.get('skip_hash_stability', None)
+if skip_hash_stability or platform.system() != 'Windows':
+  config.unsupported = True
+
+if config.unsupported == False:
+	llvm_tools_dir = getattr(config, 'llvm_tools_dir', None)
+	if not llvm_tools_dir:
+		# Check for 'clang_hash_stability_site_config' user parameter, and use that if available.
+		site_cfg = lit_config.params.get('clang_hash_stability_site_config', None)
+		if site_cfg and os.path.exists(site_cfg):
+			lit_config.load_config(config, site_cfg)
+			raise SystemExit
+		lit_config.fatal('No LLVM tools dir set!')
+
+	dxc_path = lit.util.which('dxc', llvm_tools_dir)
+	dxa_path = lit.util.which('dxa', llvm_tools_dir)
+	working_dir = config.working_dir
+
+	test_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..')
+	tmp_path = os.path.join(working_dir, 'Output')
+	print('tmp_path: ' + tmp_path)
+	# create tmp_path if it doesn't exist
+	if not os.path.exists(tmp_path):
+		try:
+			os.makedirs(tmp_path)
+		except OSError:
+			print("Creation of the directory %s for temp output failed" % tmp_path)
+			raise SystemExit
+	hash_stability_path = lit_config.params.get('hash_stability_path', None)
+	if hash_stability_path:
+		test_path = hash_stability_path
+
+	config.test_format = lit.formats.DxcHashTest(test_path, dxc_path, dxa_path, working_dir)
--- a/utils/lit/lit/formats/dxc_hash.py
+++ b/utils/lit/lit/formats/dxc_hash.py
@@ -1,0 +1,234 @@
+from __future__ import absolute_import
+import os
+import subprocess
+import platform
+import filecmp
+from glob import iglob
+
+import lit.Test
+import lit.TestRunner
+import lit.util
+import lit.ShUtil as ShUtil
+from .base import TestFormat
+
+kIsWindows = platform.system() == 'Windows'
+
+# Don't use close_fds on Windows.
+kUseCloseFDs = not kIsWindows
+
+class ShellEnvironment(object):
+
+    """Mutable shell environment containing things like CWD and env vars.
+
+    Environment variables are not implemented, but cwd tracking is.
+    """
+
+    def __init__(self, cwd, env):
+        self.cwd = cwd
+        self.env = env
+
+class DxcHashTest(TestFormat):
+    def __init__(self, test_path, dxc_path, dxa_path, working_dir):
+        self.test_path = test_path
+        self.dxc_path = dxc_path
+        self.dxa_path = dxa_path
+        self.cwd = working_dir
+
+    def getTestsInDirectory(self, testSuite, path_in_suite,
+                            litConfig, localConfig):
+        if not os.path.isdir(self.test_path):
+            hlsl_list = [self.test_path]
+        else:
+            # Scan all sub directories in self.test_path
+            rootdir = self.test_path.replace('\\','/')
+            rootdir_glob = f"{rootdir}/**/*.hlsl"
+
+            # This will return absolute paths
+            hlsl_list = [f for f in iglob(rootdir_glob, recursive=True) if os.path.isfile(f)]
+
+        # add hlsl files which has first line as %dxc test as tests.
+        for filename in hlsl_list:
+            with open(filename, 'r') as f:
+                first_line = f.readline()
+                if (first_line.find("%dxc") == -1 or first_line.find("RUN:") == -1 or
+                    first_line.find(" not ") != -1 or
+                    first_line.find(" -M ") != -1 or
+                    first_line.find(" -H ") != -1 or
+                    first_line.find(" -Fi ") != -1 or
+                    first_line.find(" -exports ") != -1 or
+                    first_line.find(" -ast-dump-implicit ") != -1 or
+                    first_line.find(" -fcgl ") != -1 or
+                    first_line.find("rootsig_1_") != -1 or
+                    first_line.find(" -verify ") != -1 or
+                    # must have a target profile
+                    first_line.find(" -T") == -1 or
+                    # skip spirv.
+                    first_line.find("spirv") != -1):
+                    continue
+            rel_dir = os.path.relpath(filename, testSuite.source_root)
+
+            yield lit.Test.Test(testSuite, [rel_dir], localConfig, filename)
+
+    def get_run_line(self, test):
+        # Read first line of the test file
+        f = open(test.getFilePath(), 'r')
+        first_line = f.readline().rstrip()
+        f.close()
+        # remove things before RUN: from first line
+        first_line = first_line[first_line.find("RUN:") + 4:]
+
+        sourcepath = test.getFilePath()
+        sourcedir = os.path.dirname(sourcepath)
+        execpath = test.getExecPath()
+        execdir,execbase = os.path.split(execpath)
+        tmpDir = os.path.join(self.cwd, 'Output')
+        tmpBase = os.path.join(tmpDir, execbase)
+        tmpFile = tmpBase + '.tmp'
+        # subsitute %s with sourcepath
+        first_line = first_line.replace("%s", sourcepath)
+        # subsitute %t with tmpDir
+        first_line = first_line.replace("%t", tmpFile)
+        # subsitute %S with sourcedir
+        first_line = first_line.replace("%S", sourcedir)
+        # subsitute %T with execdir
+        first_line = first_line.replace("%T", tmpDir)
+
+        return first_line
+
+    def extract_hash(self, dxa_path, dx_container):
+        # extract hash from dxa_path
+        hash_file = f"{dx_container}.hash"
+        args = [dxa_path, "-extractpart", "HASH", dx_container, "-o", hash_file]
+        stdin, stdout, stderr = 0, subprocess.PIPE, subprocess.PIPE
+
+        proc = subprocess.Popen(args, cwd=self.cwd,
+                                                executable = dxa_path,
+                                        stdin = stdin,
+                                        stdout = stdout,
+                                        stderr = stderr,
+                                                close_fds = kUseCloseFDs)
+        proc.communicate()
+        res = proc.returncode
+        if res != 0:
+            print(f"extract hash failed {args}")
+            # extract hash failed, return fail.
+            return None
+        return hash_file
+
+    def normal_compile(self, args, shenv, test_name, output_file):
+        normal_args = args
+        normal_args.append("-Qstrip_reflect")
+        normal_args.append("-Zsb")
+        normal_args.append("-Fo")
+        normal_args.append(output_file)
+        proc = subprocess.Popen(normal_args, cwd=shenv.cwd,
+                                # don't writ output to stdout
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE,
+                                env=shenv.env,
+                                close_fds=kUseCloseFDs)
+        proc.communicate()
+        return proc.returncode
+
+    def debug_compile(self, args, shenv, test_name, output_file):
+        debug_args = args
+        debug_args.append("-Zi")
+        debug_args.append("-Qstrip_reflect")
+        debug_args.append("-Zsb")
+        debug_args.append("-Fo")
+        debug_args.append(output_file)
+
+        proc = subprocess.Popen(debug_args, cwd=shenv.cwd,
+                                # don't writ output to stdout
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE,
+                                env=shenv.env,
+                                close_fds=kUseCloseFDs)
+        proc.communicate()
+        return proc.returncode
+
+    def execute(self, test, litConfig):
+        first_line = self.get_run_line(test)
+        cwd = self.cwd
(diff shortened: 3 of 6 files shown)


tools/clang/test/CMakeLists.txt
```diff
@@ -90,6 +95,7 @@ add_lit_testsuite(check-clang "Running the Clang regression tests"
   #LIT ${LLVM_LIT}
   PARAMS ${CLANG_TEST_PARAMS}
     skip_taef_exec=True
+    skip_hash_stability=True
```
llvm-beanz, line 98: Hash stability should be enabled by default, and we shouldn't have options to skip it:

https://github.com/microsoft/DirectXShaderCompiler/pull/5548

utils/lit/lit/formats/dxc_hash.py
```diff
@@ -0,0 +1,234 @@
...
+    """Mutable shell environment containing things like CWD and env vars.
+
+    Environment variables are not implemented, but cwd tracking is.
+    """
+
+    def __init__(self, cwd, env):
+        self.cwd = cwd
+        self.env = env
+
+class DxcHashTest(TestFormat):
```
llvm-beanz, line 30: Can we make the hash stability logic a standalone python script that we could run individually rather than buyring it inside a lit test format?

That would make it a lot easier to reproduce any individual hash stability test case failure.

python3kgae: Sure. Where should the new hash_stability.py live?

llvm-beanz: We've put scripts like this in utils/hct in the past. We've also used `CamelCase` naming for them.

tools/clang/test/CMakeLists.txt
```diff
@@ -148,5 +154,12 @@ if (WIN32)
             DEPENDS ExecHLSLTests dxexp
             ARGS ${CLANG_TEST_EXTRA_ARGS}
           )
+
+  add_lit_target("check-clang-hash-stability" "Running lit suite dxc hash stability test"
```
llvm-beanz, line 158: We shouldn't need to add a target here explicitly, the `add_lit_testsuites` call above adds targets for each subdirectory.

tools/clang/test/CMakeLists.txt
```diff
@@ -14,6 +14,11 @@ configure_lit_site_cfg(
   ${CMAKE_CURRENT_BINARY_DIR}/lit.site.cfg
   )
 
+configure_lit_site_cfg(
+  ${CMAKE_CURRENT_SOURCE_DIR}/hash_stability/lit.site.cfg.in
+  ${CMAKE_CURRENT_BINARY_DIR}/hash_stability/lit.site.cfg
```
llvm-beanz, line 19: Is it necessary to explicitly pass this? Usually lit will find the local site config by the directory it is targeted at.

utils/lit/lit/formats/dxc_hash.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def get_run_line(self, test):
+        # Read first line of the test file
+        f = open(test.getFilePath(), 'r')
+        first_line = f.readline().rstrip()
+        f.close()
+        # remove things before RUN: from first line
+        first_line = first_line[first_line.find("RUN:") + 4:]
+
+        sourcepath = test.getFilePath()
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/test/hash_stability/lit.cfg
```diff
@@ -0,0 +1,47 @@
...
+import subprocess
+
+import lit.formats
+import lit.util
+
+# name: The name of this test suite.
+config.name = 'hash-stability'
+
+skip_hash_stability = lit_config.params.get('skip_hash_stability', None)
+if skip_hash_stability or platform.system() != 'Windows':
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/test/CMakeLists.txt
python3kgae: Removed.

tools/clang/test/CMakeLists.txt
python3kgae: Cannot make it work :(
Would be nice to remove it.

utils/lit/lit/formats/dxc_hash.py
```diff
@@ -0,0 +1,234 @@
...
+import filecmp
+from glob import iglob
+
+import lit.Test
+import lit.TestRunner
+import lit.util
+import lit.ShUtil as ShUtil
+from .base import TestFormat
+
+kIsWindows = platform.system() == 'Windows'
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/test/CMakeLists.txt
python3kgae: Fixed.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+    parser = argparse.ArgumentParser(description='Description of your program')
+    parser.add_argument('-a','--argument', help='origin command line options to run dxc like \"dxc -ECSMain -Tcs_6_0 D:\\test.hlsl\"', required=True)
+    parser.add_argument('-p','--path', help='path to find dxc and dxa', required=True)
+    args = vars(parser.parse_args())
+    dxc_args = args['argument'].split(' ')
+    dxc_args[0] = "%dxc"
+    bin_dir = args['path']
+    dxc_path = os.path.join(bin_dir, 'dxc.exe')
+    dxa_path = os.path.join(bin_dir, 'dxa.exe')
+    working_dir = os.getcwd()
```
pow2clk, line 141: Aren't these getting passed in? I don't know why they are constructed twice

pow2clk (commented):
Looks good overall!

Sorry for my python-ignorant questions

tools/clang/test/hash_stability/lit.cfg
```diff
@@ -0,0 +1,42 @@
...
+# name: The name of this test suite.
+config.name = 'hash-stability'
+
+llvm_tools_dir = getattr(config, 'llvm_tools_dir', None)
+if not llvm_tools_dir:
+	# Check for 'clang_hash_stability_site_config' user parameter, and use that if available.
+	site_cfg = lit_config.params.get('clang_hash_stability_site_config', None)
+	if site_cfg and os.path.exists(site_cfg):
+		lit_config.load_config(config, site_cfg)
+		raise SystemExit
```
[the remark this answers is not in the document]
python3kgae: The SystemExit will be catched https://github.com/microsoft/DirectXShaderCompiler/blob/main/utils/lit/lit/TestingConfig.py#L157

When site config called second time, the llvm_tool_dir will exist, so the reload will not happen again.

tools/clang/test/hash_stability/lit.cfg
```diff
@@ -0,0 +1,42 @@
...
+	lit_config.fatal('No LLVM tools dir set!')
+
+dxc_path = lit.util.which('dxc', llvm_tools_dir)
+dxa_path = lit.util.which('dxa', llvm_tools_dir)
+working_dir = config.working_dir
+
+test_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..')
+tmp_path = os.path.join(working_dir, 'Output')
+print('tmp_path: ' + tmp_path)
```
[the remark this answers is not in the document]
python3kgae: just debugging code, removed.

tools/clang/test/hash_stability/lit.cfg
```diff
@@ -0,0 +1,42 @@
...
+# create tmp_path if it doesn't exist
+if not os.path.exists(tmp_path):
+	try:
+		os.makedirs(tmp_path)
+	except OSError:
+		print("Creation of the directory %s for temp output failed" % tmp_path)
+		raise SystemExit
```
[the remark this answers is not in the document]
python3kgae: DxcHashTest will calculate it again in run_hash_stability_test to avoid extra parameter required when running py HashStability.py from command line.

tools/clang/test/hash_stability/lit.cfg
```diff
@@ -0,0 +1,42 @@
...
+hash_stability_path = lit_config.params.get('hash_stability_path', None)
+if hash_stability_path:
+	test_path = hash_stability_path
+
+config.test_format = lit.formats.DxcHashTest(test_path, dxc_path, dxa_path, working_dir)
```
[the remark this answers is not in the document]
python3kgae: DxcHashTest will calculate it again in run_hash_stability_test to avoid extra parameter required when running py HashStability.py from command line.

tools/clang/test/CMakeLists.txt
python3kgae: This will also create a path for hash_stability to use as temp directory.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+    tmp_path = os.path.join(working_dir, 'Output')
+    print('tmp_path: ' + tmp_path)
```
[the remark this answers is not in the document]
python3kgae: Yes. removed.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+    print("""HashStability.py - Run hash stability test.
+
+Usage:
+HashStability.py <dxc_args> - args to run dxc like " "%dxc -ECSMain -Tcs_6_0 D:\test.hlsl"
+    <bin_dir>       - path to dxc.exe and dxa.exe
+""")
+    return 1
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Description of your program')
```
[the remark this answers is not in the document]
python3kgae: Fixed :(

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+    # run debug compilation
+    debug_out = os.path.join(working_dir, 'Output', test_name+'.dbg.out')
+    res = debug_compile(args, debug_out, working_dir, empty_env)
+    if res != 0:
+        # strip_reflect failed, return fail.
+        return False, "Adding Qstrip_reflect and Zi failed compilation."
+
+    debug_hash = extract_hash(dxa_path, debug_out, working_dir, empty_env)
+    if debug_hash is None:
+        return False, "Fail to get hash for normal compilation."
```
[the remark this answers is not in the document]
python3kgae: Fixed.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+    working_dir = os.getcwd()
+    tmp_path = os.path.join(working_dir, 'Output')
+    print('tmp_path: ' + tmp_path)
+    # create tmp_path if it doesn't exist
+    if not os.path.exists(tmp_path):
+        try:
+            os.makedirs(tmp_path)
+        except OSError:
+            print("Creation of the directory %s for temp output failed" % tmp_path)
+            exit(1)
```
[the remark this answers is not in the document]
python3kgae: Less parameter to make it easier to run through command line.

utils/hct/HashStability.py
python3kgae: Simplify the required parameter to make it easier to run thru command line.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+import shutil
+
+IS_WINDOWS = platform.system() == 'Windows'
+
+# Don't use close_fds on Windows.
+USE_CLOSE_FDS = not IS_WINDOWS
+
+
+def extract_hash(dxa_path, dx_container, working_dir, empty_env):
+    # extract hash from dxa_path
```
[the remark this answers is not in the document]
python3kgae: Fixed.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+    proc = subprocess.Popen(args, cwd=working_dir,
+                            env=empty_env,
+                            executable=dxa_path,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE,
+                            close_fds=USE_CLOSE_FDS)
+    proc.communicate()
+    res = proc.wait()
+    if res != 0:
+        print(f"extract hash failed {args}")
```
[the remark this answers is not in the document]
python3kgae: Done.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+
+def normal_compile(args, output_file, working_dir, empty_env):
+    normal_args = args
+    normal_args.append("-Qstrip_reflect")
+    normal_args.append("-Zsb")
+    normal_args.append("-Fo")
+    normal_args.append(output_file)
+    proc = subprocess.Popen(normal_args, cwd=working_dir,
+                            env=empty_env,
+                            # don't writ output to stdout
```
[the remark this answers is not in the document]
python3kgae: Done.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+                           # env=empty_env,
+                            # don't writ output to stdout
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.PIPE,
+                            close_fds=USE_CLOSE_FDS)
+    proc.communicate()
+    res = proc.wait()
+    if res != 0:
+        # original compilation failed, don't run hash test.
+        return True, "Original compilation failed."
```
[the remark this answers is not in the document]
python3kgae: Removed original compilation.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # run normal compile
+    normal_out = os.path.join(working_dir, 'Output', test_name+'.normal.out')
+    res = normal_compile(args, normal_out, working_dir, empty_env)
+    if res != 0:
+        # strip_reflect failed, return fail.
+        return False, "Adding Qstrip_reflect failed compilation."
```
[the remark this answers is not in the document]
python3kgae: Not sure, this is from the taef test.

tools/clang/test/CMakeLists.txt
llvm-beanz: If you only need it for the temp directory, there are other ways to do that, like CMake's `make_directory` command.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,142 @@
...
+import subprocess
+import platform
+import filecmp
+import argparse
+import shutil
+
+IS_WINDOWS = platform.system() == 'Windows'
+
+# Don't use close_fds on Windows.
+USE_CLOSE_FDS = not IS_WINDOWS
```
llvm-beanz, line 12: Seems like this could be simpler as: `close_fds =  platform.system() != 'Windows'`

`IS_WINDOWS` is only used in one place, so we shouldn't really need to make it a variable.

Also `USE_CLOSE_FDS` doesn't really make sense as a name, `close_fds` reads more cleanly and matches the code style elsewhere.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,142 @@
...
+        return False, "debug compilation failed."
+
+    debug_hash = extract_hash(dxa_path, debug_out, working_dir, empty_env)
+    if debug_hash is None:
+        return False, "Fail to get hash for debug compilation."
+
+    # compare normal_hash and debug_hash.
+    if filecmp.cmp(normal_hash, debug_hash):
+        # hash match, return pass.
+        return True, "Hash match."
```
llvm-beanz, line 106: nit: "Hash matches."

utils/hct/HashStability.py
```diff
@@ -0,0 +1,142 @@
...
+    else:
+        # hash mismatch
+        return False, "Hash mismatch."
```
llvm-beanz, line 109: nit: "Hash mismatches."

utils/hct/HashStability.py
```diff
@@ -0,0 +1,142 @@
...
+
+################################################
+################################################
+# For running from the command-line
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Run hash stability test')
+    parser.add_argument('-a','--argument', help='origin command line options to run dxc like \"dxc -ECSMain -Tcs_6_0 D:\\test.hlsl\"', required=True)
```
llvm-beanz, line 117: nit: `help='dxc command line options'`

utils/hct/HashStability.py
```diff
@@ -0,0 +1,142 @@
...
+    parser.add_argument('-p','--path', help='path to find dxc and dxa', required=True)
```
llvm-beanz, line 118: nit: `help='path to the directory containing dxc and dxa'`

utils/hct/HashStability.py
```diff
@@ -0,0 +1,142 @@
...
+    args = vars(parser.parse_args())
+    dxc_args = args['argument'].split(' ')
+    dxc_args[0] = "%dxc"
+    bin_dir = args['path']
+    # get dxc and dxa path when running from command line
+    dxc_path = os.path.join(bin_dir, 'dxc.exe')
```
llvm-beanz, line 124: What happens when you run this on Linux?

utils/hct/HashStability.py
```diff
@@ -0,0 +1,142 @@
+
```
llvm-beanz, line 1: Can we introduce some tests for this script?

tools/clang/test/CMakeLists.txt
python3kgae: Still cannot make it work :(
lit.cfg will fail to find llvm_tools_dir if not load site config.
Tried to load site config of parent folder, but cause other test fail.

utils/hct/HashStability.py
python3kgae: Fixed.

utils/hct/HashStability.py
python3kgae: Added one test as FileCheck test.

utils/hct/HashStability.py
python3kgae: Used is_windows for later add .exe to dxa name.
Updated to close_fds.

tools/clang/test/hash_stablity_script/hash_stability.test
```diff
@@ -0,0 +1 @@
+RUN: python3 %S/../../../../utils/hct/HashStability.py -a %S/../DXC/Inputs/smake.hlsl -p %llvm_tools_dir
```
llvm-beanz, line 1: A replacement for the hash stability script would be less gross. Also we could put this under `tools/clang/test/tools/HashStability` which at least kinda matches the patterns we have in LLVM's test directories.

tools/clang/test/hash_stability/lit.cfg
```diff
@@ -0,0 +1,42 @@
+# -*- Python -*-
```
llvm-beanz, line 1: nit: None of the test directories are `_` separated names so this should be `HashStability`.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,146 @@
...
+################################################
+################################################
+# For running from the command-line
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Run hash stability test')
+    parser.add_argument('-a','--argument', help='dxc command line options', required=True)
+    parser.add_argument('-p','--path', help='path to the directory containing dxc and dxa', required=True)
+    args = vars(parser.parse_args())
+    dxc_args = args['argument'].split(' ')
```
llvm-beanz, line 119: Relying on splitting args as a string is fragile. Instead the "argument" option could be positional and allow multiple values. That would reduce the risk of this causing problems.

utils/hct/HashStability.py
```diff
@@ -0,0 +1,144 @@
...
+
+################################################
+################################################
+# For running from the command-line
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Run hash stability test')
+    parser.add_argument('argument', nargs='+',
+                        help='command line options to run dxc')
+    parser.add_argument('-p','--path', help='path to the directory containing dxc and dxa', required=True)
```
llvm-beanz, line 115: I feel like the user interface here is still really awkward, but we can address that in a follow up change.

I think a better approach would be to have the `-p` option be optional and have it default to `os.path.dirname(os.path.abspath(__file__))`. Then we could add `#!/usr/bin/env python3` to the top of the script, mark it executable, and have a build step that copies it into the `bin` folder during the build process. That would make for a nice development experience.

llvm-beanz (approved):
One comment that can be a follow-up.

llvm-beanz (changes requested):
Wait... I think we need to revisit the LIT format.

utils/lit/lit/formats/dxc_hash.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class DxcHashTest(TestFormat):
+    def __init__(self, test_path, dxc_path, dxa_path, working_dir):
+        self.test_path = test_path
+        self.dxc_path = dxc_path
+        self.dxa_path = dxa_path
+        self.cwd = working_dir
+
+    def getTestsInDirectory(self, testSuite, path_in_suite,
```
llvm-beanz, line 32: This didn't work the way I thought it did, and I'm actually a bit concerned about it after our discussion in the morning meeting.

I assumed that this was traversing using the normal shell-test flow and respecting the lit local configs. It does not. I think that is wrong.

python3kgae: Updated.
Caught 2 hash mismatch
    tools\clang\test\HLSLFileCheck\hlsl\control_flow\return\lifetime-markers.hlsl
    and
    tools\clang\test\HLSLFileCheck\hlsl\objects\Cbuffer\retCBV3.hlsl

Disabled these tests for now, should fix in later PR.

llvm-beanz: The HLSLFileCheck directory has no test suffixes enabled. So those tests shouldn't be running if you're loading the local config.

python3kgae: It does fail.
If cannot find local config file, it will use local config from parent directory.
The root is using local config send from getTestsInDirectory.

python3kgae: HashStability has its own suffix check, it will ignore suffix in the local config.

PR #5602, opened 2023-08-24
[NFC] Clang-format DXC sources

### LINKED ISSUE #5591: clang-format DXC sources
**Description**
DXC sources are inconsistent in their style and formatting. We should clang-format the codebase to adopt consistent style.

This file lists all the files unique to DXC. We should clang-format all these files to adopt consistent style.

[dxc-source-files.txt](https://github.com/microsoft/DirectXShaderCompiler/files/12422000/dxc-source-files.txt)

**Environment**
- DXC version: `main`
- Host Operating System: all

llvm-beanz:
This applies clang-format with the LLVM style to the DXC codebase.

Fixes #5591


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -852,72 +880,45 @@ QualType GetOrCreateVectorSpecialization(ASTContext& context, Sema* sema,
...
+// replaced with clang::Type-based constructs.
 
 static const LPCSTR kBuiltinIntrinsicTableName = "op";
 
-static const ArTypeObjectKind g_ScalarTT[] =
-{
-  AR_TOBJ_SCALAR,
-  AR_TOBJ_UNKNOWN
-};
+static const ArTypeObjectKind g_ScalarTT[] = {AR_TOBJ_SCALAR, AR_TOBJ_UNKNOWN};
```
pow2clk, line 906: I feel like this is a loss of readability. If we had comments after each entry, would clang-format let us keep the multiple lines?

llvm-beanz: Yes, clang will keep it if there are trailing comments. I don't want to go through and make a bunch of hand edits here to make people happy about individual formatting changes.

This change is about adopting a policy and enforcing it.

tools/clang/unittests/HLSL/VerifierTest.cpp
```diff
@@ -122,17 +123,19 @@ class VerifierTest TEST_CLASS_DERIVATION {
...
+    char *commandLine;
 
     //
     // Very simple processing for now.
     // See utils\lit\lit\TestRunner.py for the more thorough implementation.
     //
     // The first line for HLSL tests will always look like this:
-    // // RUN: %clang_cc1 -fsyntax-only -Wno-unused-value -ffreestanding -verify %s
+    // // RUN: %clang_cc1 -fsyntax-only -Wno-unused-value -ffreestanding -verify
+    // %s
```
bogner, line 134: This scared me for a second, but it's just a comment

include/dxc/Support/Unicode.h
```diff
@@ -33,65 +33,80 @@ int WideCharToMultiByte(uint32_t CodePage, uint32_t dwFlags,
...
 // A ccp_char is a character encoded in the console code page.
 typedef char ccp_char;
 
-_Success_(return != false)
-bool UTF8ToConsoleString(_In_opt_count_(textLen) const char* text, _In_ size_t textLen, _Inout_ std::string* pValue, _Out_opt_ bool* lossy);
+_Success_(return != false) bool UTF8ToConsoleString(_In_opt_count_(textLen)
+                                                        const char *text,
+                                                    _In_ size_t textLen,
+                                                    _Inout_ std::string *pValue,
+                                                    _Out_opt_ bool *lossy);
```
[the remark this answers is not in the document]
bogner: I think this is the only one of my comments that's particularly important - the other ones look kind of bad but we can clean them up after the fact as people run into them.

bogner (approved):

include/dxc/Support/Unicode.h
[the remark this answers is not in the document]
llvm-beanz: @pow2clk, @tex3d, is there any reason not to just remove all the SAL annotations? They trip up clang format, but we also use them inconsistently enough that I wonder if they bring value.

I can maybe see keeping them on public APIs, but for internal APIs it seems like they don't add much value.

s-perron (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: You don't want to do it, but would you object if I did?

llvm-beanz: I have no objection to fixing formatting issues. You can do it before I merge the clang-format change or after. Either way is fine.

PR #5605, opened 2023-08-25
Fix crash in dominator tree construction used by DiagnoseRaytracingPayloadAccess

### LINKED ISSUE #5035: DXC crashing on while(true)-loop
# Summary
`DXC` crashes when compiling the following example with `-T lib_6_7`, apparently during a PAQ analysis that is not enabled with earlier shader models:

```HLSL
[shader("raygeneration")] void RayGen() {
   while (true) {
      break;
   }
}
```

# Fix
In `Support/GenericDomTreeConstruction.h:94`, it seems we should check for `Succ` being non-null, which can be caused by optimized-out successor nodes.

(I cannot contribute the fix myself due to the CLA).

jasilvanus:
Fix crash if a RayGen shader contains a `while(true)` loop, caused by dominator tree construction not ignoring `nullptr` successor AST nodes which occur in this case.

Also add regression test case.

Fixes #5035.

### DIFF
--- a/include/llvm/Support/GenericDomTreeConstruction.h
+++ b/include/llvm/Support/GenericDomTreeConstruction.h
@@ -92,6 +92,10 @@ unsigned DFSPass(DominatorTreeBase<typename GraphT::NodeType>& DT,
     // Visit the successor next, if it isn't already visited.
     typename GraphT::NodeType* Succ = *NextSucc;
 
+    // For clang, CFG successors can be optimized-out nullptrs. Skip those.
+    if (!Succ)
+      continue;
+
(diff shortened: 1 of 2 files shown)


jasilvanus:

jasilvanus:
Friendly ping for review.

bob80905 (approved):
LGTM

jasilvanus:
Friendly ping for review.

jasilvanus:
Friendly ping for a second review.

pow2clk (approved):
Simple enough

jasilvanus:
Thanks!

PR #5607, opened 2023-08-25
Don't reuse WaveGetLaneIndex result across thread repacking points

### LINKED ISSUE #5034: WaveGetLaneIndex() incorrectly optimized across thread repacking point
# Summary

In DXR, the results of certain wave intrinics are only valid until a potential thread repacking point (like `CallShader`) is reached.

However, `dxc` reuses the result of `@dx.op.waveGetLaneIndex` across a `CallShader` call.

# Example
Input HLSL:

```
struct MyParams
{
    int laneId;
};

RWTexture2D<float4> gOutput : register(u0);

[shader("raygeneration")]
export void RayGen()
{
    MyParams payload;
    payload.laneId = WaveGetLaneIndex();
    CallShader(0, payload);
    payload.laneId += WaveGetLaneIndex();
    gOutput[DispatchRaysIndex().xy] = float4(payload.laneId, 1.f, 1.f, 1.f);
}
```

Output DXIL compiled with `-T lib_6_6` (note how %4 is reused):

```
define void @"\01?RayGen@@YAXXZ"() #0 {
  %1 = load %dx.types.Handle, %dx.types.Handle* @"\01?gOutput@@3V?$RWTexture2D@V?$vector@M$03@@@@A", align 4
  %2 = alloca %struct.MyParams, align 4
  %3 = bitcast %struct.MyParams* %2 to i8*
  call void @llvm.lifetime.start(i64 4, i8* %3) #0
  %4 = call i32 @dx.op.waveGetLaneIndex(i32 111)  ; WaveGetLaneIndex()  <--- first WaveGetLaneIndex
  %5 = getelementptr inbounds %struct.MyParams, %struct.MyParams* %2, i32 0, i32 0
  store i32 %4, i32* %5, align 4, !tbaa !17
  call void @dx.op.callShader.struct.MyParams(i32 159, i32 0, %struct.MyParams* nonnull %2)  ; CallShader(ShaderIndex,Parameter)
  %6 = load i32, i32* %5, align 4, !tbaa !17
  %7 = add i32 %6, %4                                                   <--- reuse
  store i32 %7, i32* %5, align 4, !tbaa !17
  %8 = sitofp i32 %7 to float
  %9 = call i32 @dx.op.dispatchRaysIndex.i32(i32 145, i8 0)  ; DispatchRaysIndex(col)
  %10 = call i32 @dx.op.dispatchRaysIndex.i32(i32 145, i8 1)  ; DispatchRaysIndex(col)
  %11 = call %dx.types.Handle @dx.op.createHandleForLib.dx.types.Handle(i32 160, %dx.types.Handle %1)  ; CreateHandleForLib(Resource)
  %12 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %11, %dx.types.ResourceProperties { i32 4098, i32 1033 })  ; AnnotateHandle(res,props)  resource: RWTexture2D<4xF32>
  call void @dx.op.textureStore.f32(i32 67, %dx.types.Handle %12, i32 %9, i32 %10, i32 undef, float %8, float 1.000000e+00, float 1.000000e+00, float 1.000000e+00, i8 15)  ; TextureStore(srv,coord0,coord1,coord2,value0,value1,value2,value3,mask)
  call void @llvm.lifetime.end(i64 4, i8* %3) #0
  ret void
}
```

# Fix
In `DxilOperations.cpp`, the `WaveGetLaneIndex` op is marked as `ReadNone`, but it seems it should instead be marked as `ReadOnly`.

(I cannot submit the fix myself due to the CLA).

jasilvanus:
Wave intrinsics such as `WaveGetLaneIndex()`  are invalidated at DXR thread repacking
points such as `CallShader()`. We were however reusing the result of `WaveGetLaneIndex()`.

Fix this by marking it as `Readonly` instead of `Readnone`.

Add a test case that also covers other wave intrisics, which are handled correctly.

Fixes #5034.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/intrinsics/wave/repacking.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/intrinsics/wave/repacking.hlsl
@@ -26,9 +26,7 @@ struct Payload {
   // CHECK: call void @dx.op.callShader.struct.Payload(i32 159, i32 1
   // CHECK: call i32 @dx.op.waveGetLaneIndex(i32 111
   // CHECK: call void @dx.op.callShader.struct.Payload(i32 159, i32 1
-  // TODO: The second call is missing, we are incorrectly re-using the result
-  //       from before the repacking point.
-  // CHECK-NOT: call i32 @dx.op.waveGetLaneIndex(i32 111
+  // CHECK: call i32 @dx.op.waveGetLaneIndex(i32 111
   CallShader(1, p);
   p.value += WaveGetLaneIndex();
   CallShader(1, p);
--- a/tools/clang/test/HLSLFileCheck/hlsl/intrinsics/wave/repacking.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/intrinsics/wave/repacking.hlsl
@@ -1,0 +1,243 @@
+// RUN: %dxc -T lib_6_5 %s | FileCheck %s
+
+// Check that results of wave intrinsics are not re-used
+// cross DXR repacking points such as TraceRay() or CallShader();
+struct Payload {
+  unsigned int value;
+};
+
+[shader("miss")] void Miss(inout Payload p) {
+  // Some runtime value that we can't reason about.
+  // Freeze the current value and use it as input to
+  // all intrinsics so that we know the intrinsics' result
+  // would be the same before and after the CallShader call,
+  // except for repacking.
+  unsigned int oldValue = p.value;
+
+  // CHECK: call void @dx.op.callShader.struct.Payload(i32 159, i32 0
+  // CHECK: call i1 @dx.op.waveIsFirstLane(i32 110
+  // CHECK: call void @dx.op.callShader.struct.Payload(i32 159, i32 0
+  // CHECK: call i1 @dx.op.waveIsFirstLane(i32 110
+  CallShader(0, p);
+  p.value += WaveIsFirstLane();
+  CallShader(0, p);
+  p.value += WaveIsFirstLane();
+
+  // CHECK: call void @dx.op.callShader.struct.Payload(i32 159, i32 1
+  // CHECK: call i32 @dx.op.waveGetLaneIndex(i32 111
+  // CHECK: call void @dx.op.callShader.struct.Payload(i32 159, i32 1
+  // TODO: The second call is missing, we are incorrectly re-using the result
+  //       from before the repacking point.
+  // CHECK-NOT: call i32 @dx.op.waveGetLaneIndex(i32 111
+  CallShader(1, p);
+  p.value += WaveGetLaneIndex();
+  CallShader(1, p);
+  p.value += WaveGetLaneIndex();
+
+  // CHECK: call void @dx.op.callShader.struct.Payload(i32 159, i32 2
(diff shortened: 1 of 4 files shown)


jasilvanus:
Friendly ping for review.

jasilvanus:
Friendly ping for review. This fixes an active miscompilation.
 
(I'm planning to fix the merge conflicts quickly upon review feedback to avoid repeated resolving of them.)

tools/clang/test/HLSLFileCheck/hlsl/intrinsics/wave/repacking.hlsl
```diff
@@ -0,0 +1,241 @@
+// RUN: %dxc -T lib_6_5 %s | FileCheck %s
+
+// Check that results of wave intrinsics are not re-used
+// cross DXR repacking points such as TraceRay() or CallShader();
```
pow2clk, line 4: This mentions TraceRay, but never tests it. Are there other repacking points?

pow2clk (commented):
This definitely solves the problem. It could sacrifice some performance by recalling `WaveGetLaneIndex` in all the cases where it doesn't cross thread packing points. If all such points are in ray tracing, I wonder if there's a way to constrain it only to there. It would be more complicated and I'm not going to reject this for what might be a minimal performance hit in exchange for an obvious correctness improvement, but I wondered if it had been considered and possibly rejected. 

I think more thorough testing would involve TraceRay

tools/clang/test/HLSLFileCheck/hlsl/intrinsics/wave/repacking.hlsl
jasilvanus: I originally used CallShader because it requires the least boilerplate, but I agree we should test them all.
I've added TraceRay tests, and will later take a look at ReportHit.

jasilvanus: I'm now testing all the repacking point kinds: TraceRay, CallShader, ReportHit.

jasilvanus:
> This definitely solves the problem. It could sacrifice some performance by recalling `WaveGetLaneIndex` in all the cases where it doesn't cross thread packing points. If all such points are in ray tracing, I wonder if there's a way to constrain it only to there. It would be more complicated and I'm not going to reject this for what might be a minimal performance hit in exchange for an obvious correctness improvement, but I wondered if it had been considered and possibly rejected.

I'm not even sure re-using the result of cheap intrinsics such as `WaveGetLaneIndex()` is always a win, as it could increase register pressure, and the driver might even re-materialize to avoid that. There could be missed follow-up optimizations though due to e.g. CSE.

Also, the driver always has the option to re-use the earlier result if that is safe.

> 
> I think more thorough testing would involve TraceRay

Indeed, done. Also testing ReportHit now.

llvm-beanz (approved):
The specs and documentation for `GetWaveLaneIndex` don’t define which contexts the value might be invalidated, so I don’t think we should try to limit the scope of this change in existing shader models.

If we want to refine the spec further for future shader models so that we can consider that for a future shader model, but we should ensure that the existing shader models are correct in all cases.

PR #5615, opened 2023-08-28
Consolidate EmitInstrNote into common function

### LINKED ISSUE #5613: EmitInstrNote reports an error instead of a node, shares a lot of code and isn't tested
In DxilValidation, `EmitInstrNode` was based heavily on `EmitInstrErrorMsg`. The only real changes were to not skip the error if the same error had already been reported. Additionally, what was meant to be a note was reported like an error and there are not tests for the messages it produces

Rather than have two separate functions, modifying `EmitInstrNode` to handle notes and renaming it accordingly will share the code and allow notes to show up as they should. The completed-handle-used-* tests should be modified to include testing for this note

The original request was to consolidate a small part of `EmitInstrNode`, but I see no reason not to share nearly the entirety of the function.

**Additional context**
This is related to 6.8 work as it was added to produce errors for OutputComplete, but it could be useful in a number of scenarios

pow2clk:
EmitInstrNote reused most of the code from EmitInstrErrorMsg. This consolidates both functions into one that shares most of the code. In addition, it no longer makes the note look like an additional error in the output and adds testing for that note. Relevant tests are modified to include the notes

Fixes #5613


lib/HLSL/DxilValidation.cpp
```diff
@@ -534,70 +534,27 @@ struct ValidationContext {
...
-          break;
-        }
-        idx++;
-      }
-      Msg += "#" + std::to_string(idx);
-    }
-    Msg += "'";
-
-    // Print the function name
-    Msg += " of function '" + F->getName().str() + "'.";
```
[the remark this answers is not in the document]
pow2clk: Sorry the diff is extremely ugly and confusing. 

It's not being removed. This code was the same as the code that was originally in EmitInstrErrorMsg, which has remained unchanged below on what will be line 586 below.

lib/HLSL/DxilValidation.cpp
```diff
@@ -534,70 +534,27 @@ struct ValidationContext {
...
-  }
-
-  void EmitInstrErrorMsg(Instruction *I, ValidationRule Rule, std::string Msg) {
     Instruction *DbgI = GetDebugInstr(I);
-    const DebugLoc L = DbgI->getDebugLoc();
-    if (L) {
-      // Instructions that get scalarized will likely hit
-      // this case. Avoid redundant diagnostic messages.
-      if (Rule == LastRuleEmit && L == LastDebugLocEmit) {
-        return;
```
pow2clk, line 590: I think it's helpful to consider almost all of the above as removing EmitInstrNote outright and the below as the changes to EmitInstrErroMsg to add the ability to print a note as well as an error.

If I had it to do again, I'd have separated this into commits that alter EmitInstrErroMsg and then one that removes the  EmitInstrNote  function. I can still do that if you don't mind a force push.

bob80905 (approved):
LGTM

pow2clk:
I recommend reviewing the first commit followed by the second for more informative diff information

tools/clang/test/HLSLFileCheck/validation/completed-handle-all.ll
```diff
@@ -43,9 +43,11 @@ define void @loadStress_16() {
   ; test duplicate output complete call
   ; CHECK: error: Invalid use of completed record handle.
   ; CHECK: note: at 'call void @dx.op.outputComplete(i32 241, %dx.types.NodeRecordHandle %5)
+  ; CHECK: note: record handle invalidated by OutputComplete
+  ; CHECK: note: at 'call void @dx.op.outputComplete(i32 241, %dx.types.NodeRecordHandle %5)'
```
tex3d, line 47: Looks like an improvement, now if only we could print the instruction value assignment along with the call here, we could identify which call was being referred to!  Right now, it's not really helpful.

tex3d: I guess there's no value assignment because it returns void.  Is there any other way to uniquely identify this instruction in the stream?

tex3d (approved):
Looks like an improvement to me.

PR #5620, opened 2023-08-29
[DCE] Move output complete removal code later to DCE

### LINKED ISSUE #5363: Review issues with redundant OutputComplete elimination change
Redundant OutputComplete elilmination doesn't look like it's quite in the right place.

Moving the DxilMod init code was unnecessary, hlsl::OP can be retrieved from the m_pHLModule instead, as the others do.

See https://github.com/microsoft/DirectXShaderCompiler/blob/040e4078b3525bccef0b53464b17e0b49d518c1b/lib/HLSL/DxilGenerationPass.cpp#L281

I don't think this code should have been moved up here, above the comment that related to this step: // High-level metadata should now be turned into low-level metadata.
I think the code that gets rid of OutputComplete calls with zeroinit handle should probably be moved to some other (DCE-like) pass, later (since further optimization may produce this as well).

bob80905:
Elimination of incorrect output complete calls was located in an awkward spot. The code has been moved over to DCE. 
Fixes #5363

### DIFF
--- a/lib/DXIL/DxilOperations.cpp
+++ b/lib/DXIL/DxilOperations.cpp
@@ -648,7 +648,7 @@ bool OP::IsDxilOpFuncCallInst(const llvm::Instruction *I) {
 
 bool OP::IsDxilOpFuncCallInst(const llvm::Instruction *I, OpCode opcode) {
   if (!IsDxilOpFuncCallInst(I)) return false;
-  return llvm::cast<llvm::ConstantInt>(I->getOperand(0))->getZExtValue() == (unsigned)opcode;
+  return (unsigned)getOpCode(I) == (unsigned)opcode;
 }
 
 OP::OpCode OP::getOpCode(const llvm::Instruction *I) {
--- a/lib/DXIL/DxilOperations.cpp
+++ b/lib/DXIL/DxilOperations.cpp
@@ -651,9 +651,14 @@ bool OP::IsDxilOpFuncCallInst(const llvm::Instruction *I, OpCode opcode) {
   return llvm::cast<llvm::ConstantInt>(I->getOperand(0))->getZExtValue() == (unsigned)opcode;
 }
 
+OP::OpCode OP::getOpCode(const llvm::Instruction *I) {
+  return (OP::OpCode)llvm::cast<llvm::ConstantInt>(I->getOperand(0))
+      ->getZExtValue();
+}
+
 OP::OpCode OP::GetDxilOpFuncCallInst(const llvm::Instruction *I) {
   DXASSERT(IsDxilOpFuncCallInst(I), "else caller didn't call IsDxilOpFuncCallInst to check");
-  return (OP::OpCode)llvm::cast<llvm::ConstantInt>(I->getOperand(0))->getZExtValue();
+  return getOpCode(I);
 }
 
 bool OP::IsDxilOpWave(OpCode C) {
--- a/lib/DXIL/DxilUtil.cpp
+++ b/lib/DXIL/DxilUtil.cpp
@@ -1223,8 +1223,8 @@ bool DxilOpFunctionHasNoSideEffects(Instruction *I) {
   if (CallInst *CI = dyn_cast<CallInst>(I)) {
     if (CI->onlyReadsMemory()) return false;
 
-    bool isDxilOp = hlsl::OP::IsDxilOpFunc(CI->getCalledFunction());
-    if (!isDxilOp) return false;
+    if (!hlsl::OP::IsDxilOpFunc(CI->getCalledFunction()))
+      return false;
     switch (hlsl::OP::getOpCode(I)) {
     case hlsl::OP::OpCode::OutputComplete: {
       hlsl::DxilInst_OutputComplete OutputComplete(CI);
--- a/lib/DXIL/DxilUtil.cpp
+++ b/lib/DXIL/DxilUtil.cpp
@@ -16,6 +16,8 @@
 #include "dxc/DXIL/DxilOperations.h"
 #include "dxc/HLSL/DxilConvergentName.h"
 #include "dxc/Support/Global.h"
+#include "dxc/DXIL/DxilOperations.h"
+#include "dxc/DXIL/DxilInstructions.h"
 
 #include "llvm/IR/GetElementPtrTypeIterator.h"
 #include "llvm/ADT/StringExtras.h"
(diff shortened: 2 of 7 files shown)


lib/Transforms/Utils/Local.cpp
```diff
@@ -332,9 +335,19 @@ bool llvm::isInstructionTriviallyDead(Instruction *I,
...
 
-  // HLSL change - don't force unused convergenet markers to stay
-  if (CallInst *CI = dyn_cast<CallInst>(I))
+  // HLSL Change - don't force unused convergenet markers to stay, 
+  // remove bad OutputCompletes
+  if (CallInst *CI = dyn_cast<CallInst>(I)) {  
     if (hlsl::dxilutil::IsConvergentMarker(CI)) return true;
+    if (hlsl::OP::IsDxilOpFuncCallInst(I, hlsl::OP::OpCode::OutputComplete)) {
+      Value *NodeRecHandle =
+          CI->getArgOperand(hlsl::HLOperandIndex::kHandleOpIdx);
```
tex3d, line 344: We shouldn't be indexing into a Dxil operation using an `HLOperandIndex`.

lib/Transforms/Utils/Local.cpp
```diff
@@ -332,9 +335,19 @@ bool llvm::isInstructionTriviallyDead(Instruction *I,
     if (Constant *C = dyn_cast<Constant>(CI->getArgOperand(0)))
       return C->isNullValue() || isa<UndefValue>(C);
 
-  // HLSL change - don't force unused convergenet markers to stay
-  if (CallInst *CI = dyn_cast<CallInst>(I))
+  // HLSL Change - don't force unused convergenet markers to stay, 
+  // remove bad OutputCompletes
+  if (CallInst *CI = dyn_cast<CallInst>(I)) {  
     if (hlsl::dxilutil::IsConvergentMarker(CI)) return true;
+    if (hlsl::OP::IsDxilOpFuncCallInst(I, hlsl::OP::OpCode::OutputComplete)) {
```
tex3d, line 342: I wonder if we should instead call a utility function (perhaps defined in `dxilutil`) which returns true if the function call is an intermediate or DXIL op that has no side effects, for functions not marked ReadNone or ReadOnly.  This would incorporate both checks, but could also be expanded to handle more, such as wave ops.

tools/clang/test/HLSLFileCheck/validation/verify_output_complete_dce.ll
```diff
@@ -0,0 +1,114 @@
...
+  store i32 1, i32 addrspace(6)* %11, align 4
+  %12 = load i32, i32 addrspace(3)* getelementptr inbounds ([128 x i32], [128 x i32] addrspace(3)* @"\01?loadStressTemp@@3PAIA", i32 0, i32 0), align 4, !tbaa !22
+  %13 = getelementptr inbounds %struct.loadStressRecord.0, %struct.loadStressRecord.0 addrspace(6)* %8, i32 0, i32 1, i32 0
+  store i32 %12, i32 addrspace(6)* %13, align 4
+  %14 = load i32, i32 addrspace(3)* getelementptr inbounds ([128 x i32], [128 x i32] addrspace(3)* @"\01?loadStressTemp@@3PAIA", i32 0, i32 1), align 4, !tbaa !22
+  %15 = getelementptr inbounds %struct.loadStressRecord.0, %struct.loadStressRecord.0 addrspace(6)* %8, i32 0, i32 1, i32 1
+  store i32 %14, i32 addrspace(6)* %15, align 4
+  %16 = load i32, i32 addrspace(3)* getelementptr inbounds ([128 x i32], [128 x i32] addrspace(3)* @"\01?loadStressTemp@@3PAIA", i32 0, i32 2), align 4, !tbaa !22
+  %17 = getelementptr inbounds %struct.loadStressRecord.0, %struct.loadStressRecord.0 addrspace(6)* %8, i32 0, i32 1, i32 2
+  store i32 %16, i32 addrspace(6)* %17, align 4
```
tex3d, line 47: For the purpose of checking for the elimination of these through DCE, I don't think most of the code here is useful.  While it was generated from some HLSL, a lot of code could probably be deleted without making it invalid.  You can probably delete all instructions not involved in getting the valid handle value for the one outputComplete call you don't want deleted (`%6` through the last `store`).  Then you can delete unused function declarations too.

pow2clk (commented):
I have some additional thoughts

lib/Transforms/Utils/Local.cpp
tex3d: This is for our operations that otherwise appear to have side effects (aren't marked as (ReadOnly or ReadNone) because we want standard optimizations to treat them as if they have side effects.  But if the value is unused, or in this case, the handle is null, the operation is actually a no-op and does not have side effects.

tex3d: And `mayHaveSideEffects()` is an inexpensive inline operation on `Instruction` that we wouldn't want to pollute with DXIL operation checks and such.

lib/DXIL/DxilOperations.cpp
```diff
@@ -651,9 +651,14 @@ bool OP::IsDxilOpFuncCallInst(const llvm::Instruction *I, OpCode opcode) {
   return llvm::cast<llvm::ConstantInt>(I->getOperand(0))->getZExtValue() == (unsigned)opcode;
 }
 
+OP::OpCode OP::getOpCode(const llvm::Instruction *I) {
+  return (OP::OpCode)llvm::cast<llvm::ConstantInt>(I->getOperand(0))
+      ->getZExtValue();
+}
+
```
pow2clk, line 658: It's a nit, but if you end up making another change to this PR anyway, the `IsDxilOpFuncCallInst` function above would look a lot cleaner if it took advantage of this convenience function too.

lib/DXIL/DxilUtil.cpp
```diff
@@ -1215,6 +1217,32 @@ Value *TryReplaceBaseCastWithGep(Value *V) {
...
 }
 
+// returns true if the function call is an intermediate or DXIL op 
+// that has no side effects, for functions not marked ReadNone or ReadOnly.
+bool DxilOpFunctionHasNoSideEffects(Instruction *I) {
+  if (CallInst *CI = dyn_cast<CallInst>(I)) {
+    if (CI->onlyReadsMemory()) return false;
+
+    bool isDxilOp = hlsl::OP::IsDxilOpFunc(CI->getCalledFunction());
+    if (!isDxilOp) return false;
```
pow2clk, line 1227: Another nit. Given that the return value is only used this once, it would seem more natural to include the call in the conditional and leave out the variable entirely.

lib/DXIL/DxilUtil.cpp
```diff
@@ -1215,6 +1217,32 @@ Value *TryReplaceBaseCastWithGep(Value *V) {
...
+    switch (hlsl::OP::getOpCode(I)) {
+    case hlsl::OP::OpCode::OutputComplete: {
+      hlsl::DxilInst_OutputComplete OutputComplete(CI);
+      Value *NodeRecHandle = OutputComplete.get_output();
+      Constant *C = dyn_cast<Constant>(NodeRecHandle);
+      if (C && C->isZeroValue())
+        return true;
```
pow2clk, line 1234: I think Tex had in mind that this would include iconvergentmarker and the code that ensures it's a call instruction as well. I think the idea is one place to check HLSL-specific functions that aren't already marked readonly or readnone to see if they are known to not have side effects.

pow2clk (commented):
I'll let Tex confirm if I understood his intent for the dxilutil function correctly. I have a couple nitpicks you can take or leave besides

lib/DXIL/DxilUtil.cpp
bob80905: Thanks for the feedback, I'll wait for Tex to confirm.

pow2clk: Well, I'll go ahead and say that I think that would be a better way to structure this. Rather than having a generic named function that only checks for `OutputComplete` to determine if there are potential side effects, I think it should be more holistic and at least include the other HLSL-specific case we're aware of as well.

tex3d (approved):
LGTM!

pow2clk (approved):
Looks good!

PR #5621, opened 2023-08-29
Merge branch 'main' into staging-sm-6.8

pow2clk:
Conflicts primarily from differences in how the compiler version
mismatch on linking was implemented, some additional basic types added
in one place and the other, and some incidental clang-formatting along
the way in main


lib/HLSL/HLOperations.cpp
```diff
@@ -530,6 +558,79 @@ static void SetHLFunctionAttribute(Function *F, HLOpcodeGroup group,
...
+        if (it->isEnumAttribute()) {
+          switch (it->getKindAsEnum()) {
+          case Attribute::ReadNone:
+            ReadNone = true;
+            break;
+          case Attribute::ReadOnly:
+            ReadOnly = true;
+            break;
+          case Attribute::ArgMemOnly:
+            ArgMemOnly = true;
```
pow2clk, line 594: Merge change. Handle ArgMemOnly is introduced by sm68

lib/HLSL/HLOperations.cpp
```diff
@@ -530,6 +558,79 @@ static void SetHLFunctionAttribute(Function *F, HLOpcodeGroup group,
...
+            assert(false && "unexpected attribute for HLOperation");
+          }
+        } else if (it->isStringAttribute()) {
+          StringRef Kind = it->getKindAsString();
+          if (Kind == HLWaveSensitive) {
+            assert(it->getValueAsString() == "y" &&
+                   "otherwise, unexpected value for WaveSensitive attribute");
+            WaveSensitive = true;
+          } else {
+            assert(Kind != "ExtractRecordStructFromArray" &&
```
pow2clk, line 612: Merge change> ExtractRecordStructFromArray is introduced with sm 6.8

pow2clk (commented):
Just some notes about things I had to change to make the main code work with the 6.8 cahnges

llvm-beanz (approved):

lib/HLSL/HLOperations.cpp
```diff
@@ -471,42 +478,63 @@ HLBinaryOpcode GetUnsignedOpcode(HLBinaryOpcode opcode) {
...
   } break;
   case HLOpcodeGroup::HLAnnotateHandle: {
-    F->addFnAttr(Attribute::ReadNone);
+    addAttr(Attribute::ReadNone);
   } break;
   case HLOpcodeGroup::HLWaveMatrix_Annotate: {
-    F->addFnAttr(Attribute::ArgMemOnly);
-    F->addFnAttr(Attribute::NoUnwind);
+    addAttr(Attribute::ArgMemOnly);
+    addAttr(Attribute::NoUnwind);
```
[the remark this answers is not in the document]
tex3d: Good, but `addAttr(Attribute::NoUnwind);` is not necessary here, since it's always added at the beginning of this function.

lib/HLSL/HLOperations.cpp
tex3d: Don't you mean to assert `Kind == "ExtractRecordStructFromArray"` instead of `!=`?

lib/HLSL/HLOperations.cpp
```diff
@@ -530,6 +558,79 @@ static void SetHLFunctionAttribute(Function *F, HLOpcodeGroup group,
...
+  assert(!(ReadNone && ReadOnly && ArgMemOnly) &&
+         "ReadNone, ReadOnly, and ArgMemOnly are mutually exclusive");
+
+  // Add mangling in canonical order
+  if (NoDuplicate)
+    mangledNameStr << "nd";
+  if (ReadNone)
+    mangledNameStr << "rn";
+  if (ReadOnly)
+    mangledNameStr << "ro";
```
tex3d, line 630: Did you want to add ArgMemOnly "amo" as part of this merge, or separately?  If separately, is there an issue tracking that?

tex3d (commented):
I think there's a potential bug with `ExtractRecordStructFromArray` if I'm reading this right.

lib/HLSL/HLOperations.cpp
pow2clk: Actually, it turns out I meant Kind == "dx.hlls", which is the kind that is accompanied by that value

tex3d (approved):
LGTM!

PR #5622, opened 2023-08-30
Make NodeLaunch attribute strings case sensitive

### LINKED ISSUE #5370: Node launch types should be case-sensitive and lowercase
As per the spec, NodeLanuch string arguments should be interpreted as case-sensitive and lowercase matching other string arguments to attributes:

https://github.com/microsoft/hlsl-specs/pull/62/files#diff-f13bec8f0222a19e56d6433159f69368dd5b5fb9f7b1e0ef0a87058ced354bf9R85

tcorringham:
NodeLaunch string arguments should be interpreted as case-sensitive and lowercase matching other string arguments to attributes. The exisiting implementation followed an earlier spec which allowed case-insensitive NodeLaunch strings - this change disallows that, and amends all the tests that had strings that were not lower case.

Fixes #5370


tools/clang/test/HLSL/workgraph/node_input_compatibility.hlsl
```diff
@@ -12,7 +12,7 @@ struct RECORD
 [Shader("node")]
 [NumThreads(1024,1,1)]
 [NodeDispatchGrid(64,1,1)]
-[NodeLaunch("BrOaDcasting")]                         // expected-note  {{Launch type defined here}}
```
llvm-beanz, line 15: We should probably have tests that verify that we get an expected error with incorrect capitalizations like what was here originally.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12766,9 +12766,6 @@ static StringRef ValidateAttributeStringArg(Sema &S, const AttributeList &A,
```
llvm-beanz, line 12766: Can we restore this entire function to the version in `main`? The `isCaseSensitive` flag was added for work graphs, so we should be able to just revert this function.

llvm-beanz (commented):
Can we get a

tools/clang/lib/Sema/SemaHLSL.cpp
tcorringham: I did exactly that in the first commit - it looks as if the subsequent merge reinstated the isCaseSensitive version without me noticing!
I will sort this on return from vacation...

llvm-beanz (approved):

tex3d:
Considering the impact on all these existing tests which could indicate impact to people using the preview, should we consider adding a clang FixIt for attribute values with mismatched case?

tex3d (approved):

llvm-beanz:
A fix-it doesn't really get us much since it can't auto-apply in DXC since that infrastructure is broken.

PR #5624, opened 2023-08-30
[Test][NFC] Add shader stage attribute to test shaders

### LINKED ISSUE #5623: [NFC] Add compute shader attribute to function declarations that should be compute shaders.
Some test files in the test suite have function declarations that were intended to be used as compute shaders, but only have the numthreads attribute. This is incorrect behavior, but works because numthreads currently implies compute shader, which is an inference that will soon be removed.

To reduce the clutter of subsequent PRs when the inference is removed, and to retroactively improve the accuracy of the tests in our test suite, the compute shader attribute should be added to the tests that were intended to be interpreted as compute shaders.

Adding the attribute along with the inference change was considered, but it would be a larger PR than necessary. This large change should be isolated.

**Additional context**

bob80905:
The team has come to a decision that no inferences should be made on function declarations that have a numthreads attribute (specifically, inferences on whether or not the function represents a shader, or a compute shader, even). 
However, some tests exist that were written with the expectation that they would be treated as a compute shader, and they lack the compute shader attribute. This PR addresses the problem by correctly adding the compute shader attribute to the function declarations in these tests. The PR is necessary so that these tests won't fail when the inference behavior is removed. 
Fixes #5623

### DIFF
--- a/tools/clang/test/CodeGenHLSL/lib_cs_entry.hlsl
+++ b/tools/clang/test/CodeGenHLSL/lib_cs_entry.hlsl
@@ -23,6 +23,7 @@ void StoreOutputMat(float2x2  m, uint gidx);
 float2x2 LoadInputMat(uint x, uint y);
 float2x2 RotateMat(float2x2 m, uint x, uint y);
 
+[shader("compute")]
 [numthreads(8,8,1)]
 void entry( uint2 tid : SV_DispatchThreadID, uint2 gid : SV_GroupID, uint2 gtid : SV_GroupThreadID, uint gidx : SV_GroupIndex )
 {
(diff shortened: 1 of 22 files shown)


tex3d (approved):
Looks good.

To be clear, the design was always supposed to be: when compiling to a library target, shader functions must use the `[shader(...)]` attribute to indicate a shader entry point.

There is a bug however, which relies on some shader entry stage inference based on the other entry point attributes on a function.  This inference was originally mainly meant to stop entry points from having attributes for different shader stages at the same time.

When library target support was added, this inference leaked into defining some entry point types, even when the `[shader(...)]` attribute was absent.

This inference was also not fully correct when compiling a library.  For instance, the `[numthreads(...)]` attribute can be used on compute, mesh, amplification, and now node shaders.  But it would only detect mesh or amplification cases when compiling that entry directly to one of those shader targets (not for a library, and not for functions that are not the current entry).

This change is preparation for the removal of this buggy shader stage inference code, since a number of tests ended up accidentally relying on it.

pow2clk (approved):
Looks good!

PR #5627, opened 2023-08-30
Add profile target shader attribute to entry point function declaration

### LINKED ISSUE #5626: [Feature Request] Target profiles should be automatically added as HLSL Shader attributes to entry function declarations.
On typical compilation, the target profile and entry point is provided. There are cases where the actual entry point declaration in the source lacks an HLSLShader attribute. This PR allows for function declarations that match the entry point name specified in the compilation arguments (that is, the entry point function) to automatically obtain an HLSL Shader attribute matching the target profile that was also specified in the compilation arguments. 

The attribute should be inferable from the target profile specified on the compilation command line. 
This should mean that shaders like:

```
DXC -T cs_6_6 -E mymain

[numthreads(1,1,1)]
void mymain(){}
```

should, at the AST level, have the `compute` shader attribute attached to the `mymain` function declaration, even though the attribute wasn't specified.

bob80905:
This PR will automatically add the target profile shader stage as a HLSLShader attribute to the function declaration that represents the entry point. If the entry point doesn't exist (in the case of library shaders), then this behavior won't happen.
Fixes #5626


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15360,8 +15360,30 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
 
 void DiagnoseEntry(Sema &S, FunctionDecl *FD) {
   auto Attr = FD->getAttr<HLSLShaderAttr>();
-  if (!Attr)
-    return;
+  if (!Attr) {
+    // if this is the Entry FD, then add the target profile
```
tex3d, line 15364: Can we wrap this code in a function, then unconditionally add the attribute before this function looks it up?  Otherwise, this code will ignore if we specify a target that conflicts with an existing shader attribute.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15360,8 +15360,30 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+    // otherwise, with no shader attribute, just return
+    const std::string &EntryPointName = S.getLangOpts().HLSLEntryFunction;
+    if (EntryPointName.empty()) {
+      return;
+    }
+
+    
+    // if this FD isn't the entry point, then there's no
+    // shader attribute to work with, so just return
+    if (EntryPointName != FD->getIdentifier()->getName()) {
```
tex3d, line 15375: This will need to be more specific about the scope of the decl, otherwise we could match a function by the same name outside global scope.

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -415,6 +415,43 @@ DXIL::ShaderKind ShaderModel::KindFromFullName(llvm::StringRef Name) {
...
+    return "miss";
+  case DXIL::ShaderKind::Callable:
+    return "callable";
+  case DXIL::ShaderKind::Mesh:
+    return "mesh";
+  case DXIL::ShaderKind::Amplification:
+    return "amplification";
+  case DXIL::ShaderKind::Node:
+    return "node";
+  default:
```
tex3d, line 450: We should probably include all values from the enum and place `llvm_unreachable("unknown ShaderKind")` under `default:`.

For the `Invalid` and `Library` cases, can we return an empty `StringRef()`?  This would make it easier to detect when used.  `Library` is a valid enum value but doesn't have a corresponding long name for use in a `[shader(...)]` attribute, since that's not a shader entry type.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15360,8 +15360,30 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+      return;
+    }
+
+    std::string profile = S.getLangOpts().HLSLProfile;
+    const ShaderModel *SM = hlsl::ShaderModel::GetByName(profile.c_str());
+    const llvm::StringRef fullName = ShaderModel::FullNameFromKind(SM->GetKind());
```
tex3d, line 15381: The attribute should only be added when we have a valid shader target (as opposed to a library target, for instance).  If you follow my other suggestion about returning an empty `StringRef()`, this can be accomplished by only adding the attribute when the `fullName` returned is not empty.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15358,10 +15358,52 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+  // There's no way we should be adding a shader attribute
+  // to a Function Decl if it doesn't even have an identifer
+  if (!FD->getIdentifier()) {
+    return;
+  }
+
+  // if this FD isn't the entry point, then there's no
+  // shader attribute to work with, so just return
+  if (EntryPointName != FD->getIdentifier()->getName() || 
+    (EntryPointName == FD->getIdentifier()->getName() && !FD->isGlobal())) {
```
pow2clk, line 15380: I don't think patch constant functions can have attributes even though they are kind of entry points. I wanted to float the issue in case others know better. If they can, this code will reject them and that's a problem.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15358,10 +15358,52 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+  const ShaderModel *SM = hlsl::ShaderModel::GetByName(profile.c_str());
+  const llvm::StringRef fullName = ShaderModel::FullNameFromKind(SM->GetKind());
+
+  // don't add the attribute for an invalid profile, like library
+  if (fullName.empty()) {
+    return;
+  }
+
+  HLSLShaderAttr *pShaderAttr =
+      HLSLShaderAttr::CreateImplicit(S.Context, fullName);
```
pow2clk, line 15394: If we were to define attributes for each of the shader stages in Attr.td (we have some HLSL-specific ones there already), I think we could get away without having to pass strings around. That's something we're trying to do less of.

pow2clk (commented):
I think this could be better if we created HLSLShaderAttrs for the different stages and applied them directly instead of using the implicit string method.

pow2clk:
Hmm. on second examination, I might be confused here. I guess HLSLShader is already an attribute that takes the string as an argument. To change that would be pretty disruptive

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15358,10 +15358,46 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+  // shader attribute to the FD and carry on with validation
+  // otherwise, with no shader attribute, just return
+  if (EntryPointName.empty()) {
+    return;
+  }
+
+  // if this FD isn't the entry point, then there's no
+  // shader attribute to work with, so just return
+  if (EntryPointName != FD->getIdentifier()->getName() || 
+    (EntryPointName == FD->getIdentifier()->getName() && !FD->isGlobal())) {
```
tex3d, line 15380: I don't think `!FD->isGlobal()` is quite right, however, I don't think the condition on the call to `DiagnoseEntry()` is sufficient in the first place.  I think we need to check a few more conditions to ensure we are only calling `DiagnoseEntry()` on top-level function definitions (not declarations) that really could be entry points.  

For instance, I think this condition should be sufficient (`SemaDecl.cpp` near end of `Sema::ActOnFunctionDeclarator`):
```cpp
  if (getLangOpts().HLSL && D.isFunctionDefinition() && D.hasName() &&
      NewFD->getDeclContext()->getRedeclContext()->isTranslationUnit()) {
    hlsl::DiagnoseEntry(*this, NewFD);
  }
```

With that in place, `TryAddShaderAttrFromTargetProfile` doesn't need to check `isGlobal()`.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15358,10 +15358,46 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+  std::string profile = S.getLangOpts().HLSLProfile;
+  const ShaderModel *SM = hlsl::ShaderModel::GetByName(profile.c_str());
+  const llvm::StringRef fullName = ShaderModel::FullNameFromKind(SM->GetKind());
+
+  // don't add the attribute for an invalid profile, like library
+  if (fullName.empty()) {
+    return;
+  }
+
+  HLSLShaderAttr *pShaderAttr =
```
tex3d, line 15393: We should also check if there's an existing `shader` attribute, and if so, skip adding it after making sure it matches the `fullName` for the shader stage.

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -415,6 +416,47 @@ DXIL::ShaderKind ShaderModel::KindFromFullName(llvm::StringRef Name) {
...
+    return "vertex";
+  case DXIL::ShaderKind::Geometry:
+    return "geometry";
+  case DXIL::ShaderKind::Hull:
+    return "hull";
+  case DXIL::ShaderKind::Domain:
+    return "domain";
+  case DXIL::ShaderKind::Compute:
+    return "compute";
+  case DXIL::ShaderKind::Library:
```
tex3d, line 433: Can Library be grouped with Invalid, with a comment as to why?  Perhaps like:
```cpp
  // Library has no full name for use with shader attribute.
  case DXIL::ShaderKind::Library:
  case DXIL::ShaderKind::Invalid:
    return llvm::StringRef();
```

tools/clang/test/HLSLFileCheck/hlsl/functions/attribute/attributeAddition.hlsl
```diff
@@ -0,0 +1,10 @@
+// RUN: %dxc -T cs_6_6 -E mymain -ast-dump %s | FileCheck %s
+
+// CHECK: mymain 'void ()'
+// CHECK: -HLSLShaderAttr
+// CHECK-SAME: "compute"
+
+[numthreads(1,1,1)]
+void mymain(){
```
tex3d, line 8: We probably need some more tests for entry functions, such as ones that should be ignored, like in namespaces, classes, or inside a function, and declarations vs. definitions.

tex3d (changes requested):

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: If I'm comparing equality for target profile and any existing attributes, it begs the question of whether this function should be responsible for emitting a diagnostic if there's a mismatch. Instead of leaving it to later validation, should I emit a diagnostic if, say, the "pixel" attr is defined on the decl, but the target profile is "vertex", and delete that validation code when it happens later?

tools/clang/lib/Sema/SemaDecl.cpp
```diff
@@ -8108,7 +8108,8 @@ Sema::ActOnFunctionDeclarator(Scope *S, Declarator &D, DeclContext *DC,
     AddToScope = false;
   }
 
-  if (getLangOpts().HLSL) {
+  if (getLangOpts().HLSL && D.isFunctionDefinition() && D.hasName() &&
+      NewFD->getDeclContext()->getRedeclContext()->isTranslationUnit()) {
```
pow2clk, line 8112: The only diagnostic that seems to use the name is `DiagnoseNodeEntry` and that only for actual diagnostic errors. I don't know how a function could be missing a name though. What prompted these additions?

pow2clk (commented):
I just have a quick question

tools/clang/lib/Sema/SemaDecl.cpp
bob80905: The additional conditions were necessary to prevent accidentally matching a declaration with the same name that was outside of the global scope. Additionally, the function should only be called on top level decl's that could be the Entry function. I don't think any entry function decl will not have a name, so the condition will act as an optimization. There are also some declarations that lack names that end up appearing inside DiagnoseEntry, so the condition skips over those cases.

pow2clk (approved):
Thanks for the clarification. Looks good!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15358,10 +15358,66 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+  const std::string &EntryPointName = S.getLangOpts().HLSLEntryFunction;  
+  
+  // if there's no defined entry point, just return
+  if (EntryPointName.empty()) {
+    return;
+  }
+  
+  // There's no way we should be adding a shader attribute
+  // to a Function Decl if it doesn't even have an identifier
+  if (!FD->getIdentifier()) {
```
tex3d, line 15373: Earlier checks should have ruled out this case.

tex3d (changes requested):

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d: If we have specific diagnostics elsewhere for trying to compile an entry point marked with a conflicting shader attribute later, then we can probably skip both the diagnostic and the addition of the conflicting attribute here.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15358,10 +15358,66 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+  if (fullName.empty()) {
+    return;
+  }
+
+  HLSLShaderAttr *currentShaderAttr = FD->getAttr<HLSLShaderAttr>();
+  // don't add the attribute if it already exists as an attribute on the decl  
+  if (currentShaderAttr) {
+    llvm::StringRef currentFullName = currentShaderAttr->getStage();
+    if (currentFullName == fullName) {
+      return;
```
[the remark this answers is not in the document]
bob80905: I've added an extra diag for that case. However, the node+compute case is necessary. What it's doing is preventing the addition of the compute shader attribute to a function that already has a "node" shader attribute attached. We don't want to add the compute attribute implicitly after the spec change, and it's unnecessary to do it before the spec change too, so I think I should leave this behavior as is.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15358,10 +15358,66 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
+    }
+  }
+
+  // in the special case that the target profile is compute and the 
+  // entry decl already has a node shader attr, don't do anything
+  if (fullName == "compute" && currentShaderAttr->getStage() == "node") {
```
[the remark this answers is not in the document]
bob80905: I think the complication is necessary when considering the responsibility of the function ("Try to add the target profile as a shader attribute to the entry function") We shouldn't do this if the target is compute and there's already a node attribute, that check seems necessary to place in this function to me. 

I do see the problem that there could already be several attributes on the decl, and that we can choose the wrong one. However, the multiple attributes currently will be caught by the code in CGHLSLMS.cpp, and those diags will eventually be moved to Sema, so I don't think this function has anything wrong specifically in terms of design. To be clear, I don't think this function's responsibilties should involve checking to see that the entry decl has exactly one attribute. I've updated the code to prevent conflicting attributes from being added, along with a note that tells the user that the target profile conflicts with an existing explicit attribute. As to the multiple attributes problem, I think this function is fine and I should make an issue to move the multiple attributes detection in CGHLSLMS to a location before this function gets called.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
tex3d: The diagnostic for conflicting attributes should be earlier - before this code is run, such as in `DiagnoseHLSLDecl`.  That should invalidate the `Declarator`, preventing us from even reaching this code when there are conflicting explicit attributes.

If there's an existing attribute that conflicts with the current shader target type, then we need to diagnose that here, which I've seen you've added.  However, I think a single new diagnostic for this situation would be clearer than reusing the `err_hlsl_attribute_mismatch` plus a new `note_hlsl_attribute_mismatch`.  Not only are those diagnostic identifiers misleading in name, the error output is confusing by being structured around an internal detail about how we create an implicit attribute from the target profile.

Compiling this:
```hlsl
#define ENTRY_TYPE "vertex"
[shader(ENTRY_TYPE)]
float4 main() : OUTPUT { ... }
```

With -T ps_6_0 would result in something like this:
```
test.hlsl:2:1: error: invalid shader stage attribute combination
[shader(ENTRY_TYPE)]
^
test.hlsl:2:1: note: implicit attribute from target profile is: pixel
[shader(ENTRY_TYPE)]
^
```

When it could be something more like this:
```
[shader(ENTRY_TYPE)]
^
test.hlsl:2:1: error: entry type 'pixel' from profile 'ps_6_0' conflicts with shader attribute type 'vertex' on entry function 'main'.
```

I also don't think we should be exempting the compute+node case here.  If we do, we will validate the entry point as a node shader, not as a compute shader, when your target profile is compute.  We don't want to support this scenario going forward, and though there are likely still tests that will break, can we remove or modify those tests before merging this change?

Next, I don't think this code should ever add a shader attribute if there's an existing shader attribute.  If there was a conflict, that should be a fatal error for considering this a valid entry point to even attempt to diagnose.  Otherwise, more confusing errors could be emitted, when this error needs to be fixed before we know how to interpret and diagnose the entry function properly.

Earlier I mentioned that the diagnostics `err_hlsl_attribute_mismatch` and `note_hlsl_attribute_mismatch` are misleading in name.  That's because there are many HLSL attributes that could "mismatch", but this error is very specific to the shader attribute, even though it doesn't say 'shader' in the name.  The note is specific to a mismatch between a shader attribute and the target profile, so that name is not clear either, though I don't think we need that one if we go with what I've suggested here.

I suggest these changes to the diagnostics:
- change `err_hlsl_attribute_mismatch` to `err_hlsl_conflicting_shader_attributes`
- remove `note_hlsl_attribute_mismatch`
- add a new `err_hlsl_profile_conflicts_with_shader_attribute` diagnostic.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12836,7 +12836,8 @@ HLSLShaderAttr* ValidateShaderAttributes(Sema &S, Decl *D,
          Stage != DXIL::ShaderKind::Node) ||
         (NewStage != DXIL::ShaderKind::Compute &&
          NewStage != DXIL::ShaderKind::Node)) {
```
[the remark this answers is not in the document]
bob80905: Yes, that is the plan :)

tex3d (approved):
LGTM.

PR #5628, opened 2023-08-30
Fix asan stack use after return

amaiorano:
Fix asan use-after-return

In cases where arg 0's template id set to one of
INTRIN_TEMPLATE_FROM_TYPE, INTRIN_TEMPLATE_VARARGS,
or INTRIN_TEMPLATE_FROM_FUNCTION), this code would erroneously index
beyond Template's size.

ASAN helped to catch this error, reporting it as a use-after-return.

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -6462,8 +6462,10 @@ bool HLSLExternalSource::MatchArguments(
   // For object return types that need to match arguments, we need to slot in
   // the full type here Can't do it sooner because when return is encountered
   // above, the other arg types haven't been set
-  if (Template[pIntrinsic->pArgs[0].uTemplateId] == AR_TOBJ_OBJECT)
-    argTypes[0] = argTypes[pIntrinsic->pArgs[0].uComponentTypeId];
+  if (pIntrinsic->pArgs[0].uTemplateId < MaxIntrinsicArgs) {
+    if (Template[pIntrinsic->pArgs[0].uTemplateId] == AR_TOBJ_OBJECT)
+      argTypes[0] = argTypes[pIntrinsic->pArgs[0].uComponentTypeId];
+  }
 
   return badArgIdx == MaxIntrinsicArgs;
 #undef CAB


amaiorano:
Some more info for this fix:

We discovered this bug when running ASAN-instrumented Dawn end-to-end tests (see [bug here](https://bugs.chromium.org/p/tint/issues/detail?id=2021)). I was able to reproduce it with a version of DXC built with ASAN enabled against this shader:

```
cbuffer cbuffer_tint_symbol_8 : register(b0) {
  uint4 tint_symbol_8[1];
};
RWByteAddressBuffer tint_symbol_9 : register(u1);
RWByteAddressBuffer tint_symbol_10 : register(u2);

[numthreads(1, 1, 1)]
void main() {
  {
    for(uint tint_symbol_12 = 0u; (tint_symbol_12 < 3u); tint_symbol_12 = (tint_symbol_12 + 1u)) {
      uint tint_symbol_13 = tint_symbol_9.Load((4u * (tint_symbol_8[0].y + tint_symbol_12)));
      bool tint_tmp = (tint_symbol_8[0].z > 0u);
      if (tint_tmp) {
        tint_tmp = (tint_symbol_13 > tint_symbol_8[0].x);
      }
      if ((tint_tmp)) {
        tint_symbol_13 = 0u;
      }
      tint_symbol_10.Store((4u * tint_symbol_12), asuint(tint_symbol_13));
      if ((tint_symbol_8[0].w > 0u)) {
        tint_symbol_10.Store((4u * (tint_symbol_12 + 3u)), asuint(tint_symbol_13));
      }
    }
  }
  return;
}
```

ASAN reported a use-after-return on the following line in `HLSLExternalSource::MatchArguments`:
```
  // For object return types that need to match arguments, we need to slot in
  // the full type here Can't do it sooner because when return is encountered
  // above, the other arg types haven't been set
  if (Template[pIntrinsic->pArgs[0].uTemplateId] == AR_TOBJ_OBJECT) <--- HERE
    argTypes[0] = argTypes[pIntrinsic->pArgs[0].uComponentTypeId];
```

Debugging this, the problem is that when compiling the shader listed above, at this point, the value of `pIntrinsic->pArgs[0].uTemplateId` is 0xFD, or `INTRIN_TEMPLATE_FROM_FUNCTION`:

```
static const BYTE INTRIN_TEMPLATE_FROM_TYPE = 0xff;
static const BYTE INTRIN_TEMPLATE_VARARGS = 0xfe;
static const BYTE INTRIN_TEMPLATE_FROM_FUNCTION = 0xfd;
```

My fix here is to ensure that the templateId is within valid range:

```
  if (pIntrinsic->pArgs[0].uTemplateId < MaxIntrinsicArgs) {
    if (Template[pIntrinsic->pArgs[0].uTemplateId] == AR_TOBJ_OBJECT)
      argTypes[0] = argTypes[pIntrinsic->pArgs[0].uComponentTypeId];
  }
```

I'm not sure if this is the right fix, or whether we should be checking that the template id is not any of the three special values (`INTRIN_TEMPLATE_FROM_TYPE` , etc.), as is partially done in the code above these lines. In any case, this fixes the ASAN error, and seems to compile our shaders successfully.

llvm-beanz (approved):
This asan failure occurs against in-tree tests, so we already have tests that exercise this.

python3kgae (approved):

pow2clk:
> This asan failure occurs against in-tree tests, so we already have tests that exercise this.

Which tests are these? The example in the description isn't in tree

pow2clk (approved):
Nevermind my question about tests. I see that there are a few common methods that use this inferred type for the first argument

llvm-beanz:
Yea, there's _a lot_ of tests that hit this error. Kinda amazing that this never caused any user-visible bugs.

PR #5630, opened 2023-08-31
Tolerate argmemonly in mangled name handling

pow2clk:
The argmemonly attrib caused an assert because it wasn't handled by the mangling code. This doesn't include mangling for it yet as there are no intrinsics that use it yet, but it prevents the assert.

Followon to #5451


bob80905 (commented):
For future reference, might it be worth mentioning which file / what shader the assert was caused on?

Issue #5551, opened 2023-08-15
DXC tests: hcttest does not run ExecutionTests by default


hekota:
**Description**
After switching to LIT by default `hcttest` does not include ExecutionTests.

**Steps to Reproduce**
`hcttest`

**Actual Behavior**
Execution tests are not included in the run. Previously when `hcttest` was called without an argument it included all DXC test categories (clang, cmd, exec and dxilconv). The test runs do not list which tests were run (#5550) and that is part of the reasons which this went unnoticed for some time. To see the tests that ran one needs to change `LIT_ARGS_DEFAULT` from `-sv` to `-v`.

**Environment**
DXC main as if 8/15/2023
Windows

PR #5652, opened 2023-09-05
[lit] Enable exec test by default for hcttest

python3kgae:
Enable exec test for hcttest -all.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5551

### DIFF
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -313,7 +313,7 @@ if "%TEST_USE_LIT%"=="1" (
     set RES_DXILCONV=!RES_CLANG!
     rem check exec.
     if defined EXEC_ADAPTER (
-        echo The -adapter parameter is supported only when running just execution tests (hcttest.cmd exec)
+        echo "The -adapter parameter is supported only when running just execution tests (hcttest.cmd exec)"
       )
     set RES_EXEC=!ERRORLEVEL!
   ) else (
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -313,7 +313,7 @@ if "%TEST_USE_LIT%"=="1" (
     set RES_DXILCONV=!RES_CLANG!
     rem check exec.
     if defined EXEC_ADAPTER (
-        echo "The -adapter parameter is supported only when running just execution tests (hcttest.cmd exec)"
+        echo The -adapter parameter is supported only when running just execution tests ^(hcttest.cmd exec^)
       )
     set RES_EXEC=!ERRORLEVEL!
   ) else (
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -313,7 +313,7 @@ if "%TEST_USE_LIT%"=="1" (
     set RES_DXILCONV=!RES_CLANG!
     rem check exec.
     if defined EXEC_ADAPTER (
-        echo -adapter is not support when running all tests.
+        echo The -adapter parameter is supported only when running just execution tests (hcttest.cmd exec)
       )
     set RES_EXEC=!ERRORLEVEL!
   ) else (
--- a/utils/hct/hcttest.cmd
+++ b/utils/hct/hcttest.cmd
@@ -311,6 +311,13 @@ if "%TEST_USE_LIT%"=="1" (
     cmake --build %HLSL_BLD_DIR% --config %BUILD_CONFIG% --target check-all
     set RES_CLANG=!ERRORLEVEL!
     set RES_DXILCONV=!RES_CLANG!
+    rem check exec.
+    if defined EXEC_ADAPTER (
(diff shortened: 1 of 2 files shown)


utils/hct/hcttest.cmd
```diff
@@ -311,6 +311,13 @@ if "%TEST_USE_LIT%"=="1" (
     cmake --build %HLSL_BLD_DIR% --config %BUILD_CONFIG% --target check-all
     set RES_CLANG=!ERRORLEVEL!
     set RES_DXILCONV=!RES_CLANG!
+    rem check exec.
+    if defined EXEC_ADAPTER (
+        py %HLSL_SRC_DIR%/utils/lit/lit.py -v --no-progress-bar --param build_mode=%BUILD_CONFIG% --param clang_site_config=%HLSL_BLD_DIR%/tools/clang/test/lit.site.cfg --param clang_taef_exec_site_config=%HLSL_BLD_DIR%/tools/clang/test/taef_exec/lit.site.cfg %EXEC_ADAPTER% %HLSL_SRC_DIR%/tools/clang/test/taef_exec
+      ) else (
+        cmake --build %HLSL_BLD_DIR% --config %BUILD_CONFIG% --target check-clang-taef-exec
```
hekota, line 318: Is there a reason why check-clang-taef-exec is not included in check-all?

python3kgae: github pipeline doesn't run exec test.
check-all is matching this behavior.

hekota: GitHub pipeline is running the tests with `noexec` parameter:

[call utils\hct\hcttest.cmd -$(configuration) noexec](https://github.com/microsoft/DirectXShaderCompiler/blob/bdefcfda03e7986f5da95c64fb5f4326f7420576/azure-pipelines.yml#L35)

I think is would make sense to include `check-clang-taef-exec` into `check-all`.

llvm-beanz: Yea, `check-all` should be everything, including the exec tests.

python3kgae: Fixed.

hekota: I think we should keep skip_taef_exec, but they default value should be False. It is useful to run exec and non-exec tests separately.

utils/hct/hcttest.cmd
```diff
@@ -311,6 +311,11 @@ if "%TEST_USE_LIT%"=="1" (
     cmake --build %HLSL_BLD_DIR% --config %BUILD_CONFIG% --target check-all
     set RES_CLANG=!ERRORLEVEL!
     set RES_DXILCONV=!RES_CLANG!
+    rem check exec.
+    if defined EXEC_ADAPTER (
+        py %HLSL_SRC_DIR%/utils/lit/lit.py -v --no-progress-bar --param build_mode=%BUILD_CONFIG% --param clang_site_config=%HLSL_BLD_DIR%/tools/clang/test/lit.site.cfg --param clang_taef_exec_site_config=%HLSL_BLD_DIR%/tools/clang/test/taef_exec/lit.site.cfg %EXEC_ADAPTER% %HLSL_SRC_DIR%/tools/clang/test/taef_exec
```
hekota, line 316: Now the exec tests will run twice if EXEC_ADAPTER is set, once as part of the check-all and once with the specific adapter.

We could say specifying the adapter is not supported if running tests, it could only be supported when using hcttest exec ?

python3kgae: Fixed.

utils/hct/hcttest.cmd
python3kgae: Fixed.

utils/hct/hcttest.cmd
```diff
@@ -313,7 +313,7 @@ if "%TEST_USE_LIT%"=="1" (
     set RES_DXILCONV=!RES_CLANG!
     rem check exec.
     if defined EXEC_ADAPTER (
-        py %HLSL_SRC_DIR%/utils/lit/lit.py -v --no-progress-bar --param build_mode=%BUILD_CONFIG% --param clang_site_config=%HLSL_BLD_DIR%/tools/clang/test/lit.site.cfg --param clang_taef_exec_site_config=%HLSL_BLD_DIR%/tools/clang/test/taef_exec/lit.site.cfg %EXEC_ADAPTER% %HLSL_SRC_DIR%/tools/clang/test/taef_exec
+        echo -adapter is not support when running all tests.
```
hekota, line 316: Please change to:

The -adapter parameter is supported only when running just execution tests (hcttest.cmd exec).

python3kgae: Updated.

utils/hct/hcttest.cmd
```diff
@@ -313,7 +313,7 @@ if "%TEST_USE_LIT%"=="1" (
     set RES_DXILCONV=!RES_CLANG!
     rem check exec.
     if defined EXEC_ADAPTER (
-        echo The -adapter parameter is supported only when running just execution tests (hcttest.cmd exec)
+        echo "The -adapter parameter is supported only when running just execution tests (hcttest.cmd exec)"
```
hekota, line 316: You can escape the parenthesis with ^:

`echo Foo ^(^)`

hekota (approved):

tex3d (approved):

Issue #4594, opened 2022-08-14
[SPIRV] Incorrect error reported for non uint type template parameter: "error: alignment argument of vk::RawBufferLoad() must be a constant integer"


Cazadorro:
I'm trying to find the consequences of this feature, https://github.com/microsoft/DirectXShaderCompiler/pull/4226, thought it didn't work with template arguments, but it does, realizing that it only takes uint arguments.  Error reported made it seem like something completely different. 
```

struct StructTest{
    float a;
    float b; 
};



struct PhysicalDeviceAddress{
    uint64_t address; 
    template<typename T, const int alignment> 
    T load(){
        return vk::RawBufferLoad<T>(address, alignment);
    }
};


[numthreads(1, 1, 1)]
void main(uint tidx : SV_DispatchThreadId) {
    uint64_t test = 0; 
    PhysicalDeviceAddress address = PhysicalDeviceAddress(test); 
    StructTest x = address.load<StructTest, 8>(); 
}

```

tried in https://shader-playground.timjones.io/# using 

MicrosoftDXC, trunk, cs_6_6, entry point main, extra options: -HV 2021, output format SPIR-V, SPIR_V target vulkan 1.3

results in 

```
D:\local\Temp\9796b5d4-0e93-4f29-9a5a-883cbae55d48.hlsl:13:46: error: alignment argument of vk::RawBufferLoad() must be a constant integer
        return vk::RawBufferLoad<T>(address, alignment);
                                             ^
```

which is not the correct error.  The error **is** that I used a constant integer, instead of a constant unsigned integer.

pow2clk:
Tagging @Keenuts and @cassiebeckley for visibility.  This is a vulkan issue

natevm:
It seems like the solution to this could be to just change the compiler error message to 

“alignment argument of vk::RawBufferLoad() must be a constant unsigned integer”

Am I understanding correctly? That seems like an easy enough change.

devshgraphicsprogramming:
it seems a bit weird that `alignment` was added to the `RawBufferLoad` API as a function and not a template parameter, does LLVM 3.7 or DXC have a problem with `const int` default template args or what?

natevm:
I believe it’s discussed here in the referenced issue why a template wasn’t used. Something about template features not always being available for mobile architectures? Tbh I don’t know if I follow, seems like a default template argument would be better…

https://github.com/microsoft/DirectXShaderCompiler/pull/4226#issuecomment-1048883666

Issue #5638, opened 2023-08-31
[SPIR-V] Calling `var[0].member` as opposed to `var.operator[](0).member` made the compiler go SIGSEGV


devshgraphicsprogramming:
**Description**

Near the end of the shader there's a 
```cpp
// the following cause us to SIGSEGV
    member_ref_t<float[1],8,12> pB = pLineStyle[0].normalizedStipple;
```
comment this assignment out and it will no longer crash.

the funny thing is that I've coded `deref` and `operator[](0)` to work exactly the same, and I think that the problem might be in `operator[]`.

P.S. @natevm this works around all your issues with BDA, except for the atomics

**Steps to Reproduce**

Version with template + macro debauchery:
https://godbolt.org/z/4Gxc7rjoz

Verison with template and the macros already preprocessed:
https://godbolt.org/z/Pa119ej4n

**Actual Behavior**

Compiler Crashes without diagnostics

**Environment**
- DXC version trunk/Godbolt
- Host Operating System Linux/Godbolt

devshgraphicsprogramming:
OK I found a funny thing, without `-spirv` I get following error about `return this;`
```
error: cannot compile this aggregate expression yet
        return this;
               ^~~~
```
but I think thats a separate bug altogether

devshgraphicsprogramming:
ok but it definitely doesn't crash when emitting DXIL with `RWByteAddressBuffer` in place of BDA:
https://godbolt.org/z/KesGq3z44

but as soon as you add `-spirv` to the commandline it crashes

devshgraphicsprogramming:
Narrowed down the problem, this makes it compile fine
````cpp
    member_ref_t<uint32_t[1],8,12> pB = pLineStyle.operator[](0).normalizedStipple;
```

so this means that the problem is with `operator[]` calling

cassiebeckley:
I think there's at least three separate segfaults I've encountered while trying to create a minimal reproducible example for this. I've got a PR out to fix one of them, I'll look at the others when I have time.

cassiebeckley:
I came back to this to try to find the root cause, and I have found a smaller case that I think is touching on something close to what's going on here: https://godbolt.org/z/hnbznvz66. In this example, the fields of the base class are being ignored and not being initialized correctly. It might be an entirely separate issue, but the original case posted above fails in the debug build with an assert about initializers not being handled correctly, so I'll try fixing it first.

cassiebeckley:
[...]

PR #5653, opened 2023-09-05
[SPIR-V] Evaluate alignment argument for RawBufferLoad

cassiebeckley:
This fixes a separate segfault I encountered when looking at #5638.


sudonatalie:
I coincidentally just wrote almost the same patch today to fix #4594. I refactored slightly to fix RawBufferStore too and added 1 more test, so I'll still upload my PR now since that's probably easier making those same changes on this one.

sudonatalie:
What do you think of closing this in favor of #5654 ?

cassiebeckley:
@sudonatalie seems good :)

PR #5658, opened 2023-09-06
Pix: Fix crash getting type name of unnamed structs

jeffnn:
...and other types while I'm at it. There are tests for all of the possible unnamed types (structs, members, arrays). Don't think I can express an unnamed scalar or typededf so there's no test for those.
(This was hit in the wild while debugging a pixel shader in a WinPIX capture of a Large Well-Known Engine.)

### DIFF
--- a/lib/DxilDia/DxcPixTypes.cpp
+++ b/lib/DxilDia/DxcPixTypes.cpp
@@ -14,6 +14,13 @@
 #include "DxcPixTypes.h"
 #include "DxilDiaSession.h"
 
+static const char * GetTypeNameOrDefault(llvm::DIType *diType) {
+  auto stringRef = diType->getName();
+  if (stringRef.empty())
+    return "<unnamed>";
+  return stringRef.data();
+}
+
 HRESULT dxil_debug_info::CreateDxcPixType(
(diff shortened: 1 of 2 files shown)


tex3d (approved):
LGTM!

lib/DxilDia/DxcPixTypes.cpp
```diff
@@ -14,6 +14,13 @@
 #include "DxcPixTypes.h"
 #include "DxilDiaSession.h"
 
+static const char * GetTypeNameOrDefault(llvm::DIType *diType) {
```
pow2clk, line 17: Given that every use of this converts the output from ascii to widechar, perhaps this should return widechar instead?

pow2clk (approved):
Looks good. I just have one potential cleanup.

I think it would be clearer if you spent some time in the description explaining what you do test instead of just what you don't. I went into this thinking there were no tests at all!

lib/DxilDia/DxcPixTypes.cpp
jeffnn: Good call. Thanks.

PR #5661, opened 2023-09-06
spirv: validates parameters for InterlockedAddExch

### LINKED ISSUE #5647: [SPIR-V] InterlockedCompareExchange parameters are not validated
**Description**

DXC doesn't validates `InterlockedCompareExchange` params, and fails to report when an immediate is passed as an out param.
Instead we show a fatal error and ask to report the issue. We should produce a normal error.

**Steps to Reproduce**

```hlsl
groupshared uint value;

[numthreads(1, 1, 1)]
void main() {
  InterlockedCompareExchange(value, 1, 2, 3);
}
```

`dxc -T cs_6_6 -E main -spirv ~/repro.hlsl`

**Actual Behavior**

```
$ dxc -T cs_6_6 -E main -spirv ~/repro.hlsl
fatal error: generated SPIR-V is invalid: OpStore Pointer <id> '7[%uint_3]' is not a logical pointer.
  OpStore %uint_3 %12

```

**Environment**

- DXC version: HEAD (0f282ccabaf21c88e7dbfe0c866422842b639bc3)
- Host: linux

Keenuts:
This function takes a reference as 4th parameter. This was not validated by the backend, meaning we'd generate invalid SPIR-V and fail validation.

This PR adds a check to emit a fatal error if an immediate is used.

Fixes #5647

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -8875,7 +8875,9 @@ SpirvEmitter::processIntrinsicInterlockedMethod(const CallExpr *expr,
                                        uint32_t outputArgIndex) {
     const auto outputArg = callExpr->getArg(outputArgIndex);
     if (dyn_cast<DeclRefExpr>(outputArg) == nullptr) {
-      emitError("InterlockedCompareExchange requires a reference as output parameter", outputArg->getExprLoc());
+      emitError(
+          "InterlockedCompareExchange requires a reference as output parameter",
+          outputArg->getExprLoc());
       return;
     }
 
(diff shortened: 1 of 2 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -8874,6 +8874,11 @@ SpirvEmitter::processIntrinsicInterlockedMethod(const CallExpr *expr,
                                        const CallExpr *callExpr,
                                        uint32_t outputArgIndex) {
     const auto outputArg = callExpr->getArg(outputArgIndex);
+    if (dyn_cast<DeclRefExpr>(outputArg) == nullptr) {
+      emitError("InterlockedCompareExchange requires a reference as output parameter.", outputArg->getExprLoc());
```
sudonatalie, line 8878: nit: I think most of our error messages don't end in a period

sudonatalie (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -8874,6 +8874,11 @@ SpirvEmitter::processIntrinsicInterlockedMethod(const CallExpr *expr,
...
+      return;
```
sudonatalie, line 8879: Also return nullptr here

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts: fixed, removed the period.

PR #5663, opened 2023-09-06
spirv: remove logic to add ImageMSArray logic.

Keenuts:
Capability trimming pass is handling this logic now. Needed to fix a test as DCE now runs.
ImageMSArray is also not required for this test as no OpImageType has array=1, ms=1 and sampled=2.

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/type.texture.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/type.texture.hlsl
@@ -1,7 +1,6 @@
 // RUN: %dxc -T vs_6_0 -E main
 
 // CHECK: OpCapability Sampled1D
-// CHECK: OpCapability ImageMSArray
 
 // CHECK: %type_1d_image = OpTypeImage %float 1D 2 0 0 1 Unknown
 // CHECK: %_ptr_UniformConstant_type_1d_image = OpTypePointer UniformConstant %type_1d_image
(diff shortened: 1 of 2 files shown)


s-perron (approved):

sudonatalie:
Can you add a test that does require `OpCapability ImageMSArray`? It looks like this was our only one, so if it wasn't valid we're missing some coverage.

Keenuts:
> Can you add a test that does require `OpCapability ImageMSArray`? It looks like this was our only one, so if it wasn't valid we're missing some coverage.

Yeah, I wanted to do that, but haven't found how. I'll try a bit more, but seems every HLSL code I wrote doesn't have the `1 1 2` pattern in the OpImageType I need for this capability.

Keenuts:
Removed the capability, and fixed the documentation.

s-perron (approved):

tools/clang/test/CodeGenSPIRV/type.texture.hlsl
```diff
@@ -72,6 +73,19 @@ Texture1D<S>      sTex;
 // CHECK: %tTex = OpVariable %_ptr_UniformConstant_type_2d_image_array_1 UniformConstant
 Texture2DArray<T> tTex;
 
+// Just to prevent DCE.
```
sudonatalie, line 76: You should be able to remove all of the additions to this file since DCE isn't run on these tests

sudonatalie (approved):

tools/clang/test/CodeGenSPIRV/type.texture.hlsl
Keenuts: Right, I thought this was using lit so no `-fcgl`. Reverting.

Keenuts:
Arf auto-merge

PR #5666, opened 2023-09-06
Pix: Shader debug speedup

jeffnn:
Two related things:
-Move value/store annotations before instruction number annotations so that we can avoid giving instruction numbers to the allocas that the value-to-declare pass adds, since they are only there to implement PIX shader debugging and do not need to be debugged by the PIX end-user. And, otherwise, adding them messes up the instruction numbering for the next part:
-Add parameters that allow PIX to limit instrumentation to a smaller range of instruction numbers. This is a crude way to make instrumentation workable for very large (>100k or so instructions) shaders which otherwise will break the GPU driver either via out of memory or via TDRs when the instrumentation is run.

### DIFF
--- a/lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
+++ b/lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
@@ -160,8 +160,6 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
         if (auto *Alloca = llvm::dyn_cast<llvm::AllocaInst>(&I))
             if (PixAllocaReg::FromInst(Alloca, &unused1, &unused2))
                 continue;
-        if (PixDxilReg::FromInst(&I, &unused1))
-            continue;
         if (!llvm::isa<llvm::DbgDeclareInst>(&I)) {
           pix_dxil::PixDxilInstNum::AddMD(M.getContext(), &I, InstNum++);
           InstructionRangeEnd = InstNum;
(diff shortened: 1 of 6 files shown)


tex3d (approved):
LGTM!

python3kgae (approved):

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -133,11 +133,33 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
...
+      for (llvm::Instruction &I : block.getInstList()) {
+        AnnotateValues(&I);
+      }
+    }
+  }
+
+  for (auto *F : instrumentableFunctions) {
+    for (auto &block : F->getBasicBlockList()) {
+      for (llvm::Instruction &I : block.getInstList()) {
+        AnnotateStore(&I);
```
pow2clk, line 147: It looks as though AnnotateStore might depend on the metadata set by AnnotateValues above. Is that correct? Given how egregious the repeated looping looks, it might be worth mentioning in a comment if that's the case.

Even if it is the case, I think it's unlikely that there is ever a store with arguments that haven't been annotated even with one loop. I don't know if I'd risk it without more certainty than I have, but since this change is meant to speed things up, this might be another way to help there.

pow2clk (approved):
Just one question. It's no worse than it was for sure

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
tex3d: The performance concern is around compiling and running the instrumented shader (driver compile and runtime performance), rather than the performance of this pass.

jeffnn: Ach I went and completed the PR before noticing this comment. Darn github timeout. 

Yes, what Tex said, but yes a comment is a good idea. I'll be back in amongst this code soon, I think (for typedefs), so I'll fix it up then.

PR #5671, opened 2023-09-07
PIX: Use CA2W in original fashion, to avoid build issue on release branch

jeffnn:
This should fix a build issue in the 2380 release branch.

### DIFF
--- a/lib/DxilDia/DxcPixTypes.cpp
+++ b/lib/DxilDia/DxcPixTypes.cpp
@@ -14,11 +14,11 @@
 #include "DxcPixTypes.h"
 #include "DxilDiaSession.h"
 
-static CA2W GetTypeNameOrDefault(llvm::DIType *diType) {
+static const char * GetTypeNameOrDefault(llvm::DIType *diType) {
   auto stringRef = diType->getName();
   if (stringRef.empty())
-    return CA2W("<unnamed>");
-  return CA2W(stringRef.data());
+    return "<unnamed>";
+  return stringRef.data();
 }
 
 HRESULT dxil_debug_info::CreateDxcPixType(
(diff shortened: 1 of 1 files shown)


lib/DxilDia/DxcPixTypes.cpp
```diff
@@ -14,11 +14,11 @@
...
 #include "DxilDiaSession.h"
 
-static CA2W GetTypeNameOrDefault(llvm::DIType *diType) {
+static const char * GetTypeNameOrDefault(llvm::DIType *diType) {
   auto stringRef = diType->getName();
   if (stringRef.empty())
-    return CA2W("<unnamed>");
-  return CA2W(stringRef.data());
+    return "<unnamed>";
+  return stringRef.data();
```
pow2clk, line 21: This seems to be throwing out wide chars entirely where previously they were used in the places this function is called below. Is that okay?

jeffnn: Amusingly, CComBSTR has its own widening constructor, and that's what's getting called here. There was no need to invoke CA2W at all.

tex3d: FWIW, I don't think the WinAdapter version of CComBSTR has such a constructor, though this may never be compiled on non-win32.

jeffnn: (Now that I know about WinAdapter) yes, that's true. Well maybe I should go back to my original formulation where each caller uses CA2W. To answer Greg's question about why it doesn't return a wchar_t *: it's because CA2W (or the original CComBstr's widening constructor) allocates memory.

tex3d (approved):
LGTM!

PR #5678, opened 2023-09-09
Add outputs to the IDxcResult when linking

dmpots:
This change modifies the IDxcLinker::Link api to return outputs via the
IDxcResult interface. We now support getting the following outputs from the
IDxcResult returned from the call to Link

  DXC_OUT_REFLECTION
  DXC_OUT_SHADER_HASH
  DXC_OUT_ERRORS
  DXC_OUT_ROOT_SIGNATURE

Additionally, the DXC_OUT_OBJECT output will now set its name based on the `/Fo`
command line flag.

This makes the api more consistent with the Compile output and makes it easier
to access these outputs without having to crack open the dxil container.

Original patch by @tex3d. Tests added by @dmpots.

### DIFF
--- a/tools/clang/tools/dxcompiler/dxclinker.cpp
+++ b/tools/clang/tools/dxcompiler/dxclinker.cpp
@@ -284,6 +284,14 @@ HRESULT STDMETHODCALLTYPE DxcLinker::Link(
       return S_OK;
     }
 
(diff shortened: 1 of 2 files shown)


python3kgae (approved):

tex3d (approved):
LGTM!

PR #5679, opened 2023-09-09
Disallow multiple shader attributes from existing on a single declaration

### LINKED ISSUE #5601: Disallow multiple different HLSLShader attributes on the same function
**Description**
The work graphs spec has been updated to disallow marking shaders as both `compute` and `node`:
https://github.com/microsoft/hlsl-specs/pull/62/commits/48c683ee09da2451e1d551c1777f2ed7c6f00ef8

This will allow us to simplify aspects of our diagnostic implementation.

**Environment**
- DXC version https://github.com/microsoft/DirectXShaderCompiler/tree/staging-sm-6.8

bob80905:
There was a recent spec change on work-graphs that disallows the existence of node + compute attributes on shader declarations.  Because of this old behavior, a lot of code worked around the possibility that 2 shader attributes could exist for one declaration. With this old behavior disallowed, we no longer need to account for this possibility. The PR now adds a diagnostic preventing multiple attributes from existing. It will also change all the tests that worked with shaders with multiple shader attributes, and either remove them, or update them so that only one attribute is used.
Fixes #5601


python3kgae (approved):

pow2clk (approved):
Looks good. We already have a comment suggesting this be done earlier and there are some pending efforts in that direction

PR #5683, opened 2023-09-10
Fix compilation issues on clang

gongminmin:
ErrorOr has a implicit constructor. A move is required for construct an ErrorOr object. Then it's returned by the function. Write the construction explicit to fix the compilation on clang, and reduce the confusion.


python3kgae (approved):

pow2clk (approved):
Looks good

PR #5684, opened 2023-09-10
Fix build on ARM with non-VS generator

gongminmin:
CMAKE_GENERATOR_PLATFORM is not widely available. For example, Ninja generator doesn't define it. Use CMAKE_C_COMPILER_ARCHITECTURE_ID to avoid this problem.


hekota (approved):
LGTM!

PR #5693, opened 2023-09-12
[ClangCL] Fix warning for DxilDia

python3kgae:
1. add override for override methods.
2. remove unnecessary std::move
3. remove unused lambda capture
4. use same enum when compare.
5. fix field order in constructor.
6. add const for when cast from a const variable.

### DIFF
--- a/lib/DxilDia/DxilDiaDataSource.cpp
+++ b/lib/DxilDia/DxilDiaDataSource.cpp
@@ -61,9 +61,9 @@ getMemBufferFromStream(IStream *pStream, const llvm::Twine &BufferName) {
   size_t size = statstg.cbSize.LowPart;
   std::unique_ptr<llvm::MemoryBuffer> result(
       llvm::MemoryBuffer::getNewUninitMemBuffer(size, BufferName));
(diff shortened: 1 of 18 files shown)


tex3d (approved):
LGTM.

bogner (approved):

hekota (approved):
LGTM!

lib/DxilDia/DxilDiaDataSource.cpp
```diff
@@ -61,9 +61,9 @@ getMemBufferFromStream(IStream *pStream, const llvm::Twine &BufferName) {
   size_t size = statstg.cbSize.LowPart;
   std::unique_ptr<llvm::MemoryBuffer> result(
       llvm::MemoryBuffer::getNewUninitMemBuffer(size, BufferName));
```
pow2clk, line 63: Rather than create this uninitialized, we could create an allocation of some kind to read the stream into and then create the membuffer using a stringref to that allocation.  That would prevent the need to cast the const off on line 66.

pow2clk (approved):
Looks great! Thanks for indulging me

PR #5700, opened 2023-09-12
[NFC] Fix ubsan failure

llvm-beanz:
This variable was never initialized so it was always whatever garbage bits were left there. I've defaulted it to `true` because a bunch of tests fail if it is set to `false`, but all tests pass if it is true.

It would be great if someone with more SPIR-V experience could take a look at this.


llvm-beanz:
@sudonatalie, you're right. I must have not had a clean build directory when I ran the tests.

sudonatalie (approved):

sudonatalie:
Thanks for fixing this up!

s-perron (approved):

PR #5705, opened 2023-09-13
[ClangCL] Fix warning for rest part except clang.

python3kgae:
1. add noexcept for methods with __declspec(nothrow)
2. add override for override methods.
3. add const when cast const ptr for a const argument.
4. change 0 to 0u for immediate to avoid signed unsigned mismatch.

### DIFF
--- a/lib/DxilCompression/miniz.c
+++ b/lib/DxilCompression/miniz.c
@@ -2125,11 +2125,10 @@ void *tdefl_write_image_to_png_file_in_memory_ex(const void *pImage, int w, int 
         tdefl_output_buffer_putter(&z, 1, &out_buf);
     /* compress image data */
     tdefl_init(pComp, tdefl_output_buffer_putter, &out_buf, s_tdefl_png_num_probes[MZ_MIN(10, level)] | TDEFL_WRITE_ZLIB_HEADER);
-    for (y = 0; y < h; ++y) {
+    for (y = 0; y < h; ++y)
+    {
         tdefl_compress_buffer(pComp, &z, 1, TDEFL_NO_FLUSH);
-        tdefl_compress_buffer(
-            pComp, (const mz_uint8 *)pImage + (flip ? (h - 1 - y) : y) * bpl,
-            bpl, TDEFL_NO_FLUSH);
+        tdefl_compress_buffer(pComp, (const mz_uint8 *)pImage + (flip ? (h - 1 - y) : y) * bpl, bpl, TDEFL_NO_FLUSH);
     }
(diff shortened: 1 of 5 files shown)


lib/DxilCompression/miniz.c
```diff
@@ -2125,10 +2125,11 @@ void *tdefl_write_image_to_png_file_in_memory_ex(const void *pImage, int w, int
         tdefl_output_buffer_putter(&z, 1, &out_buf);
     /* compress image data */
     tdefl_init(pComp, tdefl_output_buffer_putter, &out_buf, s_tdefl_png_num_probes[MZ_MIN(10, level)] | TDEFL_WRITE_ZLIB_HEADER);
-    for (y = 0; y < h; ++y)
-    {
+    for (y = 0; y < h; ++y) {
         tdefl_compress_buffer(pComp, &z, 1, TDEFL_NO_FLUSH);
-        tdefl_compress_buffer(pComp, (mz_uint8 *)pImage + (flip ? (h - 1 - y) : y) * bpl, bpl, TDEFL_NO_FLUSH);
+        tdefl_compress_buffer(
+            pComp, (const mz_uint8 *)pImage + (flip ? (h - 1 - y) : y) * bpl,
```
pow2clk, line 2131: The rest of this file isn't getting clang-formatted because not only does it not belong to DXC, it doesn't belong to LLVM either.  I'm not saying we shouldn't fix the warning here, but I don't think we should clang format the parts you've changed if only to minimize the changes.

unittests/Support/ProgramTest.cpp
```diff
@@ -188,7 +188,12 @@ TEST(ProgramTest, TestExecuteNoWait) {
   ProcessInfo PI1 = ExecuteNoWait(Executable, argv, &envp[0], nullptr, 0,
                                   &Error, &ExecutionFailed);
   ASSERT_FALSE(ExecutionFailed) << Error;
+#ifdef _WIN32
+  // Pid is unsigned on windows.
+  ASSERT_NE(PI1.Pid, 0u) << "Invalid process id";
+#else
   ASSERT_NE(PI1.Pid, 0) << "Invalid process id";
+#endif
```
pow2clk, line 196: I prefer how upstream solved this: https://github.com/llvm-mirror/llvm/commit/a58e071c716c1f7c639ed63758c83b14a5ee5f61

pow2clk (commented):
A couple notes

unittests/Support/ProgramTest.cpp
python3kgae: Merged.

hekota (approved):

pow2clk (approved):
Thanks for the responses!

PR #5708, opened 2023-09-13
[ClangCL] fix warning for clang unit test.

python3kgae:
1. remove unused variable.
2. use 1u instead 1 to avoid signed unsigned mismatch.
3. use std::size to replace _countof.
4. Use LPCWSTR/LPCSTR instead of LPWSTR/LPSTR.

### DIFF
--- a/tools/clang/unittests/HLSLHost/HLSLHost.cpp
+++ b/tools/clang/unittests/HLSLHost/HLSLHost.cpp
@@ -121,7 +121,6 @@ public:
 };
 
 // Globals.
-static DWORD g_ProcessLocks;
 static const GUID
     CLSID_HLSLHostServer = // {7FD7A859-6C6B-4352-8F1E-C67BB62E774B}
     {0x7fd7a859,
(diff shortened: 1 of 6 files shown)


tools/clang/unittests/HLSLHost/HLSLHost.cpp
```diff
@@ -827,16 +821,16 @@ class ServerFactory : public IClassFactory {
 public:
   DXC_MICROCOM_ADDREF_RELEASE_IMPL(m_dwRef)
   ServerFactory() : m_dwRef(0) {}
-  HRESULT STDMETHODCALLTYPE QueryInterface(REFIID iid, void **ppvObject) {
+  HRESULT STDMETHODCALLTYPE QueryInterface(REFIID iid, void **ppvObject) override {
```
bob80905, line 824: For these override keywords, what warning went away when adding override?

python3kgae: https://clang.llvm.org/docs/DiagnosticsReference.html#winconsistent-missing-override

bob80905 (approved):

pow2clk (approved):
Looks good! Just needs to be rebased

PR #5711, opened 2023-09-13
Indicate extra metadata with unknown nodeio tag

### LINKED ISSUE #5609: Unknown NodeIO properties tag should generate a validation error
**Description**
If the metadata node IO Properties tag should somehow contain an unknown value, the compiler does not produce a validation error. 

We also need tests for this and various other similar errors

**Steps to Reproduce**
Not reproducible with the compiler itself. Would require writing DXIL by hand that uses an invalid integer for the node IO tag associated with the node shader entry point. 

**Actual Behavior**
Such a shader does not produce a validation error.

**Environment**
- DXC version 1.8.2306-preview
- Host Operating System win11

pow2clk:
When unrecognized metadata is encountered, it needs to be recorded so that validation will fail. Additionally, this removes the assert that was there so that the code can be tested. Asserts for code that can be encountered through potential input isn't ideal anyway.

Adds testing for the same which involves introducing dxv testing into the llvm test directory requiring enabling .ll tests for that subdirectory and enabling lit to understand %dxv.

Incidentally fixes some comments in unrelated workgraph tests that had outdated values for some of the metadata that might lead to confusion

Fixes #5609

### DIFF
--- a/test/DXILValidation/Metadata/known_nodeio_tags.ll
+++ b/test/DXILValidation/Metadata/known_nodeio_tags.ll
@@ -1,0 +1,84 @@
+; RUN: %dxv %s | FileCheck %s
+
+; Test that all valid node IO tags are accepted
+
+; CHECK: Validation succeeded.
+
+; shader hash: df03be6421edfc5b5480ddc4a4f8b7e3
+;
+; Buffer Definitions:
+;
+;
+; Resource Bindings:
+;
+; Name                                 Type  Format         Dim      ID      HLSL Bind  Count
+; ------------------------------ ---------- ------- ----------- ------- -------------- ------
+;
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+define void @main() {
+  ret void
+}
+
+!llvm.ident = !{!0}
+!dx.version = !{!1}
+!dx.valver = !{!1}
+!dx.shaderModel = !{!2}
+!dx.typeAnnotations = !{!3}
+!dx.entryPoints = !{!7, !8}
+
+!0 = !{!"dxc(private) 1.7.0.4023 (update-dndxc, 15d086f68-dirty)"}
+!1 = !{i32 1, i32 8}
+!2 = !{!"lib", i32 6, i32 8}
+!3 = !{i32 1, void ()* @main, !4}
+!4 = !{!5}
+!5 = !{i32 0, !6, !6}
+!6 = !{}
+!7 = !{null, !"", null, null, null}
+!8 = !{void ()* @main, !"main", null, null, !9}
+!9 = !{i32 8, i32 15, i32 13, i32 1, i32 14, i1 true, i32 15, !10, i32 16, i32 -1, i32 18, !11, i32 20, !12, i32 21, !15, i32 4, !20, i32 5, !21}
+!10 = !{!"main", i32 0}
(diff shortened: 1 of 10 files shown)


test/DXILValidation/Metadata/known_nodeio_tags.ll
```diff
@@ -0,0 +1,84 @@
+; RUN: %dxv %s | FileCheck %s
+
+; Test that all valid node IO tags are accepted
+
+; CHECK: Validation succeeded.
+
+; shader hash: df03be6421edfc5b5480ddc4a4f8b7e3
```
python3kgae, line 7: Do we need things like shader hash/Buffer Definitions in this test?

test/DXILValidation/lit.local.cfg
```diff
@@ -0,0 +1,3 @@
+# Enable .ll suffixes for this directory
+# They are disabled above to avoid tests unsupported by HLSL
+config.suffixes = ['.txt', '.td', '.test', '.ll']
```
python3kgae, line 3: why .td here?

tools/clang/test/HLSLFileCheck/hlsl/workgraph/case001_dispatchgrid_shader.hlsl
```diff
@@ -65,9 +65,9 @@ void node001_dispatchgrid_shader(DispatchNodeInputRecord<INPUT_NOGRID> input)
 // CHECK: [[DISPATCHGRID]] = !{i32 2, i32 3, i32 2}
 
 // NodeInputs
-// Arg #1: NodeIOKind Tag (1)
-// Arg #2: InputRecord(97)
-// Arg #3: NodeRecordType Tag (3)
+// Arg #1: NodeIOFlags Tag (1)
```
python3kgae, line 68: Why changes happen here?
Because  m_bExtraMetadata is set to true?

test/DXILValidation/Metadata/unknown_nodeio_tags.ll
```diff
@@ -0,0 +1,84 @@
+; RUN: not %dxv %s | FileCheck %s
```
bob80905, line 1: I'd like to ask why the `not` command is necessary here? Why not just `%dxv`?

test/DXILValidation/lit.local.cfg
pow2clk: Just matching what's already in the upper directory's `test/lit.cfg` since the intent here is only to add .ll rather than change anything else

test/DXILValidation/Metadata/unknown_nodeio_tags.ll
pow2clk: This is expected to fail validation. That makes the result of the dxv invocation produce a failing errorcode. If that happens, the test fails outright and won't pipe the results into filecheck at all.

bob80905: There are other examples of validation failing and returning bad exit codes, but the output still gets piped to filecheck. Is this required just because of how lit works?

python3kgae: Yes. taef FileCheck will take care the failure.
But lit FileCheck test will need to use 'not'.

test/DXILValidation/lit.local.cfg
python3kgae: If we only care .ll in this directory, maybe config.suffixes = ['.ll'] is what we want.

python3kgae (approved):

tex3d (approved):
LGTM!

PR #5712, opened 2023-09-13
Improve diagnostics for objects in node records

### LINKED ISSUE #5353: Make sure that objects cannot be inside the node record structs and add tests for it
Make sure that objects cannot be inside the node record structs, such as a texture field inside the struct, and add tests for it.

### LINKED ISSUE #5729: Node record should allow templated struct/class types
**Description**
It should be possible to use a suitable templated struct/class as a node input/output record type.
Currently any templated struct/class is errored as having zero size.

**Steps to Reproduce**
Simple example attached (as a .gz file as .hlsl is not accepted)
[node_record_template.hlsl.gz](https://github.com/microsoft/DirectXShaderCompiler/files/12618948/node_record_template.hlsl.gz)

**Actual Behavior**
The semantic check for a zero size record fails to handle templated types, and reports a zero sized record.
If that check is disabled, then asserts are hit during CodeGen, so bugs are present in both Sema and CodeGen.

**Environment**
- DXC version staging-sm-6.8 branch 
- Host Operating System any

tcorringham:
The diagnostics produced for the use of objects within node records were rather generic and obscure.
Improved the diagnosis of such issues by adding more specific diagnostics, and moved exisiting verifier tests from a catch-all test case to a specific test for invalid node record types and added more test cases.

It transpired that templated types within a node record are problematic - they are not currently diagnosed correctly and produce a zero size record diagnostic, and if this is bypassed they hit asserts during CodeGen. So as a lesser evil, an error diagnostic has been added to disallow templated types for now. (**_This is now fixed_**)

Fixes #5353 
Fixes #5729


tools/clang/test/HLSL/workgraph/invalid_node_record_type.hlsl
```diff
@@ -0,0 +1,160 @@
...
+typedef BAD_RECORD2 MyBadRecord;
+
+//==============================================================================
+// Check diagnostics for the various node input/output types
+
+[Shader("node")]
+[NodeLaunch("broadcasting")]
+[NumThreads(8,1,1)]
+[NodeMaxDispatchGrid(8,1,1)]
+void node01(DispatchNodeInputRecord<int> input) // expected-error {{'int' is not valid as a node record type}}
```
llvm-beanz, line 36: The message doesn't tell me why `int` isn't valid as a node record type. The old message might not have been clear, but it at least explained that a class or struct type was required. Can we adjust the wording?

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7831,6 +7832,12 @@ def err_hlsl_maxrecursiondepth_exceeded : Error<
    "NodeMaxRecursionDepth may not exceed 32">;
 def err_hlsl_too_many_node_inputs : Error<
    "Node shader '%0' may not have more than one input record">;
+def err_hlsl_node_record_type : Error<
+   "%0 is not valid as a node record type">;
+def err_hlsl_node_record_object : Error<
+   "object %0 may not appear in a node record">;
+def err_hlsl_node_record_template : Error<
+   "templated types are not currently supported in a node record">;
```
llvm-beanz, line 7840: This actually seems extremely limiting and I don't recall this being in the spec. This seems more like a bug than something we should be diagnosing on.

tools/clang/test/HLSL/workgraph/invalid_node_record_type.hlsl
tcorringham: Can do.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
tcorringham: It is indeed a bug - or rather several bugs (one with  this implementation, and several in CodeGen) . Until these bugs are fixed I consider the (temporary) diagnostic more friendly than allowing subsequent obscure asserts.

llvm-beanz: Thats fair. I don't think we have an issue tracking that. Can you file one?

tcorringham: I have created issue #5729 created to track this issue.

external/SPIRV-Headers
tex3d, line 1: I don't think you updated these submodules on purpose.  They shouldn't be part of the PR.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11349,11 +11408,11 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
-  // Check RT shader if available for their payload use and match payload access
-  // against availiable payload modifiers.
-  // We have to do it late because we could have payload access in a called
-  // function and have to check the callgraph if the root shader has the right
-  // access rights to the payload structure.
+  // Check RT shader if available for their payload use and match payload
+  // access against availiable payload modifiers. We have to do it late
+  // because we could have payload access in a called function and have to
+  // check the callgraph if the root shader has the right access rights to
+  // the payload structure.
```
tex3d, line 11415: Why the formatting changes in previously unmodified code?

tex3d: There are a ton of these, and it makes it hard to review the substantive changes.

tex3d (approved):
LGTM!

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d: Fixed!

python3kgae (approved):

Issue #5682, opened 2023-09-10
Install error


bmarques1995:
**Description**
When you try to install the dxc, an error is occuring

**Steps to Reproduce**

build.cmd: 
`
REM on dxc directory

set current_dir=%cd%

call ./utils/hct/hctstart %current_dir% %current_dir%\dependencies\dxc
call ./utils/hct/hctbuild -Release -no-parallel -installdir "%current_dir%\install"

cmake --build ./dependencies/dxc --config Release --target install
`

**Actual Behavior**
  CMake Error at tools/llvm-as/cmake_install.cmake:39 (file):
    file INSTALL cannot find
    "F:/cpp/dxc_install/modules/dxc/dependencies/dxc/Release/bin/llvm-as.exe":
    File exists.
  Call Stack (most recent call first):
    tools/cmake_install.cmake:47 (include)
    cmake_install.cmake:117 (include)
  
  
C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(161,5): error MSB3073: O comando "setlocal [F:\cpp\dxc_install\modules\dxc\dependencies\dxc\install.vcxproj]
C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(161,5): error MSB3073: "C:\Program Files\CMake\bin\cmake.exe" -DBUILD_TYPE=Release -P cmake_install.cmake [F:\cpp\dxc_install\modules\dxc\dependencies\dxc\install.vcxproj]
C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(161,5): error MSB3073: if %errorlevel% neq 0 goto :cmEnd [F:\cpp\dxc_install\modules\dxc\dependencies\dxc\install.vcxproj]
C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(161,5): error MSB3073: :cmEnd [F:\cpp\dxc_install\modules\dxc\dependencies\dxc\install.vcxproj]
[...]

PR #5713, opened 2023-09-14
Removed OS condition to install DXCompiler

bmarques1995:
added the archive(.lib) file to dxcompiler install


llvm-beanz (commented):
I think the more correct fix to this bug is to change the `add_clang_library` macro defined in `tools/clang/CMakeLists.txt` to treat `dxcompiler` the same way it treats `libclang`.

llvm-beanz:
I believe this is intended to fix #5682.

bmarques1995:
> I believe this is intended to fix #5682.

you're correct, i will look this

bmarques1995:
> I believe this is intended to fix #5682.

By the way, are you one of the analysers of this issue? I believe u've looked that I made it

bmarques1995:
> I believe this is intended to fix #5682.

I've modified my PR, now I removed the OS condition to install target, this is unnecessary, cmake automaticly manages the binaries(runtime [executable and DLLs], library [static libs on Windows, or all libs on another distros] and archive[symbols{...dll.lib}]) of all OS, depending the target

bmarques1995:
> @bmarques1995 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

llvm-beanz (changes requested):
This is much more complicated than it needs to be. You should be able to add dxcompiler here:

https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/CMakeLists.txt#L377


The stuff you're doing with the `find_*` is also not correct. I don't think we should make a DXC-specific CMake exports list because ultimately DXC is just a fork of LLVM. I think it would be better to connect these through the existing export mechanism for LLVM targets.

bmarques1995:
> This is much more complicated than it needs to be. You should be able to add dxcompiler here:
> 
> https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/CMakeLists.txt#L377
> 
> The stuff you're doing with the `find_*` is also not correct. I don't think we should make a DXC-specific CMake exports list because ultimately DXC is just a fork of LLVM. I think it would be better to connect these through the existing export mechanism for LLVM targets.

this is the only problem or don't you agree also with the new runtime syntax? if is this the case i discard the change of find_package export, but the first one for me is critical, to move the install syntax

PR #5717, opened 2023-09-14
[lit] set clang_taef_site_config in llvm-lit.py

python3kgae:
This will make it possible to run taef test from llvm-lit.py directly
 without set clang_taef_site_config and clang_taef_exec_site_config.


bob80905 (approved):

pow2clk (approved):
Very convenient

PR #5718, opened 2023-09-14
Fix ASAN memory leak in FileChecker

amaiorano:
AllocVFSIncludeHandler was returning it's IUnknown object by Detach()ing from a CComPtr, then assiging to CComPtr. This would result in the object have a ref count of 2, and when the second CComPtr destructed, it would go down to 1, and leak memory.

The fix is to return the object as a CComPtr; however, with Clang, this failed to compile because it finds no conversion from a CComPtr of child class type to a CComPtr of base class type. This compiles fine on MSVC, so I debugged it, and realized that MSVC compiles this as first a conversion to T*, then a constructor call. I suspect MSVC is wrong here. In any case, this change adds a template conversion constructor to CComPtr, very much like the existing template assignment operator.

### DIFF
--- a/include/dxc/WinAdapter.h
+++ b/include/dxc/WinAdapter.h
@@ -748,6 +748,13 @@ public:
     return *this;
   }
 
+  // NOTE: This conversion constructor is not part of the official CComPtr spec;
+  // however, it is needed to convert CComPtr<Q> to CComPtr<T> where T derives from Q on Clang.
+  // MSVC compiles this conversion as first a call to CComPtr<Q>::operator T*, followed by CComPtr<T>(T*),
+  // but Clang fails to compile with error: no viable conversion from 'CComPtr<Q>' to 'CComPtr<T>'.
+  template <typename Q>
+  CComPtr(const CComPtr<Q> &lp) throw() : CComPtrBase<T>(lp.p) {}
+
   T *operator=(const CComPtr<T> &lp) throw() {
     if (*this != lp) {
       CComPtr(lp).Swap(*this);
--- a/tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
+++ b/tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
@@ -256,7 +256,7 @@ public:
   }
 };
 
-static IncludeHandlerVFSOverlayForTest *AllocVFSIncludeHandler(IUnknown *pUnkLibrary, const FileMap *pVFS) {
+static CComPtr<IncludeHandlerVFSOverlayForTest> AllocVFSIncludeHandler(IUnknown *pUnkLibrary, const FileMap *pVFS) {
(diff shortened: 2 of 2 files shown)


include/dxc/WinAdapter.h
```diff
@@ -748,6 +748,13 @@ template <class T> class CComPtr : public CComPtrBase<T> {
     return *this;
   }
 
+  // NOTE: This conversion constructor is not part of the official CComPtr spec;
+  // however, it is needed to convert CComPtr<Q> to CComPtr<T> where T derives from Q on Clang.
+  // MSVC compiles this conversion as first a call to CComPtr<Q>::operator T*, followed by CComPtr<T>(T*),
```
amaiorano, line 753: I wonder if this has something to do with MSVC's two-phase lookup. I say this because to be able to build DXC using Clang-CL, I had to pass in `/Zc:twoPhase-`  (see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/third_party/gn/dxc/BUILD.gn;l=473?q=dxc%20BUILD.gn&ss=chromium)).

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
```diff
@@ -256,7 +256,7 @@ class IncludeHandlerVFSOverlayForTest : public IDxcIncludeHandler {
   }
 };
 
-static IncludeHandlerVFSOverlayForTest *AllocVFSIncludeHandler(IUnknown *pUnkLibrary, const FileMap *pVFS) {
+static CComPtr<IncludeHandlerVFSOverlayForTest> AllocVFSIncludeHandler(IUnknown *pUnkLibrary, const FileMap *pVFS) {
```
tex3d, line 259: Another solution would have been to modify users of `AllocVFSIncludeHandler` to properly `Attach()` to the pointer, since that's how the `Alloc*` function should be treated.

amaiorano: Would you prefer that?

amaiorano: If I do that, then we shouldn't need the conversion constructor template anymore, though I wonder if we'll hit this again in the future, since it's perfectly valid to use CComPtr that way.

tex3d: I think the conversion constructor is fine to keep.  I don't feel too strongly about the approach here, since either way is correct, and it is a static function in test support code.

tex3d (approved):
Looks good to me.

s-perron (commented):
I'll refrain from approving this because I do not know what this code is doing. I'll leave it to the Microsoft developers to review.

amaiorano:
> I'll refrain from approving this because I do not know what this code is doing. I'll leave it to the Microsoft developers to review.

No problem. We'll let @pow2clk review. Thanks.

PR #5720, opened 2023-09-14
[lit] allow for spaces in taef commands

### LINKED ISSUE #5719: lit silently skips taef tests when there is a space in the invocation command
**Description**
When running taef testing with lit, a large swath of tests are silently being skipped because my te.exe is in a path with a space in it ("Program Files (x86)" of course). When the command arguments are flattened, the execution interprets the space to mean the end of an argument. 

**Steps to Reproduce**
make sure the te.exe in your path has a space in it and then invoke lit either using the python script or hcttest. I used hcttest. 

**Actual Behavior**
The only manifestation of the error is the shorter lists of tests and that all the TAEF tests are missing. I would typically look for CompilerTests to tell.

The only other indication is that there is a warning saying: 
``` 
lit.py: discovery.py:192: warning: test suite 'clang-taef' contained no tests
```

**Environment**
- DXC version 1.7.2308
- Host Operating System win11

pow2clk:
The way taef tests were gathered and executed previously flattened the list of arguments to a string. This allowed the weird /select: flag to work correctly, but if any of the arguments had spaces, they were misinterpretted having lost the implicit quotes around each.

By removing the flattening, putting a space between the flag and the argument for /select: and removing the explicit quotes around that argument, the command executes properly whether there are spaces or not

Additionally added a way to detect when these commands failed since they were silently failing and skipping a large swath of testing when spaces were present.

Fixes #5719

### DIFF
--- a/utils/lit/lit/formats/taef.py
+++ b/utils/lit/lit/formats/taef.py
@@ -13,7 +13,6 @@ from .base import TestFormat
 # because of the way it manually processes quoted arguments in a
 # non-standard way.
 def executeCommandForTaef(command, cwd=None, env=None):
-    command = ' '.join(command)
     p = subprocess.Popen(command, cwd=cwd,
                         shell=True,
                         stdin=subprocess.PIPE,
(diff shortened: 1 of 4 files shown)


utils/lit/lit/formats/taef.py
```diff
@@ -65,17 +64,20 @@ def getTaefTests(self, dll_path, litConfig, localConfig):
...
             lines = lines.replace('\r', '')
             lines = lines.split('\n')
 
         except:
-            litConfig.error("unable to discover taef tests in %r, using %s" % (dll_path, self.te))
+            litConfig.error("unable to discover taef tests in %r, using %s. exeption encountered." % (dll_path, self.te))
+            raise StopIteration
+
+        if exitCode:
+            litConfig.error("unable to discover taef tests in %r, using %s. error: %s." % (dll_path, self.te, err))
```
pow2clk, line 80: I chose to raise an exception independently to avoid the potentially confusing error that an exception was encountered while handling the exception.  It also makes printing the error message only when present easier

python3kgae (approved):

hekota (approved):
LGTM!

tex3d (approved):
This is a good solution!

pow2clk:
dxilconv needs some adaptation to work with the new code. I'll have that soon

hekota (approved):

PR #5722, opened 2023-09-14
Remove implicit node records from being added when there are no node inputs

### LINKED ISSUE #5375: Remove implicit record types generated by compiler
Implicit input record types generated by the compiler are problematic, and a weird special case.  We can't differentiate an explicit record vs. implicit one at the DXIL level, which complicates DXIL validation.

It seems like the runtime and driver should be able to handle this case fine without a compiler generated implicit input record.  I think we should revisit this.

bob80905:
Previously, when a function declaration defined as a node shader did not have any defined node inputs, a default node input record would be created and added to the function properties' node inputs. This behavior is undesired. This PR removes the behavior of adding implicit node input records to node functions, and instead leaves the node inputs property empty. Some metadata in tests were updated as well.
Fixes #5375


tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4066,18 +4066,18 @@ TEST_F(ValidationTest, ComputeNodeCompatibility) {
     "Node node01 Coalescing launch type is not compatible with Compute");
   RewriteAssemblyCheckMsg(L"..\\DXILValidation\\compute_node_compatibility.hlsl", "lib_6_8",
     pArguments.data(), 2, nullptr, 0,
-    {"!18 = !{i32 8, i32 15"},  // original: node shader
-    {"!18 = !{i32 8, i32 5"},   // changed to: compute shader
+    {"!16 = !{i32 8, i32 15"},  // original: node shader
+    {"!16 = !{i32 8, i32 5"},   // changed to: compute shader
```
tex3d, line 4070: Should probably avoid having the numbered metadata being assigned to in the first place if possible.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/implicit_record_dispatchgrid.hlsl
```diff
@@ -1,30 +1,15 @@
 // RUN: %dxc -T lib_6_8 %s | FileCheck %s
 // RUN: %dxc -T lib_6_8 %s | %D3DReflect %s | FileCheck -check-prefix=REFL %s
 
+// The purpose of this test is to verify the behavior of the compiler on node
+// shaders with no input records
```
tex3d, line 5: Yes, and I think we now need to fail this case when no input record is defined in the HLSL entry, yet it is necessary to provide the `SV_DispatchGrid`, due to the use of `[MaxNodeDispatchGrid(...)]`.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2449,24 +2449,6 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
     }
   }
 
-  // If InputNodes is empty, add an implicit input.
```
tex3d, line 2452: We need to chek whether `SV_DispatchGrid` is necessary and fail if no input with that semantic is provided.  It's necessary when `funcProps->Node.MaxDispatchGrid[0] > 0` (`[NodeMaxDispatchGrid(...)]` was used).

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2448,22 +2448,24 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
-      funcProps->InputNodes.push_back(defaultInput);
-    } else {
-      hlsl::NodeIOProperties emptyInput(DXIL::NodeIOKind::EmptyInput);
-      funcProps->InputNodes.push_back(emptyInput);
+  
+  // Make sure that if "NodeMaxDispatchGrid" is used, that an input
+  // with the SV_DispatchGrid attribute is provided.
+  if (funcProps->Node.MaxDispatchGrid[0] > 0) {
+    bool found = false;
+    if (funcProps->InputNodes.size() != 0) {
```
tex3d, line 2456: I don't see why this check is necessary, when the loop will simply have no iterations if this is empty.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2448,22 +2448,24 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+      for (auto node : funcProps->InputNodes) {
+        if (node.RecordType.SV_DispatchGrid.NumComponents != 0) {
+          found = true;
+          break;
+        }
+      }    
+    }
+    if (!found || funcProps->InputNodes.size() == 0) {
```
tex3d, line 2464: The `funcProps->InputNodes.size() == 0` check seems redundant, since `found` will have to be false in that case.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2448,22 +2448,24 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+      unsigned DiagID = Diags.getCustomDiagID(
+          DiagnosticsEngine::Error,
+          "node shader '%0' with NodeMaxDispatchGrid attribute must have an input node with the SV_DispatchGrid semantic");
```
tex3d, line 2467: "have an input node with the SV_DispatchGrid semantic"
Is more accurately:
"declare an input record containing a field with the SV_DispatchGrid semantic"

tex3d (approved):
LGTM!

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2448,22 +2448,23 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
-      defaultInput.RecordType.SV_DispatchGrid.ComponentType = DXIL::ComponentType::U32;
-      defaultInput.RecordType.SV_DispatchGrid.NumComponents = 3;
-      funcProps->InputNodes.push_back(defaultInput);
-    } else {
-      hlsl::NodeIOProperties emptyInput(DXIL::NodeIOKind::EmptyInput);
-      funcProps->InputNodes.push_back(emptyInput);
+  
+  // Make sure that if "NodeMaxDispatchGrid" is used, that an input
+  // with the SV_DispatchGrid attribute is provided.
+  if (funcProps->Node.MaxDispatchGrid[0] > 0) {
```
llvm-beanz, line 2454: Is it too much work to put this into Sema? Ideally we shouldn’t be generating this in CodeGen.

bob80905: In this specific case, I would say yes, because I'd need to recreate funcProps in sema, and I do think the PR is already on the larger side, so I would prefer to do this later.

llvm-beanz: So, I'm okay with that answer, but I think we should file an issue to track moving this diagnostic.

llvm-beanz (approved):

PR #5728, opened 2023-09-15
Fix ASAN alloc-dealloc-mismatch (operator new vs free)

amaiorano:
The MallocAllocator originally used malloc and free, but this was changed for DXC to use operator new/delete. The test called free() on the object that is now allocated with new. Fixed the test by passing in the allocator explicitly, and destroying the object using it.

### DIFF
--- a/unittests/ADT/StringMapTest.cpp
+++ b/unittests/ADT/StringMapTest.cpp
@@ -184,15 +184,18 @@ TEST_F(StringMapTest, IterationTest) {
   }
 }
 
+// HLSL Change begin
 // Test StringMapEntry::Create() method.
 TEST_F(StringMapTest, StringMapEntryTest) {
+  MallocAllocator A;
   StringMap<uint32_t>::value_type* entry =
       StringMap<uint32_t>::value_type::Create(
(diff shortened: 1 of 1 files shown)


unittests/ADT/StringMapTest.cpp
```diff
@@ -186,12 +186,13 @@ TEST_F(StringMapTest, IterationTest) {
...
 TEST_F(StringMapTest, StringMapEntryTest) {
+  MallocAllocator A;
   StringMap<uint32_t>::value_type* entry =
       StringMap<uint32_t>::value_type::Create(
-          StringRef(testKeyFirst, testKeyLength), 1u);
+          StringRef(testKeyFirst, testKeyLength), A, 1u);
   EXPECT_STREQ(testKey, entry->first().data());
   EXPECT_EQ(1u, entry->second);
-  free(entry);
+  entry->Destroy(A);
```
llvm-beanz, line 195: Can you wrap this test in a `// HLSL Change begin/end` comment block?

amaiorano: Done.

s-perron (approved):

llvm-beanz (approved):

PR #5730, opened 2023-09-15
Fix ASAN stack-use-after-scope in unit test

amaiorano:
When StackOption is destructed, it reads the its ArgStr member, which in this test is set to local variable ArgString. Since ArgString was declared after TestOption, it would be destructed first, and thus result in a stack-use-after-scope. Fix this by swapping the declaration order.

### DIFF
--- a/unittests/Support/CommandLineTest.cpp
+++ b/unittests/Support/CommandLineTest.cpp
@@ -72,11 +72,13 @@ public:
 
 cl::OptionCategory TestCategory("Test Options", "Description");
 TEST(CommandLineTest, ModifyExisitingOption) {
-  StackOption<int> TestOption("test-option", cl::desc("old description"));
-
+  // HLSL Change begin
   const char Description[] = "New description";
   const char ArgString[] = "new-test-option";
   const char ValueString[] = "Integer";
+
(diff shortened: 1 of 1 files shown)


unittests/Support/CommandLineTest.cpp
```diff
@@ -72,12 +72,12 @@ class StackOption : public cl::opt<T> {
 
 cl::OptionCategory TestCategory("Test Options", "Description");
 TEST(CommandLineTest, ModifyExisitingOption) {
-  StackOption<int> TestOption("test-option", cl::desc("old description"));
-
   const char Description[] = "New description";
   const char ArgString[] = "new-test-option";
   const char ValueString[] = "Integer";
 
+  StackOption<int> TestOption("test-option", cl::desc("old description"));
```
llvm-beanz, line 79: Can you wrap the block from the top of this test function down to below the option declaration in a `//HLSL Change begin/end` comment block?

amaiorano: Done.

s-perron (approved):

llvm-beanz (approved):

PR #5732, opened 2023-09-15
[NFC] Integrate clang-format to hctgen

llvm-beanz:
This integrates formatting generated sources with clang-format with the hctgen generation process.

This change makes a few small adjustments to how the build-time header generation works. This change, disables automated build-time generation for C++ sources if you don't have clang-format available on your system.

CMake can automatically detect clang-format installed as part of Visual Studio, or based on your PATH. You can also explicitly set `CLANG_FORMAT_EXE` when configuring to manually override. **Developers on macOS** will need to install clang-format manually and place it on their path to generate DXC's generated sources.

If clang-format cannot be found and `HLSL_COPY_GENERATED_SOURCES` is On a fatal error will be reported at configuration time.

This change _does not_ make clang-format a requirement to build DXC, but it does make it a requirement if you are modifying the generated sources.


dmpots (approved):
LGTM

cmake/modules/HCT.cmake
```diff
@@ -49,6 +55,15 @@ function(add_hlsl_hctgen mode)
                        ${hctdb}
                        ${hctdb_helper})
 
+  get_filename_component(output_extension ${full_output} LAST_EXT)
+
+  if (output_extension MATCHES "\.h|\.cpp|\.inl")
```
damyanp, line 60: Should this also check for `.hpp` extensions?

damyanp (approved):

cmake/modules/HCT.cmake
llvm-beanz: We don't have any `.hpp` files, and I think we shouldn't add any...

damyanp: Probably not affected by codegen?

https://github.com/search?q=repo%3Amicrosoft%2FDirectXShaderCompiler%20path%3A*.hpp&type=code

llvm-beanz: Two of those are tests and the third... I wish didn't exist and am now sad. But they shouldn't be impacted here.

cmake/modules/HCT.cmake
```diff
@@ -87,6 +107,7 @@ function(add_hlsl_hctgen mode)
                      COMMAND ${PYTHON_EXECUTABLE}
                              ${hctgen} ${force_lf}
                              ${mode} --output ${temp_output} ${input_flag}
+                     ${format_cmd}
```
pow2clk, line 110: If there's no clang format, I don't expect this to end well?

cmake/modules/HCT.cmake
```diff
@@ -49,6 +58,17 @@ function(add_hlsl_hctgen mode)
                        ${hctdb}
                        ${hctdb_helper})
 
+  get_filename_component(output_extension ${full_output} LAST_EXT)
+
+  if (output_extension MATCHES "\.h|\.cpp|\.inl")
+    set(format_cmd COMMAND ${CLANG_FORMAT_EXE} -i ${temp_output})
+    if (NOT CLANG_FORMAT_EXE)
+      add_custom_target(${mode} DEPENDS ${output})
+      add_dependencies(HCTGen ${mode})
```
pow2clk, line 67: What does this do to mitigate the absence of clang-format?

pow2clk (commented):
I think I misunderstand something about the cmakery here

cmake/modules/HCT.cmake
llvm-beanz: If a variable in CMake is not set it resolves to an empty string.

cmake/modules/HCT.cmake
llvm-beanz: It creates the generation targets as empty targets so that other places that depend on the targets don’t error out. Earlier on it logs a earring for missing clang-format, and will error if you don’t have clang format and explicitly set the option to re-generate the generated sources.

Issue #5364, opened 2023-06-28
Cleanup the preview branch


llvm-beanz:
Internal work item 120 has some details, but the preview branch has a number of TODO lines, stale code from earlier iterations and other NFC code cleanups that should be performed.

pow2clk:
There is a lot to do there. This should be divided into multiple tasks. I'll look over the internal issue and get that done.

pow2clk:
Preview branch is merged. There might be a few other cleanups involved here, but they'll have their own issues.

PR #5734, opened 2023-09-15
Remove unneeded Sema equal param special case

pow2clk:
This was added to address a special case where OutputComplete was expecting a templated type, but got the wrong one and gave an unhelpful "no matching function" error instead of the error more specific to the requirements of OutputComplete.

Since OutputComplete no longer takes this as a parameter and is instead a method of the object, the hack is not needed.

Also restores testing for this by `node_emptynodeinput` in
tools/clang/test/HLSL/workgraph/outputcomplete_unsupported_nodeio.hlsl
```hlsl
[Shader("node")]
[NodeLaunch("Coalescing")]
void node_emptynodeinput([MaxRecords(5)] EmptyNodeInput emptyNodeInput)
{
  emptyNodeInput.OutputComplete(); // expected-error {{no member named 'OutputComplete' in 'EmptyNodeInput'}}
}
```

part of #5364

### DIFF
--- a/tools/clang/lib/Sema/SemaOverload.cpp
+++ b/tools/clang/lib/Sema/SemaOverload.cpp
@@ -9051,14 +9051,6 @@ static void DiagnoseBadConversion(Sema &S, OverloadCandidate *Cand,
       }
   }
 
-  // HLSL Change Starts
-  // With some intrinsics with templated parameters we can end up here
-  // with the To and From types being the same - we'll return early to
-  // avoid a weird diagnostic.
-  if (ToTy == FromTy)
-    return;
-  // HLSL Change Ends
-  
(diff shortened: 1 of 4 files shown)


llvm-beanz (approved):

python3kgae (approved):

bob80905:
Could you add the full path to case127 to the description?
I don't see this test case under tools\clang\test\HLSLFileCheck\hlsl\workgraph, and there doesn't appear to be any file that starts with case127 in the filename.

pow2clk:
Drat. Looks like #5467 deleted that test. It doesn't look like it was replaced with a verifier test like some of the others were. I'll add that to this PR. Tagging @tcorringham in case the removal without a replacement was intentional

bob80905 (approved):

PR #5735, opened 2023-09-15
StringSwitch CodeGen/HLops string->enums

pow2clk:
A number of places had if else chains to convert strings into enum values. This unapoligetically uses StringSwitch for all of them and moves one that already did in line with the rest for consistency

In HLOperations, There was a complex cascading swith on characters that was error prone. Using StringSwitch is easily maintainable and reasonably performant

This also required submitting only the substring relevant to the operation group as before it only attended to the first few characters

part of #5364


python3kgae (approved):

hekota (approved):
LGTM!

PR #5740, opened 2023-09-18
Make member expressions of bitfields emit as an RValue instead of LValue

### LINKED ISSUE #4758: [HLSL2021] Vector swizzle on bitfield crashes DXC
Hi,

the compiler crashes with this error:
```
dxc /T ps_6_6 /HV 2021 e:\temp\ps_bitfield.hlsl
Internal compiler error: access violation. Attempted to read from address 0x0000000000000008
```
When compiling this code:
```
#define USE_BIT_FIELD

struct MyStruct
{
#ifdef USE_BIT_FIELD
    uint v0: 5;
    uint v1: 15;
    uint v2: 12;
#else
    uint v2;
#endif
};

ConstantBuffer<MyStruct> myConstBuff;

[RootSignature("RootFlags(0), DescriptorTable(CBV(b0))")]
float4 main(uint addr: TEXCOORD): SV_Target
{
    // return myConstBuff.v2.x; // works fine with bitfields        
    return myConstBuff.v2.xxxx; // crash with bitfields. Using a structured buffer instead of a constant buffer also crashes   
}
```

More precisely, the bug is related to the .xxxx swizzle combined with the bit field.

Cheers,
Olivier

bob80905:
Previously, when emitting a vector splat type as a cast LValue, the only valid types that could be casted were simple and Extended vector element types. If, (in this case), a bitfield type was detected, an assert would be triggered causing an internal compiler error. This change adds cases to handle all but one of the possible expected types to cast from. A test was added specifically for the bitfield conversion process. The regression test uses the shadaer supplied in the original issue, and this PR now successfully compiles that shader.
Fixes #4758


tools/clang/lib/CodeGen/CGExpr.cpp
```diff
@@ -3498,7 +3498,16 @@ LValue CodeGenFunction::EmitCastLValue(const CastExpr *E) {
       LVal = Builder.CreateGEP(
           LV.getExtVectorAddr(),
           {Builder.getInt32(0), VecElts->getAggregateElement((unsigned)0)});
-    } else
+    } else if (LV.isBitField()) {
+      LVal = LV.getBitFieldAddr();
+    } else if (LV.isVectorElt()) {
+      LVal = LV.getVectorAddr();
+    } else if (LV.isGlobalReg()) {
+      LVal = LV.getGlobalReg();
```
pow2clk, line 3506: I think we need test cases for each of these

tools/clang/test/HLSLFileCheck/validation/bitfield_swizzle.hlsl
```diff
@@ -0,0 +1,31 @@
...
+
+// CHECK: extractvalue %dx.types.ResRet.i32
+// CHECK: extractvalue %dx.types.CBufRet.i32
+
+
+[RootSignature("RootFlags(0), DescriptorTable(CBV(b0), SRV(t0))")]
+float4 main(uint addr: TEXCOORD): SV_Target
+{
+    // return myConstBuff.v2.x; // works fine with bitfields
+    return myStructBuff[0].v2.xxxx + myConstBuff.v2.xxxx; // crash with bitfields. Using a structured buffer instead of a constant buffer also crashes
```
pow2clk, line 30: I don't think you meant to leave this comment here.

For more exhaustive testing, can we get tests for some other swizzles even if they are invalid? .xyyz for example?

tools/clang/test/HLSLFileCheck/validation/bitfield_swizzle.hlsl
```diff
@@ -0,0 +1,31 @@
+// RUN: %dxc -T ps_6_6 -HV 2021 %s | FileCheck %s
```
pow2clk, line 1: This isn't a validation test. It shouldn't be in this directory. There is an hlsl/operators/swizzle/ subdir or else there is an existing bitfields test in hlsl/types/struct/

pow2clk (commented):
Looks reasonable, I just have some testing requests

llvm-beanz (commented):
I'm going to chat with @bob80905 later today to dig into this, but I don't think this is the correct solution to the problem.

`HLSLVectorElementExpr` expressions with repeated component values should only ever be generated as `rvalue`s. I think fixing CodeGen to not crash when an invalid AST gets fed to it is a good way to sprout mis-compile bugs. We should instead fix this in AST generation to generate `rvalue` expressions.

llvm-beanz (commented):
We should probably have a test case that verifies a diagnostic is emitted if you assign to a bitfield using `.x` or try to access one with `.xy`.

Otherwise I think this looks right to me.

tools/clang/test/HLSLFileCheck/hlsl/operators/swizzle/swizzleBitfieldNotAllowed.hlsl
```diff
@@ -0,0 +1,27 @@
...
+StructuredBuffer<MyStruct> myStructBuff;
+
+
+[RootSignature("RootFlags(0), DescriptorTable(CBV(b0), SRV(t0))")]
+float4 main(uint addr: TEXCOORD): SV_Target
+{  
+    // CHECK: error: expression is not assignable
+    myConstBuff.v2.x = 4;
+    // CHECK: error: expression is not assignable
+    myStructBuff[0].v2.x = 4;
```
llvm-beanz, line 24: Checks for diagnostics should use `-verify` and the `//expected-error` syntax.

tools/clang/test/HLSLFileCheck/hlsl/operators/swizzle/swizzleBitfield.hlsl
```diff
@@ -0,0 +1,27 @@
...
+{
+    uint v0: 5;
+    uint v1: 15;
+    uint v2: 12;
+};
+
+ConstantBuffer<MyStruct> myConstBuff;
+StructuredBuffer<MyStruct> myStructBuff;
+
+// CHECK: extractvalue %dx.types.ResRet.i32
```
llvm-beanz, line 17: Because this isn’t a code generation change I’m actually less concerned that the codegen is correct and more concerned that the AST is correct. Can you write checks to verify that the AST is generating the `LValueToRValue` casts?

I’d look to see 4 AST nodes in the checks for each access:
* The HLSL vector element expression (which will be an rvalue)
* The implicit cast expression for the Vector splat T -> vector<T,1> (which will be an rvalue)
* The implicit cast expression for the LValueToRValue (which will be an rvalue)
* The member expression for the bitfield

llvm-beanz (approved):

pow2clk (approved):
Looks good!

PR #5742, opened 2023-09-18
Execution tests: allow sharing of compile arguments

hekota:
Allow sharing of shader compile arguments by referencing Arguments attribute from other shaders like this:

```
      ...
      <Shader Name="PS" Target="ps_6_0" EntryPoint="PSMain" Text="@CS" Arguments="@CS"/>
      <Shader Name="CS" Target="cs_6_0" EntryPoint="CSMain" Arguments="/Od">
        <![CDATA[
      ...
```

### DIFF
--- a/tools/clang/unittests/HLSLExec/ShaderOpTest.h
+++ b/tools/clang/unittests/HLSLExec/ShaderOpTest.h
@@ -218,6 +218,18 @@ public:
     }
     return result;
   }
+  LPCSTR GetShaderArguments(ShaderOpShader *pShader) {
+    if (!pShader || !pShader->Arguments) return nullptr;
+    LPCSTR result = pShader->Arguments;
+    if (result[0] == '@') {
+      for (auto && S : Shaders) {
+        if (S.Name && 0 == strcmp(S.Name, result + 1))
+          return S.Arguments;
+      }
+      result = nullptr;
+    }
+    return result;
+  }
   ShaderOpDescriptorHeap *GetDescriptorHeapByName(LPCSTR pName) {
     for (auto && R : DescriptorHeaps) {
       if (R.Name && 0 == strcmp(R.Name, pName))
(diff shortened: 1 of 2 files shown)


tools/clang/unittests/HLSLExec/ShaderOpTest.h
```diff
@@ -218,6 +218,18 @@ class ShaderOp {
     }
     return result;
   }
+  LPCSTR GetShaderArguments(ShaderOpShader *pShader) {
+    if (!pShader || !pShader->Arguments) return nullptr;
+    LPCSTR result = pShader->Arguments;
+    if (result[0] == '@') {
+      for (auto && S : Shaders) {
```
bob80905, line 225: May I ask why auto && instead of auto &?

tools/clang/unittests/HLSLExec/ShaderOpTest.h
```diff
@@ -218,6 +218,18 @@ class ShaderOp {
...
+        if (S.Name && 0 == strcmp(S.Name, result + 1))
```
bob80905, line 226: if `result` is really the list of arguments, are they separated by null terminators? otherwise, I imagine that the names may match, but the result string will be longer than S.name because of the rest of the arguments, and the strcmp function won't return 0.

tools/clang/unittests/HLSLExec/ShaderOpTest.h
pow2clk: I suspect this is just how it was copied from `GetShaderText` which uses this, but doesn't really need to. It's harmless anyway.

tools/clang/unittests/HLSLExec/ShaderOpTest.h
pow2clk: `result` is either the list of arguments separated by spaces or else it is the special code `@<identifier>`.  If it has this special code followed by additional arguments, it's malformed and shouldn't match since we're not going to pay attention to those arguments. To allow that might be useful, but it differs from how the original implementation did things and I personally think it adds as much confusion as convenience.

pow2clk (approved):
Please don't take my responses to Josh's comments as gospel. I just felt I should explain why I'm approving in spite of them 😄

bob80905 (approved):

tools/clang/unittests/HLSLExec/ShaderOpTest.h
hekota: Correct, this is copied from `GetShaderText` and this syntax is used in multiple places in the `ShaderOpTest` code. It means  universal reference that will bind to anything (lvalue or rvalue).

Issue #5744, opened 2023-09-18
Intrinsics ddx_fine/ddy_fine intrinsics should not be allowed to sink into flow control


hekota:
**Description**
Intrinsics ddx_fine/ddy_fine should not be allowed to sink into flow control. That can currently happen because they are marked `ReadNone` same as other unary ops, but they are reading values from neighboring threads in the same quad.

**Steps to Reproduce**
1. Open `tools/clang/unittests/HLSLExec/ShaderOpArith.xml`, find data for `HelperLaneTestNoWave` and remove `-opt-disable sink` from the shader compile arguments.
2. Run ```hcttest exec-filter *HelperLaneTest -verbose -adapter NVidia*```

**Actual Behavior**
```
StartGroup: ExecutionTest::HelperLaneTest
Verifying IsHelperLane in shader model 6.0
Using Adapter:NVIDIA GeForce GTX 1660 Ti
Error: Verify: AreEqual(pTestResults[0].is_helper_00, 0) - Values (1, 0) [File: D:\dxc2\tools\clang\unittests\HLSLExec\ExecutionTest.cpp, Function: ExecutionTest::HelperLaneTest, Line: 10939]
EndGroup: ExecutionTest::HelperLaneTest [Failed]
```

**Expected Behavior**
```
StartGroup: ExecutionTest::HelperLaneTest
Verifying IsHelperLane in shader model 6.0
Using Adapter:NVIDIA GeForce GTX 1660 Ti
Verifying IsHelperLane in shader model 6.6
Using Adapter:NVIDIA GeForce GTX 1660 Ti
EndGroup: ExecutionTest::HelperLaneTest [Passed]
```

**Environment**
- DXC version 634c2f349 or later (main branch from July 28 forward)
- Windows

simondeschenes:
I am hitting this problem, is there any known workaround?

hekota:
The compile flag `-opt-disable sink` should help with that.

PR #5745, opened 2023-09-18
Fix HelperLaneTest

hekota:
Add `-opt-disable sink` to HelperLaneTest shader compilation to prevent sinking of `ddx_fine`/`ddy_fine` intrinsics into flow control.

Separate issue has been filed to track down the root of the problem:
#5744: Intrinsics ddx_fine/ddy_fine should not be allowed to sink into flow control


pow2clk (approved):
I think it would help to have a comment that explains the reason for the argument and references the bug number. Otherwise looks good!

damyanp (approved):

tex3d (approved):
LGTM!

PR #5752, opened 2023-09-19
[NFC] Merge clang-format into 6.8 staging

llvm-beanz:
This merges the clang-formatting change into the 6.8 staging branch and formats the 6.8 sources appropriately.


cmake/modules/HCT.cmake
```diff
@@ -49,6 +58,12 @@ function(add_hlsl_hctgen mode)
                        ${hctdb}
                        ${hctdb_helper})
 
+  get_filename_component(output_extension ${full_output} LAST_EXT)
+
+  if (CLANG_FORMAT_EXE AND output_extension MATCHES "\.h|\.cpp|\.inl")
```
bob80905, line 63: We may have agreed on this already, but do we want to run clang-format on our hlsl test files? Does it even work on those files?

llvm-beanz: HLSL source files should clang-format just fine. This line isn’t about that though. This code is only used to run clang-format on the source files that contain bits of generated code that come out of the hctgen scripts.

pow2clk (approved):

PR #4571, opened 2022-08-02
NFC. Ensure that the output directory exists

llvm-beanz:
Depending on build ordering the output directory may not be created
before the output file is written. This resolves the issue by forcing
the creation.

Resolves #4570.

PR #5755, opened 2023-09-20
[NFC] Fix clang-formatting generated sources

llvm-beanz:
Clang-format determines if a header is the "main" header by looking at the header name without the last extension and matching it against the source file (with some extra handling for cuda files).

The automation generating temporary files with a `.tmp` suffix tripped this up so it wasn't matching the "main" header when sorting includes.

This corrects the issue by writing temporaries to a tmp directory instead of munging the suffix.

### DIFF
--- a/cmake/modules/HCT.cmake
+++ b/cmake/modules/HCT.cmake
@@ -44,7 +44,7 @@ function(add_hlsl_hctgen mode)
     message(FATAL_ERROR "add_hlsl_hctgen requires OUTPUT argument")
   endif()
  
-  set(temp_output ${CMAKE_CURRENT_BINARY_DIR}/${ARG_OUTPUT}.tmp)
+  set(temp_output ${CMAKE_CURRENT_BINARY_DIR}/tmp/${ARG_OUTPUT})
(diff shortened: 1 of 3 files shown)


dmpots (approved):
LGTM

damyanp (approved):

cmake/modules/HCT.cmake
```diff
@@ -44,7 +44,7 @@ function(add_hlsl_hctgen mode)
     message(FATAL_ERROR "add_hlsl_hctgen requires OUTPUT argument")
   endif()
  
-  set(temp_output ${CMAKE_CURRENT_BINARY_DIR}/${ARG_OUTPUT}.tmp)
+  set(temp_output ${CMAKE_CURRENT_BINARY_DIR}/tmp/${ARG_OUTPUT})
```
bogner, line 47: This assumes hctgen creates directories in its output paths if they don't exist. I think that's the case so no need to change it if so, but I've seen this kind of thing trip tools up before so worth double checking.

bogner (approved):

cmake/modules/HCT.cmake
llvm-beanz: Where were you last year when I made that mistake in the first place? It was fixed in #4571 😁

PR #5758, opened 2023-09-20
Allow wavesize attribute on library shaders

### LINKED ISSUE #5374: WaveSize attribute should be allowed on a compute shader in a library

bob80905:
This PR moves the wavesize diagnostics over to sema, and adjusts them so that the error messages are clearer and so that wavesize attributes are allowed on library shaders.
Fixes #5374


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12773,6 +12834,30 @@ HLSLMaxRecordsAttr *ValidateMaxRecordsAttributes(Sema &S, Decl *D,
                          A.getAttributeSpellingListIndex());
 }
 
+HLSLWaveSizeAttr *ValidateWaveSizeAttributes(Sema &S, Decl *D,
+                                             const AttributeList &A) {
+  // make sure we are in an appropriate shader model
+  const auto *SM =
+      hlsl::ShaderModel::GetByName(S.getLangOpts().HLSLProfile.c_str());
```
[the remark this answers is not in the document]
bob80905: I think I need to leave the c_str() because it won't just accept a string. GetByName requires a const char *.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11299,6 +11299,66 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
 }
 
+void ValidateCallGraphWaveSize(clang::Sema *S, FunctionDecl *FD) {
+  const auto *SM =
+      hlsl::ShaderModel::GetByName(S->getLangOpts().HLSLProfile.c_str());
+
+  const std::string &entryName = S->getLangOpts().HLSLEntryFunction;
+  StringRef functionName = "";
+  if (FD->getIdentifier())
+    functionName = FD->getIdentifier()->getName();
```
llvm-beanz, line 11309: You can just do `StringRef functionName = FD->getName();`, it will return an empty string if there is no identifier.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11299,6 +11299,66 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+      !SM->IsLib() &&
+      FD->getDeclContext()->getDeclKind() == Decl::Kind::TranslationUnit &&
+      functionName == entryName;
+
+  for (auto *pAL : FD->specific_attrs<HLSLShaderAttr>()) {
+    StringRef Literal =
+        pAL->getStage();
+    DXIL::ShaderKind Stage = ShaderModel::KindFromFullName(Literal);
+    isNode |= Stage == DXIL::ShaderKind::Node;
+    isCS |= Stage == DXIL::ShaderKind::Compute;
```
llvm-beanz, line 11324: We've eliminated the ability to have a shader that is both a compute and a node shader, so we should be able to simplify this. There should only be one `HLSLShaderAttr` on a given function.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
llvm-beanz: Gah! That's... wrong. We should fix that.

You can change GetByName to take a StringRef and it shouldn't require any additional changes since `const char *` implicitly converts to StringRef.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrNotAllowed.hlsl
```diff
@@ -0,0 +1,9 @@
+// RUN: %dxc -E S -T cs_6_0 %s | FileCheck %s
```
python3kgae, line 1: Add a run line for lib_6_7?
It can test entry in lib profile.

python3kgae (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12773,6 +12830,30 @@ HLSLMaxRecordsAttr *ValidateMaxRecordsAttributes(Sema &S, Decl *D,
...
+  }
+
+  // validate that the wavesize argument is a power of 2 between 4 and 128 inclusive
+  HLSLWaveSizeAttr *pAttr = ::new (S.Context)
+      HLSLWaveSizeAttr(A.getRange(), S.Context, ValidateAttributeIntArg(S, A),
+                       A.getAttributeSpellingListIndex());
+
+  unsigned waveSize = pAttr->getSize();
+  if (!DXIL::IsValidWaveSizeValue(waveSize)) {
+    S.Diag(A.getLoc(), diag::err_hlsl_wavesize_size);
```
pow2clk, line 12850: I have a mild preference for keeping the error message dependent on the internally defined constants. It may not be likely in the short term, but it is possible we change those at some point and it seems better design to let it all cascade from those constants.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12773,6 +12830,30 @@ HLSLMaxRecordsAttr *ValidateMaxRecordsAttributes(Sema &S, Decl *D,
                          A.getAttributeSpellingListIndex());
 }
 
+HLSLWaveSizeAttr *ValidateWaveSizeAttributes(Sema &S, Decl *D,
```
pow2clk, line 12833: Given the similar name to ValidateWaveSize above, a quick descriptive comment explaining why this function is separate might allay some future confusion.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11299,6 +11299,62 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+  const llvm::StringRef &entryName = S->getLangOpts().HLSLEntryFunction;
+  StringRef functionName = "";
+  if (FD->getIdentifier())
+    functionName = FD->getName();
+  bool isEntry = false;
+  bool isNode = false;
+  bool isCS = false;
+
+  isEntry =
+      !SM->IsLib() &&
```
pow2clk, line 11315: A more convenient way to perform this check in this environment would be to use the boolean that has already been assigned in the langopts : `self->getLangOpts().IsHLSLLibrary`. No (further) string parsing required.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11299,6 +11299,62 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+
+  auto *shaderAttribute = FD->getAttr<HLSLShaderAttr>();
+  StringRef Literal = shaderAttribute->getStage();
+  DXIL::ShaderKind Stage = ShaderModel::KindFromFullName(Literal);
+  isNode = Stage == DXIL::ShaderKind::Node;
+  isCS = Stage == DXIL::ShaderKind::Compute;
+
+  HLSLWaveSizeAttr *attr = FD->getAttr<HLSLWaveSizeAttr>();
+
+  if (!SM->IsLib() && !isEntry && !isNode) {
```
pow2clk, line 11327: This is the only use of isEntry. As a result, isEntry can be simplified to exclude the lib check. 

The effect of this and the code above is this check (lightly edited for brevity):

`if (!isLIb && (isLib || DeclKind != TranslationUnit || functionname != entryname) && !isNode)`

Since the only way to even get through short-circuiting to the second clause is for this to not be a library, the check for the library there isn't needed. 

I'm torn about the final !isNode clause. At present, the only way to have a node is to be compiling as a library, but I'm not quite confident enough about that always being the case to suggest removing it.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrAllowed4.hlsl
```diff
@@ -0,0 +1,22 @@
...
+
+// CHECK: @csmain1, !"csmain1", null, null, [[PROPS:![0-9]+]]}
+// CHECK: [[PROPS]] = !{i32 4, [[NT:![0-9]+]], i32 11, [[WS:![0-9]+]]}
+// CHECK: [[NT]] = !{i32 32, i32 1, i32 1}
+// CHECK: [[WS]] = !{i32 32}
+
+// CHECK-MAIN2: @csmain2, !"csmain2", null, null, [[PROPS:![0-9]+]]}
+// CHECK-MAIN2: [[PROPS]] = !{i32 4, [[NT:![0-9]+]], i32 11, [[WS:![0-9]+]]}
+// CHECK-MAIN2: [[NT]] = !{i32 32, i32 1, i32 1}
+// CHECK-MAIN2: [[WS]] = !{i32 32}
```
pow2clk, line 14: Given that the last three lines are the same in both of these, you could add multiple prefixes to the filecheck lines and include a CHECK that contains the common parts. These are short enough that it's kind of irrelevant, but I wanted to make sure you knew of the option.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrAllowed5.hlsl
```diff
@@ -0,0 +1,8 @@
+// RUN: %dxc -T lib_6_7 %s 
+
+// This test just checks whether compilation succeeds
+// because we shouldn't presume that N() is an entry 
+// point, so entry point attributes are ignored.
```
pow2clk, line 5: I think we at least need a CHECK-NOT for erros and warnings here. Since we're using the HLSL Filechecker, if dxc fails with an error, I don't think the test will fail.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrNotAllowedNode.hlsl
```diff
@@ -0,0 +1,12 @@
+// RUN: %dxc -T lib_6_7 %s | FileCheck %s
+// note, this test would pass if the node shader
+// launch type was NodeLaunchType::Thread
+
+
+// CHECK: error: Node shader 'N' with broadcasting launch type requires 'numthreads' attribute
```
pow2clk, line 6: This doesn't look to be the error you wanted to test here. Additionally, I would expect lib_6_7 to fail eventually for lack of node shaders.

pow2clk (commented):
I feel more strongly about the test comments. The implementing code comments you can mostly take or leave.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrNotAllowed3.hlsl
```diff
@@ -0,0 +1,18 @@
+// RUN: %dxc -T cs_6_6 -E main %s | FileCheck %s
+
+// this test should pass, because despite the fact that 
+// main2 isn't the current entry point function, 
+// the fact that it's callgraph-reachable from the entry
+// point function "main" gives it permission to have
+// "wavesize" as an "entry point attribute"
```
tex3d, line 7: No, this should pass because the WaveSize attribute should be ignored if it's not on a known entry point function for this compilation.  This has nothing to do with whether it's call-graph reachable.

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -188,10 +188,14 @@ const ShaderModel *ShaderModel::Get(Kind Kind, unsigned Major, unsigned Minor) {
   // VALRULE-TEXT:END
 }
 
-const ShaderModel *ShaderModel::GetByName(const char *pszName) {
+const ShaderModel *ShaderModel::GetByName(llvm::StringRef Name) {
```
llvm-beanz, line 191: Thank you for this. Awesome cleanup!

tex3d (changes requested):
There's a lot of structural feedback here.  Might need to sync on some things before completely addressing some of this feedback to make sure we're on the same page.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11436,6 +11492,7 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
         Diags.Report(pPatchFnDecl->getSourceRange().getBegin(), id);
       }
     }
+    ValidateEntryPointFunctionAttributes(self, pEntryPointDecl);
```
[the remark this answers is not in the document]
bob80905: The reason I can't move it under Diagnose Entry is that it doesn't always correctly identify the entry point decl. For example, in WaveSizeAttrNotAllowed2.hlsl, it will treat main2 as the entry point when really main is the entry point. This causes a diagnostic misfire.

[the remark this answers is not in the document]
bob80905: The reason is because there are non-entry point function declarations that will get DiagnoseEntry called on them. The function seems poorly named, since it will be called on non-entry point functions.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11299,6 +11299,62 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+  if (FD->getIdentifier())
+    functionName = FD->getName();
+  bool isEntry = false;
+  bool isNode = false;
+  bool isCS = false;
+
+  isEntry =
+      !SM->IsLib() &&
+      FD->getDeclContext()->getDeclKind() == Decl::Kind::TranslationUnit &&
+      functionName == entryName;
```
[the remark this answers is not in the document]
bob80905: There needs to be some code somewhere that is responsible for emitting a diagnostic when, say, the wavesize attribute is found on, say, a pixel shader. That's why I think the isNode and isCS bools are necessary. I agree that isEntry seems useless since this function is only called on entry point decls anyways.

Agree with your last statement about the extra exemption, I'll add that.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11299,6 +11299,62 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+  StringRef Literal = shaderAttribute->getStage();
+  DXIL::ShaderKind Stage = ShaderModel::KindFromFullName(Literal);
+  isNode = Stage == DXIL::ShaderKind::Node;
+  isCS = Stage == DXIL::ShaderKind::Compute;
+
+  HLSLWaveSizeAttr *attr = FD->getAttr<HLSLWaveSizeAttr>();
+
+  if (!SM->IsLib() && !isEntry && !isNode) {
+    S->Diag(attr->getRange().getBegin(),
+            diag::err_hlsl_attribute_valid_on_entry_function_only)
```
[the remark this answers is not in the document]
bob80905: I've completely removed this diagnostic.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11299,6 +11299,62 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+  bool isCS = false;
+
+  isEntry =
+      !SM->IsLib() &&
+      FD->getDeclContext()->getDeclKind() == Decl::Kind::TranslationUnit &&
+      functionName == entryName;
+
+  auto *shaderAttribute = FD->getAttr<HLSLShaderAttr>();
+  StringRef Literal = shaderAttribute->getStage();
+  DXIL::ShaderKind Stage = ShaderModel::KindFromFullName(Literal);
```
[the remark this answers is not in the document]
bob80905: I've made the parent function pass down `ShaderModel` and the `HLSLShaderAttr`. I don't think it made sense to pass the `ShaderKind` because it is possible it won't exist, and it's difficult to express a non-existent shaderkind without using invalid, which isn't an accurate representation.

For now, I haven't passed down whether or not this is an active entry point, some further discussion will need to be had.

tools/clang/test/HLSLFileCheck/validation/WaveSizeAttrAllowed.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: %dxc -T lib_6_7 %s | FileCheck %s
```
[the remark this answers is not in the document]
bob80905: I'm not sure I can move any of these into the verifier test suite, since the tests are so dependent on the target profile / entry point arguments, and the point of the test will be entirely missed if I don't involve that information. The fact that the entry point / target profile can't be passed to a verifier test discourages the use of that suite in this case.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
tex3d: `DiagnoseEntry` is called on functions that *could* be an entry, then it determines whether it *is* an entry (by looking up shader attribute, after adding it if necessary), at which point, we can diagnose entry attributes, since we know it's an entry function.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
tex3d: > As for your last point, I believe my current implementation satisfies your description, since it doesn't bother validating the wavesize attr on non-entry point decl's at all.

This is because the function is only called when this is the active entry point in a non-library compilation.  That's not correct, since we want to diagnose all known entry points (ones marked with the shader attribute).  However _if we do that_, diagnostics that depend on compilation options should not occur for entry points that aren't active, since those options could be different for a compilation where those entry points _are_ active.

Does that make sense?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11299,6 +11299,36 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+  std::string stageName = shaderAttr->getStage();
+
+  // whether or not this attribute should exist on this shader stage
+  bool isValid = false;
+  if (entryPointDecl->hasAttrs()) {
+    for (Attr *pAttr : entryPointDecl->getAttrs()) {
+      switch (pAttr->getKind()) {
+
+      case clang::attr::HLSLWaveSize: {
+        bool isValid = llvm::StringSwitch<bool>(stageName)
```
tex3d, line 11315: Can't we pass down the ShaderKind and switch on that here instead of using a StringSwitch for each attribute here?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15320,17 +15421,23 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD) {
...
+  bool isActiveEntry = false;
+  TryAddShaderAttrFromTargetProfile(S, FD, isActiveEntry);
 
-  auto Attr = FD->getAttr<HLSLShaderAttr>();
+  HLSLShaderAttr *Attr = FD->getAttr<HLSLShaderAttr>();
   if (!Attr) {
     return;
   }
 
+  if (isActiveEntry) {
```
tex3d, line 15439: This should be on inactive entries as well, so no need for this `if (isActiveEntry)` here.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15320,17 +15421,23 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD) {
...
+    DiagnoseEntryAttrAllowedOnStage(&S, FD, Attr);
```
tex3d, line 15440: If this is called after `DXIL::ShaderKind Stage = ShaderModel::KindFromFullName(Attr->getStage());`, you can pass Stage in here, and switch on the enum instead of using StringSwitch there.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15119,7 +15179,28 @@ static bool nodeInputIsCompatible(StringRef &typeName,
       .Default(false);
 }
 
-void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
+void DiagnoseComputeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr,
```
tex3d, line 15182: I don't think we should be passing the shader attribute down to these functions.  Just pass the `ShaderKind` enum value decoded from the attribute in the calling function.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15277,9 +15371,16 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr) {
...
 // shader attribute to the FD and carry on with validation
-void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD) {
+void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD, bool &isActiveEntry) {
+  isActiveEntry = false;
   const std::string &EntryPointName = S.getLangOpts().HLSLEntryFunction;
 
+  if (S.getLangOpts().IsHLSLLibrary) {
+    // TODO: Improve accuracy by analyzing -exports option
+    // to determine which entries are active for lib target
+    isActiveEntry = true;
```
tex3d, line 15383: I think this part needs to be defaulted and commented about in `DiagnoseEntry`.

tex3d: In fact, if we have a library target, we wouldn't need to call this function at all.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15320,17 +15421,23 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD) {
       HLSLShaderAttr::CreateImplicit(S.Context, fullName);
 
   FD->addAttr(pShaderAttr);
+  isActiveEntry = true;
   return;
 }
 
 void DiagnoseEntry(Sema &S, FunctionDecl *FD) {
-  TryAddShaderAttrFromTargetProfile(S, FD);
+  bool isActiveEntry = false;
```
tex3d, line 15431: A comment explaining `isActiveEntry` is needed, along with a TODO for keying it off of the exports for lib target.  I see that `TryAddShaderAttrFromTargetProfile` sets this, but we should check if it's a lib here and skip calling that entirely if so.  For example:

```cpp
if (S.getLangOpts().IsHLSLLibrary) {
  // TODO: Analyze -exports option to determine which entries
  // are active for lib target.
  // For now, assume all entries are active.
  isActiveEntry = true;
} else {
  TryAddShaderAttrFromTargetProfile(S, FD, isActiveEntry);
}
```

Note that we are setting `isActiveEntry` before we have actually confirmed that this is an entry via the shader attribute.  That's ok, as long as we keep that in mind.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15357,10 +15464,14 @@ void DiagnoseEntry(Sema &S, FunctionDecl *FD) {
   case DXIL::ShaderKind::ClosestHit: {
     return DiagnoseClosestHitEntry(S, FD, Attr);
   }
-  case DXIL::ShaderKind::Compute:
+  // TODO: Create a Compute function, move compute only diags over.
```
tex3d, line 15474: Now that you created a `DiagnoseComputeEntry` function, what is this TODO needed for?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11312,10 +11310,19 @@ void DiagnoseEntryAttrAllowedOnStage(clang::Sema *self,
...
-        bool isValid = llvm::StringSwitch<bool>(stageName)
-                           .Case("compute", true)
-                           .Case("node", true)
-                           .Default(false);
+        switch (shaderKind) {
+        case DXIL::ShaderKind::Compute: {
+          isValid = true;
+          break;
+        }
+        case DXIL::ShaderKind::Node: {
```
tex3d, line 11318: Why is this so much more code than necessary?  Why not just:
```cpp
switch (shaderKind) {
case DXIL::ShaderKind::Compute:
case DXIL::ShaderKind::Node:
  break;
default:
  self->Diag(pAttr->getRange().getBegin(),
                     diag::err_hlsl_attribute_unsupported_stage)
              << "WaveSize"
              << "compute or node";
  break;
}
```
Plus the `isValid` variable is defined and initialized outside the attribute loop, meaning it will carry a stale value around the loop which could be an issue if there's a path that doesn't update the flag for some attribute in the future.  I don't think it's needed at all.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15371,16 +15372,11 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, HLSLShaderAttr *Attr, bool isA
 
 // if this is the Entry FD, then try adding the target profile
 // shader attribute to the FD and carry on with validation
-void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD, bool &isActiveEntry) {
+void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
+                                       bool &isActiveEntry) {
   isActiveEntry = false;
```
tex3d, line 15377: We need a comment to explain this variable, I think.  For example:
> When isActiveEntry is true and this function is an entry point, this entry point is used in compilation.  This is an important distinction when diagnosing certain types of errors based on the compilation parameters.  For example, if isActiveEntry is false, diagnostics dependent on the shader model should not be performed.  That way we won't raise an error about a feature used by the inactive entry that's not available in the current shader model.  Since that entry point is not used, it may still be valid in another compilation where a different shader model is specified.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15119,16 +15177,32 @@ static bool nodeInputIsCompatible(StringRef &typeName,
...
+                          bool isActiveEntry) {
+  if (isActiveEntry) {
+    if (auto WaveSizeAttr = FD->getAttr<HLSLWaveSizeAttr>()) {
+      std::string profile = S.getLangOpts().HLSLProfile;
+      const ShaderModel *SM = hlsl::ShaderModel::GetByName(profile.c_str());
+      if (!SM->IsSM66Plus()) {
+        S.Diags.Report(WaveSizeAttr->getRange().getBegin(),
+                       diag::err_hlsl_attribute_in_wrong_shader_model)
+            << "wavesize"
+            << "6.6";
```
tex3d, line 15190: While this is fine for now, I think as these kinds of checks expand to more attributes, we can share more code, and shouldn't be doing things like translating the profile string to a shader model locally.

In other words, we would want to:
- Pass ShaderModel down to `Diagnose*Entry` functions when they need it
- Create a generic function to wrap a shader model version check for an attribute

Once that's done, the `DiagnoseNodeEntry`, or `DiagnoseComputeEntry` code for this would just look like:
```hlsl
if (auto WaveSizeAttr = FD->getAttr<HLSLWaveSizeAttr>()) {
  if (isActiveEntry)
    DiagnoseAttrRequiresShaderModel(WaveSizeAttr, 6, 6, SM);
  // do other WaveSize checks here...
  // such as ensuring WaveSize divides evenly into NumThreads when WaveMatrix is used.
}
```

tools/clang/test/HLSLFileCheck/hlsl/diagnostics/errors/WaveSizeAttrNotAllowed3.hlsl
```diff
@@ -0,0 +1,12 @@
+// RUN: %dxc -E S -T cs_6_7 %s | FileCheck %s
+
+// CHECK: error: shader attribute type 'wavesize' conflicts with shader attribute type 'wavesize'
```
tex3d, line 3: Note: This error message is unclear and could be improved in the future.

tex3d (approved):
Looks good.  I had a couple more comments that I think could be addressed in the future as we move more attributes to this path.

PR #5759, opened 2023-09-20
Fix apple-clang debug build failing to link llvm_assert

amaiorano:
When building with apple-clang on MacOS, we get a linker error:

ld64.lld: error: undefined symbol: llvm_assert

Because the llvm_assert implementation is missing the last parameter.

### DIFF
--- a/lib/Support/assert.cpp
+++ b/lib/Support/assert.cpp
@@ -25,7 +25,8 @@ void llvm_assert(const char *_Message, const char *_File, unsigned _Line,
 
 #include <assert.h>
 
-void llvm_assert(const char *message, const char *, unsigned) {
+void llvm_assert(const char *message, const char *, unsigned,
+                 const char *_Function) {
   assert(false && message);
 }
 


llvm-beanz (approved):
Weird. I have no idea why this wasn't failing before.

amaiorano:
Discussed with @llvm-beanz, and he points out that this is a C function, so the arguments are not part of the name mangling, which `ld64` (Clang) properly ignores; however, it's possible that `lld` (Apple Clang) may not.

amaiorano:
> Worth noting: there is no `lld` from Apple Clang. Apple doesn’t support `lld`. I assume you’re hitting this with a custom toolchain that includes `lld` and uses it as the linker for macOS rather than using the system linkers `ld64` or `ld64_prime`.

Actually, looking at my original commit comment, it is failing with `ld64`:
```
ld64.lld: error: undefined symbol: llvm_assert
```

So now I'm back to not being sure why this happens on our Mac builds, but not Linux.

llvm-beanz:
> Actually, looking at my original commit comment, it is failing with `ld64`:
> 
> ```
> ld64.lld: error: undefined symbol: llvm_assert
> ```

`ld64.lld` is LLD's `ld64`-compatible command line mode. The linker is `lld` for mach-o which is also _very_ different from `lld` for ELF.

PR #5761, opened 2023-09-21
Fix the clang-format workflow

llvm-beanz:
There were a few bugs that were difficult to reproduce. The pull request branch ref needs to be fetched and checked out manually in order to get the PR diff to construct. This change also fixes formatting to get a nice diff printed of the PR output.

See an example of the error output this produces here:

https://github.com/llvm-beanz/DirectXShaderCompiler/pull/2

### DIFF
--- a/.github/workflows/clang-format-checker.yml
+++ b/.github/workflows/clang-format-checker.yml
@@ -17,22 +17,22 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: install clang-format
-        run: sudo apt install -y clang-format
       - name: Comparing PR against target branch
         env:
           PR_NUMBER: ${{ github.event.number }}
           GH_TOKEN: ${{ github.token }}
         run: |
-            echo Comparing $GITHUB_BASE_REF vs HEAD
-            git diff -U0 --no-color origin/$GITHUB_BASE_REF..HEAD | clang-format-diff -p1 | tee format.diff
-            if [ -s "format.diff" ]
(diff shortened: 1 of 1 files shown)


hekota (approved):

.github/workflows/clang-format-checker.yml
```diff
@@ -17,22 +17,22 @@ jobs:
...
+            if [ -s format.diff ]
             then
               echo PR contains clang-format violations. First 50 lines of the diff: >> message.txt
-              echo ``` >> message.txt
+              echo \`\`\`diff >> message.txt
               cat format.diff | head -n 50 >> message.txt
-              echo ``` >> message.txt
-              echo See [action log]\(https://github.com/microsoft/DirectXShaderCompiler/actions/runs/$GITHUB_RUN_ID/job/$GITHUB_JOB\) for the full diff.  >> message.txt
+              echo \`\`\` >> message.txt
+              echo See [action log]\(https://github.com/microsoft/DirectXShaderCompiler/actions/runs/$GITHUB_RUN_ID/) for the full diff.  >> message.txt
```
hekota, line 35: The link in the sample provided does not work, is that expected?

See [action log](https://github.com/microsoft/DirectXShaderCompiler/actions/runs/6256158344/job/check-pr-formatting) for the full diff.

damyanp (approved):

.github/workflows/clang-format-checker.yml
llvm-beanz: Oh yes, because that URL points to DXC not my fork where the actual log is.

Issue #5738, opened 2023-09-17
DXC crashes if a type is defined within a namespace with the same name as a built-in type and is later instantiated


2aecfff4:
**Description**

DXC crashes if a type is defined within a namespace with the same name as a built-in type and is later instantiated.

**Steps to Reproduce**

This crashes with both SPIR-V and DXIL.
```hlsl
namespace test {

class Texture2D { };

}

[numthreads(16, 1, 1)]
void main() {
    test::Texture2D texture; // <- crash
}
```
https://godbolt.org/z/dhWdGr9ad

The second one only crashes during SPIR-V generation and only when resource binding is used.
```hlsl
#define CRASH 1

namespace test {

[[vk::binding(0, 1)]] Texture2D<float2> g_textures2D_float2[] : register(t0, space5);

template <typename T>
class Texture2D;

template <>                                                                          
class Texture2D<float2> {                          
    uint m_texture_index;                                                            
                                                                                        
    float2 read(const int2 pos, const int mip_level ) {                            
        const int3 loc = int3(pos.x, pos.y, mip_level);
#if CRASH
        return g_textures2D_float2[this.m_texture_index].Load(loc);
#else
        return (float2)0.f;
#endif
    }                                                                                
};

}

[numthreads(16, 1, 1)]
void main() {
    test::Texture2D<float2> t;
    t.read(int2(0, 0), 0); // <- crash
}
```
https://godbolt.org/z/GGdhbqs7d

**Actual Behavior**

`Program terminated with signal: SIGSEGV`

**Environment**
As of today, DXC on Godbolt returns `ibdxcompiler.so: 1.7(dev;1-549030f9)`.

llvm-beanz:
The crash call stack is:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00000001ac2dc764 libsystem_kernel.dylib
    frame #1: 0x00000001ac313c28 libsystem_pthread.dylib pthread_kill:`pthread_kill + 288
    frame #2: 0x000000010e16706c libdxcompiler.dylib ::raise(int):`::raise(sig=6 0x00000006) + 28 at Signals.inc:532
    frame #3: 0x000000011024918c libdxcompiler.dylib ::abort():`::abort() + 16 at Signals.inc:549
[...]

PR #5763, opened 2023-09-21
Don't string match node type names

llvm-beanz:
This change avoids the need to string match type names for node input and output types by using an unspellable AST attribute that translates the node type to an enum (and back to a string if needed).

### DIFF
--- a/tools/clang/include/clang/Basic/Attr.td
+++ b/tools/clang/include/clang/Basic/Attr.td
@@ -984,7 +984,7 @@ def HLSLNodeTrackRWInputSharing : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
-def HLSLNodeRecord : InheritableAttr {
+def HLSLNodeObject : InheritableAttr {
   let Spellings = []; // No spellings!
   let Subjects = SubjectList<[CXXRecord]>;
   let Documentation = [Undocumented];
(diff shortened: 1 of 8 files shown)


damyanp (approved):

python3kgae (approved):

tools/clang/include/clang/Basic/Attr.td
```diff
@@ -984,6 +984,40 @@ def HLSLNodeTrackRWInputSharing : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
+def HLSLNodeRecord : InheritableAttr {
```
tex3d, line 987: Not all these are technically "Record" objects, so the naming seems slightly misleading.  Maybe "HLSLNodeObject" might be more appropriate.  Also, wouldn't this be something we quickly want to expand to use on all HLSL object types generally?  Could we do that by inheriting this from a more general "HLSLObject" attribute in the future perhaps?

tools/clang/include/clang/Basic/Attr.td
```diff
@@ -984,6 +984,40 @@ def HLSLNodeTrackRWInputSharing : InheritableAttr {
...
+                           "ThreadNodeInputRecord",
+                           "RWThreadNodeInputRecord",
+                           "EmptyNodeInput",
+                           "NodeOutput",
+                           "EmptyNodeOutput",
+                           "NodeOutputArray",
+                           "EmptyNodeOutputArray",
+                           "GroupNodeOutputRecords",
+                           "ThreadNodeOutputRecords"],
+                           ["DispatchNodeInputRecord",
```
tex3d, line 1006: This list and its formatting is ... unfortunate.  And it's ugly to have to define an enum locally for these, then translate between the local attribute's enum type and the `NodeIOKind` enum.  Not that I know of a better solution that works here, mainly because I am not familiar enough with what's available in this TableGen context.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -645,15 +606,62 @@ bool IsHLSLResourceType(clang::QualType type) {
...
+    case HLSLNodeRecordAttr::ThreadNodeOutputRecords:
+      return DXIL::NodeIOKind::ThreadNodeOutputRecords;
+    }
   }
+  return DXIL::NodeIOKind::Invalid;
+}
+
+bool IsHLSLNodeInputType(clang::QualType type) {
+  if (const HLSLNodeRecordAttr *Attr = getNodeAttr(type))
+    return Attr->getType() == HLSLNodeRecordAttr::EmptyNodeInput ||
```
tex3d, line 658: It's a shame this isn't using the native `NodeIOKind` values, since those are organized such that questions like this can be answered more easily (`kind & NodeIOFlags::Input`).

tools/clang/include/clang/Basic/Attr.td
llvm-beanz: In upstream you can embed custom code into attribute definitions in tablegen, so the internal enum can be hidden entirely inside the attribute. I didn't want to make that large of a change here.

tools/clang/lib/AST/HlslTypes.cpp
llvm-beanz: I could rewrite it that way if you prefer, but I liked the explicitness of the specific types.

tools/clang/include/clang/Basic/Attr.td
llvm-beanz: Yes, we could have an inheritance structure of these attributes that would allow for more general application.

In terms of expanding this to other HLSL objects, I think that depends a lot on resourcing. There are clearly bugs caused by our string matching (#5738 is a recent example), but cost-benefit wise I'm not sure where this will fit on our schedule.

tools/clang/include/clang/Basic/Attr.td
llvm-beanz: Oh! I was actually wrong, it was just not hooked up where I expected it to be. Let me do a quick bit of cleanup...

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -704,32 +664,27 @@ bool IsHLSLObjectWithImplicitROMemberAccess(clang::QualType type) {
...
-  if (const RecordType *RT = type->getAs<RecordType>()) {
-    StringRef name = RT->getDecl()->getName();
-    if (name == "RWDispatchNodeInputRecord" ||
-        name == "RWGroupNodeInputRecords" || name == "RWThreadNodeInputRecord")
-      return true;
-  }
+  if (const HLSLNodeObjectAttr *Attr = getNodeAttr(type))
+    return Attr->getType() == HLSLNodeObjectAttr::RWDispatchNodeInputRecord ||
+           Attr->getType() == HLSLNodeObjectAttr::RWGroupNodeInputRecords ||
+           Attr->getType() == HLSLNodeObjectAttr::RWThreadNodeInputRecord;
```
pow2clk, line 670: I know you reluctantly changed the above, but for consistency we could do something like this: 

```c++
return (static_cast<uint32_t>(GetNodeIOType(type)) &
          (static_cast<uint32_t>(DXIL::NodeIOFlags::ReadWrite) |  static_cast<uint32_t>(DXIL::NodeIOFlags::Input))) ==
          (static_cast<uint32_t>(DXIL::NodeIOFlags::ReadWrite) |  static_cast<uint32_t>(DXIL::NodeIOFlags::Input));
```

hekota (approved):

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -704,32 +664,27 @@ bool IsHLSLObjectWithImplicitROMemberAccess(clang::QualType type) {
...
-  if (const RecordType *RT = type->getAs<RecordType>()) {
-    StringRef name = RT->getDecl()->getName();
-    if (name == "DispatchNodeInputRecord" || name == "GroupNodeInputRecords" ||
-        name == "ThreadNodeInputRecord")
-      return true;
-  }
+  if (const HLSLNodeObjectAttr *Attr = getNodeAttr(type))
+    return Attr->getType() == HLSLNodeObjectAttr::DispatchNodeInputRecord ||
+           Attr->getType() == HLSLNodeObjectAttr::GroupNodeInputRecords ||
+           Attr->getType() == HLSLNodeObjectAttr::ThreadNodeInputRecord;
```
pow2clk, line 678: As above, to save you the trouble: 

```c++

```c++
return (static_cast<uint32_t>(GetNodeIOType(type)) &
          (static_cast<uint32_t>(DXIL::NodeIOFlags::ReadWrite) |  static_cast<uint32_t>(DXIL::NodeIOFlags::Input))) ==
           static_cast<uint32_t>(DXIL::NodeIOFlags::Input);
```

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -704,32 +664,27 @@ bool IsHLSLObjectWithImplicitROMemberAccess(clang::QualType type) {
...
-    StringRef name = RT->getDecl()->getName();
-    if (name == "NodeOutput" || name == "NodeOutputArray" ||
-        name == "EmptyNodeOutput" || name == "EmptyNodeOutputArray")
-      return true;
-  }
+  if (const HLSLNodeObjectAttr *Attr = getNodeAttr(type))
+    return Attr->getType() == HLSLNodeObjectAttr::NodeOutput ||
+           Attr->getType() == HLSLNodeObjectAttr::NodeOutputArray ||
+           Attr->getType() == HLSLNodeObjectAttr::EmptyNodeOutput ||
+           Attr->getType() == HLSLNodeObjectAttr::EmptyNodeOutputArray;
```
pow2clk, line 687:
```c++
  return (static_cast<uint32_t>(GetNodeIOType(type)) &
          static_cast<uint32_t>(DXIL::NodeIOFlags::Output)) != 0;
```

pow2clk (commented):
Just a few suggestions. 3 of them are really extrapolating from Tex's suggestion.

tools/clang/lib/AST/HlslTypes.cpp
tex3d: I believe this needs be:
```cpp
  return (static_cast<uint32_t>(GetNodeIOType(type)) &
          (static_cast<uint32_t>(DXIL::NodeIOFlags::Output) |
           static_cast<uint32_t>(DXIL::NodeIOFlags::RecordGranularityMask))) ==
         static_cast<uint32_t>(DXIL::NodeIOFlags::Output);
```
To avoid the output record type, since this should only return true if it's an output node object type.

tools/clang/include/clang/Basic/Attr.td
tex3d: Ok, at least things that access the local enum are localized.  Still would be easy to make a mistake in lining up the enum strings/names with this formatting without easily seeing it in code review.

tex3d (approved):

tools/clang/include/clang/Basic/Attr.td
```diff
@@ -1016,6 +1016,83 @@ def HLSLNodeRecord : InheritableAttr {
...
+      case hlsl::DXIL::NodeIOKind::NodeOutputArray:
+        return NodeOutputArray;
+      case hlsl::DXIL::NodeIOKind::EmptyOutputArray:
+        return EmptyNodeOutputArray;
+      case hlsl::DXIL::NodeIOKind::GroupNodeOutputRecords:
+        return GroupNodeOutputRecords;
+      case hlsl::DXIL::NodeIOKind::ThreadNodeOutputRecords:
+        return ThreadNodeOutputRecords;
+      }
+      llvm_unreachable("all cases exhausted");
```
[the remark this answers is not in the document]
llvm-beanz: Yes, all possible cases are handled and an unreachable is thrown.

tools/clang/lib/AST/HlslTypes.cpp
llvm-beanz: I think this loses a lot of readability...

PR #5766, opened 2023-09-21
Use GITHUB_SHA instead of GITHUB_REF

llvm-beanz:
Based on GitHubs docs, I think `GITHUB_SHA` will be the last commit from the target ref that is included in the PR's base ref. That should be the right compare point rather than `GITHUB_REF`, which would be the target branch's head ref.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-w orkflows#pull_request_target


hekota (approved):

pow2clk (approved):
Looks plausible. I think we have little choice but to keep iterating here. 

I wonder if we should reword the name from "Comparing PR against target branch" to something like "Comparing PR against the reference point"?

PR #5767, opened 2023-09-21
Fixed incorrect BlockSize in PDB header

### LINKED ISSUE #3005: Generated separate PDB files have possibly invalid header
The generated separate PDB files with e.g. `-Fd` or `CompileWithDebug` seem to have an invalid PDB header.

Taking a simple shader like this:

```
float4 main() : SV_Target0
{
  return float4(0,0,0,0);
}
```

And compiling with latest dxc:

```
$ ./bin/dxc -help | grep Version
Version: dxcompiler.dll: 1.6 - 1.5.0.2616 (6ef33dce)
$ ./bin/dxc -Zi -Qstrip_debug -Zsb input.hlsl -T ps_6_0 -Fd /path/to/output/ -Fo /path/to/output/a.dxbc
warning: DXIL.dll not found.  Resulting DXIL will not be signed for use in release environments.
```

I get a PDB file `7efd93578d970ff351afc3ab6a3d8b4a.pdb` which is 5120 bytes long with this header:

```
00000000  4d 69 63 72 6f 73 6f 66  74 20 43 2f 43 2b 2b 20  |Microsoft C/C++ |
00000010  4d 53 46 20 37 2e 30 30  0d 0a 1a 44 53 00 00 00  |MSF 7.00...DS...|
00000020  00 02 00 00 01 00 00 00  09 00 00 00 30 00 00 00  |............0...|
00000030  00 00 00 00 03 00 00 00  00 00 00 00 00 00 00 00  |................|
00000040  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
```

So in particular the page count at 0x28 is 9, even though the file is 10 pages long - 5120 bytes.

The PDB is rather sparsely documented but I believe the page size (0x200 - 512 bytes) multiplied by the page count should equal the total file size.

Looking briefly at the code, the calculation of the number of pages/blocks seems off:

https://github.com/microsoft/DirectXShaderCompiler/blob/6ef33dcec6c8f7859d1998b642219e2358efa2a2/lib/DXIL/DxilPDB.cpp#L133

This accounts for 3 fixed blocks - superblock and two empty free block maps, plus the stream blocks and the stream directory blocks. However there's also at least one other block written for the block map. The list of blocks containing the stream directory which doesn't seem to have been counted:

https://github.com/microsoft/DirectXShaderCompiler/blob/6ef33dcec6c8f7859d1998b642219e2358efa2a2/lib/DXIL/DxilPDB.cpp#L201-L214

In most cases this will probably mean that the page count is off-by-one as seen here, but it could be off by more. It's not clear what readers are expected to do if this page count is wrong since it can equally be derived by `FileSize / PageSize`.

adam-yang:
Fixes #3005.

The PDB writer did not include in `NumBlocks` the blocks in the `BlockAddr` data, which points to the contents of the (not necessarily continguous) StreamDirectory data. With this bug, our PDBs' `NumBlocks` is always at least 1 less than it should be.

The purpose of wrapping shader PDB as a MSF file is to allow them to be stored in symbol servers. Symsrv does not check this property currently, but it's best to fix this in case something changes in the future.


lib/DXIL/DxilPDB.cpp
```diff
@@ -186,15 +176,29 @@ struct MSFWriter {
...
+    const uint32_t StreamDirectoryStart = BlockAddrStart + BlockAddrNumBlocks;
+    const uint32_t StreamStart =
+        StreamDirectoryStart + StreamDirectoryNumBlocks;
+
+    MSF_SuperBlock SB = {};
+    {
+      memcpy(SB.MagicBytes, kMsfMagic, sizeof(kMsfMagic));
+      SB.BlockSize = kMsfBlockSize;
+      SB.NumDirectoryBytes = StreamDirectorySizeInBytes;
+      SB.NumBlocks = 3 /*super block + FPM1 + FPM2*/ + m_NumStreamBlocks +
```
pow2clk, line 197: I see that this is the 3 explanation

lib/DXIL/DxilPDB.cpp
```diff
@@ -186,15 +176,29 @@ struct MSFWriter {
...
+                     StreamDirectoryNumBlocks + BlockAddrNumBlocks;
```
pow2clk, line 198: Just because it was a little hard to glean amongst all the renaming and refactoring of value calculations. I think the only functional change here is that BlockAddrNumBlocks is now included in this calculation. That seems consistent with the description, but I wanted to confirm my thinking.

tools/clang/test/DXC/old_pdb.test
```diff
@@ -0,0 +1,7 @@
+// RUN: %dxc /dumpbin %S/Inputs/old_pdb.pdb | FileCheck %s
+
+// CHECK: define void @main
```
pow2clk, line 3: We really want to move away from tests that just test for `@main`. Could you check the sizes or something?

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2054,6 +2056,107 @@ TEST_F(CompilerTest, CompileThenTestPdbUtilsRelativePath) {
...
+           L"/Tps_6_0",
+           L"/Zi",
+       }},
+      {source_0,
+       {
+           L"/Tps_6_0",
+           L"/Zs",
+       }},
+      {source_1, {L"/Tps_6_0", L"/Zi", L"Od"}},
+      {source_1, {L"/Tps_6_0", L"/Zs", L"Od"}},
```
pow2clk, line 2105: I really can't fathom why clang-format does these sort of things. Why are the longer lines kept on a single line? 🤦

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2054,6 +2056,107 @@ TEST_F(CompilerTest, CompileThenTestPdbUtilsRelativePath) {
...
+
+    CComPtr<IDxcBlob> pPdb;
+    VERIFY_SUCCEEDED(
+        pResult->GetOutput(DXC_OUT_PDB, IID_PPV_ARGS(&pPdb), nullptr));
+
+    static const char kMsfMagic[] = {
+        'M', 'i', 'c',  'r',  'o',    's', 'o', 'f',  't',  ' ', 'C',
+        '/', 'C', '+',  '+',  ' ',    'M', 'S', 'F',  ' ',  '7', '.',
+        '0', '0', '\r', '\n', '\x1a', 'D', 'S', '\0', '\0', '\0'};
+    struct MSF_SuperBlock {
```
pow2clk, line 2127: I wonder if we could define this in DxilPDB.h and include that header here as well as in DxilPDB.cpp?

pow2clk (commented):
As is typical for my reviews lately, I feel more strongly about the test comments than the implementation comments

tools/clang/test/DXC/old_pdb.test
adam-yang: The only thing we want to test is the ability to read the PDB. If we read the DXIL module out of the PDB, we've successfully done that. I don't think we have any command-line utilities that do any PDB specific things that we can run.

tools/clang/unittests/HLSL/CompilerTest.cpp
adam-yang: Done.

lib/DXIL/DxilPDB.cpp
adam-yang: Moved that magic number to a constant now.

lib/DXIL/DxilPDB.cpp
adam-yang: Moved the magic number to a constant.

lib/DXIL/DxilPDB.cpp
```diff
@@ -186,15 +176,29 @@ struct MSFWriter {
...
-    const uint32_t StreamStart = StreamDirectoryStart + NumDirectoryBlocks;
+    const uint32_t StreamDirectorySizeInBytes = CalculateStreamDirectorySize();
+    const uint32_t StreamDirectoryNumBlocks =
+        GetNumBlocks(StreamDirectorySizeInBytes);
+
+    const uint32_t BlockAddrSizeInBytes =
+        StreamDirectoryNumBlocks * sizeof(support::ulittle32_t);
+    const uint32_t BlockAddrNumBlocks = GetNumBlocks(BlockAddrSizeInBytes);
+
+    const uint32_t BlockAddrStart = 3;
```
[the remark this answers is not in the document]
adam-yang: See other comments

damyanp:

damyanp:

PR #5776, opened 2023-09-22
WinPIX: Find correct filename for subroutines in included headers within namespaces

jeffnn:
From a customer bug report, the previous code would skip over a DISubProgram scope if that scope was contained in a DINamespace (as in the first test case herein, wherein an include file is enclosed in a namespace). This would result in DXIL being attributed to the includer file instead of the included file.


tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -3267,6 +3276,184 @@ void main()
   VERIFY_IS_TRUE(FoundTheVariable);
 }
 
+class DxcBlobImpl : public IDxcBlob {
```
tex3d, line 3279: You can use blob creation methods of `IDxcUtils` instead of rolling your own blob type.

tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -3267,6 +3276,184 @@ void main()
...
+  }
+  virtual LPVOID STDMETHODCALLTYPE GetBufferPointer(void) override {
+    return m_data.data();
+  }
+  virtual SIZE_T STDMETHODCALLTYPE GetBufferSize(void) override {
+    return static_cast<SIZE_T>(m_data.size() * sizeof(m_data[0]));
+  }
+};
+
+class DxcIncludeHandlerForInjectedSources : public IDxcIncludeHandler {
```
tex3d, line 3298: Note: this is the same name (in global scope) used in dxc.cpp in dxclib project.  I don't think there's a collision, but maybe there could be if one were to link to that lib from the test.

tools/clang/unittests/HLSL/PixTest.cpp
jeffnn: Right you are. In fact, now that I think about it, I already have that in pixtest.

tools/clang/unittests/HLSL/PixTest.cpp
jeffnn: Indeed. Renamed!

tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -3269,6 +3278,168 @@ void main()
...
+  }
+
+  HRESULT insertIncludeFile(_In_ LPCWSTR pFilename,
+                            _In_ IDxcBlobEncoding *pBlob, _In_ UINT32 dataLen) {
+    return E_FAIL;
+  }
+
+  HRESULT STDMETHODCALLTYPE LoadSource(
+      _In_ LPCWSTR pFilename,
+      _COM_Outptr_result_maybenull_ IDxcBlob **ppIncludeSource) override {
```
python3kgae, line 3305: I thought we want to remove these _In_  and _COM_Outptr_result_maybenull_?

jeffnn: ah yes! I have removed all the SAL

python3kgae (approved):

tex3d (approved):
LGTM!

PR #5777, opened 2023-09-22
Ignore line endings for sources verification

dBagrat:
The issue arose during builds on specific macOS configurations, where the build would fail due to different line-ending conventions in compared files

```bash
[134/1556] cd /Users/user/.conan2/p/b/dxce4c427e7bc573/b/include/dxc && /usr/local/Cellar/cmake/3.27.6/bin/cmake -E compare_files /Users/user/.conan2/p/b/dxce4c427e7bc573/b/include/dxc/HlslIntrinsicOp.h.tmp /Users/user/.conan2/p/dxcb193344fb9f24/s/source_subfolder/include/dxc/HlslIntrinsicOp.h
FAILED: include/dxc/CMakeFiles/HLSLIntrinsicOp /Users/user/.conan2/p/b/dxce4c427e7bc573/b/include/dxc/CMakeFiles/HLSLIntrinsicOp
```

### DIFF
--- a/cmake/modules/HCT.cmake
+++ b/cmake/modules/HCT.cmake
@@ -93,7 +93,7 @@ function(add_hlsl_hctgen mode)
   if(NOT copy_sources)
     set(output ${temp_output})
     if (CLANG_FORMAT_EXE) # Only verify sources if clang-format is available.
-      set(verification COMMAND ${CMAKE_COMMAND} -E compare_files ${temp_output} ${full_output})
+      set(verification COMMAND ${CMAKE_COMMAND} -E compare_files --ignore-eol ${temp_output} ${full_output})
     endif()
   endif()
   if(WIN32 AND NOT HLSL_AUTOCRLF)


llvm-beanz:
I build on macOS all the time and don't encounter this issue. The file in the repo is `LF` encoded in the index, a default macOS checkout should also be `LF`, and the script should produce `LF` line endings on macOS.

Can you figure out which part of this isn't working for you?

dBagrat:
> I build on macOS all the time and don't encounter this issue. The file in the repo is `LF` encoded in the index, a default macOS checkout should also be `LF`, and the script should produce `LF` line endings on macOS.
> 
> Can you figure out which part of this isn't working for you?

While its true that the default macOS checkout is `LF`, there are still cases where the checked out file is `CRLF`. 

For example, in our local builds, we have a case where the source files can be read from a remote location, which doesn't guarantee the file to be `LF` (e.g. they are hosted on a Windows machine and accessed via SMB).

This specific case was due to the source files being a part of a `Conan` package, which for some reason decided to check them out in `CLRF` mode. We failed to find a setting to prevent this from happening.

In any case, I consider ignoring the `EOL` format to be "not worse" than the previous behavior, while also being more convenient for, albeit unusual, cases like these.

llvm-beanz:
My concern here is that the line ending verification does help us catch cases that would lead to inconsistent line endings in the git index. As one of the minority unix developers on this project I had a fair amount of pain getting the line endings corrected in the first place.

I think we really want the line-ending sensitive comparison running on our pre-merge check systems to help us keep this clean. I'd be more open to accepting a patch that adds an option to disable the verification step. Having an option to disable verification would allow users with exotic environments like yours to bypass the verification step and still build DXC.

Something like this:

```diff
diff --git a/cmake/modules/HCT.cmake b/cmake/modules/HCT.cmake
index e6fb8908a5b..0b3e4f75d25 100644
--- a/cmake/modules/HCT.cmake
+++ b/cmake/modules/HCT.cmake
@@ -1,4 +1,6 @@
 option(HLSL_COPY_GENERATED_SOURCES "Copy generated sources if different" Off)
+option(HLSL_SKIP_GENERATED_SOURCE_VERIFICATION "Skip verifying generated sources" Off)
+mark_as_advanced(HLSL_SKIP_GENERATED_SOURCE_VERIFICATION)
 
 add_custom_target(HCTGen)
 
@@ -92,7 +94,9 @@ function(add_hlsl_hctgen mode)
   # file, and define the verification command
   if(NOT copy_sources)
     set(output ${temp_output})
-    if (CLANG_FORMAT_EXE) # Only verify sources if clang-format is available.
+    # Only verify sources if clang-format is available, and verification isn't
+    # disabled explicitly.
+    if (CLANG_FORMAT_EXE AND NOT HLSL_SKIP_GENERATED_SOURCE_VERIFICATION) 
       set(verification COMMAND ${CMAKE_COMMAND} -E compare_files ${temp_output} ${full_output})
     endif()
   endif()
```

damyanp:

damyanp:

PR #5778, opened 2023-09-22
Fixing documentation code tag for BuildingAndTestingDXC.rst 

kecho:
Fixing documentation code tag for BuildingAndTestingDXC.rst doc file, Lack of new line between code and content makes the code dissapear. Removed also new line escape.
Sample sh code is now visible.

Results:
image

### DIFF
--- a/docs/BuildingAndTestingDXC.rst
+++ b/docs/BuildingAndTestingDXC.rst
@@ -32,7 +32,8 @@ for the configuration process.
 
 All of the most basic CMake configurations for DXC follow a similar format to:
 
-.. code:: sh
+.. code-block:: sh
+
   cmake <Repository Root> \
     -C <Repository Root>/cmake/caches/PredefinedParams.cmake \
     -DCMAKE_BUILD_TYPE=<Build Type> \
@@ -52,7 +53,8 @@ Generating a Visual Studio Solution
 
 Open a Visual Stuido command prompt and run:
 
-.. code:: sh
+.. code-block:: sh
+
   cmake <Repository Root> \
     -B <Path to Output> \
     -C <Repository Root>/cmake/caches/PredefinedParams.cmake \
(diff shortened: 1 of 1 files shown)


kecho:

docs/BuildingAndTestingDXC.rst
```diff
@@ -33,9 +33,10 @@ for the configuration process.
 All of the most basic CMake configurations for DXC follow a similar format to:
 
 .. code:: sh
-  cmake <Repository Root> \
-    -C <Repository Root>/cmake/caches/PredefinedParams.cmake \
-    -DCMAKE_BUILD_TYPE=<Build Type> \
+
+  cmake <Repository Root>
```
llvm-beanz, line 37: Instead of removing the `\` can you make them `\\` so they're escaped properly? The `\` needs to be preserved in order for the command lines to be able to be copy and pasted into a terminal.

kecho: on it 👌

kecho:
@llvm-beanz can you check the rich diff? i see the actual '\\'. Maybe would it be better to just put the whole thing in one line? Ugly but might be easier to cpy paste? lmk

kecho:
> Ugh. There’s something funky going on that doesn’t make sense to me.
> 
> I realize this is obnoxious, but can I ask you to try one more thing before throwing it all on one line? If you change the `code` directive to `code-block`, then I think the single `\` should work how it originally was. IIUC that will disable Sphinx’s special character handling for the block.
> 
> If that doesn’t work, throw it all on one line.
 
@llvm-beanz rebased w latest main branch, changed to code-base and is looking good from here. Lmk if ready to push.

llvm-beanz (approved):

bogner (approved):

llvm-beanz:
@kecho, thank you for your patience and help getting this right. I've set the change to merge once the pre-merge tests pass (just as a formality).

PR #5779, opened 2023-09-25
Fixed a crash in unroller when exiting blocks use switch instead of branch

adam-yang:
The code to structurize loop exits assumed exiting blocks always use BranchInst and never SwitchInst. This caused a crash. This change makes the structurizer preemptively check whether this assumption is true before attempting.

### DIFF
--- a/lib/Transforms/Scalar/DxilLoopUnroll.cpp
+++ b/lib/Transforms/Scalar/DxilLoopUnroll.cpp
@@ -137,11 +137,13 @@ public:
     GetPassOptionUnsigned(O, "MaxIterationAttempt", &MaxIterationAttempt,
                           false);
     GetPassOptionBool(O, "OnlyWarnOnFail", &OnlyWarnOnFail, false);
+    GetPassOptionBool(O, "StructurizeLoopExits", &StructurizeLoopExits, false);
   }
   void dumpConfig(raw_ostream &OS) override {
     LoopPass::dumpConfig(OS);
     OS << ",MaxIterationAttempt=" << MaxIterationAttempt;
     OS << ",OnlyWarnOnFail=" << OnlyWarnOnFail;
+    OS << ",StructurizeLoopExits=" << StructurizeLoopExits;
   }
   void RecursivelyRemoveLoopOnSuccess(LPPassManager &LPM, Loop *L);
   void RecursivelyRecreateSubLoopForIteration(LPPassManager &LPM, LoopInfo *LI,
--- a/tools/clang/test/HLSLFileCheck/hlsl/control_flow/attributes/unroll/unroll_switch_exit_crash.ll
+++ b/tools/clang/test/HLSLFileCheck/hlsl/control_flow/attributes/unroll/unroll_switch_exit_crash.ll
@@ -1,90 +1,50 @@
-; RUN: %opt %s -dxil-loop-unroll -S | FileCheck %s
+; RUN: %opt %s -dxil-loop-unroll,StructurizeLoopExits=1 -S | FileCheck %s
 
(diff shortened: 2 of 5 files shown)


tools/clang/test/HLSLFileCheck/hlsl/control_flow/attributes/unroll/unroll_switch_exit_crash.ll
```diff
@@ -0,0 +1,121 @@
+; RUN: %opt %s -dxil-loop-unroll -S | FileCheck %s
```
dmpots, line 1: It looks like we could probably simplify this bitcode quite a bit? Would be nice to have a minimal loop with the switch here to demonstrate the problem. There is lots of dxil-isms and debug info that seem unnecessary to trigger the bug.

lib/Transforms/Scalar/DxilLoopUnroll.cpp
```diff
@@ -137,11 +137,13 @@ class DxilLoopUnroll : public LoopPass {
     GetPassOptionUnsigned(O, "MaxIterationAttempt", &MaxIterationAttempt,
                           false);
     GetPassOptionBool(O, "OnlyWarnOnFail", &OnlyWarnOnFail, false);
+    GetPassOptionBool(O, "StructurizeLoopExits", &StructurizeLoopExits, false);
```
dmpots, line 140: I don't understand why this change is here. I don't see any code reading this option. Is it needed for the fix?

dmpots (approved):

lib/Transforms/Scalar/DxilLoopUnroll.cpp
adam-yang: By default, we don't structurize loop exit, and it's controlled by an -opt-enable, which translates to an argument to DxilLoopUnroll pass. When running dxopt.exe, there was no way to turn on the structurizer, so I added this option here. Plus the option should exist anyway for the argument. Without it, our -Odump is actually wrong.

adam-yang: Oops, I just straight up forgot to add it to the test.

adam-yang: Fixed now.

tools/clang/test/HLSLFileCheck/hlsl/control_flow/attributes/unroll/unroll_switch_exit_crash.ll
adam-yang: Simplified it!

tools/clang/test/HLSLFileCheck/hlsl/control_flow/attributes/unroll/unroll_switch_exit_crash.ll
```diff
@@ -0,0 +1,61 @@
+; RUN: %opt %s -dxil-loop-unroll,StructurizeLoopExits=1 -S | FileCheck %s
```
pow2clk, line 1: This passes for me without crashing with an unaltered compiler and without the StructurizeLoopExits parameter. Is that expected? From the name of the file, I got the impression that this was going to crash

adam-yang: Yes. The crash happens in the structurize code path, which is enabled in dxc.exe through `-opt-enable structurize-loop-exits-for-urnoll` (on by default), but disabled by default on dxopt since there was no way to enable that path without the new parameter.

pow2clk (approved):
Thanks for the explanation!

PR #5782, opened 2023-09-26
[SPIR-V] Add methods to record frequency mode for ROVs

cassiebeckley:
Rasterizer ordered views will need to set an execution mode of either `PixelInterlockOrderedEXT`, `SampleInterlockOrderedEXT`, or `ShadingRateInterlockOrderedEXT`. This PR records which execution mode should be used, based on which semantics are input into the shader.

1. `PixelInterlockOrderedEXT` is the default execution mode
2. `SampleInterlockOrderedEXT` will be chosen if the shader has an input with the `SV_SampleIndex` semantic
3. `ShadingRateInterlockOrderedEXT` will be chosen if the shader has an input with the `SV_ShadingRate` semantic

This PR does not actually add the execution mode to an entry point; that will need to be done only if a rasterizer ordered view is present in the program, and will be done in another PR. As such, this PR has no tests, and its behavior will be tested in the follow-up – I mainly wanted to get your input on whether this looks like a good structure for recording this information.

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -3653,6 +3653,7 @@ SpirvVariable *DeclResultIdMapper::createSpirvStageVar(
   // According to DXIL spec, the SampleIndex SV can only be used by PSIn.
   // According to Vulkan spec, the SampleId BuiltIn can only be used in PSIn.
   case hlsl::Semantic::Kind::SampleIndex: {
+    setInterlockExecutionMode(spv::ExecutionMode::SampleInterlockOrderedEXT);
     stageVar->setIsSpirvBuiltin();
     return spvBuilder.addStageBuiltinVar(type, sc, BuiltIn::SampleId, isPrecise,
                                          srcLoc);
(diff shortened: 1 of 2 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -4204,6 +4207,20 @@ void DeclResultIdMapper::decorateStageVarWithIntrinsicAttrs(
   decorateWithIntrinsicAttrs(decl, varInst, checkBuiltInLocationDecoration);
 }
 
+void DeclResultIdMapper::setInterlockExecutionMode(spv::ExecutionMode mode) {
+  interlockExecutionMode = mode;
+}
+
+spv::ExecutionMode DeclResultIdMapper::getInterlockExecutionMode() {
```
Keenuts, line 4214: Should that be an optional?
As-is, now knowing ROVs, we could understand that every declaration has by default the PixelInterlockOrdered execution mode.

sudonatalie:
I would prefer a design where we don't have additional member vars and functions added to `DeclResultIdMapper` just for this case, but it's difficult to determine why it's necessary or what alternatives might be possible without seeing how it will be used. I do appreciate the effort to split it out but I think in this case it will be easier to review with the code that uses it.

cassiebeckley:
All right, I'll close this and send in the full set of changes.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
cassiebeckley: Yeah, using an optional is probably better than my current approach.

PR #5786, opened 2023-09-26
Add test for FieldAnnotationVectorSize and similar metadata tags

### LINKED ISSUE #5697: Add test for FiledAnnotationVectorSize metadata Tag
Work graphs added a metadata tag for vector sizes to preserve that information even if the vector is converted to an array. There are no tests for it. Here is where the tag is added: 
https://github.com/microsoft/DirectXShaderCompiler/blob/58a9ac3522c899ee52f091de4fd033e12a710afc/lib/DXIL/DxilMetadataHelper.cpp#L1241

A simple test that verifies that this tag is added in the right circumstances.

**Additional context**
Internal PR 4476

hekota:
This test checks that a vector size annotation has been added for vector field. The annotation is embedded in metadata used for reflection. The test is using the `-Qkeep_reflect_in_dxil` flag to keep the metadata in the module to allow metadata check in the test, otherwise it gets stripped from the main module. 

The metadata is embedded only for DXIL version 1.8 and higher.

Added similar metadata checks for matrix and scalar annotations for completeness.

Fixes #5697


tools/clang/test/HLSLFileCheck/hlsl/types/vector/vec_size_annot.hlsl
```diff
@@ -0,0 +1,20 @@
+// RUN: %dxc -T ps_6_8 -E main -Qkeep_reflect_in_dxil -select-validator internal %s | FileCheck %s
+
+// Make sure the vector is annotated with vector size (DXIL 1.8 and higher)
+
+// CHECK: !dx.typeAnnotations
+
+// CHECK: !{i32 0, %"class.StructuredBuffer<MyStruct>" undef, !8, %struct.MyStruct undef, [[ANNOT1:![0-9]+]]}
+// CHECK: [[ANNOT1:![0-9]+]] = !{i32 16, [[ANNOT2:![0-9]+]]}
+// CHECK: [[ANNOT2:![0-9]+]] = !{i32 6, !"vec", i32 3, i32 0, i32 7, i32 9, i32 13, i32 4}
```
pow2clk, line 9: It would be helpful in the future to anyone wanting to alter or add a variant to comment on what the parameters refer to. In particular, FieldAnnotationVectorSize(13) = 4

tools/clang/test/HLSLFileCheck/hlsl/types/vector/vec_size_annot.hlsl
```diff
@@ -0,0 +1,20 @@
...
+
+struct MyStruct {
+    float4 vec;
+};
+
+StructuredBuffer<MyStruct> g_myStruct;
```
pow2clk, line 15: Does this get applied to matrices as well? Maybe we could check that and maybe ensure it doesn't get applied to scalars?

pow2clk (commented):
The title has a typo s/Filed/Field. 

Though it's a simple change, I'd still appreciate a short description of the test.

tools/clang/test/HLSLFileCheck/hlsl/types/vector/vec_size_annot.hlsl
```diff
@@ -0,0 +1,20 @@
+// RUN: %dxc -T ps_6_8 -E main -Qkeep_reflect_in_dxil -select-validator internal %s | FileCheck %s
```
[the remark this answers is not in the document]
tex3d: This might highlight a potential deficiency in our lit shell test path in comparison to our TAEF FileCheck path.  In the TAEF FileCheck path, this is automatically inferred from the -T argument on the dxc command line.  I don't think we have the equivalent in the lit shell test path.

tools/clang/test/HLSLFileCheck/hlsl/types/vector/vec_size_annot.hlsl
tex3d: BTW: Matrix has a special annotation type already.

tools/clang/test/HLSLFileCheck/hlsl/types/vector/vec_size_annot.hlsl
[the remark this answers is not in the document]
tex3d: BTW: this test has -select-validator internal, so it doesn't technically need any `%dxilver` check.  It's intended to skip external validation, and embed the reflection into the main module to allow metadata check in the test.  Otherwise, that metadata would be stripped from the main module, and only exist in the reflection module which isn't disassembled.

tex3d (approved):
LGTM.

tools/clang/test/HLSLFileCheck/hlsl/types/vector/vec_size_annot.hlsl
hekota: I have added test case for matrix and scalar as well, and also SM 6.7 variant (the vector size annotation should not be present when compiled with SM 6.7.)

dmpots:
I think there's a typo in the description

> The test it using

it -> is?

dmpots (approved):

Issue #5747, opened 2023-09-19
Library entry points are missing many diagnostics


pow2clk:
**Description**
Most entry point diagnostics are skipped in SemaHLSL when compiling to libraries. 
https://github.com/microsoft/DirectXShaderCompiler/blob/d949f78f69c0620fdba7320c2014424f4ecdd14f/tools/clang/lib/Sema/SemaHLSL.cpp#L10592

The results manifest differently, but are all incorrect and have essentially the same root cause. We should reengineer how the entry point diagnostics are detected and produced so they can be produced for library shaders too. 

Additionally, these all should have validation errors as a final catchall, but that is not going to be a single solution and may require a new shader model revision. 

**Steps to Reproduce with Actual Behavior**

Since this manifests through a number of failures that should all be fixed with a reengineering of how the entry point diagnostics are detected, I will list the link to the shaders with the results together
There are a number of failures that result from this: 
* Hull shader without patchconstantfunc attrib
   * https://godbolt.org/z/8hEao3v9E
   * result: segfaults (AV). The lack of error proceeds to codegen that expects the function to be there and derefs null if it isn't
* Geometry shader without maxvertexcount
   * https://godbolt.org/z/96bz6EWex
   * compiles successfully while it should produce an error 
* Mesh shader without numthreads
  * https://godbolt.org/z/39oz5v5nd
  * produces validation errors instead of sema errors
* Mesh shader without outputtopology
  * https://godbolt.org/z/oK9T1GYrh
  * compiles successfully instead of producing errors
* Amplification shader without numthreads
  * https://godbolt.org/z/sqoooGGGY
  * (Can we take a moment to acknowledge how great the hash in that link is? 😆)
  * compiles successfully instead of producing errors
  * I really expected this to produce validation errors like the mesh shader
* recursive functions in shader
  * https://godbolt.org/z/KahGbd6Kj
  * compiles successfully instead of producing errors
  * This one has a comment mentioning it that should be removed when this is fixed: https://github.com/microsoft/DirectXShaderCompiler/blob/d949f78f69c0620fdba7320c2014424f4ecdd14f/tools/clang/lib/Sema/SemaHLSL.cpp#L10593
  * Unlike the above, which only need be done for functions annotated with shader attributes, this would be needed for any exported functions

**Environment**
- DXC version 1.7.2308
- Host Operating System win11 + godbolt

pow2clk:
[...]

PR #5789, opened 2023-09-27
Improve library profile diagnostics

bob80905:
There are several diagnostics that are missing when a function declaration that targets a specific shader stage is missing a necessary attribute. For example, if a hull shader function declaration does not have the patchconstantfunc attribute, there is no diagnosis of the missing attribute in sema. This PR addresses this diagnostic hole by filling in some functions that check that all required attributes exist for a given shader stage. 
This PR partially addresses #5747


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15419,6 +15459,14 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
...
 }
 
+// The DiagnoseEntry function does 3 things:
+// 1. Attempt to add the target profile as an implicit shader attribute which is
+// equivalent to determining whether or not this function is the current active
+// entry point or not.
+// 2. Diagnose whether or not all entry point attributes on
+// the decl belong on the decl given the specific target profile being used.
+// 3. Given the specific target profile, verifying that all necessary attributes
+// that must accompany that target profile exist as attributes to the decl.
```
tex3d, line 15469: Point 1 doesn't sound quite right to me: "which is equivalent to ..."
Point 2 as well: "given the specific target profile being used"
Point 3 is too attribute centric.  These should be verifying any constraints we can now that we have the full entry function decl, including both parameters and attributes.

Here's an attempt at a rewrite:

Determine whether this function is the current entry point for a non-library compilation, add an implicit shader attribute if so.
For an entry point function, now identified by the shader attribute, diagnose entry point constraints:
- diagnose whether entry point attributes are allowed on the entry point type (ShaderKind) at all.
- diagnose the full entry point decl for required attributes, constraints on or between attributes and parameters, and more.

tools/clang/test/CodeGenHLSL/attributes-hs-no-pcf.hlsl
llvm-beanz, line 1: Can we move these tests into `test/SemaHLSL/attributes` and have them be run using `%dxc ... %s -verify` instead?

I strongly suspect if we did that these tests would fail because it looks to me like we're emitting errors for this twice because the old code is still there.

tex3d:
Shouldn't there be corresponding diagnostic code deleted from `AddHLSLFunctionInfo()` in `CGHLSLMS.cpp`?

tools/clang/test/CodeGenHLSL/attributes-hs-no-pcf.hlsl
tex3d: Agreed.

Also, many entry point diagnostics can be checked in a single HLSL file with a lib target, rather than having one file per scenario.  With this new codepath, the entry point checks are the same code path between lib and normal shader entry cases, making it possible to separate the unit tests for adding the shader attribute for the non-lib case from verification of all the entry point rules.

bob80905:
> Shouldn't there be corresponding diagnostic code deleted from `AddHLSLFunctionInfo()` in `CGHLSLMS.cpp`?

No, because the only thing CGHLSLMS.cpp does is check if the attribute exists, and if it does, verify that it is on a valid shader stage. It does not ever first check the shader stage, and validate that all necessary/required attributes exist.
Regardless, I agree the tests are worth moving to `test/SemaHLSL/attributes`

llvm-beanz:
`SemaHLSL.cpp` `hlsl::DiagnoseTranslationUnit` does have duplicate emission of these errors, so we probably do need some cleanup.

pow2clk (approved):
Looks good!

pow2clk:
I don't know why this can't go directly into main. I don't see any mention of nodes or other 6.8 features.

bob80905:
There seem to be some clang format issues. Perhaps I can cherry pick this over to main once it's in staging.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11388,15 +11388,7 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
-        self->Diag(pEntryPointDecl->getLocation(), diag::err_hlsl_missing_attr)
-            << "GS"
-            << "maxvertexcount";
-        return;
-      }
-    } else if (shaderModel->IsHS()) {
+    if (shaderModel->IsHS()) {
       if (const HLSLPatchConstantFuncAttr *Attr =
               pEntryPointDecl->getAttr<HLSLPatchConstantFuncAttr>()) {
         NameLookup NL = GetSingleFunctionDeclByName(
```
tex3d, line 11394: Shouldn't this code be moved into `DiagnoseHullEntry`?

bob80905: The reason I can't move this block is because pPatchFnDecl is set by this block, and is used below when checking for recursion. If I moved it into DiagnoseHullEntry, I would need to repeat this code here anyways in order to set pPatchFnDecl.

tex3d: I realize that this code also captures the patch constant function for recursion check later.  I think all this needs to be moved somewhere else eventually, since we need the recursion check on all functions (but also don't want to have to pay a cost of re-checking the same call graph over and over again).

Or is the reason this can't be moved also related to timing - maybe the patch constant function won't have been created yet when we are in `DiagnoseEntry`, for example?

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -153,8 +153,6 @@ class ValidationTest : public ::testing::Test {
   TEST_METHOD(FunctionAttributes)
   TEST_METHOD(GSMainMissingAttributeFail)
   TEST_METHOD(GSOtherMissingAttributeFail)
-  TEST_METHOD(GSMissingMaxVertexCountFail)
-  TEST_METHOD(HSMissingPCFFail)
```
tex3d, line 157: These weren't validation tests, so glad to see them moved, but nor are `GSMainMissingAttributeFail` or `GSOtherMissingAttributeFail`.  A validation test is supposed to be a test of `DxilValidation`, not of clang diagnostics, which were typically tested in `VerifierTest` or newer lit shell test with `-verify`, not `ValidationTest`.  Now these are currently diagnosed during CodeGen, so it's understandable that they can't (yet) be moved to verifier, but they should have been FileCheck tests in the first place.

If there's a task to move these in a future PR, that's fine.  I just wanted to point this out.

tools/clang/test/SemaHLSL/attributes/NecessaryAttribsInLib.hlsl
```diff
@@ -0,0 +1,52 @@
...
+
+struct myvert {
+    float4 position : SV_Position;
+};
+
+
+[shader("mesh")]
+//[NumThreads(8, 8, 2)]
+//[OutputTopology("triangle")]
+void MSmain(out vertices myvert verts[32], /* expected-error{{mesh entry point must have the numthreads attribute}} */ /* expected-error{{mesh entry point must have the outputtopology attribute}} */
```
tex3d, line 47: Tip: for better readability, you can use syntax like: `expected-error@+2`, for instance:
```hlsl
// expected-error@+2{{mesh entry point must have the numthreads attribute}}
// expected-error@+1{{mesh entry point must have the outputtopology attribute}}
void MSmain(out vertices myvert verts[32],
```

tex3d (approved):

PR #5793, opened 2023-09-27
[NFC] Remove unneeded IsComplete parameter

llvm-beanz:
This function is always called with `IsComplete = true`, so there is no need for this parameter.


tex3d (approved):
LGTM.

bob80905 (approved):

PR #5794, opened 2023-09-28
Fix UBSAN error in iterator.h

amaiorano:
Fixes UBSAN error:
```
include/llvm/ADT/iterator.h:171:7: runtime error: applying non-zero offset 8 to null pointer
```

When ULE has no decls, `decls_begin` returns a null-ptr iterator, and
adding 1 to this pointer is undefined behaviour. We fix this by adding a
check that ULE actually has decls before doing this.

Fixes 332 of these reported from running check-all.

### DIFF
--- a/tools/clang/lib/Sema/SemaOverload.cpp
+++ b/tools/clang/lib/Sema/SemaOverload.cpp
@@ -10936,7 +10936,11 @@ bool Sema::buildOverloadedCallSet(Scope *S, Expr *Fn,
     // We don't perform ADL for implicit declarations of builtins.
     // Verify that this was correctly set up.
     FunctionDecl *F;
-    if (ULE->decls_begin() + 1 == ULE->decls_end() &&
+    if (
+        // HLSL change begin
+        (ULE->getNumDecls() > 0) &&
+        // HLSL change end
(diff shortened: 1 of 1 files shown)


tools/clang/lib/Sema/SemaOverload.cpp
```diff
@@ -10936,7 +10936,8 @@ bool Sema::buildOverloadedCallSet(Scope *S, Expr *Fn,
     // We don't perform ADL for implicit declarations of builtins.
     // Verify that this was correctly set up.
     FunctionDecl *F;
-    if (ULE->decls_begin() + 1 == ULE->decls_end() &&
+    if ((ULE->getNumDecls() > 0) &&
```
llvm-beanz, line 10939: nit: Can you wrap this in a comment block to denote that it is a `// HLSL change`?

llvm-beanz (approved):
One small nit, otherwise good.

tools/clang/lib/Sema/SemaOverload.cpp
amaiorano: Done.

s-perron (approved):

PR #5795, opened 2023-09-28
Fix UBSAN error in DxilRootSignatureSerializer

amaiorano:
Fix UBSAN error:
```
lib/DxilRootSignature/DxilRootSignatureSerializer.cpp:134:5: runtime error: store to misaligned address 0x7fff812daf5f for type 'unsigned int', which requires 4 byte alignment
```

Use memcpy to write to unaligned address.

Fixes 63 of these reported from running check-all.


llvm-beanz (approved):
LGTM

s-perron (approved):

PR #5796, opened 2023-09-28
Error when declaring global operator overloads

### LINKED ISSUE #5792: Global (non-member) operator overload declarations are allowed but fail to be resolved in HLSL 2021.
**Description**
There is no error when trying to declare a global (non-member) operator overload in HLSL 2021.

**Steps to Reproduce**
Minimal reproduction: https://shader-playground.timjones.io/43640441be664659ae88414a8ad057ff

**Actual Behavior**
Both operator overloads compile, but the global operator only fails and produces an error at the callsite.

**Environment**
- DXC version - 1.7.2308.7
- Host Operating System - Windows 10 21H2

llvm-beanz:
HLSL 2021 can't resolve calls to global operator overloads and was never intended to support them, but it didn't error on them being declared. This change emits a diagnostic when declaring any operator outside class or struct scope so that users won't instead be greeted by less intuitive errors from overload resolution.

Fixes #5792

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -11175,7 +11175,9 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
     // Check if it's an entry function the hard way
     if (!FD->getDeclContext()->isNamespace() && FD->isGlobal()) {
       // Check if this is an entry function by comparing name
-      if (FD->getName() == sema.getLangOpts().HLSLEntryFunction) {
+      // TODO: Remove this once we put HLSLShaderAttr on all entry functions.
+      if (FD->getIdentifier() &&
+          FD->getName() == sema.getLangOpts().HLSLEntryFunction) {
         return true;
       }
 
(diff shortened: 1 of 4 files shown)


coopp (approved):
Looks good to me.

damyanp (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11175,7 +11175,9 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
     // Check if it's an entry function the hard way
     if (!FD->getDeclContext()->isNamespace() && FD->isGlobal()) {
       // Check if this is an entry function by comparing name
-      if (FD->getName() == sema.getLangOpts().HLSLEntryFunction) {
+      // TODO: Remove this once we put HLSLShaderAttr on all entry functions.
+      if (FD->getIdentifier() &&
```
dmpots, line 11179: Is this needed for the fix or just an opportunistic clean up?

dmpots (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: Amusingly it is needed. The new test case asserts on this line without this change because the global operator functions come through here and they don't have valid identifiers.

We didn't hit this in our existing tests because none of our tests tried to declare global operators, and users haven't hit it because we ship our release builds without asserts enabled.

PR #5797, opened 2023-09-28
Add is normal validation

bob80905:
Cherry-picking #5568 so that it goes into the latest release branch.


pow2clk:
The better way to do this would be to cherry pick the specific hash that went into main instead of the two commits that this was before it got squashed. Additionally, if you include `-x` in the cherry pick, it will populate the hash for you.

pow2clk:
`git cherry-pick -x 3974b57d12cf9c1e30d9da030820020c5a22b111`

PR #5798, opened 2023-09-28
[ExecutionTests] Add validation and signing step to isNormal Test CherryPick

bob80905:
The IsNormal tests required an addition to the testing framework
provided by ShaderOpTest.cpp, known as the MakeShaderReplacementCallback
function. This function takes a shader, compiles it to DXIL,
disassembles it into a string, replaces anything specified in the
function parameters, and reassembles the shader back into a dxil
container. The problem with the initial implementation was that after
disassembling the shader, the signature applied to the shader on the
compile step would be lost.

This PR adds a validation / signing step right after assembling the
shader to ensure the new DXIL container has a signature. This will
prevent problems with calling CreateComputePipelineState due to a
missing signature.
Verified that this fails when disabling experimental mode, and passes
only when dxil.dll is available.
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5337

(cherry picked from commit https://github.com/microsoft/DirectXShaderCompiler/commit/3974b57d12cf9c1e30d9da030820020c5a22b111)

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -11508,12 +11508,26 @@ st::ShaderOpTest::TShaderCallbackFn MakeShaderReplacementCallback(
     {
       CComPtr<IDxcAssembler> pAssembler;
       CComPtr<IDxcOperationResult> pResult;
-      VERIFY_SUCCEEDED(dllSupport.CreateInstance(CLSID_DxcAssembler, &pAssembler));
-      VERIFY_SUCCEEDED(pAssembler->AssembleToContainer(rewrittenDisassembly, &pResult));
+      CComPtr<IDxcOperationResult> pValidationResult;
+      CComPtr<IDxcValidator> pValidator;
+
       HRESULT status;
+      HRESULT validationStatus;
+      VERIFY_SUCCEEDED(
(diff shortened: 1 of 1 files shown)


tex3d:
The default commit message for this PR will be based on the PR description.  Please update to match the cherry-picked commit description (which already includes the "cherry picked from commit" detail).

tex3d:
> The default commit message for this PR will be based on the PR description. Please update to match the cherry-picked commit description (which already includes the "cherry picked from commit" detail).

Thanks for the update, but it's missing this from the cherry-pick commit:

(cherry picked from commit 3974b57)

I think you copied the description from the original commit, rather than cherry-picked commit.

tex3d (approved):

tex3d:
> > The default commit message for this PR will be based on the PR description. Please update to match the cherry-picked commit description (which already includes the "cherry picked from commit" detail).
> 
> Thanks for the update, but it's missing this from the cherry-pick commit:
> 
> (cherry picked from commit [3974b57](https://github.com/microsoft/DirectXShaderCompiler/commit/3974b57d12cf9c1e30d9da030820020c5a22b111))
> 
> I think you copied the description from the original commit, rather than cherry-pick commit.

It's fixed, thanks!

pow2clk (approved):

PR #5802, opened 2023-09-29
Allow SamplerComparisonState when CalculateLevelOfDetail

python3kgae:
New overloads are added for CalculateLevelOfDetail and CalculateLevelOfDetailUnclamped. Also update validator to allow SamplerComparisonState for shader model higher than 6.7.

Fixes https://github.com/microsoft/DirectXShaderCompiler/pull/5183

### DIFF
--- a/lib/HLSL/DxilValidation.cpp
+++ b/lib/HLSL/DxilValidation.cpp
@@ -1658,7 +1658,9 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
   case DXIL::OpCode::CalculateLOD: {
     DxilInst_CalculateLOD lod(CI);
     Value *samplerHandle = lod.get_sampler();
-    if (GetSamplerKind(samplerHandle, ValCtx) != DXIL::SamplerKind::Default) {
+    if ((DXIL::CompareVersions(ValCtx.m_DxilMajor, ValCtx.m_DxilMinor, 1, 8) <
+         0) &&
+        GetSamplerKind(samplerHandle, ValCtx) != DXIL::SamplerKind::Default) {
       ValCtx.EmitInstrError(CI, ValidationRule::InstrSamplerModeForLOD);
     }
     Value *handle = lod.get_handle();
--- a/lib/HLSL/DxilValidation.cpp
+++ b/lib/HLSL/DxilValidation.cpp
@@ -1658,7 +1658,7 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
   case DXIL::OpCode::CalculateLOD: {
     DxilInst_CalculateLOD lod(CI);
     Value *samplerHandle = lod.get_sampler();
-    DXIL::SamplerKind = samplerKind = GetSamplerKind(samplerHandle, ValCtx);
+    DXIL::SamplerKind samplerKind = GetSamplerKind(samplerHandle, ValCtx);
     if (samplerKind != DXIL::SamplerKind::Default) {
       // After SM68, Comparison is supported.
       if (!ValCtx.DxilMod.GetShaderModel()->IsSM68Plus() ||
--- a/lib/HLSL/DxilValidation.cpp
+++ b/lib/HLSL/DxilValidation.cpp
@@ -1658,10 +1658,12 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
   case DXIL::OpCode::CalculateLOD: {
     DxilInst_CalculateLOD lod(CI);
     Value *samplerHandle = lod.get_sampler();
-    if ((DXIL::CompareVersions(ValCtx.m_DxilMajor, ValCtx.m_DxilMinor, 1, 8) <
-         0) &&
-        GetSamplerKind(samplerHandle, ValCtx) != DXIL::SamplerKind::Default) {
-      ValCtx.EmitInstrError(CI, ValidationRule::InstrSamplerModeForLOD);
+    DXIL::SamplerKind = samplerKind = GetSamplerKind(samplerHandle, ValCtx);
+    if (samplerKind != DXIL::SamplerKind::Default) {
+      // After SM68, Comparison is supported.
+      if (!ValCtx.DxilMod.GetShaderModel()->IsSM68Plus() ||
+          samplerKind != DXIL::SamplerKind::Comparison)
+        ValCtx.EmitInstrError(CI, ValidationRule::InstrSamplerModeForLOD);
     }
     Value *handle = lod.get_handle();
     DXIL::ComponentType compTy;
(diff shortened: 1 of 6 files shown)


lib/HLSL/DxilValidation.cpp
```diff
@@ -1658,7 +1658,9 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
   case DXIL::OpCode::CalculateLOD: {
     DxilInst_CalculateLOD lod(CI);
     Value *samplerHandle = lod.get_sampler();
-    if (GetSamplerKind(samplerHandle, ValCtx) != DXIL::SamplerKind::Default) {
+    if ((DXIL::CompareVersions(ValCtx.m_DxilMajor, ValCtx.m_DxilMinor, 1, 8) <
+         0) &&
```
pow2clk, line 1662: Not necessarily part of this change, but seeing how ugly this gets with clang-format makes me think that `ValidationContext` should perhaps have an `isDxilVersionSM6XPlus()` function. 

I do think we should still check that the sampler kind isn't invalid for dxil version 1.8+

utils/hct/gen_intrin_main.txt
```diff
@@ -411,6 +411,8 @@ namespace Texture1DMethods {
 // Use float for DXIL don't support f16 on CalcLOD.
 float [[ro]] CalculateLevelOfDetail(in sampler1d s, in float<1> x) : tex1d_t_calc_lod;
 float [[ro]] CalculateLevelOfDetailUnclamped(in sampler s, in float<1> x) : tex1d_t_calc_lod_unclamped;
+float [[ro]] CalculateLevelOfDetail(in sampler_cmp s, in float<1> x) : tex1d_t_calc_lod;
+float [[ro]] CalculateLevelOfDetailUnclamped(in sampler_cmp s, in float<1> x) : tex1d_t_calc_lod_unclamped;
```
pow2clk, line 415: As a result of #5508, you can use `any_sampler` in the existing overloads without having to add new ones explicitly.

pow2clk (commented):
I'm surprised and a bit troubled that the only thing keeping this from working before was a validation error. I think it would be good to add a Sema error for <1.8. It's not a strict requirement, but it looks like it might slot in pretty easily in `DiagnoseHLSLMethodCall` 

I think we need a way to mark a shader as needing 6.8 when these overloads are used

lib/HLSL/DxilValidation.cpp
python3kgae: Bring back check for 1.8+.

python3kgae:
> I'm surprised and a bit troubled that the only thing keeping this from working before was a validation error. I think it would be good to add a Sema error for <1.8. It's not a strict requirement, but it looks like it might slot in pretty easily in `DiagnoseHLSLMethodCall`
> 
> I think we need a way to mark a shader as needing 6.8 when these overloads are used

Added Sema check in DiagnoseHLSLMethodCall.

lib/HLSL/DxilValidation.cpp
```diff
@@ -1658,10 +1658,12 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
     DxilInst_CalculateLOD lod(CI);
     Value *samplerHandle = lod.get_sampler();
-    if ((DXIL::CompareVersions(ValCtx.m_DxilMajor, ValCtx.m_DxilMinor, 1, 8) <
-         0) &&
-        GetSamplerKind(samplerHandle, ValCtx) != DXIL::SamplerKind::Default) {
-      ValCtx.EmitInstrError(CI, ValidationRule::InstrSamplerModeForLOD);
+    DXIL::SamplerKind samplerKind = GetSamplerKind(samplerHandle, ValCtx);
+    if (samplerKind != DXIL::SamplerKind::Default) {
+      // After SM68, Comparison is supported.
+      if (!ValCtx.DxilMod.GetShaderModel()->IsSM68Plus() ||
```
pow2clk, line 1664: As far as I can tell, this is currently equivalent to checking the dxil versions explicitly since they derive from the dxil modules shader model version. I know we've toyed with the idea of making the dxil version independent of shader model version, but this is hardly the only place in DxilValidation where this is done. 

My offhanded suggestion wasn't to use the shader model version, but to create a set of convenience functions for dxil version similar to these for the shader model version, but this is consistent too.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11287,6 +11287,22 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
         Diags.Report(Loc, diag::err_hlsl_wg_nodetrackrwinputsharing_missing);
         return true;
       }
+    } else if (opCode == hlsl::IntrinsicOp::MOP_CalculateLevelOfDetail ||
+               opCode ==
+                   hlsl::IntrinsicOp::MOP_CalculateLevelOfDetailUnclamped) {
+      const auto *shaderModel =
+          hlsl::ShaderModel::GetByName(getLangOpts().HLSLProfile.c_str());
```
pow2clk, line 11294: It's overkill for this change, but there are so many places in SemHLSL where we retrieve the profile by a string that we might as well encode at least the stage and the version numbers into the LangOpts directly

pow2clk (approved):
Thanks for the responses!

lib/HLSL/DxilValidation.cpp
```diff
@@ -1658,8 +1658,12 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
     DxilInst_CalculateLOD lod(CI);
     Value *samplerHandle = lod.get_sampler();
-    if (GetSamplerKind(samplerHandle, ValCtx) != DXIL::SamplerKind::Default) {
-      ValCtx.EmitInstrError(CI, ValidationRule::InstrSamplerModeForLOD);
+    DXIL::SamplerKind samplerKind = GetSamplerKind(samplerHandle, ValCtx);
+    if (samplerKind != DXIL::SamplerKind::Default) {
+      // After SM68, Comparison is supported.
+      if (!ValCtx.DxilMod.GetShaderModel()->IsSM68Plus() ||
+          samplerKind != DXIL::SamplerKind::Comparison)
+        ValCtx.EmitInstrError(CI, ValidationRule::InstrSamplerModeForLOD);
```
dmpots, line 1666: It looks like from the code path we could still get an error for SM6.8+ for a non-default sampler that is not a comparison sampler.

Is there some other sampler kind in hlsl that we could actually hit this error? If so it would be good to have test coverage for it.

dmpots (approved):

lib/HLSL/DxilValidation.cpp
python3kgae: No real sampler kind left.
The rest sampler kinds are SamplerKind::Invalid and SamplerKind::Mono.
SamplerKind::Mono is only from dxbc convertor.

PR #5803, opened 2023-09-29
Disable UBSAN sanitizing alignment errors

amaiorano:
The remaining errors are non-trivial to fix, so let's disable this for now.

### DIFF
--- a/cmake/modules/HandleLLVMOptions.cmake
+++ b/cmake/modules/HandleLLVMOptions.cmake
@@ -519,7 +519,7 @@ if(LLVM_USE_SANITIZER)
       endif()
     elseif (LLVM_USE_SANITIZER STREQUAL "Undefined")
       append_common_sanitizer_flags()
-      append("-fsanitize=undefined -fno-sanitize=vptr,function -fno-sanitize-recover=all"
+      append("-fsanitize=undefined -fno-sanitize=vptr,function,alignment -fno-sanitize-recover=all"
               CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
     elseif (LLVM_USE_SANITIZER STREQUAL "Thread")
       append_common_sanitizer_flags()


amaiorano:
I created this issue so that we can hopefully revert this at some point: https://github.com/microsoft/DirectXShaderCompiler/issues/5804

llvm-beanz (commented):
Can you also make this change add `Undefined` to the `LLVM_USE_SANITIZER` option in the azure Linux+Clang+Release pipeline?

amaiorano:
> Can you also make this change add `Undefined` to the `LLVM_USE_SANITIZER` option in the azure Linux+Clang+Release pipeline?

There is still one more UBSAN failure. I'll do that with the next fix.

amaiorano:
Friendly ping, can we land this?

llvm-beanz (approved):

amaiorano:
@llvm-beanz asked me offline whether it would be possible to only disable alignment sanitization for specific files, rather than across the entire project. I investigated this, and it doesn't seem so. For posterity, here's what I did:

1. I removed `alignment` from the list of `-fno-sanitize=vptr,function` in `HandleLLVMOptions.cmake` (basically reverted my change)
2. In `lib/HLSL/CMakeLists.txt`, I added `set_source_files_properties(DxilValidation.cpp PROPERTIES COMPILE_FLAGS -fno-sanitize=alignment)` to disable it only for that file.

Without these changes, when I'd run this unit test, I'd get:
```
./tools/clang/unittests/HLSL/ClangHLSLTests --gtest_filter=*DxilModuleTest.SetValidatorVersion

Note: Google Test filter = *DxilModuleTest.SetValidatorVersion
/home/amaiorano/src/external/DirectXShaderCompiler/lib/HLSL/DxilValidation.cpp:5859:33: runtime error: reference binding to misaligned address 0x55edf26d5342 for type 'const uint32_t' (aka 'const unsigned int'), which requires 4 byte alignment
0x55edf26d5342: note: pointer points here
 67 65  74 00 50 53 56 30 50 00  00 00 24 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/amaiorano/src/external/DirectXShaderCompiler/lib/HLSL/DxilValidation.cpp:5859:33 in 
```
But with the above changes to only disable the alignment check for `DxilValidation.cpp`, running this same unit test now fails differently:

```
./tools/clang/unittests/HLSL/ClangHLSLTests --gtest_filter=*DxilModuleTest.SetValidatorVersion

/usr/lib/llvm-17/bin/../include/c++/v1/__hash_table:167:12: runtime error: reference binding to misaligned address 0x56177313f3b2 for type 'const unsigned int', which requires 4 byte alignment
0x56177313f3b2: note: pointer points here
 67 65  74 00 50 53 56 30 50 00  00 00 24 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00
              ^ 
```

If I run it with `UBSAN_OPTIONS=print_stacktrace=1`, we see that actually DxilValidation.cpp is in the callstack:
```
/usr/lib/llvm-17/bin/../include/c++/v1/__hash_table:167:12: runtime error: reference binding to misaligned address 0x56177313f3b2 for type 'const unsigned int', which requires 4 byte alignment
0x56177313f3b2: note: pointer points here
 67 65  74 00 50 53 56 30 50 00  00 00 24 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00
              ^ 
    #0 0x7f0e977ae88f in std::__1::__hash_key_value_types<unsigned int>::__get_key[abi:v170000](unsigned int const&) /usr/lib/llvm-17/bin/../include/c++/v1/__hash_table:167:5
    #1 0x7f0e97d4c32c in std::__1::__hash_table<unsigned int, std::__1::hash<unsigned int>, std::__1::equal_to<unsigned int>, std::__1::allocator<unsigned int>>::__insert_unique[abi:v170000](unsigned int const&) /usr/lib/llvm-17/bin/../include/c++/v1/__hash_table:974:42
    #2 0x7f0e97cf4cfe in std::__1::unordered_set<unsigned int, std::__1::hash<unsigned int>, std::__1::equal_to<unsigned int>, std::__1::allocator<unsigned int>>::insert[abi:v170000](unsigned int const&) /usr/lib/llvm-17/bin/../include/c++/v1/unordered_set:781:26
    #3 0x7f0e97cc258c in hlsl::ValidateDxilContainerParts(llvm::Module*, llvm::Module*, hlsl::DxilContainerHeader const*, unsigned int) /home/amaiorano/src/external/DirectXShaderCompiler/lib/HLSL/DxilValidation.cpp:5864:17
...
```

So the failure is being detected within `unordered_set`, which is called from `DxilValidation.cpp`.

I suspect what's going on is that the linker is using the compiled version of unordered_set from a different translation unit, where it would have been compiled with alignment sanitization enabled. So although we no longer trigger alignment failures for code directly in DxilValidation.cpp, indirect code may still trigger it.

Another option I tried is runtime suppressions: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#runtime-suppressions. I got the same behaviour when using a suppression file: I'm able to suppress the entire file, DxilValidation.cpp, or even the specific function, but then I still get alignment failures in `unordered_set`. (Note that for runtime suppressions to work, I needed to add `-fsanitize-recover=alignment` to the file's compilation flags).

s-perron (approved):

PR #5805, opened 2023-09-29
Fix UBSAN out-of-range of representable values errors

amaiorano:
Implicit casts of double to integer types result in UB if the value cannot be represented as the target integer type. Use the APSInt facilities to round the double, and truncate as necessary.


tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -4939,23 +4939,28 @@ Value *CGMSHLSLRuntime::EmitHLSLLiteralCast(CodeGenFunction &CGF, Value *Src,
     }
   } else if (ConstantFP *CF = dyn_cast<ConstantFP>(Src)) {
     APFloat v = CF->getValueAPF();
-    double dv = v.convertToDouble();
     if (llvm::IntegerType *IT = dyn_cast<llvm::IntegerType>(DstTy)) {
+      APSInt iv;
+      bool isExact;
+      v.convertToInteger(iv, APFloat::roundingMode::rmTowardZero, &isExact);
```
amaiorano, line 4945: Not sure if `rmTowardZero` is correct. Is this defined for HLSL?

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -4939,23 +4939,28 @@ Value *CGMSHLSLRuntime::EmitHLSLLiteralCast(CodeGenFunction &CGF, Value *Src,
...
+      if (IT->getBitWidth() < iv.getBitWidth()) {
+        iv = iv.trunc(IT->getBitWidth());
+      }
       switch (IT->getBitWidth()) {
       case 32:
-        return Builder.getInt32(dv);
+        return Builder.getInt32(iv.getLimitedValue());
```
amaiorano, line 4951: See lines 4914... above for similar code used when casting to integral type.

amaiorano:
Oh, I should mention the actual failure that was getting caught by UBSAN:

```
FAIL: Clang-Unit :: HLSL/ClangHLSLTests/CompilerTest.BatchHLSL (3379 of 3381)
******************** TEST 'Clang-Unit :: HLSL/ClangHLSLTests/CompilerTest.BatchHLSL' FAILED ********************
Note: Google Test filter = CompilerTest.BatchHLSL
...
/home/amaiorano/src/external/DirectXShaderCompiler/tools/clang/lib/CodeGen/CGHLSLMS.cpp:4946:33: runtime error: 3.4e+20 is outside the range of representable values of type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/amaiorano/src/external/DirectXShaderCompiler/tools/clang/lib/CodeGen/CGHLSLMS.cpp:4946:33 in 
```

The source it points at is:
```
      case 32:
        return Builder.getInt32(dv);
                                              ^^
```

Basically, one of the tests is attempting to pass in a double value of 3.4e+20 into a function that accepts a uint32_t, so it's relying on implicit casting, and since the value is out of range, the behaviour is undefined.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
llvm-beanz: Yea, that seems like it should be right for me.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
llvm-beanz: We should specify the actual limits for each of the `getLimitedValue` calls. The default limit is effectively `std::numeric_limits<uint64_t>::max()`. With truncation, this could result in non-saturated values for these casts.

We also should probably add a release note about this since this will absolutely be a behavior change in the compiler.

llvm-beanz (commented):
@bogner, @tex3d, @pow2clk can you guys take a look at this. This fixes undefined behavior in DXC, but it does have some potential behavior change when late-converting `literal float` to integer types smaller than 64-bits.

It looks like we have some test coverage gaps in this code, which makes me concerned.

https://microsoft.github.io/DirectXShaderCompiler/coverage/home/runner/work/DirectXShaderCompiler/DirectXShaderCompiler/tools/clang/lib/CodeGen/CGHLSLMS.cpp.html#L4916

We'll definitely need new tests for this change since it does change behavior for the double->integer conversions.

bogner:
> @bogner, @tex3d, @pow2clk can you guys take a look at this. This fixes undefined behavior in DXC, but it does have some potential behavior change when late-converting `literal float` to integer types smaller than 64-bits.

I think it'd be fair to argue that any HLSL whose behaviour is changed here already had UB and it's not too concerning. If folks have shaders where there's a floating point literal outside of the integer range and then rely on the integer value it could flip signs or something, but I just can't see that being all that common an occurrence.

Regarding the tests, I definitely agree the lack of coverage is concerning, though I'm more worried about the well-defined cases.

amaiorano:
In light of the lack of testing, I spent some time today validating my change by writing a simple shader that exercises the double to integral  conversions. In doing so, I discovered an egregious bug in my original patch (see [fix diff here](https://github.com/microsoft/DirectXShaderCompiler/compare/7b681d9e8b392d47affb769a5e68fe16824ef3c9..162a162aa1a9d4ba155a07358ed107e9871a814e)), which seems to confirm that there's very little testing of these conversions.

Anyway, here's the shader:

```
RWByteAddressBuffer buff : register(u0, space2);
uint to_uint(uint a) { return a; }
int to_int(int a) { return a; }

[numthreads(1, 1, 1)]
void main() {
    // The "Before:" and "After:" values correspond to what value gets stored to the buffer before and after my
    // change respectively. These are copied from the DXIL assembly output, which are i32 values, so the uint
    // result is the signed i32 value.

    // to_int
    buff.Store(0, to_int(-2147483648.0)); // Smallest int, Before: -2147483648, After: -2147483648
    buff.Store(0, to_int(-2147483648.0 - 1.0)); // Before: 2147483647 (wrap), After: -2147483648 (clamp)
    buff.Store(0, to_int(-2147483648.0 - 2.0)); // Before: 2147483646 (wrap), After: -2147483648 (clamp)

    buff.Store(0, to_int(2147483647.0)); // Largest int, Before: 2147483647, After: 2147483647
    buff.Store(0, to_int(2147483647.0 + 1.0)); // Before: -2147483648 (wrap), After: 2147483647 (clamp)
    buff.Store(0, to_int(2147483647.0 + 2.0)); // Before: -2147483647 (wrap), After: 2147483647 (clamp)

    buff.Store(0, to_int(-1.7976931348623158e+308)); // Smallest double, Before: 0, After: -2147483648 (clamp)
    buff.Store(0, to_int(1.7976931348623158e+308)); // Largest double, Before: 0, After: 2147483647 (clamp)

    // to_uint
    buff.Store(1, to_uint(0.0)); // Smallest uint, Before: 0, After: 0
    buff.Store(1, to_uint(0.0 - 1.0)); // Before: -1, After: -1
    buff.Store(1, to_uint(0.0 - 2.0)); // Before: -2, After: -2

    buff.Store(1, to_uint(4294967295.0)); // Largest uint, Before: -1, After: 2147483647
    buff.Store(1, to_uint(4294967295.0 + 1.0)); // Before: 0 (wrap), After: 2147483647 (clamp)
    buff.Store(1, to_uint(4294967295.0 + 2.0)); // Before: 1 (wrap), After: 2147483647 (clamp)

    buff.Store(1, to_uint(-1.7976931348623158e+308)); // Smallest double, Before: 0 (?), After: -2147483648 (clamp)
    buff.Store(1, to_uint(1.7976931348623158e+308)); // Largest double, Before: 0 (?), After: 2147483647 (clamp)
}
```

I added the comments you see on each line with the before/after values by running DXC against this shader before and after my change, respectively with `./bin/dxc -T cs_6_1 test.hlsl | grep bufferStore`.

The results:

```
 ./bin/dxc -T cs_6_1 test.hlsl | grep bufferStore

Results before my change:

  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 -2147483648, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 2147483646, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 -2147483648, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 -2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 0, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 0, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 0, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 -1, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 -2, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 -1, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 0, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 1, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 0, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 0, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)

After my change:

  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 -2147483648, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 -2147483648, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 -2147483648, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 -2147483648, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 0, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 0, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 -1, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 -2, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 -2147483648, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
  call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %1, i32 1, i32 undef, i32 2147483647, i32 undef, i32 undef, i32 undef, i8 1)  ; BufferStore(uav,coord0,coord1,value0,value1,value2,value3,mask)
```

The main difference is that we now clamp (or saturate) the double value to the target integral value, whereas before it would usually wrap the value (well, at least it did so on my machine, but since it's UB, it could be doing something else on other machines).

What do you think? Does this behavior make sense?

Obviously these should be converted into proper tests, and should include literal conversions to i8, i16, and i64, if supported (as per https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#primitive-types).

amaiorano:
Looking at the output from DXC, something I had not realized is that DXC emits `-Wliteral-conversion` warnings for the out-of-range conversions:

```
test.hlsl:8:27: warning: implicit conversion from 'literal float' to 'int' changes value from 2147483648 to 2147483647 [-Wliteral-conversion]
    buff.Store(0, to_int(-2147483648.0)); // Smallest int, Before: -2147483648, After: -2147483648
                  ~~~~~~  ^~~~~~~~~~~~
test.hlsl:16:27: warning: implicit conversion from 'literal float' to 'int' changes value from 1.797693134862316E+308 to 2147483647 [-Wliteral-conversion]
    buff.Store(0, to_int(-1.7976931348623158e+308)); // Smallest double, Before: 0, After: -2147483648 (clamp)
                  ~~~~~~  ^~~~~~~~~~~~~~~~~~~~~~~
test.hlsl:17:26: warning: implicit conversion from 'literal float' to 'int' changes value from 1.797693134862316E+308 to 2147483647 [-Wliteral-conversion]
    buff.Store(0, to_int(1.7976931348623158e+308)); // Largest double, Before: 0, After: 2147483647 (clamp)
                  ~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~~
test.hlsl:28:28: warning: implicit conversion from 'literal float' to 'uint' changes value from 1.797693134862316E+308 to 4294967295 [-Wliteral-conversion]
    buff.Store(1, to_uint(-1.7976931348623158e+308)); // Smallest double, Before: 0 (?), After: -2147483648 (clamp)
                  ~~~~~~~  ^~~~~~~~~~~~~~~~~~~~~~~
test.hlsl:29:27: warning: implicit conversion from 'literal float' to 'uint' changes value from 1.797693134862316E+308 to 4294967295 [-Wliteral-conversion]
    buff.Store(1, to_uint(1.7976931348623158e+308)); // Largest double, Before: 0 (?), After: 2147483647 (clamp)
                  ~~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~~
```

Notice that in some cases, the value that the warning says we'll get is not what we get -- both before and after my change. For instance, the first one:
```
test.hlsl:8:27: warning: implicit conversion from 'literal float' to 'int' changes value from 2147483648 to 2147483647 [-Wliteral-conversion]
    buff.Store(0, to_int(-2147483648.0)); // Smallest int, Before: -2147483648, After: -2147483648
                  ~~~~~~  ^~~~~~~~~~~~
```

The input double value is `-2147483648.0`, the warning says the value `2147483648` will become `2147483647`. This is, presumably, because in C++, the unary negation is applied to the result of the converted float-to-int literal, right? If so, then should DXC do the same here?

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tex3d: Yes.  For instance, we want to saturate cast to `uint32_t` to between `0` and `std::numeric_limits<uint32_t>::max()` instead of saturating it to signed `int32_t` range.

This will also make it match fxc's behavior for casting floats/doubles.

tex3d:
Here's a modified version of your shader that, along with fixing storage offsets, includes comments for FXC behavior.  The FXC behavior looks like the correct behavior to follow for each of these cases.  The difference is that for `uint` it clamps to `uint` range, instead of `int` range.

```hlsl
RWByteAddressBuffer buff;
uint to_uint(uint a) { return a; }
int to_int(int a) { return a; }

static uint o = 0;

void store(uint u) {
    buff.Store(4 * (o++), u);
}

[numthreads(1, 1, 1)]
void main() {
    // The "Before:" and "After:" values correspond to what value gets stored to the buffer before and after my
    // change respectively. These are copied from the DXIL assembly output, which are i32 values, so the uint
    // result is the signed i32 value.

    // fxc results with -T cs_5_0 -Lx
    // store_raw u0.xyzw, l(0x00000000), l(0x80000000,0x80000000,0x80000000,0x7fffffff)
    // store_raw u0.xyzw, l(0x00000010), l(0x7fffffff,0x7fffffff,0x80000000,0x7fffffff)
    // store_raw u0.xyzw, l(0x00000020), l(0x00000000,0x00000000,0x00000000,0xffffffff)
    // store_raw u0.xyzw, l(0x00000030), l(0xffffffff,0xffffffff,0x00000000,0xffffffff)

    // to_int
    store(to_int(-2147483648.0)); // MaxNegative int, Before: -2147483648, After: -2147483648, FXC: -2147483648
    store(to_int(-2147483648.0 - 1.0)); // Before: 2147483647 (wrap), After: -2147483648 (clamp), FXC: -2147483648 (clamp int)
    store(to_int(-2147483648.0 - 2.0)); // Before: 2147483646 (wrap), After: -2147483648 (clamp), FXC: -2147483648 (clamp int)

    store(to_int(2147483647.0)); // MaxPositive int, Before: 2147483647, After: 2147483647, FXC: 2147483647
    store(to_int(2147483647.0 + 1.0)); // Before: -2147483648 (wrap), After: 2147483647 (clamp), FXC: 2147483647 (clamp int)
    store(to_int(2147483647.0 + 2.0)); // Before: -2147483647 (wrap), After: 2147483647 (clamp), FXC: 2147483647 (clamp int)

    store(to_int(-1.7976931348623158e+308)); // MaxNegative double, Before: 0, After: -2147483648 (clamp), FXC: -2147483648 (clamp int)
    store(to_int(1.7976931348623158e+308)); // MaxPositive double, Before: 0, After: 2147483647 (clamp), FXC: 2147483647 (clamp int)

    // to_uint
    store(to_uint(0.0)); // MaxNegative uint, Before: 0, After: 0, FXC: 0
    store(to_uint(0.0 - 1.0)); // Before: -1, After: -1, FXC: 0 (clamp uint)
    store(to_uint(0.0 - 2.0)); // Before: -2, After: -2, FXC: 0 (clamp uint)

    store(to_uint(4294967295.0)); // MaxPositive uint, Before: -1, After: 2147483647, FXC: 4294967295 (clamp uint)
    store(to_uint(4294967295.0 + 1.0)); // Before: 0 (wrap), After: 2147483647 (clamp), FXC: 4294967295 (clamp uint)
    store(to_uint(4294967295.0 + 2.0)); // Before: 1 (wrap), After: 2147483647 (clamp), FXC: 4294967295 (clamp uint)

    store(to_uint(-1.7976931348623158e+308)); // MaxNegative double, Before: 0 (?), After: -2147483648 (clamp), FXC: 0 (clamp uint)
    store(to_uint(1.7976931348623158e+308)); // MaxPositive double, Before: 0 (?), After: 2147483647 (clamp), FXC: 4294967295 (clamp uint)
}
```

amaiorano:
Thanks, @tex3d, for adding the FXC output. I've updated the code so that it now uses `APSInt::getExtValue()`, which returns the correctly-extended value, taking into account the bitwidth and whether the APSInt is signed/unsigned. This now returns the exact same values as FXC. In fact, I based these changes off the code in `DiagnoseFloatingLiteralImpCast()` in SemaChecking.cpp, which is where the out-of-range warnings for the float to int literal come from.

Here's the shader with the updated "After" output:

```
RWByteAddressBuffer buff;
uint to_uint(uint a) { return a; }
int to_int(int a) { return a; }

static uint o = 0;

void store(uint u) {
    buff.Store(4 * (o++), u);
}

[numthreads(1, 1, 1)]
void main() {
    // The "Before:" and "After:" values correspond to what value gets stored to the buffer before and after my
    // change respectively. These are copied from the DXIL assembly output, which are i32 values, so the uint
    // result is the signed i32 value.

    // fxc results with -T cs_5_0 -Lx
    // store_raw u0.xyzw, l(0x00000000), l(0x80000000,0x80000000,0x80000000,0x7fffffff)
    // store_raw u0.xyzw, l(0x00000010), l(0x7fffffff,0x7fffffff,0x80000000,0x7fffffff)
    // store_raw u0.xyzw, l(0x00000020), l(0x00000000,0x00000000,0x00000000,0xffffffff)
    // store_raw u0.xyzw, l(0x00000030), l(0xffffffff,0xffffffff,0x00000000,0xffffffff)

    // to_int
    store(to_int(-2147483648.0)); // MaxNegative int, Before: -2147483648, After: -2147483648, FXC: -2147483648
    store(to_int(-2147483648.0 - 1.0)); // Before: 2147483647 (wrap), After: -2147483648 (clamp), FXC: -2147483648 (clamp int)
    store(to_int(-2147483648.0 - 2.0)); // Before: 2147483646 (wrap), After: -2147483648 (clamp), FXC: -2147483648 (clamp int)

    store(to_int(2147483647.0)); // MaxPositive int, Before: 2147483647, After: 2147483647, FXC: 2147483647
    store(to_int(2147483647.0 + 1.0)); // Before: -2147483648 (wrap), After: 2147483647 (clamp), FXC: 2147483647 (clamp int)
    store(to_int(2147483647.0 + 2.0)); // Before: -2147483647 (wrap), After: 2147483647 (clamp), FXC: 2147483647 (clamp int)

    store(to_int(-1.7976931348623158e+308)); // MaxNegative double, Before: 0, After: -2147483648 (clamp), FXC: -2147483648 (clamp int)
    store(to_int(1.7976931348623158e+308)); // MaxPositive double, Before: 0, After: 2147483647 (clamp), FXC: 2147483647 (clamp int)

    // to_uint
    store(to_uint(0.0)); // MaxNegative uint, Before: 0, After: 0, FXC: 0
    store(to_uint(0.0 - 1.0)); // Before: -1, After: 0, FXC: 0 (clamp uint)
    store(to_uint(0.0 - 2.0)); // Before: -2, After: 0, FXC: 0 (clamp uint)

    store(to_uint(4294967295.0)); // MaxPositive uint, Before: -1, After: -1 (4294967295), FXC: 4294967295 (clamp uint)
    store(to_uint(4294967295.0 + 1.0)); // Before: 0 (wrap), After: -1 (4294967295) (clamp), FXC: 4294967295 (clamp uint)
    store(to_uint(4294967295.0 + 2.0)); // Before: 1 (wrap), After: -1 (4294967295) (clamp), FXC: 4294967295 (clamp uint)

    store(to_uint(-1.7976931348623158e+308)); // MaxNegative double, Before: 0 (?), After: 0 (clamp), FXC: 0 (clamp uint)
    store(to_uint(1.7976931348623158e+308)); // MaxPositive double, Before: 0 (?), After:  -1 (4294967295) (clamp), FXC: 4294967295 (clamp uint)
}
```

amaiorano:
Note that I think my latest patch is correct, but I looked again at the `-Wliteral-conversion` warnings I mentioned earlier, and noticed that what Clang outputs was fixed in version 3.9.

This is what DXC currently outputs (using code from 3.7, I think?):
```
test.hlsl:24:19: warning: implicit conversion from 'literal float' to 'int' changes value from 2147483648 to 2147483647 [-Wliteral-conversion]
    store(to_int(-2147483648.0)); // MaxNegative int, Before: -2147483648, After: -2147483648, FXC: -2147483648
          ~~~~~~  ^~~~~~~~~~~~
test.hlsl:32:19: warning: implicit conversion from 'literal float' to 'int' changes value from 1.797693134862316E+308 to 2147483647 [-Wliteral-conversion]
    store(to_int(-1.7976931348623158e+308)); // MaxNegative double, Before: 0, After: -2147483648 (clamp), FXC: -2147483648 (clamp int)
          ~~~~~~  ^~~~~~~~~~~~~~~~~~~~~~~
test.hlsl:33:18: warning: implicit conversion from 'literal float' to 'int' changes value from 1.797693134862316E+308 to 2147483647 [-Wliteral-conversion]
    store(to_int(1.7976931348623158e+308)); // MaxPositive double, Before: 0, After: 2147483647 (clamp), FXC: 2147483647 (clamp int)
          ~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~~
test.hlsl:44:20: warning: implicit conversion from 'literal float' to 'uint' changes value from 1.797693134862316E+308 to 4294967295 [-Wliteral-conversion]
    store(to_uint(-1.7976931348623158e+308)); // MaxNegative double, Before: 0 (?), After: 0 (clamp), FXC: 0 (clamp uint)
          ~~~~~~~  ^~~~~~~~~~~~~~~~~~~~~~~
test.hlsl:45:19: warning: implicit conversion from 'literal float' to 'uint' changes value from 1.797693134862316E+308 to 4294967295 [-Wliteral-conversion]
    store(to_uint(1.7976931348623158e+308)); // MaxPositive double, Before: 0 (?), After:  -1 (4294967295) (clamp), FXC: 4294967295 (clamp uint)
          ~~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~~
```

[Here's a Compiler Explorer example](https://godbolt.org/z/E5nTnx7Kq) showing what Clang 3.8 vs 3.9 does to similar code. Pay attention to the warnings emitted by the compiler, rather than the actual output (the output is mostly garbage, which isn't unexpected since these conversions are UB). We can see that the 3.8 version is what DXC is emitting.

So as a separate patch, we should update this warning code with a later version. In fact, in latest Clang, it doesn't bother saying what the literal value will be changed to since it's UB. It says, instead: 
```
<source>:28:12: warning: implicit conversion of out of range value from 'double' to 'int' is undefined [-Wliteral-conversion]
```

But for DXC, if we want to make these well-defined, as this patch is doing (and as FXC is doing), then we could go with the 3.9 changes instead.

amaiorano:
@tex3d friendly ping, can you please take a look?

tex3d (commented):
Thanks for this contribution!  It looks like we still need to add the test to the PR.  Since you have the test HLSL above, it shouldn't be hard to add this as a FileCheck test under perhaps: `tools\clang\test\HLSLFileCheck\hlsl\literal` (`literal` being a new folder).

amaiorano:
> Thanks for this contribution! It looks like we still need to add the test to the PR. Since you have the test HLSL above, it shouldn't be hard to add this as a FileCheck test under perhaps: `tools\clang\test\HLSLFileCheck\hlsl\literal` (`literal` being a new folder).

@tex3d Okay, I've added a test. Let me know if this looks good.

amaiorano:
> LGTM!

Thanks! I moved the file to `tools/clang/test/CodeGenDXIL/literal` as per @llvm-beanz in an offline conversation.

llvm-beanz (approved):

PR #5806, opened 2023-09-29
Move NodeMaxDispatchGrid diagnostic into Sema

### LINKED ISSUE #5775: Move NodeMaxDispatchGrid diagnostic into Sema
In CGHLSLMS.cpp, there is a diagnostic that is emitted when NodeMaxDispatchGrid is used on a structure that has no member with the SV_DispatchGrid semantic. The location of this diagnostic isn't ideal, it should be located in SemaHLSL.cpp. 

The problem with the current location is that the diagnostic won't be emitted when running dxc with -verify. The compiler should be able to diagnose this issue earlier than the code gen pass.

The diagnostic should be moved to SemaHLSL.cpp. The logic that's needed to store the input nodes, which is currently available in CodeGenHLSL.cpp via the funcProps structure, will somehow need to be transferred over to SemaHLSL.cpp.

hekota:
The diagnostic that is emitted when `NodeMaxDispatchGrid` attribute is used on a structure that has no members with the `SV_DispatchGrid` semantic is implemented in `CGHLSLMS.cpp`. This location isn't ideal. This PR is moving that diagnostic to an earlier compile stage into `SemaHLSL.cpp`. That way the check can happen earlier and will be emitted with `dxc -verify`.

New test cases for the `SV_DispatchGrid` diagnostic were added to `VerifierTests` and corresponding tests in codegen removed.

Since the diagnostics was implemented in codegen, many other verifier tests did not hit it and were using `NodeMaxDispatchGrid` without a valid node input record. These tests have been adjusted on a case-by-case basis either by adding a field with `SV_DispatchGrid`, changing the attribute to `NodeDispatchGrid` or adding the error to expected output.

Fixes #5775


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15169,6 +15169,60 @@ static bool nodeInputIsCompatible(DXIL::NodeIOKind IOType,
...
+                                              SourceLocation &DispatchGridLoc,
+                                              bool &Found) {
+  // Iterate over fields of the input record struct
+  for (auto FieldDecl : InputRecordDecl->fields()) {
+    // Check if any of the fields have SV_DispatchGrid annotation
+    for (const hlsl::UnusualAnnotation *it :
+         FieldDecl->getUnusualAnnotations()) {
+      if (it->getKind() == hlsl::UnusualAnnotation::UA_SemanticDecl) {
+        const hlsl::SemanticDecl *sd = cast<hlsl::SemanticDecl>(it);
+        if (sd->SemanticName.equals("SV_DispatchGrid")) {
```
bob80905, line 15186: There's been an effort to move away from explicit string compares. I'm not aware of there being any location where this semantic string exists and is stored in a variable, but maybe an issue can be filed so that in the future this could instead look something like 
`.equals( <variable that will always hold "SV_DispatchGrid"> )`

hekota: I am aware of the effort to move away from explicit string compares, but I am not sure if it makes sense for the HLSL semantics. It is not a fixed set of strings, and the comparisons are much less common than in the other cases for resources, node record types and such. In fact, there are only 3 instances of `semanticName.equals` in the code.

@llvm-beanz - what is your take on this?

tools/clang/test/HLSL/workgraph/member_write_diagnostics.hlsl
```diff
@@ -110,7 +111,7 @@ void node07(NodeOutput<RECORD> output7)
 [Shader("node")]
 [NumThreads(1024,1,1)]
 [NodeLaunch("broadcasting")]
-[NodeMaxDispatchGrid(1024,1,1)]
+[NodeDispatchGrid(1024,1,1)]
```
pow2clk, line 114: Were the changes to add `SV_DispatchGrid` to the RECORD struct and also the switching of `NodeMaxDispatchGrid` to `NodeDispatchGrid` both necessary? It seems like one or the other would have silenced the error as I understand it. 

Same for the next two test files

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: I think the way we normally look at semantics is using `Semantic::DecomposeNameAndIndex` and `Semantic::GetByName`. It would be nice if the `HLSLSemanticAttr` or `SemanticDecl` parsed the string into the enum and stored the enum value.

tools/clang/test/HLSL/workgraph/member_write_diagnostics.hlsl
hekota: Yes, they were. Without changing `NodeMaxDispatchGrid` to `NodeDispatchGrid` the node would have to have an extra argument of type `DispatchNodeInputRecord<RECORD>` added to make use of the `SV_DispatchGrid` in `RECORD`. Changing `NodeMaxDispatchGrid` to `NodeDispatchGrid` seems like an easier change and not affecting the test.

pow2clk: got it! thanks!

pow2clk (approved):
Looks good. Thanks for the explanation of the test changes

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15169,6 +15169,60 @@ static bool nodeInputIsCompatible(DXIL::NodeIOKind IOType,
...
+    }
+  }
+
+  // Walk up the inheritance chain and check all fields on base classes
+  for (CXXRecordDecl::base_class_iterator B = InputRecordDecl->bases_begin(),
+                                          BEnd = InputRecordDecl->bases_end();
+       B != BEnd; ++B) {
+
+    QualType BaseType = B->getType();
+    CXXRecordDecl *BaseTypeDecl =
```
bob80905, line 15210: I'm worried about if this dyn_cast assigns a nullptr to BaseTypeDecl, or if any caller that calls this  recursive function passes a nullptr. This is because the for loop at the top of the function would immediately dereference the pointer. Would it be possible to just do a nullptr check at the top and return early if BaseTypeDecl is null?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15288,6 +15342,42 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
...
+
+              // Get the input record struct
+              auto &TemplateArgs = templateDecl->getTemplateArgs();
+              auto &StructArg = TemplateArgs.get(0);
+              QualType StructArgType = StructArg.getAsType();
+              CXXRecordDecl *NodeInputStructDecl = dyn_cast<CXXRecordDecl>(
+                  StructArgType->getAsStructureType()->getDecl());
+
+              // Make sure there is exactly one SV_DispatchGrid semantics
+              DiagnoseMustHaveOneDispatchGridSemantics(S, NodeInputStructDecl,
```
bob80905, line 15369: How about putting the previous statement in an if condition, and the recursive call inside the if body?

bob80905 (approved):
Thanks!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15138,6 +15138,63 @@ static bool nodeInputIsCompatible(DXIL::NodeIOKind IOType,
...
+// Diagnose input node record to make sure it has exactly one SV_DispatchGrid
+// semantics. Recursivelly walk all fields on the record and all of its base
+// classes/structs
+void DiagnoseMustHaveOneDispatchGridSemantics(Sema &S,
+                                              CXXRecordDecl *InputRecordDecl,
+                                              SourceLocation &DispatchGridLoc,
+                                              bool &Found) {
+  // Iterate over fields of the input record struct
+  for (auto FieldDecl : InputRecordDecl->fields()) {
+    // Check if any of the fields have SV_DispatchGrid annotation
```
tex3d, line 15150: This doesn't handle nested or derived scenarios.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15138,6 +15138,63 @@ static bool nodeInputIsCompatible(DXIL::NodeIOKind IOType,
...
+    for (const hlsl::UnusualAnnotation *it :
+         FieldDecl->getUnusualAnnotations()) {
+      if (it->getKind() == hlsl::UnusualAnnotation::UA_SemanticDecl) {
+        const hlsl::SemanticDecl *sd = cast<hlsl::SemanticDecl>(it);
+        if (sd->SemanticName.equals("SV_DispatchGrid")) {
+          if (!Found) {
+            Found = true;
+            DispatchGridLoc = it->Loc;
```
tex3d, line 15158: Shouldn't we verify that the field is a valid type here: 32 or 16-bit uint scalar or vector with up to 3 components.

tools/clang/test/HLSL/workgraph/dispatchgrid_diags.hlsl
```diff
@@ -55,41 +62,92 @@ void node06()
...
+    uint3 grid : SV_DispatchGrid;   // expected-error {{a field with SV_DispatchGrid has already been specified}}
+};
+
+struct B : A {
+    float b;
+};
+
+struct C : B {
+    float b;
+    int3 grid2 : SV_DispatchGrid;   // expected-note {{other SV_DispatchGrid defined here}}
```
tex3d, line 145: Oh, I see that derived case does in fact work due to the way `InputRecordDecl->fields()` must iterate base class fields as well.  But I doubt the nested case does.  And what about templates?  Also, the locations of the error and note seems reversed from what I would expect.

Here's a fun one to try:
```hlsl
template<typename T, int C> struct DispatchGridField {
  vector<T, C> grid : SV_DispatchGrid;  // expected-note {{other SV_DispatchGrid defined here}}
};
template<typename T, int C> struct UseDispatchGrid {
  DispatchGridField<T, C> dg;
};
template <typename T, int C> struct Record : UseDispatchGrid<T, C> {
  uint grid2 : SV_DispatchGrid;  // expected-error {{a field with SV_DispatchGrid has already been specified}}
};

[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeMaxDispatchGrid(32, 16, 1)]
[NumThreads(32, 1, 1)]
void node1(DispatchNodeInputRecord< Record<uint, 2> > input) {}
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15138,6 +15138,63 @@ static bool nodeInputIsCompatible(DXIL::NodeIOKind IOType,
...
+          } else {
+            // There should be just one SV_DispatchGrid in per record struct
+            S.Diags.Report(
+                it->Loc,
+                diag::err_hlsl_dispatchgrid_semantic_already_specified);
+            S.Diags.Report(DispatchGridLoc, diag::note_defined_here)
```
tex3d, line 15164: Strangely, `fields()` must iterate over fields in base classes in the reverse order than what I would expect, leading to these diagnostics being reversed from expectation.  It places the duplicate diag on the base field and the note on the derived field.  That's weird, because you would expect it to be the other way around, which would be more helpful.  There's probably another way to iterate bases and fields such that they go from bases first, but it would require more complicated iteration code.

tex3d: Oh, nevermind, I hadn't read the rest of the code that handles bases after the fields here.  All you have to do is swap the ordering to iterate bases first.

tools/clang/lib/Sema/SemaHLSL.cpp
hekota: Addressed in PR #5844

tools/clang/test/HLSL/workgraph/dispatchgrid_diags.hlsl
hekota: Templates don't work at all :( 
Issue filed: #5842.

tools/clang/lib/Sema/SemaHLSL.cpp
hekota: Addressed in PR #5844

tools/clang/lib/Sema/SemaHLSL.cpp
hekota: Issue filed: #5845

PR #5815, opened 2023-10-02
Fix compilation issues on clang (#5683)

hekota:
ErrorOr has a implicit constructor. A move is required for construct an ErrorOr object. Then it's returned by the function. Write the construction explicit to fix the compilation on clang, and reduce the confusion.

(cherry picked from commit c22d103059672f4a0e40cc0f1ab6a86ee0637f86)

Plus added explicit ErrorOr template argument to fix a build error

### DIFF
--- a/lib/Support/MemoryBuffer.cpp
+++ b/lib/Support/MemoryBuffer.cpp
@@ -372,8 +372,10 @@ getOpenFileImpl(int FD, const Twine &Filename, uint64_t FileSize,
     std::unique_ptr<MemoryBuffer> Result(
         new (NamedBufferAlloc(Filename))
         MemoryBufferMMapFile(RequiresNullTerminator, FD, MapSize, Offset, EC));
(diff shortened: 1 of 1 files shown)


bob80905 (approved):

hekota:
The build is failing, it looks like we'll need to cherry-pick more clang compilation changes. I'll create a new PR.

hekota:
I am reviving this cherry-pick PR and I've added an explicit `ErrorOr` template argument to make the build work.

llvm-beanz:
Which build configuration was this failing on? I'm using clang on both macOS and Ubuntu and haven't encountered this.

hekota:
It was failing everywhere, even on my machine with MSVC. You can see the list of failed builds when you click on the red X next to my first commit. I can't find the build logs anymore though.

adam-yang (approved):

PR #5816, opened 2023-10-02
Support globallycoherent RWDispatchNodeInputRecord

### LINKED ISSUE #5813: Support globallycoherent modifier for RWDispatchNodeInputRecord
The `globallycoherent` modifier needs to be supported for `RWDispatchNodeInputRecord` in line with a recent Work Graphs spec update.  This is used to allow coherent use of this UAV-like memory across thread groups.

Implementation will require updating the `NodeIOFlags` with a new `GloballyCoherent` flag after `TrackRWInputSharing` and corresponding addition of this flag to the `RecordFlagsMask`.

tex3d:
- Add GloballyCoherent flag to NodeIOFlags
- Fix NodeFlags::IsOutputNode; add IsRecord, IsOutput, IsOutputRecord
- Improve globallycoherent diagnostic
- Update and add tests

Fixes #5813


tools/clang/test/HLSL/globallycoherent-template-errors.hlsl
```diff
@@ -2,18 +2,86 @@
...
+
+  globallycoherent T local2 = data;
+
+  doSomething3(local1);
+}
+
+// TBD: Support node types in templates.
+//  Currently, these produce:
+//  'T' cannot be used as a type parameter where a struct/class is required
+//  Is this #5729?
```
tex3d, line 52: I filed issue #5822 for this.

lib/DXIL/DxilNodeProps.cpp
```diff
@@ -81,6 +90,16 @@ bool NodeFlags::GetTrackRWInputSharing() const {
           (uint32_t)DXIL::NodeIOFlags::TrackRWInputSharing) != 0;
 }
 
+void NodeFlags::SetGloballyCoherent() {
+  m_Flags = (DXIL::NodeIOFlags)((uint32_t)m_Flags |
+                                (uint32_t)DXIL::NodeIOFlags::GloballyCoherent);
+}
+
+bool NodeFlags::GetGlobalCoherent() const {
+  return ((uint32_t)m_Flags & (uint32_t)DXIL::NodeIOFlags::GloballyCoherent) !=
```
llvm-beanz, line 99: nit: I wonder if `m_Flags` should actually be a `NodeIOFlags`, or if `NodeIOFlags` should be an enum class. One of the benefits of enum classes in some cases is that casting must be explicit, but since the flags will basically always need to be casted it is unfortunate.

lib/DXIL/DxilNodeProps.cpp
```diff
@@ -81,6 +90,16 @@ bool NodeFlags::GetTrackRWInputSharing() const {
           (uint32_t)DXIL::NodeIOFlags::TrackRWInputSharing) != 0;
 }
 
+void NodeFlags::SetGloballyCoherent() {
+  m_Flags = (DXIL::NodeIOFlags)((uint32_t)m_Flags |
```
llvm-beanz, line 94: nit: I know it is more verbose, but it is generally preferable to use `static_cast<>` over the C-style cast. The C++ casts have more clearly defined behavior so they are less prone to misuse or unexpected behavior.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2435,6 +2435,9 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
             node.MaxRecords =
                 parmDecl->getAttr<HLSLMaxRecordsAttr>()->getMaxCount();
           }
+          if (parmDecl->hasAttr<HLSLGloballyCoherentAttr>()) {
+            node.Flags.SetGloballyCoherent();
+          }
```
llvm-beanz, line 2440: nit: LLVM style prefers omitting braces for single-line bodies (https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements).

llvm-beanz (approved):
A few nits. Feel free to take or leave them as you wish.

lib/DXIL/DxilNodeProps.cpp
tex3d: If I get what you're saying, you're wondering if `m_Flags` should be `uint32_t` instead, since it is `NodeIOFlags` now, (or a regular enum, instead of enum class)?

For the ordinary enum route, we'd need to either extend the value names due to the loss of the enum class namespace or maybe place `NodeIOFlags` under the NodeFlags class, which would also need to be moved to `DxilConstants.h`.  That might not be a bad option.  We would add a `using` statement under the DXIL namespace to import `NodeIOFlags` there, without the enum values polluting the DXIL namespace.

If NodeFlags is moved to `DxilConstants.h`, we'd probably want to inline some/most of the methods as well.  `DxilConstants.h` is intended to define the necessary constants for DXIL production/consumption without having to build/link other DXIL cpp files from DXC, or even depend on LLVM, though it's ok to have some optionally-used functionality declared here and defined elsewhere.

Alternatively, or additionally, we could define global operators for `&` and `|` between types depending whether `NodeIOFlags` is changed and moved under NodeFlags.

In any case, all this seems like something that belongs outside this PR.

lib/DXIL/DxilNodeProps.cpp
tex3d: I was being consistent with the existing code here.  The use of `static_cast<>` for this kind of case is so ugly.  The noise drowns out the signal, making the code harder to read.  I think I'd rather implement the changes to the enum and/or global operators discussed in the other comment to avoid the ugly casting in the first place.

llvm-beanz: That's fair, but "ugly" code that is correct and explicit should always be preferred over brief code that has less clear meaning. Especially since the ugliness of code is extremely subjective, but the explicitness or implicitness isn't. I generally think c-style casts are ugly, which I'm guessing is a point you and I disagree on.

tex3d: Oh, I agree generally, but when it comes to code using `&`/`|` on flag values, using static_cast usually ends up making it harder to read (and notice faults in) the code.  Is there a practical advantage to it in this use case (and of using enum class instead of enum in this NodeFlags class for the flags)?

tex3d: I'll revisit the flags issue in separate PR, since this one is just adding code with exactly the same pattern, and I'd rather not mix the concern of the best way to solve that with this PR.

lib/DXIL/DxilNodeProps.cpp
tex3d: I'll revisit the flags issue in separate PR, since this one is just adding code with exactly the same pattern, and I'd rather not mix the concern of the best way to solve that with this PR.

tools/clang/test/HLSL/globallycoherent-template-errors.hlsl
tex3d: Since template fixes are going in, I might wait until those merge, then update this test.

tex3d: Looks like I need to wait a bit longer, so I'll merge this and circle back once these are supported.

Issue #5810, opened 2023-10-02
[SM 6.8] `[NodeArraySize()]` and `[AllowSparseNodes]` should only apply to some record types


llvm-beanz:
**Description**
The `[NodeArraySize()]` and attribute should error if it is applied to non-array node record types, and the `[AllowSparseNodes]` attributes should error if it is applied to array node record types.

See the comment from @tex3d here:
https://github.com/microsoft/DirectXShaderCompiler/pull/5588#discussion_r1341723232

**Steps to Reproduce**
Apply `[NodeArraySize()]` to a `NodeOutput` and observe no error reported.
Apply `[AllowSparseNodes()]` to a `NodeOutputArray` and observe no error reported.

**Environment**
- DXC version: SM 6.8 staging branch
- Host Operating System: Any

bob80905:
Fixed by #5817

PR #5817, opened 2023-10-03
Support [UnboundedSparseNodes] on node output

### LINKED ISSUE #5811: Support [UnboundedSparseNodes] on [Empty]NodeOutputArray
**Description**
Support `[UnboundedSparseNodes]` on `[Empty]NodeOutputArray` to update compiler to recent Work Graph spec updates.
This attribute is shorthand for specifying `[AllowSparseNodes]` and `[NodeArraySize((uint)-1)]`.

tex3d:
- Add [UnboundedSparseNodes] attribute
- Set AllowSparseNodes and OutputArraySize = UINT_MAX
- Improve node parameter diagnostics
- Add tests
- Update tests using MaxRecords on input with the wrong launch type

Fixes #5811


tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7809,6 +7809,12 @@ def err_hlsl_wg_nodetrackrwinputsharing_missing : Error<
 def err_hlsl_wg_input_kind : Error<
    "'%0' may not be used with %1 nodes (only %select{DispatchNodeInputRecord or RWDispatchNodeInputRecord|"
    "GroupNodeInputRecords, RWGroupNodeInputRecords, or EmptyNodeInput|ThreadNodeInputRecord or RWThreadNodeInputRecord}2)">;
+def err_hlsl_wg_attr_only_on_output : Error<
+   "attribute %0 may only be used with output nodes">;
+def err_hlsl_wg_nodearraysize_conflict_unbounded : Error<
```
python3kgae, line 7814: Will %2 change to different value or it will be only UnboundedSparseNodes?

tex3d: It's always UnboundedSparseNodes (and %0 is always NodeArraySize), but I thought using the spelling from the attribute might be the "right" way?  However, it really shouldn't matter, so I could change this to use the attribute names in the string instead of using parameters here.

pow2clk (commented):
The diagnostic typo should be fixed. Otherwise looks good!

pow2clk (approved):
Thanks for the followups!

python3kgae (approved):

bob80905:
This PR also fixes #5810

Issue #5692, opened 2023-09-12
Test common input parameters (like SV_DispatchThreadID) lowering for Node Shaders


pow2clk:
**Is your feature request related to a problem? Please describe.**
There are no tests that verify that common input parameters introduced before Shader Model 6.8 are correctly lowered 

**Describe the solution you'd like**
We need a test that verifies at least some common parameters that are processed the same way when applied to node shaders. Not only that they are accepted, but that the generated output is correct.

**Additional context**
See this untested for node shader code location: 
https://github.com/microsoft/DirectXShaderCompiler/blob/7f6d946c4bcddf1aca604e16ee6d0c0108e2d2b1/lib/HLSL/HLSignatureLower.cpp#L1744

See internal commit 2920

tex3d:
I updated the link above to a permalink for the untested node shader code location, since it didn't take me to the right place anymore.

Issue #5768, opened 2023-09-21
Declare SV_VertexID as float only get validation error


python3kgae:
**Description**

SV_VertexID should be a uint.
But declare it as float only get validation error.
Not compile error.

**Steps to Reproduce**

compile with -Tvs_6_0 

float4 main(float V : SV_VertextID) : SV_Position {
   return V;
}

https://godbolt.org/z/ahe1fjsEM

**Actual Behavior**

It should be a compile error instead of validation error.

**Environment**
- DXC version 1.7.2207
- Host Operating System Windows

PR #5818, opened 2023-10-03
Flatten non-object parameters for node shaders

python3kgae:
Enabled node shader parameters flatten for use non-object parameters in struct.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5692


tools/clang/test/HLSL/workgraph/work-graphs.hlsl
```diff
@@ -308,3 +308,24 @@ void node4_06(RWThreadNodeInputRecord<RECORD> input)
 {
   input.FinishedCrossGroupSharing(); // expected-error {{no member named 'FinishedCrossGroupSharing' in 'RWThreadNodeInputRecord<RECORD>'}}
 }
+
+struct R
+{
+  uint3 dtid: SV_DispatchThreadID;
```
bfavela, line 314: Is there a document somewhere that describes what system values are supported for node shaders? The preview at https://devblogs.microsoft.com/directx/d3d12-work-graphs-preview wasn't explicit, but I did see SV_DispatchGrid that isn't in this list.

bfavela:
> Added test for things like SV_GroupID.
> 
> The code change is for failures when use SV_GroupID in struct input.
> 
> Fixes #5560

Not sure this fixes #5560 - that issue is about sample operations, and not related to work graphs

bfavela (changes requested):
Please add SV_DispatchGrid to the tested semantics or close this if it's not relevant.

tools/clang/test/HLSL/workgraph/work-graphs.hlsl
python3kgae: This PR is only for common input before shader model 6.8.
So SV_DispatchGrid is not here.

python3kgae:
> > Added test for things like SV_GroupID.
> > The code change is for failures when use SV_GroupID in struct input.
> > Fixes #5560
> 
> Not sure this fixes #5560 - that issue is about sample operations, and not related to work graphs

Should be https://github.com/microsoft/DirectXShaderCompiler/issues/5692 . :(

python3kgae:
> Please add SV_DispatchGrid to the tested semantics or close this if it's not relevant.

This PR is for test input before shader model 6.8. So SV_DispatchGrid is not included.

bfavela (approved):

tools/clang/test/HLSL/workgraph/work-graphs.hlsl
```diff
@@ -308,3 +308,24 @@ void node4_06(RWThreadNodeInputRecord<RECORD> input)
...
+  uint gidx: SV_GroupIndex;
+  uint3 gtid: SV_GroupThreadID;
+};
+
+[Shader("node")]
+[NodeLaunch("coalescing")]
+[numthreads(4,4,4)]
+void node01(RWGroupNodeInputRecords<R> input,
+ R r,
+ uint vid : SV_VertexID ) // TODO: verify error after https://github.com/microsoft/DirectXShaderCompiler/issues/5768 got fixed.
```
pow2clk, line 325: This is the right type. I don't know what #5768 has to do with this.

lib/HLSL/HLModule.cpp
```diff
@@ -371,7 +371,7 @@ bool HLModule::IsEntryThatUsesSignatures(llvm::Function *F) {
   auto propIter = m_DxilFunctionPropsMap.find(F);
   if (propIter != m_DxilFunctionPropsMap.end()) {
     DxilFunctionProps &props = *(propIter->second);
-    return props.IsGraphics() || props.IsCS();
+    return props.IsGraphics() || props.IsCS() || props.IsNode();
```
pow2clk, line 374: This matches the equivalent function in DXModule. However, I'm a little unsure whether this is right in all cases. The way this function is used is often to determine if the shader is a library or not. Was this added for the purpose of this location? https://github.com/microsoft/DirectXShaderCompiler/blob/769b504be4748931254036e3250c1daad5a1991a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp#L4231

It's the only one that looks related, but it actually looks like returning true for this function will skip `LegalizeDxilInputOutputs`. Is that right? 

I'm particularly concerned about this usage: 
https://github.com/microsoft/DirectXShaderCompiler/blob/769b504be4748931254036e3250c1daad5a1991a/lib/HLSL/HLMatrixLowerPass.cpp#L467

pow2clk (commented):
As it happens, this overlaps with something I've been working on. I'm unsure about what IsEntryThatUsesSignatures should do in all the cases it is used.

tools/clang/test/HLSL/workgraph/work-graphs.hlsl
python3kgae: VertexID should hit error about only support vertex shader in Sema.
Now we have error but not in sema, so verifier cannot check the error.

lib/HLSL/HLModule.cpp
python3kgae: Yes. It is for SROA to flatten the struct input parameter.
I'll move the check to SROA.

python3kgae:
> Could you expand the commit message a bit more to explain what is going on here. The code change is adding support for `IsEntryThatUsesSignatures`, but the message only talks about SV_GroupID.
> 
> It looks like a more general change and would be good to explain it a bit more.

Updated.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/common_input_parameters.hlsl
```diff
@@ -0,0 +1,63 @@
...
+[Shader("node")]
+[NodeLaunch("coalescing")]
+[numthreads(4,4,4)]
+void node01(RWGroupNodeInputRecords<RECORD> input,
+ uint3 DTID : SV_DispatchThreadID,
+ uint3 GID : SV_GroupID,
+ uint GIdx : SV_GroupIndex,
+ uint3 GTID : SV_GroupThreadID )
+{
+  input.Get().dtid = DTID;
```
tex3d, line 59: Since this shader has total numthreads > 1, most writes to the RWGroupNodeInputRecords are data races.  This doesn't matter right now because we don't check for things like this yet, but if we were to (like we should), each of these tests would be expected to fail for this reason.  Perhaps we should change the code to be conditional on `GIdx == 0`, which would mean they get written only on the first thread in the group, and are no longer a data race.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -4228,7 +4228,8 @@ class SROA_Parameter_HLSL : public ModulePass {
 
       // Skip library function, except to LegalizeDxilInputOutputs
       if (&F != m_pHLModule->GetEntryFunction() &&
-          !m_pHLModule->IsEntryThatUsesSignatures(&F)) {
+          !m_pHLModule->IsEntryThatUsesSignatures(&F) &&
+          !m_pHLModule->IsNodeShader(&F)) {
```
tex3d, line 4232: This change doesn't go along with the PR title of "Mark Node as EntryThatUsesSignature for HLModule", and that title doesn't indicate what's actually been fixed either.  What about "Enable flattening of node non-object parameters" (if that's accurate)?

Can node actually be included in `EntryThatUsesSignature`, or will the one remaining use in HLMatrixLower blow up?  Can it be described why we include node in some places that call `IsEntryThatUsesSignatures` and not in another?

Also, There is a corresponding method with the same name on `DxilModule` which does include node and has different usages.  I think the names of the methods `EntryThatUsesSignature` need updating, since the meaning and purpose is not the same for each.

tex3d: Or "Flatten non-object parameters for node shaders"

tools/clang/test/HLSL/workgraph/work-graphs.hlsl
tex3d: SV_DispatchGrid is also different.  It is a special record field, instead of a system-generated input for the shader separate from the record.  It is also the only recognized system value inside a record (for now).

tex3d: The legal non-record/node object system value inputs for node shaders are described [here](https://microsoft.github.io/DirectX-Specs/d3d/WorkGraphs.html#node-shader-system-values).

tools/clang/test/HLSL/workgraph/work-graphs.hlsl
```diff
@@ -308,3 +308,24 @@ void node4_06(RWThreadNodeInputRecord<RECORD> input)
...
+struct R
+{
+  uint3 dtid: SV_DispatchThreadID;
+  uint3 gid: SV_GroupID;
+  uint gidx: SV_GroupIndex;
+  uint3 gtid: SV_GroupThreadID;
+};
+
+[Shader("node")]
+[NodeLaunch("coalescing")]
```
tex3d, line 321: I just realized another problem with this shader: `"coalescing"` launch only supports `SV_GroupIndex` and `SV_GroupThreadID`.  This must be changed to `"broadcasting"`, otherwise diagnostics that we should be implementing would cause this to fail.  I've filed #5827 to track diagnostics and validation implementation for these.

tex3d: And note: changing this to `"broadcasting"` will require a change of `RWGroupNodeInputRecords` to `RWDispatchNodeInputRecord`, and the addition of a `[NodeDispatchGrid(...)]` attribute, since you wouldn't want to add a `SV_DipatchGrid` field to the record, or you wouldn't be able to use the same struct in both places.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/common_input_parameters.hlsl
python3kgae: conditional on GIdx == 0.

tools/clang/test/HLSL/workgraph/work-graphs.hlsl
python3kgae: switched to broadcasting.

lib/HLSL/HLModule.cpp
tex3d: I don't think the usage in `HLMatrixLowerPass.cpp` is a problem.  I think we could go back to adding node to the `IsEntryThatUsesSignatures` function instead of checking separately here.

For `HLMatrixLowerPass.cpp`, I think the point is to skip casting from matrix type to lowered type if it's in a signature that gets flattened, since that would be handled elsewhere.  I don't think there's any harm in including node here because there are no scenarios where we expect argument pointers to be used for matrix where they would not be part of the regular flattened/lowered parameters, because we would never be accessing record objects through a pointer derived from the argument pointer.

tools/clang/test/HLSL/workgraph/work-graphs.hlsl
tex3d: It looks like you missed this file in particular.

python3kgae: Fixed.

tex3d (approved):
LGTM!

PR #5825, opened 2023-10-03
Update Barrier intrinsic to merge ACCESS and SYNC flags

### LINKED ISSUE #5780: Update Barrier() implementation with incoming spec change to remove SYNC_FLAG
**Description**
Minor spec changes are incoming for Barrier() intrinsic in SM 6.8.  This is the issue for tracking implementation work.

hlsl-specs issue: [[Work Graphs] Remove SYNC_FLAG from Barrier intrinsics #101](https://github.com/microsoft/hlsl-specs/issues/101) tracks the spec update.

tex3d:
- Merge ACCESS_FLAG AccessFlags and SYNC_FLAG SyncFlags into new BARRIER_SEMANTIC_FLAG SemanticFlags according to spec update
- Update lowering, DxilValidation, and tests
- Delete now-redundant validation tests

Fixes #5780


python3kgae (approved):

include/dxc/DXIL/DxilInstructions.h
```diff
@@ -8774,28 +8773,17 @@ struct DxilInst_BarrierByMemoryType {
                              llvm::IntegerType::get(Instr->getContext(), 32),
                              llvm::APInt(32, (uint64_t)val)));
   }
-  llvm::Value *get_AccessFlags() const { return Instr->getOperand(2); }
-  void set_AccessFlags(llvm::Value *val) { Instr->setOperand(2, val); }
-  int32_t get_AccessFlags_val() const {
+  llvm::Value *get_SemanticFlags() const { return Instr->getOperand(2); }
+  void set_SemanticFlags(llvm::Value *val) { Instr->setOperand(2, val); }
```
bob80905, line 8777: Wouldn't we want to use OperandIdx::arg_SemanticFlags here instead of "2" ?

tex3d: This is generated code, so the number should be correct.  The `OperandIdx::arg_SemanticFlags` are for use outside the generated code.  We could update the code generator to use these instead (and cast), but I don't think that would have any real benefit, and it certainly wouldn't be something to do in this PR.

bob80905 (approved):

PR #5829, opened 2023-10-05
[SPIRV] Implement SM6.6 implicit LOD operations in compute shaders

s-perron:
SPIRV has not yet implemented the changes in SM6.6 that allows [derivatives
in compute, mesh, and amplification shaders](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Derivatives.html). This is because there is no KHR
extension that enabled that capability in SPIR-V.

However, we have decided to use [SPV_NV_compute_shader_derivatives](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/NV/SPV_NV_compute_shader_derivatives.asciidoc)
to implement it for compute shader while we wait for a KHR extension.

This change only deals with the texture sample instructions.
The changes involve

1. modifying code that makes sure these only appear in fragment shaders
to allow compute shaders as well.
1. add the extension and capability
1. set the correct execution mode on the function when
the intrinsics are used in compute shaders.


s-perron:
This is ready to be reviewed, but I cannot merge it until the changes to spirv-opt are in, and the submodule has been updated.

sudonatalie (approved):
LGTM, just some nits

Issue #5800, opened 2023-09-28
DXC fails to report `TiledResources` feature requirement when specifying lod clamp on sample operations


tex3d:
**Description**
DXC fails to report `TiledResources` feature requirement when specifying lod clamp on sample operations.

**Steps to Reproduce**
```hlsl
// RUN: %dxc -T ps_6_0 %s | FileCheck %s

// From DXC disassembly comment:
// CHECK: Note: shader requires additional functionality:
// CHECK-NEXT: Tiled resources

// CHECK: !dx.entryPoints = !{![[entryPoints:[0-9]+]]}
// CHECK: ![[entryPoints]] = !{void ()* @main, !"main", !{{[0-9]+}}, !{{[0-9]+}}, ![[extAttr:[0-9]+]]}

// tag 0: ShaderFlags, 4096 = Tiled resources
// CHECK: ![[extAttr]] = !{i32 0, i64 4096}

Texture2D T2D;
SamplerState S;

float4 main(float2 coord : TEXCOORD, float c : CLAMP) : SV_Target {
  // LOD clamp requires TiledResources feature
  return T2D.Sample(S, coord, int2(0,0), c);
}
```

**Actual Behavior**
No required feature bit is set for Tiled Resources.

**Environment**
- DXC version: 1.7 - 1.7.2308.7 (69e54e290); dxil.dll: 1.7(101.7.2308.12)
- Host Operating System: Windows 10 22H2

PR #5830, opened 2023-10-05
Set tiled resource in shader flags.

python3kgae:
If clamp is not undef, mark tiled resource.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5800


damyanp (changes requested):

tools/clang/test/CodeGenDXIL/hlsl/shader-flags/tiled_res.hlsl
```diff
@@ -0,0 +1,19 @@
...
+// CHECK: !dx.entryPoints = !{![[entryPoints:[0-9]+]]}
+// CHECK: ![[entryPoints]] = !{void ()* @main, !"main", !{{[0-9]+}}, !{{[0-9]+}}, ![[extAttr:[0-9]+]]}
+
+// tag 0: ShaderFlags, 4096 = Tiled resources
+// CHECK: ![[extAttr]] = !{i32 0, i64 4096}
+
+Texture2D T2D;
+SamplerState S;
+
+float4 main(float2 coord : TEXCOORD, float c : CLAMP) : SV_Target {
```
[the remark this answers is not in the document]
python3kgae: More test added.

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -573,6 +573,19 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
           break;
+        case DXIL::OpCode::SampleGrad:
+        case DXIL::OpCode::SampleCmpBias:
+        case DXIL::OpCode::SampleCmpGrad:
+          if (!isa<Constant>(CI->getArgOperand(
+                  DXIL::OperandIndex::kTextureSampleOffset0OpIdx)) ||
+              !isa<Constant>(CI->getArgOperand(
+                  DXIL::OperandIndex::kTextureSampleOffset1OpIdx)) ||
+              !isa<Constant>(CI->getArgOperand(
+                  DXIL::OperandIndex::kTextureSampleOffset2OpIdx)))
```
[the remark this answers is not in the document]
python3kgae: Added static function to share the code.

tools/clang/test/CodeGenDXIL/hlsl/shader-flags/tiled_res.hlsl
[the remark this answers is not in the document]
damyanp: Thanks - I think I see two tests now, but there were two opcodes that were modified?  Do we not need coverage for the other one?

damyanp (approved):
Approving, but please see comment about whether or not all the affected opcodes are covered in the tests.

bob80905 (commented):
Nit: Would be good to add newlines at the ends of each of the test files.

python3kgae:
> Nit: Would be good to add newlines at the ends of each of the test files.

Updated.

bob80905 (approved):
Personally I think the filename should reflect the type of sampler, (e.g., tiled_res_sample_cmp_bias.hlsl, or tiled_res_sample_bias.hlsl)
But this all looks acceptable to me.

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -562,27 +572,23 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
-          if (!isa<Constant>(CI->getArgOperand(
-                  DXIL::OperandIndex::kTextureSampleOffset0OpIdx)) ||
-              !isa<Constant>(CI->getArgOperand(
-                  DXIL::OperandIndex::kTextureSampleOffset1OpIdx)) ||
-              !isa<Constant>(CI->getArgOperand(
-                  DXIL::OperandIndex::kTextureSampleOffset2OpIdx)))
-            hasAdvancedTextureOps = true;
+        case DXIL::OpCode::SampleCmpBias:
+          hasAdvancedTextureOps |= hasNonConstantSampleOffsets(CI);
+          if (!isa<UndefValue>(CI->getArgOperand(CI->getNumArgOperands() - 1)))
```
tex3d, line 590: I think this should consider a constant value of `0.0f` as not using the clamp as well.

tools/clang/test/CodeGenDXIL/hlsl/shader-flags/tiled_res.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: %dxc -T ps_6_0 %s | FileCheck %s
```
tex3d: I find the separation and naming of these tests a little opaque.  They test that the flag is set for the use of the clamp argument specifically, which isn't in the naming.  Also, the added `_01` ... numbering isn't nearly as good as having the name of the method tested.  If 'Sample' is a given, these could be `tiled_res_clamp` and `tiled_res_clamp_[cmp]_[grad|bias]`.  But even that might be unnecessary if all these tests could be the same test, with different methods triggered by different define and run lines.

I also think new cases need to be tested: make sure no clamp or using constant `0.0f` for clamp does not set the flag.

Also, all these variants could be implemented in one test with multiple RUN lines rather than a bunch of separate test files with duplicated run lines.  I don't know what others prefer, but IMO if you have a test that only varies something minor that could be varied in a run line, it's a lot easier to tell that's the case, see the breadth of coverage, and maintain the tests if written in a more templated way using multiple run lines.

tools/clang/test/CodeGenDXIL/hlsl/shader-flags/tiled_res_sample.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: %dxc -T ps_6_0 %s | FileCheck %s
```
tex3d, line 1: Apparently, there was an update that already covered some of the feedback I had in my initial comment on the tests.

Revising:

Related to my other comment, I think new cases need to be tested: make sure that both no clamp usage and using constant `0.0f` for clamp does not set the flag.  Also, if a constant value `0.0f` is tested to make sure the flag doesn't get set, then we should probably have another non-zero constant value to make sure it does get set in that case.

I think all these variants could be implemented in one test with multiple RUN lines rather than a bunch of separate test files with duplicated run lines.  I don't know what others prefer, but IMO if you have a test that only varies something minor that could be varied in a run line, it's a lot easier to tell that's the case, see the breadth of coverage, and maintain the tests if written in a more templated way using multiple run lines.

tex3d (changes requested):

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -562,27 +572,23 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
         case DXIL::OpCode::SampleBias:
         case DXIL::OpCode::SampleCmp:
-          if (!isa<Constant>(CI->getArgOperand(
-                  DXIL::OperandIndex::kTextureSampleOffset0OpIdx)) ||
-              !isa<Constant>(CI->getArgOperand(
-                  DXIL::OperandIndex::kTextureSampleOffset1OpIdx)) ||
-              !isa<Constant>(CI->getArgOperand(
-                  DXIL::OperandIndex::kTextureSampleOffset2OpIdx)))
-            hasAdvancedTextureOps = true;
+        case DXIL::OpCode::SampleCmpBias:
```
tex3d, line 588: Is there not an additional feature flag for using `SampleCmpBias` (and `SampleCmpGrad`) in the first place, or is it a required part of SM 6.8?

tools/clang/test/HLSLFileCheck/hlsl/objects/Texture/has_lod_clmap.hlsl
```diff
@@ -0,0 +1,59 @@
...
+// CHECK: ![[entryPoints]] = !{void ()* @[[name]], !"[[name]]", !{{[0-9]+}}, !{{[0-9]+}}, ![[extAttr:[0-9]+]]}
+
+// tag 0: ShaderFlags, 4096 = Tiled resources
+// CHECK: ![[extAttr]] = !{i32 0, i64 4096}
+
+Texture2D T2D;
+SamplerState S;
+
+float4 test_sample(float2 coord : TEXCOORD, float c : CLAMP) : SV_Target {
+  // LOD clamp requires TiledResources feature
```
tex3d, line 25: nit: These comments seem redundant now.

tools/clang/test/HLSLFileCheck/hlsl/objects/Texture/no_lod_clmap.hlsl
```diff
@@ -0,0 +1,93 @@
...
+// CHECK: ![[entryPoints]] = !{void ()* @[[name]], !"[[name]]", !{{[0-9]+}}, !{{[0-9]+}}, null}
+
+// tag 0: ShaderFlags, 4096 = Tiled resources
+// CHECK-NOT:!{i32 0, i64 4096}
+
+Texture2D T2D;
+SamplerState S;
+
+float4 test_sample(float2 coord : TEXCOORD) : SV_Target {
+  // LOD clamp requires TiledResources feature
```
tex3d, line 28: nit: These comments don't seem correct when checking the negative case.  One general comment for the file should suffice.

lib/DXIL/DxilShaderFlags.cpp
python3kgae: These 2 are required by SM6.8.

tex3d (approved):
LGTM!

PR #5832, opened 2023-10-05
Removing dx.temp for lib target, since it's disallowed by the validator

adam-yang:
This change makes `DxilFinalizeModule` remove `dx.temp` metadata regardless of target profile.

`DxilFinalizeModule` pass removes metadata that the validator disallows in the final DXIL. These metadata are only removed for non-lib target, however dx.temp is something that the validator does not allow even in lib target. There is no reason to keep this metadata because it's only used for inserting nops and that's not something we do when linking libraries.


dmpots (approved):
LGTM

tex3d (approved):
LGTM.

Issue #5822, opened 2023-10-03
Work Graph objects don't support dependent template arguments


tex3d:
**Description**
Work Graph node objects cannot be instantiated in a template with a dependent template argument.  If you try, you get a compilation error: "'T' cannot be used as a type parameter where a struct/class is required".

This may be related to #5729, but it's probably a separate issue.

**Steps to Reproduce**
```hlsl
// dxc -ast-dump -T lib_6_8
template <typename T> void doSomething(GroupNodeInputRecords<T> data) {}
struct Record { int id; };

[Shader("node")]
[NodeLaunch("coalescing")]
[NumThreads(4,5,6)]
void node(GroupNodeInputRecords<Record> inputData) {
  doSomething(inputData);
}
```

**Actual Behavior**
```
error: 'T' cannot be used as a type parameter where a struct/class is required
void doSomething(GroupNodeInputRecords<T> data) {
                                       ^
error: no matching function for call to 'doSomething'
  doSomething<Record>(inputData);
  ^~~~~~~~~~~~~~~~~~~
```

**Environment**
- DXC version: recent `staging-sm-6.8` branch (f40598310)
- Host Operating System: any

PR #5833, opened 2023-10-06
Allow DependentType for Node objects template argument.

python3kgae:
Allow DependentType for case like
 template < typename T > void foo(GroupNodeInputRecords< T > data) {}

Also make sure specialization is done before check template argument is not empty struct for case like
template < typename T > void bar(GroupNodeInputRecords< Record < T > > data) {}

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5822


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5377,6 +5377,10 @@ class HLSLExternalSource : public ExternalSemaSource {
         return true;
       }
       QualType argType = arg.getAsType();
+      // Allow TemplateTypeParmType for case like
+      // template <typename T> void foo(GroupNodeInputRecords<T> data) {}
+      if (argType->getAs<TemplateTypeParmType>())
```
tex3d, line 5382: Couldn't we check for `argType->isDependentType()` instead?

python3kgae: Yes. isDependentType works.
Hit other issue in test.
Debugging.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5377,6 +5377,26 @@ class HLSLExternalSource : public ExternalSemaSource {
         return true;
       }
       QualType argType = arg.getAsType();
+      // Allow DependentType for case like
+      // template <typename T> void foo(GroupNodeInputRecords<T> data) {}
+      if (argType->isDependentType())
+        return false;
```
llvm-beanz, line 5383: This will need to be validated during instantiating right? Can you add tests for failure cases?

python3kgae: Done.

tools/clang/test/SemaHLSL/hlsl/workgraph/template.hlsl
tex3d, line 1: Couldn't these be added to the existing `template-diag-deferred.hlsl`?

tools/clang/test/SemaHLSL/hlsl/workgraph/template.hlsl
```diff
@@ -0,0 +1,24 @@
...
+// GroupNodeInputRecords may not non-struct/class DependentType. 
+
+template<typename T>
+struct Record {  }; // expected-note{{zero sized record defined here}}
+
+template <typename T> void foo(GroupNodeInputRecords<T> data) {}
+
+template <typename T> void bar() {
+  GroupNodeInputRecords<Record<T> > data; // expected-error{{record used in GroupNodeInputRecords may not have zero size}}
+  foo(data);
```
tex3d, line 13: None of the calls to foo() are causing the diagnostic to fire.  I guess it's because it already fired for the local, invalidating the `data` decl?  What if something constructs a local method with an invalid argument type based on the template type, without having called that method.  For example:

```hlsl
template <typename T>
class CFoo {
  T var;
  void foo(GroupNodeInputRecords<T> data) {}
};
// ...
  CFoo<int> c;
```

Should that result in a diagnostic triggered by the `CFoo<int> c` instantiation?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5380,20 +5380,32 @@ class HLSLExternalSource : public ExternalSemaSource {
       }
 
       QualType ArgTy = Arg.getAsType();
+      // Allow DependentType for case like
+      // template <typename T> void foo(GroupNodeInputRecords<T> data) {}
+      if (ArgTy->isDependentType())
```
tex3d, line 5385: Is the `case AR_TOBJ_DEPENDENT` in `DiagnoseNodeStructArgument` still hit with this code?  If so, in what case?  Is there a test here that hits it?

tex3d (approved):
Looks good, some additional thoughts, nothing blocking.

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: Added.

tools/clang/test/SemaHLSL/hlsl/workgraph/template.hlsl
python3kgae: Added.

tools/clang/test/SemaHLSL/hlsl/workgraph/template.hlsl
python3kgae: Cannot find VerifierTest which uses template-diag-deferred.hlsl. :(
And lit Shell test is easier to add.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: Was the issue resolved?

python3kgae: Yes. It is resolved.

tools/clang/test/SemaHLSL/hlsl/workgraph/template.hlsl
pow2clk: Ugh. We really need to audit all those unused test files and get them hooked up

python3kgae: template-diag-deferred.hlsl was added in https://github.com/microsoft/DirectXShaderCompiler/commit/baae5dd1e52f5da15e829a3199e17499cb9e55c6, but never added for taef test.
Maybe we should just add it as lit Shell test in a separate PR?

tools/clang/test/HLSLFileCheck/hlsl/workgraph/template.hlsl
llvm-beanz, line 1: Can we make these test file names more descriptive?

tools/clang/test/SemaHLSL/hlsl/workgraph/template.hlsl
llvm-beanz: > Maybe we should just add it as lit Shell test in a separate PR?

If this is going to be a separate PR, can you file an issue to track it so that we don't forget about it?

tools/clang/test/SemaHLSL/hlsl/workgraph/template.hlsl
```diff
@@ -0,0 +1,45 @@
...
+
+template <typename T>
+class CBar {
+  T var;
+  void bar(GroupNodeInputRecords< Texture2D<T> > data) {} // expected-error{{'Texture2D<float>' is not valid as a node record type - struct/class required}}
+};
+
+void testCBar() {
+  CBar<float> c; // expected-note{{in instantiation of template class 'CBar<float>' requested here}}
+}
```
llvm-beanz, line 45: Can we also add a test where being zero-sized is based on the template to ensure that instantiation is happening?

Something like:
```c++
struct ZeroSize {};

template<typename T>
struct RecordTemplate { T Value; }; // expected-note{{zero sized record defined here}}

template <typename T> void oopsie() {
  GroupNodeInputRecords<RecordTemplate<ZeroSize> > data; // expected-error{{record used in GroupNodeInputRecords may not have zero size}}
  foo(data);
}
```

Additionally we should verify no diagnostic is emitted for something like:

```c++
template <typename T> void woo() {
  GroupNodeInputRecords<RecordTemplate<int> > data;
  foo(data);
}
```

llvm-beanz (commented):
A few minor comments, but mostly looks good.

tools/clang/test/SemaHLSL/hlsl/workgraph/template.hlsl
python3kgae: Added https://github.com/microsoft/DirectXShaderCompiler/issues/5870 to track it.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5380,20 +5380,32 @@ class HLSLExternalSource : public ExternalSemaSource {
       }
 
       QualType ArgTy = Arg.getAsType();
+      // Allow DependentType for case like
+      // template <typename T> void foo(GroupNodeInputRecords<T> data) {}
```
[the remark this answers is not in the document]
llvm-beanz: Instead of applying my suggestion, you kept the old comment and the new comment, which seems confusing.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/template.hlsl
llvm-beanz, line 1: You renamed one test file, but they all had bad names. Can you rename the other ones too?

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
python3kgae: :( Updated. It looks the same in github email.

llvm-beanz (approved):

PR #5834, opened 2023-10-06
Pull in llvm-project's clang-format check action

llvm-beanz:
This action and scripting was written by @tru for LLVM and addresses some of the false positives we've been seeing in our action. There is more room for improvement, but it would be better to unify than to have divergent approaches since we're solving the same problem.


dmpots (approved):

python3kgae (approved):

.github/workflows/clang-format-checker.yml
```diff
@@ -1,38 +1,54 @@
...
-          fetch-depth: 0
-      - name: Comparing PR against target branch
+          fetch-depth: 2
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v39
+        with:
+          separator: ","
+          fetch_depth: 100 # Fetches only the last 10 commits
```
tex3d, line 20: Is it 
> the last 10 commits

or 100?

PR #5843, opened 2023-10-09
D3D Linux build header changes

hekota:
Header changes to make Direct3D WSL build work:
- Move non-WIN32 IMalloc definition back to `WinAdapter.h` (it was there until commit 44f88339383018d35155ff4f337ce73eefa5cb11)
- Remove `WinIncludes.h` from `FileIOHelper.h` and `microcom.h`
- Make sure `WinIncludes.h` is included early in source files that also need `FileIOHelper.h` and/or `microcom.h`


tex3d (approved):
LGTM.

damyanp:
Do we understand why the IMalloc definition was moved previously?

damyanp (approved):

tex3d:
> Do we understand why the IMalloc definition was moved previously?

I think it was when `DxcCoGetMalloc` was added, which depended on `IMalloc`, but it could use a forward declaration, instead of moving it.  It was likely moved in an earlier iteration of that change when more code was added there, but now there's no reason it needs to be defined there.

PR #5851, opened 2023-10-11
[Sanitizer] Fix sanitizer fails in msvc

python3kgae:
1. Add -sanitizer to hctbuild.
2. Detach pCompileSource in recompile to avoid release freeed memory.
3. Fix Execution test issue about Init size.

This is for https://github.com/microsoft/DirectXShaderCompiler/issues/5876


tools/clang/tools/dxclib/dxc.cpp
```diff
@@ -803,10 +803,8 @@ void DxcContext::Recompile(IDxcBlob *pSource, IDxcLibrary *pLibrary,
                            pIncludeHandler, &pResult));
   }
 
-#ifndef _WIN32
-  // FIXME: fix crash on linux when not detach pCompileSource.
+  // Detach pCompileSource because the data is owned by pPdbUtils.
```
tex3d, line 806: I believe the proper fix is in the loop earlier - to change `pCompileSource.Attach(pSourceFile);` to `pCompileSource = pSourceFile` or `pCompileSource.Attach(pSourceFile.Detach())` to transfer the reference.

python3kgae: Updated to pCompileSource.Attach(pSourceFile.Detach())

cmake/modules/HandleLLVMOptions.cmake
```diff
@@ -533,7 +533,7 @@ if(LLVM_USE_SANITIZER)
       message(WARNING "Unsupported value of LLVM_USE_SANITIZER: ${LLVM_USE_SANITIZER}")
     endif()
   else()
-    message(WARNING "LLVM_USE_SANITIZER is not supported on this platform.")
+    append("-fsanitize=address" CMAKE_C_FLAGS CMAKE_CXX_FLAGS)
```
pow2clk, line 536: I think it would be more appropriate to check the values of LLVM_USE_SANITIZER to make sure that it is set to address as is done for Unix platforms. 

The error message comes from the original LLVM 3.7 source. I'm a bit concerned that they didn't have this enabled either. Do we know anything about what they did to enable it in later versions?

utils/hct/hctbuild.cmd
```diff
@@ -204,6 +204,13 @@ if "%1"=="-default-adapter" (
   shift /1
   shift /1 & goto :parse_args
 )
+if "%1"=="-sanitizer" (
+  set CMAKE_OPTS=%CMAKE_OPTS% -DLLVM_ENABLE_SANITIZER=Address
+  set CMAKE_OPTS=%CMAKE_OPTS% -DLLVM_USE_SANITIZER:BOOL=On
```
pow2clk, line 209: I don't see `LLVM_ENABLE_SANITIZER` used anywhere. Additionally, it looks as though `LLVM_USE_SANITIZER` is expected to be a string not a bool.

pow2clk (commented):
Just a couple notes about the Cmake defines

cmake/modules/HandleLLVMOptions.cmake
python3kgae: Changed into only support Address and report Unsupported value of LLVM_USE_SANITIZER: ${LLVM_USE_SANITIZER}.

utils/hct/hctbuild.cmd
python3kgae: Fixed.

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
```diff
@@ -1348,7 +1353,7 @@
   <ShaderOp Name="Triangle" PS="PS" VS="VS">
     <RootSignature>RootFlags(ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT)</RootSignature>
 
-    <Resource Name="VBuffer" Dimension="BUFFER" Width="1024" Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST" Init="FromBytes">
+    <Resource Name="VBuffer" Dimension="BUFFER" Width="84" Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST" Init="FromBytes">
```
bob80905, line 1356: Did you intend to make this 48?

tools/clang/tools/dxclib/dxc.cpp
```diff
@@ -778,7 +778,8 @@ void DxcContext::Recompile(IDxcBlob *pSource, IDxcLibrary *pLibrary,
     IFT(pPdbUtils->GetSourceName(i, &pFileName));
     IFT(pIncludeHandler->insertIncludeFile(pFileName, pSourceFile, 0));
     if (pMainFileName == pFileName) {
-      pCompileSource.Attach(pSourceFile);
+      // Transfer pSourceFile to avoid pPdbUtils release it again.
```
tex3d, line 781: Nit: The comment is misleading.

Assignment ('=') would work just fine too, since it would get another ref for pCompileSource so pSourceFile wouldn't release the only ref we hold outside the loop iteration.  This Attach/Detach pattern is just saving one AddRef+Release by taking pSourceFile's reference.

PdbUtils will still release its ref, but we'll have our own valid ref, so it's ok.  In any case, it's not PdbUtils we need to worry about, it's managing our refs properly, so we still have one as long as we need one.

This comment makes it sound like you need to do something special with pSourceFile when retrieved from PdbUtils.  But there's nothing special here.

You could say:
```cpp
// Transfer pSourceFile to avoid extra AddRef+Release.
```

tex3d (approved):
Looks fine, with one comment nit.

bob80905 (approved):

tools/clang/tools/dxclib/dxc.cpp
python3kgae: Updated in https://github.com/microsoft/DirectXShaderCompiler/pull/5882

PR #5853, opened 2023-10-11
Fix clang diagnostics overflow in SEMA group; add static_assert

tex3d:
Overflow of a diagnostic group could lead to a crash, and we just overflowed the diagnostics in the SEMA group.

This change bumps the count for SEMA and adds several checks to catch this in the future:
- static_asserts to prevent overflow of group IDs
- assert to catch invalid ID in group more clearly
- assert to catch failure of setSeverityForGroup in ASTUnit::LoadFromCommandLine
- keep crash on invalid DiagID to prevent late failure with NDEBUG

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticIDs.h
+++ b/tools/clang/include/clang/Basic/DiagnosticIDs.h
@@ -36,7 +36,8 @@ namespace clang {
       DIAG_START_AST           = DIAG_START_PARSE           +  500,
       DIAG_START_COMMENT       = DIAG_START_AST             +  110,
       DIAG_START_SEMA          = DIAG_START_COMMENT         +  100,
-      DIAG_START_ANALYSIS      = DIAG_START_SEMA            + 3000,
+      // HLSL Change: SEMA group length increased from 3000.
+      DIAG_START_ANALYSIS      = DIAG_START_SEMA            + 3100,
       DIAG_UPPER_LIMIT         = DIAG_START_ANALYSIS        +  100
     };
 
--- a/tools/clang/lib/Basic/DiagnosticIDs.cpp
+++ b/tools/clang/lib/Basic/DiagnosticIDs.cpp
@@ -129,8 +129,8 @@ static const StaticDiagInfoRec *GetDiagInfo(unsigned DiagID) {
   unsigned ID = DiagID - DIAG_START_COMMON - 1;
   // HLSL Change: Added static_asserts to prevent overflow in each category.
 #define CATEGORY(NAME, PREV)                                                   \
-  static_assert(NUM_BUILTIN_##PREV##_DIAGNOSTICS < DIAG_START_##NAME &&        \
-                "otherwise, previous group overflows");                        \
+  static_assert(NUM_BUILTIN_##PREV##_DIAGNOSTICS < DIAG_START_##NAME,          \
+                "otherwise, " #PREV " diagnostic group overflows");            \
   if (DiagID > DIAG_START_##NAME) {                                            \
     Offset += NUM_BUILTIN_##PREV##_DIAGNOSTICS - DIAG_START_##PREV - 1;        \
     ID -= DIAG_START_##NAME - DIAG_START_##PREV;                               \
@@ -145,8 +145,8 @@ static const StaticDiagInfoRec *GetDiagInfo(unsigned DiagID) {
   CATEGORY(SEMA, COMMENT)
   CATEGORY(ANALYSIS, SEMA)
 #undef CATEGORY
-  static_assert(NUM_BUILTIN_ANALYSIS_DIAGNOSTICS < DIAG_UPPER_LIMIT &&
-                "otherwise, ANALYSIS group overflows");
+  static_assert(NUM_BUILTIN_ANALYSIS_DIAGNOSTICS < DIAG_UPPER_LIMIT,
+                "otherwise, ANALYSIS diagnostic group overflows");
 
   // Avoid out of bounds reads.
   if (ID + Offset >= StaticDiagInfoSize)
--- a/tools/clang/lib/Basic/DiagnosticIDs.cpp
+++ b/tools/clang/lib/Basic/DiagnosticIDs.cpp
@@ -127,21 +127,26 @@ static const StaticDiagInfoRec *GetDiagInfo(unsigned DiagID) {
   // memory at all.
   unsigned Offset = 0;
   unsigned ID = DiagID - DIAG_START_COMMON - 1;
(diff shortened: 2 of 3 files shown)


tools/clang/include/clang/Basic/DiagnosticIDs.h
```diff
@@ -36,7 +36,8 @@ namespace clang {
       DIAG_START_AST           = DIAG_START_PARSE           +  500,
       DIAG_START_COMMENT       = DIAG_START_AST             +  110,
       DIAG_START_SEMA          = DIAG_START_COMMENT         +  100,
-      DIAG_START_ANALYSIS      = DIAG_START_SEMA            + 3000,
+      // HLSL Change: SEMA group length increased from 3000.
+      DIAG_START_ANALYSIS      = DIAG_START_SEMA            + 3100,
```
tex3d, line 40: Oh great, I touched a file (not ours) that doesn't currently pass clang-format check :(

What should I do?  clang-format the file?  I don't like that option.

Here's an example where the clang-formatted code is harder to read, too.
Is this really better?
```cpp
  enum {
    DIAG_START_COMMON = 0,
    DIAG_START_DRIVER = DIAG_START_COMMON + 300,
    DIAG_START_FRONTEND = DIAG_START_DRIVER + 100,
    DIAG_START_SERIALIZATION = DIAG_START_FRONTEND + 100,
    DIAG_START_LEX = DIAG_START_SERIALIZATION + 120,
    DIAG_START_PARSE = DIAG_START_LEX + 300,
    DIAG_START_AST = DIAG_START_PARSE + 500,
    DIAG_START_COMMENT = DIAG_START_AST + 110,
    DIAG_START_SEMA = DIAG_START_COMMENT + 100,
    // HLSL Change: SEMA group length increased from 3000.
    DIAG_START_ANALYSIS = DIAG_START_SEMA + 3100,
    DIAG_UPPER_LIMIT = DIAG_START_ANALYSIS + 100
  };
```

tex3d: If I start reformatting from here, the whole file will need reformatting, since it doesn't like the tabbed-in namespaces.  I really don't want to reformat the file as part of this change.  It really doesn't make sense.  I don't think we should reformat the whole file at all.

tools/clang/lib/Basic/DiagnosticIDs.cpp
```diff
@@ -540,7 +545,10 @@ static bool getDiagnosticsInGroup(diag::Flavor Flavor,
   // Add the members of the option diagnostic set.
   const int16_t *Member = DiagArrays + Group->Members;
   for (; *Member != -1; ++Member) {
-    if (GetDiagInfo(*Member)->getFlavor() == Flavor) {
+    // HLSL Change: Check result of GetDiagInfo
+    const StaticDiagInfoRec *Info = GetDiagInfo(*Member);
+    assert(Info && "otherwise, group contains invalid diag ID");
```
dmpots, line 550: Should we be using `DXASSERT` here and below?

tools/clang/include/clang/Basic/DiagnosticIDs.h
hekota: In this case I think it would be best to override the PR check for clang format if this is the only issue.

pow2clk: I concur

tools/clang/lib/Frontend/ASTUnit.cpp
```diff
@@ -2014,7 +2014,13 @@ ASTUnit *ASTUnit::LoadFromCommandLine(
...
-      Diags->setSeverityForGroup(diag::Flavor::WarningOrError, StringRef(groupNames[i]), diag::Severity::Ignored);
+      if (Diags->setSeverityForGroup(diag::Flavor::WarningOrError,
+                                     StringRef(groupNames[i]),
+                                     diag::Severity::Ignored)) {
+        assert(false &&
+               "otherwise, there is a problem with diagnostic definitions.");
+        return nullptr;
+      }
     }
     Diags->setExtensionHandlingBehavior(diag::Severity::Ignored);
```
dmpots, line 2025: Do we need to check the return from this call as well?

dmpots (approved):
Is there any way to write a test for this change?

tools/clang/lib/Basic/DiagnosticIDs.cpp
```diff
@@ -127,21 +127,26 @@ static const StaticDiagInfoRec *GetDiagInfo(unsigned DiagID) {
...
-CATEGORY(ANALYSIS, SEMA)
+  CATEGORY(DRIVER, COMMON)
+  CATEGORY(FRONTEND, DRIVER)
+  CATEGORY(SERIALIZATION, FRONTEND)
+  CATEGORY(LEX, SERIALIZATION)
+  CATEGORY(PARSE, LEX)
+  CATEGORY(AST, PARSE)
+  CATEGORY(COMMENT, AST)
+  CATEGORY(SEMA, COMMENT)
+  CATEGORY(ANALYSIS, SEMA)
```
tex3d, line 146: By the way, I verified that this static_assert caught the failure causing our break before I updated the count.

tools/clang/include/clang/Basic/DiagnosticIDs.h
pow2clk: Considering the rate at which we've been adding Sema Diagnostics, 100 more seems kind of low. Is there a reason not to add 1000?

pow2clk (approved):
I think everything here is fine provided you can silence the clang-format failure. Approving to clear the path sooner trusting everything will be worked out

tools/clang/lib/Frontend/ASTUnit.cpp
tex3d: This returns void.

tools/clang/lib/Basic/DiagnosticIDs.cpp
tex3d: This is not in HLSL-specific code, so I think it's more appropriate to use assert here.

tex3d:
@dmpots
> Is there any way to write a test for this change?

I can't think of any.

The static_assesrt will fire if diagnostic IDs from a category go higher than expected, which will cause a build break (as it should), so that can't be tested.  I don't know how else an invalid ID could get into the categories, and the ASTUnit change is to make sure we didn't make a mistake in the category names we reference.

So, I can't think of any way to test any of these changes, other than local hacks that either break the build, or break the compiler in some way.

Issue #5857, opened 2023-10-12
[Diagnostics] Only emit diagnostics for recursive functions in library shaders that are exported.


bob80905:
**Is your feature request related to a problem? Please describe.**
Diagnostics are emitted for all recursive functions inside of library shaders, when they should only be emitted for functions that are exported in library shaders.

**Describe the solution you'd like**
The logic should be changed to only emit diagnostics on exported functions. Specifically, functions that follow this rule should emit a diagnostic if it is recursive:
if the target is lib_6_x, all but 'static' functions are exported; if the target is any other lib, only 'export' functions are exported.

PR #5858, opened 2023-10-12
[Diagnostics] Improve recursion diagnostics, specifically for library shaders

bob80905:
Previously, there would be no recursion validation performed on any function declarations in library shaders. This PR adds code that validates that no function declarations that are found inside library shaders are recursive. The diagnostics aren't repeated, since there's logic to prevent the same diagnostic from being emitted by 2 similar ancestors of the same recursive function. The PR also adjusts recursion detection for shaders in non-library cases, and improves the original diagnostic by adding information about the name of the function that is recursive. The PR is meant to complete the work on issue #5789. However, more work needs to be done for full accuracy. 
The PR checks *all* function declarations in the translation unit for the library shader case, when it should really only be checking a subset of function declarations. This issue is filed here: #5857
Fixes #5789


bob80905:
The new "CheckReachability" function is unused, but was helpful during iteration, and satisfies the description of the CallGraphWithRecurseGuard class. The class advertised that checking for reachability efficiently was one feature that was provided, but there isn't a function to do that explicitly, so I figured I'd add the function here.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7802,6 +7802,10 @@ def err_hlsl_logical_binop_scalar : Error<
    "operands for short-circuiting logical binary operator must be scalar, for non-scalar types use '%select{and|or}0'">;
 def err_hlsl_ternary_scalar : Error<
    "condition for short-circuiting ternary operator must be scalar, for non-scalar types use 'select'">;
+def err_hlsl_no_recursion : Error<
+   "recursive functions are not allowed: entry (or export) function calls recursive function '%0'">;
+def err_hlsl_no_recursion_via_patch : Error<
+   "recursive functions are not allowed via patch constant functions: entry function calls recursive function '%0'">;
```
python3kgae, line 7808: patch constant function calls recursive function '%0'  ?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15690,6 +15672,98 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
...
+    CG.BuildForEntry(FD);
+    FunctionDecl *pResult = CG.CheckRecursion(FD);
+    if (pResult) {
+      return std::make_pair(pResult, false);
+    }
+    if (pPatchFnDecl) {
+      CG.BuildForEntry(pPatchFnDecl);
+      pResult = CG.CheckRecursion(pPatchFnDecl);
+      if (pResult) {
+        return std::make_pair(pResult, true);
```
pow2clk, line 15717: Rather than test these one after the other, I think it makes sense to check for the missing patch constant function after the entry function by the caller. So instead of the above single call to ValidateNoRecursion and the complicated return value, we could test the entry function, if it's HS, test if there is a patch constant function and error if not, then test that for recursion with another call to this function.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15690,6 +15672,98 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
...
+    if (FunctionDecl *FD = dyn_cast<FunctionDecl>(decl)) {
+      // returns the first recursive function declaration detected
+      // from this function declaration FD, and determines whether
+      // the recursion was detected in the patch-constant function
+      std::pair<clang::FunctionDecl *, bool> pResult =
+          ValidateNoRecursion(self, FD);
+      // if there is a function that was detected to be recursive,
+      // then make sure it is saved for later to emit a diagnostic
+      if (pResult.first) {
+        FDecls.insert(pResult);
```
pow2clk, line 15739: I was wondering why the function decl was returned. I guess this is why

tools/clang/test/HLSLFileCheck/hlsl/functions/recursion/patch_constant_func_recursion.hlsl
```diff
@@ -0,0 +1,50 @@
+// RUN: %dxc -E HSmain -T hs_6_0 %s | FileCheck %s
```
pow2clk, line 1: Given there is code that is expressly for HS shaders compiled into libraries, I think there should be a test for that with a function that is annotated with `shader("hull")` and also `patchconstantfunc("fooey")`

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -3024,6 +3024,20 @@ class CallGraphWithRecurseGuard {
...
+    auto node = m_callNodes.find(FD1);
+    if (node != m_callNodes.end()) {
+      for (FunctionDecl *Callee : node->second.CalleeFns) {
+        if (CheckReachability(Callee, FD2))
+          return true;
+      }
+    }
+    return false;
+  }
+
```
pow2clk, line 3040: This doesn't look to be used?

pow2clk (commented):
Looks like the right direction! I think too much effort was made to keep the patch constant function special in the output. I think eliminating that requirement will simplify a lot.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Sorry, I just read your comment

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7802,6 +7802,12 @@ def err_hlsl_logical_binop_scalar : Error<
    "operands for short-circuiting logical binary operator must be scalar, for non-scalar types use '%select{and|or}0'">;
 def err_hlsl_ternary_scalar : Error<
    "condition for short-circuiting ternary operator must be scalar, for non-scalar types use 'select'">;
+def err_hlsl_no_recursion : Error<
+   "recursive functions are not allowed: entry (or export) function calls recursive function '%0'">;
+def err_hlsl_no_recursion_via_patch : Error<
+   "recursive functions are not allowed: patch constant function calls recursive function '%0'">;
```
llvm-beanz, line 7808: Can we merge these two diagnostics and use a select instead?

```suggestion
def err_hlsl_no_recursion : Error<
   "recursive functions are not allowed: %select{entry|export|patch constant}1 function calls recursive function '%0'">;
```

Then you can just add `0`, `1`, or `2` to select between “entry”, “export” or “patch constant” as appropriate.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11372,6 +11386,31 @@ bool hlsl::DiagnoseNodeStructArgument(Sema *self, TemplateArgumentLoc ArgLoc,
...
+// that could serve as a candidate to that patch constant function.
+void ValidatePatchConstantFunctionsExist(clang::Sema *self) {
+  for (auto decl : self->getASTContext().getTranslationUnitDecl()->decls()) {
+    // TODO: improve condition so that only exported functions are checked,
+    // instead of all functions. Issue: #5857
+    if (FunctionDecl *FD = dyn_cast<FunctionDecl>(decl)) {
+      // If there is no patch constant function, then we don't need to validate
+      // anything.
+      if (const HLSLPatchConstantFuncAttr *Attr =
+              FD->getAttr<HLSLPatchConstantFuncAttr>()) {
```
llvm-beanz, line 11400: If I’m reading this correctly we call this function only on library shaders, then we iterate all functions looking for the patch constant attribute.

Do we somewhere else verify that the patch constant attribute is only present on entry functions marked as hull shaders?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11458,48 +11497,39 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
   }
 
-  // Validate that there is no recursion; start with the entry function.
-  // NOTE: the information gathered here could be used to bypass code generation
-  // on functions that are unreachable (as an early form of dead code
-  // elimination).
-  if (pEntryPointDecl) {
-    const auto *shaderModel =
-        hlsl::ShaderModel::GetByName(self->getLangOpts().HLSLProfile.c_str());
+  FunctionDecl *pResult = ValidateNoRecursion(self, pEntryPointDecl);
```
llvm-beanz, line 11500: nit: I have a strong aversion to introducing new code that uses Hungarian notation, and so does the LLVM style guide. This code is pretty mixed in how it names things, I don’t expect that you’ll fix that, but lets at least add new things with more style-consistent names.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -3024,6 +3024,20 @@ class CallGraphWithRecurseGuard {
     }
   }
 
+  // return true if FD2 is reachable from FD1
+  bool CheckReachability(FunctionDecl *FD1, FunctionDecl *FD2) {
```
llvm-beanz, line 3028: If this is only intended to stick around for debugging purposes, we should probably wrap the function in:

```c
#if !defined(NDEBUG) || defined(LLVM_ENABLE_DUMP)
```

So that we know it is for debugging only.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: In short, yes. It's not accessible to godbolt because it's still stuck in staging-sm-6.8, but we have catches to determine both when this is inappropriately applied to other entry points and also when it is missing from the hull entry point. 

#5789 made sure that _something_ was specified as the patch constant function. This verifies that what was specified is a real function. I really don't know for sure if those checks should be one and the same. I can see an argument for separating the checks that you've declared _something_ and that the something you declared is actually valid. My gut reaction is to leave this as it is.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: "nit:" and "strong aversion" give some pretty divergent messages here. I am emphatically indifferent. If this were to diverge from its immediately approximate Hungarian brethren, it would stand out, but would be loved just the same. If it chose to go with the flow of its surroundings, I would understand completely.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15690,6 +15720,61 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
...
+                                         clang::FunctionDecl *FD) {
+  // Validate that there is no recursion reachable by this function declaration
+  // NOTE: the information gathered here could be used to bypass code generation
+  // on functions that are unreachable (as an early form of dead code
+  // elimination).
+  if (FD) {
+    hlsl::CallGraphWithRecurseGuard CG;
+    CG.BuildForEntry(FD);
+    FunctionDecl *pResult = CG.CheckRecursion(FD);
+    if (pResult) {
```
pow2clk, line 15738: nit: If we're returning nullptr anyway by default, there's no need to check the result of pResult for null before returning it.

pow2clk (commented):
Merp. I concur with the comments that Chris made that I've thumbed. I have some pushback on at least one, I'm not prepared to approve just yet if only because I think Chris had some good points.

Thanks for the responses to my earlier comments!

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: You say it would stand out, but we're actually very inconsistent in this function already. We have `pResult` and `pEntryPointDecl`, but `shaderModel`, `Attr`, `NL`, and `self`. The changed code alone mixes at least 3 different schemes for naming variables.

We should at least make sure that the changed code is consistent with itself.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: I didn't intend for this function to only serve debugging purposes. Rather, I view this function more as an implementation of a promise that the CallGraphWithRecurseGuard class made. Though the function is unused, it implements a feature that the class should implement.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: I decided to use camelCase. Would be cool if clang-format could auto-format variable names, but I recognize that's a hard problem 😐

llvm-beanz: `clang-tidy` does support enforcing variable naming conventions, but it doesn't support applying the changes only to changed regions, so you really have to get large portions of the codebase to comply to use it effectively.

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: Should we call it from an `assert` somewhere to verify the promise?

llvm-beanz (approved):
I had one small comment, but otherwise this looks good.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: I don't see how that verifies the promise. Are you suggesting that it should have some use somewhere? That we shouldn't define a function if it's left unused?

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: @llvm-beanz, can you verify that this satisfies your concern?

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: In general we should not leave unreachable/unused code in the repository. First and foremost because that code (by definition) isn't tested and we shouldn't add untested code.

If the `CallGraphWithRecurseGuard`, and this function verifies the promise, we could call it from an assert in situations where the promise should be true. That would both make this code live, and act as a check on the promise we expect. Also, since the call would be in an `assert` we don't slow down release builds of DXC by running an unnecessary check.

pow2clk (approved):
I think some of Chris's comments still need addressing, but I'm satisfied with what's here

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: Yep. All good.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11458,48 +11499,38 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
-      CG.BuildForEntry(pPatchFnDecl);
-      Decl *pPatchFnDecl = CG.CheckRecursion(pEntryPointDecl);
-      if (pPatchFnDecl) {
-        unsigned id = Diags.getCustomDiagID(
-            clang::DiagnosticsEngine::Level::Error,
-            "recursive functions not allowed (via patch function)");
-        Diags.Report(pPatchFnDecl->getSourceRange().getBegin(), id);
-      }
+  if (pPatchFnDecl) {
+    FunctionDecl *patchResult = ValidateNoRecursion(self, pPatchFnDecl);
```
tex3d, line 11511: There is an additional check we should have for patch constant / hull shader function case.  Neither function should be reachable from the other function.  In other words, the patch constant function must not be in the call graph under the hull shader entry function, and the hull shader entry function must not be in the call graph under the patch constant function.

bob80905: Perfect, I'll add a diagnostic for that and introduce CheckReachability.

pow2clk (approved):
I just noticed that a lot of the end-to-end filecheck tests seem like they could be SemaHLSL verifier tests. Did you consider that?

Issue #5695, opened 2023-09-12
Test ReplaceUsesForLoweredUDT for address space cast user


pow2clk:
**Is your feature request related to a problem? Please describe.**
Nodes introduced a new address space with the common case being having a user-defined type in that address space. As such, some changes went in to improve address space casting. One in particular determined if the address space cast was no longer necessary after UDT lowering introducing `InsertAddrSpaceCastIfRequired`  see these locations:
https://github.com/microsoft/DirectXShaderCompiler/blob/58a9ac3522c899ee52f091de4fd033e12a710afc/lib/HLSL/HLLowerUDT.cpp#L290
https://github.com/microsoft/DirectXShaderCompiler/blob/58a9ac3522c899ee52f091de4fd033e12a710afc/lib/HLSL/HLLowerUDT.cpp#L311

There are no new tests that vet this code.

**Describe the solution you'd like**
We need a test that runs this compiler code for both the case that the address space cast must be preserved and the case where it can be eliminated and ensure that both cases are correctly handled.

**Describe alternatives you've considered**
Unless we can definitively determine why this optimization was added, we might consider removing it.

**Additional context**
See internal PR 4593

PR #5862, opened 2023-10-12
Remove InsertAddrSpaceCastIfRequired which is redundant

python3kgae:
For users of hlsl::ReplaceUsesForLoweredUDT, only EmitGetNodeRecordPtrAndUpdateUsers will has V and NewV with different address space. But there will be no addrspacecast for NodeRecordPtr from clangCodeGen, because NodeRecordPtr doesn't have address space before HLLowerOperation.

As a result, when hit addrspacecast in hlsl::ReplaceUsesForLoweredUDT, the address space of V and NewV should match. So InsertAddrSpaceCastIfRequired will always has AddrSpace != Ty->getPointerAddressSpace().

Removed InsertAddrSpaceCastIfRequired and add assert for address space mismatch.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5695

### DIFF
--- a/lib/HLSL/HLLowerUDT.cpp
+++ b/lib/HLSL/HLLowerUDT.cpp
@@ -285,11 +285,11 @@ void hlsl::ReplaceUsesForLoweredUDT(Value *V, Value *NewV) {
         // Replace bitcast argument with new value
         use.set(NewV);
       }
-
     } else if (ConstantExpr *CE = dyn_cast<ConstantExpr>(user)) {
       // Constant AddrSpaceCast, or BitCast
       if (CE->getOpcode() == Instruction::AddrSpaceCast) {
-        DXASSERT(CE->getType()->getPointerAddressSpace() != NewAddrSpace &&
+        DXASSERT(
+            CE->getType()->getPointerAddressSpace() != NewAddrSpace &&
                 OriginalAddrSpace == NewAddrSpace,
             "When replace Constant, V and NewV must have same address space");
         Constant *NewAC = ConstantExpr::getAddrSpaceCast(
--- a/lib/HLSL/HLLowerUDT.cpp
+++ b/lib/HLSL/HLLowerUDT.cpp
@@ -178,26 +178,6 @@ hlsl::TranslateInitForLoweredUDT(Constant *Init, Type *NewTy,
   return Init;
 }
 
-static Constant *InsertAddrSpaceCastIfRequired(Constant *C,
-                                               unsigned AddrSpace) {
-  Type *Ty = C->getType();
-  DXASSERT_NOMSG(Ty->isPointerTy());
-  if (AddrSpace != Ty->getPointerAddressSpace())
-    return ConstantExpr::getAddrSpaceCast(
-        C, PointerType::get(Ty->getPointerElementType(), AddrSpace));
-  return C;
-}
-
-static Value *InsertAddrSpaceCastIfRequired(IRBuilder<> &Builder, Value *V,
-                                            unsigned AddrSpace) {
-  Type *Ty = V->getType();
-  DXASSERT_NOMSG(Ty->isPointerTy());
-  if (AddrSpace != Ty->getPointerAddressSpace())
-    return Builder.CreateAddrSpaceCast(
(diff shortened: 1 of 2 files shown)


lib/HLSL/HLLowerUDT.cpp
```diff
@@ -285,11 +269,10 @@ void hlsl::ReplaceUsesForLoweredUDT(Value *V, Value *NewV) {
     } else if (AddrSpaceCastInst *AC = dyn_cast<AddrSpaceCastInst>(user)) {
       // Address space cast
       IRBuilder<> Builder(AC);
-      ReplaceUsesForLoweredUDT(
-          user, InsertAddrSpaceCastIfRequired(
-                    Builder, NewV, AC->getType()->getPointerAddressSpace()));
+      Value *NewAC = Builder.CreateAddrSpaceCast(
+          NewV, PointerType::get(Ty, AC->getType()->getPointerAddressSpace()));
+      ReplaceUsesForLoweredUDT(user, NewAC);
```
tex3d, line 274: One point of `InsertAddrSpaceCastIfRequired` was to avoid inserting the cast when *not* required.  I'm still not sure we have the testing to cover the scenario properly.

python3kgae: IRBuilder::CreateAddrSpaceCast will just return the input value if the test matches.

lib/HLSL/HLLowerUDT.cpp
```diff
@@ -302,14 +285,17 @@ void hlsl::ReplaceUsesForLoweredUDT(Value *V, Value *NewV) {
...
       if (CE->getOpcode() == Instruction::AddrSpaceCast) {
-        ReplaceUsesForLoweredUDT(
-            user,
-            InsertAddrSpaceCastIfRequired(
-                cast<Constant>(NewV), CE->getType()->getPointerAddressSpace()));
+        DXASSERT(
+            CE->getType()->getPointerAddressSpace() != NewAddrSpace &&
+                OriginalAddrSpace == NewAddrSpace,
+            "When replace Constant, V and NewV must have same address space");
+        Constant *NewAC = ConstantExpr::getAddrSpaceCast(
```
pow2clk, line 295: I'll admit to being a bit confused by `ConstantExpr::getAddrSpaceCast` but it looks like rather than return the original value when they match, barring some special cases that have nothing to do with type matching, it will return nullptr for `AddrSpaceCast`? 😮 
https://github.com/microsoft/DirectXShaderCompiler/blob/52ace1c57cb782f02ae8ae7fe011bf9c5258a295/lib/IR/ConstantFold.cpp#L723

lib/HLSL/HLLowerUDT.cpp
pow2clk: I think Xiang is referring to this code: 

https://github.com/microsoft/DirectXShaderCompiler/blob/52ace1c57cb782f02ae8ae7fe011bf9c5258a295/include/llvm/IR/IRBuilder.h#L1331

It does look like it will return the right thing when the types fully match.

lib/HLSL/HLLowerUDT.cpp
python3kgae: The DXASSERT about Constant *NewAC = ConstantExpr::getAddrSpaceCast should make sure the types don't match.

tex3d (approved):
LGTM.

lib/HLSL/HLLowerUDT.cpp
tex3d: @pow2clk:
> it will return nullptr for AddrSpaceCast? 😮

The `return nullptr` in `ConstantFoldCastInstruction` you pointed out isn't directly returned by `getAddrSpaceCast`, and `getAddrSpaceCast` only returns `nullptr` if you pass `true` in optional `OnlyIfReduced` argument.

https://github.com/microsoft/DirectXShaderCompiler/blob/52ace1c57cb782f02ae8ae7fe011bf9c5258a295/lib/IR/Constants.cpp#L1577-L1581

pow2clk (approved):
Thanks for the clarifications Tex and Xiang

PR #5873, opened 2023-10-16
[Docs] Update contributing doc for clang-format

llvm-beanz:
Specify the clang-format requirement in the contributing documentation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,6 +44,8 @@ Here's a few things you should always do when making changes to the code base:
 
 The coding, style, and general engineering guidelines follow those described in the docs/CodingStandards.rst. For additional guidelines in code specific to HLSL, see the docs/HLSLChanges.rst file.
 
+DXC has adopted a clang-format requirement for all incoming changes. PRs to DXC should have the _changed code_ clang formatted to the LLVM style, and leave the remaining portions of the file unchanged. This can be done using the `git-clang-format` tool or IDE driven workflows. A GitHub action will run on all PRs to validate that the change is properly formatted.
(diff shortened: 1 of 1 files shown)


bob80905 (approved):

damyanp (approved):

dmpots (approved):

CONTRIBUTING.md
```diff
@@ -44,6 +44,8 @@ Here's a few things you should always do when making changes to the code base:
 
 The coding, style, and general engineering guidelines follow those described in the docs/CodingStandards.rst. For additional guidelines in code specific to HLSL, see the docs/HLSLChanges.rst file.
 
+DXC has adopted a clang-format requirement for all incoming changes. PRs to DXC should have the _changed code_ clang formatted to the LLVM style, and leave the remaining portions of the file unchanged. This can be done using the `git-clang-format` tool or IDE driven workflows. A GitHub action will run on all PRs to validate that the change is properly formatted.
```
pow2clk, line 47: It might bear mentioning that if the bot finds you violated this, your change won't be merged. 

It is too much for this change, but the CodingStandards.rst mentioned above are not really consistent with the current state of affairs in terms of what C++ features are allowed, required doxygen comments, use of exceptions etc. We might consider just eliminating it rather than trying to update it and say something like C++17, use comments please, do what clang-format tells you. It has a certain simplicity to it.

llvm-beanz: I strongly prefer that we just start enforcing those coding standards. We could update them to include C++17 being allowed, but there's a lot of really good stuff in there.

PR #5847, opened 2023-10-10
[ClangCL] Cleanup rest clang-cl warnings

python3kgae:
1. printf format mismatch.
2. avoid cast from CComPtr<IncludeHandlerVFSOverlayForTest> to CComPtr<IDxcIncludeHandler>.
3. fix signed unsigned mismatch.
4. fix order of fields in constructor.
5. add override for override methods.
6. port https://github.com/llvm/llvm-project/commit/01f4209631f32ccec686b232039238066a0866d9
7. add copy assignment operator to avoid -Wdeprecated-copy-with-user-provided-copy
8. disable -Wignored-qualifiers for taef header WexTestClass.h.
9. remove unused fields.
10. add -clang-cl for hctbuild.

PR #5875, opened 2023-10-16
ATOProgOffset test fix

hekota:
The test was broken by commit 1dac9d8f518b429c3b9f95aad9fcb58d1e0d4591 that was eliminating clang compiler warnings in HLSL execution tests, specifically casting signed shift to unsigned.

In this particular case the signed shift was intentional. This PR is changing the unsigned shift to signed and disabling warning for this operation for clang.


hekota:
This has just been fixed in #5847.

PR #5880, opened 2023-10-17
Fix module loading and validation issues

### LINKED ISSUE #5879: DxilValidation fails dx.op.checkAccessFullyMapped use when declared before use.
**Description**
If the declaration of `dx.op.checkAccessFullyMapped` appears before the function that uses it, an incorrect validation error will occur.

This happens due to the way DxilOperations tries to look up functions for dxil ops by name before constructing the types used, which would fill in the type cache for existing DXIL, which is used by `DxilValidation` to detect whether the type `dx.op.checkAccessFullyMapped` is extracting a value from is a valid resource type.

However, if the types are constructed during RefreshCache as they should be, another problem is revealed: RefreshCache occurs before DxilOperations knows what low precision mode is set.

Therefore, two issues need to be fixed simultaneously like so:
1. `GetOpFunc` needs to construct types before checking for a matching function by name, and verify that any function looked up by name has the expected type.
2. `RefreshCache` (and `FixOverloadNames`) need to be called after we know what low precision mode is set (such as when we set that mode).

**Steps to Reproduce**
```llvm
; RUN: %dxv %s
target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
target triple = "dxil-ms-dx"

%dx.types.Handle = type { i8* }
%dx.types.CBufRet.i16 = type { i16, i16, i16, i16 }
%dx.types.ResRet.f32 = type { float, float, float, float, i32 }
%"hostlayout.class.StructuredBuffer<MyStruct>" = type { %hostlayout.struct.MyStruct }
%hostlayout.struct.MyStruct = type { float }
%"hostlayout.$Globals" = type { <4 x i32>, <4 x i32> }

; Function Attrs: nounwind readonly
declare i1 @dx.op.checkAccessFullyMapped.i32(i32, i32) #1

define void @main() {
  %1 = call %dx.types.Handle @dx.op.createHandle(i32 57, i8 0, i32 0, i32 0, i1 false)  ; CreateHandle(resourceClass,rangeId,index,nonUniformIndex)
  %2 = call %dx.types.Handle @dx.op.createHandle(i32 57, i8 2, i32 0, i32 0, i1 false)  ; CreateHandle(resourceClass,rangeId,index,nonUniformIndex)
  %3 = call %dx.types.CBufRet.i16 @dx.op.cbufferLoadLegacy.i16(i32 59, %dx.types.Handle %2, i32 1)  ; CBufferLoadLegacy(handle,regIndex)
  %4 = extractvalue %dx.types.CBufRet.i16 %3, 1
  %5 = sext i16 %4 to i32
  %6 = call %dx.types.ResRet.f32 @dx.op.rawBufferLoad.f32(i32 139, %dx.types.Handle %1, i32 %5, i32 0, i8 1, i32 4)  ; RawBufferLoad(srv,index,elementOffset,mask,alignment)
  %7 = extractvalue %dx.types.ResRet.f32 %6, 0
  %8 = fptrunc float %7 to half
  %9 = extractvalue %dx.types.ResRet.f32 %6, 4
  %10 = call i1 @dx.op.checkAccessFullyMapped.i32(i32 71, i32 %9)  ; CheckAccessFullyMapped(status)
  %11 = select i1 %10, half %8, half 0xHBC00
  call void @dx.op.storeOutput.f16(i32 5, i32 0, i32 0, i8 0, half %11)  ; StoreOutput(outputSigId,rowIndex,colIndex,value)
  ret void
}

; Function Attrs: nounwind
declare void @dx.op.storeOutput.f16(i32, i32, i32, i8, half) #0

; Function Attrs: nounwind readonly
declare %dx.types.CBufRet.i16 @dx.op.cbufferLoadLegacy.i16(i32, %dx.types.Handle, i32) #1

; Function Attrs: nounwind readonly
declare %dx.types.ResRet.f32 @dx.op.rawBufferLoad.f32(i32, %dx.types.Handle, i32, i32, i8, i32) #1

; Function Attrs: nounwind readonly
declare %dx.types.Handle @dx.op.createHandle(i32, i8, i32, i32, i1) #1

attributes #0 = { nounwind }
attributes #1 = { nounwind readonly }

!llvm.ident = !{!0}
!dx.version = !{!1}
!dx.valver = !{!2}
!dx.shaderModel = !{!3}
!dx.resources = !{!4}
!dx.viewIdState = !{!10}
!dx.entryPoints = !{!11}

!0 = !{!"dxc(private) 1.7.0.4154 (dxil-op-cache-init, c8642b4f6-dirty)"}
!1 = !{i32 1, i32 3}
!2 = !{i32 1, i32 7}
!3 = !{!"vs", i32 6, i32 3}
!4 = !{!5, null, !8, null}
!5 = !{!6}
!6 = !{i32 0, %"hostlayout.class.StructuredBuffer<MyStruct>"* undef, !"", i32 0, i32 0, i32 1, i32 12, i32 0, !7}
!7 = !{i32 1, i32 4}
!8 = !{!9}
!9 = !{i32 0, %"hostlayout.$Globals"* undef, !"", i32 0, i32 0, i32 1, i32 32, null}
!10 = !{[2 x i32] [i32 0, i32 1]}
!11 = !{void ()* @main, !"main", !12, !4, !17}
!12 = !{null, !13, null}
!13 = !{!14}
!14 = !{i32 0, !"OUT", i8 8, i8 0, !15, i8 2, i32 1, i8 1, i32 0, i8 0, !16}
!15 = !{i32 0}
!16 = !{i32 3, i32 1}
!17 = !{i32 0, i64 4144}
```

**Actual Behavior**
```
Function: ?main@@YA$min16f@XZ: error: CheckAccessFullyMapped should only be used on resource status.
note: at '%12 = call i1 @dx.op.checkAccessFullyMapped.i32(i32 71, i32 %11)' in block '#0' of function '?main@@YA$min16f@XZ'.
Validation failed.
```

**Environment**
- DXC version: DXIL.dll 1.7 release, as well as latest main and staging-sm-6.8 branches.
- Host Operating System: any Windows

tex3d:
Fix module loading and validation issues

There are a couple interacting issues:
1. When RefreshCache occurs, it doesn't fill in the m_pResRetType cache.
  - This is because in GetOpFunc, if it finds a function matching the
    expected name, it simply returns that, skipping function type
    construction.
  - This causes a validation error if CheckAccessFullyMapped is validated
    before a function that calls it.  This depends on the order of
    function declarations/definitions in the module.
  - The solution is to move the check after type construction, and verify
    that the function type matches expectations as well.
2. When RefreshCache() and FixOverloadNames() was called from DxilOperations
    constructor, the m_LowPrecisionMode has not yet been set, leading to
    incorrect overload types when allowing function type construction to
    proceed in GetOpFunc before looking for a matching function by name.
  - This leads to an incorrect type for 16-bit CBufRet type being used.
  - The solution is to move RefreshCache() and FixOverloadNames() calls
    to after we actually know the m_LowPrecisionMode, so that the correct
    types may be constructed.

Rename hlsl::OP::SetMinPrecision to InitWithMinPrecision to make clear what's being done.
Added validation tests to verify the fixes.
Fixed validation test that produced invalid DXIL op function signature.
Fix HLModule not setting min-precision mode when loading from metadata.

Fixes #5879

### DIFF
--- a/lib/DXIL/DxilModule.cpp
+++ b/lib/DXIL/DxilModule.cpp
@@ -123,7 +123,7 @@ void DxilModule::SetShaderModel(const ShaderModel *pSM, bool bUseMinPrecision) {
   m_pSM->GetDxilVersion(m_DxilMajor, m_DxilMinor);
   m_pMDHelper->SetShaderModel(m_pSM);
   m_bUseMinPrecision = bUseMinPrecision;
-  m_pOP->SetMinPrecision(m_bUseMinPrecision);
+  m_pOP->InitWithMinPrecision(m_bUseMinPrecision);
   m_pTypeSystem->SetMinPrecision(m_bUseMinPrecision);
 
   if (!m_pSM->IsLib()) {
--- a/lib/DXIL/DxilOperations.cpp
+++ b/lib/DXIL/DxilOperations.cpp
@@ -2980,17 +2980,6 @@ OP::OP(LLVMContext &Ctx, Module *pModule)
                            Type::getInt16Ty(m_Ctx)}; // HiHi, HiLo, LoHi, LoLo
   m_pFourI16Type =
       GetOrCreateStructType(m_Ctx, FourI16Types, "dx.types.fouri16", pModule);
-
-  // When loading a module into an existing context where types are merged,
-  // type names may change.  When this happens, any intrinsics overloaded on
-  // UDT types will no longer have matching overload names.
-  // This causes RefreshCache() to assert.
-  // This fixes the function names to they match the expected types,
-  // preventing RefreshCache() from failing due to this issue.
-  FixOverloadNames();
-
-  // Try to find existing intrinsic function.
-  RefreshCache();
 }
 
 void OP::RefreshCache() {
@@ -3078,16 +3067,6 @@ Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
   Type *obj = pOverloadType;
   Type *resProperty = GetResourcePropertiesType();
   Type *resBind = GetResourceBindingType();
-
-  std::string funcName;
-  ConstructOverloadName(pOverloadType, opCode, funcName);
-
-  // Try to find exist function with the same name in the module.
-  if (Function *existF = m_pModule->getFunction(funcName)) {
-    F = existF;
-    UpdateCache(opClass, pOverloadType, F);
-    return F;
-  }
 
 #define A(_x) ArgTypes.emplace_back(_x)
 #define RRT(_y) A(GetResRetType(_y))
@@ -4737,6 +4716,21 @@ Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
   pFT = FunctionType::get(
       ArgTypes[0], ArrayRef<Type *>(&ArgTypes[1], ArgTypes.size() - 1), false);
 
+  std::string funcName;
+  ConstructOverloadName(pOverloadType, opCode, funcName);
+
+  // Try to find existing function with the same name in the module.
+  // This needs to happen after the switch statement that constructs arguments
+  // and return values to ensure that ResRetType is constructed in the
+  // RefreshCache case.
+  if (Function *existF = m_pModule->getFunction(funcName)) {
+    DXASSERT(existF->getFunctionType() == pFT,
+             "existing function must have the expected function type");
+    F = existF;
+    UpdateCache(opClass, pOverloadType, F);
+    return F;
+  }
+
   F = cast<Function>(m_pModule->getOrInsertFunction(funcName, pFT));
 
   UpdateCache(opClass, pOverloadType, F);
@@ -4810,7 +4804,25 @@ void OP::SetMinPrecision(bool bMinPrecision) {
             m_LowPrecisionMode == DXIL::LowPrecisionMode::Undefined),
            "LowPrecisionMode should only be set once.");
 
-  m_LowPrecisionMode = mode;
+  if (mode != m_LowPrecisionMode) {
+    m_LowPrecisionMode = mode;
+
+    // The following FixOverloadNames() and RefreshCache() calls interact with
+    // the type cache, which can only be correctly constructed once we know
+    // the min precision mode.  That's why they are called here, rather than
+    // in the constructor.
+
+    // When loading a module into an existing context where types are merged,
+    // type names may change.  When this happens, any intrinsics overloaded on
+    // UDT types will no longer have matching overload names.
+    // This causes RefreshCache() to assert.
+    // This fixes the function names to they match the expected types,
+    // preventing RefreshCache() from failing due to this issue.
+    FixOverloadNames();
+
+    // Try to find existing intrinsic function.
+    RefreshCache();
+  }
 }
 
 uint64_t OP::GetAllocSizeForType(llvm::Type *Ty) {
@@ -5067,6 +5079,8 @@ Type *OP::GetCBufferRetType(Type *pOverloadType) {
   unsigned TypeSlot = GetTypeSlot(pOverloadType);
 
   if (m_pCBufferRetType[TypeSlot] == nullptr) {
+    DXASSERT(m_LowPrecisionMode != DXIL::LowPrecisionMode::Undefined,
+             "m_LowPrecisionMode must be set before constructing type.");
     string TypeName("dx.types.CBufRet.");
     TypeName += GetOverloadTypeName(TypeSlot);
     Type *i64Ty = Type::getInt64Ty(pOverloadType->getContext());
--- a/lib/DXIL/DxilOperations.cpp
+++ b/lib/DXIL/DxilOperations.cpp
@@ -4796,7 +4796,7 @@ bool OP::UseMinPrecision() {
   return m_LowPrecisionMode == DXIL::LowPrecisionMode::UseMinPrecision;
 }
 
-void OP::SetMinPrecision(bool bMinPrecision) {
+void OP::InitWithMinPrecision(bool bMinPrecision) {
   DXIL::LowPrecisionMode mode =
       bMinPrecision ? DXIL::LowPrecisionMode::UseMinPrecision
                     : DXIL::LowPrecisionMode::UseNativeLowPrecision;
--- a/lib/HLSL/HLModule.cpp
+++ b/lib/HLSL/HLModule.cpp
@@ -512,7 +512,7 @@ void HLModule::LoadHLMetadata() {
           DxilMDHelper::ConstMDToUint32(options->getOperand(1)->getOperand(0)));
   }
 
-  m_pOP->SetMinPrecision(m_Options.bUseMinPrecision);
+  m_pOP->InitWithMinPrecision(m_Options.bUseMinPrecision);
 
   m_pMDHelper->LoadRootSignature(m_SerializedRootSignature);
 
--- a/lib/HLSL/HLModule.cpp
+++ b/lib/HLSL/HLModule.cpp
@@ -512,6 +512,8 @@ void HLModule::LoadHLMetadata() {
           DxilMDHelper::ConstMDToUint32(options->getOperand(1)->getOperand(0)));
   }
 
+  m_pOP->SetMinPrecision(m_Options.bUseMinPrecision);
+
   m_pMDHelper->LoadRootSignature(m_SerializedRootSignature);
 
   // Load Subobjects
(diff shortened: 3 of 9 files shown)


tex3d:
I don't intend to reformat the whole python script for the 1-character change inside the string.  If we want to reformat these scripts to pass darker, we should do so in a separate PR.

lib/Transforms/Scalar/DxilRemoveDeadBlocks.cpp
```diff
@@ -346,8 +346,7 @@ static void EnsureDxilModule(Module *M) {
     return;
   for (Function &F : *M) {
     if (OP::IsDxilOpFunc(&F)) {
-      bool bSkipInit = true; // Metadata is not necessarily valid yet.
-      M->GetOrCreateDxilModule(bSkipInit);
+      M->GetOrCreateDxilModule();
```
python3kgae, line 349: Is metadata ready at DxilRemoveDeadBlocks?

tex3d: Either we have a DxilModule that doesn't need to be initialized (since we are compiling normally from the start), or we need to construct the DxilModule and load state from metadata because we deserialized something into an llvm module.

In the former case, we don't have metadata, but we do have the necessary state in DxilModule that would be used to emit the metadata.  The DxilModule state is what's needed, not the metadata itself.  Since we already have a DxilModule, that just gets returned, so the bSkipInit param is not used.

In the latter case, we should have metadata capturing the important properties of the module, otherwise we forgot to capture the key state from DxilModule by emitting the metadata before serialization.  If we don't have the metadata, we lost information that answers important questions, such as whether this module is min-precision mode or exact-low-precision mode.

Originally, this was written with `bSkipInit=true` because of the `delete_constant_dce.ll` IR test which was generated by dumping a module without emitting metadata first.  But that's not the right way to produce IR for a pass test, since it loses important information contained in DxilModule needed to interpret the IR.  This is why I wrote `ExtractIRForPassTest.py` to make it easy to extract IR with metadata in the state before a particular pass to turn into a test (then I checked that in separately).

**TL;DR: We do not need bSkipInit, unless we are already in an invalid state.**

dmpots:
It seems like there are three independent changes in this PR.

1. The refreshcache fix
2. The DxilRemoveDeadBlocks fix
3. The python script fix

Could we separate them out to make it easy to review them all?

pow2clk (commented):
It feels kind of shaky to refresh the cache exclusively when minprecision is set. 

I wonder if it's possible to demand that minprecision be set when the OP class is first created? I expect that will likely touch more code, but I'd feel it was safer and more intuitive.

I agree with David that this feels like three different changes.

pow2clk:
BTW, I agree that reformatting the python script isn't called for here, but we should eithe disable the test, or have a separate change that formats them "properly"

lib/DXIL/DxilOperations.cpp
```diff
@@ -4810,7 +4804,25 @@ void OP::SetMinPrecision(bool bMinPrecision) {
...
+    // When loading a module into an existing context where types are merged,
+    // type names may change.  When this happens, any intrinsics overloaded on
+    // UDT types will no longer have matching overload names.
+    // This causes RefreshCache() to assert.
+    // This fixes the function names to they match the expected types,
+    // preventing RefreshCache() from failing due to this issue.
+    FixOverloadNames();
+
+    // Try to find existing intrinsic function.
+    RefreshCache();
```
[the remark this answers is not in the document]
tex3d: What if the method was `Init()` which took the min precision mode as a required argument?  Then would it feel less weird?

lib/HLSL/HLModule.cpp
```diff
@@ -512,6 +512,8 @@ void HLModule::LoadHLMetadata() {
           DxilMDHelper::ConstMDToUint32(options->getOperand(1)->getOperand(0)));
   }
 
+  m_pOP->SetMinPrecision(m_Options.bUseMinPrecision);
+
```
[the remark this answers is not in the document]
tex3d: Because minprecision isn't set.  But refreshcache would also need to be called after initializing the DxilOperation class with the correct low-precision mode.

Before, we weren't initializing HLModule from serialized intermediate state properly because we didn't set the min precision mode, but our testing wasn't sufficient to catch this.  Now, one of our tests hits a new assert, when this isn't initialized here.  I added the assert in `GetCBufferRetType` [here](https://github.com/microsoft/DirectXShaderCompiler/pull/5880/files#diff-e85a54111eb6509ee6bf740224386600db9c32fa2fc38d9b95cbb202f0882800R5082) to catch when we try to construct a type, but don't know the min precision mode, when it would make a difference.

lib/DXIL/DxilOperations.cpp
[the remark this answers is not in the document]
tex3d: I've renamed the method to `InitWithMinPrecision`, and added more comments in this area.

lib/DXIL/DxilModule.cpp
```diff
@@ -123,7 +123,7 @@ void DxilModule::SetShaderModel(const ShaderModel *pSM, bool bUseMinPrecision) {
   m_pSM->GetDxilVersion(m_DxilMajor, m_DxilMinor);
   m_pMDHelper->SetShaderModel(m_pSM);
   m_bUseMinPrecision = bUseMinPrecision;
-  m_pOP->SetMinPrecision(m_bUseMinPrecision);
+  m_pOP->InitWithMinPrecision(m_bUseMinPrecision);
   m_pTypeSystem->SetMinPrecision(m_bUseMinPrecision);
```
dmpots, line 127: You changed the `SetMinPrecision` call to `InitiWithMinPrecision` for the operations but not the type system. Should they not be consistent?

tex3d:
> It seems like there are three independent changes in this PR.
> 
> 1. The refreshcache fix
> 2. The DxilRemoveDeadBlocks fix
> 3. The python script fix
> 
> Could we separate them out to make it easy to review them all?

This is now done, and this is the last remaining part to merge (the refreshcache fix).

tools/clang/test/DXILValidation/val-dx-type-lowprec.ll
```diff
@@ -0,0 +1,145 @@
+; RUN: %dxv %s
+
```
dmpots, line 2: It would be nice to add a comment to the test stating its purpose since it is not checking for any expected output, but just making sure it validates successfully it makes it hard to deduce what exactly is being tested here.

dmpots: Oh it looks like there is a comment in the hlsl. Maybe just a pointer to that file would be helpful? Not sure if this is a common pattern in the tests in this directory that would make it obvious to go look at the hlsl.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -943,12 +946,12 @@ TEST_F(ValidationTest, QuadOpInVS) {
       "struct PerThreadData { int "
       "input; int output; }; RWStructuredBuffer<PerThreadData> g_sb; "
       "void main(uint vid : SV_VertexID)"
-      "{ g_sb[vid].output = WaveActiveSum(g_sb[vid].input); }",
+      "{ g_sb[vid].output = WaveActiveBitAnd((uint)g_sb[vid].input); }",
```
dmpots, line 949: Why was this change needed?

dmpots (approved):

lib/DXIL/DxilModule.cpp
tex3d: Not necessarily.  I considered it, but looking into it, it doesn't really have the same justification.  There isn't initialization going on based on that in the type system, and it would have expanded the scope of changes, at which point, it looked like some other changes could be made to clean some code up that used these, but I didn't want more changes creeping into this PR.

tools/clang/test/DXILValidation/val-dx-type-lowprec.ll
tex3d: I'll comment pointing to the HLSL.

tools/clang/unittests/HLSL/ValidationTest.cpp
tex3d: The IR modification resulted in an invalid signature for the target operation, which would now be caught, causing assert and preventing this test from working.  The operation was invalid because it had the wrong number of operands.  I chose a different intrinsic so the number of operands would be the same, and IR modification would result in valid IR.

python3kgae (approved):

pow2clk (approved):
I like the new name better. Good catch with the ValidationTest change

PR #5881, opened 2023-10-17
Make hctbuild cmake gen consistent with asserts

pow2clk:
A slightly complex series of events causes invocations of hctbuild after the first to enable asserts in release builds. This sets the build type even when it isn't needed to avoid it.

Because the build type was not being set when generating for visual studio in hctbuild, when generating the cmake files using CMakeLists, it would produce an alarming message saying it was defaulting to debug, but, in fact, since it didn't care for the purposes of sln file generation, that wasn't actually a problem, just confusing. It did set a variable to enable asserts, but didn't do this for release variant builds

The real problem occurs if hctbuild is invoked again, at that point, instead of defaulting to Debug, it leaves the build type empty because of another variable set in the cache.

Additionally, since the first invocation enabled asserts, that also comes from the cache and so code that strips NDEBUG from all build variants is invoked because the build type is, in fact, not debug, it's an empty string.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 # See docs/CMake.html for instructions about how to build LLVM with CMake.
 cmake_minimum_required(VERSION 3.17.2) # HLSL Change - Require CMake 3.17.2.
 
+message(CMAKE_BUILD_TYPE0!!!!!="${CMAKE_BUILD_TYPE}")
+message(CMAKE_CONFIGURATION_TYPES!!!!!="${CMAKE_CONFIGURATION_TYPES}")
+message(CMAKE_CXX_FLAGS_RELEASE!!!!!="${CMAKE_CXX_FLAGS_RELEASE}")
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
@@ -8,10 +11,12 @@ if (NOT "${DXC_CMAKE_BEGINS_INCLUDE}" STREQUAL "")
   include(${DXC_CMAKE_BEGINS_INCLUDE})
 endif()
 
+
 if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   message(STATUS "No build type selected, default to Debug")
   set(CMAKE_BUILD_TYPE "Debug")
 endif()
+message(CMAKE_BUILD_TYPE1!!!!!="${CMAKE_BUILD_TYPE}")
 
 if(POLICY CMP0022)
   cmake_policy(SET CMP0022 NEW) # automatic when 2.8.12 is required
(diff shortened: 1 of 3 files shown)


utils/hct/hctbuild.cmd
```diff
@@ -465,10 +465,8 @@ if "%DO_SETUP%"=="1" (
     echo Running "%CMAKE_PATH%" -DCMAKE_BUILD_TYPE:STRING=%1 %CMAKE_OPTS% -G %4 %HLSL_SRC_DIR% > %3\cmake-log.txt
     "%CMAKE_PATH%" -DCMAKE_BUILD_TYPE:STRING=%1 %CMAKE_OPTS% -G %4 %HLSL_SRC_DIR% >> %3\cmake-log.txt 2>&1
   ) else (
-    rem -DCMAKE_BUILD_TYPE:STRING=%1 is not necessary for multi-config generators like VS
-    rem but need CMAKE_BUILD_TYPE to generate lit cfg.
     echo Running "%CMAKE_PATH%" -DCMAKE_BUILD_TYPE:STRING=%1  %CMAKE_OPTS% -G %4 %5 %HLSL_SRC_DIR% > %3\cmake-log.txt
-    "%CMAKE_PATH%" %CMAKE_OPTS% -G %4 %5 %HLSL_SRC_DIR% >> %3\cmake-log.txt 2>&1
```
python3kgae, line 471: Maybe add comment about why we need CMAKE_BUILD_TYPE for multi-config generators like VS?

python3kgae (approved):

bob80905 (approved):

utils/hct/hctbuild.cmd
pow2clk: Might be a long comment! I tried to be succinct here and failed. I feel like the explanation for why it wasn't needed was more comment worthy than the fact that we do. Maybe I could just say that it will still generate multiple versions in spite of that.

damyanp (approved):

utils/hct/hctbuild.cmd
pow2clk: "BUILD_TYPE is mostly ignored in this path as VS generates multiple targets. It is still needed to satisfy cmake file expectations"

I feel pretty good about that. What do you think?

pow2clk: FWIW, the latest version of LLVM CMakeLists fails if you don't set this flag, which avoids the debug by default that was the first step in this walking catastrophe

python3kgae: Looks good.
I guess upstream hit the same issue :)

CMakeLists.txt
```diff
@@ -1,17 +1,22 @@
 # See docs/CMake.html for instructions about how to build LLVM with CMake.
 cmake_minimum_required(VERSION 3.17.2) # HLSL Change - Require CMake 3.17.2.
 
+message(CMAKE_BUILD_TYPE0!!!!!="${CMAKE_BUILD_TYPE}")
+message(CMAKE_CONFIGURATION_TYPES!!!!!="${CMAKE_CONFIGURATION_TYPES}")
+message(CMAKE_CXX_FLAGS_RELEASE!!!!!="${CMAKE_CXX_FLAGS_RELEASE}")
 set(CMAKE_CXX_STANDARD 17)
```
hekota, line 7: Did you mean to leave all these messages in?

Issue #5870, opened 2023-10-16
[Test] Move Verifier test as lit test


python3kgae:
**Description**
Move verifier test as lit test.
This will help to check test not enabled like template-diag-deferred.hlsl.
And these should be shaders expected to be failed to compile, so should not affect coverage of hash stability test.

**Work Plan**
Search all shaders used in tools/clang/unittests/HLSL/VerifierTest.cpp
Search all shaders with expected-note/error/warning.
Move all these shaders into SemaHLSL and update RUN line to lit format.

PR #5885, opened 2023-10-17
[Option] add verify_ignore_unexpected_EQ for verifier test

python3kgae:
Add support -verify-ignore-unexpected= to DxcOpts.
DiagnosticLevelMask is copied to HLSLOptions.h to avoid use clang header in DxcOption.

Also change -verify to Flag to allow adding options after it.

This is for https://github.com/microsoft/DirectXShaderCompiler/issues/5870

### DIFF
--- a/tools/clang/test/SemaHLSL/array-index-out-of-bounds-HV-2016.hlsl
+++ b/tools/clang/test/SemaHLSL/array-index-out-of-bounds-HV-2016.hlsl
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -Wno-unused-value -fsyntax-only -ffreestanding -HV 2016 -verify -verify-ignore-unexpected=note %s
+// RUN: %dxc -Tlib_6_3 -Wno-unused-value -HV 2016 -verify -verify-ignore-unexpected=note %s
 
 void dead()
 {
--- a/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
+++ b/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
@@ -1337,9 +1337,18 @@ public:
     // Make sure clang::DiagnosticLevelMask and
     // hlsl::options::DiagnosticLevelMask match.
     static_assert(
-        std::is_same_v<
-            std::underlying_type_t<clang::DiagnosticLevelMask>,
-            std::underlying_type_t<hlsl::options::DiagnosticLevelMask>>,
+        static_cast<int>(clang::DiagnosticLevelMask::None) ==
+                static_cast<int>(hlsl::options::DiagnosticLevelMask::None) &&
+            static_cast<int>(clang::DiagnosticLevelMask::Note) ==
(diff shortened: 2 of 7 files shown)


tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1334,6 +1334,8 @@ class DxcCompiler : public IDxcCompiler3,
     compiler.getDiagnosticOpts().ShowOptionNames = Opts.ShowOptionNames ? 1 : 0;
     compiler.getDiagnosticOpts().Warnings = std::move(Opts.Warnings);
     compiler.getDiagnosticOpts().VerifyDiagnostics = Opts.VerifyDiagnostics;
+    compiler.getDiagnosticOpts().setVerifyIgnoreUnexpected(
+        static_cast<clang::DiagnosticLevelMask>(Opts.DiagMask));
```
dmpots, line 1338: I don't think we should be using a static cast without guaranteeing somehow that the values are the same. So either we should add `static_assert` to verify the enum values match or add a function to translate between the enums with a switch statement.

tools/clang/test/SemaHLSL/array-index-out-of-bounds-HV-2016.hlsl
```diff
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -Wno-unused-value -fsyntax-only -ffreestanding -HV 2016 -verify -verify-ignore-unexpected=note %s
+// RUN: %dxc -Tlib_6_3 -Wno-unused-value -HV 2016 -verify -verify-ignore-unexpected=note %s
```
dmpots, line 1: This test was already using `-verify-ignore-unexpected=note`. What was that doing before this change?

python3kgae: It doesn't go thru dxc before.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
python3kgae: Added static_assert.

dmpots: Does this actually check that the enum values match? Or is this just checking they have the same underlying type?

python3kgae: Updated to check enum values.

dmpots: That looks good, thanks. I think it kind of interrupts the flow a bit to have it inline with where we set the value. Might want to move it outside the function.

dmpots (approved):

llvm-beanz (changes requested):
why is this needed?

python3kgae:
Switched to remove verify_ignored_unexpected in https://github.com/microsoft/DirectXShaderCompiler/pull/5902

PR #5887, opened 2023-10-18
Enable `.ll` test suffix for LLVM tests

llvm-beanz:
This enables `.ll` as a default test suffix in the LLVM tests. To preserve the low number of unsupported tests, this changes the way we disable tests from using config.unsupported to setting the suffixes to empty.


python3kgae (approved):

dmpots (approved):
LGTM!

PR #5889, opened 2023-10-18
Make DxilSignatureElement::m_SemanticName canonical for system values

tex3d:
When DxilModule metadata is serialized, semantic strings for system values are canonicalized.  If these strings are used anywhere, this can lead to a difference in behavior between an original DxilModule that has not undergone serialization and one that has been deserialized.

This change canonicalizes the string as soon as a DxilSignatureElement is created, thereby removing the possibility of accidentally relying on the non-canonical string for a system value.

### DIFF
--- a/tools/clang/unittests/HLSL/DxilModuleTest.cpp
+++ b/tools/clang/unittests/HLSL/DxilModuleTest.cpp
@@ -612,7 +612,9 @@ TEST_F(DxilModuleTest, CanonicalSystemValueSemantic) {
   // makes sure the string gets canonicalized when the signature elsement is
   // created.
 
-  std::unique_ptr<hlsl::DxilSignatureElement> newElt = std::make_unique<hlsl::DxilSignatureElement>(hlsl::DXIL::SigPointKind::VSOut);
+  std::unique_ptr<hlsl::DxilSignatureElement> newElt =
+      std::make_unique<hlsl::DxilSignatureElement>(
+          hlsl::DXIL::SigPointKind::VSOut);
   newElt->Initialize("sV_pOsItIoN", hlsl::CompType::getF32(),
                      hlsl::InterpolationMode(
                          hlsl::DXIL::InterpolationMode::LinearNoperspective),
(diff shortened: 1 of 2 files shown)


hekota (approved):
LGTM!

tools/clang/unittests/HLSL/DxilModuleTest.cpp
```diff
@@ -598,4 +600,22 @@ TEST_F(DxilModuleTest, PayloadQualifier) {
...
+
+  // This is difficult to do from a FileCheck test because system value
+  // semantics get canonicallized during the metadata emit.  However, having a
+  // non-canonical semantic name at earlier stages can lead to inconsistency
+  // between the strings used earlier and the final canonicalized version.  This
+  // makes sure the string gets canonicalized when the signature elsement is
+  // created.
+
+  std::unique_ptr<hlsl::DxilSignatureElement> newElt = std::make_unique<hlsl::DxilSignatureElement>(hlsl::DXIL::SigPointKind::VSOut);
+  newElt->Initialize("sV_pOsItIoN", hlsl::CompType::getF32(),
```
python3kgae, line 616: Do we have to support this?
It is really ugly.

tex3d: I'm not sure what you are referring to by "this".

Semantics are case-insensitive, but system values get canonicalized when written to metadata.  This just makes sure the system value is canonicalized when DxilSignatureElement is created, instead of waiting for the write to metadata.  That way any use of the semantic string between this point and the serialized form (specifically when writing the semantic string to RDAT) will get the same value as after serialization/deserialization.

python3kgae: Sorry for not making things clearer.
I mean "sV_pOsItIoN" is ugly here.

tex3d: It's supposed to be case-insensitive.  It doesn't really matter what's here, as long as it's `"SV_Position"` with a different casing for any of the characters.  If it used `"SV_POSITION"` or `"sv_position"`, a different implementation might not properly do a case-insensitive match, instead checking for these more common possibilities only.

PR #5893, opened 2023-10-18
Remove ExtractRecordStructFromArray intrinsic

### LINKED ISSUE #5351: [SM 6.8] Eliminate hidden intrinsics: ImplicitRecordToStructCast and ExtractRecordStructFromArray
These hidden intrinsics (gen_intrin_main.txt):
$match<0,-2> void [[hidden]] ImplicitRecordToStructCast(in AnyNodeOutputRecord output);
$match<0,-2> void [[hidden]] ExtractRecordStructFromArray(in AnyNodeOutputRecord output);

Should be unnecessary, they are a vestige from an earlier implementation direction.  We should eliminate them and change the code that depends on them.

bob80905:
The ExtractRecordStructFromArray intrinsic is left over dead code that should be removed. This PR removes the intrinsic from existence. In the cases where this intrinsic was used, DefaultSubscript is used instead. 
Fixes #5351


lib/HLSL/HLOperationLower.cpp
```diff
@@ -9465,25 +9463,26 @@ void EmitGetNodeRecordPtrAndUpdateUsers(HLOperationLowerHelper &helper,
...
-  // Translate node record type here
-  auto findIt = helper.loweredTypes.find(origRecordUDT);
-  if (findIt != helper.loweredTypes.end()) {
-    getNodeRecordPtrRT = findIt->second;
-  } else {
-    getNodeRecordPtrRT = GetLoweredUDT(origRecordUDT, &helper.dxilTypeSys);
-    if (origRecordUDT != getNodeRecordPtrRT)
-      helper.loweredTypes[origRecordUDT] = getNodeRecordPtrRT;
+  PointerType *pTy = cast<PointerType>(CI->getType());
+  if (StructType *origRecordUDT = dyn_cast<StructType>(pTy->getElementType())) {
```
python3kgae, line 9467: What will happen if the element type is not struct type?

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -440,11 +440,10 @@ static void AddRecordAccessMethod(clang::ASTContext &Ctx,
...
   }
 
-  StringRef OpcodeGroup = GetHLOpcodeGroupName(HLOpcodeGroup::HLIntrinsic);
-  unsigned Opcode =
-      static_cast<unsigned>(IntrinsicOp::IOP_ExtractRecordStructFromArray);
+  StringRef OpcodeGroup = GetHLOpcodeGroupName(HLOpcodeGroup::HLSubscript);
+  unsigned Opcode = static_cast<unsigned>(HLSubscriptOpcode::DefaultSubscript);
   MethodDecl->addAttr(HLSLIntrinsicAttr::CreateImplicit(
-      Ctx, OpcodeGroup, "ExtractRecordStructFromArray", Opcode));
+      Ctx, OpcodeGroup, "DefaultSubscript", Opcode));
```
tex3d, line 446: I think `HLSubscript` group always uses empty string here.

tex3d:
This doesn't update the code in `ScalarReplAggregatesHLSL.cpp`, in `isReadOnlyResSubscriptOrLoad`:
```cpp
    } else if (opcode == HLSubscriptOpcode::DefaultSubscript) {
      DxilResourceProperties RP = GetResPropsFromHLAnnotateHandle(
          PtrCI->getArgOperand(HLOperandIndex::kSubscriptObjectOpIdx));
      if (RP.getResourceClass() == DXIL::ResourceClass::SRV)
        return true;
    }
```

This will result in `Invalid` for resource class (since the handle isn't a resource handle).  While this won't cause a failure as such, it may prevent a memcpy optimization that may otherwise be legal on read-only record types.  It should probably be investigated whether this optimization makes sense for read-only records.

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/NodeInput_type.ll
```diff
@@ -126,7 +126,7 @@ declare %dx.types.NodeHandle @"dx.hl.annotatenodehandle..%dx.types.NodeHandle (i
 declare %"struct.NodeOutput<loadStressRecord>" @"dx.hl.cast..%\22struct.NodeOutput<loadStressRecord>\22 (i32, %dx.types.NodeHandle)"(i32, %dx.types.NodeHandle) #2
 
 attributes #0 = { nounwind }
-attributes #1 = { nounwind "dx.hlls"="ExtractRecordStructFromArray" }
+attributes #1 = { nounwind "dx.hlls"="DefaultSubscript" }
```
tex3d, line 129: There should be more/different changes to this file.  The HL function using this attribute should have changed in name.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -9098,7 +9096,7 @@ void TranslateHLSubscript(CallInst *CI, HLSubscriptOpcode opcode,
     Type *HandleTy = hlslOP->GetHandleType();
     if (ptr->getType() == hlslOP->GetNodeRecordHandleType()) {
       DXASSERT(false, "Shouldn't get here, NodeRecord subscripts should have "
-                      "generated ExtractRecordStructFromArray intrinsic");
+                      "generated DefaultSubscript intrinsic");
```
tex3d, line 9099: This should say something different, like:
"NodeRecord subscripts should have been lowered in LowerRecordAccessToGetNodeRecordPtr"

tools/clang/test/HLSLFileCheck/hlsl/workgraph/ast-nodeinput.hlsl
```diff
@@ -20,7 +20,7 @@ void node01(DispatchNodeInputRecord<RECORD> input) {}
 //CHECK-NEXT: HLSLNodeObjectAttr {{0x[0-9a-fA-F]+}} <<invalid sloc>> Implicit DispatchNodeInputRecord
 //CHECK-NEXT: FieldDecl {{0x[0-9a-fA-F]+}} <<invalid sloc>> <invalid sloc> implicit h 'int'
 //CHECK-NEXT: CXXMethodDecl {{0x[0-9a-fA-F]+}} <<invalid sloc>> <invalid sloc> Get 'const recordtype &() const'
-//CHECK-NEXT: HLSLIntrinsicAttr {{0x[0-9a-fA-F]+}} <<invalid sloc>> Implicit "op" "ExtractRecordStructFromArray" 18
+//CHECK-NEXT: HLSLIntrinsicAttr {{0x[0-9a-fA-F]+}} <<invalid sloc>> Implicit "subscript" "DefaultSubscript" 0
```
tex3d, line 23: Just a comment, I don't think any change is necessary at this point:

At first, I was concerned that there may be a problem generating a subscript operation without an index parameter, but none of the code appears to depend on the index parameter for a `DefaultSubscript` until after `LowerRecordAccessToGetNodeRecordPtr` eliminates it.

Still, it feels a little odd to be generating a DefaultSubscript function without an index parameter.

tex3d (changes requested):
I think the `NodeInput_test.ll` test needs an IR update for the function, plus a few other minor things.

lib/HLSL/HLOperationLower.cpp
tex3d: Should not be possible with earlier record template checks, so we should probably keep the `cast<>` we had before.  I'm not sure why this change was made here.

bob80905: This was a remnant of a change I needed to make for some reason that can be reverted now.

tools/clang/lib/AST/ASTContextHLSL.cpp
tex3d: How was this resolved?  I meant that we should pass `""` instead of `"DefaultSubscript"`.  This will impact the AST tests as well.

bob80905: Ah, I thought you meant OpcodeGroup will always have the value of "" so I might as well pass "".
Regardless, the reason I switched it back is because leaving it an empty string will cause the preprocess destructor to throw an exception, causing some spir-v tests to fail. It's a strange bug that will take a lot of time to fix, but perhaps it should be left to a separate PR.

pow2clk: I'm confused how spir-v tests are affected by this as this is workgraph specific code

lib/HLSL/HLOperationLower.cpp
```diff
@@ -9493,16 +9491,24 @@ void LowerRecordAccessToGetNodeRecordPtr(HLModule &HLM) {
...
-        IntrinsicOp opcode = static_cast<IntrinsicOp>(hlsl::GetHLOpcode(CI));
-        if (opcode != IntrinsicOp::IOP_ExtractRecordStructFromArray)
+        HLSubscriptOpcode opcode =
+            static_cast<HLSubscriptOpcode>(hlsl::GetHLOpcode(CI));
+        if (opcode != HLSubscriptOpcode::DefaultSubscript)
+          continue;
+
+        hlsl::OP *OP = &helper.hlslOP;
+        Value *Handle = CI->getArgOperand(HLOperandIndex::kHandleOpIdx);
+        if (Handle->getType() != OP->GetNodeRecordHandleType()) {
```
pow2clk, line 9508: Should this ever happen? Maybe there should be an assert in there as well?

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/NodeInput_type.ll
```diff
@@ -8,6 +8,22 @@
...
+;
+; }
+;
+;
+; Resource Bindings:
+;
+; Name                                 Type  Format         Dim      ID      HLSL Bind  Count
+; ------------------------------ ---------- ------- ----------- ------- -------------- ------
+; $Globals                          cbuffer      NA          NA     CB0   cb4294967295     1
+;
```
pow2clk, line 26: There's at least some disagreement about whether this commented information is useful in a committed .ll test case. It's not strictly necessary.

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/NodeInput_type.ll
```diff
@@ -34,87 +50,76 @@ define void @loadStress_16(%"struct.DispatchNodeInputRecord<loadStressRecord>"*
...
   store %"struct.NodeOutput<loadStressRecord>" %2, %"struct.NodeOutput<loadStressRecord>"* %loadStressChild
   %3 = call %dx.types.NodeRecordHandle @"dx.hl.createnodeinputrecordhandle..%dx.types.NodeRecordHandle (i32, i32)"(i32 13, i32 0)
-  %4 = call %dx.types.NodeRecordHandle @"dx.hl.annotatenoderecordhandle..%dx.types.NodeRecordHandle (i32, %dx.types.NodeRecordHandle, %dx.types.NodeRecordInfo)"(i32 17, %dx.types.NodeRecordHandle %3, %dx.types.NodeRecordInfo { i32 97, i32 116 })
+  %4 = call %dx.types.NodeRecordHandle @"dx.hl.annotatenoderecordhandle..%dx.types.NodeRecordHandle (i32, %dx.types.NodeRecordHandle, %dx.types.NodeRecordInfo)"(i32 17, %dx.types.NodeRecordHandle %3, %dx.types.NodeRecordInfo { i32 97, i32 128 })
   %5 = call %"struct.DispatchNodeInputRecord<loadStressRecord>" @"dx.hl.cast..%\22struct.DispatchNodeInputRecord<loadStressRecord>\22 (i32, %dx.types.NodeRecordHandle)"(i32 11, %dx.types.NodeRecordHandle %4)
   store %"struct.DispatchNodeInputRecord<loadStressRecord>" %5, %"struct.DispatchNodeInputRecord<loadStressRecord>"* %inputData
-  %6 = alloca %"struct.GroupNodeOutputRecords<loadStressRecord>"
```
pow2clk, line 44: Where is the HLSL for this? My expectation is that somehow this change made it so that it realized it didn't need two allocas for the same struct off the bat. This led to the elimination of the memcpy's which caught my attention. I worried that they might be related to what this test is trying to capture.

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/NodeInput_type.ll
```diff
@@ -8,6 +8,22 @@
 ; CHECK: alloca %"struct.DispatchNodeInputRecord<loadStressRecord>"
 ; CHECK: load %"struct.DispatchNodeInputRecord<loadStressRecord>", %"struct.DispatchNodeInputRecord<loadStressRecord>"*
```
pow2clk, line 9: It's not strictly related, but given there is another nodeio type with an alloca `%agg.tmp` and load here, I think we ought to check for them as well.

pow2clk (commented):
I have questions! 😄

tools/clang/lib/AST/ASTContextHLSL.cpp
tex3d: I am also perplexed.  This is already used with empty string elsewhere for the DefaultSubscript case.  Could it be a bad incremental build?

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/NodeInput_type.ll
tex3d: The hlsl file is mentioned in a comment at the bottom.  I think it would be easier to notice if that comment was moved near the top of the file, instead.

See: https://github.com/microsoft/DirectXShaderCompiler/pull/5893/files#diff-e1f9916f53b273f0d0e8ded4023b2b024a5c7811294e8cd181313c6dcb36792bR197
```ll
; python3 ExtractIRForPassTest.py -p scalarrepl-param-hlsl -o outtest_without_nodeiotypecheck.ll %HLSL_SRC_DIR%\tools\clang\test\HLSLFileCheck\hlsl\workgraph\called_function_arg_record_object.hlsl -- -T lib_6_8
```

tools/clang/lib/AST/ASTContextHLSL.cpp
bob80905: I've made the change before, and the testing pipeline failed. I've run Dr. Memory to test for memory leaks and there are a couple, but it could potentially be because the exception was thrown before the code that would free the memory could execute, so I'm not sure. I do know that the phrase "bad heap allocation" shows up when running the test.
Alternatively, would it make sense to change all cases where HLSubscript uses empty string and instead make it a string that references the opcode? Why must it be an empty string?

lib/HLSL/HLOperationLower.cpp
bob80905: This condition can be true on non-NodeRecordHandle types.  DxilGenerationPass.cpp:runOnModule will unconditionally call LowerRecordAccessToGetNodeRecordPtr without checking the record type. One could argue that there should be logic preventing that function from being called in the caller function when the handle type is not a node record, but that is a separate issue. So, I think there should not be any assert here.

PR #5894, opened 2023-10-19
[SPIR-V] Initialize base classes first 

cassiebeckley:
When using an initializer list to initialize a class with base classes, the base classes should be initialized first, and then used as members of the child class.

This seems to be something of a niche case, but it is one of the asserts that are failing for https://github.com/microsoft/DirectXShaderCompiler/issues/5638.


sudonatalie:
Can you add a test case that would have failed before this change? I see no diff between the old and new tests except the move to lit.

cassiebeckley:
Hm, I had a test case in there, not sure what happened to it. I’ll add it back in.

cassiebeckley:
I've added the test case, thanks for catching that!

sudonatalie (approved):

s-perron (approved):

PR #5895, opened 2023-10-19
[bug] Fix crash when lowering multi-dimension groupshared variable

dmpots:
This commit fixes a crash in the compiler when lowering a groupshared variable with a multi-dimensional array type. The root cause of the bug was that we had a nested gep expression that could not be merged into a single gep because of an intervening addrspacecast.

The `MultiDimArrayToOneDimArray` pass flattens the multi-dimension global variables to a single dimension. It relies on the `MergeGepUse` function to flatten any nested geps into a single gep that fully dereferences a scalar element.

The fix is to modify the `MergeGepUse` function to look through addrspacecast instructions when trying to merge geps. We can now merge geps like

    gep(addrspacecast(gep(p0, gep_args0)) to p1*, gep_args1)

into

    addrspacecast(gep(p0, gep_args0+gep_args1) to p1*)

We also added a call to `removeDeadConstantUsers` before flattening multi-dimension globals because we can have some dead constants hanging around after merging geps and these constants should be ignored by the flattening pass.

### DIFF
--- a/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
+++ b/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
@@ -109,7 +109,7 @@ Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP,
   // Wrap the new gep in an addrspacecast if needed.
   if (AsCast)
     newGEP = Builder.CreateAddrSpaceCast(
-        newGEP, PointerType::get(newGEP->getType()->getPointerElementType(),
+        newGEP, PointerType::get(GEP->getType()->getPointerElementType(),
                                  AsCast->getDestAddressSpace()));
   GEP->replaceAllUsesWith(newGEP);
   if (Instruction *I = dyn_cast<Instruction>(GEP))
--- a/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
+++ b/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
@@ -48,7 +48,7 @@ namespace {
 //      addrspacecast(gep(p0, gep_args0+gep_args1) to p1*)
 //
 Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP,
-                AddrSpaceCastOperator *AsCast = nullptr) {
+                AddrSpaceCastOperator *AsCast) {
   IRBuilder<> Builder(GEP->getContext());
   StringRef Name = "";
   if (Instruction *I = dyn_cast<Instruction>(GEP)) {
--- a/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
+++ b/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
@@ -86,7 +86,7 @@ Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP,
     }
 
     // Update the GEP in place if possible.
-    if (SrcGEP->getNumOperands() == 2) {
+    if (SrcGEP->getNumOperands() == 2 && !AsCast) {
       GEP->setOperand(0, SrcGEP->getOperand(0));
       GEP->setOperand(1, Sum);
       return GEP;
--- a/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
+++ b/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
@@ -37,7 +37,18 @@ using namespace hlsl;
 
 namespace {
 
-Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP) {
+// Attempt to merge the two GEPs into a single GEP.
+//
+// If `AsCast` is non-null the merged GEP will be wrapped
+// in an addrspacecast before replacing users. This allows
+// merging GEPs of the form
+//
+//      gep(addrspacecast(gep(p0, gep_args0) to p1*), gep_args1)
+// into
+//      addrspacecast(gep(p0, gep_args0+gep_args1) to p1*)
+//
+Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP,
+                AddrSpaceCastOperator *AsCast = nullptr) {
   IRBuilder<> Builder(GEP->getContext());
   StringRef Name = "";
   if (Instruction *I = dyn_cast<Instruction>(GEP)) {
@@ -94,9 +105,61 @@ Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP) {
   DXASSERT(!Indices.empty(), "must merge");
   Value *newGEP =
       Builder.CreateInBoundsGEP(nullptr, SrcGEP->getOperand(0), Indices, Name);
+
+  // Wrap the new gep in an addrspacecast if needed.
+  if (AsCast)
+    newGEP = Builder.CreateAddrSpaceCast(
+        newGEP, PointerType::get(newGEP->getType()->getPointerElementType(),
+                                 AsCast->getDestAddressSpace()));
   GEP->replaceAllUsesWith(newGEP);
   if (Instruction *I = dyn_cast<Instruction>(GEP))
     I->eraseFromParent();
+  return newGEP;
+}
+
+// Examine the gep and try to merge it when the input pointer is
+// itself a gep. We handle two forms here:
+//
+//      gep(gep(p))
+//      gep(addrspacecast(gep(p)))
+//
+// If the gep was merged successfully then return the updated value, otherwise
+// return nullptr.
+//
+// When the gep is sucessfully merged we will delete the gep and also try to
+// delete the nested gep and addrspacecast.
+static Value *TryMegeWithNestedGEP(GEPOperator *GEP) {
+  // Sentinal value to return when we fail to merge.
+  Value *FailedToMerge = nullptr;
+
+  Value *Ptr = GEP->getPointerOperand();
+  GEPOperator *prevGEP = dyn_cast<GEPOperator>(Ptr);
+  AddrSpaceCastOperator *AsCast = nullptr;
+
+  // If there is no directly nested gep try looking through an addrspacecast to
+  // find one.
+  if (!prevGEP) {
+    AsCast = dyn_cast<AddrSpaceCastOperator>(Ptr);
+    if (AsCast)
+      prevGEP = dyn_cast<GEPOperator>(AsCast->getPointerOperand());
+  }
+
+  // Not a nested gep expression.
+  if (!prevGEP)
+    return FailedToMerge;
+
+  // Try merging the two geps.
+  Value *newGEP = MergeGEP(prevGEP, GEP, AsCast);
+  if (!newGEP)
+    return FailedToMerge;
+
+  // Delete the nested gep and addrspacecast if no more users.
+  if (AsCast && AsCast->user_empty() && isa<AddrSpaceCastInst>(AsCast))
+    cast<AddrSpaceCastInst>(AsCast)->eraseFromParent();
+
+  if (prevGEP->user_empty() && isa<GetElementPtrInst>(prevGEP))
+    cast<GetElementPtrInst>(prevGEP)->eraseFromParent();
+
   return newGEP;
 }
 
@@ -130,23 +193,14 @@ bool MergeGepUse(Value *V) {
         // merge any GEP users of the untranslated bitcast
         addUsersToWorklist(V);
       }
+    } else if (isa<AddrSpaceCastOperator>(V)) {
+      addUsersToWorklist(V);
     } else if (GEPOperator *GEP = dyn_cast<GEPOperator>(V)) {
-      if (GEPOperator *prevGEP =
-              dyn_cast<GEPOperator>(GEP->getPointerOperand())) {
-        // merge the 2 GEPs, returns nullptr if couldn't merge
-        if (Value *newGEP = MergeGEP(prevGEP, GEP)) {
-          changed = true;
-          worklist.push_back(newGEP);
-          // delete prevGEP if no more users
-          if (prevGEP->user_empty() && isa<GetElementPtrInst>(prevGEP)) {
-            cast<GetElementPtrInst>(prevGEP)->eraseFromParent();
-          }
-        } else {
-          addUsersToWorklist(GEP);
-        }
+      if (Value *newGEP = TryMegeWithNestedGEP(GEP)) {
+        changed = true;
+        worklist.push_back(newGEP);
       } else {
-        // nothing to merge yet, add GEP users
-        addUsersToWorklist(V);
+        addUsersToWorklist(GEP);
       }
     }
   }
--- a/test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
+++ b/test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
@@ -12,6 +12,7 @@ target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:3
 target triple = "dxil-ms-dx"
 
 @ArrayOfArray = addrspace(3) global [256 x [9 x float]] undef, align 4
+@ArrayOfArrayOfArray = addrspace(3) global [256 x [9 x [3 x float]]] undef, align 4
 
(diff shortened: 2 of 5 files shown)


lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
```diff
@@ -37,7 +37,18 @@ using namespace hlsl;
...
+// If `AsCast` is non-null the merged GEP will be wrapped
+// in an addrspacecast before replacing users. This allows
+// merging GEPs of the form
+//
+//      gep(addrspacecast(gep(p0, gep_args0) to p1*), gep_args1)
+// into
+//      addrspacecast(gep(p0, gep_args0+gep_args1) to p1*)
+//
+Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP,
+                AddrSpaceCastOperator *AsCast = nullptr) {
```
python3kgae, line 51: Why add default value for AsCast?
Is there use of MergeGEP somewhere else?

python3kgae (approved):

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
dmpots: That is a leftover from an earlier iteration where there were multiple calls. I removed it now since it is no longer needed.

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
```diff
@@ -0,0 +1,103 @@
...
+@ArrayOfArray = addrspace(3) global [256 x [9 x float]] undef, align 4
+
+; Test that we can merge the geps when all parts are instructions.
+; CHECK-LABEL: @merge_gep_instr_instr_instr
+; CHECK:  load float, float* addrspacecast (float addrspace(3)* getelementptr inbounds ([2304 x float], [2304 x float] addrspace(3)* @ArrayOfArray.1dim, i32 0, i32 1) to float*)
+define void @merge_gep_instr_instr_instr() {
+entry:
+  %gep0 = getelementptr inbounds [256 x [9 x float]], [256 x [9 x float]] addrspace(3)* @ArrayOfArray, i32 0, i32 0
+  %asc  = addrspacecast [9 x float] addrspace(3)* %gep0 to [9 x float]*
+  %gep1 = getelementptr inbounds [9 x float], [9 x float]* %asc, i32 0, i32 1
```
tex3d, line 23: Should we have a test where there are two GEPs after the addrspacecast to show that it successfully migrates the addrspacecast fully across all the GEPs?  It's something I wasn't sure would work properly until I thought more carefully about the algorithm.  I think it will work now, but it would be good to have a test that verifies this.

dmpots: Added a few new tests to cover this case. It uncovered a bug with an early return in the `MergeGEP` function that triggers when looking through the addrspace cast and the GEP ends in an array.

Fixed the bug and added new tests to cover both the addrspacecast case and the "plain" case.

damyanp (approved):

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
```diff
@@ -101,3 +102,97 @@ entry:
...
+; CHECK-LABEL: addrspace_cast_new_type
+; CHECK:  %0 = mul i32 %idx0, 9
+; CHECK:  %1 = add i32 %idx1, %0
+; CHECK:  %2 = getelementptr [2304 x float], [2304 x float] addrspace(3)* @ArrayOfArray.1dim, i32 0, i32 %1
+; CHECK:  %3 = addrspacecast float addrspace(3)* %2 to i32*
+; CHECK:  load i32, i32* %3
+define void @addrspace_cast_new_type(i32 %idx0, i32 %idx1) {
+entry:
+  %gep0 = getelementptr inbounds [256 x [9 x float]], [256 x [9 x float]] addrspace(3)* @ArrayOfArray, i32 0, i32 %idx0
+  %asc  = addrspacecast [9 x float] addrspace(3)* %gep0 to [3 x i32]*
```
tex3d, line 194: Wait, you're allowed to change the type in an addrspacecast?  I thought that would require a bitcast.

dmpots: Yeah, it seems a bit bonkers to me but apparently it is allowed. But for addrspace cast from/to the same addrspace you have to use a bitcast it says:

https://llvm.org/docs/LangRef.html#addrspacecast-to-instruction

> It can be a no-op cast or a complex value modification, depending on the target and the address space pair. Pointer conversions within the same address space must be performed with the bitcast instruction.

I interpreted the "complex value modification" to mean it can change the type and the bitcode was accepted as valid when I tried it. 

I was unable to get that pattern from the hlsl I tried though.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
```diff
@@ -109,7 +109,7 @@ Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP,
   // Wrap the new gep in an addrspacecast if needed.
   if (AsCast)
     newGEP = Builder.CreateAddrSpaceCast(
-        newGEP, PointerType::get(newGEP->getType()->getPointerElementType(),
+        newGEP, PointerType::get(GEP->getType()->getPointerElementType(),
```
tex3d, line 112: Are newGEP and GEP different types?  Which scenario does switching newGEP for GEP fix?

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
```diff
@@ -101,3 +102,97 @@ entry:
...
+; CHECK:  %2 = mul i32 %1, 3
+; CHECK:  %3 = add i32 %idx2, %2
+; CHECK:  %4 = getelementptr [6912 x float], [6912 x float] addrspace(3)* @ArrayOfArrayOfArray.1dim, i32 0, i32 %3
+; CHECK:  %5 = addrspacecast float addrspace(3)* %4 to float*
+; CHECK:  load float, float* %5
+define void @merge_gep_multi_level_with_addrspace(i32 %idx0, i32 %idx1, i32 %idx2) {
+entry:
+  %gep0 = getelementptr inbounds [256 x [9 x [3 x float]]], [256 x [9 x [3 x float]]] addrspace(3)* @ArrayOfArrayOfArray, i32 0, i32 %idx0
+  %asc  = addrspacecast [9 x [3 x float]] addrspace(3)* %gep0 to [9 x [3 x float]]*
+  %gep1 = getelementptr inbounds [9 x [3 x float]], [9 x [3 x float]]* %asc, i32 0, i32 %idx1
```
tex3d, line 157: What happens if there are two addrspacecast instructions, which may or may not cancel each other out?  I don't know how exactly that might be generated from HLSL, but seems like a legal form that MergeGEP might need to handle.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
dmpots: That is needed to support the change of type in the addrspacecast. The `newGEP` has the type of the result after the addrspacecast so when we generate the addrspacecase we have to use the original type.

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
dmpots: The code currently only looks back through one level of addrspacecast. If there were multiple addrspacecasts between the gep it would not merge the geps.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
```diff
@@ -75,7 +86,7 @@ Value *MergeGEP(GEPOperator *SrcGEP, GEPOperator *GEP) {
     }
 
     // Update the GEP in place if possible.
-    if (SrcGEP->getNumOperands() == 2) {
+    if (SrcGEP->getNumOperands() == 2 && !AsCast) {
```
tex3d, line 89: Do we ever hit this case where `Sum` is non-zero, especially if `GO1` was non-zero?

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
tex3d: When reading through this code, I found myself wishing there were comments with examples showing what various branches are handling.  It's not easy to follow and recognize where test coverage might be missing, or the bugs might be.

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
tex3d: If the pointer type can be changed by an addrspacecast, then won't this code break down if the shape is changed?  If you change the structure through the addrspacecast, then try to move the addrspace cast to the end, it may invalidate the GEP indices that are being moved across the addrspacecast and concatenated into one GEP.

For example, imagine a multi-dim array is changed to a single-dim array in the addrspacecast, then a GEP with a single index indexes into that array.  You can't move the addrspacecast and merge GEPs, since you can't just copy the one index into the single-dim array to index the multi-dim array before the addrspacecast.  I think if there's a type change, and you want to move the addrspacecast to the end, you have to insert a bitcast to replace the addrspacecast for the type change, and you can't merge GEPs across the bitcast.  Either that, or you have to just leave the addrspacecast where it is, and it blocks GEP merging across that operation.

To illustrate with the MergeGEP comment:
```cpp
//      gep(addrspacecast(gep(p0, gep_args0) to p1*), gep_args1)
// into
//      addrspacecast(gep(p0, gep_args0+gep_args1) to p1*)
```
This is no longer valid if addrspacecast changes the type, because `gep_args1` in `gep_args0+gep_args1` is indexing a different type than `gep_args1` was originally.

Also, as a side note, `p1*` is not the same in these two expressions.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
dmpots: I agree that more comments would be great here. I tried to leave it better than I found it.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
dmpots: I think I hit it in one of the test cases I was writing at some point. But its not clear we can reach that case from hlsl. Probably we can, but it has yet to be seen in practice.

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
dmpots: Yeah, I think it will break in a more aggressive change of type. Should I change it to not merge the geps if the addrspacecast changes types instead of trying to support the simple case?

tex3d: I think, to be safe, we should stop the merge if the addrspacecast changes the type in between GEPs.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
tex3d: Normally, we don't use the first index for anything in HLSL - it's always zero.  However, IIRC, I saw these used once when generated for array element construction or something like that.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
tex3d: Given the discussion about how we shouldn't merge GEPs around an addrspacecast that changes the pointer element type, I presume `newGEP` and `GEP` types will always be equivalent if we prevent that situation, right?

dmpots: Yes, they should be the same in that case. The GEP is RAWU newGEP just below here.

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
dmpots: Sounds good to me. I'll put up a new PR to do that.

dmpots: Note that in the case I was looking at originally (the multi-dim-to-one-dim pass) this will cause that pass to crash because it requires a single level merged gep for the accesses.

So for the test case in that PR we will be expecting a crash.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
```diff
@@ -130,23 +193,14 @@ bool MergeGepUse(Value *V) {
         // merge any GEP users of the untranslated bitcast
         addUsersToWorklist(V);
       }
+    } else if (isa<AddrSpaceCastOperator>(V)) {
+      addUsersToWorklist(V);
```
tex3d, line 197: I think generally collapsing addrspacecast chains would require an addition here to look at the pointer operand to see if it is also an addrspacecast.  If so, we can create a new merged addrspacecast, replace uses, and add it to the work list.

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
tex3d: This seems like a useful case to include a test for.  It does seem like it could be handled better as well.  

For instance:
```
This:
  addrspace(3) GEP -> addrspacecast(0) -> GEP -> addrspacecast(3)
could simplify to:
  addrspace(3) mergedGEP
instead of:
  addrspace(3) mergedGEP -> addrspacecast(0) -> addrspacecast(3)

And:
  addrspace(5) GEP -> addrspacecast(0) -> GEP -> addrspacecast(3)
could simplify to:
  addrspace(5) mergedGEP -> addrspacecast(3)
instead of:
  addrspace(5) mergedGEP -> addrspacecast(0) -> addrspacecast(3)

Testing both with and without a GEP at the end as well:
GEP -> addrspacecast -> GEP -> addrspacecast -> GEP
```

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
tex3d: In that case, was an addrspacecast changing the type in the middle of GEPs being merged?

dmpots: No, I never hit the change of type in an addrspacecast in real life hlsl code. Just in artificial IR test cases.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
dmpots: I think it is not too hard to break this code by writing some custom IR test cases, but I'm not sure its worth hardening this code against unexpected inputs.

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
dmpots: Yes, that sounds like it would work. Not sure its worth the extra complexity in the code. Those should hopefully all be folded away by instcombine. If there was an hlsl source that could generate it that would be more compelling.

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
dmpots: I'm not sure those tests actually add anything new? The addrspacecast will not be merged with the current code.

test/HLSL/passes/multi_dim_one_dim/gep_addrspacecast_gep.ll
dmpots: Made the change to stop the merge here: https://github.com/microsoft/DirectXShaderCompiler/pull/5914

lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
dmpots: I'm going to leave the code as is in case we want to support it in the future.

PR #5899, opened 2023-10-19
[Instcombine] Disable bswap match for DXIL

### LINKED ISSUE #5104: Endian swap function outputs "llvm.bswap.i32' is not a DXIL function" as an error
I only tested on commit 51777bd2b04a2d32d0d816b0dfe2145b770c2d6a with the `-HV 2021` flag. Using`-Od` eliminates the issue.

Using the following code results in this error

```
uint EndianSwapUINT32(uint i) {
    return ((i << 24) & 0xFF000000) | ((i << 8) & 0xFF0000) | ((i >> 8) & 0xFF00) | ((i >> 24) & 0xFF);
}
```

It seems LLVM is recognizing this particular pattern and generating an instruction not recognized by DXIL.

I'm not sure if more code is necessary, but the function is merely used to swap an uint loaded from an SRV.

Apologies in advance if this has been fixed in a more recent version.

llvm-beanz:
DXIL doesn't support the bswap intrinsic, normally a backend would lower this intrinsic to something else, in this case we don't have a backend so we just end up generating invalid DXIL.

The solution here is to disable bswap matching when targeting DXIL.

Fixes #5104

### DIFF
--- a/test/HLSL/passes/instcombine/disable-bswap-match.ll
+++ b/test/HLSL/passes/instcombine/disable-bswap-match.ll
@@ -1,0 +1,89 @@
+; RUN: opt -instcombine -S %s | FileCheck %s
+
+; CHECK-NOT: call i32 @llvm.bswap.i32(i32 %5)
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+%Consts = type { i32 }
+%dx.types.Handle = type { i8* }
+%dx.types.ResourceProperties = type { i32, i32 }
+%dx.types.CBufRet.i32 = type { i32, i32, i32, i32 }
+
+@Consts = external constant %Consts
+@llvm.used = appending global [1 x i8*] [i8* bitcast (%Consts* @Consts to i8*)], section "llvm.metadata"
+
+; Function Attrs: nounwind readnone
+declare %dx.types.Handle @"dx.hl.createhandle..%dx.types.Handle (i32, %Consts*, i32)"(i32, %Consts*, i32) #0
+
+; Function Attrs: nounwind readnone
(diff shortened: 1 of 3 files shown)


test/HLSL/passes/instcombine/disable-bswap-match.ll
```diff
@@ -0,0 +1,89 @@
+; RUN: opt -instcombine -S %s | FileCheck %s
```
python3kgae, line 1: Maybe put into test\HLSLFileCheck\passes.
Things under test/HLSL will not be tested automatically.

llvm-beanz: I fixed that in https://github.com/microsoft/DirectXShaderCompiler/pull/5887. These tests now run automatically as expected.

tex3d (approved):
LGTM!

test/HLSL/passes/instcombine/disable-bswap-match.ll
llvm-beanz: This test isn’t in that directory. This is `test/HLSL`, not `clang/test/HLSL`

python3kgae: I see. :( Sorry about it.
If that's the case, then maybe test/DXIL is better? Since is not about HLSL but DXIL limitation?

python3kgae (approved):

test/HLSL/passes/instcombine/disable-bswap-match.ll
llvm-beanz: The right place for this test (and the other tests in the `test/HLSL` folder) is under `Transforms`, but we currently have that directory disabled. We should fix the `Transforms` suite, then move these under. I don't think moving these to a `DXIL` directory is necessarily helpful because they aren't tests for `lib/DXIL` they're tests for `lib/Transform`.

PR #5903, opened 2023-10-20
Fix clang linker error in debug

amaiorano:
When building DXC for Chrome, although unused, it attempts to link ASTUnit::LoadFromASTFile (perhaps it's not as good at stripping unused symbolds). This patch slightly modifies the if-def'ing out of the entire definition so that it only if-def's out the function body.


llvm-beanz (commented):
I’m concerned about this fix. I don’t want us to link that function in and have people think it works since binary serialization of ASTs is _really_ broken in DXC.

What is the linker command line and error you’re encountering? I assume the linker is lld on Linux (clang isn’t a linker)?

amaiorano:
> What is the linker command line and error you’re encountering? I assume the linker is lld on Linux (clang isn’t a linker)?

Yes, this is `lld`. Here's the command line and error (this is using our GN build for Dawn):

```
ninja: Entering directory `out/debug'
[3/3] SOLINK_MODULE ./libdxcompiler.so
FAILED: libdxcompiler.so 
../../third_party/llvm-build/Release+Asserts/bin/clang++ -shared -fuse-ld=lld -Wl,--fatal-warnings -Wl,--build-id -fPIC -Wl,-z,noexecstack -Wl,-z,relro -Wl,--color-diagnostics -Wl,--undefined-version -Wl,--no-call-graph-profile-sort -m64 -no-canonical-prefixes -Wl,--gdb-index -rdynamic -Wl,-z,defs -Wl,--as-needed -nostdlib++ --sysroot=../../build/linux/debian_bullseye_amd64-sysroot -Wl,-rpath=\$ORIGIN -Werror -o "./libdxcompiler.so" -Wl,-soname="libdxcompiler.so" @"./libdxcompiler.so.rsp"
ld.lld: error: undefined symbol: clang::ASTUnit::LoadFromASTFile(std::__Cr::basic_string<char, std::__Cr::char_traits<char>, std::__Cr::allocator<char>> const&, clang::PCHContainerReader const&, llvm::IntrusiveRefCntPtr<clang::DiagnosticsEngine>, clang::FileSystemOptions const&, bool, llvm::ArrayRef<std::__Cr::pair<std::__Cr::basic_string<char, std::__Cr::char_traits<char>, std::__Cr::allocator<char>>, llvm::MemoryBuffer*>>, bool, bool, bool)
>>> referenced by ASTMerge.cpp:51 (../../third_party/dxc/tools/clang/lib/Frontend/ASTMerge.cpp:51)
>>>               clang_lib_frontend/ASTMerge.o:(clang::ASTMergeAction::ExecuteAction()) in archive obj/third_party/gn/dxc/libclang_lib_frontend.a
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.
```

I'm not sure why we get this only in our debug build on Linux. On Windows, both release and debug build just fine. On Linux, the release build is fine, but the debug one fails with this linker error. Something is causing the symbol not to be removed in this case.

If the concern is to make sure that function isn't called, maybe we can remove/disable the code that calls into it? Or disable the higher level option?

amaiorano:
I looked through the code for all call sites to `ASTUnit::LoadFromASTFile`, and there are 4, 3 of which have been if-def'd out for DXC. If I if-def out the remaining code in `ASTMergeAction::ExecuteAction()`, then everything builds and links fine. Shall I proceed with that change instead?

amaiorano:
Am going to close this, as I was able to simply remove building ASTMerge.cpp altogether from our GN build. A similar change can be done for the CMake build, so that we don't have to rely on the linker stripping unused symbols. @llvm-beanz provided the following patch which removes ASTMerge.cpp from the build, and removes the one reference to it in the codebase:

```
diff --git a/clang/lib/Frontend/CMakeLists.txt b/clang/lib/Frontend/CMakeLists.txt
index 1e5f0a859dfd..6e8eb4eaf51b 100644
--- a/clang/lib/Frontend/CMakeLists.txt
+++ b/clang/lib/Frontend/CMakeLists.txt
@@ -9,9 +9,11 @@ set(LLVM_LINK_COMPONENTS
   TargetParser
   )
 
+set(LLVM_OPTIONAL_SOURCES ASTMerge.cpp) # HLSL Change
+
 add_clang_library(clangFrontend
   ASTConsumers.cpp
-  ASTMerge.cpp
+  #ASTMerge.cpp - Disable AST serialization and merging
   ASTUnit.cpp
   ChainedDiagnosticConsumer.cpp
   ChainedIncludesSource.cpp
diff --git a/clang/lib/FrontendTool/ExecuteCompilerInvocation.cpp b/clang/lib/FrontendTool/ExecuteCompilerInvocation.cpp
index b280a1359d2f..0e2f0abbd526 100644
--- a/clang/lib/FrontendTool/ExecuteCompilerInvocation.cpp
+++ b/clang/lib/FrontendTool/ExecuteCompilerInvocation.cpp
@@ -186,11 +186,13 @@ CreateFrontendAction(CompilerInstance &CI) {
     Act = std::make_unique<WrappingExtractAPIAction>(std::move(Act));
   }
 
+#if 0 // HLSL Change Start - Disable AST Merging.
   // If there are any AST files to merge, create a frontend action
   // adaptor to perform the merge.
   if (!FEOpts.ASTMergeFiles.empty())
     Act = std::make_unique<ASTMergeAction>(std::move(Act),
                                             FEOpts.ASTMergeFiles);
+#endif // HLSL Change Start - Disable AST Merging.
 
   return Act;
 }
```

PR #5904, opened 2023-10-20
[lit] convert verify test as lit FileCheck test part1

python3kgae:
Convert first part of verify test shaders as lit FileCheck test.

This is for https://github.com/microsoft/DirectXShaderCompiler/issues/5870

CppErrorsHV2015 is skipped for cannot set hlsl 2015 thru dxc. 
CppErrors is skipped for non-expected warning when test with taef.


bob80905 (commented):
Can you explain your reasoning for why you chose to make the target profile lib_6_3? As opposed to, say, 6_0 or 6_5?

python3kgae:
> Can you explain your reasoning for why you chose to make the target profile lib_6_3? As opposed to, say, 6_0 or 6_5?

I want to use lib profile because it doesn't require entry.
And I want to keep the lowest shader model which is 6.3 for lib profile.

bob80905 (approved):

hekota (approved):
LGTM!

PR #5907, opened 2023-10-23
[SPIR-V] Fixup tests failing downstream

sudonatalie:
These tests are failing in our downstream builds after migrating to Lit due to CHECK-NOTs referencing variables before they are defined. Those CHECK-NOTs have now been replaced with an `--implicit-check-not` that checks that there are no `OpDecorate ... NoContraction` instructions other than those CHECKed in the test.

### DIFF
--- a/tools/clang/test/CodeGenSPIRV_Lit/decoration.no-contraction.stage-vars.hlsl
+++ b/tools/clang/test/CodeGenSPIRV_Lit/decoration.no-contraction.stage-vars.hlsl
@@ -1,4 +1,6 @@
-// RUN: %dxc -T vs_6_0 -E main -fspv-reflect -fcgl  %s -spirv | FileCheck %s
+// RUN: %dxc -T vs_6_0 -E main -fspv-reflect -fcgl  %s -spirv | FileCheck %s --implicit-check-not "OpDecorate {{%[0-9]+}} NoContraction"
+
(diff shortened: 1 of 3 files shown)


cassiebeckley:
These tests seem to be passing on GitHub and on my machine, do you know why that is?

tools/clang/test/CodeGenSPIRV_Lit/decoration.no-contraction.stage-vars.hlsl
```diff
@@ -18,15 +18,17 @@
 // CHECK-NEXT: OpDecorate [[cxcy_1:%[0-9]+]] NoContraction
 // CHECK-NEXT: OpDecorate [[cxcy_plus_dz_1:%[0-9]+]] NoContraction
 
-// CHECK-NOT: OpDecorate [[cxcy_2]] NoContraction
-// CHECK-NOT: OpDecorate [[cxcy_plus_dz_2]] NoContraction
+// IMPLICIT-CHECK-NOT: OpDecorate [[cxcy_2]] NoContraction
```
cassiebeckley, line 21: Does ’--implicit-check-not’ need to be passed on the command line for this?

sudonatalie: No... I didn't know that existed, but I can use it instead and it's probably more clear. You know an API is good when you make up a name for something and it actually already exists with that exact name.

cassiebeckley (approved):

Issue #5877, opened 2023-10-16
[Linux] Enable PixTest on linux


python3kgae:
**Is your feature request related to a problem? Please describe.**
Now PixTest is disable for linux because some tests depend on dia.
But many tests in PixTest don't depend on dia which could be enabled for Linux build.

**Describe the solution you'd like**
Use #ifdef _WIN32 to guard dia related code in PixTest.
Switch PixTest to gtest for Linux build.

PR #5908, opened 2023-10-23
[Test] enable rewriter test for linux

python3kgae:
Added InitSupport to setup m_dllSupport and m_pIncludeHandler.

Work around GetTempPathW and GetFileSize.

RewriterTest.RunNonUnicode is disabled for dependent on Greek language support.

This is for https://github.com/microsoft/DirectXShaderCompiler/issues/5877


tools/clang/test/HLSL/rewriter/includes.hlsl
```diff
@@ -1,6 +1,6 @@
 #include "toinclude.hlsl"
 #include "..\toinclude2.hlsl"
-#include "\for_includes_test\toinclude3.hlsl"
+#include ".\for_includes_test\toinclude3.hlsl"
```
llvm-beanz, line 3: Can you drop the `.\` entirely? The local directory should be on the include search path.
```suggestion
#include "for_includes_test\toinclude3.hlsl"
```

llvm-beanz (approved):
One small suggestion. Otherwise looks good.

tools/clang/test/HLSL/rewriter/includes.hlsl
python3kgae: Fixed.

adam-yang (approved):

PR #5909, opened 2023-10-23
[lit] convert verify test as lit FileCheck test part2

python3kgae:
Convert second part of verify test shaders as lit FileCheck test.

This is for #5870

Still same files left in HLSL folder need to check.


llvm-beanz (changes requested):
Making all the shaders lib shaders isn't really idea. Many of these are clearly not library shaders.

python3kgae:
> Making all the shaders lib shaders isn't really idea. Many of these are clearly not library shaders.

Fixed.

tools/clang/test/SemaHLSL/bitfields.hlsl
```diff
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -HV 2021 -fsyntax-only -ffreestanding -verify %s
+// RUN: %dxc -Tlib_6_3 -HV 2021 -verify %s
```
llvm-beanz, line 1: This shader looks like it is intended to be a vertex shader.

tools/clang/test/SemaHLSL/derived-to-base.hlsl
```diff
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -fsyntax-only -ffreestanding -verify %s
+// RUN: %dxc -Tlib_6_3 -verify %s
```
llvm-beanz, line 1: Also a vertex shader.

tools/clang/test/SemaHLSL/globallycoherent-mismatch.hlsl
```diff
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -fsyntax-only -ffreestanding -verify %s
+// RUN: %dxc -Tlib_6_3 -verify %s
```
llvm-beanz, line 1: This looks like a compute shader.

llvm-beanz (commented):
Please review the shaders that only have library run lines to ensure that you aren't missing the correct profiles. I commented on a few and only made it a short way through the PR.

tools/clang/test/SemaHLSL/globallycoherent-mismatch.hlsl
python3kgae: Fixed.

tools/clang/test/SemaHLSL/derived-to-base.hlsl
python3kgae: Fixed.

tools/clang/test/SemaHLSL/bitfields.hlsl
python3kgae: Fixed.

llvm-beanz:
I just want to confirm that you did make the change to the three shaders I explicitly called out, but there are other shaders that are still incorrect. Please review each of the shaders in this PR that still have only one `RUN` line to ensure that they are correct.

python3kgae:
> I just want to confirm that you did make the change to the three shaders I explicitly called out, but there are other shaders that are still incorrect. Please review each of the shaders in this PR that still have only one `RUN` line to ensure that they are correct.

Sorry. Missed the message :( Just noticed the diff.

Updated the rest.

llvm-beanz (approved):

tex3d (approved):
LGTM.

PR #5910, opened 2023-10-23
[lit] convert verify test as lit FileCheck test part3

python3kgae:
Move the verify shaders which not enabled in taef to lit.

This is for #5870

### DIFF
--- a/tools/clang/test/SemaHLSL/spec.hlsl
+++ b/tools/clang/test/SemaHLSL/spec.hlsl
@@ -1,4 +1,5 @@
-// RUN: %clang_cc1 -fsyntax-only -ffreestanding -verify %s
+// RUN: %dxc -Tlib_6_3 -HV 2018 -verify %s
+// RUN: %dxc -Tps_6_0 -HV 2018 -verify %s
 
 // This file provides test cases that are cross-references to the HLSL
 // specification.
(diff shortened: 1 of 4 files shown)


tools/clang/test/SemaHLSL/spec.hlsl
```diff
@@ -194,17 +195,17 @@ namespace ns_std_conversions {
     fn_f14(1);
 
     u = f11; // matrix single element conversion
-    u = f14; // matrix scalar truncation conversion
+    u = f14; // matrix scalar truncation conversion // expected-warning {{implicit truncation of vector type}}
```
llvm-beanz, line 198: The doubled up comments are gross. Can you clean this up by moving one of the comments to the line above?

Please look for other places where this occurs too.

llvm-beanz (approved):

tools/clang/test/SemaHLSL/template-diag-deferred.hlsl
```diff
@@ -1,3 +1,5 @@
+// RUN: %dxc -Tlib_6_6 -HV 2021 -verify %s
```
bob80905, line 1: Is lib_6_6 the lowest SM this test passes in?

python3kgae: Yes. Because it used ResourceDescriptorHeap which requires shader model 6.6.

PR #5914, opened 2023-10-24
Do not merge geps when an addrspacecast changes the underlying type.

dmpots:
We do not handle an addrspacecast that changes the shape of the underlying type when merging geps (e.g. array to non-array). Instead of trying to handle some cases we just avoid trying to merge geps in all cases where the underlying type changes shape. This causes the multi-dim-one-dim pass to crash because it expects the geps to be merged. I do not think we can hit this case from hlsl, but it may be possible.

### DIFF
--- a/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
+++ b/lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
@@ -144,6 +144,17 @@ static Value *TryMegeWithNestedGEP(GEPOperator *GEP) {
       prevGEP = dyn_cast<GEPOperator>(AsCast->getPointerOperand());
   }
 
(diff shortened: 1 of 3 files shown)


lib/DXIL/DxilUtilDbgInfoAndMisc.cpp
```diff
@@ -144,6 +144,17 @@ static Value *TryMegeWithNestedGEP(GEPOperator *GEP) {
       prevGEP = dyn_cast<GEPOperator>(AsCast->getPointerOperand());
   }
 
+  // If the addrspace cast changes the underlying type of the pointed to
```
bob80905, line 147: Nit: Would you mind changing the name of this function to TryMergeWithNestedGep ?

damyanp:

damyanp:

PR #5917, opened 2023-10-24
Reformat hct python scripts with black

tex3d:
A few syntax errors fixed along the way:
- hcttracei.py: `print` -> `print()`, `<>` -> `!=`
- hcttest-system-values.py: `os.path.join()` close paren before arguments


llvm-beanz (approved):
My only question here is do we actually run all these scripts? If not I’d prefer deleting things we don’t use, but if so this is great.

python3kgae (approved):

tex3d:
> My only question here is do we actually run all these scripts? If not I’d prefer deleting things we don’t use, but if so this is great.

I think we could sort that out as another exercise.  This just unblocks making a change to any of these files without getting whole-file formatting errors.

PR #5919, opened 2023-10-24
Make DxilRemoveDeadBlocks init DxilModule, fix test

tex3d:
DxilRemoveDeadBlocks assumed it could run with an uninitialized DxilModule, because the test case had been captured from live state without the needed DXIL metadata. DxilRemoveDeadBlocks relies on DxilValueCache, which relies on code in DxilSimplify, which assumes we have a DxilModule if there are dxil ops. Without an initialized DxilModule with the correct low-precision mode, cached ops and types may be wrong.

This change:
- modifies DxilRemoveDeadBlocks to remove bSkipInit
- updates the test to contain the metadata needed to initialize DxilModule

### DIFF
--- a/tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_constant_dce.ll
+++ b/tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_constant_dce.ll
@@ -1,4 +1,4 @@
-; RUN: %opt %s -dxil-remove-dead-blocks -S | FileCheck %s
+; RUN: %opt %s -hlsl-passes-resume -dxil-remove-dead-blocks -S | FileCheck %s
 
 ; Run the pass, to make sure that %val.0 is deleted since its only use is multiplied by 0.
 ;  %val.0 = phi float [ 1.000000e+00, %if.then ], [ 0.000000e+00, %entry ]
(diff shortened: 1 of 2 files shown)


dmpots (approved):
LGTM

pow2clk (approved):
Looks good!

tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_constant_dce.ll
```diff
@@ -102,21 +108,41 @@ attributes #3 = { nounwind readonly }
...
+!41 = !DILocation(line: 20, column: 7, scope: !42)
+!42 = distinct !DILexicalBlock(scope: !4, file: !1, line: 20, column: 7)
+!43 = !DILocation(line: 20, column: 7, scope: !4)
+!44 = !DILocation(line: 21, column: 9, scope: !42)
+!45 = !DILocation(line: 21, column: 5, scope: !42)
+!46 = !DILocation(line: 23, column: 9, scope: !4)
+!47 = !DILocalVariable(tag: DW_TAG_auto_variable, name: "zero", scope: !4, file: !1, line: 23, type: !7)
+!48 = !DILocation(line: 24, column: 19, scope: !4)
+!49 = !DILocation(line: 24, column: 9, scope: !4)
+!50 = !DILocalVariable(tag: DW_TAG_auto_variable, name: "ret", scope: !4, file: !1, line: 24, type: !7)
```
python3kgae, line 147: Do we need debug info for this test?

python3kgae (approved):

adam-yang (approved):

tools/clang/test/HLSLFileCheck/passes/dxil/dxil_remove_dead_pass/delete_constant_dce.ll
tex3d: The original test did, and it's probably good to leave in, since there are debug intrinsic uses, such as `call void @llvm.dbg.value(metadata float %val.0, i64 0, metadata !39, metadata !40), !dbg !38` that should not prevent the dead block from being eliminated.  Besides, it helps for following the IR from the original HLSL scenario.

PR #5921, opened 2023-10-25
Fix no-discard compilation error

### LINKED ISSUE #5922: Compilation errors on github build pipelines.
**Description**
The github pipelines fail to compile DXC on the linux clang debug pipeline due to an error:
error: ignoring return value of function declared with 'nodiscard' attribute [-Werror,-Wunused-result]
      FilesToRemoveRef.back().c_str();

**Steps to Reproduce**
Create a PR with a commit, and trigger the automatic git build pipelines.

**Actual Behavior**
We expect no compilation errors during the build step of DXC.

**Environment**
- DXC version 1.7.0.4181
- Host Operating System Linux

bob80905:
Signals.inc emits an error due to the value of c_str() being discarded, when the function is marked as no-discard.
This PR cherry-picks the solution from the llvm-repo, which removes the c_str() calls, so that no errors are emitted.
Fixes #5922


python3kgae:
Maybe we could merge this from upstream
https://github.com/llvm/llvm-project/commit/4135e4c47528885b7b8967c009cf356a1c0c6b1f

llvm-beanz:
We should also be making this change into `main` then merging it into `staging-sm-6.8`

bob80905:
See: https://github.com/microsoft/DirectXShaderCompiler/pull/5932

PR #5925, opened 2023-10-25
NFC: fix dereferencing of past-end iterator

Keenuts:
This worked because the value was valid by luck, but this relies on an undefined behavior.
However, it is allowed to convert the iterator to the pointer without dereferencing it.

This change will also get upstreamed in LLVM.

### DIFF
--- a/tools/clang/include/clang/Sema/CXXFieldCollector.h
+++ b/tools/clang/include/clang/Sema/CXXFieldCollector.h
@@ -66,7 +66,7 @@ public:
 
   /// getCurFields - Pointer to array of fields added to the currently parsed
   /// class.
-  FieldDecl **getCurFields() { return &*(Fields.end() - getCurNumFields()); }
+  FieldDecl **getCurFields() { return Fields.end() - getCurNumFields(); }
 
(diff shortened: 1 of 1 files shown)


Keenuts:
Upstream PR: https://github.com/llvm/llvm-project/pull/70193
Not sure if you prefer to import this patch once merged or not.

sudonatalie:
We have a convention in DXC of wrapping these changes in `// HLSL Change Begin` and `End`, see https://github.com/microsoft/DirectXShaderCompiler/commit/573b652210793368acdc81cffe50ea7071a4c202 for an example. If this isn't any rush then I'd have a slight preference for landing upstream first just in case there's any changes requested there then we can keep them consistent.

Keenuts:
Turns out this is not an actual issue. The standard does allows 2 specific edge cases, deref of null, and one past-end as long as I don't use it outside of getting its address (see linked PR and https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#232 )

The iterator checker does catches this, because we call the * operator with a bad storage, but the generated code does not do any pointer deref, only pointer arithmetic.

PR #5927, opened 2023-10-25
[SPIR-V] implement vk::spvexecutionmode attribute for inline SPIR-V

cassiebeckley:
Implements `vk::spvexecutionmode` from https://github.com/microsoft/hlsl-specs/blob/main/proposals/0011-inline-spirv.md#execution-modes.

### DIFF
--- a/tools/clang/include/clang/Basic/Attr.td
+++ b/tools/clang/include/clang/Basic/Attr.td
@@ -1142,6 +1142,14 @@ def VKExtensionExt : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
+def VKSpvExecutionMode : InheritableAttr {
+  let Spellings = [CXX11<"vk", "spvexecutionmode">];
+  let Subjects = SubjectList<[Function], ErrorDiag>;
+  let Args = [UnsignedArgument<"ExecutionMode">];
+  let LangOpts = [SPIRV];
+  let Documentation = [Undocumented];
+}
+
 def VKStorageClassExt : InheritableAttr {
   let Spellings = [CXX11<"vk", "ext_storage_class">];
   let Subjects = SubjectList<[Var, ParmVar], ErrorDiag>;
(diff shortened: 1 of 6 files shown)


s-perron (approved):

sudonatalie:
Can you add an error test where the execution mode is some value not defined in SPIRV-Headers?

Since insufficient documentation has been a problem for the existing inline SPIR-V implementation, do we want to write new documentation as we go this time? I think it should be a part of https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst.

s-perron:
> Can you add an error test where the execution mode is some value not defined in SPIRV-Headers?

I may not have looked closely enough. I'm hoping this is not an error. I want vendors to be able to use their extension before the headers have been updated.

Either way, adding a test is a good idea.

> Since insufficient documentation has been a problem for the existing inline SPIR-V implementation, do we want to write new documentation as we go this time? I think it should be a part of https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst.

We should find out how the HLSL spec will be made available. I was expecting that to be the documentation, but I was probably making some assumptions there.

cassiebeckley:
The code as written will allow vendors to use execution modes that are not in SPIRV-Headers (if and only if the validator is disabled with `-Vd`), but adding a test for that is a little tricky. This is because there is no valid disassembly for a SPIR-V module with an undefined execution mode, and in such cases dxc simply does not print the disassembly to stdout. (`-Fo` still behaves as desired, however).

If you try to add an execution mode such as

```
[[vk::spvexecutionmode(999999)]]
void main() {}
```

and validation is enabled, you will get

```
fatal error: generated SPIR-V is invalid: Invalid execution mode operand: 99999
```

which does have the desired execution mode, so I'll just add a test for that error message specifically.

sudonatalie:
That makes sense. In that case you could also just check that compilation we expect to work succeeds with `// RUN: %dxc -T ps_6_0 -E main -fcgl %s -spirv -Vd -Fo -` (no FileCheck), but either seems fine.

sudonatalie (approved):

PR #5930, opened 2023-10-25
Remove integer variants of SampleCmp

### LINKED ISSUE #5564: Remove integer variants of SampleCmp
**Description**
When adding integer sampling support in Shader Model 6.7,  SampleCmp, SampleCmpLevelZero, and SampleCmpLevel were given support for integer resources in spite of the fact that the spec limits filter modes to non-comparison variants. 

**Steps to Reproduce**

The existing test includes integer overloads for these builtins: https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/HLSLFileCheck/hlsl/objects/Texture/sampleInteger.hlsl

Just compiling it with the RUN arguments -T ps_6_7 and -DTYPE=uint (or any integer) will show a successful compile

**Actual Behavior**
DXIL with sampleCmp called on integer resources is generated. It should fail as it does with 6.6

**Environment**
- DXC version :  1.8 - 1.8.0.3959
- Host Operating System win11

python3kgae:
Disable integer resource for SampleCmp/CmpLevel/CmpLevelZero.

Fixes #5564


tools/clang/test/HLSLFileCheck/hlsl/objects/Texture/sampleInteger.hlsl
```diff
@@ -42,10 +32,12 @@ float4 main(float2 coord : TEXCOORD0) : SV_Target {
   res += g_tex0.SampleBias(g_samp, coord, 0.0);
   res += g_tex0.SampleLevel(g_samp, coord, 0.0);
   res += g_tex0.SampleGrad(g_samp, coord, 0.0, 0.0);
+#ifndef PASS
   res += g_tex0.SampleCmp(g_sampCmp, coord, 0.0);
   res += g_tex0.SampleCmpLevelZero(g_sampCmp, coord, 0.0);
 #if  __SHADER_TARGET_MAJOR > 6 || (__SHADER_TARGET_MAJOR == 6 && __SHADER_TARGET_MINOR >= 7)
```
bob80905, line 38: I don't think the line below is ever executed. All tests with SM=6.7 have -D Pass defined, and so this whole block will be skipped. The run lines without PASS defined are all shader model 6.6, so this if condition would never be true. 
Also, I wonder if there's a better more descriptive term? For example, rather than PASS, how about "SKIP_INT_SAMPLE_COMPARES"
Then it can be `#ifdef SKIP_INT_SAMPLE_COMPARES` rather than `#ifndef PASS`

hekota (approved):
I agree with [bob80905](https://github.com/bob80905)'s comment, otherwise LGTM!

tools/clang/test/HLSLFileCheck/hlsl/objects/Texture/sampleInteger.hlsl
python3kgae: Fixed.

bob80905 (approved):

PR #5932, opened 2023-10-25
Remove unnecessary in C++11 c_str() calls

bob80905:
While theoratically required in pre-C++11 to avoid re-allocation upon call, C++11 guarantees that c_str() returns a pointer to the internal array so pre-calling c_str() is no longer required.

llvm-svn: 242983

### DIFF
--- a/lib/Support/Unix/Signals.inc
+++ b/lib/Support/Unix/Signals.inc
@@ -144,9 +144,6 @@ static void RemoveFilesToRemove() {
   // memory.
   std::vector<std::string>& FilesToRemoveRef = *FilesToRemove;
   for (unsigned i = 0, e = FilesToRemoveRef.size(); i != e; ++i) {
-    // We rely on a std::string implementation for which repeated calls to
-    // 'c_str()' don't allocate memory. We pre-call 'c_str()' on all of these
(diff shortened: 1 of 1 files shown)


pow2clk (approved):
Looks clean!

llvm-beanz (approved):

lib/Support/Unix/Signals.inc
```diff
@@ -235,21 +232,7 @@ bool llvm::sys::RemoveFileOnSignal(StringRef Filename,
                                    std::string* ErrMsg) {
   {
     sys::SmartScopedLock<true> Guard(*SignalsMutex);
-    std::vector<std::string>& FilesToRemoveRef = *FilesToRemove;
-    std::string *OldPtr =
-        FilesToRemoveRef.empty() ? nullptr : &FilesToRemoveRef[0];
-    FilesToRemoveRef.push_back(Filename);
-
-    // We want to call 'c_str()' on every std::string in this vector so that if
-    // the underlying implementation requires a re-allocation, it happens here
```
python3kgae, line 244: Should we remove the c_str in sys::DontRemoveFileOnSignal too?

damyanp (approved):

PR #5933, opened 2023-10-25
Remove unnecessary in C++11 c_str() calls (#5932)

bob80905:
While theoratically required in pre-C++11 to avoid re-allocation upon call, C++11 guarantees that c_str() returns a pointer to the internal array so pre-calling c_str() is no longer required.

llvm-svn: 242983

---------


python3kgae (approved):

pow2clk (approved):

adam-yang (approved):

lib/Support/Unix/Signals.inc
```diff
@@ -264,13 +247,6 @@ void llvm::sys::DontRemoveFileOnSignal(StringRef Filename) {
   std::vector<std::string>::iterator I = FilesToRemove->end();
   if (RI != FilesToRemove->rend())
     I = FilesToRemove->erase(RI.base()-1);
```
tex3d, line 249: The reason for iterator `I` is to enable the loop below - without the loop, you don't need `I` at all.

PR #5939, opened 2023-10-27
Add execution mode for derivatives in compute shaders

s-perron:
This commit adds the execution mode to compute shaders when derivatives
are used. I only added a single test because the derivatives are all
handled in one place. The choice of execution mode was tested using the
implicit LOD instructions.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -11163,7 +11163,12 @@ SpirvInstruction *SpirvEmitter::processIntrinsicUsingSpirvInst(
     const CallExpr *callExpr, spv::Op opcode, bool actPerRowForMatrices) {
   // The derivative opcodes are only allowed in pixel shader, or in compute
   // shaderers when the SPV_NV_compute_shader_derivatives is enabled.
-  if (!spvContext.isPS())
+  if (!spvContext.isPS()) {
+    // For cases where the instructions are known to be invalid, we turn on
+    // legalization expecting the invalid use to be optimized away. For compute
+    // shaders, we add the execution mode to enable the derivatives. We legalize
+    // in this case as well because that is what we did before the extension was
+    // used, and we do not want to change previous behaviour too much.
     switch (opcode) {
     case spv::Op::OpDPdx:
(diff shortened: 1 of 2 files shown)


cassiebeckley (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11173,11 +11173,12 @@ SpirvInstruction *SpirvEmitter::processIntrinsicUsingSpirvInst(
     case spv::Op::OpFwidth:
     case spv::Op::OpFwidthFine:
     case spv::Op::OpFwidthCoarse:
+      if (spvContext.isCS())
```
sudonatalie, line 11176: I'm a bit confused. The existing comment above this block looks contradictory:

```
  // Certain opcodes are only allowed in pixel shader
  if (!spvContext.isPS())
```

and for a few of the opcodes I checked, the spec does say "This instruction is only valid in the Fragment Execution Model." ([ref](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_derivative_instructions)) which would make the comment correct and our generated SPIR-V invalid. Correcting it would also void this new addition though. Do you have any more context for how these functions should be used?

s-perron: I forgot to update the comment. The extension allows these opcodes to be used in compute shaders as well. I'll update the comment.

s-perron: I updated the comment to say:

>   // The derivative opcodes are only allowed in pixel shader, or in compute
  // shaderers when the SPV_NV_compute_shader_derivatives is enabled.

Let me know if that makes sense.

sudonatalie: The comment looks correct, but the condition is `if (!spvContext.isPS())` (emphasis on the **!**). Is that an existing bug?

sudonatalie: Got it, thanks for the comment updates, LGTM!

sudonatalie (approved):

PR #5942, opened 2023-10-27
[Linux] Enable PDBUtil test for Linux build

python3kgae:
Added CComBSTR::Length for WinAdapter.

Use VERIFY_ARE_EQUAL_WSTR to verify CComBSTR.

Use UTF8 for Linux test.

This is for #5877

### DIFF
--- a/lib/DxcSupport/WinAdapter.cpp
+++ b/lib/DxcSupport/WinAdapter.cpp
@@ -51,7 +51,15 @@ BSTR SysAllocStringLen(const OLECHAR *strIn, UINT ui) {
 
   return strOut;
 }
+//===--------------------------- BSTR Length ------------------------------===//
+unsigned int SysStringLen(const BSTR bstrString) {
+  if (!bstrString)
+    return 0;
 
+  uint32_t *blobIn = (uint32_t *)((uintptr_t)bstrString - sizeof(uint32_t));
+
+  return blobIn[0] / sizeof(OLECHAR);
+}
 //===---------------------- Char converstion ------------------------------===//
 
 const char *CPToLocale(uint32_t CodePage) {
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -1334,7 +1334,7 @@ static void VerifyPdbUtil(
   {
     CComBSTR str;
     VERIFY_SUCCEEDED(pPdbUtils->GetTargetProfile(&str));
-    VERIFY_ARE_EQUAL_WSTR(str.m_str, L"ps_6_0");
+    VERIFY_ARE_EQUAL_WSTR(L"ps_6_0", str.m_str);
   }
 
   // Entry point
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -1334,7 +1334,7 @@ static void VerifyPdbUtil(
   {
     CComBSTR str;
     VERIFY_SUCCEEDED(pPdbUtils->GetTargetProfile(&str));
-    VERIFY_ARE_EQUAL(0, wcscmp(str.m_str, L"ps_6_0"));
+    VERIFY_ARE_EQUAL_WSTR(str.m_str, L"ps_6_0");
   }
 
   // Entry point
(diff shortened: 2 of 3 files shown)


tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -1335,17 +1334,17 @@ static void VerifyPdbUtil(
   {
     CComBSTR str;
     VERIFY_SUCCEEDED(pPdbUtils->GetTargetProfile(&str));
-    VERIFY_ARE_EQUAL(str, L"ps_6_0");
+    VERIFY_ARE_EQUAL(0, wcscmp(str.m_str, L"ps_6_0"));
```
tex3d, line 1337: We defined a `VERIFY_ARE_EQUAL_STR` for string comparison case in `HlslTestUtils.h`.  `Ref` should be on left, `Src` on right (reversed from order here). That way, `Src` string will be logged when comparison fails.

python3kgae: Updated.

lib/DxcSupport/WinAdapter.cpp
```diff
@@ -51,7 +51,15 @@ BSTR SysAllocStringLen(const OLECHAR *strIn, UINT ui) {
...
   return strOut;
 }
+//===--------------------------- BSTR Length ------------------------------===//
+unsigned int SysStringLen(const BSTR bstrString) {
+  if (!bstrString)
+    return 0;
 
+  uint32_t *blobIn = (uint32_t *)((uintptr_t)bstrString - sizeof(uint32_t));
+
+  return blobIn[0] / sizeof(OLECHAR);
```
pow2clk, line 61: While I don't imagine BSTR is going to be anything but a pointer to OLECHARs any time soon, a slightly less implementation-dependent approach would be to use `sizeof(bstrString[0])`

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -1362,9 +1361,9 @@ static void VerifyPdbUtil(
 
   // Main file name
   {
-    CComBSTR pMainFileName;
-    VERIFY_SUCCEEDED(pPdbUtils->GetMainFileName(&pMainFileName));
-    VERIFY_ARE_EQUAL(pMainFileName, pMainFileName);
```
pow2clk, line 1367: Hah! This test was invalid before! A perk of porting to platforms with more demanding warning checks

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2061,11 +2063,20 @@ TEST_F(CompilerTest, CompileSameFilenameAndEntryThenTestPdbUtilsArgs) {
...
       return 0;
     }
   )x";
+#else
+  std::string shader = R"x(
+    [RootSignature("")] float PSMain() : SV_Target  {
+      return 0;
+    }
+  )x";
+#endif
```
pow2clk, line 2079: I think a simpler solution would be to just use UTF8 for both platforms. The wide char isn't really the point of this test.

pow2clk (commented):
Just a couple suggestions and a remark that I'll resolve myself

lib/DxcSupport/WinAdapter.cpp
python3kgae: We added the length to (uintptr_t)bstrString - sizeof(uint32_t) in https://github.com/microsoft/DirectXShaderCompiler/pull/5942/files#diff-387db224f410dd80e8ea3a8ebec09badd12bcecc88132f28e816f9210e209a86R42

tools/clang/unittests/HLSL/CompilerTest.cpp
python3kgae: Updated.

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -1335,17 +1334,17 @@ static void VerifyPdbUtil(
   {
     CComBSTR str;
     VERIFY_SUCCEEDED(pPdbUtils->GetTargetProfile(&str));
-    VERIFY_ARE_EQUAL(str, L"ps_6_0");
+    VERIFY_ARE_EQUAL_WSTR(str.m_str, L"ps_6_0");
```
tex3d, line 1337: Swapping order will result in a message that prints the `str.m_str` when this fails instead of re-printing the reference string.

python3kgae: Missed one :(

```suggestion
    VERIFY_ARE_EQUAL_WSTR(L"ps_6_0", str.m_str);
```

tex3d (approved):
LGTM!

pow2clk (approved):
Looks good!

PR #5943, opened 2023-10-27
Fix RDAT table versioning for ver 1.8

### LINKED ISSUE #5900: SM 6.8 RDAT has unused shader data that should be removed
**Description**
Shader data in `RuntimeDataFunctionInfo2` ended up being unused and needs to be removed before release.

Not only is this info not completely filled in (ViewID/Dependency maps, and more), but it doesn't contain everything that would be required to replace PSV0, I/O signature, and root signature parts for ordinary shader stages either.

It should not be left in for SM 6.8 release, otherwise it will have to be supported going forward.

We need to retain `MinimumExpectedWaveLaneCount`, `MaximumExpectedWaveLaneCount`, and `ShaderFlags`.  However, `DxilShaderFlags` needs to be reduced to the one flag that's still used: `NodeProgramEntry`.

tex3d:
- Reorganize part and table IDs to move shader info after ver 1.8
- Move NodeShaderInfo into RuntimeDataFunctionInfo2 shader info union, since it is mutually exclusive with other shader types
- Remove RuntimeDataFunctionInfo3
- Re-organize RDAT_LibraryTypes into 3 sections by versioning
- Format unformatted section of RDAT_LibraryTypes due to typo
- Add indentation to make definitions easier to read and disable clang-format for RDAT_LibraryTypes.inl and RDAT_SubobjectTypes.inl
- Include Node in ShaderStageFlags for 1.8+
- Added some explicit scopes due to VS IDE thinking they were ambiguous
- RDATDumper: Don't try to visit or print null records
- Fix reflection tests:
    - tests meant to check shader info use: -Vd -validator-version 0.0
    - use regex for record strides for versioning robustness
    - update ShaderStageFlags to include Node
    - use regex for RuntimeDataFunctionInfo name for versioning robustness
    - remove shader info CHECKs for cases not meant to test that

Fixes #5900

### DIFF
--- a/include/dxc/DxilContainer/DxilRuntimeReflection.h
+++ b/include/dxc/DxilContainer/DxilRuntimeReflection.h
@@ -62,6 +62,16 @@ enum class RuntimeDataPartType : uint32_t {
   Last_1_4 = SubobjectTable,
 
   // PRERELEASE-TODO: assign values explicitly to all enums before release
+  NodeIDTable = 7,
+  NodeShaderIOAttribTable = 8,
+  NodeShaderFuncAttribTable = 9,
+  IONodeTable = 10,
+  NodeShaderInfoTable = 11,
+
+  Last_1_8 = NodeShaderInfoTable, // PRERELEASE-TODO: change to last
+                                  // necessary 1.8 part before release.
+
+  // Insert experimental here.
   SignatureElementTable,
   VSInfoTable,
   PSInfoTable,
@@ -71,16 +81,6 @@ enum class RuntimeDataPartType : uint32_t {
   CSInfoTable,
   MSInfoTable,
   ASInfoTable,
-
-  NodeIDTable,
-  NodeShaderIOAttribTable,
-  NodeShaderFuncAttribTable,
-  IONodeTable,
-  NodeShaderInfoTable,
-
-  Last_1_8 = NodeShaderInfoTable, // PRERELEASE-TODO: change to last
-                                  // necessary 1.8 part before release.
-  // Insert experimental here.
 
   LastPlus1,
   LastExperimental = LastPlus1 - 1,
@@ -110,15 +110,6 @@ enum class RecordTableIndex : unsigned {
   FunctionTable,
   SubobjectTable,
 
-  SignatureElementTable,
-  VSInfoTable,
-  PSInfoTable,
-  HSInfoTable,
-  DSInfoTable,
-  GSInfoTable,
-  CSInfoTable,
-  MSInfoTable,
-  ASInfoTable,
   NodeIDTable,
   NodeShaderIOAttribTable,
   NodeShaderFuncAttribTable,
@@ -128,6 +119,16 @@ enum class RecordTableIndex : unsigned {
   DxilPdbInfoTable,
   DxilPdbInfoSourceTable,
   DxilPdbInfoLibraryTable,
+
+  SignatureElementTable,
+  VSInfoTable,
+  PSInfoTable,
+  HSInfoTable,
+  DSInfoTable,
+  GSInfoTable,
+  CSInfoTable,
+  MSInfoTable,
+  ASInfoTable,
 
   RecordTableCount
 };
--- a/include/dxc/DxilContainer/DxilRuntimeReflection.h
+++ b/include/dxc/DxilContainer/DxilRuntimeReflection.h
@@ -61,15 +61,12 @@ enum class RuntimeDataPartType : uint32_t {
   SubobjectTable = 6,
   Last_1_4 = SubobjectTable,
 
-  // PRERELEASE-TODO: assign values explicitly to all enums before release
   NodeIDTable = 7,
   NodeShaderIOAttribTable = 8,
   NodeShaderFuncAttribTable = 9,
   IONodeTable = 10,
   NodeShaderInfoTable = 11,
-
-  Last_1_8 = NodeShaderInfoTable, // PRERELEASE-TODO: change to last
-                                  // necessary 1.8 part before release.
+  Last_1_8 = NodeShaderInfoTable,
 
   // Insert experimental here.
   SignatureElementTable,
--- a/include/dxc/DxilContainer/RDAT_LibraryTypes.inl
+++ b/include/dxc/DxilContainer/RDAT_LibraryTypes.inl
@@ -24,40 +24,6 @@ RDAT_ENUM_START(DxilResourceFlag, uint32_t)
   RDAT_ENUM_VALUE(Atomics64Use, 1 << 4)
 RDAT_ENUM_END()
 
-RDAT_ENUM_START(DxilShaderFlags, uint32_t)
-  RDAT_ENUM_VALUE(None, 0)
-  RDAT_ENUM_VALUE(NodeProgramEntry, 1 << 0)
-  // End of values supported by validator version 1.8
-  RDAT_ENUM_VALUE(OutputPositionPresent, 1 << 1)
-  RDAT_ENUM_VALUE(DepthOutput, 1 << 2)
-  RDAT_ENUM_VALUE(SampleFrequency, 1 << 3)
-  RDAT_ENUM_VALUE(UsesViewID, 1 << 4)
-RDAT_ENUM_END()
-
-RDAT_ENUM_START(NodeFuncAttribKind, uint32_t)
-  RDAT_ENUM_VALUE(None, 0)
-  RDAT_ENUM_VALUE(ID, 1)
-  RDAT_ENUM_VALUE(NumThreads, 2)
-  RDAT_ENUM_VALUE(ShareInputOf, 3)
-  RDAT_ENUM_VALUE(DispatchGrid, 4)
-  RDAT_ENUM_VALUE(MaxRecursionDepth, 5)
-  RDAT_ENUM_VALUE(LocalRootArgumentsTableIndex, 6)
-  RDAT_ENUM_VALUE(MaxDispatchGrid, 7)
-  RDAT_ENUM_VALUE_NODEF(LastValue)
-RDAT_ENUM_END()
-
-RDAT_ENUM_START(NodeAttribKind, uint32_t)
-  RDAT_ENUM_VALUE(None, 0)
-  RDAT_ENUM_VALUE(OutputID, 1)
-  RDAT_ENUM_VALUE(MaxRecords, 2)
-  RDAT_ENUM_VALUE(MaxRecordsSharedWith, 3)
-  RDAT_ENUM_VALUE(RecordSizeInBytes, 4)
-  RDAT_ENUM_VALUE(RecordDispatchGrid, 5)
-  RDAT_ENUM_VALUE(OutputArraySize, 6)
-  RDAT_ENUM_VALUE(AllowSparseNodes, 7)
-  RDAT_ENUM_VALUE_NODEF(LastValue)
-RDAT_ENUM_END()
-
 #endif // DEF_RDAT_ENUMS
 
 #ifdef DEF_DXIL_ENUMS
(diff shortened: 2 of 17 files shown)


tex3d:
Note: this change breaks runtime compatibility for validator version 1.8 (unreleased).

tools/clang/test/HLSLFileCheck/d3dreflect/mesh-groupshared.hlsl
```diff
@@ -1,11 +1,11 @@
-// RUN: %dxilver 1.8 | %dxc -T lib_6_7 %s | %D3DReflect %s | FileCheck %s
+// RUN: %dxilver 1.8 | %dxc -T lib_6_7 -Vd -validator-version 0.0 %s | %D3DReflect %s | FileCheck %s
```
python3kgae, line 1: Should we have another RUN line test validator 1.7?

include/dxc/DxilContainer/RDAT_SubobjectTypes.inl
```diff
@@ -74,99 +82,104 @@ RDAT_ENUM_END()
...
-RDAT_BYTES(Data)
+  RDAT_BYTES(Data)
 RDAT_STRUCT_END()
 #undef RECORD_TYPE
 
 #define RECORD_TYPE SubobjectToExportsAssociation_t
 RDAT_STRUCT(SubobjectToExportsAssociation_t)
-RDAT_STRING(Subobject)
-RDAT_STRING_ARRAY_REF(Exports)
+  RDAT_STRING(Subobject)
```
python3kgae, line 97: Why do we need those space only changes? Clang format?

python3kgae:
There're so many things in this PR, is it possible to split it into smaller PRs?

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1591,7 +1587,7 @@ class DxilRDATWriter : public DxilPartWriter {
 
   uint32_t AddShaderNodeInfo(const DxilModule &DM, llvm::Function &function,
                              const DxilEntryProps &entryProps,
-                             RuntimeDataFunctionInfo3 &funcInfo,
+                             RuntimeDataFunctionInfo2 &funcInfo,
```
python3kgae, line 1590: Where is RuntimeDataFunctionInfo3 defined? Should we remove the definition of RuntimeDataFunctionInfo3 too?

tools/clang/test/HLSLFileCheck/d3dreflect/mesh-groupshared.hlsl
tex3d: No, this is data that is only in the additional RDAT beyond 1.8.  The whole purpose of this test is to test that data, so the test requires this validator version to allow the data to be set in the RDAT.  Same story with other ones I added this to.

include/dxc/DxilContainer/RDAT_SubobjectTypes.inl
tex3d: We don't really need them.  The formatting was messed up for `RDAT_LibraryTypes.inl` due to typo, but fixing that results in flattening the entire file, which was even harder to read.  I thought it would be better to clang-format to wrap lines, but to indent for readability, as a compromise.  I couldn't find any other way to get clang-format to allow me to indent these for readability.

Since I changed that file this way, I took a look at the other files, and noticed this one was also harder to read from clang-format flattening, so I did the same thing here.  It's not necessary, and I could revert, but I'd rather have it this way, if that's acceptable to others.

lib/DxilContainer/DxilContainerAssembler.cpp
tex3d: It has been removed from the `RDAT_LibraryTypes.inl` where it was defined.  See deleted definition here:
https://github.com/microsoft/DirectXShaderCompiler/pull/5943/files/258e3f19ccab8788b6e80f0f620dbaa826c5a6f2#diff-2ea78813faaf946be274cbe2d57a3e174718a09199d9f90e5563aa713fa334adL552

include/dxc/DxilContainer/RDAT_SubobjectTypes.inl
python3kgae: It is OK to do the change for readability.
But it would be nice to do that in separate PR so it is easier to review.

tex3d:
> There're so many things in this PR, is it possible to split it into smaller PRs?

That would be a lot more work, I think.  And more to review.  For instance, trying to break up `RDAT_LibraryTypes.inl` changes reveals each dependent on the priro: typo `clang-format onm` caused most of the file to skip clang-formatting, re-indenting and disabling clang-format around the contents, re-organizing all the contents to make clear version sequence, merging Info2 and Info3.  Each step would conflict with each other, so cannot be done independently.  And this is probably the biggest source of the problem you're identifying.

Doing separate steps now would also be quite a pain due to interdependencies of different parts of the change.  For instance, tests have several changes at once to deal with these changes: ShaderStageFlags update, 1.8 version excluding shader info, Merged Info2 and Info3, stride regex for better versioning robustness.  If other changes were broken up, I'd have to break up these changes into multiple stages as well.  This will be a lot more work than updating for all the changes in one go.

If it helps, there are separate commits you can look at for different phases, but these separate commits aren't separable into independent PRs as is either (tests would break at the very least).

python3kgae:
> > There're so many things in this PR, is it possible to split it into smaller PRs?
> 
> That would be a lot more work, I think. And more to review. For instance, trying to break up `RDAT_LibraryTypes.inl` changes reveals each dependent on the priro: typo `clang-format onm` caused most of the file to skip clang-formatting, re-indenting and disabling clang-format around the contents, re-organizing all the contents to make clear version sequence, merging Info2 and Info3. Each step would conflict with each other, so cannot be done independently. And this is probably the biggest source of the problem you're identifying.
> 
> Doing separate steps now would also be quite a pain due to interdependencies of different parts of the change. For instance, tests have several changes at once to deal with these changes: ShaderStageFlags update, 1.8 version excluding shader info, Merged Info2 and Info3, stride regex for better versioning robustness. If other changes were broken up, I'd have to break up these changes into multiple stages as well. This will be a lot more work than updating for all the changes in one go.
> 
> If it helps, there are separate commits you can look at for different phases, but these separate commits aren't separable into independent PRs as is either (tests would break at the very least).

Yeah. It is really hard to understand change in RDAT_LibraryTypes.inl with format change mixed with real change.
Is it possible to do the format change in a separate PR first, then, rebase current PR after the format PR is in?

tools/clang/test/HLSLFileCheck/d3dreflect/empty_broadcasting_nodes.hlsl
```diff
@@ -4,8 +4,8 @@
 // CHECK:  StringBuffer (size = {{[0-9]+}} bytes)
```
pow2clk, line 4: Should this one get `-Vd -validator-version 0.0` too?

tools/clang/test/HLSLFileCheck/d3dreflect/empty_thread_nodes.hlsl
```diff
@@ -5,8 +5,8 @@
 // CHECK:  StringBuffer (size = {{[0-9]+}} bytes)
```
pow2clk, line 5: Here too?

include/dxc/DxilContainer/DxilRuntimeReflection.h
```diff
@@ -62,6 +62,16 @@ enum class RuntimeDataPartType : uint32_t {
...
 
   // PRERELEASE-TODO: assign values explicitly to all enums before release
+  NodeIDTable = 7,
+  NodeShaderIOAttribTable = 8,
+  NodeShaderFuncAttribTable = 9,
+  IONodeTable = 10,
+  NodeShaderInfoTable = 11,
+
+  Last_1_8 = NodeShaderInfoTable, // PRERELEASE-TODO: change to last
+                                  // necessary 1.8 part before release.
```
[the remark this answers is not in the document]
tex3d: Thanks, I will clean up these comments.

include/dxc/DxilContainer/DxilRuntimeReflection.h
```diff
@@ -129,6 +120,16 @@ enum class RecordTableIndex : unsigned {
   DxilPdbInfoSourceTable,
   DxilPdbInfoLibraryTable,
```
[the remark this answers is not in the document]
tex3d: Values in this enum are only used internally, so they need not be stable between builds or versions.  The values in the `RuntimeDataPartType` enum are part of the protocol and are fixed for shipped versions.

I think the idea was to leave these at the end of all the tables used for runtime, once the additional tables are removed.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1686,9 +1682,12 @@ class DxilRDATWriter : public DxilPartWriter {
     // We must select the appropriate shader mask for the validator version,
     // so we don't set any bits the validator doesn't recognize.
     unsigned ValidShaderMask =
-        (1 << ((unsigned)DXIL::ShaderKind::Amplification + 1)) - 1;
+        (1 << ((unsigned)DXIL::ShaderKind::Node + 1)) - 1;
```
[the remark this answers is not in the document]
tex3d: Yeah, I could add enum values to `ShaderKind` to capture these in a safer way.

include/dxc/DxilContainer/RDAT_LibraryTypes.inl
```diff
@@ -9,48 +9,19 @@
...
-RDAT_ENUM_START(NodeAttribKind, uint32_t)
-RDAT_ENUM_VALUE(None, 0)
-RDAT_ENUM_VALUE(OutputID, 1)
-RDAT_ENUM_VALUE(MaxRecords, 2)
-RDAT_ENUM_VALUE(MaxRecordsSharedWith, 3)
-RDAT_ENUM_VALUE(RecordSizeInBytes, 4)
-RDAT_ENUM_VALUE(RecordDispatchGrid, 5)
-RDAT_ENUM_VALUE(OutputArraySize, 6)
-RDAT_ENUM_VALUE(AllowSparseNodes, 7)
-RDAT_ENUM_VALUE_NODEF(LastValue)
```
pow2clk, line 53: These three enums were moved to the section requiring 1.8 or above around line 174

include/dxc/DxilContainer/RDAT_LibraryTypes.inl
```diff
@@ -65,175 +36,65 @@ RDAT_ENUM_END()
...
-  RDAT_ENUM_VALUE_NODEF(ViewID)
-  RDAT_ENUM_VALUE_NODEF(Barycentrics)
-  RDAT_ENUM_VALUE_NODEF(ShadingRate)
-  RDAT_ENUM_VALUE_NODEF(CullPrimitive)
-  RDAT_ENUM_VALUE_NODEF(Invalid)
-  // SemanticKind-ENUM:END
-RDAT_ENUM_END()
-
-RDAT_DXIL_ENUM_START(hlsl::DXIL::ComponentType, uint32_t)
   RDAT_ENUM_VALUE_NODEF(Invalid)
```
pow2clk, line 51: I think this one mistaken match makes the diff actively unhelpful in trying to review this. If only the whitespace change could be separated from the rest, I think this would be a lot easier to review. I know it's some additional work, but I think it needs to be done in order to review this already tricky file. I'm inclined to do it myself. Is that okay?

pow2clk (commented):
First try at reviewing RDAT_LibraryTypes and I gave up :(

include/dxc/DxilContainer/RDAT_LibraryTypes.inl
pow2clk: Just a comment to try to help reviewers. Can be resolved by anyone.

include/dxc/DxilContainer/RDAT_LibraryTypes.inl
tex3d: This is also a problem because I reordered definitions.  I can try to separate changes into three phases: whitespace only, then reorder top-level definitions, then the modification of enum value order and RuntimeDataFunctionInfo2.  This way the first two should have no functional impact (NFC), while the last contains all the functional impact.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1788,14 +1782,23 @@ class DxilRDATWriter : public DxilPartWriter {
             payloadSizeInBytes = props.ShaderProps.Ray.paramSizeInBytes;
           }
           shaderKind = (uint32_t)props.shaderKind;
-          if (pInfo2 && DM.HasDxilEntryProps(&function)) {
+          if (DM.HasDxilEntryProps(&function)) {
```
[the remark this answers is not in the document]
tex3d: Ah, yes, looking at the final change, it looks odd.  This is an artifact of a change in an intermediate version that separated info2 and info3.  I'll clean that up.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1788,14 +1782,23 @@ class DxilRDATWriter : public DxilPartWriter {
...
+              unsigned waveSize = entryProps.props.waveSize;
+              if (waveSize) {
+                pInfo2->MinimumExpectedWaveLaneCount = waveSize;
+                pInfo2->MaximumExpectedWaveLaneCount = waveSize;
+              }
+              pInfo2->ShaderFlags = 0;
+              if (entryProps.props.IsNode()) {
+                shaderInfo = AddShaderNodeInfo(DM, function, entryProps,
+                                               *pInfo2, TGSMInFunc[&function]);
+              } else if (DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 8) >
```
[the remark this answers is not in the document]
tex3d: Yes it's intentional.  AddShaderInfo adds the (now experimental) info we can do without for 1.8, since we capture the wave size and shader flags outside that function now.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1804,8 +1807,6 @@ class DxilRDATWriter : public DxilPartWriter {
         info.ShaderKind = shaderKind;
         if (pInfo2)
           pInfo2->RawShaderRef = shaderInfo;
-        if (pInfo3)
-          pInfo3->Node = nodeInfo;
```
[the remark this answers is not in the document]
tex3d: Since other shader info is mutually exclusive with node shader info, it made sense to integrate node and the other shader info types into the same union in info2.  Then there's no need for any info3 version.  This is a modification from the original plan that I realized made even more sense once I was modifying it.

tools/clang/test/HLSLFileCheck/d3dreflect/empty_broadcasting_nodes.hlsl
tex3d: No, this tests the node shader data, which must be valid when targeting lib_6_8 (validator version 1.8 required).

tools/clang/test/HLSLFileCheck/d3dreflect/empty_thread_nodes.hlsl
tex3d: Same answer.

include/dxc/DxilContainer/RDAT_LibraryTypes.inl
```diff
@@ -304,129 +167,101 @@ RDAT_STRUCT_TABLE(RuntimeDataFunctionInfo, FunctionTable)
...
+#endif // DEF_DXIL_ENUMS
+
+#ifdef DEF_RDAT_TYPES
 
 #define RECORD_TYPE RecordDispatchGrid
 RDAT_STRUCT(RecordDispatchGrid)
   RDAT_VALUE(uint16_t, ByteOffset)
-  RDAT_VALUE(uint16_t, ComponentNumAndType) // 0:2 = NumComponents (0-3), 3:15 = hlsl::DXIL::ComponentType enum
+  RDAT_VALUE(uint16_t, ComponentNumAndType) // 0:2 = NumComponents (0-3), 3:15 =
+                                          // hlsl::DXIL::ComponentType enum
```
pow2clk, line 255: Another formatting change you might reconsider

include/dxc/DxilContainer/RDAT_LibraryTypes.inl
```diff
@@ -438,67 +273,94 @@ RDAT_STRUCT_TABLE(NodeID, NodeIDTable)
...
       RDAT_RECORD_VALUE(RecordDispatchGrid, RecordDispatchGrid)
-    RDAT_UNION_ELIF(OutputArraySize, getAttribKind() == hlsl::RDAT::NodeAttribKind::OutputArraySize)
+    RDAT_UNION_ELIF(OutputArraySize,
+                    getAttribKind() ==
+                        hlsl::RDAT::NodeAttribKind::OutputArraySize)
       RDAT_VALUE(uint32_t, OutputArraySize)
-    RDAT_UNION_ELIF(AllowSparseNodes, getAttribKind() == hlsl::RDAT::NodeAttribKind::AllowSparseNodes)
+    RDAT_UNION_ELIF(AllowSparseNodes,
+                    getAttribKind() ==
+                        hlsl::RDAT::NodeAttribKind::AllowSparseNodes)
```
pow2clk, line 339: This run of multiline macros makes me wonder if formatting could be reconsidered here too.

pow2clk (approved):
In the end, I was surprised by how few functional changes there are to RDAT_LibraryTypes. The only way I was able to make sense of it was to create some commits of my own that first revert the reordering and whitespace changes and then restore them. I'm willing to push them to this PR, but for now they can be accessed here:  
https://github.com/pow2clk/DirectXShaderCompiler/tree/tex-fix-rdat-versions

pow2clk:
I have a few formatting suggestions you might consider, but you can consider all of my comments resolvable when you're ready to merge

python3kgae (approved):

include/dxc/DxilContainer/RDAT_LibraryTypes.inl
```diff
@@ -538,22 +409,248 @@ RDAT_STRUCT_TABLE_DERIVED(RuntimeDataFunctionInfo2, RuntimeDataFunctionInfo, Fun
...
+  RDAT_ENUM(uint8_t, hlsl::DXIL::InterpolationMode, InterpolationMode)
+  RDAT_VALUE(
+      uint8_t,
+      StartRow) // Starting row of packed location if allocated, otherwise 0xFF
+  // TODO: use struct with bitfields or accessors for ColsAndStream and
+  // UsageAndDynIndexMasks
+  RDAT_VALUE(uint8_t, ColsAndStream) // 0:2 = (Cols-1) (0-3), 2:4 = StartCol
+                                     // (0-3), 4:6 = OutputStream (0-3)
+  RDAT_VALUE(uint8_t,
+             UsageAndDynIndexMasks) // 0:4 = UsageMask, 4:8 = DynamicIndexMask
```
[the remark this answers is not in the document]
tex3d: While I disabled clang-format, I still wanted to have lines wrapped by clang-format to preserve the value of column limits.  So I actually manually reformatted lines as I went along after indentation, which worked in most cases, but had a few tricky ones where clang-format wouldn't respect the indentation I had, even though I was doing a single-line reformat.  The main thing I wanted for readability was the indentation for nested elements.  Flattening indentation made it hard to see where one thing ended and another began.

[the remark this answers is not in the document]
tex3d: At this point, I don't think I want to make additional formatting changes.  I'd rather not delay merge any more than I have to now.

PR #5946, opened 2023-10-28
Add EntryFunctionName to PSV0 data

### LINKED ISSUE #5944: Add entry name to PSV0 data for runtime visibility
Upcoming generic program support in D3D12 State Object API relies on function names to put together programs.  Since we are not supporting library targets for graphics and compute shaders, we must supply these through ordinary non-library shader blobs.  These shader blobs have no visibility of the entry name that the runtime can see without cracking open the DXIL bitcode.  This prevents you from referencing the shaders when constructing the generic program state object.

We need to add an entry name field to the PSV0 data that the runtime consumes so it knows the function name being added to a collection, which can then be referred to when constructing a generic program state object.

Implementation details are as follows:
- In `DxilPipelineStateValidation.h`, add `PSVRuntimeInfo3` derived from `PSVRuntimeInfo2`, containing one `uint32_t EntryFunctionName` member.
- Update `MAX_PSV_VERSION` and `RuntimeInfoSize()` for new version.
- Add pointer to `DxilPipelineStateValidation` and assign in `ReadOrWrite` along with other versioned pointers.
- Add `DxilPipelineStateValidation::GetEntryFunctionName()` accessor to look up name in string table.
- Update PSV0 versioning to key off validator version 1.8 in `DxilContainerAssembler.cpp`
- Add entry name to string table and set `EntryFunctionName` to offset.

1. Default names based on the shader type, such as `VSMain` or `PSMain` can be assigned, and will be when this entry name is unavailable, but this limits you to a single shader of each type upon import to a collection.  This severely limits the capabilities of collections.
2. Add a symbol table to DxilContainer, formatted after a part of some standard format we may want to use in the future.  There is a [proposal on hlsl-specs](https://github.com/microsoft/hlsl-specs/pull/110) for this.  I have extensively written about my reservations with this proposal in comments there.

tex3d:
- In `DxilPipelineStateValidation.h`, add `PSVRuntimeInfo3` derived from `PSVRuntimeInfo2`, containing one `uint32_t EntryFunctionName` member.
- Update `MAX_PSV_VERSION` and `RuntimeInfoSize()` for new version.
- Add pointer to `DxilPipelineStateValidation` and assign in `ReadOrWrite` along with other versioned pointers.
- Add `DxilPipelineStateValidation::GetEntryFunctionName()` accessor to look up name in string table.
- Update PSV0 versioning to key off validator version 1.8 in `DxilContainerAssembler.cpp`
- Add entry name to string table and set `EntryFunctionName` to offset.
- Add initialization to DxilPSVWriter members
- Add asserts in DxilPSVWriter::write() to ensure buffer sizes not changed after m_PSVInitInfo is set.
- Add test to DxilContainerTest, adding/updating common functions as needed

Note: Testing the HS passthrough case revealed problems blocking that scenario.
Filed issue #6001 to track this.

Fixes #5944

### DIFF
--- a/lib/DxilContainer/DxilContainerAssembler.cpp
+++ b/lib/DxilContainer/DxilContainerAssembler.cpp
@@ -678,16 +678,17 @@ unsigned hlsl::LoadViewIDStateFromPSV(unsigned *pOutputData,
 class DxilPSVWriter : public DxilPartWriter {
 private:
   const DxilModule &m_Module;
-  unsigned m_ValMajor, m_ValMinor;
+  unsigned m_ValMajor = 0, m_ValMinor = 0;
   PSVInitInfo m_PSVInitInfo;
   DxilPipelineStateValidation m_PSV;
-  uint32_t m_PSVBufferSize;
+  uint32_t m_PSVBufferSize = 0;
   SmallVector<char, 512> m_PSVBuffer;
   SmallVector<char, 256> m_StringBuffer;
   SmallVector<uint32_t, 8> m_SemanticIndexBuffer;
   std::vector<PSVSignatureElement0> m_SigInputElements;
   std::vector<PSVSignatureElement0> m_SigOutputElements;
   std::vector<PSVSignatureElement0> m_SigPatchConstOrPrimElements;
+  unsigned EntryFunctionName = 0;
 
   void SetPSVSigElement(PSVSignatureElement0 &E,
                         const DxilSignatureElement &SE) {
@@ -758,6 +759,9 @@ public:
     else if (PSVVersion > 1 &&
              DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 6) < 0)
       m_PSVInitInfo.PSVVersion = 1;
+    else if (PSVVersion > 2 &&
+             DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 8) < 0)
+      m_PSVInitInfo.PSVVersion = 2;
     else if (PSVVersion > MAX_PSV_VERSION)
       m_PSVInitInfo.PSVVersion = MAX_PSV_VERSION;
 
@@ -794,6 +798,16 @@ public:
       for (auto &SE : m_Module.GetPatchConstOrPrimSignature().GetElements()) {
         SetPSVSigElement(m_SigPatchConstOrPrimElements[i++], *(SE.get()));
       }
+
+      // Add entry function name to string table in version 3 and above.
+      if (m_PSVInitInfo.PSVVersion > 2) {
+        EntryFunctionName = (uint32_t)m_StringBuffer.size();
+        StringRef Name(m_Module.GetEntryFunctionName());
+        m_StringBuffer.append(Name.size() + 1, '\0');
+        memcpy(m_StringBuffer.data() + EntryFunctionName, Name.data(),
+               Name.size());
+      }
+
       // Set String and SemanticInput Tables
       m_PSVInitInfo.StringTable.Table = m_StringBuffer.data();
       m_PSVInitInfo.StringTable.Size = m_StringBuffer.size();
@@ -821,6 +835,9 @@ public:
   uint32_t size() const override { return m_PSVBufferSize; }
 
   void write(AbstractMemoryStream *pStream) override {
+    // Do not add any data in write() which wasn't accounted for already in the
+    // constructor, where we compute the size based on m_PSVInitInfo.
+
     m_PSVBuffer.resize(m_PSVBufferSize);
     if (!m_PSV.InitNew(m_PSVInitInfo, m_PSVBuffer.data(), &m_PSVBufferSize)) {
       DXASSERT(false, "PSV InitNew failed!");
@@ -831,9 +848,13 @@ public:
     PSVRuntimeInfo0 *pInfo = m_PSV.GetPSVRuntimeInfo0();
     PSVRuntimeInfo1 *pInfo1 = m_PSV.GetPSVRuntimeInfo1();
     PSVRuntimeInfo2 *pInfo2 = m_PSV.GetPSVRuntimeInfo2();
+    PSVRuntimeInfo3 *pInfo3 = m_PSV.GetPSVRuntimeInfo3();
     const ShaderModel *SM = m_Module.GetShaderModel();
     pInfo->MinimumExpectedWaveLaneCount = 0;
     pInfo->MaximumExpectedWaveLaneCount = (UINT)-1;
+
+    if (pInfo3)
+      pInfo3->EntryFunctionName = EntryFunctionName;
 
     switch (SM->GetKind()) {
     case ShaderModel::Kind::Vertex: {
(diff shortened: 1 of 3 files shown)


lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -831,10 +834,19 @@ class DxilPSVWriter : public DxilPartWriter {
     PSVRuntimeInfo0 *pInfo = m_PSV.GetPSVRuntimeInfo0();
     PSVRuntimeInfo1 *pInfo1 = m_PSV.GetPSVRuntimeInfo1();
     PSVRuntimeInfo2 *pInfo2 = m_PSV.GetPSVRuntimeInfo2();
+    PSVRuntimeInfo3 *pInfo3 = m_PSV.GetPSVRuntimeInfo3();
     const ShaderModel *SM = m_Module.GetShaderModel();
     pInfo->MinimumExpectedWaveLaneCount = 0;
     pInfo->MaximumExpectedWaveLaneCount = (UINT)-1;
 
+    if (pInfo3) {
+      pInfo3->EntryFunctionName = (uint32_t)m_StringBuffer.size();
```
bob80905, line 843: At what point does the string table get set so that this uint32 that is being assigned here will correspond to the actual entry function name? Also, is it possible that the key value pair in the string table could be overwritten by another string of the same length? I don't understand how we can reliably get the entry point name from the string table at this point. It seems like m_Module has already built a string table that contains the a set of key (uint32_t) value (char*) pairs, but the key is just the length of the string, and if there are multiple entries, that seems very risky.

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -1154,6 +1155,60 @@ TEST_F(DxilContainerTest, CompileGS_CheckPSV0_ViewID) {
...
+    VERIFY_ARE_EQUAL_STR(EntryPointName, PSV.GetEntryFunctionName());
+  };
+
+  const char src_PSMain[] = "float MyPSMain() : SV_Target { return 0; }";
+  const char src_VSMain[] = "float4 MyVSMain() : SV_Position { return 0; }";
+  const char src_CSMain[] = "[numthreads(1,1,1)] void MyCSMain() {}";
+
+  CheckEntryFunctionName("MyPSMain", src_PSMain, "ps_6_0");
+  CheckEntryFunctionName("MyVSMain", src_VSMain, "vs_6_0");
+  CheckEntryFunctionName("MyCSMain", src_CSMain, "cs_6_0");
```
bob80905, line 1209: Would it be possible to test for library shaders where no entry point exists? Would it be impossible to get a PSV3 object from a library? Maybe we could test that like this in the library case:
```
PSVRuntimeInfo3 *pInfo3 = PSV.GetPSVRuntimeInfo3();
    VERIFY_IS_NULL(pInfo3); // IS NULL
```

lib/DxilContainer/DxilContainerAssembler.cpp
tex3d: > At what point does the string table get set so that this uint32 that is being assigned here will correspond to the actual entry function name?

Right here.  The current size of the string buffer is the location we are about to place this string in the code below this line. This is exactly the same pattern used for the semantic names here:

https://github.com/microsoft/DirectXShaderCompiler/blob/05094636736eaf10f964545a6873d4b31c06d739/lib/DxilContainer/DxilContainerAssembler.cpp#L697

> is it possible that the key value pair in the string table could be overwritten by another string of the same length

What key value pair?  I think you might be misinterpreting the code.  There are no key value pairs in this code.  The string table is just a `SmallVector<char, 256>` that will be written as-is to the container.  Offsets into the string table are byte offsets into this buffer.  A reader must identify where the string table is in the PSV0 container and offset into that location when looking at the offset written to the EntryFunctionName field.

https://github.com/microsoft/DirectXShaderCompiler/blob/05094636736eaf10f964545a6873d4b31c06d739/lib/DxilContainer/DxilContainerAssembler.cpp#L686

A string table could be like: `"\0FirstString\0Second\0"`.  The value written for `"FirstString"` is the offset `1`.  The value written for the `"Second"` is the offset `13`.

Here is how the offset is looked up in the string table by the reader:
https://github.com/microsoft/DirectXShaderCompiler/blob/05094636736eaf10f964545a6873d4b31c06d739/include/dxc/DxilContainer/DxilPipelineStateValidation.h#L215

tools/clang/unittests/HLSL/DxilContainerTest.cpp
tex3d: `PSV0` is only used for non-library shaders.  For libraries, `RDAT` replaces `PSV0` for supplying runtime data.

Also, `PSV3` isn't a thing.  That `0` in `PSV0` is a higher level versioning for the part protocol itself, in case we needed to make a breaking change to that protocol.  There's no need to version that at this point.

tex3d: I guess `PSV3` was just shorthand for `PSVRuntimeInfo3` though...

tex3d: The test case is not relevant, because there's no `PSV0` part to load into PSV in the first place.  This change doesn't affect any of that, and the validator would fail if you had a library with an unexpected `PSV0` part.

bob80905: Ok, that resolves my concern

lib/DxilContainer/DxilContainerAssembler.cpp
bob80905: Understood, the confusion came from my assumption that GetEntryFunctionName referred to the function you defined earlier under the DxilPipelineStateValidation class, rather than the DxilModule class.
But I still don't see how the string table gets filled with the entry function name, only the string buffer. It looks like DxilPipelineStateValidation::ReadOrWrite is what initializes it, so I suppose it will have already been set before we get to this point.

bob80905 (approved):

lib/DxilContainer/DxilContainerAssembler.cpp
tex3d: The "string table" *is* just the string buffer.  The word "table" might be misleading you here.  The string buffer gets written to the PSV0 part, and offset values in fields referring to it work by offsetting into that string buffer in the PSV0 part.

tex3d: Ah-ha, you actually identified a bug!  I was going to walk through the steps, and realized that the string buffer must be complete before we compute the size in the constructor.  We can't add to it in the write() method, because the buffer we write will still not 

I was thinking "how could the test pass?" but now I realize that it wasn't running in this branch yet due to the validator version requirement (it should now that staging has merged to main, bumping the validator version to 1.8).

I'll fix this and update the PR.

Thanks!

python3kgae (approved):

tools/clang/unittests/HLSL/DxilContainerTest.cpp
```diff
@@ -442,6 +443,25 @@ class DxilContainerTest : public ::testing::Test {
   }
 #endif // _WIN32 - DXBC Unsupported
 
+  void CheckResult(LPCWSTR operation, IDxcOperationResult *pResult) {
+    HRESULT hrStatus = S_OK;
+    VERIFY_SUCCEEDED(pResult->GetStatus(&hrStatus));
+    if (FAILED(hrStatus)) {
+      LogErrorFmt(L"%s failed with hr %d.", operation);
```
bob80905, line 450: Shouldn't there be another object here? I don't see what %d is formatting, but I see operation goes into %s.

bob80905 (approved):

PR #5952, opened 2023-10-30
Add a test dxil lib for testing loadability

### LINKED ISSUE #5394: [Feature Request] Add test for DXIL shared library Loadability

pow2clk:
This creates a dxil library that doesn't sign anything, but does perform the validation very much as the internal validator does. This is used to test that the compiler and validator objects can properly locate, load, and make use of the external dxil library.

The DxilLib test verifies both that loading the compiler with the external dxil library available will load that library by default, resulting in compiles without the usual warning when it is not found and also that explicitly loading the library can be done and the correct API calls can be used.

Fixes #5394


tools/testdxildll/DxilValidator.cpp
```diff
@@ -0,0 +1,242 @@
...
+// DxilValidator.cpp                                                         //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Implements the DirectX Validator object for testing dxil lib              //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+#include "llvm/Bitcode/ReaderWriter.h"
```
python3kgae, line 12: Bitcode readerwriter is not used in this file.

tools/testdxildll/DxilValidator.cpp
```diff
@@ -0,0 +1,242 @@
...
+#include "llvm/IR/DiagnosticPrinter.h"
+#include "llvm/IR/LLVMContext.h"
```
python3kgae, line 14:
```suggestion

```

tools/testdxildll/DxilValidator.cpp
```diff
@@ -0,0 +1,242 @@
...
+#include "dxc/HLSL/DxilValidation.h"
+#include "dxc/Support/WinIncludes.h"
+
+#include "dxc/DxilRootSignature/DxilRootSignature.h"
+#include "dxc/Support/FileIOHelper.h"
+#include "dxc/Support/Global.h"
+#include "dxc/Support/dxcapi.impl.h"
+#include "dxc/Support/microcom.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/MSFileSystem.h"
```
python3kgae, line 26:
```suggestion

```

unittests/DXILLib/DxilLibTest.cpp
```diff
@@ -0,0 +1,152 @@
...
+  HRESULT status1;
+  CComPtr<IDxcBlobEncoding> errBuf1;
+  ASSERT_TRUE(SUCCEEDED(validator1->Validate(blob, DxcValidatorFlags_Default,
+                                             &validationResult1)));
+  ASSERT_TRUE(SUCCEEDED(validationResult1->GetStatus(&status1)));
+  EXPECT_TRUE(SUCCEEDED(status1));
+  ASSERT_TRUE(SUCCEEDED(validationResult1->GetErrorBuffer(&errBuf1)));
+  EXPECT_EQ(errBuf1->GetBufferSize(), 0U);
+
+  CComPtr<IDxcOperationResult> validationResult2;
```
python3kgae, line 143: What is the difference between validationResult1 and validationResult2?

unittests/DXILLib/DxilLibTest.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  // Compile trivial, valid shader and check for warnings
+  CComPtr<IDxcOperationResult> result;
+  CompileSomething(compilerLib, &result);
+
+  CComPtr<IDxcBlobEncoding> errBuf;
+  ASSERT_TRUE(SUCCEEDED(result->GetErrorBuffer(&errBuf)));
+
+  EXPECT_EQ(errBuf->GetBufferSize(), 0U)
+      << "Unexpected warning found in compiler error buffer: "
```
python3kgae, line 103: Could this hit error instead of warning?

pow2clk: I guess it's possible, but given how simple the only shader that is compiled here is, it would indicate a serious problem in the compiler.

If this is to fail, the diagnostic I expect to find is the one that says you don't have a dxil lib and so your shaders won't be signed. This dxil lib won't sign your shader libs, but it avoids this warning. If we should get some other warning or error in this buffer, that would be unexpected, but this would catch it.

unittests/DXILLib/DxilLibTest.cpp
pow2clk: One is created by the plain `validator1` that uses the `CreateInstance` API call and the other is created by `validator2` which uses the `CreateInstance2` API call that also provides a custom allocator. I kept the variables used fully separate to minimize potential for interference between them like a potential stale value from an earlier call.

python3kgae (approved):
Do we need a test case for validation fail?

pow2clk:
The intent of this is not to test validation really, but just that the library is loadable and usable. We have a lot of tests for validation failures. Originally, my intent was to use the complete suite of validation tests on the loadable dxil lib, but was convinced that was overkill

tex3d:
> The intent of this is not to test validation really, but just that the library is loadable and usable. We have a lot of tests for validation failures. Originally, my intent was to use the complete suite of validation tests on the loadable dxil lib, but was convinced that was overkill

Using this library to validate more content is actually useful because it takes a different validation path than the internal validator.  This validation path can and has been broken multiple times before while internal validation path did not catch the problem.  This is because the internal validation path uses the main and debug llvm modules from in-memory form without going through serialization and deserialization.  Something that doesn't get preserved exactly the same way through serialization/deserialization can cause a failure with the external validation path that isn't caught by the internal path.

In short, I think we should run all content against this external validation path when possible.  We don't even need to run content twice to do this, we can simply use this as the available external validator when running tests in GitHub pipelines.

tools/testdxildll/DxilContainerBuilder.cpp
```diff
@@ -0,0 +1,33 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// dxcontainerbuilder.cpp                                                    //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Implements the Dxil Container Builder                                     //
+//                                                                           //
```
hekota, line 9: The comment is incorrect, it does not really implement the containers builds. Any reason why this is not part of the DxilLib.cpp?

tools/testdxildll/DxilContainerBuilder.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "dxc/Support/WinIncludes.h"
+#include "dxc/Support/dxcapi.impl.h"
+#include "dxc/dxcapi.h"
+
+#include <algorithm>
+
+using namespace hlsl;
+
+HRESULT CreateDxcContainerBuilder(_In_ REFIID riid, _Out_ LPVOID *ppv) {
+  // Call dxil.dll's containerbuilder
```
tex3d, line 26: Is this comment accurate?  Isn't this creating the local container builder instead?

tools/testdxildll/DxilValidator.cpp
```diff
@@ -0,0 +1,236 @@
...
+#include "dxc/Support/microcom.h"
+
+#ifdef _WIN32
+#include "dxcetw.h"
+#endif
+
+using namespace llvm;
+using namespace hlsl;
+
+class DxcValidator : public IDxcValidator2, public IDxcVersionInfo {
```
tex3d, line 29: Wait, why can't we move DxcValidator.cpp out of dxcompiler project into shared lib, then you can simply use the same one instead reimplementing so much here?

tools/testdxildll/DxilContainerBuilder.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "dxc/Support/Global.h"
+#include "dxc/Support/WinIncludes.h"
+#include "dxc/Support/dxcapi.impl.h"
+#include "dxc/dxcapi.h"
+
+#include <algorithm>
+
+using namespace hlsl;
+
+HRESULT CreateDxcContainerBuilder(_In_ REFIID riid, _Out_ LPVOID *ppv) {
```
tex3d, line 25: Couldn't this be implemented in DxilLib.cpp, since we really aren't supplying another implementation here?  One might almost mix it up with `DxcContainerBuilder.cpp` where we implement the internal builder that this code uses.

tex3d (commented):
Mainly: I think we might be able to move `dxcvalidator.cpp` into shared lib instead of reimplementing most of it here.

tools/testdxildll/DxilContainerBuilder.cpp
pow2clk: Good catch. The reason it's not part of DxilLib.cpp is because of some inclusions this requires that confuse the code in DxilLib.cpp. I spent a little while trying to resolve them, but decided it wasn't worth it as this is the structure of similar code elsewhere likely for the same reason.

tools/testdxildll/DxilLib.cpp
```diff
@@ -0,0 +1,150 @@
+///////////////////////////////////////////////////////////////////////////////
```
hekota, line 1: Would you mind renaming DxilLib.* files to TestDxilLib.*? I am often searching through the whole code base and since this code is similar to the actual DxilLib.cpp it would be nice to have it under a different name to it can be clearly seen in the search results that this is the test dll.

tools/testdxildll/DxilValidator.cpp
pow2clk: Done

tools/testdxildll/DxilContainerBuilder.cpp
pow2clk: There were some compilation errors when I tried to do this before that had to do with conflicting defines. I'll take another run at it.

tools/testdxildll/DxilLib.cpp
pow2clk: Will do

tools/no-sig-dxillib/DxilLibValidator.cpp
```diff
@@ -0,0 +1,65 @@
...
+  if (pMajor == nullptr || pMinor == nullptr)
+    return E_INVALIDARG;
+  hlsl::GetValidationVersion(pMajor, pMinor);
+  return S_OK;
+}
+
+HRESULT STDMETHODCALLTYPE DxilLibValidator::GetFlags(UINT32 *pFlags) {
+  if (pFlags == nullptr)
+    return E_INVALIDARG;
+  *pFlags = DxcVersionInfoFlags_None;
```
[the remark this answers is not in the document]
tex3d: `DxcVersionInfoFlags_Internal` can also be used to determine whether we should expect the line-only debug info supplied to the validator when the validator is called automatically from Compile().

tools/clang/test/CodeGenHLSL/mesh-val/asOversizePayload.hlsl
```diff
@@ -2,7 +2,7 @@
 // RUN: %dxilver 1.6 | %dxc -E main -T as_6_5 %s | FileCheck %s -check-prefix=CHK_NODB
 
 // CHK_DB: 23:5: error: For amplification shader with entry 'main', payload size 16400 is greater than maximum size of 16384 bytes.
-// CHK_NODB: 23:5: error: For amplification shader with entry 'main', payload size 16400 is greater than maximum size of 16384 bytes.
+// CHK_NODB: error: For amplification shader with entry 'main', payload size 16400 is greater than maximum size of 16384 bytes.
```
tex3d, line 5: This change shouldn't have been necessary.  IDxcValidator2::ValidateWithDebug allows you to provide the serialized debug module, which is normally provided to the external validator when present.  This must work the normal DXIL.dll because these don't normally fail when testing against that.  So perhaps there's something wrong with the new test validator?

damyanp:
Now that we've open sourced the validator/hashing we'll likely address this a different way.

PR #5953, opened 2023-10-30
[Diag] report error when SV_DispatchGrid has illegal type.

### LINKED ISSUE #5845: Work graphs: SV_DispatchGrid field type in node input records is not validated
**Description**
We need to verify that the input record field with `SV_DispatchGrid` semantic is a valid type: 32 or 16-bit uint scalar or vector with up to 3 components.

**Steps to Reproduce**

```
struct Record {
  uint4 grid : SV_DispatchGrid;  
  float data;
};


[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeMaxDispatchGrid(32, 16, 1)]
[NumThreads(32, 1, 1)]
void node1(DispatchNodeInputRecord<Record> input) {}
```

**Actual Behavior**

No errors. 
The compiler crashes if matrix is used: `uint4x4 grid : SV_DispatchGrid;`

**Expected Behavior**

Error message saying that the field with `SV_DispatchGrid` semantic must be a 32 or 16-bit uint scalar or vector with up to 3 components.

**Environment**
- DXC version `staging-sm-6.8` branch as of 10/9

python3kgae:
verify that the input record field with SV_DispatchGrid semantic is a valid type: 32 or 16-bit uint scalar or array/vector with up to 3 components.

Fixes #5845

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7861,7 +7861,7 @@ def err_hlsl_missing_dispatchgrid_semantic : Error<
 def err_hlsl_dispatchgrid_semantic_already_specified : Error<
    "a field with SV_DispatchGrid has already been specified">;
 def err_hlsl_incompatible_dispatchgrid_semantic_type : Error<
-   "SV_DispatchGrid can only be applied to a declaration of unit/uint16_t scalar, vector, or array up to 3 elements, not %0">;
+   "SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not %0">;
 def err_hlsl_maxrecursiondepth_exceeded : Error<
    "NodeMaxRecursionDepth may not exceed 32">;
 def err_hlsl_too_many_node_inputs : Error<
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7861,7 +7861,7 @@ def err_hlsl_missing_dispatchgrid_semantic : Error<
 def err_hlsl_dispatchgrid_semantic_already_specified : Error<
    "a field with SV_DispatchGrid has already been specified">;
 def err_hlsl_incompatible_dispatchgrid_semantic_type : Error<
-   "SV_DispatchGrid should be 32/16 bit uint scalar or vector/array up to 3 elements">;
+   "SV_DispatchGrid can only be applied to a declaration of unit/uint16_t scalar, vector, or array up to 3 elements, not %0">;
 def err_hlsl_maxrecursiondepth_exceeded : Error<
    "NodeMaxRecursionDepth may not exceed 32">;
 def err_hlsl_too_many_node_inputs : Error<
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7860,6 +7860,8 @@ def err_hlsl_missing_dispatchgrid_semantic : Error<
    "Broadcasting node shader '%0' with NodeMaxDispatchGrid attribute must declare an input record containing a field with SV_DispatchGrid semantic">;
 def err_hlsl_dispatchgrid_semantic_already_specified : Error<
    "a field with SV_DispatchGrid has already been specified">;
+def err_hlsl_incompatible_dispatchgrid_semantic_type : Error<
+   "SV_DispatchGrid should be 32/16 bit uint scalar or vector/array up to 3 elements">;
 def err_hlsl_maxrecursiondepth_exceeded : Error<
    "NodeMaxRecursionDepth may not exceed 32">;
 def err_hlsl_too_many_node_inputs : Error<
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -15273,6 +15273,32 @@ void DiagnoseMustHaveOneDispatchGridSemantics(Sema &S,
         if (sd->SemanticName.equals("SV_DispatchGrid")) {
           if (!Found) {
             Found = true;
+            QualType Ty = FD->getType();
+            if (hlsl::IsVectorType(&S, Ty)) {
+              unsigned NumElt = hlsl::GetElementCount(Ty);
+              if (NumElt > 3) {
+                S.Diags.Report(
+                    it->Loc,
+                    diag::err_hlsl_incompatible_dispatchgrid_semantic_type);
+              }
+              Ty = hlsl::GetHLSLVecElementType(Ty);
+            } else if (const ArrayType *AT = Ty->getAsArrayTypeUnsafe()) {
+              if (auto *CAT = dyn_cast<ConstantArrayType>(AT)) {
+                if (CAT->getSize().getZExtValue() > 3) {
+                  S.Diags.Report(
+                      it->Loc,
+                      diag::err_hlsl_incompatible_dispatchgrid_semantic_type);
+                }
+              }
+              Ty = AT->getElementType();
+            }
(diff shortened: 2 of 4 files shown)


hekota (approved):
LGTM!

tools/clang/test/SemaHLSL/hlsl/workgraph/dispatch_grid.hlsl
```diff
@@ -0,0 +1,84 @@
...
+
+
+[Shader("node")]
+[NodeLaunch("broadcasting")]
+[NodeMaxDispatchGrid(32, 16, 1)]
+[NumThreads(32, 1, 1)]
+void node5(DispatchNodeInputRecord<Record5> input) {}
+
+struct T {
+  float a;
```
pow2clk, line 71: I think a few other test cases might be interesting: A scalar float, a float3, a user-defined struct with uint3 as the only member.

tools/clang/test/HLSL/workgraph/dispatchgrid_diags.hlsl
```diff
@@ -153,6 +153,7 @@ void node20(DispatchNodeInputRecord<C> input)
 { }
 
 struct D {
+  // expected-error@+1 {{SV_DispatchGrid should be 32/16 bit uint scalar or vector/array up to 3 elements}}
   uint4 grid1 : SV_DispatchGrid;   // expected-note {{other SV_DispatchGrid defined here}}
```
pow2clk, line 157: Since this isn't what this was meant to test, I'm inclined to change the type to uint3 instead of adding another expected diagnostic as the test you've added does that quite well.

pow2clk (commented):
Looks good! I have some minor test suggestions you can take or leave.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15273,6 +15273,32 @@ void DiagnoseMustHaveOneDispatchGridSemantics(Sema &S,
...
+            if (hlsl::IsVectorType(&S, Ty)) {
+              unsigned NumElt = hlsl::GetElementCount(Ty);
+              if (NumElt > 3) {
+                S.Diags.Report(
+                    it->Loc,
+                    diag::err_hlsl_incompatible_dispatchgrid_semantic_type);
+              }
+              Ty = hlsl::GetHLSLVecElementType(Ty);
+            } else if (const ArrayType *AT = Ty->getAsArrayTypeUnsafe()) {
+              if (auto *CAT = dyn_cast<ConstantArrayType>(AT)) {
```
llvm-beanz, line 15286: nit:
```suggestion
              if (const auto *CAT = dyn_cast<ConstantArrayType>(AT)) {
```

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7860,6 +7860,8 @@ def err_hlsl_missing_dispatchgrid_semantic : Error<
    "Broadcasting node shader '%0' with NodeMaxDispatchGrid attribute must declare an input record containing a field with SV_DispatchGrid semantic">;
 def err_hlsl_dispatchgrid_semantic_already_specified : Error<
    "a field with SV_DispatchGrid has already been specified">;
+def err_hlsl_incompatible_dispatchgrid_semantic_type : Error<
+   "SV_DispatchGrid should be 32/16 bit uint scalar or vector/array up to 3 elements">;
```
llvm-beanz, line 7864: Should we make the diagnostic print the type found? That will make it easier to understand if this were to occur under template instantiation where the type may not appear in the code.

```suggestion
   "SV_DispatchGrid can only be applied to a declaration of 32 or 16-bit integer scalar or 3-element vector or array, not %0">;
```

tools/clang/test/SemaHLSL/hlsl/workgraph/dispatch_grid.hlsl
```diff
@@ -0,0 +1,126 @@
+// RUN: %dxc -Tlib_6_8 -verify %s
```
llvm-beanz, line 1: None of the tests here verify the `uint16_t` case.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7860,6 +7860,8 @@ def err_hlsl_missing_dispatchgrid_semantic : Error<
    "Broadcasting node shader '%0' with NodeMaxDispatchGrid attribute must declare an input record containing a field with SV_DispatchGrid semantic">;
 def err_hlsl_dispatchgrid_semantic_already_specified : Error<
    "a field with SV_DispatchGrid has already been specified">;
+def err_hlsl_incompatible_dispatchgrid_semantic_type : Error<
+   "SV_DispatchGrid can only be applied to a declaration of unit/uint16_t scalar, vector, or array up to 3 elements, not %0">;
```
llvm-beanz, line 7864: It would be better to not list the explicit type since these are type aliases and may end up sugared ore de-sugared during diagnostics or rewriting. `uint` is equivalent to `unsigned int` and `uint16_t` should be equivalent to `unsigned int16_t` (although the later seems to crash DXC 🤕)

```suggestion
   "SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not %0">;
```

tools/clang/test/SemaHLSL/hlsl/workgraph/dispatch_grid.hlsl
python3kgae: Added uint16_t and int16_t cases.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
python3kgae: Updated.

tools/clang/test/SemaHLSL/hlsl/workgraph/dispatch_grid.hlsl
python3kgae: Done.

pow2clk: Wow! well I think we have enough test cases now! 👍

pow2clk (approved):
So many test cases!

tex3d (approved):
I think this change looks good.  However, there are a couple more potential follow-up changes to make still:
- minor: name `DiagnoseMustHaveOneDispatchGridSemantics` could probably be `DiagnoseDispatchGridSemantics`, since it does more now.  Plus, it actually only diagnoses if there is an error in type or more than one.  You diagnose separately if no SV_DispatchGrid is found (which is good, see the next point).
- This appears to only be used for the one input record scenario.  This should also be used for output record types.  It's not required that any output record has the SV_DispatchGrid semantic, but if it does, there should be at most one field that conforms to the type requirements.

llvm-beanz (approved):

tex3d:
> I think this change looks good. However, there are a couple more potential follow-up changes to make still:
> 
> * minor: name `DiagnoseMustHaveOneDispatchGridSemantics` could probably be `DiagnoseDispatchGridSemantics`, since it does more now.  Plus, it actually only diagnoses if there is an error in type or more than one.  You diagnose separately if no SV_DispatchGrid is found (which is good, see the next point).
> * This appears to only be used for the one input record scenario.  This should also be used for output record types.  It's not required that any output record has the SV_DispatchGrid semantic, but if it does, there should be at most one field that conforms to the type requirements.

I filed #5958 to capture this follow-up.

python3kgae:
> I think this change looks good. However, there are a couple more potential follow-up changes to make still:
> 
> * minor: name `DiagnoseMustHaveOneDispatchGridSemantics` could probably be `DiagnoseDispatchGridSemantics`, since it does more now.  Plus, it actually only diagnoses if there is an error in type or more than one.  You diagnose separately if no SV_DispatchGrid is found (which is good, see the next point).
> * This appears to only be used for the one input record scenario.  This should also be used for output record types.  It's not required that any output record has the SV_DispatchGrid semantic, but if it does, there should be at most one field that conforms to the type requirements.

Updated the name.
Will work on output record types next.

PR #5954, opened 2023-10-30
WinPix: variables in subroutines called from multiple functions

jeffnn:
From the comment in one of the changed files:

// If a function is inlined, then the scope of variables within that function
// will be that scope. Since many such callers may inline the function, we
// actually need a "scope" that's unique to each "instance" of that function.
// The caller's scope would be unique, but would have the undesirable
// side-effect of smushing all of the function's variables together with the
// caller's variables, at least from the point of view of PIX. The pair of
// scopes (the function's and the caller's) is both unique to that particular
// inlining, and distinct from the caller's scope by itself.

### DIFF
--- a/lib/DxilDia/DxcPixLiveVariables.cpp
+++ b/lib/DxilDia/DxcPixLiveVariables.cpp
@@ -29,6 +29,35 @@
 
 #include <unordered_map>
 
+namespace std {
+template <> struct hash<std::pair<llvm::DIScope *, llvm::DIScope *>> {
+  size_t
+  operator()(std::pair<llvm::DIScope *, llvm::DIScope *> const &p) const {
+    const uint64_t f = reinterpret_cast<uint64_t>(p.first);
+    const uint64_t s = reinterpret_cast<uint64_t>(p.second);
+    std::hash<uint64_t> h;
+    return (h(f) ^ h(s));
+  }
+};
+} // namespace std
+
+// If a function is inlined, then the scope of variables within that function
+// will be that scope. Since many such callers may inline the function, we
+// actually need a "scope" that's unique to each "instance" of that function.
+// The caller's scope would be unique, but would have the undesirable
+// side-effect of smushing all of the function's variables together with the
+// caller's variables, at least from the point of view of PIX. The pair of
+// scopes (the function's and the caller's) is both unique to that particular
+// inlining, and distinct from the caller's scope by itself.
+static std::pair<llvm::DIScope *, llvm::DIScope *>
+GetUniqueScopeForPossiblyInlinedFunction(llvm::DebugLoc const &DbgLoc,
+                                         llvm::DIScope *VariableScope) {
+  llvm::DIScope *inlinedScope =
+      llvm::dyn_cast<llvm::DIScope>(DbgLoc.getInlinedAtScope());
+  return std::pair<llvm::DIScope *, llvm::DIScope *>(VariableScope,
+                                                     std::move(inlinedScope));
+}
+
 // ValidateDbgDeclare ensures that all of the bits in
 // [FragmentSizeInBits, FragmentOffsetInBits) are currently
 // not assigned to a dxil alloca register -- i.e., it
@@ -61,7 +90,9 @@ struct dxil_debug_info::LiveVariables::Impl {
   using VariableInfoMap =
       std::unordered_map<llvm::DIVariable *, std::unique_ptr<VariableInfo>>;
 
-  using LiveVarsMap = std::unordered_map<llvm::DIScope *, VariableInfoMap>;
+  using LiveVarsMap =
+      std::unordered_map<std::pair<llvm::DIScope *, llvm::DIScope *>,
+                         VariableInfoMap>;
 
   IMalloc *m_pMalloc;
   DxcPixDxilDebugInfo *m_pDxilDebugInfo;
@@ -115,8 +146,9 @@ void dxil_debug_info::LiveVariables::Impl::Init_DbgDeclare(
     return;
   }
 
-  auto *S = Variable->getScope();
-  if (S == nullptr) {
+  auto S = GetUniqueScopeForPossiblyInlinedFunction(DbgDeclare->getDebugLoc(),
+                                                    Variable->getScope());
+  if (S.first == nullptr) {
     return;
   }
 
@@ -204,13 +236,13 @@ HRESULT dxil_debug_info::LiveVariables::GetLiveVariablesAtInstruction(
     return E_FAIL;
   }
 
-  llvm::DIScope *S = DL->getScope();
-  if (S == nullptr) {
+  auto S = GetUniqueScopeForPossiblyInlinedFunction(DL, DL->getScope());
+  if (S.first == nullptr) {
     return E_FAIL;
   }
 
   const llvm::DITypeIdentifierMap EmptyMap;
-  while (S != nullptr) {
+  while (S.first != nullptr) {
     auto it = m_pImpl->m_LiveVarsDbgDeclare.find(S);
     if (it != m_pImpl->m_LiveVarsDbgDeclare.end()) {
       for (const auto &VarAndInfo : it->second) {
@@ -232,7 +264,8 @@ HRESULT dxil_debug_info::LiveVariables::GetLiveVariablesAtInstruction(
         LiveVars.emplace_back(VarAndInfo.second.get());
       }
     }
-    S = S->getScope().resolve(EmptyMap);
+    S.second = nullptr;
+    S.first = S.first->getScope().resolve(EmptyMap);
   }
   for (const auto &VarAndInfo : m_pImpl->m_LiveGlobalVarsDbgDeclare) {
     if (!LiveVarsName.insert(VarAndInfo.first->getName()).second) {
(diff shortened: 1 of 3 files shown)


lib/DxilDia/DxcPixLiveVariables.cpp
```diff
@@ -232,7 +264,8 @@ HRESULT dxil_debug_info::LiveVariables::GetLiveVariablesAtInstruction(
         LiveVars.emplace_back(VarAndInfo.second.get());
       }
     }
-    S = S->getScope().resolve(EmptyMap);
+    S.second = nullptr;
```
python3kgae, line 267: Is there user of S.second?

lib/DxilDia/DxcPixLiveVariables.cpp
```diff
@@ -29,6 +29,35 @@
...
+// caller's variables, at least from the point of view of PIX. The pair of
+// scopes (the function's and the caller's) is both unique to that particular
+// inlining, and distinct from the caller's scope by itself.
+static std::pair<llvm::DIScope *, llvm::DIScope *>
+GetUniqueScopeForPossiblyInlinedFunction(llvm::DebugLoc const &DbgLoc,
+                                         llvm::DIScope *VariableScope) {
+  llvm::DIScope *inlinedScope =
+      llvm::dyn_cast<llvm::DIScope>(DbgLoc.getInlinedAtScope());
+  return std::pair<llvm::DIScope *, llvm::DIScope *>(VariableScope,
+                                                     std::move(inlinedScope));
```
python3kgae, line 58: Why std::move here?

jeffnn: OOps that was a leftover form an earlier attempt.

lib/DxilDia/DxcPixLiveVariables.cpp
jeffnn: S.second is only there to be part of the key for the m_LiveVarsDbgDeclare map. Seems like the "inlined at" function is always the top-level function even when the inlined is more than one function deep, so it's OK to discard "S.second": it has done its job (unique-ifying the subroutine AND where it was inlined).

python3kgae: I see. It would be nice to have some comment for this.

python3kgae (approved):

lib/DxilDia/DxcPixLiveVariables.cpp
```diff
@@ -29,6 +29,35 @@
...
+
+// If a function is inlined, then the scope of variables within that function
+// will be that scope. Since many such callers may inline the function, we
+// actually need a "scope" that's unique to each "instance" of that function.
+// The caller's scope would be unique, but would have the undesirable
+// side-effect of smushing all of the function's variables together with the
+// caller's variables, at least from the point of view of PIX. The pair of
+// scopes (the function's and the caller's) is both unique to that particular
+// inlining, and distinct from the caller's scope by itself.
+static std::pair<llvm::DIScope *, llvm::DIScope *>
```
adam-yang, line 52: I'm not sure this could uniquely identify an inlined scope. If `B` calls `A`, and `C` and `D` both call `B`, can this scheme distinguish scopes of the two inlined `A`'s?

tools/clang/unittests/HLSL/PixDiaTest.cpp
```diff
@@ -2755,4 +2756,159 @@ void main()
...
+      dxilDebugger, instructionOffset, L"ClosestHitShader1");
+  instructionOffset = AdvanceUntilFunctionEntered(
+      dxilDebugger, instructionOffset, L"DeeperInlinedFunction");
+  DWORD RegisterNumber1 = GetRegisterNumberForVariable(
+      dxilDebugger, instructionOffset, L"ret", L"x");
+  instructionOffset = AdvanceUntilFunctionEntered(
+      dxilDebugger, instructionOffset, L"InlinedFunction");
+  DWORD RegisterNumber3 = GetRegisterNumberForVariable(
+      dxilDebugger, instructionOffset, L"color2", L"x");
+  VERIFY_ARE_NOT_EQUAL(RegisterNumber0, RegisterNumber1);
```
adam-yang, line 2910: Can we also test that the register numbers are not -1?

lib/DxilDia/DxcPixLiveVariables.cpp
jeffnn: I think that's the case in the included test? A == DeeperInlinedFunction, B == InlinedFunction and C/D are ClosestHitShader0/1...
Plus, I have only been able to come up with situations in which the deeper-inlined function's llvm::DebugLoc::getInlinedAtScope returns the outermost caller, i.e. the deeper never returns the next-further-up-the-stack inlined function... Would there be any tricks you can think of where I could get deeper inlined scopes than this?

tools/clang/unittests/HLSL/PixDiaTest.cpp
jeffnn: Sure. Any -1 will eventually cause the final test to fail, but it makes sense to test early and often: more debuggable and readable, at least.

lib/DxilDia/DxcPixLiveVariables.cpp
```diff
@@ -29,6 +29,35 @@
...
+GetUniqueScopeForPossiblyInlinedFunction(llvm::DebugLoc const &DbgLoc,
+                                         llvm::DIScope *VariableScope) {
+  llvm::DIScope *inlinedScope =
+      llvm::dyn_cast<llvm::DIScope>(DbgLoc.getInlinedAtScope());
```
adam-yang, line 56: I think this better be `dyn_cast_or_null`, because `DbgLoc` is not guaranteed to be inlined.

lib/DxilDia/DxcPixLiveVariables.cpp
```diff
@@ -29,6 +29,35 @@
 
 #include <unordered_map>
 
+namespace std {
+template <> struct hash<std::pair<llvm::DIScope *, llvm::DIScope *>> {
```
adam-yang, line 33: Since we need to define a custom hashing function anyway, can we use a struct instead of a `std::pair`? It'll be easier to read at a glance what the two scopes are.

jeffnn: That makes sense...

tools/clang/unittests/HLSL/PixDiaTest.cpp
adam-yang: If one of `RegisterNumber0` or `RegisterNumber1` is -1, the test will just pass.

lib/DxilDia/DxcPixLiveVariables.cpp
adam-yang: Oh I see what's going on. I assumed `getInlinedAtScope` returns the parent caller, but it actually returns the outer-most caller. I bet if you have ClosestHitShader0 call both `InlinedFunction` and `DeeperInlinedFunction`, the two inlined `ret` variable would end up with the same scope pair.

jeffnn: I think the _pair_ would be unique because the first element (i.e. the inlined function's own scope) would be different. Where this _does_ break is the same inline called from two different places from the top-level function. Urf. Need to think about that one.

adam-yang: Sorry I mean the two `ret` from the two `DeeperInlinedFunction`'s. Would they not both have the pair `{ DeeperInlinedFunction, ClosestHitShader2 }`?

I was thinking we might need to keep a whole callstack of scopes to uniquely identify a variable, but that wouldn't work either, because the two instances of inlined `DeeperInlinedFunction` would have the exact same call stack.
```
void DeeperInlinedFunction() {}
void InlinedFunction() {
   DeeperInlinedFunction();
   DeeperInlinedFunction();
}
void ClosestHitShader2() {
   InlinedFunction();
}
```

Maybe the actual identifier needs to be a stack of `DILocation`s instead of `DIScope`s. 🤔

jeffnn: We don't have to make each variable unique to make PIX happy. The issue comes from PIX thinking that two instances of the same variable (i.e. a local in an inlined function that has been inlined twice or more) are the same, and so only a single storage area is allocated for that variable in the output UAV of the debug instrumentation pass.
If a function is inlined into two different top-level functions in the same module, then prior to this change, PIX would associate the inlined function's locals with one UAV storage area, with the result in PIX that one invocation's variables will leak into the other function. That was happening because the prior code used just the inlined function's scope to look up the UAV storage. (The actual experience in PIX is that you get the same 
If a function is inlined twice into the same top-level, then it's OK to have the same UAV storage, since any individual thread that executes that inline function will write first one value into the UAV storage, and later write another. PIX is perfectly happy with this, and in fact you could argue that it SHOULD be this way, since the inline really is conceptually reusing memory.
The double inline case (i.e. one function inlined into two different functions that are then also inlined in the top-level function) is, from PIX's point of view, the same as the single-inline case: all we need to do is make the UAV storage unique per upper-level function (and, of course, unique per inlined function).

(Some mostly-irrelevant complexity is omitted from this discussion which I'm happy to supply.)

adam-yang: Thanks for the exaplanation! I think I get it now 👍

adam-yang (approved):

PR #5956, opened 2023-11-01
[SPIR-V] Emit error for unsupported HLSL res obj

### LINKED ISSUE #5913: [SPIR-V] Internal shader compilation error with ResourceDescriptorHeap
**Description**
DXC Shader Compiler crashes with internal error when compiling shader code with ResourceDescriptorHeap to SPIR-V.

**Steps to Reproduce**
_bindlessCompute.hlsl_
```hlsl
[numthreads(256,1,1)]
void main(uint3 DTid: SV_DispatchThreadID) {
    StructuredBuffer<uint> src = ResourceDescriptorHeap[0];
    RWStructuredBuffer<uint> dst = ResourceDescriptorHeap[1];
    dst[DTid.x] = src[DTid.x];
}
```
_Command:_
```powershell
dxc.exe -spirv -T cs_6_7 -E main -O0 bindlessCompute.hlsl -Fo compiled.spirv
```
_Error:_
```
Internal compiler error: access violation. Attempted to read from address 0x0000000000000000
```

**Actual Behavior**
As I know SPIR-V does not support ```ResourceDescriptorHeap``` and dynamic descriptor casting for now. But compiler crashes trying to compile code that contains ```ResourceDescriptorHeap``` instead of printing validation error message.

**Environment**
- DXC version  1.7 - 1.7.0.4004

sudonatalie:
Emit an explicit error message when HLSL `ResourceDescriptorHeap` or `SamplerDescriptorHeap` are used with `-spirv` as they are not yet supported.

Fixes #5913

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -956,8 +956,9 @@ DeclResultIdMapper::getDeclSpirvInfo(const ValueDecl *decl) const {
 SpirvInstruction *DeclResultIdMapper::getDeclEvalInfo(const ValueDecl *decl,
                                                       SourceLocation loc,
                                                       SourceRange range) {
-  if (hlsl::IsHLSLDynamicResourceType(decl->getType())) {
-    emitError("HLSL resource %0 not yet supported with -spirv",
+  if (hlsl::IsHLSLDynamicResourceType(decl->getType()) ||
+      hlsl::IsHLSLDynamicSamplerType(decl->getType())) {
+    emitError("HLSL object %0 not yet supported with -spirv",
               decl->getLocation())
         << decl->getName();
     return nullptr;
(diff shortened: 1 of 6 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -875,6 +875,13 @@ DeclResultIdMapper::getDeclSpirvInfo(const ValueDecl *decl) const {
 SpirvInstruction *DeclResultIdMapper::getDeclEvalInfo(const ValueDecl *decl,
                                                       SourceLocation loc,
                                                       SourceRange range) {
+  if (hlsl::IsHLSLDynamicResourceType(decl->getType())) {
```
s-perron, line 878: Can you explain to me why this will impact `ResourceDescriptorHeap` and not other types?

sudonatalie: To be honest though I did also rely on just seeing that that was all code that was added in the big SM 6.6 patch, and that all tests passed still passed when I added this, but if that feels like logic that's a bit too shaky I can change it to string checking for `"ResourceDescriptorHeap"`.

cassiebeckley: In my opinion it's better to use the existing `IsHLSLDynamicResourceType`, which seems to be explicitly for checking for this type, than using a string comparison.

cassiebeckley (approved):
Will we need something like this for `SamplerDescriptorHeap` as well?

sudonatalie:
> Will we need something like this for `SamplerDescriptorHeap` as well?

Good catch. Updated.

@llvm-beanz Can you review the `HlslTypes.*` changes?

llvm-beanz (approved):
LGTM

s-perron (approved):

PR #5974, opened 2023-11-03
Fix ASAN heap-buffer-overflow in PixTests

amaiorano:
PixStructAnnotation tests were converting an IDxcBlobEncoding to a std::string, but this isn't valid if the underlying object is an InternalDxcBlobEncoding_Impl, which holds a buffer that may not be null-terminated. Fix this by using the BlobToUtf8 helper, which handles this case gracefully.

Fixes 23 ASAN failures.


python3kgae:
Should the target branch be main?

amaiorano:
> Should the target branch be main?

Yes, thank you. Fixed.

python3kgae (approved):

llvm-beanz (approved):

Issue #5971, opened 2023-11-03
ASAN alloc_dealloc_mismatch false positive on Ubuntu Linux when using libc++ package


amaiorano:
Build and run check-all on Linux with ASAN enabled, and at least one test fails with alloc_dealloc_mismatch ASAN failure:

```
FAIL: Clang :: DXC/recompile.test (925 of 3612)
...
==477155==ERROR: AddressSanitizer: alloc-dealloc-mismatch (operator new vs free) on 0x506000002f60
    #0 0x562b4361fd06 in free (/home/user/src/external/DirectXShaderCompiler/build-clang/bin/dxc-3.7+0x589d06) (BuildId: 85a1466c7f2cb376)
    #1 0x7f5b2a8a80d4 in std::invalid_argument::~invalid_argument() (/lib/x86_64-linux-gnu/libc++abi.so.1+0x250d4) (BuildId: 5abde9e03185be802c843f06145fa1ac038513fd)
    #2 0x7f5b2a8aa3b1 in __cxa_end_catch (/lib/x86_64-linux-gnu/libc++abi.so.1+0x273b1) (BuildId: 5abde9e03185be802c843f06145fa1ac038513fd)
    #3 0x562b4368ca23 in DxcIncludeHandlerForInjectedSources::LoadSource(wchar_t const*, IDxcBlob**) /home/user/src/external/DirectXShaderCompiler/tools/clang/tools/dxclib/dxc.cpp:709:5
    #4 0x7f5b2e2e7d95 in dxcutil::DxcArgsFileSystemImpl::TryFindOrOpen(wchar_t const*, unsigned long&) /home/user/src/external/DirectXShaderCompiler/tools/clang/tools/dxcompiler/dxcfilesystem.cpp:305:37
    #5 0x7f5b2e2e35a5 in dxcutil::DxcArgsFileSystemImpl::GetFileAttributesW(wchar_t const*) /home/user/src/external/DirectXShaderCompiler/tools/clang/tools/dxcompiler/dxcfilesystem.cpp:583:19
    #6 0x7f5b2e2e52ed in dxcutil::DxcArgsFileSystemImpl::Stat(char const*, stat*) /home/user/src/external/DirectXShaderCompiler/tools/clang/tools/dxcompiler/dxcfilesystem.cpp:811:18

...

0x506000002f60 is located 0 bytes inside of 57-byte region [0x506000002f60,0x506000002f99)
allocated by thread T0 here:
[...]

PR #5976, opened 2023-11-03
Fix enable asan on pipeline

amaiorano:
Two commits in this one:

 azure-pipelines: Fix ASAN not being enabled on Linux bot
    
    Unfortunately, the following CL did not correctly spell out the CMake
    variable to enable ASAN properly:
    https://github.com/microsoft/DirectXShaderCompiler/pull/5799
    
    This commit fixes that.

 azure-pipelines: disable ASAN alloc_dealloc_mismatch checks
    
    We are currently hitting false positives because of this. See the issue
    I opened: https://github.com/microsoft/DirectXShaderCompiler/issues/5971
    
    Perhaps a future Linux image will include a build of libc++ that does
    not exhibit this false positive, at which point this workaround can be
    reverted.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -49,7 +49,7 @@ stages:
           configuration: Release
           CC: clang
           CXX: clang++
-          CMAKE_OPTS: -DLLVM_ENABLE_WERROR=On -DLLVM_ENABLE_SANITIZER=Address -DLLVM_ENABLE_LIBCXX=On -DLLVM_USE_LINKER=lld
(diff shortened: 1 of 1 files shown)


amaiorano:
This one depends on the following three ASAN fixes landing first:
https://github.com/microsoft/DirectXShaderCompiler/pull/5972
https://github.com/microsoft/DirectXShaderCompiler/pull/5974
https://github.com/microsoft/DirectXShaderCompiler/pull/5975

sudonatalie (approved):

PR #5979, opened 2023-11-04
Fix downlevel validation testing failures

tex3d:
Moving some tests from CodeGenDXIL to HLSLFileCheck because they will be properly gated on validator version here. These can be moved back once we have support for this in lit shell test.

Added %dxilver for %dxv tests requiring new validator version.

Moved `DXILValidation/Metadata` tests to `HLSLFileCheck/validation/Metadata` to allow %dxilver.  Needed to add RUN line for the HLSL file there.

Added -select-validator internal to lit shell test that checks diagnostics, thus does not run through validation, and shouldn't be limited by external validator version.

Updated version requirement in ValidationTest::AtomicsInvalidDests because the message was changed in the latest validator version.


hekota (approved):
LGTM!

tools/clang/test/HLSLFileCheck/hlsl/workgraph/invalid_barrier_1.ll
```diff
@@ -1,4 +1,4 @@
-; RUN: %dxv %s | FileCheck %s
+; RUN: %dxilver 1.8 | %dxv %s | FileCheck %s
```
pow2clk, line 1: I'm surprised this is the only one that requires this. invalid_barrier_2/12.ll both use dx.op.barrierByMemoryType.

pow2clk (approved):
I wonder why more wasn't required, but there's nothing wrong here.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/invalid_barrier_1.ll
tex3d: Strange, you're right, how could these pass?  Maybe I missed something, let me double check.

tex3d: Yeah, there were more failures, working on it.  Basically, the failure I hit were the first ones and would crash the old validator, preventing more tests in the same group from running.  Then when testing locally, hcttest.cmd -dxil-loc mechanism has been completely broken by the lit testing addition.  So after fixing and trying this locally, everything passed, because it wasn't actually testing against the validator I selected.

tex3d: Ok, I've pushed a bunch of additional test fixes.  This should be good now.

PR #5983, opened 2023-11-07
Only diagnose exported library function declarations

### LINKED ISSUE #5857: [Diagnostics] Only emit diagnostics for recursive functions in library shaders that are exported.

bob80905:
Previously, diagnosing the translation unit would loop through every function declaration found in a library shader. This behavior is undesired, as there are some functions that are not intended to be exported, and thus shouldn't emit diagnostics (maybe the library compilation target is for a shader model that wouldn't mesh well with some other function declarations in the translation unit). This PR adjusts the diagnostic emission code so that only exported functions are diagnosed.
Fixes #5857

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -11464,14 +11464,12 @@ bool IsExported(Sema *self, clang::FunctionDecl *FD) {
 void ValidatePatchConstantFunctionsExist(clang::Sema *self) {
 
   for (auto decl : self->getASTContext().getTranslationUnitDecl()->decls()) {
-    // TODO: improve condition so that only exported functions are checked,
-    // instead of all functions. Issue: #5857
     if (FunctionDecl *FD = dyn_cast<FunctionDecl>(decl)) {
-      // If there is no patch constant function, then we don't need to validate
-      // anything.
       if (!IsExported(self, FD)) {
         continue;
       }
+      // If there is no patch constant function, then we don't need to validate
+      // anything.
       if (const HLSLPatchConstantFuncAttr *Attr =
               FD->getAttr<HLSLPatchConstantFuncAttr>()) {
         NameLookup NL =
@@ -15859,12 +15857,13 @@ clang::FunctionDecl *ValidateNoRecursion(clang::Sema *self,
   return nullptr;
 }
 
+// There is an assumption being made that this function will only be called
+// on library shader stages.Thus, there is no entry point, only exported
+// functions
 void ValidateNoRecursionInTranslationUnit(clang::Sema *self) {
   std::set<FunctionDecl *> FDecls;
   std::vector<FunctionDecl *> FDeclsVec;
   for (auto decl : self->getASTContext().getTranslationUnitDecl()->decls()) {
-    // TODO: improve condition so that only exported functions are checked,
-    // instead of all functions. Issue: #5857
     if (FunctionDecl *FD = dyn_cast<FunctionDecl>(decl)) {
       // returns the first recursive function declaration detected
       // from this function declaration FD, and determines whether
@@ -15889,7 +15888,7 @@ void ValidateNoRecursionInTranslationUnit(clang::Sema *self) {
       continue;
     }
     self->Diag(fdecl->getSourceRange().getBegin(), diag::err_hlsl_no_recursion)
-        << 0 << fdecl->getName();
+        << 1 << fdecl->getName();
 
     FDecls.erase(fdecl);
   }
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -11400,17 +11400,17 @@ bool hlsl::DiagnoseNodeStructArgument(Sema *self, TemplateArgumentLoc ArgLoc,
   }
 }
 
-bool IsExported(Sema *self, clang::FunctionDecl *FD) {        
+bool IsExported(Sema *self, clang::FunctionDecl *FD) {
   if (FD->hasAttr<HLSLShaderAttr>()) {
-		return true;
-	}
-  
+    return true;
+  }
+
   bool isMarkedStatic = false;
   bool isMarkedExport = false;
 
   if (auto cDecl = dyn_cast<CXXMethodDecl>(FD)) {
     if (cDecl->isStatic()) {
-        isMarkedStatic = true;
+      isMarkedStatic = true;
     }
   }
 
@@ -11419,30 +11419,26 @@ bool IsExported(Sema *self, clang::FunctionDecl *FD) {
   }
 
   if (isMarkedStatic && isMarkedExport) {
-    self->Diag(FD->getLocation(), diag::err_hlsl_varmodifiersna)
-        << "static"
-        << "export"
-        << "function";
-		return false;
-	}
-
+    self->Diag(FD->getLocation(), diag::err_hlsl_varmodifiersna) << "static"
+                                                                 << "export"
+                                                                 << "function";
+    return false;
+  }
 
   Linkage L = FD->getLinkageAndVisibility().getLinkage();
 
-  
   const auto *shaderModel =
       hlsl::ShaderModel::GetByName(self->getLangOpts().HLSLProfile.c_str());
   // case 1 requires special assignments to linkage
   if (L != ExternalLinkage && L != InternalLinkage) {
     if (shaderModel->IsSM60Plus()) {
       L = ExternalLinkage;
-    }
-		else {
+    } else {
       L = InternalLinkage;
     }
-  }   
-  
-  // now case 2 and 3 can apply and we can determine 
+  }
+
+  // now case 2 and 3 can apply and we can determine
   // whether the function is exported.
   if (L == InternalLinkage && isMarkedExport) {
     return true;
@@ -11467,7 +11463,7 @@ bool IsExported(Sema *self, clang::FunctionDecl *FD) {
 // references a patch constant function, there exists a function declaration
 // that could serve as a candidate to that patch constant function.
 void ValidatePatchConstantFunctionsExist(clang::Sema *self) {
-  
+
   for (auto decl : self->getASTContext().getTranslationUnitDecl()->decls()) {
     // TODO: improve condition so that only exported functions are checked,
     // instead of all functions. Issue: #5857
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -11405,18 +11405,8 @@ bool IsExported(Sema *self, clang::FunctionDecl *FD) {
     return true;
   }
 
-  bool isMarkedStatic = false;
-  bool isMarkedExport = false;
-
-  if (auto cDecl = dyn_cast<CXXMethodDecl>(FD)) {
-    if (cDecl->isStatic()) {
-      isMarkedStatic = true;
-    }
-  }
-
-  if (FD->hasAttr<HLSLExportAttr>()) {
-    isMarkedExport = true;
-  }
+  const bool isMarkedStatic = FD->getStorageClass() == SC_Static;
+  const bool isMarkedExport = FD->hasAttr<HLSLExportAttr>();
 
   if (isMarkedStatic && isMarkedExport) {
     self->Diag(FD->getLocation(), diag::err_hlsl_varmodifiersna) << "static"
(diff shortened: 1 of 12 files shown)


tools/clang/test/SemaHLSL/recursion/library_recursion_unexported.hlsl
```diff
@@ -0,0 +1,59 @@
...
+
+// expected-error@+1{{recursive functions are not allowed: entry function calls recursive function 'unexported_recurse'}}
+static void unexported_recurse(inout float4 f, float a) 
+{
+    if (a > 1)
+      unexported_recurse(f, a-1);
+    f = abs(f+a);
+}
+
+// expected-error@+1{{recursive functions are not allowed: entry function calls recursive function 'exported_recurse'}}
```
python3kgae, line 40: exported_recurse is not called by entry function here.
We need different error message here.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11400,16 +11400,78 @@ bool hlsl::DiagnoseNodeStructArgument(Sema *self, TemplateArgumentLoc ArgLoc,
...
 // that could serve as a candidate to that patch constant function.
 void ValidatePatchConstantFunctionsExist(clang::Sema *self) {
+
   for (auto decl : self->getASTContext().getTranslationUnitDecl()->decls()) {
     // TODO: improve condition so that only exported functions are checked,
     // instead of all functions. Issue: #5857
     if (FunctionDecl *FD = dyn_cast<FunctionDecl>(decl)) {
       // If there is no patch constant function, then we don't need to validate
       // anything.
+      if (!IsExported(self, FD)) {
```
python3kgae, line 11472: What if a function is not exported, but used by exported function?

bob80905: My understanding during implementation was that exported functions and all functions reachable from exported functions should be checked for recursion. I know for sure that exported functions should be checked for recursion. So I do think everything reachable should be checked too because otherwise we would be missing a diagnostic on actual recursion that takes place during execution, right?

tools/clang/test/SemaHLSL/recursion/library_recursion_unexported.hlsl
bob80905: True, I'll fix that.

tools/clang/test/SemaHLSL/recursion/library_recursion_unexported.hlsl
```diff
@@ -0,0 +1,59 @@
...
+// unexported_recurse, so we expect an error.
+// Thirdly, unreachable functions with internal linkage should not 
+// be tested for recursion in library shaders. 
+// In this case, unreachable_unexported_recurse is an example, and we 
+// do not expect any compilation errors for that function.
+// Finally, functions with unspecified linkage will default to external
+// for library shaders, so we expect errors on unreachable functions
+// that are recursive
+
+// expected-error@+1{{recursive functions are not allowed: export function calls recursive function 'unreachable_unexported_recurse_external'}}
```
python3kgae, line 17: There is no export function which called unreachable_unexported_recurse_external ?

bob80905: My understanding is that this function has external linkage, so it will be exported. The function calls itself, so the diagnostic is accurate. Is this incorrect?

python3kgae: Only export function be exported unless -default-linkage is set to external.
We need add hlsl spec about library profile to make things clear.
The name 'external' is really make things hard to understand here :(

tex3d (changes requested):
Generally, I think we need to rework some things:

`IsExported` isn't correct.  We must handle `-exports`, `-export-shaders-only` and `-default-linkage` options.  I've suggested ways of handling `-export-shaders-only` and `-default-linkage` options, but more work will need to be done for the `-exports` option.  We may need to split `dxilutil::ExportMap` into a part that parses and validates the `-exports` option, making the needed details available to Sema, and the part used during CodeGen with `llvm::Function` pointers.

In addition, we need to reuse `CallGraphWithRecurseGuard` to avoid redundant traversals, and check patch constant functions.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15797,16 +15857,20 @@ clang::FunctionDecl *ValidateNoRecursion(clang::Sema *self,
...
-    // TODO: improve condition so that only exported functions are checked,
-    // instead of all functions. Issue: #5857
     if (FunctionDecl *FD = dyn_cast<FunctionDecl>(decl)) {
       // returns the first recursive function declaration detected
       // from this function declaration FD, and determines whether
       // the recursion was detected in the patch-constant function
+      if (!IsExported(self, FD)) {
+        continue;
+      }
       FunctionDecl *pResult = ValidateNoRecursion(self, FD);
```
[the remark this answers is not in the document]
tex3d: We also need to add a map to `CallGraphWithRecurseGuard` to keep track of recursive functions we have already identified, since this was originally designed with only one entry point in mind.  We can use this map at the beginning of CheckRecursion to early out if we already checked this function before.

Something like this:
```diff
@@ -2916,8 +2916,9 @@ struct CallNode {
 typedef ::llvm::DenseMap<FunctionDecl *, CallNode> CallNodes;
 typedef ::llvm::SmallPtrSet<Decl *, 8> FnCallStack;
 typedef ::llvm::SmallPtrSet<FunctionDecl *, 128> FunctionSet;
 typedef ::llvm::SmallVector<FunctionDecl *, 32> PendingFunctions;
+typedef ::llvm::DenseMap<FunctionDecl *, FunctionDecl *> FunctionMap;
 
 // Returns the definition of a function.
 // This serves two purposes - ignore built-in functions, and pick
 // a single Decl * to be used in maps and sets.
@@ -2990,21 +2991,28 @@ public:
 class CallGraphWithRecurseGuard {
 private:
   CallNodes m_callNodes;
   FunctionSet m_visitedFunctions;
+  FunctionMap m_functionsCheckedForRecursion;
 
   FunctionDecl *CheckRecursion(FnCallStack &CallStack, FunctionDecl *D) const {
+    auto it = m_functionsCheckedForRecursion.find(D);
+    if (it != m_functionsCheckedForRecursion.end())
+      return it->second;
     if (CallStack.insert(D).second == false)
       return D;
     auto node = m_callNodes.find(D);
     if (node != m_callNodes.end()) {
       for (FunctionDecl *Callee : node->second.CalleeFns) {
         FunctionDecl *pResult = CheckRecursion(CallStack, Callee);
-        if (pResult)
+        if (pResult) {
+          m_functionsCheckedForRecursion[D] = pResult;
           return pResult;
+        }
       }
     }
     CallStack.erase(D);
+    m_functionsCheckedForRecursion[D] = nullptr;
     return nullptr;
   }
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15797,16 +15857,20 @@ clang::FunctionDecl *ValidateNoRecursion(clang::Sema *self,
...
+// There is an assumption being made that this function will only be called
+// on library shader stages.Thus, there is no entry point, only exported
+// functions
 void ValidateNoRecursionInTranslationUnit(clang::Sema *self) {
   std::set<FunctionDecl *> FDecls;
   std::vector<FunctionDecl *> FDeclsVec;
   for (auto decl : self->getASTContext().getTranslationUnitDecl()->decls()) {
-    // TODO: improve condition so that only exported functions are checked,
-    // instead of all functions. Issue: #5857
     if (FunctionDecl *FD = dyn_cast<FunctionDecl>(decl)) {
```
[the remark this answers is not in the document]
bob80905: The intention was to iterate over all potentially exported function decls, including inside namespaces, or class methods. I did not test those cases, thanks for the recommendation.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11400,14 +11400,74 @@ bool hlsl::DiagnoseNodeStructArgument(Sema *self, TemplateArgumentLoc ArgLoc,
...
+  if (L == InternalLinkage) {
+    return false;
+  }
+
+  if (L == ExternalLinkage) {
+    return true;
+  }
+  // unreachable but necessary to prevent warnings
+  return true;
+}
```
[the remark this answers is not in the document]
bob80905: Talked offline with @tex3D, we decided that it was too complex to use ExportMap by copying it over from DXCOpts to LangOpts (since dxilutil's ExportMap implementation has a memebr that cannot be copied, which is an llvm string map).
We might consider re-addressing this in upstream, but there's also the issue where we don't have mangled names in Sema, and so the export map may not be sufficient information to tell which functions are exported (in case of a name collision).

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
bob80905: Added some test cases, classes / structs / namespaces should be tested for recursion when the default is external linkage.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15790,41 +15888,71 @@ clang::FunctionDecl *ValidateNoRecursion(clang::Sema *self,
...
+                                      e = declContext->decls_end();
+           j != e; ++j) {
+        // Check if the declaration is a function declaration
+        if (FunctionDecl *functionDecl = dyn_cast<FunctionDecl>(*j)) {
+          if (!functionDecl->hasBody() ||
+              !IsExported(self, functionDecl,
+                          getDefaultLinkageExternal(self))) {
+            continue;
+          }
+          AllExportedFDecls.push_back(functionDecl);
```
tex3d, line 15925: This gathers one level deep, but I think a recursive worklist pattern should be used to go deeper, and it's less code.

Basically, that's:
```cpp
  llvm::SmallVector<DeclContext *, 8> Worklist;
  Worklist.push_back(self->getASTContext().getTranslationUnitDecl());
  while (Worklist.size()) {
    DeclContext *DC = Worklist.pop_back_val();
    for (auto *D : DC->decls()) {
      if (auto *FD = dyn_cast<FunctionDecl>(D)) {
        if (FD->hasBody() &&
            IsExported(self, FD, getDefaultLinkageExternal(self)))
          AllExportedFDecls.push_back(FD);
      } else if (auto *DC2 = dyn_cast<DeclContext>(D)) {
        Worklist.push_back(DC2);
      }
    }
  }
```

Also, how is the main function checked for non-lib case?  I think this function could be split into a function that verifies recursion on a set of decls, and things that gather those decls we need to check.  Unification of code checking recursion and so on for all functions, reachability between hull shaders and patch constant functions, between library and non-library compilations is also possible, and desirable.  It's still a bit of a mess right now.

It appears to me that we have:
- if library
  - iterate all function decls, skipping no-body and internal linkage
    - if patch constant attr
      - look up patch constant function, error if not found
      - check patch constant function for recursion
      - check hull shader for recursion
      - check reachability between these
  - iterate all function decls, skipping no-body and internal linkage (again, but this time, go into decl context one level deep)
    - add function to AllExportedFDecls
  - check all functions in AllExportedFDecls for recursion (this will redundantly check hull shaders)
- else (not library)
  - lookup entry function
  - check for recursion
  - if hull shader
    - look up patch constant function, error if not found
    - check patch constant function for recursion
    - check reachability between these

This could be simplified to:
- gather function phase
  - if library
    - use recursive worklist pattern to gather functions
  - else
    - look up and add entry function
- check functions phase
  - check for recursion
  - if hull shader
    - get hull shader function, error if not found
    - check for recursion
    - check for reachability between these

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15790,41 +15888,71 @@ clang::FunctionDecl *ValidateNoRecursion(clang::Sema *self,
...
     }
   }
 
+  for (auto FD : AllExportedFDecls) {
+    // returns the first recursive function declaration detected
+    // from this function declaration FD, and determines whether
+    // the recursion was detected in the patch-constant function
+    if (!FD->hasBody() ||
+        !IsExported(self, FD, getDefaultLinkageExternal(self))) {
+      continue;
```
tex3d, line 15937: Presumably, this was already checked when adding to `AllExportedFDecls`, so you don't need to do it here again.

bob80905:
clang format checker doesn't seem to function properly when the base branch is changed, so I need to make another PR, which is here:
https://github.com/microsoft/DirectXShaderCompiler/pull/6009

PR #5984, opened 2023-11-07
[CI] add workflow to apply clang-format diff

python3kgae:
When hit clang-format error in a pull request, edit the comment with diff.  
Add @apply-format at the beginning of the comment and Update comment.  
 

New added "Apply code format" workflow will be triggered.  
It will checkout the code from compare branch of the pull request, apply the diff,  
then commit and push it to the compare branch to fix the format issue.  
  
The process should be get clang-format diff comment first.  
Then review the diff in the comment.
If the diff looks good, edit and add @apply-format to apply the diff.

### DIFF
--- a/utils/git/code-format-diff-apply.py
+++ b/utils/git/code-format-diff-apply.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+#
+# ====- code-format-diff-apply, apply diff from comment --*- python -*-------==#
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# ==-------------------------------------------------------------------------==#
+
+import argparse
+import os
+import re
+import subprocess
+import sys
+import tempfile
+from functools import cached_property
+
+import github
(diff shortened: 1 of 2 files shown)


utils/git/code-format-diff-apply.py
```diff
@@ -0,0 +1,93 @@
...
+        # run git apply tmp.name
+        apply_cmd = [
+            "git",
+            "apply",
+            tmp.name
+        ]
+        run_cmd(apply_cmd)
+
+    # run git add .
+    add_cmd = [
```
bob80905, line 70: One thing I ran into when manually clang formatting was outdated SPIR-V headers. In fact, I think it is quite common  to accidentally add modified SPIR-V headers through git add . when it's not intended. Is there a way to loop through the actual diff text and collect all the files that are mentioned in the diff, rather than running git add .? Maybe with a regex? Then we can run git add <filename> for each one.

bob80905 (changes requested):

bob80905 (approved):

utils/git/code-format-diff-apply.py
bob80905: The build machines should checkout correctly, and in theory it shouldn't have any out-of-date spir-v headers or any other files that may be floating around that weren't impacted by clang-format, so my concern is resolved.

PR #5992, opened 2023-11-09
[Sema] Move more diagnostics over

### LINKED ISSUE #5990: Move remaining SM 6.8 diagnostics to Sema and add test coverage
**Description**
The following diagnostics should be moved from CodeGen to Sema and need to have tests added: 

https://github.com/microsoft/DirectXShaderCompiler/blob/d5ddb1e9dff6fdd5f8f0f6e874970b4ad9a63c58/tools/clang/lib/CodeGen/CGHLSLMS.cpp#L1405

https://github.com/microsoft/DirectXShaderCompiler/blob/d5ddb1e9dff6fdd5f8f0f6e874970b4ad9a63c58/tools/clang/lib/CodeGen/CGHLSLMS.cpp#L1555

https://github.com/microsoft/DirectXShaderCompiler/blob/d5ddb1e9dff6fdd5f8f0f6e874970b4ad9a63c58/tools/clang/lib/CodeGen/CGHLSLMS.cpp#L2519

https://github.com/microsoft/DirectXShaderCompiler/blob/d5ddb1e9dff6fdd5f8f0f6e874970b4ad9a63c58/tools/clang/lib/CodeGen/CGHLSLMS.cpp#L2680

For more context see discussion about `tools/clang/lib/CodeGen/CGHLSLMS.cpp` in PR https://github.com/microsoft/DirectXShaderCompiler/pull/5982.

bob80905:
This PR moves over some more diagnostics from CGHLSLMS.cpp. Some diagnostics are moved into DiagnoseEntryAttrAllowedOnStage, to specifically diagnose when an attribute appears on a shader stage that it shouldn't appear on. Other diagnostics that are covered in Sema are removed in CGHLSLMS.
Fixes #5990


tools/clang/test/HLSL/workgraph/node_track_rw_input_sharing.hlsl
```diff
@@ -0,0 +1,32 @@
+// RUN: %clang_cc1 -HV 2021 -verify %s
```
python3kgae, line 1: verify test should be OK to move to SemaHLSL/hlsl/workgraph as lit test since it not run hash stability test.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11448,6 +11464,103 @@ void DiagnoseEntryAttrAllowedOnStage(clang::Sema *self,
...
+        switch (shaderKind) {
+        case DXIL::ShaderKind::Node:
+          break;
+        default:
+          self->Diag(pAttr->getRange().getBegin(),
+                     diag::err_hlsl_attribute_unsupported_stage)
+              << "NodeMaxRecursionDepth"
+              << "node";
+        }
+        break;
```
llvm-beanz, line 11563:
```suggestion
      case clang::attr::HLSLNodeLaunch:
      case clang::attr::HLSLNodeIsProgramEntry:
      case clang::attr::HLSLNodeId:
      case clang::attr::HLSLNodeLocalRootArgumentsTableIndex:
      case clang::attr::HLSLNodeShareInputOf:
      case clang::attr::HLSLNodeDispatchGrid:
      case clang::attr::HLSLNodeMaxDispatchGrid:
      case clang::attr::HLSLNodeMaxRecursionDepth: {
        switch (shaderKind) {
        case DXIL::ShaderKind::Node:
          break;
        default:
          self->Diag(pAttr->getRange().getBegin(),
                     diag::err_hlsl_attribute_unsupported_stage)
              << pAttr->getName()
              << "node";
        }
        break;
```

We should be able to collapse this all down.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/missing_node_attribute_error_msg.hlsl
```diff
@@ -13,11 +13,11 @@
...
 void secondNode()
 {
-    // CHECK-DAG: Attribute nodeisprogramentry only applies to node shaders (indicated with '[shader("node")]')
-    // CHECK-DAG: Attribute nodeid only applies to node shaders (indicated with '[shader("node")]')
-    // CHECK-DAG: Attribute nodelocalrootargumentstableindex only applies to node shaders (indicated with '[shader("node")]')
-    // CHECK-DAG: Attribute nodeshareinputof only applies to node shaders (indicated with '[shader("node")]')
-    // CHECK-DAG: Attribute nodemaxrecursiondepth only applies to node shaders (indicated with '[shader("node")]')
-    // CHECK-DAG: Attribute nodedispatchgrid only applies to node shaders (indicated with '[shader("node")]')
-    // CHECK-DAG: Attribute nodemaxdispatchgrid only applies to node shaders (indicated with '[shader("node")]')
+    // CHECK-DAG: attribute NodeIsProgramEntry only allowed on node shaders
```
python3kgae, line 16: And this test too.
Could it be a verify test?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15756,15 +15891,18 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
...
+  // Don't add the attribute if it already exists as an attribute on the decl,
+  // and emit an error. In the special case that the target profile is compute
+  // and the entry decl already has a node shader attr, don't do anything
   if (currentShaderAttr) {
     llvm::StringRef currentFullName = currentShaderAttr->getStage();
     if (currentFullName != fullName) {
-      S.Diag(currentShaderAttr->getLocation(),
-             diag::err_hlsl_profile_conflicts_with_shader_attribute)
-          << fullName << profile << currentFullName << EntryPointName;
+      if (!(SM->IsCS() && strstr(currentFullName.data(), "node"))) {
```
llvm-beanz, line 15900: I'm a little confused by this logic. We don't allow node entries to be used as compute entries.

bob80905: This logic was meant to be the move from CGHLSLMS.cpp, where this diagnostic is emitted:
`Diags.getCustomDiagID(DiagnosticsEngine::Error,
                                    "Node function as global entry point must "
                                    "be compiled to compute shader target");`
I believe my logic accurately detects when this case happens, and emits a diagnostic. Is it true that we don't allow a Node function as a global entry point to be compiled to a compute shader target?

tools/clang/test/HLSLFileCheck/hlsl/workgraph/entry_node_attr_on_non_compute_stage.hlsl
```diff
@@ -0,0 +1,21 @@
...
+
+void loadStressWorker(NodeOutput<loadStressRecord> x) {
+    return;
+}
+
+[Shader("node")]
+[NodeMaxDispatchGrid(3, 1, 1)]
+[NumThreads(16, 1, 1)]
+void main(
+    // CHECK: error: entry type 'pixel' from profile 'ps_6_6' conflicts with shader attribute type 'node' on entry function 'main'.
```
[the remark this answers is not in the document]
bob80905: This needs to be a filecheck test because filecheck tests provide a way to input the target shader stage. I need the shader stage to be valid, and to not be compute, in order to run this test. AFAIK, there's no way to specify a valid, non-compute shader stage in a verifier test.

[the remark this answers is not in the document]
python3kgae: You can use %dxc  -verify in lit test which has all the options.

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: We disallowed mixing the entry points and removed the implicit input record arguments, so I believe this code path also shouldn’t be supported.

@tex3d & @pow2clk, thoughts?

pow2clk: It is slightly different from the decision we made before to disallow node and compute shader attributes on the same entry point, but I think we should keep the existing behavior that disallows using an entry point for a target that doesn't match its shader attribute.

bob80905: Fixed.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1877,40 +1845,30 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
     }
   }
   if (const auto *pAttr = FD->getAttr<HLSLNodeMaxRecursionDepthAttr>()) {
     if (isNode)
       funcProps->Node.MaxRecursionDepth = pAttr->getCount();
-    else
-      ReportMissingNodeDiag(Diags, pAttr);
   }
   if (!FD->getAttr<HLSLNumThreadsAttr>()) {
     if (isNode) {
```
llvm-beanz, line 1874: GitHub won't let me write a suggestion for this because this block spans across a lot of code, but there's a really gross pattern here that we should clean up.

We have a chain of statements here that are basically:

```c++
if (/*getAttr succeedes*/)
  if (isNode)
    // do something
```

Clang's getAttr isn't crazy slow, but it is silly to do a bunch of attribute lookups if we're never going to do anything with the result.

Can we instead rewrite this to remove all the inner `if (isNode)` checks and just fold all the code up under the `if (isNode)` block on line 1816 above?

bob80905: Done

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11430,12 +11430,28 @@ void ValidatePatchConstantFunctionsExist(clang::Sema *self) {
...
-                                     DXIL::ShaderKind shaderKind) {
+                                     DXIL::ShaderKind shaderKind,
+                                     bool isActiveEntry) {
 
   if (entryPointDecl->hasAttrs()) {
     for (Attr *pAttr : entryPointDecl->getAttrs()) {
       switch (pAttr->getKind()) {
-
+      // if this is a "node" shader and the target stage isn't compute nor
+      // library, emit a diagnostic
```
tex3d, line 11440: No, we don't support node shader entry points compiled to a compute shader target.

Note: "library" isn't a shader "stage".  It may contain entry points for various shader stages, but it isn't a stage itself.

In `DiagnoseEntryAttrAllowedOnStage`, `shaderKind` comes from the attribute on the function.  Either this attribute was provided by the user, or it was added to the function if you're compiling directly to a non-library target and there's no existing or conflicting attribute already present.

If you try to compile a node shader to a compute target, code in `TryAddShaderAttrFromTargetProfile` should already have emitted a diagnostic that the profile conflicts with the shader attribute on the function.  Note that the comment above that diagnostic is incorrect: the two lines talking about a special case for compute+node should be removed.

No conflicting attribute should exist at this point, because that would have been caught earlier.

Therefore, there is no need to do anything with the `HLSLShader` attribute here.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15684,12 +15718,11 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
               << HLSLNodeObjectAttr::ConvertRecordTypeToStr(Kind);
       }
     }
-
+    unsigned int ArgIdx = 0;
```
tex3d, line 15721: `ArgIdx` is basically used as a for loop variable.  I think the original purpose was to do something with this index, such as track whether the parameter referenced from `[MaxRecordsSharedWith(...)]` was the parameter with this attribute.  But as the comment says, this is checked elsewhere.  So, this `ArgIdx` is just used as a variable to look up each parameter in the function decl.

I would recommend the while loops using this be converted to for loops.  Since FunctionDecl has a `params()` method that gets the params, you can do away with `ArgIdx` entirely by using a different iteration pattern: `for (auto *ParamDecl : FD->params())`.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15716,6 +15749,56 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
...
+              if (const ClassTemplateSpecializationDecl *templateDecl =
+                      dyn_cast<ClassTemplateSpecializationDecl>(CXXRD)) {
+
+                auto &TemplateArgs = templateDecl->getTemplateArgs();
+                DXASSERT(
+                    TemplateArgs.size() == 1,
+                    "Input record types need to have one template argument");
+                auto &Rec = TemplateArgs.get(0);
+                clang::QualType RecType = Rec.getAsType();
+                if (RecordDecl *RD = RecType->getAs<RecordType>()->getDecl()) {
```
tex3d, line 15780: I think this functionality of getting the RecordDecl for the user defined record type of `NodeOutput` or `NodeInputRecord` types should be wrapped in a utility function in `HLSLTypes.cpp`.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15716,6 +15749,56 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
                diag::err_hlsl_maxrecordssharedwith_references_invalid_arg);
       }
     }
+
+    // Make sure NodeTrackRWInputSharing attribute cannot be applied to
+    // Input Records that are not RWDispatchNodeInputRecord
+    ArgIdx = 0;
+    while (ArgIdx < FD->getNumParams()) {
```
tex3d, line 15756: This is already inside the argument loop.  Why are we iterating through all the arguments for each argument here?  I believe you just want to check the current argument.  Take a look at how `ParamTy` is used, which is only set for the current argument in the outer loop.  The inner loop code doesn't make sense based on this.

tools/clang/test/SemaHLSL/hlsl/workgraph/node_track_rw_input_sharing.hlsl
```diff
@@ -0,0 +1,31 @@
...
+  uint a;
+  bool b;
+  uint3 grid : SV_DispatchGrid;
+};
+
+struct RECORD2
+{
+  uint a;
+  bool b;
+};
```
tex3d, line 14: It doesn't look like `RECORD` or `RECORD2` are used.  Should probably delete them.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -68,6 +68,29 @@ bool IsHLSLVecMatType(clang::QualType type) {
   return false;
 }
 
+clang::RecordDecl *getRecordTypeFromParameterType(clang::QualType ParamTy) {
```
tex3d, line 71: I think the name of this function should be more like "GetRecordTypeFromNodeObjectType" instead (and ParamTy could be ObjectTy).  The type being passed in is some node object type with a record template parameter.  This function does not operate on a "ParameterType".

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -68,6 +68,29 @@ bool IsHLSLVecMatType(clang::QualType type) {
...
+  if (ParamTy->isStructureOrClassType()) {
+    if (const CXXRecordDecl *CXXRD =
+            ParamTy.getCanonicalType()->getAsCXXRecordDecl()) {
+
+      if (const ClassTemplateSpecializationDecl *templateDecl =
+              dyn_cast<ClassTemplateSpecializationDecl>(CXXRD)) {
+
+        auto &TemplateArgs = templateDecl->getTemplateArgs();
+        DXASSERT(TemplateArgs.size() == 1,
+                 "Input record types need to have one template argument");
```
tex3d, line 82: The type passed in might not be an "Input record type", it could be a NodeOutput or even a NodeOutputRecord type.

We should probably also have an assert at the beginning of this function that the type passed in is a node object type, so it isn't misused for another type.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -68,6 +68,48 @@ bool IsHLSLVecMatType(clang::QualType type) {
...
+            ObjectTy.getCanonicalType()->getAsCXXRecordDecl()) {
+
+      if (const ClassTemplateSpecializationDecl *templateDecl =
+              dyn_cast<ClassTemplateSpecializationDecl>(CXXRD)) {
+
+        auto &TemplateArgs = templateDecl->getTemplateArgs();
+        DXASSERT(TemplateArgs.size() == 1,
+                 "Record types need to have one template argument");
+        auto &Rec = TemplateArgs.get(0);
+        clang::QualType RecType = Rec.getAsType();
```
tex3d, line 103: Actually, there's no need for the DXASSERT, we really care that the args are >= 1, and there is an assert in `get()` already for that.  Also, `operator[]` is equivalent to `get()`, and you don't need the intermediate result, so you could `getAsType()` in the same expression.  So these lines could be:

```suggestion
        clang::QualType RecType = TemplateArgs[0].getAsType();
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11430,12 +11430,12 @@ void ValidatePatchConstantFunctionsExist(clang::Sema *self) {
 // should exist on this shader stage
 void DiagnoseEntryAttrAllowedOnStage(clang::Sema *self,
                                      FunctionDecl *entryPointDecl,
-                                     DXIL::ShaderKind shaderKind) {
+                                     DXIL::ShaderKind shaderKind,
+                                     bool isActiveEntry) {
```
tex3d, line 11434: What's the point of adding `isActiveEntry` now?  It isn't used, so I don't think we should add it until it has a real purpose.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15756,15 +15793,17 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
   }
 
   HLSLShaderAttr *currentShaderAttr = FD->getAttr<HLSLShaderAttr>();
-  // Don't add the attribute if it already exists as an attribute on the decl.
-  // In the special case that the target profile is compute and the
-  // entry decl already has a node shader attr, don't do anything
+  // Don't add the attribute if it already exists as an attribute on the decl,
+  // and emit an error. In the special case that the target profile is compute
+  // and the entry decl already has a node shader attr, don't do anything
```
tex3d, line 15798: > ... In the special case that the target profile is compute
  // and the entry decl already has a node shader attr, don't do anything

There is no special case anymore.  We should get rid of this sentence entirely.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15756,15 +15793,17 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
...
   if (currentShaderAttr) {
     llvm::StringRef currentFullName = currentShaderAttr->getStage();
     if (currentFullName != fullName) {
+
       S.Diag(currentShaderAttr->getLocation(),
              diag::err_hlsl_profile_conflicts_with_shader_attribute)
           << fullName << profile << currentFullName << EntryPointName;
+      return;
```
tex3d, line 15806: We would have returned already in the outer `if` block, so I'm not sure what difference this return makes.

Also of note, below that outer return, we set `isActiveEntry = true`.  I think this should be set before looking for a current shader attribute, since at that point, we know this is the entry point referred to by the HLSLEntryFunction setting.  Right now, if you have a shader attribute on your function that you compile through an ordinary shader target, this code will set isActiveEntry to false, which is not the desired behavior.

tex3d (changes requested):

python3kgae (approved):

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -859,6 +859,27 @@ bool GetHLSLSubobjectKind(clang::QualType type,
...
+            ObjectTy.getCanonicalType()->getAsCXXRecordDecl()) {
+
+      if (const ClassTemplateSpecializationDecl *templateDecl =
+              dyn_cast<ClassTemplateSpecializationDecl>(CXXRD)) {
+
+        auto &TemplateArgs = templateDecl->getTemplateArgs();
+        clang::QualType RecType = TemplateArgs[0].getAsType();
+        if (RecordDecl *RD = RecType->getAs<RecordType>()->getDecl()) {
+          return RD;
+        }
```
tex3d, line 876:
```suggestion
        if (const RecordType *RT = RecType->getAs<RecordType>())
          return RT->getDecl();
```
`getAs<RecordType>()` is the part that may return nullptr if not the right type, so the if should be on that.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15746,6 +15782,8 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
     return;
   }
 
+  isActiveEntry = true;
+
   std::string profile = S.getLangOpts().HLSLProfile;
   const ShaderModel *SM = hlsl::ShaderModel::GetByName(profile.c_str());
   const llvm::StringRef fullName = ShaderModel::FullNameFromKind(SM->GetKind());
```
tex3d, line 15789: Below here we have:
```cpp
  // don't add the attribute for an invalid profile, like library
  if (fullName.empty()) {
    return;
  }
```

However, I believe this should be an assert or llvm_unreachable, since we shouldn't have called this function in that case.

pow2clk (approved):
Looks like I never submitted my review 😣

Sorry about that. You can take or leave my suggestions

tex3d (approved):
LGTM!

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1841,79 +1816,46 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
       funcProps->Node.MaxDispatchGrid[1] = pAttr->getY();
       funcProps->Node.MaxDispatchGrid[2] = pAttr->getZ();
-    } else {
-      ReportMissingNodeDiag(Diags, pAttr);
     }
-  }
-  if (const auto *pAttr = FD->getAttr<HLSLNodeMaxRecursionDepthAttr>()) {
-    if (isNode)
+    if (const auto *pAttr = FD->getAttr<HLSLNodeMaxRecursionDepthAttr>()) {
       funcProps->Node.MaxRecursionDepth = pAttr->getCount();
```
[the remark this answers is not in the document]
bob80905: I don't think there needs to be an assert since isNode is checked on line 1816 and this is all contained inside that if.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11448,6 +11448,25 @@ void DiagnoseEntryAttrAllowedOnStage(clang::Sema *self,
...
+      }
+      case clang::attr::HLSLNodeLaunch:
+      case clang::attr::HLSLNodeIsProgramEntry:
+      case clang::attr::HLSLNodeId:
+      case clang::attr::HLSLNodeLocalRootArgumentsTableIndex:
+      case clang::attr::HLSLNodeShareInputOf:
+      case clang::attr::HLSLNodeDispatchGrid:
+      case clang::attr::HLSLNodeMaxDispatchGrid:
+      case clang::attr::HLSLNodeMaxRecursionDepth: {
+        switch (shaderKind) {
```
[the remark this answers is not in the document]
bob80905: Will fix

PR #5995, opened 2023-11-09
[Test] add test for different matrix representations in refection

### LINKED ISSUE #5696: Test different matrix representations in reflection
The introduction of a metadata field to preserve the vector size when converted to an array added a list of different matrix/vector representations that reflection has to deal with: 
https://github.com/microsoft/DirectXShaderCompiler/blob/58a9ac3522c899ee52f091de4fd033e12a710afc/lib/HLSL/DxilContainerReflection.cpp#L954

There isn't obvious testing for all these variants

We need comprehensive testing for these variants to ensure they are handled correctly and appropriately represented in reflection.

There may be existing testing for some of these, but there doesn't seem to be any for the vector size field annotation that was introduced specifically. Though that is likely a separate issue.

**Additional context**
See internal PR 4476

python3kgae:
Not all matrix representations could be generated with current dxc, so use IR test here.

Fixes #5696


tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_array_of_vector.ll
```diff
@@ -0,0 +1,114 @@
...
+; CHECK-NEXT:                    Rows: 3
+; CHECK-NEXT:                    Columns: 1
+; CHECK-NEXT:                    Members: 0
+; CHECK-NEXT:                    Offset: 148
+; CHECK-NEXT:              }
+; CHECK-NEXT:            CBuffer: s
+
+%hostlayout.s = type { %hostlayout.struct.S2 }
+
+%hostlayout.struct.S2 = type { [4 x <3 x i32>], [2 x [3 x <1 x i32>]], <3 x i32> }
```
tex3d, line 69: `[2 x [3 x <1 x i32>]]` looks like it has rows and cols reversed.  Shouldn't it be `[2 x [1 x <3 x i32>]]`?

tex3d: Though, this reveals an interesting case, which is the native layout in col major will be transposed like this.  Still, it's not consistent with the layout of the other matrices here.  Maybe there should be both versions.  Part of the point of the 1x3 case is to catch the folding of 1d array dimensions (where there was a bug interpreting this at some point): like `[2 x [1 x <3 x int>]]` was changed to `[2 x <3 x int>]`.

The reflection code must count components to determine where the boundary between array and matrix is so it can be robust to different forms and even row major and column major layouts.  The matrix dimensions are then read from the type annotation, so the inner component layout doesn't impact the reflection.

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_array_of_array.ll
```diff
@@ -0,0 +1,114 @@
...
+; CHECK-NEXT:                    Rows: 3
+; CHECK-NEXT:                    Columns: 1
+; CHECK-NEXT:                    Members: 0
+; CHECK-NEXT:                    Offset: 148
+; CHECK-NEXT:              }
+; CHECK-NEXT:            CBuffer: s
+
+%hostlayout.s = type { %hostlayout.struct.S2 }
+
+%hostlayout.struct.S2 = type { [4 x [3 x i32]], [2 x [3 x [1 x i32]]], [3 x i32] }
```
tex3d, line 69: Also second (`[2 x [3 x [1 x i32]]]`) and third (`[3 x i32]`) are col-major layout, while first matrix is in row-major layout.  Is this on purpose to mix-in some layout change testing, or was it an accident?

tex3d: Oh, wait, I see the first is row_major, so you're expecting these to be in the storage layout.  We should test them in opposite layout from storage layout to ensure robustness against that as well, since they normally would be in logical layout, not necessarily storage layout.

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_array_of_vector.ll
python3kgae: Fixed the 1 x <3 x i32> part.
2 x <3 x i32> is in type_sys_dump_mat_as_vector.ll.

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_array_of_array.ll
python3kgae: Fixed and add test for different orientation.

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_array_of_vector.ll
```diff
@@ -0,0 +1,114 @@
...
+; CHECK-NEXT:                    Rows: 3
+; CHECK-NEXT:                    Columns: 1
+; CHECK-NEXT:                    Members: 0
+; CHECK-NEXT:                    Offset: 148
+; CHECK-NEXT:              }
+; CHECK-NEXT:            CBuffer: s
+
+%hostlayout.s = type { %hostlayout.struct.S2 }
+
+%hostlayout.struct.S2 = type { [4 x <3 x i32>], [2 x [1 x <3 x i32>]], <3 x i32> }
```
tex3d, line 69: Above, the comment says "one row: <cols x float>", but the third matrix here is 3 rows, 1 column, in column major.  Shouldn't this be `[3 x <1 x i32>]`?  Most other cases test the `<3 x i32>`, but not `[3 x <1 x i32>]` or `[3 x [1 x i32]]`.

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_flattened_array.ll
```diff
@@ -0,0 +1,114 @@
...
+; CHECK-NEXT:                    Rows: 3
+; CHECK-NEXT:                    Columns: 1
+; CHECK-NEXT:                    Members: 0
+; CHECK-NEXT:                    Offset: 148
+; CHECK-NEXT:              }
+; CHECK-NEXT:            CBuffer: s
+
+%hostlayout.s = type { %hostlayout.struct.S2 }
+ 
+%hostlayout.struct.S2 = type { [12 x i32], [2 x [3 x i32]], <3 x i32> }
```
tex3d, line 69: Shouldn't `<3 x i32>` be `[3 x i32]` here?

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_flattened_array.ll
```diff
@@ -0,0 +1,114 @@
...
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+;struct S2 {
+;    row_major int4x3 m;
+;	int1x3 m1[2];
+;	int3x1 m2;
+;};
+
+; Make sure cbuffer reflection is correct with Flattened vector as array: [(rows*cols) x float] and High level struct stripped, one row: <cols x float>
```
tex3d, line 13: I don't think `, one row: <cols x float>` in this comment should be here.

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_flattened_array_2.ll
```diff
@@ -0,0 +1,114 @@
...
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+;struct S2 {
+;   int4x3 m;
+;	row_major int1x3 m1[2];
+;	row_major int3x1 m2;
+;};
+
+; Make sure cbuffer reflection is correct with Flattened vector as array: [(rows*cols) x float] and High level struct stripped, one row: <cols x float>
```
tex3d, line 13: I don't think `, one row: <cols x float>` in this comment should be here.

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_vector_2.ll
```diff
@@ -0,0 +1,114 @@
+; RUN: %dxa %s -o %t
+; RUN: %dxa %t -dumpreflection | FileCheck %s
+
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+;struct S2 {
+;   int4x3 m;
+;	row_major int1x3 m1[2];
+;	row_major int3x1 m2;
```
tex3d, line 10: Since row_major/column_major makes no difference to the type at the moment, shouldn't we verify that it works whether or not the llvm type is expressed in a transposed way?  You were doing that (accidentally) before.  I'd hate to increase the number of variants of this test case, since it's already getting unwieldy.  The _2 variants appear to just swap row_major and column_major (default) on fields in HLSL, but the llvm form is identical  That's good to test once, but could be tested separately from other dimension tests, right?  In other words, we could have one test that ensures type annotation has correct layout for row or column major, and that reflection is handled whether that layout is expressed in llvm as the row or column major.  Then we shouldn't need each test to have the `_2` variant that tests with the HLSL row/column major swapped, but same IR otherwise for each test scenario.

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_array_of_vector.ll
```diff
@@ -0,0 +1,114 @@
...
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+;struct S2 {
+;    row_major int4x3 m;
+;	int1x3 m1[2];
+;	int3x1 m2;
+;};
+
+; Make sure cbuffer reflection is correct with High level struct stripped: [rows x <cols x float>] and High level struct stripped, one row: <cols x float>
```
tex3d, line 13: I don't think this part of the comment is right: `one row: <cols x float>`.  This applies to the _2 variant too.

We will definitely test the single-vector case, so if we have a chance to test a `[3 x <1 x i32>]` or `[1 x <3 x i32>]`, we shouldn't squander it, right?

tools/clang/test/CodeGenDXIL/reflection/type_sys_dump_mat_as_array_of_array.ll
```diff
@@ -0,0 +1,114 @@
...
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+;struct S2 {
+;    row_major int4x3 m;
+;	int1x3 m1[2];
+;	int3x1 m2;
+;};
+
+; Make sure cbuffer reflection is correct with Vector as array: [rows x [cols x float]] and Vector as array, one row: [cols x float]
```
tex3d, line 13: The comment `one row: [cols x float]` seems to refer to the `[3 x i32]`, which is one column, not one row, right?

tex3d (approved):

pow2clk (approved):
Looks good!

PR #5996, opened 2023-11-09
[SM6.8] Diagnostic for output records with SV_DispatchGrid

### LINKED ISSUE #5958: [SM6.8] SV_DispatchGrid diagnostics needed for output records
**Description**
The current version of `DiagnoseDispatchGridSemantics` is used only to check input records when the SV_DispatchGrid field is actively used.  It would also be useful for the producers of such records to apply the same rules and diagnostics.  Note that for output records, you do not know the downstream launch type, so you have to assume SV_DispatchGrid may or may not be present, but if it is, it should be subject to the same rules (only one field with allowed type).

**Steps to Reproduce**
```hlsl
struct OutputRecord {
// expected-error@+1{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'uint4'}}
  uint4 grid : SV_DispatchGrid;
  float data;
};

[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeDispatchGrid(1,1,1)]
[NumThreads(32, 1, 1)]
void node(NodeOutput<OutputRecord> output) {}
```

**Actual Behavior**
No error produced.

**Environment**
- DXC version: latest `staging-sm-6.8` branch
- Host Operating System: any

(edited based on updated name in [pending PR](https://github.com/microsoft/DirectXShaderCompiler/pull/5953))

python3kgae:
Check both Input and Output node type when diagnose Semantic of DispatchGrid.

Fixes #5958

### DIFF
--- a/tools/clang/lib/AST/HlslTypes.cpp
+++ b/tools/clang/lib/AST/HlslTypes.cpp
@@ -975,6 +975,26 @@ QualType GetHLSLResourceTemplateParamType(QualType type) {
 QualType GetHLSLInputPatchElementType(QualType type) {
   return GetHLSLResourceTemplateParamType(type);
 }
+
+CXXRecordDecl *GetHLSLNodeRecordDecl(clang::QualType type) {
+  type = type.getCanonicalType();
+  const RecordType *RT = cast<RecordType>(type);
+  const ClassTemplateSpecializationDecl *templateDecl =
+      dyn_cast<ClassTemplateSpecializationDecl>(RT->getAsCXXRecordDecl());
+  if (!templateDecl)
+    return nullptr;
+  const TemplateArgumentList &argList = templateDecl->getTemplateArgs();
+  DXASSERT_NOMSG(argList.size() >= 1);
+  QualType TemplateParamTy = argList[0].getAsType();
+  if (const RecordType *NodeStructType =
+          TemplateParamTy->getAsStructureType()) {
+    if (CXXRecordDecl *NodeStructDecl =
+            dyn_cast<CXXRecordDecl>(NodeStructType->getDecl()))
+      return NodeStructDecl;
+  }
+  return nullptr;
+}
+
 unsigned GetHLSLInputPatchCount(QualType type) {
   type = type.getCanonicalType();
   const RecordType *RT = cast<RecordType>(type);
--- a/tools/clang/lib/AST/HlslTypes.cpp
+++ b/tools/clang/lib/AST/HlslTypes.cpp
@@ -976,25 +976,6 @@ QualType GetHLSLInputPatchElementType(QualType type) {
   return GetHLSLResourceTemplateParamType(type);
 }
 
-CXXRecordDecl *GetHLSLNodeRecordDecl(clang::QualType type) {
-  type = type.getCanonicalType();
-  const RecordType *RT = cast<RecordType>(type);
-  const ClassTemplateSpecializationDecl *templateDecl =
-      dyn_cast<ClassTemplateSpecializationDecl>(RT->getAsCXXRecordDecl());
-  if (!templateDecl)
-    return nullptr;
-  const TemplateArgumentList &argList = templateDecl->getTemplateArgs();
-  DXASSERT_NOMSG(argList.size() >= 1);
-  QualType TemplateParamTy = argList[0].getAsType();
-  if (const RecordType *NodeStructType =
-          TemplateParamTy->getAsStructureType()) {
-    if (CXXRecordDecl *NodeStructDecl =
-            dyn_cast<CXXRecordDecl>(NodeStructType->getDecl()))
-      return NodeStructDecl;
-  }
-  return nullptr;
-}
-
 unsigned GetHLSLInputPatchCount(QualType type) {
   type = type.getCanonicalType();
   const RecordType *RT = cast<RecordType>(type);
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -15666,37 +15666,21 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
     }
     // Diagnose dispatch grid semantics.
     bool Found = false;
-    for (FunctionDecl::param_iterator I = FD->param_begin(),
-                                      E = FD->param_end();
-         I != E; ++I) {
-      QualType ParamType = (*I)->getType().getCanonicalType();
+    for (ParmVarDecl *PD : FD->params()) {
+      QualType ParamType = PD->getType().getCanonicalType();
 
       // Find parameter that is the node input record
       if (hlsl::IsHLSLNodeType(ParamType)) {
         // Node records are template types
-        if (const RecordType *NodeRT = dyn_cast<RecordType>(ParamType)) {
-          if (const ClassTemplateSpecializationDecl *templateDecl =
-                  dyn_cast<ClassTemplateSpecializationDecl>(
-                      NodeRT->getDecl())) {
-
-            // Get the record struct
-            auto &TemplateArgs = templateDecl->getTemplateArgs();
-            DXASSERT_NOMSG(TemplateArgs.size() >= 1);
-            QualType Arg0Type = TemplateArgs.get(0).getAsType();
-            if (const RecordType *NodeStructType =
-                    Arg0Type->getAsStructureType()) {
-              if (CXXRecordDecl *NodeStructDecl =
-                      dyn_cast<CXXRecordDecl>(NodeStructType->getDecl())) {
-                bool OutputFound = false;
-                // Make sure there is exactly one SV_DispatchGrid semantics
-                // and it has correct type.
-                DiagnoseDispatchGridSemantics(
-                    S, NodeStructDecl,
-                    hlsl::IsHLSLNodeOutputType(ParamType) ? OutputFound
-                                                          : Found);
-              }
-            }
-          }
+        if (CXXRecordDecl *NodeStructDecl =
+                hlsl::GetHLSLNodeRecordDecl(ParamType)) {
+          // Diagnose any SV_DispatchGrid semantics used in record.
+          // OutputFound provides the output needed by
+          // DiagnoseDispatchGridSemantics, but isn't used otherwise.
+          bool OutputFound = false;
+          DiagnoseDispatchGridSemantics(
+              S, NodeStructDecl,
+              hlsl::IsHLSLNodeOutputType(ParamType) ? OutputFound : Found);
         }
       }
     }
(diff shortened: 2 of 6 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15541,46 +15541,48 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
...
+            if (nullptr != NodeStructType) {
+              CXXRecordDecl *NodeStructDecl =
+                  dyn_cast<CXXRecordDecl>(NodeStructType->getDecl());
+              if (nullptr != NodeStructDecl) {
+                bool OutputFound = false;
+                // Make sure there is exactly one SV_DispatchGrid semantics
+                // and it has correct type.
+                DiagnoseDispatchGridSemantics(
+                    S, NodeStructDecl,
+                    hlsl::IsHLSLNodeOutputType(ParamType) ? OutputFound
```
bob80905, line 15573: In this code, within this for loop, Found is set to false and never re-assigned, and OutputFound is set to false and never re-assigned. Then, they're both used at the bottom in DiagnoseDispatchGridSemantics(). I think both variables can be removed, and we can substitute `false` instead? We may need a refactor or to re-inspect the logic here, because I don't think these variables are being used correctly.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15541,46 +15541,48 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
...
+          if (const ClassTemplateSpecializationDecl *templateDecl =
+                  dyn_cast<ClassTemplateSpecializationDecl>(
+                      NodeRT->getDecl())) {
+
+            // Get the record struct
+            auto &TemplateArgs = templateDecl->getTemplateArgs();
+            DXASSERT_NOMSG(TemplateArgs.size() >= 1);
+            QualType Arg0Type = TemplateArgs.get(0).getAsType();
+            const RecordType *NodeStructType = Arg0Type->getAsStructureType();
+            if (nullptr != NodeStructType) {
```
bob80905, line 15564: Nit: I think the convention here is to put the cast in the if condition, like:
`if (CXXRecordDecl *NodeStructDecl = dyn_cast<CXXRecordDecl>(NodeStructType->getDecl())`
Same as above.

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: DiagnoseDispatchGridSemantics will update Found for input node types.

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: Fixed.

bob80905 (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15541,46 +15541,47 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
...
+                      NodeRT->getDecl())) {
+
+            // Get the record struct
+            auto &TemplateArgs = templateDecl->getTemplateArgs();
+            DXASSERT_NOMSG(TemplateArgs.size() >= 1);
+            QualType Arg0Type = TemplateArgs.get(0).getAsType();
+            if (const RecordType *NodeStructType =
+                    Arg0Type->getAsStructureType()) {
+              if (CXXRecordDecl *NodeStructDecl =
+                      dyn_cast<CXXRecordDecl>(NodeStructType->getDecl())) {
```
tex3d, line 15566: As I commented on another PR, getting the record type from a `NodeOutput` or `NodeInputRecord` should be a helper function in HLSLTypes.cpp.  That would greatly simplify this code.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15541,46 +15541,47 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
...
-                if (nullptr != NodeInputStructDecl) {
-                  // Make sure there is exactly one SV_DispatchGrid semantics
-                  // and it has correct type.
-                  DiagnoseDispatchGridSemantics(S, NodeInputStructDecl, Found);
-                }
+    // Diagnose dispatch grid semantics.
+    bool Found = false;
+    for (FunctionDecl::param_iterator I = FD->param_begin(),
+                                      E = FD->param_end();
+         I != E; ++I) {
```
tex3d, line 15548: Existing code, but wouldn't something like `for (ParamVarDecl *PD : FD->params())` work to simplify this a bit?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15541,46 +15541,47 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
...
+            auto &TemplateArgs = templateDecl->getTemplateArgs();
+            DXASSERT_NOMSG(TemplateArgs.size() >= 1);
+            QualType Arg0Type = TemplateArgs.get(0).getAsType();
+            if (const RecordType *NodeStructType =
+                    Arg0Type->getAsStructureType()) {
+              if (CXXRecordDecl *NodeStructDecl =
+                      dyn_cast<CXXRecordDecl>(NodeStructType->getDecl())) {
+                bool OutputFound = false;
+                // Make sure there is exactly one SV_DispatchGrid semantics
+                // and it has correct type.
```
tex3d, line 15569:
```suggestion
                // Diagnose any SV_DispatchGrid semantics used in record.
                // OutputFound provides the output needed by
                // DiagnoseDispatchGridSemantics, but isn't used otherwise.
                bool OutputFound = false;
```

This comment isn't quite right.  This code doesn't make sure there's exactly one, it makes sure there is at most one and that it's on a field of a supported type.  The later code makes sure one was found on the input when it's required.

I've suggested a change, and moved `OutputFound`, making it clear why it's needed, and why it isn't used elsewhere.

tex3d: Strangely, GitHub won't show this comment in file view.  Is it because it includes or ends in lines that aren't changed in the commit?

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: Fixed.

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: Fixed.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -949,6 +949,26 @@ QualType GetHLSLResourceTemplateParamType(QualType type) {
 QualType GetHLSLInputPatchElementType(QualType type) {
   return GetHLSLResourceTemplateParamType(type);
 }
+
+CXXRecordDecl *GetHLSLNodeRecordDecl(clang::QualType type) {
```
tex3d, line 953: It looks like both PRs I commented about adding this now have two independent implementations of the same function.  See: https://github.com/microsoft/DirectXShaderCompiler/pull/5992/files#diff-bb5c401f39f10a7cc1f6d80ee2eecf7aa94e6d7d6ce9508a2034793e34b74067R90

These changes should sync on the method used.

tex3d: With the changes I suggested there, I think we can go with the version in the other PR.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15541,46 +15541,31 @@ void DiagnoseNodeEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName,
...
+        // Node records are template types
+        if (CXXRecordDecl *NodeStructDecl =
+                hlsl::GetHLSLNodeRecordDecl(ParamType)) {
+          // Diagnose any SV_DispatchGrid semantics used in record.
+          // OutputFound provides the output needed by
+          // DiagnoseDispatchGridSemantics, but isn't used otherwise.
+          bool OutputFound = false;
+          DiagnoseDispatchGridSemantics(
+              S, NodeStructDecl,
+              hlsl::IsHLSLNodeOutputType(ParamType) ? OutputFound : Found);
```
tex3d, line 15560: I jus realized that this isn't in the right location for diagnosing output nodes, since we are under a broadcasting-only branch (the else branch of `if (NodeLaunchTy != DXIL::NodeLaunchType::Broadcasting)`).

So, we either need to move this param iteration out of this branch, or copy it for outputs after this branch.  If we move this loop out of the branch, for inputs, we only want to check if it's broadcasting, then we have to check `if (NodeMDG && !Found)` only if broadcasting as well.

So maybe an independent loop for checking outputs after this branch would be simpler to read.

tools/clang/test/SemaHLSL/hlsl/workgraph/dispatch_grid.hlsl
```diff
@@ -240,3 +240,58 @@ struct Record18 {
...
+// expected-error@+3{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'matrix<int16_t, 2, 2>'}}
+// expected-error@+2{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'S<int>'}}
+// expected-error@+1{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'S<float>'}}
+  T grid : SV_DispatchGrid;
+  float data;
+};
+
+#define OUTPUT_NODE_ENTRY(T,N) \
+[Shader("node")] \
+[NodeLaunch("broadcasting")] \
```
tex3d, line 268: Given the issue with not checking non-broadcasting outputs, we should probably test across launch modes as another test case.

tools/clang/test/SemaHLSL/hlsl/workgraph/dispatch_grid.hlsl
```diff
@@ -240,3 +240,58 @@ struct Record18 {
...
+// expected-error@+6{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'vector<uint16_t, 4>'}}
+// expected-error@+5{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'uint16_t [4]'}}
+// expected-error@+4{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'vector<int16_t, 4>'}}
+// expected-error@+3{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'matrix<int16_t, 2, 2>'}}
+// expected-error@+2{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'S<int>'}}
+// expected-error@+1{{SV_DispatchGrid can only be applied to a declaration of unsigned 32 or 16-bit integer scalar, vector, or array up to 3 elements, not 'S<float>'}}
```
tex3d, line 261: For these, shouldn't we have a note for where the record instantiation was?  This template usage highlights the problem of not knowing where this came from.

tex3d (approved):
LGTM.

PR #6000, opened 2023-11-09
[SM68] convert workgraph verify test as lit test

### LINKED ISSUE #5964: Convert Verifier tests that were added in SM 6.8 to lit FileCheck tests
**Description**
VerifierTests in the `main` branch have been mostly converted to lit FileCheck tests. However, we did add new tests in SM 6.8 branch (`staging-sm-6.8`) using the old VerifierTests style. We should convert these tests to lit FileCheck tests after they are merged to `main`.

**Environment**
- DXC version - staging-sm-6.8 branch as of Nov 2, 2023

python3kgae:
Fixes #5964

### DIFF
--- a/tools/clang/test/SemaHLSL/hlsl/workgraph/attribute_diags.hlsl
+++ b/tools/clang/test/SemaHLSL/hlsl/workgraph/attribute_diags.hlsl
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -fsyntax-only -verify %s
+// RUN: %dxc -Tlib_6_8 -verify %s
 
 // NodeDispatchGrid and NodeMaxDispatchGrid may only be used with Broadcasting launch nodes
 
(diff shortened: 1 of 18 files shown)


llvm-beanz (approved):

hekota (approved):
LGTM!

tools/clang/test/SemaHLSL/hlsl/workgraph/attribute_diags.hlsl
```diff
@@ -1,4 +1,4 @@
-// RUN: %clang_cc1 -fsyntax-only -verify %s
+// RUN: %dxc -Tlib_6_8 -verify %s
```
tex3d, line 1: Do we ignore attached validator versioning when using `-verify` option?  If not, these will cause failures when we test against released validators.

tex3d: Ok, it looks like validator version will only impact `CodeGenOpts`, which won't be used in this case, so I think we're fine.

tex3d (approved):

PR #6008, opened 2023-11-13
Add IsNode helper to ShaderModel class

sooknarine:


bob80905 (commented):
The reason this helper doesn't exist is because it's impossible to have a shader target a node stage specifically. Here is a relevant part of the spec:
image
This set of functions determine whether or not the shader was targeted to any specific stage, and node stages aren't among the set of valid stages to target.

llvm-beanz (changes requested):
In general we don't accept code without use cases and test coverage.

sooknarine:
> The reason this helper doesn't exist is because it's impossible to have a shader target a node stage specifically.

Thanks for the clarification.  Looks like `ShaderModel::m_Kind == ShaderKind::Library`.  Need to check `DxilFunctionProps::shaderKind` for the details.

PR #6009, opened 2023-11-13
Only diagnose exported library function declarations

bob80905:
Previously, diagnosing the translation unit would loop through every function declaration found in a library shader. This behavior is undesired, as there are some functions that are not intended to be exported, and thus shouldn't emit diagnostics (maybe the library compilation target is for a shader model that wouldn't mesh well with some other function declarations in the translation unit). This PR adjusts the diagnostic emission code so that only exported functions are diagnosed.
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5857


tools/clang/include/clang/Basic/LangOptions.h
```diff
@@ -39,8 +40,10 @@ class LangOptionsBase {
 
 #else
 
-#define LANGOPT(Name, Bits, Default, Description) static const unsigned Name = Default;
-#define LANGOPT_BOOL(Name, Default, Description) static const bool Name = static_cast<bool>( Default );
+#define LANGOPT(Name, Bits, Default, Description)                              \
+  static const unsigned Name = Default;
+#define LANGOPT_BOOL(Name, Default, Description)                               \
+  static const bool Name = static_cast<bool>(Default);
```
tex3d, line 46: How were all these unrelated formatting changes necessary for LangOptions.h?
Could this have been left-over from trying to accommodate clang-format checker errors after switching base branch on the previous PR?  Maybe it wouldn't have complained about these in this branch had you only added the original modification?

tools/clang/test/HLSLFileCheck/hlsl/functions/recursion/lib_struct_member_unconditional.hlsl
```diff
@@ -6,14 +6,14 @@
 // The SCCP pass replaces the recursive call with an undef value,
 // which is why validation fails with a non-obvious error.
 
-// CHECK: error: recursive functions are not allowed: entry function calls recursive function 'func'
+// CHECK: error: recursive functions are not allowed: export function calls recursive function 'func'
```
tex3d, line 9: I feel like this error could be improved.

Originally, it just checked the entry function, so saying "entry function" was sufficient.  But now, we should probably say the function name that calls the recursive function, and I'm not sure saying 'export' here is helpful (might be confusing since we use it for entry functions as well).  So this could be: `"... allowed: function '<name>' calls ..."`

Next, I wonder if function name (`func`) could use a fully-qualified name instead (`S::func` in this case).
Finally, it could be useful to place a "note:" diag on the recursive function (or even better, on the recursive call site, but that'll be more work).

tools/clang/test/HLSLFileCheck/hlsl/functions/recursion/lib_struct_member_unconditional.hlsl
tex3d, line 1: Can't these be converted to verifier tests?

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1564,13 +1564,17 @@ class DxcCompiler : public IDxcCompiler3,
...
 
     if (Opts.DefaultLinkage.empty()) {
       compiler.getCodeGenOpts().DefaultLinkage = DXIL::DefaultLinkage::Default;
+      compiler.getLangOpts().DefaultLinkage = DXIL::DefaultLinkage::Default;
     } else if (Opts.DefaultLinkage.equals_lower("internal")) {
       compiler.getCodeGenOpts().DefaultLinkage = DXIL::DefaultLinkage::Internal;
+      compiler.getLangOpts().DefaultLinkage = DXIL::DefaultLinkage::Internal;
     } else if (Opts.DefaultLinkage.equals_lower("external")) {
       compiler.getCodeGenOpts().DefaultLinkage = DXIL::DefaultLinkage::External;
+      compiler.getLangOpts().DefaultLinkage = DXIL::DefaultLinkage::External;
```
tex3d, line 1577: Alternatively, you could set one option from the other after this translation from string to enum. Like: `compiler.getLangOpts().DefaultLinkage = compiler.getCodeGenOpts().DefaultLinkage`.  Another option would be to change HLSLOptions to parse and store the enum value in the first place.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -4145,6 +4155,14 @@ class HLSLExternalSource : public ExternalSemaSource {
 
   Sema *getSema() { return m_sema; }
 
+  void BuildCallGraphForEntry(FunctionDecl *EntryFnDecl) {
+    m_callGraph.BuildForEntry(EntryFnDecl);
+  }
+
+  FunctionDecl *CheckForCallGraphRecursion(FunctionDecl *EntryFnDecl) {
```
tex3d, line 4162: These two methods just wrap methods on `m_callGraph`.  Why not just use `getCallGraph.<method>()` instead?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5794,6 +5812,8 @@ class HLSLExternalSource : public ExternalSemaSource {
     return method;
   }
 
+  CallGraphWithRecurseGuard getCallGraph() { return m_callGraph; }
```
tex3d, line 5815: This should *definitely* return a reference, not a copy.  There could be const version too, if that's necessary.

tools/clang/include/clang/Basic/LangOptions.h
bob80905: Yes, I think the most likely explanation or cause was the base branch switch.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11454,6 +11485,38 @@ void DiagnoseEntryAttrAllowedOnStage(clang::Sema *self,
...
+  std::vector<FunctionDecl *> AllExportedFDecls;
+
+  std::deque<DeclContext *> Worklist;
+  Worklist.push_back(self->getASTContext().getTranslationUnitDecl());
+  while (Worklist.size()) {
+    DeclContext *DC = Worklist.front();
+    Worklist.pop_front();
+    if (auto *FD = dyn_cast<FunctionDecl>(DC)) {
+      if (!FD->hasBody() ||
+          !IsExported(self, FD, getDefaultLinkageExternal(self))) {
```
tex3d, line 11492: Why is this check necessary?  Isn't it checked before adding a FunctionDecl to the worklist?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11534,34 +11607,35 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
-    if (!result && !patchResult) {
-      hlsl::CallGraphWithRecurseGuard CG;
-      CG.BuildForEntry(pPatchFnDecl);
-      if (CG.CheckReachability(pPatchFnDecl, pEntryPointDecl)) {
-        self->Diag(pEntryPointDecl->getSourceRange().getBegin(),
-                   diag::err_hlsl_patch_reachability_not_allowed)
-            << 1 << pEntryPointDecl->getName() << 0 << pPatchFnDecl->getName();
+      // In this case, recursion was detected in the patch-constant function
+      if (patchResult) {
+        self->Diag(patchResult->getSourceRange().getBegin(),
```
tex3d, line 11641: Earlier, we use `DiagnosedDecls` to prevent duplicate recursion diagnostics for the same function.  We should do the same here.  That will also solve the issue that multiple hull shaders could use the same patch constant function, leading to duplicate recursion diagnostics for that patch constant function.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: Originally I thought there would be an edge case where we enter the loop and the first thing in the worklist is a function decl. But, I don't think the translationunitdecl (which starts off the worklist on iteration 0) is ever a function decl, so we are guaranteed it will enter the else case. So, I can see now that the condition isn't necessary. Thanks!

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: Good point!

tex3d (approved):
LGTM!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11475,56 +11557,68 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
-  }
+  // Now check for recursion, and check for patch constant function
+  // reachabililty Validation methods differ depending on whether this is a
+  // library shader or not.
 
   // TODO: make these error 'real' errors rather than on-the-fly things
   // Validate that the entry point is available.
   DiagnosticsEngine &Diags = self->getDiagnostics();
   FunctionDecl *pEntryPointDecl = nullptr;
   FunctionDecl *pPatchFnDecl = nullptr;
```
python3kgae, line 11568: pPatchFnDecl should be moved into the for (FunctionDecl *FDecl : FDeclsToCheck) { loop.

tools/clang/test/CodeGenHLSL/recursive2.hlsl
```diff
@@ -1,6 +1,6 @@
 // RUN: %dxc -E main -T ps_6_0 %s | FileCheck %s
 
-// CHECK: error: recursive functions are not allowed: entry function calls recursive function 'test_inout'
+// CHECK: error: recursive functions are not allowed: function 'main' calls recursive function 'test_inout'
```
python3kgae, line 3: Would be nice to change to verify test and move to SemaHLSL/recursion.

PR #6011, opened 2023-11-14
[SPIR-V] Fix push_constant and shader_record_ext to work with LowerTypeVisitor

### LINKED ISSUE #5808: [SPIR-V] Validation error when push constant is passed into a function
**Description**
Getting a compiler error (thrown by SPIR-V validation) when a push constant is passed straight into a function. This error is thrown on trunk but not on DXC 1.7. ~~Please note that the SPIR-V generated between trunk and 1.7 is identical so either the problem was always there and the updated spirv-val caught it or newer spir-val is broken~~ (scratch that, seems I was wrong).

**Steps to Reproduce**
Simple repro case
```
struct Foo
{
    float m_x;
};
 
[[vk::push_constant]] ConstantBuffer<Foo> g_pc;
RWStructuredBuffer<float> g_buff;

float mul1(Foo m, float4 v)
{
	return m.m_x + v.x;
}
 
[numthreads(1, 1, 1)] void main()
{
    g_buff[0] = mul1(g_pc, float4(1, 0, 0, 1));
}
```

And the error is:
```
fatal error: generated SPIR-V is invalid: OpLoad Result Type <id> '7[%type_ConstantBuffer_Foo]' does not match Pointer <id> '4[%g_pc]'s type.
  %21 = OpLoad %type_ConstantBuffer_Foo %g_pc

```

**Actual Behavior**
There shouldn't be an error.

**Environment**
- DXC version: Trunk (6b4b0eb5)
- Host Operating System: Linux

cassiebeckley:
Since `ConstantBuffer` is now lowered in `LowerTypeVisitor`, lower `push_constant` and `shader_record_*` with `ConstantBuffer` types there as well.

Fixes #5808.

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1297,7 +1297,7 @@ SpirvVariable *DeclResultIdMapper::createStructOrStructArrayVarOfExplicitLayout(
   const bool forShaderRecordNV =
       usageKind == ContextUsageKind::ShaderRecordBufferNV;
   const bool forShaderRecordEXT =
-      usageKind == ContextUsageKind::ShaderRecordBufferEXT;
+      usageKind == ContextUsageKind::ShaderRecordBufferKHR;
 
   const auto &declGroup = collectDeclsInDeclContext(decl);
 
@@ -1438,7 +1438,7 @@ SpirvVariable *DeclResultIdMapper::createPushConstant(const VarDecl *decl) {
   const QualType type = decl->getType();
   const auto *recordType = type->getAs<RecordType>();
 
-  SpirvVariable *var;
+  SpirvVariable *var = nullptr;
 
   if (isConstantBuffer(type)) {
     // Constant buffers already have Block decoration. The variable will need
@@ -1480,10 +1480,10 @@ DeclResultIdMapper::createShaderRecordBuffer(const VarDecl *decl,
       hlsl::GetHLSLResourceResultType(type)->getAs<RecordType>();
   assert(recordType);
 
-  assert(kind == ContextUsageKind::ShaderRecordBufferEXT ||
+  assert(kind == ContextUsageKind::ShaderRecordBufferKHR ||
          kind == ContextUsageKind::ShaderRecordBufferNV);
 
-  SpirvVariable *var;
+  SpirvVariable *var = nullptr;
   if (isConstantBuffer(type)) {
     // Constant buffers already have Block decoration. The variable will need
     // the appropriate storage class.
@@ -1503,8 +1503,8 @@ DeclResultIdMapper::createShaderRecordBuffer(const VarDecl *decl,
     var->setHlslUserType("");
     var->setLayoutRule(layoutRule);
   } else {
-    const auto typeName = kind == ContextUsageKind::ShaderRecordBufferEXT
-                              ? "type.ShaderRecordBufferEXT."
+    const auto typeName = kind == ContextUsageKind::ShaderRecordBufferKHR
+                              ? "type.ShaderRecordBufferKHR."
                               : "type.ShaderRecordBufferNV.";
 
     const std::string structName =
@@ -1526,11 +1526,11 @@ DeclResultIdMapper::createShaderRecordBuffer(const VarDecl *decl,
 SpirvVariable *
 DeclResultIdMapper::createShaderRecordBuffer(const HLSLBufferDecl *decl,
                                              ContextUsageKind kind) {
-  assert(kind == ContextUsageKind::ShaderRecordBufferEXT ||
+  assert(kind == ContextUsageKind::ShaderRecordBufferKHR ||
          kind == ContextUsageKind::ShaderRecordBufferNV);
 
-  const auto typeName = kind == ContextUsageKind::ShaderRecordBufferEXT
-                            ? "type.ShaderRecordBufferEXT."
+  const auto typeName = kind == ContextUsageKind::ShaderRecordBufferKHR
+                            ? "type.ShaderRecordBufferKHR."
                             : "type.ShaderRecordBufferNV.";
 
   const std::string structName = typeName + decl->getName().str();
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1347,9 +1347,6 @@ SpirvVariable *DeclResultIdMapper::createStructOrStructArrayVarOfExplicitLayout(
     }
   }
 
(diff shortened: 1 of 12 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1441,18 +1438,30 @@ SpirvVariable *DeclResultIdMapper::createPushConstant(const VarDecl *decl) {
   const QualType type = decl->getType();
   const auto *recordType = type->getAs<RecordType>();
 
+  SpirvVariable *var;
```
Keenuts, line 1441:
```suggestion
  SpirvVariable *var = nullptr;
```

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1466,22 +1475,44 @@ SpirvVariable *DeclResultIdMapper::createPushConstant(const VarDecl *decl) {
...
+      hlsl::GetHLSLResourceResultType(type)->getAs<RecordType>();
   assert(recordType);
 
   assert(kind == ContextUsageKind::ShaderRecordBufferEXT ||
          kind == ContextUsageKind::ShaderRecordBufferNV);
 
-  const auto typeName = kind == ContextUsageKind::ShaderRecordBufferEXT
-                            ? "type.ShaderRecordBufferEXT."
-                            : "type.ShaderRecordBufferNV.";
+  SpirvVariable *var;
```
Keenuts, line 1486:
```suggestion
  SpirvVariable *var = nullptr;
```

tools/clang/test/CodeGenSPIRV_Lit/vk.layout.shader-record-ext.std430.hlsl
```diff
@@ -44,7 +44,7 @@ struct S {
 [[vk::shader_record_ext]]
 ConstantBuffer<S> cbuf;
 
-// CHECK: OpDecorate %type_ShaderRecordBufferEXT_S Block
```
Keenuts, line 47: Maybe I'm paranoid, but shouldn't we check those decoration do apply to the type used for the cbuf variable declaration?
Since we are in fcgl, a bug could make us generate this type correctly, but not use it for the cbuf variable.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1466,22 +1475,44 @@ SpirvVariable *DeclResultIdMapper::createPushConstant(const VarDecl *decl) {
...
-  const std::string structName =
-      typeName + recordType->getDecl()->getName().str();
-  SpirvVariable *var = createStructOrStructArrayVarOfExplicitLayout(
-      recordType->getDecl(), /*arraySize*/ 0, kind, structName,
-      decl->getName());
+    var->setHlslUserType("");
+    var->setLayoutRule(layoutRule);
+  } else {
+    const auto typeName = kind == ContextUsageKind::ShaderRecordBufferEXT
+                              ? "type.ShaderRecordBufferEXT."
```
Keenuts, line 1507: nit: name is EXT, but storage class is now KHR

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1466,22 +1475,44 @@ SpirvVariable *DeclResultIdMapper::createPushConstant(const VarDecl *decl) {
...
-                            ? "type.ShaderRecordBufferEXT."
-                            : "type.ShaderRecordBufferNV.";
+  SpirvVariable *var;
+  if (isConstantBuffer(type)) {
+    // Constant buffers already have Block decoration. The variable will need
+    // the appropriate storage class.
+
+    const auto sc = kind == ContextUsageKind::ShaderRecordBufferNV
+                        ? spv::StorageClass::ShaderRecordBufferNV
+                        : spv::StorageClass::ShaderRecordBufferKHR;
```
Keenuts, line 1493: I don't see any test checking we do add the required capabilities/extensions. Shall this be added?
I do believe changing EXT into KHR is a good change, but maybe we should now check it.

s-perron (approved):
Nothing beyond what Nathan already suggested.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
cassiebeckley: It looks like there are tests that the ray tracing capability and extension are added for the `RaytracingAccelerationStructure` type, but not for the `vk::shader_record_ext` attribute. I'll add them into the new tests.

tools/clang/test/CodeGenSPIRV_Lit/vk.layout.shader-record-ext.std430.hlsl
cassiebeckley: Sure, added.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
cassiebeckley: I've changed the typename and the enum identifier.

cassiebeckley:
@Keenuts did you want any other changes?

tools/clang/test/CodeGenSPIRV_Lit/vk.layout.shader-record-ext.std430.hlsl
```diff
@@ -44,7 +44,7 @@ struct S {
 [[vk::shader_record_ext]]
 ConstantBuffer<S> cbuf;
```
Keenuts, line 45: I'm a bit confused, this test is supposed to only use the KHR variant of the ShaderRecordBuffer extensions?
Building this test with the DXC at this PR ends up generating this kind of lines:
```
 %_ptr_ShaderRecordBufferNV_type_ShaderRecordBufferKHR_block = OpTypePointer ShaderRecordBufferNV %type_ShaderRecordBufferKHR_block

...

%block = OpVariable %_ptr_ShaderRecordBufferNV_type_ShaderRecordBufferKHR_block ShaderRecordBufferNV
```

Aren't the NV version not supposed to be used?

Keenuts:
> @Keenuts did you want any other changes?

Sorry missed this PR. Just a question about observed generated code, otherwise nothing specific.

tools/clang/test/CodeGenSPIRV_Lit/vk.layout.shader-record-ext.std430.hlsl
cassiebeckley: The NV and KHR constants have the same value; the SPIRV-Tools disassembler is defaulting to using the NV string for that constant.

Keenuts (approved):

PR #6012, opened 2023-11-14
SPIRV: Fix segfault when compiling hlsl intrinsics with variadic argu…

ChristianReinbold:
For hlsl intrinsics with variadic argument lists (in my case: printf), the number of param modifiers and function arguments did not match, resulting in segfaults when creating a FunctionProtoType. The presumed oversight: The paramsMods was made twice as large as intended, by first resizing to the expected number of arguments, and then pushing back the same number of arguments again. A fix that seems to resolve this issue comes with this PR. Please have a look.

Edit: The suggested modifications do not feel as if they are the proper way to handle modParams initialization. They only defuse the issue I had with printf. If you think this issue needs a better solution and/or a proper bug report before suggesting a PR, please let me know - and please apologize in case I opened up a PR too early.

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -1649,26 +1649,24 @@ ParamModsFromIntrinsicArg(const HLSL_INTRINSIC_ARGUMENT *pArg) {
   return hlsl::ParameterModifier(hlsl::ParameterModifier::Kind::In);
 }
 
-static void InitParamMods(const HLSL_INTRINSIC *pIntrinsic,
-                          SmallVectorImpl<hlsl::ParameterModifier> &paramMods,
-                          size_t uNumArgsWithVariadic = 0u) {
+static SmallVector<hlsl::ParameterModifier, g_MaxIntrinsicParamCount>
+createParamMods(const HLSL_INTRINSIC *pIntrinsic,
+                size_t uVariadicArgumentCount = 0u) {
+  SmallVector<hlsl::ParameterModifier, g_MaxIntrinsicParamCount> paramMods;
   // The first argument is the return value, which isn't included.
-  UINT i = 1;
-  for (; i < pIntrinsic->uNumArgs; ++i) {
+  for (UINT i = 1; i < pIntrinsic->uNumArgs; ++i) {
     // Once we reach varargs we can break out of this loop.
     if (IsVariadicArgument(pIntrinsic->pArgs[i]))
       break;
     paramMods.push_back(ParamModsFromIntrinsicArg(&pIntrinsic->pArgs[i]));
   }
-
   // For variadic functions, any argument not explicitly specified will be
   // considered an input argument.
-  if (IsVariadicIntrinsicFunction(pIntrinsic)) {
-    for (; i < uNumArgsWithVariadic; ++i) {
-      paramMods.push_back(
-          hlsl::ParameterModifier(hlsl::ParameterModifier::Kind::In));
-    }
+  for (UINT i = 0; i < uVariadicArgumentCount; ++i) {
+    paramMods.push_back(
+        hlsl::ParameterModifier(hlsl::ParameterModifier::Kind::In));
   }
+  return paramMods;
 }
 
 static bool IsBuiltinTable(LPCSTR tableName) {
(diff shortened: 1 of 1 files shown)


ChristianReinbold:

Keenuts:
Hi!
Thanks for the contribution!

To best help us review, and also make sure we don't regress, could you add at least a test? (The case that was triggering the segfault would be great) 😊.

Thanks!

ChristianReinbold:
A reproducer could look like

```
[numthreads(64,1,1)]
void main() {
  float f = 0;
  printf("%f%f%f%f%f%f%f%f%f%f", f, f, f, f, f, f, f, f, f, f);
}
```

The problem with that approach though is that it will not fail due to the custom memory allocation strategy of clang. It allocates memory slabs of size >= 4096B, and the segfault will only happen if the `FunctionProtoType` created when parsing `printf` is allocated close to the end of a slab, such that after allocation not more than 40B are left unallocated in the slab. Then we actually try to access memory outside of the slab for this example, resulting in a segfault.

I could now try bloat the example such that the printf is processed when being close to the end of a slab, but this is somehow pointless, since arbitrary code changes influencing allocation patterns will "resolve" the segfault. By reducing the default `SlabSize` of `BumpPtrAllocatorImpl` from 4096 to 32, I was able to reliably reproduce the segfault with the hlsl code above. However, clearly I do not want to check in this modification.

I think the cleanest solution would be to disable slab allocation for tests completely. Maybe this will uncover more segfaults that currently are hidden by the custom allocation strategy, and only occur sporadically in production.

ChristianReinbold:
Actually, this test already should do the job, assuming that allocation behavior is modified to be less "forgiving".

https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/CodeGenSPIRV_Lit/intrinsics.printf.hlsl

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -1650,9 +1650,10 @@ ParamModsFromIntrinsicArg(const HLSL_INTRINSIC_ARGUMENT *pArg) {
 }
 
 static void InitParamMods(const HLSL_INTRINSIC *pIntrinsic,
-                          SmallVectorImpl<hlsl::ParameterModifier> &paramMods) {
+                          SmallVectorImpl<hlsl::ParameterModifier> &paramMods,
+                          size_t uNumArgsWithVariadic = 0u) {
```
Keenuts, line 1654: nit: what about `uVariadicArgumentCount = 0`? 
As-is, it feels like we should set the number of arguments everytime, but default to 0, while it's only useful for variadic.

This of course requires to add a new size inside the function:
```
const size_t argumentCount = pIntrinsic->uNumArgs + uVariadicArgumentCount;
```

As for the loop, fine leaving it as you did, or if you have some time, maybe squash them into a single loop from 1 to argumentCount? But that's my personal preference to have a simple linear counter, not jumping around across multiple loops.

Keenuts:
Thanks for looking into it!
I agree, seems like testing this is not that easy, sorry for wasting your time on that front!

I took a look at the good, and looks good, just a few suggestions, but it's more my personal prefrerences than real issues 😊

tools/clang/lib/Sema/SemaHLSL.cpp
ChristianReinbold: I gave it a shot. The "numArgs" parameter now is closer to what it actually represents. We still have two loops though, but at least not with a shared index.

Keenuts (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
Keenuts: Thanks!

s-perron:
This is not SPIRV specific code. I think someone from MS should look into it. @llvm-beanz can you bring it to the attention of the correct person?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -3031,8 +3025,7 @@ class HLSLExternalSource : public ExternalSemaSource {
         &templateParamNamedDeclsCount);
     // }
 
-    SmallVector<hlsl::ParameterModifier, g_MaxIntrinsicParamCount> paramMods;
-    InitParamMods(intrinsic, paramMods);
+    auto paramMods = createParamMods(intrinsic);
```
llvm-beanz, line 3028: Revert back to passing the SmallVector as an argument. SmallVectors are kinda cool, but can be _really_ big. In this case the SmallVector's inline storage is big enough to store `g_MaxIntrinsicParamCount` elements. That means the copy on return is large and not just a `std::move` like you might have with a `std::vector`.

ChristianReinbold: Reverted. By default I assume return value optimization is in place, which is why I made this change as a slight readability improvement. Otherwise, this line of discussion would require to never return anything larger than `uintpr_t`. Admittedly, I have not checked if the extra copy actually was elided.

bob80905 (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
Keenuts: Good to know.
note: just looked at the generated assembly in release, and the whole smallvector and createParamMods implem get's inlined, and looking at the debug build, it seems that we do get the copy elision

llvm-beanz (approved):

PR #6014, opened 2023-11-15
WinPix: Inline functions test update

jeffnn:
I had a couple of test updates that should have gone in my previous PR but got left out due to me forgetting, due to a two-week mental context switch :-). The new test adds some useful regression testing, so here it is.

### DIFF
--- a/tools/clang/unittests/HLSL/PixDiaTest.cpp
+++ b/tools/clang/unittests/HLSL/PixDiaTest.cpp
@@ -177,7 +177,7 @@ public:
   TEST_METHOD(DxcPixDxilDebugInfo_UnnamedField)
   TEST_METHOD(DxcPixDxilDebugInfo_SubProgramsInNamespaces)
   TEST_METHOD(DxcPixDxilDebugInfo_SubPrograms)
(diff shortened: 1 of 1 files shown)


adam-yang (approved):

tools/clang/unittests/HLSL/PixDiaTest.cpp
```diff
@@ -2920,4 +2923,57 @@ float4 DeeperInlinedFunction(in BuiltInTriangleIntersectionAttributes attr, int
                        ColorRegisterNumberWhenCalledFromOuterForDeeper);
 }
 
-#endif
\ No newline at end of file
+TEST_F(PixDiaTest,
+       DxcPixDxilDebugInfo_InlinedFunctions_CalledTwiceInSameCaller) {
+  if (m_ver.SkipDxilVersion(1, 2))
```
tex3d, line 2928: You compile to `lib_6_6` in this code, so I think the minimum required DXIL & validator should be `(1, 6)`, not `(1, 2)`.

jeffnn: Ah, copy-pasta. Thanks.

tex3d (approved):

PR #6015, opened 2023-11-15
[SPIRV] allow location index to be > than 64

### LINKED ISSUE #3735: [SPIR-V] Using very large vert/hull/domain IO produce a silent compiler crash (works with DXIL backend)
### Title
[SPIR-V] Using very large vert/hull/domain IO produce a silent compiler crash (works with DXIL backend)

### Functional impact
Accidentally having too much data in stage IO will crash the compiler without a clear indication of what happened.

### Minimal repro steps

1. Compile the following shader with either of:
 `dxc.exe -T vs_6_0 -E Vert-spirv repro_shader.txt`
 `dxc.exe -T hs_6_0 -E Hull -spirv repro_shader.txt`
 `dxc.exe -T ds_6_0 -E Domain -spirv repro_shader.txt` 
[repro_shader.txt](https://github.com/microsoft/DirectXShaderCompiler/files/6414941/repro_shader.txt)

### Expected result
The SPIR-V backend should produce user facing errors like the DXIL backend does:
```
repro_shader.txt:21:1: error: Failed to allocate all input signature elements in available space.
ControlPoint Hull (InputPatch<ControlPoint,13> v, uint id : SV_OutputControlPointID) { return v[id]; }
^
repro_shader.txt:21:1: error: Failed to allocate all output signature elements in available space.
```

### Actual result
The SPIR-V backend crashes silently

### Further technical details

Keenuts:
There was an arbitrary limit to set the maximum number of locations to 64.
This commit allows the set to be resized, to allow arbitrarily large location indices.

Curiously, one test, vk.location.large.hlsl started to fail. This test assumed 123456 was an invalid location. From the SPIR-V spec, I don't think this is true, as it only days Location should be an arbitraty number.
In addition, it seems like this test was triggering an out-of-bound access, which is now solved. Only drawback: we now end-up with a 123456 sized bitset.
I believe this is OK, as WHY would someone only require this index.

Fixes #3735

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1828,7 +1828,8 @@ public:
     if (loc == nextAvailableLocation[index]) {
       ++nextAvailableLocation[index];
     } else {
-      nextAvailableLocation[index] = std::max(loc + 1, nextAvailableLocation[index]);
+      nextAvailableLocation[index] =
+          std::max(loc + 1, nextAvailableLocation[index]);
     }
   }
 
@@ -1841,9 +1842,11 @@ public:
 
     // Simple case: no hole large enough left, resizing.
     if (res == std::nullopt) {
-      const uint32_t spaceLeft = locations.size() - nextAvailableLocation[index];
+      const uint32_t spaceLeft =
+          locations.size() - nextAvailableLocation[index];
       assert(spaceLeft < count && "There is a bug.");
-      const uint32_t requiredAlloc = count - spaceLeft;;
+      const uint32_t requiredAlloc = count - spaceLeft;
+      ;
       locations.resize(locations.size() + requiredAlloc);
       res = nextAvailableLocation[index];
     }
@@ -1851,7 +1854,8 @@ public:
     for (uint32_t i = res.value(); i < res.value() + count; i++) {
       locations.set(i);
     }
-    nextAvailableLocation[index] = std::max(res.value() + count, nextAvailableLocation[index]);
+    nextAvailableLocation[index] =
+        std::max(res.value() + count, nextAvailableLocation[index]);
     return res.value();
   }
 
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -9,6 +9,8 @@
 
 #include "DeclResultIdMapper.h"
 
+#include <algorithm>
+#include <optional>
 #include <sstream>
 
 #include "dxc/DXIL/DxilConstants.h"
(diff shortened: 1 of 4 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1802,51 +1804,85 @@ class LocationSet {
...
 
-  /// Uses the next |count| available location.
-  int useNextLocs(uint32_t count, uint32_t index = 0) {
-    assert(index < kMaxIndex);
-    auto &locs = usedLocs[index];
-    auto &next = nextLoc[index];
-    while (locs[next])
-      next++;
+    auto &set = usedLocations[index];
+    if (loc <= set.size()) {
```
s-perron, line 1821:
```suggestion
    if (loc >= set.size()) {
```

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1802,51 +1804,85 @@ class LocationSet {
...
+    uint32_t slot_size = 0;
+    for (uint32_t i = nextAvailableLocation[index]; i < locations.size(); i++) {
+      if (locations[i])
+        slot_size = 0;
+      else
+        ++slot_size;
+
+      if (slot_size == size)
+        return i + 1 - slot_size;
+    }
```
s-perron, line 1877: What happens if the size of `locations` is 1024 and is all empty and you want a range of size 1025?

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1802,51 +1804,85 @@ class LocationSet {
...
+    if (loc <= set.size()) {
+      set.resize(std::max<size_t>(loc + 1, set.size() * 2));
+    }
+    set.set(loc);
 
-    int toUse = next;
+    // Maybe the bit marked is the next available one.
+    // Otherwise, there is a hole, ignoring.
+    if (loc == nextAvailableLocation[index]) {
+      ++nextAvailableLocation[index];
```
s-perron, line 1829: You don't keep `nextAvailableLocation` up-to-date properly. In this case, the value could already be used. When you call `useNextLocs`, I do not see it getting updated at all.

s-perron (changes requested):
I want to double check that there are no limits on the value of the location. There are definitely limits on the number of locations that can be used. See the [VkPhysicalDeviceLimits](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceLimits.html). I don't know what those values are on real hardware.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
Keenuts: Ah yes, thanks!

Keenuts:
> I want to double check that there are no limits on the value of the location. There are definitely limits on the number of locations that can be used. See the [VkPhysicalDeviceLimits](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceLimits.html). I don't know what those values are on real hardware.

Yes, I was also wondering about that.
With optimizations, this end up creating a struct with 100 fields, but we have a 16k limit on the number of members for the struct, so 16k for inputs I believe.
But outside of this, I haven't seen a limit on the SPIR-V spec side.

As for the Vulkan side, since this is driver specific, I suppose we cannot handle that at compile time, but it's more the responsibility from the writer to not use thousands of inputs in their shaders?

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
Keenuts: Ah yes, thanks. I'd like to unit test this TBH...

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
Keenuts: We bump the vector size to 1024+1025. Added a quick way to handle this specific case which doesn't add that much complexity. Would like to keep the logic simple as I believe those are corner cases. We were fine a fixed, unchecked 64 bits vector for quite some time, so don't believe we might hit those cases often.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1802,51 +1804,94 @@ class LocationSet {
...
-  llvm::SmallBitVector usedLocs[kMaxIndex]; ///< All previously used locations
-  uint32_t nextLoc[kMaxIndex];              ///< Next available location
+  // Find the first unused range of size |size| in the given set.
+  // If the set contains such range, returns the first usable index.
+  // Otherwise, nullopt is returned.
+  std::optional<uint32_t> find_unused_range(uint32_t index, uint32_t size) {
+    assert(index < kMaxIndex);
+    const auto &locations = usedLocations[index];
+    uint32_t slot_size = 0;
+    for (uint32_t i = nextAvailableLocation[index]; i < locations.size(); i++) {
```
s-perron, line 1878: If `nextAvailableLocation` points to one past the last bit set, then this function could be simplified, or removed altogether, right?

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1802,51 +1804,94 @@ class LocationSet {
...
+      if (slot_size == size)
+        return i + 1 - slot_size;
+    }
+
+    return std::nullopt;
+  }
+
+  /// All previously used locations
+  llvm::SmallBitVector usedLocations[kMaxIndex];
+  /// Next available location
```
s-perron, line 1893: What does "available" mean? Is it the first unset bit in the bitvector? It is one past the last bit that is set?

s-perron: It might be useful to make the comment more precise.

Keenuts: Rewrote this class and expanded comments to make this more clear

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
Keenuts: Yes, this was a bit messy. Now the function is way simpler, and we handle holes correctly. Not the most efficient (since it's a bitvector, could be faster), but at least it's easy to review, and I don't this location allocation is a critical path at all.

s-perron (approved):
I think this looks good. I just realized that we should try to keep the location allocation the same as before. We do not want shaders to start failing because the locations changed. I believe this will be true for cases that would not have asserted. Could you double check that I am right about that?

cassiebeckley (approved):

PR #6018, opened 2023-11-15
[SPIR-V] Fix crash with GetAttributeAtVertex

### LINKED ISSUE #2955: [SPIR-V] Silent compiler crash when using GetAttributeAtVertex
### Title
 Silent compiler crash when using GetAttributeAtVertex

### Functional impact
Compiler crashes, no result gets printed to stdout.

### Minimal repro steps

1. Compile [repro_shader.txt](https://github.com/microsoft/DirectXShaderCompiler/files/4746158/repro_shader.txt) with `dxc.exe -E frag -T ps_6_1 repro_shader.txt -spirv`
```
struct PSInput
{
	nointerpolation float3 color : COLOR;
};

cbuffer constants : register(b0)
{
	float4 g_constants;
}

float4 frag(PSInput input) : SV_TARGET
{
	uint cmp = (uint)(g_constants[0]);

	float colorAtV0 = GetAttributeAtVertex(input.color, 0)[cmp];
	float colorAtV1 = GetAttributeAtVertex(input.color, 1)[cmp];
	float colorAtV2 = GetAttributeAtVertex(input.color, 2)[cmp];

	return float4(colorAtV0, colorAtV1, colorAtV2, 0);
}
```

### Expected result
The compiler outputs valid SPIR-V

### Actual result
DXC crashes on null pointer dereference (see [call_stack.txt](https://github.com/microsoft/DirectXShaderCompiler/files/4746508/call_stack.txt))

### Further technical details
DXC built from changeset 5ebc67c618d7061407cca039573108078f601ee6

Keenuts:
The code assumed all the types we handled had an AST type. But when the struct is created in the SPIR-V backend, there are no AST type left to represent it.
In such case, the lowering step should handle the interpolation attributes, so I think we shouldn't have to do anything.

Fixes #2955 again.

### DIFF
--- a/tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
+++ b/tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
@@ -56,6 +56,11 @@ int PervertexInputVisitor::appendIndexZeroAt(
 ///< treated as nointerpolated too.
 bool PervertexInputVisitor::expandNointerpVarAndParam(
     SpirvInstruction *spvInst) {
+  // If the spirv type has no AST type (hybrid struct for ex), no need to expand
+  // it.
+  if (!spvInst->hasAstResultType())
+    return true;
+
   QualType type = spvInst->getAstResultType();
   bool isExpanded = false;
   auto typePtr = type.getTypePtr();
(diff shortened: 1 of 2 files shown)


s-perron (changes requested):
It is hard for me to know if it is correct. I want to see that the expansion of `GetAttributeAtVertex` is still correct. I also do not understand why the type does not have to be expanded. Please update the comment with more details. Where does it get expanded?

ShchchowAMD:
I'll help to take a look.

cassiebeckley (approved):
LGTM once Steven's comments are addressed

Keenuts:
Was wrong, the repro case I had is fixed, but not the issue case.
I need to look into this PR again to understand if the fix I suggested is really a fix, or just works because "luck".

Keenuts:
Replaced by https://github.com/microsoft/DirectXShaderCompiler/pull/6459

PR #6019, opened 2023-11-15
[CI] fix issue when apply format

python3kgae:
Fix issues hit when apply format.
1. tmp diff file is not opened binary form, don't need to encode.
2. add contents write to "Check code formatting" workflow for commit and push change.
3. put main branch to different path to avoid issue when checkout head branch to same path.

### DIFF
--- a/.github/workflows/clang-format-checker.yml
+++ b/.github/workflows/clang-format-checker.yml
@@ -6,6 +6,7 @@ on:
     types: edited
 permissions:
   pull-requests: write
+  contents: write
 
 jobs:
   code_formatter:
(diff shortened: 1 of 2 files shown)


hekota (approved):

bob80905 (approved):

.github/workflows/clang-format-checker.yml
```diff
@@ -84,23 +85,24 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 2
+          path: main
```
tex3d, line 88: Should this really be `main` under the source root of the tree?  Won't that cause problems too?  Could this be changed to be under a `.gitignore` directory, such as `build`?  Like `build/main_src`?  Also, how does this action select the branch to check out?  Is it just defaulting to `main`?  I admit I'm not that familiar with the command structure here, but there appears to be some magic/defaults that don't appear explicitly defined in this file.  Like where is the PR branch placed?  Is it placing us at the source root of that checkout by default?

python3kgae: build/main_src should be fine.
I'll update it.
The default for actions/checkout@v4 should be current repo with default branch.
And you can change it by set ref and repository.

PR #6021, opened 2023-11-15
[Diag] Diagnose missing numthreads in compute shaders

bogner:
Currently we error out in validation if a compute shader doesn't have
a numthreads attribute:

```
<source>:2: error: Declared Thread Group X size 2833376888 outside valid range [1..1024].
<source>:2: error: Declared Thread Group Y size 21956 outside valid range [1..1024].
<source>:2: error: Declared Thread Group Z size 2833381336 outside valid range [1..64].
<source>:2: error: Declared Thread Group Count 3863291136 (X*Y*Z) is beyond the valid maximum of 1024.
```

Instead, diagnose this in Sema like we do for mesh and amplitude shaders

### DIFF
--- a/tools/clang/test/CodeGenSPIRV_Lit/attribute.numthreads.missing.hlsl
+++ b/tools/clang/test/CodeGenSPIRV_Lit/attribute.numthreads.missing.hlsl
@@ -1,4 +1,4 @@
 // RUN: not %dxc -T cs_6_0 -E main -fcgl  %s -spirv  2>&1 | FileCheck %s
 
-// CHECK: 4:6: error: thread group size [numthreads(x,y,z)] is missing from the entry-point function
+// CHECK: 4:6: error: compute entry point must have the numthreads attribute
 void main() {}
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -12584,19 +12584,12 @@ void SpirvEmitter::processPixelShaderAttributes(const FunctionDecl *decl) {
 }
 
 void SpirvEmitter::processComputeShaderAttributes(const FunctionDecl *decl) {
-  // If not explicitly specified, x, y, and z should be defaulted to 1.
-  uint32_t x = 1, y = 1, z = 1;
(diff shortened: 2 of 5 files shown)


llvm-beanz (approved):

tools/clang/test/CodeGenSPIRV_Lit/attribute.numthreads.missing.hlsl
```diff
@@ -1,4 +1,4 @@
 // RUN: not %dxc -T cs_6_0 -E main -fcgl  %s -spirv  2>&1 | FileCheck %s
 
-// CHECK: 4:6: error: thread group size [numthreads(x,y,z)] is missing from the entry-point function
```
llvm-beanz, line 3: Should we be deleting the code that generated this error message?

bob80905 (approved):

damyanp (approved):

tools/clang/test/CodeGenSPIRV_Lit/attribute.numthreads.missing.hlsl
bogner: Unclear. It's in the SPIR-V emitter, so it might technically still be reachable given invalid dxil. Do you have any insight here?

llvm-beanz: My read of the code in the SpirvEmitter is that it shouldn't be reachable.

Any thoughts from @cassiebeckley, @sudonatalie, @Keenuts, or @s-perron?

bogner: Looking at it I'm convinced you're right - I've replaced it with an assert

pow2clk (approved):
It looks good, but I'm making an argument over on #6023 that maybe you should reword the error message that this change uses. Do with that information what you will!

llvm-beanz (approved):

tools/clang/test/CodeGenSPIRV_Lit/attribute.numthreads.missing.hlsl
s-perron: Yes, this looks fine to me.

PR #6022, opened 2023-11-15
[Diag] Treat entry points as active even if they already have a stage

bogner:
If we explicitly annotate an entry point with a shader stage we should
treat it as active.


llvm-beanz (approved):

bob80905:
I already move the isActiveEntry boolean higher in this PR:
https://github.com/microsoft/DirectXShaderCompiler/pull/5992/files
Would it be possible to merge that one in instead to prevent merge conflicts?

bogner:
> I already move the isActiveEntry boolean higher in this PR: https://github.com/microsoft/DirectXShaderCompiler/pull/5992/files Would it be possible to merge that one in instead to prevent merge conflicts?

I guess it depends on how close that one is to being ready to merge, and if we did do that I think we'd want to add this test anyways. Personally I'd prefer to merge this targeted fix on it's own

bob80905 (approved):

PR #6023, opened 2023-11-15
[Diag] Warn instead of error on invalid numthreads

bogner:
If an invalid numthreads attribute isn't used, erroring out on it is a
breaking change that might be annoying to deal with. Instead we make
invalid numthreads a warning and drop the attribute - we can rely on
the existing diagnostics for missing numthreads to error out in cases
that aren't recoverable.


damyanp (approved):

bob80905 (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -13319,15 +13319,20 @@ void hlsl::HandleDeclAttributeForHLSL(Sema &S, Decl *D, const AttributeList &A,
...
+    int X = ValidateAttributeIntArg(S, A, 0);
+    int Y = ValidateAttributeIntArg(S, A, 1);
+    int Z = ValidateAttributeIntArg(S, A, 2);
+    int N = X * Y * Z;
+    if (N > 0 && N <= 1024) {
+      auto numThreads = ::new (S.Context) HLSLNumThreadsAttr(
+          A.getRange(), S.Context, X, Y, Z, A.getAttributeSpellingListIndex());
+      declAttr = numThreads;
+    } else {
+      // If the number of threads is invalid, diagnose and drop the attribute.
```
pow2clk, line 13331: I think it's your intent by this that since numthreads are only valid where they are required and active, this will produce a warning for non-entry points that might have invalid numthreads, but for current entry points, removing the numthreads will produce an error that says they provided no numthreads attribute as well?

I'm wondering if the messaging could be a bit more clear since the first error they see will likely be something that says they didn't provide an attribute that they did and if there's an ocean of warnings, they may not connect it to this one. Maybe we can change the error for the missing attrib can say "%0entry point has no valid %1 attribute" instead of "%0 entry point must have the %1 attribute". Or maybe we could have a custom error for numthreads since we are conceding to a warning here.

bogner: The error messages do come out in order, so I'm not too worried about confusion:
```
numthrads.hlsl:2:2: warning: Group size of 4096 (64 * 64 * 1) is outside of valid range [1..1024] - attribute will be ignored [-Wignored-attributes]
[numthreads(64, 64, 1)]
 ^~~~~~~~~~~~~~~~~~~~~
numthrads.hlsl:3:6: error: compute entry point must have the numthreads attribute
void main() {}
     ^
```
That said, I'm not opposed to your wording, and if it can sometimes make things clearer then I don't see a downside really. Would you prefer I include that in this change or that I do it as a follow up? It's entirely mechanical but it does update a lot of tests, so it might be clearer as a separate change.

bogner: Follow up that rewords the error here: https://github.com/microsoft/DirectXShaderCompiler/pull/6024

llvm-beanz (approved):

PR #6024, opened 2023-11-15
[Diag] Reword `err_hlsl_missing_attr`

bogner:
Instead of "must have the X attribute" we say "must have a valid X
attribute". This should make things clearer in cases where there was
something wrong with the attribute and we've dropped it, like we do if
`numthreads` is specified incorrectly.


llvm-beanz (approved):

bob80905 (approved):

pow2clk (approved):
Thanks!

PR #6027, opened 2023-11-16
[SPIR-V] Implement vk::ext_builtin_input and vk::ext_builtin_output

### LINKED ISSUE #4217: [SPIRV] Missing gl_NumWorkGroups
I'm currently using this to atomically count the workgroups as such:
```glsl
if( ( gl_LocalInvocationID.x == 0 ) && ( workgrAtomicCounterShared == gl_NumWorkGroups.x - 1 ) )
//...... DO some work on the final value of a counter
```
In SPIRV gl_NumWorkGroups would be:
```spirv
OpDecorate %gl_NumWorkGroups BuiltIn NumWorkgroups
```

How do I access the same BuiltIn in HLSL as it's not exposed through any SV ?

cassiebeckley:
I definitely think it would look better if we allowed these attributes on variables, ie https://github.com/microsoft/hlsl-specs/issues/76. I haven't fully investigated how involved it would be to implement, but my intuition is that it wouldn't take that much more work.

Fixes #4217.


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -3459,15 +3459,15 @@ void DeclResultIdMapper::decorateInterpolationMode(
 
 SpirvVariable *DeclResultIdMapper::getBuiltinVar(spv::BuiltIn builtIn,
                                                  QualType type,
-                                                 SourceLocation loc) {
+                                                 SourceLocation loc,
+                                                 spv::StorageClass sc) {
```
s-perron, line 3463: I like to avoid default parameters when possible. Could we split this into two function. One takes the storage class, and does not include the switch that sets the storage class. The other does not take the storage class. It infers the stroage class and then calls the first function.

s-perron:
> I definitely think it would look better if we allowed these attributes on variables

That could be better. Let me know if you think the implementation would be to complicated. We can go back and change HLSL specs.

s-perron (commented):
Looks good. Just a little style comment.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
cassiebeckley: Sure, I'll push up a change with this in it before looking into allowing the attribute on variables.

cassiebeckley:
@s-perron lmk what you think of this approach.

tools/clang/test/CodeGenSPIRV_Lit/spv.inline.builtin.output.hlsl
```diff
@@ -0,0 +1,16 @@
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv | FileCheck %s
+
+// CHECK: OpEntryPoint Fragment %main "main" [[fragStencilVar:%[0-9]+]]
+// CHECK: OpDecorate [[fragStencilVar]] BuiltIn FragStencilRefEXT
+
+// CHECK: [[fragStencilVar]] = OpVariable %_ptr_Output_int Output
+
+//[[vk::ext_extension("SPV_EXT_shader_stencil_export")]]
+[[vk::ext_builtin_output(/* FragStencilRefEXT */ 5014)]]
```
Keenuts, line 9: Just by curiosity I tried this:

```
  9 [[vk::ext_builtin_input(/* FragStencilRefEXT */ 5014)]]
 10 [[vk::ext_builtin_output(/* FragStencilRefEXT */ 5014)]]
 11 static int gl_FragStencilRefARB;
```

Because the input condition is first, the input decoration is chosen. But there are no diagnostic to say I'm doing something wrong.
Also, if I set it to input, I get a validation error as the input with a type float or int must be marked Flat.

```
fatal error: generated SPIR-V is invalid: [VUID-StandaloneSpirv-Flat-04744] Fragment OpEntryPoint operand 2 with Input interfaces with integer or float type must have a Flat decoration for Entry Point id 1.
  %2 = OpVariable %_ptr_Input_int Input

```

(To make this trigger, I replaced the line `gl_FragStencilRefARB = 10;` with `gl_FragStencilRefARB;`

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -3467,7 +3476,6 @@ SpirvVariable *DeclResultIdMapper::getBuiltinVar(spv::BuiltIn builtIn,
     return builtInVar->second;
   }
   bool mayNeedFlatDecoration = false;
-  spv::StorageClass sc = spv::StorageClass::Max;
   // Valid builtins supported
   switch (builtIn) {
```
s-perron, line 3480: I would move this code to the version of the `getBuiltinVar` that does not take a storage class.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -3536,6 +3545,12 @@ SpirvVariable *DeclResultIdMapper::getBuiltinVar(spv::BuiltIn builtIn,
   return var;
 }
 
+SpirvVariable *DeclResultIdMapper::getBuiltinVar(spv::BuiltIn builtIn,
+                                                 QualType type,
+                                                 SourceLocation loc) {
+  return getBuiltinVar(builtIn, type, spv::StorageClass::Max, loc);
```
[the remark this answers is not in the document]
s-perron: I would move the code that infers the storage class based on `builtIn` into this function, and pass in that storage class instead. There is no need to pass in `Max`.

s-perron:
> @s-perron lmk what you think of this approach.

I think this looks good. We should probably add more testing. For example, what happens if the builtin is passed as a parameter to a function? 

Making the variable static is not a problem. The input and output variables are not shared, and static tells you that. I wonder if we should enforce that. Add an error if we try to apply the attribute to a variable that is not static. We should update the spec to say that.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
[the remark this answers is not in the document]
cassiebeckley: That's fair, I think I was trying to avoid needing to pass in `needsLegalization` and `mayNeedFlatDecoration` as well, but looks like I'll need to look closer at `mayNeedFlatDecoration` too.

cassiebeckley:
> I think this looks good. We should probably add more testing. For example, what happens if the builtin is passed as a parameter to a function?

Do you mean something like `someFunction(gl_NumWorkGroups)`? If so, I've added a test case. I've also added an error message enforcing `static` for `ext_builtin_input` and `ext_builtin_output`.

tools/clang/test/CodeGenSPIRV_Lit/spv.inline.builtin.output.hlsl
cassiebeckley: I've created an error diagnostic for the first case, and I've handled the second automatically adding the `Flat` decoration in the necessary cases.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
cassiebeckley: Done.

Keenuts:
Hello!

Found another cases where the diagnostic should be improved

```hlsl
[[vk::ext_builtin_output(/* FragStencilRefEXT */ 5014)]]
static int output;

[[vk::ext_builtin_input(/* FragStencilRefEXT */ 5014)]]
static int input;

[[vk::ext_extension("SPV_EXT_shader_stencil_export")]]
void main() {
  int tmp = input;
  output = 10;
}
```

```
$ dxc -T ps_6_0 -E main -fcgl -spirv repro.hlsl
repro.hlsl:10:3: error: cannot assign to input variable
  output = 10;
```

If input is not read, it works. The error seems wrong as it puts the blame on output, which is marked as output. We may want to error out before, if 2 variables share the same builtin but with different I/O.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -1641,6 +1641,28 @@ bool SpirvEmitter::validateVKAttributes(const NamedDecl *decl) {
     }
   }
 
+  // a VarDecl should have only one of vk::ext_builtin_input or
```
Keenuts, line 1644: Shall this case be handled?

```hlsl
cbuffer Constants {
    [[vk::ext_builtin_output(/* FragStencilRefEXT */ 5014)]]
    int builtin;
}
```

Right now it works, and builtin becomes constant (since in a cbuffer) so cannot be assigned.

Keenuts (approved):
Fine with this.
It's easy to make the validation fails, but I'd assume that's like inline assembly, you are on your own to make sure you use this correctly.

tools/clang/test/CodeGenSPIRV_Lit/spv.inline.builtin.input.hlsl
```diff
@@ -0,0 +1,23 @@
...
+}
+
+[numthreads(32,1,1)]
+void main() {
+  // CHECK: {{%[0-9]+}} = OpLoad %v3uint %gl_NumWorkGroups
+  uint3 numWorkgroups = gl_NumWorkGroups;
+// CHECK: [[nwg:%[0-9]+]] = OpLoad %v3uint %gl_NumWorkGroups
+// CHECK:                   OpStore %param_var_v [[nwg]]
+// CHECK:                   OpFunctionCall %uint %square_x %param_var_v
+  square_x(gl_NumWorkGroups);
```
s-perron, line 22: Can you also add a test for an output variable as the out parameter of a function? What should we do with inout parameters?

s-perron (commented):
Just a couple more test to add.

tools/clang/test/CodeGenSPIRV_Lit/spv.inline.builtin.input.hlsl
cassiebeckley: Sure, I've added one. I've also added an error for using a builtin input as an inout parameter, since by definition the function will assign to it. This doesn't happen in all cases in practice, since the SPIR-V backend just passes a pointer instead of actually doing copy-in copy-out.

cassiebeckley:
> If input is not read, it works. The error seems wrong as it puts the blame on output, which is marked as output. We may want to error out before, if 2 variables share the same builtin but with different I/O.

Sure, I've added an error for that.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: I think the error message for assigning to a constant is sufficient; I don't think there needs to be a special error message for declaring an output as constant.

s-perron (commented):
Okay, we let this sit until we update HLSL specs, and hear back from the community on some of the finer details.

cassiebeckley:
I've added a `const` requirement for `ext_builtin_input`

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -6408,6 +6479,10 @@ void SpirvEmitter::storeValue(SpirvInstruction *lhsPtr,
   if (!lhsPtr || !rhsVal)
     return;
 
+  if (lhsPtr->getStorageClass() == spv::StorageClass::Input) {
+    emitError("cannot assign to input variable", loc);
+  }
+
```
s-perron, line 6485: Is this still needed? Is the const qualification?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -2923,6 +2988,12 @@ SpirvInstruction *SpirvEmitter::processCall(const CallExpr *callExpr) {
     if (const auto *declRefExpr = dyn_cast<DeclRefExpr>(arg)) {
       argInfo = declIdMapper.getDeclEvalInfo(declRefExpr->getDecl(),
                                              arg->getLocStart());
+
+      if (canActAsOutParmVar(param) &&
+          declRefExpr->getDecl()->hasAttr<VKExtBuiltinInputAttr>()) {
+        emitError("cannot use builtin input as inout argument",
+                  arg->getExprLoc());
+      }
```
s-perron, line 2996: Is this still needed? Is the const qualification?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: It's not needed; I forgot to remove it.

s-perron (approved):

PR #6029, opened 2023-11-16
[SPIR-V] Fix interlock output parameter validation

### LINKED ISSUE #5772: [SPIR-V] Compile failure Interlocked*() functions with output a struct member
**Description**

Interlocked*() functions with the out parameter as part of struct fail to compile for SPIR-V

**Steps to Reproduce**

Example Shader
```
groupshared uint lds;

struct Struct
{
    uint member;
};

[numthreads(64, 1, 1)]
void main()
{
    Struct s;
    InterlockedAdd(lds, 1, s.member);
}
```

Error:

**Actual Behavior**
```
error: InterlockedCompareExchange requires a reference as output parameter
    InterlockedAdd(lds, 1, s.member);
```
Compiles correctly for DXIL.  Failure is only SPIRV.

**Environment**
- DXC version  1.7 - 1.7.0.4123 (8c201e0bd) (head)
- Host Operating System Win10

Keenuts:
Interlock* functions takes output parameters. Because HLSL had no real concept of references, this is allowed:

```
  int value;
  takeAnOutput(value);
```

This weirdness makes using `isLValue()` to validate output parameters a bit tricky. If a function returns a local array, and we take the subscript of this r-value, the r-value is now marked as l-value, but this shouldn't be allowed.

This forces us to walk the expression, and add heuristics to determine if the output parameter is valid or not. Not pretty, but I think this is required until HLSL gets proper references.

Fixes #5772

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -8978,12 +8978,15 @@ SpirvEmitter::processIntrinsicFirstbit(const CallExpr *callExpr,
                                        srcRange);
 }
 
-// Determines if the given expression is a valid output parameter to pass to
-// interlock* functions. Ideally, we could just check if the passed value is a
-// reference, but the HLSL frontend allows integers to be passed as output
-// parameters, without being marked as reference in the AST. This means we need
-// to walk the expression a bit to determine if it it a valid output parameter.
-// This function might miss some cases.
+// Returns true is the given expression can be used as an output parameter.
+//
+// Warning: this function could return false negatives.
+// HLSL had no references, and the AST handling of lvalues and rvalues is not
+// to be trusted for this usage.
+//  - output variables can be passed but marked as rvalues.
+//  - rvalues can becomes lvalues due to an operator call.
+// This means we need to walk the expression, and explicitly allow some cases.
+// I might have missed a valid use-case, hence the risk of false-negative.
 bool isValidOutputArgument(const Expr *expr) {
   // This could be either a member from an R-value, or an L-value. Checking
   // struct.
@@ -9004,7 +9007,7 @@ bool isValidOutputArgument(const Expr *expr) {
     return isValidOutputArgument(cast->getSubExpr());
 
   // For call operators, we trust the LValue() method.
-  // Haven't find cases where this one lies.
+  // Haven't found a cases where this is not true.
   if (const CXXOperatorCallExpr *call = dyn_cast<CXXOperatorCallExpr>(expr))
     return call->isLValue();
   if (const CallExpr *call = dyn_cast<CallExpr>(expr))
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -8978,6 +8978,48 @@ SpirvEmitter::processIntrinsicFirstbit(const CallExpr *callExpr,
                                        srcRange);
 }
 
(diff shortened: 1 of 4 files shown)


tools/clang/test/CodeGenSPIRV_Lit/intrinsics.interlocked-methods.compareexchange.reference.hlsl
```diff
@@ -2,8 +2,31 @@
 
 groupshared uint value;
 
+uint getValue() {
+  return 0;
+}
+
+uint getVector() {
```
s-perron, line 9: Should this be returning `uint2`?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -8978,6 +8978,48 @@ SpirvEmitter::processIntrinsicFirstbit(const CallExpr *callExpr,
                                        srcRange);
 }
 
+// Determines if the given expression is a valid output parameter to pass to
+// interlock* functions. Ideally, we could just check if the passed value is a
+// reference, but the HLSL frontend allows integers to be passed as output
+// parameters, without being marked as reference in the AST. This means we need
+// to walk the expression a bit to determine if it it a valid output parameter.
+// This function might miss some cases.
```
s-perron, line 8986: Please clarify "miss some cases." Hopefully we could say something like

> Returns false if `expr` is definitely an invalid output parameter for the interlock functions. Note that `expr` might be invalid even if the function returns true.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -8978,6 +8978,48 @@ SpirvEmitter::processIntrinsicFirstbit(const CallExpr *callExpr,
...
+  // Accessor to HLSL vectors.
+  if (const HLSLVectorElementExpr *vec = dyn_cast<HLSLVectorElementExpr>(expr))
+    return isValidOutputArgument(vec->getBase());
+
+  // Going through implicit casts.
+  if (const ImplicitCastExpr *cast = dyn_cast<ImplicitCastExpr>(expr))
+    return isValidOutputArgument(cast->getSubExpr());
+
+  // For call operators, we trust the LValue() method.
+  // Haven't find cases where this one lies.
```
s-perron, line 9007:
```suggestion
  // Haven't found a case where this is not true.
```

s-perron (approved):
Looks good. Just some suggestions for the comments.

cassiebeckley (approved):

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.interlocked-methods.compareexchange.reference.hlsl
Keenuts: Oh right. Also noticed I hadn't commited my changed to the loop...

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts: Reworded the whole comment to have a short usage sentence, and then the explanation around false-negatives.

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.interlocked-methods.compareexchange.reference.hlsl
Keenuts: Turns out those were unrelated changes... Fixed the 2 small nits and working, no change to the loop.

s-perron (approved):

PR #6032, opened 2023-11-16
[SM66,SPIRV] Implement InterlockedExchange for floats

s-perron:
This implements the atomic exchange for floating point values that was
added in SM6.6. See https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_Int64_and_Float_Atomics.html#interlockedexchange.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -9012,9 +9012,10 @@ SpirvEmitter::processIntrinsicInterlockedMethod(const CallExpr *expr,
   const auto srcLoc = expr->getExprLoc();
   const auto baseType = dest->getType()->getCanonicalTypeUnqualified();
 
-  if (!baseType->isIntegerType()) {
-    emitError("can only perform atomic operations on scalar integer values",
-              dest->getLocStart());
+  if (!baseType->isIntegerType() && !baseType->isFloatingType()) {
+    emitError(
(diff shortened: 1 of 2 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -9012,9 +9012,10 @@ SpirvEmitter::processIntrinsicInterlockedMethod(const CallExpr *expr,
   const auto srcLoc = expr->getExprLoc();
   const auto baseType = dest->getType()->getCanonicalTypeUnqualified();
 
-  if (!baseType->isIntegerType()) {
-    emitError("can only perform atomic operations on scalar integer values",
-              dest->getLocStart());
+  if (!baseType->isIntegerType() && !baseType->isFloatingType()) {
```
s-perron, line 9015: I did not add more testing that this error is emitted, because in general is should never be. It is the job of the parser to determine if the correct types are used for the atomic operations.

bob80905: Would it make sense to make this an llvm_unreachable?

cassiebeckley (approved):

Keenuts (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
s-perron: Good call. I'll do that.

PR #6033, opened 2023-11-16
PIX: Instrument ret instructions for shader debugging

jeffnn:
Got a couple reasons to do this: 1) so that users can step over the last value assignment and actually see that value in the debugger and 2) so that I can easily distinguish the final instruction in DXR traces wherein two different function's traces might be intermingled in the debug output data in PIX's UAV.


tex3d (approved):
LGTM.

adam-yang (approved):

PR #6041, opened 2023-11-20
Fix for an issue when noninterpolation is decorated on structure inputs.

ShchchowAMD:
**Fix for**

Fix issue under:
https://github.com/microsoft/DirectXShaderCompiler/issues/2955
For structure type, when `nointerpolation` is decorated on a structure input, this flag should be broadcast to its members.

PR (https://github.com/microsoft/DirectXShaderCompiler/pull/6018) also helps to resolve an issue when SPIRV backend generated variable has no AST type.

**Test**
Please see same case under:
https://github.com/microsoft/DirectXShaderCompiler/pull/6018

Besides, example as below should be invalid as the first parameter of function `compute` should has only one spirv-type.

If its parameter type declaration has been expanded to an array implicitly, it should not accept other interpolated inputs as its input parameter.

```
struct S {
  float4 a : COLOR;
};

float compute(float4 a) {
  return GetAttributeAtVertex(a, 2)[0];
}

float4 main(nointerpolation S s, float4 b : COLOR2) : SV_TARGET
//float4 main(nointerpolation S s) : SV_TARGET
{
  return float4(0, 0, compute(b), compute(s.a));
  //return float4(0, 0, 0, compute(s.a));
}
```
I added an error report point in this commit and gets following reports:
```
1.hlsl:12:31: error: Current function could only use noninterpolated variable as input.
  return float4(0, 0, compute(b), compute(s.a));
                              ^
fatal error: generated SPIR-V is invalid: OpFunctionCall Argument <id> '37[%param_var_a_0]'s type does not match Function <id> '24[%_ptr_Function_v4float]'s parameter type.
  %45 = OpFunctionCall %float %compute %param_var_a_0

```
Any idea to let the error only reported before spirv validator?

**Ref**:
https://github.com/microsoft/DirectXShaderCompiler/issues/2955
https://github.com/microsoft/DirectXShaderCompiler/pull/6018


ShchchowAMD:
@s-perron @Keenuts

Keenuts:
> Any idea to let the error only reported before spirv validator?

emitFatal should be enough. Looks like we don't check for error in the SpirvEmitter once we got the module.

I'll add a check to early return in such cases.

Keenuts:
Sent https://github.com/microsoft/DirectXShaderCompiler/pull/6044 to solve this.

s-perron:
Sorry, I'm confused. Is there a test that this PR is fixing which we can add as a test case? Does it add to #6018 or does it replace it?

ShchchowAMD:
> Sorry, I'm confused. Is there a test that this PR is fixing which we can add as a test case? Does it add to #6018 or does it replace it?

I think I make a mistake. For this fix, test case should be a little different from Keenuts' test case (as showed in PR message). I'll add another one for this fix. Thanks for reminding.

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.get-attribute-at-vertex.s.funcParam.hlsl
```diff
@@ -0,0 +1,21 @@
+// RUN: %dxc -T ps_6_1 -E main  %s -spirv -Od 2>&1 | FileCheck %s
```
s-perron, line 1: If you use `-Od` the legalization pass could still run. Using `-fcgl` will make sure the legalization passes do not run either.

```suggestion
// RUN: %dxc -T ps_6_1 -E main  %s -spirv -fcgl 2>&1 | FileCheck %s
```

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.get-attribute-at-vertex.s.funcParam.hlsl
```diff
@@ -0,0 +1,21 @@
...
+{
+  return float4(0, 0, 0, compute(s.a));
+}
+
+// CHECK: [[param_var_a:%[a-zA-Z0-9_]+]] = OpVariable %_ptr_Function__arr_v4float_uint_3 Function
+// CHECK: [[inst32:%[0-9_]+]] = OpAccessChain %_ptr_Function_v4float [[param_var_a]] %uint_0
+// CHECK: [[inst33:%[0-9_]+]] = OpAccessChain %_ptr_Function_v4float [[s:%[a-zA-Z0-9_]+]] %int_0 %uint_0
+// CHECK: [[inst34:%[0-9_]+]] = OpLoad %v4float [[inst33]]
+// CHECK: OpStore [[inst32]] [[inst34]]
+// CHECK: [[inst35:%[0-9_]+]] = OpFunctionCall %float %compute [[param_var_a]]
```
s-perron, line 21: I don't understand why this is correct. It seems to only be storing the attribute at vertex 0. Should it be copying the attribute at all three vertices?

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.get-attribute-at-vertex.s.funcParam.neg.hlsl
```diff
@@ -0,0 +1,16 @@
...
+float compute(float4 a) {
+  return GetAttributeAtVertex(a, 2)[0];
+}
+
+float4 main(nointerpolation S s, float4 b : COLOR2) : SV_TARGET
+{
+  return float4(0, 0, compute(b), compute(s.a));
+}
+
+// CHECK: error: Current function could only use noninterpolated variable as input.
```
s-perron, line 16: Could we improve the error message by adding the name of the current function? In more complicated cases, it could be hard to pick out which function requires noninterpolated variable.

tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
```diff
@@ -306,11 +306,35 @@ bool PervertexInputVisitor::visit(SpirvFunctionCall *inst) {
...
+    if (funcParam->isNoninterpolated()) {
+      /// Error: this broadcast process is from top to lower, hence this
+      ///        argument should be noninterpolated (will be expanded) here.
+      ///        When any matched param is noninterpolated, it means one or more
+      ///        noninterpolated variable will be passed as an expanded array
+      for (auto caller : paramCaller[funcParam])
+        if (!caller->isNoninterpolated()) {
+          emitError("Current function could only use noninterpolated variable "
+                    "as input.",
+                    caller->getSourceLocation());
```
s-perron, line 333: Please make the function name part of the error.

s-perron (approved):

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.get-attribute-at-vertex.s.funcParam.hlsl
ShchchowAMD: Will fix.

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.get-attribute-at-vertex.s.funcParam.neg.hlsl
ShchchowAMD: Will fix.

tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
ShchchowAMD: Will fix.

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.get-attribute-at-vertex.s.funcParam.hlsl
ShchchowAMD: Yes, this is wrong. Current visitor doesn't add structure variable and its relationship with local function parameter in new visitor. I'll make a fix.

ShchchowAMD:
@Keenuts  Please help to review, thanks!

tatus
s-perron, line 1: Was this checked in by mistake?

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.get-attribute-at-vertex.s.funcParam.hlsl
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// CHECK: [[param_var_a:%[a-zA-Z0-9_]+]] = OpVariable %_ptr_Function__arr_v4float_uint_3 Function
+// CHECK: [[inst32:%[0-9_]+]] = OpAccessChain %_ptr_Function_v4float [[param_var_a]] %uint_0
+// CHECK: [[inst33:%[0-9_]+]] = OpAccessChain %_ptr_Function__arr_v4float_uint_3 [[s:%[a-zA-Z0-9_]+]] %int_0
+// CHECK: [[inst34:%[0-9_]+]] = OpLoad %_arr_v4float_uint_3 [[inst33]]
+// CHECK: OpStore [[param_var_a]] [[inst34]]
+// CHECK: [[inst35:%[0-9_]+]] = OpAccessChain %_ptr_Function_v4float [[s]] %int_0 %uint_0
+// CHECK: [[inst36:%[0-9_]+]] = OpLoad %v4float [[inst35]]
+// CHECK: OpStore [[inst32]] [[inst36]]
```
s-perron, line 23: Why do you need both stores? It seems like the second is redundant.

tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
```diff
@@ -310,8 +317,28 @@ bool PervertexInputVisitor::visit(SpirvFunctionCall *inst) {
   unsigned argIndex = 0;
   for (auto *arg : inst->getArgs()) {
     if (currentFunc->getMappedFuncParam(arg)) {
-      createVertexStore(arg,
-                        createVertexLoad(currentFunc->getMappedFuncParam(arg)));
+      auto paramVar = currentFunc->getMappedFuncParam(arg);
```
s-perron, line 320: Could you move this outsize of the `if`, and use `paramVar` in the condition? This avoid two calls to the same function.

tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
```diff
@@ -310,8 +317,28 @@ bool PervertexInputVisitor::visit(SpirvFunctionCall *inst) {
...
+      if (isa<SpirvAccessChain>(paramVar)) {
+        auto tempVar = paramVar;
+        while (isa<SpirvAccessChain>(tempVar)) {
+          tempVar = dyn_cast<SpirvAccessChain>(tempVar)->getBase();
+        }
+        if (tempVar->isNoninterpolated()) {
+          /// For Structure type, expanded param needs restore to
```
s-perron, line 327: I find the wording in the comment confusing. As someone not too familiar, I do not understand what the means.

Do you mean something like:

> If the base variable is not interpolated, then we need to copy the values for all of the vertices to the local variable. This means copying an entire array. This is accomplished by removing the last index in the access chain for the parameter variable to get a pointer to the entire array.

s-perron (changes requested):
Generally looks good. I just have a couple minor suggestions. Also, the one test still confuses me. I don't know where the second store it coming from.

tatus
ShchchowAMD: Sorry, will remove it.

tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
ShchchowAMD: Will fix.

tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
ShchchowAMD: Yes. A minor difference is we need to recreate (clone) a new access chain instruction instead of just remove original last zero index, I'll show an example in following comments.

tools/clang/test/CodeGenSPIRV_Lit/intrinsics.get-attribute-at-vertex.s.funcParam.hlsl
ShchchowAMD: I'll show a detailed example below.

ShchchowAMD:
> Why do you need both stores? It seems like the second is redundant.

The first access chain is created in PerVertex Visitor, as you commented above, to help get a pointer to the entire array and pass it to function call as its argument.

The second one is created in original SPIRV generation step, which has been changed to access to first vertex value, which will be used to let all the 's.a' equal to 's.a[0]' within current function.

To make it simple, here's the conclusion of what I made for noninterpolated inputs:

1. If it is used within any function (not entry block), every access to this variable will be replaced as access to the first vertex value, except for function call and GetAttributeAt.

2. For function call and GetAttributeAt, it will still be treated as an array to make sure we have entire array values.

An example could be as following:

**1. interpolated inputs:**
HLSL:
```
struct S {
  float4 a : COLOR;
};

float compute(float4 a) {
  return a[0];
}

float4 main(S s) : SV_TARGET
{
  return float4(0, 0, 0, compute(s.a));
}
```
SPIRV:
```
...
%src_main = OpFunction %v4float None %23
          %s = OpFunctionParameter %_ptr_Function_S
   %bb_entry = OpLabel
%param_var_a = OpVariable %_ptr_Function_v4float Function
         %28 = OpAccessChain %_ptr_Function_v4float %s %int_0
         %29 = OpLoad %v4float %28
               OpStore %param_var_a %29
         %30 = OpFunctionCall %float %compute %param_var_a
         %32 = OpCompositeConstruct %v4float %float_0 %float_0 %float_0 %30
               OpReturnValue %32
               OpFunctionEnd
    %compute = OpFunction %float None %33
          %a = OpFunctionParameter %_ptr_Function_v4float
 %bb_entry_0 = OpLabel
         %37 = OpAccessChain %_ptr_Function_float %a %uint_0
         %38 = OpLoad %float %37
               OpReturnValue %38
               OpFunctionEnd
```

**2. non-interpolated inputs:**
HLSL:
```
struct S {
  float4 a : COLOR;
};

float compute(float4 a) {
  return GetAttributeAtVertex(a, 2)[0];
}

float4 main(noninterpolation S s) : SV_TARGET
{
  return float4(0, 0, 0, compute(s.a));
}
```
SPIRV:
```
...
   %src_main = OpFunction %v4float None %26
          %s = OpFunctionParameter %_ptr_Function_S
   %bb_entry = OpLabel
%param_var_a = OpVariable %_ptr_Function__arr_v4float_uint_3 Function
         %32 = OpAccessChain %_ptr_Function_v4float %param_var_a %uint_0
         %33 = OpAccessChain %_ptr_Function__arr_v4float_uint_3 %s %int_0
         %34 = OpLoad %_arr_v4float_uint_3 %33
               OpStore %param_var_a %34
         %35 = OpAccessChain %_ptr_Function_v4float %s %int_0 %uint_0
         %36 = OpLoad %v4float %35
               OpStore %32 %36
         %37 = OpFunctionCall %float %compute %param_var_a
         %39 = OpCompositeConstruct %v4float %float_0 %float_0 %float_0 %37
               OpReturnValue %39
               OpFunctionEnd
    %compute = OpFunction %float None %40
          %a = OpFunctionParameter %_ptr_Function__arr_v4float_uint_3
 %bb_entry_0 = OpLabel
%temp_var_vector = OpVariable %_ptr_Function_v4float Function
%a_perVertexParam0 = OpVariable %_ptr_Function_v4float Function
         %45 = OpAccessChain %_ptr_Function_v4float %a %uint_0
         %46 = OpLoad %v4float %45
               OpStore %a_perVertexParam0 %46
         %47 = OpAccessChain %_ptr_Function_v4float %a %uint_2
         %48 = OpLoad %v4float %47
               OpStore %temp_var_vector %48
         %50 = OpAccessChain %_ptr_Function_float %temp_var_vector %uint_0
         %51 = OpLoad %float %50
               OpReturnValue %51
               OpFunctionEnd
```

You can see,
        `%28 = OpAccessChain %_ptr_Function_v4float %s %int_0`
has been modified to
        `%35 = OpAccessChain %_ptr_Function_v4float %s %int_0 %uint_0`

Hence, following codes are added in the front of current block to make sure 'param_var_a' could be passed as an array to function call to 'compute':


 ```
%32 = OpAccessChain %_ptr_Function_v4float %param_var_a %uint_0
         %33 = OpAccessChain %_ptr_Function__arr_v4float_uint_3 %s %int_0
         %34 = OpLoad %_arr_v4float_uint_3 %33
               OpStore %param_var_a %34
```

If there's any usage to **s.a** in **main**,
        ` %35 = OpAccessChain %_ptr_Function_v4float %s %int_0 %uint_0`
Will still be used for replacement. Otherwise, it will be removed after optimization.

**Comment by Keenuts** [comment]
Hello!

Thanks for putting this together!
I don't know you, but I do feel like we might want to first look back at what's allowed, and document the proper usage for nointerpolation and GetAttributeAtVertex. Because I do think the example I shared is actually invalid. 

From https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html, it says if I use GetAttributeAtVertex, I should mark it as nointerpolation.


```hlsl
struct S {
  float4 a : COLOR;
};

float compute(float4 a) {
  return GetAttributeAtVertex(a, 0)[0];
}

float4 main(nointerpolation S s) : SV_TARGET
{
  return float4(0, 0, 0, compute(s.a));
}
```

Here, the main param `s` is indeed marked as nointerpolation. But the parameter of the function compute is not. So I do think using GetAttributeAtVertex should not be legal here.
What is legal is to call GetAttributeAtVertex in the main.

Keenuts:
> It works for DXIL (https://godbolt.org/z/n443jj4E9).

Yes, I know, but should we this time?
What bugs me is the type of the `float param` here is not `float`, but `float[3]` because of a statement in the function, and not an attribute. Seems weird to define the type from the usage and not its declaration.

If at least DXC had one HLSL->DXIL test covering this case, I'd say "sure, that's the design", but the only tests I see that aren't SPIR-V are calling GetAttributeAtVertex directly in the main, where the params are directly decorated with `nointerpolation` (or the field in the struct).

ShchchowAMD:
@Keenuts 

Hi Nathan, as detailed spec for pervertex inputs doesn't say too much about compile check or error rules, I'll share my personal thinking here:

> it says if I use GetAttributeAtVertex, I should mark it as nointerpolation.

From my personal understanding, the main idea is:
An input attributes used in GetAttributeAtVertex should be marked as **noninterpolation**, no matter it is from a Mesh or Vertex shader.

In pixel shader stage, the most important thing is we should make sure the first argument is an **input** attribute decorated with **noninterpolation** qualifier.

For example, a temp local variable, marked as noninterpolation will be meaningless when it is used as the first argument in a GetAttributeAtVertex function call like (maybe this needs an error check later if we all agree?):

```
float compute(float4 a) {
  float4 b = (1.0, 1.0, 0.2, 0.2);
  return GetAttributeAtVertex(b, 0)[0] + a[0];
}
```


> What is legal is to call GetAttributeAtVertex in the main.

I'm afraid this is not a good way for developers, including following concerns:

1. We should try to keep 'compiler error rules' as same as DXIL. This is for developers to re-use their HLSL codes or something. Not just for current case, they may get confused if DXIL could pass but SPIRV backend fails with same HLSL codes.

2. Restrict usage of a intrinsic only in the main block sounds not like a good idea.

Like above explanation, even if we may need more works in HLSL-SPIRV compiler module, if it is not definitely abandoned from spec, users should be allowed to use GetAttributeAtVertex in a stage local function call.

What we need to do is to report errors when compiler find the first argument is interpolated or not an input from final function call chain.

> What bugs me is the type of the float4 a here is not float4, but float4[3] because of a statement in the function, and not an attribute. Seems weird to define the type from the usage and not its declaration.

Yes, that's true. The key problem for **per-vertex inputs** and **noninterpolation qualifier** now, is that we need to keep legacy usage still legal.

So we can't directly require all the **noninterpolation** inputs to be an array now, that's the key question.

We now need to implicitly expand them to an three-element array in new feature. The first draft of implementation tried to let all the noninterpolated inputs be an array and received many errors reported here.

Keenuts:
> ```
> 
> ```
> 
> 
>     
>   
> 

> 
> 
> I'm afraid this is not a good way for developers, including following concerns:
> 
> 
>     1. We should try to keep 'compiler error rules' as same as DXIL. This is for developers to re-use their HLSL codes or something. Not just for current case, they may get confused if DXIL could pass but SPIRV backend fails with same HLSL codes.
> 
> 


Hi,

Thanks for the input!

> > What is legal is to call GetAttributeAtVertex in the main.
>     2. Restrict usage of a intrinsic only in the main block sounds not like a good idea.

To be clear, my case was not to restrict `GetAttributeAtVertex` to the main function.
But to require the function signature to have `nointerpolation`.
It was only to say "in this example, only legal usage should be in `main`"
The example in questions being:

```hlsl
void foo(float4 a) {
    GetAttributeAtVertex(a, 0);
}

int main(nointerpolation float4 a : COLOR) {
    foo(a);
}
```

But I do see this as valid:

```hlsl
void foo(nointerpolation float4 a) {
    GetAttributeAtVertex(a, 0);
}

int main(nointerpolation float4 a : COLOR) {
    foo(a);
}
```

> For example, a temp local variable, marked as noninterpolation will be meaningless when it is used as the first argument in a GetAttributeAtVertex function call like (maybe this needs an error check later if we all agree?):

Yes, I do think this should trigger an error. Or if we stay coherent with the `precise` keyword, accept this pattern, and propagate the `nointerpolation` annotation upstream into the main function call (a bit like `precise` would do).

```hlsl
float compute(nointerpolation float4 a) {
  return a.r;
}

float4 main(float4 a : COLOR) : SV_TARGET
{
  return float4(0, 0, 0, compute(s.a));
}
```

Same as this, which returns a validation error right now:

```hlsl
float compute(nointerpolation float4 a) {
  return GetAttributeAtVertex(a, 0)[0];
}

float4 main(float4 a : COLOR) : SV_TARGET
{
  return float4(0, 0, 0, compute(a));
}
```

s-perron:
@Keenuts The defacto standard is what is implemented in DXC for dxil. This means that there is probably lots of code that assumes the current behaviour. We should do what happens in DXIL. If you want to make changes to the spec the explicitly allow or disallow something, that is a dicussion for HLSL specs. You can bring it up there, but I'm sure your TL will tell you it is not worth it. :wink:

ShchchowAMD:
@s-perron

Should I wait for feedback from Nathan related to current discussion or should I merge this PR?

I think current fix won't against additional detailed usage description of `noninterpolation` later (another error check for examples he shared above).

Keenuts:
> Should I wait for feedback from Nathan related to current discussion or should I merge this PR?

I think we can move forward, and follow blindly what the DXIL backend allows.
But we should not crash with a validation error, so the examples above might need to be handled gracefuly.

Keenuts:
A few other examples:

```hlsl
float foo(int c, float v) {
  if (c == 0)
    return GetAttributeAtVertex(v, 1);
  else
    return v;
}

float4 main(nointerpolation float a : A) : SV_Target
{
  float tmp1 = GetAttributeAtVertex(a, 2);
  float tmp2 = foo(0, a);
  float tmp3 = foo(1, a);
  return float4(0, 0, 0, tmp1 + tmp2 + tmp3);
}
```

DXIL seems to only call the intrinsic for `tmp1` and `tmp2`, but `tmp3` gets loaded as usual, but with an undefined vertex index. While for SPIR-V, we default to the index 0.
Shall a warning be emitted to say you probably wanted to use "GetAttributeAtVertex"? Or leave as-is? (not sure if the behavior is defined for DXIL)

Also, noticed we crash for this usage of `GetAttributeAtVertex`

```hlsl
float foo(float v, uint index) {
    return GetAttributeAtVertex(v, index);
}

float4 main(nointerpolation float a : A) : SV_Target
{
  float tmp = foo(a, 1);
  return float4(0, 0, 0, tmp);
}
```

```
dxc -T ps_6_1 -E main repro.hlsl -spirv
```

Another example `GetAttributeAtVertex` causes issue is this one:

```hlsl
float foo(float a, float b, uint index) {
  if (index == 0)
    return GetAttributeAtVertex(a, 0);
  else
    return GetAttributeAtVertex(b, 0);
}

float4 main(nointerpolation float a : A) : SV_Target
{
  float b = 0;
  float tmp = foo(a, b, 0);
  return float4(0, 0, 0, tmp);
}
```

Depending if foo is called with `0` or `1`, we crash with a validation error.
And we fail in both cases with optimization disabled. Seems like this should be handled with a diagnostic early on when marking foo as taking a `nointerpolation` from its usage.

Keenuts:
Just discusses with @s-perron, and we might want to force legalization if `GetAttributeAtVertex` is used, that could solve move of those weird cases.

sudonatalie:
@Keenuts I've resolved the merge conflicts on this change and retested all (I think) of the cases mentioned in the comments here and I don't think there are any cases which this PR makes *worse*, so what do you think about merging it as-is, and leaving follow-up issues for error message improvements? I've created:
* https://github.com/microsoft/DirectXShaderCompiler/issues/6382
* https://github.com/microsoft/DirectXShaderCompiler/issues/6383
* https://github.com/microsoft/DirectXShaderCompiler/issues/6384

sudonatalie (approved):

PR #6047, opened 2023-11-20
Revert #1128

python3kgae:
Revert #1128 for cannot make a test.
All internal tests pass with it reverted.

### DIFF
--- a/lib/Transforms/Scalar/Scalarizer.cpp
+++ b/lib/Transforms/Scalar/Scalarizer.cpp
@@ -626,16 +626,6 @@ bool Scalarizer::visitShuffleVectorInst(ShuffleVectorInst &SVI) {
       Res[I] = Op0[Selector];
     else
       Res[I] = Op1[Selector - Op0.size()];
-    // HLSL Change Begins: (fix bug in upstream llvm)
-    if (ExtractElementInst *EA = dyn_cast<ExtractElementInst>(Res[I])) {
-      // Clone extractelement here, since it is associated with another inst.
-      // Otherwise it will be added to our Gather, and after the incoming
-      // instruction is processed, it will be replaced without updating our
-      // Gather entry.  This dead instruction will be accessed by finish(),
-      // causing assert or crash.
-      Res[I] = IRBuilder<>(&SVI).Insert(EA->clone());
-    }
-    // HLSL Change Ends
   }
   gather(&SVI, Res);
   return true;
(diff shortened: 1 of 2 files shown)


dmpots:
Why are we reverting this change?

python3kgae:
> Why are we reverting this change?

We thought the code is useless.

tex3d:
I think there was a real bug being worked around here with this clone of the extractelement inst.

At first, I thought this bug was fixed in upstream in this change:
https://github.com/llvm/llvm-project/commit/db05f2e34a5e9380ddcc199d6687531108d795e4
with the test case that could likely trigger this shuffle scenario being: `llvm/test/Transforms/Scalarizer/crash-bug.ll`

I think the issue is caused by deleting component instructions when overriding the value vector inside gather().  When the shuffle is processed before an instruction used in the shuffle's operands, a Scatterer is created for that operand to get scalar values, which will create extractelement instructions from the other instruction's vector result (in the crash-bug.ll test this is the PHI).  Later, when visiting the PHI instruction, we replace these values in the Scattered for that PHI, deleting them right away, yet, the Scattered/Gathered entry for the select still points to these deleted instructions.

It's also possible that hitting an actual crash or something that produces a reliable failure requires using the shuffle vector value, forcing it to reconstruct it from the dead values in the cache, and using those values in a way that causes it to crash, perhaps by running under some heap verifier that will cause a fault if freed memory is accessed.  Then this still might not trigger until you do something with the dead instructions later that causes it to access that freed memory.

However, following the history of this test revealed some additional related changes, one of which was the real change that would have fixed the issue with select originally, by using ReversePostOrderTraversal:
https://github.com/llvm/llvm-project/commit/3d4579829e85c108e729ade64d778e614b702e16

Perhaps we could apply the ReversePostOrderTraversal fix while reverting the change that cloned the extractelement inst.  And maybe we can build a test based off of the `llvm/test/Transforms/Scalarizer/crash-bug.ll` from upstream llvm.

dmpots:
I am still confused on why we are trying to revert it now. Based on the comment from Tex it sounds like there was a real bug being fixed here? If so then I think we should have a sequence of commits that

1. Add a test that demonstrates the bug
2. Revert the original fix (the test should now fail)
3. Add the "better" fix from upstream

It's also not clear to me why the upstream fix is better.

If it's not actually fixing a bug then I still don't see a strong motivation to remove it. It seems like the code works as is and reverting it now just adds more risk.

Is there a problem with the way it is currently written?

damyanp:

damyanp:

Issue #4338, opened 2022-03-18
[SPIR-V] Error with switch statement with constant input


simontaylor81:
If I try to create a switch statement with a literal value as the input, I get an error like the following when compiling with spirv:
```
file.hlsl:5:7: error: non-32bit integer case value in switch statement unimplemented
        case 0:
             ^
```

This link should reproduce the issue: https://shader-playground.timjones.io/6fd31f9898afb861d22de36c0f989a85

Shader code, in case that link goes stale:
```
float PSMain() : SV_TARGET
{
    switch (0)
	{
	case 0:
        return 1;
    }
	return 0;
}
```

(Obviously you generally would not hand-write code like that, but in our case the code is generated).

The issue can be worked-around by assigning the value to a dummy variable, or introducing an (identity) cast, e.g. `switch ((int)0)`

PR #6049, opened 2023-11-21
[SPIR-V] Add support for 64-bit switch sel

sudonatalie:
Allow 64-bit integer variables as switch statement selectors and clarify the error message for literal selectors (related to #4338).

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -25,8 +25,10 @@
 #include "clang/SPIRV/AstTypeProbe.h"
 #include "clang/SPIRV/String.h"
 #include "clang/Sema/Sema.h"
+#include "llvm/ADT/APInt.h"
 #include "llvm/ADT/SetVector.h"
 #include "llvm/ADT/StringExtras.h"
+#include "llvm/Support/Casting.h"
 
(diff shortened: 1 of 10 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -13530,6 +13527,13 @@ void SpirvEmitter::processSwitchStmtUsingSpirvOpSwitch(
     doDeclStmt(condVarDeclStmt);
 
   auto *cond = switchStmt->getCond();
+  if (llvm::dyn_cast<IntegerLiteral>(cond)) {
+    emitError(
+        "integer literal selectors in switch statements not yet implemented",
+        cond->getLocStart());
+    return;
+  }
```
s-perron, line 13535: Under which conditions does this error come up? I see an error as a condition that could happen if the programmer did something wrong. There should be a test that these are emitted correctly. An assert or llvm unreachable is a case that should never happen. If it does, then there is something wrong in the code.

Please add a test for this error.

sudonatalie: Test added. It used to fail with `error: non-32bit integer case value in switch statement unimplemented` which was a bit confusing since it was a 32-bit integer literal. I haven't gotten to actually implementing that, but that was my original motivation that led me to this fix first on the way there.

s-perron (approved):

PR #6050, opened 2023-11-21
CMake dxc as subproject fixes

amaiorano:
These changes allow us to use DXC as as CMake subproject (via add_subdirectory).

Commit comments from each commit:

===

cmake: allow CMAKE_INSTALL_RPATH to be defined in superproject on APPLE platforms

As on Linux, allow CMAKE_INSTALL_RPATH to be defined in a CMake superproject.

===

cmake: Allow LLVM_ENABLE_ASSERTIONS=OFF to disable assertions in Debug builds

Currently, this flag is only used to enable assertions in non-Debug
builds, but is not respected in Debug builds. That is, if set to false,
Debug builds will still assert. This change fixes that.

Note that by default, LLVM_ENABLE_ASSERTIONS is true in Debug builds, so
unless it's explicitly set to false, this is a no-op for most people.

===

cmake: Allow DIRECTX_HEADERS_INCLUDE_DIR to be defined from superproject

Useful for when DXC is included as a subdirectory. In Dawn/Chrome, we
use our own mirror of DirectX-Headers instead of checking out the repo's
submodules, so we need to override this variable.


amaiorano:
Let me know if you prefer I split this into three PRs, one per commit.

s-perron (commented):
I don't have any opinion on this PR. From an implementation perspective it looks fine to me. My only question is if we really want to disable assertions in debug build by default. I think that is what is happening. Many developers probably expect that behaviour, and will be caught off guard.

amaiorano:
> My only question is if we really want to disable assertions in debug build by default. I think that is what is happening. Many developers probably expect that behaviour, and will be caught off guard.

This change does not disable asserts by default. All it does is make it possible to do so by setting `LLVM_ENABLE_ASSERTION` to OFF; however, this flag is ON by default in Debug builds.

s-perron (approved):

hekota (approved):
LGTM!

PR #6056, opened 2023-11-23
CMake: fix FindD3D12.cmake when using non-VS generators

amaiorano:
FindD3D12.cmake sets WIN10_SDK_VERSION to
CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION; however, this latter variable is only defined when using a Visual Studio CMake generator. When using Ninja instead, for example, this variable will be empty, and WIN10_SDK_VERSION will remain unset.

This change detects when CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION and attempts to retrieve the latest SDK by listing directories under ${WIN10_SDK_PATH}/Include.

### DIFF
--- a/cmake/modules/FindD3D12.cmake
+++ b/cmake/modules/FindD3D12.cmake
@@ -1,7 +1,18 @@
 # Find the Win10 SDK path.
 if ("$ENV{WIN10_SDK_PATH}$ENV{WIN10_SDK_VERSION}" STREQUAL "" )
   get_filename_component(WIN10_SDK_PATH "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots;KitsRoot10]" ABSOLUTE CACHE)
-  set (WIN10_SDK_VERSION ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION})
+  if (CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION)
+    set (WIN10_SDK_VERSION ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION})
+  else()
+    # CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION may not be defined if, for example,
+    # the Ninja generator is used instead of Visual Studio. Attempt to retrieve the
+    # most recent SDK version from the list of paths under "${WIN10_SDK_PATH}/Include/".
+    file(GLOB sdk_dirs RELATIVE "${WIN10_SDK_PATH}/Include/" "${WIN10_SDK_PATH}/Include/10.*")
+    if (sdk_dirs)
+      list(POP_BACK sdk_dirs WIN10_SDK_VERSION)
+    endif()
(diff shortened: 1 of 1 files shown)


amaiorano:
This is also needed for our CMake built of Dawn, as we use Ninja to build faster. I suspect this would also fix issues when building DXC using "Open Folder" which also uses Ninja instead of MSBuild, but I haven't tested this path.

cmake/modules/FindD3D12.cmake
```diff
@@ -1,7 +1,18 @@
...
+    set (WIN10_SDK_VERSION ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION})
+  else()
+    # CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION may not be defined if, for example,
+    # the Ninja generator is used instead of Visual Studio. Attempt to retrieve the
+    # most recent SDK version from the list of paths under "${WIN10_SDK_PATH}/Include/".
+    file(GLOB sdk_dirs RELATIVE "${WIN10_SDK_PATH}/Include/" "${WIN10_SDK_PATH}/Include/10.*")
+    if (sdk_dirs)
+      list(POP_BACK sdk_dirs WIN10_SDK_VERSION)
+    endif()
+    unset(sdk_dirs)
```
amaiorano, line 14: Note that this makes use of `file(GLOB...)` to get the list of directories from the file system. There is a way to actually get this list from the registry, but it requires CMake 3.24 to make use of the [windows registry query function to get the list of subkeys](https://cmake.org/cmake/help/latest/command/cmake_host_system_information.html#query-windows-registry). The code would look like this:

```
  # Using the Ninja generator will not set CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION.
  # Attempt to retrieve the most recent SDK version by retrieving the last subkeys under "Installed Roots"
  cmake_host_system_information(RESULT INSTALLED_ROOTS QUERY WINDOWS_REGISTRY "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots" SUBKEYS ERROR_VARIABLE ERROR_RESULT)
  list(POP_BACK INSTALLED_ROOTS WIN10_SDK_VERSION)
```

But since DXC requires CMake 3.17.2, I went with file globbing instead.

hekota (approved):
LGTM!

cmake/modules/FindD3D12.cmake
llvm-beanz: In CMake 3.17 (and many earlier versions) there is an undocumented feature of `get_filename_component` to read registry values. We actually use that in this file a few lines above your change.

amaiorano: > In CMake 3.17 (and many earlier versions) there is an undocumented feature of `get_filename_component` to read registry values. We actually use that in this file a few lines above your change.

Yeah, but unfortunately it cannot be used to read subkeys. I looked at the source code for CMake to see if it was possible.

llvm-beanz (approved):

PR #6058, opened 2023-11-23
Propagate precise from formal parameter to actual parameter

s-perron:
It is unclear what should happen if an input parameter on a function is
marked as precise. I am choosing to propagate it to the actual
parameters.

Note that it is not wrong to propagate precise more than we should since
it will only stop some optimizations.


tools/clang/test/CodeGenSPIRV_Lit/decoration.no-contraction.parameters.hlsl
```diff
@@ -0,0 +1,20 @@
...
+precise float4 MakePrecise(precise float4 v) { return v; }
+
+// CHECK: OpDecorate [[mul:%[0-9]+]] NoContraction
+
+void VSMain(VSInput input, out precise float4 OutputPos : SV_Position)
+{
+  // CHECK: [[mul]] = OpMatrixTimesVector
+  // CHECK: OpStore %param_var_v [[mul]]
+  // CHECK: OpFunctionCall %v4float %MakePrecise %param_var_v
+  OutputPos = MakePrecise(mul(input.pos, Proj));
```
s-perron, line 19: A user that this use of `MakePrecise` should cause (or at least allow) the multiple to be precise. I find the documentation for precise to be unclear, but I think it is reasonable to say that the precise on the parameter `v` should cause the expression that is computing `v`'s value should be governed by that use.

@pow2clk Do you have any opinion on this?

Keenuts: To me the documentation here is clear: https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-variable-syntax

Every operation up to the precise store should be marked as NoContraction.
But every use of the precise variable doesn't have to (as long as it doesn't contribute to another precise down the line).
So here, the mul, and all operations leading to the the value of `v` should be NoContraction, but if OutputPos is used later, doesn't have to no?

Keenuts (approved):

tools/clang/test/CodeGenSPIRV_Lit/decoration.no-contraction.parameters.hlsl
s-perron: That documentation does seem to be clearer. I was looking at https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/precise.  However, I am concerned that there is something else that needs to be done. Even that documentation is not clear. I says that the keywords on a variable restricts "any calculations used to produce the value assigned to that variable". It is a clear statement, but one that is not easily implemented. Some of the calculations could be hidden in function calls, maybe even function defined in another module that will linked. Is there an implicit understanding that we are talking about calculations in the same function that obviously contribute or not? Remember that DXIL supports shader linking at runtime. 

In this case, `OutputPos` is marked as precise. Should the multiple be nocontract simply because of that, and the precise on `v` means nothing? It is used to produce the value for `OutputPos`, but only through a function. In the SPIR-V backend, we propagate the "precise" before optimization.

Note the other test that was changed. If the output is marked as precise, and a parameter contributes to that output, then the formal parameter is marked as precise as well. This will be propagated to the actual parameter. That is not "wrong", but is possible not necessary.

Keenuts: Ahh, good point, I haven't thought about module linking. I wonder if it wouldn't be simpler if `precise` was applied downstream.. But that's irrelevant.

Given this code, I'd assume the `precise` in the param is indeed useless, as the output is downstream, and is `precise`. But this clearly makes me wonder about module linking and how that would work..

s-perron: I think this is reasonable. I think @keenuts agrees. I will merge this, and see if there are any issues.

PR #6060, opened 2023-11-23
Python: fix calls to os.makedirs to avoid existence check races

amaiorano:
In parallel builds of DXC, hctgen.py in particuilar would sometimes throw an exception from os.makedirs because the directory would already exist. The conditional check for file existence introduces a race when running the script in parallel. Fix this, and all other cases where such conditional checks were done, by using the `exist_ok = True` argument.


s-perron (approved):

llvm-beanz (approved):

hekota (approved):

amaiorano:
I rebased and force pushed to see if I can kick off the build again

PR #6061, opened 2023-11-23
[CMake] add another hint for DIASDK

python3kgae:
Fail to find DIASDK when using VisualStudio "Open a local folder". Both VS_DIA_INC_PATH and VSWHERE_LATEST are empty. But MSVC_DIA_SDK_DIR is set correctly.

Add MSVC_DIA_SDK_DIR/include as hint for DIASDK_INCLUDE_DIR.


dmpots (approved):
LGTM

PR #6077, opened 2023-11-29
[lit][SPIRV] convert runCodeTest to lit and remove dependence on effcee

python3kgae:
Convert runCodeTest to lit FileCheck test and remove all the code related to runCodeTest.

Remove effcee and re2 from git submodule since nothing is dependent on them after the change.


s-perron:
FYI: #6085 wants to add a new codegentest. This is possibly one that cannot be done by calling dxc from the command line.

python3kgae:
> FYI: #6085 wants to add a new codegentest. This is possibly one that cannot be done by calling dxc from the command line.

So I should not remove FileTestFixture and FileTestUtils?

sudonatalie (approved):
nit: Can you move these tests from `CodeGenSPIRV/legacy-runCodeTest` to the base `CodeGenSPIRV` dir? We don't need to maintain a distinction between old runCodeTest/FileTest.

Otherwise looks good, thanks for making these changes!

python3kgae:
> nit: Can you move these tests from `CodeGenSPIRV/legacy-runCodeTest` to the base `CodeGenSPIRV` dir? We don't need to maintain a distinction between old runCodeTest/FileTest.
> 
> Otherwise looks good, thanks for making these changes!

Updated.
Thanks for the review.

Keenuts (approved):

SteveUrquhart:
> FYI: #6085 wants to add a new codegentest. This is possibly one that cannot be done by calling dxc from the command line.

The test in #6085 does require dxcompiler.dll, as does InlinedCodeTest and several of the tests affected here.  How is it planned to continue this testing?

s-perron:
@SteveUrquhart If something that you need was removed by this PR, you can bring it back in, but it will need to be changed to not rely on effcee. I can help get the infrastructutre that you need back, if you can tell me exactly what you need. We could easily add classes that would take the HLSL as a string, compile it, and then return the spir-v assembly as a string. We no longer have effcee and re2 to do string matching. What would you need?

The tests that were removed by this PR were either replaced with new tests or no longer needed. For example the InlinedCode tests only tested infrastructure, and did not test any DXC features.

SteveUrquhart:
We need a test.exe that links against dxcompiler.dll to test dll-only bugs such as https://github.com/microsoft/DirectXShaderCompiler/pull/6085.  Many toolchains use the dll interface instead of dxc.exe. The input string is simply passed through, dll users don’t need string matching.

PR #6078, opened 2023-11-29
Update DXIL.rst

bharadwajy:
Make the table with supported shader model, major and minor numbers visible.

### DIFF
--- a/docs/DXIL.rst
+++ b/docs/DXIL.rst
@@ -88,21 +88,29 @@ The shader model is specified as a named metadata in DXIL::
   !dx.shaderModel = !{ !0 }
   !0 = !{ !"<shadelModelName>", i32 <major>, i32 <minor> }
 
-The following values of <shaderModelName>_<major>_<minor> are supported:
+The following values of ``<shaderModelName>``, ``<major>``, ``<minor>`` are supported:
 
-====================      ===================================== ===========
-Target                    Legacy Models                         DXIL Models
-====================      ===================================== ===========
-Vertex shader (VS)        vs_4_0, vs_4_1, vs_5_0, vs_5_1        vs_6_0
-Hull shader (HS)          hs_5_0, hs_5_1                        hs_6_0
-Domain shader (DS)        ds_5_0, ds_5_1                        ds_6_0
-Geometry shader (GS)      gs_4_0, gs_4_1, gs_5_0, gs_5_1        gs_6_0
-Pixel shader (PS)         ps_4_0, ps_4_1, ps_5_0, ps_5_1        ps_6_0
-Compute shader (CS)       cs_5_0 (cs_4_0 is mapped onto cs_5_0) cs_6_0
-Shader library            no support                            lib_6_1
-Mesh shader (MS)          no support                            ms_6_5
-Amplification shader (AS) no support                            as_6_5
-========================= ===================================== ===========
++-----------------------+---------------------------------------+-------------+
+| Target                |   Legacy Models                       | DXIL Models |
++=======================+=======================================+=============+
+| Vertex shader (VS)    | vs_4_0, vs_4_1, vs_5_0, vs_5_1        | vs_6_0      |
++-----------------------+---------------------------------------+-------------+
+| Hull shader (HS)      | hs_5_0, hs_5_1                        | hs_6_0      | 
++-----------------------+---------------------------------------+-------------+
+| Domain shader (DS)    | ds_5_0, ds_5_1                        | ds_6_0      | 
++-----------------------+---------------------------------------+-------------+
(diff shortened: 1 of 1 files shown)


llvm-beanz (approved):

docs/DXIL.rst
```diff
@@ -88,21 +88,29 @@ The shader model is specified as a named metadata in DXIL::
...
-Hull shader (HS)          hs_5_0, hs_5_1                        hs_6_0
-Domain shader (DS)        ds_5_0, ds_5_1                        ds_6_0
-Geometry shader (GS)      gs_4_0, gs_4_1, gs_5_0, gs_5_1        gs_6_0
-Pixel shader (PS)         ps_4_0, ps_4_1, ps_5_0, ps_5_1        ps_6_0
-Compute shader (CS)       cs_5_0 (cs_4_0 is mapped onto cs_5_0) cs_6_0
-Shader library            no support                            lib_6_1
-Mesh shader (MS)          no support                            ms_6_5
-Amplification shader (AS) no support                            as_6_5
-========================= ===================================== ===========
+The following values of ``<shaderModelName>``, ``<major>``, ``<minor>`` are supported:
```
tex3d, line 91: Thanks for noticing and updating this!

I think the formatting with `_` between these was meant to match the values in the table below, **however this section is quite out of date anyway**.

Here are some problems with the table as it stands:
- We don't actually support these "Legacy Models" in DXIL.  When this was originally written, we did, but later decided to upgrade the shader model in dxc.exe, and not support legacy values at the API, nor capture them into the DXIL.
- The "DXIL Models" column only lists the very first shader model supported for various shader stages.
- lib_6_1 was added here before we officially supported libraries as a target the runtime would accept at lib_6_3, so while you can technically compile to lib_6_1 and lib_6_2, you must disable validation to do so.

Given all of this, I feel like we should do away with this specific table format.  Perhaps we should change columns to something like this:
```
| Shader Kind | shaderModelName | Minimum major, minor |
+=======================+=======================================+=============+
| Vertex shader (VS)    | vs | 6, 0 |
...
| Dxil library | lib | 6, 3 |
```
... and so on ...
(note: corrected library line here)

What do you think?

coopp (approved):
Looks good.

docs/DXIL.rst
bharadwajy: Thanks @tex3d, for the suggestions regarding the correctness of the content. Happy to update the content.

Are Mesh Shader (MS) and Amplification Shader (AS) supported? If not, do they need to be listed as `unsupported` (for completeness??)?

docs/DXIL.rst
```diff
@@ -88,21 +88,29 @@ The shader model is specified as a named metadata in DXIL::
...
+| Compute shader (CS)   | cs_5_0 (cs_4_0 is mapped onto cs_5_0) | cs_6_0      |
++-----------------------+---------------------------------------+-------------+
+| Shader library        | no support                            | lib_6_1     |
++-----------------------+---------------------------------------+-------------+
+| Mesh shader (MS)      | no support                            | ms_6_5      |
++-----------------------+---------------------------------------+-------------+
+| Amplification shader (AS) | no support                        | as_6_5      |
++-----------------------+---------------------------------------+-------------+
 
 The DXIL verifier ensures that DXIL conforms to the specified shader model.
```
tex3d, line 115: Side note: "verifier" should really be "validator"

Plus, the following paragraph seems odd when you remove the legacy shader models from the picture.
> For shader models prior to 6.0, only the rules applicable to the DXIL representation are valid. For example, the limits on maximum number of resources is honored, but the limits on registers aren't because DXIL does not have a representation for registers.

Maybe we should just delete this?

bharadwajy: OK.

Let me know of your feedback on the accuracy of [this changed version of the section](https://github.com/bharadwajy/DirectXShaderCompiler/blob/main/docs/DXIL.rst#shader-model).

bharadwajy: > Side note: "verifier" should really be "validator"

Is the "verifier" now referred to as "validator" now? If so, should the term be changed throughout the document?

tex3d: That looks good to me!

tex3d: Oh, missed that last comment.  Yes, where it uses "verifier" elsewhere in the document, it should probably say "validator".

I think "verifier" was just the earliest terminology used when drafting this document before the validator actually existed.  I believe we chose "validator" partly because that's what we used prior to DXC, and also because verifier is already an overloaded term used to refer to verifying that an llvm module is well-formed, as well as the clang verifier test which checks front-end diagnostics.

bharadwajy: @tex3d please review #6080 with changes suggested above along with additional cleanup. Thx!

PR #6086, opened 2023-12-01
[Diagnostics] Add diagnostics for incompatible combination of System Values and node Launch Types.

### LINKED ISSUE #5827: Implement compute shader SV diagnostics for node shader wrt. launch modes
We are currently missing parameter diagnostics for compute shader system value inputs for node shader that are restricted per launch mode.  Legal system value semantics given launch mode are described [here](https://microsoft.github.io/DirectX-Specs/d3d/WorkGraphs.html#node-shader-system-values).

**Summary of Rules**

> For Thread launch, no compute shader system values are supported. For Coalescing launch, SV_GroupThreadID and SV_GroupIndex are supported. For Broadcasting launch, all compute shader system values are supported.

Note that `SV_DispatchID` is a different category of system value, as it is a special field of the record.  It is not used, described or diagnosed in the same way as these other compute shader system values.

**Complication for Implementation**

Diagnostics for these will be complicated by the requirement that non-record/node object parameters must be flattened in the general case to get the signature elements corresponding to these system values.  We may want to put off diagnosing this in Sema until we have a path for proper flattened signature construction and diagnostics implemented there.

In the meantime, we need at least to implement `DxilValidation` coverage of these rules first.  After that, we can implement some interim diagnostics in `HLSignatureLower`.

bob80905:
According to this section of the spec:
https://microsoft.github.io/DirectX-Specs/d3d/WorkGraphs.html#node-shader-system-values
There should be diagnostics announcing the unsupported use of certain system value semantics in combination with certain supported launch modes. There were previously no diagnostics for these incompatible combinations. this PR adds the diagnostics into CodeGen. The diagnostics aren't added to Sema (yet) because, according to issue #5827, there is no path for proper flattened signature construction. Parameters won't be flattened at the time that Sema runs, but system values should be accessible by CodeGen.
Fixes #5827

### DIFF
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -1864,6 +1864,50 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
       funcProps->numThreads[1] = 1;
       funcProps->numThreads[2] = 1;
     }
+
+    // Validate Compute Shader system value inputs per launch mode
+    for (ParmVarDecl *param : FD->parameters()) {
+      for (const hlsl::UnusualAnnotation *it : param->getUnusualAnnotations()) {
+        if (it->getKind() == hlsl::UnusualAnnotation::UA_SemanticDecl) {
+          const hlsl::SemanticDecl *sd = cast<hlsl::SemanticDecl>(it);
+          // if the node launch type is Thread, then there are no system values
+          // allowed
+          if (funcProps->Node.LaunchType == DXIL::NodeLaunchType::Thread) {
+            if (sd->SemanticName.equals("SV_GroupThreadID") ||
+                sd->SemanticName.equals("SV_GroupIndex") ||
+                sd->SemanticName.equals("SV_GroupID") ||
+                sd->SemanticName.equals("SV_DispatchThreadID")) {
+              // emit diagnostic
+              unsigned DiagID = Diags.getCustomDiagID(
+                  DiagnosticsEngine::Error,
+                  "Invalid system value semantic '%0' for launchtype '%1'");
+              Diags.Report(param->getLocation(), DiagID)
+                  << sd->SemanticName << "Thread";
+            }
+          }
+
+          // if the node launch type is Coalescing, then
+          // SV_GroupIndex and SV_GroupThreadID are allowed
+          else if (funcProps->Node.LaunchType ==
+                   DXIL::NodeLaunchType::Coalescing) {
+            if (sd->SemanticName.equals("SV_GroupID") ||
+                sd->SemanticName.equals("SV_DispatchThreadID")) {
+              // emit diagnostic
+              unsigned DiagID = Diags.getCustomDiagID(
+                  DiagnosticsEngine::Error,
+                  "Invalid system value semantic '%0' for launchtype '%1'");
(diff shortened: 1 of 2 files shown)


tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1864,6 +1864,50 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
       funcProps->numThreads[1] = 1;
       funcProps->numThreads[2] = 1;
     }
+
+    // Validate Compute Shader system value inputs per launch mode
+    for (ParmVarDecl *param : FD->parameters()) {
```
tex3d, line 1869: Checking this properly is more complicated.  You can supply a struct as input with these fields.  The struct could have these fields in nested struct or base class.  I think we need a more general function that walks entry function parameters, identifying the inputs and outputs that they declare, with their semantics, types and locations, for general use in entry function validation, rather than ad-hoc checks that don't catch all possible cases.

tex3d (changes requested):
While this catches basic cases for Node shaders, it would miss other cases.  I don't think it's the right approach to add these diagnostics ad-hoc like this.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tex3d: I think the right approach would be to write a few CollectSignatureElements function overloads to solve this.

Here's a rough outline:

- Gathers into an info structure containing a list of elements for each SigPoint.  Elements have:
    - a component type, rows, and cols
    - the semantic name, and a starting semantic index.  Maybe we can capture the indexing pattern later, but don't need to yet.
    - interpolation modifier, and source location, if any
    - a source location for the parameter or field associated with this signature element
    - a source location for the associated semantic annotation (may not be on this field)
- First level:
    - input: entry function and ShaderKind.
    - Processes parameters, return type and function semantic.
- Second level recursive:
    - input: type, semantic name and index (if any), and interpolation modifiers (if any), array information (if any) - note that arrays complicate things, so details need to be worked out here.
    - Collects array dimensions, walks struct bases and fields with recursion
    - Outer semantic assignments override inner ones, with semantic indexing based on the flattening shown in [this example](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst#signature-record-metadata "https://github.com/microsoft/directxshadercompiler/blob/main/docs/dxil.rst#signature-record-metadata").
- During traversal, diagnostics can be emitted when there's a clear error based on the hctdb-specified element definitions.
- After traversal, additional rules can be diagnosed from the collected elements, such as the launch mode specific constraints.

llvm-beanz: Why are we doing this in CodeGen instead of Sema?

tex3d: Yes, this should be in Sema instead of CodeGen as well.  I think when I looked at this code and wrote the recommended approach, I was imagining it happening in Sema, since it was operating on the clang types.  It makes virtually no difference to what needs to be done, but Sema is the right place to do it.

bob80905:
The implementation of the semantic index assignment spec here is needed as a prerequisite before this PR can be considered.
https://github.com/microsoft/hlsl-specs/pull/156/files

PR #6088, opened 2023-12-01
[NFC] Remove misleading function

bob80905:
The IsRay() function defined in the ShaderModel class can be misleading. It is included in a set of functions that are used to test whether or not the shader in question was targeted to a specific shader stage. However, it is impossible for a shader to be targeted to, say, the "raygeneration" stage, or any other stages associated with the kinds that the IsRay() function checks for. Like the "node" shader kind, a shader can only have these shader kinds if the associated attribute is found on the function declaration. To prevent confusion, and to keep the Is*() functions inside the ShaderModel class restricted to only those targetable shader stages, the IsRay() function can be removed. This PR comes as a response to #6008

### DIFF
--- a/lib/DXIL/DxilShaderModel.cpp
+++ b/lib/DXIL/DxilShaderModel.cpp
@@ -76,7 +76,8 @@ bool ShaderModel::IsValidForDxil() const {
 
 bool ShaderModel::IsValidForModule() const {
   // Ray tracing shader model should only be used on functions in a lib
-  return IsValid() && !IsRay();
+  return IsValid() &&
+         !(m_Kind >= Kind::RayGeneration && m_Kind <= Kind::Callable);
 }
 
 const ShaderModel *ShaderModel::Get(Kind Kind, unsigned Major, unsigned Minor) {
--- a/lib/DXIL/DxilShaderModel.cpp
+++ b/lib/DXIL/DxilShaderModel.cpp
@@ -74,12 +74,6 @@ bool ShaderModel::IsValidForDxil() const {
   return false;
 }
 
-bool ShaderModel::IsValidForModule() const {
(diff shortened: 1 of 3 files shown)


lib/DXIL/DxilShaderModel.cpp
```diff
@@ -76,7 +76,8 @@ bool ShaderModel::IsValidForDxil() const {
 
 bool ShaderModel::IsValidForModule() const {
   // Ray tracing shader model should only be used on functions in a lib
-  return IsValid() && !IsRay();
+  return IsValid() &&
+         !(m_Kind >= Kind::RayGeneration && m_Kind <= Kind::Callable);
```
tex3d, line 80: Technically, this exposed a hole.  Node is not a valid kind for module either, so we should update the expression to reflect that, and update the comment, since it's not just raytracing.

tex3d: Actually, there's more.  Notice how `IsValid()` is asserting that it's one of the supported kinds?  `IsValidForModule()` calls `IsValid()`.  If it was ray or node, it would assert (on debug) in `IsValid()`.

It looks like the only way you could have a kind outside the valid or the explicit `Kind::Invalid`, is to manually construct the ShaderModel, but that constructor is private.  So, I think the assert is sufficient, and there's no need for additional checks in this method beyond the current IsValid().

Then, if `IsValidForModule()` is identical to `IsValid()`, why have it?  There's no reason to keep it.  The only use of this function is in an assert in `DxilModule::SetShaderModel`.  Since the prior assert already calls `IsValidForDxil()`, which calls `IsValid()` as well, there's no need for the assert that calls `IsValidForModule()`.

In short, we should be able to remove this method and the assert where it's called.

tex3d (approved):
LGTM!

PR #6089, opened 2023-12-01
Check WaveMatrix support; locally declare feature struct for now

tex3d:
WaveMatrix wasn't checking WaveMMATier before running tests (assuming it was always supported for SM 6.8). This change fixes this.

D3D12_SDK_VERSION check for WAVE_MMA feature structure and enum definitions assumed they will be defined in SDK version 613. This isn't accurate, so this block of local definitions will always be enabled until we have the correct version in the future.

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -758,6 +758,12 @@ public:
       return nullptr;
     }
 
+    if (!DoesDeviceSupportWaveMatrix(pDevice)) {
+      LogCommentFmt(L"WaveMatrix not supported on this device.");
+      WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+      return nullptr;
+    }
+
     CComPtr<IStream> pStream;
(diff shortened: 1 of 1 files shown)


dmpots (approved):
LGTM

python3kgae (approved):

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -758,6 +758,12 @@ class ExecutionTest {
       return nullptr;
     }
 
+    if (!DoesDeviceSupportWaveMatrix(pDevice)) {
+      LogCommentFmt(L"WaveMatrix not supported on this device.");
+      WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+      return nullptr;
```
amarpMSFT, line 764: Does the caller gracefully handle not getting a device, even if a skip was logged here? (e.g. exit test)

From my web view not easy for me to quickly check myself I don't think.

amarpMSFT (approved):

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
tex3d: Yes.

PR #6090, opened 2023-12-02
[Sema] Change data type in MaxDispatchGrid validation calculation to uint64_t to prevent overflow

### LINKED ISSUE #5988: NodeDispatchGrid diagnostic failure
**Description**
The product of the components of the NodeDispatchGrid or  NodeMaxDispatchGrid attributes may not exceed 16,777,215 (2^24-1)     but this is not diagnosed correctly in cases where the product overflowed 32 bits.

This can be fixed simply by increasing the width of the calculation of the product to 64 bits (that can't overflow without at least one of the components exceeding 65,535 which is diagnosed earlier).

**Steps to Reproduce**
[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeDispatchGrid(65535, 65535, 65535)] // expected-error {{'NodeDispatchGrid' X * Y * Z product may not exceed 16,777,215 (2^24-1)}}
[NumThreads(1, 1, 1)]
void myNode() { }

**Actual Behavior**
Currently no error diagnostic is generated for cases where the product overflows 32 bits.

**Environment**
Using staging-sm-6.8 branch

bob80905:
Previously, in order to determine whether the three arguments provided to the MaxDispatchGrid attribute exceeded the maximum product value (which is 16,777,215 (2^24-1)), each of the three components would be multiplied together and compared to this maximum. However, the three arguments would be read in as unsigned ints. In certain conditions, the multiplication would cause an overflow, which would programmatically result in a number less than the maximum product limit, despite mathematically being above this limit. The diagnostic should trigger in this case, but it doesn't.
This PR changes the data type of the three arguments to unsigned long long, which can hold the maximum possible product that can be achieved without triggering the per-multiplicand limit of 65535, and won't overflow in this case. Now, the compiler will trigger this diagnostic. Regression test has been added.
Fixes #5988


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -13134,7 +13134,7 @@ void Sema::DiagnoseGloballyCoherentMismatch(const Expr *SrcExpr,
 
 void ValidateDispatchGridValues(DiagnosticsEngine &Diags,
                                 const AttributeList &A, Attr *declAttr) {
-  unsigned x = 1, y = 1, z = 1;
+  unsigned long long int x = 1, y = 1, z = 1;
```
tex3d, line 13137: Should we use `uint64_t` instead?  I think that's more commonly used, isn't it?
I don't think it's necessary for the types of x,y,z to be changed, since this is the type retrieved.  Instead, we could use a `uint64_t product = (uint64_t)x * (uint64_t)y * (uint64_t)z;` later, right?

bob80905: Yes, that's logical.

tex3d (approved):
LGTM.

python3kgae (approved):

PR #6091, opened 2023-12-02
PIX: Fix scope ascension for variable lookup

jeffnn:
In DxcPixLiveVariables.cpp, UniqueScopeForInlinedFunctions::AscendScopeHierarchy tries to keep track of a variable's scope and the scope at which its containing function was inlined, if any.

This scope duo has to be the same in two places for things to work out: the scopes inferred from dbg.declare, and then again when the scope hierarchy is ascended in order to discover all variables accessible from a given point in the program.

AscendScopeHierarchy was forgetting to update the inlined part of the scope, resulting in, for example, PIX's debugger losing track of variables outside of, for example, a for loop, even though the code can see those variables at that point.

Just setting the inlined scope to be the same as the function scope fixes this problem.

(In addition to these new tests, existing PIX tests exercise this situation happily)

### DIFF
--- a/tools/clang/unittests/HLSL/PixDiaTest.cpp
+++ b/tools/clang/unittests/HLSL/PixDiaTest.cpp
@@ -3050,7 +3050,8 @@ TEST_F(PixDiaTest, DxcPixDxilDebugInfo_VariableScopes_ForScopes) {
   DWORD instructionOffset =
       AdvanceUntilFunctionEntered(dxilDebugger, 0, L"CSMain");
 
-  instructionOffset = Labels->FindInstructionOffsetForLabel(L"CheckVariableExistsHere");
+  instructionOffset =
+      Labels->FindInstructionOffsetForLabel(L"CheckVariableExistsHere");
   CheckVariableExistsAtThisInstruction(dxilDebugger, instructionOffset,
                                        L"zero");
 
@@ -3195,14 +3196,12 @@ void CSMain()
       Labels->FindInstructionOffsetForLabel(L"inside member fn");
   CheckVariableDoesNOTExistsAtThisInstruction(dxilDebugger, instructionOffset,
                                               L"s");
-  CheckVariableExistsAtThisInstruction(dxilDebugger, instructionOffset,
-                                       L"q");
+  CheckVariableExistsAtThisInstruction(dxilDebugger, instructionOffset, L"q");
 
   instructionOffset = Labels->FindInstructionOffsetForLabel(L"Stop here");
   CheckVariableDoesNOTExistsAtThisInstruction(dxilDebugger, instructionOffset,
                                               L"i");
-  CheckVariableExistsAtThisInstruction(dxilDebugger, instructionOffset,
-      L"j");
+  CheckVariableExistsAtThisInstruction(dxilDebugger, instructionOffset, L"j");
 }
 
 #endif
--- a/tools/clang/unittests/HLSL/PixDiaTest.cpp
+++ b/tools/clang/unittests/HLSL/PixDiaTest.cpp
@@ -129,6 +129,11 @@ static HRESULT UnAliasType(IDxcPixType *MaybeAlias,
   return S_OK;
 }
 
+struct DebuggerInterfaces {
+  CComPtr<IDxcPixDxilDebugInfo> debugInfo;
+  CComPtr<IDxcPixCompilationInfo> compilationInfo;
+};
+
 class PixDiaTest {
 public:
   BEGIN_TEST_CLASS(PixDiaTest)
@@ -177,8 +182,14 @@ public:
   TEST_METHOD(DxcPixDxilDebugInfo_UnnamedField)
   TEST_METHOD(DxcPixDxilDebugInfo_SubProgramsInNamespaces)
   TEST_METHOD(DxcPixDxilDebugInfo_SubPrograms)
-  TEST_METHOD(DxcPixDxilDebugInfo_InlinedFunctions_TwiceInlinedFunctions)
-  TEST_METHOD(DxcPixDxilDebugInfo_InlinedFunctions_CalledTwiceInSameCaller)
+  TEST_METHOD(
+      DxcPixDxilDebugInfo_VariableScopes_InlinedFunctions_TwiceInlinedFunctions)
+  TEST_METHOD(
+      DxcPixDxilDebugInfo_VariableScopes_InlinedFunctions_CalledTwiceInSameCaller)
+  TEST_METHOD(DxcPixDxilDebugInfo_VariableScopes_ForScopes)
+  TEST_METHOD(DxcPixDxilDebugInfo_VariableScopes_ScopeBraces)
+  TEST_METHOD(DxcPixDxilDebugInfo_VariableScopes_Function)
+  TEST_METHOD(DxcPixDxilDebugInfo_VariableScopes_Member)
 
   dxc::DxcDllSupport m_dllSupport;
   VersionSupportInfo m_ver;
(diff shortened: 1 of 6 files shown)


tools/clang/unittests/HLSL/PixDiaTest.cpp
```diff
@@ -2777,10 +2790,71 @@ static DWORD AdvanceUntilFunctionEntered(IDxcPixDxilDebugInfo *dxilDebugger,
...
+      DWORD sourceOrdinal = 0;
+      CComBSTR fileName;
+      CComBSTR fileContent;
+      while (SUCCEEDED(debuggerInterfaces.compilationInfo->GetSourceFile(
+          sourceOrdinal++, &fileName, &fileContent))) {
+        if (fileName == CurrentNameAndNumber.FileName) {
+          auto lines = strtok(std::wstring(fileContent), L"\n");
+          // VERIFY_IS_TRUE(CurrentNameAndNumber.LineNumber >= 1);
+          // VERIFY_IS_TRUE(CurrentNameAndNumber.LineNumber - 1 <
+          //               static_cast<DWORD>(lines.size()));
```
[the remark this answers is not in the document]
jeffnn: Yep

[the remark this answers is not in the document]
jeffnn: (Was commented out during development-  forgot to reenable)

tools/clang/unittests/HLSL/PixDiaTest.cpp
```diff
@@ -2777,10 +2790,71 @@ static DWORD AdvanceUntilFunctionEntered(IDxcPixDxilDebugInfo *dxilDebugger,
...
+          std::wostringstream s;
+          s << "Seeking line number " << CurrentNameAndNumber.LineNumber
+            << " at Instruction offsset " << instructionOffset
+            << ": Checking line " << lines[CurrentNameAndNumber.LineNumber - 1]
+            << "\n";
+          OutputDebugStringW(s.str().c_str());
```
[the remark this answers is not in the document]
jeffnn: Forgot to delete!

tools/clang/unittests/HLSL/PixDiaTest.cpp
```diff
@@ -2777,10 +2790,71 @@ static DWORD AdvanceUntilFunctionEntered(IDxcPixDxilDebugInfo *dxilDebugger,
...
+                                      DWORD instructionOffset,
+                                      wchar_t const *lineContents) {
+  for (;;) {
+    auto CurrentNameAndNumber = GetFileNameAndLineNumberForInstructionOffset(
+        debuggerInterfaces, instructionOffset);
+    if (CurrentNameAndNumber.FileName != nullptr) {
+      DWORD sourceOrdinal = 0;
+      CComBSTR fileName;
+      CComBSTR fileContent;
+      while (SUCCEEDED(debuggerInterfaces.compilationInfo->GetSourceFile(
```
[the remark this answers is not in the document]
jeffnn: Implemented as suggested. Thanks!

tools/clang/unittests/HLSL/PixTestUtils.cpp
```diff
@@ -274,6 +274,57 @@ PassOutput RunAnnotationPasses(dxc::DxcDllSupport &dllSupport, IDxcBlob *dxil,
...
+    while (SUCCEEDED(debuggerInterfaces.compilationInfo->GetSourceFile(
+        SourceFileOrdinal, &fileName, &fileContent))) {
+      auto lines = strtok(std::wstring(fileContent), L"\n");
+      for (size_t line = 0; line < lines.size(); ++line) {
+        const wchar_t DebugLocLabel[] = L"debug-loc(";
+        auto DebugLocPos = lines[line].find(DebugLocLabel);
+        if (DebugLocPos != std::wstring::npos) {
+          auto StartLabelPos = DebugLocPos + (_countof(DebugLocLabel) - 1);
+          auto CloseDebugLocPos = lines[line].find(L")", StartLabelPos);
+          VERIFY_ARE_NOT_EQUAL(CloseDebugLocPos, std::string::npos);
```
tex3d, line 295: I wonder about the utility of using VERIFY macros deep in test utility code, especially within loops and such.  These should only fail on a malformed test or an internal test bug.  We certainly do use it for some things of this class (probably too much), but these add to the verification noise, since it's not what the test case is focused on testing.  Logging a failure with some detail is important, but when using VERIFY macros, every passing verification is logged for use if something fails later in the test, or if verbose logging is enabled.

tools/clang/unittests/HLSL/PixTestUtils.cpp
```diff
@@ -274,6 +274,57 @@ PassOutput RunAnnotationPasses(dxc::DxcDllSupport &dllSupport, IDxcBlob *dxil,
...
+        }
+      }
+      SourceFileOrdinal++;
+      fileName = nullptr;
+      fileContent = nullptr;
+    }
+  }
+
+  virtual DWORD FindInstructionOffsetForLabel(const wchar_t *label) override {
+    return m_labelToInstructionOffset[label];
```
tex3d, line 319: I think this is a good place for a VERIFY macro to ensure that the requested label is in the map first.  Otherwise, this will just add something to the map and return `0`.

tools/clang/unittests/HLSL/PixTestUtils.cpp
jeffnn: Sure. It was pretty noisy.

tools/clang/unittests/HLSL/PixTestUtils.cpp
jeffnn: Good point. (The caller would then fail, but this is the right place to check.)

tools/clang/unittests/HLSL/PixTestUtils.cpp
```diff
@@ -274,6 +274,57 @@ PassOutput RunAnnotationPasses(dxc::DxcDllSupport &dllSupport, IDxcBlob *dxil,
...
+                               ->InstructionOffsetsFromSourceLocation(
+                                   fileName, line + 1, 0, &InstructionOffsets));
+          auto InstructionOffsetCount = InstructionOffsets->GetCount();
+          VERIFY_IS_TRUE(InstructionOffsetCount > 0);
+          // No duplicate labels:
+          VERIFY_IS_TRUE(m_labelToInstructionOffset.find(Label) ==
+                         m_labelToInstructionOffset.end());
+          // Just the last offset is sufficient:
+          m_labelToInstructionOffset[Label] =
+              InstructionOffsets->GetOffsetByIndex(InstructionOffsetCount - 1);
```
tex3d, line 316: Isn't source location to instruction offset a many-to-many mapping, or at least a one to many?  Shouldn't this capture all the offsets associated with the source location?  Why not just capture the `IDxcPixDxilInstructionOffsets`?

If capturing only one instruction offset, doesn't that create a problem if you wanted to step and stop at the first offset associated with the location (in this case you're only storing the last)?  For instance, you could have an unrolled loop, and you want to stop on the first instruction associated with that location at each iteration.

jeffnn: Should such a need arise, it seems an easy thing to add. A speculative test-support feature doesn't seem worth it for now.

tex3d (approved):
LGTM!

tools/clang/unittests/HLSL/PixTestUtils.cpp
```diff
@@ -274,6 +274,66 @@ PassOutput RunAnnotationPasses(dxc::DxcDllSupport &dllSupport, IDxcBlob *dxil,
...
+            VERIFY_FAIL(L"Duplicate label found!");
+          }
+          // Just the last offset is sufficient:
+          m_labelToInstructionOffset[Label] =
+              InstructionOffsets->GetOffsetByIndex(InstructionOffsetCount - 1);
+        }
+      }
+      SourceFileOrdinal++;
+      fileName.Empty();
+      fileContent.Empty();
```
tex3d, line 321: I think you'll need to add the `Empty()` method to the CComBSTR implementation in `WinAdapter.h` for this to work.  That implementation is super minimal, and doesn't seem to have a way to clear the string other than the destructor.
See here:
https://github.com/microsoft/DirectXShaderCompiler/blob/main/include/dxc/WinAdapter.h#L1000

python3kgae (approved):

PR #6095, opened 2023-12-04
Revert "Add /bigobj flag on big spirvtest object (#3858)"

### LINKED ISSUE #3867: [SPIR-V] Break up CodeGenSpirvTest.cpp into multiple files
As we discussed in #3858, the big object file built from `CodeGenSpirvTest.cpp` matters.
I guess the complicated macros of the googletest framework significantly amplifies the object file size.

We have to consider splitting the test file into multiple smaller files.

sudonatalie:
This reverts commit 36bfa748784966e1406f8c1823e379bb97f4dc81.

CodeGenSpirvTest.cpp has been cut down significantly from its original >10,000 line size, so this flag should no longer be necessary.

Fixes #3867

### DIFF
--- a/tools/clang/unittests/SPIRV/CMakeLists.txt
+++ b/tools/clang/unittests/SPIRV/CMakeLists.txt
@@ -51,10 +51,6 @@ configure_file(
     ${CMAKE_CURRENT_BINARY_DIR}/SpirvTestOptions.h
     )
 
-if(WIN32)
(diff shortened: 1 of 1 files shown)


Keenuts (approved):
CI seems to pass, so I guess that's fine.

cassiebeckley (approved):

Issue #5762, opened 2023-09-21
Add docs for building on Linux/macOS


ApoorvaJ:
When I looked for how to build DXC on MacOS, I found a few scarce resources:
- [This link](https://github.com/google/DirectXShaderCompiler/blob/linux/docs/DxcOnUnix.rst) from an archived fork
- [ShaderConductor](https://github.com/microsoft/ShaderConductor), which seems to be defunct, with broken pre-built binaries.

I eventually figured it out, and it worked quite well. So would it be possible to add something along these lines to [this page](https://github.com/microsoft/DirectXShaderCompiler/wiki/Building-Sources)?

## How to build on MacOS/Linux

`mkdir build && cd build`

For debug mode, run
`cmake -C ../cmake/caches/PredefinedParams.cmake ..`
For release mode, run
`cmake -C ../cmake/caches/PredefinedParams.cmake -DCMAKE_BUILD_TYPE=Release ..`

Build using `make -j<num of worker threads>`

There should now be bin and lib folders in your current directory.

arnamoy10:
Thanks for the info, very useful.  Just wanted to add something for people who tried to build this just after cloning the repo.  You may have to run:
`git submodule init;
git submodule update`

Else you might get error like:  `external/SPIRV-Headers Does not contain CMakeLists.txt`

gongminmin:
One trick I always use is checking the CI script. Since the CI of dxc builds on macOS and Linux, it must contains instructions of how to do it. So open [appveyor.yml](https://github.com/microsoft/DirectXShaderCompiler/blob/main/appveyor.yml), it's just right there.

gl-yziquel:
> Thanks for the info, very useful. Just wanted to add something for people who tried to build this just after cloning the repo. You may have to run: `git submodule init; git submodule update`
> 
> Else you might get error like: `external/SPIRV-Headers Does not contain CMakeLists.txt`

Yes. Please add this to the documentation.

gl-yziquel:
The documentation would need some love.

After checking out the submodules as indicated by @gongminmin I get the following build failure:

It complains about a missing `CHROMIUM_BUILDTOOLS_PATH` environment variable. This is frustrating as I already have a built chromium from source on my box, and would thus need instructions as to where this environment variable should point to.

```console
minime@virtucon ~/h/c/D/build (main)> ninja
[8/1432] Building DxilCounters.h...
[...]

PR #6102, opened 2023-12-10
Documenting submodule initialisation in experimental build. This avoid a cmake error about missing `external/SPRIV-Headers`.

gl-yziquel:


docs/BuildingAndTestingDXC.rst
```diff
@@ -93,6 +93,15 @@ You may substitute ``Ninja`` for ``Unix Makefiles`` to generate a makefile
 build. After generation completes you can run ``ninja`` or ``make`` as
 appropriate.
 
+To avoid some build failures such as missing `external/SPIRV-Headers`, make
+sure that you initialise the git submodules before running `ninja` or `make`:
+
+.. code-block:: sh
+
+  git submodule init
+  git submodule update
```
s-perron, line 102: This seems out of place. I would expect this to apply to all of the cmake commands above as well. However, it might be worth while mentioning the submodule somewhere. @llvm-beanz Where would you suggest we add the instruction to check out with code with the submodules?

gl-yziquel: I indeed felt it was out of place. But it is in the place where **I** needed the information and that would have saved me quite a lot of time.

Please do feel free to put that information anywhere else.

llvm-beanz: Separately I put together a PR cleaning up the documentation for building DXC in relation to #5762 which also added information about the submodules (see #6185).

s-perron:
This is fixed in another PR

gl-yziquel:
> This is fixed in another PR

cool.

PR #6108, opened 2023-12-14
[Validation] Add validation and tests for invalid system values in incompatible node launch type

### LINKED ISSUE #6104: [Validation] Validate SV values in combination with node launch types in DXIL
There are a couple of rules that aren't validated in DxilValidaiton.cpp, that should be:
`For Thread launch, no compute shader system values are supported. For Coalescing launch, SV_GroupThreadID and SV_GroupIndex are supported. For Broadcasting launch, all compute shader system values are supported.`

We need to implement DxilValidation coverage of these rules. After that, we can solve #5827

bob80905:
#5827 first requires validation at the DXIL level. The validation will check for functions that have incompatible node launch types, and intrinsics that represent certain system values.
This PR implements this validation and adds some tests that exercise this validation.
Fixes #6104


utils/hct/hctdb.py
```diff
@@ -7552,6 +7552,11 @@ def build_valrules(self):
         self.add_valrule(
             "Instr.AtomicIntrinNonUAV", "Non-UAV destination to atomic intrinsic."
         )
+        self.add_valrule_msg(
+            "Instr.InvalidSVInFunction",
```
python3kgae, line 7556: Cannot see node in the name.
Maybe Instr.InvalidSVInNodeFunction?

lib/HLSL/DxilValidation.cpp
```diff
@@ -3494,7 +3626,7 @@ static void ValidateFunction(Function &F, ValidationContext &ValCtx) {
           }
         }
         break;
-      }
+      }
```
python3kgae, line 3629: Useless space?

lib/HLSL/DxilValidation.cpp
```diff
@@ -2345,7 +2345,83 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
           CI, ValidationRule::SmOpcodeInInvalidFunction,
           {"CreateHandle", "Shader model 6.5 and below"});
     }
+  } break;
+    
+  // this intrinsic is generated when the SV_DispatchThreadID system value is used
+  case DXIL::OpCode::ThreadId: {  
+    // get the node launch type
+    DxilModule &DM = ValCtx.DxilMod;
+    if (!DM.HasDxilEntryProps(F)) {
```
tex3d, line 2354: First, we need to check whether `shaderKind == DXIL::ShaderKind::Node`, since these intrinsics are used for other shader types as well.

Next, `DXIL::NodeLaunchType` should also be gathered at the top of this function when appropriate, since it will be useful for multiple verifications.  Also, at some point, we will also need to update the validator to handle noinline functions potentially called from entry points (a current gap), where this can be inherited through the call graph, rather than looked up directly on this function.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2345,7 +2345,83 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
...
+      break;
+    }
+
+    DxilEntryProps &entryProps = DM.GetDxilEntryProps(F);
+    DXIL::NodeLaunchType NLT = entryProps.props.Node.LaunchType;
+    if (NLT == DXIL::NodeLaunchType::Thread || NLT == DXIL::NodeLaunchType::Coalescing) {
```
tex3d, line 2360: I feel this approach is a bit fragile, since it will permit any new launch type if we forget to update the code here.  If we checked for the specific launch types where it's legal instead, that would force us to update the code if necessary in the future, instead of opening up a validation gap.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2345,7 +2345,83 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
...
+          {"ThreadId", NLT ==
+              DXIL::NodeLaunchType::Thread
+          ? "Thread"
+          : "Coalescing"});
+		}
+	} break;                
+  // this intrinsic is generated when the SV_GroupId
+  // system value is used
+  
+	case DXIL::OpCode::GroupId: {
```
tex3d, line 2371: nits: tab used for indentation here, and the comment above looks a little out of place.

lib/HLSL/DxilValidation.cpp
tex3d, line 3629: Spurious end-of-line whitespace added here.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2345,7 +2345,83 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
           CI, ValidationRule::SmOpcodeInInvalidFunction,
           {"CreateHandle", "Shader model 6.5 and below"});
     }
+  } break;
+    
+  // this intrinsic is generated when the SV_DispatchThreadID system value is used
+  case DXIL::OpCode::ThreadId: {
```
tex3d, line 2351: Using a minimal comment (like `// SV_DispatchThreadID`) at the end of the case statement line would probably suffice for each of these.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3463,6 +3539,62 @@ static void ValidateNodeInputRecord(Function *F, ValidationContext &ValCtx) {
   }
 }
 
+static void ValidateFunctionArgs(Function& F, unsigned numUDTShaderArgs, DXIL::ShaderKind shaderKind, ValidationContext& ValCtx) {
```
tex3d, line 3542: I don't think the PR should include unrelated refactor work, should it?

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_broadcast.ll
```diff
@@ -0,0 +1,50 @@
...
+; Function Attrs: nounwind readnone
+declare i32 @dx.op.groupId.i32(i32, i32) #1
+
+; Function Attrs: nounwind readnone
+declare i32 @dx.op.threadIdInGroup.i32(i32, i32) #0
+
+; Function Attrs: nounwind readnone
+declare i32 @dx.op.flattenedThreadIdInGroup.i32(i32) #0
+
+; this file comes from compiling launch_types.hlsl with:
```
tex3d, line 50: This is incomplete, and I don't think there's a `launch_types.hlsl` currently in the project.

Also, can we move lines like this to the top so generated/modified code follows any informative comments like this?

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_broadcast.ll
```diff
@@ -0,0 +1,50 @@
+; RUN: %dxv %s 
+
+; There should be no errors!
```
tex3d, line 3: If we have a succeeding case for broadcasting, it seems we should have ones for coalescing and thread launch as well.

However, this could be considered covered by tests specifically for these intrinsics in those launch modes that should exist elsewhere, rather needing validation tests which don't fail validation here.

tex3d (changes requested):

lib/HLSL/DxilValidation.cpp
```diff
@@ -2134,6 +2134,24 @@ static void ValidateBarrierFlagArg(ValidationContext &ValCtx, CallInst *CI,
...
+  case DXIL::NodeLaunchType::Thread:
+    return "Thread";
+  case DXIL::NodeLaunchType::Invalid:
+    return "Invalid";
+  case DXIL::NodeLaunchType::LastEntry:
+    return "LastEntry";
+	
+	default:
+		return "Unknown";
+	}
```
tex3d, line 2152: I believe all of these could be considered `"<invalid>"` which should never be seen unless some other part of validation is not performed.  Also, there's weird line spacing and tabbed indentation here.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2156,6 +2174,15 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
   // Is called from a library function
   bool isLibFunc = shaderKind == DXIL::ShaderKind::Library;
 
+  // get the node launch type
+  DxilModule &DM = ValCtx.DxilMod;
+  DXIL::NodeLaunchType NLT = DXIL::NodeLaunchType::Invalid;
+
+  if (DM.HasDxilEntryProps(F)) {
+    DxilEntryProps &entryProps = DM.GetDxilEntryProps(F);
```
tex3d, line 2182: This can be merged with earlier code that gets the properties under this branch:
https://github.com/microsoft/DirectXShaderCompiler/pull/6108/files#diff-3d74ec29820bb5a94491fb4d46c80d37c89b0e72c55df5394fea58236c1f5300R2163

lib/HLSL/DxilValidation.cpp
```diff
@@ -2156,6 +2174,15 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
   // Is called from a library function
   bool isLibFunc = shaderKind == DXIL::ShaderKind::Library;
 
+  // get the node launch type
+  DxilModule &DM = ValCtx.DxilMod;
+  DXIL::NodeLaunchType NLT = DXIL::NodeLaunchType::Invalid;
```
tex3d, line 2179: Given the lifetime of `NLT` now, I think it needs a longer name, like `nodeLaunchType`.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2333,7 +2360,7 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
             DXIL::ResourceClass::UAV)
       ValCtx.EmitInstrError(CI, ValidationRule::InstrAtomicIntrinNonUAV);
   } break;
-  case DXIL::OpCode::CreateHandle:
+  case DXIL::OpCode::CreateHandle: {
```
tex3d, line 2363: Was there any particular reason for this change adding the extra block scope?

lib/HLSL/DxilValidation.cpp
```diff
@@ -2345,7 +2372,65 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
...
+  } break;
+      
+  case DXIL::OpCode::ThreadId: { // SV_DispatchThreadID
+    if (shaderKind != DXIL::ShaderKind::Node) {
+      break;
+    }
+
+    if (NLT == DXIL::NodeLaunchType::Broadcasting) {
+      // good
+    } else {
```
tex3d, line 2384: It's ok, but why this pattern instead of `!= DXIL::NodeLaunchType::Broadcasting`?

lib/HLSL/DxilValidation.cpp
```diff
@@ -3510,8 +3595,9 @@ static void ValidateFunction(Function &F, ValidationContext &ValCtx) {
     // Shader functions should return void.
     if (isShader && !F.getReturnType()->isVoidTy())
       ValCtx.EmitFnFormatError(&F, ValidationRule::DeclShaderReturnVoid,
-                               {F.getName()});
+                               {F.getName()});
```
tex3d, line 3598: nit: trailing whitespace.  I would have expected the clang-format checker to catch this, but for some reason that didn't appear to run against this PR.  I wonder why?

lib/HLSL/DxilValidation.cpp
```diff
@@ -3510,8 +3595,9 @@ static void ValidateFunction(Function &F, ValidationContext &ValCtx) {
     // Shader functions should return void.
     if (isShader && !F.getReturnType()->isVoidTy())
       ValCtx.EmitFnFormatError(&F, ValidationRule::DeclShaderReturnVoid,
-                               {F.getName()});
+                               {F.getName()});    
 
+    // Validate parameter type.
```
tex3d, line 3600: I think the move of this comment is left over from reverting an unrelated refactor here.  It's probably best to leave it alone in this PR to reduce unrelated, unnecessary churn.

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_broadcast.ll
tex3d: I'm not sure what the resolution to this comment was.  I don't think this test should be here.  If anything, just make sure we have tests elsewhere that check the appropriate system value intrinsics are called when the system values are used (not as a validation test).

lib/HLSL/DxilValidation.cpp
bob80905: I found it easier to read to do the pattern shown above. If I instead specified the failure case logically, we would need to write something like ` ! (NLT == coalescing || NLT == thread)`, just a tad bit more complicated.

lib/HLSL/DxilValidation.cpp
bob80905: I think the link may be broken?

tex3d: It links to above code in this file, which is collapsed by default.  Just look around line 2163 for `if (ValCtx.DxilMod.HasDxilFunctionProps(F))`.  That branch can be merged with this.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2134,17 +2134,38 @@ static void ValidateBarrierFlagArg(ValidationContext &ValCtx, CallInst *CI,
...
+  case DXIL::NodeLaunchType::Coalescing:
+		return "Coalescing";
+  case DXIL::NodeLaunchType::Thread:
+    return "Thread";
+  case DXIL::NodeLaunchType::Invalid:
+  case DXIL::NodeLaunchType::LastEntry:
+    return "Invalid";
+
+  default:
+    return "Invalid";
```
bogner, line 2150: Since you're covering all of the cases, it'd be better to skip the `default` label so that `-Wswitch` can warn if any enum values are missed (ie, if they're added). You'll need to add an `llvm_unreachable` call outside of the switch to quiet warnings about not all control paths returning a value on some compilers.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2156,6 +2177,8 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
   // Is called from a library function
   bool isLibFunc = shaderKind == DXIL::ShaderKind::Library;
 
+  // get the node launch type
+
```
bogner, line 2181: Is this comment leftover from a previous iteration?

lib/HLSL/DxilValidation.cpp
```diff
@@ -2134,17 +2134,38 @@ static void ValidateBarrierFlagArg(ValidationContext &ValCtx, CallInst *CI,
   }
 }
 
+std::string GetLaunchTypeStr(DXIL::NodeLaunchType LT) {
+  switch (LT) {
+	case DXIL::NodeLaunchType::Broadcasting:
+		return "Broadcasting";
+  case DXIL::NodeLaunchType::Coalescing:
+		return "Coalescing";
+  case DXIL::NodeLaunchType::Thread:
```
bogner, line 2143: The formatting looks a little messed up here, please check.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2134,17 +2134,38 @@ static void ValidateBarrierFlagArg(ValidationContext &ValCtx, CallInst *CI,
...
       pSM ? pSM->GetKind() : DXIL::ShaderKind::Invalid;
   llvm::Function *F = CI->getParent()->getParent();
+  DXIL::NodeLaunchType nodeLaunchType = DXIL::NodeLaunchType::Invalid;
   if (DXIL::ShaderKind::Library == shaderKind) {
-    if (ValCtx.DxilMod.HasDxilFunctionProps(F))
+    if (ValCtx.DxilMod.HasDxilFunctionProps(F)) {
+      DxilModule &DM = ValCtx.DxilMod;
+      DxilEntryProps &entryProps = DM.GetDxilEntryProps(F);
+      nodeLaunchType = entryProps.props.Node.LaunchType;
       shaderKind = ValCtx.DxilMod.GetDxilFunctionProps(F).shaderKind;
```
bogner, line 2167: Doesn't matter too much in practice, but it feels weirdly inconsistent to capture `ValCtx.DxilMod` in `DM` for calculating `entryProps` but then use `ValCtx.DxilMod` directly again when getting `shaderKind`

lib/HLSL/DxilValidation.cpp
bogner: I also found this pattern confusing. Maybe it would be clearest if you use break instead of an empty if block?
```c++
    if (nodeLaunchType == DXIL::NodeLaunchType::Broadcasting)
      break;
    ValCtx.EmitInstrFormatError(...);
```

lib/HLSL/DxilValidation.cpp
```diff
@@ -2346,6 +2369,67 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
...
+
+    if (nodeLaunchType == DXIL::NodeLaunchType::Broadcasting) {
+      // good
+    } else {
+      ValCtx.EmitInstrFormatError(CI,
+                                  ValidationRule::InstrSVConflictingLaunchMode,
+                                  {"ThreadId", "SV_DispatchThreadID",
+                                   GetLaunchTypeStr(nodeLaunchType)});
+    }
+  } break;
```
bogner, line 2386: Another formatting nitpick - `break` on the same line as a closing brace looks weird.

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_broadcast.ll
llvm-beanz: @bob80905, it looks like @tex3d had further comments here. How has this been addressed?

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_broadcast.ll
llvm-beanz: @bob80905, this too.

lib/HLSL/DxilValidation.cpp
bob80905: Clang format will place, and continue replacing, the break  on the same line as the closing brace. I think I'll leave this as is.

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_broadcast.ll
bob80905: I resolved this comment by removing the broadcast test from the PR. I agreed with Tex; it doesn't need to be tested, because it already works and is tested by the other tests we have. This PR doesn't add any extra diagnostics in the broadcast launch type case.

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_broadcast.ll
bob80905: I resolved this by removing the comment entirely. When I get around to implementing these diagnostics in Sema, then I will add launch_types.hlsl, and add a reference in these .ll files to refer back to that file. But, it doesn't make sense to have this reference in this PR because the file isn't being added yet. It can't be added because the test would fail, since diagnostic implementation in Sema should happen separately. launch_types.hlsl tests each launch type for sema diagnostics, and if the sema implementation happens, then validation code won't get to execute, so these files need to go in first without any knowledge of launch_types.hlsl

lib/HLSL/DxilValidation.cpp
bogner: Oh, this break should probably be inside the block for the label, no?
```c++
case abc: {
  // ...
  break;
}
case xyz: ...
```
rather than
```c++
case abc: {
  // ...
}
break;
case xyz: ...
```

bogner: Actually, why are we using braces for these case labels at all? They don't need to introduce a scope.

bogner (approved):

lib/HLSL/DxilValidation.cpp
```diff
@@ -2134,17 +2134,35 @@ static void ValidateBarrierFlagArg(ValidationContext &ValCtx, CallInst *CI,
...
+    return "Broadcasting";
+  case DXIL::NodeLaunchType::Coalescing:
+    return "Coalescing";
+  case DXIL::NodeLaunchType::Thread:
+    return "Thread";
+  case DXIL::NodeLaunchType::Invalid:
+  case DXIL::NodeLaunchType::LastEntry:
+    return "Invalid";
+  }
+  llvm_unreachable("invalid launch type");
```
tex3d, line 2149: I think within the context of the validator, we should not use llvm_unreachable here, since you certainly could have an invalid value that you're passing to validation.  Instead, we should probably have Invalid/LastEntry/default case that simply returns "<invalid>".  That is unless we check for invalid launch type elsewhere first, in which case all of these should be unreachable.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2134,17 +2134,35 @@ static void ValidateBarrierFlagArg(ValidationContext &ValCtx, CallInst *CI,
...
                                                ValidationContext &ValCtx) {
   DXIL::ShaderKind shaderKind =
       pSM ? pSM->GetKind() : DXIL::ShaderKind::Invalid;
   llvm::Function *F = CI->getParent()->getParent();
+  DXIL::NodeLaunchType nodeLaunchType = DXIL::NodeLaunchType::Invalid;
   if (DXIL::ShaderKind::Library == shaderKind) {
-    if (ValCtx.DxilMod.HasDxilFunctionProps(F))
+    if (ValCtx.DxilMod.HasDxilFunctionProps(F)) {
+      DxilEntryProps &entryProps = ValCtx.DxilMod.GetDxilEntryProps(F);
+      nodeLaunchType = entryProps.props.Node.LaunchType;
```
tex3d, line 2163: Should check for `shaderKind == DXIL::ShaderKind::Node` before reading the `Node.LaunchType` property.  While not currently the case, `NodeProps` could be integrated into the `ShaderProps` union now that it's mutually exclusive with other shader types, at which point, checking would be necessary to avoid reading garbage.

lib/HLSL/DxilValidation.cpp
```diff
@@ -2134,17 +2134,39 @@ static void ValidateBarrierFlagArg(ValidationContext &ValCtx, CallInst *CI,
...
+  switch (LT) {
+  case DXIL::NodeLaunchType::Broadcasting:
+    return "Broadcasting";
+  case DXIL::NodeLaunchType::Coalescing:
+    return "Coalescing";
+  case DXIL::NodeLaunchType::Thread:
+    return "Thread";
+  case DXIL::NodeLaunchType::Invalid:
+  case DXIL::NodeLaunchType::LastEntry:
+    return "";
```
python3kgae, line 2147: Should this be "Invalid"?

tex3d: That would be because of a bad rendering of my last comment.  It was written as `"<invalid>"` but without the back-ticks around it, so when rendering the comment, it dropped what it thought was an HTML tag 😜.  Also, we shouldn't need the separate cases, we just need a single `default:` case.

lib/HLSL/DxilValidation.cpp
tex3d: FYI: "" was mis-rendered.  It was written as `"<invalid>"`, and the middle was dropped when rendering the comment!

tex3d (approved):
LGTM.

PR #6111, opened 2023-12-18
Fix DxilDescriptorRangeFlags::ValidFlags

austinEng:
it was missing a zero, so DescriptorsStaticKeepingBufferBoundsChecks was not considered valid.


austinEng:
sorry for the noise - this is not the correct fix.

PR #6112, opened 2023-12-19
[CMAKE] Cleanup logic for enabling SPIR-V

llvm-beanz:
This just cleans up the logic for enabling and disabling building SPIR-V and the SPIR-V tests. It is _extremely_ unusable to ignore a user-defined value that is specified as an option. In the current code if a user enables SPIR-V tests but disables SPIR-V codegen, the user-specified option to enable codegen gets ignored.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,21 +129,24 @@ endif()
 # HLSL Change Ends
 
 # SPIRV change starts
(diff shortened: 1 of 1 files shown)


Keenuts (approved):

s-perron (approved):

s-perron:
The changes look good to me. I cannot see why the bots failed.

dnovillo:
@llvm-beanz do you still want this in? Not sure why it's still not submitted.

PR #6122, opened 2024-01-03
Fix bad arg indexing after array of struct in semantic assignment

### LINKED ISSUE #6121: Assert/crash during semantic index assignment after array of struct
**Description**
Argument Indexing is incorrect after an array of struct in a signature is encountered, leading to incorrect indexing, assert, and/or crash.

[This code](https://github.com/microsoft/DirectXShaderCompiler/blob/7d2f9c74d53dcbb9de4e148b68c99e5986875be6/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp#L4566) is meant to increment the parameter index after a struct field of a basic type, but misinterprets any array as a basic type, so increments the `argIdx` when it shouldn't.

**Steps to Reproduce**
```hlsl
// RUN: %dxc -E main -T vs_6_0 %s

struct Foo {
    float a : A;
    float b[2] : B;
};
struct INPUT {
    float4 pos : SV_Position;
    Foo foo[3] : FOO;
};
struct OUTPUT {
    Foo foo[3] : FOO;
    float4 pos : SV_Position;
};

void main(INPUT In : IN, out OUTPUT Out : OUT) {
    Out.pos = In.pos;
    Out.foo = In.foo;
}
```

**Actual Behavior**
Crash or assert.

**Environment**
- DXC version: recent main build
- Host Operating System: any

tex3d:
Semantic index assignment for flattened arguments would incorrectly increment the argument index after an array of struct field was encountered. It would misinterpret any array of struct as a basic type because it wasn't a struct type.

This fixes the bug by drilling into the array element before checking the type.

Fixes #6121

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -4563,6 +4563,10 @@ static unsigned AllocateSemanticIndex(
       Type *EltTy = Ty->getStructElementType(i);
       argIdx = AllocateSemanticIndex(EltTy, semIndex, argIdx, endArgIdx,
                                      FlatAnnotationList);
+      // Unwrap array types when checking whether this is a leaf node,
+      // otherwise, array of struct will be misinterpreted as a leaf node.
+      while (EltTy->isArrayTy())
+        EltTy = EltTy->getArrayElementType();
       if (!(EltTy->isStructTy() && !HLMatrixType::isa(EltTy))) {
         // Update argIdx only when it is a leaf node.
         argIdx++;
--- a/tools/clang/test/HLSLFileCheck/hlsl/signature/sig-flatten-struct-array.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/signature/sig-flatten-struct-array.hlsl
@@ -1,0 +1,52 @@
+// RUN: %dxc -E main -T vs_6_0 %s | FileCheck %s
+
+// Check correct assignment of semantic indices to parameter with array of
+// struct followed by another argument, which was triggering an assert
+// and invalid indexing for updating the index in the annotation.
+
+// CHECK: ; Output signature:
+// CHECK-NEXT: ;
+// CHECK-NEXT: ; Name                 Index   Mask Register SysValue  Format   Used
+// CHECK-NEXT: ; -------------------- ----- ------ -------- -------- ------- ------
+// CHECK-NEXT: ; OUT                      0   x           0     NONE   float   x
+// CHECK-NEXT: ; OUT                      1    y          0     NONE   float    y
+// CHECK-NEXT: ; OUT                      3   x           1     NONE   float   x
+// CHECK-NEXT: ; OUT                      2    y          1     NONE   float    y
+// CHECK-NEXT: ; OUT                      6   x           2     NONE   float   x
+// CHECK-NEXT: ; OUT                      4    y          2     NONE   float    y
+// CHECK-NEXT: ; OUT                      5    y          3     NONE   float    y
+// CHECK-NEXT: ; OUT                      7    y          4     NONE   float    y
+// CHECK-NEXT: ; OUT                      8    y          5     NONE   float    y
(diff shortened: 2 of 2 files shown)


tools/clang/test/HLSLFileCheck/hlsl/signature/sig-flatten-struct-array.hlsl
```diff
@@ -0,0 +1,52 @@
...
+// CHECK-NEXT: ; OUT                      0   x           0     NONE   float   x
+// CHECK-NEXT: ; OUT                      1    y          0     NONE   float    y
+// CHECK-NEXT: ; OUT                      3   x           1     NONE   float   x
+// CHECK-NEXT: ; OUT                      2    y          1     NONE   float    y
+// CHECK-NEXT: ; OUT                      6   x           2     NONE   float   x
+// CHECK-NEXT: ; OUT                      4    y          2     NONE   float    y
+// CHECK-NEXT: ; OUT                      5    y          3     NONE   float    y
+// CHECK-NEXT: ; OUT                      7    y          4     NONE   float    y
+// CHECK-NEXT: ; OUT                      8    y          5     NONE   float    y
+// CHECK-NEXT: ; OUT                      9   xyzw        6     NONE   float   xyzw
```
python3kgae, line 20: Would be nice to add comment about each line like
// Out.foo[0].a
// CHECK-NEXT: ; OUT                      0   x           0     NONE   float   x
// Out.foo[0].b[0]
// CHECK-NEXT: ; OUT                      1    y          0     NONE   float    y
...
// Out.postion
// CHECK-NEXT: ; OUT                      9   xyzw        6     NONE   float   xyzw

tex3d: done.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -4563,6 +4563,10 @@ static unsigned AllocateSemanticIndex(
       Type *EltTy = Ty->getStructElementType(i);
       argIdx = AllocateSemanticIndex(EltTy, semIndex, argIdx, endArgIdx,
                                      FlatAnnotationList);
+      // Unwrap array types when checking whether this is a leaf node,
+      // otherwise, array of struct will be misinterpreted as a leaf node.
+      while (EltTy->isArrayTy())
```
bob80905, line 4568: Similar to this circumstance, I notice that we only ever go one level deep if a struct is detected, and recurse. I would expect something like 
`
 while (EltTy->isStructTy())
        EltTy = EltTy->getStructElementType();
`
but instead we use reccursion. Is it because every struct that is encountered requires an increment in argIdx? If not, why is drilling into an Array treated differently than drilling into a struct here?

tex3d: The logic is a bit complicated, but I think it is sound.

The argument index keeps track of which flattened argument we are currently assigning indices to.  The only time we are changing arguments is when we advance past a struct field of basic type, since each of these are what become separate arguments after flattening.  You don't want to increment on a field of struct type because that will double-increment: an increment from the inner struct field, then an increment from the outer struct field, but the outer field doesn't have an argument of its own, unless the field is a "leaf node", hence the bug.

When we iterate into an array, if it contains a struct, we might have to add indices to multiple final flattened arguments, but since semantic index assignment and increment is depth-first, we have to visit those fields/arguments for each element of the array, adding a few semantic indices at a time across any fields in that element of the array.  This is why the argument index is incremented only when processing a struct inside the array, and returned, but then the argument index starts at the beginning again for each iteration of array elements, finally returning the incremented argument index from the last iteration.  If the array did not contain a struct, no argument increment will occur when handling the array, so if it was a field of a struct, any needed increment will be performed when advancing past the field.

In theory, you could assign blocks of indices without iterating array elements, but it would be even more complicated to compute (requiring index expansion size information from all other struct fields at each level when computing indexing for one field), and you still need to add each index individually, so you can't save much of the work required.

I think the best future improvement might be to abstract the semantic/index assignment logic so that it may use the same code here as we would want to use in Sema to validate the signature elements.

bob80905 (approved):

PR #6123, opened 2024-01-03
Allow node input record to be a class

### LINKED ISSUE #6119: A class as a node input record causes compiler to crash
**Description**
A node input record that is a class rather than a struct causes an assert in CodeGen.

**Steps to Reproduce**

```
class ClassInputRecord
{
  uint a;
};

[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeDispatchGrid(2,3,2)]
[NumThreads(1024,1,1)]
void node01(DispatchNodeInputRecord<ClassInputRecord> input)
{ }
```
Save the example shader as class_input_record.hlsl and compile using:
dxc -T lib_6_8 class_input_record.hlsl

**Actual Behavior**
Results in an access violation during CodeGen, due to expecting the record to be a struct rather than a class

**Environment**
- DXC version dxcompiler.dll 1.8 - 1.7.0.14388 (main, 7d2f9c74d)
- Host Operating System 
The failure can also be reproduced on https://gobolt.org using DXC(trunc)

tcorringham:
Node input records may be a struct or a class, but CodeGen failed to correctly handle the case when a class was used. This change fixes the error, and adds a test case.

Fixes #6119

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/workgraph/class_input_record.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/workgraph/class_input_record.hlsl
@@ -1,0 +1,20 @@
+// RUN: %dxc -T lib_6_8 `%s | FileCheck %s
+// ==================================================================
+// Broadcasting launch node with class input record
+// ==================================================================
+
+class ClassInputRecord
+{
+  uint a;
+};
+
+// CHECK-NOT: error
+// CHECK: define void @node01()
+// CHECK-NOT: error
+
+[Shader("node")]
+[NodeLaunch("broadcasting")]
+[NodeDispatchGrid(2,3,2)]
+[NumThreads(1024,1,1)]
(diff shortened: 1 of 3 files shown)


hekota (approved):
LGTM!

tools/clang/test/HLSLFileCheck/hlsl/workgraph/class_input_record.hlsl
llvm-beanz, line 1: Can you instead move this test to SemaHLSL and have the RUN line be `-verify` instead of FileCheck? Then you can have `//expected-no-diagnostics` to signify that the verifier shouldn't find any diagnostics rather than relying on FileCheck directives.

tcorringham: Done.

llvm-beanz (approved):

Issue #5855, opened 2023-10-11
SemaHLSL diag for CalculateLevelOfDetail too strict when code requiring SM 6.8 is unreachable


tex3d:
**Description**
PR #5802 added diagnostics for the use of `SamplerComparisonState` with `CalculateLevelOfDetail`, requiring SM 6.8 or above.  However, this kind of diagnostic needs to be performed only on active code paths, otherwise it will still produce an error when compiling a different entry point to a lower shader model, when this code path is not reachable.

See section added below:
https://github.com/microsoft/DirectXShaderCompiler/blob/d7d0d66764a8ee188527a4378c84bc90cf0a0da4/tools/clang/lib/Sema/SemaHLSL.cpp#L11295-L11311

Instead, we need to only do these sorts of shader target dependent checks on active entries and exports to avoid misdiagnosing usages that would be legal when under another compilation when those code paths are active.

Normally, this kind of verification is left up to `DxilValidation`, but it would be a lot more convenient for the developer to get an error earlier in the compilation pipeline, where richer diagnostic information is available.
 
I believe this will require call graph traversal in `DiagnoseTranslationUnit` from each active entry or export.  This is already necessary for the recursion check, so it should be possible to do the same for checking for reachable code.  It will also be necessary to do constant branching analysis to avoid considering statically disabled blocks reachable.

Since the immediate change does introduce a potential regression now, the first step may be to remove this check from SemaHLSL.cpp until we have a more robust solution in place.

**Steps to Reproduce**
```hlsl
// RUN: %dxc -E main -T ps_6_6 | FileCheck %s

// CHECK: %[[SAMP:[0-9]+]] = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %{{.*}}, %dx.types.ResourceProperties { i32 14, i32 0 })
// CHECK: call float @dx.op.calculateLOD.f32(i32 81, %dx.types.Handle %{{.*}}, %dx.types.Handle %[[SAMP]],

SamplerState S;
SamplerComparisonState SC;
Texture2D<float> t2d;

float4 main(float2 coord : TEXCOORD) : SV_Target {
  return t2d.CalculateLevelOfDetail(S, coord);
}

// Unused entry point function which uses new SM 6.8 overload
float4 main_SC(float2 coord : TEXCOORD) : SV_Target {
  return t2d.CalculateLevelOfDetail(SC, coord);
}
```

**Actual Behavior**
```
[...]

PR #6126, opened 2024-01-04
[NFC] Move hlsl::DiagnoseTranslationUnit to separate file.

python3kgae:
Since DiagnoseTranslationUnit is after Sema, it doesn't need to be part of SemaHLSL. 
This is in preparation for moving DiagnoseHLSLMethodCall to hlsl::DiagnoseTranslationUnit for access to the call graph.

Only move code to new file, not other changes except replace OutputDebugFormat with llvm::dbgs for CallGraphWithRecurseGuard::dump.

For https://github.com/microsoft/DirectXShaderCompiler/issues/5855

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
+++ b/tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
@@ -29,7 +29,7 @@ using namespace hlsl;
 // - merge graph visiting with checking for recursion
 // - track global variables and types used (NYI)
 //
-namespace hlsl {
+namespace {
 struct CallNode {
   FunctionDecl *CallerFn;
   ::llvm::SmallPtrSet<FunctionDecl *, 4> CalleeFns;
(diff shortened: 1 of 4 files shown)


bob80905 (approved):

tex3d (approved):

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -0,0 +1,429 @@
...
+  FunctionDecl *CheckRecursion(FunctionDecl *EntryFnDecl) {
+    FnCallStack CallStack;
+    EntryFnDecl = getFunctionWithBody(EntryFnDecl);
+    return CheckRecursion(CallStack, EntryFnDecl);
+  }
+
+  const CallNodes &GetCallGraph() { return m_callNodes; }
+
+  void dump() const {
+    llvm::dbgs() << "Call Nodes:\r\n";
```
tex3d, line 179: Oh, should `\r\n` be switched to `\n`?

python3kgae: Updated.

PR #2503, opened 2019-10-03
[spirv] Don't emit extension for using fp16 in GLSLstd450 instructions.

ehsannas:
With the exception of interpolation instructions.

New revision of SPIR-V Extended Instructions for GLSL allows fp16
except for interpolation instructions.

See: https://www.khronos.org/registry/spir-v/specs/1.0/GLSL.std.450.html

Fixes #2500.

PR #6127, opened 2024-01-04
[SPIR-V] Remove dead code and correct Float16 docs

sudonatalie:
The docs were out of date stating that the `SPV_AMD_gpu_shader_half_float` extension was added when half types were used. This is now corrected to show that the Float16 capability is added and no extension.

The remaining code that could have added that extension in certain cases has been dead code since #2503 since none of the instructions checked in that list are ones that are emitted by DXC, so it is removed to clean up.

Evidence that this is dead code:
1. `git grep "InterpolateAt"` finds no uses other than those removed here.
2. All tests still pass when it's removed.
3. The comments in #2503 also say they were not able to find a mapping from any HLSL intrinsics to these instructions, which is why no tests were added.

Follow-up from: https://github.com/KhronosGroup/SPIRV-Tools/pull/5519

### DIFF
--- a/tools/clang/lib/SPIRV/CapabilityVisitor.cpp
+++ b/tools/clang/lib/SPIRV/CapabilityVisitor.cpp
@@ -736,27 +736,6 @@ bool CapabilityVisitor::visit(SpirvExtInstImport *instr) {
   return true;
 }
 
-bool CapabilityVisitor::visit(SpirvExtInst *instr) {
-  // OpExtInst using the GLSL extended instruction allows only 32-bit types by
-  // default for interpolation instructions. The AMD_gpu_shader_half_float
-  // extension adds support for 16-bit floating-point component types for these
-  // instructions:
-  // InterpolateAtCentroid, InterpolateAtSample, InterpolateAtOffset
-  if (SpirvType::isOrContainsType<FloatType, 16>(instr->getResultType()))
-    switch (instr->getInstruction()) {
-    case GLSLstd450::GLSLstd450InterpolateAtCentroid:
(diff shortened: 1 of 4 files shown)


Keenuts (approved):

tools/clang/lib/SPIRV/CapabilityVisitor.cpp
```diff
@@ -736,27 +736,6 @@ bool CapabilityVisitor::visit(SpirvExtInstImport *instr) {
...
-  // OpExtInst using the GLSL extended instruction allows only 32-bit types by
-  // default for interpolation instructions. The AMD_gpu_shader_half_float
-  // extension adds support for 16-bit floating-point component types for these
-  // instructions:
-  // InterpolateAtCentroid, InterpolateAtSample, InterpolateAtOffset
-  if (SpirvType::isOrContainsType<FloatType, 16>(instr->getResultType()))
-    switch (instr->getInstruction()) {
-    case GLSLstd450::GLSLstd450InterpolateAtCentroid:
-    case GLSLstd450::GLSLstd450InterpolateAtSample:
-    case GLSLstd450::GLSLstd450InterpolateAtOffset:
```
s-perron, line 749: Do we not generate these instructions anymore? Why can we get rid of this code?

sudonatalie: The PR description includes the justification, let me know if that seems sufficient

s-perron: Right, sorry I did not read the description closely.

s-perron (approved):

PR #6131, opened 2024-01-05
Correct sampling param error in DXIL docs

pow2clk:
The documentation for sample incorrectly implied that textureCUBE resources should have 3 active offsets. Actually, cube textures have no sample method overloads that include offsets.

This corrects the table indicating the number of active elements for sample and also for sampleGrad as it incorrectly stated that the number of active gradients is the same as the number of offsets. Again for cube textures this is inaccurate. Cube textures take 3 gradients, but still allow no active offsets.


bharadwajy (approved):
LGTM!

tex3d (approved):
Looks like a set of good fixes!

PR #6134, opened 2024-01-08
Remove LLVM_C_FALLTHROUGH instances that cause warnings on clang-12

hekota:
The clang-12 warning is:

    error: fallthrough annotation in unreachable code [-Werror,-Wimplicit-fallthrough]

These LLVM_C_FALLTHROUGH were originally added to silence GCC warnings, but the GCC PR validation build is has succeeded, so it looks like they are not needed.


llvm-beanz (approved):

pow2clk (approved):
Obviously unreachable

PR #6137, opened 2024-01-09
Require outputtopology attribute on Hull shader

### LINKED ISSUE #3740: [SPIR-V] Omitting outputtopology Hull entry point attribute does not get caught (is caught by DXIL backend)
### Title
[SPIR-V] Omitting outputtopology Hull entry point attribute does not get caught and reported (is caught by DXIL backend)

### Functional impact
When this is attribute is omitted the vertex winding order is not set (i.e. no `VertexOrderCw` or `VertexOrderCcw` execution mode is in the SPIR-V), the [tessellation vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/chap23.html) states that `VertexOrderCw and VertexOrderCcw. These control the orientation of triangles generated by the tessellator. One mode must be set in at least one of the tessellation shader stages.`. So if I didn't misunderstand that part, the generated SPIR-V is invalid. Also I'm guessing this is a required HLSL Hull stage attribute as DXC fails validation for the DXIL backend.

### Minimal repro steps
[repro_shader.txt](https://github.com/microsoft/DirectXShaderCompiler/files/6415572/repro_shader.txt)

### Expected result
Compilation should fail with an error message stating that the `outputtopology` Hull stage attribute is missing. Similar to the DXIL backend error:
```
error: validation errors
repro_shader.txt:14: error: Invalid Tessellator Output Primitive specified. Must be point, line, triangleCW or triangleCCW.
Validation failed.
```

### Actual result
Compilation completes without errors, the generated SPIR-V has neither `VertexOrderCw` nor `VertexOrderCcw` execution model.

### Further technical details
**N.B.**: the DXIL backend error message could probably also be improved as the line `error: validation errors` and `Validation failed.` are not useful

sudonatalie:
There was an existing check in DXIL validation that failed when the outputtopology was not set for a Hull shader. Rather than duplicate this check in the SPIR-V backend/validator, it seems correct to check that it is present in Sema.

This PR needs review from someone who is more familiar with the HLSL spec for hull shaders to verify that this check is correct though. I can't say for sure from the HLSL docs.

Fixes #3740


sudonatalie:
@llvm-beanz Is the `outputtopology` attribute required on Hull shaders? It's not completely clear to me, and the HLSL tests that required it to be added here were only so that the expected error would occur in a later stage of compilation.

tools/clang/test/HLSLFileCheck/hlsl/functions/attribute/outputtopology_missing.hlsl
```diff
@@ -0,0 +1,17 @@
+// RUN: %dxc -T hs_6_0 -E Hull %s | FileCheck %s
```
s-perron, line 1: Do we want a second run command with `-spirv`? Theoretical, but not likely, concern is that someone will move the check to a place on a DXIL only path, and break SPIR-V again.

s-perron (approved):
LGTM, I'm just wondering if we should have a spir-v specific test.

We need to hear from MS to be sure they do not expect optimizations to somehow hide some cases where this would trigger. I personally do not see it.

Keenuts (approved):

tools/clang/test/HLSLFileCheck/hlsl/functions/attribute/outputtopology_missing.hlsl
sudonatalie: Done (created as a separate test though since SPIR-V codegen is off by default, so it wouldn't necessarily be enabled for these tests on other builds)

tools/clang/test/CodeGenSPIRV/attribute.outputtopology.error.hlsl
```diff
@@ -0,0 +1,17 @@
+// RUN: not %dxc -T hs_6_0 -E Hull -spirv %s 2>&1 | FileCheck %s
```
llvm-beanz, line 1: This should also be a verify test.

tools/clang/test/HLSLFileCheck/hlsl/functions/recursion/library_hull_recursion.hlsl
llvm-beanz, line 1: Can we also make this a `-verify` test?

llvm-beanz:
> @llvm-beanz Is the `outputtopology` attribute required on Hull shaders? It's not completely clear to me, and the HLSL tests that required it to be added here were only so that the expected error would occur in a later stage of compilation.

Yea. It should be required, but I think the error was implemented as a validator error instead of a sema error:

https://godbolt.org/z/hTq1MK5xM

We've been working on moving errors earlier. I have a big stocked up rant about how much I hate that we diagnose things in the validator instead of emitting warnings/errors in Sema. IMO, any validator error should at least be a Sema warning and possibly an error. Last night I posted a related spec for review to propose a different solution in Clang (see: https://github.com/microsoft/hlsl-specs/pull/153).

llvm-beanz (approved):
One comment below, but overall this is awesome. Thank you!

tools/clang/test/HLSLFileCheck/hlsl/functions/recursion/library_hull_recursion.hlsl
```diff
@@ -1,4 +1,4 @@
-// RUN: %dxc -T lib_6_5 %s | FileCheck %s
+// RUN: %dxc -T lib_6_5 -verify %s
```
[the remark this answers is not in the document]
sudonatalie: Thanks, I over-optimistically pushed the commit before `check-all` was quite done.

PR #6140, opened 2024-01-09
[Sema] Not report error when invalid intrinsic call is not used.

python3kgae:
Added DiagnoseUsedHLSLMethodCall which only report error when the MethodCall is used.
DiagnoseUsedHLSLMethodCall is used in DiagnoseTranslationUnit. 
In DiagnoseTranslationUnit, traverse all visited functions in call graph, diagnose illegal intrinsic call inside these functions.

Shader kind is also checked in DiagnoseUsedHLSLMethodCall.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5855


tools/clang/lib/Sema/SemaHLSLHelper.h
llvm-beanz, line 1: Since this method takes a pointer to Sema, shouldn't this just be a method on the Sema object? Added benefit there is that we then don't need an extra header.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15610,4 +15595,28 @@ void DiagnoseEntry(Sema &S, FunctionDecl *FD) {
   }
   }
 }
+
+// Check if intrinsic function is illegal in current shader model.
+bool isIllegalIntrinsic(const CXXMethodDecl *MD, IntrinsicOp opCode,
```
llvm-beanz, line 15600: nit: What about `isInvalidForHLSLTarget`?

"Illegal" sounds a bit strong wording and the name doesn't really capture that it is only contextually invalid.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -295,6 +301,39 @@ clang::FunctionDecl *ValidateNoRecursion(CallGraphWithRecurseGuard &callGraph,
   return nullptr;
 }
 
+bool DiagnoseHLSLMethodCall(CXXMemberCallExpr *CE, clang::Sema *self) {
```
bob80905, line 304: I do not like how this function is named exactly the same as the function "DiagnoseHLSLMethodCall" in SemaHLSL.cpp. I would think we'd either want to make one function, or leave these two functions as they are except rename them so that we know how they differ.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -295,6 +301,39 @@ clang::FunctionDecl *ValidateNoRecursion(CallGraphWithRecurseGuard &callGraph,
...
+  const CXXMethodDecl *MD = CE->getMethodDecl();
+  if (MD->hasAttr<clang::HLSLIntrinsicAttr>()) {
+    hlsl::IntrinsicOp opCode =
+        (IntrinsicOp)MD->getAttr<clang::HLSLIntrinsicAttr>()->getOpcode();
+    if (opCode == hlsl::IntrinsicOp::MOP_CalculateLevelOfDetail ||
+        opCode == hlsl::IntrinsicOp::MOP_CalculateLevelOfDetailUnclamped) {
+      if (isIllegalIntrinsic(MD, opCode, self)) {
```
llvm-beanz, line 311: The splitting of logic here is a bit odd. `isIllegalIntrinsic` has the logic for why the intrinsic is not allowed, but the reporting of the error is here disjointed from the logic. That makes this code significantly harder to follow, reuse and extend.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11138,22 +11139,6 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
         Diags.Report(Loc, diag::err_hlsl_wg_nodetrackrwinputsharing_missing);
         return true;
       }
-    } else if (opCode == hlsl::IntrinsicOp::MOP_CalculateLevelOfDetail ||
```
bob80905, line 11141: I would think the call-site of this function, wherever it's called, would need to be updated to account for the case of the opcode being one of these two. The logic has moved, so wherever DiagnoseHLSLMethodCall is called, wouldn't we also need to call isIllegalIntrinsic now?

bob80905:
From reading the changes, I don't think the title of the PR matches what's being done. You are separating out the logic for identifying an error, and actually emitting a diagnostic. Maybe a better title would be:
"Separate diagnostic detection logic from diagnostic emission in Sema for invalid method calls."

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
python3kgae: Fixed. Now there're only 1 DiagnoseHLSLMethodCall.

tools/clang/lib/Sema/SemaHLSLHelper.h
python3kgae: Fixed. Extra header removed.

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: The function is removed.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
python3kgae: Fixed.

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: Removed isIllegalIntrinsic.

python3kgae:
> From reading the changes, I don't think the title of the PR matches what's being done. You are separating out the logic for identifying an error, and actually emitting a diagnostic. Maybe a better title would be: "Separate diagnostic detection logic from diagnostic emission in Sema for invalid method calls."

The implementation is updated.
Let me know if the title still needs to be change.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11117,40 +11117,44 @@ void hlsl::DiagnoseRegisterType(clang::Sema *self, clang::SourceLocation loc,
...
-      auto &TemplateArgs = templateDecl->getTemplateArgs();
-      DXASSERT(TemplateArgs.size() == 1,
-               "Input record types need to have one template argument");
-      auto &Rec = TemplateArgs.get(0);
-      clang::QualType RecType = Rec.getAsType();
-      RecordDecl *RD = RecType->getAs<RecordType>()->getDecl();
-      if (!RD->hasAttr<HLSLNodeTrackRWInputSharingAttr>()) {
-        Diags.Report(Loc, diag::err_hlsl_wg_nodetrackrwinputsharing_missing);
-        return true;
+    if (SkipUnused) {
```
tex3d, line 11125: It seems like we wouldn't want code executing from both cases, so why can't we make this case a separate function?  That way arguments can be added as needed to handle the call graph traversal case without impacting parse-time diagnostics.

python3kgae: Fixed.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11138,25 +11138,44 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+bool Sema::DiagnoseUsedHLSLMethodCall(const CXXMethodDecl *MD,
+                                      SourceLocation Loc,
+                                      const hlsl::ShaderModel *SM,
+                                      DXIL::ShaderKind EntrySK) {
+  bool bError = false;
+  if (MD->hasAttr<HLSLIntrinsicAttr>()) {
+    hlsl::IntrinsicOp opCode =
+        (IntrinsicOp)MD->getAttr<HLSLIntrinsicAttr>()->getOpcode();
+
+    if (opCode == hlsl::IntrinsicOp::MOP_CalculateLevelOfDetail ||
```
bob80905, line 11156: Should we make this a switch statement instead for future additions to this function?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15610,4 +15629,5 @@ void DiagnoseEntry(Sema &S, FunctionDecl *FD) {
   }
   }
 }
+
```
bob80905, line 15632: This whitespace doesn't belong

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -295,6 +301,25 @@ clang::FunctionDecl *ValidateNoRecursion(CallGraphWithRecurseGuard &callGraph,
...
 
+class HLSLMethodCallDiagnoseVisitor
+    : public RecursiveASTVisitor<HLSLMethodCallDiagnoseVisitor> {
+public:
+  explicit HLSLMethodCallDiagnoseVisitor(Sema *S, const hlsl::ShaderModel *SM,
+                                         DXIL::ShaderKind EntrySK)
+      : sema(S), SM(SM), EntrySK(EntrySK) {}
+
+  bool VisitCXXMemberCallExpr(CXXMemberCallExpr *CE) {
+    sema->DiagnoseUsedHLSLMethodCall(CE->getMethodDecl(), CE->getExprLoc(), SM,
```
bob80905, line 312: The boolean return value for this function is discarded / unused. Are you sure you always want to return true here? Do you not want to use the return value for this change?

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -354,10 +379,14 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
     }
   }
 
-  CallGraphWithRecurseGuard callGraph;
```
bob80905, line 357: Wouldn't it be more efficient to leave the declaration outside of the for loop and reuse the callgraph?

python3kgae: The visited functions need to limit to current entry so we can use the entry shader kind when available.
If reuse callgraph, the visited functions will be mixed, cannot know the shader kind anymore.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
python3kgae: Return true is OK here.
We already emit the error in DiagnoseUsedHLSLMethodCall.
Maybe we could change DiagnoseUsedHLSLMethodCall to return void?

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: Updated.

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: Removed.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
llvm-beanz: If the only place `DiagnoseUsedHLSLMethodCall` is used we ignore the return value, it seems like it should return void.

python3kgae: Updated.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7771,6 +7771,8 @@ def err_hlsl_unsupported_for_version_lower : Error<
    "%0 is only allowed for HLSL %1 and lower.">;
 def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
+def err_hlsl_derivatives_intrinsic_in_wrong_shader_kind : Error<
```
llvm-beanz, line 7774: Can we instead make this a warning that is `DefaultError`? Since we still need to have the validator catch these cases this would allow developers to disable the error (or make it a warning) if it interferes with their code in the ways that @tex3d has mentioned.

bob80905 (approved):
LGTM

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
python3kgae: Updated.

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -494,6 +494,18 @@ const llvm::StringRef ShaderModel::FullNameFromKind(DXIL::ShaderKind sk) {
   }
 }
 
+bool ShaderModel::AllowDerivatives(DXIL::ShaderKind sk) {
```
tex3d, line 497: I think this function needs shader model as well, since derivatives were enabled in compute+amp+mesh in SM 6.6.

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -494,6 +494,18 @@ const llvm::StringRef ShaderModel::FullNameFromKind(DXIL::ShaderKind sk) {
...
+  switch (sk) {
+  case DXIL::ShaderKind::Pixel:
+  case DXIL::ShaderKind::Library:
+  case DXIL::ShaderKind::Compute:
+  case DXIL::ShaderKind::Amplification:
+  case DXIL::ShaderKind::Mesh:
+    return true;
```
tex3d, line 504: Missing `DXIL::ShaderKind::Node`, right?

lib/DXIL/DxilShaderModel.cpp
python3kgae: Fixed.

lib/DXIL/DxilShaderModel.cpp
python3kgae: Added node.
Should all the ray tracing shader kind be added here as well?

tex3d: No, derivative ops are not allowed in DXR shaders.

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -494,6 +494,20 @@ const llvm::StringRef ShaderModel::FullNameFromKind(DXIL::ShaderKind sk) {
...
 
+bool ShaderModel::AllowDerivatives(DXIL::ShaderKind sk) const {
+  switch (sk) {
+  case DXIL::ShaderKind::Pixel:
+  case DXIL::ShaderKind::Library:
+    return true;
+  case DXIL::ShaderKind::Compute:
+  case DXIL::ShaderKind::Amplification:
+  case DXIL::ShaderKind::Mesh:
+  case DXIL::ShaderKind::Node:
```
tex3d, line 505: While this wouldn't cause a problem, node doesn't need the shader model check, since it only exists in SM 6.8+.  There should be an independent check for trying to use Node in a lower shader model.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7771,8 +7771,12 @@ def err_hlsl_unsupported_for_version_lower : Error<
    "%0 is only allowed for HLSL %1 and lower.">;
 def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
-def err_hlsl_intrinsic_overload_in_wrong_shader_model : Error<
-   "overload of intrinsic %0 requires shader model %1 or greater">;
+def err_hlsl_derivatives_intrinsic_in_wrong_shader_kind : Warning<
+   "Derivatives intrinsic %0 only works in lib/ps/cs/as/ms">,
```
tex3d, line 7775: With the more subtle check combining stage and version, the message either needs revision, or we need different messages depending on whether we failed due to stage or shader model.

We also need to add node, and I think including 'lib' here might just add confusion, since it's not clear what this means to the user.

lib/DXIL/DxilShaderModel.cpp
python3kgae: Updated.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
python3kgae: Updated.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7771,8 +7771,12 @@ def err_hlsl_unsupported_for_version_lower : Error<
    "%0 is only allowed for HLSL %1 and lower.">;
 def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
-def err_hlsl_intrinsic_overload_in_wrong_shader_model : Error<
-   "overload of intrinsic %0 requires shader model %1 or greater">;
+def err_hlsl_derivatives_intrinsic_in_wrong_shader_kind : Warning<
+   "Derivatives intrinsic %0 only works in ps and cs_6.6+/as_6.6+/ms_6.6+/node">,
+    DefaultError, InGroup<HLSLIntrinsicCall>;
+def err_hlsl_intrinsic_overload_in_wrong_shader_model : Warning<
```
llvm-beanz, line 7777: These are now warnings, not errors, so they shouldn't start with the `err_` prefix.

python3kgae: Fixed.

tools/clang/include/clang/Sema/Sema.h
```diff
@@ -3807,6 +3808,9 @@ class Sema {
                                         QualType TargetType,
                                         SourceLocation Loc);
   bool DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc);
+  void DiagnoseUsedHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc,
```
llvm-beanz, line 3811: nit: this method name doesn't really convey what you're diagnosing. This isn't diagnosing that a method call is unused, but rather diagnosing _availability_ of methods regardless of liveness.

llvm-beanz (commented):
We should also probably add a test that verifies the `-Wno-error=` flag functionality (if it works). I don't think we have any tests running in tree that exercise that since we don't support the C++ tests.

tools/clang/include/clang/Basic/DiagnosticGroups.td
```diff
@@ -800,4 +800,5 @@ def HLSLPayloadAccessQualifer: DiagGroup<"payload-access-qualifier", [
 def HLSLSemanticIdentifierCollision : DiagGroup<"semantic-identifier-collision">;
 def HLSLStructurizeExitsLifetimeMarkersConflict: DiagGroup<"structurize-exits-lifetime-markers-conflict">;
 def HLSLParameterUsage : DiagGroup<"parameter-usage">;
+def HLSLIntrinsicCall : DiagGroup<"intrinsic-call">;
```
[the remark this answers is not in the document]
python3kgae: Updated.

tools/clang/include/clang/Sema/Sema.h
python3kgae: Only method call which could reached from entry or exported function will be diagnosed.
How about DiagnoseReachableHLSLMethodCall?

python3kgae:
> We should also probably add a test that verifies the `-Wno-error=` flag functionality (if it works). I don't think we have any tests running in tree that exercise that since we don't support the C++ tests.

-Wno-error= not work :(

llvm-beanz:
> -Wno-error= not work :(

Can you file an issue for that?

python3kgae:
> > -Wno-error= not work :(
> 
> Can you file an issue for that?

It works. Thought it does not work when hit lit test fail.
Added test for -Wno-error.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -354,10 +379,14 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
+  const auto *shaderModel =
+      hlsl::ShaderModel::GetByName(self->getLangOpts().HLSLProfile.c_str());
+  DXIL::ShaderKind shaderKind = shaderModel->GetKind();
```
tex3d, line 384: How about `profileKind`, as it pertains to the `HLSLProfile` option, or perhaps just use `shaderModel->GetKind()` directly later, since keeping this as a separate local variable isn't really necessary.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -295,6 +301,25 @@ clang::FunctionDecl *ValidateNoRecursion(CallGraphWithRecurseGuard &callGraph,
...
+  bool VisitCXXMemberCallExpr(CXXMemberCallExpr *CE) {
+    sema->DiagnoseUsedHLSLMethodCall(CE->getMethodDecl(), CE->getExprLoc(), SM,
+                                     EntrySK);
+    return true;
+  }
+
+private:
+  clang::Sema *sema;
+  const hlsl::ShaderModel *SM;
+  DXIL::ShaderKind EntrySK;
```
tex3d, line 320: We should keep track of the root entry decl for use with a new note which identifies the entry that leads to the call of this method. This may also be used to check for other compatibility constraints, such as numthreads and broadcast launch constraints.  We could gather facts about the root entry once upon construction of this visitor class.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7771,8 +7771,12 @@ def err_hlsl_unsupported_for_version_lower : Error<
    "%0 is only allowed for HLSL %1 and lower.">;
 def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
-def err_hlsl_intrinsic_overload_in_wrong_shader_model : Error<
-   "overload of intrinsic %0 requires shader model %1 or greater">;
+def err_hlsl_derivatives_intrinsic_in_wrong_shader_kind : Warning<
+   "Derivatives intrinsic %0 only works in ps and cs_6.6+/as_6.6+/ms_6.6+/node">,
```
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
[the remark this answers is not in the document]
python3kgae: Fixed.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
python3kgae: Fixed.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11138,25 +11138,79 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+      }
+      switch (EntrySK) {
+      default:
+        break;
+      case DXIL::ShaderKind::Compute:
+      case DXIL::ShaderKind::Amplification:
+      case DXIL::ShaderKind::Mesh: {
+        if (!SM->IsSM66Plus()) {
+          Diags.Report(Loc, diag::warn_hlsl_derivatives_in_wrong_shader_model)
+              << MD->getNameAsString();
```
tex3d, line 11181: Isn't this unreachable due to the `AllowDerivatives` method checking `IsSM66Plus` already?

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -494,6 +494,20 @@ const llvm::StringRef ShaderModel::FullNameFromKind(DXIL::ShaderKind sk) {
...
+bool ShaderModel::AllowDerivatives(DXIL::ShaderKind sk) const {
+  switch (sk) {
+  case DXIL::ShaderKind::Pixel:
+  case DXIL::ShaderKind::Library:
+  case DXIL::ShaderKind::Node:
+    return true;
+  case DXIL::ShaderKind::Compute:
+  case DXIL::ShaderKind::Amplification:
+  case DXIL::ShaderKind::Mesh:
+    return IsSM66Plus();
```
tex3d, line 506: Now, since this checks shader model, but doesn't differentiate result between failure due to kind vs. model, the later code that would emit `warn_hlsl_derivatives_in_wrong_shader_model` is unreachable.  Perhaps now we should rename this function to AllowDerivativesInShaderKind, and there's no need for it to be a method on shader model.

Alternatively, we could encapsulate all the required logic in one diagnostic function (not as a method of ShaderModel), which can be called from the switch on derivative ops in `DiagnoseUsedHLSLMethodCall`.

If this logic is needed outside `SemaHLSLDiagnoseTU.cpp`, like for reuse in DxilValidation, then it could be given extracted values (such as for NodeLaunch and numthreads), located here, and could return an enum for the source of incompatibility with derivatives (translated to a diagnostic by the caller).

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
tex3d: Looks good, however, we still don't have a note pointing to an entry that was incompatible with the call, which would be useful when the call graph is deeper and there are multiple entries.

tools/clang/lib/Sema/SemaHLSL.cpp
python3kgae: Moved AllowDerivatives check into the switch.

lib/DXIL/DxilShaderModel.cpp
python3kgae: Moved AllowDerivatives check into the switch.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7772,10 +7772,10 @@ def err_hlsl_unsupported_for_version_lower : Error<
 def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
 def warn_hlsl_derivatives_in_wrong_shader_kind: Warning<
-   "Intrinsic %0 requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders.">,
+   "Intrinsic %0 called by %1 requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders.">,
```
tex3d, line 7775: I feel like ", potentially used by %1, " might be more appropriate than "called by", since called by may not be technically accurate.

python3kgae: Updated.

tex3d (approved):
I think we'll want to eventually move the derivatives requirements checks into a separate function, to simplify `DiagnoseUsedHLSLMethodCall` when more checks are needed, but this is fine for now.

tools/clang/include/clang/Sema/Sema.h
llvm-beanz: @python3kgae, were you going to update this name?

python3kgae: Fixed.

PR #6145, opened 2024-01-10
[SPIR-V] Handle vk::ext_capability and vk::ext_extension on entry point

cassiebeckley:
`ext_capability` and `ext_extension` were previously only handled when translating a CallExpr. Since the entry point is never called, they were ignored when attached to it.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -13196,6 +13196,10 @@ bool SpirvEmitter::emitEntryFunctionWrapper(const FunctionDecl *decl,
   auto *entryLabel = spvBuilder.createBasicBlock();
   spvBuilder.setInsertPoint(entryLabel);
 
+  // Handle vk::execution_mode, vk::ext_extension and vk::ext_capability
+  // attributes. Uses pseudo-instructions for extensions and capabilities, which
+  // are added to the beginning of the entry basic block, so must be called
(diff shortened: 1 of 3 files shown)


sudonatalie (approved):
Looks good to me!

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -13191,6 +13196,8 @@ bool SpirvEmitter::emitEntryFunctionWrapper(const FunctionDecl *decl,
   auto *entryLabel = spvBuilder.createBasicBlock();
   spvBuilder.setInsertPoint(entryLabel);
 
+  processInlineSpirvAttributes(decl);
+
```
s-perron, line 13200: Why did you move the call to this function? I'm fine with it here, but if it needs to be called after something else happens, then we should document it with a comment.

s-perron (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: Sure, I added a comment:

> Handle vk::execution_mode, vk::ext_extension and vk::ext_capability attributes. Uses pseudo-instructions for extensions and capabilities, which are added to the beginning of the entry basic block, so must be called after the basic block is created and insert point is set.

s-perron (approved):

PR #6146, opened 2024-01-11
Add ShaderFeatureInfo flags for SM6.8 comparison sampling / extended command info

jenatali:
These will be optional in shader model 6.8, so we'll need to detect their usage and set these flags accordingly, so that D3D can validate against driver capabilities. This change simply reserves the bits; the actual detection to set them in the shader still needs to be done.


jenatali:
Closing since this superseded by the actual implementation.

PR #6152, opened 2024-01-15
Add testing infrastructure to test SPIR-V codegen from lib call

s-perron:
We recently removed the FileTest infrstructure because we were able to
replace all of the tests with LIT tests. However, we now have to test
the a case were the source code is in memory and not in a file (see
`FileTest`, buti specializes it so that it cannot be given a file name, and
remove the uses of EFFCEE.

One test is added as an example on how to use a REGEX matcher to check
the assemble output.

Other options we considers:

1. Adding new infrastructure based on the code in HLSL/CompilerTest.cpp.
   However that was more complex that we need. I wanted a simple
   interface so that it is easy for people to add new tests. I also wanted
   to make it spir-v specific.
2. Modifying the DXC executable to not pass along the file name. This
   was rejected because it would place an extra option, just for
   testing, the public facing executable.
3. Have the dxc executable accept `-` at the input file to have it read
   the input file from stdin. Similar to 2. Rejected for similar
   reasons.


sudonatalie (approved):

tools/clang/unittests/SPIRV/LibTestUtils.cpp
```diff
@@ -0,0 +1,228 @@
...
+  std::vector<std::string> tokens(start, end);
+  if (tokens.size() < 3 || tokens[1].find("RUN") == std::string::npos ||
+      tokens[2].find("%dxc") == std::string::npos) {
+    fprintf(stderr, "The only supported format is: \"// RUN: %%dxc -T "
+                    "<profile> -E <entry>\"\n");
+    return false;
+  }
+
+  std::ostringstream rest;
+  for (uint32_t i = 3; i < tokens.size(); ++i) {
```
Keenuts, line 77: maybe use `spvParseTargetEnv`?

tools/clang/unittests/SPIRV/LibTestUtils.cpp
```diff
@@ -0,0 +1,228 @@
...
+                                    const llvm::StringRef entryPoint,
+                                    const llvm::StringRef targetProfile,
+                                    const std::vector<std::string> &restArgs,
+                                    std::vector<uint32_t> *generatedBinary,
+                                    std::string *errorMessages) {
+  std::wstring srcFile(L"");
+  std::wstring entry(entryPoint.begin(), entryPoint.end());
+  std::wstring profile(targetProfile.begin(), targetProfile.end());
+
+  std::vector<std::wstring> rest;
```
Keenuts, line 140: Isn't it possible to use:
```cpp
std::vector<std::wstring> rest(restArgs.begin(), restArgs.end())
```
? Or is the wstring to string conversion causing issues?

tools/clang/unittests/SPIRV/LibTestUtils.cpp
```diff
@@ -0,0 +1,228 @@
...
+    CComPtr<IDxcCompiler> pCompiler;
+    CComPtr<IDxcOperationResult> pResult;
+    CComPtr<IDxcBlobEncoding> pErrorBuffer;
+    CComPtr<IDxcBlob> pCompiledBlob;
+    CComPtr<IDxcIncludeHandler> pIncludeHandler;
+    HRESULT resultStatus;
+
+    bool requires_opt = false;
+    for (const auto &arg : rest)
+      if (arg == L"-O3" || arg == L"-O0" || arg.substr(0, 8) == L"-Oconfig")
```
Keenuts, line 163: What about `-O1`, `-O2`?
Also, why is `-fcgl` added if no `-Ox` is specified? Wouldn't that change the default behavior we observe vs invoking the commandline (where `-O3` is the default?)
Why not forwarding those arguments directly?

tools/clang/unittests/SPIRV/LibTestUtils.cpp
```diff
@@ -0,0 +1,228 @@
...
+    flags.push_back(L"-T");
+    flags.push_back(profile.c_str());
+    flags.push_back(L"-spirv");
+    // Disable legalization and optimization for testing, unless the caller
+    // wants to run a specific optimization recipe (with -Oconfig).
+    if (!requires_opt)
+      flags.push_back(L"-fcgl");
+    // Disable validation. We'll run it manually.
+    flags.push_back(L"-Vd");
+    for (const auto &arg : rest)
```
Keenuts, line 181: Shall the logic be more "forward the given arguments as-is"?
(and adding `-Vd` to be able to manually validate later on)

tools/clang/unittests/SPIRV/LibTestUtils.cpp
```diff
@@ -0,0 +1,228 @@
...
+    IFT(pResult->GetErrorBuffer(&pErrorBuffer));
+    const std::string diagnostics((char *)pErrorBuffer->GetBufferPointer(),
+                                  pErrorBuffer->GetBufferSize());
+    *errorMessages = diagnostics;
+
+    if (SUCCEEDED(resultStatus)) {
+      CComPtr<IDxcBlobEncoding> pStdErr;
+      IFT(pResult->GetResult(&pCompiledBlob));
+      convertIDxcBlobToUint32(pCompiledBlob, generatedBinary);
+      success = true;
```
Keenuts, line 213: return true?

tools/clang/unittests/SPIRV/LibTestUtils.cpp
```diff
@@ -0,0 +1,228 @@
...
+      success = false;
+    }
+  } catch (...) {
+    // An exception has occurred while running the compiler with SPIR-V
+    // Generation
+    success = false;
+  }
+
+  return success;
+}
```
Keenuts, line 224:
```suggestion
      return true;
    }
  } catch (...) {
    // An exception has occurred while running the compiler with SPIR-V
    // Generation
  }

  return false;
}
```

tools/clang/unittests/SPIRV/CodeGenSpirvTest.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+namespace {
+using clang::spirv::LibTest;
+
+using ::testing::ContainsRegex;
+
+// This test is purely for demonstrating how to use `LibTest`, and does not
+// test anything specific in DXC itself.
+TEST_F(LibTest, SourceCodeWithoutFilePath) {
+  const std::string command(R"(// RUN: %dxc -T ps_6_0 -E PSMain -Zi)");
```
Keenuts, line 22: Why keep the `// RUN:` pattern?
Is that to read filechecks tests and feed them as-is? Or would this be designed to only be called in tests like that?

If so, shall the command be passed to `compileCodeAndGetSpirvAsm(` in its own argument, saving us form the parsing logic?

What about an interface like:
```
compileCodeAndGetSpirvAsm("<shader-code>", "-Tcs_6_0 -Emain -fcgl -fspv-target-env=vulkan1.2")
```

s-perron: Even with that you will still have to keep the logic to part the options, which is most of the parsing code. 

I did it this way because I simply brought back the old code, and updated it to not use effcee. By minimizing the changes, I lower the chances of introducing a bug.

To me this is 6 of one and half a dozen of another. If you really insist I can change it, but I don't think either way is better or worse.

Also, this infra is to only be used when the test cannot be tested using the command line. We have had two cases for this in the past 6 years. I do not expect there to be many others, so do not want to put too much work into this.

tools/clang/unittests/SPIRV/LibTestFixture.h
```diff
@@ -0,0 +1,71 @@
...
+    Success, // Success (with or without warnings) - check disassembly
+    Warning, // Success (with warnings) - check warning message
+    Failure, // Failure (with errors) - check error message
+  };
+
+  LibTest()
+      : targetEnv(SPV_ENV_VULKAN_1_0), beforeHLSLLegalization(false),
+        glLayout(false), dxLayout(false), scalarLayout(false) {}
+
+  void setBeforeHLSLLegalization() { beforeHLSLLegalization = true; }
```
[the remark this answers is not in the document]
s-perron: What would the documentation be if not repeating the function name? That is why the google style guide says the it is [not needed when it is obvious](https://google.github.io/styleguide/cppguide.html#Function_Comments:~:text=when%20it%20is%20non%2Dobvious). I know we do not follow the google style guide, but I think the same applies.

tools/clang/unittests/SPIRV/LibTestUtils.cpp
s-perron: That does not compile because there is not constructor for std::wstring that takes an std::string, which is why `arg` is not passed directly to `emplace_back` below, but we use `begin` and `end`.

tools/clang/unittests/SPIRV/LibTestUtils.cpp
s-perron: This was the existing behaviour. I was looking to change as little as possible from what was there before this code was removed.

tools/clang/unittests/SPIRV/CodeGenSpirvTest.cpp
s-perron: For this PR, I just want to resurrect the code that was removed with as few changes as possible. If you want to redesign it you can do that in a separate PR, but I don't think it is worth it.

tools/clang/unittests/SPIRV/LibTestFixture.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  checkTestResult(compileOk, expect, runValidation);
+  return generatedSpirvAsm;
+}
+
+void LibTest::checkTestResult(const bool compileOk, Expect expect,
+                              bool runValidation) {
+  generatedSpirvAsm = "";
+  if (expect == Expect::Failure) {
+    EXPECT_FALSE(compileOk);
```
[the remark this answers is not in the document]
s-perron: I added a return afterwards. ASSERT_ only return if the condition fails, so I need a return statement either way to removing the nesting for the else.

tools/clang/unittests/SPIRV/LibTestUtils.cpp
s-perron: Again, I will keep the original behaviour. If we want to change it, it can be done in another PR. 

The reason was probably that we should default to the behaviour that we want most often, which is -fcgl for testing.

tools/clang/unittests/SPIRV/LibTestUtils.cpp
s-perron: For test, you want to caller to deal with validation so that it can parse the errors if they are expected. Again, this is bringing back the old code with as few changes as possible. If we want to change this it can be done later.

tools/clang/unittests/SPIRV/LibTestUtils.cpp
s-perron: done

tools/clang/unittests/SPIRV/CodeGenSpirvTest.cpp
Keenuts: Right, I see the logic, and I'm fine with it.

tools/clang/unittests/SPIRV/LibTestFixture.h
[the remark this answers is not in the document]
Keenuts: The issue IMO is not we don't understand what the function is doing, but more what is it used for.
If a contributor adds a test today, they'd need to look at this option, see it's being forwarded to `spirv-opt`, call `--help` on it, just to be told:
```
Forwards this option to the validator.  See the validator help for details.
```

Then call `--help` on `spirv-val` to be finally told:
```
Allows code patterns that are intended to be fixed by spirv-opt's legalization passes.
```

This line is exactly the comment I'd see valuable for this function. So we know what the desired effect is.

Keenuts (approved):

tools/clang/unittests/SPIRV/LibTestFixture.h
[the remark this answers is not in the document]
s-perron: Now I see what you mean. Those types of comments belong on the variable that the setter/getter accesses. I added them to the variables.

s-perron:
I thought I pushed the changes before resolving the last comment, but I did not. Automerge did its thing before I could push. I opened https://github.com/microsoft/DirectXShaderCompiler/pull/6215 to fix the comments on the variables.

PR #6155, opened 2024-01-16
Update DXIL.rst

bharadwajy:
Fix description of DXIL log operator.

### DIFF
--- a/docs/DXIL.rst
+++ b/docs/DXIL.rst
@@ -2721,7 +2721,7 @@ Loads the value from shader input
 Log
 ~~~
 
-Returns log base 2. Note that hlsl log intrinsic returns natural log.
+Returns log base e. Note that hlsl log intrinsic returns natural log.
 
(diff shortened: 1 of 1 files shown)


docs/DXIL.rst
```diff
@@ -2721,7 +2721,7 @@ Loads the value from shader input
 Log
 ~~~
 
-Returns log base 2. Note that hlsl log intrinsic returns natural log.
```
damyanp, line 2724: Are you sure?  I notice that the "Note that hlsl log instrinsic returns natural log" is there to suggest that the "log 2" is surprising, but correct.

dmpots: I think the  docs are correct here and the DXIL log is log_2.

If you compile a shader with `log(x)` you get

```
  %2 = call float @dx.op.unary.f32(i32 23, float %1)  ; Log(value)
  %3 = fmul fast float %2, 0x3FE62E4300000000
```

The `0x3FE62E4300000000` is `0.693147` which is log_e(2).

Issue #5751, opened 2023-09-19
[SPIR-V] Variable Templates and `using` in global scope not implemented


devshgraphicsprogramming:
**Description**
DXIL emission implements there just fine:
https://godbolt.org/z/638TrrhWe
obviously emitting warnings

**Steps to Reproduce**
Adding `-spirv` to the commandline breaks both template using aliases
```cpp
template <bool val>
using bool_constant = integral_constant<bool, val>;
```
and variable templates
```cpp
template <class, class>
static const bool is_same_v = false;

template <class T>
static const bool is_same_v<T, T> = true;
```
https://godbolt.org/z/5ooerq9bY

**Actual Behavior**
Compiler emits errors

**Environment**
- DXC version: Latest Godbolt trunk
- Host Operating System: Linux

devshgraphicsprogramming:
@sudonatalie is this already in current `master` ?

sudonatalie:
No, this issue hasn't been addressed and we can't provide an estimate for when it will be looked at unfortunately.

devshgraphicsprogramming:
> No, this issue hasn't been addressed and we can't provide an estimate for when it will be looked at unfortunately.

thx, I think you told me on another issue that `Done` in HLSL Triage means triaged, not resolved or fixed.

cassiebeckley:
I'll likely be looking at this as part of https://github.com/microsoft/hlsl-specs/blob/main/proposals/0011-inline-spirv.md#types.

devshgraphicsprogramming:
> I'll likely be looking at this as part of https://github.com/microsoft/hlsl-specs/blob/main/proposals/0011-inline-spirv.md#types.

already using the new builtin syntax, looking forward to this next improvement for the types

Issue #5554, opened 2023-08-05
[Feature & Bug Repor] Add support for Scoped Enum (`enum class`)


devshgraphicsprogramming:
**Is your feature request related to a problem? Please describe.**

Thanks to HLSL2021 I can share enumeration declarations between C++ and HLSL202x.

**Bug Report:** However whenever I declare the `enum class` HLSL does not complain until I actually try to use the scoped enum. Seems that `enum class` passes through your tokenizer unharmed, but then the rest of the logic is gone.

**Describe the solution you'd like**

For the compiler to support scope enum lookup `enum_t::` without errors.

**Describe alternatives you've considered**

Not using scoped enums but a lot of my C++ code I want to expose to HLSL already does, so not fun.

**Additional context**

Here's a repro:
https://godbolt.org/z/1TdYh9E1a

llvm-beanz:
I’ve transferred this issue over to DXC because this is a bug not a language feature request.

devshgraphicsprogramming:
hmm it works with correct `static_cast` substitute of a C-style cast
https://godbolt.org/z/M6Kna6r7s

The only annoying thing is that DXC doesn't recognize `enum class` as being something you can template on
https://godbolt.org/z/8hqrj1ezr

devshgraphicsprogramming:
btw also seems that concepts like `integral_constant<Enum,Enum Val>` are busted
https://godbolt.org/z/EGaesxvE1

llvm-beanz:
Oops... wrong window. Sorry for the false hope.

PR #6156, opened 2024-01-17
[SPIR-V] Implement SpirvType and SpirvOpaqueType

cassiebeckley:
Implements hlsl-specs proposal 0011, adding `vk::SpirvType` and `vk::SpirvOpaqueType` templates which allow users to define and use SPIR-V level types.


tools/clang/lib/SPIRV/ConstEvaluator.h
s-perron, line 1: I can't tell if you have added new functionality or only moved it from SpirvEmitter. Could you make the refactoring a separate PR? This is a fairly large PR, so separating it into smaller one where possible would be nice.

s-perron (approved):

s-perron (changes requested):
Generally look okay to me. I have a request for a few extra comments, and if you can split out the ConstEvaluator refactor into a separate PR, that would be nice.

Also @llvm-beanz should look at this as well (or someone else for MS) since you are making significant chagnes to SEMA.

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -1241,6 +1241,26 @@ CXXRecordDecl *hlsl::DeclareNodeOrRecordType(
   return Builder.getRecordDecl();
 }
 
+#ifdef ENABLE_SPIRV_CODEGEN
+CXXRecordDecl *hlsl::DeclareInlineSpirvType(clang::ASTContext &context,
+                                            clang::DeclContext *declContext,
+                                            llvm::StringRef typeName,
+                                            bool opaque) {
+  // template<uint opcode> vk::SpirvType { ... }
```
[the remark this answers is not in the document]
cassiebeckley: I believe I was following the format of the other methods in this file. This comment basically just shows the syntax of the type declaration that is added to the AST. I forgot to update the comment when adding size and alignment – I've now updated it.

tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
```diff
@@ -604,10 +627,115 @@ const SpirvType *LowerTypeVisitor::lowerType(QualType type,
...
+      /* PrevDecl */ nullptr);
+  QualType type = astContext.getTypeDeclType(specialized);
+
+  return type;
+}
+
+const SpirvType *LowerTypeVisitor::lowerVkTypeInVkNamespace(
+    QualType type, llvm::StringRef name, SpirvLayoutRule rule,
+    llvm::Optional<bool> isRowMajor, SourceLocation srcLoc) {
+  if (name == "SpirvType" || name == "SpirvOpaqueType") {
```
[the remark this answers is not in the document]
cassiebeckley: Sure, done.

s-perron (approved):

tools/clang/lib/Sema/SemaTemplate.cpp
```diff
@@ -3983,8 +3983,27 @@ bool Sema::CheckTemplateArgumentList(TemplateDecl *Template,
...
+    // TODO(6246): this breaks if the frontend tries to desugar the type
+    NamedDecl *TemplatedDecl = Template->getTemplatedDecl();
+    llvm::StringRef TemplateName = TemplatedDecl->getName();
+    llvm::StringRef Namespace = "";
+    if (const auto *ND =
+            dyn_cast<NamespaceDecl>(TemplatedDecl->getDeclContext()))
+      Namespace = ND->getName();
+
+    if (Namespace != "vk" ||
+        (TemplateName != "SpirvType" && TemplateName != "SpirvOpaqueType"))
```
llvm-beanz, line 4001: I'm a little nervous about this. Why is this needed? Can't we just have default values for the additional template arguments?

utils/hct/gen_intrin_main.txt
```diff
@@ -393,6 +393,7 @@ void [[]] RawBufferStore(in u64 addr, in $funcT value);
 void [[]] RawBufferStore(in u64 addr, in $funcT value, in uint alignment);
 void [[]] ext_execution_mode(in uint mode, ...);
 void [[]] ext_execution_mode_id(in uint mode, ...);
+void [[]] ext_literal(in any<> value);
```
llvm-beanz, line 396: This also feels odd and unwieldy. Don't we know that the template parameters need to be literals, and that they need to be encoded as immediate values? Why do we need users to specify it this way?

llvm-beanz: Is that not something we should capture as a type requirement rather than an intrinsic? If we use an intrinsic to act as a cast, it seems like it should produce a value with a type annotation rather than returning void where we implicitly "look through" to the argument.

s-perron: We used the intrinsic because it already existed for this purpose.

I'm not sure what you mean by "type requirment", but we could define a new type `vk::Literal` and then cast the HLSL literal to that type. Is that what you are thinking about?

tools/clang/lib/Sema/SemaTemplate.cpp
cassiebeckley: There's currently no way to represent `vk::SpirvType` in HLSL syntax or in the AST. The closest representation would probably be something like

```hlsl
template <uint OpCode, uint size, uint alignment, typename... args>
struct SpirvType {}
```

However, that would only allow us to pass types to `SpirvType`, whereas we would like to be able to pass either types or values of any type. So instead this PR defines `SpirvType` as

```hlsl
template <uint OpCode, uint size, uint alignment>
struct SpirvType {}
```

and modifies Sema to ignore any additional arguments. This actually doesn't work for every situation where someone might reasonably want to use `SpirvType`, so I've created #6246 as a follow-up issue.

The options to fix it will likely be either to add `typename... args` to the `SpirvType` template and specifically disable validation for that parameter pack, or to add a new type of parameter pack which will allow `SpirvType` to be accurately represented in the AST.

cassiebeckley:
@sudonatalie @llvm-beanz I've updated this PR based on the design changes to 0011.

tools/clang/test/CodeGenSPIRV/spv.inline.type.enum-class.hlsl
```diff
@@ -0,0 +1,39 @@
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv | FileCheck %s
```
sudonatalie, line 1: Did you mean to clean up the TODOs in this test?

tools/clang/test/CodeGenSPIRV/spv.inline.type.hlsl
```diff
@@ -0,0 +1,32 @@
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv | FileCheck %s
+
+// TODO(5751): enable Array test when `using` has been implemented
```
sudonatalie, line 3: #5751 is closed now, can these be uncommented?

tools/clang/test/CodeGenSPIRV/spv.inline.type.enum-class.hlsl
cassiebeckley: Hm, this is something we should support once #5554 is fixed. Do you think it would be better to commit this with the TODOs, or to simply remove the test altogether?

tools/clang/test/CodeGenSPIRV/spv.inline.type.hlsl
cassiebeckley: Hm, this still doesn't work completely. Something to do with how Texture2D is something of a special case, I think.

tools/clang/test/CodeGenSPIRV/spv.inline.type.enum-class.hlsl
sudonatalie: Hm, that's unfortunate. If we're going to keep the test in this PR then let's leave the TODO but uncomment the test itself and make it expected to fail, so that if/when #5554 is fixed it will get caught for an update.

tools/clang/test/CodeGenSPIRV/spv.inline.type.hlsl
cassiebeckley: I'm *really* not sure what's causing this. It seems to be an error from `SemaHLSL.cpp`, but it doesn't happen for non-SpirvType code (https://godbolt.org/z/KeoGMaEqs).

I've filed a bug at #6498. I'll reference the bug in this test.

tools/clang/test/CodeGenSPIRV/spv.inline.type.enum-class.hlsl
cassiebeckley: Done.

sudonatalie:
LGTM

@llvm-beanz Can you take another look?

llvm-beanz (approved):
Sorry for the delay in reviewing this. It looks great to me. Thank you!

Issue #6125, opened 2024-01-04
[Feature Request] Finalize DXIL 1.8 operations


pow2clk:
**Is your feature request related to a problem? Please describe.**
Final placement of the number of opcodes allocated to 1.8 determines which are actually allowed in that version

**Describe the solution you'd like**
When final opcode numbers are determined, set the 1.8 value accordingly.

pow2clk:
Fixed by #6163

PR #6163, opened 2024-01-19
Add DXIL 1.8 op code cap and move WaveMatrix intrisics to SM 6.9

### LINKED ISSUE #6133: WaveMatrix should not be allowed in SM 6.7
**Description**
We currently allow WaveMatrix types in SM 6.7 shaders.  This is an experimental feature.
See: https://github.com/microsoft/DirectXShaderCompiler/blob/main/utils/hct/hctdb.py#L649

**Steps to Reproduce**
https://godbolt.org/z/633oG497n

**Actual Behavior**
The above shader should produce an error.

**Environment**
- DXC version: Current main
- Host Operating System: All

hekota:
Adds max opcode value for DXIL 1.8 and moves WaveMatrix intrinsics into future shader model 6.9.

Contributes to #6133
Related to #6125

### DIFF
--- a/lib/DXIL/DxilOperations.cpp
+++ b/lib/DXIL/DxilOperations.cpp
@@ -3160,18 +3160,6 @@ void OP::GetMinShaderModelAndMask(OpCode C, bool bWithTranslation,
     minor = 7;
     return;
   }
-  // Instructions: WaveMatrix_Annotate=226, WaveMatrix_Depth=227,
-  // WaveMatrix_Fill=228, WaveMatrix_LoadRawBuf=229,
-  // WaveMatrix_LoadGroupShared=230, WaveMatrix_StoreRawBuf=231,
-  // WaveMatrix_StoreGroupShared=232, WaveMatrix_Multiply=233,
-  // WaveMatrix_MultiplyAccumulate=234, WaveMatrix_ScalarOp=235,
-  // WaveMatrix_SumAccumulate=236, WaveMatrix_Add=237
-  if ((226 <= op && op <= 237)) {
-    major = 6;
-    minor = 7;
-    mask = SFLAG(Library) | SFLAG(Compute);
-    return;
-  }
   // Instructions: QuadVote=222
   if (op == 222) {
     if (bWithTranslation) {
@@ -3217,6 +3205,18 @@ void OP::GetMinShaderModelAndMask(OpCode C, bool bWithTranslation,
     major = 6;
     minor = 8;
     mask = SFLAG(Vertex);
+    return;
+  }
+  // Instructions: WaveMatrix_Annotate=226, WaveMatrix_Depth=227,
+  // WaveMatrix_Fill=228, WaveMatrix_LoadRawBuf=229,
+  // WaveMatrix_LoadGroupShared=230, WaveMatrix_StoreRawBuf=231,
+  // WaveMatrix_StoreGroupShared=232, WaveMatrix_Multiply=233,
+  // WaveMatrix_MultiplyAccumulate=234, WaveMatrix_ScalarOp=235,
+  // WaveMatrix_SumAccumulate=236, WaveMatrix_Add=237
+  if ((226 <= op && op <= 237)) {
+    major = 6;
+    minor = 9;
+    mask = SFLAG(Library) | SFLAG(Compute);
     return;
   }
   // OPCODE-SMMASK:END
--- a/tools/clang/test/HLSLFileCheck/hlsl/objects/WaveMatrix/lit.local.cfg
+++ b/tools/clang/test/HLSLFileCheck/hlsl/objects/WaveMatrix/lit.local.cfg
@@ -1,2 +1,0 @@
-# HLSL Change - Disable WaveMatrix tests, shader model 6.9 is not enabled yet
-config.unsupported = True
--- a/tools/clang/test/HLSLFileCheck/hlsl/objects/WaveMatrix/lit.local.cfg
+++ b/tools/clang/test/HLSLFileCheck/hlsl/objects/WaveMatrix/lit.local.cfg
@@ -1,2 +1,2 @@
 # HLSL Change - Disable WaveMatrix tests, shader model 6.9 is not enabled yet
-config.unsupported = False
+config.unsupported = True
--- a/tools/clang/test/HLSLFileCheck/hlsl/objects/WaveMatrix/lit.local.cfg
+++ b/tools/clang/test/HLSLFileCheck/hlsl/objects/WaveMatrix/lit.local.cfg
@@ -1,0 +1,2 @@
+# HLSL Change - Disable WaveMatrix tests, shader model 6.9 is not enabled yet
+config.unsupported = False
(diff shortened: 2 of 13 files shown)


lib/DXIL/DxilOperations.cpp
```diff
@@ -3219,6 +3207,18 @@ void OP::GetMinShaderModelAndMask(OpCode C, bool bWithTranslation,
...
+  // Instructions: WaveMatrix_Annotate=226, WaveMatrix_Depth=227,
+  // WaveMatrix_Fill=228, WaveMatrix_LoadRawBuf=229,
+  // WaveMatrix_LoadGroupShared=230, WaveMatrix_StoreRawBuf=231,
+  // WaveMatrix_StoreGroupShared=232, WaveMatrix_Multiply=233,
+  // WaveMatrix_MultiplyAccumulate=234, WaveMatrix_ScalarOp=235,
+  // WaveMatrix_SumAccumulate=236, WaveMatrix_Add=237
+  if ((226 <= op && op <= 237)) {
+    major = 6;
+    minor = 9;
+    mask = SFLAG(Library) | SFLAG(Compute);
```
pow2clk, line 3219: It wasn't supposed to be part of this change, but these should be supported in node shaders too. Can you add that to the hctdb.py file around line 650?

hekota: Either way, it should probably not be part of this PR as it would need additional tests added.
I have created a new issue to track this: #6165

pow2clk:
Ironically, the failures reveal success. The shaders that were targeting earlier shader models are failing.  It would be nice to have a simple test that verifies those failures.

pow2clk (approved):
Looks good! Thanks for the quick work!

tools/clang/test/HLSLFileCheck/hlsl/objects/WaveMatrix/lit.local.cfg
```diff
@@ -0,0 +1,2 @@
+# HLSL Change - Disable WaveMatrix tests, shader model 6.9 is not enabled yet
+config.unsupported = False
```
tex3d, line 2: Seems like unsupported should be `True`?  Or am I reading this wrong?

hekota: Yes, it should!

hekota: I was testing this with `hcttest -spirv` and that somehow succeeded, and it did not seem to use lit at all. I'll verify again.

pow2clk: Actually, I realized this isn't needed here as these tests aren't run through lit (yet). So we can remove this file entirely. It is harmless though.

pow2clk:
I changed the description from "fixes" to "contributes to" in order to keep the bug open since we'd still like a test for the original failure of allowing these in earlier shader models.

llvm-beanz (approved):

tex3d (approved):
Looks good!

MarijnS95:
Hi! Is there any possible way to test the new `cs_6_9` profile now? Judging from the commit [Remove lit.local.config - this directory is not scanned by lit](https://github.com/microsoft/DirectXShaderCompiler/pull/6163/commits/406bfa1511d788660a00c4f639f7f3c5b51492f4) it is not intended to be reachable at all, and I get `error: invalid profile cs_6_9` out of `CGHLSLMS.cpp`, presumably because `hlsl::ShaderModel` is invalid as there is no parser variant for the number `'9'` yet, not even on `main` as of today:

https://github.com/microsoft/DirectXShaderCompiler/blob/ef043e90f3f3c5b2db899380591b5091b05f002e/lib/DXIL/DxilShaderModel.cpp#L300-L314

Is this intended to be reenabled for testing at some point? It is blocking us from upgrading DXC as we have experimental shaders with `WaveMatrix` intrinsics.

hekota:
Sorry @MarijnS95, WaveMatrix support is getting removed in https://github.com/microsoft/DirectXShaderCompiler/pull/6807.

llvm-beanz:
> Hi! Is there any possible way to test the new `cs_6_9` profile now? Judging from the commit [Remove lit.local.config - this directory is not scanned by lit](https://github.com/microsoft/DirectXShaderCompiler/pull/6163/commits/406bfa1511d788660a00c4f639f7f3c5b51492f4)

Just to clarify something here. The `HLSLFileCheck` test directory has never been scanned by LIT. Those tests have a separate runner, and are scanned by the unit test implemented [here](https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/unittests/HLSL/CompilerTest.cpp). It's an odd legacy of the way DXC's Windows-only testing strategy was done.

MarijnS95:
> Sorry @MarijnS95, WaveMatrix support is getting removed in #6807.

@hekota thanks for the update, but is there anywhere I can read up on what this is about? Are the ops deprecated and will they be replaced by a different spec in the future, or just out of scope for DXC at the time?

---

> Just to clarify something here. The `HLSLFileCheck` test directory has never been scanned by LIT. Those tests have a separate runner, and are scanned by the unit test implemented [here](https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/unittests/HLSL/CompilerTest.cpp). It's an odd legacy of the way DXC's Windows-only testing strategy was done.

@llvm-beanz thanks for poinging that out, looks like it was a red herring because the file was only [temporarily introduced](https://github.com/microsoft/DirectXShaderCompiler/pull/6163/commits/53f008c0703b79fe9bc5aefb4922ab07b266673b#diff-0b6ff5c71f0cb36120d61bb211e82707dc5f0de3318734c3b3f20387b12cf471) in this PR.

The thing making sure that the tests don't run at all with the unsupported `_6_9` profile must be https://github.com/microsoft/DirectXShaderCompiler/pull/6807 then.

hekota:
> @hekota thanks for the update, but is there anywhere I can read up on what this is about? Are the ops deprecated and will they be replaced by a different spec in the future, or just out of scope for DXC at the time?

AFAIK the ops are marked reserved. I am not sure if they would be reused in the future or stay reserved forever. @damyanp?

damyanp:
> > @hekota thanks for the update, but is there anywhere I can read up on what this is about? Are the ops deprecated and will they be replaced by a different spec in the future, or just out of scope for DXC at the time?
> 
> AFAIK the ops are marked reserved. I am not sure if they would be reused in the future or stay reserved forever. @damyanp?

It's unclear what the exact future here is at the moment.  However, I would strongly advise against making any assumptions about the meaning of these opcodes being stable.

MarijnS95:
@damyanp thanks, I'm not specifically looking at these opcodes but more about the `WaveMatrix` feature in general. I haven't written this shader code and am completely out of the loop about this experimental feature, and am merely seeking for advice on what to tell my coworkers about this future of their shader and the impact on updating our shader compiler. If this is something that is expected to continue working on the driver end, we'll likely hold off DXC upgrades or include an older DXC to keep using it.

damyanp:
I can't speak for the GPU vendors and their plans for supporting this experimental feature that was never released.

We are continuing to look into possible future features that support the scenarios and hardware that WaveMatrix was intended for, but I would be very surprised if whatever we end up with allows these experimental shaders to continue working without modification.

PR #6166, opened 2024-01-19
Remove SPIRV test case from Sema tests related to DS/HS patch size

hekota:
The tests are for an error message that originates from Sema HLSL, it is a not a spirv-specific code. The `-spirv` test case does not bring any additional value. And since these tests are not marked spirv-only (as they should not be), they are failing when the compiler is built without SPIRV support.


pow2clk:
We're in a hurry to get this in, but we'll talk to @keenuts about it next week to make sure everything is okay on his side.

tex3d (approved):
LGTM!

pow2clk (approved):
Looks good!

PR #6167, opened 2024-01-19
Implement wave size range

bob80905:
Wave Size should be able to take a range of possible wavesizes. This PR aims to implement what is described in this spec:
https://github.com/microsoft/hlsl-specs/pull/149/files


tools/clang/test/HLSLFileCheck/hlsl/workgraph/wavesize_range.hlsl
```diff
@@ -0,0 +1,120 @@
+// RUN: %dxc -T lib_6_8 %s | FileCheck %s
+
+// Check the WaveSize attribute is accepted by work graph nodes
+// and appears in the metadata
```
tex3d, line 4: This looks like diagnostics verification, belonging under a different path using -verify, no?

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1812,11 +1812,10 @@ class DxilRDATWriter : public DxilPartWriter {
...
             const auto &entryProps = DM.GetDxilEntryProps(&function);
-            unsigned waveSize = entryProps.props.waveSize;
-            if (waveSize) {
-              pInfo2->MinimumExpectedWaveLaneCount = waveSize;
-              pInfo2->MaximumExpectedWaveLaneCount = waveSize;
-            }
+            pInfo2->MinimumExpectedWaveLaneCount = entryProps.props.waveMinSize;
+            pInfo2->MaximumExpectedWaveLaneCount =
+                entryProps.props.waveMaxSize > 0 ? entryProps.props.waveMaxSize
+                                                 : entryProps.props.waveMinSize;
```
tex3d, line 1818: Ok, but missing use of `m_Module.GetWaveSize()` for actually filling in the PSV0 Min/Max values.  That's a good reason that method needed updating or removal.

tex3d: By this, I mean look around line 944 for where PSV0 is filled in.  That needs to be updated as well.

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
```diff
@@ -0,0 +1,126 @@
...
+[WaveSize(4, 16, 32)] /* expected-error{{Preferred WaveSize value 32 must be between 4 and 16}} */
+void node01(DispatchNodeInputRecord<INPUT_RECORD> input) { }
+
+
+[Shader("node")]
+[NodeLaunch("broadcasting")]
+[NumThreads(1,1,1)]
+[NodeMaxDispatchGrid(32,1,1)]
+/* expected-error@+2{{Preferred WaveSize value 32 must be between 16 and 16}} */
+/* expected-error@+1{{Minimum WaveSize value 16 must be less than Maximum WaveSize value 16}} */
```
tex3d, line 27: I find this error ("Minimum WaveSize value 16 must be less than Maximum WaveSize value 16") confusing and misleading.  Min==Max is a valid degenerate case.  I opened an [issue](https://github.com/microsoft/hlsl-specs/issues/157) about this in order to resolve a comment in the spec proposal, but it seems to have been ignored.

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
```diff
@@ -0,0 +1,126 @@
+// RUN: %dxc -T lib_6_8 -verify %s
+
+// Check the WaveSize attribute is accepted by work graph nodes
+// and appears in the metadata
```
tex3d, line 4: This comment isn't correct, since we are checking diagnostics for error cases.

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
tex3d: Also: this doesn't check error cases of WaveSize with default preferred (`[WaveSize(min,max)]` form), or passing a `0` value explicitly in different combinations of parameters.

I would prefer we allow and translate degenerate cases into the canonical form.  Pseudo code:
- if max==min, set max=0
- if max==0 && preferred!=0:
  - if preferred==min, set preferred=0, else error.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12620,21 +12620,39 @@ HLSLWaveSizeAttr *ValidateWaveSizeAttributes(Sema &S, Decl *D,
                                              const AttributeList &A) {
   // validate that the wavesize argument is a power of 2 between 4 and 128
   // inclusive
-  HLSLWaveSizeAttr *pAttr = ::new (S.Context)
-      HLSLWaveSizeAttr(A.getRange(), S.Context, ValidateAttributeIntArg(S, A),
-                       A.getAttributeSpellingListIndex());
+  HLSLWaveSizeAttr *pAttr = ::new (S.Context) HLSLWaveSizeAttr(
```
llvm-beanz, line 12623: We will need to emit an error if the WaveSize attribute has more than one argument and the shader model target is less than 6.8.

tex3d: We don't need to do this now.  Besides, it must be checked based on active entry points.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -941,10 +941,11 @@ class DxilPSVWriter : public DxilPartWriter {
       break;
     }
     case ShaderModel::Kind::Compute: {
-      UINT waveSize = (UINT)m_Module.GetWaveSize();
-      if (waveSize != 0) {
-        pInfo->MinimumExpectedWaveLaneCount = waveSize;
-        pInfo->MaximumExpectedWaveLaneCount = waveSize;
+      UINT waveMinSize = (UINT)m_Module.GetMinWaveSize();
+      UINT waveMaxSize = (UINT)m_Module.GetMaxWaveSize();
```
tex3d, line 945: Ok, now we have another ordering of the words, Wave, Size, and Min/Max/Preferred.  We can fix this later, since it's internal, but I'd prefer we use WaveSize(Min/Max/Preferred), and even better, roll these into a struct that you can access directly.  Then struct can have methods, like HasRange(), which abstracts the degenerate encoding away from each bit of code that has to deal with this.

tools/clang/test/SemaHLSL/attributes/wavesize_range.ll
```diff
@@ -0,0 +1,51 @@
+; RUN: %dxv %s | FileCheck %s
```
tex3d, line 1: Validation tests need to be under `tools/clang/test/HLSLFileCheck/validation` for now because there's no way to predicate certain tests on validator version here.
Also, you need to add `%dxilver 1.8 | ` at the beginning of the RUN line.

tools/clang/test/SemaHLSL/ds-error-outputpatch-size.hlsl
```diff
@@ -1,5 +1,4 @@
 // RUN: %dxc -T ds_6_0 -E main -verify %s
-// RUN: %dxc -T ds_6_0 -E main -verify %s -spirv
```
tex3d, line 2: I think merging or rebasing will make it so this diff doesn't show in your PR.

bob80905: That was an accident, will be fixed on next commit.

lib/DXIL/DxilMetadataHelper.cpp
```diff
@@ -1821,9 +1831,18 @@ void DxilMDHelper::LoadDxilEntryProperties(const MDOperand &MDO,
...
     } break;
     case DxilMDHelper::kDxilWaveSizeTag: {
-      DXASSERT(props.IsCS() || props.IsNode(), "else invalid shader kind");
       MDNode *pNode = cast<MDNode>(MDO.get());
-      props.waveSize = ConstMDToUint32(pNode->getOperand(0));
+      props.waveMinSize = ConstMDToUint32(pNode->getOperand(0));
+    } break;
+    case DxilMDHelper::kDxilRangedWaveSizeTag: {
+      // if we're here, we're using the range variant.
+      // Extra metadata is used is SM < 6.8
```
tex3d, line 1839: nit: typo "is used is" where I think you meant "is used if"

tools/clang/test/HLSLFileCheck/validation/wavesize_range2.ll
```diff
@@ -0,0 +1,51 @@
+; RUN: %dxv %s %dxilver 1.8 | FileCheck %s
+
+; This tests the validator on emitting errors for invalid 
+; arguments in the wavesize range attribute for entry point functions.
+; This test tests that when the preferred value lies outside of the given
+; minimum and maximum value range, an error is emitted.
```
tex3d, line 6: This comment was copied from the first case, but doesn't match the test.

tools/clang/test/HLSLFileCheck/validation/wavesize_range3.ll
```diff
@@ -0,0 +1,51 @@
+; RUN: %dxv %s %dxilver 1.8 | FileCheck %s
+
+; This tests the validator on emitting errors for invalid 
+; arguments in the wavesize range attribute for entry point functions.
+; This test tests that when the preferred value lies outside of the given
+; minimum and maximum value range, an error is emitted.
```
tex3d, line 6: This comment was copied from the first case, but doesn't match the test.

tools/clang/test/SemaHLSL/attributes/wavesize_range_valid.hlsl
```diff
@@ -0,0 +1,31 @@
+// RUN: %dxc -T lib_6_8 %s | FileCheck %s
+
+// Check the WaveSize attribute emits the right metadata in DXIL
```
tex3d, line 3: As a `SemaHLSL` test, I would have expected a check for -ast-dump only instead of end-to-end compilation.

For end-to-end compilation, we need to put the test somewhere under HLSLFileCheck where we can avoid running the test when the current validator doesn't support SM 6.8.

tools/clang/test/HLSLFileCheck/validation/wavesize_range.ll
```diff
@@ -0,0 +1,51 @@
+; RUN: %dxv %s %dxilver 1.8 | FileCheck %s
```
tex3d, line 1: This syntax isn't right.  RUN line should look like:
```
; RUN: %dxilver 1.8 | %dxv %s | FileCheck %s
```

tools/clang/test/SemaHLSL/attributes/wavesize_range_lib_6_6.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: %dxc -E main -T lib_6_6 %s -verify
+
+[wavesize(4)]
+[numthreads(1,1,8)]
+[shader("compute")]
+void one() {
+}
+
+[wavesize(4, 8)] // expected-error{{wavesize attribute takes no more than 1 argument}}
```
tex3d, line 9: This doesn't seem right:
1. It's not true: WaveSize can take more than 1 argument.
2. Is no preferred size really an invalid form?

tex3d: Ah, this is because the shader model is 6.6.  In that case, we should report that it requires SM 6.8, not that the attribute doesn't take more than 1 argument.

llvm-beanz: Under SM 6.6?

bob80905: Shouldn't wavesize take exactly one argument given the shader model here is 6.6?

llvm-beanz (approved):

tools/clang/test/SemaHLSL/attributes/wavesize_range_valid.hlsl
bob80905: I moved the test to a place that makes sense for end-to-end compilation.

tex3d (approved):
I think this is good to merge with follow up for some minor internal things later.

PR #6169, opened 2024-01-20
Update WaveMatrix test shader model requirement to 6.9

tex3d:
WaveMatrix is moved out of SM 6.8, so the requirement in the test needs to be updated to SM 6.9 to exclude the test from execution until at least experimental SM 6.9 is supported by the device.


pow2clk (approved):
Looks good! Nice catch!

llvm-beanz (approved):

Issue #157, opened 2017-03-28
Strange #include/file system behaviour


simontaylor81:
I'm getting some strange behaviour with #includeing files from the file system using the dxc commandline app, which are somwhat hard to describe.

First, the repro:

First, download this file, rename .zip (github doesn't like zip files apparently), and extract somewhere. [OddIncludes.txt](https://github.com/Microsoft/DirectXShaderCompiler/files/875286/OddIncludes.txt)
The repro relies on using absolute paths, so I can't just give a command line to run directly, but assuming the zip was extracted to c:\temp, run the following command (assumes dxc.exe is on the path):

```dxc /T ps_6_0 c:\temp\OddIncludes\main.hlsl /I c:\temp```

This should generate an error:

```
c:\temp\OddIncludes\main.hlsl:1:10: fatal error: 'include.hlsl' file not found
#include "include.hlsl"
         ^
```

This is erroneous -- include.hlsl is in the same directory as main.hlsl, so this should work fine. However, you can "fix" this error by changing the case of the directories. So this command works (note capital T):

```dxc /T ps_6_0 c:\temp\OddIncludes\main.hlsl /I c:\Temp```

Ok, so now the explanation. Apologies if this is very confusing! The first thing to note is that the (in this case superfluous) additional include directory specification is critical to the repro. Additional include directories are process before any source files (this matters because the file system has lots of caching). When it tries to get the file info for this path, it ends up in `DxcArgsFileSystem::CreateFileW`. That code does a substring search of the path against the input filename. If it matches (in this case it does), it assumes it is the parent of the source and returns the special handle `SourceParentDirHandle`. This is added to the cache with the path c:\temp.

Later on, we come to look up the directory for including the file. This is _also_ a parent of the source file, and so also matches the substring match, and so also returns `SourceParentDirHandle`. This is then a cache hit, so instead of returning the directory asked for, it returns c:\temp again. Include.hlsl file is not found there, giving the error.

Now, the reason changing the case "fixes" things is that all the substring comparisons in dxcompilerobj.cpp are case sensitive! So by changing the case, the substring comparisons fail, and it falls through to just looking up the file in the file system, and things work correctly.

Hopefully that made some sense! Basically, I think the logic surrounding `SourceParentDirHandle` is broken. I do not know the best way to fix it, though, I'm afraid.

marcelolr:
[...]

PR #6172, opened 2024-01-21
Fix bad WaveSize non-range PSV0 and improve things

tex3d:
PSV0 MaximumExpectedWaveLaneCount was incorrectly set to 0 for non-range.

Create struct for WaveSize in DxilFunctionProps.h.
Centralize encoding and validation logic there.
Use validation logic in both SemaHLSL and DxilValidation.

Remove test requiring newer shader model in CodeGenHLSL.
Add comprehensive test for compute and node, cs and lib targets, SM 6.6 and 6.8, testing ast, metadata and RDAT.
Add PSV0 tests to catch incorrect runtime data.

Update validation rules and test for more cases.


include/dxc/DXIL/DxilFunctionProps.h
```diff
@@ -25,6 +25,93 @@ class Constant;
...
+    if (max == 0 && preferred == min)
+      preferred = 0;
+    return DxilWaveSize(min, max, preferred);
+  }
+
+  // Valid non-zero values are powers of 2 between 4 and 128, inclusive.
+  static bool IsValidValue(unsigned Value) {
+    return (Value >= 4 && Value <= 128 && ((Value & (Value - 1)) == 0));
+  }
+  // Valid encodings:
```
llvm-beanz, line 59: This comment makes it sound like these _encodings_ are valid. IIUC the old wave size didn't encode anything if the values were `0`, I assume we should do the same here.

There is also an outstanding question for the spec about whether or not we're deprecating the old metadata tag or not. I think that impacts which other formulations are valid. I filed this issue to track that:
https://github.com/microsoft/hlsl-specs/issues/160

include/dxc/DXIL/DxilShaderFlags.h
```diff
@@ -160,7 +168,7 @@ class ShaderFlags {
   void SetDerivativesInMeshAndAmpShaders(bool flag) {
     m_bDerivativesInMeshAndAmpShaders = flag;
   }
-  bool GetDerivativesInMeshAndAmpShaders() {
+  bool GetDerivativesInMeshAndAmpShaders() const {
```
llvm-beanz, line 171: This is completely unrelated. Including unrelated changes makes it harder to review.

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -792,3 +795,31 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
 void ShaderFlags::CombineShaderFlags(const ShaderFlags &other) {
   SetShaderFlagsRaw(GetShaderFlagsRaw() | other.GetShaderFlagsRaw());
 }
+
+void ShaderFlags::GetMinimumShaderModelBasedOnFlags(unsigned &major,
```
llvm-beanz, line 799: The introduction of this function introduces a significant and separate behavior change. We really should be doing this separately from the rest of this PR.

include/dxc/DXIL/DxilFunctionProps.h
tex3d: "encodings" is referring to the encoding of different forms *in this structure*.  What would you prefer?

Regarding deprecation of the old metadata, The prior change already made that decision (not to deprecate it), and I already had an issue filed for that question: microsoft/hlsl-specs#158.

We can change the decision, but this PR doesn't change the current state.  Besides, these "encodings" still only pertain to the encoding of various valid states into this structure, so that issue doesn't impact this code - only how to write these states into the metadata in `DxilMetadataHelper.cpp`.

include/dxc/DXIL/DxilShaderFlags.h
tex3d: I'm working on separating this change out.

tex3d: Change removed from this PR.

utils/hct/hctdb.py
```diff
@@ -7656,17 +7656,34 @@ def build_valrules(self):
         self.add_valrule(
             "Sm.TGSMUnsupported", "Thread Group Shared Memory not supported %0."
         )
-        self.add_valrule(
+        self.add_valrule_msg(
             "Sm.WaveSizeValue",
-            "Declared WaveSize %0 outside valid range [%1..%2], or not a power of 2.",
+            "WaveSize value must be a power of 2 in range [4..128]",
+            "WaveSize %0 (%1) outside valid range [%2..%3], or not a power of 2.",
```
bob80905, line 7662: Is there a good reason that we merge these two diagnostic messages into one? I'm just curious why we don't have a diagnostic specifically for when a value is not a power of 2.

include/dxc/DXIL/DxilFunctionProps.h
llvm-beanz: valid "values" or "representations" would be more clear. "Encodings" implies that it is being encoded, not that the values are just stored.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12626,26 +12627,49 @@ HLSLWaveSizeAttr *ValidateWaveSizeAttributes(Sema &S, Decl *D,
...
   pAttr->setSpelledArgsCount(A.getNumArgs());
-  int minWave = pAttr->getMin();
-  int maxWave = pAttr->getMax();
-  int prefWave = pAttr->getPreferred();
 
-  // validate attribute arguments.
-  if (!DXIL::IsValidWaveSizeValue(minWave, maxWave, prefWave)) {
-    S.Diag(A.getLoc(), diag::err_hlsl_wavesize_size)
-        << DXIL::kMinWaveSize << DXIL::kMaxWaveSize;
-  }
```
llvm-beanz, line 12637: I have a strong preference for the way this code was before. The way it is written now I find much harder to read. I understand that this reduces duplicated logic, but it is at the expense of readability.

Feel free to ignore this feedback because I don't feel strongly about it.

include/dxc/DXIL/DxilFunctionProps.h
pow2clk: When writing the spec, I had no intention of deprecating the old tag. In fact, the spec says that the only valid use of the new tag is with at least two values, which this deviates from. I don't really care enough to insist, but that's what is currently specced.

pow2clk: Okay. reread my own words and I see that I didn't actually write that. It's what I was envisioning, but I'm not insistent. I should clarify the spec though.

include/dxc/DXIL/DxilFunctionProps.h
```diff
@@ -25,6 +25,93 @@ class Constant;
...
+  DxilWaveSize &operator=(const DxilWaveSize &other) = default;
+  bool operator==(const DxilWaveSize &other) const {
+    return Min == other.Min && Max == other.Max && Preferred == other.Preferred;
+  }
+
+  // Encode valid DxilWaveSize, handling potential degenerate cases.
+  static DxilWaveSize Encode(unsigned min, unsigned max = 0,
+                             unsigned preferred = 0) {
+    if (max == min)
+      max = 0;
```
pow2clk, line 49: Per the spec, max values aren't allowed to be zero and min must be less than max. This turns one invalid state to another.

include/dxc/DXIL/DxilFunctionProps.h
```diff
@@ -25,6 +25,93 @@ class Constant;
...
+    MaxLessThanMin,
+    PreferredOutOfRange,
+  };
+  ValidationResult Validate() const {
+    if (Min == 0) { // Not defined
+      if (Max != 0 || Preferred != 0)
+        return ValidationResult::MaxOrPreferredWhenUndefined;
+    } else if (!IsValidValue(Min)) {
+      return ValidationResult::InvalidMin;
+    } else if (Max == 0) { // single WaveSize (SM 6.6/6.7)
```
pow2clk, line 80: per the spec, max == 0 should be an error regardless

include/dxc/DXIL/DxilFunctionProps.h
```diff
@@ -25,6 +25,93 @@ class Constant;
...
+    InvalidMax,
+    InvalidPreferred,
+    MaxOrPreferredWhenUndefined,
+    PreferredWhenNoRange,
+    MaxEqualsMin,
+    MaxLessThanMin,
+    PreferredOutOfRange,
+  };
+  ValidationResult Validate() const {
+    if (Min == 0) { // Not defined
```
pow2clk, line 75: Per the spec, min == 0 should be an error regardless

include/dxc/DXIL/DxilFunctionProps.h
```diff
@@ -25,6 +25,93 @@ class Constant;
...
+        return ValidationResult::InvalidPreferred;
+      if (Preferred < Min || Preferred > Max)
+        return ValidationResult::PreferredOutOfRange;
+    }
+    return ValidationResult::Success;
+  }
+  bool IsValid() const { return Validate() == ValidationResult::Success; }
+
+  bool IsDefined() const { return Min != 0; }
+  bool IsRange() const { return Max != 0; }
```
pow2clk, line 100: Per the spec, this tag can only represent ranged wave sizes

pow2clk (commented):
As promised, pointing out where we have spec discrepancies.

include/dxc/DXIL/DxilFunctionProps.h
tex3d: I can replace the word "encodings" in the comment with "representations" if that will satisfy.  I will also have to rename two methods, but now I don't have a good name for them (encode/decode naming no longer works).

All the different states for WaveSize must be "encoded" into the values of this structure in a specific way for it to be valid, including the state where it is undefined.  You could also say they must be "represented" in the values of this structure in a specific way for it to be valid.  It seems to be a minor semantic point.

Also: there are multiple "encodings": The internal representation in this structure, one in llvm metadata, one in PSV0, and one in RDAT.  This does not describe any "encodings" besides the one in this structure.

The spec issue is **not** something that impacts this code, or the comment here.  This isn't referring to just the "encoding" specifically into metadata.  Note: "encodings" for PSV0 and RDAT are different than the one for metadata as well.

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
```diff
@@ -22,7 +22,6 @@ void node01(DispatchNodeInputRecord<INPUT_RECORD> input) { }
 [NodeLaunch("broadcasting")]
 [NumThreads(1,1,1)]
 [NodeMaxDispatchGrid(32,1,1)]
-/* expected-error@+2{{Preferred WaveSize value 32 must be between 16 and 16}} */
```
pow2clk, line 25: I guess we can only produce one error at a time now? Could we get a new case to verify the previous error?

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
```diff
@@ -40,8 +39,7 @@ void node03(DispatchNodeInputRecord<INPUT_RECORD> input) { }
 [NodeLaunch("broadcasting")]
 [NumThreads(1,1,1)]
 [NodeMaxDispatchGrid(32,1,1)]
-/* expected-error@+2{{WaveSize arguments must be between 4 and 128 and a power of 2}} */
-/* expected-error@+1{{Preferred WaveSize value 32 must be between 15 and 17}} */
```
pow2clk, line 44: Here as well

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
```diff
@@ -76,7 +74,9 @@ void node07(DispatchNodeInputRecord<INPUT_RECORD> input) { }
 [NodeLaunch("broadcasting")]
 [NumThreads(1,1,1)]
 [NodeMaxDispatchGrid(32,1,1)]
-[WaveSize(4, 8, node07)] /* expected-error{{'WaveSize' attribute requires an integer constant}} */
+// expected-error@+2{{WaveSize arguments must be between 4 and 128 and a power of 2}}
```
pow2clk, line 77: Conversely, this error seems unhelpful. No one would argue that a function call is between 4 and 128, but it seems an odd thing to assert.

pow2clk (commented):
Sorry it's so long. I have some previous comments to resolve now.

include/dxc/DXIL/DxilFunctionProps.h
pow2clk: Nevermind, I wasn't clear on the use of this struct yet.

lib/DXIL/DxilModule.cpp
```diff
@@ -395,32 +395,16 @@ unsigned DxilModule::GetNumThreads(unsigned idx) const {
...
 
-void DxilModule::SetWaveSize(unsigned size) {
-  DXASSERT(m_DxilEntryPropsMap.size() == 1 && m_pSM->IsCS(),
-           "only works for CS profile");
-  DxilFunctionProps &props = m_DxilEntryPropsMap.begin()->second->props;
-  DXASSERT_NOMSG(m_pSM->GetKind() == props.shaderKind);
-  props.waveMinSize = size;
+DxilWaveSize &DxilModule::GetWaveSize() {
+  return const_cast<DxilWaveSize &>(
+      static_cast<const DxilModule *>(this)->GetWaveSize());
```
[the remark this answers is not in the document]
pow2clk: I guess you're constifying and then deconstifying it, but I don't know why that's needed either.

include/dxc/DXIL/DxilFunctionProps.h
tex3d: > This turns one invalid state to another.

No, it doesn't.  It translates the degenerate cases, which are valid cases, but aren't represented that way in this structure or in metadata.  Truly invalid cases will still result in invalid state, which is validated elsewhere.
Min == Max is the meaning of max = 0, and the correct encoding for PSV0 and RDAT.  Missing this was the reason for the regression (assuming that Max = 0 was correct for PSV0).

And I don't see how this is translating into an invalid state.

You seem to be thinking about this struct as if it *is* the metadata.  It isn't.  It is the internal representation for any possible state for WaveSize.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -941,11 +941,20 @@ class DxilPSVWriter : public DxilPartWriter {
...
+      // DXASSERT(waveSize.IsValid(), "wave size should be valid");
+      DXASSERT(!waveSize.IsDefined() || m_PSVInitInfo.PSVVersion >= 2,
+               "wave size requires SM 6.6 or above");
+      DXASSERT(!waveSize.IsRange() || m_PSVInitInfo.PSVVersion >= 3,
+               "wave size range requires SM 6.8 or above");
+      if (waveSize.IsDefined() && m_PSVInitInfo.PSVVersion >= 2) {
+        unsigned waveSizeMin = 0, waveSizeMax = 0;
+        waveSize.DecodeMinMax(waveSizeMin, waveSizeMax);
+        if (m_PSVInitInfo.PSVVersion < 3)
+          waveSizeMax = waveSizeMin;
```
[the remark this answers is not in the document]
llvm-beanz: We probably should have an assert here to verify that `waveSizeMax == 0`, since I believe anything else shouldn't be possible.

include/dxc/DXIL/DxilFunctionProps.h
tex3d: You seem to be interpreting what this struct is for in a very different way than intended.

It is not a definition of the input attribute, nor is it a definition of a single encoding in dxil metadata, PSV0, or RDAT.  It is an internal canonical representation of all states for WaveSize.

The valid states for this representation include the undefined, single wave size, and range forms.  In this internal canonical representation, Min == 0 means no wave size is defined.

include/dxc/DXIL/DxilFunctionProps.h
tex3d: No, in the internal canonical representation, Max == 0 indicates a single wave size (no range).
This is just as it was in the prior version, but wrapped in a struct.

Validation or diagnostics for additional constraints on other forms exists elsewhere, where it belongs.

In SemaHLSL the rules for the attribute are checked, leveraging this method for commonalities.
For the cases where you have the attribute and used explicit Min, Max, or Preferred == 0, there is explicit code to translate that to invalid Min/Max/Preferred:
https://github.com/microsoft/DirectXShaderCompiler/pull/6172/files#diff-b5ff0eb7c221e9ad11a754e889ea567c35d83ad9f51e89e77ec844ea4fd8eebbR12638

For the case where you have metadata, we could argue whether it's legal to have metadata that explicitly defines a degenerate case, but if we disallow, it means you couldn't use just one structure of metadata for WaveSize and minimally modify the module in the latest metadata form to represent each valid state.  We don't prevent this kind of thing in other cases - we don't force you to omit an extended attribute definition when the value in the definition is equal to the default.  In any case, that discussion doesn't impact this code, but impacts whether DxilValidation should have extra checks for metadata.

There are no changes from the prior version in this regard.

include/dxc/DXIL/DxilFunctionProps.h
tex3d: No, this is not a "tag" this is a state structure that every entry has, which must be able to represent each possible state: no wave size defined, wave size with no range defined, and wave size with range defined.

pow2clk (approved):
Approving with suggestions in the interest of time. Feel free to resolve all my comments to get this in.

lib/DXIL/DxilModule.cpp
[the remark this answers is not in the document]
tex3d: So it's not calling (and recursing into) *this* method.

By defining the const version and casting in this way, we ensure the const version cannot internally modify the `const DxilModule*`.  If we implemented it the other way around, with a `const_cast` on `DxilModule*`, there's no guarantee that the non-const method doesn't modify the `DxilModule` in some way, which would violate the contract for the const method.  Of course, we could just copy paste the contents into both methods, but then you end up with potential divergence and subtle bugs.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -941,11 +941,20 @@ class DxilPSVWriter : public DxilPartWriter {
...
     }
     case ShaderModel::Kind::Compute: {
-      UINT waveMinSize = (UINT)m_Module.GetMinWaveSize();
-      UINT waveMaxSize = (UINT)m_Module.GetMaxWaveSize();
-      if (waveMinSize != 0) {
-        pInfo->MinimumExpectedWaveLaneCount = waveMinSize;
-        pInfo->MaximumExpectedWaveLaneCount = waveMaxSize;
+      const DxilWaveSize &waveSize = m_Module.GetWaveSize();
+      // Can't assert valid or validation tests will assert during assembly
+      // DXASSERT(waveSize.IsValid(), "wave size should be valid");
```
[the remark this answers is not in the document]
tex3d: It's something we might be able to fix and enable the assert for, if/when we move to lit, after we solve the validator versioning problem.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -941,11 +941,20 @@ class DxilPSVWriter : public DxilPartWriter {
...
+      DXASSERT(!waveSize.IsDefined() || m_PSVInitInfo.PSVVersion >= 2,
+               "wave size requires SM 6.6 or above");
+      DXASSERT(!waveSize.IsRange() || m_PSVInitInfo.PSVVersion >= 3,
+               "wave size range requires SM 6.8 or above");
+      if (waveSize.IsDefined() && m_PSVInitInfo.PSVVersion >= 2) {
```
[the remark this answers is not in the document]
tex3d: Asserts won't be enabled on release build, so this prevents us from writing an invalid form into PSV0 for the given version.  We also check the version loaded from metadata in the validator, so we won't pass validation anyway.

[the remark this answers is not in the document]
tex3d: Actually, this raises a point that we are missing another test case for validation: downlevel IR and explicit validator version with WaveSize states that require a higher shader model and validator version.  If we add these tests, we will have to remove (comment out) the DXASSERTS here, for the same reason the first one is commented out.  That is until we move to lit and stop doing `AssembleToContainer` for IR based validation tests, unless done explicitly.

lib/DxilContainer/DxilContainerAssembler.cpp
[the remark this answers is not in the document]
tex3d: @llvm-beanz We have an assert that checks this case above.
https://github.com/microsoft/DirectXShaderCompiler/pull/6172/files#diff-26cf31df90958893d0ad9cfd8ac55fd8a844f8adc643ec7d2843997dfe26ae29R949

However, as my previous comment indicated, I think we have to disable these asserts in order to add a couple more missing validation tests for versioning.

I don't especially think this code should be doing a lot of validation work, which DxilValidation handles.
Perhaps I'll just remove the DXASSERT code and add the new validator test cases for PSV0 versioning.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1811,12 +1821,21 @@ class DxilRDATWriter : public DxilPartWriter {
             payloadSizeInBytes = props.ShaderProps.Ray.paramSizeInBytes;
           }
           shaderKind = (uint32_t)props.shaderKind;
+          waveSize = props.WaveSize;
+          // Can't assert valid or validation tests will assert during assembly
+          // DXASSERT(waveSize.IsValid(), "wave size should be valid");
```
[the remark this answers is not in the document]
tex3d: `DecodeMinMax()` should probably just avoid calling `IsValid()`, so validation only occurs when explicitly called.
I'll make this change.

lib/DXIL/DxilModule.cpp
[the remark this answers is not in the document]
llvm-beanz: Duplicating an accessor with a const and non-const version seems vastly preferable to const_cast. Generally speaking I think that const_cast pretty much always indicates a design flaw, and I think most C++ coding guides agree on that point.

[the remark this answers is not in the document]
pow2clk: It's not a principal I'll die on a hill for, but that's my basic reaction whenever I see a const_cast. An unfortunate necessity at times for sure.

lib/DxilContainer/DxilContainerAssembler.cpp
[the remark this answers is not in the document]
pow2clk: I'll add it to the spec

lib/DxilContainer/DxilContainerAssembler.cpp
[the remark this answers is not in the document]
pow2clk: Sounds good to me. Do you want to do this in this PR or a later one?

utils/hct/hctdb.py
tex3d: They weren't separate before, so I'm didn't separate them in this PR.

For validation, I don't see much value separating them.  If you want separate diagnostics for the normal path, that would require separating them in DiagnosticSemaKinds.td.  Still, I don't see much value in doing that.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12626,26 +12627,49 @@ HLSLWaveSizeAttr *ValidateWaveSizeAttributes(Sema &S, Decl *D,
...
+  if (validationResult == DxilWaveSize::ValidationResult::Success &&
+      !waveSize.IsDefined())
+    validationResult = DxilWaveSize::ValidationResult::InvalidMin;
 
-  if (maxWave != 0 && minWave >= maxWave) {
+  // It is invalid to explicitly specify degenerate cases.
+  if (A.getNumArgs() > 1 && waveSize.Max == 0)
+    validationResult = DxilWaveSize::ValidationResult::InvalidMax;
+  else if (A.getNumArgs() > 2 && waveSize.Preferred == 0)
+    validationResult = DxilWaveSize::ValidationResult::InvalidPreferred;
```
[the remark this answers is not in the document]
tex3d: I believe it is better to keep the HLSL attribute specific rules in Sema, rather than polluting `DxilWaveSize` with these extra rules, only checked when the suggested parameter is supplied, which only occurs when called from Sema.

In any case, this pertains to the open issue #157.  If that is resolved in a different way than currently implemented, these special cases for Max and Preferred would be deleted anyway.

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d: This catches and diagnoses things in a less confusing way than before.  For instance, the previous code would not catch the max==0 case.  And in that case, if preferred was non-zero, the error it would give is that it isn't in range [min..0].  Reviewing the previous code to ensure it catches all the cases correctly was actually harder, because some cases would be blocked almost as a side effect of other things, such as the max==0 cases.

The new approach could also improve the diagnostics a bit without changing the validation logic.  For instance by explicitly identifying degenerate cases that should have been written in different syntax in HLSL, which isn't even hinted at by the diagnostics we have now.  The previous logic collapsed cases like this and gives the confusing diagnostic that min must be less than max, when min==max, instead of saying you must omit the max parameter to indicate this case.  Similar issue with preferred==0.

I didn't change these diagnostics, but they could be improved in the future based on the more specific validation failures available now.  Plus the explicit NumArgs checks could be changed to emit more useful diagnostics, in which case, centralizing these checks in DXILWaveSize::Validate would make even less sense.
 
Here are some cases I notice that the prior code didn't handle well (might not be complete).  These weren't caught in review because it wasn't as straightforward as it looked.
- Sema: wouldn't catch when max is set to 0 explicitly
- DxilValidation: wouldn't catch invalid preferred when max is 0
- DxilValidation: incorrect validation error when max is invalid (would complain about Min)
- DxilValidation: incorrect validation error when preferred is invalid but in range (would complain about Min)

tex3d: One more I just thought of:
- Sema: wouldn't catch when preferred is set to 0 explicitly.

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
tex3d: How useful are multiple errors on the same attribute?

The error coming from validation should highlight the first error that needs fixing.  Other errors might be moot or change once you fix the first issue.

For instance, in this case, you need to omit the max parameter to the attribute, which means preferred must also be omitted.  If we improve the diagnostics for cases like that, emitting additional errors is basically adding noise that distracts from the primary thing that must be addressed.

I could change the validation method to flag multiple issues, but that seems to be of limited value.

Note: We should still add more test cases here, partly because I identified some cases that wouldn't have been caught or handled very well in Sema by the previous version, and partly to make sure we independently test each invalid scenario

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
tex3d: See my other response.

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
tex3d: The only reason this wasn't emitted before is because it didn't catch the explicit preferred==0 case either, so this was due to missing diagnostic coverage.

With that coverage, it's quite a bit more difficult to avoid emitting this specific diagnostic, without some way of stopping diagnostics as soon as something fails when it's checking for int args (which we currently have no way of knowing outside that function).

In any case, along with the `'WaveSize' attribute requires an integer constant`, this is useful because it more specifically defines the allowed values, isn't it?

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
llvm-beanz: If there are multiple legitimately wrong things (as in these examples) printing all the errors allows the user to have the full list of errors after a single compile. That’s definitely useful rather than having to compile multiple times to get and resolve the errors.

utils/hct/hctdb.py
```diff
@@ -7656,17 +7656,34 @@ def build_valrules(self):
...
+        self.add_valrule_msg(
+            "Sm.WaveSizeAllZeroWhenUndefined",
+            "WaveSize Max and Preferred must be 0 when Min is 0",
+            "WaveSize Max (%0) and Preferred (%1) must be 0 when Min is 0",
+        )
+        self.add_valrule_msg(
+            "Sm.WaveSizeMaxAndPreferredZeroWhenNoRange",
+            "WaveSize Max and Preferred must be 0 to encode min==max",
+            "WaveSize Max (%0) and Preferred (%1) must be 0 to encode min==max",
+        )
```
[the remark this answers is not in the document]
tex3d: That is an open issue whether we should prevent the new tag in that case.  I don't think we should prevent the new form for defining any valid WaveSize state because it's simpler.  Currently the validator does not prevent the new form of metadata from being used.

Also, note that the spec proposal was reviewed and approved for getting added, but not all issues were resolved.  So, implementing some things to the letter when decisions haven't been finalized will result in more work in the short run, and may result in more work in the long run as well.  As long as the decisions don't impact things in a way that makes currently valid generated DXIL invalid, we should be good.

tools/clang/test/HLSLFileCheck/hlsl/entry/attributes/wavesize-compute-node.hlsl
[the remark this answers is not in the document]
tex3d: Because the newer infrastructure does not handle the versioning required yet.

Side point: I disagree with "provides significantly more friendly error output on failure", since in my experience, it has been significantly worse for many scenarios.

lib/DxilContainer/DxilContainerAssembler.cpp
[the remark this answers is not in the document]
tex3d: I've changed the way this is handled.  I no longer try to validate here.

lib/DxilContainer/DxilContainerAssembler.cpp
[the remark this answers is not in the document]
tex3d: I've updated the logic to handle encoding here, keep validation elsewhere.

lib/DxilContainer/DxilContainerAssembler.cpp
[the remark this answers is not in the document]
tex3d: I've updated the logic to do validation elsewhere and handle encoding of valid cases here.

tools/clang/test/CodeGenHLSL/attributes_wavesize.hlsl
```diff
@@ -1,11 +0,0 @@
...
-
-// CHECK: @main, !"main", null, null, [[PROPS:![0-9]+]]}
-// CHECK: [[PROPS]] = !{i32 4, [[NT:![0-9]+]], i32 11, [[WS:![0-9]+]]}
-// CHECK: [[NT]] = !{i32 1, i32 1, i32 8}
-// CHECK: [[WS]] = !{i32 32}
-
-[wavesize(32)]
-[numthreads(1,1,8)]
-void main() {
-}
```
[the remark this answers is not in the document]
tex3d: Yes.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
pow2clk: That's fine

tools/clang/test/SemaHLSL/attributes/wavesize_range.hlsl
pow2clk: I don't understand where preferred == 0 comes in. Because the param is invalid, it defaults to 0? Is that it?

tex3d: See where the arg types are checked and the attribute is created here:
https://github.com/microsoft/DirectXShaderCompiler/pull/6172/files#diff-b5ff0eb7c221e9ad11a754e889ea567c35d83ad9f51e89e77ec844ea4fd8eebbL12625

The function `ValidateAttributeIntArg` that checks for int argument type always returns an int.  It just returns `0` if it was invalid.  There's no indication that it emitted an error as far as `ValidateWaveSizeAttributes` is concerned.

The rest of the diagnostic code that looks at this attribute cannot tell the difference between a type error on an argument and an explicit `0` supplied by the user.

lib/DXIL/DxilMetadataHelper.cpp
```diff
@@ -1596,19 +1596,18 @@ MDTuple *DxilMDHelper::EmitDxilEntryProperties(uint64_t rawShaderFlag,
     NumThreadVals.emplace_back(Uint32ToConstMD(props.numThreads[2]));
     MDVals.emplace_back(MDNode::get(m_Ctx, NumThreadVals));
 
-    if (props.waveMinSize != 0) {
-      bool UseRange = props.waveMaxSize != 0;
-      if (UseRange)
+    if (props.WaveSize.IsDefined()) {
+      if (props.WaveSize.IsRange())
```
pow2clk, line 1600: This conditional should be worked into the assert below.

Issue #6175, opened 2024-01-22
Replace binary release artifact links


sudonatalie:
Now that the AppVeyor subscription has ended, we should update the links for downloading binary release artifacts in the README to the appropriate Azure bucket when they're available.

MarijnS95:
That's unfortunate (timing), I just wanted to download some artifacts to test/import a recent bugfix but AppVeyor is no longer hosting them and there don't seem to be any artifacts emitted on Azure either (latest build at https://dev.azure.com/DirectXShaderCompiler/public/_build/results?buildId=5428&view=results).

Any suggestions, or should I build it on my own for the time being?

s-perron:
Sorry about that. Please build on your own for now. We are working on a replacement.

MarijnS95:
@s-perron thanks for the info. This is not super high priority, I'm more so curious to your approximate time-frame to provide a replacement, before spending some time to put together a CI build myself.

Not that such a thing is too complicated (for Linux I'm familiar at least), we already have one for MacOS and you might want to consider providing artifacts out of the box for that platform too (x86 and aarch64)?

s-perron:
@pow2clk Will have a better idea on the timeline.

tgjones:
I'm also interested in this. For [Shader Playground](https://shader-playground.timjones.io/), I've relied on downloading the artifacts from AppVeyor. 

I could change my build pipeline to build from source, if necessary, but it would help to have some idea about the timeline for making artifacts available from Azure Pipelines.

PR #6176, opened 2024-01-22
Remove AppVeyor

sudonatalie:
The AppVeyor subscription has now ended, so latest builds are failing with timeouts and downloads of last-known-good are failing with download size limits. This change removes the AppVeyor run scripts and documentation references.

Follow-up to add new links for downloading latest build artifacts in Issue #6175


README.md
```diff
@@ -1,19 +1,22 @@
...
 
 For more information, see the [Wiki](https://github.com/microsoft/DirectXShaderCompiler/wiki).
 
 Visit the [DirectX Landing Page](https://devblogs.microsoft.com/directx/landing-page/) for more resources for DirectX developers.
 
 ## Downloads
-You can download the latest successful build's artifacts (built by Appveyor) for the main branch:
+
+> [!CAUTION]
+> Latest project builds are temporarily unavailable. See [Issue #6175](https://github.com/microsoft/DirectXShaderCompiler/issues/6175).
```
Keenuts, line 12: Shall we redirect them to the Vulkan SDK?

Keenuts (approved):

README.md
sudonatalie: I don't think so, there's still the "Pre-built Releases" below for folks want a stable known-good rather than a latest tip-of-tree build. It's probably better to just remove this section entirely for now.

hekota (approved):

PR #6177, opened 2024-01-22
Make clang-format happy

sudonatalie:
The build is failing due to clang-format modifying these lines. I've changed them to what clang-format wants even though it's a bit odd, and updated the CMake script to make the error message more clear when this type of failure occurs.


sudonatalie:
I'm not sure if this is due to a change in clang-format, so we'll see if the bots fail this now. @Keenuts, @cassiebeckley and I are all getting local build failures, but I think we're all working with a tip-of-tree build of clang-format, and maybe we don't want to try to keep up with that if that's the cause...

llvm-beanz (approved):
Yea... this is because of the switch nesting.

Issue #6151, opened 2024-01-12
Build failure on macOS and some Linux distros


richard-lunarg:
Description
DXC is failing to build from source on the latest macOS and XCode versions, and at least one Linux distro. This is happening even for old commits, so it is likely either a compiler "bug", or a DXC bug that has "gotten past" previous compilers.

Steps to Reproduce
Download source or clone from main branch.
Extract.
Create a new folder for building and build:
mkdir dxc-build-dir
cd dxc-build-dir
cmake .. -DCMAKE_BUIILD_TYPE=Release -C ../cmake/caches/PredefinedParams.cmake
make 

(Note, same results from Ninja generator)

Actual Behavior
The following failure occurs:
[372/1514] cd /home/marky/Downloads/vulkansdk/1.3.275.0_backup/source/DirectXShaderCompiler/build/lib/D...me/marky/Downloads/vulkansdk/1.3.275.0_backup/source/DirectXShaderCompiler/lib/DXIL/DxilShaderModel.cpp
FAILED: lib/DXIL/CMakeFiles/DxilShaderModel /home/marky/Downloads/vulkansdk/1.3.275.0_backup/source/DirectXShaderCompiler/build/lib/DXIL/CMakeFiles/DxilShaderModel 
cd /home/marky/Downloads/vulkansdk/1.3.275.0_backup/source/DirectXShaderCompiler/build/lib/DXIL && /usr/bin/cmake -E compare_files /home/marky/Downloads/vulkansdk/1.3.275.0_backup/source/DirectXShaderCompiler/build/lib/DXIL/tmp/DxilShaderModel.cpp /home/marky/Downloads/vulkansdk/1.3.275.0_backup/source/DirectXShaderCompiler/lib/DXIL/DxilShaderModel.cpp

Environment
macOS 14.2.1
XCode 15.1 (and 15.2)
CMake 3.28.1 (also tried 3.28.0)
- DXC version <main or any recent commit>

It's also failing on "a" linux distro (reported to me internally). I'll have to amend this later when I can get those details. The failure on macOS is easy enough to reproduce however and might be more immediately useful.

On an older machine, running an older version of XCode (15.01) it works.

llvm-beanz:
@richard-lunarg, do you have local changes? This failure means that there is a difference between the locally generated versions of source files and the versions in the source tree.

[...]

PR #6185, opened 2024-01-23
[Docs] Update Readme and build instruction docs

llvm-beanz:
This workflow has been around long enough to stop treating it as the "experimental" workflow and just update the documentation to use it.

We've gotten numerous issues lately related to problems building, so this will hopefully help.

Fixes: #5762
Related: #6151

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,58 +43,7 @@ As an example of community contribution, this project can also target the [SPIR-
 
 ## Building Sources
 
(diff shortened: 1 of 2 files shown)


s-perron (approved):

coopp (approved):
Looks good to me.

dmpots (approved):

PR #6186, opened 2024-01-23
[SPIR-V] Add VulkanMemoryModelDeviceScope

### LINKED ISSUE #6066: [SPIR-V] Missing VulkanMemoryModelDeviceScopeKHR Capability when using wave intrinsics
**Description**
When building a raytracing shader (closestHit, miss or raygeneration) with a device atomic operation **and** wave intrinsic operations, the compilation fails due to a missing SPIR-V capability declaration. 

**Steps to Reproduce**
1. Create a closestHit, miss or rayGeneration shader
2. Add an atomic operation on a device buffer
3. Add a waveintrinsic operation
4. Compile for Vulkan 1.2

compiled with `-HV 2021 -fspv-target-env=vulkan1.2 spirv1.4 -T lib_6_4 -E main -spirv -O0 `

A reproducer can be found here:
https://godbolt.org/z/Mrxb7z3K7

**Actual Behavior**
The compiler fails with the following error.

```
fatal error: generated SPIR-V is invalid: Use of device scope with VulkanKHR memory model requires the VulkanMemoryModelDeviceScopeKHR capability
  %21 = OpAtomicIAdd %int %18 %uint_1 %uint_0 %20

Compiler returned: 5
```

The created SPIR-V is obviously missing the required capability which should be added by dxc in this case.

**Environment**
- DXC version: trunk (7b222ff6)
- Host Operating System Windows 11 64-bit

sudonatalie:
The VulkanMemoryModelDeviceScope capability is now covered by the SPIRV-Tools capability trimming pass, so it should be added here unconditionally now.

Fixes #6066

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -592,6 +592,7 @@ public:
 
   // === SPIR-V Module Structure ===
   inline void setMemoryModel(spv::AddressingModel, spv::MemoryModel);
+  inline SpirvMemoryModel *getMemoryModel();
 
   /// \brief Adds an entry point for the module under construction. We only
   /// support a single entry point per module for now.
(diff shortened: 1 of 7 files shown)


Keenuts (commented):
Should this be handled using the capability trimming pass?
We haven't done a bulk cleanup, but now seems this can be handled the following way:

- add the capability support in SPIRV-Tools
- add the capability unconditionally in CapabilityVisitor.cpp:855

cassiebeckley:
Using the capability trimming pass here seems like a good idea to me as well.

sudonatalie:
Fair enough, adding it to SPIRV-Tools here: https://github.com/KhronosGroup/SPIRV-Tools/pull/5544
and I'll update this PR after it lands.

sudonatalie:
Updated and ready for re-review

cassiebeckley (approved):

Keenuts (approved):

PR #6189, opened 2024-01-23
[SPIRV] Reimplement flat conversion.

### LINKED ISSUE #4906: [SPIR-V] fatal error: generated SPIR-V is invalid: Expected total number of given components to be equal to the size of Result Type vector
### fatal error: generated SPIR-V is invalid: Expected total number of given components to be equal to the size of Result Type vector
This error occurs when compiling for SPIR-V, but does not occur for DXIL.

### Functional impact
Cannot compile shaders for SPIR-V.

#### test.i
```
#line 1 "test.hlsl"
#line 1 "./sh.hlsli"



#line 1 "./common.hlsli"



static const float pi = 3.141592653589793;
static const float sqrt_pi = 1.772453850905516;
static const float sqrt_2 = 1.414213562373095;
static const float sqrt_3 = 1.732050807568877;

static const int max_int_v = 0x7FFFFFFF;
static const int min_int_v = -0x80000000;
static const uint max_uint_v = 0xFFFFFFFFu;
static const float max_float_v = 3.4028234664e+38f;

float squared(float x) {
 return x * x;
}

#line 4 "./sh.hlsli"


namespace zh {
 typedef float2 zh2;
 typedef float3 zh3;
}

namespace sh {
 namespace constants {
  static const float band0_factor = 0.5f / sqrt_pi;
  static const float band1_factor = 0.5f * sqrt_3 / sqrt_pi;

  static const float zh0_rotation_factor = 2.0f * sqrt_pi;
  static const float zh1_rotation_factor = 2.0f * sqrt_pi / sqrt_3;
 }

 typedef float band0;
 typedef float3 band1;

 band0 coefficients_band0(float3 dir) {
  return constants::band0_factor;
 }
 band1 coefficients_band1(float3 dir) {
  return constants::band1_factor * float3(-dir.y, dir.z, -dir.x);
 }

 float integrate(band0 lhs, band0 rhs) {
  return lhs * rhs;
 }
 float integrate(band1 lhs, band1 rhs) {
  return dot(lhs, rhs);
 }

 band0 rotate_zh_band0(float zh_v, float3 z) {
  return constants::zh0_rotation_factor * zh_v * coefficients_band0(z);
 }
 band1 rotate_zh_band1(float zh_v, float3 z) {
  return constants::zh1_rotation_factor * zh_v * coefficients_band1(z);
 }


 struct sh2 {
  band0 b0;
  band1 b1;

  sh2 operator+(sh2 rhs) {
   sh2 result;
   result.b0 = b0 + rhs.b0;
   result.b1 = b1 + rhs.b1;
   return result;
  }
  sh2 operator-(sh2 rhs) {
   sh2 result;
   result.b0 = b0 - rhs.b0;
   result.b1 = b1 - rhs.b1;
   return result;
  }
  sh2 operator*(float rhs) {
   sh2 result;
   result.b0 = b0 * rhs;
   result.b1 = b1 * rhs;
   return result;
  }
 };

 sh2 coefficients_sh2(float3 dir) {
  return (sh2)float4(coefficients_band0(dir), coefficients_band1(dir));
 }

 sh2 rotate_zh(zh::zh2 v, float3 z) {
  return (sh2)float4(rotate_zh_band0(v.x, z), rotate_zh_band1(v.y, z));
 }

 float integrate(sh2 lhs, sh2 rhs) {
  return integrate(lhs.b0, rhs.b0) + integrate(lhs.b1, rhs.b1);
 }


 namespace clamped_cosine {
  static const float band0_zh = 0.5f * sqrt_pi;
  static const float band1_zh = sqrt_pi / sqrt_3;

  band0 eval_band0(float3 v) {
   return rotate_zh_band0(band0_zh, v);
  }
  band1 eval_band1(float3 v) {
   return rotate_zh_band1(band1_zh, v);
  }

  sh2 eval_sh2(float3 v) {
   return (sh2)float4(eval_band0(v), eval_band1(v));
  }
 }

 namespace impulse {
  static const float band0_zh = 0.5f / sqrt_pi;
  static const float band1_zh = 0.5f * sqrt_3 / sqrt_pi;

  band0 eval_band0(float3 v) {
   return rotate_zh_band0(band0_zh, v);
  }
  band1 eval_band1(float3 v) {
   return rotate_zh_band1(band1_zh, v);
  }

  sh2 eval_sh2(float3 v) {
   return (sh2)float4(eval_band0(v), eval_band1(v));
  }
 }
}
#line 1 "test.hlsl"


StructuredBuffer<float4> input1;
StructuredBuffer<float4> input2;
RWStructuredBuffer<float4> output;


[numthreads(64, 1, 1)]
void main_cs(uint id : SV_DispatchThreadID) {
 sh::sh2 in1 = (sh::sh2)input1[id];
 sh::sh2 in2 = (sh::sh2)input2[id];
 output[id] = (float4)sh::integrate(in1, in2);
}
```

Compile with `dxc test.i -HV 2021 -T cs_6_5 -E main_cs` and `dxc test.i -HV 2021 -T cs_6_5 -E main_cs -spirv`.

### Expected result
The shader should compile.

### Actual result
The shader compiles for the first command line, but fails for the second with the given error:
```
fatal error: generated SPIR-V is invalid: Expected total number of given components to be equal to the size of Result Type vector
  %28 = OpCompositeConstruct %v3float %27 %27 %27

```

### Further technical details
This comes from the version of DXC bundled with Vulkan SDK 1.3.236.0.
```
dxcompiler.dll: 1.7 - 1.7.0.3738 (cafbbed4d)
```

s-perron:
The implementation of flat conversion in the spir-v codegen is very
adhoc. This changes the implemetation to match what is done in DXIL. For
now I have only reimplemented to catch all case. I will leave the
special cases as they seem to be correct for what they do, and will
generate less verbose code.

Fixes #4906

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -3540,132 +3540,34 @@ SpirvInstruction *SpirvEmitter::doCastExpr(const CastExpr *expr,
 SpirvInstruction *SpirvEmitter::processFlatConversion(
     const QualType type, const QualType initType, SpirvInstruction *initInstr,
     SourceLocation srcLoc, SourceRange range) {
-  if (isConstantTextureBuffer(initType)) {
-    return reconstructValue(initInstr, type, SpirvLayoutRule::Void, srcLoc,
-                            range);
+
+  // If the same literal is used in multiple instructions, then the literal
+  // visitor may not be able to pick the correct type for the literal. The
+  // happens when say one instruction uses the literal as a float and another
+  // uses it as a double. We solve this by setting the type for the literal to
+  // its 32-bit equivalent.
+  //
+  // TODO(6188): This is wrong when the literal it too large to be held in
+  // the the 32-bit type. We do this because it is consistent with the long
+  // standing behaviour. Changing now would result in more 64-bit arithmetic,
+  // which the optimizer does not handle as well.
+  QualType resultType = initInstr->getAstResultType();
+  if (resultType->isSpecificBuiltinType(BuiltinType::LitFloat)) {
+    initInstr->setAstResultType(astContext.FloatTy);
+  } else if (resultType->isSpecificBuiltinType(BuiltinType::LitInt)) {
+    if (resultType->isSignedIntegerType())
+      initInstr->setAstResultType(astContext.IntTy);
+    else
+      initInstr->setAstResultType(astContext.UnsignedIntTy);
   }
 
-  // Try to translate the canonical type first
-  const auto canonicalType = type.getCanonicalType();
-  if (canonicalType != type)
-    return processFlatConversion(canonicalType, initType, initInstr, srcLoc,
-                                 range);
+  // Decompose `initInstr`.
+  std::vector<SpirvInstruction *> flatValues = decomposeToScalars(initInstr);
 
-  // Primitive types
-  {
-    QualType ty = {};
-    if (isScalarType(type, &ty)) {
-      if (const auto *builtinType = ty->getAs<BuiltinType>()) {
-        switch (builtinType->getKind()) {
-        case BuiltinType::Void: {
-          emitError("cannot create a constant of void type", srcLoc);
-          return 0;
-        }
-        case BuiltinType::Bool:
-          return castToBool(initInstr, initType, ty, srcLoc, range);
-        // Target type is an integer variant.
-        case BuiltinType::Int:
-        case BuiltinType::Short:
-        case BuiltinType::Min12Int:
-        case BuiltinType::Min16Int:
-        case BuiltinType::Min16UInt:
-        case BuiltinType::UShort:
-        case BuiltinType::UInt:
-        case BuiltinType::Long:
-        case BuiltinType::LongLong:
-        case BuiltinType::ULong:
-        case BuiltinType::ULongLong:
-        case BuiltinType::Int8_4Packed:
-        case BuiltinType::UInt8_4Packed:
-          return castToInt(initInstr, initType, ty, srcLoc, range);
-        // Target type is a float variant.
-        case BuiltinType::Double:
-        case BuiltinType::Float:
-        case BuiltinType::Half:
-        case BuiltinType::HalfFloat:
-        case BuiltinType::Min10Float:
-        case BuiltinType::Min16Float:
-          return castToFloat(initInstr, initType, ty, srcLoc, range);
-        default:
-          emitError("flat conversion of type %0 unimplemented", srcLoc)
-              << builtinType->getTypeClassName();
-          return 0;
-        }
-      }
-    }
+  if (flatValues.size() == 1) {
+    return SplatScalarToGenerate(type, flatValues[0], SpirvLayoutRule::Void);
   }
-  // Vector types
-  {
-    QualType elemType = {};
-    uint32_t elemCount = {};
-    if (isVectorType(type, &elemType, &elemCount)) {
-      auto *elem =
-          processFlatConversion(elemType, initType, initInstr, srcLoc, range);
-      llvm::SmallVector<SpirvInstruction *, 4> constituents(size_t(elemCount),
-                                                            elem);
-      return spvBuilder.createCompositeConstruct(type, constituents, srcLoc,
-                                                 range);
-    }
-  }
-
-  // Matrix types
-  {
-    QualType elemType = {};
-    uint32_t rowCount = 0, colCount = 0;
-    if (isMxNMatrix(type, &elemType, &rowCount, &colCount)) {
-      // By default HLSL matrices are row major, while SPIR-V matrices are
-      // column major. We are mapping what HLSL semantically mean a row into a
-      // column here.
-      const QualType vecType = astContext.getExtVectorType(elemType, colCount);
-      auto *elem =
-          processFlatConversion(elemType, initType, initInstr, srcLoc, range);
-      const llvm::SmallVector<SpirvInstruction *, 4> constituents(
-          size_t(colCount), elem);
-      auto *col = spvBuilder.createCompositeConstruct(vecType, constituents,
-                                                      srcLoc, range);
-      const llvm::SmallVector<SpirvInstruction *, 4> rows(size_t(rowCount),
-                                                          col);
-      return spvBuilder.createCompositeConstruct(type, rows, srcLoc, range);
-    }
-  }
-
-  // Struct type
-  if (const auto *structType = type->getAs<RecordType>()) {
-    const auto *decl = structType->getDecl();
-    llvm::SmallVector<SpirvInstruction *, 4> fields;
-
-    for (const auto *field : decl->fields()) {
-      // There is a special case for FlatConversion. If T is a struct with only
-      // one member, S, then (T)<an-instance-of-S> is allowed, which essentially
-      // constructs a new T instance using the instance of S as its only member.
-      // Check whether we are handling that case here first.
-      if (!field->isBitField() &&
-          field->getType().getCanonicalType() == initType.getCanonicalType()) {
-        fields.push_back(initInstr);
-      } else {
-        fields.push_back(processFlatConversion(field->getType(), initType,
-                                               initInstr, srcLoc, range));
-      }
-    }
-
-    return spvBuilder.createCompositeConstruct(type, fields, srcLoc, range);
-  }
-
-  // Array type
-  if (const auto *arrayType = astContext.getAsConstantArrayType(type)) {
-    const auto size =
-        static_cast<uint32_t>(arrayType->getSize().getZExtValue());
-    auto *elem = processFlatConversion(arrayType->getElementType(), initType,
-                                       initInstr, srcLoc, range);
-    llvm::SmallVector<SpirvInstruction *, 4> constituents(size_t(size), elem);
-    return spvBuilder.createCompositeConstruct(type, constituents, srcLoc,
-                                               range);
-  }
-
-  emitError("flat conversion of type %0 unimplemented", {})
-      << type->getTypeClassName();
-  type->dump();
-  return 0;
+  return GenerateFromScalars(type, flatValues, SpirvLayoutRule::Void);
 }
 
 SpirvInstruction *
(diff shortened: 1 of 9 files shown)


tools/clang/test/CodeGenSPIRV/cast.struct-to-int.hlsl
```diff
@@ -83,22 +83,21 @@ void main()
 // CHECK: [[COLORS:%[^ ]*]] = OpLoad [[TWOCOLORS]]
 // CHECK: [[COLORS0:%[^ ]*]] = OpCompositeExtract [[COLORRGBA]] [[COLORS]] 0
 // CHECK: [[COLORS00:%[^ ]*]] = OpCompositeExtract [[UINT]] [[COLORS0]] 0
-// CHECK: [[COLORS000:%[^ ]*]] = OpBitFieldUExtract [[UINT]] [[COLORS00]] [[U0]] [[U8]]
 // CHECK: [[BUF00:%[^ ]*]] = OpAccessChain %{{[^ ]*}} [[BUF]] [[I0]] [[U0]]
-// CHECK: OpStore [[BUF00]] [[COLORS000]]
+// CHECK: OpStore [[BUF00]] [[COLORS00]]
```
s-perron, line 87: I looked at the DXIL code, and this seems to be the correct behaviour. Could someone else check? We assign the scalar to the bit container instead of to the individual bit fields.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -14630,5 +14532,263 @@ SpirvEmitter::doUnaryExprOrTypeTraitExpr(const UnaryExprOrTypeTraitExpr *expr) {
...
+  uint32_t elementCount = 0;
+  uint32_t numOfRows = 0;
+  uint32_t numOfCols = 0;
+
+  QualType resultType = inst->getAstResultType();
+  if (hlsl::IsHLSLResourceType(resultType)) {
+    resultType = hlsl::GetHLSLResourceResultType(resultType);
+  }
+
+  // switch(inst->getAstResultType()->getKind())
```
sudonatalie, line 14547: Clean up this line?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -14630,5 +14532,263 @@ SpirvEmitter::doUnaryExprOrTypeTraitExpr(const UnaryExprOrTypeTraitExpr *expr) {
...
+          // See how we can improve the performance by avoiding this copy.
+          result.insert(result.end(), decomposedField.begin(),
+                        decomposedField.end());
+          return true;
+        });
+    return result;
+  }
+
+  else {
+    resultType->dump();
```
sudonatalie, line 14612: Did you mean to leave this here?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -14630,5 +14532,263 @@ SpirvEmitter::doUnaryExprOrTypeTraitExpr(const UnaryExprOrTypeTraitExpr *expr) {
...
+    SpirvInstruction *result = spvBuilder.createCompositeConstruct(
+        type, elements, scalars[0]->getSourceLocation());
+    result->setLayoutRule(layoutRule);
+    scalars.erase(scalars.begin(), scalars.begin() + elementCount);
+    return result;
+  } else if (isMxNMatrix(type, &elementType, &numOfRows, &numOfCols)) {
+    std::vector<SpirvInstruction *> rows;
+    SourceLocation loc = scalars[0]->getSourceLocation();
+    QualType rowType = astContext.getExtVectorType(elementType, numOfCols);
+    for (uint32_t i = 0; i < numOfRows; i++) {
```
sudonatalie, line 14660: Since HLSL is row-major and SPIR-V is column-major, I would have expected this to be transposed when reconstructing, but it's difficult for me to tell for sure if that's correct or not without a test. Unfortunately I don't see an existing matrix flat conversion test that would have caught it if it did change. Can you add one?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -14630,5 +14532,263 @@ SpirvEmitter::doUnaryExprOrTypeTraitExpr(const UnaryExprOrTypeTraitExpr *expr) {
...
+
+  QualType resultType = inst->getAstResultType();
+  if (hlsl::IsHLSLResourceType(resultType)) {
+    resultType = hlsl::GetHLSLResourceResultType(resultType);
+  }
+
+  // switch(inst->getAstResultType()->getKind())
+  if (isScalarType(resultType)) {
+    return {inst};
+  } else if (isVectorType(resultType, &elementType, &elementCount)) {
```
sudonatalie, line 14550: Can you switch these else ifs after returns to ifs for readability? (https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html)

Same for the other examples below

tools/clang/lib/SPIRV/SpirvEmitter.h
```diff
@@ -1248,6 +1248,26 @@ class SpirvEmitter : public ASTConsumer {
...
+  /// Returns a vector of SpirvInstruction that is the decompostion of `inst`
+  /// into scalars. This is recursive. For example, a struct of a 4 element
+  /// vector will return 4 scalars.
+  std::vector<SpirvInstruction *> decomposeToScalars(SpirvInstruction *inst);
+
+  /// Returns a spirv instruction with the value of the given type and layout
+  /// rule that is obtained by assigning each scalar in `type` to corresponding
+  /// value in `scalars`. This is the inverse of `decomposeToScalars`.
+  SpirvInstruction *
+  GenerateFromScalars(QualType type, std::vector<SpirvInstruction *> &scalars,
```
sudonatalie, line 1260: nit: Switch function name to lowerCamelCase (I know we're inconsistent with the style guide in general, but might as well be consistent with our inconsistency within the file)

Same for Splat below

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -3540,132 +3540,34 @@ SpirvInstruction *SpirvEmitter::doCastExpr(const CastExpr *expr,
...
-      return spvBuilder.createCompositeConstruct(type, constituents, srcLoc,
-                                                 range);
-    }
+  // If the same literal is used in multiple instructions, then the literal
+  // visitor may not be able to pick the correct type for the literal. The
+  // happens when say one instruction uses the literal as a float and another
+  // uses it as a double. We solve this by setting the type for the literal to
+  // its 32-bit equivalent.
+  //
+  // TODO(6188): This is wrong when the literal it too large to be held in
```
Keenuts, line 3550:
```suggestion
  // TODO(6188): This is wrong when the literal is too large to be held in
```

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -14630,5 +14532,263 @@ SpirvEmitter::doUnaryExprOrTypeTraitExpr(const UnaryExprOrTypeTraitExpr *expr) {
...
+SpirvEmitter::GenerateFromScalars(QualType type,
+                                  std::vector<SpirvInstruction *> &scalars,
+                                  SpirvLayoutRule layoutRule) {
+  QualType elementType;
+  uint32_t elementCount = 0;
+  uint32_t numOfRows = 0;
+  uint32_t numOfCols = 0;
+
+  if (isScalarType(type)) {
+    // If the type if bool with a non-void layout rule, then it should be
```
Keenuts, line 14628:
```suggestion
    // If the type is bool with a non-void layout rule, then it should be
```

tools/clang/test/CodeGenSPIRV/cast.flat-conversion.literal-initializer.hlsl
```diff
@@ -47,4 +52,11 @@ void main() {
 // CHECK-NEXT:           {{%[0-9]+}} = OpCompositeConstruct %S [[a2_float]] [[a_float_0]] [[a2_double]] [[a]] [[a_int64]] [[a_uint64]]
   double a;
   S s1 = (S)(a);
+
+// TODO(6188): This is wrong because we lose most significant bits in the literal.
+// CHECK: [[lit:%[0-9]+]] = OpIAdd %int %int_0 %int_1
+// CHECK: [[longLit:%[0-9]+]] = OpSConvert %long [[lit]]
+// CHECK: [[t:%[0-9]+]] = OpCompositeConstruct %T [[lit]] [[longLit]]
+// CHECK: OpStore %t [[t]]
+  T t = (T)(0x100000000+1);
```
Keenuts, line 61: Could you add lines to test suffixes?
```
T t = (T)(0x100000000ul+1);
T t = (T)(0x100000000ull+1);
```

There are behavior differences between C++ and DXIL compilation, and SPIR-V compilation and DXIL, so we might want to be careful to not diverge from DXIL (which already diverges from C++)

tools/clang/test/CodeGenSPIRV/cast.struct-to-int.hlsl
Keenuts: Seems to me that this is correct.
The DXIL code is a bit weird, they load 3 times the single uint, maybe because a code part sees 3 fields, but overall, the generated SPIR-V code seems to match: the whole uint is stored in the array.

s-perron: I think my implementation is correct for this case. I tried another test (https://godbolt.org/z/PPjGWfcKG), and I think I have that one "wrong". I personally think the DXIL behaviour is incomprensible.

It seems like each bit field will count for a different scalar  when flattening, but the value of the scalar is not the value of the bitfield. It is the value of the contain holding the bitfiled. So in the new test case, `buf[0]` gets every element set to the same value `524255231` not `{{255, 127}, {63, 31}}` as you might expect.

s-perron: I added that example as a test, and made it work.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
s-perron: I added at test. Please double check that my understanding is correct. It is hard to follow exactly what is going on. Basicly, it seems like I row or col major matricies can be treated the same way in that part of the code.

sudonatalie: That seems correct to me, thanks for double checking and adding a test.

sudonatalie (approved):
LGTM

Keenuts (approved):
LGTM, thanks for the fix!

PR #6190, opened 2024-01-23
[SPIR-V] Allow zero outputcontrolpoints

### LINKED ISSUE #3733: [SPIR-V] [outputcontrolpoints(0)] or missing outputcontrolpoints Hull stage attribute crash the compiler (works with DXIL backend)
### Title
[SPIR-V] [outputcontrolpoints(0)] or missing outputcontrolpoints Hull stage attribute crash the compiler (works with DXIL backend)

### Functional impact
Cannot produce hull shaders that don't have any control point output (e.g. when only patch data is needed for the domain stage)

### Minimal repro steps
[repro_shader.txt](https://github.com/microsoft/DirectXShaderCompiler/files/6414759/repro_shader.txt)
or do `dxc.exe -T hs_6_0 -E Hull -spirv repro_shader_no_attribute.txt`
[repro_shader_no_attribute.txt](https://github.com/microsoft/DirectXShaderCompiler/files/6414789/repro_shader_no_attribute.txt)

### Expected result
Compilation succeeds with control point output count being zero

### Actual result
Compiler crashes with output:
`Internal compiler error: access violation. Attempted to read from address 0x0000000000000020`

### Further technical details

sudonatalie:
It is valid HLSL for a Hull shader to have zero outputcontrolpoints or have the outputcontrolpoints attribute unset, based on the tests added here passing when compiling for DXIL. This change removes the checks in the SPIR-V backend that required a strictly positive number of control points and adds tests to verify the result.

Fixes #3733

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -2836,7 +2836,8 @@ void DeclResultIdMapper::storeToShaderOutputVariable(
   // element in the per-vertex data array: the one indexed by
   // SV_ControlPointID.
   else if (stageVarData.invocationId.hasValue() &&
-           stageVarData.invocationId.getValue() != nullptr) {
+           stageVarData.invocationId.getValue() != nullptr &&
+           stageVarData.arraySize != 0) {
     // Remove the arrayness to get the element type.
     assert(isa<ConstantArrayType>(varInstr->getAstResultType()));
     const auto elementType =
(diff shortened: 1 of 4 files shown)


s-perron:
This is awkward. Vulkan does not allow a zero for the number of output vertices. See https://docs.vulkan.org/spec/latest/chapters/pipelines.html#VUID-VkPipelineShaderStageCreateInfo-stage-00713.

I can't tell of the code generated when there is no output vertices is correct. GLSL does not generate anything https://shader-playground.timjones.io/7f6252b21a9b099056aaccd511ec7046.

sudonatalie:
Hm... upon closer review of the D3D documentation for Hull shaders it seems like it also shouldn't be allowed there: "The shader output is between 1 and 32 control points".

https://learn.microsoft.com/en-us/windows/win32/direct3d11/direct3d-11-advanced-stages-tessellation#hull-shader-stage

I'll look into whether disallowing it on the frontend is a reasonable option.

PR #6195, opened 2024-01-24
[SPIR-V] Move constant evaluation code into a new class

cassiebeckley:
This moves functions for evaluating constant expressions and translating LLVM AP types to SPIR-V constants into a new class. These functions need to be used from LowerTypeVisitor in #6156, so I am factoring them out to avoid duplication.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.h
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.h
@@ -33,6 +33,7 @@
 #include "clang/SPIRV/SpirvContext.h"
 #include "llvm/ADT/STLExtras.h"
 
+#include "ConstEvaluator.h"
 #include "DeclResultIdMapper.h"
 
 namespace spvtools {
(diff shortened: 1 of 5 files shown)


sudonatalie (approved):

tools/clang/lib/SPIRV/SpirvEmitter.h
```diff
@@ -767,42 +768,6 @@ class SpirvEmitter : public ASTConsumer {
                                           SourceLocation,
                                           SourceRange range = {});
 
-private:
-  /// Translates the given frontend APValue into its SPIR-V equivalent for the
```
Keenuts, line 771: Shall those comments also be moved to the header? Especially for function like `translateAsFloat` where we allow casting to some extent.

Keenuts (changes requested):
LGTM, just the comments part.

tools/clang/lib/SPIRV/SpirvEmitter.h
cassiebeckley: Ah, yes, they should have been moved. I've fixed it.

Keenuts (approved):

PR #6202, opened 2024-01-24
[Sema] Add final touches to wave size range implementation

### LINKED ISSUE #6161: [Feature Request] Implement Wave Range Size

Shader Model 6.6 provides no mechanism for specifying multiple wave sizes that a
 single shader might be able to support.
Consequently, if more than one target wave size is of interest, separate entry
 points will have to be created for each wave size.
These will each have to be compiled and shipped separately and, at run time,
 the appropriate shaders will have to be selected.
This increases the size of the shipped product and potentially slows down
 runtime shader loading.

Ranges of wave sizes are already present in areas of the D3D API.
The query to determine what wave sizes are supported takes a range and similar
 information is transmitted from the shader to the runtime, though not the
 driver.

Modifying the parameters of the `WaveSize` attribute that was introduced by
 Shader Model 6.6 to take additional parameters will allow the shader author to
 specify fully what the shader supports and what it prefers.

By adding an optional second `WaveSize` parameter, shader authors can provide
 two values that represent the range of wave sizes that the shader can support.
This allows the same shader to be used in the event of the availability of a
 wave size that the shader has optimized for specifically as well as other sizes
 that are supported, but might not be so optimal.

Useful as it is, the preferred wave size parameter adds some level of testing,
 diagnostic, and other implementation complexity.

Instead of modifying the existing attribute, an additional range attribute taking
 only two parameters might be provided to specify the range while keeping the
 existing `WaveSize` attribute to indicate the preferred wave size.

bob80905:
Upon further discussion, the team has agreed that in certain degenerate cases, the current diagnostics are insufficient.
In the case that min == max in the wave size range attribute, a defaultError warning should be emitted. Additionally, there should be an explicit way to handle the case where 0,0,0 is passed to the wavesize range attribute. 
This PR directly handles and tests both of these cases.
Fixes #6161


bob80905:
This PR inspired the creation of this issue:
https://github.com/microsoft/DirectXShaderCompiler/issues/6211

pow2clk (commented):
Mostly nits with one concern about whether the two places that an error can be generated will result in duplicate errors. 

There are still a few tests mentioned in the spec that I don't see here. 

1. Use the following invalid parameters each parameter location to `WaveSize`
  and ensure that an appropriate error is produced:
   1. Negative power-of-two integer
   2. Floating-point value
   3. non-compile-time constant integer
   4. Power-of-two integer less than 4
   5. Power-of-two integer greater than 128
   6. A non-power-of-two integer between 4 and 128
2. 2. Add the following invalid `WaveSize` attributes to an compute shader entry
  point and ensure that an appropriate error is produced:
   1. No parameter list  --- By this I mean `[WaveSize]`
   2. An empty parameter list "()"  -- DONE in wavesize_range.hlsl
   3. Four parameters
3. DONE
4. Combine multiples of the 1, 2 and 3 parameter `WaveSize` attribute variants
  with different values on the same entry point and ensure that an attribute
  conflict error is produced.  -- This should just work. Only needs testing. SHould also verify that when the values are the same, there is no failure, but that's a bonus.


Test that the following produce validation errors:

1. The wave size range tag pointing to anything but a tuple of 3.
2. A tuple value is not an integer.
3. A range tuple value is -4, 0, 1, 2, 3, 127, 129, or 256.
4. A preferred tuple value is -4, 1, 2, 3, 127, 129, or 256.
5. The minimum wave size value is greater than the maximum, but otherwise valid. -- DONE in max-lt-min.ll
6. The preferred wave size is outside the specified range, but otherwise valid.
7. Multiple metadata `kDxilRangedWaveSizeTag`s are in the same compiled shader.

As a matter of fact, I noticed a mistake in the validation error list as I was writing this :) we should remove the validation error for when max==min since we intend to allow users to force that state if they disable the ErrorDefault warning. I think that means max-equals-min.ll should go.


If I've missed places where this testing exists, please point them out. If you'd rather address that with a followup, that's fine, but we either need a new bug or I'm afraid for this one to not close the bug.

tools/clang/test/HLSLFileCheck/hlsl/entry/attributes/wavesize-compute-node-rdat.hlsl
```diff
@@ -0,0 +1,36 @@
...
+}
+
+#ifdef NODE
+[Shader("node")]
+[NodeLaunch("broadcasting")]
+[NumThreads(1,1,8)]
+[NodeDispatchGrid(1,1,1)]
+[WaveSize(16 RANGE)]
+void node() { }
+#endif
```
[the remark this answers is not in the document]
pow2clk: Resolve this when you've read it as it's just an idle comment.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -1812,6 +1812,8 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
   if (const HLSLWaveSizeAttr *Attr = FD->getAttr<HLSLWaveSizeAttr>()) {
     funcProps->WaveSize = DxilWaveSize::Translate(
         Attr->getMin(), Attr->getMax(), Attr->getPreferred());
+    // The legacy wavesize tag was used if there is only one spelled argument.
+    funcProps->WaveSize.SetIsLegacy(Attr->getSpelledArgsCount() == 1);
```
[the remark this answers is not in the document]
bob80905: Originally I thought to preserve that information for the purpose of identifying when the legacy wavesize tag was used, and forgot to remove it once I learned about metadata tags. I think I can completely rip out all the logic for this safely, unless there's some other potential usefulness for this information later on.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
   }
 
   const hlsl::ShaderModel *SM = ValCtx.DxilMod.GetShaderModel();
+  // validate that any wavesize tags don't appear outside their expected shader
+  // models. Validate only 1 tag exists per entry point.
+  // Don't do this validation if the shader is non-compute
```
pow2clk, line 3993: It's an extremely small nit, but a space between the comments describing the next block of code and the comment describing the single conditional line might make it a bit less confusing. 

As long and as nested as this is though, perhaps a separate function call would be useful to keep all the wave size range code in the same place

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+  if (SM->IsCS() || SM->IsLib()) {
+
+    NamedMDNode *TA = pModule->getNamedMetadata("dx.entryPoints");
```
pow2clk, line 3996: I'm curious why the variable is `TA`? Was it just copied from another loaded MD node?

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+            TANode->getOperand(TANode->getNumOperands() - 1);
+        // the final operand to the entry points tuple should be a constant int.
+        if (mOp == nullptr ||
+            (mOp.get())->getMetadataID() != Metadata::MDTupleKind) {
+          continue;
+        }
+
+        // get access to the node that stores entry properties
+        MDTuple *TANode2 =
+            dyn_cast<MDTuple>(TANode->getOperand(TANode->getNumOperands() - 1));
```
pow2clk, line 4016: This is retrieving the same value as `mOp` above, but casting it to MDTuple. Is that right? 

Again I'm unsure what "TA" refers to. This should be the entry properties tuple and naming it accordingly would make the code easier to follow.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+        MDTuple *TANode = dyn_cast<MDTuple>(TA->getOperand(i));
+        if (TANode->getNumOperands() < 5) {
+          ValCtx.EmitMetaError(TANode, ValidationRule::MetaWellFormed);
+          return;
+        }
+        // get access to the digit that represents the metadata number that
+        // would store entry properties
+        const llvm::MDOperand &mOp =
+            TANode->getOperand(TANode->getNumOperands() - 1);
+        // the final operand to the entry points tuple should be a constant int.
```
pow2clk, line 4008: By the code, it looks like it should be a tuple not a constant int.

If the last operand isn't a tuple, should we just ignore it? Perhaps another well formed metadata error or at least an assert?

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+        // get access to the node that stores entry properties
+        MDTuple *TANode2 =
+            dyn_cast<MDTuple>(TANode->getOperand(TANode->getNumOperands() - 1));
+        // find any incompatible tags inside the entry properties
+        // increment j by 2 to only analyze tags, not values
+        bool foundTag = false;
+        for (unsigned j = 0, end2 = TANode2->getNumOperands(); j < end2;
+             j += 2) {
+          // if this tag is already present in the current entry point, emit an
+          // error
```
pow2clk, line 4023: If you do break this code out into a separate function, you might take a pass at awkward comment breaks like this since they should be less deeply embedded in that case.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+
+          const MDOperand &mOp2 = TANode2->getOperand(j);
```
pow2clk, line 4025: a name like `propetyTagOp` would be clearer.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+          // note, we are only looking for tags, which will be a constant
+          // integer
+          if (mOp2 == nullptr || (mOp2.get())->getMetadataID() !=
+                                     Metadata::ConstantAsMetadataKind) {
+            continue;
```
pow2clk, line 4030: Unless there's a reason why this is expected to happen, we should at least have an assert here.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+          }
+          ConstantInt *tag = mdconst::extract<ConstantInt>(mOp2);
+          uint64_t tagValue = tag->getZExtValue();
+
+          // legacy wavesize is only supported between 6.6 and 6.7, so we
+          // should fail if we find the ranged wave size metadata tag
+          if (tagValue == DxilMDHelper::kDxilRangedWaveSizeTag) {
+            if (foundTag) {
+              ValCtx.EmitFormatError(ValidationRule::SmWaveSizeTagDuplicate,
+                                     {});
```
pow2clk, line 4040: another awkward line break that might get cleared up with a separate function.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+            foundTag = true;
+            if (SM->IsSM66Plus() && !SM->IsSM68Plus()) {
+
+              ValCtx.EmitFormatError(
+                  ValidationRule::SmWaveSizeRangeNeedsSM68Plus, {});
+              return;
+            }
+            // get the metadata that contains the parameters to the wavesize
+            // attribute
+            MDTuple *TANode3 = dyn_cast<MDTuple>(TANode2->getOperand(j + 1));
```
pow2clk, line 4052: Something like `waveSizeParams` or `waveSizeParamsNode` would be clearer.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3988,6 +3988,114 @@ static void ValidateMetadata(ValidationContext &ValCtx) {
...
+              if (param->getMetadataID() != Metadata::ConstantAsMetadataKind) {
+                ValCtx.EmitFormatError(
+                    ValidationRule::SmWaveSizeNeedsConstantOperands, {});
+                return;
+              }
+            }
+
+          } else if (tagValue == DxilMDHelper::kDxilWaveSizeTag) {
+            // if the shader model is anything but 6.6 or 6.7, then we do not
+            // expect to encounter the legacy wave size tag.
```
pow2clk, line 4069: I think this comment would be more useful above the conditional that checks for this.

tools/clang/test/CodeGenDXIL/hlsl/wavesize/wavesize-range-not-int.ll
```diff
@@ -0,0 +1,42 @@
+; XFAIL:*
```
pow2clk, line 1: A comment explaining why this is expected to fail and referencing a bug would be helpful

tools/clang/test/CodeGenDXIL/hlsl/wavesize/wavesize-range-not-three-params.ll
```diff
@@ -1,3 +1,4 @@
+; XFAIL:*
```
pow2clk, line 1: As above. I think we should have comments for all XFAILs

pow2clk (commented):
Acknowledging that there are still some weird failures on non-windows platforms, as far as I can see the implementation and tests look good. My comments are all about readability and mostly variable name suggestions that you might include with the fix for those failures.

lib/HLSL/DxilValidation.cpp
bob80905: I copied / was inspired to write the code the way it was by ValidateTypeAnnotation(). I suppose TA stands for TypeAnnotation. I'll perform some renames, because this has nothing to do with the type annotations :)

lib/HLSL/DxilValidation.cpp
bob80905: Correct, I've adjusted the comment.
I believe the check is necessary on the final operand because there are some entry point functions that don't have entry point properties, and so the operand is null. In fact, that is how I decided to add that line of code, since I was working on an example that had an entry point without entry point properties. By adding an assert, 53 other tests will fail, so I will leave the continue; as is.

lib/HLSL/DxilValidation.cpp
bob80905: Added.

pow2clk (approved):
Thanks for all the responses! Looks good!

tools/clang/test/HLSLFileCheck/validation/wavesize/outdated-metadata-tag.ll
```diff
@@ -0,0 +1,42 @@
...
+; CHECK: error: WaveSize is valid only for Shader Model 6.6 and 6.7.
+!10 = !{i32 16}
+!11 = !{!"node01", i32 0}
+!12 = !{i32 32, i32 1, i32 1}
+!13 = !{!14}
+!14 = !{i32 1, i32 97, i32 2, !15}
+!15 = !{i32 0, i32 12, i32 1, !16}
+!16 = !{i32 0, i32 5, i32 1}
+!17 = !{i32 1, i32 1, i32 1}
+!18 = !{i32 0}
```
llvm-beanz, line 42: nit: See if you can configure your editor to add newlines at the end of files. For IR test cases this matters less, but for C/C++ sources this will produce warnings when compiled with Clang & GCC because the C standard dictates that files must end in a newline.

llvm-beanz (commented):
I threw some nitpick comments and one real comment (about the use of asserts).

You can take or leave the "nit" comments, but it would be nice to fix the assert usage.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3958,111 +3958,105 @@ static void ValidateBitcode(ValidationContext &ValCtx) {
...
+    for (unsigned j = 0, end2 = EPropNode->getNumOperands(); j < end2; j += 2) {
+      const MDOperand &propertyTagOp = EPropNode->getOperand(j);
+      // note, we are only looking for tags, which will be a constant
+      // integer
+      bool invalidPropertyTag =
+          propertyTagOp == nullptr || (propertyTagOp.get())->getMetadataID() !=
+                                          Metadata::ConstantAsMetadataKind;
+      DXASSERT(!invalidPropertyTag,
+               "tag operand should be a constant integer.");
+      if (invalidPropertyTag) {
```
llvm-beanz, line 3997: I think we're misusing `assert` here still. We're asserting that a condition is never true, but then we have code that handles the possibility of it being true.

If this condition really can't happen we should have an `assert` and we shouldn't need the continue. If it can happen, we should do something to report the error.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3994,9 +3994,7 @@ static void ValidateWaveSize(ValidationContext &ValCtx,
                                           Metadata::ConstantAsMetadataKind;
       DXASSERT(!invalidPropertyTag,
```
llvm-beanz, line 3995: This should cause a warning/error on an unused variable in Release builds. You'll either need to collapse the condition into the assert or apply `[[maybe_unused]]` to the variable declaration to silence the warning/error/

llvm-beanz (approved):

bob80905:
This issue surfaced and needs to be addressed:
https://github.com/microsoft/DirectXShaderCompiler/issues/6390

PR #6204, opened 2024-01-25
[SM6.8] Add validation for sampleCmpBias/Grad.

python3kgae:
Add sampleCmpBias/Grad to ValidateImmOperandForMathDxilOp.  
And add validation test.

### DIFF
--- a/tools/clang/test/CodeGenDXIL/validation/val-SampleCmpBiasGrad-shader-model-vs67.ll
+++ b/tools/clang/test/CodeGenDXIL/validation/val-SampleCmpBiasGrad-shader-model-vs67.ll
@@ -25,9 +25,6 @@
 ;
 ;
 
-; CHECK: Opcode SampleCmpBias not valid in shader model vs_6_7
-; CHECK: Opcode SampleCmpGrad not valid in shader model vs_6_7
-; CHECK: Opcode CalculateLOD not valid in shader model vs_6_7
 
 target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
 target triple = "dxil-ms-dx"
@@ -62,6 +59,9 @@ define void @main() {
   %14 = extractvalue %dx.types.CBufRet.f32 %12, 0
   %15 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %2, %dx.types.ResourceProperties { i32 2, i32 1028 })  ; AnnotateHandle(res,props)  resource: Texture2D<4xI32>
   %16 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %4, %dx.types.ResourceProperties { i32 32782, i32 0 })  ; AnnotateHandle(res,props)  resource: SamplerComparisonState
+
+; CHECK: Opcode SampleCmpBias not valid in shader model vs_6_7
+
   %17 = call %dx.types.ResRet.f32 @dx.op.sampleCmpBias.f32(i32 255, %dx.types.Handle %15, %dx.types.Handle %16, float %7, float %8, float undef, float undef, i32 -5, i32 7, i32 undef, float %14, float %13, float %11)  ; SampleCmpBias(srv,sampler,coord0,coord1,coord2,coord3,offset0,offset1,offset2,compareValue,bias,clamp)
   %18 = extractvalue %dx.types.ResRet.f32 %17, 0
   %19 = extractvalue %dx.types.CBufRet.f32 %10, 0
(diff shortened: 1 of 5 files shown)


pow2clk (approved):
I looked over the tests and compared them to the spec. I'm unsure how much benefit the vs tests provide since most samples are disallowed there anyway, but it can't hurt.

tools/clang/test/CodeGenDXIL/validation/val-SampleCmpBiasGrad-shader-model-vs67.ll
```diff
@@ -0,0 +1,148 @@
...
+  %10 = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle %6, i32 1)  ; CBufferLoadLegacy(handle,regIndex)
+  %11 = extractvalue %dx.types.CBufRet.f32 %10, 2
+  %12 = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle %6, i32 0)  ; CBufferLoadLegacy(handle,regIndex)
+  %13 = extractvalue %dx.types.CBufRet.f32 %12, 1
+  %14 = extractvalue %dx.types.CBufRet.f32 %12, 0
+  %15 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %2, %dx.types.ResourceProperties { i32 2, i32 1028 })  ; AnnotateHandle(res,props)  resource: Texture2D<4xI32>
+  %16 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %4, %dx.types.ResourceProperties { i32 32782, i32 0 })  ; AnnotateHandle(res,props)  resource: SamplerComparisonState
+  %17 = call %dx.types.ResRet.f32 @dx.op.sampleCmpBias.f32(i32 255, %dx.types.Handle %15, %dx.types.Handle %16, float %7, float %8, float undef, float undef, i32 -5, i32 7, i32 undef, float %14, float %13, float %11)  ; SampleCmpBias(srv,sampler,coord0,coord1,coord2,coord3,offset0,offset1,offset2,compareValue,bias,clamp)
+  %18 = extractvalue %dx.types.ResRet.f32 %17, 0
```
bob80905, line 66: Maybe the rest of the lines down to %31 aren't needed? Also, I think it would be a good thing to move the check lines down to the relevant line, maybe near %17?

tools/clang/test/CodeGenDXIL/validation/val-SampleCmpBiasGrad-shader-model-vs68.ll
```diff
@@ -0,0 +1,147 @@
+; RUN: not %dxv %s 2>&1 | FileCheck %s
```
bob80905, line 1: Won't filecheck's exit code be the code that lit uses to determine whether the test passed? why is 'not' needed here?

tools/clang/test/CodeGenDXIL/validation/val-SampleCmpBiasGrad.ll
```diff
@@ -0,0 +1,170 @@
...
+  %22 = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle %7, i32 1)  ; CBufferLoadLegacy(handle,regIndex)
+  %23 = extractvalue %dx.types.CBufRet.f32 %22, 2
+  %24 = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle %7, i32 0)  ; CBufferLoadLegacy(handle,regIndex)
+  %25 = extractvalue %dx.types.CBufRet.f32 %24, 0
+  %26 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %3, %dx.types.ResourceProperties { i32 2, i32 1033 })  ; AnnotateHandle(res,props)  resource: Texture2D<4xF32>
+  %27 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %5, %dx.types.ResourceProperties { i32 14, i32 0 })  ; AnnotateHandle(res,props)  resource: SamplerComparisonState
+  %28 = call %dx.types.ResRet.f32 @dx.op.sampleCmpBias.f32(i32 255, %dx.types.Handle %21, %dx.types.Handle %27, float %8, float %9, float undef, float undef, i32 -5, i32 7, i32 undef, float %25, float 1.500000e+2, float %23)  ; SampleCmpBias(srv,sampler,coord0,coord1,coord2,coord3,offset0,offset1,offset2,compareValue,bias,clamp)
```
bob80905, line 83: Again, I think if possible, lines that aren't needed should be removed, and I think every line before this point can be removed. And, I would prefer having the check line here, right under %28

tools/clang/test/CodeGenDXIL/validation/val-SampleCmpBiasGrad-shader-model-vs67.ll
python3kgae: Updated.

tools/clang/test/CodeGenDXIL/validation/val-SampleCmpBiasGrad-shader-model-vs68.ll
python3kgae: lit doesn't like command fail. not will make the command pass.
taef allows command to fail, so we don't need not for taef.

tools/clang/test/CodeGenDXIL/validation/val-SampleCmpBiasGrad.ll
python3kgae: Updated.

bob80905 (approved):

PR #6207, opened 2024-01-25
RDAT: Check flags and WaveSize for min SM; fix flag detection and merging

### LINKED ISSUE #6218: Per-function feature flags and min target in RDAT are incorrect
**Description**
RDAT has per-function feature requirement flags, compatible shader stage flags, and a minimum shader target that should take into account internal function calls and should adjust the minimum shader targets and legal shader stages based on optional feature usage.

Several issues were preventing these from correctly representing the requirements for the function:
- per-function DXIL Shader Flag collection has bugs:
  - missing some feature usage scenarios
  - using global resources to indicate flags that may not be applicable to the function.
  - can't account for stage-specific flags (`DerivativesInMeshAndAmplificationShaders`)
- Features used in internally called functions were not accounted for in the caller.
- Min shader model wasn't adjusted by any feature usage flags introduced since SM 6.2.
- Min shader model didn't take into account newer shader stages (ms/as introduced in SM 6.5).
- Min shader model couldn't account for stage-specific feature availability differences (derivatives used in CS, MS, or AS).
- WaveSize and WaveSizeRange fail to impact minimum shader target.

This also impacts module-level flags, since they are just the union of flags computed for each function, plus a few module-scope checks.

Fixing these issues can cause differences in the `DFCC_RuntimeData` (`RDAT`) part as well as DxilModule ShaderFlags (in metadata) and the DFCC_FeatureInfo (`SFI0`) parts.  This would be considered a validator-breaking change, since the validator uses the same code to compute these details, then makes sure the container matches these expectations.

**Steps to Reproduce**
The following reflection test demonstrates some of these bugs based on the difference between expected result from 1.7 and fixed 1.8 validator:
```hlsl
// RUN: %dxilver 1.8 | %dxc -T lib_6_8 %s | %D3DReflect %s | %FileCheck %s -check-prefixes=RDAT,RDAT18
// RUN: %dxilver 1.7 | %dxc -T lib_6_7 -validator-version 1.7 %s | %D3DReflect %s | FileCheck %s -check-prefixes=RDAT,RDAT17

// Ensure min shader target incorporates optional features used

// RDAT: FunctionTable[{{.*}}] = {

// SM 6.7+

// ShaderFeatureInfo_WriteableMSAATextures (0x40000000) = 1073741824

// RDAT-LABEL: UnmangledName: "rwmsaa"
// ShaderFeatureInfo_WriteableMSAATextures (0x40000000) = 1073741824
// RDAT:   FeatureInfo1: 1073741824
// RDAT:   FeatureInfo2: 0
// MinShaderTarget: (Library(6) << 16) + (SM 6.7 ((6 << 4) + 7)) = 0x60067 = 393319
// RDAT18: MinShaderTarget: 393319
// Old: 6.0
// RDAT17: MinShaderTarget: 393312

RWByteAddressBuffer BAB : register(u1, space0);
RWTexture2DMS<float, 1> T2DMS : register(u2, space0);

[noinline] export
void rwmsaa() {
  // Use dynamic resource to avoid MSAA flag on all functions issue in 1.7
  // RWTexture2DMS<float, 1> T2DMS = ResourceDescriptorHeap[0];
  BAB.Store(0, T2DMS.sample[1][uint2(1, 2)]);
}

// RDAT-LABEL: UnmangledName: "no_rwmsaa"
// RDAT18: FeatureInfo1: 0
// 1.7 Incorrectly reports feature use for function
// RDAT17: FeatureInfo1: 1073741824
// RDAT:   FeatureInfo2: 0
// MinShaderTarget: (Library(6) << 16) + (SM 6.0 ((6 << 4) + 0)) = 0x60060 = 393312
// RDAT: MinShaderTarget: 393312

export void no_rwmsaa() {
  BAB.Store(0, 0);
}

// RDAT-LABEL: UnmangledName: "rwmsaa_in_raygen"
// ShaderFeatureInfo_WriteableMSAATextures (0x40000000) = 1073741824
// Old: set because of global
// RDAT: FeatureInfo1: 1073741824
// RDAT:   FeatureInfo2: 0
// MinShaderTarget: (RayGeneration(7) << 16) + (SM 6.7 ((6 << 4) + 7)) = 0x70067 = 458855
// RDAT18: MinShaderTarget: 458855
// Old: 6.0
// RDAT17: MinShaderTarget: 458848

[shader("raygeneration")]
void rwmsaa_in_raygen() {
  rwmsaa();
}

// RDAT-LABEL: UnmangledName: "rwmsaa_not_used_in_raygen"
// ShaderFeatureInfo_WriteableMSAATextures (0x40000000) = 1073741824
// RDAT18: FeatureInfo1: 0
// Old: set because of global
// RDAT17: FeatureInfo1: 1073741824
// RDAT:   FeatureInfo2: 0
// MinShaderTarget: (RayGeneration(7) << 16) + (SM 6.3 ((6 << 4) + 3)) = 0x70063 = 458851
// RDAT18: MinShaderTarget: 458851
// Old: 6.0
// RDAT17: MinShaderTarget: 458848

[shader("raygeneration")]
void rwmsaa_not_used_in_raygen() {
  no_rwmsaa();
}
```

**Expected Behavior**
Optional feature use and potential optional feature use is always captured into flags.  Min shader model is correct when combining all called functions and any special-case logic required for a final entry point function of a particular kind.
For the example above, when expected is different than actual, expected values are shown in the RDAT18 checks.

**Actual Behavior**
See above description of cases not captured correctly, as well as the RDAT17 sections of the example test case above.

**Environment**
- DXC version: recent `main` branch (`dcb618a2f1b7f4265fc2b77c12291f76eb334ae3`).
- Host Operating System: any

tex3d:
Add ShaderKind::Last_1_8 for shader mask
Add shader model comments before flag groupings in DxilConstants.h and DxilShaderFlags.

Add missing flag checks for min shader model in RDAT function info. Move ShaderCompatInfo computation into DxilModule, propagate callee info.

Move computation of shader model requirements based on flags into DxilModule. Finalize requirements for entry functions in AdjustMinimumShaderModelAndFlags.

Fixes for function level flag tracking:
- DerivativesInMeshAndAmpShaders: use flag to track deriv use, then adjust for entry functions.
- hasUAVs: based on resource use in function instead of global resources.
- WriteableMSAATextures: based on use in function instead of global resources.
  - Also catch cases for dynamic res from any use by looking at create/annotate handle, not just the TextureStoreSample op.
- RaytracingTier1_1: move module-level detection to CollectShaderFlagsForModule
- Marked deriv and quad ops as being supported in node.
- Fixed SampleCmpBias to be considered gradient op.
- Update RDAT definitions to dump more useful info for testing

Fixes #6218.


tex3d:
If you start [here](https://github.com/microsoft/DirectXShaderCompiler/pull/6207/commits/7f7f2af0c9c0746c23cc2337b72485a96246cf51), then step through commits with [Next], it's a lot more digestible.

include/dxc/DXIL/DxilConstants.h
```diff
@@ -60,6 +60,22 @@ inline int CompareVersions(unsigned Major1, unsigned Minor1, unsigned Major2,
   return 0;
 }
 
+// Utility for updating major,minor to max of current and new.
+inline bool MaxOfShaderModels(unsigned &major, unsigned &minor,
```
python3kgae, line 64: Maybe move this to DxilShaderModel.h?

include/dxc/DXIL/DxilModule.h
```diff
@@ -404,6 +404,31 @@ class DxilModule {
...
+  void ComputeShaderCompatInfo();
+
+  // Clear previous ShaderCompatInfo for recomputation.
+  void ClearShaderCompatInfo();
+
+  // ShaderCompatInfo tracks requirements per-function, subsequently merged into
+  // final entry function requirements.
+  struct ShaderCompatInfo {
+    unsigned minMajor = 6, minMinor = 0;
+    unsigned mask = ((unsigned)1 << (unsigned)DXIL::ShaderKind::Invalid) - 1;
```
python3kgae, line 419: Comment for what the mask is would be nice.

include/dxc/DXIL/DxilConstants.h
```diff
@@ -1882,12 +1892,13 @@ const uint64_t ShaderFeatureInfo_AdvancedTextureOps = 0x20000000;
...
 
 // SM 6.8+
-// WaveMMA slots in between two SM 6.6 feature bits.
-const uint64_t ShaderFeatureInfo_WaveMMA = 0x8000000;
-
 const uint64_t ShaderFeatureInfo_SampleCmpGradientOrBias = 0x80000000;
 const uint64_t ShaderFeatureInfo_ExtendedCommandInfo = 0x100000000;
 
+// Experimental SM 6.9+ - Reserved, not yet supported.
+// WaveMMA slots in between two SM 6.6 feature bits.
```
coopp, line 1915: Is this comment necessary?  Do developers need to know where feature bits are taken from?

coopp: Oh.. I see that this was an existing code comment, not a new one.

lib/DXIL/DxilOperations.cpp
```diff
@@ -3249,23 +3249,24 @@ void OP::GetMinShaderModelAndMask(const llvm::CallInst *CI,
...
-    unsigned mode = barrier.get_barrierMode_val();
-    if (mode != (unsigned)DXIL::BarrierMode::UAVFenceGlobal) {
-      mask =
-          SFLAG(Library) | SFLAG(Compute) | SFLAG(Amplification) | SFLAG(Mesh);
+    // Barrier mode should be a constant, but be robust to non-constants here.
+    if (isa<ConstantInt>(CI->getArgOperand(DxilInst_Barrier::arg_barrierMode))) {
+      DxilInst_Barrier barrier(const_cast<CallInst *>(CI));
+      unsigned mode = barrier.get_barrierMode_val();
+      if (mode != (unsigned)DXIL::BarrierMode::UAVFenceGlobal) {
+        mask &= SFLAG(Library) | SFLAG(Compute) | SFLAG(Amplification) |
```
coopp, line 3268: Just double checking here.  The original code just assigned an OR'd set of flags directly 'mask = flag | flag | flag'.  This new logic is ANDing this set of bits now. 'mask &= flag | flag | flag'.  Is this expected?

include/dxc/DXIL/DxilConstants.h
coopp, line 64: MaxOfShaderModels( ) returns a bool, but it looks like the code below never checks that return.  Could this just be void or is there really something the code below needs to be aware of if 'false' is returned?

coopp: I see in other areas that use this helper, they do in fact check the bool which means that values have changed.

include/dxc/DXIL/DxilModule.h
```diff
@@ -404,6 +404,31 @@ class DxilModule {
...
+    ShaderFlags shaderFlags;
+    bool Merge(ShaderCompatInfo &other);
+  };
+
+  const ShaderCompatInfo *
+  GetCompatInfoForFunction(const llvm::Function *F) const;
+
+private:
+  typedef std::unordered_map<const llvm::Function *, ShaderCompatInfo>
```
coopp, line 428: Consider using llvm::DenseMap instead of std::unordered_map?  ( I have been reading coding guidelines for the llvm project and it suggests that replacement for map usage.. I see other llvm:: objects are being used in this code, so I wanted to make this suggestion.

include/dxc/DXIL/DxilConstants.h
tex3d: Maybe I should consider a rename instead.  It really isn't specific to DxilShaderModel, it's just trying to do an in-place max for a major/minor version pair.

How about `UpdateToMaxOfVersions`?

tex3d: Yeah, the way I tried to separate different changes made it so the first one didn't need the return value, but other later uses did.

include/dxc/DXIL/DxilModule.h
tex3d: I can do this, but there are several other pre-existing collections used in DxilModule for which the same argument stands.  This code uses a template already used elsewhere in DxilModule, and this was a move of existing code that used this map type, so perhaps a decision to change the collection types in DxilModule could be made separately?  What do you think?

lib/DXIL/DxilOperations.cpp
tex3d: I believe the `&=` is a safer pattern to use whenever further restricting the mask based on any special case following the `GetMinShaderModelAndMask` call, which initializes the mask and should set it based on specific instruction requirements.

Special cases like this need to restrict the mask, not accidentally expand it.  The generated code for the DXIL op should have set the maximally permissive mask for the operation, which we might need to restrict with some special cases here.

Assignment would have been equivalent in this case, but it's not a pattern this code should use unless we are trying to match a prior validator in a special case for compatibility, like the earlier code does for wave ops.

However, this comment made take a second look at another case in this function, and I realized it needed updating, so thanks for that!  I'll update that in a little while after I can craft and verify a test that hits that case.

coopp (approved):

lib/DXIL/DxilModule.cpp
```diff
@@ -2081,4 +2111,157 @@ bool DxilModule::IsPrecise(const Instruction *inst) const {
...
+  if (DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 5) < 0) {
+    ValidShaderMask = (1 << ((unsigned)DXIL::ShaderKind::Last_1_4 + 1)) - 1;
+  } else if (DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 8) < 0) {
+    ValidShaderMask = (1 << ((unsigned)DXIL::ShaderKind::Last_1_7 + 1)) - 1;
+  } else if (DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 8) >= 0) {
+    ValidShaderMask = (1 << ((unsigned)DXIL::ShaderKind::Last_1_8 + 1)) - 1;
+  }
+
+  for (auto &function : GetModule()->getFunctionList()) {
+    if (!function.isDeclaration()) {
```
pow2clk, line 2190: Nit: It would be a bit easier to follow the indentation if this were accomplished using something like ` if (function.isDeclaration()) continue;`

lib/DXIL/DxilModule.cpp
```diff
@@ -2081,4 +2111,157 @@ bool DxilModule::IsPrecise(const Instruction *inst) const {
...
+      }
+
+      if (shaderKind == DXIL::ShaderKind::Library)
+        info.mask &= ValidShaderMask;
+      else
+        info.mask &= (1U << static_cast<unsigned>(shaderKind));
+
+      // Increase min target based on features used:
+      ShaderFlags &flags = info.shaderFlags;
+      if (DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 8) >= 0) {
```
pow2clk, line 2206: There are a few of these checks for greater than 1.8 sprinkled in this function. I can't tell quickly if this function has any effect with a version less than 1.8. Regardless, can any be consolidated for maintainability and legibility?

include/dxc/DXIL/DxilConstants.h
```diff
@@ -1881,14 +1908,30 @@ const uint64_t ShaderFeatureInfo_AdvancedTextureOps = 0x20000000;
...
+const uint64_t ShaderFeatureInfo_WaveMMA = 0x8000000;
+
+// Maximum count without rolling over into another 64-bit field is 40,
+// so the last flag we can use for a feature requirement is: 0x8000000000
+// This is because of the following set of flags, considered optional
+// and ignored by the runtime if not recognized:
+// D3D11_OPTIONAL_FEATURE_FLAGS 0x7FFFFF0000000000
 const unsigned ShaderFeatureInfoCount = 33;
```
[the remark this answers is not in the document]
tex3d: I don't know how we can get an assert for this.  The comment is here where we define these flags so when a new flag is defined, hopefully this will let us know.  I'm open to suggestions for a static assert if you have any ideas, of course.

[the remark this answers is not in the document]
tex3d: Oh, sorry, I realize you're talking about `ShaderFeatureInfoCount`, duh.  I can static_assert that.  I also have to double-check that it's ok to have the optional flag outside this set.

lib/DXIL/DxilModule.cpp
```diff
@@ -2081,4 +2111,157 @@ bool DxilModule::IsPrecise(const Instruction *inst) const {
...
+    changed = true;
+  }
+  uint64_t rawBefore = shaderFlags.GetShaderFlagsRaw();
+  shaderFlags.CombineShaderFlags(other.shaderFlags);
+  if (rawBefore != shaderFlags.GetShaderFlagsRaw())
+    changed = true;
+  return changed;
+}
+
+void DxilModule::ClearShaderCompatInfo() { m_FuncToShaderCompat.clear(); }
```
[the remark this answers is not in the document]
tex3d: Yeah, it was intended to be used after changes are made to the functions in the module that might require recomputation, such as after a PIX pass or something like that, but I'll call it at the beginning of `CollectShaderFlagsForModule`, since that will also guarantee it's fresh when computing the overall module flags.

lib/DXIL/DxilModule.cpp
```diff
@@ -342,6 +349,29 @@ void DxilModule::CollectShaderFlagsForModule(ShaderFlags &Flags) {
...
+      for (const auto &it : pSubobjects->GetSubobjects()) {
+        switch (it.second->GetKind()) {
+        case DXIL::SubobjectKind::RaytracingPipelineConfig1:
+          Flags.SetRaytracingTier1_1(true);
+          break;
+        case DXIL::SubobjectKind::StateObjectConfig: {
+          uint32_t configFlags;
+          if (it.second->GetStateObjectConfig(configFlags) &&
+              ((configFlags &
+                ~(unsigned)DXIL::StateObjectFlags::ValidMask_1_4) != 0))
```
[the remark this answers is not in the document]
tex3d: I moved this code unchanged, but I can update it.  I'll add a new mask that include the 1.1 flags I think.

lib/DXIL/DxilModule.cpp
```diff
@@ -342,6 +349,29 @@ void DxilModule::CollectShaderFlagsForModule(ShaderFlags &Flags) {
...
+            Flags.SetRaytracingTier1_1(true);
+        } break;
+        default:
+          break;
+        }
+        if (Flags.GetRaytracingTier1_1())
+          break;
```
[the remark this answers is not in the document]
tex3d: Again, I moved this code unchanged.  I'll move it into a separate static function.

lib/DXIL/DxilModule.cpp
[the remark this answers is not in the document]
tex3d: Or maybe I'll just test for the one 1.1 flag.

lib/DXIL/DxilModule.cpp
```diff
@@ -2081,4 +2111,157 @@ bool DxilModule::IsPrecise(const Instruction *inst) const {
...
+  shaderFlags.CombineShaderFlags(other.shaderFlags);
+  if (rawBefore != shaderFlags.GetShaderFlagsRaw())
+    changed = true;
+  return changed;
+}
+
+void DxilModule::ClearShaderCompatInfo() { m_FuncToShaderCompat.clear(); }
+
+void DxilModule::ComputeShaderCompatInfo() {
+  // Initialize worklist with functions with callers
```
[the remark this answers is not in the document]
tex3d: ok

lib/DXIL/DxilModule.cpp
```diff
@@ -2081,4 +2111,157 @@ bool DxilModule::IsPrecise(const Instruction *inst) const {
...
+  SmallSetVector<llvm::Function *, 8> worklist;
+
+  for (auto &function : GetModule()->getFunctionList()) {
+    if (function.isDeclaration() && !function.isIntrinsic() &&
+        function.getLinkage() ==
+            llvm::GlobalValue::LinkageTypes::ExternalLinkage &&
+        OP::IsDxilOpFunc(&function)) {
+      // update min shader model and shader stage mask per function
+      UpdateFunctionToShaderCompat(&function);
+    } else if (!function.isDeclaration()) {
```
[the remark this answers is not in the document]
tex3d: ok

lib/DXIL/DxilModule.cpp
```diff
@@ -2081,4 +2111,157 @@ bool DxilModule::IsPrecise(const Instruction *inst) const {
...
+      // Initialize worklist with functions with callers.
+      // only used for validator version 1.8+
+      if (DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 8) >= 0 &&
+          !function.user_empty())
+        worklist.insert(&function);
+
+      // Skip functions with shader flags already set
+      if (m_FuncToShaderCompat.count(&function) != 0)
```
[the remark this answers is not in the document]
tex3d: Yes, we still want to add it to the worklist, as another function that doesn't have flags set may call this one.  We go from callee to caller when combining flags.

lib/DXIL/DxilModule.cpp
```diff
@@ -2081,4 +2111,157 @@ bool DxilModule::IsPrecise(const Instruction *inst) const {
...
+        continue;
+
+      // Collect shader flags for function
+      ShaderCompatInfo &info = m_FuncToShaderCompat[&function];
```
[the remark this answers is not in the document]
tex3d: We do this quite a bit elsewhere, but I'll add a comment.

lib/DXIL/DxilModule.cpp
tex3d: ok

lib/DXIL/DxilModule.cpp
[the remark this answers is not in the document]
tex3d: I'm removing this because I realize it could cause functions to skip flag collection, if an intrinsic used in the function was iterated first.  That also means I'm not bothering with retaining results from a prior run, so I'll clear the map each time this is called and get rid of the `ClearShaderCompatInfo` function as well.

lib/DXIL/DxilModule.cpp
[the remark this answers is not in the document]
tex3d: Actually, this *may* add info, or the info may already be added when traversing an intrinsic users first.

lib/DXIL/DxilModule.cpp
tex3d: It does have an effect for < 1.8.  The 1.8 checks are to make sure we generate the same result in validator 1.7 and below.

The effects when < 1.8 are:
- CollectShaderFlags for each function
- Updates each functions min target and valid shader stage flags based on intrinsic usage via UpdateFunctionToShaderCompat on intrinsics to update users.
- Masks shader stage flags based on ShaderKind for a shader entry function, or based on stages that exist in the shader model for a library function.
- Updates min target based on use of a few optional feature flags (6.2 for native low precision, 6.1 for barycentrics or ViewID)

With >= 1.8, we also:
- propagate flags from internal calees to callers
- use AdjustMinimumShaderModelAndFlags to account for all feature flags, shader stages, WaveSize use, and handle the special case for derivative use.

lib/DXIL/DxilModule.cpp
```diff
@@ -2081,4 +2111,157 @@ bool DxilModule::IsPrecise(const Instruction *inst) const {
...
+        if (flags.GetUseNativeLowPrecision() && flags.GetLowPrecisionPresent())
+          DXIL::UpdateToMaxOfVersions(info.minMajor, info.minMinor, 6, 2);
+        else if (flags.GetBarycentrics() || flags.GetViewID())
+          DXIL::UpdateToMaxOfVersions(info.minMajor, info.minMinor, 6, 1);
+      }
+    }
+  }
+}
+
+void DxilModule::UpdateFunctionToShaderCompat(const llvm::Function *dxilFunc) {
```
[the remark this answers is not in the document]
tex3d: Yeah, good point for next time.  For this one, I believe I just added ` | SFLAG(Node)`, which was missing before, and encapsulated the version update in `DXIL::UpdateToMaxOfVersions`.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm67_writeablemsaa.hlsl
```diff
@@ -0,0 +1,52 @@
...
+// ShaderFeatureInfo_WriteableMSAATextures (0x40000000) = 1073741824
+// ResourceDescriptorHeapIndexing (0x2000000) = 33554432
+// 0x40000000 + 0x2000000 = 0x42000000 = 1107296256
+// RDAT18:   FeatureInfo1: 1107296256
+// Old: missed use of WriteableMSAATextures
+// RDAT17:   FeatureInfo1: 33554432
+// RDAT:   FeatureInfo2: 0
+// MinShaderTarget: (Library(6) << 16) + (SM 6.7 ((6 << 4) + 7)) = 0x60067 = 393319
+// RDAT18: MinShaderTarget: 393319
+// Old: 6.6
```
pow2clk, line 23:
```suggestion
// Old: 6.6 (Because of dynamic resources)
```

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm67_writeablemsaa_global.hlsl
```diff
@@ -0,0 +1,74 @@
...
+// Old: 6.0
+// RDAT17: MinShaderTarget: 393312
+
+RWByteAddressBuffer BAB : register(u1, space0);
+RWTexture2DMS<float, 1> T2DMS : register(u2, space0);
+
+[noinline] export
+void rwmsaa() {
+  // Use dynamic resource to avoid MSAA flag on all functions issue in 1.7
+  // RWTexture2DMS<float, 1> T2DMS = ResourceDescriptorHeap[0];
```
pow2clk, line 29: These seem irrelevant here. 
```suggestion
```

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm68_node_barrier.hlsl
```diff
@@ -0,0 +1,52 @@
+// RUN: %dxilver 1.8 | %dxc -T lib_6_8 %s | %D3DReflect %s | %FileCheck %s -check-prefixes=RDAT
+
+// Ensure that barrier is allowed for node shaders.
```
pow2clk, line 3: I guess it has that side effect, but the checks seem to be, like the other tests to determine that RDAT data correctly sets the version due to node shaders.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm68_node_deriv.hlsl
```diff
@@ -0,0 +1,222 @@
+// RUN: %dxilver 1.8 | %dxc -T lib_6_8 %s | %D3DReflect %s | %FileCheck %s -check-prefixes=RDAT
+
+// Ensure that categories of deriv ops are allowed for node shaders.
```
pow2clk, line 3: As with the previous, this at least seems to check that and that the OptFeatureInfo_UsesDerivatives  flag gets set. 
```suggestion
// Ensure that the OptFeatureInfo_UsesDerivatives flag gets set
// for different categories of deriv ops in node shaders.
```

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm6x_wavemma.hlsl
```diff
@@ -0,0 +1,53 @@
...
+// + WaveOps (0x4000) = 0x8004000 = 134234112
+// RDAT:   FeatureInfo1: 134234112
+// RDAT:   FeatureInfo2: 0
+// MinShaderTarget: (Library(6) << 16) + (SM 6.9 ((6 << 4) + 9)) = 0x60069 = 393321
+// RDAT: MinShaderTarget: 393321
+
+[noinline] export
+void call_use_wavematrix() {
+  use_wavematrix();
+}
```
pow2clk, line 39: I'm curious why a second level of indirection was useful here and nowhere else.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/uavs_at_every_stage.hlsl
pow2clk, line 1: This looks to be an exact copy of sm67_advancedtextureops.hlsl. I think you meant to modify it.

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
```diff
@@ -783,8 +783,15 @@ FileRunCommandPart::RunD3DReflect(dxc::DxcDllSupport &DllSupport,
   IFT(DllSupport.CreateInstance(CLSID_DxcLibrary, &pLibrary));
   IFT(DllSupport.CreateInstance(CLSID_DxcAssembler, &pAssembler));
 
-  IFT(pLibrary->CreateBlobWithEncodingFromPinned(
-      (LPCVOID)Prior->StdOut.c_str(), Prior->StdOut.size(), CP_UTF8, &pSource));
+  // It would be better to properly honour %s for input, but for now, if prior
+  // output is empty, use the file as input.
```
pow2clk, line 787: I'd like to understand why this was necessary

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -420,16 +438,22 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
 
   bool hasAdvancedTextureOps = false;
-  bool hasWriteableMSAATextures = false;
+  bool hasWriteableMSAATexturesGlobal = false;
   bool hasSampleCmpGradientOrBias = false;
 
   bool hasWaveMMA = false;
   bool hasExtendedCommandInfo = false;
 
+  // UsesDerivatives is used to indicate any derivative use per-function, before
```
pow2clk, line 450: Typo I think
```suggestion
  // hasDerivatives is used to indicate any derivative use per-function, before
```

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -778,17 +840,86 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
   flag.SetSamplerDescriptorHeapIndexing(hasSamplerDescriptorHeapIndexing);
   flag.SetAtomicInt64OnHeapResource(hasAtomicInt64OnHeapResource);
   flag.SetAdvancedTextureOps(hasAdvancedTextureOps);
-  flag.SetWriteableMSAATextures(hasWriteableMSAATextures);
+  flag.SetWriteableMSAATextures(setWriteableMSAATexturesBasedOnGlobal
+                                    ? hasWriteableMSAATexturesGlobal
+                                    : hasWriteableMSAATextures);
```
pow2clk, line 845: This check seems unnecessary and inconsistent as there's already a place above that will set `hasWriteableMSAATextures` to true if `setWiteableMSAATexturesBasedOnGlobal` is true and there are any RWMS textures in the module. 

Is there a case where there aren't RWMS textures in the module and yet there are somehow handle creation calls for a RWMS texture? That seems the only case where `hasWriteableMSAATexturesGlobal` would be false, but somehow `hasWriteableMSAATextures` would be true and we want the false because `setWiteableMSAATexturesBasedOnGlobal` is set. 

Said another way, I thinkt hat if `hasWriteableMSAATextures` is true, then `hasWriteableMSAATexturesGlobal` _must_ be true, so I don't know why you wouldn't just use `hasWriteableMSAATextures` here without any contidional. 
```suggestion
  flag.SetWriteableMSAATextures(hasWriteableMSAATextures);
```

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -778,17 +840,86 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
+  flag.SetUsesDerivatives(hasDerivatives);
 
   return flag;
 }
 
 void ShaderFlags::CombineShaderFlags(const ShaderFlags &other) {
   SetShaderFlagsRaw(GetShaderFlagsRaw() | other.GetShaderFlagsRaw());
 }
+
+ShaderFlags
```
pow2clk, line 861: As many times as I've cursed when I came to a called function without any description above it, I would like a description above this function. 
```suggestion
// Use the function properties `props` to determine the minimum shader model and
// flag requirements based on shader stage and feature usage.
// Compare that minimum required version to the values passed in with
// `minMajor` and `minMinor` and pass the maximum of those back through those
// same variables.
// Return a copy of the `ShaderFlags` object with any additional flags required
// by `props` set.
ShaderFlags
```
Note that if the returned value of this changes because of my comment below, the last line will have to change too.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1829,50 +1768,26 @@ class DxilRDATWriter : public DxilPartWriter {
                                              TGSMInFunc[&function]);
             } else if (DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 8) >
                        0) {
```
pow2clk, line 1770: Is this right or should it be `>=`? Maybe it's part of an experimental feature? 
```suggestion
            } else if (DXIL::CompareVersions(m_ValMajor, m_ValMinor, 1, 8) >=
                       0) {
```

pow2clk (commented):
I think with a long change like this, my comments tend toward complaining about how hard it is to read and review. Sorry about that. 

I tried to make suggested changes where possible. Overall, this improves more than I expected to fix this bug. Now that it's done, that's great!

I don't think much needs to change, I just have a couple concerns and many readability suggestions.

lib/DxilContainer/DxilContainerAssembler.cpp
tex3d: It's not included for 1.8, it's part of the experimental/potential next version.

lib/DXIL/DxilOperations.cpp
```diff
@@ -3230,42 +3225,42 @@ void OP::GetMinShaderModelAndMask(const llvm::CallInst *CI,
...
+    }
   }
 
   if (DXIL::CompareVersions(valMajor, valMinor, 1, 5) < 0) {
     // validator 1.4 didn't exclude wave ops in mask
     if (IsDxilOpWave(opcode))
-      mask = ((unsigned)1 << (unsigned)DXIL::ShaderKind::Invalid) - 1;
-    // These shader models don't exist before 1.5
-    mask &= ~(SFLAG(Amplification) | SFLAG(Mesh));
+      mask = ((unsigned)1 << (unsigned)DXIL::ShaderKind::Mesh) - 1;
```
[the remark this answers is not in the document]
tex3d: This is merely to maintain compatibility with validator version < 1.4.  It had a bug where it said wave ops were available to all shader stages.  However, the previous workaround using `((unsigned)1 << (unsigned)DXIL::ShaderKind::Invalid) - 1` would produce flags that included mesh and amp shaders when we added them.  That was incorrect, since in 1.4, mesh and amp shaders didn't exist.  So instead of including all stages, I'm including only the ones that existed at the time.

lib/DXIL/DxilShaderFlags.cpp
tex3d: `UsesDerivatives` is the name of the flag, and intentional.  `hasDerivatives` is the name of the bool variable that we use locally to keep track of whether we found derivative usage, but may or may not result in the UsesDerivatives flag being set.

lib/DXIL/DxilShaderFlags.cpp
tex3d: It *is* possible to use a dynamic resource so `hasWriteableMSAATextures` can be true with `hasWriteableMSAATexturesGlobal` being false.

However, revisiting this, I think the compatibility code won't quite get it right either.  Before, the flag would be set if either there was a global resource, or if [TextureStoreSample was used](https://github.com/microsoft/DirectXShaderCompiler/pull/6207/files#diff-d35f0d00e33e546a948baeacb7cbd7b6e5d3ebf0650bf93bd354e0f47821c20aR710).  But there are more potential uses of the resource than `TextureStoreSample`, so I added code to catch any other use.  However, since the `TextureStoreSample` isn't counted separately now, neither flag captures the exact behavior of 1.7.

I think I'm going to rename `hasWriteableMSAATexturesGlobal` to `hasWriteableMSAATextures_1_7`, then set this based on global, and set it when `TextureStoreSample` is encountered, then switch between the flags based on that.  Looks like another subtle test case is needed, ugh..

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm66_atomicint64.hlsl
```diff
@@ -0,0 +1,96 @@
...
+  GroupMemoryBarrierWithGroupSync();
+  uint64_t original;
+  InterlockedExchange(gs, tidx, original);
+  BAB.Store(tidx * 4, original);
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// ShaderFeatureInfo_AtomicInt64OnHeapResource (0x10000000) = 268435456
+
+// TBD: should it be set on bound heap resource?
```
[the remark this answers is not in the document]
tex3d: Oh yeah, we only set this if a dynamic heap resource is used with AtomicInt64.  I remember that was necessary, and was wondering why it wasn't set when you used a bound heap resource instead.  But we set a flag on the resource itself indicating this usage when it's a bound resource, and this flag makes it into PSV0 and RDAT resource flags, so I guess it wasn't necessary.  It was just a bit confusing because the feature can be used without incurring the feature flag requirement here.  I think it's by-design, but it made me go hmmm?

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm66_derivatives.hlsl
```diff
@@ -0,0 +1,131 @@
...
+
+// Ensure min shader target incorporates optional features used
+
+// RDAT: FunctionTable[{{.*}}] = {
+
+// SM 6.6+
+
+///////////////////////////////////////////////////////////////////////////////
+// ShaderFeatureInfo_DerivativesInMeshAndAmpShaders (0x1000000) = 16777216
+// OptFeatureInfo_UsesDerivatives (0x0000010000000000) = FeatureInfo2: 256
```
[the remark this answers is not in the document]
tex3d: No, not to reflection API actually.  This is an RDAT dump which is part of our D3DReflect dump for testing purposes, but is not exposed through our reflection API.  The flag is explicitly masked off for the reflection API, so there is no API surface impacted.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm66_derivatives_lib_6_6.ll
```diff
@@ -0,0 +1,218 @@
...
+; RDAT18: ShaderStageFlag: 57441
+; Old would not report Compute, Mesh, Amplification, or Node compatibility.
+; Pixel(0), Library(6) = 0x41 = 65
+; RDAT16: ShaderStageFlag: 65
+; RDAT17: ShaderStageFlag: 65
+; MinShaderTarget: (Library(6) << 16) + (SM 6.0 ((6 << 4) + 0)) = 0x60060 = 393312
+; RDAT17: MinShaderTarget: 393312
+; RDAT18: MinShaderTarget: 393312
+; Old: Didn't set min target properly for lib function
+; RDAT16: MinShaderTarget: 393318
```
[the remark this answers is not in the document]
tex3d: The issue is that I had to split these into three files for different versions, and they needed the identical checks that the HLSL file had.  I wanted to preserve the checks as-is so copy-paste between would work.  If I tailor each to each version, it would be a lot harder to edit and maintain.  I wish there was a way to do multiple RUN lines with preprocessing and ifdef/macro expansion on the source for .ll tests.  Then I could use an include file for common checks and/or code.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm66_derivatives_lib_6_8_decl_intrin_first.ll
```diff
@@ -0,0 +1,223 @@
...
+; RDAT18: ShaderStageFlag: 57441
+; Old would not report Compute, Mesh, Amplification, or Node compatibility.
+; Pixel(0), Library(6) = 0x41 = 65
+; RDAT16: ShaderStageFlag: 65
+; RDAT17: ShaderStageFlag: 65
+; MinShaderTarget: (Library(6) << 16) + (SM 6.0 ((6 << 4) + 0)) = 0x60060 = 393312
+; RDAT17: MinShaderTarget: 393312
+; RDAT18: MinShaderTarget: 393312
+; Old: Didn't set min target properly for lib function
+; RDAT16: MinShaderTarget: 393318
```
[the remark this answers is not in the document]
tex3d: See earlier comment regarding the ability to copy-paste between these that have different versions.  It's much easier to maintain if they have the same checks, and it's much easier to see the expected version differences side-by-side.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm68_node_barrier.hlsl
tex3d: This one is to test the bug where node wasn't added to the mask of allowed shader stages for this barrier operation.  This test ensures the function is still considered compatible with node shader when the barrier is used.

In fact, this is the fix that's being waited on so that Amar can enable the RDAT validation in the runtime, which currently breaks node shader tests due to this bug.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm68_node_deriv.hlsl
tex3d: I can add that it checks that the `OptFeatureInfo_UsesDerivatives` flag gets set as well, but this is also to explicitly regression test that node shaders are considered compatible with different categories of derivative ops, since that was another bug fixed (some deriv op categories didn't say that they were supported in node shaders, see update in hctdb.py).

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/uavs_at_every_stage.hlsl
tex3d: I didn't mean to add it.  I was going to use it to test the change to that flag, but it turns out putting the metadata check in the sm66_derivatives.hlsl test did double duty checking this and making sure the UsesDerivatives flag was masked away from the global flags.  I have an update deleting this file incoming soon.

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
tex3d: It was necessary to input raw IR.  Normally, `%D3DReflect` follows a compilation (`%dxc`) or link (`%dxl`) where the resulting IR is dumped to the `Prior->StdOut`.  But if I want to start with IR, there is no prior stage dumping this IR to `Prior->StdOut`.  Ideally, I might do `%dxa` but that doesn't exist here, so I did this as a workaround for the problem.  The right fix would be to use the file if `%s` was specified, otherwise use the `Prior->StdOut`.  But to make that change, I'd have to change a whole lot more tests to remove the `%s` from `%D3DReflect %s` in their RUN lines.  I was thinking of doing that in a follow-up change.

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -778,17 +840,86 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
+    DXIL::UpdateToMaxOfVersions(minMajor, minMinor, 6, 7);
+  else if (GetAtomicInt64OnTypedResource() || GetAtomicInt64OnGroupShared() ||
+           GetAtomicInt64OnHeapResource() ||
+           GetResourceDescriptorHeapIndexing() ||
+           GetSamplerDescriptorHeapIndexing())
+    DXIL::UpdateToMaxOfVersions(minMajor, minMinor, 6, 6);
+  else if (GetRaytracingTier1_1() || GetSamplerFeedback())
+    DXIL::UpdateToMaxOfVersions(minMajor, minMinor, 6, 5);
+  else if (GetShadingRate())
+    DXIL::UpdateToMaxOfVersions(minMajor, minMinor, 6, 4);
```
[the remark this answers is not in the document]
tex3d: I'm not sure what you mean.

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -778,17 +840,86 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
+    // entry point we have. Then, we may need to clear the flag, when it doesn't
+    // apply.
+
+    if (GetUsesDerivatives()) {
+      if (props->IsCS()) {
+        // Always supported if SM 6.6+.
+        DXIL::UpdateToMaxOfVersions(minMajor, minMinor, 6, 6);
+      } else if (props->IsMS() || props->IsAS()) {
+        // Requires flag for support on SM 6.6+.
+        flags.SetDerivativesInMeshAndAmpShaders(true);
```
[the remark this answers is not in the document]
tex3d: I suppose it could be elsewhere, but where?  Perhaps DxilModule?

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -2837,59 +2847,66 @@ HRESULT CFunctionReflection::GetDesc(D3D12_FUNCTION_DESC *pDesc) {
...
+  // Unset: UINT MacroInstructionCount; // Number of macro instructions used
+  // Unset: UINT ArrayInstructionCount; // Number of array instructions used
+  // Unset: UINT MovInstructionCount; // Number of mov instructions used
+  // Unset: UINT MovcInstructionCount; // Number of movc instructions used
+  // Unset: UINT ConversionInstructionCount; // Number of type conversion
+  //                                            instructions used
+  // Unset: UINT BitwiseInstructionCount; // Number of bitwise arithmetic
+  //                                         instructions used
+  // Unset: D3D_FEATURE_LEVEL MinFeatureLevel; // Min target of the function
+  //                                              byte code
```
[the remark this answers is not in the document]
tex3d: Yeah, I thought that at the time, but I saw it and felt I needed to fix it while I was here.  I could try to extract into another PR, if you'd like.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm6x_wavemma.hlsl
tex3d: I honestly don't recall at the moment.  I think I wanted to make sure the exported function picked up the called function as well in this case, but I don't really remember now.  It seems useful to have at least one test here that does two levels of indirection though, so I'm inclined to leave it.

tools/clang/test/HLSLFileCheck/d3dreflect/empty_broadcasting_nodes.hlsl
```diff
@@ -16,7 +16,7 @@
 // CHECK:      FeatureInfo1: 0
 // CHECK:      FeatureInfo2: 0
 // CHECK:      ShaderStageFlag: 32768
-// CHECK:      MinShaderTarget: 983136
+// CHECK:      MinShaderTarget: 983144
```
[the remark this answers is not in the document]
tex3d: I changed the dump output to make this much easier to check.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm66_atomicint64.hlsl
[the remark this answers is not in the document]
tex3d: I removed this comment.

lib/DXIL/DxilShaderFlags.cpp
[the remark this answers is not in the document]
pow2clk: We missed the minimum version required in a lot of cases. Will any full shaders require later versions as a result of this or is it just functions showing up in RDAT?

lib/DXIL/DxilShaderFlags.cpp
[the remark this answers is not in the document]
pow2clk: I don't have something in mind. It's probably not worth it. I was just thinking out loud.

lib/DXIL/DxilModule.cpp
```diff
@@ -288,11 +288,39 @@ unsigned DxilModule::GetGlobalFlags() const {
...
-  };
+  ComputeShaderCompatInfo();
+  for (auto &itInfo : m_FuncToShaderCompat)
+    Flags.CombineShaderFlags(itInfo.second.shaderFlags);
+
+  // Clear UsesDerivatives flag for module, making sure
+  // DerivativesInMeshAndAmpShaders is set for MS/AS.
+  if (Flags.GetUsesDerivatives()) {
+    Flags.SetUsesDerivatives(false);
+    if (m_pSM->IsMS() || m_pSM->IsAS())
```
python3kgae, line 321: Why compute is not here?

pow2clk: Advanced Texture Ops made derivatives in mesh and amplification shaders a separate feature bit. This is to check if that bit is needed.

pow2clk (approved):
Love the pretty printing! Thanks for all the responses!

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
```diff
@@ -783,8 +783,15 @@ FileRunCommandPart::RunD3DReflect(dxc::DxcDllSupport &DllSupport,
   IFT(DllSupport.CreateInstance(CLSID_DxcLibrary, &pLibrary));
   IFT(DllSupport.CreateInstance(CLSID_DxcAssembler, &pAssembler));
 
-  IFT(pLibrary->CreateBlobWithEncodingFromPinned(
-      (LPCVOID)Prior->StdOut.c_str(), Prior->StdOut.size(), CP_UTF8, &pSource));
+  // It would be better to properly honour %s for input, but for now, if prior
```
python3kgae, line 786: Why we need this?

lib/HLSL/DxilContainerReflection.cpp
```diff
@@ -2837,59 +2847,66 @@ HRESULT CFunctionReflection::GetDesc(D3D12_FUNCTION_DESC *pDesc) {
...
 
   pDesc->RequiredFeatureFlags =
       m_FeatureFlags & ~(UINT64)D3D_SHADER_REQUIRES_EARLY_DEPTH_STENCIL;
+  // Also Mask off function-level derivatives flag.
```
python3kgae, line 2895: Why mask off function-level flag?

pow2clk: It's a temporary flag that isn't meant to make it to the final feature flags as runtime has no need of it. It's used to determine derivative usage for the sake of cs/as/ms shaders. Once it has been used to determine what the entry points use, it can and should be discarded.

lib/DXIL/DxilShaderFlags.cpp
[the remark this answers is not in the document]
pow2clk: I'll incorporate what I mean in my release notes and ask that Tex look them over.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm65_ms_as.hlsl
```diff
@@ -0,0 +1,67 @@
...
+// works.  If that requirement were to be enforced, we would have to declare
+// these outputs.  However, if we do, there should also be a requirement that
+// mesh shader vertex output has SV_Position (not enforced currently either).
+// If that were to be enforced, and added to the struct, then the validator
+// will fail unless you write to all components of SV_Position.  This should
+// probably only be the case if OutputCounts are set to anything other than 0.
+// In any case, all this means that if some rules start to be enforced, we
+// will be forced to use things which will produce intrinsic calls in the mesh
+// shader which will cause the min target to be set to SM 6.5+ even for
+// validator version 1.7 and below.
```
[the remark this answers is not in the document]
pow2clk: I filed a bug: https://github.com/microsoft/DirectXShaderCompiler/issues/6267

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
tex3d: I'll give the response I gave @pow2clk [here](https://github.com/microsoft/DirectXShaderCompiler/pull/6207#discussion_r1483829104):

> It was necessary to input raw IR.  Normally, `%D3DReflect` follows a compilation (`%dxc`) or link (`%dxl`) where the resulting IR is dumped to the `Prior->StdOut`.  But if I want to start with IR, there is no prior stage dumping this IR to `Prior->StdOut`.  Ideally, I might do `%dxa` but that doesn't exist here, so I did this as a workaround for the problem.  The right fix would be to use the file if `%s` was specified, otherwise use the `Prior->StdOut`.  But to make that change, I'd have to change a whole lot more tests to remove the `%s` from `%D3DReflect %s` in their RUN lines.  I was thinking of doing that in a follow-up change.

python3kgae: You mean for the ll tests?
You can do it in lit with
// RUN: %dxa %s -o %t
// RUN: %dxa %t -dumpreflection | FileCheck %s

python3kgae: Or -dumpreflection does not dump what we need here?

python3kgae: maybe -dumprdat?

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
tex3d: Filed #6271 to track.

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
tex3d: Filed #6271 to track.

PR #6208, opened 2024-01-26
Gate local WaveMatrix runtime definitions on D3D12_EXPERIMENTAL_WAVE_MATRIX define

hekota:
The D3D12_EXPERIMENTAL_WAVE_MATRIX define will be removed in the future and replaced with WinSDK version that includes runtime support for WaveMatrix.

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -531,11 +531,9 @@ public:
                        L"Table:ShaderOpArithTable.xml#PackUnpackOpTable")
   END_TEST_METHOD()
 
-#ifdef D3D12_EXPERIMENTAL_WAVE_MATRIX
   TEST_METHOD(WaveMatrixLoadStoreTests);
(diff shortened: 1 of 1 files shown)


python3kgae (approved):

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -531,9 +531,11 @@ class ExecutionTest {
                        L"Table:ShaderOpArithTable.xml#PackUnpackOpTable")
   END_TEST_METHOD()
 
+#ifdef D3D12_EXPERIMENTAL_WAVE_MATRIX
```
tex3d, line 534: I thought the idea was just to replace the `#if !defined(D3D12_SDK_VERSION) || (D3D12_SDK_VERSION < 613)` with `#ifndef D3D12_EXPERIMENTAL_WAVE_MATRIX`, rather than disable the whole test based on the lack of this define.

tex3d (approved):
LGTM!

PR #6213, opened 2024-01-28
Added a ground truth path normalization.

adam-yang:
This change adds a ground truth path normalization helper:
- All slashes are changed to system native (`\` for windows, `/` for all others)
- All repeated slashes are removed (except for leading slashes, so windows UNC paths are not broken)
- All relative paths (including the main file, and ones that begin with `..`) are prepended with `./` or `.\` if not already

The path normalization is applied in the following places:
- Main file path in dxcompilerobj
- Paths passed into IDxcIncludeHandler
- Paths written into all DI* debug info
- Paths written into dx.content metadata and Pdb SourceInfo
- All paths loaded up by DxcPdbUtils

The reason for this change is to make it easier to tools authors to implement recompilation (example: PIX edit-and-continue). When the paths in all the above places match, the files can be matched with a normal string equal instead of having to path normalization.

### DIFF
--- a/include/dxc/Support/Path.h
+++ b/include/dxc/Support/Path.h
@@ -50,7 +50,7 @@ inline bool IsAbsoluteOrCurDirRelative(const char *Path) {
 }
 
 template <typename CharT, typename StringTy>
-inline StringTy NormalizePathImpl(const CharT *Path, size_t Length,
+StringTy NormalizePathImpl(const CharT *Path, size_t Length,
                                   bool PrefixWithDot) {
   StringTy PathCopy(Path, Length);
 
--- a/include/dxc/Support/Path.h
+++ b/include/dxc/Support/Path.h
@@ -1,8 +1,8 @@
 #pragma once
 
-#include <string>
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Path.h"
+#include <string>
 
 namespace hlsl {
 
--- a/include/dxc/Support/Path.h
+++ b/include/dxc/Support/Path.h
@@ -6,7 +6,7 @@
 namespace hlsl {
 
 template <typename CharTy>
-bool IsAbsoluteOrCurDirRelativeShared(const CharTy *Path) {
+bool IsAbsoluteOrCurDirRelativeImpl(const CharTy *Path) {
   if (!Path || !Path[0])
     return false;
   // Current dir-relative path.
@@ -43,15 +43,14 @@ bool IsAbsoluteOrCurDirRelativeShared(const CharTy *Path) {
 }
 
 inline bool IsAbsoluteOrCurDirRelativeW(const wchar_t *Path) {
-  return IsAbsoluteOrCurDirRelativeShared<wchar_t>(Path);
+  return IsAbsoluteOrCurDirRelativeImpl<wchar_t>(Path);
 }
 inline bool IsAbsoluteOrCurDirRelative(const char *Path) {
-  return IsAbsoluteOrCurDirRelativeShared<char>(Path);
+  return IsAbsoluteOrCurDirRelativeImpl<char>(Path);
 }
 
 template <typename CharT, typename StringTy>
-StringTy NormalizePathImpl(const CharT *Path, size_t Length,
-                                  bool PrefixWithDot) {
+StringTy NormalizePathImpl(const CharT *Path, size_t Length) {
   StringTy PathCopy(Path, Length);
 
 #ifdef _WIN32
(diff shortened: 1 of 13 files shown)


include/dxc/Support/Path.h
```diff
@@ -0,0 +1,90 @@
...
+    if (PathCopy[i] == '\\')
+      PathCopy[i] = '/';
+#endif
+  }
+  if (IsAbsoluteOrCurDirRelativeShared<CharT>(PathCopy.c_str())) {
+    return PathCopy;
+  }
+
+  if (PrefixWithDot)
+    return StringTy(1, CharT('.')) + StringTy(1, CharT('/')) + PathCopy;
```
python3kgae, line 70: '/' was replaced with '\\' in the loop for WIN32, why add '/' always here?

adam-yang: That is a mistake and looks like we don't have a test to exercise that path :/. That part is specifically for upgrading old PDBs with paths not prefixed with ".\". The reason the slash is not dependent on the platform is because when I wrote this part I hadn't decided that it's important to use backslash on windows.

I'll add a test for this.

adam-yang: The new test case I added actually managed to exercise this path. Also fixed this code.

include/dxc/Support/Path.h
```diff
@@ -0,0 +1,94 @@
+#pragma once
+
+#include "llvm/ADT/StringRef.h"
+#include <string>
+
+namespace hlsl {
+
+template <typename CharTy>
+bool IsAbsoluteOrCurDirRelativeShared(const CharTy *Path) {
```
damyanp, line 9: I spent a while trying to figure out what this is meant to do.  The "Shared" threw me quite a bit because I thought it was something to do with "CurDirRelativeShared", but I think the "Shared" just means that this same function is called by both the W and A versions?

In the absence of any unit tests, a comment giving some example inputs and the expected results would be helpful to understand and validate this.

Also, something somewhere explaining the overall strategy of path normalization would be helpful.  From the code we can see the somewhat arbitrary set of things that happen without really understanding why any of it is done.

include/dxc/Support/Path.h
```diff
@@ -0,0 +1,94 @@
...
+
+inline bool IsAbsoluteOrCurDirRelativeW(const wchar_t *Path) {
+  return IsAbsoluteOrCurDirRelativeShared<wchar_t>(Path);
+}
+inline bool IsAbsoluteOrCurDirRelative(const char *Path) {
+  return IsAbsoluteOrCurDirRelativeShared<char>(Path);
+}
+
+template <typename CharT, typename StringTy>
+inline StringTy NormalizePathImpl(const CharT *Path, size_t Length,
```
damyanp, line 53: Don't need the `inline` here.

tools/clang/tools/dxcompiler/dxcshadersourceinfo.cpp
```diff
@@ -313,19 +314,18 @@ static std::vector<SourceFile> ComputeFileList(clang::CodeGenOptions &cgOpts,
       if (it->first->isValid() && !it->second->IsSystemFile) {
         // If main file, write that to metadata first.
         // Add the rest to filesMap to sort by name.
-        llvm::SmallString<128> NormalizedPath;
-        llvm::sys::path::native(it->first->getName(), NormalizedPath);
         if (cgOpts.MainFileName.compare(it->first->getName()) == 0) {
           SourceFile file;
-          file.Name = NormalizedPath.str();
+          file.Name = hlsl::NormalizePath(it->first->getName(),
```
damyanp, line 319: I can't figure out from the code why one of these cases need the `.` prefix, but the other one doesn't.

include/dxc/Support/Path.h
```diff
@@ -0,0 +1,94 @@
...
+                                  bool PrefixWithDot) {
+  StringTy PathCopy(Path, Length);
+
+#ifdef _WIN32
```
damyanp, line 57: So this means that I'll get different looking paths if compiled on windows vs non-windows, even if all of my paths are relative.  Will this cause any problems with debuggers that only run on one platform?

damyanp (changes requested):

include/dxc/Support/Path.h
adam-yang: Removed.

tools/clang/tools/dxcompiler/dxcshadersourceinfo.cpp
adam-yang: I went ahead and got rid of the differences here. All paths regardless of main file or not is now normalized the same way. This makes things simpler in most cases.

include/dxc/Support/Path.h
adam-yang: That's right. The main reason I'm doing this is because of UNC paths on windows. If we normalize all paths to forward slash, UNC paths stop working. If we just leave UNC paths unchanged just for windows, we'll get different looking paths anyway.

I don't think debuggers are going to have any problems. All the paths they get will go through the same normalization process.

include/dxc/Support/Path.h
adam-yang: I changed the "Shared" to "Impl", just like the Normalize function below.

Also added explanation and examples in the Normalize function below.

tools/clang/lib/CodeGen/ModuleBuilder.cpp
```diff
@@ -277,15 +278,17 @@ namespace {
...
+              AddFile(hlsl::NormalizePath(it->first->getName()),
+                      it->second->getRawBuffer()->getBuffer());
               bFoundMainFile = true;
             } else {
-              filesMap[NormalizedPath.str()] =
-                  it->second->getRawBuffer()->getBuffer();
+              // We want the include file paths to match the values passed into
+              // the include handlers exactly. The SourceManager entries should
+              // match it except the call to MakeAbsoluteOrCurDirRelative.
+              std::string path = hlsl::NormalizePath(it->first->getName());
```
bob80905, line 290: Consider declaring this variable before the if condition, and then using it in both the if and the else clause.

include/dxc/Support/Path.h
```diff
@@ -0,0 +1,128 @@
...
+template <typename CharTy>
+bool IsAbsoluteOrCurDirRelativeImpl(const CharTy *Path) {
+  if (!Path || !Path[0])
+    return false;
+  // Current dir-relative path.
+  if (Path[0] == '.') {
+    return Path[1] == '\0' || Path[1] == '/' || Path[1] == '\\';
+  }
+  // Disk designator, then absolute path.
+  if (Path[1] == ':' && (Path[2] == '\\' || Path[2] == '/')) {
```
bob80905, line 27: In the case that the path passed is malformed, like "C", Path[1] would be an out of bounds access at this point. Would we want to check that Path[1] exists first? Likewise for Path[2]

tools/clang/tools/dxcompiler/dxcpdbutils.cpp
```diff
@@ -622,7 +615,9 @@ struct DxcPdbUtils : public IDxcPdbUtils2
                    hlsl::DxilMDHelper::kDxilSourceMainFileNameOldMDName) {
         MDTuple *tup = cast<MDTuple>(node.getOperand(0));
         MDString *str = cast<MDString>(tup->getOperand(0));
-        IFR(Utf8ToBlobWide(str->getString(), &m_MainFileName));
+        std::string normalized = hlsl::NormalizePath(str->getString());
+        m_MainFileName = nullptr;
```
bob80905, line 619: You added this nullptr assignment. Should there be a check after Utf8ToblobWide to see if m_MainFileName is still nullptr?

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -1939,7 +1959,7 @@ TEST_F(CompilerTest, CompileThenTestPdbUtilsWarningOpt) {
     CComBSTR pMainFileName;
     VERIFY_SUCCEEDED(pPdbUtils->GetMainFileName(&pMainFileName));
     std::wstring mainFileName = static_cast<const wchar_t *>(pMainFileName);
-    VERIFY_ARE_EQUAL(mainFileName, L"hlsl.hlsl");
+    VERIFY_ARE_EQUAL(mainFileName, L"." SLASH_W L"hlsl.hlsl");
```
bob80905, line 1962: Shouldn't there be a + here?

include/dxc/Support/Path.h
adam-yang: Since "C" is null-terminated, Path[1] would be '\0' and the subsequent Path[2] would not be checked. Although this is pretty confusing and unsafe... So I should probably change it.

tools/clang/lib/CodeGen/ModuleBuilder.cpp
adam-yang: Done.

tools/clang/tools/dxcompiler/dxcpdbutils.cpp
adam-yang: Setting it to nullptr is just to release the previous value if it was assigned. It's not a danger that `Utf8ToBlobWide` leaves it as null. Added a comment here to clarify.

tools/clang/unittests/HLSL/CompilerTest.cpp
adam-yang: String literals can be concatenated without putting +.

include/dxc/Support/Path.h
adam-yang: I made it take a length and check the length instead.

bob80905 (approved):

PR #6219, opened 2024-01-30
WinPIX: Changes to support bitfields in PIX's shader debugger

jeffnn:
Two related sets of changes:
1) Modify the value-to-declare pass to tolerate bitfields, which are detected by their habit of being smaller than the type in which they are contained. The pass produces dbg.declare statements that map each HLSL variable to some storage. With bitfields, that mapping becomes many-to-one, because many bitfields can be stored in the same basic type. Most of the changes are to preserve the size of the bitfield as the code descends the type hierarchy into the underlying basic type, which is what motivates these new "type overrride" parameters to some functions. Complications to watch out for are min16 types, which may or may not be their own purpose-made (16-bit) types depending on whether or not they are enabled.
2) Modify the PIX debug-info API to return enough data to PIX to allow it to extract (by bit shifting) a bitfield from the underlying type. This bit is also about preserving knowledge of the field's size as opposed to its containing type's size, hence the new dxcapi interface/method.


include/dxc/dxcpix.h
```diff
@@ -47,12 +47,23 @@ struct __declspec(uuid("9ba0d9d3-457b-426f-8019-9f3849982aa2")) IDxcPixArrayType
 };
 
 struct __declspec(uuid("6c707d08-7995-4a84-bae5-e6d8291f3b78"))
+    IDxcPixStructField0 : public IUnknown {
```
adam-yang, line 50: This gave me a pause, but I see what's happening here now. Might be worth adding a comment.

Also why doesn't `IDxcPixStructField` just subtype from `IDxcPixStructField0`?

jeffnn: I didn't inherit from the old because JP's too-clever-for-me wrapper stuff can't handle it, and after half a day trying to make it do so, I gave up. One of these days I should just delete all that in favor of explicit error checking. It's hard to maintain and such a pain to step through.

adam-yang (approved):

austinkinross (approved):

PR #6227, opened 2024-01-31
Add type annotations for missing HLSL object cases

tex3d:
- Node Object method return types
- Type template arguments

TODO: Add tests

### DIFF
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -931,9 +931,12 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(
             annotation->GetTemplateArgAnnotation(i);
         const clang::TemplateArgument &arg = args[i];
         switch (arg.getKind()) {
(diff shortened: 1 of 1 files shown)


tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -931,9 +931,12 @@ unsigned CGMSHLSLRuntime::ConstructStructAnnotation(
             annotation->GetTemplateArgAnnotation(i);
         const clang::TemplateArgument &arg = args[i];
         switch (arg.getKind()) {
-        case clang::TemplateArgument::ArgKind::Type:
-          argAnnotation.SetType(CGM.getTypes().ConvertType(arg.getAsType()));
-          break;
+        case clang::TemplateArgument::ArgKind::Type: {
+          auto argTy = arg.getAsType();
+          unsigned arrayEltSize = 0;
+          AddTypeAnnotation(argTy, dxilTypeSys, arrayEltSize);
```
anupamachandra, line 937: I'm curious why this hasn't come up before for other templated types.

PR #6232, opened 2024-01-31
[SPIR-V] Mark array reference to aliased var as unaliased

### LINKED ISSUE #6110: [SPIR-V] Crash when accessing buffer from array of buffers
**Description**
The SPIR-V optimizer crashes when accessing a buffer from an array of buffers which has been assigned to a variable.

**Steps to Reproduce**
Minimal example derived from a real world shader that suffers from the crash:
```hlsl
ByteAddressBuffer inputBuffers[16] : register(t0);

struct MyStruct
{
	ByteAddressBuffer myBuffers[16];
};

[numthreads(32, 1, 1)]
void main()
{
	MyStruct ms;
	ms.myBuffers = inputBuffers;
	ByteAddressBuffer buffer = ms.myBuffers[0]; // < This line causes crash
}
```

I managed to strip the example even more to just these two lines, but the location of the crash is slightly different so I'll include both examples for completeness:
```hlsl
[numthreads(32, 1, 1)]
void main()
{
	ByteAddressBuffer buffers[16];
	ByteAddressBuffer buffer = buffers[0]; // < This line causes crash
}
```

Everything is compiled with command line options: `-E "main" -T cs_6_0 -spirv`

**Actual Behavior**
Both examples causes a crash in SPIRV-Tools `Pass::GenerateCopy()`
The first example produces an assert `"Don't know how to copy this type.  Code is likely illegal."`
The second example attempts to read nullptr.

**Environment**
- DXC version `dxcompiler.dll: 1.8 - 1.7.0.14383 (main, ad3958c2d)`
- Host Operating System Windows 10 64-bit

s-perron:
An "aliased" variable is on where the compiler has implicitly added an
extra level indirection. So when we access one of these variable, we
have to do an extra load.

When we do an array access on an aliased variable, we do the extra load
before doing the access chain, but we still treat the result of the
access chain as if it is an aliased value. This causes an extra load and
generates invalid spir-v.

The fix is to change `isReferencingNonAliasStructuredOrByteBuffer` so
that we do not distinguish between externally and internally visible
variable if we see an array subscript.

Fixes #6110

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
@@ -26,7 +26,7 @@ RWByteAddressBuffer BufferOut;
 void main() {
 // CHECK: %LocalArray = OpVariable %_ptr_Function__ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Function
 // CHECK: %Local = OpVariable %_ptr_Function__ptr_Uniform_type_ByteAddressBuffer Function
-  ByteAddressBuffer LocalArary[2];
+  ByteAddressBuffer LocalArray[2];
 
 // CHECK: OpStore %LocalArray %BufferArray
   LocalArray = BufferArray;
--- a/tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
@@ -14,25 +14,25 @@
 // CHECK: %type_RWByteAddressBuffer = OpTypeStruct %_runtimearr_uint
 // CHECK: %_ptr_Uniform_type_RWByteAddressBuffer = OpTypePointer Uniform %type_RWByteAddressBuffer
 // CHECK: %Buffer0 = OpVariable %_ptr_Uniform_type_ByteAddressBuffer Uniform
-// CHECK: %BufferArary = OpVariable %_ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Uniform
+// CHECK: %BufferArray = OpVariable %_ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Uniform
 // CHECK: %BufferOut = OpVariable %_ptr_Uniform_type_RWByteAddressBuffer Uniform
 
 ByteAddressBuffer Buffer0;
-ByteAddressBuffer BufferArary[2];
+ByteAddressBuffer BufferArray[2];
 RWByteAddressBuffer BufferOut;
 
 // CHECK: %src_main = OpFunction
 [numthreads(1, 1, 1)]
 void main() {
-// CHECK: %LocalArary = OpVariable %_ptr_Function__ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Function
+// CHECK: %LocalArray = OpVariable %_ptr_Function__ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Function
 // CHECK: %Local = OpVariable %_ptr_Function__ptr_Uniform_type_ByteAddressBuffer Function
   ByteAddressBuffer LocalArary[2];
 
-// CHECK: OpStore %LocalArary %BufferArary
-  LocalArary = BufferArary;
+// CHECK: OpStore %LocalArray %BufferArray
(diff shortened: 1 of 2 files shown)


tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
```diff
@@ -14,10 +14,25 @@
 // CHECK: %type_RWByteAddressBuffer = OpTypeStruct %_runtimearr_uint
 // CHECK: %_ptr_Uniform_type_RWByteAddressBuffer = OpTypePointer Uniform %type_RWByteAddressBuffer
 // CHECK: %Buffer0 = OpVariable %_ptr_Uniform_type_ByteAddressBuffer Uniform
+// CHECK: %BufferArary = OpVariable %_ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Uniform
```
Keenuts, line 17:
```suggestion
// CHECK: %BufferArray = OpVariable %_ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Uniform
```

tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
```diff
@@ -14,10 +14,25 @@
...
 // CHECK: %BufferOut = OpVariable %_ptr_Uniform_type_RWByteAddressBuffer Uniform
 
 ByteAddressBuffer Buffer0;
+ByteAddressBuffer BufferArary[2];
```
Keenuts, line 21:
```suggestion
ByteAddressBuffer BufferArray[2];
```

tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
```diff
@@ -14,10 +14,25 @@
...
 RWByteAddressBuffer BufferOut;
 
+// CHECK: %src_main = OpFunction
 [numthreads(1, 1, 1)]
-void main() {}
+void main() {
+// CHECK: %LocalArary = OpVariable %_ptr_Function__ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Function
```
Keenuts, line 27:
```suggestion
// CHECK: %LocalArray = OpVariable %_ptr_Function__ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 Function
```

tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
```diff
@@ -14,10 +14,25 @@
...
+// CHECK: %Local = OpVariable %_ptr_Function__ptr_Uniform_type_ByteAddressBuffer Function
+  ByteAddressBuffer LocalArary[2];
+
+// CHECK: OpStore %LocalArary %BufferArary
+  LocalArary = BufferArary;
```
Keenuts, line 32:
```suggestion
// CHECK: OpStore %LocalArray %BufferArray
  LocalArray = BufferArray;
```

tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
```diff
@@ -14,10 +14,25 @@
...
+
+// CHECK: [[array:%[0-9]+]] = OpLoad %_ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 %LocalArary
```
Keenuts, line 34:
```suggestion
// CHECK: [[array:%[0-9]+]] = OpLoad %_ptr_Uniform__arr_type_ByteAddressBuffer_uint_2 %LocalArray
```

tools/clang/test/CodeGenSPIRV/type.byte-address-buffer.hlsl
```diff
@@ -14,10 +14,25 @@
...
+// CHECK: [[ac:%[0-9]+]] = OpAccessChain %_ptr_Uniform_type_ByteAddressBuffer [[array]] %int_0
+// CHECK: OpStore %Local [[ac]]
+  ByteAddressBuffer Local = LocalArary[0];
```
Keenuts, line 37:
```suggestion
  ByteAddressBuffer Local = LocalArray[0];
```

Keenuts (approved):

PR #6243, opened 2024-02-05
Warn on ignored attributes on library non-entry points

### LINKED ISSUE #6100: Add warning for entry-only attribute without the shader attribute when compiling to library target

When Compiling to a library target, such as `lib_6_7`, DXC used to infer that a function was an entry point of a specific kind if it used an attribute that was only meaningful on an entry point of a particular shader kind.  This was broken for at least compute vs. amplification vs. mesh, since it assumed that the `[numthreads(...)]` attribute implied compute shader, even though it is now used with other shader kinds as well.

This shader kind inference was not meant to be used to indicate entry points in libraries, since the HLSLShaderAttribute, or `[shader("<kind>")]` in HLSL, was introduced for this purpose.  This inference was originally meant for determining whether you had attributes indicating a conflicting shader kind when compiling an entry point to a certain shader target.

These facts led us to remove this shader entry point inference for libraries, since it can cause unintended entry points to be included in your library, and it can make an incorrect assumption in ambiguous cases (ex. numthreads).

For shader kinds that are supported in libraries by the D3D runtime (DXR entry points, and node shaders in the future), you have always been required to use the shader attribute (`[shader("<kind>")]`) to identify a function as an entry point.  So that change did not cause a problem for library compilations used for DXR.

However, there may be users of offline linking that relied on this entry point inference in intermediate libraries used for linking to different final shader targets.  In these scenarios that used to rely on this inference, the entry point will now be seen as an ordinary internal function, which will be eliminated as unused in the output when compiling to a library target.  This would cause subsequent linking to fail.  The fix is to add the `[entry("kind")]` attribute to the function, but the root cause and this fix may not be obvious when upgrading to the latest compiler.

Add a warning when compiling to a library target (`lib_6_*`), when a function lacks the shader attribute (`[shader("<kind>")]`), and includes any other attributes which are only meaningful when applied to an entry point.

Example attributes used to indicate an entry point function: `numthreads`, `rootsignature`, `clipplanes`, `outputcontrolpoints`, and many more.

Some of these attributes are only used with a specific shader kind (like `clipplanes` indicating "vertex" shader), or a limited subset of shader kinds (like `numthreads` indicating "compute", "amplification", "mesh", or "node").  It might be useful to identify the legal kinds usable with the attribute, or the intersecting set from such attributes found, and use this in a suggestion to add the `[shader("<kind>")]` attribute with the identified shader kinds that are compatible with the other attributes found.

It's important that this warning is assigned a group to allow it to be turned off, so that intentional cases that cause the warning to fire can be muted when treating warnings as errors (-WX).

- Rely only on a release note about the change in entry point inference for library compilations.
  - People often don't read or remember all the release notes or may not really comprehend their full impact when reading them.  Also, the people who run into the problem may not have been the ones upgrading the compiler and reading those notes.  It's best to emit an explicit warning with a recommended fix when we detect a case where we may not be interpreting things the way the developer intended.
- Make this an error instead of a warning
  - This is problematic because it's legal to have shader entry points in HLSL that you don't intend to use in a library compilation, or for these attributes to be used and ignored on non-entry functions for a given compilation.  For instance, you might call the function from an entry point in your library, and for another compilation you use the same function as the direct entry point.
  - Options can be used to turn all warnings, or just this warning, into an error as desired, so this seems unnecessary.

pow2clk:
For every attribute that is only meaningful on an entry-point, produce a warning indicating that this could be a problem.
The attributes are ignored, so it's harmless, but it still might indicate a mistake on the part of the user. 
Previously, some of these would cause the function to be implicitly made a certain kind of entry point.
These warnings will soften the impact of that change. 

Fixes #6100

### DIFF
--- a/lib/HLSL/HLSignatureLower.cpp
+++ b/lib/HLSL/HLSignatureLower.cpp
@@ -479,7 +479,7 @@ void HLSignatureLower::CreateDxilSignatures() {
                     isPatchConstantFunctionFalse, bForOutFasle, bHasClipPlane);
   }
 
-  if (bHasClipPlane) {
+  if (bHasClipPlane && !pSM->IsLib()) {
     dxilutil::EmitErrorOnFunction(HLM.GetModule()->getContext(), Entry,
                                   "Cannot use clipplanes attribute without "
                                   "specifying a 4-component SV_Position "
--- a/lib/HLSL/HLSignatureLower.cpp
+++ b/lib/HLSL/HLSignatureLower.cpp
@@ -479,7 +479,7 @@ void HLSignatureLower::CreateDxilSignatures() {
                     isPatchConstantFunctionFalse, bForOutFasle, bHasClipPlane);
   }
 
-  if (bHasClipPlane && !pSM->IsLib()) {
+  if (bHasClipPlane) {
     dxilutil::EmitErrorOnFunction(HLM.GetModule()->getContext(), Entry,
                                   "Cannot use clipplanes attribute without "
                                   "specifying a 4-component SV_Position "
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -15678,38 +15678,37 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
 // (where the compiler would infer a specific shader kind based on
 // a present entry-point-only attribute).
 void WarnOnEntryAttrWithoutShaderAttr(Sema &S, FunctionDecl *FD) {
-  if (FD->hasAttrs()) {
-    for (Attr *A : FD->getAttrs()) {
-      switch (A->getKind()) {
-        // Entry-Function-only attributes
-      case clang::attr::HLSLClipPlanes:
-      case clang::attr::HLSLDomain:
-      case clang::attr::HLSLEarlyDepthStencil:
-      case clang::attr::HLSLInstance:
-      case clang::attr::HLSLMaxTessFactor:
-      case clang::attr::HLSLNumThreads:
-      case clang::attr::HLSLRootSignature:
-      case clang::attr::HLSLOutputControlPoints:
-      case clang::attr::HLSLOutputTopology:
-      case clang::attr::HLSLPartitioning:
-      case clang::attr::HLSLPatchConstantFunc:
-      case clang::attr::HLSLMaxVertexCount:
-      case clang::attr::HLSLWaveSize:
-      case clang::attr::HLSLNodeLaunch:
-      case clang::attr::HLSLNodeIsProgramEntry:
-      case clang::attr::HLSLNodeId:
-      case clang::attr::HLSLNodeLocalRootArgumentsTableIndex:
-      case clang::attr::HLSLNodeShareInputOf:
-      case clang::attr::HLSLNodeDispatchGrid:
-      case clang::attr::HLSLNodeMaxDispatchGrid:
-      case clang::attr::HLSLNodeMaxRecursionDepth:
-        S.Diag(A->getLocation(),
-               diag::warn_hlsl_entry_attribute_without_shader_attribute)
-            << A->getSpelling();
-        break;
-      }
-    }
+  if (!FD->hasAttrs())
     return;
+  for (Attr *A : FD->getAttrs()) {
+    switch (A->getKind()) {
+      // Entry-Function-only attributes
+    case clang::attr::HLSLClipPlanes:
+    case clang::attr::HLSLDomain:
+    case clang::attr::HLSLEarlyDepthStencil:
+    case clang::attr::HLSLInstance:
+    case clang::attr::HLSLMaxTessFactor:
+    case clang::attr::HLSLNumThreads:
+    case clang::attr::HLSLRootSignature:
+    case clang::attr::HLSLOutputControlPoints:
+    case clang::attr::HLSLOutputTopology:
+    case clang::attr::HLSLPartitioning:
+    case clang::attr::HLSLPatchConstantFunc:
+    case clang::attr::HLSLMaxVertexCount:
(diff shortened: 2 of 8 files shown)


lib/HLSL/HLSignatureLower.cpp
```diff
@@ -479,7 +479,7 @@ void HLSignatureLower::CreateDxilSignatures() {
                     isPatchConstantFunctionFalse, bForOutFasle, bHasClipPlane);
   }
 
-  if (bHasClipPlane) {
+  if (bHasClipPlane && !pSM->IsLib()) {
```
tex3d, line 482: What's the reason for this?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15672,6 +15672,49 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
...
+      case clang::attr::HLSLEarlyDepthStencil:
+      case clang::attr::HLSLInstance:
+      case clang::attr::HLSLMaxTessFactor:
+      case clang::attr::HLSLNumThreads:
+      case clang::attr::HLSLRootSignature:
+      case clang::attr::HLSLOutputControlPoints:
+      case clang::attr::HLSLOutputTopology:
+      case clang::attr::HLSLPartitioning:
+      case clang::attr::HLSLPatchConstantFunc:
+      case clang::attr::HLSLExperimental:
```
tex3d, line 15696: [experimental] attribute is used to develop experimental features.  It is not an entry point-specific attribute.

tools/clang/test/SemaHLSL/attributes.hlsl
```diff
@@ -350,11 +350,11 @@ int uav() {
   return i;
 }
 
-[domain] int domain_fn_missing() { return 1; }          // expected-error {{'domain' attribute takes one argument}} fxc-pass {{}}
+[domain] [shader("pixel")]int domain_fn_missing() { return 1; }          // expected-error {{'domain' attribute takes one argument}} fxc-pass {{}}
```
tex3d, line 353: Why `[shader("pixel")]`?  That would be incorrect for this shader stage.  I would expect we would want to emit some error in the future for a case like this.  On top of that, with the addition of shader attributes and a lib target, we might do additional attribute diagnostics that these sorts of tests are not prepared for.  It would be more work to make these error-free if we were doing the diagnostics in Sema that we would want to do.

I think a better option is to just add the expected-warning message to these, rather than add shader attributes.

If the issue is the addition of the lib target run line and this test behaves differently in that case, then we should split the test into lib/non-lib files.

tools/clang/test/SemaHLSL/attributes.hlsl
```diff
@@ -855,66 +867,77 @@ bool Test_Loop() {
 // Test ClipPlanes
 float4 ClipPlanesVal;
 [ClipPlanes(ClipPlanesVal)]
+[shader("pixel")]
```
tex3d, line 870: I bet this case is why you added the ` && !pSM->IsLib()` to the ClipPlanes check.  ClipPlanes is a vertex attribute, with the additional requirement for the `SV_Position` output.  I think that's the wrong fix.

bob80905: Yeah, the diagnostic was being emitted on certain tests and I didn't understand why, so I just added the attribute. Turns out removing the condition didn't cause the problem to resurface, so I can just resolve this.

tools/clang/test/SemaHLSL/attributes.hlsl
bob80905: I've added warnings instead of shader attributes.

pow2clk (commented):
I approve, but I can't actually click the button since it still looks like my change to the system 🤷

bob80905 (approved):

bob80905:
> I approve, but I can't actually click the button since it still looks like my change to the system 🤷

I can approve in your place then!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15672,6 +15672,48 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
   return;
 }
 
+// The compiler should emit a warning when an entry-point-only attribute
+// is detected without the presence of a shader attribute,
+// to prevent reliance on deprecated behavior
+// (where the compiler would infer a specific shader kind based on
+// a present entry-point-only attribute).
+void WarnOnEntryAttrWithoutShaderAttr(Sema &S, FunctionDecl *FD) {
+  if (FD->hasAttrs()) {
```
llvm-beanz, line 15681: nit: we could early exit here to reduce nesting.

See: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15672,6 +15672,48 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
...
+      case clang::attr::HLSLNodeDispatchGrid:
+      case clang::attr::HLSLNodeMaxDispatchGrid:
+      case clang::attr::HLSLNodeMaxRecursionDepth:
+        S.Diag(A->getLocation(),
+               diag::warn_hlsl_entry_attribute_without_shader_attribute)
+            << A->getSpelling();
+        break;
+      }
+    }
+    return;
```
llvm-beanz, line 15712: nit: this looks like an unnecessary return since control flow only directly flows from here to the end of the function and the function is `void(...)` so `return` shouldn't be necessary.

llvm-beanz (approved):
two nits, otherwise looks good.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: I'm happy to learn that this preference is documented!

PR #6244, opened 2024-02-05
[DIAG] Limit derivative warning to intrinsic overload added in sm6.8

### LINKED ISSUE #6235: SM 6.6 ExecutionTest::ComputeSampleTest test is failing
**Description**
Regression in  ExecutionTest::ComputeSampleTest that happened somewhere between DXC commit f94fe829 (good) and 9196dfb0 (bad).

**Steps to Reproduce**
`hcttest exec-filter ComputeSampleTest -adapter "NVidia*"`

**Actual Behavior**
```
StartGroup: ExecutionTest::ComputeSampleTest
Using Adapter:NVIDIA GeForce GTX 1660 Ti
Failed to compile shader:  CS2:23:24: error: Intrinsic CalculateLevelOfDetail potentially used by CSMain2D requires derivatives - when used in compute/amplification/mesh shaders or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or both X and Y must be multiples of 2. [-Whlsl-availability]
          return uint4(g_tex.CalculateLevelOfDetail(g_samp, float2(left, 0.5)) * (~ix&1) +
                       ^
CS2:132:14: note: declared here
        void CSMain2D(uint3 id : SV_GroupThreadID) {
             ^
CS2:24:24: error: Intrinsic CalculateLevelOfDetail potentially used by CSMain2D requires derivatives - when used in compute/amplification/mesh shaders or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or both X and Y must be multiples of 2. [-Whlsl-availability]
                       g_tex.CalculateLevelOfDetail(g_samp, float2(right, 0.5)) * (ix&1),
                       ^
CS2:132:14: note: declared here
        void CSMain2D(uint3 id : SV_GroupThreadID) {
             ^
CS2:27:24: error: Intrinsic CalculateLevelOfDetail potentially used by CSMain2D requires derivatives - when used in compute/amplification/mesh shaders or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or both X and Y must be multiples of 2. [-Whlsl-availability]
                       g_tex.CalculateLevelOfDetail(g_samp, float2(0.5, top)) * (~iy&1) +
                       ^
CS2:132:14: note: declared here
        void CSMain2D(uint3 id : SV_GroupThreadID) {
             ^
CS2:28:24: error: Intrinsic CalculateLevelOfDetail potentially used by CSMain2D requires derivatives - when used in compute/amplification/mesh shaders or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or both X and Y must be multiples of 2. [-Whlsl-availability]
                       g_tex.CalculateLevelOfDetail(g_samp, float2(0.5, bot)) * (iy&1),
                       ^
CS2:132:14: note: declared here
        void CSMain2D(uint3 id : SV_GroupThreadID) {
             ^


Error: Verify: SUCCEEDED(resultCode) - Value (0x80004005) [File: D:\dxc\tools\clang\unittests\HLSLExec\ShaderOpTest.cpp, Function: st::ShaderOpTest::CreateShaders, Line: 876]
EndGroup: ExecutionTest::ComputeSampleTest [Failed]

Summary of Non-passing Tests:
    ExecutionTest::ComputeSampleTest [Failed]
```

**Environment**
- The regression happened somewhere between DXC commit f94fe829 (good) and 9196dfb0 (bad).
- Windows 11 x64

### LINKED ISSUE #6249: Sm 6.8 derivative-related DefaultError warnings fire on existing pre-6.8 shaders
**Description**
We decided to introduce a DefaultError warning for potential uses of operations that require derivatives in unsupported shaders due to stage or numthreads characteristics. However, this got emitted for previous shader models which would break existing shaders.

**Steps to Reproduce**
https://godbolt.org/z/r9MbeKG4Y

Just by using a derivative-dependent op in a vs shader should trigger this. Even without the early diagnostic, validation should fail, but it's a simple way to check the issue without having to devise an optimization-dependent way to let validation pass.

**Actual Behavior**
clang diagnostics are emitted such as: 
"Intrinsic CalculateLevelOfDetail potentially used by vs requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders"

**Environment**
- DXC version dxcompiler.dll: 1.8 - 1.7.0.4313
- Host Operating System win11

python3kgae:
This change is made to prevent errors in unused code that might impact existing shaders.

Fixes #6249
Fixes #6235

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7787,7 +7787,7 @@ def warn_hlsl_derivatives_wrong_numthreads : Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - when used in compute/amplification/mesh shaders"
     " or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or"
     " both X and Y must be multiples of 2.">,
-    InGroup<HLSLAvailability>;
+    DefaultError, InGroup<HLSLAvailability>;
 def warn_hlsl_intrinsic_overload_in_wrong_shader_model : Warning<
    "overload of intrinsic %0 requires shader model %1 or greater">, 
     DefaultError, InGroup<HLSLAvailability>;
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7779,10 +7779,10 @@ def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
 def warn_hlsl_derivatives_in_wrong_shader_kind: Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders.">,
-    InGroup<HLSLAvailability>;
+    DefaultError, InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_in_wrong_shader_model: Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - shader model 6.6 or greater is required for use in compute, amplification, or mesh shaders.">,
-    InGroup<HLSLAvailability>;
+    DefaultError, InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_wrong_numthreads : Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - when used in compute/amplification/mesh shaders"
     " or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or"
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7779,15 +7779,15 @@ def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
 def warn_hlsl_derivatives_in_wrong_shader_kind: Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders.">,
-    DefaultError, InGroup<HLSLAvailability>;
+    InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_in_wrong_shader_model: Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - shader model 6.6 or greater is required for use in compute, amplification, or mesh shaders.">,
-    DefaultError, InGroup<HLSLAvailability>;
+    InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_wrong_numthreads : Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - when used in compute/amplification/mesh shaders"
     " or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or"
     " both X and Y must be multiples of 2.">,
-    DefaultError, InGroup<HLSLAvailability>;
+    InGroup<HLSLAvailability>;
 def warn_hlsl_intrinsic_overload_in_wrong_shader_model : Warning<
    "overload of intrinsic %0 requires shader model %1 or greater">, 
     DefaultError, InGroup<HLSLAvailability>;
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7778,13 +7778,13 @@ def err_hlsl_unsupported_for_version_lower : Error<
 def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
 def warn_hlsl_derivatives_in_wrong_shader_kind: Warning<
-   "Intrinsic %0 potentially used by %1 requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders.">,
+   "Intrinsic %0 potentially used by '%1' requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders.">,
     DefaultError, InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_in_wrong_shader_model: Warning<
-   "Intrinsic %0 potentially used by %1 requires derivatives - shader model 6.6 or greater is required for use in compute, amplification, or mesh shaders.">,
+   "Intrinsic %0 potentially used by '%1' requires derivatives - shader model 6.6 or greater is required for use in compute, amplification, or mesh shaders.">,
     DefaultError, InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_wrong_numthreads : Warning<
-   "Intrinsic %0 potentially used by %1 requires derivatives - when used in compute/amplification/mesh shaders"
+   "Intrinsic %0 potentially used by '%1' requires derivatives - when used in compute/amplification/mesh shaders"
     " or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or"
     " both X and Y must be multiples of 2.">,
     DefaultError, InGroup<HLSLAvailability>;
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -11156,65 +11156,68 @@ void Sema::DiagnoseReachableHLSLMethodCall(const CXXMethodDecl *MD,
     switch (opCode) {
     case hlsl::IntrinsicOp::MOP_CalculateLevelOfDetail:
     case hlsl::IntrinsicOp::MOP_CalculateLevelOfDetailUnclamped: {
-      if (!SM->IsSM68Plus()) {
-        QualType SamplerComparisonTy =
-            HLSLExternalSource::FromSema(this)->GetBasicKindType(
-                AR_OBJECT_SAMPLERCOMPARISON);
-        if (MD->getParamDecl(0)->getType() == SamplerComparisonTy) {
+      QualType SamplerComparisonTy =
+          HLSLExternalSource::FromSema(this)->GetBasicKindType(
+              AR_OBJECT_SAMPLERCOMPARISON);
+      if (MD->getParamDecl(0)->getType() == SamplerComparisonTy) {
+
+        if (!SM->IsSM68Plus()) {
+
           Diags.Report(Loc,
                        diag::warn_hlsl_intrinsic_overload_in_wrong_shader_model)
               << MD->getNameAsString() + " with SamplerComparisonState"
               << "6.8";
-        }
-      } else {
+        } else {
 
-        switch (EntrySK) {
-        default: {
-          if (!SM->AllowDerivatives(EntrySK)) {
-            Diags.Report(Loc, diag::warn_hlsl_derivatives_in_wrong_shader_kind)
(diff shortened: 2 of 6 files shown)


tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7779,15 +7779,15 @@ def err_hlsl_unsupported_keyword_for_min_precision : Error<
...
+    InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_wrong_numthreads : Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - when used in compute/amplification/mesh shaders"
     " or broadcast nodes, numthreads must be either 1D with X as a multiple of 4 or"
     " both X and Y must be multiples of 2.">,
-    DefaultError, InGroup<HLSLAvailability>;
+    InGroup<HLSLAvailability>;
 def warn_hlsl_intrinsic_overload_in_wrong_shader_model : Warning<
    "overload of intrinsic %0 requires shader model %1 or greater">, 
     DefaultError, InGroup<HLSLAvailability>;
```
llvm-beanz, line 7793: Do we also need to change this to a warning?

python3kgae: Nice catch.
I'll keep it as error.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7779,10 +7779,10 @@ def err_hlsl_unsupported_keyword_for_min_precision : Error<
    "%0 is only supported with -enable-16bit-types option">;
 def warn_hlsl_derivatives_in_wrong_shader_kind: Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders.">,
-    DefaultError, InGroup<HLSLAvailability>;
+    InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_in_wrong_shader_model: Warning<
    "Intrinsic %0 potentially used by %1 requires derivatives - shader model 6.6 or greater is required for use in compute, amplification, or mesh shaders.">,
-    DefaultError, InGroup<HLSLAvailability>;
+    InGroup<HLSLAvailability>;
 def warn_hlsl_derivatives_wrong_numthreads : Warning<
```
tex3d, line 7786: This one and the next are in the same group and are still `DefaultError`.  This one could still impact existing pre-SM 6.8 usage.

The next one is `warn_hlsl_intrinsic_overload_in_wrong_shader_model` which is the only one that (for now) seems to be limited to diagnosing the SM 6.8 overload, and is the case where I think you wanted DefaultError.  But the warning itself is generic enough that you could certainly see it being applied to other overload cases that have different shader model requirements.

It feels weird to have warnings in the same group with some DefaultError and others not, when the granularity you have to change the default is per-group.

What's the consistent policy here?  We haven't applied this to SM 6.8 ops and overloads generally, nor are we limiting it to when you compile to SM 6.8, nor is it triggered by a language version update.  It's not easy to explain the policy to users now.  I'm not even sure what it's supposed to be now.

Is it this?
> Some new SM 6.8 operation overloads may be diagnosed early.  This may cause a compilation failure even if the operation would not have been used in the final shader after optimizations and dead code elimination.
_Followed by mitigation instructions..._

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
tex3d: Why was the `warn_hlsl_derivatives_wrong_numthreads` one undone?  This could impact existing usage on pre-SM 6.8.

I think @llvm-beanz question might have been misinterpreted.  Were you asking if `warn_hlsl_intrinsic_overload_in_wrong_shader_model` should also be changed to warning?  If so, we might need another group and warning to support the `DefaultError` case covering just the new SM 6.8 overload.  In fact, we should be doing all the checks for the new overload as well, right?  So it seems we need two groups of these, one that `DefaultError` and are only used for SM 6.8 new intinsics/overloads (only used when comparison sampler is used) and the other group used for pre-existing things.  Either that, or instead of removing `DefaultError`, we only diagnose them for the new SM 6.8 overload of the intrinsic, and do not use them for existing overloads.

tools/clang/test/SemaHLSL/hlsl/objects/texture/CalculateLODWithSamplerComparisonState.hlsl
```diff
@@ -35,12 +35,3 @@ float ps(float a:A) : SV_Target {
...
 
-// expected-note@+3{{declared here}}
-// expected-note@+2{{declared here}}
-[shader("vertex")]
-float4 vs(float a:A) : SV_Position {
-  // expected-error@+1 {{Intrinsic CalculateLevelOfDetail potentially used by vs requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders}}
-  return t.CalculateLevelOfDetail(s, a) + // expected-error {{overload of intrinsic CalculateLevelOfDetail with SamplerComparisonState requires shader model 6.8 or greater}}
-  // expected-error@+1 {{Intrinsic CalculateLevelOfDetailUnclamped potentially used by vs requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders}}
-    t.CalculateLevelOfDetailUnclamped(s, a); // expected-error {{overload of intrinsic CalculateLevelOfDetailUnclamped with SamplerComparisonState requires shader model 6.8 or greater}}
-}
```
pow2clk, line 46: The bug we are fixing is that this code produced defaultError warnings before on previous shader models. We don't have a test here that verifies that they don't anymore. Something like this, but using a non-comparison sampler that just verifies there are no diagnostics should be sufficient.

pow2clk:
I belatedly created an issue for this #6249. I'll add it to the description.

tools/clang/test/SemaHLSL/hlsl/objects/texture/CalculateLODWithSamplerComparisonState.hlsl
python3kgae: Added.

python3kgae:
> I belatedly created an issue for this #6249. I'll add it to the description.

Thanks a lot.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11166,52 +11166,55 @@ void Sema::DiagnoseReachableHLSLMethodCall(const CXXMethodDecl *MD,
               << MD->getNameAsString() + " with SamplerComparisonState"
               << "6.8";
         }
-      }
+      } else {
```
tex3d, line 11169: I didn't think this is what we were going to do.  I thought all the checks were going to be limited to the new overload by first checking the intrinsic overload, then if it was the new one, do all the other checks.  Like so:

```
if it's SamplerComparisonTy:
  if < SM 6.8:
    emit diagnostic
  check shader stage
  check threadgroup dimensions for compute-like stages
```

This code looks like it does this:
```
if < SM 6.8:
  if new intrinsic:
    emit diagnostic
else: // SM is 6.8+
  do the rest of the diagnostics for shader stage and threadgroup dimensions
```

With this code, existing intrinsic usage will hit the new diagnostics when compiling to SM 6.8.

python3kgae: Updated.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11156,60 +11156,66 @@ void Sema::DiagnoseReachableHLSLMethodCall(const CXXMethodDecl *MD,
...
+
+        if (!SM->IsSM68Plus()) {
+
           Diags.Report(Loc,
                        diag::warn_hlsl_intrinsic_overload_in_wrong_shader_model)
               << MD->getNameAsString() + " with SamplerComparisonState"
               << "6.8";
-        }
-      }
+        } else {
```
pow2clk, line 11170: I don't think this is quite what Tex was suggesting in the pseudo-code [here](https://github.com/microsoft/DirectXShaderCompiler/pull/6244#discussion_r1480848992). I think there was some meaningful whitespace missed. Here's how I interpretted it with braces for clarity: 
```C++
if < SM 6.8 {
  if new intrinsic {
    // emit diagnostic
  }
} else { // SM is 6.8+
   //do the rest of the diagnostics for shader stage and threadgroup dimensions
}
```

@tex3d can you confirm my undertanding?

tools/clang/test/SemaHLSL/hlsl/objects/texture/CalculateLODExtraDiag.hlsl
```diff
@@ -2,7 +2,7 @@
 
 // Check cs/as/mesh and node.
 
-SamplerState ss : register(s2);
+SamplerComparisonState ss : register(s2);
```
pow2clk, line 5: I think this change was made to make these diagnostics show up in the earlier understanding of the pseudo-code. I think this can stay a SamplerState, though perhaps we should have both and mix up uses a bit.

tools/clang/test/SemaHLSL/hlsl/objects/texture/CalculateLODWithSamplerComparisonState.hlsl
pow2clk: I love the new test that was added, but I didn't mean to imply that this code should be removed. It's perfectly fine, I just think that we needed another test.

pow2clk (commented):
I feel there's been a misunderstanding here. I hope I'm able to clarify it, but I'd appreciate @tex3d's confirmation.

tools/clang/test/SemaHLSL/hlsl/objects/texture/CalculateLODExtraDiag.hlsl
```diff
@@ -76,3 +84,24 @@ void node02()
     o[0] = t1.CalculateLevelOfDetail(ss, 0.5);
 }
 
+// expected-note@+2 {{declared here}}
+[Shader("vertex")]
+float4 vs(float2 a :A) :SV_POSTION {
+  float r = 0;
+  if (1>3)
+    // expected-error@+1 {{Intrinsic CalculateLevelOfDetail potentially used by vs requires derivatives - only available in pixel, compute, amplification, mesh, or broadcast node shaders}}
```
bob80905, line 92: Nit: I think this diagnostic message would be best written with single quotes surrounding the variable name, like:
`'vs'`
Not sure how many tests this would affect?

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: In conversation, I was corrected that the intent was to only produce these diagnostics on the new overloads so that old code can be compiled in 6.8 without any issue, so I withdraw my complaint. 

Nevermind I converted the wrong version of the pseudocode above 🤦

tools/clang/test/SemaHLSL/hlsl/objects/texture/CalculateLODExtraDiag.hlsl
pow2clk: Per above, this is no longer relevant.

pow2clk (approved):
Per my resolved issue, I think this is ready

tex3d (approved):
LGTM!

tools/clang/test/SemaHLSL/hlsl/objects/texture/CalculateLODExtraDiag.hlsl
python3kgae: Updated.

bob80905 (approved):
LGTM

Issue #6222, opened 2024-01-30
[SPIR-V] Wave intrinsicts are not guarenteed to have the correct active lanes


s-perron:
**Description**

SPIR-V does not guarantee that correct threads in the wave are active for wave operations.

Consider this example: https://godbolt.org/z/aWqf1a41d.

**Steps to Reproduce**

https://godbolt.org/z/aWqf1a41d

**Actual Behavior**

Problem 1; spirv-opt is merging the break block containing the wave operation with the merge block for the loop, and making the new block the merge block. This means that all threads will have reconverged and execute the wave operation together. We need to fix spirv-opt. (Note that this is technically not incorrect for spirv-opt to do given the convergence requirements. See problem 2.)

Problem 2: The SPIR-V spec allows implementation to decide when the threads reconverge:

> Control flow may partially or fully reconverge before reaching the merge block so long as it converges by the time the merge block is reached.

This makes it impossible to guarantee that an implementation will given the result that is expected from the HLSL. This has been fixed by the  [SPV_KHR_maximal_reconvergence](https://github.com/KhronosGroup/SPIRV-Registry/pull/232) extension. We should start using this in DXC. Based on the discussion in https://github.com/microsoft/hlsl-specs/issues/164, we want to add a new option to DXC that will set the new execution mode. Initially this should default to not setting the execution mode, but eventually the default should switch.

- [x] Fix merge blocks in spirv-opt.
- [x] Add command line option to add the MaximallyReconvergesKHR execution mode.

**Environment**
- DXC version: all
- Host Operating System: all

s-perron:
Submodules have been updated. All of the tasks are done.

PR #6248, opened 2024-02-06
[SPIR-V] Add fspv-enable-maximal-reconvergence opt

Keenuts:
This option is not enabled by default. When enabled, this option adds the MaximallyReconvergesKHR execution mode to the module, and the associated extension.

Related to #6222

### DIFF
--- a/tools/clang/lib/SPIRV/CapabilityVisitor.cpp
+++ b/tools/clang/lib/SPIRV/CapabilityVisitor.cpp
@@ -718,6 +718,10 @@ bool CapabilityVisitor::visit(SpirvExecutionMode *execMode) {
     addExtension(Extension::EXT_shader_stencil_export,
                  "[[vk::stencil_ref_less_equal_back]]", execModeSourceLocation);
     break;
+  case spv::ExecutionMode::MaximallyReconvergesKHR:
+    addExtension(Extension::KHR_maximal_reconvergence, "",
+                 execModeSourceLocation);
+    break;
   default:
     break;
   }
--- a/tools/clang/test/CodeGenSPIRV/spv.enable-maximal-reconvergence.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/spv.enable-maximal-reconvergence.hlsl
@@ -1,8 +1,11 @@
-// RUN: %dxc -T cs_6_0 -E main -fspv-enable-maximal-reconvergence %s -spirv | FileCheck %s
+// RUN: %dxc -T cs_6_0 -E main -fspv-enable-maximal-reconvergence %s -spirv | FileCheck %s -check-prefix=CHECK-ENABLED
+// RUN: %dxc -T cs_6_0 -E main %s -spirv | FileCheck %s -check-prefix=CHECK-DISABLED
 
-// CHECK:       OpExtension "SPV_KHR_maximal_reconvergence"
-// CHECK:       OpEntryPoint GLCompute %main "main"
-// CHECK:       OpExecutionMode %main MaximallyReconvergesKHR
+// CHECK-ENABLED:            OpExtension "SPV_KHR_maximal_reconvergence"
+// CHECK-DISABLED-NOT:       OpExtension "SPV_KHR_maximal_reconvergence"
+// CHECK:                    OpEntryPoint GLCompute %main "main"
+// CHECK-ENABLED:            OpExecutionMode %main MaximallyReconvergesKHR
+// CHECK-DISABLED-NOT:       OpExecutionMode %main MaximallyReconvergesKHR
 [numthreads(1, 1, 1)]
 void main() {}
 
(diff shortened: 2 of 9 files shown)


tools/clang/lib/SPIRV/CapabilityVisitor.cpp
```diff
@@ -718,6 +718,10 @@ bool CapabilityVisitor::visit(SpirvExecutionMode *execMode) {
     addExtension(Extension::EXT_shader_stencil_export,
                  "[[vk::stencil_ref_less_equal_back]]", execModeSourceLocation);
     break;
+  case spv::ExecutionMode::MaximallyReconvergesKHR:
+    addExtension(Extension::KHR_maximal_reconvergence, "",
```
sudonatalie, line 722: Is the recommended approach to unconditionally add extensions and then use the capability trimming pass? I assume that would work here.

tools/clang/test/CodeGenSPIRV/spv.enable-maximal-reconvergence.hlsl
```diff
@@ -0,0 +1,8 @@
+// RUN: %dxc -T cs_6_0 -E main -fspv-enable-maximal-reconvergence %s -spirv | FileCheck %s
```
s-perron, line 1: Could you add another run command that checks that the extension and execution mode are not there when the option is not used?

tools/clang/lib/SPIRV/CapabilityVisitor.cpp
s-perron: Yup, we should be able to remove it if the execution mode is not used.

s-perron (changes requested):

tools/clang/lib/SPIRV/CapabilityVisitor.cpp
Keenuts: Yeah, I wanted to do that, as it seems natural, but the issue is: MaximallyReconvergesKHR is not a capability, nor requires one.
It's only an operand and an extension.

The way the capability trimming pass is built is it relies on capabilities first.
- we have a list of allowed capabilities to trim.
- we figure out what capabilities from that list can be removed.
- we list the extensions linked to those capabilities.
- we trim the capabilities & extensions found in the previous steps.

For this to be trimmed by the pass, the pass would need to be changed:
 - we'd need to also have a list of extensions we are allowed to remove
 - and a way to link operands usage to extension requirement.

tools/clang/test/CodeGenSPIRV/spv.enable-maximal-reconvergence.hlsl
Keenuts: Sure, done

s-perron: You don't need a separate file. You can add multiple run commands on one file with different patterns to look for.
```suggestion
// RUN: %dxc -T cs_6_0 -E main -fspv-enable-maximal-reconvergence %s -spirv | FileCheck %s -check-prefix=ENABLED
// RUN: %dxc -T cs_6_0 -E main %s -spirv | FileCheck %s -check-prefix=DISABLED
```

Then you can add the checks for using `ENABLED` or `DISABLED` instead of `CHECK`.

Keenuts: Ah yes, better to keep them close. Merged the 2 tests.

tools/clang/test/CodeGenSPIRV/spv.enable-maximal-reconvergence.hlsl
```diff
@@ -0,0 +1,11 @@
+// RUN: %dxc -T cs_6_0 -E main -fspv-enable-maximal-reconvergence %s -spirv | FileCheck %s -check-prefix=CHECK-ENABLED
+// RUN: %dxc -T cs_6_0 -E main %s -spirv | FileCheck %s -check-prefix=CHECK-DISABLED
+
+// CHECK-ENABLED:            OpExtension "SPV_KHR_maximal_reconvergence"
+// CHECK-DISABLED-NOT:       OpExtension "SPV_KHR_maximal_reconvergence"
+// CHECK:                    OpEntryPoint GLCompute %main "main"
```
s-perron, line 6: Does this get checked? I through that "CHECK" was the default, but was not checked if you specify a prefix.

Keenuts: Good catch. The command is not additive. re-added the CHECK prefix too to have the shared one. Checked by modifying the lines, and the 2 CHECKS are checked, flags are not exclusive.

sudonatalie (approved):

s-perron (approved):

PR #6252, opened 2024-02-07
[Sema] Add diagnostics for incompatible combination of System Values and node Launch Types. 

bob80905:
According to this section of the spec:
https://microsoft.github.io/DirectX-Specs/d3d/WorkGraphs.html#node-shader-system-values
There should be diagnostics announcing the unsupported use of certain system value semantics in combination with certain supported launch modes. There were previously no diagnostics for these incompatible combinations. 
This PR adds the diagnostics into Sema. 
According to issue https://github.com/microsoft/DirectXShaderCompiler/issues/5827, there is no path for proper flattened signature construction. Parameters won't be fully flattened at the time that Sema runs, so the diagnostics in this PR only trigger on simple parameters with low dimensionality.
The team has agreed to leave the full implementation (which can address higher-dimensional parameters) to the implementation of a proper parameter parsing infrastructure, which will take place in the modernization effort.
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5827


tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -475,6 +477,18 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
                                             DiagnosedCalls);
       Visitor.TraverseDecl(FD);
+
+      // diagnose any node functions that have incompatible launch types with
+      // the present SV semantics on each parameter.
+      if (EntrySK == DXIL::ShaderKind::Node) {
+        if (const auto *NodeLaunchAttr =
+                FDecl->getAttr<clang::HLSLNodeLaunchAttr>()) {
+          llvm::StringRef NodeLaunchTyStr = NodeLaunchAttr->getLaunchType();
+          hlsl::DXIL::NodeLaunchType NodeLaunchTy =
```
python3kgae, line 487: Is it possible to do this in Sema instead of hlsl::DiagnoseTranslationUnit?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11144,6 +11144,51 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+    for (const hlsl::UnusualAnnotation *it : param->getUnusualAnnotations()) {
+      if (it->getKind() == hlsl::UnusualAnnotation::UA_SemanticDecl) {
+        const hlsl::SemanticDecl *sd = cast<hlsl::SemanticDecl>(it);
+        // if the node launch type is Thread, then there are no system values
+        // allowed
+        if (LaunchTy == DXIL::NodeLaunchType::Thread) {
+          if (sd->SemanticName.equals("SV_GroupThreadID") ||
+              sd->SemanticName.equals("SV_GroupIndex") ||
+              sd->SemanticName.equals("SV_GroupID") ||
+              sd->SemanticName.equals("SV_DispatchThreadID")) {
```
pow2clk, line 11160: Question: As long as we're string comparing, can we go broader and simpler by just checking the "SV_" prefix? Do we need to check this at all? Are all semantic decls invalid here?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11144,6 +11144,51 @@ bool Sema::DiagnoseHLSLMethodCall(const CXXMethodDecl *MD, SourceLocation Loc) {
...
+            Diags.Report(param->getLocation(), DiagID)
+                << sd->SemanticName << "Thread";
+          }
+        }
+
+        // if the node launch type is Coalescing, then
+        // SV_GroupIndex and SV_GroupThreadID are allowed
+        else if (LaunchTy == DXIL::NodeLaunchType::Coalescing) {
+          if (sd->SemanticName.equals("SV_GroupID") ||
+              sd->SemanticName.equals("SV_DispatchThreadID")) {
```
pow2clk, line 11174: Similar to the above, should we be checking for the few that are invalid and expected or the two that are valid and erroring on all that aren't those?

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -475,6 +477,18 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
       HLSLMethodCallDiagnoseVisitor Visitor(self, shaderModel, EntrySK, FDecl,
                                             DiagnosedCalls);
       Visitor.TraverseDecl(FD);
+
+      // diagnose any node functions that have incompatible launch types with
+      // the present SV semantics on each parameter.
+      if (EntrySK == DXIL::ShaderKind::Node) {
+        if (const auto *NodeLaunchAttr =
+                FDecl->getAttr<clang::HLSLNodeLaunchAttr>()) {
```
pow2clk, line 485: Question: Can we not assume that there is a launch attr for a node shader at this point? Do we catch that?

pow2clk (approved):
Looks good overall. I suggested a comment and have some questions. If those questions don't inspire additional changes, then I'm fine with how it is. 

Thanks!

tools/clang/test/HLSLFileCheck/hlsl/diagnostics/errors/launch_types.hlsl
```diff
@@ -0,0 +1,39 @@
+// RUN: %dxc -T lib_6_3 -D e1 %s | FileCheck -check-prefix=CHECK_ENTRY1 %s
```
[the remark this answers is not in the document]
pow2clk: Oh I approve this message too.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: According to the spec, all *compute* System values are invalid. I'll see if just checking SV causes failures in other tests, and if not, leave the change in.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: Excellent suggestion!

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
bob80905: I don't think we can do this in sema because we have to account for library diagnostics. Sema has DiagnoseEntry, which runs on shaders that are not the library shader kind. SemaHLSLDiagnoseTU runs with library shader kinds. So we have to add this here to account for the case where a shader in a library file has the node shader attribute and an invalid nodeLaunch type.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
bob80905: We don't catch it before this point, and we cannot assume there will be a launch attr.
One example of this is tools\clang\test\SemaHLSL\shader_attribute_no_mismatch.hlsl:
```
[shader("node")] 
[shader("node")] 
[nodedispatchgrid(8,1,1)]
[ numthreads( 64, 2, 2 ) ]  /* expected-no-diagnostics */
void NNMain() {
}
```
which will cause an ```Internal compiler error: access violation. Attempted to read from address 0x0000000000000010```
So I must leave the check here.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: I've left both recommended changes in.

python3kgae (approved):

PR #6255, opened 2024-02-07
[SPIR-V] Process array subscript individually

s-perron:
In the current implementation, a single OpAccessChain is generated for
all array subscripts and member accesses. This causes a problem if the
member access is an aliased variable, and we need to generate and
OpAccessChains for the member, a load, and then another OpAccessChain
to index into the aliased array.

This change cause an array subscript to be process as a single
OpAccessChian. Note that we still process array sub scripts with the
member access when it is a subnode of the member access. We will
probably need to fix that up later as well, but I want to keep the
change small.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/6110.


cassiebeckley (commented):
Will splitting `OpAccessChain` instructions like this have any performance impact? Does the optimizer know to combine these?

cassiebeckley (approved):

sudonatalie (approved):

Issue #6203, opened 2024-01-25
DXC 1.8.2403 Release


llvm-beanz:
The proposed schedule for this release is below:

![DXC 1.8.2403 Schedule](https://github.com/microsoft/DirectXShaderCompiler/assets/15788759/d5bae1ec-061d-4011-b1d8-b1ce2ad1ae60)

* 1/26/2024 - Create the release branch from `main`
* 2/12/2024 - Release Candidate 1 (begin Ask Mode[^1] for release branch)
* 3/4/2024 - Final Release Candidate
* 3/11/2024 - Final Release Date

Tasks: 
* [x] 1/25/2024 - Create release branch from `main` (@hekota)
* [x] 1/27/2024 - Update README.md (@pow2clk)
* [x] 3/1/2024 - Create draft of Release post on GitHub (@pow2clk)
* [x] 3/11/2024 - Tag final release and post binaries. (@hekota)

Quality Sign-off
* [x] 3/3/2024 - Microsoft Testing Sign-off (@pow2clk)
* [x] 3/3/2024 - Google Testing Sign-off (@sudonatalie)

[^1]: [Ask Mode](https://devblogs.microsoft.com/oldnewthing/20140722-00/?p=433) is a Microsoft-ism to denote when changes require approval before accepting merges. For DXC this will require approval from @microsoft/hlsl-release

sudonatalie:
Internal Google testing passed on the `release-1.8.2403` branch. Feel free to ping me if/when re-testing is necessary assuming more changes come in before the final release candidate.

sudonatalie:
Signing off that `v1.8.2403` has passed internal testing.

PR #6257, opened 2024-02-07
Add CODEOWNERS to release

llvm-beanz:
As per the schedule in #6203, we will begin AskMode on 2/12, at which time all PRs will be required to have at least one sign off from a member of @microsoft/hlsl-release.

Adding this CODEOWNERS file allows us to have GitHub enforce that requirement.

Note: This PR does not enable AskMode, it just puts the infrastructure in place. AskMode will not be enabled until 2/12.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,4 @@
+# DXC release branches require approval from `hlsl-release` group starting from
+# the release AskMode date.
+
+* @microsoft/hlsl-release


dmpots:
The PR message may be out of date. It says it won't enable ask mode until 2/12, but today is 2/12.

dmpots (approved):

pow2clk (approved):

PR #6268, opened 2024-02-10
[SPIR-V] Remove duplication of raytracing stage vars

### LINKED ISSUE #6265: [SPIR-V] All interface variables for all entry points are added to the interface for each entry point
This issue was found when a check was added to spirv-val to prevent [multiple interface variables with the same storage class](https://github.com/KhronosGroup/SPIRV-Tools/pull/4969#issuecomment-1334918901).

For the test [vk.layout.shader-record-ext.std430.hlsl](https://github.com/microsoft/DirectXShaderCompiler/blob/c997ea02605d2bc30176cfab14fa205ece2f5af9/tools/clang/test/CodeGenSPIRV/vk.layout.shader-record-ext.std430.hlsl), the closest hit entry points

```hlsl
[shader("closesthit")]
void chs1(inout Payload P, in Attr A) {
    P.p = cbuf.f1;
}

[shader("closesthit")]
void chs2(inout Payload P, in Attr A) {
    P.p = f1;
}
```

are lowered to the interface

```
               OpEntryPoint ClosestHitNV %chs1 "chs1" %cbuf %block %P %A %P_0 %A_0
               OpEntryPoint ClosestHitNV %chs2 "chs2" %cbuf %block %P %A %P_0 %A_0
```

because `%P` and `%A` are the arguments to `chs1`, and `%P_0` and `%A_0` are the arguments to `chs2`.

cassiebeckley:
Only add raytracing stage var to `OpEntryPoint` interface if it is actually used by that entry point.

Fixes #6265.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -757,6 +757,11 @@ SpirvEmitter::getInterfacesForEntryPoint(SpirvFunction *entryPoint) {
   for (auto *moduleVar : spvBuilder.getModule()->getVariables()) {
     if (moduleVar->getStorageClass() != spv::StorageClass::Input &&
         moduleVar->getStorageClass() != spv::StorageClass::Output) {
+      if (auto *varEntry =
+              declIdMapper.getRayTracingStageVarEntryFunction(moduleVar)) {
+        if (varEntry != entryPoint)
+          continue;
+      }
       interfaces.insert(moduleVar);
     }
   }
--- a/tools/clang/test/CodeGenSPIRV/vk.layout.shader-record-ext.std430.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/vk.layout.shader-record-ext.std430.hlsl
@@ -1,5 +1,8 @@
 // RUN: %dxc -T lib_6_3 -fspv-target-env=vulkan1.2 -fvk-use-gl-layout -fcgl  %s -spirv | FileCheck %s
 
+// CHECK: OpEntryPoint ClosestHitNV %chs1 "chs1" %cbuf %block %P %A
+// CHECK: OpEntryPoint ClosestHitNV %chs2 "chs2" %cbuf %block %P_0 %A_0
+
 // CHECK: OpDecorate %_arr_v2float_uint_3 ArrayStride 8
 // CHECK: OpDecorate %_arr_mat3v2float_uint_2 ArrayStride 32
 // CHECK: OpDecorate %_arr_v2int_uint_3 ArrayStride 8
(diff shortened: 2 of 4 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -757,6 +757,11 @@ SpirvEmitter::getInterfacesForEntryPoint(SpirvFunction *entryPoint) {
   for (auto *moduleVar : spvBuilder.getModule()->getVariables()) {
     if (moduleVar->getStorageClass() != spv::StorageClass::Input &&
         moduleVar->getStorageClass() != spv::StorageClass::Output) {
```
s-perron, line 759: Is this only a problem for ray tracing? What happens if we have multiple fragment shaders in a `lib`?

tools/clang/test/CodeGenSPIRV/vk.layout.shader-record-ext.std430.hlsl
```diff
@@ -1,5 +1,8 @@
 // RUN: %dxc -T lib_6_3 -fspv-target-env=vulkan1.2 -fvk-use-gl-layout -fcgl  %s -spirv | FileCheck %s
 
+// CHECK: OpEntryPoint ClosestHitNV %chs1 "chs1" %cbuf %block %P %A
+// CHECK: OpEntryPoint ClosestHitNV %chs2 "chs2" %cbuf %block %P_0 %A_0
```
s-perron, line 4: Does this change the location decorations for the input and output variables?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: I believe it's only a problem for ray tracing, since other shader types keep track of their stage variables in the `stageVars` vector, but ray tracing shaders do not – https://github.com/microsoft/DirectXShaderCompiler/blob/30f45bcf0f56fe675d797bde700891f4a6a816c3/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp#L4524.

tools/clang/test/CodeGenSPIRV/vk.layout.shader-record-ext.std430.hlsl
cassiebeckley: I don't think so; I don't think closest hit shaders have any variables in the `Input` or `Output` storage class. (`%cbuf` and `%block` are `ShaderRecordBufferNV`, `%P` is `IncomingRayPayloadNV`, and `%A` is `HitAttributeNV`)

s-perron (approved):

sudonatalie (approved):

PR #6277, opened 2024-02-10
Correct metadata load for NodeRecordType for robustness

### LINKED ISSUE #6276: DxilMetadata loading code for NodeRecordType properties is incorrect/fragile
**Description**
The DxilMetadata loading code for loading NodeRecordType from key,value property list ignores the tag keys and just assumes positions for the values.  This is incorrect because this pattern is supposed to iterate through pairs of key,value and switch on the key, loading the value when the key is recognized, and setting m_bExtraMetadata flag if a key is not recognized.

This doesn't currently create a symptom, but would not be robust to a different ordering or additional properties being added.  This needs to be fixed ASAP if drivers use this code to load metadata, since they will be fragile otherwise.

**Environment**
- DXC version: latest main and release branches
- Host Operating System any

tex3d:
Metadata loading for node record properties assumed fixed positions for values in a (key,value) property list, skipping and ignoring the keys.

This was not correct, fragile, and not the intended implementation of the key,value property pattern.

There is no functional change, but this makes the loader robust to additional properties instead of crashing or loading the wrong thing.

Fixes #6276

### DIFF
--- a/lib/DXIL/DxilMetadataHelper.cpp
+++ b/lib/DXIL/DxilMetadataHelper.cpp
@@ -2815,6 +2815,40 @@ DxilMDHelper::EmitDxilNodeIOState(const hlsl::NodeIOProperties &Node) {
   return MDNode::get(m_Ctx, MDVals);
 }
 
+NodeRecordType
+DxilMDHelper::LoadDxilNodeRecordType(const llvm::MDOperand &MDO) {
+  const MDTuple *pTupleMD = dyn_cast<MDTuple>(MDO.get());
+  IFTBOOL(pTupleMD != nullptr, DXC_E_INCORRECT_DXIL_METADATA);
+  IFTBOOL((pTupleMD->getNumOperands() & 0x1) == 0,
+          DXC_E_INCORRECT_DXIL_METADATA);
+
+  NodeRecordType Record = {};
+  for (unsigned iNode = 0; iNode < pTupleMD->getNumOperands(); iNode += 2) {
+    unsigned Tag = DxilMDHelper::ConstMDToUint32(pTupleMD->getOperand(iNode));
+    const MDOperand &MDO = pTupleMD->getOperand(iNode + 1);
+    IFTBOOL(MDO.get() != nullptr, DXC_E_INCORRECT_DXIL_METADATA);
+
+    switch (Tag) {
+    case DxilMDHelper::kDxilNodeRecordSizeTag: {
+      Record.size = ConstMDToUint32(MDO);
+    } break;
+    case DxilMDHelper::kDxilNodeSVDispatchGridTag: {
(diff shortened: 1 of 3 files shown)


lib/DXIL/DxilMetadataHelper.cpp
```diff
@@ -2815,6 +2815,40 @@ DxilMDHelper::EmitDxilNodeIOState(const hlsl::NodeIOProperties &Node) {
...
+    unsigned Tag = DxilMDHelper::ConstMDToUint32(pTupleMD->getOperand(iNode));
+    const MDOperand &MDO = pTupleMD->getOperand(iNode + 1);
+    IFTBOOL(MDO.get() != nullptr, DXC_E_INCORRECT_DXIL_METADATA);
+
+    switch (Tag) {
+    case DxilMDHelper::kDxilNodeRecordSizeTag: {
+      Record.size = ConstMDToUint32(MDO);
+    } break;
+    case DxilMDHelper::kDxilNodeSVDispatchGridTag: {
+      MDTuple *pSVDTupleMD = cast<MDTuple>(MDO.get());
```
pow2clk, line 2836: Could have another check here to ensure that the tuple has 3 elements.

pow2clk (approved):
Looks pretty simple. I have one possible suggestion.

python3kgae (approved):

tools/clang/test/HLSLFileCheck/hlsl/workgraph/record_type_metadata.ll
```diff
@@ -0,0 +1,46 @@
...
+
+; RDAT: FunctionTable[{{.*}}] = {
+; RDAT-LABEL: UnmangledName: "node01"
+; RDAT: Inputs: <11:RecordArrayRef<IONode>[1]>  = {
+; RDAT: AttribKind: RecordSizeInBytes
+; RDAT-NEXT: RecordSizeInBytes: 12
+; RDAT: AttribKind: RecordDispatchGrid
+; RDAT-NEXT: RecordDispatchGrid: <RecordDispatchGrid>
+; RDAT-NEXT: ByteOffset: 0
+; RDAT-NEXT: ComponentNumAndType: 23
```
bob80905, line 13: Where does this 23 come from?

tex3d: This is a combination of `NumComponents` (`3`) and `DXIL::ComponentType::U32` (`5`) `<< 2`:
`0:2 = NumComponents (0-3)`
`3:15 = hlsl::DXIL::ComponentType enum`

https://github.com/microsoft/DirectXShaderCompiler/blob/2872d2d0e7eeec9c4caec91ff3858f4f54f31f7b/include/dxc/DxilContainer/RDAT_LibraryTypes.inl#L322-L323

PR #6279, opened 2024-02-10
Add alignment to NodeRecordType including DXIL metadata update

### LINKED ISSUE #6270: RDAT needs Node Record Alignment
**Description**
Record alignment info for node records is needed in RDAT to enforce pointer alignment in the runtime.  This is currently missing from the existing RDAT info for node records.  Adding this to RDAT is pretty simple, but it will also need to be added to DxilMetadata, making it a DXIL breaking change.

Adding this to RDAT will require a new `NodeAttribKind::RecordAlignmentInBytes` enum and corresponding union case in the `NodeShaderIOAttrib` record.

It looks like extending the `NodeRecordType` property isn't possible without fixing the loading code that implemented the tag,value pattern incorrectly for `NodeRecordType`.  See: #6276.

It's possible we could do something to avoid metadata modification in this version:
Upon load of DxilModule, restore alignment property by walking from GetRecordPtr up handles to identify input or output, then fill in the alignment from the llvm structure returned from GetRecordPtr.  DxilValidation will also need to restore alignment property for unused records from RDAT.

tex3d:
This change adds NodeRecordType alignment field to RDAT to make it possible to validate pointer and stride alignment in the runtime.

This includes a change to DXIL metadata to preserve the record alignment without requiring recovery by looking for GetNodeRecordPtr.

Fixes #6270

### DIFF
--- a/include/dxc/DXIL/DxilMetadataHelper.h
+++ b/include/dxc/DXIL/DxilMetadataHelper.h
@@ -624,6 +624,7 @@ private:
                        unsigned &payloadSizeInBytes);
 
   llvm::MDTuple *EmitDxilNodeIOState(const NodeIOProperties &Node);
+  llvm::MDTuple *EmitDxilNodeRecordType(const NodeRecordType &RecordType);
   hlsl::NodeIOProperties LoadDxilNodeIOState(const llvm::MDOperand &MDO);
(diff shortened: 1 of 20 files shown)


pow2clk (approved):
Looks good!

PR #6280, opened 2024-02-10
NFC: Fix DXIL op is_const for Barrier and node create handle ops

tex3d:
DXIL operations should use is_const=True for constant arguments.  This allows for convenience methods to retrieve the constant value, and could (should, but currently doesn't) result in validation that the argument is constant.

`BarrierByNodeRecordHandle` SemanticFlags argument must be constant.
`MetadataIdx` for both `createNodeOutputHandle` and `CreateNodeInputRecordHandle` must be constant.


pow2clk (approved):
Looks good!

python3kgae (approved):

PR #6281, opened 2024-02-11
PIX: Optimize debug instrumentation for fewest emitted instructions

jeffnn:
This change switches from instrumentation per instruction to instrumentation per basic block. Furthermore, not every instruction in a basic block needs to write debug data to the output UAV- it's enough to know that the basic block was entered, as long as the calling application can figure out what instructions were in that block. To support that knowledge, the pass now emits a text "precis" of each basic block.

Also, the previous branchless UAV bounds enforcement was replaced with something similar that emits fewer instructions at the cost of a larger UAV. This tradeoff is WELL worth it.

Additionally, the debug pass used to add extra blocks in order to solidify the arguments to phi instructions. This work was unnecessary, and added a lot of complexity to the resulting instrumented shader. The debugger application is only interested in the value of the phi itself and the actual value produced via the actual preceding edge.
 
Full details are in the comments in the code.

This change reduces driver-side compilation overhead from "overnight" to 2 minutes on a 160k-instruction shader.


lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -34,69 +35,103 @@ using namespace hlsl;
...
+// because several threads may satisfy the selection criteria (e.g. a pixel
+// shader may be invoked several times for a given pixel coordinate if the model
+// has overlapping triangles).
+//
+// A picture of the UAV layout:
+// <--------------power-of-two-size-of-UAV---------------->
+// [1           ][2           ][3           ][4           ]
+// <------A----->             ^                           ^
+//                            B                           C
+//                            <------D------>
```
python3kgae, line 69: What does the A/B/C/D mean in this table?

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -793,7 +929,8 @@ void DxilDebugInstrumentation::addDebugEntryValue(BuilderContext &BC,
                      WriteMask_X});
 
     m_RemainingReservedSpaceInBytes -= 4;
-    assert(m_RemainingReservedSpaceInBytes < 1024); // check for underflow
+    assert(m_RemainingReservedSpaceInBytes <
+           1024 * 1024); // check for underflow
```
python3kgae, line 933: Does this mean we could reserve smaller space?

jeffnn: The m_RemainingReservedSpaceInBytes is expected to be exact for the calculated debug payload for the given block. 
So, yes, this assert would be wrong if we tried to instrument a block with > 1M payload. Good catch! I'll reformulate that to be more obvious.

python3kgae (approved):

PR #6283, opened 2024-02-12
[SPIRV] Using 64-bit int literals in flat conversion

### LINKED ISSUE #6188: [SPIR-V] Does not handle flat conversion from large literal
**Description**

**Steps to Reproduce**

```
struct Error {
  int32_t i;
  int64_t j;
};
  

RWStructuredBuffer<Error> b;

[numthreads(1, 1, 1)]
void main()
{
   b[0] = (Error)(0x100000000+1);
}

```

https://godbolt.org/z/Es35cGcdj

**Actual Behavior**

The current behaviour is a spir-v validation error because of an incorrect type.

**Environment**
- DXC version all
- Host Operating System all

s-perron:
During flat convertion, we have to pick the bit width for literals
because they could be used in multiple places as different sizes, and
that breaks the literal visitor.  We currently use 32-bit types because
that is the most common use case, and we don't want to unnecessarily
add a capability.

However, if we pick a 64-bit width, we should be able to fold all of
the operations on literals. If they are only used as 32-bit value, then
ultimatlly fold them to 32-bit constants and remove the 64-bit
capabilities.

This commit updates the spir-v submodule to pull in the 64-bit interger
folding changes, and then starts to use 64-bit integer literal in flat conversion.

Note that the problem still exists for floating point values. Spirv-opt
does not fold OpFConvert yet, we still treat the literals like a float.

Fixes #6188

mend

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/cast.flat-conversion.literal-initializer.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/cast.flat-conversion.literal-initializer.hlsl
@@ -59,8 +59,8 @@ void main() {
   S s1 = (S)(a);
 
 // TODO(6188): This is wrong because we lose most significant bits in the literal.
-// CHECK: [[lit:%[0-9]+]] = OpIAdd %int %int_0 %int_1
-// CHECK: [[longLit:%[0-9]+]] = OpSConvert %long [[lit]]
+// CHECK: [[longLit:%[0-9]+]] = OpIAdd %long %long_4294967296 %long_1
+// CHECK: [[lit:%[0-9]+]] = OpSConvert %int [[longLit]]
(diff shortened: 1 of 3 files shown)


tools/clang/test/CodeGenSPIRV/cast.flat-conversion.literal-initializer.hlsl
```diff
@@ -59,8 +59,8 @@ void main() {
   S s1 = (S)(a);
 
 // TODO(6188): This is wrong because we lose most significant bits in the literal.
```
sudonatalie, line 61:
```suggestion
```

sudonatalie (approved):

cassiebeckley (approved):

PR #6284, opened 2024-02-12
[Sema] Only emit warnings on entry point attrs with missing shader attr for libraries

### LINKED ISSUE #6100: Add warning for entry-only attribute without the shader attribute when compiling to library target

bob80905:
In implementing this PR:
https://github.com/microsoft/DirectXShaderCompiler/pull/6243
somehow, the condition to only emit the warning for library shaders was left out.
This PR is a follow up on #6243 to properly restrict the emission of this warning for non-library shaders.
This should properly fix #6100


python3kgae (approved):

tools/clang/include/clang/Basic/DiagnosticGroups.td
```diff
@@ -783,6 +783,7 @@ def HLSLEffectsSyntax : DiagGroup<"effects-syntax">;
 def HLSLForRedefinition : DiagGroup<"for-redefinition">;
 def HLSLAttributeStatement : DiagGroup<"attribute-statement">;
 def HLSLAttributeType : DiagGroup<"attribute-type">;
+def HLSLEntryAttributeWithoutShaderAttrType : DiagGroup<"entry-attribute-type-without-shader-attribute-type">;
```
pow2clk, line 786: It's kinda long. Maybe something as simple as "non-entry-attributes", "misplaced-attributes"?

pow2clk (approved):
I think the group name is a bit unwieldy. Otherwise this is good.

pow2clk (approved):
Looks good!

tex3d:
Sorry I missed the review of this earlier.  While `misplaced-attributes` is an improvement, I think something like `ignored-entry-attributes` might have been more accurate/appropriate.

I also realize that the warning could have been worded a little more clearly for the impact.  It's currently "attribute '%0' ignored without accompanying shader attribute", but I think this would have been better "entry attribute %0 ignored; entry functions require [shader(...)] attribute for inclusion in library".

PR #6288, opened 2024-02-13
NFC: Fix switch with missing default case in DxilShaderModel.cpp

### LINKED ISSUE #6287: Build: switch missing default case breaks build when building with /Wall and /WX
**Description**
This switch is missing a default case, which cases builds with stricter settings to fail when not all values from the enum are handled explicitly.

https://github.com/microsoft/DirectXShaderCompiler/blob/dbf60d777156b98ea2f8cafc8336bc62bf65213c/lib/DXIL/DxilShaderModel.cpp#L498-L507

**Steps to Reproduce**
For Visual C++ compiler: build DXC with `/Wall` and `/WX`

Expect build to succeed.

**Actual Behavior**
`C4062 enumerator'hlsl::DXIL::ShaderKind::Vertex' in switch of enum 'hlsl::DXIL::ShaderKind' is not handled`
... and so on ...

**Environment**
- DXC version: latest `main`
- Host Operating System: any

tex3d:
This missing default case statement would cause build break when compiling with additional warning settings, such as /Wall /WX in VC++.

Fixes #6287.

### DIFF
--- a/lib/DXIL/DxilShaderModel.cpp
+++ b/lib/DXIL/DxilShaderModel.cpp
@@ -504,8 +504,9 @@ bool ShaderModel::AllowDerivatives(DXIL::ShaderKind sk) const {
   case DXIL::ShaderKind::Amplification:
   case DXIL::ShaderKind::Mesh:
     return IsSM66Plus();
+  default:
+    return false;
   }
-  return false;
 }
 
 typedef ShaderModel SM;
--- a/lib/DXIL/DxilShaderModel.cpp
+++ b/lib/DXIL/DxilShaderModel.cpp
@@ -504,9 +504,22 @@ bool ShaderModel::AllowDerivatives(DXIL::ShaderKind sk) const {
   case DXIL::ShaderKind::Amplification:
   case DXIL::ShaderKind::Mesh:
     return IsSM66Plus();
-  default:
+  case DXIL::ShaderKind::Vertex:
+  case DXIL::ShaderKind::Geometry:
+  case DXIL::ShaderKind::Hull:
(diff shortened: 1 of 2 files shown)


lib/DXIL/DxilShaderModel.cpp
```diff
@@ -504,8 +504,9 @@ bool ShaderModel::AllowDerivatives(DXIL::ShaderKind sk) const {
   case DXIL::ShaderKind::Amplification:
   case DXIL::ShaderKind::Mesh:
     return IsSM66Plus();
+  default:
+    return false;
   }
-  return false;
```
llvm-beanz, line 508: If this is unreachable we should have an `llvm_unreachable` here, otherwise as code changes if a case breaks out instead of returning this becomes UB.

llvm-beanz (approved):
One small change requested. Otherwise looks good.

lib/DXIL/DxilShaderModel.cpp
tex3d: How about I just change this to handle all cases (no default) and with the `llvm_unreachable` to be sure we added all our cases here?

Also, we don't appear to be building with this warning enabled, but I thought we hit this before in some other scenario.  Perhaps we aren't warning clean across the project with this one, but it would be good to be warning clean for the DXIL headers/cpp files that are consumed by other projects at a minimum.

adam-yang (approved):

pow2clk:
I remember fixing warnings like this during the original Linux port. I see that the switch warnings were disabled by #1628. Without it, I do see a lot of warnings from generated file Attrs.inc. I wonder how upstream LLVM deals with that

Issue #6256, opened 2024-02-07
SM6.8 Barrier ops: not fully validated for stage or correctly represent constraints for RDAT


tex3d:
**Description**
New barrier operations in SM 6.8 are not validated for cases only allowed in certain shader models.

The new Barrier operation is underspecified, and has some problems.
- issues with the definitions of equivalent new barrier operations for old barrier intrinsics:
  - it uses `ALL_MEMORY` where the old intrinsic would not have had any concept of node record memory
  - `DeviceMemoryBarrier*` intrinsics should exclude `GROUP_SHARED_MEMORY`
- issues with DXIL definitions:
  - only one incorrect definition for intrinsic receiving handle (mixes name for UAV intrinsic with node record handle type).
  - it's not defined what we do with old intrinsics for SM 6.8: map to new?  keep old?
  - it's not defined whether you can use new intrinsics to define equivalent barriers to old intrinsics on lower shader models.
- overly broad/loose definitions
  - Missing rules to constrain legal uses to certain argument combinations and contexts that make sense.

Incoming proposed spec updates:

- Uniform control flow only required for `GROUP_SYNC`. `DeviceMemoryBarrier*()` excludes `GROUP_SHARED_MEMORY`.
- `MemoryTypeFlags` masked for shader stage, allowing you to use ALL_MEMORY when some memory types are not applicable.
- More detailed rules for valid Barrier() use.
- Fix DXIL ops for UAV Handle vs. NodeRecordHandle.
- Old intrinsics should produce the new DXIL op on SM 6.8.
- Uses of the new HLSL intrinsic equivalent to the old HLSL intrinsics will map to the existing barrier DXIL op on prior shader models.
- A set of flag combinations for translation between SM 6.8 and the older barrier DXIL operation

**DXC Sub-Tasks**

- [x] #6266
- [x] #6292
- [x] #6293

**Missing Validation Example**
Expect validation errors when using Barrier with `GROUP_SYNC` or `GROUP_SCOPE` when shader stage does not have access to that scope.

Example test case:
```ll
; RUN: %dxilver 1.8 | %dxv %s | FileCheck %s

; CHECK-DAG: Function: main: error: sync in a non-Compute/Amplification/Mesh/Node Shader must only sync UAV (sync_uglobal).
; CHECK-DAG: note: at 'call void @dx.op.barrierByMemoryType(i32 244, i32 1, i32 3)' in block '#0' of function 'main'.

[...]

PR #6291, opened 2024-02-13
New barrier DXIL Op fixes for Validation and RDAT function compatibility info

### LINKED ISSUE #6266: Barrier ops: fix legal models and stages for RDAT
`OP::GetMinShaderModelAndMask`: account for new barrier ops for legal shader stages and minimum shader model, including potential for translation.  Add D3DReflect RDAT tests for scenarios.  This impacts the validated RDAT blob, so it's a DXIL container breaking change and should be prioritized first.

tex3d:
Fix barrier allowed ops and flags by shader kind

New barrier operations lacked validation and for RDAT info: had incorrect min target and shader stage flags.

- Identify barrier DXIL operations with new `is_barrier` in `hctdb.py` and generated `OP::IsDxilOpBarrier`.
- Identify when a barrier op requires shader stage with group (compute-like stage), or when it requires node memory.
- Add new `OptFeatureInfo_RequiresGroup` to identify function only compatible with a shader stage with a visible group for access to groupshared memory or use of group sync.
- Translate to original `BarrierMode` when compatible; adds `BarrierMode::Invalid` to identify invalid cases.
- Account for `DXIL::MemoryTypeFlags::AllMemory` being allowed and auto-masked by driver.
- Properly set min shader model and compatible shader stage flags.
- Validate barrier for shader stage.
- Added new barriers to counters which were missing.

Adressing parts of: #6256 and #6292
Fixes #6266


lib/DXIL/DxilOperations.cpp
```diff
@@ -2876,6 +2876,128 @@ bool OP::IsDxilOpFeedback(OpCode C) {
...
+    if (isa<ConstantInt>(barrier.get_MemoryTypeFlags())) {
+      memoryTypeFlags = barrier.get_MemoryTypeFlags_val();
+    }
+    if (isa<ConstantInt>(barrier.get_SemanticFlags())) {
+      semanticFlags = barrier.get_SemanticFlags_val();
+    }
+    // If the memory type is AllMemory, masking inapplicable flags is allowed.
+    if (memoryTypeFlags != (unsigned)DXIL::MemoryTypeFlag::AllMemory) {
+      if (memoryTypeFlags & ((unsigned)DXIL::MemoryTypeFlag::NodeInputMemory |
+                             (unsigned)DXIL::MemoryTypeFlag::NodeOutputMemory))
```
pow2clk, line 2978: Nit: Having this exception in multiple places makes me a little nervous. It might be safer to have a utility function that does this for you.

lib/DXIL/DxilOperations.cpp
```diff
@@ -3249,20 +3381,23 @@ void OP::GetMinShaderModelAndMask(const llvm::CallInst *CI,
...
-  if (opcode == DXIL::OpCode::Barrier) {
-    // Barrier mode should be a constant, but be robust to non-constants here.
-    if (isa<ConstantInt>(
-            CI->getArgOperand(DxilInst_Barrier::arg_barrierMode))) {
-      DxilInst_Barrier barrier(const_cast<CallInst *>(CI));
-      unsigned mode = barrier.get_barrierMode_val();
-      if (mode != (unsigned)DXIL::BarrierMode::UAVFenceGlobal) {
-        mask &= SFLAG(Library) | SFLAG(Compute) | SFLAG(Amplification) |
-                SFLAG(Mesh) | SFLAG(Node);
+  // Barrier requiring node or group limit shdaer kinds.
```
pow2clk, line 3384: typo: shdaer -> shader

lib/DXIL/DxilOperations.cpp
```diff
@@ -3249,20 +3381,23 @@ void OP::GetMinShaderModelAndMask(const llvm::CallInst *CI,
...
+  if (IsDxilOpBarrier(opcode)) {
+    // TODO: if BarrierByMemoryType, check if translatable, or set min to 6.8.
```
pow2clk, line 3386: Looks like that's what the code does? Is something still missing here?

tools/clang/test/HLSLFileCheck/d3dreflect/amp-groupshared.hlsl
```diff
@@ -38,7 +38,7 @@
 // CHECK:      PayloadSizeInBytes: 0
 // CHECK:      AttributeSizeInBytes: 0
 // CHECK:      FeatureInfo1: 0
-// CHECK:      FeatureInfo2: 0
+// CHECK:      FeatureInfo2: (Opt_RequiresGroup)
```
pow2clk, line 41: Reaping the benefits of pretty printing! 😄

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm68_samplecmpgradbias.hlsl
```diff
@@ -60,7 +60,7 @@ void samplecmpbias() {
 // RDAT:   MinShaderTarget: 0x50068
 
 [shader("compute")]
-[numthreads(1,1,1)]
+[numthreads(4,1,1)]
```
pow2clk, line 63: I see why this is required. I'm not sure what about this change enforced that requirement.

tools/clang/tools/dxcompiler/dxcdisassembler.cpp
```diff
@@ -355,6 +355,7 @@ static_assert(_countof(g_pFeatureInfoNames) == ShaderFeatureInfoCount,
 
 PCSTR g_pOptFeatureInfoNames[] = {
     "Function uses derivatives",
+    "Function requires group",
```
pow2clk, line 358: nit: more bikeshedding, but maybe "visible group"?

pow2clk (approved):
Looks good. I have some very minor nits you might choose to address or not.

include/dxc/DXIL/DxilInstructions.h
```diff
@@ -8856,6 +8856,15 @@ struct DxilInst_BarrierByNodeRecordHandle {
   void set_object(llvm::Value *val) { Instr->setOperand(1, val); }
   llvm::Value *get_SemanticFlags() const { return Instr->getOperand(2); }
   void set_SemanticFlags(llvm::Value *val) { Instr->setOperand(2, val); }
```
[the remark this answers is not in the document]
tex3d: `is_const - True` in `hctdb.py` doesn't mean you can't set the argument `llvm::Value`, it means the value must be an immediate `llvm::Constant` value  We don't do additional validation upon setting the value to make sure it's `llvm::Constant` though.  More validation in these generated helpers could be useful, but I don't think it's within the scope of this change.

utils/hct/hctdb.py
```diff
@@ -685,12 +692,20 @@ def populate_categories_and_models(self):
...
+            self.name_idx[i].is_barrier = True
+            self.name_idx[i].shader_model = 6, 8
+            self.name_idx[i].shader_model_translated = 6, 0
+        for i in "BarrierByMemoryHandle".split(","):
+            # included in Synchronization category
+            self.name_idx[i].is_barrier = True
+            self.name_idx[i].shader_model = 6, 8
+        for i in "BarrierByNodeRecordHandle".split(","):
+            # included in Synchronization category
+            self.name_idx[i].is_barrier = True
```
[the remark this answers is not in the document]
tex3d: True for these, the prior one was to capture the flag for all barrier ops, including the pre-existing one.  We could remove it here if you want.  The reason the earlier addition is where it is was due to where `Barrier` was placed in the `Synchronization` category.  That could be moved, but it wouldn't make any difference of course.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm68_samplecmpgradbias.hlsl
tex3d: Ah, the change that enforces this requirement was separated into another branch for independent PR, but forgot to move this change I guess.  Should I remove it or leave it (it is correct).

lib/DXIL/DxilModule.cpp
```diff
@@ -314,15 +314,15 @@ void DxilModule::CollectShaderFlagsForModule(ShaderFlags &Flags) {
   for (auto &itInfo : m_FuncToShaderCompat)
     Flags.CombineShaderFlags(itInfo.second.shaderFlags);
 
-  // Clear UsesDerivatives flag for module, making sure
-  // DerivativesInMeshAndAmpShaders is set for MS/AS.
+  const ShaderModel *SM = GetShaderModel();
```
bob80905, line 317: Why not throw out this variable entirely and use m_pSM?

utils/hct/hctdb.py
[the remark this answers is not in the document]
bob80905: I also don't like this assignment method, the for loop isn't really being used. Personally I would just say b = "Barrier" and sub in b for the name_idx index, so that the for loops can all be removed. Different variables for the different strings too.

PR #6296, opened 2024-02-13
Made the `this` variable in templated struct member functions references

adam-yang:
Made the debug info type for the `this` argument a reference type. `DxilRewriteOutputArgDebugInfo` then turns the reference type into a value type to preserve the value mapping when the alloca that `this` points to becomes registers.


llvm-beanz (approved):
I don't love the way this is all built. Specifically, I wish we had a way to inspect the initial DI code generation, and could test the mutating pass in isolation. That said, I think the change here is correct and it is as well tested as we can manage.

dmpots (approved):
LGTM

PR #6299, opened 2024-02-13
CP: WinPIX: Changes to support bitfields in PIX's shader debugger (#6219)

jeffnn:
Two related sets of changes:
1) Modify the value-to-declare pass to tolerate bitfields, which are detected by their habit of being smaller than the type in which they are contained. The pass produces dbg.declare statements that map each HLSL variable to some storage. With bitfields, that mapping becomes many-to-one, because many bitfields can be stored in the same basic type. Most of the changes are to preserve the size of the bitfield as the code descends the type hierarchy into the underlying basic type, which is what motivates these new "type overrride" parameters to some functions. Complications to watch out for are min16 types, which may or may not be their own purpose-made (16-bit) types depending on whether or not they are enabled.
2) Modify the PIX debug-info API to return enough data to PIX to allow it to extract (by bit shifting) a bitfield from the underlying type. This bit is also about preserving knowledge of the field's size as opposed to its containing type's size, hence the new dxcapi interface/method.

---------

Co-authored-by: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
(cherry picked from commit a23d58663f9055dea1ddcb32ad4539eaccb08d5b)

### DIFF
--- a/lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
+++ b/lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
@@ -51,6 +51,44 @@ using namespace llvm;
 namespace {
 using OffsetInBits = unsigned;
 using SizeInBits = unsigned;
+struct Offsets {
+  OffsetInBits Aligned;
+  OffsetInBits Packed;
+};
+
+// DITypePeelTypeAlias peels const, typedef, and other alias types off of Ty,
+// returning the unalised type.
+static llvm::DIType *DITypePeelTypeAlias(llvm::DIType *Ty) {
+  if (auto *DerivedTy = llvm::dyn_cast<llvm::DIDerivedType>(Ty)) {
+    const llvm::DITypeIdentifierMap EmptyMap;
+    switch (DerivedTy->getTag()) {
+    case llvm::dwarf::DW_TAG_restrict_type:
+    case llvm::dwarf::DW_TAG_reference_type:
+    case llvm::dwarf::DW_TAG_const_type:
+    case llvm::dwarf::DW_TAG_typedef:
+    case llvm::dwarf::DW_TAG_pointer_type:
+      return DITypePeelTypeAlias(DerivedTy->getBaseType().resolve(EmptyMap));
+    case llvm::dwarf::DW_TAG_member:
+      return DITypePeelTypeAlias(DerivedTy->getBaseType().resolve(EmptyMap));
+    }
+  }
+
+  return Ty;
+}
+
+llvm::DIBasicType *BaseTypeIfItIsBasicAndLarger(llvm::DIType *Ty) {
+  // Working around problems with bitfield size/alignment:
+  // For bitfield types, size may be < 32, but the underlying type
+  // will have the size of that basic type, e.g. 32 for ints.
+  // By contrast, for min16float, size will be 16, but align will be 16 or 32
+  // depending on whether or not 16-bit is enabled.
+  // So if we find a disparity in size, we can assume it's not e.g. min16float.
+  auto *baseType = DITypePeelTypeAlias(Ty);
+  if (Ty->getSizeInBits() != 0 &&
+      Ty->getSizeInBits() < baseType->getSizeInBits())
+    return llvm::dyn_cast<llvm::DIBasicType>(baseType);
+  return nullptr;
+}
 
 // OffsetManager is used to map between "packed" and aligned offsets.
 //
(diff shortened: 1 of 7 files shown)


lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -1207,24 +1231,28 @@ static llvm::Type *GetLLVMTypeFromDIBasicType(llvm::IRBuilder<> &B,
...
+  const auto offsets = m_Offsets.Add(Ty, sizeOverride);
 
   llvm::Type *AllocaTy = llvm::ArrayType::get(AllocaElementTy, 1);
-  llvm::AllocaInst *&Alloca = m_AlignedOffsetToAlloca[AlignedOffset];
-  Alloca = m_B.CreateAlloca(AllocaTy, m_B.getInt32(0));
-  Alloca->setDebugLoc(llvm::DebugLoc());
+  llvm::AllocaInst *&Alloca = m_AlignedOffsetToAlloca[offsets.Aligned];
+  if (Alloca == nullptr) {
+    Alloca = m_B.CreateAlloca(AllocaTy, m_B.getInt32(0));
+    Alloca->setDebugLoc(llvm::DebugLoc());
```
pow2clk, line 1248: Is this debug loc going to arbitrarily be assigned to the first member to get here? That is probably okay, but the first in the struct might be best and I don't know if that's guaranteed.

pow2clk (approved):
I recognize that this is a cherry pick and I'm not asking you to address any of my comments here, but since I missed this on the first go around, I thought I'd point out some potential concerns that might be helpful to address in some later followups. 

Though I would welcome responses to my inline comments, please resolve the conversations once you've read them.

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -51,6 +51,44 @@ using namespace llvm;
...
+  return Ty;
+}
+
+llvm::DIBasicType *BaseTypeIfItIsBasicAndLarger(llvm::DIType *Ty) {
+  // Working around problems with bitfield size/alignment:
+  // For bitfield types, size may be < 32, but the underlying type
+  // will have the size of that basic type, e.g. 32 for ints.
+  // By contrast, for min16float, size will be 16, but align will be 16 or 32
+  // depending on whether or not 16-bit is enabled.
+  // So if we find a disparity in size, we can assume it's not e.g. min16float.
```
tex3d, line 85: Shouldn't you be switching expectations for min/low precision types on the basis of the `UseNativeLowPrecision` flag in the module?

tex3d (approved):

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
tex3d: BTW: This is just a comment/question and can be resolved to complete the PR of course.

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -63,28 +101,28 @@ using SizeInBits = unsigned;
...
-          llvm::DIType *baseType = DerivedTy->getBaseType().resolve(EmptyMap);
-          if (baseType != nullptr) {
-            if (baseType->getAlignInBits() == 0) {
-              (void)baseType->getAlignInBits();
+    if (BaseTypeIfItIsBasicAndLarger(Ty))
+      AlignMask = 0;
+    else {
+      auto *DerivedTy = llvm::dyn_cast<llvm::DIDerivedType>(Ty);
+      if (DerivedTy != nullptr) {
+        // Working around a bug where byte size is stored instead of bit size
```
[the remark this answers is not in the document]
jeffnn: Well none of the existing compiler-side unit tests or PIX tests exercise this case (anymore?). I found the original customer pix file that was involved in the fix that added this change, and THAT doesn't exercise this case either. So maybe this is no longer (never was?) an issue...

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -63,28 +101,28 @@ using SizeInBits = unsigned;
...
+          AlignMask = 32;
+        }
+        if (AlignMask == 0) {
+          const llvm::DITypeIdentifierMap EmptyMap;
+          switch (DerivedTy->getTag()) {
+          case llvm::dwarf::DW_TAG_restrict_type:
+          case llvm::dwarf::DW_TAG_reference_type:
+          case llvm::dwarf::DW_TAG_const_type:
+          case llvm::dwarf::DW_TAG_typedef: {
+            llvm::DIType *baseType = DerivedTy->getBaseType().resolve(EmptyMap);
```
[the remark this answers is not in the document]
jeffnn: Noted. It was only these alias-type types that had a zero alignment, but can't hurt to be consistent here.

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
jeffnn: I think this is OK. Multiple bitfields can live at the same aligned offset (that is, the offset of the actual storage). We'd have to choose some DebugLoc to associate the alloca with, so the first seems like as good a choice as any.

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
jeffnn: Actually I think the comment should probably be tweaked. Alignment is no longer part of this decision here. 
But to answer the question, min16float (etc.) always have the same size as their basic type (cuz there are explicit types added just for them) so they can be treated like any other derived type that is the same size as its basic-type storage. So they're not really special here at all.

PR #6300, opened 2024-02-13
CP: PIX: Optimize debug instrumentation for fewest emitted instructions (…

jeffnn:
…#6281)

This change switches from instrumentation per instruction to instrumentation per basic block. Furthermore, not every instruction in a basic block needs to write debug data to the output UAV- it's enough to know that the basic block was entered, as long as the calling application can figure out what instructions were in that block. To support that knowledge, the pass now emits a text "precis" of each basic block.

Also, the previous branchless UAV bounds enforcement was replaced with something similar that emits fewer instructions at the cost of a larger UAV. This tradeoff is WELL worth it.

Additionally, the debug pass used to add extra blocks in order to solidify the arguments to phi instructions. This work was unnecessary, and added a lot of complexity to the resulting instrumented shader. The debugger application is only interested in the value of the phi itself and the actual value produced via the actual preceding edge.
 
Full details are in the comments in the code.

This change reduces driver-side compilation overhead from "overnight" to 2 minutes on a 160k-instruction shader.

(cherry picked from commit 446da195b4f37a8aaa20fcd0d8c7202133cd93f6)


lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -850,99 +986,199 @@ void DxilDebugInstrumentation::addStepEntryForType(
...
+    if (Inst->getOpcode() == Instruction::Ret) {
+      if (BC != nullptr)
+        addStepEntryForType<void>(DebugShaderModifierRecordTypeDXILStepRet, *BC,
+                                  InstNum, nullptr, 0, 0);
+      InstructionAndType ret{};
+      ret.Inst = Inst;
+      ret.InstructionOrdinal = InstNum;
+      ret.Type = DebugShaderModifierRecordTypeDXILStepRet;
+      return ret;
+    } else if (Inst->isTerminator()) {
```
pow2clk, line 1104: It's a very late nitpick, but just as an FYI, Instruction::Ret is a terminator and will be true here as well. You might have checked terminator first and included all of the common code and only checked for the ret opcode for the last type assignment.

tools/clang/test/HLSLFileCheck/pix/DebugVSParameters.hlsl
```diff
@@ -8,7 +8,6 @@
 // CHECK: %CompareToVertId = icmp eq i32 %VertId, 1
 // CHECK: %CompareToInstanceId = icmp eq i32 %InstanceId, 2
 // CHECK: %CompareBoth = and i1 %CompareToVertId, %CompareToInstanceId
-// CHECK: %OffsetMultiplicand = zext i1 %CompareBoth to i32
```
pow2clk, line 11: I'm curious why test cases such as this and the deleted files were removed.

pow2clk (approved):
A few FYIs and one question out of interest

tools/clang/test/HLSLFileCheck/pix/DebugVSParameters.hlsl
jeffnn: The cases with "phi" in their names were deleted because the pass itself no longer does anything special with PHIs. (It used to add new blocks and edges for all incoming values, for no really good reason, and the extra complexity of the instrumented shader was inducing yet more run-time overhead in the GPU driver.)

This particular line isn't part of the new scheme anymore, and the point of this test file is to validate the checking of the vertex-selection parameters anyway, not the actual per-block debug instrumentation.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -957,11 +1193,16 @@ void DxilDebugInstrumentation::addStepDebugEntryValue(
   case Type::TypeID::PPC_FP128TyID:
     assert(false);
```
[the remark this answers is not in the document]
jeffnn: I filed this: https://github.com/microsoft/DirectXShaderCompiler/issues/6307
#6307

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -981,10 +1222,159 @@ bool DxilDebugInstrumentation::runOnModule(Module &M) {
...
+      count += (*datum)->PayloadSize;
+  }
+  return count;
+}
+
+const char *TypeString(InstructionAndType const &IandT) {
+  auto datum = FindDatum(IandT.Type);
+  if (datum)
+    return (*datum)->AsString;
+  assert(false);
```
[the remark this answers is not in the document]
jeffnn: Mentioned in the issue.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -1061,100 +1424,81 @@ bool DxilDebugInstrumentation::RunOnFunction(Module &M, DxilModule &DM,
...
+                          BlockInstrumentation.FirstInstructionOrdinalInBlock));
+      for (auto &Inst : BlockInstrumentation.Instructions) {
+        Instruction *BuilderInstruction;
+        if (Inst.InstructionAfterWhichToAddInstrumentation != nullptr)
+          BuilderInstruction =
+              Inst.InstructionAfterWhichToAddInstrumentation->getNextNode();
+        else if (Inst.InstructionBeforeWhichToAddInstrumentation != nullptr)
+          BuilderInstruction = Inst.InstructionBeforeWhichToAddInstrumentation;
+        else {
+          assert(false);
```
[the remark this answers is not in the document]
jeffnn: Mentioned in the issue.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -850,99 +986,199 @@ void DxilDebugInstrumentation::addStepEntryForType(
...
   const Type::TypeID ID = V->getType()->getTypeID();
 
   switch (ID) {
   case Type::TypeID::StructTyID:
   case Type::TypeID::VoidTyID:
-    addStepEntryForType<void>(DebugShaderModifierRecordTypeDXILStepVoid, BC,
-                              InstNum, V, ValueOrdinal, ValueOrdinalIndex);
-    break;
+    if (BC != nullptr)
+      addStepEntryForType<void>(DebugShaderModifierRecordTypeDXILStepVoid, *BC,
```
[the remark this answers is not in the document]
jeffnn: The original thought was that addStepEntryForType takes a reference and doesn't have to worry about null, but I think you're right and that putting the check there would be cleaner in total. I field #6308 to address this.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
jeffnn: Thanks. Yes, this could be telescoped a teeny bit.

jeffnn: Made a note to fix this in another checkin.

austinkinross (approved):

PR #6302, opened 2024-02-13
Fix crash in DXIL.dll caused by illegal DXIL intrinsic.

### LINKED ISSUE #6168: DXIL.dll crashes on intrinsics that are not defined for a known shader model
**Description**
Given DXIL that declares itself as shader model 6.7 / DXIL 1.7, but uses intrinsic functions defined by shader model 6.8, attempting to validate that DXIL module will crash with an access violation.

**Steps to Reproduce**
See the relevant code in `dxil::OP`:

```c++
void OP::RefreshCache() {
  for (Function &F : m_pModule->functions()) {
    if (OP::IsDxilOpFunc(&F) && !F.user_empty()) {
      CallInst *CI = cast<CallInst>(*F.user_begin());
      OpCode OpCode = OP::GetDxilOpFuncCallInst(CI);
      Type *pOverloadType = OP::GetOverloadType(OpCode, &F);
      Function *OpFunc = GetOpFunc(OpCode, pOverloadType);
      (void)(OpFunc);
      DXASSERT_NOMSG(OpFunc == &F);
    }
  }
}
```

For every intrinsic function, `GetOpFunc` is called to instantiate it in the cache. Inside `GetOpFunc`:
```c++
Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
  DXASSERT(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes,
           "otherwise caller passed OOB OpCode");
  assert(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes);
  DXASSERT(IsOverloadLegal(opCode, pOverloadType),
           "otherwise the caller requested illegal operation overload (eg HLSL "
           "function with unsupported types for mapped intrinsic function)");
  OpCodeClass opClass = m_OpCodeProps[(unsigned)opCode].opCodeClass;
  Function *&F =
      m_OpCodeClassCache[(unsigned)opClass].pOverloads[pOverloadType];
```

Those asserts for "otherwise caller..." should be validation, because the caller clearly didn't do any validation. As a result, indexing into `m_OpCodeProps` can use an out-of-bounds index, meaning that accessing `opCodeClass` could be an access violation directly, or at best could return garbage. I was seeing it return a pointer-type value, which was then used to index into `m_OpCodeClassCache`, which failed catastrophically.

My repro was done by accidentally inserting SampleCmpBias instructions into a SM6.7 shader in the Mesa DXIL backend. DXC will not emit such a shader, since it has validation against doing so earlier on in the compilation process.

python3kgae:
Replace assert on illegal DXIL op with return illegal value. Check the illegal cases in validation.

Fixes #6168

### DIFF
--- a/lib/DXIL/DxilOperations.cpp
+++ b/lib/DXIL/DxilOperations.cpp
@@ -2737,8 +2737,8 @@ llvm::Attribute::AttrKind OP::GetMemAccessAttr(OpCode opCode) {
 }
 
 bool OP::IsOverloadLegal(OpCode opCode, Type *pType) {
-  DXASSERT(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes,
-           "otherwise caller passed OOB index");
+  if (opCode >= OpCode::NumOpCodes)
+    return false;
   unsigned TypeSlot = GetTypeSlot(pType);
   return TypeSlot != UINT_MAX &&
          m_OpCodeProps[(unsigned)opCode].bAllowOverload[TypeSlot];
@@ -2814,8 +2814,10 @@ bool OP::IsDxilOpFuncCallInst(const llvm::Instruction *I, OpCode opcode) {
 }
 
 OP::OpCode OP::getOpCode(const llvm::Instruction *I) {
-  return (OP::OpCode)llvm::cast<llvm::ConstantInt>(I->getOperand(0))
-      ->getZExtValue();
+  auto *OpConst = llvm::dyn_cast<llvm::ConstantInt>(I->getOperand(0));
+  if (!OpConst)
+    return OpCode::NumOpCodes;
+  return (OP::OpCode)OpConst->getZExtValue();
 }
 
 OP::OpCode OP::GetDxilOpFuncCallInst(const llvm::Instruction *I) {
@@ -3383,9 +3385,7 @@ void OP::RefreshCache() {
       CallInst *CI = cast<CallInst>(*F.user_begin());
       OpCode OpCode = OP::GetDxilOpFuncCallInst(CI);
       Type *pOverloadType = OP::GetOverloadType(OpCode, &F);
-      Function *OpFunc = GetOpFunc(OpCode, pOverloadType);
-      (void)(OpFunc);
-      DXASSERT_NOMSG(OpFunc == &F);
+      GetOpFunc(OpCode, pOverloadType);
     }
   }
 }
@@ -3404,11 +3404,13 @@ void OP::FixOverloadNames() {
       CallInst *CI = cast<CallInst>(*F.user_begin());
       DXIL::OpCode opCode = OP::GetDxilOpFuncCallInst(CI);
       llvm::Type *Ty = OP::GetOverloadType(opCode, &F);
-      if (isa<StructType>(Ty) || isa<PointerType>(Ty)) {
-        std::string funcName;
-        if (OP::ConstructOverloadName(Ty, opCode, funcName)
-                .compare(F.getName()) != 0) {
-          F.setName(funcName);
+      if (OP::IsOverloadLegal(opCode, Ty)) {
+        if (isa<StructType>(Ty) || isa<PointerType>(Ty)) {
+          std::string funcName;
+          if (OP::ConstructOverloadName(Ty, opCode, funcName)
+                  .compare(F.getName()) != 0) {
+            F.setName(funcName);
+          }
         }
       }
     }
@@ -3421,12 +3423,11 @@ void OP::UpdateCache(OpCodeClass opClass, Type *Ty, llvm::Function *F) {
 }
 
 Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
-  DXASSERT(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes,
-           "otherwise caller passed OOB OpCode");
-  assert(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes);
-  DXASSERT(IsOverloadLegal(opCode, pOverloadType),
-           "otherwise the caller requested illegal operation overload (eg HLSL "
-           "function with unsupported types for mapped intrinsic function)");
+  if (opCode >= OpCode::NumOpCodes)
+    return nullptr;
+  if (!IsOverloadLegal(opCode, pOverloadType))
+    return nullptr;
+
   OpCodeClass opClass = m_OpCodeProps[(unsigned)opCode].opCodeClass;
   Function *&F =
       m_OpCodeClassCache[(unsigned)opClass].pOverloads[pOverloadType];
@@ -5369,6 +5370,8 @@ Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
   // and return values to ensure that ResRetType is constructed in the
   // RefreshCache case.
   if (Function *existF = m_pModule->getFunction(funcName)) {
+    if (existF->getFunctionType() != pFT)
+      return nullptr;
     DXASSERT(existF->getFunctionType() == pFT,
              "existing function must have the expected function type");
     F = existF;
@@ -5489,7 +5492,8 @@ llvm::Type *OP::GetOverloadType(OpCode opCode, llvm::Function *F) {
   case OpCode::CallShader:
   case OpCode::Pack4x8:
   case OpCode::WaveMatrix_Fill:
-    DXASSERT_NOMSG(FT->getNumParams() > 2);
+    if (FT->getNumParams() <= 2)
+      return Ty;
     return FT->getParamType(2);
   case OpCode::MinPrecXRegStore:
   case OpCode::StoreOutput:
@@ -5499,7 +5503,8 @@ llvm::Type *OP::GetOverloadType(OpCode opCode, llvm::Function *F) {
   case OpCode::StoreVertexOutput:
   case OpCode::StorePrimitiveOutput:
   case OpCode::DispatchMesh:
-    DXASSERT_NOMSG(FT->getNumParams() > 4);
+    if (FT->getNumParams() <= 4)
+      return Ty;
     return FT->getParamType(4);
   case OpCode::IsNaN:
   case OpCode::IsInf:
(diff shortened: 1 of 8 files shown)


lib/DXIL/DxilOperations.cpp
```diff
@@ -2814,8 +2814,10 @@ bool OP::IsDxilOpFuncCallInst(const llvm::Instruction *I, OpCode opcode) {
 }
 
 OP::OpCode OP::getOpCode(const llvm::Instruction *I) {
-  return (OP::OpCode)llvm::cast<llvm::ConstantInt>(I->getOperand(0))
-      ->getZExtValue();
+  auto *OpConst = llvm::dyn_cast<llvm::ConstantInt>(I->getOperand(0));
+  if (!OpConst)
+    return OpCode::NumOpCodes;
+  return (OP::OpCode)OpConst->getZExtValue();
```
llvm-beanz, line 2820:
```suggestion
  uint64_t OpCodeVal = OpConst->getZExtValue();
  if (OpCodeVal >= NumOpCodes)
    return NumOpCodes;
  return static_cast<OP::OpCode>(OpCodeVal);
```
Casting an integer value to an enum that is out-of-range is undefined behavior. This approach ensures that the resulting `Op::OpCode` is valid and in range for the enumeration.

lib/DXIL/DxilOperations.cpp
```diff
@@ -2737,8 +2737,8 @@ llvm::Attribute::AttrKind OP::GetMemAccessAttr(OpCode opCode) {
 }
 
 bool OP::IsOverloadLegal(OpCode opCode, Type *pType) {
-  DXASSERT(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes,
-           "otherwise caller passed OOB index");
+  if (opCode >= OpCode::NumOpCodes)
```
llvm-beanz, line 2740:
```suggestion
  if (opCode == OpCode::NumOpCodes)
```

It is undefined behavior for a value of enumeration type to be out of range. (see my next suggestion for more on how to fix this case).

lib/DXIL/DxilOperations.cpp
```diff
@@ -3404,11 +3404,13 @@ void OP::FixOverloadNames() {
       CallInst *CI = cast<CallInst>(*F.user_begin());
       DXIL::OpCode opCode = OP::GetDxilOpFuncCallInst(CI);
       llvm::Type *Ty = OP::GetOverloadType(opCode, &F);
-      if (isa<StructType>(Ty) || isa<PointerType>(Ty)) {
-        std::string funcName;
-        if (OP::ConstructOverloadName(Ty, opCode, funcName)
-                .compare(F.getName()) != 0) {
-          F.setName(funcName);
+      if (OP::IsOverloadLegal(opCode, Ty)) {
```
llvm-beanz, line 3407: nit: We can eliminate at least one or two levels of indentation using `continue` expressions.

(see: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code)

lib/DXIL/DxilOperations.cpp
```diff
@@ -5369,6 +5370,8 @@ Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
   // and return values to ensure that ResRetType is constructed in the
   // RefreshCache case.
   if (Function *existF = m_pModule->getFunction(funcName)) {
+    if (existF->getFunctionType() != pFT)
+      return nullptr;
     DXASSERT(existF->getFunctionType() == pFT,
```
llvm-beanz, line 5375: This assert seems unnecessary now.

lib/DXIL/DxilOperations.cpp
```diff
@@ -3421,12 +3423,11 @@ void OP::UpdateCache(OpCodeClass opClass, Type *Ty, llvm::Function *F) {
 }
 
 Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
-  DXASSERT(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes,
-           "otherwise caller passed OOB OpCode");
-  assert(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes);
```
llvm-beanz, line 3426: There are a few other places in this file where we replicate this assert (which itself is duplicated). Can we remove those cases? These asserts seem silly since they're basically asserting "we don't have undefined behavior".

lib/DXIL/DxilOperations.cpp
```diff
@@ -5489,7 +5492,8 @@ llvm::Type *OP::GetOverloadType(OpCode opCode, llvm::Function *F) {
   case OpCode::CallShader:
   case OpCode::Pack4x8:
   case OpCode::WaveMatrix_Fill:
-    DXASSERT_NOMSG(FT->getNumParams() > 2);
+    if (FT->getNumParams() <= 2)
+      return Ty;
```
llvm-beanz, line 5496: Should these return a value, or should they return an error of some sort (nullptr)?

python3kgae: Changed to return nullptr.

lib/DXIL/DxilOperations.cpp
python3kgae: Removed.

lib/DXIL/DxilOperations.cpp
python3kgae: Removed.

lib/DXIL/DxilOperations.cpp
python3kgae: Updated.

lib/DXIL/DxilOperations.cpp
python3kgae: Updated.

lib/DXIL/DxilOperations.cpp
python3kgae: Updated.

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -572,13 +572,9 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
           continue;
-        Value *opcodeArg = CI->getArgOperand(DXIL::OperandIndex::kOpcodeIdx);
-        ConstantInt *opcodeConst = dyn_cast<ConstantInt>(opcodeArg);
-        DXASSERT(opcodeConst, "DXIL opcode arg must be immediate");
-        unsigned opcode = opcodeConst->getLimitedValue();
-        DXASSERT(opcode < static_cast<unsigned>(DXIL::OpCode::NumOpCodes),
-                 "invalid DXIL opcode");
-        DXIL::OpCode dxilOp = static_cast<DXIL::OpCode>(opcode);
+        DXIL::OpCode dxilOp = hlsl::OP::getOpCode(CI);
+        if (dxilOp >= DXIL::OpCode::NumOpCodes)
```
llvm-beanz, line 576: nit:
```suggestion
        if (dxilOp == DXIL::OpCode::NumOpCodes)
```

lib/HLSL/DxilValidation.cpp
```diff
@@ -3239,6 +3241,29 @@ static void ValidateFunctionBody(Function *F, ValidationContext &ValCtx) {
           unsigned opcode = OpcodeConst->getLimitedValue();
           DXIL::OpCode dxilOpcode = (DXIL::OpCode)opcode;
```
llvm-beanz, line 3242: This is undefined behavior if `opcode > NumOpCodes`. We should handle that case on the `unsigned` value before casting.

llvm-beanz (commented):
One thing I missed on my last pass, and a cleanup nit.

lib/DXIL/DxilShaderFlags.cpp
python3kgae: Fixed.

lib/HLSL/DxilValidation.cpp
python3kgae: Updated to unsigned compare.

lib/DXIL/DxilOperations.cpp
```diff
@@ -2719,26 +2717,22 @@ const char *OP::GetAtomicOpName(DXIL::AtomicBinOpCode OpCode) {
...
 
 const char *OP::GetOpCodeClassName(OpCode opCode) {
-  DXASSERT(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes,
-           "otherwise caller passed OOB index");
   return m_OpCodeProps[(unsigned)opCode].pOpCodeClassName;
 }
 
 llvm::Attribute::AttrKind OP::GetMemAccessAttr(OpCode opCode) {
-  DXASSERT(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes,
-           "otherwise caller passed OOB index");
```
pow2clk, line 2735: Given the lack of a robust response to an invalid `opCode` like what `IsOverloadLegal` has, I don't know why removing these asserts makes sense. Perhaps it was in response to this comment? https://github.com/microsoft/DirectXShaderCompiler/pull/6302/files#r1489723490 That was in a case where an OOB opcode was handled well and there were two asserts at least one of which was clearly redundant. That's not the case here. If I'm developing the compiler and I send an invalid opcode to one of these functions, I'd rather see an assert than have an access violation that may crash immediately in the best case or just corrupt things int eh worst case.

lib/DXIL/DxilOperations.cpp
pow2clk: I'm not sure if this is what Chris meant. In this location we have `DXASSERT` and `assert`, which are entirely redundant. In this case, the robust responses below make both asserts less useful, but in many cases where asserts got removed entirely including those I commented on above, that is not the case.

lib/DXIL/DxilOperations.cpp
```diff
@@ -5389,9 +5387,6 @@ Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
 
 const SmallMapVector<llvm::Type *, llvm::Function *, 8> &
 OP::GetOpFuncList(OpCode opCode) const {
-  DXASSERT(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes,
-           "otherwise caller passed OOB OpCode");
-  assert(0 <= (unsigned)opCode && opCode < OpCode::NumOpCodes);
```
pow2clk, line 5394: Another complete removal of asserts that I don't think is needed.

lib/HLSL/DxilValidation.cpp
```diff
@@ -3237,8 +3239,30 @@ static void ValidateFunctionBody(Function *F, ValidationContext &ValCtx) {
           }
 
           unsigned opcode = OpcodeConst->getLimitedValue();
+          if (opcode >= static_cast<unsigned>(DXIL::OpCode::NumOpCodes)) {
```
pow2clk, line 3242: Is this right or should we verify it doesn't exceed the number of opcodes for the current version? We have some custom checks about such things and if doing that silences them, then we probably shouldn't, but it might be kind of nice to have a broad catchall that saves us the trouble of making those individual checks.

pow2clk (commented):
Mostly speaking up for asserts

lib/HLSL/DxilValidation.cpp
llvm-beanz: We need to verify that the value is in-range before we can cast it to the `DXIL::OpCode` type because casting an out-of-range value is undefined behavior.

llvm-beanz (approved):

lib/HLSL/DxilValidation.cpp
pow2clk: I'm realizing now that we can't actually use the versioned numopcodes to indicate version support anyway because we consciously decided to allow some intermingling there.

lib/DXIL/DxilOperations.cpp
pow2clk: It being an enum satisfies me

pow2clk (approved):
Looks good given Chris's responses to my comments.

PR #6309, opened 2024-02-14
PIX: Preclude debug output of ray query handle from phi or select (since the value isn't really an int, it's a handle)

jeffnn:
AllocateRayQuery returns an int, but it's not really an int: it's a handle to the query. Thus, it's not sensible for the PIX debugging instrumentation to attempt to write it out to the debug-data-UAV. The previous code did an explicit check against the type of the value to be written, but if that value is actually a phi itself, then that check would fail. So now we recursively run through the phi values looking to see if any of its antecedents are a phi, and if so, refuse to send its value to the UAV.


tools/clang/test/HLSLFileCheck/pix/TraceRayInlineDebug.hlsl
```diff
@@ -0,0 +1,30 @@
+// RUN: %dxc -T vs_6_5 -Od -E main %s | %opt -S -dxil-annotate-with-virtual-regs -hlsl-dxil-debug-instrumentation | FileCheck %s
+ 
+// CHECK: [[RAYQUERY0:%.*]] = call i32 @dx.op.allocateRayQuery
+// CHECK: [[RAYQUERY1:%.*]] = call i32 @dx.op.allocateRayQuery
+
+// CHECK: [[PHIQUERY:%.*]] = phi i32 [ [[RAYQUERY0:%.*]], {{.*}} ], [ [[RAYQUERY0:%.*]], {{.*}} ]
+
+// The debug instrumentation should NOT try to store this phi value: it's not an i32! (It's an opaque handle).
```
tex3d, line 8: If the DXIL validator let you do this, I think validation is deficient for RayQuery handle.  I'm not surprised, but an issue should be filed about that.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -41,8 +41,16 @@ bool IsAllocateRayQueryInstruction(llvm::Value const *Val) {
   if (Val != nullptr) {
     if (llvm::Instruction const *Inst =
             llvm::dyn_cast<llvm::Instruction>(Val)) {
-      return hlsl::OP::IsDxilOpFuncCallInst(Inst,
-                                            hlsl::OP::OpCode::AllocateRayQuery);
+      if (hlsl::OP::IsDxilOpFuncCallInst(Inst,
+                                         hlsl::OP::OpCode::AllocateRayQuery))
+        return true;
+      if (auto PHI = llvm::dyn_cast<llvm::PHINode>(Inst)) {
```
tex3d, line 47: Incoming value for PHI might be another PHI, so I think this might need to be recursive?
It may be possible to see a select instead of PHI as well.

So, I think a robust solution requires a little more work.

Also, the name `IsAllocateRayQueryInstruction` seems like it should be something more like `IsRayQueryHandle`.

tex3d: For testing, you might need an IR test to get a select scenario if it's hard to produce in HLSL.

jeffnn: Woah, didn't know select existed! Thanks!
Also, it seems like phis can recursively refer to each other, which breaks my brain.

tools/clang/test/HLSLFileCheck/pix/TraceRayInlineDebug.hlsl
jeffnn: #6312

adam-yang: In DXIL it is just an i32 value though, is that something that validator should/could check?

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: Addressed the recursive phi issue and added select checking. switch() with -O3 exercises it.

tools/clang/test/HLSLFileCheck/pix/TraceRayInlineDebug.hlsl
jeffnn: That's what I filed #6312 for...

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -37,17 +37,49 @@ using namespace llvm;
...
-      return hlsl::OP::IsDxilOpFuncCallInst(Inst,
-                                            hlsl::OP::OpCode::AllocateRayQuery);
+      if (hlsl::OP::IsDxilOpFuncCallInst(Inst,
+                                         hlsl::OP::OpCode::AllocateRayQuery)) {
+        return true;
+      } else if (auto PHI = llvm::dyn_cast<llvm::PHINode>(Inst)) {
+        for (unsigned IVOrdinal = 0; IVOrdinal < PHI->getNumIncomingValues();
+             ++IVOrdinal) {
+          // phi instructions can recursively refer to each other
+          if (std::find(PreviousPhis.begin(), PreviousPhis.end(), Val) ==
```
jeffnn, line 53: This is really gross, so I hope somebody has a better idea about how to cope with circular phi references...

tex3d: Can't you keep a set of visited instructions and pass a reference to that set to IsRayQueryHandleRecursive?  Skip the incoming value if it's in the set.

tex3d: You could use a `llvm::SmallPtrSet`, and you could just try adding it when you have a PHI, returning false if it was already in the set.  Like:
```
if (!VisitedPhis.insert(PHI).second)
  return false;
```

jeffnn: Maybe I missed your point, but that seems equivalent to what I have? And if I don't clean up the set as I go back up the stack, the set could grow to many thousands (not really a "small" pointer set) and the lookup could bog down.

tex3d: Another approach is to just make a map of PHI instructions that are RayQuery handles by following the AllocateRayQuery users.  I think this might be a lot better approach.  This should end up being a small map or empty map most of the time, and you just look up the PHI in the map to see if it is one.

jeffnn: Good idea. Worked well.

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -319,6 +320,7 @@ class DxilDbgValueToDbgDeclare : public llvm::ModulePass {
 
   std::unordered_map<llvm::DIVariable *, std::unique_ptr<VariableRegisters>>
       m_Registers;
+  std::vector<llvm::Value const *> m_PhiDescentHistory;
```
tex3d, line 323: What's this used for?

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -37,15 +37,33 @@ using namespace llvm;
...
+    iterator_range<Value::user_iterator> Users = Use->users();
+    for (User *User : Users) {
+      if (std::find(Handles.begin(), Handles.end(), User) == Handles.end())
+        FindRayQueryHandlesFromUse(User, Handles);
     }
   }
-  return false;
+}
+
+std::vector<Value *> FindRayQueryHandlesForFunction(llvm::Function *F) {
```
tex3d, line 53: Why use a `std::vector`?  Couldn't this be a `llvm::SmallPtrSet` instead?  This should be fast when small and scale well too.
Recommend using argument of type `SmallPtrSetImpl<Value*> &`, and using like so:
```cpp
llvm::SmallPtrSet<Value*, 16> RayQueryHandles;
FindRayQueryHandlesForFunction(F, RayQueryHandles);
```
That way, calling code determines the "small" size, and this function isn't locked to a particular "small" size.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -37,15 +37,33 @@ using namespace llvm;
...
-                                            hlsl::OP::OpCode::AllocateRayQuery);
+static void FindRayQueryHandlesFromUse(Value *U,
+                                       std::vector<Value *> &Handles) {
+  Handles.push_back(U);
+  auto RayQueryHandleUses = U->uses();
+  for (Use &Use : RayQueryHandleUses) {
+    iterator_range<Value::user_iterator> Users = Use->users();
+    for (User *User : Users) {
+      if (std::find(Handles.begin(), Handles.end(), User) == Handles.end())
+        FindRayQueryHandlesFromUse(User, Handles);
```
tex3d, line 48: Not every user of anything that accepts a RayQuery handle on input is itself a ray query handle!  This should be careful only to follow users of specific instructions: PHI and Select for now.  No other instructions should have their users followed recursively.

tex3d (changes requested):

lib/DxilPIXPasses/PixPassHelpers.cpp
tex3d: Another approach would be a helper object that handles the details for you, like a `RayQueryHandleHelper` class.  It would manage the SmallPtrSet and have an accessor for whether some value is a RayQuery handle.  That way specific details like this don't get proliferated though various parameter types and such.

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
jeffnn: Oops: left over from previous.

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: Right-o

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: D'Oh! Of course... Thanks.

adam-yang (approved):

adam-yang:
Do we need the SPIR-V submodule updates?

jeffnn:
> Do we need the SPIR-V submodule updates?

Ach! That snuck in. Undone...

tex3d (approved):
LGTM!

PR #6310, opened 2024-02-14
[lit] Add dxil version to available_features for lit

python3kgae:
This will allow lit test check dxil version to avoid running on unsupported dxil.dll.

### DIFF
--- a/tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/node-objects-metdata.hlsl
+++ b/tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/node-objects-metdata.hlsl
@@ -1,8 +1,22 @@
-// RUN: %dxc -T lib_6_8 %s | FileCheck %s --check-prefixes=MD
-// RUN: %dxc -T lib_6_8 -Od %s | FileCheck %s --check-prefixes=MD
-// RUN: %dxc -T lib_6_8 -Zi %s | FileCheck %s --check-prefixes=MD
+// RUN: %if dxil-1-8  \
+// RUN:   %{          \
+// RUN: %dxc -T lib_6_8 %s | FileCheck %s --check-prefixes=MD \
+// RUN:   %}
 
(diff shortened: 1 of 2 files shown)


adam-yang (approved):

pow2clk (commented):
FWIW, I looked over this too and it's a great addition!

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/node-objects-metdata.hlsl
```diff
@@ -1,8 +1,22 @@
-// RUN: %dxc -T lib_6_8 %s | FileCheck %s --check-prefixes=MD
-// RUN: %dxc -T lib_6_8 -Od %s | FileCheck %s --check-prefixes=MD
-// RUN: %dxc -T lib_6_8 -Zi %s | FileCheck %s --check-prefixes=MD
-
-// RUN: %dxc -T lib_6_8 -fcgl %s | FileCheck %s --check-prefix=FCGLMD
+// RUN: %if dxil-1-8  \
+// RUN:   %{          \
+// RUN: %dxc -T lib_6_8 %s | FileCheck %s --check-prefixes=MD \
+// RUN:   %}
```
tex3d, line 4: Unfortunately, this is really ugly.

Since this is implied for our FileCheckerTest based on -T target passed to a %dxc RUN part, it means a *ton* of RUN lines in existing tests would have to be translated to these expressions in order to convert them into shell tests.  Is there any way to perform custom feature checking based on reading the RUN line in the lit path?

python3kgae: If all RUN lines share same dxil version, you can just add ; REQUIRES: dxil-1-8 to the top of the test.
The 'if' is for case where each single RUN line needs different dxil version.
Just use it as a reference in the future.

PR #6311, opened 2024-02-14
DerivativesTest fix

hekota:


pow2clk (approved):
I wrote a lot of comments that I ended up deleting as I reacquainted myself with the code and remembered how this was laid out. In the end I have a suggestion for an additional dimension to check and a potential simplification that might make things clearer or possibly not. I think this is easier to understand than it was, but it's still kind of tricky to grasp.

tex3d (approved):

PR #6313, opened 2024-02-15
CP: PIX: Deduplicate globals when referenced in multiple library fns (#6305)

jeffnn:
PIX's code for parsing debug data operates at the module level. When the same global is referenced by multiple functions in a module, that variable is referred to by multiple dbg.value/dbg.declare statements, and those are mapped (by the PIX passes) to multiple fake allocas using its usual scheme. This code was written before libraries were a thing, and wasn't expecting this duplication. A little more attention to the variable's scope fixes the issue.
Also, the changed code's original "return false" broke the whole process of discovering variables with the results that PIX's shader debugger locals window was completely empty. Makes more sense to ignore the one variable and keep going.

(cherry picked from commit 64cdb9cfc07ce37a8156c219c04f2b00f5beb547)

### DIFF
--- a/lib/DxilDia/DxcPixLiveVariables.cpp
+++ b/lib/DxilDia/DxcPixLiveVariables.cpp
@@ -17,6 +17,7 @@
 #include "DxcPixLiveVariables_FragmentIterator.h"
 #include "DxilDiaSession.h"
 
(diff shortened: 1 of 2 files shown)


python3kgae (approved):

pow2clk (approved):
Looks good!

PR #6316, opened 2024-02-16
NFC: Fix DXIL op is_const for Barrier and node create handle ops (#6280)

tex3d:
DXIL operations should use is_const=True for constant arguments. This allows for convenience methods to retrieve the constant value, and could (should, but currently doesn't) result in validation that the argument is constant.

`BarrierByNodeRecordHandle` SemanticFlags argument must be constant. `MetadataIdx` for both `createNodeOutputHandle` and `CreateNodeInputRecordHandle` must be constant.

(cherry picked from commit df588beb485958a3ece128bfbeda5fb888e2dfce)


tex3d:
Requesting this NFC cherry-pick because other changes intended to rely on it will be incoming.

pow2clk (approved):
Still looks good!

bob80905 (approved):

PR #6317, opened 2024-02-16
Fixed PdbUtils not actually normalizing the file paths

adam-yang:
There was a typo in dxcpdbutils where the normalized file path wasn't actually being used. This change fixes that and adds thorough testing for the DxcPdbUtils path normalization. Also fixed another case where `/dir/my_file` gets normalized to `.\\dir\my_file`.

### DIFF
--- a/include/dxc/Support/Path.h
+++ b/include/dxc/Support/Path.h
@@ -64,6 +64,24 @@ inline bool IsAbsoluteOrCurDirRelative(const char *Path) {
   return IsAbsoluteOrCurDirRelativeImpl<char>(Path, strlen(Path));
 }
 
+template <typename CharT, typename StringTy>
+void RemoveDoubleSlashes(StringTy &Path, CharT Slash) {
+  // Remove double slashes.
+  bool SeenNonSlash = false;
+  for (unsigned i = 0; i < Path.size();) {
+    // Remove this slash if:
+    // 1. It is preceded by another slash.
+    // 2. It is NOT part of a series of leading slashes. (E.G. \\, which on
+    // windows is a network path).
+    if (Path[i] == Slash && i > 0 && Path[i - 1] == Slash && SeenNonSlash) {
+      Path.erase(Path.begin() + i);
+      continue;
+    }
+    SeenNonSlash |= Path[i] != Slash;
+    i++;
+  }
+}
+
 // This is the new ground truth of how paths are normalized. There had been
 // many inconsistent path normalization littered all over the code base.
 // 1. All slashes are changed to system native: `\` for windows and `/` for all
@@ -96,28 +114,16 @@ StringTy NormalizePathImpl(const CharT *Path, size_t Length) {
       PathCopy[i] = SlashTo;
   }
 
-  // Remove double slashes.
-  bool SeenNonSlash = false;
-  for (unsigned i = 0; i < PathCopy.size();) {
-    // Remove this slash if:
-    // 1. It is preceded by another slash.
-    // 2. It is NOT part of a series of leading slashes. (E.G. \\, which on
-    // windows is a network path).
-    if (PathCopy[i] == SlashTo && i > 0 && PathCopy[i - 1] == SlashTo &&
-        SeenNonSlash) {
-      PathCopy.erase(PathCopy.begin() + i);
-      continue;
-    }
-    SeenNonSlash |= PathCopy[i] != SlashTo;
-    i++;
-  }
+  RemoveDoubleSlashes<CharT, StringTy>(PathCopy, SlashTo);
 
   // If relative path, prefix with dot.
   if (IsAbsoluteOrCurDirRelativeImpl<CharT>(PathCopy.c_str(),
                                             PathCopy.size())) {
     return PathCopy;
   } else {
-    return StringTy(1, CharT('.')) + StringTy(1, SlashTo) + PathCopy;
+    PathCopy = StringTy(1, CharT('.')) + StringTy(1, SlashTo) + PathCopy;
+    RemoveDoubleSlashes<CharT, StringTy>(PathCopy, SlashTo);
+    return PathCopy;
   }
 }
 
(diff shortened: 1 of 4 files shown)


python3kgae (approved):

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2884,50 +2885,207 @@ TEST_F(CompilerTest, CompileWhenIncludeThenLoadUsed) {
...
+
+  TestCase tests[] = {
+      {R"(main.hlsl)", R"(include.h)"},
+      {R"(.\5Cmain.hlsl)", R"(.\5Cinclude.h)"},
+      {R"(/path/main.hlsl)", R"(/path/include.h)"},
+      {R"(..\5Cmain.hlsl)", R"(..\5Cinclude.h)"},
+      {R"(..\5Cdir\5Cmain.hlsl)", R"(..\5Cdir\5Cinclude.h)"},
+      {R"(F:\5C\5Cdir\5Cmain.hlsl)", R"(F:\5C\5Cdir\5Cinclude.h)"},
+      {R"(\5C\5Cdir\5Cmain.hlsl)", R"(\5C\5Cdir\5Cinclude.h)"},
+      {R"(\5C\5C\5Cdir\5Cmain.hlsl)", R"(\5C\5C\5Cdir\5Cinclude.h)"},
```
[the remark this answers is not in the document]
adam-yang: Double leading forward slash? Yeah I'll add it.

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2884,50 +2885,207 @@ TEST_F(CompilerTest, CompileWhenIncludeThenLoadUsed) {
...
+    !22 = !{i32 0, !2, !2}
+    !23 = !{[2 x i32] [i32 0, i32 1]}
+    !24 = !{void ()* @main, !"main", !25, null, null}
+    !25 = !{null, !26, null}
+    !26 = !{!27}
+    !27 = !{i32 0, !"SV_Target", i8 9, i8 16, !28, i8 0, i32 1, i8 1, i32 0, i8 0, !29}
+    !28 = !{i32 0}
+    !29 = !{i32 3, i32 1}
+    !30 = !DILocation(line: 2, column: 28, scope: !4)
+  )x";
```
pow2clk, line 2993: That's a lot to inline. Might it be loaded from a .ll file?

include/dxc/Support/Path.h
```diff
@@ -96,28 +114,16 @@ StringTy NormalizePathImpl(const CharT *Path, size_t Length) {
...
 
   // If relative path, prefix with dot.
   if (IsAbsoluteOrCurDirRelativeImpl<CharT>(PathCopy.c_str(),
                                             PathCopy.size())) {
     return PathCopy;
   } else {
-    return StringTy(1, CharT('.')) + StringTy(1, SlashTo) + PathCopy;
+    PathCopy = StringTy(1, CharT('.')) + StringTy(1, SlashTo) + PathCopy;
+    RemoveDoubleSlashes<CharT, StringTy>(PathCopy, SlashTo);
+    return PathCopy;
```
pow2clk, line 126: It really feels like all of this should be available in the llvm::sys::path utilities, but I sure can't find it 😕

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2884,50 +2885,207 @@ TEST_F(CompilerTest, CompileWhenIncludeThenLoadUsed) {
...
-  public:
-    DXC_MICROCOM_ADDREF_RELEASE_IMPL(m_dwRef)
-    dxc::DxcDllSupport &m_dllSupport;
-    HRESULT m_defaultErrorCode = E_FAIL;
-    SimpleIncludeHanlder(dxc::DxcDllSupport &dllSupport)
-        : m_dwRef(0), m_dllSupport(dllSupport) {}
-    HRESULT STDMETHODCALLTYPE QueryInterface(REFIID iid,
-                                             void **ppvObject) override {
-      return DoBasicQueryInterface<IDxcIncludeHandler>(this, iid, ppvObject);
+class SimpleIncludeHanlder : public IDxcIncludeHandler {
```
pow2clk, line 2904: pre-existing typo: SimpleIncludeHanlder -> SimpleIncludeHandler

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -2884,50 +2885,207 @@ TEST_F(CompilerTest, CompileWhenIncludeThenLoadUsed) {
...
-  };
+  std::wstring ret = s;
+  for (wchar_t &c : ret) {
+    if (c == From)
+      c = To;
+  }
+  return ret;
+};
 
-  class SimpleIncludeHanlder : public IDxcIncludeHandler {
```
pow2clk, line 2904: Question: Can you confirm that the only change to this was indentation?

pow2clk (approved):
I don't have any objections really, but you might fix the typo in the class that got the indentation changed.

tools/clang/unittests/HLSL/CompilerTest.cpp
adam-yang: Actually there was a call to `NormalizeForPlatform` that was removed. It would have actually defeated the purpose of the test during the recompile.

tools/clang/unittests/HLSL/CompilerTest.cpp
adam-yang: I put the IR in a separate header file and included it here, like another test did.

tools/clang/unittests/HLSL/CompilerTest.cpp
adam-yang: Fixed.

include/dxc/Support/Path.h
adam-yang: There's `llvm::sys::path::native`. However `native` replaces all double backslash with single backslash, which is not really what we want for UNC paths. The `RemoveDoubleSlashes` functionality is definitely not in there. Even if it is, I would like to keep this functionality header only to avoid linking problems.

PR #6320, opened 2024-02-16
Don't set RaytracingTier1_1 based on subobjects; enable flag validation

### LINKED ISSUE #6321: RaytracingTier1_1 flag should not be set in module based on subobject
**Description**

**Steps to Reproduce**
```hlsl
// dxc -T lib_6_5
```

**Actual Behavior**

**Environment**
- DXC version: any
- Host Operating System: any

tex3d:
In validator version 1.7 and below, the RaytracingTier1_1 module flag was set on every function if any StateObjectConfig subobject had the AllowStateObjectAdditions flag set. This was incorrect, as the requirement is validated in the runtime based on the use of the subobject instead. Subobjects are removed from the module and placed in RDAT during container serialization, so the requirement would be lost when recomputing the flags in validation. This didn't break validation because flag validation was completely disabled for libraries!

This change fixes this problem, and allows DxilValidation to validate ShaderFlags because they will no longer mismatch due to this issue. Running CollectShaderFlagsForModule is also necessary for collecting per-function shader flags which will be used by call graph validation in a subsequent change, so enabling flag validation unblocks that as well.

Fixes #6321

### DIFF
--- a/lib/DXIL/DxilModule.cpp
+++ b/lib/DXIL/DxilModule.cpp
@@ -288,27 +288,6 @@ unsigned DxilModule::GetGlobalFlags() const {
   return Flags;
 }
 
-static bool RequiresRaytracingTier1_1(const DxilSubobjects *pSubobjects) {
-  if (!pSubobjects)
-    return false;
-  for (const auto &it : pSubobjects->GetSubobjects()) {
-    switch (it.second->GetKind()) {
-    case DXIL::SubobjectKind::RaytracingPipelineConfig1:
-      return true;
-    case DXIL::SubobjectKind::StateObjectConfig: {
-      uint32_t configFlags;
-      if (it.second->GetStateObjectConfig(configFlags) &&
-          ((configFlags &
-            (unsigned)DXIL::StateObjectFlags::AllowStateObjectAdditions) != 0))
-        return true;
-    } break;
-    default:
-      break;
-    }
-  }
-  return false;
-}
-
 void DxilModule::CollectShaderFlagsForModule(ShaderFlags &Flags) {
   ComputeShaderCompatInfo();
   for (auto &itInfo : m_FuncToShaderCompat)
@@ -380,10 +359,6 @@ void DxilModule::CollectShaderFlagsForModule(ShaderFlags &Flags) {
   bool hasCSRawAndStructuredViaShader4X =
       hasRawAndStructuredBuffer && m_pSM->GetMajor() == 4 && m_pSM->IsCS();
   Flags.SetCSRawAndStructuredViaShader4X(hasCSRawAndStructuredViaShader4X);
-
-  if (!Flags.GetRaytracingTier1_1()) {
-    Flags.SetRaytracingTier1_1(RequiresRaytracingTier1_1(GetSubobjects()));
-  }
 }
 
 void DxilModule::CollectShaderFlagsForModule() {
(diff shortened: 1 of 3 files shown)


tools/clang/test/HLSLFileCheck/shader_targets/raytracing/subobjects_raytracingPipelineConfig1.hlsl
```diff
@@ -1,4 +1,9 @@
-// RUN: %dxilver 1.5 | %dxc -T lib_6_3 %s | FileCheck %s
+// RUN: %dxilver 1.5 | %dxc -T lib_6_3 %s | FileCheck %s -check-prefixes=CHECK
```
python3kgae, line 1: Why check-prefixes here?

pow2clk (commented):
I think it would be helpful to acknowledge #6207 and the changes it made to this code. The description refers to 1.7 behavior, but is actually removing code that was added by that change which confused me at first.

In particular, I'd appreciate an explanation of why any consideration of the subobjects isn't needed to determine whether raytracing 1.1 is required. After this change, it seems that the only way to set the raytracing 1.1 flag is if these opcodes are used: 
https://github.com/microsoft/DirectXShaderCompiler/blob/df588beb485958a3ece128bfbeda5fb888e2dfce/lib/DXIL/DxilShaderFlags.cpp#L634-L636

That doesn't sound like checking for the use of subobjects to me and I think there are other ops that make use of them?

tools/clang/test/HLSLFileCheck/shader_targets/raytracing/subobjects_raytracingPipelineConfig1.hlsl
tex3d: Oops, leftover from intermediate change, I can remove it.

python3kgae (approved):

lib/DXIL/DxilModule.cpp
```diff
@@ -380,10 +359,6 @@ void DxilModule::CollectShaderFlagsForModule(ShaderFlags &Flags) {
   bool hasCSRawAndStructuredViaShader4X =
       hasRawAndStructuredBuffer && m_pSM->GetMajor() == 4 && m_pSM->IsCS();
   Flags.SetCSRawAndStructuredViaShader4X(hasCSRawAndStructuredViaShader4X);
-
-  if (!Flags.GetRaytracingTier1_1()) {
-    Flags.SetRaytracingTier1_1(RequiresRaytracingTier1_1(GetSubobjects()));
-  }
```
pow2clk, line 386: Without this, the only indication of raytracing tier 1.1 requirement is the opcodes check that I referenced in my global comment.

tex3d:
> I think it would be helpful to acknowledge #6207 and the changes it made to this code. The description refers to 1.7 behavior, but is actually removing code that was added by that change which confused me at first.
> 
> In particular, I'd appreciate an explanation of why any consideration of the subobjects isn't needed to determine whether raytracing 1.1 is required. After this change, it seems that the only way to set the raytracing 1.1 flag is if these opcodes are used:
> 
> https://github.com/microsoft/DirectXShaderCompiler/blob/df588beb485958a3ece128bfbeda5fb888e2dfce/lib/DXIL/DxilShaderFlags.cpp#L634-L636
> 
> That doesn't sound like checking for the use of subobjects to me and I think there are other ops that make use of them?

The code was basically moved, not exactly added.  I added a [note](https://github.com/microsoft/DirectXShaderCompiler/issues/6321#issuecomment-1949322225) to issue #6321.  Let me know if that helps regarding the churn of that code.

That issue should also explain why this flag should not be set from the subobject, since for subobjects the feature requirement is checked independently in the runtime, and the subobjects are part of RDAT, and not part of the module.

There are multiple things that could require raytracing 1.1 though.  As you see based on the link, a function could use a DXR 1.1 feature directly, in which case, the function that uses it *will* reflect the raytracing 1.1 requirement flag, and this requirement will bubble up to the module flags as well.

lib/DXIL/DxilModule.cpp
tex3d: And that's by-design.  The subobject dependency on DXR 1.1 is independently checked by the runtime.  If this flag is not set on the module, then nothing in the module sent to the driver has a DXR 1.1 dependency.  Only if the runtime uses a subobject with that dependency (and checks the requirement), will it send a subobject to the driver requiring DXR 1.1.

pow2clk (approved):
Thanks for the explanation!

PR #6322, opened 2024-02-16
Don't set RaytracingTier1_1 based on subobjects; enable flag validation (#6320)

### LINKED ISSUE #6321: RaytracingTier1_1 flag should not be set in module based on subobject

tex3d:
In validator version 1.7 and below, the RaytracingTier1_1 module flag was set on every function if any StateObjectConfig subobject had the AllowStateObjectAdditions flag set. This was incorrect, as the requirement is validated in the runtime based on the use of the subobject instead. Subobjects are removed from the module and placed in RDAT during container serialization, so the requirement would be lost when recomputing the flags in validation. This didn't break validation because flag validation was completely disabled for libraries!

This change fixes this problem, and allows DxilValidation to validate ShaderFlags because they will no longer mismatch due to this issue. Running CollectShaderFlagsForModule is also necessary for collecting per-function shader flags which will be used by call graph validation in a subsequent change, so enabling flag validation unblocks that as well.

Fixes #6321

(cherry picked from commit fafbc4272540c09cd070ca8a7a8dcda2788bc108)

### DIFF
--- a/lib/DXIL/DxilModule.cpp
+++ b/lib/DXIL/DxilModule.cpp
@@ -288,27 +288,6 @@ unsigned DxilModule::GetGlobalFlags() const {
   return Flags;
 }
 
-static bool RequiresRaytracingTier1_1(const DxilSubobjects *pSubobjects) {
-  if (!pSubobjects)
-    return false;
-  for (const auto &it : pSubobjects->GetSubobjects()) {
-    switch (it.second->GetKind()) {
-    case DXIL::SubobjectKind::RaytracingPipelineConfig1:
-      return true;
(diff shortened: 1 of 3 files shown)


bob80905 (approved):
I understand this is a cherry-pick, but I would think a test case that validates that shader flags are validated in library shaders (that checks for a validation error, for example) would be a valuable addition. Maybe worth an issue, maybe not.

pow2clk (approved):

PR #6325, opened 2024-02-17
NFC: ExtractIRForPassTest.py error reporting when pass not found

tex3d:
This just adds convenient error reporting to `ExtractIRForPassTest.py` when the specified pass is not found.

No functional change.


bob80905 (approved):

tex3d:
In between making this PR and now, another change was merged (PR #6519) to throw when the pass name was not found.  I still think this PR is valuable since it provides more detailed and useful reporting than the current state.  If the pass name wasn't found, seeing the list of valid passes is *very* useful.  Additionally, if the `n`th instance of a pass is desired, but not found, this provides much more useful reporting, instead of just saying that it wasn't found.

pow2clk (approved):

pow2clk:
I think this counts as a functional change

PR #6330, opened 2024-02-20
[Test] Limit NodeObjects test to dxil-1-8

python3kgae:
Add lit.local.cfg to set unsupported when not dxil-1-8.


pow2clk (approved):
Simple enough

llvm-beanz (approved):

PR #6332, opened 2024-02-20
DxilValidation: add ShaderCompatInfo call graph checking

### LINKED ISSUE #6292: DxilValidation: Stricter validation for new DXIL barrier ops.
`DxilValidation`: Stricter validation for new DXIL barrier ops.  Add validation tests for scenarios.  Technically, this is not a DXIL-breaking change, so can be done last, but should be done before validator release.

tex3d:
ShaderCompatInfo identifies compatibility for functions that could be called from an entry point.
Currently, this checking detects compatibility problems between entry points and internal function calls that the validator otherwise misses.

This change adds a check for ShaderCompatInfo, recursing into called functions looking for a source of conflict when not compatible with the current entry point properties.
Errors are emitted for the entry point and at one source of each type of conflict that ShaderCompatInfo detects.
A function is considered the source of a conflict when it has the conflict but none of the functions called from this function have this conflict.

Removed early exit for ShaderFlags validation when module is a library, since these flags should be validated for libraries, and running CollectShaderFlagsForModule fills in the ShaderCompatInfo data we need for validation.

Also fixed tests for illegal barrier and derivative use, as well as replacing the ignored MaxOutputRecords with the correct MaxRecords attribute.

Fixes #6292.


utils/hct/hctdb.py
```diff
@@ -8081,6 +8081,56 @@ def build_valrules(self):
...
+        )
+        self.add_valrule_msg(
+            "Sm.IncompatibleShaderModel",
+            "functions may only use features available in the current shader model",
+            "Function uses features incompatible with the shader model.",
+        )
+        self.add_valrule_msg(
+            "Sm.IncompatibleThreadGroupDim",
+            "When derivatives are used in compute-model shaders, the thread group dimensions must be compatible",
+            "Function uses derivatives in compute-model shader with NumThreads (%0, %1, %2); derivatives require NumThreads to be 1D and a multiple of 4, or 2D/3D with X and Y both being a multiple of 2.",
```
bob80905, line 8116: Looks like there isn't a test for this diagnostic.

utils/hct/hctdb.py
```diff
@@ -8081,6 +8081,56 @@ def build_valrules(self):
...
+        )
+        self.add_valrule_msg(
+            "Sm.IncompatibleRequiresGroup",
+            "Functions requiring groupshared memory must be called from shaders with a visible group",
+            "Function requires a visible group, but is called from a shader without one.",
+        )
+        self.add_valrule_msg(
+            "Sm.IncompatibleDerivLaunch",
+            "Node shaders only support derivatives in broadcasting launch mode",
+            "Function called from %0 launch node shader uses derivatives; only broadcasting launch supports derivatives.",
```
bob80905, line 8131: None for this one either.

bob80905 (commented):
Some questions and concerns

lib/HLSL/DxilValidation.cpp
```diff
@@ -5372,6 +5380,211 @@ static void ValidateEntrySignatures(ValidationContext &ValCtx) {
...
+          continue;
+        maskForCalls |= Visit(called, remainingMask, visited, CG);
+        if (remainingMask == 0)
+          return 0; // Nothing left to search for.
+      }
+    }
+
+    // Mask of incompatibilities introduced by this function.
+    uint32_t conflictsIntroduced =
+        remainingMask & maskForThisFunction & ~maskForCalls;
```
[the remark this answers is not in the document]
tex3d: We are computing `conflictsIntroduced`, the mask of conflicts introduced by this function **that we still need to diagnose**, and that weren't introduced by other functions called by this one.

`remainingMask` has the remaining conflicts that have not yet been diagnosed.  We don't want to diagnose multiple sources of the same conflict under the same call graph, since it would mainly just add noise at that point.  So we remove bits from `remainingMask` when we diagnose that issue, and once there are no remaining conflicts to diagnose, we are done.

Without ANDing in `remainingMask`, we could end up diagnosing the same issue repeatedly.

lib/HLSL/DxilValidation.cpp
```diff
@@ -5372,6 +5380,211 @@ static void ValidateEntrySignatures(ValidationContext &ValCtx) {
...
+    // flags introduced by this function from the working mask so we don't
+    // report this conflict again.
+    // - When the remainingMask is 0, we are done.
+
+    if (remainingMask == 0)
+      return 0; // Nothing left to search for.
+    if (!visited.insert(F).second)
+      return 0; // Already visited.
+
+    const DxilModule::ShaderCompatInfo *compatInfo =
```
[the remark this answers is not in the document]
tex3d: If compat info wasn't computed for the function.  This info is computed in `ComputeShaderCompatInfo()` which is called from `CollectShaderFlagsForModule()`, which occurs earlier in the validation flow.  If we moved the sequence around, we could end up missing compat info for a function.  We shouldn't be missing this, but I wanted the release build validator to be robust to it, which is why it asserts and 0 is returned when not found if the assert is disabled.

lib/HLSL/DxilValidation.cpp
```diff
@@ -5372,6 +5380,211 @@ static void ValidateEntrySignatures(ValidationContext &ValCtx) {
...
+    Diagnose(F, conflictMask, ConflictKind::DerivThreadGroupDim,
+             ValidationRule::SmIncompatibleThreadGroupDim,
+             {std::to_string(props.numThreads[0]),
+              std::to_string(props.numThreads[1]),
+              std::to_string(props.numThreads[2])});
+    Diagnose(F, conflictMask, ConflictKind::DerivInComputeShaderModel,
+             ValidationRule::SmIncompatibleDerivInComputeShaderModel);
+    Diagnose(F, conflictMask, ConflictKind::RequiresGroup,
+             ValidationRule::SmIncompatibleRequiresGroup);
+    Diagnose(F, conflictMask, ConflictKind::DerivLaunch,
```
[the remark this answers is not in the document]
tex3d: I suppose you want it in the same order as the enum, right?

utils/hct/hctdb.py
tex3d: Good catch, working on these.

lib/HLSL/DxilValidation.cpp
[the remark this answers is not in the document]
bob80905: Yes, that is what I was expecting though I don't think it matters.

lib/HLSL/DxilValidation.cpp
```diff
@@ -5372,6 +5380,211 @@ static void ValidateEntrySignatures(ValidationContext &ValCtx) {
...
+              std::to_string(props.numThreads[2])});
+    Diagnose(F, conflictMask, ConflictKind::DerivInComputeShaderModel,
+             ValidationRule::SmIncompatibleDerivInComputeShaderModel);
+    Diagnose(F, conflictMask, ConflictKind::RequiresGroup,
+             ValidationRule::SmIncompatibleRequiresGroup);
+  }
+
+  // Visit function and all functions called by it.
+  // Emit diagnostics for incompatibilities found in a function when no
+  // called functions introduced the conflict.
```
pow2clk, line 5513: I don't understand this sentence. It sounds like we're emitting incompatibility diagnostics when no called functions introduced the conflict? Maybe this is the intent? 
```suggestion
  // Emit diagnostics for incompatibilities found in a function when no
  // functions called by that function introduced the conflict.
  // In those cases, the called functions themselves will emit the diagnostic.
```

lib/HLSL/DxilValidation.cpp
```diff
@@ -5372,6 +5380,211 @@ static void ValidateEntrySignatures(ValidationContext &ValCtx) {
...
+    uint32_t maskForThisFunction = IdentifyConflict(*compatInfo);
+
+    uint32_t maskForCalls = 0;
+    if (CallGraphNode *CGNode = CG[F]) {
+      for (auto &Call : *CGNode) {
+        Function *called = Call.second->getFunction();
+        if (called->isDeclaration())
+          continue;
+        maskForCalls |= Visit(called, remainingMask, visited, CG);
+        if (remainingMask == 0)
```
pow2clk, line 5543: Nit: It seems more natural to me to return the value that continued recursion depends on and pass the other value back as a parameter. That way you wouldn't need to `|=` it and could just modify it by reference.

tools/clang/test/HLSLFileCheck/validation/callgraph/barrier-group-in-nested-fn-vs-lib68.hlsl
```diff
@@ -0,0 +1,56 @@
...
+
+// RDAT-LABEL:   UnmangledName: "main"
+// RDAT:   FeatureInfo1: 0
+// RDAT:   FeatureInfo2: (Opt_RequiresGroup)
+// ShaderStageFlag indicates no compatible entry type after masking for vertex.
+// RDAT:   ShaderStageFlag: 0
+// MinShaderTarget still indicates vertex shader.
+// RDAT:   MinShaderTarget: 0x10060
+
+//[numthreads(8, 8, 1)]
```
pow2clk, line 52: Intended inclusion?
```suggestion
```

tools/clang/test/HLSLFileCheck/validation/callgraph/barrier-group-in-nested-fn-vs-lib68.hlsl
```diff
@@ -0,0 +1,56 @@
+// RUN: %dxc -T lib_6_8 -Vd %s | %D3DReflect %s | %FileCheck %s -check-prefixes=RDAT
+
```
pow2clk, line 2: Should validation be disabled here? 

I like to have explanations of the intent of a test at the top. 

```suggestion

// Verifies that use of a group barrier in a nested call in a vs library shader
// produces no diagnostics and serves as the source for like-named .ll files

```

tools/clang/test/HLSLFileCheck/validation/callgraph/deriv-in-nested-fn-cs-lib65.hlsl
```diff
@@ -0,0 +1,55 @@
+// RUN: %dxilver 1.8 | %dxc -T lib_6_5 -Vd %s | %D3DReflect %s | %FileCheck %s -check-prefixes=RDAT
+
```
pow2clk, line 2: Again, should validation be disabled here? 

```suggestion

// Verifies that use of a group barrier in a nested call in a cs library shader
// produces no diagnostics and serves as the source for like-named .ll files

```

tools/clang/test/HLSLFileCheck/validation/callgraph/deriv-in-nested-fn-node-lib68-launch.hlsl
```diff
@@ -0,0 +1,55 @@
+// RUN: %dxilver 1.8 | %dxc -T lib_6_8 -Vd %s | %D3DReflect %s | %FileCheck %s -check-prefixes=RDAT
+
```
pow2clk, line 2: Again, should validation be disabled here? 

```suggestion

// Verifies that use of a derivative in a nested call in a node library shader
// produces no diagnostics and serves as the source for like-named .ll files

```

tools/clang/test/HLSLFileCheck/validation/callgraph/deriv-in-nested-fn-vs-lib68.hlsl
```diff
@@ -0,0 +1,60 @@
+// RUN: %dxc -T lib_6_8 -Vd %s | %D3DReflect %s | %FileCheck %s -check-prefixes=RDAT
+
+// Check ShaderCompatInfo for illegal scenario with verttex entry point calling
+// a function using derivatives (through Sample()).
+// Used to generate deriv-in-nested-fn-vs.ll
```
pow2clk, line 5: This is probably a better description than what I had above except this hlsl file doesn't check that, the generated .ll file does. So maybe put the last line at the top? 
```suggestion
// Used to generate deriv-in-nested-fn-vs.ll which will
// check ShaderCompatInfo for illegal scenario with verttex entry point calling
// a function using derivatives (through Sample()).

```

pow2clk (commented):
Comments and suggestions

lib/HLSL/DxilValidation.cpp
```diff
@@ -36,6 +36,7 @@
 #include "dxc/HLSL/DxilSpanAllocator.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/BitVector.h"
+#include "llvm/ADT/SmallSet.h"
```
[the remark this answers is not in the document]
tex3d: I think I added this during an intermediate version and forgot to remove it again.  But strangely, there are uses of SmallPtrSet, and no inclusion of `SmallPtrSet.h`, which `SmallSet.h` will also include.  I wonder where it's getting the necessary header from.  Must be from one of the other headers somewhere.  Should I just change this to `SmallPtrSet.h`, since that one is directly used in this file?

lib/HLSL/DxilValidation.cpp
```diff
@@ -5372,6 +5380,211 @@ static void ValidateEntrySignatures(ValidationContext &ValCtx) {
...
 
+// CompatibilityChecker is used to identify incompatibilities in an entry
+// function and any functions called by that entry function.
+struct CompatibilityChecker {
+  ValidationContext &ValCtx;
+  Function *EntryFn;
+  const DxilFunctionProps &props;
+  DXIL::ShaderKind shaderKind;
+
+  // Precomputed masks OR'd in based on flag set in ShaderCompatInfo.
```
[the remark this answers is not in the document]
tex3d: Ok.  Also looks like I can skip the code that sets these masks if I pass in the entry functions ShaderCompatInfo and only set them when it indicates deriv or group usage.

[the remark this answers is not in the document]
tex3d: Although that will complicate the logic without any real benefit, so I'll just leave it alone.

tools/clang/test/HLSLFileCheck/validation/callgraph/barrier-group-in-nested-fn-vs-lib68.hlsl
tex3d: Yes, Validation is disabled because this HLSL is making sure the ShaderCompatInfo has the expected flags set to produce an illegal call graph which will be rejected by validation, which is used to generate .ll tests.

I'll add a comment, but the intent of the test isn't quite as stated in your suggestion, so I'll write it differently.

tools/clang/test/HLSLFileCheck/validation/callgraph/barrier-group-in-nested-fn-vs-lib68.hlsl
tex3d: Yeah, I had been using and modifying a file to generate .ll before copying to here and forgot to remove the commented attribute.

tools/clang/test/HLSLFileCheck/validation/callgraph/deriv-in-nested-fn-cs-lib65.hlsl
tex3d: Yes, validation needs to be disabled so we can produce the RDAT and check that the flags were propagated as expected.  I'll add a comment.

tools/clang/test/HLSLFileCheck/validation/callgraph/deriv-in-nested-fn-vs-lib68.hlsl
tex3d: This file checks that the ShaderCompatInfo contains the expected state that the validator will catch, which is why validation is disabled on this file, otherwise we wouldn't have RDAT to check.  I'll try to clarify the comment.

lib/HLSL/DxilValidation.cpp
tex3d: So your suggestion is to add the `maskForCalls` as a reference param, which gets ORd in the Visit call, and `remainingMask` will be passed in by-value, then Visit will return the masked result after diagnosing a conflict?

That's one additional parameter and doesn't really seem to simplify things, since you would need to track two separate values for `maskForCalls` inside Visit: one for the parameter which you OR bits for this functions into, and a new one on the stack for the inner calls to Visit, which is initialized to 0 and gets OR'd into by those inner calls.  That spreads the logic for `maskForCalls` between an outer and inner Visit call, which seems more complicated than just doing it where we want to see the union of the conflict masks for the inner calls.

The current definition returns a simple value: the function's conflict mask.  It also updates remainingMask for the remaining diagnostics.

I'm not sure I want to refactor this code at this point.  If I do, I would need to revisit the decision of whether or not to diagnose the same conflict multiple times when it shows up from different called functions.  This would also influence the approach with `remainingMask` as well.

bob80905 (approved):

pow2clk (approved):
Thanks for the responses!

PR #6333, opened 2024-02-21
Fix downlevel validator compatibility with DXR 1.1 flag

### LINKED ISSUE #6321: RaytracingTier1_1 flag should not be set in module based on subobject

tex3d:
It turns out that in the prior validator version, if a subobject required DXR 1.1, the DXR 1.1 flag would be set on each function in RDAT output, as well as the global flags. It didn't appear this was the case through a D3DReflect test because that goes through disassembly to assembly step, where subobjects are lost because they aren't in the llvm IR. The previous change assumed this was not the case when the subobjects were removed, but this removal occurs after the RDATWriter constructor, which does the full RDAT serialization since size if required right away.

This restores the detection code and hooks it into DxilModule::ComputeShaderCompatInfo when validator version is in range [1.5, 1.7]. DXR 1.1 was introduced in 1.5, and we no longer tag every function as requiring DXR 1.1 based on subobjects in 1.8.

At the moment, there is no way to CHECK the subobject RDAT in D3DReflect because they get stripped from the module (even reflection and debug modules) before serialization, and the test path for D3DReflect goes through a disassemble/re-assemble step between the prior stage and the D3DReflect stage.

Fixes #6321 (really).

### DIFF
--- a/lib/HLSL/DxilValidation.cpp
+++ b/lib/HLSL/DxilValidation.cpp
@@ -4494,7 +4494,8 @@ static void ValidateShaderFlags(ValidationContext &ValCtx) {
   // which are gone by this point.  Set the flag and the rest should match.
   unsigned valMajor, valMinor;
   ValCtx.DxilMod.GetValidatorVersion(valMajor, valMinor);
-  if (DXIL::CompareVersions(valMajor, valMinor, 1, 8) < 0 &&
+  if (DXIL::CompareVersions(valMajor, valMinor, 1, 5) >= 0 &&
+      DXIL::CompareVersions(valMajor, valMinor, 1, 8) < 0 &&
       ValCtx.DxilMod.m_ShaderFlags.GetRaytracingTier1_1() &&
       !calcFlags.GetRaytracingTier1_1()) {
     calcFlags.SetRaytracingTier1_1(true);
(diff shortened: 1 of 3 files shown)


lib/HLSL/DxilValidation.cpp
```diff
@@ -4487,6 +4487,19 @@ static void ValidateResources(ValidationContext &ValCtx) {
...
   ShaderFlags calcFlags;
   ValCtx.DxilMod.CollectShaderFlagsForModule(calcFlags);
+
+  // Special case for validator version prior to 1.8.
+  // If DXR 1.1 flag is set, but our computed flags do not have this set, then
+  // this is due to prior versions setting the flag based on DXR 1.1 subobjects,
+  // which are gone by this point.  Set the flag and the rest should match.
+  unsigned valMajor, valMinor;
+  ValCtx.DxilMod.GetValidatorVersion(valMajor, valMinor);
+  if (DXIL::CompareVersions(valMajor, valMinor, 1, 8) < 0 &&
```
bob80905, line 4497: Do we also need to check that the validator version is >1.5 here?

tex3d: Sure, we could, thanks.

bob80905 (approved):

hekota (approved):
LGTM!

PR #6336, opened 2024-02-21
Fix duplicate output of "globallycoherent" attribute in HLSL rewriter.

LukasBanana:
This attribute is already handled in TypePrinter::printAttributedBefore and must not be emitted hlsl::CustomPrintHLSLAttr prior to that call.


damyanp (approved):
Could you add some test coverage for this please?

PR #6338, opened 2024-02-21
[SPIR-V] Allow SamplerComparisonState in combinedImageSampler

### LINKED ISSUE #4696: [SPIR-V] `[[vk::combinedImageSampler]]` refuses to work with `SamplerComparisonState`
Hey, I tried doing this:

```hlsl
[[vk::combinedImageSampler]]
Texture2D foo: register(t0);
[[vk::combinedImageSampler]]
SamplerComparisonState foo_sampler: register(s0);
```

And got this error:

> 'combinedImageSampler' attribute only applies to Textures (e.g., Texture2D) and SamplerState

Interestingly, there is a workaround, which even causes SPIR-V Cross to produce the correct GLSL 😀: make a dummy `SamplerState` using the same register, and apply [[vk::combinedImageSampler]] to that instead!

```hlsl
[[vk::combinedImageSampler]]
Texture2D foo: register(t0);
[[vk::combinedImageSampler]]
SamplerState foo_sampler_UNUSED: register(s0);
SamplerComparisonState foo_sampler: register(s0);
```

sudonatalie:
Previously, only `SamplerState` was allowed to be used as the sampler type with `[[vk::combinedImageSampler]]`, but I don't see any reason that `SamplerComparisonState` shouldn't also be permitted. Since `SamplerComparisonState` is a sampler state plus a comparison state, `SampleCmp` calls are translated to an `OpImageSampleDrefImplicitLod` (sampling with depth-comparison) in Vulkan SPIR-V, which does accept an `OpTypeSampledImage` (combined image sampler) as an operand.

Fixes #4696

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/vk.combined-image-sampler.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/vk.combined-image-sampler.hlsl
@@ -25,6 +25,13 @@ Texture2D<float4> tex2 : register(t2);
 // CHECK: OpDecorate %sam3 Binding 0
 SamplerState sam3 : register(s0);
 
+// CHECK: OpDecorate %tex4 DescriptorSet 0
+// CHECK: OpDecorate %tex4 Binding 4
+[[vk::combinedImageSampler]]
+Texture2D tex4 : register(t4);
+[[vk::combinedImageSampler]]
+SamplerComparisonState sam4 : register(s4);
+
 Texture2D<float4> getTexture(int i) {
   if (i == 0) return tex0;
   if (i == 1) return tex1;
@@ -58,6 +65,10 @@ float4 main(int3 offset: A) : SV_Target {
 // CHECK: [[tex2:%[a-zA-Z0-9_]+]] = OpLoad %type_sampled_image %tex2
 // CHECK: OpImageSampleExplicitLod %v4float [[tex2]]
   ret += sampleLevel(2);
+
+// CHECK: [[tex4:%[a-zA-Z0-9_]+]] = OpLoad %type_sampled_image %tex4
+// CHECK: OpImageSampleDrefImplicitLod %float [[tex4]]
+  ret += tex4.SampleCmp(sam4, float2(1, 2), 10, 2);
 
 // CHECK: [[tex0_0:%[a-zA-Z0-9_]+]] = OpLoad %type_sampled_image %tex0
 // CHECK: [[img_extracted_from_tex0:%[a-zA-Z0-9_]+]] = OpImage %type_2d_image [[tex0_0]]
--- a/tools/clang/include/clang/Basic/Attr.td
+++ b/tools/clang/include/clang/Basic/Attr.td
@@ -1258,14 +1258,13 @@ def ArrayOfBuffer
                   S->getType()->getAsArrayTypeUnsafe()->getElementType()->getAs<RecordType>()->getDecl()->getName() ==
                       "RasterizerOrderedBuffer")}]>;
 
-// Global variable with "Texture" or "SamplerState" type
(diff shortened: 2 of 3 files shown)


sudonatalie:
@cassiebeckley @Keenuts I've reviewed the spec and think this change seems reasonable, but please go ahead and double-check references if you see any reason this change might not make sense.

cassiebeckley (approved):
Seems reasonable to me.

tools/clang/test/CodeGenSPIRV/vk.combined-image-sampler.hlsl
```diff
@@ -25,6 +25,13 @@ Texture2D<float4> tex2 : register(t2);
 // CHECK: OpDecorate %sam3 Binding 0
 SamplerState sam3 : register(s0);
 
+// CHECK: OpDecorate %tex4 DescriptorSet 0
+// CHECK: OpDecorate %tex4 Binding 4
+[[vk::combinedImageSampler]]
+Texture2D tex4 : register(t4);
+[[vk::combinedImageSampler]]
+SamplerComparisonState sam4 : register(s4);
```
Keenuts, line 33: The vulkan specs seems to allow using `sam4` with `tex2` for ex, but this crashes at the legalization step in the current state. Maybe that's why it was disabled?

sudonatalie: Good catch. That's an invalid use case because once a sampler is used in a combined image sampler, it can't be used with other textures at different descriptor bindings. That's because a SPIR-V sampler can't be extracted from the combined sampler. We should fail with something more helpful though, so I'll take a closer look.

Keenuts: Wait, is that the case? My understanding of the Vulkan spec is the opposite:
(https://registry.khronos.org/vulkan/specs/1.2/html/chap14.html#descriptorsets-combinedimagesampler)

```
If the descriptor refers to a sampler that performs [Y′CBCR conversion](https://registry.khronos.org/vulkan/specs/1.2/html/chap13.html#samplers-YCbCr-conversion), the sampler must only be used to sample the image in the same descriptor. Otherwise, the sampler and image in this type of descriptor can be used freely with any other samplers and images.
```

sudonatalie: Oh, we just don't handle that error case at all. `tex0.SampleLevel(sam1, float2(1, 2), 10, 2)` also fails at legalization. I opened a new issue to track it: https://github.com/microsoft/DirectXShaderCompiler/issues/6346.

I don't think it needs to block this PR though since it's an existing bug that isn't affected by this change.

Keenuts: Oh right, thanks for checking!

Keenuts (approved):

tools/clang/test/CodeGenSPIRV/vk.combined-image-sampler.hlsl
sudonatalie: Sorry, I didn't refresh the page to see your new comment before posting mine. I'll take a look at the spec later today, maybe we shouldn't have that restriction.

sudonatalie: Thanks, haha, I'm all out of sync on the threads today. I guess either way it's not blocking this PR, but we should investigate further... I'm not actually sure how we'd differentiate between whether or not the sampler performs Y′CBCR conversion, so I think it still might be the right call to disallow it in HLSL. But either way we should fail more gracefully than we do now.

PR #6343, opened 2024-02-22
[lit] support dxil version check in lit test

python3kgae:
dxil-1-* will be added as available_features for lit config.

There're 3 ways to use it.
1. Whole file limit use required dxil-1-8
2. Single RUN line with %if dxil-1-8 %{ %}
3. Whole folder with check dxil-1-8 to set config.unsupported in lit.local.cfg


pow2clk (approved):

tex3d (approved):
I believe we want to merge, not squash, this.  Looks good.

PR #6348, opened 2024-02-22
[ExecutionTest] Add Wave Range Size HLK Tests

### LINKED ISSUE #6347: [ExecutionTest] WaveRangeSize is missing HLK tests
WaveRangeSize was recently implemented, but there are no accompanying HLK tests to test that the implementation behaves correctly with drivers. We need to make sure that the waverangesize values are communicated and interpreted correctly by GPU drivers.

ExecutionTest.cpp should add testing for WaveRangeSize values, and have coverage on a variety of ranges, from ranges that don't overlap with wave sizes available on the hardware, to partially overlapping ranges, to ranges that are subsets of the allowed hardware range.

None

**Additional context**
This PR essentially implements the Execution Testing paragraph of the wave range size spec in hlsl-specs, described here:
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0013-wave-size-range.md#execution-testing

bob80905:
This PR adds execution testing to wave range size attributes, on top of the already existing WaveSizeTest HLK test.
It implements the necessary coverage of the new wave range size attribute, as described in the hlsl-specs here: https://github.com/microsoft/hlsl-specs/blob/main/proposals/0013-wave-size-range.md#execution-testing
Fixes #6347

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -13160,7 +13160,7 @@ void RunWaveSizeTest(UINT minWaveSize, UINT maxWaveSize,
       "[numthreads(" stringify(
           MAX_WAVESIZE) "*2,1,1)]\r\n"
                         "void main(uint3 tid : SV_DispatchThreadID ) { \r\n"
-                        "  data[tid.x].count = WaveActiveSum(1); \r\n"
+                        "  data[tid.x].count = WaveGetLaneCount(); \r\n"
                         "}\r\n";
 
   struct WaveSizeTestData {
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -13312,7 +13312,7 @@ void RunWaveSizeRangeTest(UINT minWaveSize, UINT maxWaveSize,
         // Only allow valid shader wave ranges
         bool AcceptedByRuntime = TestShaderRangeAgainstRequirements(
             minShaderWaveSize, maxShaderWaveSize, minWaveSize, maxWaveSize);
-        if (!AcceptedByRuntime) {          
+        if (!AcceptedByRuntime) {
           continue;
         }
 
@@ -13332,7 +13332,8 @@ void ExecutionTest::WaveSizeTest() {
       WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
 
   CComPtr<ID3D12Device> pDevice;
-  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_6, /*skipUnsupported*/ false)) {
+  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_6,
+                    /*skipUnsupported*/ false)) {
     return;
   }
 
@@ -13371,7 +13372,8 @@ void ExecutionTest::WaveSizeRangeTest() {
       WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
 
   CComPtr<ID3D12Device> pDevice;
-  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_8, /*skipUnsupported*/ false)) {
+  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_8,
+                    /*skipUnsupported*/ false)) {
     return;
   }
 
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -328,6 +328,7 @@ public:
   TEST_METHOD(WaveIntrinsicsDDITest);
   TEST_METHOD(WaveIntrinsicsInPSTest);
   TEST_METHOD(WaveSizeTest);
+  TEST_METHOD(WaveSizeRangeTest);
   TEST_METHOD(PartialDerivTest);
   TEST_METHOD(DerivativesTest);
   TEST_METHOD(ComputeSampleTest);
@@ -13248,17 +13249,10 @@ void RunWaveSizeRangeTest(UINT minWaveSize, UINT maxWaveSize,
       for (UINT prefShaderWaveSize = minShaderWaveSize;
            prefShaderWaveSize <= maxShaderWaveSize; prefShaderWaveSize *= 2) {
 
-        // For now, only allow valid shader wave ranges
-        // TODO: Create callback function for CreateComputePipelineState to
-        // handle expected errors.
+        // Only allow valid shader wave ranges
         bool AcceptedByRuntime = TestShaderRangeAgainstRequirements(
             minShaderWaveSize, maxShaderWaveSize, minWaveSize, maxWaveSize);
-        if (!AcceptedByRuntime) {
-          LogCommentFmt(
-              L"Wave size range [%d, %d, %d] not supported by runtime, "
-              L"which only supports range [%d, %d]",
-              minShaderWaveSize, maxShaderWaveSize, prefShaderWaveSize,
-              minWaveSize, maxWaveSize);
+        if (!AcceptedByRuntime) {          
           continue;
         }
 
@@ -13289,11 +13283,9 @@ void RunWaveSizeRangeTest(UINT minWaveSize, UINT maxWaveSize,
         MappedData dataUav;
         WaveSizeTestData *pOutData;
 
-        LogCommentFmt(
-            L"Verifying test result for these shader wave size ranges:");
-        LogCommentFmt(L"Minimum shader wave size: %d", minShaderWaveSize);
-        LogCommentFmt(L"Maximum shader wave size: %d", maxShaderWaveSize);
-        LogCommentFmt(L"Preferred shader wave size: %d", prefShaderWaveSize);
+        LogCommentFmt(L"Verifying wave size range test results for (min, max, "
+                      L"preferred): (%d, %d, %d)",
+                      minShaderWaveSize, maxShaderWaveSize, prefShaderWaveSize);        
 
         // at this point we assume that the waverange size that
         // the shader specifies is legal.
@@ -13322,14 +13314,14 @@ void ExecutionTest::WaveSizeTest() {
       WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
 
   CComPtr<ID3D12Device> pDevice;
-  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_6)) {
+  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_6, /*skipUnsupported*/ false)) {
     return;
   }
 
   // Check Wave support
   if (!DoesDeviceSupportWaveOps(pDevice)) {
     // Optional feature, so it's correct to not support it if declared as such.
-    WEX::Logging::Log::Comment(L"Device does not support wave operations.");
+    WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
     return;
   }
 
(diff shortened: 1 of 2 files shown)


tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,164 @@ void ExecutionTest::WaveSizeTest() {
...
+                        "}\r\n";
+
+  struct WaveSizeTestData {
+    uint32_t count;
+  };
+
+  for (UINT minShaderWaveSize = 4; minShaderWaveSize <= maxWaveSize;
+       minShaderWaveSize *= 2) {
+    for (UINT maxShaderWaveSize = minShaderWaveSize * 2;
+         maxShaderWaveSize <= 128; minShaderWaveSize *= 2) {
```
hekota, line 13251: Did you mean `maxShaderWaveSize *= 2`?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,164 @@ void ExecutionTest::WaveSizeTest() {
...
+            minShaderWaveSize, maxShaderWaveSize, minWaveSize, maxWaveSize);
+        if (AcceptedByRuntime) {
+
+          test->Test->GetReadBackData("UAVBuffer0", &dataUav);
+          VERIFY_ARE_EQUAL(sizeof(WaveSizeTestData) * MAX_WAVESIZE,
+                           dataUav.size());
+          pOutData = (WaveSizeTestData *)dataUav.data();
+
+          for (unsigned i = 0; i < MAX_WAVESIZE; i++) {
+            if (!VERIFY_IS_TRUE(pOutData[i].count >= minWaveSize &&
```
hekota, line 13297: I don't think `VERIFY_IS_TRUE` needs the `if` and `break` (although the original WaveSize test is written that way). If the verify condition is false the test fails and stops executing any further.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,164 @@ void ExecutionTest::WaveSizeTest() {
...
+                                pOutData[i].count <= maxWaveSize))
+              break;
+            if (prefShaderWaveSize >= minWaveSize &&
+                prefShaderWaveSize <= maxWaveSize) {
+
+              if (!VERIFY_ARE_EQUAL(pOutData[i].count, prefShaderWaveSize))
```
hekota, line 13303: Is it mandatory to honor the preferred wave size, or optional? If optional, we should not be checking it and failing the test if it does not match.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,164 @@ void ExecutionTest::WaveSizeTest() {
...
+    return;
+  }
+
+  // Check Wave support
+  if (!DoesDeviceSupportWaveOps(pDevice)) {
+    // Optional feature, so it's correct to not support it if declared as such.
+    WEX::Logging::Log::Comment(L"Device does not support wave operations.");
+    return;
+  }
+  RunWaveSizeRangeTest(minWaveSize, maxWaveSize, ShaderOpSet, pDevice,
```
hekota, line 13362: It would be good to add a logging comment before the `RunWaveSizeRangeTest` call, something like:

`LogCommentFmt(L"Testing WaveSize attribute with ranges for shader model 6.8.");`

and similar comment above for shader model 6.6:

`LogCommentFmt(L"Testing WaveSize attribute for shader model 6.6.");`

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,164 @@ void ExecutionTest::WaveSizeTest() {
...
+              break;
+            if (prefShaderWaveSize >= minWaveSize &&
+                prefShaderWaveSize <= maxWaveSize) {
+
+              if (!VERIFY_ARE_EQUAL(pOutData[i].count, prefShaderWaveSize))
+                break;
+            }
+          }
+        } else {
+          VERIFY_ARE_EQUAL(0u, dataUav.size());
```
hekota, line 13308: I would add a logging comment here that the wave size combination was not supported.

Also, I am not sure if the test will be rejected silently to even get here. I believe if the shader is not accepted it will fail the test and stop executing any further, but I might be wrong. If that happens, we might need to adjust the test to run just on the supported wave size range.

hekota (changes requested):

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
bob80905: Yes, thanks!

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
bob80905: Ok, I've removed the if condition for the wave size range test.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
bob80905: My understanding is that if the device allows the preferred wave size, it is mandatory that the preferred wave size is supported.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
bob80905: Done!

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
bob80905: Done!

hekota (approved):
LGTM!

bob80905:
For future reference, the hardware I tested on only allowed waverange size of [32,32]. So I could only test wave range sizes that included 32. CreateComputePipelineState() fails if the shader expresses a waverange size outside of what's required, so a proposal to test all possible combinations between 4,128 would be to create a callback function that runs CreateComputePipelineState(), and encapsulates it in a try/catch. Then, we can expect the error that is thrown when CreateComputePipelineState() is run with incompatible wavesize, and have full test coverage.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,173 @@ void ExecutionTest::WaveSizeTest() {
...
+
+  CComPtr<ID3D12Device> pDevice;
+  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_6)) {
+    return;
+  }
+
+  // Check Wave support
+  if (!DoesDeviceSupportWaveOps(pDevice)) {
+    // Optional feature, so it's correct to not support it if declared as such.
+    WEX::Logging::Log::Comment(L"Device does not support wave operations.");
```
tex3d, line 13332: At this point, we can't run any tests, so we should explicitly report skip, right?
```cpp
    WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,173 @@ void ExecutionTest::WaveSizeTest() {
...
+  CComPtr<IStream> pStream;
+  std::shared_ptr<st::ShaderOpSet> ShaderOpSet =
+      std::make_shared<st::ShaderOpSet>();
+  ReadHlslDataIntoNewStream(L"ShaderOpArith.xml", &pStream);
+  st::ParseShaderOpSetFromStream(pStream, ShaderOpSet.get());
+
+  LogCommentFmt(L"Testing WaveSize attribute for shader model 6.6.");
+  RunWaveSizeTest(minWaveSize, maxWaveSize, ShaderOpSet, pDevice, m_support);
+  pDevice.Release();
+  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_8)) {
```
tex3d, line 13358: For this case, if we fail to create SM 6.8 device, we don't want to report skip for the overall result of this test, so we need to pass false to the `skipUnsupported` parameter, like:
```cpp
  if (!CreateDevice(&pDevice, D3D_SHADER_MODEL_6_8, /*skipUnsupported*/ false)) {
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,173 @@ void ExecutionTest::WaveSizeTest() {
...
+    return;
+  }
+
+  // Check Wave support
+  if (!DoesDeviceSupportWaveOps(pDevice)) {
+    // Optional feature, so it's correct to not support it if declared as such.
+    WEX::Logging::Log::Comment(L"Device does not support wave operations.");
+    return;
```
tex3d, line 13366: This should not happen here, given the previous check when creating the device with SM 6.6.

In any case, I think we really should be separating the tests into two test groups, since one is required for devices that support SM 6.6 and the other for devices that support SM 6.8.  These should be different requirements in the HLK metadata and should include different tests in the playlist as a result.

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
```diff
@@ -2487,6 +2487,17 @@
       </RootValues>
       <Shader Name="CS" Target="cs_6_6">
           <![CDATA[// Shader source code will be set at runtime]]>
+      </Shader>      
+  </ShaderOp>>
+  
+  <ShaderOp Name="WaveSizeRangeTest" CS="CS">
```
tex3d, line 2493: There doesn't appear to be any reason to have this copy beyond the WaveSizeTest template here.  You shouldn't need an independent entry unless there are independent configuration requirements defined in this file.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,173 @@ void ExecutionTest::WaveSizeTest() {
...
+  for (UINT minShaderWaveSize = 4; minShaderWaveSize <= maxWaveSize;
+       minShaderWaveSize *= 2) {
+    for (UINT maxShaderWaveSize = minShaderWaveSize * 2;
+         maxShaderWaveSize <= 128; maxShaderWaveSize *= 2) {
+      for (UINT prefShaderWaveSize = minShaderWaveSize;
+           prefShaderWaveSize <= maxShaderWaveSize; prefShaderWaveSize *= 2) {
+
+        // For now, only allow valid shader wave ranges
+        // TODO: Create callback function for CreateComputePipelineState to
+        // handle expected errors.
```
tex3d, line 13253: A test for the runtime API validation should be separate from this test, so we should simply skip any incompatible sizes without needing to comment about them in the log.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,173 @@ void ExecutionTest::WaveSizeTest() {
...
+
+        for (unsigned i = 0; i < MAX_WAVESIZE; i++) {
+          VERIFY_IS_TRUE(pOutData[i].count >= minWaveSize &&
+                         pOutData[i].count <= maxWaveSize);
+
+          if (prefShaderWaveSize >= minWaveSize &&
+              prefShaderWaveSize <= maxWaveSize) {
+
+            VERIFY_ARE_EQUAL(pOutData[i].count, prefShaderWaveSize);
+          }
```
tex3d, line 13313: Seems like the first VERIFY should be in an else clause when not checking `prefShaderWaveSize`.
Also, we should use the appropriate comparison VERIFY macros so you see value being compared when it fails.

```suggestion
          unsigned count = pOutData[i].count;
          if (prefShaderWaveSize >= minWaveSize &&
              prefShaderWaveSize <= maxWaveSize) {
            VERIFY_ARE_EQUAL(count, prefShaderWaveSize);
          } else {
            VERIFY_IS_GREATER_THAN_OR_EQUAL(count, minWaveSize);
            VERIFY_IS_LESS_THAN_OR_EQUAL(count, maxWaveSize);
          }
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,173 @@ void ExecutionTest::WaveSizeTest() {
...
+      "RWStructuredBuffer<TestData> data : register(u0); \r\n"
+      "\r\n"
+      "// Note: The 3 WAVESIZE variables below will be defined via compiler "
+      "option -D\r\n"
+      "[wavesize(MINWAVESIZE, MAXWAVESIZE, PREFWAVESIZE)]\r\n"
+      "[numthreads(" stringify(MAX_WAVESIZE) "*2,1,1)]\r\n"
+                                             "void main(uint3 tid : "
+                                             "SV_DispatchThreadID ) { \r\n"
+                                             "  data[tid.x].count = "
+                                             "WaveGetLaneCount(); \r\n"
```
tex3d, line 13237: What's the value of writing the WaveGetLaneCount() value from every thread?  It should be a constant value from the device, and could be written to one location by one thread ` if (WaveIsFirstLane())`.  Same for the single-value test.

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
bob80905: The extra configuration requirement is setting Target="cs_6_8"
I need to make it separate because of the nature of how the source is generated (wavesize parameters being defined via compile define flags).

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13236,6 +13205,173 @@ void ExecutionTest::WaveSizeTest() {
...
+  // format shader source
+  const char waveSizeTestShader[] =
+      "struct TestData { \r\n"
+      "  uint count; \r\n"
+      "}; \r\n"
+      "RWStructuredBuffer<TestData> data : register(u0); \r\n"
+      "\r\n"
+      "// Note: The 3 WAVESIZE variables below will be defined via compiler "
+      "option -D\r\n"
+      "[wavesize(MINWAVESIZE, MAXWAVESIZE, PREFWAVESIZE)]\r\n"
```
tex3d, line 13232: Shouldn't we test the form with no preferred size specified at all?

tex3d:
> For future reference, the hardware I tested on only allowed waverange size of [32,32]. So I could only test wave ranges that included 32. CreateComputePipelineState() fails if the shader expresses a waverange size outside of what's required, so a proposal to test all possible combinations between 4,128 would be to create a callback function that runs CreateComputePipelineState(), and encapsulates it in a try/catch. Then, we can expect the error that is thrown when CreateComputePipelineState() is run with incompatible wavesize, and have full test coverage.

This additional testing described for `CreateComputePipelineState()` validation belongs in a different test.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -13188,11 +13158,11 @@ void ExecutionTest::WaveSizeTest() {
       "\r\n"
       "// Note: WAVESIZE will be defined via compiler option -D\r\n"
       "[wavesize(WAVESIZE)]\r\n"
-      "[numthreads(" strinfigy(
-          MAX_WAVESIZE) "*2,1,1)]\r\n"
-                        "void main(uint3 tid : SV_DispatchThreadID ) { \r\n"
-                        "  data[tid.x].count = WaveActiveSum(1); \r\n"
-                        "}\r\n";
+      "[numthreads(1,1,1)]\r\n"
```
tex3d, line 13161: I suppose with the different approach of using `WaveGetLaneCount()`, a group size of 1 is _technically_ sufficient to output the value from the first lane, however, it is a degenerate case.  While this hopefully won't be handled any differently than other dimensions, it's pretty likely that group sizes do impact the picking of wave sizes for drivers.  One could argue that you shouldn't really request a wave size larger than your thread group size anyway, since how would you possibly use that larger wave size?

I think the point of my rambling is that it might be better to test with a threadgroup size that's a multiple of the largest possible wave size of 128.

tex3d: And if doing so, you'll want to use comparison with the SV_DispatchThreadID to only write a result out on ID (0,0,0).

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
```diff
@@ -2487,7 +2487,7 @@
       </RootValues>
       <Shader Name="CS" Target="cs_6_6">
           <![CDATA[// Shader source code will be set at runtime]]>
-      </Shader>
+      </Shader>
```
tex3d, line 2490: Just added whitespace at end of line now, should probably just revert.

tex3d (approved):

PR #6351, opened 2024-02-22
PIX: Emit correct input sig elements and view ID state into PSV0 when input sigs are changed by debug instrumentation

jeffnn:
Two intertwined issues:

1:
The shader-debug-instrumentation and pixel-hit-counter passes add SV_VertexId+SV_InstanceId to the input sig for a VS if they are not present, and SV_Position for the PS. The changes in PixPassHelpers.cpp centralize the SV_Position case and fix up a few missing fields whose absence was breaking PIX shader debugging on WARP. The calller now has to pass the upstream shader's SV_Position row, if it exists. (PIX has been updated to do this.)

2:
However, independent of these tweaks, adding these new values means that the view id metadata would be incorrect.  A long-standing assert would fire, but is herein fixed, since the above work exacerbated this problem. The assert in question is this one in CopyViewIDStateForOutputToPSV that previously fired when a debug-instrumented shader was subsequently wrapped into a container:
```     DXASSERT_NOMSG((InputScalars <= IOTable.InputVectors * 4) &&
                   (IOTable.InputVectors * 4 - InputScalars < 4)); 
```
(InputScalars, which comes from the serialized state in PSV0, was too small, since it did not include the added system values.)

To fix up these data, the caller now has to invoke the generate-view-id-state pass, and follow that with the emit-metadata pass, since emitting the metadata in the debug pass itself is now too early (and that call has been deleted). (PIX has been updated to do invoke these passes.)

These changes together are sufficient to allow PIX shader debugging to operate on WARP in those cases where these system-values were not included by the application, with correct view-id metadata, and the assert is happy.


lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -455,13 +458,13 @@ DxilDebugInstrumentation::addRequiredSystemValues(BuilderContext &BC,
...
     // If not present, we add it.
     if (Existing_SV_Position == InputElements.end()) {
+      unsigned int Index = static_cast<unsigned int>(InputElements.size());
       auto Added_SV_Position =
           llvm::make_unique<DxilSignatureElement>(DXIL::SigPointKind::PSIn);
       Added_SV_Position->Initialize("Position", hlsl::CompType::getF32(),
-                                    hlsl::DXIL::InterpolationMode::Linear, 1,
-                                    4);
+                                    hlsl::DXIL::InterpolationMode::Linear, 1, 4,
+                                    Index, 0);
```
tex3d, line 466: I think there are potential problems with this:

Index is being passed to the `StartRow` parameter, which is supposed to the be the packing location starting row, not the index of the element.  This row must match the upstream vertex shader's output row for `SV_Position`, whatever that is.

You should not be relying on default parameter values for `ID` and `IndexVector` when constructing an element for final DXIL.  These initial states are meant for an earlier phase, before element flattening, index assignment, and signature packing.

lib/DxilPIXPasses/DxilPixEmitResources.cpp
```diff
@@ -0,0 +1,43 @@
...
+using namespace llvm;
+using namespace hlsl;
+
+class EmitDxilResourceToMeta : public ModulePass {
+public:
+  static char ID;
+  explicit EmitDxilResourceToMeta() : ModulePass(ID) {}
+  bool runOnModule(Module &M) override {
+    DxilModule &DM = M.GetOrCreateDxilModule();
+    DM.ReEmitDxilResources();
```
tex3d, line 31: I find the name of this is misleading.  Originally, when named, it was for the purpose of re-emitting resource metadata, but now it re-emits *all* metadata.  The same issue extends to this pass name and the descriptions of the pass.  I think it all should be changed to indicate that all metadata is being re-emitted.  The change could be made first in just the names and descriptions around this pass, with a follow-up to change the name of the method in DxilModule, since that will have broader impact.  Or it all could be done at the same time, if it's not too much of a pain.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -396,11 +396,14 @@ static unsigned FindOrAddInputSignatureElement(
 
   if (ExistingElement == InputElements.end()) {
     auto AddedElement = llvm::make_unique<DxilSignatureElement>(sigPointKind);
+    unsigned int Index = static_cast<unsigned int>(InputElements.size());
     AddedElement->Initialize(name, hlsl::CompType::getF32(),
-                             hlsl::DXIL::InterpolationMode::Undefined, 1, 1);
+                             hlsl::DXIL::InterpolationMode::Constant, 1, 1,
+                             Index, 0);
```
tex3d, line 402: Same comment as for `SV_Position`, but depending on the SemanticKind interpretation at this SigPoint, packing location may be required (and must match connecting stage), or may need to be unassigned.

I see the existing code only used this to add VertexID and InstanceID to `VSIn` SigPoint.  `VSIn` represents a special case, handled by the input assembler, where the input row matches the element index in the signature, and elements are not packed as they would be in other stages.  You also lucked out on not breaking anything you were relying on when not supplying the `IndexVector` (which will break some things).  That's why this approach appeared to work for these values.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
tex3d: By the way, `ID` and `IndexVector` are easy to supply, they are the index of the element in the signature (`InputElements.size()`) and a single-element vector containing the single semantic index for your 1-row element, which for each of your use cases is `0`.

tex3d (changes requested):
The main issue I have is the incorrect initialization of added signature elements.  Can there also be a test for added signature elements which ensures the metadata is as expected?

I suspect we have a validator gap here, since an empty `IndexVector` shouldn't have been considered valid.

lib/DxilPIXPasses/DxilPixEmitResources.cpp
jeffnn: Changed the pass name. Will change the fn in later PR as suggested.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
jeffnn: I think this latest change does it... I have some PIX-side tests that do the round-trip from the new input sig reader pass back to the debug instrumentation pass.

jeffnn:
> The main issue I have is the incorrect initialization of added signature elements. Can there also be a test for added signature elements which ensures the metadata is as expected?
> 
> I suspect we have a validator gap here, since an empty `IndexVector` shouldn't have been considered valid.

I think it's good now. A tale that grew in the telling...

lib/DxilPIXPasses/DxilAddPixelHitInstrumentation.cpp
```diff
@@ -32,22 +32,26 @@ class DxilAddPixelHitInstrumentation : public ModulePass {
...
   bool runOnModule(Module &M) override;
+  std::string m_upstreamSVPositionIndices;
 };
 
 void DxilAddPixelHitInstrumentation::applyOptions(PassOptions O) {
   GetPassOptionBool(O, "force-early-z", &ForceEarlyZ, false);
   GetPassOptionBool(O, "add-pixel-cost", &AddPixelCost, false);
   GetPassOptionInt(O, "rt-width", &RTWidth, 0);
   GetPassOptionInt(O, "num-pixels", &NumPixels, 0);
-  GetPassOptionInt(O, "sv-position-index", &SVPositionIndex, 0);
```
jeffnn, line 50: The previous index option wasn't used!

lib/DxilPIXPasses/DxilPixEmitMetadata.cpp
```diff
@@ -0,0 +1,43 @@
...
+using namespace hlsl;
+
+class DxilPixEmitMetadata : public ModulePass {
+public:
+  static char ID;
+  explicit DxilPixEmitMetadata() : ModulePass(ID) {}
+  bool runOnModule(Module &M) override {
+    DxilModule &DM = M.GetOrCreateDxilModule();
+    DM.ReEmitDxilResources();
+    return true;
```
jeffnn, line 32: Is "true" appropriate here, given that only the metadata is changed?

tex3d: I believe that counts!

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -397,15 +402,14 @@ static unsigned FindOrAddInputSignatureElement(
...
+    auto AddedElement =
+        llvm::make_unique<DxilSignatureElement>(DXIL::SigPointKind::VSIn);
     unsigned int Index = static_cast<unsigned int>(InputElements.size());
-    AddedElement->Initialize(name, hlsl::CompType::getF32(),
-                             hlsl::DXIL::InterpolationMode::Constant, 1, 1,
-                             Index, 0);
-    AddedElement->AppendSemanticIndex(0);
+    AddedElement->Initialize(
+        hlsl::Semantic::Get(semanticKind)->GetName(), hlsl::CompType::getU32(),
+        hlsl::DXIL::InterpolationMode::Constant, 1, 1, Index, 0, Index, {0});
```
tex3d, line 410: I think this `FindOrAddVSInSignatureElement` function sounds more generic than it is.  The first use of `Index` in the arguments here is only correct given a particular 1:1 layout between elements in the signature and packing row, which only applies to `VSIn`.  You could assert that the `sigPointKind == DXIL::SigPointKind::VSIn` and mention why in a comment, otherwise, if this assumption is violated, you could end up with weird problems.

tex3d: Actually, it still might not be true, since it's possible to have multi-row elements on input, which could cause the Index to overlap a packing location for another element.  So, perhaps you really should iterate through the elements and find the next available row to place this element into.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -450,6 +450,56 @@ void ReplaceAllUsesOfInstructionWithNewValueAndDeleteInstruction(
...
+                   });
+
+  // SV_Position, if present, has to have full mask, so we needn't worry
+  // about the shader having selected components that don't include x or y.
+  // If not present, we add it.
+  if (Existing_SV_Position == InputElements.end()) {
+
+    unsigned int Index = static_cast<unsigned int>(InputElements.size());
+
+    unsigned int StartRow = Index;
```
tex3d, line 472: `StartRow` should never just take the Index.  Is this path actually used, or is it always overridden by `UpStream` read below?  Even if we rely on it always being overridden, we should fail here somehow if not.  If it's possible you want to just add one without parsing from `UpStream`, then at least look for the next available row in the signature, rather than just taking the next signature element index.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -450,6 +450,56 @@ void ReplaceAllUsesOfInstructionWithNewValueAndDeleteInstruction(
...
+        RowCount = atoi(indices[2].str().c_str());
+        ColumnCount = atoi(indices[3].str().c_str());
+      }
+    }
+
+    auto Added_SV_Position =
+        llvm::make_unique<DxilSignatureElement>(DXIL::SigPointKind::PSIn);
+    Added_SV_Position->Initialize("Position", hlsl::CompType::getF32(),
+                                  hlsl::DXIL::InterpolationMode::Linear,
+                                  RowCount, ColumnCount, StartRow, StartColumn);
```
tex3d, line 492: Still need to add the ID, and if `RowCount > 1`, then you'll need to append sequentially incrementing index... but how would that be?  If that should never be the case, then why is the option available in `UpStream` that you're parsing?  If `UpStream` is more generic than for `SV_Position`, then we should fail on unexpected scenarios, like `RowCount > 1`.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -450,6 +450,56 @@ void ReplaceAllUsesOfInstructionWithNewValueAndDeleteInstruction(
...
+    Added_SV_Position->AppendSemanticIndex(0);
+    Added_SV_Position->SetKind(hlsl::DXIL::SemanticKind::Position);
+
+    auto index = InputSignature.AppendElement(std::move(Added_SV_Position));
+    return InputElements[index]->GetID();
```
tex3d, line 497: This ID was never set, so it's not going to be valid.  It should be the same as the `Index` computed previously, which should match `index` here.

lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -450,6 +450,56 @@ void ReplaceAllUsesOfInstructionWithNewValueAndDeleteInstruction(
...
+    unsigned int ColumnCount = 4;
+    if (!UpStream.empty()) {
+      SmallVector<StringRef, 4> indices;
+      UpStream.split(indices, "-");
+      if (indices.size() == 4) {
+        // These are encoded in ./DxilPixReadOutputSig.cpp
+        StartRow = atoi(indices[0].str().c_str());
+        StartColumn = atoi(indices[1].str().c_str());
+        RowCount = atoi(indices[2].str().c_str());
+        ColumnCount = atoi(indices[3].str().c_str());
```
tex3d, line 484: I think this is only valid if all of the following conditions are true, right?  Shouldn't we check for these and bail otherwise?
- `StartColumn == 0`
- `RowCount == 1`
- `ColumnCount == 4`

tex3d (commented):
I had some follow up comments, but think I'm finished reviewing this current state for now.

lib/DxilPIXPasses/PixPassHelpers.cpp
tex3d: Oh yeah, given that these all have to be these values, there's probably no value in passing anything but an int row index, similar to the prior option.

lib/DxilPIXPasses/DxilPixReadOutputSig.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "dxc/DxilPIXPasses/DxilPIXPasses.h"
+#include "dxc/DxilPIXPasses/DxilPIXVirtualRegisters.h"
+#include "dxc/HLSL/DxilGenerationPass.h"
+
+#include "llvm/IR/Module.h"
+
+using namespace llvm;
+using namespace hlsl;
+
+const char *SemanticKindName(DXIL::SemanticKind Kind) {
```
tex3d, line 25: Why not use `Semantic::Get(Kind)->GetName()` instead?  You don't have to keep a separate list up to date then.

tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -820,8 +1073,9 @@ static bool FindStructMemberFromStore(llvm::StoreInst *S,
   return false;
 }
 
-// This function lives in lib\DxilPIXPasses\DxilAnnotateWithVirtualRegister.cpp
-// Declared here so we can test it.
+// This function lives in
+// lib\DxilPIXPasses\DxilAnnotateWithVirtualRegister.cpp Declared here so we
+// can test it.
```
tex3d, line 1078: These don't look like what clang-format would have done, since they were previously 79 characters long, so why the changes here?

lib/DxilPIXPasses/DxilPixReadOutputSig.cpp
```diff
@@ -0,0 +1,128 @@
...
+public:
+  static char ID;
+  explicit DxilPixReadOutputSig() : ModulePass(ID) {}
+  bool runOnModule(Module &M) override {
+    if (OSOverride != nullptr) {
+      DxilModule &DM = M.GetOrCreateDxilModule();
+      auto const &Meta = DM.GetOutputSignature();
+      auto const &Elements = Meta.GetElements();
+      for (auto const &Element : Elements) {
+        *OSOverride << "OutputSigElement:" << Element->GetName() << ":"
```
tex3d, line 108: I'm not sure where `OSOverride` comes from, how it's used, and I'm not sure why we have to run a pass just to get the row for `SV_Position` output in upstream shader.  You could read this from the output signature part from a compiled shader, for instance.  Do you have need for the rest of the signature elements here?

lib/DxilPIXPasses/DxilPixReadOutputSig.cpp
```diff
@@ -0,0 +1,128 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DxilPixEmitMetadata.cpp                                                   //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Emit Dxil resource meta data                                              //
```
tex3d, line 8: Nit: Copy-pasted from different pass without changing filename or description.

lib/DxilPIXPasses/DxilPixEmitMetadata.cpp
```diff
@@ -0,0 +1,43 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DxilPixEmitResources.cpp                                                  //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Emit Dxil resource meta data                                              //
```
tex3d, line 8: Nit: Didn't update filename and description.

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: DxilSignature::AppendElement sets the ID to the previous size of the element array and returns that value. So yes I could just return index, but how it is is a bit more readable.

lib/DxilPIXPasses/DxilPixReadOutputSig.cpp
jeffnn: PIX needs to read the sig to get the row for SV_Position somehow so it can pass it back in to the debug instrumentation pass. This way keeps all the knowledge of the format of the signature in the compiler binary. 

I have a dastardly plan to use this also for mesh shader output visualization in PIX.

tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -2333,8 +2587,8 @@ float4 main(VS_OUTPUT_ENV input) : SV_Target
 }
 )";
 
-    // This is little more than a crash test, designed to exercise a previously
-    // over-active assert..
+    // This is little more than a crash test, designed to exercise a
+    // previously over-active assert..
```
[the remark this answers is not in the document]
jeffnn: Don't know how that happened!

lib/DxilPIXPasses/DxilPixReadOutputSig.cpp
jeffnn: Doh. Didn't find the GetName() function early in this process. Did find it later. Forgot to come back here and clean up.

tools/clang/unittests/HLSL/PixTest.cpp
jeffnn: (Probably from the clang-format setting in VS being global, and me switching back and forth between dxc and PIX and forgetting to toggle the setting.)

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: Fair point. At first I thought it was better to rely entirely on the read-back-signature pass to get all these values precisely as they were in the upstream output sig. But if it's really true that these system values always start at col 0, then it's better to restrict the input to the one value (start row).

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: As commented, the ID is added for you by AppendElement...

tex3d (commented):
One last thing.  Why won't `"hlsl-dxilemit"` work in place of adding the new DxilPixEmitMetadata?

lib/DxilPIXPasses/DxilPixEmitMetadata.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "dxc/DxilPIXPasses/DxilPIXPasses.h"
+#include "dxc/DxilPIXPasses/DxilPIXVirtualRegisters.h"
+#include "dxc/HLSL/DxilGenerationPass.h"
+
+#include "llvm/IR/Module.h"
+
+using namespace llvm;
+using namespace hlsl;
+
+class DxilPixEmitMetadata : public ModulePass {
```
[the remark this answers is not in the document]
jeffnn: It does indeed. Thanks for pointing it out.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -1515,8 +1511,6 @@ bool DxilDebugInstrumentation::RunOnFunction(Module &M, DxilModule &DM,
     }
   }
 
-  DM.ReEmitDxilResources();
```
jeffnn, line 1518: It is now the caller's (PIX's) responsibility to call the emit-metadata pass after invoking this pass.

austinkinross (approved):

adam-yang (approved):

tools/clang/test/HLSLFileCheck/pix/DebugBasic.hlsl
```diff
@@ -22,7 +22,12 @@
 // CHECK: and i32 
 // CHECK: or i32
 
-
+// Check that the correct metadata was emitted for the requested SV_Position at row 2, col 0, 1 row, 4 cols.
+// See DxilMDHelper::EmitSignatureElement for the meaning of these entries:
+//             ID                 TypeF32 SemKin Sem-Idx-Vec           interp  Rows Cols   Row    Col
+//              |                     |     |       |                    |      |     |      |     |
+// CHECK: !{i32 0, !"SV_Position", i8 9, i8 3, ![[SEMIDXVEC:[0-9]*]], i8 2, i32 1, i8 4, i32 2, i8 0, null}
```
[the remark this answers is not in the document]
adam-yang: Yep that's awesome.

PR #6353, opened 2024-02-22
[Sema] Allow int arguments for attributes to be enum values

### LINKED ISSUE #6337: [SPIR-V] REGRESSION: Getting lots of `warning: attribute `ext_instruction` must have a uint literal argument [-Wattribute-type]` when using enums
**Description**
I recently merged some upstream commits to my DXC branch

**Steps to Reproduce**
We use the SPIR-V Headers (the enums) to keep our SPIR-V Intrinsic headers legible like so
```cpp
[[vk::ext_builtin_input(spv::BuiltInHelperInvocation)]]
static const bool HelperInvocation;
```
where
```cpp
namespace spv
{

enum BuiltIn {
    BuiltInPosition = 0,
    BuiltInPointSize = 1,
    BuiltInClipDistance = 3,
    BuiltInCullDistance = 4,
    BuiltInVertexId = 5,
    BuiltInInstanceId = 6,
    BuiltInPrimitiveId = 7,
    BuiltInInvocationId = 8,
    BuiltInLayer = 9,
    BuiltInViewportIndex = 10,
    BuiltInTessLevelOuter = 11,
    BuiltInTessLevelInner = 12,
    BuiltInTessCoord = 13,
    BuiltInPatchVertices = 14,
    BuiltInFragCoord = 15,
    BuiltInPointCoord = 16,
    BuiltInFrontFacing = 17,
    BuiltInSampleId = 18,
    BuiltInSamplePosition = 19,
    BuiltInSampleMask = 20,
    BuiltInFragDepth = 22,
    BuiltInHelperInvocation = 23,
```

We can obviously fix by doing the following
```cpp
[[vk::ext_builtin_input(uint(spv::BuiltInHelperInvocation))]]
static const bool HelperInvocation;
```

but it would be nice for the compiler not to moan when the value can be converted to `uint` without any narrowing etc.

I tried using but DXC doesn't understand clang pragmas
```
#pragma clang diagnostic ignored "-Wattribute-type"
```

**Actual Behavior**

Lots of `-Wattribute-type` spam that wasn't there before we merged that commit range.

**Environment**
- DXC version trunk/main as of 21st of February 2024

cassiebeckley:
The SPIR-V backend uses `ValidateAttributeIntArg` for the `vk::ext_builtin_input` and `vk::ext_builtin_output` attrs, as well as some others. It is convenient for users to be able to pass in an enum value to this attribute, for example:

```
enum class BuiltIn {
  HelperInvocation = 23
};

[[vk::ext_builtin_input(BuiltIn::HelperInvocation)]]
static const bool gl_HelperInvocation;
```

If you think `ValidateAttributeIntArg` *should* warn if a user passes in an enum value, I'll create a separate function for validating `vk::ext_builtin_input` instead.

Fixes #6337

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/spv.inline.builtin.enum.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/spv.inline.builtin.enum.hlsl
@@ -1,0 +1,19 @@
+// RUN: %dxc -T cs_6_0 -E main -fcgl  %s -spirv 2>&1 | FileCheck %s
+
+// CHECK-NOT: warning
+// CHECK: OpDecorate %gl_HelperInvocation BuiltIn HelperInvocation
+
+enum class BuiltIn {
+  HelperInvocation = 23
+};
+
+[[vk::ext_builtin_input(BuiltIn::HelperInvocation)]]
+static const bool gl_HelperInvocation;
+
+uint square_x(uint3 v) {
+  return v.x * v.x;
+}
+
+[numthreads(32,1,1)]
+void main() {
+}
--- a/tools/clang/test/CodeGenSPIRV/spv.inline.builtin.enum.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/spv.inline.builtin.enum.hlsl
@@ -1,7 +1,6 @@
-// RUN: %dxc -T cs_6_0 -E main -fcgl  %s -spirv 2>&1 | FileCheck %s
+// RUN: %dxc -T cs_6_0 -E main -fcgl  %s -spirv -verify
 
-// CHECK-NOT: warning
-// CHECK: OpDecorate %gl_HelperInvocation BuiltIn HelperInvocation
+// expected-no-diagnostics
 
 enum class BuiltIn {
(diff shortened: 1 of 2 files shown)


tools/clang/test/CodeGenSPIRV/spv.inline.builtin.enum.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: %dxc -T cs_6_0 -E main -fcgl  %s -spirv 2>&1 | FileCheck %s
+
+// CHECK-NOT: warning
```
sudonatalie, line 3: Maybe add a `-verify` RUN line to check there's no unexpected errors/warnings?

sudonatalie (approved):
This seems reasonable to me, as long as someone from MS approves.

pow2clk (approved):
Makes sense to me, I agree that a verify run would be the more appropriate way to test this.

llvm-beanz (approved):

bob80905 (approved):

PR #6355, opened 2024-02-24
dxexp: add support for mesh shaders, derivatives in mesh/amp shaders and wave matrix

hekota:
These features were not included in the dxexp output and they are important to query for HLK test reports.


tex3d (approved):
Looks good.  Have you tried this with AgilitySDK headers?  Our way of determining whether something is defined with `NTDDI_*` is based on the Windows SDK headers, but things may be defined in AgilitySDK which aren't defined there.

I suppose that will only matter if building against an older Windows SDK that doesn't support one of these, and using Agility SDK headers that do (then you get duplicate definitions).

pow2clk (approved):
Looks good!

damyanp (approved):
@hekota - can you have a look at Tex's comment about trying with Agility SDK headers before completing please?

PR #6356, opened 2024-02-24
Add -fdiagnostics-format-msvc to output diagnostics in MSCV mode

jonasmr:


jonasmr:

jonasmr:
I just realized the suggestion from @llvm-beanz  that I just merged breaks everything.
Closing while I fix.

PR #6363, opened 2024-02-28
sema: check input patch size mismatch earlier.

### LINKED ISSUE #3739: [SPIR-V] No error when Hull and patchconstantfunc input point counts don't match
### Title
[SPIR-V] No error when Hull and patchconstantfunc input point counts don't match (errors with DXIL backend)

### Functional impact
The InputPatch point count argument is ignored for the patch constant function. This will result in undetected/silent errors, when a shader writer might believe they modified the input point count but actually don't.

### Minimal repro steps

[repro_shader.txt](https://github.com/microsoft/DirectXShaderCompiler/files/6415304/repro_shader.txt)

### Expected result
Compilation should fail with error as produced by the DXIL backend:
```
repro_shader.txt:7:14: error: Patch constant function's input patch input should have 3 elements, but has 16.
HullPatchOut HullConst (InputPatch<ControlPoint,16> v) { return (HullPatchOut)0; }
             ^
```
### Actual result
Compilation completes with the point count in the patch constant function being ignored (cannot find the patch function's input point count in the generated SPIR-V)

### Further technical details

Keenuts:
The check was done in the DXIL backend, and not in the SPIR-V backend. This causes the issue to be ignored on the SPIR-V side. Moved the check earlier to SEMA so a single check works for both backends.

I had to split the existing HLSL check as it doesn't support having 2 run lines (AFAIK).

Fixes #3739

### DIFF
--- a/tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
+++ b/tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
@@ -1,6 +1,7 @@
 // RUN: %dxc -E main -T hs_6_0  %s | FileCheck %s
 
-// CHECK-DAG: Patch constant function's input patch input should have 3 elements, but has 12.
+// CHECK-DAG: Patch Constant function HSPerPatchFunc should not have inout param
+// CHECK-DAG: Patch constant function's output patch input should have 3 elements, but has 5.
 
 //--------------------------------------------------------------------------------------
 // SimpleTessellation.hlsl
@@ -44,7 +45,7 @@ float4 HSPerPatchFunc()
     return 1.8;
 }
 
-HSPerPatchData HSPerPatchFunc( const InputPatch< PSSceneIn, 12 > points, OutputPatch<HSPerVertexData, 3> outp)
+HSPerPatchData HSPerPatchFunc( const InputPatch< PSSceneIn, 12 > points, OutputPatch<HSPerVertexData, 5> outp, inout float x)
 {
     HSPerPatchData d;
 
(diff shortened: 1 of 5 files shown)


tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// hull per-control point shader
+[domain("tri")]
+[partitioning("fractional_odd")]
+[outputtopology("triangle_cw")]
+[patchconstantfunc("HSPerPatchFunc")]
+[outputcontrolpoints(3)]
+HSPerVertexData main(const uint id : SV_OutputControlPointID,
+                      const InputPatch< PSSceneIn, 12 > points )
```
bob80905, line 67: Would it not be possible to change the input patch number here, to catch both diagnostics at the same time with one run line?

Keenuts: When I tried it didn't work, as the error checked in the other file makes the compile process abort earlier, and the 2 other messages don't show.

sudonatalie: It seems a bit odd that these are caught at different points now so I'd probably want to move all 3 checks to Sema, or at least track the remaining 2 in a follow-up bug, but I'll leave it to MS folks to decide.

Keenuts: Created 2 issues to track those. One is handled by the SPIR-V backend, but weirdly, the other also causes us to accept bad code.

bob80905 (approved):

PR #6368, opened 2024-02-28
Fix auto-dxilver logic in FileCheckerTest

### LINKED ISSUE #6367: FileCheckerTest incorrectly skips validator version check when -Vd specified
**Description**
In FileCheckerTest.cpp, the -T target of a %dxc part is checked for compatibility. However, if you use -Vd, it skips the version check for the validator. This isn't quite right since the validator version will be set from the attached validator, unless you explicitly set -validator-version or use -select-validator internal. Also, if -validator-version is explicitly set, we should check the validator against that version, not just the one based on shader model.

**Steps to Reproduce**
Run downlevel validator test pass with test that uses a newer shader target such as lib_6_8 with the -Vd option.

**Actual Behavior**
Test is run with a validator version set from the external validator that does not support the shader model.

**Environment**
- DXC version: latest main
- Host Operating System: any

tex3d:
In FileCheckerTest.cpp, the -T target of a %dxc part is checked for compatibility. However, if you use -Vd, it skips the version check for the validator. This isn't quite right since the validator version will be set from the attached validator, unless you explicitly set -validator-version or use -select-validator internal. Also, if -validator-version is explicitly set, we should check the validator against that version, not just the one based on shader model.

This change fixes the initial check logic to only skip validator check when using -select-validator internal or -validator-version. It also checks against an explicitly requested -validator-version when an external validator will be used.

Fixes #6367

### DIFF
--- a/tools/clang/test/HLSLFileCheck/infra/auto-dxilver.hlsl
+++ b/tools/clang/test/HLSLFileCheck/infra/auto-dxilver.hlsl
@@ -1,0 +1,28 @@
+// Make sure each run line either works, or implicitly requires dxilver 1.8.
+// If an implicit dxilver check is missing, we should see one of these errors.
+
+// RUN: %dxc -T vs_6_8 %s | FileCheck %s
+// This should implicitly require dxilver 1.8.
+
+// RUN: %dxc -T vs_6_8 -Vd %s | FileCheck %s
+// Even though this is using -Vd, the validator version is set by the available
+// validator.  If that isn't version 1.8 or above, we'll see an error.
+// The implicit dxilver logic should not skip the check when -Vd is used.
+// CHECK-NOT: error: validator version {{.*}} does not support target profile.
+
+// RUN: %dxc -T vs_6_0 -validator-version 1.8 %s | FileCheck %s
+// Even though target is 6.0, the explicit -validator-version should add an
+// implicit dxilver 1.8 requirement.
+// CHECK-NOT: error: The module cannot be validated by the version of the validator currently attached.
+
+// This error would occur if run against wrong compiler.
+// CHECK-NOT: error: invalid profile
+
+// Catch any other unexpected error cases.
+// CHECK-NOT: error
+
+// RUN: %dxc -T vs_6_8 -select-validator internal %s | FileCheck %s
+// This should always be run, and always succeed.
+// CHECK: define void @main()
+
+void main() {}
--- a/tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
+++ b/tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
@@ -518,12 +518,32 @@ FileRunCommandPart::RunDxc(dxc::DxcDllSupport &DllSupport,
     if (RequiredDxilMinor != 0xF && stage.compare("rootsig") != 0) {
       // Convert stage to minimum dxil/validator version:
       RequiredDxilMajor = std::max(RequiredDxilMajor, (unsigned)6) - 5;
+
+      bool bInternalValidator =
+          opts.SelectValidator == hlsl::options::ValidatorSelection::Internal;
+      bool bValVerExplicit = opts.ValVerMajor != UINT_MAX;
+
+      // Normally we must check the validator version as well, but there are
+      // two scenarios where the validator version doesn't need to be checked
+      // against the version based on the shader model:
+      // 1. The test selects internal validator.
+      // 2. The test explicitly requests a specific validator version.
       FileRunCommandResult result =
           CheckDxilVer(DllSupport, RequiredDxilMajor, RequiredDxilMinor,
-                       !opts.DisableValidation);
-      if (result.AbortPipeline) {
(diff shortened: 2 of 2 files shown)


pow2clk (approved):
I believe that the lack of support for loading external test validators prevents us from committing this with any testing really, but I think we have some internal testing that will allow it. Is that correct? Will there be testing there? The state of #5952 doesn't currently allow lower version validators, but it could. Do you think we should polish that off and get it in for the sake of this?

python3kgae:
Does lit FileCheck with things like
```
 REQUIRES: dxil-1-8 
```
have the same issue?

damyanp:
> Will there be testing there? 

Seems there should be some test coverage for this.  If we can't test is then that presumably means that there's no observable change from this at all?

tex3d:
@pow2clk
> I believe that the lack of support for loading external test validators prevents us from committing this with any testing really, but I think we have some internal testing that will allow it. Is that correct? Will there be testing there? The state of #5952 doesn't currently allow lower version validators, but it could. Do you think we should polish that off and get it in for the sake of this?

As you say, #5952 wouldn't help in its current state.  I'm not sure how we would add testing for lower versions, other than perhaps building multiple validators and versioning them in this way.  That seems like overkill for testing this mechanism in our FileCheckerTest.  I could add a test that doesn't necessarily trip this scenario in CI testing here, but would verify that this mechanism works upon integration testing internally.  Would that be acceptable instead?

@damyanp
> Seems there should be some test coverage for this. If we can't test is then that presumably means that there's no observable change from this at all?

There's an observable change in our internal downlevel validator coverage upon integration.  I can add a test that would trigger a failure there without the fix.

@python3kgae
> Does lit FileCheck with things like
> 
> ```
>  REQUIRES: dxil-1-8 
> ```
> 
> have the same issue?

No, this does not have the same issue.

---

I will add a test that would fail our internal integration testing.  Hopefully that is sufficient.

tools/clang/unittests/HLSLTestLib/FileCheckerTest.cpp
```diff
@@ -518,12 +518,32 @@ FileRunCommandPart::RunDxc(dxc::DxcDllSupport &DllSupport,
...
+      // Normally we must check the validator version as well, but there are
+      // two scenarios where the validator version doesn't need to be checked
+      // against the version based on the shader model:
+      // 1. The test selects internal validator.
+      // 2. The test explicitly requests a specific validator version.
       FileRunCommandResult result =
           CheckDxilVer(DllSupport, RequiredDxilMajor, RequiredDxilMinor,
-                       !opts.DisableValidation);
-      if (result.AbortPipeline) {
+                       !(bInternalValidator || bValVerExplicit));
```
python3kgae, line 533: Why not check  opts.DisableValidation here along with bInternalValidator || bValVerExplicit?

tex3d: The assumption of the original code that it didn't need to check the validator version when `-Vd` was used was incorrect.

That's because `opts.DisableValidation` (or `-Vd`) doesn't disable the behavior of setting the validator version based on the attached validator.  This is the issue hit by tests.  They had used `-Vd` with a SM 6.8 target and were still run when an older validator was attached, which initialized the validator version to one that didn't support the shader model, causing the tests to fail.

However, these other conditions were not checked.  If you select the internal validator, the attached one is irrelevant.  If you explicitly set the version, it needs to be checked against that instead, which is done next.

tools/clang/test/HLSLFileCheck/infra/auto-dxilver.hlsl
```diff
@@ -0,0 +1,28 @@
...
+// If an implicit dxilver check is missing, we should see one of these errors.
+
+// RUN: %dxc -T vs_6_8 %s | FileCheck %s
+// This should implicitly require dxilver 1.8.
+
+// RUN: %dxc -T vs_6_8 -Vd %s | FileCheck %s
+// Even though this is using -Vd, the validator version is set by the available
+// validator.  If that isn't version 1.8 or above, we'll see an error.
+// The implicit dxilver logic should not skip the check when -Vd is used.
+// CHECK-NOT: error: validator version {{.*}} does not support target profile.
```
bogner, line 11: Do the run lines need `2>&1` to make sure that FileCheck sees stderr for these to work?

bogner (approved):

tools/clang/test/HLSLFileCheck/infra/auto-dxilver.hlsl
tex3d: The point of this test is to verify a behavior that only exists in the TAEF FileCheckerTest.  It would never make sense to run this as a lit shell test.

One of the differences for FileCheckerTest is that stderr is passed to the next part automatically when the operation fails, and `2>&1` is not handled in any special way - instead, it will be passed to the compiler API and parsed as if it's a regular argument, causing it to set the `InputFile` in `hlsl::options::DxcOpts` to `"2>&1"`.

Another difference is that the lit shell test would never even get past the dxc part in any of these failure cases to even run the FileCheck in the first place.

PR #6369, opened 2024-02-29
Add dxilver to tests using -Vd to work around auto-dxilver bug

tex3d:
Issue #6367 identifies a problem with the auto-dxilver logic based on -T target option in %dxc part when using FileCheckerTest.cpp.

This change adds a manual %dxilver 1.8 to two tests that run into this problem, as a workaround.


pow2clk (approved):
Appreciate the simpler solution for the sake of the release branch

dmpots (approved):

bob80905 (approved):

PR #6373, opened 2024-02-29
Fixed invalid debug metadata for LexicalBlock in include files.

adam-yang:
No longer normalizing paths that are baked into `DIFile` (as well as the debug module file list and PDB source info).

The previous path normalization change introduced a bug in debug info where instead of generating a normal `DILexicalBlock`, we generated a `DILexicalBlockFile`. This was because the file paths written into the `DIFile`'s were normalized, but the file list in `SourceManager` was left unchanged. `CGDebugInfo::setLocation` checks whether the `DIFile` path matches the current file, and due to the normalization differences, assume the file has changes and compensates by generating a new `DILexicalBlockFile` on the scope stack.

An alternative to this PR would be to change `CGDebugInfo::setLocation` to normalize the `SourceManager` path before comparing; however, there may be other unknown code paths that could have similar issues. Another alternative is to normalize the path list in `SourceManager`; however, this has other ramifications and require many more drastic changes.

By no longer normalizing the paths written into `DIFile`, we revert to the behaviour where the paths in `DIFile` come directly from `SourceManager`, so there is no chance of a mismatch. Furthermore, by avoiding normalizing any paths that are written to PDBs, we do not bake any potential normalization bugs into any test content.

### DIFF
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -3088,14 +3088,6 @@ TEST_F(CompilerTest, CompileWhenAllIncludeCombinations) {
   std::string commonIncludeFile = "float foo() { return 10; }";
 
   TestCase tests[] = {
-
-      {L"\\my_path\\main.hlsl",
-       {L"\\my_path\\main.hlsl",
-        R"(#include "\my_include_dir\\second_dir/include.h" // <-- Network path
-           float main() : SV_Target { return foo(); } )"},
-       {L"\\my_include_dir\\second_dir\\include.h", commonIncludeFile},
-       {}},
-
       {L"main.hlsl",
        {L".\\main.hlsl",
         R"(#include "include.h"


tools/clang/tools/dxcompiler/dxcshadersourceinfo.cpp
```diff
@@ -312,19 +312,21 @@ static std::vector<SourceFile> ComputeFileList(clang::CodeGenOptions &cgOpts,
...
           file.Content = it->second->getRawBuffer()->getBuffer();
           ret.push_back(file);
           assert(!bFoundMainFile &&
                  "otherwise, more than one file matches main filename");
           bFoundMainFile = true;
         } else {
-          std::string NormalizedBuf = hlsl::NormalizePath(it->first->getName());
-          filesMap[NormalizedBuf] = it->second->getRawBuffer()->getBuffer();
+          filesMap[Path.str()] =
+              it->second->getRawBuffer()->getBuffer(); // NOCHECKIN
```
dmpots, line 329: Is the `NOCHECKIN` comment still valid?

adam-yang: Removed.

damyanp (approved):

dmpots (approved):
LGTM

PR #6378, opened 2024-03-01
Implement -ftime-trace-granularity option to set time granularity value for the time profiler

### LINKED ISSUE #6372: TimeTraceProfiler is missing the configurable setting for TimeTraceGranularity
**Description**
It has been observed that the unittest run for ftime-trace in Xbox compiler
branch fails randomly.  This test uses TimeTraceProfiler and expects specific
backend timings to be present in the resulting JSON.

***Why?***
Debugging this shows that the version of TimeTraceProfiler from the llvm support
library is throwing out timings that have a < 500us duration. This causes
the unittest to not find expect timings for the run in the resulting JSON and fail..

**Steps to Reproduce**
Difficult to reproduce consistantly as it is depends on the load of the test machine.
Machines with a higher load delay the test along enough to pass making the timings
be slightly longer than 500us.  Fast test runs fail.

**LLVM-PROJECT contains the fix already**
The fix for this is already present and tested in the upstream llvm-project repo.  
The plan is to only bring over the specific change that adds this missing ability.
A full cherry pick cannot be done as other clang test files are in the commit, but
the other files can be brought over as-is.

This change is small and contained.  This is not fully upgrading TimeTraceProfiler
but just bringing in the bits that we require.

***From the llvm-project repo:***
'Time profiler: small fixes and optimizations' 
Author:		Anton Afanasyev <anton.a.afanasyev@gmail.com>
Date:		5 years ago (4/15/2019 2:02:47 PM)
Commit hash:	26536728591d5fdac373ef535ae122b873f73292

**Environment**

coopp:
This commit adds a new optional -ftime-trace-granularity option that is already implemented in llvm-project. This change is a surgical port of an existing feature from the upstream llvm-project repo into the DXC codebase.

The following commits in the llvm-project repo were copied and followed for this change.  Clean cherry picks were not possible due to the differences in repos like change of file locations and other dependant changes made in the repo.

*** Adds the granularity configuration setting ***
'Time profiler: small fixes and optimizations'
Commit: 26536728591d5fdac373ef535ae122b873f73292

*** Wires up the commandline option -ftime-trace-granularity to the TraceProiler code ***
'[Support] Fix `-ftime-trace-granularity` option'
Commit: 4fdcabf259c4ab94654e6cd5d95d0e0313159c70

Fixes #6372

### DIFF
--- a/lib/DxcSupport/HLSLOptions.cpp
+++ b/lib/DxcSupport/HLSLOptions.cpp
@@ -826,11 +826,12 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
     opts.TimeTrace = Args.getLastArgValue(OPT_ftime_trace_EQ);
   if (Args.hasArg(OPT_ftime_trace_granularity_EQ)) {
     if (Arg *A = Args.getLastArg(OPT_ftime_trace_granularity_EQ)) {
-      if (llvm::StringRef(A->getValue()).getAsInteger(10, opts.TimeTraceGranularity)) {
+      if (llvm::StringRef(A->getValue())
+              .getAsInteger(10, opts.TimeTraceGranularity)) {
         opts.TimeTraceGranularity = 500;
         errors << "Warning: Invalid value for -ftime-trace-granularity option "
-                  "specified, defaulting to " << opts.TimeTraceGranularity
-               << " microseconds.";
+                  "specified, defaulting to "
+               << opts.TimeTraceGranularity << " microseconds.";
       }
     }
   }
--- a/lib/DxcSupport/HLSLOptions.cpp
+++ b/lib/DxcSupport/HLSLOptions.cpp
@@ -824,6 +824,17 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
   opts.VerifyDiagnostics = Args.hasFlag(OPT_verify, OPT_INVALID, false);
   if (Args.hasArg(OPT_ftime_trace_EQ))
     opts.TimeTrace = Args.getLastArgValue(OPT_ftime_trace_EQ);
+  if (Args.hasArg(OPT_ftime_trace_granularity_EQ)) {
+    if (Arg *A = Args.getLastArg(OPT_ftime_trace_granularity_EQ)) {
+      if (llvm::StringRef(A->getValue()).getAsInteger(10, opts.TimeTraceGranularity)) {
+        opts.TimeTraceGranularity = 500;
+        errors << "Warning: Invalid value for -ftime-trace-granularity option "
+                  "specified, defaulting to " << opts.TimeTraceGranularity
+               << " microseconds.";
+      }
+    }
+  }
+
   opts.EnablePayloadQualifiers =
       Args.hasFlag(OPT_enable_payload_qualifiers, OPT_INVALID,
                    DXIL::CompareVersions(Major, Minor, 6, 7) >= 0);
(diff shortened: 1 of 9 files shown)


lib/Support/TimeProfiler.cpp
```diff
@@ -115,23 +122,23 @@ struct TimeTraceProfiler {
     // Emit totals by section name as additional "thread" events, sorted from
     // longest one.
     int Tid = 1;
-    std::vector<NameAndDuration> SortedTotals;
-    SortedTotals.reserve(TotalPerName.size());
-    for (const auto &E : TotalPerName) {
-      SortedTotals.push_back(E);
-    }
+    std::vector<NameAndCountAndDurationType> SortedTotals;
```
damyanp, line 125: Are all these changes required to add the granularity flag?  Kind of surprised at the amount of churn here to turn a hard coded 500 into something controlled by a command line option.

damyanp (changes requested):
I'm like to understand why there seems to be more changes in here than are strictly necessary.

lib/Support/TimeProfiler.cpp
coopp: I followed the changes in llvm-project.  The Granularity setting was added... but not connnected initially to any external configuration.  Then there were follow up PRs that added the ftime-trace-granularity option to the compiler to configure that.  That is where these changes come in.  The way options are connected differs in the llvm-project branch the the DXC branch, to I had to follow how DXC wires in the option.

coopp: I think all of the defaulting to 500us might be overkill.  I am open to better suggestions here.  A default is required if the option is not specified.

tools/clang/test/DXC/ftime-trace-granularity.hlsl
```diff
@@ -0,0 +1,7 @@
+// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace | FileCheck %s
+// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace=%t.json -ftime-trace-granularity=0
```
damyanp, line 2: I don't think there's anything actually testing that the option does anything?

Suggestion: try setting it to some super larger value which means that all the trace events get filtered out.

coopp: Will do!

lib/Support/TimeProfiler.cpp
damyanp: If that's hard to do then maybe just making the changes to the granularity checking and not modifying anything else would reduce the risk of this.

damyanp: My point is that this commit is doing much more than just adding that option.

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -824,6 +824,17 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
   opts.VerifyDiagnostics = Args.hasFlag(OPT_verify, OPT_INVALID, false);
   if (Args.hasArg(OPT_ftime_trace_EQ))
     opts.TimeTrace = Args.getLastArgValue(OPT_ftime_trace_EQ);
+  if (Args.hasArg(OPT_ftime_trace_granularity_EQ)) {
+    if (Arg *A = Args.getLastArg(OPT_ftime_trace_granularity_EQ)) {
+      if (llvm::StringRef(A->getValue()).getAsInteger(10, opts.TimeTraceGranularity)) {
```
pow2clk, line 829: Some of this could be simplified by using `getLastArgValue` upfront instead of checking for the arg, getting arg and then converting it to a stringref.

coopp:
I need to find a good way to run clang-format on these files.  I saw that a check failed which was related to the code not being properly formatted.

PR #6380, opened 2024-03-04
fix non-portable include path

ukai:
d3dcommon.h seems to be canonical path name in Win SDK

Windows Kits/10/Include/10.0.22621.0/um/d3dcommon.h

so use d3dcommon.h rather than D3Dcommon.h

### DIFF
--- a/lib/Support/MSFileSystemBasic.cpp
+++ b/lib/Support/MSFileSystemBasic.cpp
@@ -16,7 +16,7 @@
 
 #include "dxc/Support/WinIncludes.h"
 
-#include <D3Dcommon.h>
+#include <d3dcommon.h>
 #include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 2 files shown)


hekota (approved):
LGTM!

ukai:

bob80905 (approved):

philwo:
Is there anything left to do here to get this merged? :)

pow2clk:
Thanks for the contribution @ukai! While this is clearly a correct change, I was curious what made this necessary given that they are windows-specific files. That's why they were missed from previous efforts to make building portable.

philwo:
@pow2clk Thanks for asking :)

@ukai and I are working on Chromium's build system. We want to make it as easy as possible for developers to work on Chromium for all supported platforms, no matter what OS they use on their developer machine. Sometimes we stumble upon portability issues in a dependency used by our build that way, like this one or https://github.com/microsoft/DirectXShaderCompiler/pull/6379.

We have some docs for contributors who want to build Chromium for Windows from their Linux or Mac here, if you're interested: https://chromium.googlesource.com/chromium/src/+/master/docs/win_cross.md

PR #6381, opened 2024-03-04
Implement -ftime-trace-granularity option to set time granularity value for the time profiler

### LINKED ISSUE #6372: TimeTraceProfiler is missing the configurable setting for TimeTraceGranularity

coopp:
This commit adds a new optional -ftime-trace-granularity option that is already implemented in llvm-project. This change is a surgical port of an existing feature from the upstream llvm-project repo into the DXC codebase.

The following commits in the llvm-project repo were copied and followed for this change. Clean cherry picks were not possible due to the differences in repos like change of file locations and other dependant changes made in the repo.

*** Adds the granularity configuration setting ***
'Time profiler: small fixes and optimizations'
Commit: 26536728591d5fdac373ef535ae122b873f73292

*** Wires up the commandline option -ftime-trace-granularity to the TraceProfiler code ***
'[Support] Fix -ftime-trace-granularity option'
Commit: 4fdcabf259c4ab94654e6cd5d95d0e0313159c70

Fixes #6372

### DIFF
--- a/tools/clang/test/DXC/ftime-trace-granularity.hlsl
+++ b/tools/clang/test/DXC/ftime-trace-granularity.hlsl
@@ -1,0 +1,7 @@
+// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace | FileCheck %s
+// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace=%t.json -ftime-trace-granularity=0
+// RUN: cat %t.json | FileCheck %s
+
+// CHECK: { "traceEvents": [
+
+void main() {}
--- a/tools/clang/test/DXC/ftime-trace-granularity.hlsl
+++ b/tools/clang/test/DXC/ftime-trace-granularity.hlsl
@@ -2,6 +2,7 @@
 // RUN: %dxc -E main -T vs_6_0 %s -ftime-trace=%t.json -ftime-trace-granularity=99999
 // RUN: cat %t.json | FileCheck %s
 
+// This test runs both stdout and file output paths.
 // Validate that we do not output named stats
 // but still continue to output the Totals
 // versions of them which is expected.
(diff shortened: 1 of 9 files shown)


tools/clang/test/DXC/ftime-trace-granularity.hlsl
```diff
@@ -0,0 +1,12 @@
+// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace -ftime-trace-granularity=99999 | FileCheck %s
```
damyanp, line 1: Would it be possible to add a comment here explaining what the two test cases are?  I think that it is checking that stdout works and that file output works?  Is that the idea?

damyanp (approved):

tex3d (approved):
Looks good.

Note: there's a small typo in the PR description: `TraceProiler`

tools/clang/test/DXC/ftime-trace-granularity.hlsl
coopp: Yes.  That is the idea.  I can add some comments to the test.

damyanp (approved):

PR #6388, opened 2024-03-05
[SPIR-V] Evaluate void return expressions

### LINKED ISSUE #5464: [SPIR-V] Wrong code is generated when returning a function call
**Description**
We noticed recently that DXC does not create correct code in some cases.
I have created a simple reproducer for such a case.

**Steps to Reproduce**

```
struct InputParameters
{
	unsigned int myImageSizeX;
	unsigned int myImageSizeY;
};

ConstantBuffer<InputParameters> myInputData : register(b0);
RWStructuredBuffer<float4> myResultHdrBuffer : register(u1);

class RayMarchingResultData
{
	void init(const float4 aRgba_in)
	{
		myRgba = aRgba_in;
	}

	float4 myRgba;
};

void doRayMarchingOnBlock_Step(inout RayMarchingResultData aRayData_inout)
{
	aRayData_inout.myRgba = float4(1.0, 0.0, 0.0, 1.0);
}

void doRayMarchingOnBlock(inout RayMarchingResultData aRayData_inout)
{
	//doRayMarchingOnBlock_Step(aRayData_inout);
	return doRayMarchingOnBlock_Step(aRayData_inout);
}

[numthreads(8, 8, 1)]
void main(uint3 LaunchID : SV_DispatchThreadID, uint3 LocalInvocationID : SV_GroupThreadID)
{
	if ((LaunchID[0] >= myInputData.myImageSizeX) || (LaunchID[1] >= myInputData.myImageSizeY))
	{
		return;
	}

	uint aRayIndex = LaunchID[1] * myInputData.myImageSizeX + LaunchID[0];

	RayMarchingResultData aResult;
	aResult.init(float4(0.0, 0.0, 0.0, 1.0));

	doRayMarchingOnBlock(aResult);

	myResultHdrBuffer[aRayIndex] = aResult.myRgba;
}
```

**Actual Behavior**
When running this shader one will always get a wrong result image (image is black).
However, when replacing the line 
`	return doRayMarchingOnBlock_Step(aRayData_inout);`
with 
`	doRayMarchingOnBlock_Step(aRayData_inout);`
and hence removing the "return" instruction, the result is as expected (image is red).

This issue can also be reproduced when comparing the created SPIR-V code of those two cases.
The case that uses "return" creates wrong SPIR-V code.

**Environment**
- DXC version 
dxcompiler.dll: 1.7 - 1.7.0.3991 (b97f9b938)

- Host Operating System
Windows 10

sudonatalie:
Even when a function's return type is void, it is still valid to return an void expression, and that expression should be evaluated before the function is terminated to match DXIL behavior.

Although I don't see this explicitly mentioned in HLSL documention, the cpp reference states (abridged):

> `return expression;`
> Evaluates the expression, terminates the current function and returns the result of the expression to the caller, after implicit conversion to the function return type.
> In a function returning void, the return statement with expression can be used, if the expression type is void.

Fixes #5464

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -2633,6 +2633,9 @@ void SpirvEmitter::doReturnStmt(const ReturnStmt *stmt) {
                                    {stmt->getReturnLoc(), retVal->getLocEnd()});
     }
   } else {
+    if (retVal) {
+      loadIfGLValue(retVal);
+    }
     spvBuilder.createReturn(stmt->getReturnLoc());
   }
 
--- a/tools/clang/test/CodeGenSPIRV/cf.return.void.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/cf.return.void.hlsl
@@ -4,6 +4,10 @@ void A() {
 }
 
 void main() {
-  // CHECK: OpReturn
+  // CHECK: [[type:%[0-9]+]] = OpTypeFunction %void
+  // CHECK:        %src_main = OpFunction %void None [[type]]
+  // CHECK:      {{%[0-9]+}} = OpFunctionCall %void %A
+  // CHECK:                    OpReturn
+  // CHECK:                    OpFunctionEnd
   return A();
 }


cassiebeckley (approved):
LGTM

Keenuts (approved):

PR #6389, opened 2024-03-05
[SPIR-V] Disallow variable offsets in texture Load

### LINKED ISSUE #6149: [SPIR-V] Texture2D.Load with incorrect operands generates invalid spir-v instead of usable error message
**Description**
An easy newbie mistake is to call `Texture2D.Load(x, y)` instead of `Texture2D.Load(int3(x, y, 0))`.  This mistake should be caught by the type checker, and an error raised saying something like "Texture2D.Load() parameter 0 must be int3".  Instead, DXC generates invalid spir-v.  Luckily this triggers an error in the validator, rather than making it into the app and crashing in the driver.  But a better error would be nice.

**Steps to Reproduce**
```hlsl
Texture2D SourceTexture;
void Main(in float4 SvPosition : SV_Position, out float4 OutColor : SV_Target0) {
    OutColor = SourceTexture.Load(SvPosition.x, SvPosition.y);
}
```
Compile the above code with the following command line:
```
dxc -HV 2021 -T ps_6_2 -E Main -spirv -fspv-target-env=vulkan1.1 -Fo broken.spv broken.hlsl
```

[Link to godbolt repro](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:hlsl,selection:(endColumn:1,endLineNumber:5,positionColumn:1,positionLineNumber:5,selectionStartColumn:1,selectionStartLineNumber:5,startColumn:1,startLineNumber:5),source:'Texture2D+SourceTexture%3B%0Avoid+Main(in+float4+SvPosition+:+SV_Position,+out+float4+OutColor+:+SV_Target0)+%7B%0A++++OutColor+%3D+SourceTexture.Load(SvPosition.x,+SvPosition.y)%3B%0A%7D%0A'),l:'5',n:'0',o:'HLSL+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:dxc_1_7_2207,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:hlsl,libs:!(),options:'-HV+2021+-T+ps_6_2+-E+Main+-spirv+-fspv-target-env%3Dvulkan1.1',overrides:!(),selection:(endColumn:45,endLineNumber:3,positionColumn:45,positionLineNumber:3,selectionStartColumn:45,selectionStartLineNumber:3,startColumn:45,startLineNumber:3),source:1),l:'5',n:'0',o:'+DXC+1.7.2207+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'DXC+1.7.2207',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+DXC+1.7.2207+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

**Actual Behavior**
```
fatal error: generated SPIR-V is invalid: [VUID-StandaloneSpirv-Offset-04663] Image Operand Offset can only be used with OpImage*Gather operations
  %27 = OpImageFetch %v4float %26 %22 Lod|Offset %20 %25
```

**Environment**
- DXC version `dxcompiler.dll: 1.7(dev); dxil.dll: 1.7(101.7.2212.14)` (forked version included in Unreal 5.2.1)
- Host Operating System: Windows 10 Pro 22H2 build 19045.3803
- Also reproduces on godbolt, see link above

sudonatalie:
The `Load` function's optional `Offset` parameter must be an immediate value to match the behavior of the DXIL backend. Further, the existing logic in the SPIR-V backend to evaluate it as a variable offset generated invalid SPIR-V that violated the `VUID-StandaloneSpirv-Offset-04663` rule:
> Image operand Offset must only be used with OpImage*Gather instructions

The logic for processing the argument as a variable value has been removed and replaced with an error message that matches DXIL's.

Note that `handleOffsetInMethodCall` is left unmodified because it is also used to process HLSL `.Gather()` calls, which are lowered to `OpImageGather` SPIR-V instructions and can therefore accept non-const `Offset`s as image operands.

It would be nice to move this error message earlier in execution and share between backends, but I found that to be non-trivial to implement. Since both backends already have all of the logic to identify whether the function call has an offset argument, at which index, and evaluate its const-ness, I think this is the most straightforward implementation at this point. 

Fixes #6149

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/method.buffer.load.offset.error.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/method.buffer.load.offset.error.hlsl
@@ -1,0 +1,7 @@
+// RUN: not %dxc -T ps_6_2 -E main -fcgl -spirv %s 2>&1 | FileCheck %s
+
+Texture2D SourceTexture;
+void main(in float4 SvPosition : SV_Position, out float4 OutColor : SV_Target0) {
+    // CHECK: error: Offsets to texture access operations must be immediate values.
+    OutColor = SourceTexture.Load(SvPosition.x, SvPosition.y);
+}
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -322,9 +322,9 @@ public:
       bool doImageFetch, QualType texelType, QualType imageType,
       SpirvInstruction *image, SpirvInstruction *coordinate,
       SpirvInstruction *lod, SpirvInstruction *constOffset,
-      SpirvInstruction *varOffset, SpirvInstruction *constOffsets,
-      SpirvInstruction *sample, SpirvInstruction *residencyCode,
-      SourceLocation loc, SourceRange range = {});
+      SpirvInstruction *constOffsets, SpirvInstruction *sample,
+      SpirvInstruction *residencyCode, SourceLocation loc,
+      SourceRange range = {});
 
   /// \brief Creates SPIR-V instructions for writing to the given image.
   void createImageWrite(QualType imageType, SpirvInstruction *image,
--- a/tools/clang/lib/SPIRV/SpirvBuilder.cpp
+++ b/tools/clang/lib/SPIRV/SpirvBuilder.cpp
@@ -622,14 +622,13 @@ SpirvInstruction *SpirvBuilder::createImageFetchOrRead(
     bool doImageFetch, QualType texelType, QualType imageType,
     SpirvInstruction *image, SpirvInstruction *coordinate,
     SpirvInstruction *lod, SpirvInstruction *constOffset,
-    SpirvInstruction *varOffset, SpirvInstruction *constOffsets,
-    SpirvInstruction *sample, SpirvInstruction *residencyCode,
-    SourceLocation loc, SourceRange range) {
+    SpirvInstruction *constOffsets, SpirvInstruction *sample,
+    SpirvInstruction *residencyCode, SourceLocation loc, SourceRange range) {
   assert(insertPoint && "null insert point");
 
   const auto mask = composeImageOperandsMask(
       /*bias*/ nullptr, lod, std::make_pair(nullptr, nullptr), constOffset,
-      varOffset, constOffsets, sample, /*minLod*/ nullptr);
+      /*varOffset*/ nullptr, constOffsets, sample, /*minLod*/ nullptr);
 
   const bool isSparse = (residencyCode != nullptr);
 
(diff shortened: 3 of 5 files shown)


cassiebeckley (approved):

tools/clang/test/CodeGenSPIRV/method.buffer.load.offset.error.hlsl
```diff
@@ -0,0 +1,7 @@
+// RUN: not %dxc -T ps_6_2 -E main -fcgl -spirv %s 2>&1 | FileCheck %s
+
+Texture2D SourceTexture;
+void main(in float4 SvPosition : SV_Position, out float4 OutColor : SV_Target0) {
+    // CHECK: error: Offsets to texture access operations must be immediate values.
+    OutColor = SourceTexture.Load(SvPosition.x, SvPosition.y);
```
Keenuts, line 6: DXIL behavior seems slight different than what the error says:
- this code does compile with `-fcgl`
- this code throws an error with any other optimization level.

The following code works:
```
float foo() {
  return 0;
}

Texture2D SourceTexture;
void main(in float4 SvPosition : SV_Position, out float4 OutColor : SV_Target0) {
  // CHECK: error: Offsets to texture access operations must be immediate values.
  OutColor = SourceTexture.Load(SvPosition.x, foo());
}
```

Seems like they don't require an immediate, but something that evaluates to a constant, and always legalize the code if a `Load` is used. (The snippets also compiles with `-O0`).

Keenuts (approved):
LGTM for this as it improves from the current state, but seems like we don't match DXIL behavior just yet

tools/clang/test/CodeGenSPIRV/method.buffer.load.offset.error.hlsl
sudonatalie: > DXIL behavior seems slight different than what the error says:
> 
> * this code does compile with `-fcgl`
> * this code throws an error with any other optimization level.

It looks like the pass in which DXIL does this validation is skipped with `-fcgl`, so I wouldn't really consider it relevant.

> The following code works:
> 
> ```
> float foo() {
>   return 0;
> }
> 
> Texture2D SourceTexture;
> void main(in float4 SvPosition : SV_Position, out float4 OutColor : SV_Target0) {
>   // CHECK: error: Offsets to texture access operations must be immediate values.
>   OutColor = SourceTexture.Load(SvPosition.x, foo());
> }
> ```
>
> Seems like they don't require an immediate, but something that evaluates to a constant, and always legalize the code if a `Load` is used. (The snippets also compiles with `-O0`).

That's correct. In reality both backends attempt to evaluate the expression to a constant int, and that behavior likely has some edge cases that are slightly different. Since we don't have any formal `constexpr` though, it's a bit vague to define what's accepted so I think it's alright to keep the error message consistent and more narrow than what we actually support.

PR #6393, opened 2024-03-06
[SPIR-V] Fix null pointer crash in short-circuit ternary op.

LukasBanana:
`castToType` can return null after emitting an error, so `doShortCircuitedConditionalOperator` should exit when true or false branches failed to compile, otherwise it will crash inside `createStore`.

This can be reproduced with in the main branch with a ternary operator selectin a `SamplerState`, which should only emit an error but crashes with a SIGSEGV:
https://godbolt.org/z/j9EGGxWTh

`$ dxc -spirv -E PSMain -T ps_6_0 -HV 2021`
```hlsl
SamplerState SamplerA;
SamplerState SamplerB;
Texture2D Tex;
bool WhichSampler;
float4 SampleTex() {
    return Tex.Sample(WhichSampler ? SamplerA : SamplerB, (float2)0);
}
float4 PSMain() : SV_Target0 {
    return SampleTex();
}
```

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/ternary-op.short-circuited-cond-op.sampler-state.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/ternary-op.short-circuited-cond-op.sampler-state.hlsl
@@ -1,0 +1,11 @@
+// RUN: not %dxc -T ps_6_0 -E main -HV 2021 -fcgl  %s -spirv 2>&1 | FileCheck %s
+
+SamplerState sampler_a;
+SamplerState sampler_b;
+Texture2D tex;
+bool selector;
+
+float4 main() : SV_Target0 {
+// CHECK: error: casting to type 'SamplerState' unimplemente
+  return tex.Sample(selector ? sampler_a : sampler_b, (float2)0);
+}
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -3608,6 +3608,8 @@ SpirvInstruction *SpirvEmitter::doShortCircuitedConditionalOperator(
   SpirvInstruction *trueVal = loadIfGLValue(trueExpr);
   trueVal = castToType(trueVal, trueExpr->getType(), type,
                        trueExpr->getExprLoc(), range);
+  if (!trueVal)
+    return nullptr;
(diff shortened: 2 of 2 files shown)


cassiebeckley (changes requested):
Looks good, but can you add your godbolt sample as a test case?

LukasBanana:
I added [ternary-op.short-circuited-cond-op.sampler-state.hlsl](https://github.com/microsoft/DirectXShaderCompiler/pull/6393/files#diff-a9c513604bc02320f844a3dbe491c721716296dd6f49409c63f0e85f0db28229) and adopted the syntax from *ternary-op.short-circuited-cond-op.hlsl*.

tools/clang/test/CodeGenSPIRV/ternary-op.short-circuited-cond-op.sampler-state.hlsl
```diff
@@ -0,0 +1,10 @@
+// RUN: %dxc -T ps_6_0 -E main -HV 2021 -fcgl  %s -spirv | FileCheck %s
```
cassiebeckley, line 1: This will fail since dxc will report an error and return a non-zero value. You'll need to replace it with
```suggestion
// RUN: not %dxc -T ps_6_0 -E main -HV 2021 -fcgl  %s -spirv 2>&1 | FileCheck %s
```

tools/clang/test/CodeGenSPIRV/ternary-op.short-circuited-cond-op.sampler-state.hlsl
```diff
@@ -0,0 +1,10 @@
...
+
+SamplerState sampler_a;
+SamplerState sampler_b;
+Texture2D tex;
+bool selector;
+
+float4 main() : SV_Target0 {
+  return tex.Sample(selector ? sampler_a : sampler_b, (float2)0);
```
cassiebeckley, line 9: `FileCheck` will expect some `// CHECK:` comments to verify; you can either remove FileCheck from the RUN line above or add

```suggestion
// CHECK: error: casting to type 'SamplerState' unimplemented
  return tex.Sample(selector ? sampler_a : sampler_b, (float2)0);
```

cassiebeckley (approved):
LGTM once the comments are addressed

sudonatalie (approved):
Thank you for the fix!

PR #6394, opened 2024-03-06
Move 2 new SPIR-V tests to CodeGenSPIRV

hekota:
CodeGenSPIRV_Lit directory is redundant and should not be used anymore. Also, the CodeGenSPIRV directory has a lit.local.cfg that makes sure these tests are exluded from run for builds without SPIR-V support.


pow2clk (approved):
Looks correct! Just tagging @sudonatalie as an FYI since this is now a fixup to #6041  CodeGenSPIRV_Lit is no more since everything in CodeGenSPIRV is lit now

Keenuts (approved):

sudonatalie:
Thanks for catching that!

PR #6396, opened 2024-03-07
[SPIR-V] Block semantic annotation reuse for input

### LINKED ISSUE #3737: [SPIR-V] Reapeated SV_DomainLocation semantics don't produce an error and can crash the compiler (works with the DXIL backend)
### Title
[SPIR-V] Reapeated `SV_DomainLocation` semantics don't produce an error and can crash the compiler (works with the DXIL backend)

### Functional impact
Declaring repeated `SV_DomainLocation` input semantics in the same Domain shader entry point does not produce a compiler error and crashes in some cases.

### Minimal repro steps
1. Compile the following shader with `dxc.exe -T ds_6_0 -E Domain -spirv repro_shader.txt`
[repro_shader.txt](https://github.com/microsoft/DirectXShaderCompiler/files/6415171/repro_shader.txt)

### Expected result
The SPIR-V backend should produce error similar to that of the DXIL backend:
```
float4 Domain (const OutputPatch<HullOut,3> vi,
^
repro_shader.txt:4:1: error: Parameter with semantic SV_DomainLocation has overlapping semantic index at 0
```

### Actual result
The SPIR-V backend crashes with an access violation exception:
`Internal compiler error: access violation. Attempted to read from address 0x0000000000000010`

### Further technical details
The `repro_shader.txt` in the `Minimal repro steps` section crashes the compiler with 3 repeats, if repeating only twice it still compiles, which is still an issue as it should instead produce a compiler error (like the DXIL backend does)

Keenuts:
The MSDN spec is not very clear regarding input parameter aliasing, BUT DXIL & MS agrees (See #3737) using the same semantic annotation twice should be disallowed.

DXIL currently disallows it for most, and due to a bug, allows some compute related semantics to be aliased.
SPIR-V did allowed aliasing, but it was a implemented as a hack causing typing issues if the type changed between the 2 parameters using the same semantic annotation.

To align more closely with DXIL, and the "spec", we are now disallowing all semantic attribute reuse for input parameters.

Fixes #3737


sudonatalie (approved):
Even though it's a bit breaking, this seems reasonable to me. Thanks for all the background work looking into this.

> To align more closely with DXIL, and the "spec", we are not disallowing all semantic attribute reuse for input parameters.

s/not/now?

Keenuts:
> Even though it's a bit breaking, this seems reasonable to me. Thanks for all the background work looking into this.

If users complain, maybe we could limit that to some specific shader model like 6+ (as the SPIR-V doc does says we support all shader model < 5..)

> > To align more closely with DXIL, and the "spec", we are not disallowing all semantic attribute reuse for input parameters.
> 
> s/not/now?

Ah yes thanks!

cassiebeckley (approved):

PR #6399, opened 2024-03-08
[SPIR-V] Allow literal selectors in switch stmts

### LINKED ISSUE #4338: [SPIR-V] Error with switch statement with constant input

sudonatalie:
Adds support for literal ints as selectors in switch statements.

Fixes #4338

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvInstruction.h
+++ b/tools/clang/include/clang/SPIRV/SpirvInstruction.h
@@ -821,10 +821,10 @@ private:
 /// \brief Switch instruction
 class SpirvSwitch : public SpirvBranching {
 public:
-  SpirvSwitch(SourceLocation loc, SpirvInstruction *selector,
-              SpirvBasicBlock *defaultLabel,
-              llvm::MutableArrayRef<std::pair<llvm::APInt, SpirvBasicBlock *>>
-                  &targetsVec);
+  SpirvSwitch(
+      SourceLocation loc, SpirvInstruction *selector,
+      SpirvBasicBlock *defaultLabel,
+      llvm::ArrayRef<std::pair<llvm::APInt, SpirvBasicBlock *>> &targetsVec);
 
   DEFINE_RELEASE_MEMORY_FOR_CLASS(SpirvSwitch)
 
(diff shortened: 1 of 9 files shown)


tools/clang/include/clang/SPIRV/SpirvInstruction.h
```diff
@@ -837,7 +837,7 @@ class SpirvSwitch : public SpirvBranching {
 
   SpirvInstruction *getSelector() const { return selector; }
   SpirvBasicBlock *getDefaultLabel() const { return defaultLabel; }
-  llvm::ArrayRef<std::pair<llvm::APInt, SpirvBasicBlock *>> getTargets() const {
+  std::vector<std::pair<llvm::APInt, SpirvBasicBlock *>> &getTargets() {
```
Keenuts, line 840: You can revert all the `std::vector` changes if you change the `llvm::ArrayRef` here by a `llvm::MutableArrayRef`.

Keenuts (approved):

tools/clang/include/clang/SPIRV/SpirvInstruction.h
sudonatalie: Good call, thanks!

tools/clang/include/clang/SPIRV/SpirvInstruction.h
```diff
@@ -837,7 +837,8 @@ class SpirvSwitch : public SpirvBranching {
 
   SpirvInstruction *getSelector() const { return selector; }
   SpirvBasicBlock *getDefaultLabel() const { return defaultLabel; }
-  llvm::ArrayRef<std::pair<llvm::APInt, SpirvBasicBlock *>> getTargets() const {
+  llvm::MutableArrayRef<std::pair<llvm::APInt, SpirvBasicBlock *>>
```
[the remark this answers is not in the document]
sudonatalie: Good point, done.

Keenuts (approved):

cassiebeckley (approved):

PR #6401, opened 2024-03-11
Sema: warn when POSITION[0] is used intead of SV_Position

### LINKED ISSUE #3742: No warning/error when SV_Position semantic is missing from Vert/Domain/Geom output and POSITION is used instead
### Title
No warning/error when SV_Position semantic is missing from Vert/Domain/Geom output and POSITION is used instead

### Functional impact
Shaders written for FXC which use the d3d9 style POSITION semantic for the vertex pipeline output will compile with DXC but will produce no output rendering. The POSITION semantic is just interpreted as a user semantic but the fact that SV_Position is not written out is not reported, which is most often a user error. P.S. this gotcha is also not documented in the [Porting shaders from FXC to DXC](https://github.com/microsoft/DirectXShaderCompiler/wiki/Porting-shaders-from-FXC-to-DXC) wiki

### Minimal repro steps
1. Compile the following shader with `dxc.exe -T vs_6_0 -E Vert repro_shader.txt` (or `with dxc.exe -T vs_6_0 -E Vert -spirv repro_shader.txt` for the SPIR-V backend)
```
float4 Vert(float4 position : POS) : POSITION
{
	return position;
}
```

### Expected result
An error or at least a warning is printed when there is no SV_Position in the Vertex/Domain/Geometry shader output, possibly pointing to the user defined POSITION semantic as the place for user correction. Similar to what DXC does for uses of the legacy COLOR or DEPTH semantics in the pixel shader output.

### Actual result
Shader compiles without any `POS` SysValue (for the DXIL backend) and no `gl_Position` (for the SPIR-V backend) in the output. The resulting shader produces no rendering at runtime.

### Further technical details
**N.B.** To my understanding DXC should probably only not warn/error when using the stream output stage in the vertex pipeline but we at Unity don't ever use stream output and would prefer if DXC had a compiler flag to enable this warning when we know before compilation that the shader will never be used for stream output.

Keenuts:
Before SM4, POSITION was equivalent to today's SV_Position. However, as HLSL moved to system semantics (SV_), the support for those old semantics has been deprecated.
Some like VFACE are now completely forbidden, but others like POSITION or COLOR have been handled differently by FXC and DXC.

Today, usage of POSITION in place of SV_Position is probably a mistake as POSITION is now considered to be a user-defined semantic.

This commit adds a warning when POSITION is used as an output semantic for a vertex shader, as most cases should be the result of a mistake. If this is expected, the warning can be disabled by using `-Wno-dx9-deprecation`.

Fixes #3742

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7712,6 +7712,9 @@ def warn_hlsl_incorrect_bind_semantic: Warning< // Is an error in fxc in some ca
   "invalid register specification, expected %0 binding">;
 def err_hlsl_incorrect_bind_semantic: Error< // Is an error in fxc in some cases, but is sometimes ignored
   "invalid register specification, expected %0 binding">;
+def warn_hlsl_semantic_attribute_position_misuse_hint: Warning<
+  "'POSITION' is not considered equal to SV_Position but is treated as a user-defined semantic attribute">,
+  InGroup<DiagGroup<"dx9-deprecation">>;
 def warn_hlsl_unary_negate_unsigned : Warning<
   "unary negate of unsigned value is still unsigned">,
   InGroup<Conversion>, DefaultWarn;
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7713,7 +7713,7 @@ def warn_hlsl_incorrect_bind_semantic: Warning< // Is an error in fxc in some ca
 def err_hlsl_incorrect_bind_semantic: Error< // Is an error in fxc in some cases, but is sometimes ignored
   "invalid register specification, expected %0 binding">;
 def warn_hlsl_semantic_attribute_position_misuse_hint: Warning<
-  "'POSITION' is not considered equal to SV_Position but is treated as a user-defined semantic attribute">,
+  "'%0' is a user-defined semantic; did you mean 'SV_Position'?">,
   InGroup<DiagGroup<"dx9-deprecation">>;
 def warn_hlsl_unary_negate_unsigned : Warning<
   "unary negate of unsigned value is still unsigned">,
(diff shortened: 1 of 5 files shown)


Keenuts:
@pow2clk @tex3d

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7712,6 +7712,9 @@ def warn_hlsl_incorrect_bind_semantic: Warning< // Is an error in fxc in some ca
   "invalid register specification, expected %0 binding">;
 def err_hlsl_incorrect_bind_semantic: Error< // Is an error in fxc in some cases, but is sometimes ignored
   "invalid register specification, expected %0 binding">;
+def warn_hlsl_semantic_attribute_position_misuse_hint: Warning<
+  "'POSITION' is not considered equal to SV_Position but is treated as a user-defined semantic attribute">,
```
sudonatalie, line 7716: optional: I think a slightly more idiomatic clang phrasing might be: `'POSITION' is a user-defined semantic; did you mean 'SV_Position'?`

Keenuts: Yes, way better message!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15295,6 +15295,20 @@ void DiagnoseAmplificationEntry(Sema &S, FunctionDecl *FD,
   return;
 }
 
+void DiagnoseVertexEntry(Sema &S, FunctionDecl *FD, llvm::StringRef StageName) {
+  for (auto *annotation : FD->getUnusualAnnotations()) {
+    if (auto *sema = dyn_cast<hlsl::SemanticDecl>(annotation)) {
+      if (sema->SemanticName.equals_lower("POSITION") ||
+          sema->SemanticName.equals_lower("POSITION0")) {
+        S.Diags.Report(FD->getLocation(),
```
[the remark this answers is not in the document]
Keenuts: Yes, that's better, done

sudonatalie (approved):
LGTM pending approval from someone at MS

Keenuts:
@llvm-beanz ping? (or @tex3d?)

PR #6405, opened 2024-03-12
Fix verification in ExecutionTest::ComputeSampleTest

### LINKED ISSUE #4788: ExecutionTest::ComputeSampleTest has invalid verification logic causing failures with mesh shaders
The relevant function:
```c++
void VerifySampleResults(const UINT *pPixels, UINT width) {
  UINT xlod = 0;
  UINT ylod = 0;
  // Each pixel contains 4 samples and 4 LOD calculations.
  // 2 of these (called 'left' and 'right') have X values that vary and a constant Y
  // 2 others (called 'top' and 'bot') have Y values that vary and a constant X
  // Only of the X variant sample results and one of the Y variant results
  // are actually reported for the pixel.
  // The other 2 serve as "helpers" to the other pixels in the quad.
  // On the left side of the quad, the 'left' samples are reported.
  // Op the top of the quad, the 'top' samples are reported and so on.
  // The varying coordinate values alternate between zero and a
  // value whose magnitude increases with the index.
  // As a result, the LOD level should steadily increas.
  // Due to vagaries of implementation, the same derivatives
  // in both directions might result in different levels for different locations
  // in the quad. So only comparisons between sample results and LOD calculations
  // and ensuring that the LOD increased and reaches the max can be tested reliably.
  for (unsigned i = 0; i < width; i++) {
    // CalculateLOD and Sample from texture with mip levels containing LOD index should match
    VERIFY_ARE_EQUAL(pPixels[4*i + 0], pPixels[4*i + 1]);
    VERIFY_ARE_EQUAL(pPixels[4*i + 2], pPixels[4*i + 3]);
    // Make sure LODs are ever climbing as magnitudes increase
    VERIFY_IS_TRUE(pPixels[4*i] >= xlod);
    xlod = pPixels[4*i];
    VERIFY_IS_TRUE(pPixels[4*i + 2] >= ylod);
    ylod = pPixels[4*i + 2];
  }
  // Make sure we reached the max lod level for both tracks
  VERIFY_ARE_EQUAL(xlod, 6u);
  VERIFY_ARE_EQUAL(ylod, 6u);
}
```

This iterates "pixel" by pixel ("pixel" because it's a 1D dispatch, which is laid out as if it were operating on quads) and verifies the 4 components, as:
1. The LOD that would be computed based on the X derivative between that pixel and 0.
2. The actual LOD that a sample operation used with the same coordinates of 1, by loading from a texture with different values in each mip.
3. Same as 1 but for Y dimension
4. Same as 2 but for Y dimension

Since it's iterating pixel by pixel, this loop will go from top-left of a quad, to top-right, to bottom-left, to bottom-right, and then back to top-left of a new quad. So when going from pixel 0 to pixel 1, it's expected that the X-derivative-LOD should **not decrease**, since pixel 1's X coordinate increased, and therefore that pixel's derivative vs 0 increased. But going from pixel 1 to 2... it is absolutely valid for the X-derivative-LOD to decrease, because pixel 2's X coordinate *is less than* pixel 1's.

This logic is incorrect for both compute and mesh, but for compute, it coincidentally works, because pixel 1's X-derivative-LOD is also 0, because the 1/336 difference between 0 and pixel 1's X coordinate is not sufficient to change LODs. For mesh, however, since the limit of a thread group size is 128, the validation is done using the smaller range of 116, and 1/116 *is* sufficient to change LOD for pixel 1.

This validation should instead be iterating quad by quad, and ensuring that when *not* moving right->left that the X-derivative-LOD *doesn't decrease* (i.e. ignoring the edge of 1->2 for pixel indices mod 4).

For Y-derivative-LODs, since the logic doesn't strictly iterate row-by-row by quads, but instead along a diagonal, there's no situation where increasing thread index will cause a decrease in the Y coordinate, so there's no real need to change the validation for Y values.

### LINKED ISSUE #6181: ExecutionTest::ComputeSampleTest is failing on latest Microsoft Basic Render Driver
**Description**
`ExecutionTest::ComputeSampleTest` is failing on latest Microsoft Basic Render Driver.

**Steps to Reproduce**
On Windows build 26016.1001 or higher:
`hcttest exec-filter ComputeSampleTest -adapter "NVidia*"`

**Actual Behavior**
```
******************** TEST 'clang-taef-exec :: ExecHLSLTests.dll/ExecutionTest::ComputeSampleTest' FAILED ********************
Test Authoring and Execution Framework v10.68 for arm64

Warning: TAEF will not listen for crashes from Windows Error Reporting because it failed to set up the listener. ([HRESULT 0x80070005] Failed to enable the Microsoft.Windows.FaultReporting TraceLogging provider in an ETW trace.)

Agility SDK not enabled.

Experimental mode enabled.

Debug layer enabled.



StartGroup: ExecutionTest::ComputeSampleTest

Using Adapter:Microsoft Basic Render Driver

Error: Verify: IsTrue(pPixels[4 * i] >= xlod) [File: C:\__w\1\s\DXC\tools\clang\unittests\HLSLExec\ExecutionTest.cpp, Function: VerifySampleResults, Line: 4150]

EndGroup: ExecutionTest::ComputeSampleTest [Failed]
```

**Environment**
- DXC version: `main` as of 1/22/2024
- Windows 11 build 26016.1001 or higher

hekota:
The `ComputeSampleTest` is verifying `Sample` and `CalculateLevelOfDetail` intrinsics calls on a texture in compute, mesh and amplification shaders. The test is set up in a way that the reported LOD values should be increasing as the X and Y coordinates increase.

However, since the test results are stored in quad z-order, and it is perfectly valid for the X-derivative-LOD to decrease when going from top-right to bottom-left pixel in a quad because in this case the X coordinate decreases. Therefore, the test needs verify the LOD values are increasing everywhere except for the bottom-left quad pixel (every output with `index % 4 == 2`).

Previously, the test coincidentally worked for compute shaders because it uses thread group of 336 and the 1/336 difference between 0 and pixel 1's X coordinate is not sufficient to change LODs. For mesh, however, since the limit of a thread group size is 128, the validation is done using the smaller range of 116, and 1/116 is sufficient to change LOD for pixel 1.

This PR also fixes a couple of mismatched thread group dimensions in `VerifySampleResults`.

The shader code for this test is at [ShaderOpArith.xml](https://github.com/microsoft/DirectXShaderCompiler/blob/130877392c263888ef06bab768856d3dab1f1c9a/tools/clang/unittests/HLSLExec/ShaderOpArith.xml#L454) line 454.

Fixes #4788 
Fixes #6181

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -4210,25 +4210,36 @@ void VerifySampleResults(const UINT *pPixels, UINT width) {
   UINT ylod = 0;
   // Each pixel contains 4 samples and 4 LOD calculations.
   // 2 of these (called 'left' and 'right') have X values that vary and a
-  // constant Y 2 others (called 'top' and 'bot') have Y values that vary and a
-  // constant X Only of the X variant sample results and one of the Y variant
-  // results are actually reported for the pixel. The other 2 serve as "helpers"
-  // to the other pixels in the quad. On the left side of the quad, the 'left'
-  // samples are reported. Op the top of the quad, the 'top' samples are
-  // reported and so on. The varying coordinate values alternate between zero
-  // and a value whose magnitude increases with the index. As a result, the LOD
-  // level should steadily increas. Due to vagaries of implementation, the same
-  // derivatives in both directions might result in different levels for
+  // constant Y. 2 others (called 'top' and 'bot') have Y values that vary and a
+  // constant X. Only one of the X variant sample results and one of the Y
+  // variant results are actually reported for the pixel. The other 2 serve as
+  // "helpers" to the other pixels in the quad. On the left side of the quad,
+  // the 'left' samples are reported. On the top of the quad, the 'top' samples
+  // are reported and so on. The varying coordinate values alternate between
+  // zero and a value whose magnitude increases with the index. As a result, the
+  // LOD level should steadily increase. Due to vagaries of implementation, the
+  // same derivatives in both directions might result in different levels for
   // different locations in the quad. So only comparisons between sample results
   // and LOD calculations and ensuring that the LOD increased and reaches the
   // max can be tested reliably.
+
+  // The results are stored in quad z-order (top-left (#0), top-right (#1),
+  // bottom-left (#2), bottom-right (#3)) and need to be evaluated as such.
+  // The X-derivative-LOD should not decrease when going from quad pixel #0->#1
+  // and #2->#3 or #3->#0, but not for #1->#2 where is is absolutely valid for
+  // the X-derivative-LOD to decrease, because pixel #2's X coordinate is less
+  // than pixel #1's. Therefore the test skips verification of X-derivative-LOD
+  // on quad pixel #2.
+
   for (unsigned i = 0; i < width; i++) {
     // CalculateLOD and Sample from texture with mip levels containing LOD index
     // should match
     VERIFY_ARE_EQUAL(pPixels[4 * i + 0], pPixels[4 * i + 1]);
     VERIFY_ARE_EQUAL(pPixels[4 * i + 2], pPixels[4 * i + 3]);
     // Make sure LODs are ever climbing as magnitudes increase
-    VERIFY_IS_TRUE(pPixels[4 * i] >= xlod);
+    if (i % 4 != 2) { // skip X-derivative-LOD verification on quad pixel #2
+      VERIFY_IS_TRUE(pPixels[4 * i] >= xlod);
+    }
     xlod = pPixels[4 * i];
     VERIFY_IS_TRUE(pPixels[4 * i + 2] >= ylod);
     ylod = pPixels[4 * i + 2];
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -4210,25 +4210,36 @@ void VerifySampleResults(const UINT *pPixels, UINT width) {
...
   // and LOD calculations and ensuring that the LOD increased and reaches the
   // max can be tested reliably.
+
+  // The results are stored in quad z-order (top-left (#0), top-right (#1),
+  // bottom-left (#2), bottom-right (#3)) and need to be evaluated as such.
+  // The X-derivative-LOD should not decrease when going from quad pixel #0->#1
+  // and #2->#3 or #3->#0, but not for #1->#2 where is is absolutely valid for
+  // the X-derivative-LOD to decrease, because pixel #2's X coordinate is less
+  // than pixel #1's. Therefore the test skips verification of X-derivative-LOD
+  // on quad pixel #2.
```
pow2clk, line 4232: The "not" on line 4229 is negating the earlier "not decrease", which is a bit confusing. I would make a new sentence that states outright that the `#1->#2` transition should involve a decrease.
  
There are simpler ways to update the text. I got a bit carried away.

```suggestion
  // The X-derivative-LOD should not decrease when going from quad pixel #0->#1,
  // #2->#3, or #3->#0. For #1->#2 the end of the typewriter "line" is reached and
  // zags left resulting in a smaller x value. So, it is absolutely valid for the 
  // X-derivative-LOD to decrease. Therefore, the test skips verification of X-derivative-LOD
  // on quad pixel #2.
```

pow2clk (approved):
The functional changes are all valid. I have a suggestion to clarify the comment that could be simplified if you prefer

damyanp (approved):

PR #6428, opened 2024-03-18
azure-pipelines: Fix the fact that neither asan+ubsan were not actually enabled

amaiorano:
Also make the message about the unsupported value of LLVM_USE_SANITIZER an error rather than a warning, so we don't make this mistake in the future.


amaiorano:
This change depends on the following landing first:
https://github.com/microsoft/DirectXShaderCompiler/pull/6429
https://github.com/microsoft/DirectXShaderCompiler/pull/6430
https://github.com/microsoft/DirectXShaderCompiler/pull/6431/files

PR #6429, opened 2024-03-18
Fix ubsan failure on uninitialized read of bool param to WideCharToMultiByte

amaiorano:
If last arg lpUsedDefaultChar is non-null, the function must set this value to true or false. WideToEncodedString conditionally passes in a pointer to 'usedDefaultChar', which was not getting written to, and causing ubsan to fail on reading an invalid boolean value:
```
.../lib/DxcSupport/Unicode.cpp:156:14: runtime error: load of value 208, which is not a valid value for type 'BOOL' (aka 'bool')
```


llvm-beanz (approved):

bob80905:
Why not initialize the relevant BOOL variable to FALSE on line 130?

amaiorano:
> Why not initialize the relevant BOOL variable to FALSE on line 130?

Because that would only fix the asan error for this one instance. The real problem isn't that we didn't initialize the input bool, it's that we're supposed to write to the bool if it's passed to `WideToEncodedString`, which we were not.

Issue #6404, opened 2024-03-12
[Feature Request] Implement experimental mesh nodes


pow2clk:
**Is your feature request related to a problem? Please describe.**
Work graphs are limited to compute shaders at present. There is discussion in the [work graphs spec](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/WorkGraphs.md#graphics-nodes) itself about adding graphics leaf nodes to give the calculations performed by work graphs a path to the screen. The mesh node approach is formally proposed [here](https://github.com/microsoft/hlsl-specs/pull/155)

**Describe the solution you'd like**
Support for mesh nodes could be done by replacing the input payload of a mesh shader with a node input record for a mesh node. Such nodes would be denoted using a new lauch type: `[NodeLaunch("mesh")]`.

**Describe alternatives you've considered**
While more complete support for graphics leaf nodes would be useful, mesh shader interfaces present a relatively simple modification that allows feeding work graphs output into the graphics pipeline without too many interface changes.

PR #6436, opened 2024-03-19
Add mesh launch type for node shaders

### LINKED ISSUE #6432: [Feature Request] Allow the mesh launch type for node shaders
Essentially a renamed broadcast node. set up a skeleton. including metadata and RDAT that captures launch type and additional properties. Allows defining and refining interfaces early. Tested with blank entry points without much content.

In order to support mesh nodes, the compiler must accept the HLSL code that specified mesh nodes. See parent #6404

Add the ability to specify a mesh node as little more than a renamed broadcasting node. This will produce the metadata and RDAT with the launch type and a few additional properties. Test with blank entry points without much additional content or complications.

pow2clk:
Adds the ability to specify a mesh node which functions largely like a broadcasting node. This will generate code, metadata, and rdat info for a minimal mesh node. Diagnostics are changed to allow the mesh nodes to be generated. Tests for RDAT and metadata are included. Errors referencing only broadcasting nodes now mention mesh nodes where appropriate.

Fixes #6432


tools/clang/test/HLSLFileCheck/hlsl/workgraph/case070_noinput.hlsl
```diff
@@ -20,6 +20,12 @@ void node070_coalescing() { }
 [NumThreads(1,1,1)]
 void node070_thread() { }
 
+[Shader("node")]
+[NodeLaunch("mesh")]
+[NumThreads(1,1,1)]
```
python3kgae, line 25: Any limit on number threads for mesh node?

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -641,6 +641,7 @@ DXIL::NodeLaunchType ShaderModel::NodeLaunchTypeFromName(llvm::StringRef name) {
       .Case("broadcasting", DXIL::NodeLaunchType::Broadcasting)
       .Case("coalescing", DXIL::NodeLaunchType::Coalescing)
       .Case("thread", DXIL::NodeLaunchType::Thread)
+      .Case("mesh", DXIL::NodeLaunchType::Mesh)
```
tex3d, line 644: Should we have had a static_assert here as well to make sure this StringSwitch gets updated?

tools/clang/test/HLSLFileCheck/hlsl/workgraph/case070_noinput.hlsl
pow2clk: The same limits as on regular mesh shaders, which is that the maximum threads must be less than 128. As for lower limits, this is fine.

tex3d: Yes, we should limit these to what's allowed for mesh shaders.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/case070_noinput.hlsl
```diff
@@ -53,6 +59,22 @@ void node070_thread() { }
 // CHECK-NOT: i32 20, i32 {{![-=9]+}}
 // CHECK-SAME: }
 
+// Metadata for node070_mesh
```
tex3d, line 62: I think we should put mesh node tests in a separate file from other node types if possible.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/noinput_rdat.hlsl
```diff
@@ -0,0 +1,371 @@
+// RUN: %dxc -T lib_6_8 %s | %D3DReflect %s | FileCheck %s
```
tex3d, line 1: See above about separating mesh nodes from launch types in test files, and mesh nodes should require `lib_6_9`.

Plus, we should add any additional non-mesh node tests to main.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/noinput_rdat.hlsl
```diff
@@ -0,0 +1,371 @@
...
+// CHECK:      ShaderStageFlag: (Node)
+// CHECK:      MinShaderTarget: 0xf0068
+// CHECK:      MinimumExpectedWaveLaneCount: 0
+// CHECK:      MaximumExpectedWaveLaneCount: 0
+// CHECK:      ShaderFlags: 0 (None)
+// CHECK:      Node: <1:NodeShaderInfo> = {
+// CHECK:        LaunchType: Coalescing
+// CHECK:        GroupSharedBytesUsed: 0
+// CHECK:        Attribs: <{{[0-9]+}}:RecordArrayRef<NodeShaderFuncAttrib>[2]>  = {
+// CHECK:          [0]: <3:NodeShaderFuncAttrib> = {
```
tex3d, line 70: Do we need every line of RDAT dump?  Note that the `3:` here is fragile, since it indicates the index of this attribute in the `NodeShaderFuncAttribTable`, which could change if attribute change for some function in this library.

This recursively dumps the contents of the RecordRef (from other tables), the first time referenced from each top-level record.  So it's redundant to check the contents of those same records in that corresponding table later.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/noinput_rdat.hlsl
```diff
@@ -0,0 +1,371 @@
...
+// CHECK:            AttribKind: DispatchGrid
+// CHECK:            DispatchGrid: <4:array[3]> = { 4, 1, 1 }
+// CHECK:          }
+// CHECK:        }
+// CHECK:        Outputs: <RecordArrayRef<IONode>[0]> = {}
+// CHECK:        Inputs: <RecordArrayRef<IONode>[0]> = {}
+// CHECK:      }
+// CHECK:    }
+// CHECK:  }
+// CHECK:  RecordTable (stride = {{[0-9]+}} bytes) NodeIDTable[4] = {
```
tex3d, line 161: Checking the contents of this table and the rest of the tables here is redundant if you already checked the nested record dumps in the `FunctionTable`.

tex3d (commented):
Mainly, I think we should:
- Separate test files for mesh node launch from other launch modes.
- Use `lib_6_9` target and make sure we don't allow the use of mesh launch in lower shader models in `DiagnoseNodeEntry`.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/case070_noinput.hlsl
pow2clk: I saw that as being a part of the second task in #6404. Do you think it needs doing now?

tools/clang/test/HLSLFileCheck/hlsl/workgraph/noinput_rdat.hlsl
pow2clk: I modeled it after empty_broadcasting_nodes.hlsl, which includes every line. I'm not particular though, it was just easier this way.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/noinput_rdat.hlsl
pow2clk: Adding the additional tests was the plan. I think there will be others.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/case070_noinput.hlsl
pow2clk: Done

tools/clang/test/HLSLFileCheck/hlsl/workgraph/noinput_rdat.hlsl
pow2clk: Separated. I have a raft of tests that will go along to main after I'm sure I have them all.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/noinput_rdat.hlsl
pow2clk: Removed

lib/DXIL/DxilShaderModel.cpp
pow2clk: do you just mean to throw in a static_assert that suggests the LaunchTypeNames hasn't been updated similar to this line? 

https://github.com/microsoft/DirectXShaderCompiler/blob/7f12aa652602bfc5f2510b04c6806fd10e52f3ee/lib/DXIL/DxilShaderModel.cpp#L633

That will only work if something is added to that list after mesh, which is likely, but I'm not sure it's what you're suggesting.

pow2clk: I added what I think you want

tools/clang/test/HLSLFileCheck/hlsl/workgraph/case070_noinput.hlsl
pow2clk: I've mentioned this limit in the mesh nodes attribute issue #6443

tex3d (approved):
LGTM.

include/dxc/DXIL/DxilConstants.h
```diff
@@ -29,7 +29,7 @@ namespace DXIL {
 const unsigned kDxilMajor = 1;
 /* <py::lines('VALRULE-TEXT')>hctdb_instrhelp.get_dxil_version_minor()</py>*/
 // VALRULE-TEXT:BEGIN
-const unsigned kDxilMinor = 8;
+const unsigned kDxilMinor = 9;
```
tex3d, line 32: Oh, I just recalled that there may be other things gated on SM 6.9 that need to be moved to a new version to prevent them from being enabled in this experiment.

One is the general RDAT for regular shader entry points in libraries.  I'm not sure what other things off the top of my head, but that's probably not the only one.

I also realize that separating PRs between the one that just bumps the version and the rest would be really helpful.

damyanp (approved):

include/dxc/DXIL/DxilConstants.h
pow2clk: Tex and I talked about this issue at some length. The main concern is losing the changes to accommodate the version bump.  I mean to take the commits from this PR and put them into main as the test changes facilitate 6.9, but don't require it.

pow2clk:
I don't know why the PR description bot is out to lunch, but the description meets our criteria. I'm merging anyway.

Issue #6419, opened 2024-03-14
[DXIL] Null pointer exception due to double floating-point literals in fmod intrinsic (in 1.8.2403 release)


damyanp:
This issue tracks #6410, but specifically in the release branch.  See the main issue for full information.

damyanp:
Fixed by #6418

PR #6437, opened 2024-03-20
Add tests for literal overloads of fmod, length, and normalize

tex3d:
Certain HL ops with no double overload will generate double overloads for literal types. These are lowered to double DXIL overloads, some of which are not legal for these ops. We currently rely on illegal intermediate DXIL op overloads for constant evaluation.  If constant evaluation fails, we end up with illegal DXIL overloads in final DXIL, which is caught by the validator.

The prior revert restored the ability to run these scenarios (with asserts disabled).

This change adds tests for #6419 so automated testing will test across configurations and prevent regressions of this scenario in this branch.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/intrinsics/basic/fmod_const_eval.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/intrinsics/basic/fmod_const_eval.hlsl
@@ -1,0 +1,55 @@
+// RUN: %dxc -T vs_6_0 %s -E main | %FileCheck %s
+// RUN: %dxc -T vs_6_0 %s -E main -DNO_FOLD | %FileCheck %s -check-prefixes=NO_FOLD
+
+// Ensure fmod is constant evaluated during codegen, or dxil const eval
+// TODO: handle fp specials properly!
+
+RWBuffer<float4> results : register(u0);
+
+[shader("vertex")]
+void main(bool b : B) {
+    uint i = 0;
+
+    // Literal float
+    // 2.5, -2.5, 2.5, -2.5
+    // CHECK: call void @dx.op.bufferStore.f32(i32 69, %dx.types.Handle %{{.+}}, i32 0, i32 undef, float 2.500000e+00, float -2.500000e+00, float 2.500000e+00, float -2.500000e+00, i8 15)
+    results[i++] = float4(fmod(5.5, 3.0),
+                          fmod(-5.5, 3.0),
+                          fmod(5.5, -3.0),
+                          fmod(-5.5, -3.0));
+
+    // Explicit float
+    // 2.5, -2.5, 2.5, -2.5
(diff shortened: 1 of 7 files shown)


tools/clang/test/HLSLFileCheck/hlsl/intrinsics/basic/fmod_const_eval.hlsl
```diff
@@ -0,0 +1,55 @@
...
+    // 2.5, -2.5, 2.5, -2.5
+    // CHECK: call void @dx.op.bufferStore.f32(i32 69, %dx.types.Handle %{{.+}}, i32 1, i32 undef, float 2.500000e+00, float -2.500000e+00, float 2.500000e+00, float -2.500000e+00, i8 15)
+    results[i++] = float4(fmod(5.5f, 3.0f),
+                          fmod(-5.5f, 3.0f),
+                          fmod(5.5f, -3.0f),
+                          fmod(-5.5f, -3.0f));
+
+#ifdef SPECIALS
+    // Literal float
+    // 0.0, -0.0, NaN, -NaN
```
pow2clk, line 31: I realize this isn't run yet, but could there be CHECKs for when it is?

tools/clang/test/HLSLFileCheck/hlsl/intrinsics/basic/fmod_const_eval.hlsl
```diff
@@ -0,0 +1,55 @@
+// RUN: %dxc -T vs_6_0 %s -E main | %FileCheck %s
+// RUN: %dxc -T vs_6_0 %s -E main -DNO_FOLD | %FileCheck %s -check-prefixes=NO_FOLD
+
+// Ensure fmod is constant evaluated during codegen, or dxil const eval
+// TODO: handle fp specials properly!
```
pow2clk, line 5: The way we've been dealing with such is with XFAILS and bugs filed to account for them. I think that would require this to be a lit test, but I don't see any reason why it can't be.

tools/clang/test/HLSLFileCheck/hlsl/intrinsics/basic/fmod_const_eval.hlsl
tex3d: I'm not positive these are the expected values yet for `-Gis`.  I think there are also more cases to add for FP specials.  I had left this in because I didn't want to lose track of the scenario so far, but maybe it's better to just delete this code and have an issue tracking the work to handle and test FP specials in constant evaluation later in a different solution merged to main.  What do you think?

pow2clk (approved):
I suppose it's fine, but it has no effect at present.

python3kgae (approved):

tex3d:
> I suppose it's fine, but it has no effect at present.

Other than testing this scenario across configurations in our build pipeline, and preventing any further changes in the branch from regressing it, right?

PR #6439, opened 2024-03-20
[Sema] Check FunctionDecl has identifier before getName.

### LINKED ISSUE #6426: Regression, SIGSEGV instead of diagnostics when encountering bool operator==(const T&, const T&)
Description
SIGSEGV instead of diagnostics when encountering `bool operator==(const T&, const T&)`

Steps to reproduce:
Godbolt link: https://godbolt.org/z/5hsxeK5aE

python3kgae:
Use identifier name without check the identifier exists will cause crash.

Fixes #6426

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -15690,7 +15690,7 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
 
   // if this FD isn't the entry point, then we shouldn't add
   // a shader attribute to this decl, so just return
-  if (EntryPointName != FD->getIdentifier()->getName()) {
+  if (FD->getIdentifier() && EntryPointName != FD->getIdentifier()->getName()) {
     return;
   }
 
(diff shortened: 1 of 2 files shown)


bob80905:
For your test case, I would think only one function declaration is needed. Is it possible to get rid of the rest of them?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15690,7 +15690,7 @@ void TryAddShaderAttrFromTargetProfile(Sema &S, FunctionDecl *FD,
 
   // if this FD isn't the entry point, then we shouldn't add
   // a shader attribute to this decl, so just return
-  if (EntryPointName != FD->getIdentifier()->getName()) {
+  if (FD->getIdentifier() && EntryPointName != FD->getIdentifier()->getName()) {
```
tex3d, line 15693: I don't think this is right.  We should return if there is no identifier as well, since it can't be an entry point.

```suggestion
    if (!FD->getIdentifier() || EntryPointName != FD->getIdentifier()->getName()) {
```

llvm-beanz (approved):

tex3d (approved):
LGTM.

PR #6445, opened 2024-03-22
[SPIR-V] Implement dot2add intrinsic function

### LINKED ISSUE #6442: [SPIR-V] Add support for dot2add
The `dot2add` function is currently unimplemented in the SPIR-V backend. This was a Shader Model 6.4 feature so we should close this gap.

sudonatalie:
Add support for the Shader Model 6.4 intrinsic dot2add which was missing from the SPIR-V backend.

Fixes #6442

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -11535,17 +11535,25 @@ SpirvInstruction *SpirvEmitter::processIntrinsicDP2a(const CallExpr *callExpr) {
 
   assert(callExpr->getNumArgs() == 3u);
 
-  QualType vecType = callExpr->getArg(0)->getType();
+  const Expr *arg0 = callExpr->getArg(0);
+  const Expr *arg1 = callExpr->getArg(1);
+  const Expr *arg2 = callExpr->getArg(2);
+
+  QualType vecType = arg0->getType();
   QualType componentType = {};
   uint32_t vecSize = {};
   bool isVec = isVectorType(vecType, &componentType, &vecSize);
 
   assert(isVec && vecSize == 2);
+  (void)isVec;
+
+  SpirvInstruction *arg0Instr = doExpr(arg0);
+  SpirvInstruction *arg1Instr = doExpr(arg1);
+  SpirvInstruction *arg2Instr = doExpr(arg2);
 
   // Create the dot product of the half2 vectors.
   SpirvInstruction *dotInstr = spvBuilder.createBinaryOp(
-      spv::Op::OpDot, componentType, doExpr(callExpr->getArg(0)),
-      doExpr(callExpr->getArg(1)), loc, range);
+      spv::Op::OpDot, componentType, arg0Instr, arg1Instr, loc, range);
 
   // Convert dot product (half type) to result type (float).
   QualType resultType = callExpr->getType();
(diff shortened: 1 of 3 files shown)


cassiebeckley (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11499,6 +11503,44 @@ SpirvInstruction *SpirvEmitter::processIntrinsicDP4a(const CallExpr *callExpr,
...
+  auto range = callExpr->getSourceRange();
+
+  assert(callExpr->getNumArgs() == 3u);
+
+  QualType vecType = callExpr->getArg(0)->getType();
+  QualType componentType = {};
+  uint32_t vecSize = {};
+  bool isVec = isVectorType(vecType, &componentType, &vecSize);
+
+  assert(isVec && vecSize == 2);
```
Keenuts, line 11526: Shouldn't both arguments be checked? (or all 3?)

tools/clang/test/CodeGenSPIRV/intrinsics.dot2add.hlsl
```diff
@@ -0,0 +1,14 @@
...
+
+// CHECK: [[inputs:%[0-9]+]] = OpLoad %v4half %in_var_Inputs0
+// CHECK: [[acc:%[0-9]+]] = OpLoad %int %in_var_Acc0
+// CHECK: [[b:%[0-9]+]] = OpVectorShuffle %v2half [[inputs]] [[inputs]] 2 3
+// CHECK: [[a:%[0-9]+]] = OpVectorShuffle %v2half [[inputs]] [[inputs]] 0 1
+// CHECK: [[dot:%[0-9]+]] = OpDot %half [[a]] [[b]]
+// CHECK: [[dot2:%[0-9]+]] = OpFConvert %float [[dot]]
+// CHECK: [[acc2:%[0-9]+]] = OpConvertSToF %float [[acc]]
+// CHECK: [[res:%[0-9]+]] = OpFAdd %float [[dot2]] [[acc2]]
+  return dot2add(inputs.xy, inputs.zw, acc);
```
Keenuts, line 13: Shall we test calls with implicit conversion, and other cases to make sure we handled them correctly?

Seems like passing a single `float` for the first 2 params works by broadcasting it to a `half2`, and it seems to allow any type, just warn about narrowing cast.
Sending larger vectors is also allowed, an extraction is done.
However, the accumulator HAS to be a single scalar (any type)

tools/clang/lib/SPIRV/SpirvEmitter.cpp
sudonatalie: This is just an extra assert on things that are already type checked earlier in execution. We don't typically re-check every attribute of every arg here, but if there's anything specific you think might not be caught earlier or should be rechecked I can add it.

tools/clang/test/CodeGenSPIRV/intrinsics.dot2add.hlsl
sudonatalie: Added some more tests

tools/clang/lib/SPIRV/SpirvEmitter.cpp
sudonatalie: Actually this needs another fixup for the release build, since isVec is an unused variable. Working on it.

sudonatalie: Fixed

Keenuts: Right. Not a strong opinion, but to me it seems weird to check only one parameter, and trust the other. I'd either check none (assuming some check upstream exists), or check all 3 for the correct type.

sudonatalie: Ah, this bit of logic is mainly to get the component type, which is used for the dot product. I see how it seems a bit random otherwise.

Issue #6383, opened 2024-03-04
[SPIR-V] Verify behavior of GetAttributeAtVertex


sudonatalie:
**Description**
Follow-up from https://github.com/microsoft/DirectXShaderCompiler/pull/6041#issuecomment-1829775849

**Steps to Reproduce**

```
dxc -T ps_6_6 -spirv
```

```hlsl
float foo(int c, float v) {
  if (c == 0)
    return GetAttributeAtVertex(v, 1);
  else
    return v;
}

float4 main(nointerpolation float a : A) : SV_Target
{
  float tmp1 = GetAttributeAtVertex(a, 2);
  float tmp2 = foo(0, a);
  float tmp3 = foo(1, a);
  return float4(0, 0, 0, tmp1 + tmp2 + tmp3);
}
```

**Actual Behavior**
Once #6041 is merged, the above sample will successfully compile for both DXIL and SPIR-V, however, as @Keenuts mentioned:

> DXIL seems to only call the intrinsic for tmp1 and tmp2, but tmp3 gets loaded as usual, but with an undefined vertex index. While for SPIR-V, we default to the index 0.

**Expected Behavior**
We should determine whether this behavior is well-defined for DXIL then verify that that SPIR-V backend implementation is matches.

pow2clk:
I'm surprised we don't see errors for this since it's calling GetAttributeAtVertex from a subfunction.

sudonatalie:
@pow2clk If calling `GetAttributeAtVertex` outside the main entrypoint function is illegal, adding a check and error messaging for that would resolve a lot of our issues, but we didn't find any docs explicitly stating that. It looks like the compiler allows it (with various degrees of functionality) for both DXIL and SPIR-V right now.

Keenuts:
This is now fixed (https://github.com/microsoft/DirectXShaderCompiler/commit/27b87b75937f4d3fff0c486bc9bba6540c3bddb7)

Issue #6382, opened 2024-03-04
[SPIR-V] GetAttributeAtVertex(0,0) fails validation


sudonatalie:
**Description**
Follow-up from https://github.com/microsoft/DirectXShaderCompiler/pull/6041#issuecomment-1829775849

This invalid usage of `GetAttributeAtVertex` fails at validation rather than providing a useful error message.

**Steps to Reproduce**
```
dxc -T ps_6_0 -spirv
```

```
float main() : SV_Target
{
  return GetAttributeAtVertex(0, 0);
}
```

**Actual Behavior**
SPIR-V compilation fails with a validation error message:

```
fatal error: generated SPIR-V is invalid: The Base <id> '7[%int_0]' in OpAccessChain instruction must be a pointer.
  %14 = OpAccessChain %_ptr_Function_int %int_0 %uint_0
```

DXIL fails with:

```
<source>:6:10: error: attribute evaluation can only be done on values taken directly from inputs.
  return GetAttributeAtVertex(0, 0);
         ^
``` 

The SPIR-V backend should produce a similarly helpful error message.

sudonatalie:
Related: https://github.com/microsoft/hlsl-specs/issues/181

Keenuts:
This is now fixed:

```
./build/bin/dxc -T ps_6_0 -spirv ./test.hlsl
./test.hlsl:3:31: error: parameter 0 of GetAttributeAtVertex must have a 'nointerpolation' attribute
  return GetAttributeAtVertex(0, 0);
                              ^
```

Issue #6220, opened 2024-01-30
GetAttributeAtVertex() doesn't work on function parameters


damyanp:
**Description**

When using an array member of a VS interpolant structure, `GetAttributeAtVertex` fails to compile.

**Steps to Reproduce**

```hlsl
// dxc.exe repro.hlsl /Tps_6_1 /EMainPS

cbuffer Globals : register (b0)
{
    int GIndex;
};

struct Interpolants
{
	float3 BaryWeights : SV_Barycentrics;
	nointerpolation float4 PackedTexCoords[1] : COLOR0;
	nointerpolation uint PrimitiveId : PRIMITIVE_ID;
};

float4 GetUV(Interpolants vs, int index)
{
	float4 a1 = GetAttributeAtVertex(vs.PackedTexCoords[index], 0);
	float4 a2 = GetAttributeAtVertex(vs.PackedTexCoords[index], 1);
	float4 a3 = GetAttributeAtVertex(vs.PackedTexCoords[index], 2);

	return a1 * vs.BaryWeights.x + a2 * vs.BaryWeights.y + a3 * vs.BaryWeights.z;
}

[RootSignature("CBV(b0)")]
float4 MainPS(Interpolants vs) : SV_Target0
{
    float4 x = GetUV(vs, 0); // FAIL
    return x;
}
```

**Actual Behavior**

```
gaav-repro.hlsl:19:14: error: attribute evaluation can only be done on values taken directly from inputs.
        float4 a3 = GetAttributeAtVertex(vs.PackedTexCoords[index], 2);
                    ^
gaav-repro.hlsl:18:14: error: attribute evaluation can only be done on values taken directly from inputs.
        float4 a2 = GetAttributeAtVertex(vs.PackedTexCoords[index], 1);
                    ^
gaav-repro.hlsl:17:14: error: attribute evaluation can only be done on values taken directly from inputs.
        float4 a1 = GetAttributeAtVertex(vs.PackedTexCoords[index], 0);
                    ^
```

Note that moving the GetAttributeAtVertex calls to be inside `MainPS` can compile successfully.

**Environment**
- DXC version dxcompiler.dll: 1.7 - 1.7.2308.7 (69e54e290); dxil.dll: 1.7(101.7.2308.12)
- Host Operating System Windows 11

tex3d:
Unfortunately, this one is not sound language-wise, and tricky to support.  It's likely going to be considered by-design.

[...]

PR #6453, opened 2024-03-25
[SPIR-V][Sema] Requires nointerpolation on GetAttributeAtVertex param #0

### LINKED ISSUE #6384: [SPIR-V] GetAttributeAtVertex on struct member without nointerpolation fails validation
**Description**
Follow-up from #6041

**Steps to Reproduce**
```
dxc -T ps_6_6 -spirv
```

```
struct S {
  float4 a : COLOR;
};

float compute(nointerpolation float4 a) {
  return GetAttributeAtVertex(a, 2)[0];
}

float4 main(S s) : SV_TARGET
{
  return float4(0, 0, 0, compute(s.a));
}
```

**Actual Behavior**
Once https://github.com/microsoft/DirectXShaderCompiler/pull/6041 is merged, the above sample will fail at validation for both DXIL and SPIR-V, but the SPIR-V backend needs better error messaging.

DXIL:
```
error: validation errors
<source>:6:10: error: Attribute COLOR must have nointerpolation mode in order to use GetAttributeAtVertex function.
note: at '%1 = call float @dx.op.attributeAtVertex.f32(i32 137, i32 0, i32 0, i8 0, i8 2)' in block '#0' of function 'main'.
Validation failed.
```

SPIR-V:
```
fatal error: generated SPIR-V is invalid: Result type (OpTypeVector) does not match the type that results from indexing into the composite (OpTypeFloat).
  %13 = OpCompositeExtract %v4float %12 2
```

Keenuts:
This commit adds new sema checks for SPIR-V to restrict the scope of the feature: the `nointerpolation` attribute is now required on all parameters used in GetAttributeAtVertex, and doesn't propagage magically to parent/child functions.

The current accepted cases this now blocks were mostly broken later in the codegen, so this should be an acceptable 'regression'.

Related to #6220
Fixes #6384, #6383, #6382

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7735,6 +7735,8 @@ def warn_hlsl_implicit_vector_truncation  : Warning<
   InGroup<Conversion>, DefaultWarn;
 def err_hlsl_nointerpolation_and_linear : Error<
     "nointerpolation cannot be used with any other interpolation mode specifier">;
+def err_hlsl_parameter_requires_attribute : Error<
+    "parameter %0 of %1 must have a a '%2' attribute">;
 def warn_hlsl_duplicate_specifier : Warning<
     "duplicate HLSL specifier %0">,
 	InGroup<IgnoredAttributes>, DefaultWarn;
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7736,7 +7736,7 @@ def warn_hlsl_implicit_vector_truncation  : Warning<
 def err_hlsl_nointerpolation_and_linear : Error<
     "nointerpolation cannot be used with any other interpolation mode specifier">;
 def err_hlsl_parameter_requires_attribute : Error<
-    "parameter %0 of %1 must have a a '%2' attribute">;
+    "parameter %0 of %1 must have a '%2' attribute">;
 def warn_hlsl_duplicate_specifier : Warning<
     "duplicate HLSL specifier %0">,
 	InGroup<IgnoredAttributes>, DefaultWarn;
--- a/tools/clang/lib/Sema/SemaExpr.cpp
+++ b/tools/clang/lib/Sema/SemaExpr.cpp
@@ -5290,8 +5290,6 @@ Sema::BuildResolvedCallExpr(Expr *Fn, NamedDecl *NDecl,
       return ExprError();
     if (CheckHLSLFunctionCall(FDecl, TheCall))
       return ExprError();
-    if (hlsl::IsIntrinsicOp(FDecl) && CheckHLSLIntrinsicCall(FDecl, TheCall))
-      return ExprError();
     if (BuiltinID)
       return CheckBuiltinFunctionCall(FDecl, BuiltinID, TheCall);
   } else if (NDecl) {
--- a/tools/clang/lib/Sema/SemaExpr.cpp
+++ b/tools/clang/lib/Sema/SemaExpr.cpp
@@ -5288,7 +5288,10 @@ Sema::BuildResolvedCallExpr(Expr *Fn, NamedDecl *NDecl,
   if (FDecl) {
     if (CheckFunctionCall(FDecl, TheCall, Proto))
       return ExprError();
-
+    if (CheckHLSLFunctionCall(FDecl, TheCall))
+      return ExprError();
+    if (hlsl::IsIntrinsicOp(FDecl) && CheckHLSLIntrinsicCall(FDecl, TheCall))
+      return ExprError();
     if (BuiltinID)
       return CheckBuiltinFunctionCall(FDecl, BuiltinID, TheCall);
   } else if (NDecl) {
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -15227,8 +15227,10 @@ bool Sema::CheckHLSLIntrinsicCall(FunctionDecl *FDecl, CallExpr *TheCall) {
     // to limit the scope, and fail gracefully in some cases.
     if (!getLangOpts().SPIRV)
       return false;
-    if (FDecl->getName() != "GetAttributeAtVertex")
-      return false;
+    // This should never happen for SPIR-V. But on the DXIL side, extension can be added by
+    // inserting new intrinsics, meaning opcodes can collide with existing ones.
+    // See the ExtensionTest.EvalAttributeCollision test.
+    assert(FDecl->getName() == "GetAttributeAtVertex");
     return CheckIntrinsicGetAttributeAtVertex(this, FDecl, TheCall);
   default:
     break;
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -15184,6 +15184,83 @@ QualType Sema::getHLSLDefaultSpecialization(TemplateDecl *Decl) {
   return QualType();
 }
 
+static bool isRelatedDeclMarkedNointerpolation(Expr *E) {
+  if (!E)
+    return false;
+  E = E->IgnoreCasts();
+  if (auto *DRE = dyn_cast<DeclRefExpr>(E))
+    return DRE->getDecl()->hasAttr<HLSLNoInterpolationAttr>();
+
+  if (auto *ME = dyn_cast<MemberExpr>(E))
+    return ME->getMemberDecl()->hasAttr<HLSLNoInterpolationAttr>() ||
+           isRelatedDeclMarkedNointerpolation(ME->getBase());
+
+  if (auto *HVE = dyn_cast<HLSLVectorElementExpr>(E))
+    return isRelatedDeclMarkedNointerpolation(HVE->getBase());
+
+  if (auto *ASE = dyn_cast<ArraySubscriptExpr>(E))
+    return isRelatedDeclMarkedNointerpolation(ASE->getBase());
+
+  return false;
+}
+
+static bool CheckIntrinsicGetAttributeAtVertex(Sema *S, FunctionDecl *FDecl,
+                                               CallExpr *TheCall) {
+  assert(TheCall->getNumArgs() > 0);
+  auto argument = TheCall->getArg(0)->IgnoreCasts();
+
+  if (!isRelatedDeclMarkedNointerpolation(argument)) {
+    S->Diag(argument->getExprLoc(), diag::err_hlsl_parameter_requires_attribute)
+        << 0 << FDecl->getName() << "nointerpolation";
+    return true;
+  }
+
+  return false;
+}
+
+bool Sema::CheckHLSLIntrinsicCall(FunctionDecl *FDecl, CallExpr *TheCall) {
+  auto attr = FDecl->getAttr<HLSLIntrinsicAttr>();
+
+  switch (hlsl::IntrinsicOp(attr->getOpcode())) {
+  case hlsl::IntrinsicOp::IOP_GetAttributeAtVertex:
+    // See #hlsl-specs/issues/181. Feature is broken. For SPIR-V we want
+    // to limit the scope, and fail gracefully in some cases.
+    if (!getLangOpts().SPIRV)
+      return false;
+    if (FDecl->getName() != "GetAttributeAtVertex")
+      return false;
+    return CheckIntrinsicGetAttributeAtVertex(this, FDecl, TheCall);
+  default:
+    break;
+  }
+
+  return false;
+}
+
+bool Sema::CheckHLSLFunctionCall(FunctionDecl *FDecl, CallExpr *TheCall) {
+  // See #hlsl-specs/issues/181. Feature is broken. For SPIR-V we want
+  // to limit the scope, and fail gracefully in some cases.
+  if (!getLangOpts().SPIRV)
+    return false;
+
+  bool error = false;
+  for (unsigned i = 0; i < FDecl->getNumParams(); i++) {
+    assert(i < TheCall->getNumArgs());
+
+    if (!FDecl->getParamDecl(i)->hasAttr<HLSLNoInterpolationAttr>())
+      continue;
+
+    if (!isRelatedDeclMarkedNointerpolation(TheCall->getArg(i))) {
+      Diag(TheCall->getArg(i)->getExprLoc(),
+           diag::err_hlsl_parameter_requires_attribute)
+          << i << FDecl->getName() << "nointerpolation";
+      error = true;
+    }
+  }
+
+  return error;
+}
+
 namespace hlsl {
 
 static bool nodeInputIsCompatible(DXIL::NodeIOKind IOType,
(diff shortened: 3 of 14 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15176,6 +15176,83 @@ QualType Sema::getHLSLDefaultSpecialization(TemplateDecl *Decl) {
...
+bool Sema::CheckHLSLIntrinsicCall(FunctionDecl *FDecl, CallExpr *TheCall) {
+  auto attr = FDecl->getAttr<HLSLIntrinsicAttr>();
+
+  switch (hlsl::IntrinsicOp(attr->getOpcode())) {
+  case hlsl::IntrinsicOp::IOP_GetAttributeAtVertex:
+    // See #hlsl-specs/issues/181. Feature is broken. For SPIR-V we want
+    // to limit the scope, and fail gracefully in some cases.
+    if (!getLangOpts().SPIRV)
+      return false;
+    if (FDecl->getName() != "GetAttributeAtVertex")
```
llvm-beanz, line 15222: Is this possible? I assume if the intrinsic op is set it should be this function.

Keenuts: It just for a weird test in DXC, if an extension registers an instrinsic and there is a name conflict.

llvm-beanz: I'm confused. Is there a test case where this fails? If we've assigned an intrinsic op then we're telling the compiler to treat it as that intrinsic regardless of the name of the function. I don't see anywhere else in DXC where we also separately need to verify the name of the function.

Keenuts: IIRC the test was `ExtensionTest.EvalAttributeCollision`, which seems to register an intrinsic (collide_proc), with the same opcode as a registered intrinsic.
So here I hit the condition because the opcodes are the same, but the name mismatch.

Keenuts:
Rebases on main to fix the bots.

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: I see the test now. @adam-yang wrote it. I'm unsure what the context there is (although it terrifies me).

It shouldn't actually matter here though since that is a DXIL-only test and right above here you return if you're targeting DXIL, so this code shouldn't be possible to hit. We could make this an assert if you're concerned that it might come up for SPIR-V, but I think the inline SPIR-V proposal provides a more sane way to expose extensions than inserting a table of overridden intrinsics.

Keenuts: Of course, forgot I gated this on `isSPIRV`. Cleaning that up now, adding an assert.

tools/clang/test/SemaHLSL/getattribute-at-vertex.missing.nointerpolation.call.hlsl
```diff
@@ -0,0 +1,11 @@
+// RUN: dxc -T ps_6_2 -E main %s -spirv -verify
```
hekota, line 1: The SemaHLSL tests need to run on both spirv and dxil backends. Could you please remove the -spirv argument?

hekota: I see that this is a sema check specific to SPIR-V. In that case it would probably be best to have these tests in a subfolder with `lit.local.cfg` file that would have `config.unsupported = not config.spirv`.

llvm-beanz: We can solve this by putting `// REQUIRES: spirv` in the test. That will make it only run if the spir-v code generator is enabled.

Keenuts: There is one thing I don't get:
If we never add tests with `-spirv`, like in https://github.com/microsoft/DirectXShaderCompiler/pull/6467, we essentially lose coverage for the whole SPIR-V compilation side no?

As in, today, SEMA is shared, but if one day something goes bad, and for some reason sema diagnostics don't show up with `-spirv`, none of those tests would catch it.
Shouldn't all the sema tests also have a run line with `-spirv`?

llvm-beanz: Counter point: If you look at clang tests for parsing and semantic analysis of C++ code don't have separate run lines for every target that LLVM supports.

Adding loads of duplicated test coverage that provides no additional meaningful test coverage does nothing but slow down development.

Keenuts: I see, yes, that would be unscalable. Thanks!

Keenuts: Added the `// REQUIRED: spirv` to the 2 impacted SEMA tests.

llvm-beanz (approved):

tools/clang/test/CodeGenSPIRV/intrinsics.get-attribute-at-vertex.cbuf.var.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: dxc -T ps_6_2 -E main -spirv %s | FileCheck %s
```
sudonatalie, line 1: Should be `RUN: %dxc`. Same applies for a few more tests below.

damyanp:
> Hello! Limited the scope of this change to the SPIR-V backend. Added some tests to make sure the DXIL side always catch those as part of the validation.

Could you update the PR description to reflect the change in scope please?

damyanp:
Also, can you confirm that with this change scoped to just affect SPIR-V, does it really address #6220?, or should we remove that from the description?

damyanp (changes requested):
Can you update the description to clarify if this impacts just SPIR-V or DXIL as well?

If it is just SPIR-V, can you double check the list of issues fixed by this, since one of them is a DXIL issue.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7732,6 +7732,8 @@ def warn_hlsl_implicit_vector_truncation  : Warning<
   InGroup<Conversion>, DefaultWarn;
 def err_hlsl_nointerpolation_and_linear : Error<
     "nointerpolation cannot be used with any other interpolation mode specifier">;
+def err_hlsl_parameter_requires_attribute : Error<
+    "parameter %0 of %1 must have a a '%2' attribute">;
```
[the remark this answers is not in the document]
Keenuts: Ah nice catch.. Thanks!

tools/clang/test/CodeGenSPIRV/intrinsics.get-attribute-at-vertex.cbuf.var.hlsl
Keenuts: Ah yes, fixed!

Keenuts:
> Can you update the description to clarify if this impacts just SPIR-V or DXIL as well?
> 
> If it is just SPIR-V, can you double check the list of issues fixed by this, since one of them is a DXIL issue.

Right, forgot about the description.
Updated the issue list to only reference the one impacting both DXIL and SPIR-V, and changed title+description to say this only affects SPIR-V.
Thanks!

tools/clang/test/SemaHLSL/getattribute-at-vertex.legal.hlsl
```diff
@@ -0,0 +1,41 @@
+// RUN: %dxc -T ps_6_2 -E main %s -verify
+
+// expected-no-diagnostics
+struct S1 {
+  nointerpolation float3 f1;
```
tex3d, line 5: I think there are missing cases where a struct field is marked nointerpolation and the field is passed to a function.  Then if you create a temporary of that structure in main and pass that field to a function.  What happens in these cases?

tex3d (commented):
I have some concerns with the approach, but realize it is (likely too) late to voice them.  Since it's only enabled for SPIR-V, we have less grounds for objection, but we should be clear that this is not likely to be the solution for validating this scenario going forward.

`GetAttributeAtVertex` issues are in a class shared with `EvaluateAttribute*` intrinsics.  This change can't impact/help those intrinsics for those similar use cases.

At a high-level, I would prefer a warning diagnostic (default error on SPIR-V) that catches any use of this class of intrinsic outside the entry point (and reachable from an active entry point).  This would stop catch a lot more invalid use cases that we could not easily diagnose in SemaHLSL.  It would hit some cases that work today, but as a warning defaulting to error, it could be disabled for cases that are known to work for someone's specific scenario.  The idea is to put the warning up where we know the language can't handle this reliably today, even though it works in some cases today due to some unreliable/unpredictable "magic".

The language implications aren't consistent with HLSL's copy-in parameter passing, or the limited scope where interpolation modifiers have any meaning.  I think there are legitimate scenarios this would block, and other illegal ones that would be allowed through this check.  I think it would be good to at least make the diagnostic a warning, which defaults to error (and has a unique warning group).  This would allow someone to work around a blocked scenario when it doesn't impact `GetAttributeAtVertex`.

Also, how does this deal with implicitly `nointerpolation` inputs - any values of non-float types.  There's no requirement to use `nointerpolation` for those.

I can think of scenarios that might pass the diagnostic, and I don't know what the behavior will be for these.  I think we could use a few more some test cases for this.

damyanp (approved):
Thanks for updating the description.  It looks like Tex has also added some notes on the code.

tools/clang/lib/Sema/SemaExpr.cpp
```diff
@@ -5288,7 +5288,10 @@ Sema::BuildResolvedCallExpr(Expr *Fn, NamedDecl *NDecl,
   if (FDecl) {
     if (CheckFunctionCall(FDecl, TheCall, Proto))
       return ExprError();
-
+    if (CheckHLSLFunctionCall(FDecl, TheCall))
+      return ExprError();
+    if (hlsl::IsIntrinsicOp(FDecl) && CheckHLSLIntrinsicCall(FDecl, TheCall))
+      return ExprError();
```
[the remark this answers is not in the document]
Keenuts: What do you mean? Is it called below it, only gated by the `IsInstrinsicOp` call.
Am I missing something?

tools/clang/test/SemaHLSL/getattribute-at-vertex.legal.hlsl
Keenuts: Do you mean this case:

```hlsl
struct S6 {
  nointerpolation uint f1;
};

float4 main(S6 s6 : G) : SV_Target {
  S6 s6_copy = s6;
  float v8 = GetAttributeAtVertex(s6_copy.f1, 0);
  
  
  S6 s6_copy_2;
  s6_copy_2.f1 = s6.f1;
  float v8 = GetAttributeAtVertex(s6_copy_2.f1, 0);
}
```

If so, I can add those 2 cases. They pass the check as legal, but on the SPIR-V side, they crash (but that's another story 🙃)

Keenuts:
> At a high-level, I would prefer a warning diagnostic (default error on SPIR-V) that catches any use of this class of intrinsic outside the entry point (and reachable from an active entry point). This would stop catch a lot more invalid use cases that we could not easily diagnose in SemaHLSL.

I'd be fine with blocking that too.

> I think there are legitimate scenarios this would block, and other illegal ones that would be allowed through this check. 

My point of view on that is: the feature cannot be implemented. It is ill-defined to have something clean.
We tried to implement it still, but we have many edge cases we fail to handle, and as a result, the compiler either crashes, or errors with a validation error (which on the the SPIR-V side, is considered a compiler bug).

If I'm building an engine, and use DXC, I'd rather see a feature be disabled with a clear error message, than having to deal with validation errors, compiler instabilities, or guess in which scenario can this feature be used.

But I understand removing the feature is a bit extreme, so what I'm trying is to at least drastically reduce the scope of the feature, limiting the magic required to make it work.

Only allowing the function in the entrypoint would for example go in that direction: remove most of the ill-defined scenarios, to have something predictable and usable.

tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
```diff
@@ -56,6 +56,11 @@ int PervertexInputVisitor::appendIndexZeroAt(
 ///< treated as nointerpolated too.
 bool PervertexInputVisitor::expandNointerpVarAndParam(
     SpirvInstruction *spvInst) {
+  // If there is no AST type, it means the type was already lowered from some
+  // expression/construct. If it required expansion, it should be done already.
+  if (!spvInst->hasAstResultType())
+    return spvInst->isNoninterpolated();
```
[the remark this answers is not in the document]
Keenuts: Rebased on main to clean this up.

sudonatalie:
> My point of view is: the feature is ill-defined, and cannot be implemented as is. Knowing that, we tried to implement it. As a result, we have many edge cases we fail to handle: the compiler either crashes, or displays a validation error (which on the the SPIR-V side, is considered a compiler bug).

+1. We've spent a lot of time trying to figure out a reasonable way to match the DXIL behavior in SPIR-V and it's just not feasible with the current semantics. This PR will get the SPIR-V backend to a better state of erroring instead of crashing or throwing unparseable validation errors. If we get to a re-design in https://github.com/microsoft/hlsl-specs/issues/181 then we'll have a good foundation to implement the SPIR-V backend from but I think this is the best option in the meantime.

sudonatalie (approved):

tools/clang/lib/Sema/SemaExpr.cpp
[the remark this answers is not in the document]
tex3d: I meant that we could have one hook function into SemaHLSL (CheckHLSLFunctionCall), and the knowledge of using IsIntrinsicOp and CheckHLSLIntrinsicCall would be localized to CheckHLSLFunctionCall in SemaHLSL, instead of requiring this to be known in SemaExpr.  Also, CheckHLSLIntrinsicCall wouldn't have to be part of Sema.  That is unless you can think of a place that CheckHLSLIntrinsicCall would need to be called from Sema without calling CheckHLSLFunctionCall.

tools/clang/test/SemaHLSL/getattribute-at-vertex.legal.hlsl
tex3d: You don't seem to have the common, legal case of passing `s6.f1` directly to a function that uses `GetAttributeAtVertex`, unless that was covered by some other existing test case I didn't see.  The temporary variable case was just an extension of this to see how it's handled.

tools/clang/lib/Sema/SemaExpr.cpp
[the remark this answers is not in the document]
Keenuts: Ah right, now I understand the 'under' 😊 Moved the call inside the CheckHLSLFunctionCall!

tools/clang/test/SemaHLSL/getattribute-at-vertex.legal.hlsl
Keenuts: Right, added one line to check this.

PR #6459, opened 2024-03-26
[SPIR-V] Fix crash in GetAttributeAtVertex impl.

### LINKED ISSUE #2955: [SPIR-V] Silent compiler crash when using GetAttributeAtVertex

Keenuts:
Some SPIR-V types are not direct translations of an AST type, hence they don't have an AST type attached to it.
The code didn't checked for this possibility when visiting all variables.

Fixes #2955

### DIFF
--- a/tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
+++ b/tools/clang/lib/SPIRV/PervertexInputVisitor.cpp
@@ -56,6 +56,11 @@ int PervertexInputVisitor::appendIndexZeroAt(
 ///< treated as nointerpolated too.
 bool PervertexInputVisitor::expandNointerpVarAndParam(
     SpirvInstruction *spvInst) {
+  // If there is no AST type, it means the type was already lowered from some
+  // expression/construct. If it required expansion, it should be done already.
+  if (!spvInst->hasAstResultType())
+    return spvInst->isNoninterpolated();
+
   QualType type = spvInst->getAstResultType();
   bool isExpanded = false;
   auto typePtr = type.getTypePtr();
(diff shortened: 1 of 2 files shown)


sudonatalie (approved):
TBH I don't know for sure if the "If it required expansion, it should be done already." is true, but since it only affects shaders that would otherwise crash and this bug is super hairy, LGTM. Thanks for investigating this!

cassiebeckley (approved):

Keenuts:
The reason is:
For constants, the AST has multiple variables with their own type.
But what we do is create a struct which contains all those constants.

Because we created this struct from scratch to contain the multiple constants we have, there is no AST type to match against.
Each field has an AST type, but the struct itself has none.

Hence why this condition: if there is no AST type, it means it's a "synthetic" type we created in the backend.

PR #6460, opened 2024-03-26
[Sema] Move outputcontrolpoints check to Sema

### LINKED ISSUE #6387: [SPIR-V] Move check to verify PatchConstant output patch element count to Sema
The DXIL backend checks the PatchConstant output patch element count.
This is not checked on the SPIR-V backend.

This means we accept code DXIL doesn't. This should not happen, and in this case, the check should be moved to Sema so both backends can benefit from it.

Follow up from https://github.com/microsoft/DirectXShaderCompiler/pull/6363

Keenuts:
DXIL had a check in the backend for those values.
Moving it to Sema to share it between DXIL and SPIR-V.

Fixes #6387

### DIFF
--- a/tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
+++ b/tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
@@ -1,7 +1,6 @@
 // RUN: %dxc -E main -T hs_6_0  %s | FileCheck %s
 
 // CHECK-DAG: Patch Constant function HSPerPatchFunc should not have inout param
-// CHECK-DAG: Patch constant function's output patch input should have 3 elements, but has 5.
 
 //--------------------------------------------------------------------------------------
 // SimpleTessellation.hlsl
@@ -62,7 +61,7 @@ HSPerPatchData HSPerPatchFunc( const InputPatch< PSSceneIn, 12 > points, OutputP
 [partitioning("fractional_odd")]
 [outputtopology("triangle_cw")]
 [patchconstantfunc("HSPerPatchFunc")]
-[outputcontrolpoints(3)]
+[outputcontrolpoints(5)]
 HSPerVertexData main(const uint id : SV_OutputControlPointID,
                       const InputPatch< PSSceneIn, 12 > points )
 {
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7837,8 +7837,8 @@ def err_hlsl_missing_patch_constant_function : Error<
     "patch constant function '%0' must be defined">;
 def err_hlsl_patch_reachability_not_allowed : Error<
     "%select{patch constant|entry}0 function '%1' should not be reachable from %select{patch constant|entry}2 function '%3'">;
-def err_hlsl_patch_input_size_mismatch : Error<
-    "Patch constant function's input patch input should have %0 elements, but has %1.">;
+def err_hlsl_patch_size_mismatch : Error<
+    "Patch constant function's %0 patch input should have %1 elements, but has %2.">;
 def warn_hlsl_structurize_exits_lifetime_markers_conflict : Warning <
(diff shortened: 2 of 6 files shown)


tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
```diff
@@ -62,7 +61,7 @@ HSPerPatchData HSPerPatchFunc( const InputPatch< PSSceneIn, 12 > points, OutputP
 [partitioning("fractional_odd")]
 [outputtopology("triangle_cw")]
 [patchconstantfunc("HSPerPatchFunc")]
-[outputcontrolpoints(3)]
```
sudonatalie, line 65: Assuming this had to be changed just because it's an earlier return now?

sudonatalie (approved):
LGTM, but should have a review from someone at MS too

cassiebeckley (approved):

tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
Keenuts: Yes that's correct. One is Sema, the other is at a later stage in the DXC side.

damyanp:
@bob80905 - as you've been moving various things into Sema you might have some useful context here.  Can you have a look please?

tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
pow2clk: I don't follow that explanation. This is an end-to-end compilation. Sema is still being invoked, why isn't the error firing anymore?

Keenuts: Sorry, mixed up things:
the error is indeed firing as expected. But the check being moved to Sema, I added a new test (`tools/clang/test/CodeGenDXIL/semantic.hull-output.size-mismatch.hs.hlsl`) which covers it from `-verify`, hence didn't wanted to check it again here using `FileCheck`

bob80905:
I think the implementation (the diagnostic emission part) should be moved to SemaHLSL.cpp:DiagnoseHullEntry.
My understanding is that the code in SemaHLSLDiagnoseTU is for generic checks on constraints we've placed on HLSL, like recursion. SemaHLSL.cpp has finer scopes on when to diagnose what, and I think DiagnoseHullEntry is exactly the place where a Hull-only attribute should be validated / diagnosed. 

I also think there should be a test that tests for the diagnostic: 
`/* expected-error{{Patch constant function's input patch input should have 5 elements, but has 2.}} */`
unless there already exists a test that tests for the input patch variant of that diagnostic.

tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
Keenuts: Makes sense, reverted this.

Keenuts:
> I think the implementation (the diagnostic emission part) should be moved to SemaHLSL.cpp:DiagnoseHullEntry. 

The issue with that is we are parsing function declaration at this state.
This means I don't have yet the function definition of the other functions. Meaning I cannot check the output/input size matches the other function input/output size.

This is probably why this was done at the TU level, later.


> I also think there should be a test that tests for the diagnostic: `/* expected-error{{Patch constant function's input patch input should have 5 elements, but has 2.}} */`

Yes, and there is already one (`test/CodeGenSPIRV/semantic.hull-input.size-mismatch.hs.hlsl`) added when the input check was moved to SEMA.

tools/clang/test/HLSLFileCheck/samples/SimpleHs12.hlsl
Keenuts: So, turns out I confused my confusion:
The error doesn't fire anymore if I don't fix the `outputcontrolpoint` first.
The `outputcontrolpoint` is a SEMA error, while the other is a DXIL codegen error.
This means the compilation stops earlier, so we don't show both errors at the same time.

bob80905 (approved):

PR #6465, opened 2024-03-26
Ready tests for 6.9 version

pow2clk:
In experimenting with bumping the max shader model version in the mesh nodes branch, many unrelated tests started failing due to being overly-sensitive to ordering, particularly for metadata. This uses regular expressions or order-independent checks to make these tests more robust so they will be able to handle the change to 6.9 when it comes to main, whatever form that takes.

Followup to #6432


tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4310,88 +4320,88 @@ TEST_F(ValidationTest, NodeInputCompatibility) {
   RewriteAssemblyCheckMsg(
       L"..\\DXILValidation\\node_input_compatibility.hlsl", "lib_6_8",
       pArguments.data(), 2, nullptr, 0,
-      {"!13 = !{i32 1, i32 97"}, // DispatchNodeInputRecord
-      {"!13 = !{i32 1, i32 65"}, // GroupNodeInputRecords
+      {"= !{i32 1, i32 97"}, // DispatchNodeInputRecord
```
damyanp, line 4323: Is there a reason this uses a different pattern to the ones above that were verifying that the thing on the left of the `=` matched?

pow2clk: If by the above, you mean the pattern matching in ComputeNodeCompatibility, that one requires repeating some of the metadata references in the replacement text, so they need to be captured. In this case, we're just changing an integer on the right side of the equation. The shader is simpler and this is sufficient to match it.

pow2clk: Actually, that's true of the below, but not so much the above. I suppose it could be simpler there.

pow2clk: Okay, I've reminded myself that in the above case, we have several node entries that have the same metadata. To distinguish and alter them independently, we need to bring in more information including the metadata that identifies the function in question. 

Without this level of specificity, each replacement will alter the shader or launch type data for all matching disassembly. That happens to work, but it's not really what's meant to be tested.

damyanp (approved):

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/node-objects-metadata.hlsl
```diff
@@ -59,11 +59,11 @@ struct RECORD1
...
-// MD: !{void ()* @node_DispatchNodeInputRecord, !"node_DispatchNodeInputRecord", null, null, ![[DispatchNodeInput:[0-9]+]]}
-// MD: ![[DispatchNodeInput]] = !{i32 8, i32 15, i32 13, i32 1, i32 15, !{{[0-9]+}}, i32 16, i32 -1, i32 18, !{{[0-9]+}}, i32 20, ![[EntryInputs:[0-9]+]], i32 4, !{{[0-9]+}}, i32 5, !{{[0-9]+}}}
-// MD: ![[EntryInputs]] = !{![[EntryInputs0:[0-9]+]]}
-// MD: ![[EntryInputs0]] = !{i32 1, i32 97, i32 2, ![[EntryInputs0Record:[0-9]+]]}
-// MD: ![[EntryInputs0Record]] = !{i32 0, i32 16, i32 2, i32 4}
+// MD-DAG: !{void ()* @node_DispatchNodeInputRecord, !"node_DispatchNodeInputRecord", null, null, ![[DispatchNodeInput:[0-9]+]]}
+// MD-DAG: ![[DispatchNodeInput]] = !{i32 8, i32 15, i32 13, i32 1, i32 15, !{{[0-9]+}}, i32 16, i32 -1, i32 18, !{{[0-9]+}}, i32 20, ![[EntryInputs:[0-9]+]], i32 4, !{{[0-9]+}}, i32 5, !{{[0-9]+}}}
+// MD-DAG: ![[EntryInputs]] = !{![[EntryInputs0:[0-9]+]]}
+// MD-DAG: ![[EntryInputs0]] = !{i32 1, i32 97, i32 2, ![[EntryInputs0Record:[0-9]+]]}
+// MD-DAG: ![[EntryInputs0Record]] = !{i32 0, i32 16, i32 2, i32 4}
```
tex3d, line 66: I *really* wish we had a better way to check structured data than what FileCheck gives us.

tools/clang/unittests/HLSL/ValidationTest.cpp
tex3d: Did we need to copy any updates to these changes back into the mesh nodes branch?

tex3d (approved):

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/node-objects-metadata.hlsl
pow2clk: Agreed. I'm assuming I can resolve this conversation.

tools/clang/test/SemaHLSL/hlsl/objects/texture/CalculateLODExtraDiag.hlsl
```diff
@@ -86,7 +86,7 @@ void node02()
 
 // expected-note@+2 {{declared here}}
 [Shader("vertex")]
-float4 vs(float2 a :A) :SV_POSTION {
+float4 vs(float2 a :A) :SV_Position {
```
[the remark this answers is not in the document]
pow2clk: The original misspelled "POSITION". It was an incidental change. I altered the capitalization as I think it makes the error easier to spot.

tools/clang/unittests/HLSL/ValidationTest.cpp
pow2clk: There are no unique updates here. These commits are all copied from those that were part of #6436. Nothing has changed. This is the copy from there to here.

pow2clk (commented):
@tex3d, if you're satisfied with my answers, please resolve the conversations.

PR #6467, opened 2024-03-26
Remove -spirv from new HLSL Sema tests

hekota:
These are semantic verification tests, the `-spir-v` option is not necessary. 
It also breaks our internal test passes for builds that do include SPIR-V backend.


python3kgae (approved):

cassiebeckley (approved):

Keenuts:
Hello!
to understand, `-verify` sets the langopts to both values? or what's happening?
Example: https://github.com/microsoft/DirectXShaderCompiler/pull/6453 checks if we are building SPIR-V to emit sema errors. So seems like the flag it required there. Would that break your tests?

hekota:
I've already replied in #6453. Ideally, sema check should not be dependent on a backend, but it looks like there are exceptions where this is necessary. In that case these tests should be moved into a subfolder that would run only when SPIR-V is enabled.

PR #6468, opened 2024-03-27
Allow disabling payload access qualifiers in 6.8+

### LINKED ISSUE #6462: `-disable-payload-qualifiers` not functioning in SM6.8
**Description**
Defining a ray payload structure without adding Payload Access Qualifiers to each member while compiling `-disable-payload-qualifiers` fails to compile with errors as if the commandline argument is not functioning
Are payload access qualifiers perhaps no longer optional in SM6.8? 

**Steps to Reproduce**
Link to repro: https://godbolt.org/z/9xx78cEYd
Compile arguments: `-T lib_6_8 -disable-payload-qualifiers`

**Actual Behavior**
```
payload type 'PayloadType' requires that all fields carry payload access qualifiers.
```

**Environment**
- DXC version `dxcompiler.dll: 1.8 - 1.8.2403.34 (c9660a8c0); dxil.dll: 1.8(101.8.2403.18)`
- Host Operating System: W11 - 19045.4170

pow2clk:
The original payload access qualifiers change disallowed disabling them on 6.8+ versions. This wasn't an intended feature of 6.8. We'd prefer to maintain the ability to disable these when needed to avoid issues that might crop up.

This change extends respect of the disable-payload-qualifiers flag for all versions. It rewords the warning that is produced when they are encountered when disabled to be more applicable. Tests are also added to verify that the implicit enabling in 6.7+ works in the basic cases and also that disabling in the 6.7+ cases works as well.

Fixes #6462

### DIFF
--- a/tools/clang/test/DXC/disable_paq.hlsl
+++ b/tools/clang/test/DXC/disable_paq.hlsl
@@ -1,0 +1,18 @@
+// RUN: %dxc -T lib_6_7 -disable-payload-qualifiers %s 2>&1 | FileCheck %s
+// RUN: %dxc -T lib_6_8 -disable-payload-qualifiers %s 2>&1 | FileCheck %s
+
+// Ensures that -disable-payload-qualifiers works properly for SM6.6+ versions
+
+
+// CHECK: warning: payload access qualifiers ignored. These are only supported for lib_6_7+ targets and lib_6_6 with with the -enable-payload-qualifiers flag.
+
+// CHECK-NOT: !dx.dxrPayloadAnnotations
+
+struct [raypayload] Payload {
+    int a : read(caller) : write(caller, miss);
+};
+
+[shader("miss")]
+void main( inout Payload payload ) {
+  payload.a = 0.0;
+}
--- a/tools/clang/test/HLSLFileCheck/hlsl/payload_qualifier/access.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/payload_qualifier/access.hlsl
@@ -10,6 +10,18 @@
 // RUN: %dxc -T lib_6_6 main %s -enable-payload-qualifiers -D TEST_NUM=9 | FileCheck -check-prefix=CHK9 -input-file=stderr  %s
 // RUN: %dxc -T lib_6_6 main %s -enable-payload-qualifiers -D TEST_NUM=10 | FileCheck -check-prefix=CHK10 -input-file=stderr  %s
 // RUN: %dxc -T lib_6_6 main %s -enable-payload-qualifiers -D TEST_NUM=11 | FileCheck -check-prefix=CHK11 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=0 | FileCheck -check-prefix=CHK0 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=1 | FileCheck -check-prefix=CHK1 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=2 | FileCheck -check-prefix=CHK2 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=3 | FileCheck -check-prefix=CHK3 -input-file=stderr  %s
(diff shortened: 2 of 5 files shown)


tools/clang/test/HLSLFileCheck/hlsl/payload_qualifier/access.hlsl
```diff
@@ -10,6 +10,18 @@
...
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=2 | FileCheck -check-prefix=CHK2 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=3 | FileCheck -check-prefix=CHK3 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=4 | FileCheck -check-prefix=CHK4 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=5 | FileCheck -check-prefix=CHK5 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=6 | FileCheck -check-prefix=CHK6 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=7 | FileCheck -check-prefix=CHK7 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=8 | FileCheck -check-prefix=CHK8 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=9 | FileCheck -check-prefix=CHK9 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=10 | FileCheck -check-prefix=CHK10 -input-file=stderr  %s
+// RUN: %dxc -T lib_6_7 main %s -D TEST_NUM=11 | FileCheck -check-prefix=CHK11 -input-file=stderr  %s
```
damyanp, line 25: Are there no lib_6_8 tests here because of #6462?

tools/clang/test/DXC/disable_paq.hlsl
```diff
@@ -0,0 +1,18 @@
+// RUN: %dxc -T lib_6_7 -disable-payload-qualifiers %s 2>&1 | FileCheck %s
+// RUN: %dxc -T lib_6_8 -disable-payload-qualifiers %s 2>&1 | FileCheck %s
+
+// Ensures that -disable-payload-qualifiers works properly for SM6.6+ versions
+
+
+// CHECK: warning: payload access qualifiers ignored. These are only supported for lib_6_7+ targets and lib_6_6 with with the -enable-payload-qualifiers flag.
```
python3kgae, line 7: Why we have warning for lib_6_7 and lib_6_8?

damyanp (approved):
LGTM.  If we are skipping the lib_6_8 tests because of a bug then a TODO might be helpful in there, plus a note on the issue to ensure that those tests are added.

coopp (approved):
Looks good to me.

tools/clang/test/HLSLFileCheck/hlsl/payload_qualifier/access.hlsl
pow2clk: No, this fixes #6462. The point where it makes a difference in treatment is 6.7. I could add 6.8 tests to make doubly sure, but I was trying to balance my usual propensity to test everything against the equally common pushback that I was using up test time with minimal benefit.

tools/clang/test/DXC/disable_paq.hlsl
pow2clk: In both of these cases, the payload qualifiers are being explicitly disabled, so we want a warning letting the user know that the annotations they are making are being ignored. The warning message doesn't mention that possibility, but it was already pretty long and I thought disabling them was both uncommon enough to skip mentioning and explicit enough that it wouldn't require mentioning.

pow2clk:
Just to reiterate. targeting 6.8 will work just as well as targeting 6.7. No change was made to implicitly enabling payload access qualifiers through the shader model version. Both scenarios worked fine before this fix and they work fine after. The only reason I added additional testing was because I noted we didn't have any tests that the implicit enabling was working. Since 6.7 is the threshold for that implicit enablement, that's what I targeted. I know we've talked about having broad testing in later shader models that aren't expected to regress anything, but I doubt that's a per-CI change and beyond the scope here.

PR #6471, opened 2024-03-28
[SPIR-V] Fix struct size computation with bitfields

### LINKED ISSUE #5648: [SPIR-V] Wrong value for ArrayStride decoration when it involves bitfields
**Description**

The `ArrayStride` decoration is incorrectly generated if the buffer contains squashed bitfields.

**Steps to Reproduce**

```hlsl
struct PTE {
  uint valid : 1;
  uint dirty : 1;
};

struct Good {
  uint value;
};

RWStructuredBuffer<PTE>  bad;
RWStructuredBuffer<Good> good;

[numthreads(1, 1, 1)]
void main() {
  if (bad[1].dirty == 1) {
    good[1].value = 0;
  }
}
```

`dxc -T cs_6_6 -E main -spirv ~/repro.hlsl`

**Actual Behavior**

The `ArrayStride` decoration for the `bad` `RWStructuredBuffer` is set to `8` instead of `4`.
This leads to wrong indices being computed (at least different indices that what the host computes when using bitfields.

The root cause seem to be that we compute the `ArrayStride` decoration using the AST nodes, and not the lowered representation. The AST nodes has 2 int, marked as bitfields. SPIR-V representation squashes those 2 bitfields into a single int.

The `good` buffer has the correct stride (4).

**Environment**
- DXC version: HEAD (0f282ccabaf21c88e7dbfe0c866422842b639bc3)
- Host Operating System: linux

Keenuts:
The struct size & alignment computations were wrong when
bitfields were in use. This was due to a mismatch between size
computations at the AST level, and at the SPIR-V level.

This commit adds logic to correctly compute bitfields offsets on
an AST struct.

**note: there are 2 commits in this PR, one that moves code around with no change, and one that adds the logic. To help you review.**

Fixes #5648

### DIFF
--- a/tools/clang/lib/SPIRV/AlignmentSizeCalculator.cpp
+++ b/tools/clang/lib/SPIRV/AlignmentSizeCalculator.cpp
@@ -63,6 +63,93 @@ void AlignmentSizeCalculator::alignUsingHLSLRelaxedLayout(
 }
 
 std::pair<uint32_t, uint32_t> AlignmentSizeCalculator::getAlignmentAndSize(
+    QualType type, const RecordType *structType, SpirvLayoutRule rule,
+    llvm::Optional<bool> isRowMajor, uint32_t *stride) const {
+  bool hasBaseStructs = type->getAsCXXRecordDecl() &&
+                        type->getAsCXXRecordDecl()->getNumBases() > 0;
+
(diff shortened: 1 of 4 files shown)


sudonatalie (approved):
LGTM

cassiebeckley (approved):

PR #6477, opened 2024-03-29
[Test] skip test for older validator which check different value.

python3kgae:
Error message was updated in #6196 which cause test fail on older validator.

Since this issue is only caused by error message update, the validation itself is still the same.
The compiler didn't generate anything that behaves badly with older validators.

So just limit these tests to sm6.8+ to make it work on older validator.

### DIFF
--- a/tools/clang/unittests/HLSL/ValidationTest.cpp
+++ b/tools/clang/unittests/HLSL/ValidationTest.cpp
@@ -1036,6 +1036,8 @@ TEST_F(ValidationTest, SigOverlapFail) {
        "signature element"});
 }
 TEST_F(ValidationTest, SimpleHs1Fail) {
+  if (m_ver.SkipDxilVersion(1, 8))
+    return;
   RewriteAssemblyCheckMsg(
       L"..\\DXILValidation\\SimpleHs1.hlsl", "hs_6_0",
       {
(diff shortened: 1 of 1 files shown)


damyanp:
Can you update the description please?  I think there's a bunch of context required to understand why it's ok to just skip this test like this and why we don't need need to test it for older SM's?  We assume that the person reading the commit message it hasn't been following the recent internal chat.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1057,6 +1059,8 @@ TEST_F(ValidationTest, SimpleHs1Fail) {
       });
 }
 TEST_F(ValidationTest, SimpleHs3Fail) {
+  if (m_ver.SkipDxilVersion(1, 8))
+    return;
```
pow2clk, line 1063: SimpleHs1Fail was the test that was altered in #6196. I think this one is fine and that one needs this protection.

python3kgae: :( Fixed.

pow2clk (approved):
Thanks for responding to this!

damyanp (approved):

PR #6482, opened 2024-03-30
Implement mesh node output params

### LINKED ISSUE #6475: Mesh node output parameters
Need to be allowed for this launch mode and no others. SROA components.

Mesh nodes require mesh shader type outputs to pass along the results to the usual mesh shader interface.

Leverage existing code paths in the compiler to allow output indices, vertices, and primitives to be output just as they are in mesh shaders. 

ensure that assigning of these outputs forces the inclusion of `SetMeshOutputCounts()` just as it does in mesh shaders

**Additional context**
subtask of #6404

pow2clk:
To enable reuse of the mesh code generation for output parameters, mesh nodes now share the mesh shader function properties instead of having their own copies. This allows us to rely on the same code to detect errors and pass along parameter information. Since nodes don't store their info in the shader info union, there isn't a conflict.

Similar sharing with mesh signature lowering is possible by calling the mesh signature lowering stages for mesh nodes and lightly altering those functions to accomodate for node shaders.

Some changes were made to validation to allow for the new parameters and DXIL intrinsics in node shaders.

Created a convenience method in function properties to identify mesh nodes as I was tired of typing out the conditionals to identify it.

Adds tests for invalid output parameters, misuse of SetMeshOutputCounts, and a thorough end-to-end node record-to-mesh output test.

Fixes #6475

Add a raft of tests for mesh node out param fails

enable signature lowering more tests

### DIFF
--- a/lib/HLSL/HLSignatureLower.cpp
+++ b/lib/HLSL/HLSignatureLower.cpp
@@ -467,12 +467,13 @@ void HLSignatureLower::CreateDxilSignatures() {
     if (HLModule::IsStreamOutputPtrType(Ty))
       continue;
 
-    // Skip OutIndices and InPayload
+    // Skip OutIndices, InPayload, and NodeIO
     DxilParameterAnnotation &paramAnnotation =
         EntryAnnotation->GetParameterAnnotation(arg.getArgNo());
     hlsl::DxilParamInputQual qual = paramAnnotation.GetParamInputQual();
     if (qual == hlsl::DxilParamInputQual::OutIndices ||
-        qual == hlsl::DxilParamInputQual::InPayload)
+        qual == hlsl::DxilParamInputQual::InPayload ||
+        qual == hlsl::DxilParamInputQual::NodeIO)
       continue;
 
     ProcessArgument(Entry, EntryAnnotation, arg, props, pSM,
@@ -523,7 +524,7 @@ void HLSignatureLower::AllocateDxilInputOutputs() {
         "Failed to allocate all input signature elements in available space.");
   }
 
-  if (props.shaderKind != DXIL::ShaderKind::Amplification) {
+  if (!props.IsAS()) {
     hlsl::PackDxilSignature(EntrySig.OutputSignature, packing);
     if (!EntrySig.OutputSignature.IsFullyAllocated()) {
       llvm_unreachable("Failed to allocate all output signature elements in "
@@ -531,9 +532,7 @@ void HLSignatureLower::AllocateDxilInputOutputs() {
     }
   }
 
-  if (props.shaderKind == DXIL::ShaderKind::Hull ||
-      props.shaderKind == DXIL::ShaderKind::Domain ||
-      props.shaderKind == DXIL::ShaderKind::Mesh) {
+  if (props.IsHS() || props.IsDS() || props.IsMS() || props.IsMeshNode()) {
     hlsl::PackDxilSignature(EntrySig.PatchConstOrPrimSignature, packing);
     if (!EntrySig.PatchConstOrPrimSignature.IsFullyAllocated()) {
       llvm_unreachable("Failed to allocate all patch constant signature "
@@ -1130,7 +1129,7 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
     break;
   case DXIL::SignatureKind::Output:
     opcode =
-        props.IsMS() ? OP::OpCode::StoreVertexOutput : OP::OpCode::StoreOutput;
+       props.IsMS() || props.IsMeshNode() ? OP::OpCode::StoreVertexOutput : OP::OpCode::StoreOutput;
     break;
   case DXIL::SignatureKind::PatchConstOrPrim:
     opcode = OP::OpCode::StorePrimitiveOutput;
@@ -1141,7 +1140,8 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
   bool bInput = SK == DXIL::SignatureKind::Input;
   bool bNeedVertexOrPrimID =
       bInput && (props.IsGS() || props.IsDS() || props.IsHS());
-  bNeedVertexOrPrimID |= !bInput && props.IsMS();
+  bNeedVertexOrPrimID |= !bInput && (props.IsMS() || props.IsMeshNode());
+
 
   Constant *OpArg = hlslOP->GetU32Const((unsigned)opcode);
 
@@ -1155,7 +1155,7 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
 
   Constant *constZero = hlslOP->GetU32Const(0);
 
-  Value *undefVertexIdx = props.IsMS() || !bInput
+  Value *undefVertexIdx = props.IsMS() || props.IsMeshNode() || !bInput
                               ? nullptr
                               : UndefValue::get(Type::getInt32Ty(HLM.GetCtx()));
 
@@ -1258,10 +1258,18 @@ void HLSignatureLower::GenerateDxilComputeAndNodeCommonInputs() {
     DxilParameterAnnotation &paramAnnotation =
         funcAnnotation->GetParameterAnnotation(arg.getArgNo());
 
+    DxilParamInputQual inputQual = paramAnnotation.GetParamInputQual();
+    // Skip mesh node out params
+    if (funcProps.IsMeshNode() &&
+        (inputQual == DxilParamInputQual::OutIndices ||
+         inputQual == DxilParamInputQual::OutVertices ||
+         inputQual == DxilParamInputQual::OutPrimitives))
+      continue;
+
     llvm::StringRef semanticStr = paramAnnotation.GetSemanticString();
 
     if (semanticStr.empty()) {
-      if (funcProps.IsNode() && paramAnnotation.IsParamInputQualNode())
+      if (funcProps.IsNode() && (paramAnnotation.IsParamInputQualNode()))
         continue;
       dxilutil::EmitErrorOnFunction(HLM.GetModule()->getContext(), Entry,
                                     "Semantic must be defined for all "
@@ -1802,6 +1810,13 @@ void HLSignatureLower::Run() {
       GenerateDxilPrimOutputs();
     }
   } else if (props.IsCS() || props.IsNode()) {
+    if (props.IsMeshNode()) {
+      GenerateEmitIndicesOperations();
+      CreateDxilSignatures();
+      AllocateDxilInputOutputs();
+      GenerateDxilOutputs();
+      GenerateDxilPrimOutputs();
+    }
     GenerateDxilComputeAndNodeCommonInputs();
   }
 
--- a/tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
+++ b/tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
@@ -5,15 +5,16 @@
 // Test all valid mesh node input parameters work
 
 // CHECK: define void @node01()
-// CHECK:  %[[tid_x:.+]] = call i32 @dx.op.threadId.i32(i32 93, i32 0)  ; ThreadId(component)
-// CHECK:  %[[tid_y:.+]] = call i32 @dx.op.threadId.i32(i32 93, i32 1)  ; ThreadId(component)
-// CHECK:  %[[tid_z:.+]] = call i32 @dx.op.threadId.i32(i32 93, i32 2)  ; ThreadId(component)
-
-// CHECK:  %[[ftid:.+]] = call i32 @dx.op.flattenedThreadIdInGroup.i32(i32 96)  ; FlattenedThreadIdInGroup()
 
 // CHECK:  %[[tid_group_x:.+]] = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 0)  ; ThreadIdInGroup(component)
 // CHECK:  %[[tid_group_y:.+]] = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 1)  ; ThreadIdInGroup(component)
 // CHECK:  %[[tid_group_z:.+]] = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 2)  ; ThreadIdInGroup(component)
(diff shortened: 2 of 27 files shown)


tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node.hlsl
```diff
@@ -0,0 +1,130 @@
...
+  // CHECK: %[[AHDL:[0-9]+]] = call %dx.types.NodeRecordHandle @dx.op.annotateNodeRecordHandle(i32 251, %dx.types.NodeRecordHandle %[[CHDL]], %dx.types.NodeRecordInfo { i32 97, i32 40 })
+
+  MeshPerVertex ov;
+  ov.position = float4(14.0,15.0,16.0,17.0);
+  ov.color[0] = 14.0;
+  ov.color[1] = 15.0;
+  ov.color[2] = 16.0;
+  ov.color[3] = 17.0;
+
+  if (tig % 3) {
```
tex3d, line 65: Not that we'd ever run this code, but it doesn't make much sense (should be `(!(tig % 3))`), and has data race problems (with `gsMem`).  I see this appears to be based on some other mesh shader test code copied all over the place, so whatever. 🤷

Basically, if we were to implement some useful diagnostics for obvious data races such as this, it would not be happy with this code.

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
tex3d, line 1: I consider the path `objects/NodeObjects` to indicate tests pertaining to node object types (`*NodeInputRecord*` and `*NodeOutput*` objects), not just for everything in node shaders.

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-multi-setoutcounts.hlsl
```diff
@@ -0,0 +1,46 @@
+// RUN: not %dxc -T lib_6_9 %s 2>&1 | FileCheck %s
+
+// REQUIRES: dxil-1-9
+
+// Ensure multiple setMeshOutputCounts produce a validation error
```
tex3d, line 5: I think validation tests should be implemented as IR tests, shouldn't they?  Presumably, we would eventually want Sema or something else to catch some of these errors before validation if possible, which would break any validation test that relies on compiling invalid HLSL.

`CodeGenDXIL` doesn't seem like the right place to put validator tests, and see previous comment about usage of the path `objects/NodeObjects`.

tex3d (commented):
Some general things (see comments for specifics):
- Gate mesh node metadata on `IsMeshNode()` rather than things like validator version for HL IR or the field of a union member (like `ShaderProps.MS.outputTopology`).
- Don't access `ShaderProps.MS` union member without being sure it's valid (`IsMS() || IsMeshNode()`).
- Validation tests should probably be IR based, not HLSL based, since we'd likely want to prevent some of these cases in HLSL before reaching validation at some point.
- Test file paths could be improved.

lib/HLSL/HLSignatureLower.cpp
```diff
@@ -523,17 +524,15 @@ void HLSignatureLower::AllocateDxilInputOutputs() {
         "Failed to allocate all input signature elements in available space.");
   }
 
-  if (props.shaderKind != DXIL::ShaderKind::Amplification) {
+  if (!props.IsAS()) {
     hlsl::PackDxilSignature(EntrySig.OutputSignature, packing);
     if (!EntrySig.OutputSignature.IsFullyAllocated()) {
       llvm_unreachable("Failed to allocate all output signature elements in "
```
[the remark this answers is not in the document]
bob80905: Filed an issue, won't fix here.
https://github.com/microsoft/DirectXShaderCompiler/issues/6485

lib/HLSL/HLSignatureLower.cpp
```diff
@@ -523,17 +524,15 @@ void HLSignatureLower::AllocateDxilInputOutputs() {
...
     }
   }
 
-  if (props.shaderKind == DXIL::ShaderKind::Hull ||
-      props.shaderKind == DXIL::ShaderKind::Domain ||
-      props.shaderKind == DXIL::ShaderKind::Mesh) {
+  if (props.IsHS() || props.IsDS() || props.IsMS() || props.IsMeshNode()) {
     hlsl::PackDxilSignature(EntrySig.PatchConstOrPrimSignature, packing);
     if (!EntrySig.PatchConstOrPrimSignature.IsFullyAllocated()) {
       llvm_unreachable("Failed to allocate all patch constant signature "
```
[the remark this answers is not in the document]
bob80905: Likewise, Filed an issue, won't fix here.
https://github.com/microsoft/DirectXShaderCompiler/issues/6485

lib/HLSL/HLSignatureLower.cpp
```diff
@@ -1141,7 +1140,7 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
   bool bInput = SK == DXIL::SignatureKind::Input;
   bool bNeedVertexOrPrimID =
       bInput && (props.IsGS() || props.IsDS() || props.IsHS());
-  bNeedVertexOrPrimID |= !bInput && props.IsMS();
+  bNeedVertexOrPrimID |= !bInput && (props.IsMS() || props.IsMeshNode());
```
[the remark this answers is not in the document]
bob80905: Removed the extra param from the prototype, and the extraneous args from the calls.

lib/HLSL/HLSignatureLower.cpp
```diff
@@ -1155,7 +1154,7 @@ void HLSignatureLower::GenerateDxilInputsOutputs(DXIL::SignatureKind SK) {
 
   Constant *constZero = hlslOP->GetU32Const(0);
 
-  Value *undefVertexIdx = props.IsMS() || !bInput
+  Value *undefVertexIdx = props.IsMS() || props.IsMeshNode() || !bInput
```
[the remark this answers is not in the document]
bob80905: I'll get back to this.

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
bob80905: Ok, how about I rename the directory to NodeObjectTypes, and make a new directory called NodeShaders in which I will put the new test files?

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node.hlsl
bob80905: Note taken!

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
damyanp: Does this really need to be addressed in this experimental branch?

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-multi-setoutcounts.hlsl
damyanp: Longer-term this may be the case, but is this important to fix for this experimental release?

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
```diff
@@ -5,16 +5,17 @@
...
 // CHECK:  %[[tid_group_x:.+]] = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 0)  ; ThreadIdInGroup(component)
 // CHECK:  %[[tid_group_y:.+]] = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 1)  ; ThreadIdInGroup(component)
 // CHECK:  %[[tid_group_z:.+]] = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 2)  ; ThreadIdInGroup(component)
 
+// CHECK:  %[[ftid:.+]] = call i32 @dx.op.flattenedThreadIdInGroup.i32(i32 96)  ; FlattenedThreadIdInGroup()
+
+// CHECK:  %[[tid_x:.+]] = call i32 @dx.op.threadId.i32(i32 93, i32 0)  ; ThreadId(component)
+// CHECK:  %[[tid_y:.+]] = call i32 @dx.op.threadId.i32(i32 93, i32 1)  ; ThreadId(component)
+// CHECK:  %[[tid_z:.+]] = call i32 @dx.op.threadId.i32(i32 93, i32 2)  ; ThreadId(component)
+
```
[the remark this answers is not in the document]
bob80905: Added checks for groupId, but I am not sure why this reordering happened.

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
damyanp: Could see moving new files to a new directory, but it'll probably make our lives a bit easier in the future if this branch is as close as possible to main.

damyanp (approved):

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
tex3d: No, it does not have to be addressed in this branch now, though it should be easy to move some files to a different path and rename some of them which don't appear to be named properly.  All this could be related to a cleanup issue for later, though it contributes to the test organization chaos in the meantime.

I don't think we want to change the name of the `NodeObjects` folder.  Everything under the `hlsl/objects` path should have to do with testing special HLSL object types.

I would recommend putting this test under a new `hlsl/shaders/node` path.  `hlsl/shaders` can be our new location for shader kind specific entry function tests like these under `CodeGenDXIL`.  The analogue to this under `HLSLFileCheck` would be under `HLSLFileCheck/shader_targets`, though most work graphs content ended up being placed under `HLSLFileCheck/hlsl/workgraph` instead.

DXC tests are disorganized, and getting more so.  We still don't have good guidance written somewhere for where to place various types of tests yet.  There was an organization attempt under the `HLSLFileCheck` folder at some point, but as new features have been added, there seems to be no consistent pattern being applied to where new tests are placed, especially now that there are more roots to start from (like `CodeGenDXIL`).

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-multi-setoutcounts.hlsl
tex3d: It's fine for this branch for now.

lib/HLSL/HLSignatureLower.cpp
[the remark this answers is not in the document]
tex3d: No need to do this now.  Probably should file an issue about it.  I'm just not sure this change actually makes any difference for the mesh node case.

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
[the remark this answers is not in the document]
tex3d: I wish I knew.  I wonder if it's the ordering of this `GenerateDxilComputeAndNodeCommonInputs` call after the other calls for mesh node specifically:
https://github.com/microsoft/DirectXShaderCompiler/pull/6482/files#diff-9406e37b16c71f4146d9c673b5e6dc8408024b2411bdd3212095df9c375d7907R1817

Can we call `GenerateDxilComputeAndNodeCommonInputs` before the `if (props.IsMeshNode()) {` branch?  Does it make a difference to the ordering of the output instructions for some reason?

tools/clang/test/HLSLFileCheck/shader_targets/mesh-node/notArrayOutIndicesNode.hlsl
tex3d, line 1: Small nit: now that this is under `mesh-node` folder, I'm not sure why each file has to have `Node` appended to the name.

tex3d (approved):
I think this is basically fine to merge for now.  Follow-ups can be done later.

tools/clang/test/CodeGenDXIL/hlsl/objects/NodeObjects/mesh-node-input-parameters.hlsl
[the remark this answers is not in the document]
bob80905: Yes, it does result in a significant difference:
```
%1 = load %dx.types.Handle, %dx.types.Handle* @"\01?outbuf@@3V?$RWStructuredBuffer@V?$vector@I$02@@@@A", align 4
  %2 = call i32 @dx.op.threadId.i32(i32 93, i32 0)  ; ThreadId(component)
  %3 = call i32 @dx.op.threadId.i32(i32 93, i32 1)  ; ThreadId(component)
  %4 = call i32 @dx.op.threadId.i32(i32 93, i32 2)  ; ThreadId(component)
  %5 = call i32 @dx.op.flattenedThreadIdInGroup.i32(i32 96)  ; FlattenedThreadIdInGroup()
  %6 = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 0)  ; ThreadIdInGroup(component)
  %7 = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 1)  ; ThreadIdInGroup(component)
  %8 = call i32 @dx.op.threadIdInGroup.i32(i32 95, i32 2)  ; ThreadIdInGroup(component)
```
The ordering is restored, but now theadIdInGroup comes AFTER instead of BEFORE. So now there's another reordering issue!
I'll leave things as they are.

lib/HLSL/HLSignatureLower.cpp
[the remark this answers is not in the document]
bob80905: Issue filed here:
https://github.com/microsoft/DirectXShaderCompiler/issues/6486

PR #6484, opened 2024-04-01
Update version to 1.8.2404

bob80905:
Update version to 1.8.2404 for experimental mesh nodes branch.
The branch will just be used for internal experimentation.


utils/version/latest-release.json
```diff
@@ -2,7 +2,7 @@
     "version" : {
         "major" : "1",
         "minor" : "8",
-        "rev"   : "2403"
+        "rev"   : "2404"
     },
-    "sha" : "809d71c4f33ae1b69c4cb3ac457286ff34008582"
+    "sha" : "9196dfb0fd50aa354b72389281f732ea949f5c73"
```
damyanp, line 7: How is this SHA normally picked?

Wondering why it isn't `72e0fa4ed86ad30eca87053c5993002ba2287437`, which would be the parent of this PR.

damyanp (approved):

utils/version/latest-release.json
bob80905: I ran `git merge-base release-preview-mesh-nodes main` to get the nearest common ancestor of the two branches.

damyanp: Oh, maybe this preview branched from release-1.8.2403, which would give you, I think:

```
> git merge-base origin/release-preview-mesh-nodes origin/release-1.8.2403
11ee8de197007231fe3fde6a78f8a289627a0380
```

hekota: This sha is supposed to be in the `main` branch and it is used by the `gen_version.py` script to generate the rev version on the binaries to be the number of commits from the branch point in `main`.

hekota (approved):

utils/version/latest-release.json
hekota: The goal is to be able to identify the commit when the binary was from built from just the version `1.8.2404.20`. So if the rev number is not calculated from a sha in main it will not work.

Issue #6201, opened 2024-01-24
DXC clang tests fail on Linux Mariner 


hekota:
**Description**
Several tests under OptionsTest and RewriterTest are failing on the Linux Mariner distro.

**Steps to Reproduce**
Repros on internal OB build pipeline. Reach out to @hekota for more info.

**Actual Behavior**
```
[----------] 16 tests from OptionsTest
/__w/1/s/DXC/tools/clang/unittests/HLSL/OptionsTest.cpp:293: Failure
      Expected: true
To be equal to: Unicode::UTF8ToWideString("\xC3\xB1", &wstr)
      Which is: false
/__w/1/s/DXC/tools/clang/unittests/HLSL/OptionsTest.cpp:294: Failure
      Expected: 0
To be equal to: wcscmp(L"\x00F1", wstr.data())
      Which is: 1
[  FAILED  ] OptionsTest.ConvertWhenFailThenThrow (0 ms)
[----------] 16 tests from OptionsTest (2 ms total)

[----------] 37 tests from RewriterTest
/__w/1/s/DXC/tools/clang/unittests/HLSL/RewriterTest.cpp:511: Failure
Value of: SUCCEEDED(pRewriter->RewriteUnchanged(source, 0, 0, &pRewriteResult))
  Actual: false
Expected: true
```

**Environment**
- DXC `main` as of 1/24/2024
- cbl-mariner/build:2.0

hekota:
That is entirely up to you.

damyanp:
Leaving this unfixed means that we ship releases that have known test failures.  We should fix this before the next release.

damyanp:
@Coopp - there's a possibility that there may be a....."wider" task here to work through how wide strings work in linux vs windows here.

coopp:
Capturing some more information on this issue:

Chris did some debugging and noticed that there might be an issue with CPToLocale returning a bad locale value and making incorrect assumptions about the platform.  A starting point would be to see how to make this more robust with a bit of a refactor.  A quick search on github shows others obtain a locale setting by asking for a locale repeatedly until a valid one is returned.  Could be an approach to at least this issue, but not fully settled yet.

igaryhe:
Hi I'm trying to run tests via `make check-all` on GNU Guix, and got similar test failure report:
```
-- Testing: 4271 tests, 16 threads --
Testing: 0 .. 10
[...]

PR #6488, opened 2024-04-02
Update locale setting logic to support UTF-8 only for Apple and Linux distros

coopp:
This commit fixes the setlocale( ) logic to include the UTF-8 supported string value for Mariner distros.  

It also introduces a new RAII class `ScopedLocale' which ensure that the locale setting is set/reset during string conversion operations.

Fixes: #6201

### DIFF
--- a/include/dxc/WinAdapter.h
+++ b/include/dxc/WinAdapter.h
@@ -925,7 +925,6 @@ public:
       setlocale(LC_ALL, m_locale);
     }
   }
-  bool IsSupported() { return (m_locale != nullptr); }
 
 private:
   // Sets a locale for the specified Windows codepage
@@ -934,26 +933,16 @@ private:
       assert(false && "Support for Linux only handles UTF8 code pages");
     }
 #ifdef __APPLE__
-    switch (CodePage) {
-    case CP_UTF8:
-      return setlocale(Category, "en_US.UTF-8");
-    default:
-      return nullptr;
-    }
+    return setlocale(Category, "en_US.UTF-8");
 #else
     const char *utf8LocaleOptions[] = {"en_US.UTF-8", "en_US.utf8"};
 
-    switch (CodePage) {
-    case CP_UTF8: {
-      for (int i = 0; i < _countof(utf8LocaleOptions); ++i) {
-        const char *locale = setlocale(Category, utf8LocaleOptions[i]);
-        if (locale != nullptr)
-          return locale;
-      }
+    for (int i = 0; i < _countof(utf8LocaleOptions); ++i) {
+      const char *locale = setlocale(Category, utf8LocaleOptions[i]);
+      if (locale != nullptr)
+        return locale;
     }
-    default:
-      return nullptr;
-    }
+    return nullptr;
 #endif
   }
 };
@@ -962,14 +951,8 @@ private:
 // used here.
 template <int t_nBufferLength = 128> class CW2AEX {
 public:
-  CW2AEX(LPCWSTR psz, UINT nCodePage = CP_UTF8) {
-    ScopedLocale locale(nCodePage);
-    if (!locale.IsSupported()) {
-      // Current Implementation only supports CP_UTF8, and CP_ACP
-      assert(false && "CW2AEX implementation for Linux only handles "
-                      "UTF8 and ACP code pages");
-      return;
-    }
+  CW2AEX(LPCWSTR psz) {
+    ScopedLocale locale(CP_UTF8);
 
     if (!psz) {
       m_psz = NULL;
@@ -993,14 +976,8 @@ typedef CW2AEX<> CW2A;
 // used here.
 template <int t_nBufferLength = 128> class CA2WEX {
 public:
-  CA2WEX(LPCSTR psz, UINT nCodePage = CP_UTF8) {
-    ScopedLocale locale(nCodePage);
-    if (!locale.IsSupported()) {
-      // Current Implementation only supports CP_UTF8, and CP_ACP
-      assert(false && "CA2WEX implementation for Linux only handles "
-                      "UTF8 and ACP code pages");
-      return;
-    }
+  CA2WEX(LPCSTR psz) {
+    ScopedLocale locale(CP_UTF8);
 
     if (!psz) {
       m_psz = NULL;
--- a/include/dxc/WinAdapter.h
+++ b/include/dxc/WinAdapter.h
@@ -930,7 +930,9 @@ public:
 private:
   // Sets a locale for the specified Windows codepage
   const char *SetLocaleForCodePage(int Category, uint32_t CodePage) {
-    assert(false && "Support for Linux only handles UTF8 code pages");
+    if (CodePage != CP_UTF8) {
+      assert(false && "Support for Linux only handles UTF8 code pages");
+    }
 #ifdef __APPLE__
     switch (CodePage) {
     case CP_UTF8:
--- a/include/dxc/WinAdapter.h
+++ b/include/dxc/WinAdapter.h
@@ -932,18 +932,7 @@ private:
     if (CodePage != CP_UTF8) {
       assert(false && "Support for Linux only handles UTF8 code pages");
     }
-#ifdef __APPLE__
     return setlocale(Category, "en_US.UTF-8");
-#else
-    const char *utf8LocaleOptions[] = {"en_US.UTF-8", "en_US.utf8"};
-
-    for (size_t i = 0; i < _countof(utf8LocaleOptions); ++i) {
-      const char *locale = setlocale(Category, utf8LocaleOptions[i]);
-      if (locale != nullptr)
-        return locale;
-    }
-    return nullptr;
-#endif
   }
 };
 
--- a/include/dxc/WinAdapter.h
+++ b/include/dxc/WinAdapter.h
@@ -126,7 +126,7 @@
 // Used by HRESULT <--> WIN32 error code conversion
 #define SEVERITY_ERROR 1
 #define FACILITY_WIN32 7
-#define HRESULT_CODE(hr) ((hr)&0xFFFF)
+#define HRESULT_CODE(hr) ((hr) & 0xFFFF)
 #define MAKE_HRESULT(severity, facility, code)                                 \
   ((HRESULT)(((unsigned long)(severity) << 31) |                               \
              ((unsigned long)(facility) << 16) | ((unsigned long)(code))))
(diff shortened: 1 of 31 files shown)


coopp:
Pushed an update that ran clang-format over my changes.

include/dxc/WinAdapter.h
```diff
@@ -912,16 +912,31 @@ unsigned int SysStringLen(const BSTR bstrString);
 #define CP_ACP 0
 #define CP_UTF8 65001 // UTF-8 translation.
 
-// Convert Windows codepage value to locale string
-const char *CPToLocale(uint32_t CodePage);
+// Sets a locale for the specified Windows codepage
+const char *SetLocaleForCodePage(int Category, uint32_t CodePage);
```
damyanp, line 916: Maybe SetLocaleForCodePage could be moved inside ScopedLocale, further encapsulating all this stuff.

coopp: I can do that.  I remove all places where callers were manually setting the locale.  They should all be using the ScopedLocale class now.

lib/DxcSupport/WinAdapter.cpp
```diff
@@ -62,21 +62,40 @@ unsigned int SysStringLen(const BSTR bstrString) {
 }
 //===---------------------- Char converstion ------------------------------===//
 
-const char *CPToLocale(uint32_t CodePage) {
+const char *SetLocaleForCodePage(int Category, uint32_t CodePage) {
 #ifdef __APPLE__
```
damyanp, line 66: The stuff inside the `#ifdef __APPLE__` is starting to follow an extremely similar pattern to the others.  Could possibly be unified a little more here.

In fact, it's suspicious to me that the Apple one follows the same convention as Mariner for the UTF-8 locale, but not for the ISO88591-1 locale.  Wondering if that's working on Mariner by luck (eg it is failing but the default happens to work).

damyanp: | Platform | ACP | UTF8 |
|------|---|---|
| Apple | en_US.ISO8859-1 | en_US.UTF-8 |
| Mariner | en_US.iso88591 | en_US.UTF-8 |
| other | en_US.iso88591 | en_US.utf8 |

lib/DxcSupport/WinAdapter.cpp
```diff
@@ -62,21 +62,40 @@ unsigned int SysStringLen(const BSTR bstrString) {
...
+          return setlocale(Category, "en_US.UTF-8");
+      default:
+          return nullptr;
+    }
 #else
-  static const char *utf8 = "en_US.utf8";
-  static const char *iso88591 = "en_US.iso88591";
+    const char* utf8LocaleOptions[] = {
+      "en_US.utf8", // supported on Ubuntu
+      "en_US.UTF-8" // supported on Mariner
```
damyanp, line 78: Is it right to talk about Ubuntu and Mariner here?  There are more distros than that available.

coopp: I accidently resolved this conversation.

coopp: I can remove these comments and just treat this list as possible options that work on different distros.  Nothing too specific.

lib/DxcSupport/WinAdapter.cpp
damyanp: I'm not 100% convinced this is working as expected.  On my Mariner:

```
$ locale -a
C
POSIX
en_US.UTF-8
```

And on Ubuntu:

```
$ locale -a
C
C.UTF-8
POSIX
en_US.utf8
```

damyanp:
```
setlocale(LC_ALL, "en_US.iso88591");
setlocale(LC_ALL, "en_US.ISO8859-1");
```

Both return nullptr for me.

damyanp: For me.  Ubuntu:
```
en_US.UTF-8 ok!
en_US.utf8 ok!
en_US.ISO8859-1 bad.
en_US.iso88591 bad.
```

Mariner:

```
en_US.UTF-8 ok!
en_US.utf8 bad.
en_US.ISO8859-1 bad.
en_US.iso88591 bad.
```

So we could probably use en_US.UTF-8 everywhere.

include/dxc/WinAdapter.h
```diff
@@ -912,16 +912,60 @@ unsigned int SysStringLen(const BSTR bstrString);
...
+  ~ScopedLocale() {
+    if (m_locale != nullptr) {
+      setlocale(LC_ALL, m_locale);
+    }
+  }
+  bool IsSupported() { return (m_locale != nullptr); }
+
+private:
+  // Sets a locale for the specified Windows codepage
+  const char *SetLocaleForCodePage(int Category, uint32_t CodePage) {
```
tex3d, line 932: Is there a reason the implementation isn't located in WinAdapter.cpp?

include/dxc/WinAdapter.h
```diff
@@ -912,16 +912,60 @@ unsigned int SysStringLen(const BSTR bstrString);
...
+#ifdef __APPLE__
+    switch (CodePage) {
+    case CP_ACP:
+      return setlocale(Category, "en_US.ISO8859-1");
```
tex3d, line 936: Existing, but isn't CP_ACP supposed to refer to the system default ANSI code page, rather than this specific code page?  For the compiler, and on Apple/Linux platforms, shouldn't that be considered whatever's already set in locale?

include/dxc/WinAdapter.h
```diff
@@ -912,16 +912,60 @@ unsigned int SysStringLen(const BSTR bstrString);
...
-// Convert Windows codepage value to locale string
-const char *CPToLocale(uint32_t CodePage);
+// RAII style mechanism for setting/unsetting a locale for the specified Windows
+// codepage
+class ScopedLocale {
+  const char *m_locale;
+
+public:
+  explicit ScopedLocale(uint32_t codePage)
+      : m_locale(SetLocaleForCodePage(LC_ALL, codePage)) {}
```
tex3d, line 922: Why do we use `LC_ALL` instead of `LC_CTYPE` to only set the encoding?  Couldn't this potentially fail on a system without `en_US` installed/available?

include/dxc/WinAdapter.h
```diff
@@ -912,16 +912,60 @@ unsigned int SysStringLen(const BSTR bstrString);
...
+    switch (CodePage) {
+    case CP_ACP:
+      return setlocale(Category, "en_US.ISO8859-1");
+    case CP_UTF8:
+      return setlocale(Category, "en_US.UTF-8");
+    default:
+      return nullptr;
+    }
+#else
+    const char *utf8LocaleOptions[] = {"en_US.utf8", "en_US.UTF-8"};
```
tex3d, line 943: `"en_US.UTF-8"` seems like it might be the standard one, so why not put it first?  In fact, do we need any other option?

Looks like @damyanp already suggested that here:
https://github.com/microsoft/DirectXShaderCompiler/pull/6488#discussion_r1548651097

include/dxc/WinAdapter.h
damyanp: It turns out that CP_ACP is _never_ used with this code!

include/dxc/WinAdapter.h
```diff
@@ -912,16 +912,57 @@ unsigned int SysStringLen(const BSTR bstrString);
...
+    if (m_locale != nullptr) {
+      setlocale(LC_ALL, m_locale);
+    }
+  }
+  bool IsSupported() { return (m_locale != nullptr); }
+
+private:
+  // Sets a locale for the specified Windows codepage
+  const char *SetLocaleForCodePage(int Category, uint32_t CodePage) {
+    assert(false && "Support for Linux only handles UTF8 code pages");
```
damyanp, line 933: Doesn't this assert fire every time SetLocaleForCodePage is called?

coopp:
Remove CP_ACP path in all references in the code?  Everywhere?

damyanp:
> Remove CP_ACP path in all references in the code? Everywhere?

Only in the implementation of CA2WEX, this is the path that is not used and doesn't work.

include/dxc/WinAdapter.h
```diff
@@ -934,26 +933,16 @@ class ScopedLocale {
...
 #ifdef __APPLE__
-    switch (CodePage) {
-    case CP_UTF8:
-      return setlocale(Category, "en_US.UTF-8");
-    default:
-      return nullptr;
-    }
+    return setlocale(Category, "en_US.UTF-8");
 #else
     const char *utf8LocaleOptions[] = {"en_US.UTF-8", "en_US.utf8"};
```
damyanp, line 938: I don't think we need `en_US.utf8` anywhere.  `en_US.UTF-8` looks like it is pretty universal, it's just an outlier that Ubuntu also supports the lowercase version.

damyanp (approved):
Looking good. I think that the PR title and description needs updating to more describe what this change does now.

llvm-beanz (approved):

include/dxc/WinAdapter.h
tex3d: How was this resolved?  Is setting all other locale properties to `en_US` really not a problem?  Why is it necessary?

include/dxc/WinAdapter.h
```diff
@@ -912,32 +912,45 @@ unsigned int SysStringLen(const BSTR bstrString);
...
+      setlocale(LC_ALL, m_locale);
+    }
+  }
+
+private:
+  // Sets a locale for the specified Windows codepage
+  const char *SetLocaleForCodePage(int Category, uint32_t CodePage) {
+    if (CodePage != CP_UTF8) {
+      assert(false && "Support for Linux only handles UTF8 code pages");
+    }
```
tex3d, line 934: Why isn't this just:
```suggestion
    assert(CodePage == CP_UTF8 && "Support for Linux only handles UTF8 code pages");
```

include/dxc/WinAdapter.h
coopp: I think the overall assumptions on en-US in the system needs some attention.  I am choosing the path of least resisitance at the moment, but leaving the existing behaviors untouched.  I'll file an Issue for this so we can track as a known thing we need to think about and possibly address.

coopp: Added Issue #6497 to track a full revisit of en-us assumptions here.

include/dxc/WinAdapter.h
coopp: It could be.  Just a style choice I made that the time.  I am going to leave it as-is for now.

bogner:
It would be quite a bit easier to see and understand the functional change here if the introduction of the RAII ScopedLocale class was a separate NFC change.

coopp (commented):
Trying to complete my review.  Comments went to wrong section and were set as pending.

include/dxc/WinAdapter.h
bogner: Please don't - `assert(false)` is an antipattern. It's better to assert the condition directly.

damyanp (approved):

Issue #6327, opened 2024-02-19
[SPIR-V] Fatal error with validation on using bitfield uint64_t


deprilula28:
**Description**

When using a shader with bitfield structs and uint64_t, getting invalid SPIR-V and failing.

**Steps to Reproduce**

For example, when using the following struct:
```hlsl
struct TransferRequest
{
    // This represents a transfer command/request
    uint64_t srcAddr;
    uint64_t dstAddr;
    uint64_t srcIndexAddr; // IOTA default
    uint64_t dstIndexAddr; // IOTA default
    uint64_t elementCount : 35; // allow up to 64GB IGPUBuffers
    uint64_t propertySize : 24; // all the leftover bits (just use bytes now)
    uint64_t fill;
    // 0=uint8, 1=uint16, 2=uint32, 3=uint64
    uint64_t srcIndexSizeLog2 : 2;
    uint64_t dstIndexSizeLog2 : 2;
};
```
And trying to set values to it.

(Error happens on this shader code: https://github.com/Devsh-Graphics-Programming/Nabla/tree/adc4d576d034b3e1fb31356645c0a0915beee817/include/nbl/builtin/hlsl/property_pool)

**Actual Behavior**
Invalid SPIR-V causing fatal error on validation.
```
fatal error: generated SPIR-V is invalid: [VUID-StandaloneSpirv-Base-04781] Expected 32-bit int type for Base operand: BitFieldInsert
%58 = OpBitFieldInsert %ulong %42 %57 %uint_0 %uint_35
```

**Environment**
- DXC version: From commit 7d2f9c74d53dcbb9de4e148b68c99e5986875be6 
- Host Operating System: Windows 11 version 22H2 build 22621.3155

cassiebeckley:
Minimal reproducible example: https://godbolt.org/z/86j5fcKGf

cassiebeckley:
I think this might actually be a spirv-val error. If I'm reading https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpBitFieldInsert correctly `OpBitFieldInsert` is allowed to have a result type of any scalar or vector integer type, and base and insert must have the same type. There's nothing stating that it must have a specific bit width. I'll double check with some of the people who work on spirv-val.

devshgraphicsprogramming:
[...]

PR #6491, opened 2024-04-03
[SPIR-V] Emulate OpBitFieldInsert for type != i32

Keenuts:
SPIR-V supported OpBitFieldInsert on all integer types. But Vulkan requires the operands to be 32bit integers.
This means we need to emulate this instruction for types other then i32.

This PR only adds emulation for OpBitFieldInsert. The next PR will add support for emulating OpBitFieldExtract,

Related to #6327

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -444,8 +444,8 @@ public:
   /// \brief Creates an OpBitFieldInsert SPIR-V instruction for the given
   /// arguments.
   SpirvInstruction *createBitFieldInsert(QualType resultType,
-                                         SpirvInstruction *dst,
-                                         SpirvInstruction *src,
+                                         SpirvInstruction *base,
+                                         SpirvInstruction *insert,
                                          unsigned bitOffset, unsigned bitCount,
                                          SourceLocation, SourceRange);
 
@@ -836,7 +836,7 @@ private:
   /// arguments.
   SpirvInstruction *
   createEmulatedBitFieldInsert(QualType resultType, uint32_t baseTypeBitwidth,
-                               SpirvInstruction *dst, SpirvInstruction *src,
+                               SpirvInstruction *base, SpirvInstruction *insert,
                                unsigned bitOffset, unsigned bitCount,
                                SourceLocation, SourceRange);
 
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -443,10 +443,11 @@ public:
 
   /// \brief Creates an OpBitFieldInsert SPIR-V instruction for the given
   /// arguments.
-  SpirvBitFieldInsert *
-  createBitFieldInsert(QualType resultType, SpirvInstruction *base,
-                       SpirvInstruction *insert, SpirvInstruction *offset,
-                       SpirvInstruction *count, SourceLocation);
+  SpirvInstruction *createBitFieldInsert(QualType ResultType,
+                                         SpirvInstruction *Dst,
+                                         SpirvInstruction *Src,
+                                         unsigned BitOffset, unsigned BitCount,
+                                         SourceLocation, SourceRange);
 
   /// \brief Creates an OpBitFieldUExtract or OpBitFieldSExtract SPIR-V
   /// instruction for the given arguments.
@@ -830,6 +831,14 @@ private:
   SpirvVariable *createCloneVarForFxcCTBuffer(QualType astType,
                                               const SpirvType *spvType,
                                               SpirvInstruction *var);
+
+  /// \brief Emulates OpBitFieldInsert SPIR-V instruction for the given
+  /// arguments.
+  SpirvInstruction *
+  createEmulatedBitFieldInsert(QualType ResultType, uint32_t BaseTypeBitwidth,
+                               SpirvInstruction *Dst, SpirvInstruction *Src,
+                               unsigned BitOffset, unsigned BitCount,
+                               SourceLocation loc, SourceRange range);
 
 private:
   ASTContext &astContext;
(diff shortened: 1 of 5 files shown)


tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -882,15 +875,77 @@ void SpirvBuilder::createBarrier(spv::Scope memoryScope,
...
-    QualType resultType, SpirvInstruction *base, SpirvInstruction *insert,
-    SpirvInstruction *offset, SpirvInstruction *count, SourceLocation loc) {
+SpirvInstruction *SpirvBuilder::createEmulatedBitFieldInsert(
+    QualType ResultType, uint32_t BaseTypeBitwidth, SpirvInstruction *Dst,
+    SpirvInstruction *Src, unsigned BitOffset, unsigned BitCount,
+    SourceLocation loc, SourceRange range) {
+
+  assert(BitCount <= 64 &&
+         "Bitfield insertion emulation can only insert at most 64 bits.");
+  auto *DstTy = dyn_cast<IntegerType>(Dst->getResultType());
```
sudonatalie, line 885: I know it's inconsistent with the LLVM style guide, but to be consistent with the rest of this file and the SPIR-V backend, can you switch to lowerCamelCase variable names?

tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -882,15 +875,77 @@ void SpirvBuilder::createBarrier(spv::Scope memoryScope,
...
+  uint64_t MaskValue = 0;
+  for (unsigned i = 0; i < BitCount; i++)
+    MaskValue = (MaskValue << 1) | 1;
+
+  auto MaskTy =
```
sudonatalie, line 890: Can you add some comments to explain the logic? It's not easy to follow how it works just reading the SpirvBuilder calls.

tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -882,15 +875,77 @@ void SpirvBuilder::createBarrier(spv::Scope memoryScope,
...
+SpirvInstruction *SpirvBuilder::createEmulatedBitFieldInsert(
+    QualType ResultType, uint32_t BaseTypeBitwidth, SpirvInstruction *Dst,
+    SpirvInstruction *Src, unsigned BitOffset, unsigned BitCount,
+    SourceLocation loc, SourceRange range) {
+
+  assert(BitCount <= 64 &&
+         "Bitfield insertion emulation can only insert at most 64 bits.");
+  auto *DstTy = dyn_cast<IntegerType>(Dst->getResultType());
+  uint64_t MaskValue = 0;
+  for (unsigned i = 0; i < BitCount; i++)
```
sudonatalie, line 887: To create a mask of `n` 1s, I think you can simply do `2^n - 1` = `(1 << n) - 1`

tools/clang/test/CodeGenSPIRV/op.struct.access.bitfield.sized.hlsl
```diff
@@ -0,0 +1,148 @@
...
+};
+// CHECK-DAG: %S4 = OpTypeStruct %long
+
+[numthreads(1, 1, 1)]
+void main() {
+  S1 s1;
+  s1.f1 = 3;
+// CHECK:   [[ptr:%[0-9]+]] = OpAccessChain %_ptr_Function_ulong %s1 %int_0
+// CHECK:   [[tmp:%[0-9]+]] = OpLoad %ulong [[ptr]]
+// CHECK:  [[mask:%[0-9]+]] = OpShiftLeftLogical %ulong %ulong_4294967295 %uint_0
```
sudonatalie, line 35: Can you add comments with the binary/hex notation for all the values and masks? Without that it's hard to verify that any of these values are correct.

tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -443,10 +443,11 @@ class SpirvBuilder {
 
   /// \brief Creates an OpBitFieldInsert SPIR-V instruction for the given
   /// arguments.
-  SpirvBitFieldInsert *
-  createBitFieldInsert(QualType resultType, SpirvInstruction *base,
-                       SpirvInstruction *insert, SpirvInstruction *offset,
-                       SpirvInstruction *count, SourceLocation);
+  SpirvInstruction *createBitFieldInsert(QualType resultType,
+                                         SpirvInstruction *dst,
```
sudonatalie, line 447: Can we keep the names as base and insert to match the SPIR-V spec? I also think it's a bit more clear than src and dst, both here and for the emulated version.

tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -876,76 +876,92 @@ void SpirvBuilder::createBarrier(spv::Scope memoryScope,
...
-  Dst->setResultType(DstTy);
-  return Dst;
+  auto maskTy =
+      astContext.getIntTypeForBitwidth(baseTypeBitwidth, /* signed= */ 0);
+  const uint64_t MaskValue = (1ull << bitCount) - 1ull;
+  auto *wordMask =
+      getConstantInt(maskTy, llvm::APInt(baseTypeBitwidth, MaskValue));
+  auto *shiftOffset =
+      getConstantInt(astContext.UnsignedIntTy, llvm::APInt(32, bitOffset));
+  auto *mask = createBinaryOp(spv::Op::OpShiftLeftLogical, maskTy, wordMask,
```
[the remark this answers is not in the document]
Keenuts: I was thinking maybe there would be some differences between the SPIR-V bitwidth, and the compile-time bitwidth, but if the codegen makes sense, we should never end up in a case where this is an issue (we should at most truncate, so should be fine).
Changed the mask to a constant (`shift` + `not` removed)

sudonatalie (approved):
LGTM!

cassiebeckley (approved):

PR #6493, opened 2024-04-03
Set supported validator version on DXC/disable_paq.hlsl

hekota:
This test is failing in the internal validator back-compat suite because the feature is not supported on older validators.

### DIFF
--- a/tools/clang/test/DXC/disable_paq.hlsl
+++ b/tools/clang/test/DXC/disable_paq.hlsl
@@ -1,3 +1,4 @@
+// REQUIRES: dxil-1-8
 // RUN: %dxc -T lib_6_7 -disable-payload-qualifiers %s 2>&1 | FileCheck %s
 // RUN: %dxc -T lib_6_8 -disable-payload-qualifiers %s 2>&1 | FileCheck %s
 


tools/clang/test/DXC/disable_paq.hlsl
```diff
@@ -1,3 +1,4 @@
+// REQUIRES: dxil-1-8
 // RUN: %dxc -T lib_6_7 -disable-payload-qualifiers %s 2>&1 | FileCheck %s
 // RUN: %dxc -T lib_6_8 -disable-payload-qualifiers %s 2>&1 | FileCheck %s
```
tex3d, line 3: I think this change is fine, and this is just a side question based on seeing this test.  So feel free to resolve in order to merge the PR.

This only requires 1.8 due to the run line compiling to lib_6_8.  Is there a reason this target is needed?  As in, are there code changes that make a difference to this path between lib_6_7 and lib_6_8?

tex3d (approved):

tools/clang/test/DXC/disable_paq.hlsl
python3kgae: It wants to test all the SM6.6+ versions.
So both lib_6_7 and lib_6_8 are added here.

python3kgae (approved):

PR #6494, opened 2024-04-03
Gate experimental mesh test on validator version 1.9

bob80905:
The multipleSetMeshOutputCounts.hlsl test file needs to be gated on dxil validator version 1.9

### DIFF
--- a/tools/clang/test/CodeGenHLSL/mesh-val/multipleSetMeshOutputCounts.hlsl
+++ b/tools/clang/test/CodeGenHLSL/mesh-val/multipleSetMeshOutputCounts.hlsl
@@ -1,4 +1,4 @@
-// RUN: %dxc -E main -T ms_6_5 %s | FileCheck %s
+// RUN: not %dxc -E main -T ms_6_5 %s 2>&1 | FileCheck %s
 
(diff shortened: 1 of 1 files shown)


python3kgae (approved):

tex3d (approved):

tools/clang/test/CodeGenHLSL/mesh-val/multipleSetMeshOutputCounts.hlsl
```diff
@@ -1,4 +1,4 @@
-// RUN: %dxc -E main -T ms_6_5 %s | FileCheck %s
+// RUN: %dxilver 1.9 | %dxc -E main -T ms_6_5 %s | FileCheck %s
```
tex3d, line 1: Hang on, is CodeGenHLSL run via lit shell testing?  If so, this needs to be changed to `// REQUIRES: dxil-1-9` instead, right?

bob80905: There is a file in CodeGenHLSL that disables the directory from being discovered by lit, which implies this is not a lit shell testing directory.

PR #6496, opened 2024-04-04
DerivativesTest: Remove invalid thread group size test cases 

hekota:
Invalid thread group size testing is included in the compiler/validator tests (PR #6332, commit 696a13a). It does not belong in the execution tests because is it not verifying the driver execution results. The shaders with invalid thread group sizes fail validation and cannot be executed.

Also fixing compile warning and override defines in the derivatives test HLSL code.

Note that the test is incorrectly using the term `dispatch size` instead of `thread group size`.

Fixes internal bug #49806539. Reported from RITP pass.

### DIFF
--- a/tools/clang/unittests/HLSLExec/ShaderOpArith.xml
+++ b/tools/clang/unittests/HLSLExec/ShaderOpArith.xml
@@ -233,7 +233,7 @@
         void CSMain(uint3 id : SV_GroupThreadID, uint ix : SV_GroupIndex) {
           if (DISPATCHY == 1 && DISPATCHZ == 1) {
             g_bufMain[ix] = DerivTest(ix);
-            g_bufDbg[ix] = uint4(ix, ConvertGroupIdx(id), 0);
+            g_bufDbg[ix] = uint4(ix, ConvertGroupIdx(id.x), 0);
            }
           else {
             g_bufMain[convert2Dto1D(id.x, id.y, DISPATCHX)] = DerivTest(id.xy);
@@ -246,7 +246,7 @@
 #undef DISPATCHY
 #undef DISPATCHZ
 
-#define DISPATCHX 1
+#define DISPATCHX 4
 #define DISPATCHY 1
 #define DISPATCHZ 1
 #endif
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -4021,7 +4021,7 @@ TEST_F(ExecutionTest, DerivativesTest) {
                                           {4, 16, 2}, {4, 16, 2}};
 
   pShaderOp->UseWarpDevice = GetTestParamUseWARP(true);
-  
+
   MappedData data;
 
   for (Dispatch &D : dispatches) {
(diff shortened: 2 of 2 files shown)


tools/clang/unittests/HLSLExec/ShaderOpArith.xml
```diff
@@ -246,7 +246,7 @@
 #undef DISPATCHY
 #undef DISPATCHZ
 
-#define DISPATCHX 1
+#define DISPATCHX 4
```
tex3d, line 249: "dispatch" is the wrong term.  This isn't the dispatch dimensions but the threadgroup dimensions that we're talking about.

tex3d: Presumably, this block of code is just to make the MSMain compile when thread dimensions are too high?
Also, presumably we aren't using ASMain or MSMain under these conditions?

First, I don't know if ASMain has this same constraint that MSMain has.  Though, the spec doesn't appear to mention AS in the restriction described [here](https://microsoft.github.io/DirectX-Specs/d3d/MeshShader.html#numthreads).

Second, can't the test avoid compiling ASMain and/or MSMain when supplying invalid dimensions specified in the first place?

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
```diff
@@ -233,7 +233,7 @@
         void CSMain(uint3 id : SV_GroupThreadID, uint ix : SV_GroupIndex) {
           if (DISPATCHY == 1 && DISPATCHZ == 1) {
             g_bufMain[ix] = DerivTest(ix);
-            g_bufDbg[ix] = uint4(ix, ConvertGroupIdx(id), 0);
+            g_bufDbg[ix] = uint4(ix, ConvertGroupIdx(id.x), 0);
```
tex3d, line 236: Shouldn't this be passing `ix` instead?

tex3d (approved):
I think this change is fine, but this test is weird and confusing.  Feel free to resolve my comment to merge, but the test could still use clarification/improvement in the future.

tex3d:
The one thing that really does bug me though, is the improper use of "dispatch" instead of "thread" or "thread group" when referring to these dimensions.  Can the PR title and description at least be updated to reflect the correct terminology, with a note that the code uses "dispatch dimensions" when referring to thread group dimensions, and should be updated in the future to avoid further confusion from this?  Or you could update the code to change that too, which might not be too much work (all mechanical).

llvm-beanz (approved):

hekota:
> The one thing that really does bug me though, is the improper use of "dispatch" instead of "thread" or "thread group" when referring to these dimensions. Can the PR title and description at least be updated to reflect the correct terminology, with a note that the code uses "dispatch dimensions" when referring to thread group dimensions, and should be updated in the future to avoid further confusion from this? Or you could update the code to change that too, which might not be too much work (all mechanical).

The term 'dispatch' is used in several other execution tests as well, not just this one. Since this PR will need to be cherry-picked to the release branch in order to update the actual HLK test, I would like to limit it to just the functional test change. I will update the PR description and title.

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
hekota: Correct. This define overrides are used when the shader source code is compiled as a compute shader which allows larger thread group sizes than AS or MS. The ASMain and MSMain are not used in this case. This is the tax for having a single HLSL source code for the test. We could separate the code into two sections in the `ShaderOpArith.xml`, one for CS and one for MS & AS, but then some of the shared code would have to be duplicated in both sections.

PR #6499, opened 2024-04-04
[RDAT] Add mesh node signature elements to mesh node case for RDAT

### LINKED ISSUE #6495: [mesh nodes] Add missing RDAT for MS output signatures
Mesh Nodes are missing output signature descriptions in RDAT.  This should have been part of [#6475](https://github.com/microsoft/DirectXShaderCompiler/issues/6475), but was missed.

Changes are required in `DxilContainerAssembler.cpp` to set `Sig*Elements` using `AddSigElements` in the MSInfo used for mesh nodes, like the code does under the `case ShaderKind::Mesh` in `DxilRDATWriter::AddShaderInfo`.  D3DReflect tests need to be added/updated to verify mesh node RDAT contains all the expected details.

bob80905:
Previously, mesh nodes would be missing two details in the RDAT output: SigOutputElements and SigPrimOutputElements.
The signature data was missing from RDAT because lib\DxilContainer\DxilContainerAssembler.cpp:AddShaderNodeInfo only added MeshOutputTopology, MaxOutputVertices, and MaxOutputPrimitives data to the RDAT. 
In this PR, more data gets added to RDAT, including the previously missing signature elements.
A test was added that verifies that the RDAT includes these new signatue element details. 
Fixes #6495


bob80905:
The new test does have some data inside SigOutputElements / SigPrimOutputElements, and ideally the test would just check for that data. But other rdat tests tend to check for lots of surrounding structure. The test is large, since it tests ALL RDAT output, and I would like to remove some check lines, but I am unsure what the general guidelines are for what doesn't need to be checked in RDAT tests.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1657,9 +1657,25 @@ class DxilRDATWriter : public DxilPartWriter {
...
       nAttrib.AttribKind = (uint32_t)RDAT::NodeFuncAttribKind::MeshShaderInfo;
       RDAT::MSInfo info = {};
+      uint32_t shaderFlags = 0;
+      info.SigOutputElements = AddSigElements(sig.OutputSignature, shaderFlags);
+      funcInfo.ShaderFlags |=
+          (uint16_t)(shaderFlags &
+                     (uint16_t)DxilShaderFlags::OutputPositionPresent);
+      info.SigPrimOutputElements =
+          AddSigElements(sig.PatchConstOrPrimSignature, shaderFlags);
+      // TODO: Fill in ViewID related masks
```
damyanp, line 1671: What's the story with this TODO?  Are the ViewID things the thing we decided not to implement for the preview?  If so that could be explained in this comment.

Oh, I see that it looks like this TODO comment is repeated through this function, so this may be pre-existing.  
@llvm-beanz - looks like you added those TODOs when this code was originally added.  Do you know if they're still relevant?

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1657,9 +1657,25 @@ class DxilRDATWriter : public DxilPartWriter {
...
+      info.NumThreads =
+          Builder.InsertArray(&props.numThreads[0], &props.numThreads[0] + 3);
+      info.GroupSharedBytesUsed = tgsmSizeInBytes;
+      info.GroupSharedBytesDependentOnViewID =
+          (uint32_t)0; // TODO: same thing (note: this isn't filled in for PSV!)
```
damyanp, line 1676: "same thing"?  Is this referring to line 1671?  Suggest making the TODO comment self-contained in case some future revision updates one without spotting the other.

lib/DxilContainer/DxilContainerAssembler.cpp
```diff
@@ -1657,9 +1657,25 @@ class DxilRDATWriter : public DxilPartWriter {
     }
 
     if (props.Node.LaunchType == NodeLaunchType::Mesh) {
+      const DxilEntrySignature &sig = entryProps.sig;
       nAttrib = {};
       nAttrib.AttribKind = (uint32_t)RDAT::NodeFuncAttribKind::MeshShaderInfo;
       RDAT::MSInfo info = {};
+      uint32_t shaderFlags = 0;
+      info.SigOutputElements = AddSigElements(sig.OutputSignature, shaderFlags);
```
damyanp, line 1665: An aside: I found this pretty confusing until I realized that "shaderFlags" is a write-only out param that's taken as a non-const reference.  Totally beyond the scope of this PR, just wanted to have my say somewhere!

damyanp (approved):
LGTM - I have some questions around TODO comments.

lib/DxilContainer/DxilContainerAssembler.cpp
bob80905: Yes, I copied the code directly from AddShaderInfo, which had this comment that I'm not sure is still relevant or not.

damyanp:
> ... other rdat tests tend to check for lots of surrounding structure. The test is large, since it tests ALL RDAT output, and I would like to remove some check lines, but I am unsure what the general guidelines are for what doesn't need to be checked in RDAT tests.

Yeah, I was going to comment on that but I remember seeing other tests like this.  It strikes me as being pretty brittle if anything is ever added to the output here, but I don't think you're too far from existing patterns here.

For the preview branch this is probably ok and we don't need to dig too deep into it.

lib/DxilContainer/DxilContainerAssembler.cpp
tex3d: These were TODO for exposing regular graphics entry points in libraries for the state object API (which was part of DSL), instead of requiring them to be compiled to the native shader stage targets instead (like `vs_6_8`).  We decided not to try to expose existing graphics/compute entries in libraries because there were too many incompletes and unknowns there, and we didn't have the time/resources to get right.

lib/DxilContainer/DxilContainerAssembler.cpp
tex3d: The mesh nodes spec identifies these as set to zero for the preview (ViewID not supported in preview).  We could group the ViewID related properties and have a comment that reflects this decision from the spec instead.

tools/clang/test/HLSLFileCheck/d3dreflect/mesh-node-rdat.hlsl
```diff
@@ -0,0 +1,398 @@
...
+// CHECK:                AttribKind: RecordAlignmentInBytes
+// CHECK:                RecordAlignmentInBytes: 4
+// CHECK:              }
+// CHECK:            }
+// CHECK:          }
+// CHECK:        }
+// CHECK:      }
+// CHECK:    }
+// CHECK:  }
+// CHECK:  RecordTable (stride = {{[0-9]+}} bytes) NodeIDTable[1] = {
```
tex3d, line 107: I don't think we need checks for all the remaining RDAT tables, since all the details are expanded in-line the first time they are used in the `FunctionTable`.

tools/clang/test/HLSLFileCheck/hlsl/workgraph/mesh_node_attribs_rdat.hlsl
```diff
@@ -217,6 +201,8 @@ void node02_mesh_maxdispatch(DispatchNodeInputRecord<RECORD> input) {}
 // CHECK:      }
 // CHECK:    }
 // CHECK:  }
+// Lots of intermediate RDAT output that is skipped,
+// we jump to checking for ID3D12Library Reflection here.
```
tex3d, line 205: This is ok, since these table elements are expanded in-line and checked inside the `RuntimeDataFunctionInfo*` record.

bob80905: Understandable, just thought I'd leave that comment there to prevent any confusion in the future.

tex3d (approved):
Looks good!  The one thing to watch out for when initializing a record to zero is if it has any ArrayRef fields, which must be set to `(uint)-1` (`RDAT_NULL_REF`) for any null/empty arrays.  This was the cause of the corrupt values for the signature elements before this change.  I feel like RDAT validation should have caught that, but maybe there's a gap here.  I'll look into that later.

PR #6500, opened 2024-04-05
[SPIR-V] Emulate OpBitFieldExtract for type != i32

### LINKED ISSUE #6327: [SPIR-V] Fatal error with validation on using bitfield uint64_t

Keenuts:
This is a follow up to #6491 which adds emulation for OpBitField*Extract
This is required because Vulkan requires OpBitField*Extract operands to
be 32-bit integers.

Fixes #6327

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvBuilder.cpp
+++ b/tools/clang/lib/SPIRV/SpirvBuilder.cpp
@@ -226,17 +226,11 @@ SpirvInstruction *SpirvBuilder::createLoad(QualType resultType,
   if (!bitfieldInfo.hasValue())
     return instruction;
 
-  auto *offset = getConstantInt(
-      astContext.UnsignedIntTy,
-      llvm::APInt(32, static_cast<uint64_t>(bitfieldInfo->offsetInBits),
-                  /* isSigned= */ false));
-  auto *count = getConstantInt(
-      astContext.UnsignedIntTy,
-      llvm::APInt(32, static_cast<uint64_t>(bitfieldInfo->sizeInBits),
-                  /* isSigned= */ false));
   return createBitFieldExtract(
-      resultType, instruction, offset, count,
-      pointer->getAstResultType()->isSignedIntegerOrEnumerationType(), loc);
+      resultType, instruction, bitfieldInfo->offsetInBits,
+      bitfieldInfo->sizeInBits,
+      pointer->getAstResultType()->isSignedIntegerOrEnumerationType(), loc,
+      range);
 }
 
 SpirvCopyObject *SpirvBuilder::createCopyObject(QualType resultType,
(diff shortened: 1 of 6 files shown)


Keenuts:
Applied the same feedback as the related PR, and also fixed an issue with signed bitfields.
Ready for review!

tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -980,12 +974,70 @@ SpirvBuilder::createBitFieldInsert(QualType resultType, SpirvInstruction *base,
...
+    auto baseType = dyn_cast<IntegerType>(base->getResultType());
+    leftShift->setResultType(baseType);
+    rightShift->setResultType(baseType);
+  }
+
+  return rightShift;
+}
+
+SpirvInstruction *SpirvBuilder::createBitFieldExtract(
+    QualType resultType, SpirvInstruction *Src, unsigned bitOffset,
```
sudonatalie, line 1018: nit: s/Src/base

tools/clang/lib/SPIRV/SpirvBuilder.cpp
```diff
@@ -980,12 +974,70 @@ SpirvBuilder::createBitFieldInsert(QualType resultType, SpirvInstruction *base,
   return inst;
 }
 
-SpirvBitFieldExtract *SpirvBuilder::createBitFieldExtract(
-    QualType resultType, SpirvInstruction *base, SpirvInstruction *offset,
-    SpirvInstruction *count, bool isSigned, SourceLocation loc) {
+SpirvInstruction *SpirvBuilder::createEmulatedBitFieldExtract(
```
sudonatalie, line 977: Do we need to emulate `OpBitFieldUExtract` differently when `isSigned = false`?

sudonatalie (commented):
Can you update the PR description for OpBitFieldExtract?

tools/clang/lib/SPIRV/SpirvBuilder.cpp
Keenuts: ah yes, thanks

tools/clang/lib/SPIRV/SpirvBuilder.cpp
Keenuts: I don't think so:
If the type was unsigned, we wouldn't have to do sign extension, to masking could be enough. But this requires also 2 instructions (shift + masking).
Since using 2 shifts should yield the same result, I think both can use this method.

(Changed the instruction from ShiftRightLogical to ShiftRightArithmetic, as the second does sign-extension when required vs the first)

sudonatalie: Sorry, I wasn't very clear. I meant that the `createBitFieldExtract` function has a distinct `isSigned` parameter, and for non-emulated versions, we determine whether or not to do sign extension based on its value, but for the emulated version, it's now based on the signedness of `base`.

If we know those are equivalent for all the existing usages, we should remove the `isSigned` param because it's now a misleading interface. If we're not certain, then we should pass `isSigned` onto `createEmulatedBitFieldExtract` and use it there too.

Keenuts: Right, got it.
I checked, and it is invalid to have the resultType != baseType on OpBitFieldExtract instructions. This means we can use the AST result type to get the `isSigned`, and there is no risk of mismatch.
Cleanup those.

sudonatalie (approved):
Thanks for the updates, LGTM!

cassiebeckley (approved):

PR #6504, opened 2024-04-05
[ClangCL] fix x86 only warning

python3kgae:
SIZE_T is unsigned long for x86 while size_t is unsigned int for x86. Cast SIZE_T to size_t to avoid integer mismatch warning.

### DIFF
--- a/tools/clang/tools/dxa/dxa.cpp
+++ b/tools/clang/tools/dxa/dxa.cpp
@@ -303,7 +303,7 @@ bool DxaContext::ExtractPart(const char *pName) {
 
   WriteBlobToFile(pContent, StringRefWide(OutputFilename),
                   DXC_CP_UTF8); // TODO: Support DefaultTextCodePage
-  printf("%zu bytes written to %s\n", pContent->GetBufferSize(),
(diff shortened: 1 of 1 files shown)


damyanp (approved):

dmpots:
Why is `SIZE_T` 64-bits for an x86 target? Is that intended or could it be a configuration error?

If I look at some doce
https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types

SIZE_T is defined as ULONG_PTR, which is 

```
#if defined(_WIN64)
 typedef unsigned __int64 ULONG_PTR;
#else
 typedef unsigned long ULONG_PTR;
#endif
```

So it seems like SIZE_T should be 32 bits here.

dmpots:
> The error is complaining unsigned long and unsigned int. (format specifies type 'size_t' (aka 'unsigned int') but the argument has type 'SIZE_T' (aka 'unsigned long') [-Werror,-Wformat]
> )

Ah, ok that makes more sense. It looks like for some reason on windows x86 they use `size_t == unsigned int`, but `SIZE_T == unsigned long`. The storage should be the same (32-bits), but the type is different.

LGTM.

dmpots (approved):

PR #6507, opened 2024-04-05
Allow derivatives on mesh nodes

### LINKED ISSUE #6480: [Mesh Nodes] Ensure derivatives are allowed in mesh nodes
Derivatives are allowed in broadcasting and mesh shaders. They should be allowed in mesh nodes as well at least on platforms that support it. 

At least where derivatives in mesh shaders are allowed, derivatives should be available in mesh nodes. Testing should confirm this and possibly the mesh derivative feature flag should be set accordingly. 

I'm unsure if the mesh derivatives flag should be required here as broadcasting node derivatives are already available wherever work graphs are

bob80905:
Previously, mesh nodes (that is, node shaders with mesh launch types) would emit diagnostics when they contain a derivative operation. However, mesh nodes should allow derivative operations. This PR adjusts this rule for mesh nodes, so that derivative operations are allowed. An accompanying regression test ensures that diagnostics that would previously be emitted are no longer emitted.

Fixes #6480.


python3kgae (approved):

damyanp:
>  mesh nodes should allow derivative operations. 

Citation needed :)

(Just checking that we're sure about this!)

tex3d (commented):
We need test(s) for the DXIL to make sure we don't block derivative intrinsics in validation as well.  We also need test(s) checking RDAT, like an adapted version of `tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm68_node_deriv.hlsl` for SM 6.9 mesh node cases.

bob80905:
@damyanp 
Here's the citation, it's officially in the specs repo now.
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0019-mesh-nodes.md#derivative-operations

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm69_mesh_node_deriv.hlsl
```diff
@@ -0,0 +1,202 @@
...
+
+Texture2D<float4> T2D : register(t0, space0);
+SamplerState Samp : register(s0, space0);
+RWByteAddressBuffer BAB : register(u1, space0);
+
+///////////////////////////////////////////////////////////////////////////////
+// Category: derivatives ddx/ddy/ddx_coarse/ddy_coarse/ddx_fine/ddy_fine
+
+// RDAT-LABEL: UnmangledName: "node_deriv"
+// RDAT: FeatureInfo1: 0
```
tex3d, line 16: This should be reporting the `DerivativesInMeshAndAmpShaders` feature flag requirement.  In order to do that, you need to update `AdjustMinimumShaderModelAndFlags` in `DxilModule.cpp` under the `if (flags.GetUsesDerivatives()) {` block to add a node case, and check the launch mode, then set `flags.SetDerivativesInMeshAndAmpShaders(true);`.

tools/clang/test/HLSLFileCheck/d3dreflect/rdat_mintarget/sm69_mesh_node_deriv.hlsl
```diff
@@ -0,0 +1,202 @@
+// RUN: %dxilver 1.8 | %dxc -T lib_6_9 %s | %D3DReflect %s | %FileCheck %s -check-prefixes=RDAT
```
tex3d, line 1: %dxilver should be 1.9, not 1.8.

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_mesh.ll
```diff
@@ -0,0 +1,307 @@
+; RUN: %dxilver 1.8 | %dxv %s
+; This test's content is from the compilation output of 
+; \tools\clang\test\HLSLFileCheck\d3dreflect\rdat_mintarget\sm69_mesh_node_deriv.hlsl
+; compiling with -T lib_6_9. 
+; The purpose of the test is to detect validation errors when using 
+; derivative operations on node shaders with the mesh launch type.
+; No diagnostics are expected, because derivatives are expected to be allowed in mesh nodes.
+
+; Ensure that categories of deriv ops are allowed for mesh node shaders.
+; Ensure that the OptFeatureInfo_UsesDerivatives flag is set as well.
```
tex3d, line 10: What is this test for exactly?  You can test this from HLSL compiling through to DXIL and check the DXIL for expected properties, right?  It's not a validator test, because there's nothing expected from the validator here.  An end-to-end HLSL to DXIL test will verify that the validator passes as well.  The hlsl file referenced does this already.

It may be fine to keep this, but I'm not sure it's necessary, and it would need re-generating anyway:

We expect the `DerivativesInMeshAndAmpShaders` flag to be set, which should expect to see `"Derivatives in mesh and amplification shaders"` in a comment in the DXIL disassembly after `Note: shader requires additional functionality:`.

You say here "Ensure that the OptFeatureInfo_UsesDerivatives flag is set as well.", but we do not expect that flag to be set in module flags (only in RDAT).  That flag is filtered out of the module flags (it's a flag for functions used in RDAT for linkage validation, but isn't present in the global module flags).  This test doesn't check that, and it isn't present in the module.

The module flags are later in the metadata: `8590458896` or hex: `0x200080010`, which breaks down into the DxilShaderFlags: `ResMayNotAlias` (default when not passing -res-may-alias), `WaveOps` (due to Quad op), and `EnableRawAndStructuredBuffers` (using a raw buffer).

tools/clang/test/SemaHLSL/derivatives-allowed-on-mesh-nodes.hlsl
```diff
@@ -0,0 +1,27 @@
+// RUN: %dxc -Tlib_6_9 %s -verify
+
+// expected-no-diagnostics
+// previously, we expected 2 diagnostics: 
+// {{overload of intrinsic CalculateLevelOfDetail with SamplerComparisonState requires shader model 6.8 or greater}}
+// {{overload of intrinsic CalculateLevelOfDetailUnclamped with SamplerComparisonState requires shader model 6.8 or greater}}
```
tex3d, line 6: I don't see why it's necessary to have these, it's just confusing.  We never actually expected these diagnostics for mesh node shaders did we?

tools/clang/test/SemaHLSL/derivatives-allowed-on-mesh-nodes.hlsl
```diff
@@ -0,0 +1,27 @@
...
+[NodeIsProgramEntry]
+[NodeID("some_call_me_tim", 7)]
+[NodeLocalRootArgumentsTableIndex(13)]
+void node01_mesh_dispatch() {
+  float a = 3.0;
+  (void)(t.CalculateLevelOfDetail(s, a) + 
+    t.CalculateLevelOfDetailUnclamped(s, a));
+}
+
+// Make sure direct call to CalculateLevelOfDetail and CalculateLevelOfDetailUnclamped get error.
```
tex3d, line 27: This comment doesn't seem to apply at all.

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_mesh.ll
```diff
@@ -0,0 +1,307 @@
+; RUN: %dxilver 1.8 | %dxv %s
```
tex3d, line 1: dxilver would need to be 1.9, not 1.8.

tools/clang/test/SemaHLSL/derivatives-allowed-on-mesh-nodes.hlsl
```diff
@@ -0,0 +1,27 @@
+// RUN: %dxc -Tlib_6_9 %s -verify
```
tex3d, line 1: should have a:
// REQUIRES: dxil-1-9

tools/clang/test/SemaHLSL/derivatives-allowed-on-mesh-nodes.hlsl
bob80905: Not technically necessary, I thought it was extra information that could be useful. Those diagnostics I mentioned would be emitted without my logical changes. I can remove them.

tools/clang/test/SemaHLSL/derivatives-allowed-on-mesh-nodes.hlsl
bob80905: Thanks, this saves a lot of time

tools/clang/test/HLSLFileCheck/dxil/intrinsics/launch_types_mesh.ll
bob80905: Yes, the validator does run on sm69_mesh_node_deriv, so I shouldn't need this test at all. I'll get rid of it.

tools/clang/test/SemaHLSL/derivatives-allowed-on-mesh-nodes.hlsl
```diff
@@ -0,0 +1,22 @@
+// REQUIRES: dxil-1-9
```
tex3d, line 1: Did you verify that this actually runs (like `dxil-1-9`) is recognized and works?  You can insert something that will cause an error and run it to make sure it fails if not.

bob80905: Yes, running hcttest file-check on this test produces an error when I introduce a syntax error.

tex3d (approved):
Looks good!

PR #6513, opened 2024-04-09
Add barrier MemoryTypeFlags and SemanticFlags diagnostics

tex3d:
Context-specific diagnostics are warnings with DefaultError, since in theory they could be dead code at this stage.

There were some changes to existing Sema diagnostic hooks along the way.

SemaHLSLDiagnoseTU.cpp - HLSLMethodCallDiagnoseVisitor
- extend to non-method calls and pass CallExpr
- instead of skipping when call reached from multiple entry/export functions, pass locallyVisited flag for skipping of redundant diags not dependent on entry properties.
- use llvm::SmallPtrSet for DiagnosticCalls
- capture launch type for entry and pass it down

Replace DiagnoseHLSLMethodCall hook with CheckHLSLFunctionCall:

Move hook into Sema::CheckFunctionCall in SemaChecking.cpp to be general to either a function call or a method call.
Refactor existing code in DiagnoseHLSLMethodCall into CheckFinishedCrossGroupSharingCall called in the hook for this method.
Add CheckBarrierCall for context-free flag validity check, called by CheckHLSLFunctionCall.

Refactor CalculateLOD checking into more generic parts, dispatched based on IntrinsicOp from top-level DiagnoseReachableHLSLCall function.

Switch to using new note: "entry function defined here", instead of "declared here".

Fix intrinsic checking to use IsBuiltinTable on the intrinsic attribute's group.

**Preparation Changes**
While these could be broken out, the barrier diagnostic changes are dependent on them, and testing for some of these is dependent on the other parts of the barrier diagnostic changes.

Correct errors in tests, check MaxRecords for EmptyNodeOutput in AST
Note that Barrier(..., 3) means SemanticFlags: GroupSync | GroupScope.
These are only valid in a shader with a visible group, and GroupSync on memory with at least group scope, so they have been changed to 0 in specific cases in these tests.

Make DiagnoseSVForLaunchType a static function, add FIXME notes.

Use StringRef for IsBuiltinTable, compare strings instead of pointers.

Fix FileCheckerTest for -verify output.

New ShaderModel::HasVisibleGroup uses ShaderKind and NodeLaunchType to determine whether shader type has a visible group.

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -1853,8 +1853,8 @@ static void InitParamMods(const HLSL_INTRINSIC *pIntrinsic,
   }
 }
 
-static bool IsBuiltinTable(LPCSTR tableName) {
-  return tableName == kBuiltinIntrinsicTableName;
+static bool IsBuiltinTable(StringRef tableName) {
+  return tableName.compare(kBuiltinIntrinsicTableName) == 0;
 }
 
 static bool HasUnsignedOpcode(LPCSTR tableName, IntrinsicOp opcode) {
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -11249,7 +11249,7 @@ void Sema::CheckHLSLFunctionCall(FunctionDecl *FDecl, CallExpr *TheCall,
   switch (opCode) {
   case hlsl::IntrinsicOp::MOP_FinishedCrossGroupSharing:
     CheckFinishedCrossGroupSharingCall(*this, cast<CXXMethodDecl>(FDecl),
-                                              TheCall->getLocStart());
+                                       TheCall->getLocStart());
     break;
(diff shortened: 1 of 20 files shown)


tex3d:
For reviewers: It will likely be helpful to look at the commits in order, since it starts with several isolated changes that are mostly depended on by the last/main change adding all the diagnostics.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11236,26 +11236,28 @@ static bool CheckBarrierCall(Sema &S, FunctionDecl *FD, CallExpr *CE) {
...
+    break;
   case hlsl::IntrinsicOp::IOP_Barrier:
-    return CheckBarrierCall(*this, FDecl, TheCall);
+    CheckBarrierCall(*this, FDecl, TheCall);
+    break;
   default:
     break;
   }
-  return false;
+  return;
```
damyanp, line 11260: nit: redundant empty return

damyanp (approved):

coopp (approved):
Looks good to me.  I reread the warning messages to make sure they matched conditions.  They do :)

PR #6516, opened 2024-04-09
Fix recursive zero replacement bottoming out in scalars.

dneto0:
During memcpy replacement, the scalarrepl-param-hlsl pass will replace certain uses of a zero-initialized global variable.  The pass uses a recursive algorithm to replace uses of that global by a zero value. The current code expects the recursion to bottom out in an aggregate type. But it can bottom out in a scalar integer or scalar float.

This patch fixes the scalar cases.

Lit-based pass tests are included.

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -3694,7 +3694,7 @@ static bool ReplaceUseOfZeroInit(Instruction *I, Value *V, DominatorTree &DT,
         continue;
     } else if (LoadInst *LI = dyn_cast<LoadInst>(UI)) {
       // Replace uses of the load with a constant zero.
-      Constant* replacement = Constant::getNullValue(LI->getType());
+      Constant *replacement = Constant::getNullValue(LI->getType());
       DXASSERT(replacement != nullptr, "missing case when replacing with zero");
       LI->replaceAllUsesWith(replacement);
       LI->eraseFromParent();
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -3695,7 +3695,6 @@ static bool ReplaceUseOfZeroInit(Instruction *I, Value *V, DominatorTree &DT,
     } else if (LoadInst *LI = dyn_cast<LoadInst>(UI)) {
       // Replace uses of the load with a constant zero.
       Constant *replacement = Constant::getNullValue(LI->getType());
-      DXASSERT(replacement != nullptr, "missing case when replacing with zero");
       LI->replaceAllUsesWith(replacement);
       LI->eraseFromParent();
       continue;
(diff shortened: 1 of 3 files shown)


dneto0:
cc: @amaiorano

dneto0:
Let me know if I made the lit-based test correctly.  I think I followed the intended pattern, using the `utils/hct/ExtractIRForPass.py` script.  I had to modify `%opt` to `%dxopt` but otherwise it was straightforward

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3693,7 +3693,18 @@ static bool ReplaceUseOfZeroInit(Instruction *I, Value *V, DominatorTree &DT,
       if (ReplaceUseOfZeroInit(I, UI, DT, Reachable))
         continue;
     } else if (LoadInst *LI = dyn_cast<LoadInst>(UI)) {
-      LI->replaceAllUsesWith(ConstantAggregateZero::get(LI->getType()));
+      // Replace uses of the load with a constant zero.
+      Type *ty = LI->getType();
+      Value *replacement = nullptr;
```
ben-clayton, line 3698: [`llvm::Constant::getNullValue()`](https://llvm.org/doxygen/classllvm_1_1Constant.html#aa6574d526b3e38a28f688a7bb4325c2c) ?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3693,7 +3693,10 @@ static bool ReplaceUseOfZeroInit(Instruction *I, Value *V, DominatorTree &DT,
       if (ReplaceUseOfZeroInit(I, UI, DT, Reachable))
         continue;
     } else if (LoadInst *LI = dyn_cast<LoadInst>(UI)) {
-      LI->replaceAllUsesWith(ConstantAggregateZero::get(LI->getType()));
+      // Replace uses of the load with a constant zero.
+      Constant *replacement = Constant::getNullValue(LI->getType());
+      DXASSERT(replacement != nullptr, "missing case when replacing with zero");
```
ben-clayton, line 3698: nit: [LLVM already asserts for invalid types](https://llvm.org/doxygen/Constants_8cpp_source.html#l00370), so this assert is unnecessary.

ben-clayton: LGTM - thanks

tools/clang/test/DXC/scalarrepl-param-hlsl-replace-zero-recurse-to-float.ll
```diff
@@ -0,0 +1,131 @@
+; RUN: %dxopt %s -hlsl-passes-resume -scalarrepl-param-hlsl -S | FileCheck %s
+
+; The pass replaces a memcpy from a zero-initialized global that does not have an intervening store.
+; When tracing through geps and bitcasts of uses of that global, the algorithm might
+; bottom out at replacing a load of a scalar float.  Verify this works.
+
+; In the following code, %2 should be replace by i32 0.
+;    %2 = load float, float* %arrayidx.i...
+
+; CHECK: store float 0.000000e+00, float* %arrayidx2.i
```
llvm-beanz, line 10: You'll need to use a regular expression for the name of the SSA value, those are not preserved in release builds, so this will fail the PR tests.

Also given how complicated the input here is (and the aforementioned need to fuzzy-match the instruction) we should probably have more than one check line to protect against any incorrect matches.

dneto0: Thanks for the feedback.
It's addressed in the latest commit.

Issue #6479, opened 2024-03-29
Disallow prohibited node elements from mesh nodes


pow2clk:
**Is your feature request related to a problem? Please describe.**
Mesh nodes were based on broadcasting shaders. As such, they initially had all the features that broadcasting shaders did. We need to verify through testing that nothing is allowed where it shouldn't be. 

**Describe the solution you'd like**
Features include: no node outputs, and no node recursion. This may not be a complete list. Testing should verify that they cannot be used in mesh nodes and that helpful errors are produced.

**Alternatives Considered**
I'm far less concerned about mesh-node-specific features leaking into other node types, but if they come up in the process and missing testing is identified, it might be added as part of this.

bob80905:
Closed with #6517

PR #6517, opened 2024-04-10
Add errors for prohibited features in mesh nodes, add validation for allowed features

bob80905:
Tests need to be added to mesh nodes preview to ensure that the features we would expect to disallow for broadcast node shaders are also disallowed for mesh nodes.
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/6479


python3kgae (approved):

tools/clang/test/CodeGenDXIL/hlsl/shaders/node/mesh-node-no-cross-group-sharing.hlsl
```diff
@@ -0,0 +1,23 @@
...
+
+[Shader("node")]
+[OutputTopology("line")]
+// expected-note@+1{{Launch type defined here}}
+[NodeLaunch("mesh")]
+[NumThreads(4,5,6)]
+[NodeDispatchGrid(2,2,2)]
+void myFancyNode(
+    // expected-error@+1{{'RWDispatchNodeInputRecord' may not be used with mesh nodes (only DispatchNodeInputRecord)}}
+    globallycoherent RWDispatchNodeInputRecord<MY_MATERIAL_RECORD> myProgressCounter
```
tex3d, line 20: Where does it say this is disallowed?  It is listed in the table of input records here:
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0019-mesh-nodes.md#input-record

tools/clang/test/CodeGenDXIL/hlsl/shaders/node/mesh-node-no-outputs.hlsl
```diff
@@ -0,0 +1,45 @@
...
+        NodeOutput<MY_RECORD> myFascinatingNode, // expected-error{{parameter myFascinatingNode is a Node Output type, which is not allowed for mesh node shaders.}}
+
+        [NodeID("myNiftyNode",3)] [MaxRecords(4)] NodeOutput<MY_RECORD> myRecords, // expected-error{{parameter myRecords is a Node Output type, which is not allowed for mesh node shaders.}}
+
+        [MaxRecordsSharedWith(myRecords)]
+        [AllowSparseNodes]
+        [NodeArraySize(63)] NodeOutputArray<MY_MATERIAL_RECORD> myMaterials, // expected-error{{parameter myMaterials is a Node Output type, which is not allowed for mesh node shaders.}}
+
+        // an output that has empty record size
+        [MaxRecords(20)] EmptyNodeOutput myProgressCounter // expected-error{{parameter myProgressCounter is a Node Output type, which is not allowed for mesh node shaders.}}
```
tex3d, line 42: No EmptyNodeOutputArray case.

tools/clang/test/CodeGenDXIL/hlsl/shaders/node/mesh-node-writable-input-record.hlsl
```diff
@@ -0,0 +1,23 @@
+// RUN: %dxc -Tlib_6_9 %s -verify 
+
+// Make sure writable input records aren't allowed for mesh node shaders
```
tex3d, line 3: The issue isn't writable input records, but Group and Thread input records meant for coalescing and thread launch only.  Should have an entry for each of those objects used as inputs (`[RW](Group|Thread)NodeInputRecord(s)`), which could all be in the same test.

tools/clang/test/SemaHLSL/recursion-on-mesh-nodes.hlsl
```diff
@@ -0,0 +1,21 @@
...
+[OutputTopology("line")]
+[NumThreads(4,1,1)]
+[NodeDispatchGrid(4,5,6)]
+[NodeIsProgramEntry]
+[NodeID("some_call_me_tim", 7)]
+[NodeLocalRootArgumentsTableIndex(13)]
+// expected-error@+2{{recursive functions are not allowed: function 'node01_mesh_dispatch' calls recursive function 'node01_mesh_dispatch'}}
+// expected-note@+1{{recursive function located here:}}
+void node01_mesh_dispatch() { 
+ node01_mesh_dispatch();
```
tex3d, line 20: This was not what was meant by recursive nodes.  We prevent this pattern from any entry point or export generally already.

What was meant by recursion was a node that outputs to an output that targets this same node, allowing it to feed itself.  Since no node output objects are allowed for mesh nodes, you are already prevented from setting up a recursive graph.

tools/clang/test/CodeGenDXIL/hlsl/shaders/node/mesh-node-no-outputs.hlsl
tex3d: Also, we should be checking that the validator catches incorrect input record type in metadata, as well as any node outputs.

tools/clang/test/CodeGenDXIL/hlsl/shaders/node/mesh-node-no-cross-group-sharing.hlsl
bob80905: According to https://github.com/microsoft/DirectXShaderCompiler/issues/6479, "writable input records" are not allowed. I assumed RWDispatchNodeInputRecord qualified as a writable input record. Are writable input records allowed? If not, what exactly qualifies as a writable input record?

tools/clang/test/SemaHLSL/recursion-on-mesh-nodes.hlsl
bob80905: Ah, then I will delete the recursion test, since no outputs is already tested.

tools/clang/test/CodeGenDXIL/hlsl/shaders/node/mesh-node-no-cross-group-sharing.hlsl
bob80905: Resolving, since cross group sharing has no indication of not being allowed in the mesh nodes spec.

damyanp (approved):

utils/hct/hctdb.py
```diff
@@ -8080,6 +8080,11 @@ def build_valrules(self):
             "Invalid input record type for node launch type",
             "%0 node shader '%1' has incompatible input record type (should be %2)",
         )
+        self.add_valrule_msg(
+            "Decl.MeshNodeOutputRecord",
+            "A mesh shader may not have any output records,",
+            "Mesh node shader '%0' has unexpected output record '%1'",
```
tex3d, line 8086: nit: I think the wording isn't right, though it might be fine for now.  I think the term we would want to use is "node outputs" rather than "output records".
Maybe:
```suggestion
            "Decl.NodeOutputOnMeshLaunch",
            "Node outputs are not allowed on mesh launch node shaders.",
            "node shader '%0' has disallowed node output '%1' for 'mesh' launch",
```

tools/clang/test/CodeGenDXIL/hlsl/shaders/node/mesh-node-no-invalid-input-record.hlsl
```diff
@@ -0,0 +1,65 @@
...
+};
+
+[Shader("node")]
+[OutputTopology("line")]
+// expected-note@+1{{Launch type defined here}}
+[NodeLaunch("mesh")]
+[NumThreads(4,5,6)]
+[NodeDispatchGrid(2,2,2)]
+void myFancyNode1(
+    // expected-error@+1{{'ThreadNodeInputRecord' may not be used with mesh nodes (only DispatchNodeInputRecord)}}
```
tex3d, line 19: Seems the diagnostic needs adjustment, since RWDispatchNodeinputRecord is allowed as well, maybe just change it to `[RW]DispatchNodeinputRecord`?

tex3d: And is there a test case that verifies that the RW case is valid?

tex3d (approved):

docs/DXIL.rst
```diff
@@ -3038,6 +3038,7 @@ DECL.EXTRAARGS                            Extra arguments not allowed for shader
 DECL.FNATTRIBUTE                          Functions should only contain known function attributes
 DECL.FNFLATTENPARAM                       Function parameters must not use struct types
 DECL.FNISCALLED                           Functions can only be used by call instructions
+DECL.MESHNODEOUTPUTRECORD                 A mesh shader may not have any output records,
```
pow2clk, line 3041: I wrote a comment that I didn't like the wording here, but it seems someone beat me to it and this is stale, which is causing the CI build failures.

tools/clang/test/CodeGenDXIL/hlsl/shaders/node/mesh-node-no-invalid-input-record.hlsl
pow2clk: I agree. A test for globallycoherent too since it seems that was allowed as well.

pow2clk:
I'd update the title since it's not just testing, it's adding a new validation error for when outputs are present. This is actually prohibiting one thing and allowing another

PR #6519, opened 2024-04-10
ExtractIRForPassTest: use %dxopt on the RUN line

dneto0:
Also, emit a better message when the desired pass does not exist (or is mis-spelled). Before this change, there would only be a confusing message about 'after' being None.


llvm-beanz (approved):

sudonatalie (approved):

sudonatalie:
Sorry a second review for this got lost in the shuffle, but I'll assume this should be merged now!

PR #6520, opened 2024-04-10
Add CODEOWNERS to main

llvm-beanz:
In release branches we've added enforcement of AskMode where approval is required before merging. This just adds the requisite file to the `main` branch so that it will appear in future release branches. It DOES NOT apply the branch protection policy, which is a setting controlled in GitHub.

There is no policy change to contributions to main with this change.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @microsoft/hlsl-release


damyanp (approved):

tex3d (approved):

coopp (approved):
I saw this type of thing when creating the new repo in ADO for compiling llvm-project.  I think I will keep the codeowner file in there now! 

Change looks good!

Issue #6521, opened 2024-04-11
[Mesh Nodes] Mesh nodes should have the usage masks set in RDAT 


bob80905:
**Is your feature request related to a problem? Please describe.**
The compiler doesn't set the usage mask for mesh node signature elements.
Without this information, in the runtime, SignatureElement_Reader->GetUsageMask() is returning 0 all the time.
This causes the runtime to produce Mesh Node -> Pixel Shader linkage errors because it thinks
the pixel shader is always reading an input field that the Mesh Node never writes.

**Describe the solution you'd like**
The compiler should set usage masks for mesh nodes in library shaders. After setting the mask, the data should appear in RDAT, which should resolve the issue of GetUsageMask() returning 0.

**Describe alternatives you've considered**
There has been discussion to unify the way usage masks are determined, by simply looking at the entire module, and looking for instructions that interact with signatures. By finding all the users of such instructions, we can form an expected usage mask, and assign that to each signature element.

bob80905:
Closed with #6522

PR #6522, opened 2024-04-11
Write usage masks for mesh nodes in libraries into RDAT

bob80905:
Previously, the step that is typically taken to set the usage mask for signature elements was not executed, due to mesh nodes being defined in library shaders. The mask was only written for non-lib shaders. Now that mesh nodes exist in library shaders, we need to allow usage masks to be set for mesh nodes, even though they are in library shaders.

This PR adds an alternative case for mesh nodes. In this case, usage masks are written for all mesh nodes defined in a library shader, by iterating over all functions declared in a library module. An accompanying test was written that shows the RDAT now contains the usage mask, and it is non-null. 

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/6521

### DIFF
--- a/lib/HLSL/DxilPreparePasses.cpp
+++ b/lib/HLSL/DxilPreparePasses.cpp
@@ -848,7 +848,7 @@ public:
         const DxilEntryProps &entryProps = DM.GetDxilEntryProps(F);
         MarkUsedSignatureElements(F, entryProps);
         if (entryProps.props.IsHS() && entryProps.props.ShaderProps.HS.patchConstantFunc)
-          MarkUsedSignatureElements(DM.GetPatchConstantFunction(), DM, IsLib);
+          MarkUsedSignatureElements(props.ShaderProps.HS.patchConstantFunc, entryProps);
       } else {
         for (auto &function : M.getFunctionList()) {
           if (!function.isDeclaration()) {
--- a/lib/HLSL/DxilPreparePasses.cpp
+++ b/lib/HLSL/DxilPreparePasses.cpp
@@ -287,7 +287,7 @@ static bool GetUnsignedVal(Value *V, uint32_t *pValue) {
   return true;
 }
 
-static void MarkUsedSignatureElements(Function *F, const DxilEntryProps &entryProps) {
+static void MarkUsedSignatureElements(Function *F, DxilEntryProps &entryProps) {
   DXASSERT_NOMSG(F != nullptr);
   // For every loadInput/storeOutput, update the corresponding ReadWriteMask.
   // F is a pointer to a Function instance
@@ -309,7 +309,7 @@ static void MarkUsedSignatureElements(Function *F, const DxilEntryProps &entryPr
         continue;
       if (!GetUnsignedVal(LI.get_rowIndex(), &row))
         bDynIdx = true;
-      pSig = entryProps.sig.InputSignature;
+      pSig = &entryProps.sig.InputSignature;
     } else if (SO) {
       if (!GetUnsignedVal(SO.get_outputSigId(), &sigId))
         continue;
@@ -317,7 +317,7 @@ static void MarkUsedSignatureElements(Function *F, const DxilEntryProps &entryPr
         continue;
       if (!GetUnsignedVal(SO.get_rowIndex(), &row))
         bDynIdx = true;
-      pSig = entryProps.sig.OutputSignature;
+      pSig = &entryProps.sig.OutputSignature;
     } else if (SPC) {
       if (!GetUnsignedVal(SPC.get_outputSigID(), &sigId))
         continue;
@@ -325,7 +325,7 @@ static void MarkUsedSignatureElements(Function *F, const DxilEntryProps &entryPr
         continue;
       if (!GetUnsignedVal(SPC.get_row(), &row))
         bDynIdx = true;
-      pSig = entryProps.sig.PatchConstOrPrimSignature;
+      pSig = &entryProps.sig.PatchConstOrPrimSignature;
     } else if (LPC) {
       if (!GetUnsignedVal(LPC.get_inputSigId(), &sigId))
         continue;
@@ -333,7 +333,7 @@ static void MarkUsedSignatureElements(Function *F, const DxilEntryProps &entryPr
         continue;
       if (!GetUnsignedVal(LPC.get_row(), &row))
         bDynIdx = true;
-      pSig = entryProps.sig.PatchConstOrPrimSignature;
+      pSig = &entryProps.sig.PatchConstOrPrimSignature;
     } else if (SVO) {
       if (!GetUnsignedVal(SVO.get_outputSigId(), &sigId))
         continue;
@@ -341,7 +341,7 @@ static void MarkUsedSignatureElements(Function *F, const DxilEntryProps &entryPr
         continue;
       if (!GetUnsignedVal(SVO.get_rowIndex(), &row))
         bDynIdx = true;
-      pSig = entryProps.sig.OutputSignature;
+      pSig = &entryProps.sig.OutputSignature;
     } else if (SPO) {
       if (!GetUnsignedVal(SPO.get_outputSigId(), &sigId))
         continue;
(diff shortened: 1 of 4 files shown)


lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -846,6 +846,17 @@ class DxilFinalizeModule : public ModulePass {
         MarkUsedSignatureElements(DM.GetEntryFunction(), DM);
         if (DM.GetShaderModel()->IsHS())
           MarkUsedSignatureElements(DM.GetPatchConstantFunction(), DM);
+      } else {
+        for (auto &function : M.getFunctionList()) {
+          if (!function.isDeclaration()) {
+            if (DM.HasDxilEntryProps(&function)) {
+              const auto &entryProps = DM.GetDxilEntryProps(&function);
+              if (entryProps.props.IsMeshNode()) {
```
tex3d, line 854: Is there a reason this needs to be limited to mesh node shaders?

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -309,47 +309,47 @@ static void MarkUsedSignatureElements(Function *F, DxilModule &DM) {
         continue;
       if (!GetUnsignedVal(LI.get_rowIndex(), &row))
         bDynIdx = true;
-      pSig = &DM.GetInputSignature();
+      pSig = &DM.GetDxilEntryProps(F).sig.InputSignature;
```
tex3d, line 312: It would be better to either pass the entry props for this entry into this function, or to look it up once outside the loop.

tools/clang/test/HLSLFileCheck/d3dreflect/RDATUsageMaskRepro.hlsl
tex3d, line 1: This test seems to be massive overkill, and it's checking a heck of a lot more than what's necessary to verify that the signatures have valid usage masks.

Can we have a more focused test instead?

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -287,7 +287,7 @@ static bool GetUnsignedVal(Value *V, uint32_t *pValue) {
   return true;
 }
 
-static void MarkUsedSignatureElements(Function *F, DxilModule &DM) {
+static void MarkUsedSignatureElements(Function *F, DxilModule &DM, bool IsLib) {
```
tex3d, line 290: I know what your problem was: it was the patch constant shader case.  That could be solved more elegantly by passing the entry props to this function.  For the hull shader case outside this function, we would simply pass the hull shader props along with the patch constant function.  That would allow you to avoid all the special casing below.

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -843,9 +849,20 @@ class DxilFinalizeModule : public ModulePass {
 
       if (!IsLib) {
         // Set used masks for signature elements
-        MarkUsedSignatureElements(DM.GetEntryFunction(), DM);
+        MarkUsedSignatureElements(DM.GetEntryFunction(), DM, IsLib);
         if (DM.GetShaderModel()->IsHS())
```
tex3d, line 853: Here is how you would just pass the entry props, see next comment for hull shader function.
```suggestion
        Function *F = DM.GetEntryFunction();
        // Assume entry props are available for entry function.
        const DxilEntryProps &entryProps = DM.GetDxilEntryProps(F);
        MarkUsedSignatureElements(F, entryProps);
        if (props.IsHS() && props.ShaderProps.HS.patchConstantFunc)
```
Note: don't need to pass DxilModule anymore.

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -843,9 +849,20 @@ class DxilFinalizeModule : public ModulePass {
...
-          MarkUsedSignatureElements(DM.GetPatchConstantFunction(), DM);
+          MarkUsedSignatureElements(DM.GetPatchConstantFunction(), DM, IsLib);
```
tex3d, line 854:
```suggestion
          MarkUsedSignatureElements(props.ShaderProps.HS.patchConstantFunc, entryProps);
```

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -843,9 +849,20 @@ class DxilFinalizeModule : public ModulePass {
...
+      } else {
+        for (auto &function : M.getFunctionList()) {
+          if (!function.isDeclaration()) {
+            if (DM.HasDxilEntryProps(&function)) {
+              const auto &entryProps = DM.GetDxilEntryProps(&function);
+              if (entryProps.props.IsMeshNode()) {
+                MarkUsedSignatureElements(&function, DM, IsLib);
```
tex3d, line 861:
```suggestion
                MarkUsedSignatureElements(&function, entryProps);
```

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -309,47 +309,53 @@ static void MarkUsedSignatureElements(Function *F, DxilModule &DM) {
...
     } else if (SPC) {
       if (!GetUnsignedVal(SPC.get_outputSigID(), &sigId))
         continue;
       if (!GetUnsignedVal(SPC.get_col(), &col))
         continue;
       if (!GetUnsignedVal(SPC.get_row(), &row))
         bDynIdx = true;
-      pSig = &DM.GetPatchConstOrPrimSignature();
+      pSig = IsLib ? &DM.GetDxilEntryProps(F).sig.PatchConstOrPrimSignature
+                   : &DM.GetPatchConstOrPrimSignature();
```
tex3d, line 331:
```suggestion
      pSig = entryProps.sig.PatchConstOrPrimSignature;
```

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -309,47 +309,53 @@ static void MarkUsedSignatureElements(Function *F, DxilModule &DM) {
...
     } else if (LPC) {
       if (!GetUnsignedVal(LPC.get_inputSigId(), &sigId))
         continue;
       if (!GetUnsignedVal(LPC.get_col(), &col))
         continue;
       if (!GetUnsignedVal(LPC.get_row(), &row))
         bDynIdx = true;
-      pSig = &DM.GetPatchConstOrPrimSignature();
+      pSig = IsLib ? &DM.GetDxilEntryProps(F).sig.PatchConstOrPrimSignature
+                   : &DM.GetPatchConstOrPrimSignature();
```
tex3d, line 340:
```suggestion
      pSig = entryProps.sig.PatchConstOrPrimSignature;
```

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -309,47 +309,53 @@ static void MarkUsedSignatureElements(Function *F, DxilModule &DM) {
...
     } else if (SVO) {
       if (!GetUnsignedVal(SVO.get_outputSigId(), &sigId))
         continue;
       if (!GetUnsignedVal(SVO.get_colIndex(), &col))
         continue;
       if (!GetUnsignedVal(SVO.get_rowIndex(), &row))
         bDynIdx = true;
-      pSig = &DM.GetOutputSignature();
+      pSig = IsLib ? &DM.GetDxilEntryProps(F).sig.OutputSignature
+                   : &DM.GetOutputSignature();
```
tex3d, line 349:
```suggestion
      pSig = entryProps.sig.OutputSignature;
```

lib/HLSL/DxilPreparePasses.cpp
```diff
@@ -309,47 +309,53 @@ static void MarkUsedSignatureElements(Function *F, DxilModule &DM) {
...
     } else if (SPO) {
       if (!GetUnsignedVal(SPO.get_outputSigId(), &sigId))
         continue;
       if (!GetUnsignedVal(SPO.get_colIndex(), &col))
         continue;
       if (!GetUnsignedVal(SPO.get_rowIndex(), &row))
         bDynIdx = true;
-      pSig = &DM.GetPatchConstOrPrimSignature();
+      pSig = IsLib ? &DM.GetDxilEntryProps(F).sig.PatchConstOrPrimSignature
+                   : &DM.GetPatchConstOrPrimSignature();
```
tex3d, line 358:
```suggestion
      pSig = entryProps.sig.PatchConstOrPrimSignature;
```

lib/HLSL/DxilPreparePasses.cpp
tex3d: Note, if you were to open it up to other shader stages, you would want to use the same pattern for hull shader I suggested.  In fact that little bit could be integrated into the beginning of the MarkUsedSignatureElements function.  Something like:
```cpp
// For HS, also traverse patch constant function
if (entryProps.props.IsHS() && entryProps.props.ShaderProps.HS.patchConstantFunc && entryProps.props.ShaderProps.HS.patchConstantFunc != F)
  MarkUsedSignatureElements(entryProps.props.ShaderProps.HS.patchConstantFunc, entryProps);
```

bob80905: Typically, shaders in libs are never linked in a pipeline in the runtime, which is the reasoning for why we didn't set the usage mask for library shaders as far as I know. So, I think it makes sense to limit this to mesh node shaders only.

tex3d (approved):
LGTM.

python3kgae (approved):

PR #6532, opened 2024-04-15
Increase filter time value for failing ftime-trace-granularity test

### LINKED ISSUE #6528: ftime-trace-granularity.hlsl test fails periodically, possibly due to timing condition
**Description**
Test runs for the nightly build are seeing test failures occuring with ftime-trace-granularity.  Outputs are present that shouldn't be which indicates that the filtering mechanism may not be working as expected.

**Steps to Reproduce**
This is difficult to reproduce as it appears to be a timing issue.  The build lab that runs this test seems to be able to hit it consistantly so that may have to be leveraged to verify the fix

**Actual Behavior**
FILECHECK test logic is failing because outputs are present that should not be there.

**Environment**
- DXC version lastest

coopp:
ftime-trace-granularity test was failing in test runs due to the configured time filter value being too small.  The latest test run was recording duration times of 322308 which is larger than the test's specified filter value of 99999 causing timings to not be filtered out.

The test is now updated to have a filter value of 999999, which should properly filter out timings.

Fixes #6528

### DIFF
--- a/tools/clang/test/DXC/ftime-trace-granularity.hlsl
+++ b/tools/clang/test/DXC/ftime-trace-granularity.hlsl
@@ -1,5 +1,5 @@
-// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace -ftime-trace-granularity=99999 | FileCheck %s
-// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace=%t.json -ftime-trace-granularity=99999
+// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace -ftime-trace-granularity=999999 | FileCheck %s
+// RUN: %dxc -E main -T vs_6_0 %s -ftime-trace=%t.json -ftime-trace-granularity=999999
 // RUN: cat %t.json | FileCheck %s
 
 // This test runs both stdout and file output paths.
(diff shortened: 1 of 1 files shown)


python3kgae (approved):
Do we have timeout setting for the pipeline ?

bogner (commented):
This feels like it's just pushing the problem around and won't actually solve the flakiness of the test - it could still in theory run too fast or too slow to hit the right granularity.

We're probably better off using a smaller granularity and just checking that the filtering does "something". That is, we could have a couple of different run lines, one with granularity zero and one with nonzero and check that the size of the output of the filtered one is less than the greater.

The difficulty here is that checking file size in a portable way is a bit awkward. We may be able to use `wc -l`

coopp:
@bogner Wouldn't testing two outputs sizes (filtered against non-filtered) have the same issues here?  The filter value I choose to filter the output would still have to be large enough to cause filtering.  Another idea would be to put in some sort of MAX int value for the granularity value ensuring that a number so large would in fact cause filtering.  Right?

bogner:
The thing that comparing different sizes improves is that we don't need to look for a particular counter being filtered out. The test as is is sensitive to the duration of "Frontend" specifically, so if the frontend time increases for some reason we might need to tweak the value repeatedly. The size idea simplifies things a little because we don't need to hardcode any particular counter, and it's probably fairly safe to assume "some" counters have very short durations (on my machine even `-ftime-trace-granularity=1` manages to filter out around 50 out of 200 counters).

OTOH if it's simpler to just crank the granularity, then let's just use something like int32 max (ie, 2147483647) just to be sure.

bogner (approved):

PR #6533, opened 2024-04-15
[SPIR-V] Remove bitcast from float->int conversion

### LINKED ISSUE #6501: [SPIR-V] Casting float to int16_t looses precision in unexpected way
**Description**

When casting a `float` value to `int16_t`, generated SPIR-V code is equivalent to first casting to `half`, and then casting to `int16_t`. This breaks the assumption that casting to an integral type always floors the float value, because the mantissa of the `half` type has less precision than `int16_t`. For instance, `int16_t(0.999855876f)` results in `1`, whereas `int16_t(int(0.999855876f))` maps to `0`.

**Steps to Reproduce**

See https://godbolt.org/z/7YWfWeGM3

Compile the following hlsl shader with `-T lib_6_4 -spirv -fspv-target-env=vulkan1.2 -enable-16bit-types -HV 2021`

```
RWStructuredBuffer<int> a;

[shader("raygeneration")]
void main() {
    float fvalue = 0.99985586;
    int16_t ivalue1 = int16_t(fvalue);
    int16_t ivalue2 = int16_t(int(fvalue));
    int16_t ivalue3 = int16_t(half(fvalue));
    a[0] = ivalue1;
    a[1] = ivalue2;
    a[2] = ivalue3;
}
```

**Actual Behavior**

Generated code is

```
%21 = OpFConvert %half %float_0_999855876
%22 = OpConvertFToS %short %21
%23 = OpSConvert %int %22
%24 = OpAccessChain %_ptr_StorageBuffer_int %a %int_0 %uint_0
           OpStore %24 %23
%25 = OpAccessChain %_ptr_StorageBuffer_int %a %int_0 %uint_1
           OpStore %25 %int_0
%26 = OpAccessChain %_ptr_StorageBuffer_int %a %int_0 %uint_2
           OpStore %26 %23
```

Note how in %22 and %23 the float value first is converted to `half` before doing the `int16_t` conversion. Note how `int16_t(int(fvalue));` is correctly propagated to a `0`.

I think the issue is lurking at [SpirvEmitter.cpp:8222](https://github.com/microsoft/DirectXShaderCompiler/blob/14ec4b49d419195b787d41464d4a4489bba87bb2/tools/clang/lib/SPIRV/SpirvEmitter.cpp#L8222). The `convertBitwidth` seems to first convert the source type to the bitwidth of the destination type, and only after `OpConvertFToS` is generated.

**Environment**

- DXC version: libdxcompiler.so: 1.8(dev;1-aa4d3746)
- Host Operating System: Compiler Explorer

sudonatalie:
OpConvertFToU and FToS convert numerically from floating point to integer, with rounding toward 0.0, matching the defined behavior of HLSL without any need for an initial bitwidth conversion from floating point to floating point, which can result in incorrect rounding behavior (see #6501).

Note that behavior is undefined if the target type is not wide enough to hold the converted value, however this was also true with an initial bitcast (because an N-bit FP can hold values outside the range of an N-bit int), and I can't come up with any case where an intial truncation would result in correct defined behavior where this straight conversion would be undefined since the precision will inevitably be lost anyways.

Fixes #6501

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -8229,17 +8229,14 @@ SpirvInstruction *SpirvEmitter::castToInt(SpirvInstruction *fromVal,
   }
 
   if (isFloatOrVecOfFloatType(fromType)) {
-    // First convert the source to the bitwidth of the destination if necessary.
-    fromVal = convertBitwidth(fromVal, srcLoc, fromType, toIntType, nullptr,
-                              srcRange);
     if (isSintOrVecOfSintType(toIntType)) {
       return spvBuilder.createUnaryOp(spv::Op::OpConvertFToS, toIntType,
                                       fromVal, srcLoc, srcRange);
-    } else if (isUintOrVecOfUintType(toIntType)) {
+    }
+
+    if (isUintOrVecOfUintType(toIntType)) {
       return spvBuilder.createUnaryOp(spv::Op::OpConvertFToU, toIntType,
                                       fromVal, srcLoc, srcRange);
-    } else {
-      emitError("casting from floating point to integer unimplemented", srcLoc);
     }
   }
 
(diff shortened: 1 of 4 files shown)


Keenuts (approved):
I think this is OK. What would be ideal is an `OpSatConvertFToS`, but seems like this doesn't exist.

cassiebeckley (approved):

PR #6534, opened 2024-04-16
Add missing REQUIRES: dxil-1-9 to tests

tex3d:
These mesh nodes tests were missing REQUIRES: dxil-1-9 to prevent them from running on down-level validator test pass.  This change fixes that and should unblock the internal build.


hekota (approved):
LGTM!

jenatali (approved):

PR #6542, opened 2024-04-17
[SPIR-V] Warn when wave size is used

sudonatalie:
Specifying a wave size is not supported in Vulkan SPIR-V, so warn when it is used in a shader.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -12577,6 +12577,19 @@ void SpirvEmitter::processComputeShaderAttributes(const FunctionDecl *decl) {
 
   spvBuilder.addExecutionMode(entryFunction, spv::ExecutionMode::LocalSize,
                               {x, y, z}, decl->getLocation());
+
+  auto *waveSizeAttr = decl->getAttr<HLSLWaveSizeAttr>();
+  if (waveSizeAttr) {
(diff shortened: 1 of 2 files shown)


cassiebeckley (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -12577,6 +12577,19 @@ void SpirvEmitter::processComputeShaderAttributes(const FunctionDecl *decl) {
...
+  auto *waveSizeAttr = decl->getAttr<HLSLWaveSizeAttr>();
+  if (waveSizeAttr) {
+    // Not supported in Vulkan SPIR-V, warn and ignore.
+
+    // SPIR-V SubgroupSize execution mode would work but it is Kernel only
+    // (requires the SubgroupDispatch capability, which implies the
+    // DeviceEnqueue capability, which is Kernel only). Subgroup sizes can be
+    // specified in Vulkan on the application side via
+    // VK_EXT_subgroup_size_control.
+    emitWarning("Wave size is not supported by SPIR-V",
```
Keenuts, line 12590: Shall the warning hint toward `VK_EXT_subgroup_size_control`? This comment is great, but the user would never see it.

sudonatalie: Good point, done

Keenuts (approved):

PR #6543, opened 2024-04-17
[DxilOp] Allow generation of illegal DXIL operations.

### LINKED ISSUE #6410: [DXIL] Null pointer exception due to `double` floating-point literals in `fmod` intrinsic
**Description**
There is a null pointer exception during the translation of FMod (`TranslateFMod`) when the arguments are floating-point literals, because the latest DXC version does not reduce them to `float` during compilation. The null pointer occurs inside `TrivialDxilUnaryOperation(OP::OpCode::Frc, absDiv, hlslOP, Builder)` because `Frc` won't find the respective function for double-precision floats. Previous versions had this assertion in *OP::GetOpFunc*:
```cpp
DXASSERT(IsOverloadLegal(opCode, pOverloadType), "otherwise the caller requested illegal operation overload (eg HLSL function with unsupported types for mapped intrinsic function)");
```
But the latest version only returns a null pointer that is not handled correctly:
```cpp
if (!IsOverloadLegal(opCode, pOverloadType))
  return nullptr;
```

**Steps to Reproduce**
This issue can be reproduced with the latest version of the main branch and compiling with `dxc -E PSMain -T ps_6_0` (or by following https://godbolt.org/z/beK7oPsx6):
```hlsl
float4 PSMain() : SV_Target {
    return (float4)fmod(1.0, 2.0);
}
```
Making the floating-point literals explicitly single-precision with the `f` suffix compiles fine:
```hlsl
float4 PSMain() : SV_Target {
    return (float4)fmod(1.0f, 2.0f);
}
```
This does not reproduce with other intrinsics such as `sin`.

python3kgae:
This change removes the IsOverloadLegal check in OP::GetOpFunc.

It will permit the generation of illegal DXIL operations. Subsequently, the validation should catch these illegal DXIL operations if they are not optimized later in SimplifyDxilCall.

Fixes #6410


tools/clang/test/lit.cfg
```diff
@@ -503,6 +503,11 @@ if config.enable_backtrace == "1":
 if config.spirv:
     config.available_features.add("spirv")
 
+build_mode = lit_config.params.get('build_mode', None)
+
+if build_mode == "Release" or build_mode == "RelWithDebInfo":
+    config.available_features.add("release_build")
```
tex3d, line 509: This change isn't needed if using `UNSUPPORTED: asserts` instead, which I think is more correct.

lib/DXIL/DxilOperations.cpp
```diff
@@ -3564,8 +3564,9 @@ void OP::UpdateCache(OpCodeClass opClass, Type *Ty, llvm::Function *F) {
 Function *OP::GetOpFunc(OpCode opCode, Type *pOverloadType) {
   if (opCode == OpCode::NumOpCodes)
     return nullptr;
-  if (!IsOverloadLegal(opCode, pOverloadType))
-    return nullptr;
+  DXASSERT(IsOverloadLegal(opCode, pOverloadType),
+           "otherwise the caller requested illegal operation overload (eg HLSL "
+           "function with unsupported types for mapped intrinsic function)");
```
tex3d, line 3569: There is the question of whether to keep the assert here.

I think if we do not have a more general solution planned for the near future that doesn't rely on this code path, we should not have the assert, since we know we rely on violating this assert for perfectly legal code.

In my original change to add tests in #6437, I was going to remove the assert and add a comment to explain why:
https://github.com/microsoft/DirectXShaderCompiler/pull/6437/commits/f72d2e2702c7f058e8b755cddd04def2332edbc7#diff-e85a54111eb6509ee6bf740224386600db9c32fa2fc38d9b95cbb202f0882800L3570

I think this change makes sense for main now (removing the assert and adding the comment).

tex3d: And if we do remove the assert, we won't need the `UNSUPPORTED: asserts` in the tests either.

python3kgae: Updated.

tools/clang/test/lit.cfg
python3kgae: Updated.

tools/clang/test/CodeGenDXIL/hlsl/types/intrinsic_call_with_all_literal_args.hlsl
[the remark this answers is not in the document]
python3kgae: Updated.

tex3d (approved):

pow2clk:
The description is no longer accurate as we're not moving it into an active DXASSERT

tools/clang/test/CodeGenDXIL/literal/fmod_const_eval.hlsl
```diff
@@ -0,0 +1,43 @@
+// RUN: %dxc -T vs_6_0 %s -E main | FileCheck %s
+// RUN: not %dxc -T vs_6_0 %s -E main -DNO_FOLD 2>&1 | FileCheck %s --check-prefixes=NO_FOLD
+
+// The code path for this case asserts, but the assert does not indicate a
```
pow2clk, line 4: I don't think this is accurate anymore either. You might just say that this code path produces an invalid overload

python3kgae:
> The description is no longer accurate as we're not moving it into an active DXASSERT

Fixed.

tools/clang/test/CodeGenDXIL/literal/fmod_const_eval.hlsl
```diff
@@ -0,0 +1,43 @@
...
+// serious problem with internal state.  The invalid overload will either be
+// constant folded away, or caught by the validator.
+
+// Ensure fmod is constant evaluated during codegen, or dxil const eval
+// TODO: handle fp specials properly!
```
pow2clk, line 9: The fact that we produce an invalid overload with user-writable code is troubling.  Is this TODO about fixing that? The comments in #6410 suggested that we would keep it open to fix the root cause even after this partial revert. If that's the case, it should be cited here and the "fixes" line should be removed from the description. If not, we need another bug to cite here.

tools/clang/test/CodeGenDXIL/literal/length_const_eval.hlsl
```diff
@@ -0,0 +1,41 @@
+// RUN: %dxc -T vs_6_0 %s -E main | FileCheck %s
+// RUN: not %dxc -T vs_6_0 %s -E main -DNO_FOLD 2>&1 | FileCheck %s --check-prefixes=NO_FOLD
+
+// The code path for this case asserts, but the assert does not indicate a
+// serious problem with internal state.  The invalid overload will either be
+// constant folded away, or caught by the validator.
+
+// Ensure length is constant evaluated during codegen, or dxil const eval
+// TODO: handle fp specials properly!
```
pow2clk, line 9: Updates to the comments in the previous test should be propagated here too.

tools/clang/test/CodeGenDXIL/literal/normalize_const_eval.hlsl
```diff
@@ -0,0 +1,40 @@
+// RUN: %dxc -T vs_6_0 %s -E main | FileCheck %s
+// RUN: not %dxc -T vs_6_0 %s -E main -DNO_FOLD 2>&1 | FileCheck %s --check-prefixes=NO_FOLD
+
+// The code path for this case asserts, but the assert does not indicate a
+// serious problem with internal state.  The invalid overload will either be
+// constant folded away, or caught by the validator.
+
+// Ensure normalize is constant evaluated during codegen, or dxil const eval
+// TODO: handle fp specials properly!
```
pow2clk, line 9: Here as well. Good tests though!

pow2clk (commented):
Per this comment: https://github.com/microsoft/DirectXShaderCompiler/issues/6410#issuecomment-1998254537 This revert should not "fix" #6410. I think the description and comments should be updated to reflect the removal of the assert as well.

python3kgae:
Created https://github.com/microsoft/DirectXShaderCompiler/issues/6566 and https://github.com/microsoft/DirectXShaderCompiler/issues/6565 to track non-literal crashes.

tex3d:
> Created #6566 and #6565 to track non-literal crashes.

Looks good, though I think they shouldn't be considered crashes, since that's a temporary state of affairs that will change once this PR merges.

Also these issues don't cover the TODO comments in the test (`// TODO: handle fp specials properly!`).  These are about the fact that we won't constant-fold generated literal DXIL ops if they produce FP exceptions (or FP specials), and we need to test specific expected results for operations involving FP specials.  While issue #6557 is likely caused by this problem, it's about more than just this issue.  I think we need another issue to track this problem.  I had added a section for these to the fmod test when #6437 was initially created against the release branch.  But I removed that section because it was disabled/unused and I wasn't sure if I got the expected results correct.  This code could be used to craft a repro for an issue covering this TODO.

Here was the code:
https://github.com/microsoft/DirectXShaderCompiler/pull/6437/commits/d93ffade25e7339c11e9595a628510d6455f988d#diff-557ef4666494c6d42b516718ea7ed954e7e7153efc844f9dfd0678b9b53edb39R29-R42

llvm-beanz:
Are #6566 & #6565 fixed with this change? From the surface those look like duplicates of #6410, but for different intrinsics, so I'm unsure why we have extra issues filed.

python3kgae:
> Are #6566 & #6565 fixed with this change? From the surface those look like duplicates of #6410, but for different intrinsics, so I'm unsure why we have extra issues filed.

No. current PR will not fix #6566 and #6565.
They fail for different reason which looks similar.

llvm-beanz:
To clarify as we discussed in triage this morning. Once this change goes in the issues described originally as crashes in #6566 & #6565, revert to the pre-existing behavior of producing invalid DXIL that fails validation with a validator error.

Both issues are closed as per @damyanp's comment:
> We have no plans to make changes in this area in DXC without considering a design that takes into account what the design is in Clang. For this reason I'm closing this issue as not planned.

python3kgae:
> > Created #6566 and #6565 to track non-literal crashes.
> 
> Looks good, though I think they shouldn't be considered crashes, since that's a temporary state of affairs that will change once this PR merges.
> 
> Also these issues don't cover the TODO comments in the test (`// TODO: handle fp specials properly!`). These are about the fact that we won't constant-fold generated literal DXIL ops if they produce FP exceptions (or FP specials), and we need to test specific expected results for operations involving FP specials. While issue #6557 is likely caused by this problem, it's about more than just this issue. I think we need another issue to track this problem. I had added a section for these to the fmod test when #6437 was initially created against the release branch. But I removed that section because it was disabled/unused and I wasn't sure if I got the expected results correct. This code could be used to craft a repro for an issue covering this TODO.
> 
> Here was the code: [d93ffad#diff-557ef4666494c6d42b516718ea7ed954e7e7153efc844f9dfd0678b9b53edb39R29-R42](https://github.com/microsoft/DirectXShaderCompiler/commit/d93ffade25e7339c11e9595a628510d6455f988d#diff-557ef4666494c6d42b516718ea7ed954e7e7153efc844f9dfd0678b9b53edb39R29-R42)

Created https://github.com/microsoft/DirectXShaderCompiler/issues/6567 to track the fp special value issue.

tools/clang/test/CodeGenDXIL/literal/fmod_const_eval.hlsl
python3kgae: The fp specials issue is tracked with #6567.

pow2clk: I think we decided to punt on fixes, we need to reference the bugs in the code. I'd like to see it mentioend in the comments.

lib/DXIL/DxilOperations.cpp
```diff
@@ -3564,8 +3564,18 @@ void OP::UpdateCache(OpCodeClass opClass, Type *Ty, llvm::Function *F) {
...
+  // Remove this assert on illegal overload for now.
+  // Illegal overloads are generated and eliminated by DXIL op constant
+  // evaluation for a number of cases where a double overload of an HL intrinsic
+  // that otherwise does not support double is used for literal values, when
+  // there is no constant evaluation for the intrinsic in CodeGen.
+  // Illegal overloads of DXIL intrinsics may survive through to final DXIL,
+  // but these will be caught by the validator, and this is not a regression.
+  // DXASSERT(IsOverloadLegal(opCode, pOverloadType),
+  //      "otherwise the caller requested illegal operation overload (eg HLSL "
+  //      "function with unsupported types for mapped intrinsic function)");
```
pow2clk, line 3578: Since we've closed the issues that would have uncommented this, I think we just need to remove the commented code entirely since we're not fixing it.

pow2clk (commented):
Just some changes to commments really.

pow2clk (approved):
Thanks!

PR #6548, opened 2024-04-19
Fixed crash in loop unroll caused by bug in structurize loop exits

adam-yang:
Fixed a bug in `hlsl::RemoveUnstructuredLoopExits` where when a new exiting block is created from splitting, it was added to the current loop being processed, when it could also part of an inner loop. Not adding the new block to inner loops that it's part of makes the inner loops malformed, and causes crash.

This fix adds the new block to the inner most loop that it should be part of. Also adds the `StructurizeLoopExits` option to `loop-unroll` pass for dxopt, which was missing before.

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/control_flow/loops/struct_exit_inner_latch_regression.ll
+++ b/tools/clang/test/HLSLFileCheck/hlsl/control_flow/loops/struct_exit_inner_latch_regression.ll
@@ -5,18 +5,17 @@
 ; CHECK: mul nsw i32
 ; CHECK-NOT: mul nsw i32
 
-; This is a regression test for a crash in loop unroll.
-; When there are multiple exits, the compiler will run
-; hlsl::RemoveUnstructuredLoopExits to try to avoid
-; unstructured code.
+; This is a regression test for a crash in loop unroll. When there are multiple
+; exits, the compiler will run hlsl::RemoveUnstructuredLoopExits to try to
+; avoid unstructured code.
 ;
-; In this test, -loop-unroll will try to unroll the middle
-; loop. The exit edge from %land.lhs.true to %if.then will
-; be removed, and %if.end will be split at the beginning,
-; and branch to %if.end instead.
+; In this test, the compiler will try to unroll the middle loop. The exit edge
+; from %land.lhs.true to %if.then will be removed, and %if.end will be split at
+; the beginning, and branch to %if.end instead.
 ;
-; Since the new split block at %if.end becomes the new latch of the inner-most loop, it needs to be added to the
-; Loop analysis structure of the inner loop. However, it was only added to the current middle loop that is being unrolled.
+; Since the new split block at %if.end becomes the new latch of the inner-most
+; loop, it needs to be added to the Loop analysis structure of the inner loop.
+; However, it was only added to the current middle loop that is being unrolled.
 
 target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
 target triple = "dxil-ms-dx"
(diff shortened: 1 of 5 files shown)


utils/hct/hctdb.py
```diff
@@ -6680,6 +6680,12 @@ def add_pass(name, type_name, doc, opts):
                     "t": "unsigned",
                     "d": "Unrolled size limit for loops with an unroll(full) or unroll_count pragma.",
                 },
+                {
+                    "n": "StructurizeLoopExits",
```
damyanp, line 6684: Can you help me understand what this change does?  Does the test above provide coverage for this change?

adam-yang: This script generates some boilerplate code for the pass arguments. I'll have to double check what exact code it generates, but I think it's automatically called by cmake.

dmpots (commented):
LGTM

utils/hct/hctdb.py
llvm-beanz: IIUC, without this change you can’t set `StructureizeLoopExits` on the `dxopt` command line, which the test case above does.

tools/clang/test/HLSLFileCheck/hlsl/control_flow/loops/struct_exit_inner_latch_regression.ll
```diff
@@ -0,0 +1,75 @@
+; RUN: %opt %s -S -loop-unroll,StructurizeLoopExits=1 | FileCheck %s
+; RUN: %opt %s -S -dxil-loop-unroll,StructurizeLoopExits=1 | FileCheck %s
```
llvm-beanz, line 2: These are `dxopt` command lines not `opt`. Can we try putting this test somewhere that it runs with LIT instead of the legacy infrastructure?

adam-yang: I don't think opt can run passes with arguments like `StructurizeLoopExits`. These options are off by default.

llvm-beanz: This isn’t using a `cl::opt` option, it is running with the `dxopt` option handling. I’m pretty sure that’s what the python script generates.

adam-yang: Yeah it's using the `applyOptions` function added to the pass. the hctdb.py must be generating the `build\DXC\lib\HLSL\DxcOptimizer.inc` file that gets included in `DxcOptimizer.cpp`.

dmpots (approved):

tools/clang/test/HLSLFileCheck/hlsl/control_flow/loops/struct_exit_inner_latch_regression.ll
llvm-beanz: I don't want to hold this change up, but I'm unsure why you marked this as resolved.

When run in the `HLSLFileCheck` folder `%opt` is actually `dxopt` not LLVM's `opt`. I'm pretty sure we can move this test into the `tools/clang/test/DXC/*` folder and change the `RUN` lines to `%dxopt`, and they should work just fine.

That makes this test run through LIT, using LIT's test scheduler which better distributes test execution.

adam-yang: Oh I see what you mean now. Change incoming.

llvm-beanz (approved):

tools/clang/test/HLSLFileCheck/hlsl/control_flow/loops/struct_exit_inner_latch_regression.ll
adam-yang: Done.

PR #6551, opened 2024-04-19
PIX: Cope with arbitrary <32 integer widths in debugger pass

jeffnn:
Who knew codegen could generate i25 as part of a switch argument? Not me!


damyanp (approved):

tools/clang/test/HLSLFileCheck/pix/Debugi25.hlsl
```diff
@@ -0,0 +1,32 @@
+// RUN: %dxc -Emain -Tps_6_1 %s | %opt -S -dxil-dbg-value-to-dbg-declare -dxil-annotate-with-virtual-regs -hlsl-dxil-debug-instrumentation,UAVSize=65536 | %FileCheck %s
+
+// Expect an i25 cast, or this test isn't testing anything:
+// CHECK: [[CAST:%.*]] = trunc i32 %{{.*}} to i25
```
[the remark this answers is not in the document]
tex3d: I've filed issues #6563 for validation and #6564 for the optimization producing this i25 value.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -1394,6 +1407,7 @@ bool DxilDebugInstrumentation::RunOnFunction(Module &M, DxilModule &DM,
   case DXIL::ShaderKind::AnyHit:
   case DXIL::ShaderKind::ClosestHit:
   case DXIL::ShaderKind::Miss:
+  case DXIL::ShaderKind::Node:
```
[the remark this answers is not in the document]
jeffnn: Arg, too many projects going on at once. Removed.

hekota (approved):
LGTM!

tools/clang/test/HLSLFileCheck/pix/Debugi25.hlsl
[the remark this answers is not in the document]
jeffnn: Moved to a .ll-based check. 
Unfortunately, this required some changes to opt.exe. Is this OK?

tools/clang/test/DXC/Passes/PIX/debugi25.ll
```diff
@@ -0,0 +1,103 @@
+; RUN: %dxopt %s -dxil-annotate-with-virtual-regs -hlsl-dxil-debug-instrumentation -S | FileCheck %s
+
+; Expect an i25 cast, or this test isn't testing anything:
+; CHECK: [[CAST:%.*]] = trunc i32 %{{.*}} to i25
+;
+; Check that we correctly z-extended that i25 before trying to write it to i32
+; [[ZEXT:%.*]] = zext i25 [[CAST]] to i32
+; call void @dx.op.bufferStore.i32(i32 69, %dx.types.Handle %PIX_DebugUAV_Handle, i32 %{{.*}}, i32 undef, i32 [[ZEXT]]
```
tex3d, line 8: These lines are missing `CHECK:` prefix, so they won't be checked.

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
```diff
@@ -1151,13 +1150,23 @@ DxilDebugInstrumentation::addStepDebugEntryValue(BuilderContext *BC,
                                  ValueOrdinalIndex);
     return DebugShaderModifierRecordTypeDXILStepFloat;
   case Type::TypeID::IntegerTyID:
+    if (V->getType()->getIntegerBitWidth() > 64) {
+      // If this actually happens, we'd like to catch it in debug
+      assert(false);
```
tex3d, line 1155: `assert(false)` is an anti-pattern, since it's not locally informative.  Could an assert be written before these if conditions that checks the size against the set of expected sizes instead?  Something like:
```cpp
assert(V->getType()->getIntegerBitWidth() == 32 ||
       V->getType()->getIntegerBitWidth() == 64);
```
Using DXASSERT with a message would also seem helpful, but I'm not sure if that's fallen out of favor for use in passes.

tools/clang/test/DXC/Passes/PIX/debugi25.ll
jeffnn: D'Oh! Musta deleted them when copying the file. Still passes with them :-)

lib/DxilPIXPasses/DxilDebugInstrumentation.cpp
jeffnn: Right you are. Laziness on my part.

PR #6553, opened 2024-04-22
[Sema] Report warning when mix packoffset with nonpackoffset

### LINKED ISSUE #3724: cbuffer packoffset insufficiently validated
Some invalid cases below with fxc's errors in comments.  In fxc, repro later errors by fixing earlier ones.

```HLSL
// dxc -T vs_6_0
struct Foo {
  float f;
  uint2 u2;
};

struct Vec4 {
  float4 v4;
};

cbuffer CB0 {
  float f : packoffset(c17.y);

  // structs must start at 16-byte alignment (.x)
  // error X3530: register or offset bind c2.y not valid
  Foo foo : packoffset(c2.y);

  // also, dxc crashes if you offset a structure with a vector such that
  // the vector straddles the alignment in an illegal way, like so:
  // Vec4 v4 : packoffset(c1.y);

  // error X3530: cannot mix packoffset elements with nonpackoffset elements in a cbuffer
  uint u1;

  // error X4019: multiple variables found with the same user-specified location
  uint u3 : packoffset(c2.y);

  // additionally, reported cbuffer size will be wrong, but that part may be fixed soon.
}

uint main() : OUT {
  return foo.u2.x * u3;
}
```

python3kgae:
Check  mix packoffset elements with nonpackoffset elements in a cbuffer at Sema::ActOnFinishHLSLBuffer. Also check packoffset for aggregate type.

Fixes #3724

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7544,7 +7544,7 @@ def err_hlsl_objectintemplateargument : Error<
   "%0 is an object and cannot be used as a type parameter">;
 def err_hlsl_packoffset_requires_cbuffer : Error<
   "packoffset is only allowed in a constant buffer">;
-def err_hlsl_packoffset_mix : Error<
+def warn_hlsl_packoffset_mix : Warning<
   "cannot mix packoffset elements with nonpackoffset elements in a cbuffer">;
 def err_hlsl_packoffset_overlap : Error<"packoffset overlap between %0, %1">;
 def err_hlsl_register_semantics_conflicting : Error<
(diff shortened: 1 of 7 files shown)


damyanp:
What was the previous behavior with mixed offsets?  Is this change likely to break any existing code that has been working under the assumption that the previous behavior was the correct behavior?

python3kgae:
> What was the previous behavior with mixed offsets? Is this change likely to break any existing code that has been working under the assumption that the previous behavior was the correct behavior?

The previous behavior is scanning all the packoffset first, then allocate the elements without packoffset after all the packoffset elements.
```
cbuffer A {
  float3 a : packoffset(c2);
  float   b;  // b will be at offset c2.w
  float   c: packoffset(c1);
}
```

This change will report error on the above code which compiles OK in dxc before.
It will get error in fxc.

damyanp (approved):
LGTM. It might be good to try and get a review from the SPIR-V maintainers as well.

sudonatalie (approved):
LGTM on the SPIR-V side, thanks!

PR #6556, opened 2024-04-22
Fix invalid IR from scalarrepl-param-hlsl in ReplaceConstantWithInst

amaiorano:
ReplaceConstantWithInst(C, V) replaces uses of C in the current function with V. If such a use C is an instruction I, the it replaces uses of C in I with V. However, this function did not make sure to only perform this replacement if V dominates I. As a result, it may end up replacing uses of C in instructions before the definition of V.

The fix is to lazily compute the dominator tree in ReplaceConstantWithInst so that we can guard the replacement with that dominance check.

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -3271,15 +3271,35 @@ bool SROA_Helper::DoScalarReplacement(GlobalVariable *GV,
   return true;
 }
 
-static void ReplaceConstantWithInst(Constant *C, Value *V,
+// Replaces uses of constant C in the current function
+// with V, when those uses are dominated by V.
+// Returns true if it was completely replaced.
+static bool ReplaceConstantWithInst(Constant *C, Value *V,
                                     IRBuilder<> &Builder) {
+  bool bReplacedAll = true;
   Function *F = Builder.GetInsertBlock()->getParent();
+  Instruction *VInst = dyn_cast<Instruction>(V);
+  // Lazily calculate dominance
+  DominatorTree DT;
+  bool Calculated = false;
+  auto Dominates = [&](llvm::Instruction *Def, llvm::Instruction *User) {
+    if (!Calculated) {
+      DT.recalculate(*F);
+      Calculated = true;
+    }
+    return DT.dominates(Def, User);
+  };
+
   for (auto it = C->user_begin(); it != C->user_end();) {
     User *U = *(it++);
     if (Instruction *I = dyn_cast<Instruction>(U)) {
       if (I->getParent()->getParent() != F)
         continue;
-      I->replaceUsesOfWith(C, V);
+      if (VInst && Dominates(VInst, I)) {
+        I->replaceUsesOfWith(C, V);
+      } else {
+        bReplacedAll = false;
+      }
     } else {
       // Skip unused ConstantExpr.
       if (U->user_empty())
@@ -3288,10 +3308,13 @@ static void ReplaceConstantWithInst(Constant *C, Value *V,
       Instruction *Inst = CE->getAsInstruction();
       Builder.Insert(Inst);
       Inst->replaceUsesOfWith(C, V);
-      ReplaceConstantWithInst(CE, Inst, Builder);
+      if (!ReplaceConstantWithInst(CE, Inst, Builder)) {
+        bReplacedAll = false;
+      }
     }
   }
   C->removeDeadConstantUsers();
+  return bReplacedAll;
 }
 
 static void ReplaceUnboundedArrayUses(Value *V, Value *Src) {
@@ -3531,7 +3554,9 @@ static bool ReplaceMemcpy(Value *V, Value *Src, MemCpyInst *MC,
       } else {
         // Replace Constant with a non-Constant.
         IRBuilder<> Builder(MC);
-        ReplaceConstantWithInst(C, Src, Builder);
+        if (!ReplaceConstantWithInst(C, Src, Builder)) {
+          return false;
+        }
       }
     } else {
       // Try convert special pattern for cbuffer which copy array of float4 to
@@ -3539,7 +3564,9 @@ static bool ReplaceMemcpy(Value *V, Value *Src, MemCpyInst *MC,
       if (!tryToReplaceCBVec4ArrayToScalarArray(V, TyV, Src, TySrc, MC, DL)) {
         IRBuilder<> Builder(MC);
         Src = Builder.CreateBitCast(Src, V->getType());
-        ReplaceConstantWithInst(C, Src, Builder);
+        if (!ReplaceConstantWithInst(C, Src, Builder)) {
+          return false;
+        }
       }
     }
   } else {
@@ -3678,13 +3705,15 @@ static bool ReplaceUseOfZeroInit(Instruction *I, Value *V, DominatorTree &DT,
       continue;
 
     // Skip properly dominated users
-    if (DT.properlyDominates(BB, UI->getParent()))
+    if (DT.properlyDominates(BB, UI->getParent())) {
       continue;
+    }
 
     // If user is found in memcpy successor list
     // then the user is not safe to replace with zeroinitializer.
-    if (Reachable.count(UI->getParent()))
+    if (Reachable.count(UI->getParent())) {
       return false;
+    }
 
     // Remaining cases are where I:
     // - is at the end of the same block
@@ -5449,9 +5478,9 @@ void SROA_Parameter_HLSL::flattenArgument(
         if (Ty->isPointerTy())
           Ty = Ty->getPointerElementType();
         unsigned size = DL.getTypeAllocSize(Ty);
-#if 0  // HLSL Change
+#if 0 // HLSL Change
         DIExpression *DDIExp = DIB.createBitPieceExpression(debugOffset, size);
-#else  // HLSL Change
+#else // HLSL Change
         Type *argTy = Arg->getType();
         if (argTy->isPointerTy())
           argTy = argTy->getPointerElementType();
--- a/tools/clang/test/DXC/scalarrepl-param-hlsl-const-to-local-and-back.hlsl
+++ b/tools/clang/test/DXC/scalarrepl-param-hlsl-const-to-local-and-back.hlsl
@@ -1,0 +1,46 @@
+// RUN: %dxc -T ps_6_2 %s | FileCheck %s
+
+// Validate that copying from static array to local, then back to static
+// array does not crash the compiler. This was resulting in an invalid
+// ReplaceConstantWithInst from ScalarReplAggregatesHLSL, which would
+// result in referenced deleted instruction in a later pass.
+
+// This test is expected to fail with "error: Loop must have a break."
+// XFAIL: *
+
+static int arr1[10] = (int[10])0;
+static int arr2[10] = (int[10])0;
+static float result = 0;
+ByteAddressBuffer buff : register(t0);
+
+void foo() {
+  int i = 0;
+  if (buff.Load(0u)) {
+    return;
+  }
+  arr2[i] = arr1[i];
+  result = float(arr1[0]);
+}
+
+struct tint_symbol {
+  float4 value : SV_Target0;
+};
+
+float main_inner() {
+  foo();
+  bool cond = false;
+  while (true) {
+    if (cond) { break; }
+  }
+  int arr1_copy[10] = arr1; // constant to local
+  arr1 = arr1_copy; // local to constant
+  foo();
+  return result;
+}
+
+tint_symbol main() {
+  float inner_result = main_inner();
+  tint_symbol wrapper_result = (tint_symbol)0;
+  wrapper_result.value.x = inner_result;
+  return wrapper_result;
+}
--- a/tools/clang/test/DXC/scalarrepl-param-hlsl-const-to-local-and-back.ll
+++ b/tools/clang/test/DXC/scalarrepl-param-hlsl-const-to-local-and-back.ll
@@ -1,0 +1,253 @@
+; RUN: %dxopt %s -hlsl-passes-resume -scalarrepl-param-hlsl -S | FileCheck %s
+
+; The first memcpy, from arr1 to arr1_copy.i, should be replaced by a series of 10 loads and stores,
+; while the second memcpy, from arr1_copy.i back to arr1, should be removed:
+;   %19 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33 ; line:25 col:23
+;   call void @llvm.memcpy.p0i8.p0i8.i64(i8* %19, i8* bitcast ([10 x i32]* @arr1 to i8*), i64 40, i32 1, i1 false) #0, !dbg !33 ; line:25 col:23
+;   %20 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !34 ; line:26 col:10
+;   call void @llvm.memcpy.p0i8.p0i8.i64(i8* bitcast ([10 x i32]* @arr1 to i8*), i8* %20, i64 40, i32 1, i1 false) #0, !dbg !34 ; line:26 col:10
+;   store i32 0, i32* %i.i.1.i, align 4, !dbg !35, !tbaa !12 ; line:7 col:7
+
+; CHECK:  [[DEST0:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 0
+; CHECK-NEXT:  [[SRC0:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 0)
+; CHECK-NEXT:  store i32 [[SRC0:%[a-z0-9\.]+]], i32* [[DEST0:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST1:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 1
+; CHECK-NEXT:  [[SRC1:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 1)
+; CHECK-NEXT:  store i32 [[SRC1:%[a-z0-9\.]+]], i32* [[DEST1:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST2:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 2
+; CHECK-NEXT:  [[SRC2:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 2)
+; CHECK-NEXT:  store i32 [[SRC2:%[a-z0-9\.]+]], i32* [[DEST2:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST3:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 3
+; CHECK-NEXT:  [[SRC3:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 3)
+; CHECK-NEXT:  store i32 [[SRC3:%[a-z0-9\.]+]], i32* [[DEST3:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST4:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 4
+; CHECK-NEXT:  [[SRC4:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 4)
+; CHECK-NEXT:  store i32 [[SRC4:%[a-z0-9\.]+]], i32* [[DEST4:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST5:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 5
+; CHECK-NEXT:  [[SRC5:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 5)
+; CHECK-NEXT:  store i32 [[SRC5:%[a-z0-9\.]+]], i32* [[DEST5:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST6:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 6
+; CHECK-NEXT:  [[SRC6:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 6)
+; CHECK-NEXT:  store i32 [[SRC6:%[a-z0-9\.]+]], i32* [[DEST6:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST7:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 7
+; CHECK-NEXT:  [[SRC7:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 7)
+; CHECK-NEXT:  store i32 [[SRC7:%[a-z0-9\.]+]], i32* [[DEST7:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST8:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 8
+; CHECK-NEXT:  [[SRC8:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 8)
+; CHECK-NEXT:  store i32 [[SRC8:%[a-z0-9\.]+]], i32* [[DEST8:%[a-z0-9\.]+]]
+; CHECK-NEXT:  [[DEST9:%[a-z0-9\.]+]] = getelementptr inbounds [10 x i32], [10 x i32]* %arr1_copy.i, i32 0, i32 9
+; CHECK-NEXT:  [[SRC9:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds ([10 x i32], [10 x i32]* @arr1, i32 0, i32 9)
+; CHECK-NEXT:  store i32 [[SRC9:%[a-z0-9\.]+]], i32* [[DEST9:%[a-z0-9\.]+]]
+
+;
+; Buffer Definitions:
+;
+; cbuffer $Globals
+; {
+;
+;   [0 x i8] (type annotation not present)
+;
+; }
+;
+;
+; Resource Bindings:
+;
+; Name                                 Type  Format         Dim      ID      HLSL Bind  Count
+; ------------------------------ ---------- ------- ----------- ------- -------------- ------
+; $Globals                          cbuffer      NA          NA     CB0   cb4294967295     1
+; buff                              texture    byte         r/o      T0             t0     1
+;
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+%struct.ByteAddressBuffer = type { i32 }
+%ConstantBuffer = type opaque
+%struct.tint_symbol = type { <4 x float> }
+%dx.types.Handle = type { i8* }
+%dx.types.ResourceProperties = type { i32, i32 }
+
+@"\01?buff@@3UByteAddressBuffer@@A" = external global %struct.ByteAddressBuffer, align 4
+@arr1 = internal global [10 x i32] zeroinitializer, align 4
+@arr2 = internal global [10 x i32] zeroinitializer, align 4
+@"$Globals" = external constant %ConstantBuffer
+
+; Function Attrs: nounwind
+define void @main(%struct.tint_symbol* noalias sret %agg.result) #0 {
+  %1 = alloca float
+  store float 0.000000e+00, float* %1
+  %i.i.1.i = alloca i32, align 4
+  %i.i.i = alloca i32, align 4
+  %cond.i = alloca i32, align 4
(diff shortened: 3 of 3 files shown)


amaiorano:
Note that this was a joint effort with @dneto0. See the test file in the commit. Also see this [compiler explorer link](https://godbolt.org/z/aqn8YcdT8) with the same example that segfaults.

The following explains the fix:

The bug is in `ReplaceConstantWIthInst(Constant *C, Value *V, Builder..)`.

Here's the relevant IR:
```
while.end.i:                                      ; preds = %while.body.i
  %10 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %10, i8* bitcast ([10 x i32]* @arr1 to i8*), i64 40, i32 1, i1 false) #0, !dbg !33
  %11 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !34
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* bitcast ([10 x i32]* @arr1 to i8*), i8* %11, i64 40, i32 1, i1 false) #0, !dbg !34
  store i32 0, i32* %i.i.1.i, align 4, !dbg !35, !tbaa !12
  %12 = load %struct.ByteAddressBuffer, %struct.ByteAddressBuffer* @"\01?buff@@3UByteAddressBuffer@@A", !dbg !37
```

There are two memcpys. (dest is first arg, src is second arg)
* `ReplaceMemcpy(Value=@arr1,Src=%11)` tries to replace the second memcpy: roughly `memcpy(@arr1, %11)` a.k.a. `memcpy(@arr1, %arr1_copy.i)`, where `%arr1_copy.i` is the function-local array variable.
* It sees that Value is a constant (because its the address of global var `@arr`)
* It sees that Src is not a constant, because it's the result of an alloca (a.k.a. stack variable address)
* So it falls into the code path with the comment ` // Replace Constant with a non-Constant. and calls ReplaceConstantWithInst`

Now let's enter `ReplaceConstantWithInst(C=@arr, V=%arr1_copy.i)`. Its job is to replace uses of C, in the current function, with the value V. It traverses the uses of C, skipping over any that aren't in the current function (i.e. the function containing the instruction that generates V). If such a use C is an instruction I, then it replaces uses of C in I with V. Remember C=@arr, and one of the uses is the first memcpy: `memcpy(bitcast of %arr1_copy.i, bitcast of @arr)`.

After these replacements, the first part of that basic block is this:
```
while.end.i:                                      ; preds = %while.body.i
  %10 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !32
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %10, i8* %12, i64 40, i32 1, i1 false) #0, !dbg !32
  %11 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
  %12 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %12, i8* %11, i64 40, i32 1, i1 false) #0, !dbg !33
  store i32 0, i32* %i.i.1.i, align 4, !dbg !34, !tbaa !12
```

We see that %12 is in the first memcpy, but defined two instructions later. That's bad.

The bug is in this statement: If such a use C is an instruction I, then it replaces uses of C in I with V. That's only safe to do if V dominates I.

So the fix uses a big hammer: compute the dominator tree so we can guard the replacement with that dominance check. Although it may seem heavyweight to recompute the dominator tree repeatedly, it's the safe thing to do. Earlier in the code it says:
```cpp
 // SROA_Parameter_HLSL has no access to a domtree, if one is needed, it'll
 // be generated
```

So it seems in the spirit to compute the dominator tree on demand in this pass.

Note that early in ReplaceMemcpy, it has a promising comment and mechanism to early-out:
```cpp
  // If the source of the memcpy (Src) doesn't dominate all users of dest (V),
  // full replacement isn't possible without complicated PHI insertion
  // This will likely replace with ld/st which will be replaced in mem2reg
  if (Instruction *SrcI = dyn_cast<Instruction>(Src))
    if (!DominateAllUsers(SrcI, V, DT))
      return false;
```

But DominateAllUsers doesn't quite do the job for us. It calls DominateAllUsersDom (where it has a dominator tree), and it does this code:
```cpp
// Use `DT` to trace all users and make sure `I`'s BB dominates them all
static bool DominateAllUsersDom(Instruction *I, Value *V, DominatorTree *DT) {
  BasicBlock *BB = I->getParent();
  Function *F = I->getParent()->getParent();
  for (auto U = V->user_begin(); U != V->user_end();) {
    Instruction *UI = dyn_cast<Instruction>(*(U++));
    // If not an instruction or from a differnt function, nothing to check, move
    // along.
    if (!UI || UI->getParent()->getParent() != F)
      continue;

    if (!DT->dominates(BB, UI->getParent()))
      return false;
      
    if (isa<GetElementPtrInst>(UI) || isa<BitCastInst>(UI)) {
      if (!DominateAllUsersDom(I, UI, DT))
        return false;
    }
  }
  return true;
}
```

First, in our case value V is a constant (`@arr`), and its use in the first memcpy is `i8* bitcast ([10 x i32]* @arr1 to i8*` which is itself an llvm::Constant. So it early outs because UI here is null because that llvm::Constant is not an instruction. So I think it returns true, and that's why we get into trouble later.

One local fix here could be if the use is a constant, then recurse and follow the users of that enclosing constant, until you hit an instruction. This has a problem that it could blow up the search.

And even if you got out far enough to reach an instruction you would get to the first memcpy. And then the test `if (!DT->dominates(BB, UI->getParent()))` only compares at a basic block level. But both memcpys are in the same basic block, and a node always dominates itself. So again it would decide incorrectly. You have to check dominance on an instruction level.

tools/clang/test/DXC/scalarrepl-param-hlsl-const-to-local-and-back.hlsl
```diff
@@ -0,0 +1,46 @@
+// RUN: %dxc -T ps_6_2 %s | FileCheck %s
```
llvm-beanz, line 1:
```suggestion
// RUN: not %dxc -T ps_6_2 %s 2>&1 | FileCheck %s
```
LIT `RUN` lines are basically just shell commands with some extra variable expansions. If the sell command returns success the test passes if it returns failure it fails.

Prefacing the call with `not` inverts the return code of the process (success becomes failure, failure becomes success). This allows LIT to run commands that are expected to fail and report an error if they succeed.

Adding `2>&1` is fun POSIX shell for redirect standard error to standard out, which we then pipe into FileCheck, so that below we can check the output.

tools/clang/test/DXC/scalarrepl-param-hlsl-const-to-local-and-back.hlsl
```diff
@@ -0,0 +1,46 @@
...
+
+// Validate that copying from static array to local, then back to static
+// array does not crash the compiler. This was resulting in an invalid
+// ReplaceConstantWithInst from ScalarReplAggregatesHLSL, which would
+// result in referenced deleted instruction in a later pass.
+
+// This test is expected to fail with "error: Loop must have a break."
+// XFAIL: *
```
llvm-beanz, line 9:
```suggestion
// CHECK: error: Loop must have a break.
```

And here we can check for the expected error string instead of `XFAIL`.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3288,10 +3308,13 @@ static void ReplaceConstantWithInst(Constant *C, Value *V,
       Instruction *Inst = CE->getAsInstruction();
       Builder.Insert(Inst);
       Inst->replaceUsesOfWith(C, V);
-      ReplaceConstantWithInst(CE, Inst, Builder);
+      if (!ReplaceConstantWithInst(CE, Inst, Builder)) {
+        bReplacedAll = false;
+      }
```
llvm-beanz, line 3313: nit:
```suggestion
      if (!ReplaceConstantWithInst(CE, Inst, Builder))
        bReplacedAll = false;
```
see: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3531,15 +3554,19 @@ static bool ReplaceMemcpy(Value *V, Value *Src, MemCpyInst *MC,
       } else {
         // Replace Constant with a non-Constant.
         IRBuilder<> Builder(MC);
-        ReplaceConstantWithInst(C, Src, Builder);
+        if (!ReplaceConstantWithInst(C, Src, Builder)) {
+          return false;
+        }
```
llvm-beanz, line 3559: nit:
```suggestion
        if (!ReplaceConstantWithInst(C, Src, Builder))
          return false;
```
see: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3531,15 +3554,19 @@ static bool ReplaceMemcpy(Value *V, Value *Src, MemCpyInst *MC,
...
     } else {
       // Try convert special pattern for cbuffer which copy array of float4 to
       // array of float.
       if (!tryToReplaceCBVec4ArrayToScalarArray(V, TyV, Src, TySrc, MC, DL)) {
         IRBuilder<> Builder(MC);
         Src = Builder.CreateBitCast(Src, V->getType());
-        ReplaceConstantWithInst(C, Src, Builder);
+        if (!ReplaceConstantWithInst(C, Src, Builder)) {
+          return false;
+        }
```
llvm-beanz, line 3569: nit:
```suggestion
        if (!ReplaceConstantWithInst(C, Src, Builder))
          return false;
```
see: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3678,13 +3705,15 @@ static bool ReplaceUseOfZeroInit(Instruction *I, Value *V, DominatorTree &DT,
       continue;
 
     // Skip properly dominated users
-    if (DT.properlyDominates(BB, UI->getParent()))
+    if (DT.properlyDominates(BB, UI->getParent())) {
       continue;
+    }
```
llvm-beanz, line 3710: nit:
```suggestion
    if (DT.properlyDominates(BB, UI->getParent()))
      continue;
```
see: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

llvm-beanz (commented):
To avoid the `tools/clang/test/DXC` directory becoming a dumping ground, can you move these tests into a new subdirectory, maybe `tools/clang/test/DXC/Passes/ScalarReplHLSL`?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3271,15 +3271,35 @@ bool SROA_Helper::DoScalarReplacement(GlobalVariable *GV,
...
+    return DT.dominates(Def, User);
+  };
+
   for (auto it = C->user_begin(); it != C->user_end();) {
     User *U = *(it++);
     if (Instruction *I = dyn_cast<Instruction>(U)) {
       if (I->getParent()->getParent() != F)
         continue;
-      I->replaceUsesOfWith(C, V);
+      if (VInst && Dominates(VInst, I)) {
```
pow2clk, line 3298: This repeats a check performed here: https://github.com/microsoft/DirectXShaderCompiler/blob/1c66c4d798b02c62510a7b751303bea4cfd10849/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp#L3521

Can you explain why it's necessary? Maybe there's a recalculation missed somewhere? The intent of the original code was to minimize the number of times recalculation was needed. The new code here will recalculate it every time `ReplaceConstantWithInst` is called.

At the very least, the original check shouldn't be necessary

pow2clk (commented):
I'd like to understand why the existing dominance check was insufficient.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3531,15 +3554,19 @@ static bool ReplaceMemcpy(Value *V, Value *Src, MemCpyInst *MC,
...
+        }
       }
     } else {
       // Try convert special pattern for cbuffer which copy array of float4 to
       // array of float.
       if (!tryToReplaceCBVec4ArrayToScalarArray(V, TyV, Src, TySrc, MC, DL)) {
         IRBuilder<> Builder(MC);
         Src = Builder.CreateBitCast(Src, V->getType());
-        ReplaceConstantWithInst(C, Src, Builder);
+        if (!ReplaceConstantWithInst(C, Src, Builder)) {
```
pow2clk, line 3567: I suspect this is where the additional dominance check is needed as we've created a new Src here. Alternately, we could create the src where it is sure to dominate uses of C by placing it just after Src instead of just after MC.  This would prevent the need to recalculate the dominance tree. Alternately, we could recalculate the tree and check again for total dominance here, but I prefer the first option.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
amaiorano: The original check doesn't do the trick because the very first condition in `DominateAllUsers` is true:
```cpp
  // The Entry Block dominates everything, trivially true
  if (&F->getEntryBlock() == I->getParent())
    return true;
```
Even if we comment this out, and execute the code that computes the DT and calls `DominateAllUsersDom` below that, it still ends up returning true:
```cpp
  if (!DT) {
    DominatorTree TempDT;
    TempDT.recalculate(*F);
    return DominateAllUsersDom(I, V, &TempDT);
  } else {
    return DominateAllUsersDom(I, V, DT);
  }
```

The reason, I think, is because unlike in `ReplaceConstantWithInst`, `DominateAllUsersDom` doesn't handle the case of users of `V` being `ConstantExpr`. In the former, when the user is a `ConstantExpr`, it recurses with the `CE->getAsInstruction`, although it makes sure to insert it into the BB so that it's parent's parent is the current Function. I tried very hard to make `DominateAllUsersDom` do this, but it's tricky because in this case, we don't want to insert and instructions.

> At the very least, the original check shouldn't be necessary

Maybe? The thing is, this fix is only for the case of replacing a constant with a non-constant. I'm not sure if the other cases are properly avoided by this original check.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
amaiorano: When running against the example that causes the crash, we don't actually execute this line. Instead, we enter the call to `ReplaceConstantWithInst` in the first condition block about 10 lines above this one.

amaiorano:
@pow2clk, you asked good questions yesterday about why this change is needed. Here's a deeper dive into what's happening when we run DXC against the example code, `scalarrepl-param-hlsl-const-to-local-and-back.hlsl`. Hopefully this makes the rationale for the patch more clear.

You asked about why the initial call to DominateAllUsers wasn't good enough to determine that we shouldn't proceed with replacing the memcpy. I set a breakpoint in `DominateAllUsers`.

Call stack:
```
> dxcompiler.dll!DominateAllUsers(llvm::Instruction * I=0x0000028f0f546768, llvm::Value * V=0x0000028f0d72a318, llvm::DominatorTree * DT=0x0000000000000000) Line 3857	C++
 	dxcompiler.dll!ReplaceMemcpy(llvm::Value * V=0x0000028f0d72a318, llvm::Value * Src=0x0000028f0f546768, llvm::MemCpyInst * MC=0x0000028f0f546da0, hlsl::DxilFieldAnnotation * annotation=0x0000000000000000, hlsl::DxilTypeSystem & typeSys={...}, const llvm::DataLayout & DL={...}, llvm::DominatorTree * DT=0x0000000000000000) Line 3556	C++
 	dxcompiler.dll!`anonymous namespace'::SROA_Helper::LowerMemcpy(llvm::Value * V=0x0000028f0d72a318, hlsl::DxilFieldAnnotation * annotation=0x0000000000000000, hlsl::DxilTypeSystem & typeSys={...}, const llvm::DataLayout & DL={...}, llvm::DominatorTree * DT=0x0000000000000000, bool bAllowReplace=true) Line 4074	C++
 	dxcompiler.dll!`anonymous namespace'::SROAGlobalAndAllocas(hlsl::HLModule & HLM={...}, bool bHasDbgInfo=false) Line 1950	C++
 	dxcompiler.dll!`anonymous namespace'::SROA_Parameter_HLSL::runOnModule(llvm::Module & M={...}) Line 4381	C++
 	dxcompiler.dll!`anonymous namespace'::MPPassManager::runOnModule(llvm::Module & M={...}) Line 1669	C++
...
```

Looking at the call stack, we see that we are processing globals and allocas:
```
GV->dump()
@arr1 = internal global [10 x i32] zeroinitializer, align 4
```
Which corresponds to:
```
static int arr1[10] = (int[10])0;
```

In `SROA_Helper::LowerMemcpy`, we determine that GV has a non-undef initializer, and is `MemcopyDestOnce`, so we will replace the source of the memcpy. The memcpy and Src are:
```
MC->dump()
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* bitcast ([10 x i32]* @arr1 to i8*), i8* %11, i64 40, i32 1, i1 false) #0, !dbg !34

Src->dump()
  %arr1_copy.i = alloca [10 x i32], align 4
```

Note that `Src` is an `llvm::AllocaInst` of `arr1_copy`.

In HLSL, this refers to:
```
  int arr1_copy[10] = arr1; // constant to local
  arr1 = arr1_copy; // local to constant
```

In IR, this looks like:
```
@arr1 = internal global [10 x i32] zeroinitializer, align 4
...

; Function Attrs: nounwind
define void @main(<4 x float>* noalias) #0 {
entry:
...
  %arr1_copy.i = alloca [10 x i32], align 4
...
while.end.i:                                      ; preds = %while.body.i
  %10 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %10, i8* bitcast ([10 x i32]* @arr1 to i8*), i64 40, i32 1, i1 false) #0, !dbg !33
  %11 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !34
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* bitcast ([10 x i32]* @arr1 to i8*), i8* %11, i64 40, i32 1, i1 false) #0, !dbg !34
...
```
We're currently lowering the 2nd memcpy - the one from `arr1_copy` to `arr1`.

In `SROA_Helper::LowerMemcpy` we are here, in the call to `ReplaceMemcpy`:
```cpp
        } else if (!isa<CallInst>(Src)) {
          // Resource ptr should not be replaced.
          // Need to make sure src not updated after current memcpy.
          // Check Src only have 1 store now.
          // If Src has more than 1 store but only used once by memcpy, check if
          // the stores dominate the memcpy.
          hlutil::PointerStatus SrcPS(Src, size, /*bLdStOnly*/ false);
          SrcPS.analyze(typeSys, bStructElt);
          if (SrcPS.storedType != hlutil::PointerStatus::StoredType::Stored ||
              (SrcPS.loadedType ==
                   hlutil::PointerStatus::LoadedType::MemcopySrcOnce &&
               allStoresDominateInst(Src, MC, DT))) {
            if (ReplaceMemcpy(V, Src, MC, annotation, typeSys, DL, DT)) { <------------ HERE
              if (V->user_empty())
                return true;
              return LowerMemcpy(V, annotation, typeSys, DL, DT, bAllowReplace);
            }
          }
```
It's true that Src (`arr1_copy`) is not updated after the current memcpy, and that all stores of Src dominate the memcpy, as the only store to `arr1_copy` is made just before it from `arr1` (the first memcpy). Here's `SrcPS`:
```
SrcPS
{storedType=MemcopyDestOnce (3) loadedType=MemcopySrcOnce (1) StoredOnceValue=0x0000000000000000 <NULL> ...}
    storedType: MemcopyDestOnce (3)
    loadedType: MemcopySrcOnce (1)
    StoredOnceValue: 0x0000000000000000 <NULL>
    memcpySet: {set_={...} vector_={ size=2 } }
    StoringMemcpy: 0x0000028f0f546cb0 68
    LoadingMemcpy: 0x0000028f0f546da0 68
    AccessingFunction: 0x0000028f0f483a18 FunctionVal (2)
    HasMultipleAccessingFunctions: false
    Size: 40
    Ptr: 0x0000028f0f546768 45
    bLoadStoreOnly: false
```

So now we're in `ReplaceMemcpy`.

```
  // If the source of the memcpy (Src) doesn't dominate all users of dest (V),
  // full replacement isn't possible without complicated PHI insertion
  // This will likely replace with ld/st which will be replaced in mem2reg
  if (Instruction *SrcI = dyn_cast<Instruction>(Src))
    if (!DominateAllUsers(SrcI, V, DT))
      return false;
```
So this is supposed to check whether Src, the `AllocaInst` of `arr1_copy` (`%arr1_copy.i = alloca [10 x i32], align 4`), dominates all users of the dest value V, `arr1` (`@arr1 = internal global [10 x i32] zeroinitializer, align 4`). If it doesn't, we bail on full replacement.

At the very start of the call to `DominateAllUsers`, there is this check:
```
  // The Entry Block dominates everything, trivially true
  if (&F->getEntryBlock() == I->getParent())
    return true;
```
Indeed, the `alloca` of `arr1_copy` is in the `entry` block of the function F:
```
; Function Attrs: nounwind
define void @main(<4 x float>* noalias) #0 {
entry:
  %agg.result.0 = alloca <4 x float>
  %1 = alloca float
  store float 0.000000e+00, float* %1
  %i.i.1.i = alloca i32, align 4
  %i.i.i = alloca i32, align 4
  %cond.i = alloca i32, align 4
  %arr1_copy.i = alloca [10 x i32], align 4 <------ HERE
```

I assume the reason for this is that `alloca`s are usually the first thing in a function's entry block, so if it's there, it necessarily dominates all users of the memcpy's destination.

Our use-case is one where there are 2 memcpys, where the first copies from `arr1` to `arr1_copy`, then the second from `arr1_copy` back to `arr1`:
```
  %10 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %10, i8* bitcast ([10 x i32]* @arr1 to i8*), i64 40, i32 1, i1 false) #0, !dbg !33
  %11 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !34
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* bitcast ([10 x i32]* @arr1 to i8*), i8* %11, i64 40, i32 1, i1 false) #0, !dbg !34
```
We're processing the second memcpy's destination, `arr1`, and sure enough, the src, `arr1_copy` does indeed dominate all *uses* of `arr1`, including that first memcpy. So we return `true` from `DominateAllUsers`.

The problem isn't the fact that Src doesn't dominate uses of Dest, but rather, than in replacing Dest `arr1`  with Src `arr1_copy`, we end up doing so to the memcpy that comes _before_ the one we're currently processing. The bug is that without any changes, the following:
```
while.end.i:                                      ; preds = %while.body.i
  %10 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %10, i8* bitcast ([10 x i32]* @arr1 to i8*), i64 40, i32 1, i1 false) #0, !dbg !33
  %11 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !34
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* bitcast ([10 x i32]* @arr1 to i8*), i8* %11, i64 40, i32 1, i1 false) #0, !dbg !34
```

Turns into this, right after the call to `ReplaceConstantWithInst(C = arr1, Src = alloca of arr1_copy)`:
```
while.end.i:                                      ; preds = %while.body.i
  %10 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !32
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %10, i8* %12, i64 40, i32 1, i1 false) #0, !dbg !32
  %11 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
  %12 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* %12, i8* %11, i64 40, i32 1, i1 false) #0, !dbg !33
```
Note that the first call to memcpy changed so that it's source went from `i8* bitcast ([10 x i32]* @arr1 to i8*)` to the new `%12`, but `%12` is defined below it, which is bad.

Now let's see _why_ does `ReplaceConstantWithInst` end up wanting to replace the source of that first memcpy with `%12` defined below it. In `ReplaceConstantWithInst(C = arr1, Src = alloca of arr1_copy, Builder)`, we iterate over the users of C, `arr1`. At one point, one such user is:
```
U->dump()
i8* bitcast ([10 x i32]* @arr1 to i8*)
```
This use is the one shown above that is the source of that first memcpy. This is not, itself, an `llvm::Instruction`, but is, instead, an `llvm::UnaryConstantExpr`. As such, we fall into the else branch of the user iteration loop:

```cpp
static bool ReplaceConstantWithInst(Constant *C, Value *V,
                                    IRBuilder<> &Builder) {
  bool replacedAll = true;
  Function *F = Builder.GetInsertBlock()->getParent();
  Instruction *VInst = dyn_cast<Instruction>(V);
  DominatorTree DT;
  if (VInst) {
    DT.recalculate(*F);
  }
  for (auto it = C->user_begin(); it != C->user_end();) {
    User *U = *(it++);
    if (Instruction *I = dyn_cast<Instruction>(U)) {
      if (I->getParent()->getParent() != F)
        continue;
      if (VInst && DT.dominates(VInst, I)) {
        I->replaceUsesOfWith(C, V);
      } else {
        replacedAll = false;
      }
    } else {
      <----------------------------------- HERE
      // Skip unused ConstantExpr.
      if (U->user_empty())
        continue;
      ConstantExpr *CE = cast<ConstantExpr>(U);
      Instruction *Inst = CE->getAsInstruction();
      Builder.Insert(Inst);
      Inst->replaceUsesOfWith(C, V);
      if (!ReplaceConstantWithInst(CE, Inst, Builder)) {
        replacedAll = false;
      }
    }
  }
  C->removeDeadConstantUsers();
  return replacedAll;
}
```

Since the use, U, is a `llvm::UnaryConstantExpr`, we use `getAsInstruction`, which creates a new `llvm::Instruction` that "implements the same operation as the ConstantExpr", and does not belong to a BB. Right after the call to `CE->getAsInstruction()`, (and inserting into the Builder - more on this below) `Inst` looks like this:
```
Inst->dump()
  %12 = bitcast [10 x i32]* @arr1 to i8*, !dbg !33
```
We then call `Inst->replaceUsesOfWith(C, V);` on it to replace the use of `arr1` with `arr1_copy`, and now it looks like this:
```
Inst->dump()
  %12 = bitcast [10 x i32]* %arr1_copy.i to i8*, !dbg !33
```

Before we recurse into `ReplaceConstantWithInst`, lets talk about _where_ this new instruction got inserted. We end up inserting this instruction at the current position in the current BB, which is `while.end.i`, right before the *2nd memcpy*, because the `IRBuilder` was built from that one in `ReplaceMemcpy`:
```cpp
  if (Constant *C = dyn_cast<Constant>(V)) {
    updateLifetimeForReplacement(V, Src);
    if (TyV == TySrc) {
      if (isa<Constant>(Src)) {
        V->replaceAllUsesWith(Src);
      } else {
        // Replace Constant with a non-Constant.
        IRBuilder<> Builder(MC);   <---------------------------- HERE
        if (!ReplaceConstantWithInst(C, Src, Builder)) {
          return false;
        }
      }
    } else {
```
So this new instruction is inserted _after_ the first memcpy in which we're going to replace it soon. We now recurse, calling `ReplaceConstantWithInst(C = CE, V = Inst, Builder)`, so that we can replace the `llvm::UnaryConstantExpr`, that is `bitcast [10 x i32]* @arr1 to i8*`, with our new instruction, ` %12 = bitcast [10 x i32]* %arr1_copy.i to i8*`. This is what ultimately leads to the first memcpy's source being changed to `%12`, which is bad.

So the bug is that we should not replace uses of a value C in I, with another one, V, if I dominates V; or conversely, we should only replace uses of value C in I with another one V if V dominates I. This is exactly what this patch does:

```cpp
static bool ReplaceConstantWithInst(Constant *C, Value *V,
                                    IRBuilder<> &Builder) {
  bool replacedAll = true;
  Function *F = Builder.GetInsertBlock()->getParent();
  Instruction *VInst = dyn_cast<Instruction>(V);
  DominatorTree DT;
  if (VInst) {
    DT.recalculate(*F);
  }
  for (auto it = C->user_begin(); it != C->user_end();) {
    User *U = *(it++);
    if (Instruction *I = dyn_cast<Instruction>(U)) {
      if (I->getParent()->getParent() != F)
        continue;
      if (VInst && DT.dominates(VInst, I)) { <------------ HERE
        I->replaceUsesOfWith(C, V);
      } else {
        replacedAll = false;
      }
    } else {
      // Skip unused ConstantExpr.
      if (U->user_empty())
        continue;
      ConstantExpr *CE = cast<ConstantExpr>(U);
      Instruction *Inst = CE->getAsInstruction();
      Builder.Insert(Inst);
      Inst->replaceUsesOfWith(C, V);
      if (!ReplaceConstantWithInst(CE, Inst, Builder)) {
        replacedAll = false;
      }
    }
  }
  C->removeDeadConstantUsers();
  return replacedAll;
}
```

Now maybe there's another way to solve this by making sure to insert our new instruction above all uses? Maybe instead of creating a builder around the 2nd memcpy, we should create it around the alloca of the Src `arr1_copy`? I'm not sure, as there are three places in `ReplaceMemcpy` that creates builders, and they all do so around the current memcpy instruction.

Hopefully this clarifies what's going on with this bug, and why our patch works this way. Basically, while replacing constants, we may end up injecting new instructions above the current memcpy, but then have to be careful not to replace references to it in instructions _before_ that memcpy.

tools/clang/test/DXC/scalarrepl-param-hlsl-const-to-local-and-back.hlsl
amaiorano: Done.

amaiorano:
@llvm-beanz addressed all your comments and 👍 each one.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
pow2clk: I've done some additional experimentation and you are correct that the problem is replacing constants, which dominate everything, with non-constants, which don't unless you put them in the right place. At present, the new instructions are being placed after the memcpy, but that doesn't always dominate everything. It prevents the crash, but has some changes in the output of the pass. I'll get back to you with it shortly.

amaiorano: Hi Greg, any update on this? Thanks!

llvm-beanz (approved):

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
pow2clk: Yeah. I was wrong about this supposition.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
pow2clk: I've uploaded my incomplete take on this change in this draft PR #6591, we can discuss it there particularly if this causes compilation time regressions.

pow2clk (approved):

Issue #6538, opened 2024-04-16
Update specs on new Barrier and block new Barrier on SM < 6.8


tex3d:
The new SM 6.8 `BarrierByMemoryType` DXIL operation is not translated down to prior DXIL operations on lower shader models, and the older barrier/sync HLSL intrinsics are not up-translated to the new operation (canonical representation) for intermediate IR or to the new `BarrierByMemoryType` DXIL operation for final DXIL on SM 6.8+.

A decision has been made to keep this behavior (for now), so specs need updating and Sema diagnostics need updating to diagnose on the use of the new Barrier op on earlier shader models.

- [x] Update spec to remove language about Barrier translation for different shader models
- [x] Add Sema diagnostic for new Barrier op when reachable from entry compiled to lower shader model.

PR #6558, opened 2024-04-23
Add diag for new Barrier intrinsic on SM < 6.8

tex3d:
Tests are split between one containing cases that could potentially be translated to legacy barrier op in the future, and a test without translatable barrier ops.

Part of #6538.

### DIFF
--- a/tools/clang/test/SemaHLSL/hlsl/intrinsics/barrier/barrier-cs-sm67-compat.hlsl
+++ b/tools/clang/test/SemaHLSL/hlsl/intrinsics/barrier/barrier-cs-sm67-compat.hlsl
@@ -1,0 +1,103 @@
+// RUN: %dxc -Tlib_6_7 -verify %s
+// REQUIRES: dxil-1-8
+
+// Test translatable barriers in compute shader model.
+// These are currently not translated, so we expect errors for now.
+
+struct RECORD { uint a; };
+RWBuffer<uint> buf0;
+static uint i = 7;
+
+// Barriers not requiring visible group, compatible with any shader stage.
+// expected-note@+4 {{entry function defined here}}
+// expected-note@+3 {{entry function defined here}}
+// expected-note@+2 {{entry function defined here}}
+[noinline] export
+void DeviceBarriers() {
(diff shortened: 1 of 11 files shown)


damyanp (approved):
LGTM, can you be sure to associate this with an issue please.

tools/clang/test/SemaHLSL/hlsl/intrinsics/barrier/barrier-cs-sm67-compat.hlsl
```diff
@@ -0,0 +1,103 @@
+// RUN: %dxc -Tlib_6_7 -verify %s
+// REQUIRES: dxil-1-8
```
llvm-beanz, line 2: Is this correct? This is a validator test so it doesn't seem like it should require the 1.8 validator.

tools/clang/test/SemaHLSL/hlsl/intrinsics/barrier/barrier-cs-sm67-errors.hlsl
```diff
@@ -0,0 +1,94 @@
+// RUN: %dxc -Tlib_6_7 -verify %s
+// REQUIRES: dxil-1-8
```
llvm-beanz, line 2: Same question here.

llvm-beanz (commented):
One set of questions below. Otherwise looks good.

tools/clang/test/SemaHLSL/hlsl/intrinsics/barrier/barrier-cs-sm67-compat.hlsl
llvm-beanz: Why do we have any requirement on the validator? The validator doesn't run as part of this test. `-verify` stops at parsing.

llvm-beanz (approved):

tools/clang/test/SemaHLSL/hlsl/intrinsics/barrier/barrier-cs-sm67-compat.hlsl
tex3d: I think these can be resolved, since I've modified the tests as planned.  Re-activate if you think otherwise.

PR #6559, opened 2024-04-23
Disable non-legacy cbuffer flags

### LINKED ISSUE #6306: Deprecate/remove -not_use_legacy_cbuf_load
**Description**
This flag does not work and has never worked. We've got bugs on it, but the correct fix requires a shader model change and significant design and implementation work. We should deprecate and remove the flag in the meantime to not confuse our users further.

**Steps to Reproduce**
Compile any shader that uses cbuffer data with and without the flag and notice nothing changes...

**Environment**
- DXC version: All
- Host Operating System: All

pow2clk:
These flags had some effect on load code generation, but never worked right and were not tested in the simple compiler let alone with drivers. This removes any notion of the flags beyond option processing and produces a warning message that the flag is not supported and will be ignored. The flags are both not listed in help anymore either. A simple OptionsTest verifies that the warning message is produced for either flag and existing tests that used the flag are updated or removed.

Fixes #6306


lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -762,10 +762,13 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
...
-  opts.NotUseLegacyCBufLoad =
+  bool NotUseLegacyCBufLoad =
       Args.hasFlag(OPT_no_legacy_cbuf_layout, OPT_INVALID, false);
-  opts.NotUseLegacyCBufLoad = Args.hasFlag(
-      OPT_not_use_legacy_cbuf_load_, OPT_INVALID, opts.NotUseLegacyCBufLoad);
+  NotUseLegacyCBufLoad = Args.hasFlag(OPT_not_use_legacy_cbuf_load_,
+                                      OPT_INVALID, NotUseLegacyCBufLoad);
+  if (NotUseLegacyCBufLoad)
+    errors << "Warning: -no-legacy-cbuf-layout"
+           << " is no longer supported and will be ignored.\n";
```
llvm-beanz, line 771: Maybe we should also mention that the flag will be removed in a future release? That way we can remove it and it can become an error in the future.

lib/DxcSupport/HLSLOptions.cpp
```diff
@@ -762,10 +762,13 @@ int ReadDxcOpts(const OptTable *optionTable, unsigned flagsToInclude,
...
   opts.Link = Args.hasFlag(OPT_link, OPT_INVALID, false);
-  opts.NotUseLegacyCBufLoad =
+  bool NotUseLegacyCBufLoad =
       Args.hasFlag(OPT_no_legacy_cbuf_layout, OPT_INVALID, false);
-  opts.NotUseLegacyCBufLoad = Args.hasFlag(
-      OPT_not_use_legacy_cbuf_load_, OPT_INVALID, opts.NotUseLegacyCBufLoad);
+  NotUseLegacyCBufLoad = Args.hasFlag(OPT_not_use_legacy_cbuf_load_,
+                                      OPT_INVALID, NotUseLegacyCBufLoad);
+  if (NotUseLegacyCBufLoad)
+    errors << "Warning: -no-legacy-cbuf-layout"
```
llvm-beanz, line 770: nit: clang prints warning in lowercase.
```suggestion
    errors << "warning: -no-legacy-cbuf-layout"
```

llvm-beanz (commented):
We should add a test that verifies the new warning text when the flag is passed. Otherwise this looks good with some nits.

pow2clk:
@llvm-beanz the optionstest verifies the warning when the flag is provided. Do you want an end-to-end lit test that does the same?

lib/DxcSupport/HLSLOptions.cpp
pow2clk: I think the most accurate way to phrase how much more it will be released is "Future releases will not recognize it". Does that sound alright?

damyanp (approved):

llvm-beanz:
> @llvm-beanz the optionstest verifies the warning when the flag is provided. Do you want an end-to-end lit test that does the same?

Ah, sorry I didn't notice that. I have a personal preference for LIT-driven tests because they are simpler to read, but the unit test is also fine.

pow2clk:
> Ah, sorry I didn't notice that. I have a personal preference for LIT-driven tests because they are simpler to read, but the unit test is also fine.

Well I added one anyway. 🙃

PR #6572, opened 2024-04-30
[202x] Implementaiton of conforming literals

### LINKED ISSUE #6537: [0017] Implement conforming literals proposal
**Overview**
Implement HLSL 202x conforming literals feature. 
Full documentation for the feature proposal is [here](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0017-conforming-literals.md).

**Additional Context**
An experimental implementation is available [here](https://github.com/llvm-beanz/DirectXShaderCompiler/tree/cbieneman/conforming-literals). This implementation exposed the new behavior in all language modes for testing of the impact on existing shaders. The implementation in DXC should gate on HLSL 202x.

llvm-beanz:
This change implements the conforming literals proposal as documented in proposal 0017:

https://github.com/microsoft/hlsl-specs/blob/main/proposals/0017-conform ing-literals.md

Fixes #6537

### DIFF
--- a/tools/clang/lib/Sema/SemaExpr.cpp
+++ b/tools/clang/lib/Sema/SemaExpr.cpp
@@ -3458,7 +3458,8 @@ ExprResult Sema::ActOnNumericConstant(const Token &Tok, Scope *UDLScope) {
     else if (Literal.isLong)
       Ty = Context.DoubleTy;
     else if (Literal.isHalf) {
-      if (getLangOpts().HLSLVersion < hlsl::LangStd::v202x)
+      if (getLangOpts().HLSL &&
+          getLangOpts().HLSLVersion < hlsl::LangStd::v202x)
         Ty = getLangOpts().UseMinPrecision ? Context.FloatTy : Context.HalfTy;
       else
         Ty = getLangOpts().UseMinPrecision ? Context.HalfFloatTy
@@ -3487,7 +3488,8 @@ ExprResult Sema::ActOnNumericConstant(const Token &Tok, Scope *UDLScope) {
     return ExprError();
 
   // HLSL Change Starts
-  } else if (getLangOpts().HLSLVersion < hlsl::LangStd::v202x) {
+  } else if (getLangOpts().HLSL &&
+             getLangOpts().HLSLVersion < hlsl::LangStd::v202x) {
     QualType Ty;
     unsigned Width = 64;
     llvm::APInt ResultVal(Width, 0);
--- a/tools/clang/lib/Sema/SemaExpr.cpp
+++ b/tools/clang/lib/Sema/SemaExpr.cpp
@@ -3257,7 +3257,7 @@ ExprResult Sema::ActOnIntegerConstant(SourceLocation Loc, uint64_t Val) {
   unsigned IntSize = Context.getTargetInfo().getIntWidth();
   // HLSL Change Starts - HLSL literal int
   QualType Ty;
-  if (getLangOpts().HLSL) {
+  if (getLangOpts().HLSL && getLangOpts().HLSLVersion < hlsl::LangStd::v202x) {
     IntSize = 64;
     Ty = Context.LitIntTy;
   } else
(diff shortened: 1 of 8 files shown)


tools/clang/lib/Sema/SemaExpr.cpp
```diff
@@ -3519,8 +3525,9 @@ ExprResult Sema::ActOnNumericConstant(const Token &Tok, Scope *UDLScope) {
   } else {
     QualType Ty;
 
+    // HLSL Chagne: disable the warning below.
```
damyanp, line 3528: typo

```suggestion
    // HLSL Change: disable the warning below.
```

tools/clang/test/SemaHLSL/v202x/conforming-literals/valid-literals.hlsl
```diff
@@ -0,0 +1,247 @@
...
+
+_Static_assert(is_same<__decltype(0), int>::value, "0 is int");
+_Static_assert(is_same<__decltype(1), int>::value, "1 is int");
+
+// Decimal literals are always signed.
+_Static_assert(is_same<__decltype(2147483647), int>::value, "2147483647 is int");
+_Static_assert(is_same<__decltype(2147483648), int64_t>::value, "2147483648 is int64_t");
+_Static_assert(is_same<__decltype(4294967296), int64_t>::value, "4294967296 is int64_t");
+
+// This is an anomoly that exists in C as well as HLSL. This value can't be
```
damyanp, line 43:
```suggestion
// This is an anomaly that exists in C as well as HLSL. This value can't be
```

tools/clang/test/SemaHLSL/v202x/conforming-literals/valid-literals.hlsl
```diff
@@ -0,0 +1,247 @@
...
+// represented as a signed integer, but base-10 literals are always signed.
+// Clang emits a wanring that it is interpreting it as unsigned because that is
```
damyanp, line 45:
```suggestion
// Clang emits a warning that it is interpreting it as unsigned because that is
```

tools/clang/lib/Sema/SemaExpr.cpp
coopp, line 3528: nit: "Chagne"?  Sounds exotic! :)

tools/clang/test/SemaHLSL/v202x/conforming-literals/valid-literals.hlsl
```diff
@@ -0,0 +1,247 @@
...
+
+template <typename T>
+struct is_same<T, T> {
+  static const bool value = true;
+};
+
+bool B; // Used for ternary operator tests below
+
+#if !defined(__HLSL_ENABLE_16_BIT)
+min16int i16;
```
damyanp, line 23: Would these tests all still pass if `-enable-16bit-types` was getting ignored?

damyanp (approved):
A few typos and a question about the tests, but otherwise LGTM.

tools/clang/test/SemaHLSL/v202x/conforming-literals/valid-literals.hlsl
```diff
@@ -0,0 +1,247 @@
...
+
+_Static_assert(is_same<__decltype(B ? 1l : 1), int64_t>::value, "B ? 1l : 1 is int64_t");
+_Static_assert(is_same<__decltype(B ? 1 : 1l), int64_t>::value, "B ? 1 : 1l is int64_t");
+
+_Static_assert(is_same<__decltype(B ? 1ul : 1), uint64_t>::value, "B ? 1ul : 1 is uint64_t");
+_Static_assert(is_same<__decltype(B ? 1 : 1ul), uint64_t>::value, "B ? 1 : 1ul is uint64_t");
+
+#if !defined(__HLSL_ENABLE_16_BIT)
+_Static_assert(is_same<__decltype(B ? i16 : i16), min16int>::value, "B ? (min16int) : (min16int) is int"); // maybe a bug
+_Static_assert(is_same<__decltype(B ? u16 : u16), min16uint>::value, "B ? (min16uint) : (min16uint) is int"); // maybe a bug
```
tex3d, line 133: What's the bug?  Looks like the text description doesn't match the check.

coopp (approved):
Small typos, but looks good to me.

tools/clang/test/SemaHLSL/v202x/conforming-literals/valid-literals.hlsl
llvm-beanz: I mean... kinda by design. I wrote the test so it covers both when the flag is passed and when it isn't.

If the flag is completely ignored, this test would still pass, but a lot of other tests which test behaviors of that flag specifically would fail.

tools/clang/test/SemaHLSL/v202x/conforming-literals/valid-literals.hlsl
llvm-beanz: Oh! Sorry, that was a note I left for myself that was a bug in my implementation which I fixed. I'll correct the test and remove the comment.

tools/clang/lib/Sema/SemaExpr.cpp
```diff
@@ -3451,13 +3451,19 @@ ExprResult Sema::ActOnNumericConstant(const Token &Tok, Scope *UDLScope) {
 
       Ty = Context.FloatTy;
     // HLSL Change Starts
-    else if (getLangOpts().HLSL && !Literal.isLong && !Literal.isHalf)
+    else if (getLangOpts().HLSLVersion < hlsl::LangStd::v202x &&
```
tex3d, line 3454: Why the removal of `getLangOpts().HLSL` conditions around here?  Just the assumption that it's never false for DXC now?  There are some that retain this condition though.

tex3d: I have a preference to keep it, mainly for consistency with the surrounding code and the existing convention.  But I won't die on that hill 😉.  Part of its purpose is to clearly indicate branches that were changed in HLSL-specific ways, and to attempt to limit those change to when this is set.  We did used to have a config where we could build without it enabled, I think, but it's been a long time since any of that would work, I imagine.

It feels weird to partly change to the assumption that it's always true and doesn't need to be checked for just some small subset of the code.  But I don't have a strong reason why you must restore it, since obviously we don't actually support or test any scenario where it's not true.

Also, if comparing the version, HLSL should be known to be true already, otherwise, the result of `getLangOpts().HLSLVersion < hlsl::LangStd::v202x` wouldn't be meaningful, so I don't see how that factors into whether we omit the HLSL check at all.  Perhaps you mean that the version check indicates that this path is HLSL specific to the reader of the code, though I don't think that applies to every omission of the `getLangOpts().HLSL` check.

PR #6582, opened 2024-05-03
Fix invalid module bitcode when indexing a swizzled bool vector

amaiorano:
When indexing a swizzled bool vector, some HLSL-specific code in EmitCXXMemberOrOperatorMemberCallExpr kicks in to handle the HLSLVecType. In this case, we’re dealing with an ExtVectorElt because of the swizzle, so this function creates a GEP, Load, and Store on the vector. However, boolean scalars are returned as type i11 while the store is storing to a bool, which is an i32, so we need to insert a cast before the store.

### DIFF
--- a/tools/clang/lib/CodeGen/CGExpr.cpp
+++ b/tools/clang/lib/CodeGen/CGExpr.cpp
@@ -1137,6 +1137,11 @@ llvm::MDNode *CodeGenFunction::getRangeForLoadFromType(QualType Ty) {
   return MDHelper.createRange(Min, End);
 }
 
+bool ShouldEmitRangeMD(llvm::Value *Value, QualType Ty) {
+  return hasBooleanRepresentation(Ty) &&
+         cast<llvm::IntegerType>(Value->getType())->getBitWidth() != 1;
+}
+
 llvm::Value *CodeGenFunction::EmitLoadOfScalar(llvm::Value *Addr, bool Volatile,
                                                unsigned Alignment, QualType Ty,
                                                SourceLocation Loc,
@@ -1236,7 +1241,8 @@ llvm::Value *CodeGenFunction::EmitLoadOfScalar(llvm::Value *Addr, bool Volatile,
       EmitCheck(std::make_pair(Check, Kind), "load_invalid_value", StaticArgs,
                 EmitCheckValue(Load));
     }
-  } else if (CGM.getCodeGenOpts().OptimizationLevel > 0)
+  } else if (CGM.getCodeGenOpts().OptimizationLevel > 0 &&
+             ShouldEmitRangeMD(Load, Ty))
     if (llvm::MDNode *RangeInfo = getRangeForLoadFromType(Ty))
       Load->setMetadata(llvm::LLVMContext::MD_range, RangeInfo);
 
--- a/tools/clang/lib/CodeGen/CGExpr.cpp
+++ b/tools/clang/lib/CodeGen/CGExpr.cpp
@@ -1137,9 +1137,10 @@ llvm::MDNode *CodeGenFunction::getRangeForLoadFromType(QualType Ty) {
   return MDHelper.createRange(Min, End);
 }
 
-bool ShouldEmitRangeMD(llvm::Value *Value, QualType Ty) {
-  return hasBooleanRepresentation(Ty) &&
-         cast<llvm::IntegerType>(Value->getType())->getBitWidth() != 1;
+static bool ShouldEmitRangeMD(llvm::Value *Value, QualType Ty) {
+  if (hasBooleanRepresentation(Ty))
+    return cast<llvm::IntegerType>(Value->getType())->getBitWidth() != 1;
+  return true;
 }
 
 llvm::Value *CodeGenFunction::EmitLoadOfScalar(llvm::Value *Addr, bool Volatile,
--- a/tools/clang/lib/CodeGen/CGExprCXX.cpp
+++ b/tools/clang/lib/CodeGen/CGExprCXX.cpp
@@ -235,12 +235,17 @@ RValue CodeGenFunction::EmitCXXMemberOrOperatorMemberCallExpr(
 
           llvm::Constant *zero = Builder.getInt32(0);
           llvm::Value *TmpThis = CreateTempAlloca(Ty);
+          QualType ElTy = hlsl::GetElementTypeOrType(Base->getType());
+          bool IsBool = ElTy->isSpecificBuiltinType(BuiltinType::Bool);
           for (unsigned i = 0; i < Ty->getVectorNumElements(); i++) {
             llvm::Value *EltIdx = Elts->getAggregateElement(i);
             llvm::Value *EltGEP = Builder.CreateGEP(This, {zero, EltIdx});
             llvm::Value *TmpEltGEP =
                 Builder.CreateGEP(TmpThis, {zero, Builder.getInt32(i)});
             llvm::Value *Elt = Builder.CreateLoad(EltGEP);
+            if (IsBool)
+              Elt = Builder.CreateTrunc(
+                  Elt, llvm::Type::getInt1Ty(getLLVMContext()));
             Builder.CreateStore(Elt, TmpEltGEP);
           }
           This = TmpThis;
--- a/tools/clang/test/CodeGenDXIL/operators/swizzle/indexSwizzledBoolVec.hlsl
+++ b/tools/clang/test/CodeGenDXIL/operators/swizzle/indexSwizzledBoolVec.hlsl
@@ -1,0 +1,30 @@
+// Test indexing a swizzled bool vector
+// RUN: %dxc -fcgl -T cs_6_0 %s | FileCheck %s
+
+// This was asserting in Instructions.cpp with:
+// void llvm::StoreInst::AssertOK(): Assertion `getOperand(0)->getType() == cast<PointerType>(getOperand(1)->getType())->getElementType() && "Ptr must be a pointer to Val type!"' failed.
+
+// Make sure load of i32 gets truncated to i1 when indexing bool vectors
+// CHECK:       [[TMP:%[a-z0-9\.]+]] = alloca <2 x i1>
+// CHECK:       [[VA0:%[a-z0-9\.]+]] = getelementptr <2 x i1>, <2 x i1>* [[TMP]], i32 0, i32 0,
+// CHECK-NEXT:  [[VA1:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds (<4 x i32>, <4 x i32>* @"\01?v_bool4@?1??main@@YAXXZ@3V?$vector@_N$03@@B", i32 0, i32 2),
+// CHECK-NEXT:  [[VA2:%[a-z0-9\.]+]] = trunc i32 [[VA1]] to i1,
+// CHECK-NEXT:  store i1 [[VA2]], i1* [[VA0]],
+// CHECK-NEXT:  [[VB0:%[a-z0-9\.]+]] = getelementptr <2 x i1>, <2 x i1>* [[TMP]], i32 0, i32 1,
+// CHECK-NEXT:  [[VB1:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds (<4 x i32>, <4 x i32>* @"\01?v_bool4@?1??main@@YAXXZ@3V?$vector@_N$03@@B", i32 0, i32 3),
+// CHECK-NEXT:  [[VB2:%[a-z0-9\.]+]] = trunc i32 [[VB1]] to i1,
+// CHECK-NEXT:  store i1 [[VB2]], i1* [[VB0]],
+
+
+cbuffer cbuffer_tint_symbol_3 : register(b0) {
+    uint4 global_uint4[1];
+};
+
+[numthreads(1, 1, 1)]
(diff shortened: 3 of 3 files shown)


amaiorano:
See this example of the failure: https://godbolt.org/z/WsajscbT9

tools/clang/test/CodeGenDXIL/operators/swizzle/indexSwizzledBoolVec.hlsl
```diff
@@ -0,0 +1,20 @@
+// Test indexing a swizzled bool vector
+// RUN: %dxc -T cs_6_0 %s | FileCheck %s
+
+// This was asserting in Instructions.cpp with:
+// void llvm::StoreInst::AssertOK(): Assertion `getOperand(0)->getType() == cast<PointerType>(getOperand(1)->getType())->getElementType() && "Ptr must be a pointer to Val type!"' failed.
+
+// CHECK: @main
```
llvm-beanz, line 7: Checking that it doesn't crash isn't a good enough test. We tend to expect that inputs both don't crash _and_ produce valid output.

You should be able to add the `-fcgl` flag to get the pre-optimized IR, and do some checks against that to verify that the IR generation is correct.

tools/clang/lib/CodeGen/CGExprCXX.cpp
```diff
@@ -217,6 +217,27 @@ RValue CodeGenFunction::EmitCXXMemberOrOperatorMemberCallExpr(
...
+        }
+        if (isBool) {
+          // bool pointer is not i1 *.
+          llvm::Type *BoolTy = llvm::IntegerType::get(
+              getLLVMContext(), getContext().getTypeSize(CE->getType()));
+          Val = Builder.CreateBitCast(
+              Val, llvm::PointerType::get(
+                       BoolTy, Val->getType()->getPointerAddressSpace()));
+        }
+        return Val;
```
llvm-beanz, line 238: nit:
```suggestion
        Type *ElTy = This->getType()->getPointerElementType()
                                    ->getVectorElementType();
        if (llvm::IntegerType *IT = dyn_cast<llvm::IntegerType>(ThisTy))
          if (IT->getBitWidth() != 1)
            return Val;

        // bool pointer is not i1 *.
        llvm::Type *BoolTy = llvm::IntegerType::get(
            getLLVMContext(), getContext().getTypeSize(CE->getType()));
        Val = Builder.CreateBitCast(
            Val, llvm::PointerType::get(
                     BoolTy, Val->getType()->getPointerAddressSpace()));
        return Val;
```

amaiorano: My original patch actually looked like this, but I wanted to make it clear that I had simply refactored the existing code mostly as-is. Will apply this change.

tools/clang/test/CodeGenDXIL/operators/swizzle/indexSwizzledBoolVec.hlsl
amaiorano: That's fair. I had based this on other examples I saw of hlsl tests that seem to validate that dxc doesn't crash. But a better test is a good thing, so I'll do this.

tools/clang/lib/CodeGen/CGExprCXX.cpp
amaiorano: Uploaded new patch based on our conversation.

tools/clang/test/CodeGenDXIL/operators/swizzle/indexSwizzledBoolVec.hlsl
amaiorano: Done.

tools/clang/lib/CodeGen/CGExpr.cpp
```diff
@@ -1236,7 +1241,8 @@ llvm::Value *CodeGenFunction::EmitLoadOfScalar(llvm::Value *Addr, bool Volatile,
       EmitCheck(std::make_pair(Check, Kind), "load_invalid_value", StaticArgs,
                 EmitCheckValue(Load));
     }
-  } else if (CGM.getCodeGenOpts().OptimizationLevel > 0)
+  } else if (CGM.getCodeGenOpts().OptimizationLevel > 0 &&
+             ShouldEmitRangeMD(Load, Ty))
```
llvm-beanz, line 1245: Just as a comment for any additional reviewers. This matches what modern clang does. Basically we can skip inserting value range metadata for bools that are represented as i1 because the valid range of a bool is the full range of an i1.

llvm-beanz (commented):
I think this looks good, but since I contributed a bit to it I want to get some additional eyes on it.

farzonl:
LGTM

farzonl (approved):

tools/clang/lib/CodeGen/CGExpr.cpp
```diff
@@ -1137,6 +1137,11 @@ llvm::MDNode *CodeGenFunction::getRangeForLoadFromType(QualType Ty) {
   return MDHelper.createRange(Min, End);
 }
 
+bool ShouldEmitRangeMD(llvm::Value *Value, QualType Ty) {
```
tex3d, line 1140: Shouldn't this be static?  Plus the name is generic, but seems to only apply to bool?  Is that correct?

tools/clang/lib/CodeGen/CGExpr.cpp
tex3d: Is bool the only type we would want to emit range information for?  `ShouldEmitRangeMD` won't return true unless hasBooleanRepresentation(Ty) returns true AND has bit size > 1.  It isn't obvious to me that this is correct.

tools/clang/lib/CodeGen/CGExprCXX.cpp
```diff
@@ -235,12 +235,17 @@ RValue CodeGenFunction::EmitCXXMemberOrOperatorMemberCallExpr(
...
+          bool IsBool = ElTy->isSpecificBuiltinType(BuiltinType::Bool);
           for (unsigned i = 0; i < Ty->getVectorNumElements(); i++) {
             llvm::Value *EltIdx = Elts->getAggregateElement(i);
             llvm::Value *EltGEP = Builder.CreateGEP(This, {zero, EltIdx});
             llvm::Value *TmpEltGEP =
                 Builder.CreateGEP(TmpThis, {zero, Builder.getInt32(i)});
             llvm::Value *Elt = Builder.CreateLoad(EltGEP);
+            if (IsBool)
+              Elt = Builder.CreateTrunc(
+                  Elt, llvm::Type::getInt1Ty(getLLVMContext()));
```
tex3d, line 248: Will this impact conversion of bool that could come from host memory?  Trunc is only valid if you know that the source is already a bit pattern constrained by the compiler.  Otherwise, we must use comparison with zero (not-equal) to translate from i32 in storage to i1.

hekota (approved):
LGTM!

tools/clang/test/CodeGenDXIL/operators/swizzle/indexSwizzledBoolVec.hlsl
```diff
@@ -0,0 +1,30 @@
...
+// RUN: %dxc -fcgl -T cs_6_0 %s | FileCheck %s
+
+// This was asserting in Instructions.cpp with:
+// void llvm::StoreInst::AssertOK(): Assertion `getOperand(0)->getType() == cast<PointerType>(getOperand(1)->getType())->getElementType() && "Ptr must be a pointer to Val type!"' failed.
+
+// Make sure load of i32 gets truncated to i1 when indexing bool vectors
+// CHECK:       [[TMP:%[a-z0-9\.]+]] = alloca <2 x i1>
+// CHECK:       [[VA0:%[a-z0-9\.]+]] = getelementptr <2 x i1>, <2 x i1>* [[TMP]], i32 0, i32 0,
+// CHECK-NEXT:  [[VA1:%[a-z0-9\.]+]] = load i32, i32* getelementptr inbounds (<4 x i32>, <4 x i32>* @"\01?v_bool4@?1??main@@YAXXZ@3V?$vector@_N$03@@B", i32 0, i32 2),
+// CHECK-NEXT:  [[VA2:%[a-z0-9\.]+]] = trunc i32 [[VA1]] to i1,
```
tex3d, line 11: `trunc` is ok if original bool was not host-visible and guaranteed to be a known bit representation, but what if this `v_bool4` array is in host-visible memory?  In that case we have to convert with comparisons against zero.  Are we sure that path works?  We should have a test for that case, I think.

tools/clang/lib/CodeGen/CGExpr.cpp
llvm-beanz: Good catch!

tools/clang/lib/CodeGen/CGExpr.cpp
```diff
@@ -1137,6 +1137,11 @@ llvm::MDNode *CodeGenFunction::getRangeForLoadFromType(QualType Ty) {
...
+  return hasBooleanRepresentation(Ty) &&
+         cast<llvm::IntegerType>(Value->getType())->getBitWidth() != 1;
```
llvm-beanz, line 1142:
```suggestion
  if(hasBooleanRepresentation(Ty))
    return cast<llvm::IntegerType>(Value->getType())->getBitWidth() != 1;
  return true;
```

tools/clang/lib/CodeGen/CGExpr.cpp
llvm-beanz:
```suggestion
static bool ShouldEmitRangeMD(llvm::Value *Value, QualType Ty) {
```

tools/clang/lib/CodeGen/CGExpr.cpp
llvm-beanz: That’s a bug in the code I gave @amaiorano. I’ve posted suggestions to correct.

tools/clang/lib/CodeGen/CGExprCXX.cpp
llvm-beanz: It is undefined behavior in C/C++ and HLSL for bool to have any value other than 0 or 1. Given that, trunc is valid in all cases converting between lengths of bool.

amaiorano: I've applied Chris' fixes based on your comments @tex3d. I'm waiting for your approval before merging.

llvm-beanz: To fully cite the spec. This is the wording in the C++ spec:

> Values of type bool are either true or false.
[HTML Citation](https://timsong-cpp.github.io/cppwp/n4140/basic.fundamental#6)

Since HLSL defines `true` to `1` and `false` to `0` (the same as pretty much all C/C++ compilers), any different value is UB.

tex3d: While I can't find documentation to back this up now, HLSL compilers (FXC and DXC) have always required a translation for bools coming from host visible memory (cbuffer, etc...) by interpreting any non-zero value as true, and zero value as false.  It's something we did in FXC, and had to do in DXC to match FXC.  Trunc will convert any bit pattern with the lowest bit unset to `0`.  This would be inconsistent with the way bools coming from host memory have been handled thus far for HLSL.

tex3d: For one thing, DXBC instructions are defined in a way that an incoming bool value from a constant buffer may be used in an `if` instruction, and the `if` instruction interprets the 32-bit value with any bit set as `true`, and no bits set as `false`.
See the instruction spec [here](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#22.7.1%20Branch%20based%20on%20boolean%20condition:%20if_condition).

tex3d: Though, as luck would have it, under `Restrictions:` `2)` it gets `_z` and `_nz` modifiers exactly backwards.  The earlier description of the operation gets it right though.

llvm-beanz: I don’t see how the behavior of the DXBC instruction is relevant here. This is absolutely not a behavior you can rely on LLVM to match. Clang (and by extension DXC) have been assuming that the only valid values for bool are 0 and 1 (see the bit where this PR changes the range check).

If a bool value comes in from somewhere and it is anything other than 0 or 1, it’s going to break in a lot of places. The advantage to using a `trunc` here is that it is _way_ easier for LLVM to optimize and eliminate unnecessary `trunc`, `zext` and `sext` instructions because the don’t always depend on value analysis.

If you feel strongly enough about this to hold up the PR we should have a meeting to discuss.

llvm-beanz: We had a lengthy discussion internally and I think we all went back and forth about different thoughts here. In practice since the existing code causes a crash, and this causes correct behavior where the only possible error cases are very firmly in the UB, I think we should take the change as-is.

If @tex3d feels strongly about this we can discuss this more and revisit.

There are a lot of broken behaviors around HLSL boolean vectors that we're going to need to revisit and reconsider in our Clang implementation and we can discuss longer term solutions in that context.

llvm-beanz (approved):

PR #6589, opened 2024-05-06
Retain alignment when lowering groupshared matrices

### LINKED ISSUE #6416: dxc generates invalid alignment on groupshared matrix load/store instructions
**Description**
When loading or storing a matrix to `groupshared` dxc is using an alignment on the instruction that is not guaranteed by the underlying global variable.

**Steps to Reproduce**
https://godbolt.org/z/9c9os31MW

```
// dxc /Tcs_6_0 alignbug.hlsl
struct Data {
   float4x4 m;
};

groupshared Data GData;
StructuredBuffer<Data> input : register(t0);
RWStructuredBuffer<Data> output : register(u0);

[RootSignature("SRV(t0), UAV(u0)")]
[numthreads(128,1,1)]
void main(uint Id : SV_DispatchThreadId, uint g : SV_GroupID)
{
   GData = input[0];
   GroupMemoryBarrierWithGroupSync();
   output[Id] = GData;

}
```

**Actual Behavior**
Compiling this shader produces loads and stores that do not match the alignment of the underlying global.

The global does not specify an alignment so it will get the default alignment of 4 for this type
```
@"\01?GData@@3UData@@A.0.v.v" = addrspace(3) global [16 x float] undef
```

But the load
```
%22 = load float, float addrspace(3)* getelementptr inbounds ([16 x float], [16 x float] addrspace(3)* @"\01?GData@@3UData@@A.0.v.v", i32 0, i32 0), align 16
```

and store
```
store float %3, float addrspace(3)* getelementptr inbounds ([16 x float], [16 x float] addrspace(3)* @"\01?GData@@3UData@@A.0.v.v", i32 0, i32 0), align 16
```
both specify an alignment of `align 16` for the first element (other elements are overaligned as well).

From the  [llvm language reference manual](https://llvm.org/docs/LangRef.html#load-instruction), the aligment on the instruction should be guaranteed by somehow, but the alignment of the global does not do it.

> The optional constant align argument specifies the alignment of the operation (that is, the alignment of the memory address). It is the responsibility of the code emitter to ensure that the alignment information is correct.

**Environment**
- DXC version dxc main as of 3/14/2024
- Host Operating System windows

pow2clk:
When flattening the global for a groupshared matrix, the alignment information was getting lost. As a result, the alignments of the loads and stores were calculating their own alignment based on preferred alignment and trailing zeros of the index. The preferred alignment switched to 16 when the type size was over 128 bits due to a heuristic whose rationale is lost to time. When the global has its own alignment, that gets used, so by retaining it through lowering, the alignments are consistent and reliable.

Includes testing for a few matrix variants

fixes #6416

### DIFF
--- a/tools/clang/test/HLSLFileCheck/hlsl/types/modifiers/groupshared/groupshared-member-matrix-subscript-align.hlsl
+++ b/tools/clang/test/HLSLFileCheck/hlsl/types/modifiers/groupshared/groupshared-member-matrix-subscript-align.hlsl
@@ -1,15 +1,28 @@
-// RUN: %dxc -DTYPE=Data /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK
-// RUN: %dxc -DTYPE=float4x4 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK
-// RUN: %dxc -DTYPE=float4x1 /Tcs_6_0 %s | FileCheck %s
-// RUN: %dxc -DTYPE=float1x4 /Tcs_6_0 %s | FileCheck %s
-// RUN: %dxc -DTYPE=float2x2 /Tcs_6_0 %s | FileCheck %s
+// RUN: %dxc -DSTYPE=float4x4 -DTYPE=Data /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECK16,LCHECKF
+// RUN: %dxc -DSTYPE=double4x4 -DTYPE=Data /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECKD16,LCHECKF
+// RUN: %dxc -DSTYPE=float16_t4x4 -DTYPE=Data /Tcs_6_2 %s -enable-16bit-types | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECKH16,LCHECKH
+// RUN: %dxc -DTYPE=float4x4 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECK4,LCHECKF
+// RUN: %dxc -DTYPE=float4x1 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,CHECK4
+// RUN: %dxc -DTYPE=float1x4 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,CHECK4
+// RUN: %dxc -DTYPE=float2x2 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,CHECK4
+// RUN: %dxc -DTYPE=double4x4 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECK8,LCHECKF
+// RUN: %dxc -DTYPE=float16_t4x4 /Tcs_6_2 %s -enable-16bit-types | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECK2,LCHECKF
+
+#ifndef STYPE
+#define STYPE float4x4
+#endif
 
 struct Data {
-   float4x4 m;
+   STYPE m;
 };
 
 // Ensure that the global has an alignment
-//CHECK: @"[[GV:\\01\?GData@@.*]]" = addrspace(3) global [[DIM:\[[0-9]+]] x float] undef, align 4
+//CHECK16: @"[[GV:\\01\?GData@@.*]]" = addrspace(3) global [[DIM:\[[0-9]+]] x [[ELTY:float]]] undef, align [[ALIGN:16]]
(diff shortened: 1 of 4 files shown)


tools/clang/test/HLSLFileCheck/hlsl/types/modifiers/groupshared/groupshared-member-matrix-subscript-align.hlsl
```diff
@@ -0,0 +1,59 @@
+// RUN: %dxc -DTYPE=Data /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK
+// RUN: %dxc -DTYPE=float4x4 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK
+// RUN: %dxc -DTYPE=float4x1 /Tcs_6_0 %s | FileCheck %s
+// RUN: %dxc -DTYPE=float1x4 /Tcs_6_0 %s | FileCheck %s
+// RUN: %dxc -DTYPE=float2x2 /Tcs_6_0 %s | FileCheck %s
```
dmpots, line 5: Should we include tests for 16-bit and 64-bit types as well?

tools/clang/test/HLSLFileCheck/hlsl/types/modifiers/groupshared/groupshared-member-matrix-subscript-align.hlsl
llvm-beanz, line 1: Since this is a change in an IR pass can we write this as an IR test that verifies the specific change?

dmpots (commented):
There is a possibility of a perf regression with this change as it may prevent generating larger loads in some backends. Should we set the alignment to what it was assumed to be previously?

pow2clk:
> There is a possibility of a perf regression with this change as it may prevent generating larger loads in some backends. Should we set the alignment to what it was assumed to be previously?

It's certainly an option. What I have here preserves the alignment that is present from the start. I'm a bit more nervous about interjecting the alignment that the zero index happened to get when loads and stores were generated from the beginning. I fear that might have more unforeseen consequences than this does, but I can experiment with it.

dmpots:
> > There is a possibility of a perf regression with this change as it may prevent generating larger loads in some backends. Should we set the alignment to what it was assumed to be previously?
> 
> It's certainly an option. What I have here preserves the alignment that is present from the start. I'm a bit more nervous about interjecting the alignment that the zero index happened to get when loads and stores were generated from the beginning. I fear that might have more unforeseen consequences than this does, but I can experiment with it.

In the updated PR we are using the datalayout to compute the preferred alignment. This looks like the correct way to do it to me.

dmpots (approved):

llvm-beanz (approved):

tools/clang/test/HLSLFileCheck/hlsl/types/modifiers/groupshared/groupshared-member-matrix-subscript-align.hlsl
```diff
@@ -0,0 +1,86 @@
+// RUN: %dxc -DSTYPE=float4x4 -DTYPE=Data /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECK16,LCHECKF
+// RUN: %dxc -DSTYPE=double4x4 -DTYPE=Data /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECKD16,LCHECKF
+// RUN: %dxc -DSTYPE=float16_t4x4 -DTYPE=Data /Tcs_6_2 %s -enable-16bit-types | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECKH16,LCHECKH
+// RUN: %dxc -DTYPE=float4x4 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECK4,LCHECKF
+// RUN: %dxc -DTYPE=float4x1 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,CHECK4
+// RUN: %dxc -DTYPE=float1x4 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,CHECK4
+// RUN: %dxc -DTYPE=float2x2 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,CHECK4
+// RUN: %dxc -DTYPE=double4x4 /Tcs_6_0 %s | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECK8,LCHECKF
+// RUN: %dxc -DTYPE=float16_t4x4 /Tcs_6_2 %s -enable-16bit-types | FileCheck %s -check-prefixes=CHECK,LCHECK,CHECK2,LCHECKF
```
tex3d, line 9: I think it might be important to also test with a matrix array, non-zero index, and odd matrix sizes, like (3x3, 1x3, 3x1).  Since the alignment of the matrix is considered to be the alignment of the element, different array elements for a matrix whose size isn't an even multiple of 16 bytes won't be aligned by that amount.

PR #6592, opened 2024-05-06
PIX: Correct the disambiguation of AS+MS threads for mesh shader output

jeffnn:
PIX requires that all vertex information generated by these passes be uniquely identified by vertex id and MS thread id.
This change fixes the MS thread id part in two places: the amplification shader and the mesh shader.
To be unique across an entire DispatchMesh call, we must uniquify the AS thread group, the AS thread, the MS thread group and the MS thread. This is a lot of multiplying and adding, and there wasn't quite enough math going on here before.
In the AS case, we now generate a unique "flat" thread id from the flat-thread-id-in-group (the already-available system value) and the "flat group id", which we synthesize by multiplying together the group id components with the DispatchMesh API's thread group counts, and then multiplying that by the number of threads each AS group launches, then add the flat-thread-id-in-group. (This flat id then goes into an expanded version of the AS->MS payload, the code for which was pre-existing.)
The MS will either treat the incoming AS thread id as its unique thread-group-within-the-whole-dispatch id.
If the AS is not active, the instrumentation herein will synthesize a flat id in the same way as the AS did before it passed that id through the payload, again from the DispatchMesh parameters (newly-added params to that pass) and the flat-thread-in-group.

In addition to the new filecheck tests for this, there is also a new filecheck test to cover coercion of non-i32 types to i32 before being written to PIX's output UAV, which I happened to notice wasn't adequately tested.

### DIFF
--- a/lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
+++ b/lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
@@ -117,29 +117,51 @@ bool DxilPIXAddTidToAmplificationShaderPayload::runOnModule(Module &M) {
 
     llvm::IRBuilder<> B(UserInstruction);
 
-    auto ThreadIdFunc =
-        HlslOP->GetOpFunc(DXIL::OpCode::ThreadId, Type::getInt32Ty(Ctx));
-    Constant *Opcode = HlslOP->GetU32Const((unsigned)DXIL::OpCode::ThreadId);
     Constant *Zero32Arg = HlslOP->GetU32Const(0);
     Constant *One32Arg = HlslOP->GetU32Const(1);
     Constant *Two32Arg = HlslOP->GetU32Const(2);
 
-    auto ThreadIdX =
-        B.CreateCall(ThreadIdFunc, {Opcode, Zero32Arg}, "ThreadIdX");
-    auto ThreadIdY =
-        B.CreateCall(ThreadIdFunc, {Opcode, One32Arg}, "ThreadIdY");
-    auto ThreadIdZ =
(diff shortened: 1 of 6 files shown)


damyanp (approved):

lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
```diff
@@ -117,29 +117,51 @@ bool DxilPIXAddTidToAmplificationShaderPayload::runOnModule(Module &M) {
...
+    Constant *GroupIdOpcode =
+        HlslOP->GetU32Const((unsigned)DXIL::OpCode::GroupId);
+    auto *GroupIdX =
+        B.CreateCall(GroupIdFunc, {GroupIdOpcode, Zero32Arg}, "GroupIdX");
+    auto *GroupIdY =
+        B.CreateCall(GroupIdFunc, {GroupIdOpcode, One32Arg}, "GroupIdY");
+    auto *GroupIdZ =
+        B.CreateCall(GroupIdFunc, {GroupIdOpcode, Two32Arg}, "GroupIdZ");
+
+    auto *GroupYxNumZ = B.CreateMul(
```
tex3d, line 135: Following this code could benefit from a comment with the formula being constructed.
I would have expected something like: `gid.X + gid.Y * DimX + gid.Z * (DimX * DimY)`
But this looks like: `gid.X * (DimY * DimZ) + gid.Z + gid.Y * DimZ`
That just makes it flatten in a different way.  Instead of minor-to-major dims being X, Y, Z, they are: Z, Y, X.

Is that the intent?

jeffnn: That is the intent. Yes, this definitely needs some comments. Added. Thanks!

coopp (approved):
Looks good to me.  Glad to see the tests too!

PR #6596, opened 2024-05-07
[NFC][SPIR-V] Refactor SpirvGroupNonUniformOps

sudonatalie:
A follow-up change will use the PartitionedExclusiveScanNV GroupOperation, which requires that an additional operand is added to all GroupNonUniformArithmetic instructions. This means that some of the SPIR-V opcodes which are currently categorized as unary will become either unary or binary depending on the GroupOp. Since the arity distinctions between the OpGroupNonUniform* instructions were already somewhat arbitrary, I'm prefacing that change by refactoring them into a single SpirvGroupNonUniformOp instruction type for better reusability.

Follow up: #6608


Keenuts (approved):

tools/clang/include/clang/SPIRV/SpirvInstruction.h
```diff
@@ -1495,102 +1491,43 @@ class SpirvFunctionCall : public SpirvInstruction {
...
   void replaceOperand(
       llvm::function_ref<SpirvInstruction *(SpirvInstruction *)> remapOp,
       bool inEntryFunctionWrapper) override {
-    arg = remapOp(arg);
-    if (inEntryFunctionWrapper)
-      setAstResultType(arg->getAstResultType());
+    for (auto operand : getOperands()) {
+      operand = remapOp(operand);
+      if (inEntryFunctionWrapper)
+        setAstResultType(operand->getAstResultType());
```
amaiorano, line 1524: This will `setAstResultType` for each operand, where formerly it was only doing so for the one arg (for `SpirvNonUniformUnaryOp`). Is this okay?

tools/clang/lib/SPIRV/LiteralTypeVisitor.cpp
```diff
@@ -294,17 +294,9 @@ bool LiteralTypeVisitor::visit(SpirvVectorShuffle *inst) {
...
-  tryToUpdateInstLitType(inst->getArg(), inst->getAstResultType());
-  return true;
-}
-
-bool LiteralTypeVisitor::visit(SpirvNonUniformBinaryOp *inst) {
-  // Went through each non-uniform unary operation and made sure the following
-  // does not result in a wrong type deduction.
-  tryToUpdateInstLitType(inst->getArg1(), inst->getAstResultType());
+bool LiteralTypeVisitor::visit(SpirvGroupNonUniformOp *inst) {
+  for (auto *operand : inst->getOperands())
```
amaiorano, line 298: nit: In some of the chages above, we use `auto operand` while here it's `auto *operand`.  Should we be consistent?

amaiorano, line 298: In the former `SpirvNonUniformBinaryOp`, only the 2nd arg (`inst->getArg1()`) was used as the argument to `tryToUpdateInstLitType`. Is the idea to use the last argument, rather than iterate through all of them?

tools/clang/lib/SPIRV/PreciseVisitor.cpp
```diff
@@ -233,14 +233,10 @@ bool PreciseVisitor::visit(SpirvUnaryOp *inst) {
...
-bool PreciseVisitor::visit(SpirvNonUniformBinaryOp *inst) {
-  inst->getArg1()->setPrecise(inst->isPrecise());
-  inst->getArg2()->setPrecise(inst->isPrecise());
-  return true;
-}
-
-bool PreciseVisitor::visit(SpirvNonUniformUnaryOp *inst) {
-  inst->getArg()->setPrecise(inst->isPrecise());
+bool PreciseVisitor::visit(SpirvGroupNonUniformOp *inst) {
+  if (inst->isPrecise())
```
amaiorano, line 237: Comparing to the old code, shouldn't this function always call `setPrecise` on each operand, and set the value to `inst->isPrecise()`? In this case, if `inst->isPrecise()` returns false, and any of operands current have precise set to `true`, they will not be updated. In other words, shouldn't this be:

```
  for (auto *operand : inst->getOperands())
    operand->setPrecise(inst->isPrecise());
  return true;
```

tools/clang/lib/SPIRV/SpirvInstruction.cpp
```diff
@@ -662,65 +660,66 @@ SpirvFunctionCall::SpirvFunctionCall(QualType resultType, SourceLocation loc,
...
-         op == spv::Op::OpGroupNonUniformBitwiseXor ||
-         op == spv::Op::OpGroupNonUniformLogicalAnd ||
-         op == spv::Op::OpGroupNonUniformLogicalOr ||
-         op == spv::Op::OpGroupNonUniformLogicalXor);
+SpirvGroupNonUniformOp::SpirvGroupNonUniformOp(
+    spv::Op op, QualType resultType, spv::Scope scope,
+    llvm::ArrayRef<SpirvInstruction *> operandsVec, SourceLocation loc,
+    llvm::Optional<spv::GroupOperation> group)
+    : SpirvInstruction(IK_GroupNonUniformOp, op, resultType, loc),
+      execScope(scope), operands(operandsVec.begin(), operandsVec.end()),
```
amaiorano, line 668: Given that the constructor always copies the operands, would it make more sense to make the parameter a `llvm::SmallVector<SpirvInstruction *, 4>` and use `std::move` to move the args into the `operands` data member?

Actualy, after some thought, I suspect the answer is that it doesn't really matter, since the number of args is expected to be small, so we won't be heap allocating, and thus never benefit from moving. Just leaving this comment here in case you can teach me more about the DXC codebase.

amaiorano (changes requested):
I've posted a number of comments.

tools/clang/lib/SPIRV/LiteralTypeVisitor.cpp
sudonatalie: This is another bit of cargo culting of untested code that I think was either buggy or possibly totally unnecessary. I will try to come up with some HLSL to trigger it and include a test...

tools/clang/lib/SPIRV/PreciseVisitor.cpp
sudonatalie: It's a default-false flag that get propagated down by this visitor when it's true, so either should be equivalent but switched just to be closer to the original.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -9544,14 +9545,13 @@ SpirvEmitter::processWaveCountBits(const CallExpr *callExpr,
...
+      spv::Op::OpGroupNonUniformBallot, v4u32Type, spv::Scope::Subgroup,
+      {predicate}, srcLoc);
 
-  return spvBuilder.createGroupNonUniformUnaryOp(
-      srcLoc, spv::Op::OpGroupNonUniformBallotBitCount, retType,
-      spv::Scope::Subgroup, ballot,
-      llvm::Optional<spv::GroupOperation>(groupOp));
+  return spvBuilder.createGroupNonUniformOp(
+      spv::Op::OpGroupNonUniformBallotBitCount, retType, spv::Scope::Subgroup,
+      {ballot}, srcLoc, llvm::Optional<spv::GroupOperation>(groupOp));
```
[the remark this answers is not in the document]
sudonatalie: It does, done.

tools/clang/lib/SPIRV/SpirvInstruction.cpp
sudonatalie: Since SmallVector avoids heap allocations and is `std::is_trivially_copyable`, I think you're right that there's no benefit to be gained unless the size is sometimes >4, which it never should be in this case. Some more info here: https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h

sudonatalie (commented):
I will try to come up with a couple test cases that exercise the LiteralTypeVisitor and PervertexInputVisitor code because the necessity/correctness of both before and after code is still a bit unclear to me.

tools/clang/include/clang/SPIRV/SpirvInstruction.h
sudonatalie: Update: This is the bit I was referring to as the "PervertexInputVisitor" code (the only place this function is called from), because yeah, this isn't a great answer.

amaiorano: It's fine, I don't want to hold up the change. Just wanted to call your attention to this, in case it's a problem.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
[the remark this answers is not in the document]
amaiorano: I think there are a few more locations where this could be applied, but will not hold up the change for this. Can be done in follow up.

amaiorano (approved):
LGTM. You can decide to add more tests if you wish.

sudonatalie:
> LGTM. You can decide to add more tests if you wish.

Thanks! I couldn't trigger these code paths easily, so going ahead with it as-is since I think it's not in a better or worse state, and makes the follow up change cleaner.

PR #6597, opened 2024-05-07
Add compiler option -Qsource_in_debug_module for embedding source in debug builds to documented help text

### LINKED ISSUE #6028: Compiler option for embedding source in debug builds not documented
**Description**
The compiler option `-Qsource_in_debug_module` is not documented in `-help` where users are likely to look for it when trying to use debug info as generated by `-Zi`.

**Steps to Reproduce**
When building a module with separate debug info using `-Zi -Fd ...`, if `-Qsource_in_debug_module` is not specified then the debug info module doesn't have any source for the hlsl files in the debug metadata.

This flag is only needed if separate debug info is used - if the debug info is embedded in the module then the source is there as expected. The flag isn't documented anywhere though as far as I can tell, at least it's not mentioned in -help along with the other debug options - I'd expect it near `-Qembed_debug`.

```
$ dxc -Od -Zi a.hlsl -Qembed_debug -Fo a.dxbc -T cs_6_0
$ dxc -dumpbin a.dxbc | grep 'void main'
!38 = !{!"a.hlsl", !"RWByteAddressBuffer a : register(u0);\0D\0A\0D\0A[numthreads(1,1,1)]\0D\0Avoid main() {\0D\0A\09a.Store(0, uint(4));\0D\0A\09a.Store(128, float(4));\0D\0A\09a.Store(256, float4(4,5,6,7));\0D\0A\09a.Store(512, double(8));\0D\0A\09a.Store(768, double4(8,9,10,11));\0D\0A}\0D\0A"}
$ dxc -Od -Zi a.hlsl -Fd a.pdb -Fo a.dxbc -T cs_6_0
$ dxc -dumpbin a.dxbc | grep 'void main'
$ dxc -dumpbin a.pdb | grep 'void main'
```

**Actual Behavior**
Since the source is a big part of the value of the debug data I would expect the necessary flag to be documented in `-help` and in the documentation like [SourceLevelDebuggingHLSL.rst](https://github.com/microsoft/DirectXShaderCompiler/blob/cfd47e9806907db055959a8cd56cde7aa48bd7b3/docs/SourceLevelDebuggingHLSL.rst).

**Environment**
- DXC version dxcompiler.dll: 1.7 - 1.7.2308.7 (69e54e290); dxil.dll: 1.7(101.7.2308.12)
- Host Operating System Windows 10 19045

coopp:
This commit adds the compiler option (-Qsource_in_debug_module) to the help text.

New help text output below:

```
Utility Options:
  -dumpbin              Load a binary file rather than compiling
  -extractrootsignature Extract root signature from shader bytecode (must be used with /Fo <file>)
  -getprivate <file>    Save private data from shader blob
  -link                 Link list of libraries provided in <inputs> argument separated by ';'
  -P                    Preprocess to file
  -Qembed_debug         Embed PDB in shader container (must be used with /Zi)
  -Qsource_in_debug_module
                        Embed source info in PDB
  -Qstrip_debug         Strip debug information from 4_0+ shader bytecode  (must be used with /Fo <file>)
  -Qstrip_priv          Strip private data from shader bytecode  (must be used with /Fo <file>)
  -Qstrip_reflect       Strip reflection data from shader bytecode  (must be used with /Fo <file>)
  -Qstrip_rootsignature Strip root signature data from shader bytecode  (must be used with /Fo <file>)
  -setprivate <file>    Private data to add to compiled shader blob
  -setrootsignature <file>
                        Attach root signature to shader bytecode
  -verifyrootsignature <file>
                        Verify shader bytecode with root signature
```

fixes #6028

### DIFF
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -499,7 +499,7 @@ def Qembed_debug : Flag<["-", "/"], "Qembed_debug">, Flags<[CoreOption]>, Group<
 def Qstrip_priv : Flag<["-", "/"], "Qstrip_priv">, Flags<[CoreOption, DriverOption]>, Group<hlslutil_Group>,
   HelpText<"Strip private data from shader bytecode  (must be used with /Fo <file>)">;
 def Qsource_in_debug_module : Flag<["-", "/"], "Qsource_in_debug_module">, Flags<[CoreOption]>, Group<hlslutil_Group>,
-  HelpText<"Embed source info in PDB">;
+  HelpText<"Embed source code in PDB">;
 def Qpdb_in_private : Flag<["-", "/"], "Qpdb_in_private">, Flags<[CoreOption, HelpHidden]>, Group<hlslutil_Group>,
   HelpText<"Store PDB in private user data.">;
 
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -498,8 +498,8 @@ def Qembed_debug : Flag<["-", "/"], "Qembed_debug">, Flags<[CoreOption]>, Group<
   HelpText<"Embed PDB in shader container (must be used with /Zi)">;
 def Qstrip_priv : Flag<["-", "/"], "Qstrip_priv">, Flags<[CoreOption, DriverOption]>, Group<hlslutil_Group>,
   HelpText<"Strip private data from shader bytecode  (must be used with /Fo <file>)">;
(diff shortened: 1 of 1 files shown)


python3kgae (approved):

damyanp:
There's no tests for this.  I'm wondering, though, if we have any tests for help text for any of the other options?

coopp:
There are existing tests for this already in the repo.  I looked before putting up this PR

damyanp:
There may be tests for the command-line option, but there can't be any tests covering your change otherwise they'd have been failing before your change.

Maybe we're ok skipping tests for command-line help, but we should make a conscious decision to do so.

hekota (approved):
LGTM!

Issue #6593, opened 2024-05-07
[Optimization] DXC is missing common factor optimization in some cases


lizhengxing:
**Description**
DXC is missing common factor optimization in some cases. Here's an example code:
```
cbuffer TemporalAAData : register ( b10 )
{
    float2 viewportRelativeSize ;
    float4 outputDimensions ;
}

Texture2D < float4 > HistoryColor : register ( t2 ) ;
SamplerState s_Linear : register ( s1 ) ;
RWTexture1D < float3 > outColorBuffer : register ( u0 ) ;

float4 Test ( in Texture2D < float4 > tex , in SamplerState linearSampler , in float2 uv )
{
    float2 samplePos = uv;
    float2 texPos1 = floor ( samplePos - 0.5f ) + 0.5f ;

    float2 texPos0 = texPos1 - 1 ;
    float2 texPos3 = texPos1 + 2 ;
    float2 texPos12 = texPos1 + samplePos;

    // DXC should recognize (outputDimensions . zw * viewportRelativeSize) is a common factor.
    texPos0 *= outputDimensions . zw * viewportRelativeSize ;
    texPos3 *= outputDimensions . zw * viewportRelativeSize ;
    texPos12 *= outputDimensions . zw * viewportRelativeSize ;

    float4 result = 0.0f ;
    result += tex . SampleLevel ( linearSampler , float2 ( texPos0 . x , texPos0 . y ) , 0.0f );
    result += tex . SampleLevel ( linearSampler , float2 ( texPos12 . x , texPos0 . y ) , 0.0f );
    result += tex . SampleLevel ( linearSampler , float2 ( texPos3 . x , texPos0 . y ) , 0.0f );
    result += tex . SampleLevel ( linearSampler , float2 ( texPos0 . x , texPos12 . y ) , 0.0f );
    result += tex . SampleLevel ( linearSampler , float2 ( texPos12 . x , texPos12 . y ) , 0.0f );
    result += tex . SampleLevel ( linearSampler , float2 ( texPos3 . x , texPos12 . y ) , 0.0f );
    result += tex . SampleLevel ( linearSampler , float2 ( texPos0 . x , texPos3 . y ) , 0.0f );
    result += tex . SampleLevel ( linearSampler , float2 ( texPos12 . x , texPos3 . y ) , 0.0f );
    result += tex . SampleLevel ( linearSampler , float2 ( texPos3 . x , texPos3 . y ) , 0.0f );
    return result ;
}

[...]

PR #6598, opened 2024-05-08
Reassociate: add global reassociation algorithm

lizhengxing:
This PR pulls the upstream change, Reassociate: add global reassociation algorithm (https://github.com/llvm/llvm-project/commit/b8a330c42ab43879119cd3a305756d28aefe9fe6), into DXC with miminal changes.

For the code below:
  foo = (a * b) * c
  bar = (a * d) * c

As the upstream change states, it can identify the a*c is a common factor and redundant.

This is part 1 of the fix for #6593.

### DIFF
--- a/lib/Transforms/Scalar/Reassociate.cpp
+++ b/lib/Transforms/Scalar/Reassociate.cpp
@@ -20,11 +20,11 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "llvm/Transforms/Scalar.h"
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/PostOrderIterator.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SetVector.h"
+#include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/Statistic.h"
 #include "llvm/IR/CFG.h"
 #include "llvm/IR/Constants.h"
@@ -37,6 +37,7 @@
 #include "llvm/Pass.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Transforms/Scalar.h"
 #include "llvm/Transforms/Utils/Local.h"
 #include <algorithm>
 using namespace llvm;
@@ -161,6 +162,13 @@ namespace {
     DenseMap<BasicBlock*, unsigned> RankMap;
     DenseMap<AssertingVH<Value>, unsigned> ValueRankMap;
     SetVector<AssertingVH<Instruction> > RedoInsts;
+
+    // Arbitrary, but prevents quadratic behavior.
+    static const unsigned GlobalReassociateLimit = 10;
+    static const unsigned NumBinaryOps =
+        Instruction::BinaryOpsEnd - Instruction::BinaryOpsBegin;
+    DenseMap<std::pair<Value *, Value *>, unsigned> PairMap[NumBinaryOps];
+
     bool MadeChange;
   public:
     static char ID; // Pass identification, replacement for typeid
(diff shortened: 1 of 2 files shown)


python3kgae (approved):

dmpots:
What is the performance impact of this change?

lizhengxing:
> What is the performance impact of this change?

I haven't run the perf tests. But I believe the impact's smaller than the one I got before. That's why I want to merge the 2 PRs together or merge them consecutively.

dmpots:
> > What is the performance impact of this change?
> 
> I haven't run the perf tests. But I believe the impact's smaller than the one I got before. That's why I want to merge the 2 PRs together or merge them consecutively.

It would be good to get the perf numbers for this change to understand the impact of each part. Also, since it will be possible to run this change by itself with a flag it would be good to understand what the perf looks like.

lizhengxing:
> > > What is the performance impact of this change?
> > 
> > 
> > I haven't run the perf tests. But I believe the impact's smaller than the one I got before. That's why I want to merge the 2 PRs together or merge them consecutively.
> 
> It would be good to get the perf numbers for this change to understand the impact of each part. Also, since it will be possible to run this change by itself with a flag it would be good to understand what the perf looks like.

There's no flag for this upstream change. I thought to use a flag, but it needs to change the interface of Reassociate Pass to pass in the flag. I'm not sure if it's acceptable.
`FunctionPass *llvm::createReassociatePass() { return new Reassociate(); }`

dmpots:
> > > > What is the performance impact of this change?
> > > 
> > > 
> > > I haven't run the perf tests. But I believe the impact's smaller than the one I got before. That's why I want to merge the 2 PRs together or merge them consecutively.
> > 
> > 
> > It would be good to get the perf numbers for this change to understand the impact of each part. Also, since it will be possible to run this change by itself with a flag it would be good to understand what the perf looks like.
> 
> There's no flag for this upstream change. I thought to use a flag, but it needs to change the interface of Reassociate Pass to pass in the flag. I'm not sure if it's acceptable. `FunctionPass *llvm::createReassociatePass() { return new Reassociate(); }`
> 
> Anyway, I started to collect the perf number.

I don't think we need a flag for this change, but I do think we need a sanity check on the perf numbers that it is an overall positive change. For the flag, I was referring to the flag you added in the other PR (was something like `-disable-agressive-reassoc`). That flag was not controlling this modification, so I was just asking that we understand the perf of this change in isolation.

dmpots (changes requested):
I'd like to see the perf data for this change.

dmpots:
> I'd like to see the perf data for this change.

Synced offline. Looks like a nice win overall in our test suite. Reduced ALU in ~40% of shaders (increased it in ~3%). A small number of shaders had occupancy impacts (~1%). There were an equal number of regressions and improvements.

Overall, this change looks to be positive.

dmpots (approved):

tools/clang/test/DXC/Passes/Transforms/Reassociate/basictest.ll
llvm-beanz, line 1: Can I suggest that we re-enable the Reassociate pass tests and move these new tests under there?

I posted a PR to enable those tests: https://github.com/microsoft/DirectXShaderCompiler/pull/6605

dmpots: I think that is a good idea!

llvm-beanz (approved):

llvm-beanz (commented):
I approved the PR because I think this is good, but it would be nice to rebase this on main and put the tests with the other reassociate tests in LLVM.

lizhengxing:
> I approved the PR because I think this is good, but it would be nice to rebase this on main and put the tests with the other reassociate tests in LLVM.

Done. I rebased this PR on main branch and updated the tests.

tex3d (approved):

PR #6601, opened 2024-05-08
Add test to validate that -Qsource_in_debug_module is present in help…

coopp:
This commit adds a new test that validates the help text output contains the -Qsource_in_debug_module option recently enabled by a previous PR #6597 .

This is a follow up commit for Issue: #6028

### DIFF
--- a/tools/clang/test/DXC/help_text.test
+++ b/tools/clang/test/DXC/help_text.test
@@ -1,0 +1,7 @@
+// Test /? and /help option
+// RUN: %dxc /? | FileCheck %s
+// RUN: %dxc /help | FileCheck %s
+
+// Test to ensure the help text for '-Qsource_in_debug_module' is present
+// CHECK:-Qsource_in_debug_module
+// CHECK:Embed source code in PDB


tools/clang/test/DXC/help_text.test
```diff
@@ -0,0 +1,7 @@
+// Test /? and /help option
+// RUN: %dxc /? | FileCheck %s
+// RUN: %dxc /help | FileCheck %s
```
damyanp, line 3:
```suggestion
// Test -? and -help option
// RUN: %dxc -? | FileCheck %s
// RUN: %dxc -help | FileCheck %s
```

Seems dxc generally prefers to use `-` rather than `/` for flags.

damyanp (approved):

tools/clang/test/DXC/help_text.test
coopp: Do you suggest that I change the / usage here?  It works?  Should I also be testing - ?

bob80905 (approved):

PR #6602, opened 2024-05-08
Remove all WARP skips

jenatali:
Now that WARP is actually getting bugfixes and features, and has a way for folks to get new versions without waiting years for an OS update, stop hiding the places where it's failing. Having these hardcoded skips in place makes it impossible to run the tests to see if it's still failing. If we really need these skips, we at least need some kind of command line option to override them for debugging and fixing the failures.


hekota (approved):
LGTM!

damyanp (approved):

PR #6608, opened 2024-05-10
[SPIR-V] Implement WaveMutliPrefix*

### LINKED ISSUE #6600: [SPIR-V] Implement WaveMutliPrefix* intrinsic support
The [WaveMultiPrefix*() Functions](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_ShaderModel6_5.html#wavemultiprefix-functions) were introduced in shader model 6.5 but are not yet supported by the SPIR-V backend. They should be implemented using: https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/NV/SPV_NV_shader_subgroup_partitioned.html

Related to #6545

sudonatalie:
Implements the Shader Model 6.5 WaveMultiPrefix* intrinsic functions
using the group operation from SPV_NV_shader_subgroup_partitioned,
PartitionedExclusiveScanNV, which performs a partitioned exclusive scan
operation across a subset of invocations ("lanes") in a subgroup
("wave"). The subset of the partition is determined by the provided
ballot ("mask") parameter, which follows the same requirements for
valid partitioning and active invocations/lanes as the HLSL parameter.

Note that WaveMultiPrefixCountBits remains unimplemented because it does
not directly map to a SPIR-V GroupNonUniformArithmetic instruction that
accepts the PartitionedExclusiveScanNV Group Operation.

DirectX Spec: https://microsoft.github.io/DirectX-Specs/d3d/HLSL_ShaderModel6_5.html#wavemultiprefix-functions
SPIR-V Extension: https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/NV/SPV_NV_shader_subgroup_partitioned.html

Depends on #6596
Fixes #6600


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -8709,6 +8709,18 @@ SpirvEmitter::processIntrinsicCallExpr(const CallExpr *callExpr) {
...
+  case hlsl::IntrinsicOp::IOP_WaveMultiPrefixSum:
+  case hlsl::IntrinsicOp::IOP_WaveMultiPrefixUProduct:
+  case hlsl::IntrinsicOp::IOP_WaveMultiPrefixProduct:
+  case hlsl::IntrinsicOp::IOP_WaveMultiPrefixBitAnd:
+  case hlsl::IntrinsicOp::IOP_WaveMultiPrefixBitOr:
+  case hlsl::IntrinsicOp::IOP_WaveMultiPrefixBitXor: {
+    const auto retType = callExpr->getCallReturnType(astContext);
+    retVal = processWaveReductionOrPrefix(
+        callExpr, translateWaveOp(hlslOpcode, retType, srcLoc),
+        spv::GroupOperation::PartitionedExclusiveScanNV);
```
Keenuts, line 8722: Why is this using `PartitionedExclusiveScanNV` which is NV specific vs `ExclusiveScan` which is brought by the more generic `GroupNonUniformBallot`?
Seems like for exemple, WavePrefixSum could be implemented using `OpGroupNonUniformIAdd` and the `ExclusiveScan` group operation or am I missing something?

sudonatalie: The ballot (mask) operand is added as an operand to GroupNonUniform arithmetic instructions by the `SPV_NV_shader_subgroup_partitioned` extension, and then `PartitionedExclusiveScanNV` is the group operation that makes that operand required and and allows us to specify an arbitrary subset of the subgroup. Without it, you're doing an exclusive scan on the whole subgroup which is what the non-"Multi" WavePrefix HLSL intrinsics do.

Keenuts: Ahh, for some reason I only looked at the WavePrefixSum part...
Thanks for the details! Yes, agree, don't see a way to represent this type of decoupage in the core part of the spec.

Keenuts (approved):

s-perron (approved):

PR #6610, opened 2024-05-10
Fix dxil-remove-dead-blocks removing switch with multiple same successor

amaiorano:
Given a switch with a constant condition and all cases the same (branching to the same successor), dxil-remove-dead-blocks would incorrectly remove the switch when replacing it with a branch, by forgetting to remove the N-1 incoming values to the PHIs in the successor block.

### DIFF
--- a/tools/clang/test/DXC/Passes/DxilRemoveDeadBlocks/switch-with-multiple-same-successor.ll
+++ b/tools/clang/test/DXC/Passes/DxilRemoveDeadBlocks/switch-with-multiple-same-successor.ll
@@ -1,0 +1,369 @@
+; RUN: %dxopt %s -hlsl-passes-resume -dxil-remove-dead-blocks -S | FileCheck %s
+
+; Validate that a switch with a constant condition and multiple of the same successor
+; is correctly removed, ensuring that PHIs in the successor are properly updated.
+; For instance, in:
+;
+;
+; if.end.1:                                         ; preds = %for.inc
+;   switch i32 1, label %sw.epilog.1 [
+;     i32 1, label %dx.struct_exit.new_exiting.1
+;     i32 2, label %dx.struct_exit.new_exiting.1
+;     i32 3, label %dx.struct_exit.new_exiting.1
+;     i32 5, label %dx.struct_exit.new_exiting.1
+;   ], !dbg !31 ; line:23 col:5
+; 
+; sw.epilog.1:                                      ; preds = %if.end.1
+;   br label %dx.struct_exit.new_exiting.1
+; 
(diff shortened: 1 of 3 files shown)


tools/clang/test/DXC/Passes/DxilRemoveDeadBlocks/switch-with-multiple-same-successor.ll
```diff
@@ -0,0 +1,369 @@
...
+;
+; After dxil-remove-dead-blocks, the multiple `%if.end.1` in preds and in the two phi instructions should be removed,
+; and only one instance should be left.
+
+; CHECK:      if.end.1:                                         ; preds = %for.inc
+; CHECK-NEXT:   br label %dx.struct_exit.new_exiting.1
+
+; CHECK:     dx.struct_exit.new_exiting.1:                     ; preds = %if.end.1, %for.inc
+; CHECK-NEXT:  %do_discard.2.1 = phi i32 [ 1, %if.end.1 ], [ %do_discard.2, %for.inc ]
+; CHECK-NEXT:  %g.2.i0.1 = phi float [ 0x3FF921FB60000000, %if.end.1 ], [ %g.2.i0, %for.inc ]
```
amaiorano, line 34: I'm not sure exactly how to generalize this test. I tried replacing the specific labels with variables, but ran into problems of lit not being able to match. I think I over-generalized. How would you do this?

amaiorano:
Also want to mention that this crash only occured with SM 6.6 enabled, although I don't believe it's specific to 6.6. Here's a Compiler Explorer link with the same test code that demonstrates the failure: https://godbolt.org/z/KPsrdPeM6

tools/clang/test/DXC/Passes/DxilRemoveDeadBlocks/switch-with-multiple-same-successor.ll
adam-yang: I don't understand, what is the issue? The test looks good to me.

amaiorano: Most tests seem to replace specific variable/label names with regex-based variables like `[[name:<regex>]]`. If this looks good, then great.

adam-yang: Usually when we have HLSL tests, it's not guaranteed what the LLVM values will be named, so using patterns makes the tests less brittle. In this case dxil-remove-dead-blocks won't rename any values or labels, so it's not necessary to use patterns.

adam-yang (approved):

llvm-beanz (approved):

PR #6617, opened 2024-05-13
Amend template specialization DXASSERT conditions

### LINKED ISSUE #4875: Assertion reached in SEMA related to builtin template specialization
Hi,

Reached this assert while trying to reproduce https://github.com/microsoft/DirectXShaderCompiler/issues/4856

revision: 2c3d965b2fc734c09e76631de65f8ffbe4e3e68b
command line: `./build/bin/dxc -T cs_6_0 -E main repro.hlsl`

code:
```hlsl
[[vk::binding(0)]] Texture2D<float> g_depth_buffer : register(t0);

#define A_GPU
#define A_HLSL
#include "ffx_a.h"

groupshared float g_group_shared_depth_values[16][16];

AF4 SpdLoadIntermediate(AU1 x, AU1 y) {
	float f = g_group_shared_depth_values[x][y];
	return f.xxxx;
}


[numthreads(32, 8, 1)]
void main(uint3 dispatch_thread_id : SV_DispatchThreadID, uint3 group_id : SV_GroupID, uint group_index : SV_GroupIndex){
}
```

I do not have a `ffx_a.h` file on my machine, at least not that I know of. Removing the first line makes DXC fails with the "ffx_a.h not found".

Error:
```
dxc: SemaHLSL.cpp:788: clang::QualType GetOrCreateTemplateSpecialization(clang::ASTContext &, clang::Sema &, clang::ClassTemplateDecl *, ArrayRef<clang::TemplateArgument>): Assertion `false == sema.InstantiateClassTemplateSpecialization( NoLoc, specializationDecl, TemplateSpecializationKind::TSK_ImplicitInstantiation, true)' failed.
Aborted
```

### LINKED ISSUE #6615: DXASSERT condition for failed template specialization needs amending
**Description**
A large codebase that is successfully compiled in many projects sometimes triggers a DXASSERT when specializing a template type - with no reason being apparent.

**Steps to Reproduce**
// bug.hlsl - a very artificial example that illustrates the issue
float a;
#include "a.h"  // include file doesn't exist - this should produce an fatal error diagnostic
void b() {};
int3 c(int X)
{
    return X.xxx;  // assert triggered if include file a.h doesn't exist, compilation succeeds if it does
}
// dxc -T lib_6_8 bug.hlsl

**Actual Behavior**
An investigation revealed the DXASSERT was only triggered if there was some error detected earlier in the compilation - e.g. an include file is not found. 
Clang suppresses template specialization if a fatal error has been reported in order to avoid the risk of a cascade of secondary error diagnostics from a failed specialization.
The DXASSERTs only check for a failed specialization, without allowing for the failure being due to an earlier unrelated fatal error - this has the unintended result of hiding the fatal error, and providing no indication of what the  problem is.

The DXASSERTs should not trigger if a fatal error has been registered.

**Environment**
- DXC version :libdxcompiler.so: 1.8(dev;4588-f6d17595)
- Host Operating System : Any

tcorringham:
Clang suppresses template specialization if a fatal error has been reported in order to reduce the risk of a cascade of secondary error diagnostics.

However, DXC DXASSERTs if template specialization fails - even if that is due to an unrelated fatal error - which has the unintended result of hiding the fatal error and hence providing no indication of what the problem is.

The DXASSERT conditions have been amended so they are no longer raised if a fatal error has been registered.

Fixes #6615 
Fixes #4875


llvm-beanz (commented):
This is all a grotesque misuse of asserts. We shouldn't be asserting for things that can actually fail.

My gut feeling here is that if we're going to spend the time fixing this we should actually refactor the code so that we propagate the error condition and gracefully exit early.

tcorringham:
Asserts are used widely to detect exceptional error conditions that should never occur in general use - this style of programming has advantages and disadvantages, which could be debated at length (I'm not a fan...)

In this instance, it was expected that the HLSL template specialization can never fail (it is straightforward, and well exercised) - however, when a fatal error has already been registered template specialization is suppressed leading to the failure of the specialization, and thereby triggering the assert. 

There could be an argument for changing the general approach of using asserts, and possibly replacing them with a fatal error - that should be possible for most, but I suspect there would be cases where continuing compilation to the point where error diagnostics would be generated wouldn't be possible.

In this specific case, the assert could easily be replaced by a fatal error - but the only expected trigger is when another fatal error has already been registered. Simply suppressing the assert in that case does allow the compilation to continue, and the fatal error is then generated. Adding a second fatal error to replace the assert seems unnecessary, though a fatal error could be generated if specialization failed due to some other reason.
 
Simply changing the assert condition as does solve the immediate issue, but doesn't address the wider issue of the use of asserts. I'm open to suggestions of how to progress this...

llvm-beanz:
I'm not asking you to fix DXC's misuse of asserts everywhere, just in the code that you're modifying in this PR already. I don't think any best-practices for C/C++ programming recommends using them in this manner, but more importantly they're not consistent with LLVM convention and style (which we've adopted for all future changes).

You've already had to update a chain of asserts across all the use sites, we could instead just clean them up. It isn't actually any more work.

Here's a patch that accomplishes the same thing you're doing here, but cleans up the code it touches along the way:

```diff
diff --git a/tools/clang/lib/Sema/SemaHLSL.cpp b/tools/clang/lib/Sema/SemaHLSL.cpp
index 4df32f9a659..7e9ba3f1334 100644
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -869,11 +869,10 @@ GetOrCreateTemplateSpecialization(ASTContext &context, Sema &sema,
     if (specializationDecl->getInstantiatedFrom().isNull()) {
       // InstantiateClassTemplateSpecialization returns true if it finds an
       // error.
-      DXVERIFY_NOMSG(false ==
-                     sema.InstantiateClassTemplateSpecialization(
-                         NoLoc, specializationDecl,
-                         TemplateSpecializationKind::TSK_ImplicitInstantiation,
-                         true));
+      if (sema.InstantiateClassTemplateSpecialization(
+              NoLoc, specializationDecl,
+              TemplateSpecializationKind::TSK_ImplicitInstantiation, true))
+        return QualType();
     }
     return context.getTemplateSpecializationType(
         TemplateName(templateDecl), templateArgs.data(), templateArgs.size(),
@@ -930,16 +929,15 @@ static QualType GetOrCreateMatrixSpecialization(
       context, *sema, matrixTemplateDecl,
       ArrayRef<TemplateArgument>(templateArgs));
 
-#ifndef NDEBUG
-  // Verify that we can read the field member from the template record.
-  DXASSERT(matrixSpecializationType->getAsCXXRecordDecl(),
+  if (!matrixSpecializationType.isNull()) {
+    assert(matrixSpecializationType->getAsCXXRecordDecl() &&
            "type of non-dependent specialization is not a RecordType");
-  DeclContext::lookup_result lookupResult =
-      matrixSpecializationType->getAsCXXRecordDecl()->lookup(
-          DeclarationName(&context.Idents.get(StringRef("h"))));
-  DXASSERT(!lookupResult.empty(),
+    [[maybe_unused]] DeclContext::lookup_result lookupResult =
+        matrixSpecializationType->getAsCXXRecordDecl()->lookup(
+            DeclarationName(&context.Idents.get(StringRef("h"))));
+    assert(!lookupResult.empty() &&
            "otherwise matrix handle cannot be looked up");
-#endif
+  }
 
   return matrixSpecializationType;
 }
@@ -965,16 +963,16 @@ GetOrCreateVectorSpecialization(ASTContext &context, Sema *sema,
       context, *sema, vectorTemplateDecl,
       ArrayRef<TemplateArgument>(templateArgs));
 
-#ifndef NDEBUG
-  // Verify that we can read the field member from the template record.
-  DXASSERT(vectorSpecializationType->getAsCXXRecordDecl(),
+  if (!vectorSpecializationType.isNull()) {
+    // Verify that we can read the field member from the template record.
+    assert(vectorSpecializationType->getAsCXXRecordDecl() &&
            "type of non-dependent specialization is not a RecordType");
-  DeclContext::lookup_result lookupResult =
-      vectorSpecializationType->getAsCXXRecordDecl()->lookup(
-          DeclarationName(&context.Idents.get(StringRef("h"))));
-  DXASSERT(!lookupResult.empty(),
+    [[maybe_unused]] DeclContext::lookup_result lookupResult =
+        vectorSpecializationType->getAsCXXRecordDecl()->lookup(
+            DeclarationName(&context.Idents.get(StringRef("h"))));
+    assert(!lookupResult.empty() &&
            "otherwise vector handle cannot be looked up");
-#endif
+  }
 
   return vectorSpecializationType;
 }
@@ -1021,16 +1019,16 @@ GetOrCreateNodeOutputRecordSpecialization(ASTContext &context, Sema *sema,
   QualType specializationType = GetOrCreateTemplateSpecialization(
       context, *sema, templateDecl, ArrayRef<TemplateArgument>(templateArgs));
 
-#ifdef DBG
-  // Verify that we can read the field member from the template record.
-  DXASSERT(specializationType->getAsCXXRecordDecl(),
+  if (!specializationType.isNull()) {
+    // Verify that we can read the field member from the template record.
+    assert(specializationType->getAsCXXRecordDecl() &&
            "type of non-dependent specialization is not a RecordType");
-  DeclContext::lookup_result lookupResult =
-      specializationType->getAsCXXRecordDecl()->lookup(
-          DeclarationName(&context.Idents.get(StringRef("h"))));
-  DXASSERT(!lookupResult.empty(),
+    [[maybe_unused]] DeclContext::lookup_result lookupResult =
+        specializationType->getAsCXXRecordDecl()->lookup(
+            DeclarationName(&context.Idents.get(StringRef("h"))));
+    assert(!lookupResult.empty() &&
            "otherwise *NodeOutputRecords handle cannot be looked up");
-#endif
+  }
 
   return specializationType;
 }
@@ -5242,12 +5240,13 @@ public:
           // This is a bit of a special case we need to handle. Because the
           // buffer types don't use their template parameter in a way that would
           // force instantiation, we need to force specialization here.
-          GetOrCreateTemplateSpecialization(
+          QualType Ty = GetOrCreateTemplateSpecialization(
               *m_context, *m_sema,
               cast<ClassTemplateDecl>(
                   TST->getTemplateName().getAsTemplateDecl()),
               llvm::ArrayRef<TemplateArgument>(TST->getArgs(),
                                                TST->getNumArgs()));
+          return Ty.isNull();
         }
         if (const RecordType *recordType = argType->getAs<RecordType>()) {
           if (!recordType->getDecl()->isCompleteDefinition()) {

```

tcorringham:
I have adopted the suggested approach - however, use of hasFatalErrorOccurred() is still required in some places to avoid conditions which trigger subsequent asserts.

tcorringham:
Are there any further comments on this change? It will fix two issues, and get rid of an assert that causes confusion quite frequently.

tex3d (approved):
Thanks for this fix.

I don't understand why we continue to parse after a fatal error in preprocessing such as a missing include file.  On top of that, I don't understand the reasoning behind specifically failing template specializations, for no other legitimate reason, after an unrelated fatal error has occurred.

llvm-beanz (approved):

tcorringham:
Can anyone tell me how to resolve the pending required microsoft.DirectXShaderCompiler check please? - I either don't know or have forgotten!

tcorringham:
Could someone with the necessary super-power please merge this? Thanks!

damyanp (approved):

PR #6618, opened 2024-05-14
[SPIR-V] Add support for SampleCmpLevel

### LINKED ISSUE #6613: [SPIR-V][Feature Request][SM6.7] Implement SampleCmpLevel
SM6.7 added [SampleCmlLevel](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_7_Advanced_Texture_Ops.html#samplecmplevel) which is similar to `SampleCmp`, but which fixes the LOD at which the comparison is done.

```hlsl
Texture2D<float> tex;
SamplerComparisonState samplerComparisonState;

float4 main() : SV_Target {
  float tmp = tex.SampleCmpLevel(samplerComparisonState, float2(0, 0), 2, 3);
  return tmp.xxxx;
}
```

```
$ dxc -T ps_6_7 shader.hlsl -spirv -fspv-target-env=vulkan1.3
shader.hlsl:7:19: error: intrinsic 'SampleCmpLevel' method unimplemented
  float tmp = tex.SampleCmpLevel(samplerComparisonState, float2(0, 0), 2, 3);
```

Seems like this could be implemented with a small change in DXC: `SampleCmpLevelZero` uses the correct instruction, but fixes the LOD to 0. But passing a register is possible, so seems like implementation will be straightforward.

Keenuts:
SampleCmpLevel is similar to SampleCmpLevel0, except the LOD level can be specified using either a const-offset, or a variable. This should be available starting SM6.7

Fixes #6613

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -5125,10 +5125,13 @@ SpirvEmitter::processIntrinsicMemberCall(const CXXMemberCallExpr *expr,
     retVal = processTextureSampleGrad(expr);
     break;
   case IntrinsicOp::MOP_SampleCmp:
-    retVal = processTextureSampleCmpCmpLevelZero(expr, /*isCmp=*/true);
+    retVal = processTextureSampleCmp(expr);
     break;
   case IntrinsicOp::MOP_SampleCmpLevelZero:
-    retVal = processTextureSampleCmpCmpLevelZero(expr, /*isCmp=*/false);
+    retVal = processTextureSampleCmpLevel(expr, /* hasExplicitLod= */ false);
+    break;
+  case IntrinsicOp::MOP_SampleCmpLevel:
+    retVal = processTextureSampleCmpLevel(expr, /* hasExplicitLod= */ true);
     break;
   case IntrinsicOp::MOP_GatherRed:
     retVal = processTextureGatherRGBACmpRGBA(expr, /*isCmp=*/false, 0);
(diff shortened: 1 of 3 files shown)


sudonatalie (approved):
LGTM!

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5577,10 +5579,56 @@ SpirvEmitter::processTextureSampleCmpCmpLevelZero(const CXXMemberCallExpr *expr,
...
+  SpirvInstruction *constOffset = nullptr, *varOffset = nullptr;
+
+  // Subtract 1 for clamp (if it exists), 1 for status (if it exists),
+  // and 3 for sampler_state, location, and compare_value.
+  const bool hasOffsetArg = numArgs - hasStatusArg - hasClampArg - 3 > 0;
+  if (hasOffsetArg)
+    handleOffsetInMethodCall(expr, 3, &constOffset, &varOffset);
+
+  const auto retType = expr->getDirectCallee()->getReturnType();
+  const auto imageType = imageExpr->getType();
```
s-perron, line 5610: Is it worth moving this code into a separate function so that it is not repeated in the next function? I'm on the fence personally, so I'll go with whatever you decide. I just want to know that you thought about it.

The pros are that we will

1. Not repeat the code.
2. We could split the next function into two functions to remove the flag, and have a third function easily.

Cons

1. This code essentially loads operands to `createImageSample`, so the return value would have to be struct that gets unpacked. This could make the code harder to read, but I'm not sure about that.

s-perron (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts: Each intrinsic has its own function. Not great for duplication, but the code stays simple & easy to understand.

s-perron (approved):

PR #6619, opened 2024-05-14
Pix: Cope with group shared AS->MS payload

jeffnn:
This change copes with the AS->MS payload being placed in group-shared by the application (and MSFT's samples do indeed do this). (TIL, thanks to pow2clk, that the spec says that the payload counts against the group-shared total, implying, if not explicitly stating, that at least on some platforms, the payload will be in group-shared anyway.)

The MS pass needs to be given data from the AS about the AS's thread group topology, and this is done by extending the payload struct to add three uints. This can't be done when the payload is resident in group-shared, of course, because that would change the layout of group-shared memory. 
So the new approach here is to copy the payload to a new alloca (in the default address space) struct with the members of the base struct plus the extended data the MS needs, and then to copy piece-wise because llvm.memcpy isn't appropriate for group-shared-to-normal address space copies.


tools/clang/test/HLSLFileCheck/pix/DebugAsGroupSharedTrickyTypesPayload.hlsl
```diff
@@ -0,0 +1,28 @@
...
+
+// And no more:
+// CHECK-NOT: [[LOAD2:%.*]] = load {{.*}}, {{.*}} addrspace(3)* getelementptr inbounds
+
+struct MyPayload {
+  uint i;
+  void Init() { i = 27; }
+struct {
+  int bf0 : 7;
+  int bf1 : 11;
```
jeffnn, line 19: Oops: Shower thought: add an embedded resource here. Not sure if that's legal for a payload, but need to check.

jeffnn: Compiler crashes in this case. Filed #6620

lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
```diff
@@ -45,143 +45,154 @@ void DxilPIXAddTidToAmplificationShaderPayload::applyOptions(PassOptions O) {
...
+      auto *GroupYxNumZ = B.CreateMul(
+          GroupIdY, HlslOP->GetU32Const(m_DispatchArgumentZ), "GroupYxNumZ");
+      auto *FlatGroupNumZY =
+          B.CreateAdd(GroupIdZ, GroupYxNumZ, "FlatGroupNumZY");
+      auto *GroupXxNumYZ = B.CreateMul(
+          GroupIdX,
+          HlslOP->GetU32Const(m_DispatchArgumentY * m_DispatchArgumentZ),
+          "GroupXxNumYZ");
+      auto *FlatGroupID =
+          B.CreateAdd(GroupXxNumYZ, FlatGroupNumZY, "FlatGroFlatGroupIDupNum");
```
bob80905, line 142: "FlatGroFlatGroupIDupNum"
I think this might be a typo

lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
```diff
@@ -45,143 +45,154 @@ void DxilPIXAddTidToAmplificationShaderPayload::applyOptions(PassOptions O) {
...
-    AddValueToExpandedPayload(
-        HlslOP, B, expanded, NewStructAlloca,
-        OriginalPayloadStructType->getStructNumElements() + 2,
-        DispatchMesh.get_threadGroupCountZ());
-  }
-
-  DM.ReEmitDxilResources();
 
-  return true;
+  return false;
```
bob80905, line 195: Why  change the return value?

jeffnn: Because there's always exactly one DispatchMesh, and if we don't find one, then this change will return false indicating that the pass didn't change anything.

lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
jeffnn: Woops, thanks! Looks like I pasted in a new name but didn't select the old one first, resulting in the new name inside the old!
I tend not to enforce names for DXIL values because they aren't actually stable anyway. In fact, the tests tend to go out of their way to match a regex to the actual name...

tools/clang/test/HLSLFileCheck/pix/DebugAsGroupSharedTrickyTypesPayload.hlsl
```diff
@@ -0,0 +1,28 @@
+// RUN: %dxc -enable-16bit-types -Od -Emain -Tas_6_6 %s | %opt -S -hlsl-dxil-PIX-add-tid-to-as-payload,dispatchArgY=3,dispatchArgZ=7 | %FileCheck %s
```
pow2clk, line 1: Don't see why enable-16bit-types is needed. Harmless though.

lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
```diff
@@ -45,143 +45,154 @@ void DxilPIXAddTidToAmplificationShaderPayload::applyOptions(PassOptions O) {
...
+      StructIndices.back() = B.getInt32(j);
+      CopyAggregate(B, AT->getArrayElementType(), Source, Dest, StructIndices);
+    }
+  } else {
+    auto *SourceGEP = B.CreateGEP(Source, GEPIndices, "CopyStructSourceGEP");
+    Value *Val = B.CreateLoad(SourceGEP, "CopyStructLoad");
+    auto *DestGEP = B.CreateGEP(Dest, GEPIndices, "CopyStructDestGEP");
+    B.CreateStore(Val, DestGEP, "CopyStructStore");
+  }
+}
```
pow2clk, line 84: I could swear I've seen something like this before, but the closest I came was the memcpy splitting code, which isn't quite right. Was this modified from anything existing?

pow2clk (approved):
Looks good, just a minor nit and a question

lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
jeffnn: No, it was coded fresh.

tools/clang/test/HLSLFileCheck/pix/DebugAsGroupSharedTrickyTypesPayload.hlsl
jeffnn: Just to exercise the "half" properly.

jeffnn: Oh right, this file has no halfs/(halves?)! That's a copy-paste from the other file that does have one half. A whole half, that is.

lib/DxilPIXPasses/DxilPIXAddTidToAmplificationShaderPayload.cpp
jeffnn: (The nearest thing I've seen descends a DIType hierarchy. Didn't find an example of iterating through a real type like this)

coopp (approved):
Looks good to me..

PR #6623, opened 2024-05-14
Fix false positive assert in SuccIterator::operator+=

amaiorano:
The implementation of index_is_valid was incorrect returning false when the input index was equal to Term->getNumSuccessors(). The end iterator will have such an index, and it is valid to construct such an iterator.

For example, assume we have a block with two successors:

```
BasicBlock* bb = ...;
SuccIterator b(bb); // Index 0
SuccIterator e(bb, true); // Index 2, for example

SuccIterator v = b;
b += 2; // Without this fix, this asserts
assert(b == e);
```

Note that this was also fixed upstream in https://reviews.llvm.org/D47467


amaiorano:
I also want to add that the reason I think this assertion has not been much of a problem before is that `SuccIterator::operator+=` isn't called often. In my test case, however, the failure occurs in `LazyValueInfoCache::threadEdge`, where we build a `worklist` of blocks, and perform a depth-first search on it to process all successors:

```
  std::vector<BasicBlock*> worklist;
  worklist.push_back(OldSucc);
...
  while (!worklist.empty()) {
    BasicBlock *ToUpdate = worklist.back();
    worklist.pop_back();
...
    worklist.insert(worklist.end(), succ_begin(ToUpdate), succ_end(ToUpdate));
```

Note how we insert the range of successors of `ToUpdate` at the end of `worklist` at the bottom of the loop. The assert only triggers when `worklist` has enough capacity to copy in the range of values. In that case, at least in libc++, the implementation of std::vector::insert ultimately ends up doing something like:

```cpp
                __m = __first;
                difference_type __diff = this->__end_ - __p;
                std::advance(__m, __diff);
```
And `std::advance` will ends up effectively calling `__m += _diff` on the iterator.

I think that in general, the allocation pattern for the vectors is such that it typically doesn't have enough capacity when a range of values is added, so it grows the vector, and avoids calling `operator+=`. I was able to make it happen more often by modifying `LazyValueInfoCache::threadEdge` so that the `worklist` reserves 100 values:
```cpp
  std::vector<BasicBlock*> worklist;
  worklist.reserve(100); // HACK: test
  worklist.push_back(OldSucc);
```

This made it fail on the very first iteration of the loop, where in my test case, it happens after a few iterations. Sadly, though, this change did not make any other tests fail (when running `ninja check-all`).

llvm-beanz (commented):
nit: we tend to wrap modified code in the inherited LLVM codebase inside comment blocks so we know it was intentionally modified in DXC.

Can you add `// HLSL change` comments around the code you're adding and modifying?

Otherwise looks good to me.

include/llvm/IR/CFG.h
```diff
@@ -130,7 +130,7 @@ class SuccIterator {
   typedef SuccIterator<Term_, BB_> Self;
 
   inline bool index_is_valid(int idx) {
-    return idx >= 0 && (unsigned) idx < Term->getNumSuccessors();
+    return idx >= 0 && (unsigned)idx <= Term->getNumSuccessors();
```
sudonatalie, line 133:
```suggestion
    // HLSL Change Begin
    return idx >= 0 && (unsigned)idx <= Term->getNumSuccessors();
    // HLSL Change End
```

sudonatalie (approved):
LGTM (with suggestion to also add the HLSL wrapper comments around the change itself)

llvm-beanz (approved):

PR #6625, opened 2024-05-15
Add a flag for the upstream global reassociation algorithm change

lizhengxing:
This PR (https://github.com/microsoft/DirectXShaderCompiler/pull/6598) pulls the upstream global reassociation algorithm change in DXC and can reduce redundant calculations obviously.

However, from the testing result of a large offline suite of shaders, some shaders got worse compilation results and couldn't benefit from this upstream change.

This PR adds a flag for the upstream global reassociation change. It would be easier to roll back if a shader get worse compilation result due to this upstream change.

This is part 2 of the fix for #6593.

### DIFF
--- a/include/dxc/Support/DxcOptToggles.h
+++ b/include/dxc/Support/DxcOptToggles.h
@@ -38,6 +38,8 @@ enum {
 static const Toggle TOGGLE_GVN = {"gvn", DEFAULT_ON};
 static const Toggle TOGGLE_LICM = {"licm", DEFAULT_ON};
 static const Toggle TOGGLE_SINK = {"sink", DEFAULT_ON};
+static const Toggle TOGGLE_ENABLE_AGGRESSIVE_REASSOCIATION = {
+    "aggressive-reassociation", DEFAULT_ON};
 static const Toggle TOGGLE_LIFETIME_MARKERS = {"lifetime-markers", DEFAULT_ON};
 static const Toggle TOGGLE_PARTIAL_LIFETIME_MARKERS = {
     "partial-lifetime-markers", DEFAULT_OFF};
--- a/lib/Transforms/Scalar/Reassociate.cpp
+++ b/lib/Transforms/Scalar/Reassociate.cpp
@@ -176,6 +176,22 @@ namespace {
       initializeReassociatePass(*PassRegistry::getPassRegistry());
     }
 
+    // HLSL Change - begin
+    // Enable global reassociation when HLSLEnableAggressiveReassociation is
+    // set
+    bool HLSLEnableAggressiveReassociation = true;
+    Reassociate(bool HLSLEnableAggressiveReassociation) : Reassociate() {
+      this->HLSLEnableAggressiveReassociation =
+          HLSLEnableAggressiveReassociation;
+    }
+
+    void applyOptions(PassOptions O) override {
+      GetPassOptionBool(O, "EnableAggressiveReassociation",
+                        &HLSLEnableAggressiveReassociation,
+                        /*defaultValue*/ true);
+    }
+    // HLSL Change - end
+
     bool runOnFunction(Function &F) override;
 
     void getAnalysisUsage(AnalysisUsage &AU) const override {
@@ -241,6 +257,13 @@ INITIALIZE_PASS(Reassociate, "reassociate",
 
 // Public interface to the Reassociate pass
 FunctionPass *llvm::createReassociatePass() { return new Reassociate(); }
+
+// HLSL Change - begin
+FunctionPass *
+llvm::createReassociatePass(bool HLSLEnableAggressiveReassociation) {
+  return new Reassociate(HLSLEnableAggressiveReassociation);
+}
+// HLSL Change - end
 
 /// Return true if V is an instruction of the specified opcode and if it
 /// only has one use.
@@ -2243,7 +2266,8 @@ void Reassociate::ReassociateExpression(BinaryOperator *I) {
     return;
   }
 
-  if (Ops.size() > 2 && Ops.size() <= GlobalReassociateLimit) {
+  if (HLSLEnableAggressiveReassociation && // HLSL Change
+      (Ops.size() > 2 && Ops.size() <= GlobalReassociateLimit)) {
     // Find the pair with the highest count in the pairmap and move it to the
     // back of the list so that it can later be CSE'd.
     // example:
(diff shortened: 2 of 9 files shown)


tools/clang/test/HLSLFileCheck/passes/transforms/reassociate/reassociation-flag.hlsl
```diff
@@ -0,0 +1,31 @@
+// RUN: %dxc -T cs_6_3 -E cs_main %s -opt-enable enable-aggressive-reassociation  | FileCheck %s -check-prefixes=CHECK,COMMON_FACTOR
```
dmpots, line 1: This is good for validating that the flag gets passed through from the driver, but I think we should also add a way to control it for unit tests. 

To add it for unit testing we need to modify the hctdb.py file. This will hook it up for using in dxcopt. You can see an example of how it is done for the instcombine [NoSink ](https://github.com/microsoft/DirectXShaderCompiler/blob/d9caef5289dc7fee304817963ea55d229a1eae4d/utils/hct/hctdb.py#L6500)flag

```
        add_pass(
            "instcombine",
            "InstructionCombiningPass",
            "Combine redundant instructions",
            [
                {"n": "NoSink", "t": "bool", "c": 1},
            ],
        )
```

And then how it is used for testing [tools/clang/test/HLSLFileCheck/passes/llvm/instcombine/nosink.ll](https://github.com/microsoft/DirectXShaderCompiler/blob/d9caef5289dc7fee304817963ea55d229a1eae4d/tools/clang/test/HLSLFileCheck/passes/llvm/instcombine/nosink.ll#L1)

Then we can take the unit test from the previous PR and run it with dxopt passing the flag on and off to show the difference. You will have to put the test in a new file under HLSLFileCheck since that is how we run a dxopt vs. opt test.

dmpots (changes requested):

include/dxc/Support/DxcOptToggles.h
```diff
@@ -38,6 +38,8 @@ enum {
 static const Toggle TOGGLE_GVN = {"gvn", DEFAULT_ON};
 static const Toggle TOGGLE_LICM = {"licm", DEFAULT_ON};
 static const Toggle TOGGLE_SINK = {"sink", DEFAULT_ON};
+static const Toggle TOGGLE_ENABLE_AGGRESSIVE_REASSOCIATION = {
+    "enable-aggressive-reassociation", DEFAULT_ON};
```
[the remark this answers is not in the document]
lizhengxing: Done.

lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -176,6 +176,16 @@ namespace {
       initializeReassociatePass(*PassRegistry::getPassRegistry());
     }
 
+    // HLSL Change - begin
+    // Enable global reassociation when HLSLEnableAggressiveReassociation is
+    // set
+    bool m_HLSLEnableAggressiveReassociation = true;
+    Reassociate(bool HLSLEnableAggressiveReassociation) : FunctionPass(ID) {
+      initializeReassociatePass(*PassRegistry::getPassRegistry());
```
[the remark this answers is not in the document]
lizhengxing: Done.

lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -2243,7 +2260,8 @@ void Reassociate::ReassociateExpression(BinaryOperator *I) {
     return;
   }
 
-  if (Ops.size() > 2 && Ops.size() <= GlobalReassociateLimit) {
+  if (m_HLSLEnableAggressiveReassociation && // HLSL Change
```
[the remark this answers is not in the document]
lizhengxing: Done. I guard the PairMap clear code on this flag as well.

tools/clang/test/HLSLFileCheck/passes/transforms/reassociate/reassociation-flag.hlsl
lizhengxing: Done.

dmpots (approved):
LGTM

lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -161,13 +162,35 @@ namespace {
...
   public:
     static char ID; // Pass identification, replacement for typeid
     Reassociate() : FunctionPass(ID) {
       initializeReassociatePass(*PassRegistry::getPassRegistry());
     }
 
+    // HLSL Change - begin
+    // Enable global reassociation when HLSLEnableAggressiveReassociation is
+    // set
+    bool m_HLSLEnableAggressiveReassociation = true;
```
llvm-beanz, line 182: Please follow LLVM naming conventions for new code. LLVM convention does not use Hungarian notation prefixes:

lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -2234,18 +2265,132 @@ void Reassociate::ReassociateExpression(BinaryOperator *I) {
...
+      }
+      // Skip extremely long expressions.
+      if (Ops.size() > GlobalReassociateLimit)
+        continue;
+
+      // Add all pairwise combinations of operands to the pair map.
+      unsigned BinaryIdx = I.getOpcode() - Instruction::BinaryOpsBegin;
+      SmallSet<std::pair<Value *, Value *>, 32> Visited;
+      for (unsigned i = 0; i < Ops.size() - 1; ++i) {
+        for (unsigned j = i + 1; j < Ops.size(); ++j) {
```
llvm-beanz, line 2349: nit: `j` should be `J` and `i` should be something else capitalized since `I` is already used. You could rename `I` to `Inst` as well if you prefer `I` and `J` for indices.

lib/Transforms/Scalar/Reassociate.cpp
```diff
@@ -2234,18 +2265,132 @@ void Reassociate::ReassociateExpression(BinaryOperator *I) {
...
+          // Canonicalize operand orderings.
+          Value *Op0 = Ops[i];
+          Value *Op1 = Ops[j];
+          if (std::less<Value *>()(Op1, Op0))
+            std::swap(Op0, Op1);
+          if (!Visited.insert({Op0, Op1}).second)
+            continue;
+          auto res = PairMap[BinaryIdx].insert({{Op0, Op1}, 1});
+          if (!res.second)
+            ++res.first->second;
```
llvm-beanz, line 2359: nit:
```suggestion
          auto Res = PairMap[BinaryIdx].insert({{Op0, Op1}, 1});
          if (!Res.second)
            ++Res.first->second;
```

tools/clang/test/HLSLFileCheck/passes/llvm/reassociate/reassociation-flag.hlsl
llvm-beanz, line 1: Can we pull these tests out of the `HLSLFileCheck` folder and put them under `tests/DXC/Passes/Reassociate`?

The true lit shell tests parallelize better.

lib/Transforms/Scalar/Reassociate.cpp
lizhengxing: @llvm-beanz This is from the original upstream change. Do we still need to change them in this PR?

lib/Transforms/Scalar/Reassociate.cpp
lizhengxing: @llvm-beanz This is from the original upstream change as well. Do we still need to change them in this PR?

llvm-beanz: Oh... Sorry, it is tough to tell what is coming in with this change vs what is from the other change.

lizhengxing: Yes, I guard most of the upstream change on a flag as David suggests. It's necessary but makes it hard to tell what's coming from upstream change.

lib/Transforms/Scalar/Reassociate.cpp
lizhengxing: Done.

tools/clang/test/HLSLFileCheck/passes/llvm/reassociate/reassociation-flag.hlsl
lizhengxing: Done.

lib/Transforms/Scalar/Reassociate.cpp
lizhengxing: @llvm-beanz I updated the PR. I use the `// HLSL Change` instead of `// HLSL Change - begin` and `// HLSL Change - end` for the upstream change. I hope this would help you review this PR.

llvm-beanz (approved):

tex3d (approved):

PR #6626, opened 2024-05-15
More aggressive reassociations

lizhengxing:
Although DXC applied the upstream change, Reassociate: add global reassociation algorithm (https://github.com/llvm/llvm-project/commit/b8a330c) in this PR (https://github.com/microsoft/DirectXShaderCompiler/pull/6598), it still might overlook some obvious common factors.

One case has been observed is:
```
  %Float4_0   = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle %1, i32 1)
  %Float4_0.w = extractvalue %dx.types.CBufRet.f32 %Float4_0, 3
  %Float2_0   = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle %1, i32 0)
  %Float2_0.y = extractvalue %dx.types.CBufRet.f32 %Float2_0, 1

  /* %Float4_1 is redundant with %Float4_0 since they invokes cbufferLoadLegacy with same parameters */
  %Float4_1   = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle %1, i32 1)

  /* %Float4_1.w is redundant with %Float4_0.w */
  %Float4_1.w = extractvalue %dx.types.CBufRet.f32 %Float4_1, 3 

  /* %Float2_1 is redundant with %Float2_0 since they invokes cbufferLoadLegacy with same parameters */
  %Float2_1   = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle %1, i32 0)

  /* %Float2_1.y is redundant with %Float2_0.y */
  %Float2_1.y = extractvalue %dx.types.CBufRet.f32 %Float2_1, 1

  ....
  %11 = fmul fast float %Float4_0.w, %10
  %12 = fmul fast float %11, %Float2_0.y
  ....
  %14 = fmul fast float %Float4_1.w, %13
  %15 = fmul fast float %14, %Float2_1.y 

(%Float4_0.w * %Float2_0.y) equals to (%Float4_1.w * %Float2_1.y), they should be reassociated to a common factor

```
The upstream change can't identify this common factor because DXC doesn't know (%Float4_0.w, %Float4_1.w) and (%Float2_0.y, %Float2_1.y) are redundant when running Reassociate pass. Those redundancies will be eliminated in GVN pass.

For DXC can identify more common factors, this PR will aggressively run Reassociate pass again after GVN pass and then run GVN pass again to remove the redundancies generared in this run of Reassociate pass.

Changing the order of floating point operations causes the precision issue. In case some shaders get unexpected results due to this PR, use "-opt-disable aggressive-reassociation" to disable this PR and roll back.

This is part 3 of the fix for https://github.com/microsoft/DirectXShaderCompiler/issues/6593.

### DIFF
--- a/lib/Transforms/IPO/PassManagerBuilder.cpp
+++ b/lib/Transforms/IPO/PassManagerBuilder.cpp
@@ -504,6 +504,16 @@ void PassManagerBuilder::populateModulePassManager(
   }
 
   // HLSL Change Begins.
+  {
+    // Run reassociate pass again after GVN since GVN will expose more
+    // opportunities for reassociation.
+    if (HLSLEnableAggressiveReassociation) {
+      MPM.add(createReassociatePass(true)); // Reassociate expressions
+      if (EnableGVN)
+        MPM.add(createGVNPass(DisableGVNLoadPRE)); // Remove redundancies
+    }
+  }
+
   // Use value numbering to figure out if regions are equivalent, and branch to only one.
   MPM.add(createDxilSimpleGVNEliminateRegionPass());
   // HLSL don't allow memcpy and memset.
--- a/tools/clang/test/DXC/Passes/reassociate/fp-common-factor-optimization.hlsl
+++ b/tools/clang/test/DXC/Passes/reassociate/fp-common-factor-optimization.hlsl
@@ -1,0 +1,69 @@
+// RUN: %dxc -T cs_6_3 -E cs_main %s -opt-enable aggressive-reassociation | FileCheck %s -check-prefixes=CHECK,COMMON_FACTOR
+// RUN: %dxc -T cs_6_3 -E cs_main %s -opt-disable aggressive-reassociation | FileCheck %s -check-prefixes=CHECK,NO_COMMON_FACTOR
+
+// Make sure DXC recognize the common factor and generate optimized dxils.
+
+// CHECK: [[TMP_CBUF:%.*]] = call %dx.types.Handle @dx.op.createHandle(i32 57, i8 2, i32 0, i32 10, i1 false)
+
+// CHECK: [[ELEM_1:%.*]] = call %dx.types.CBufRet.f32 @dx.op.cbufferLoadLegacy.f32(i32 59, %dx.types.Handle [[TMP_CBUF]], i32 1)
+// CHECK: [[FACTOR_SRC1:%.*]] = extractvalue %dx.types.CBufRet.f32 [[ELEM_1]], 2
+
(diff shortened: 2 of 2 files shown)


dmpots:
In your message you say

>  DXC doesn't know (%3, %7) and (%7, %9) are redundant 

How are %7 and %9 redundant here? Should it be %5 and %9?

It would be easier to read this if you gave them names I think.

dmpots (changes requested):
We should look at the performance impact of this change before committing it.

lizhengxing:
> In your message you say
> 
> > DXC doesn't know (%3, %7) and (%7, %9) are redundant
> 
> How are %7 and %9 redundant here? Should it be %5 and %9?
> 
> It would be easier to read this if you gave them names I think.

Yes, %5 and %9 are redundant. I simplified the dxils and renamed the virtual register number. I'll update the description.

lizhengxing:
> In your message you say
> 
> > DXC doesn't know (%3, %7) and (%7, %9) are redundant
> 
> How are %7 and %9 redundant here? Should it be %5 and %9?
> 
> It would be easier to read this if you gave them names I think.

Done.

lib/Transforms/IPO/PassManagerBuilder.cpp
```diff
@@ -504,6 +504,17 @@ void PassManagerBuilder::populateModulePassManager(
...
 
   // HLSL Change Begins.
+  {
+    // Run reassociate pass again after GVN since GVN will expose more
+    // opportunities for reassociation.
+    if (HLSLEnableAggressiveReassociation) {
+      MPM.add(createReassociatePass(true)); // Reassociate expressions
+      if (EnableGVN) {
+        MPM.add(createGVNPass(DisableGVNLoadPRE)); // Remove redundancies
+      }
```
llvm-beanz, line 514: nit:
```suggestion
      if (EnableGVN)
        MPM.add(createGVNPass(DisableGVNLoadPRE)); // Remove redundancies
```

See: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

lizhengxing: Done.

tex3d (approved):

lizhengxing:
> We should look at the performance impact of this change before committing it.

Overall, compared to the upstream change (https://github.com/microsoft/DirectXShaderCompiler/pull/6598), this change can reduce ALU in ~37% of shaders.  The occupancy of a small number of shaders are improved (~1%). There were an equal number of regressions and improvements.

Overall, this change looks to be positive.

dmpots (approved):
LGTM

PR #6628, opened 2024-05-15
Fix use-after-free in SimplifyCFG

amaiorano:
When SimplifySwitchOnSelect calls SimplifyTerminatorOnSelect, it holds onto the select's condition value to use for the conditional branch it replaces the switch with. When removing the switch's unused predecessors, it must make sure not to delete PHIs in case one of them is used by the condition value, otherwise the condition value itself may get deleted, resulting in an use-after-free.

Note that this was fixed in LLVM as well:
https://github.com/llvm/llvm-project/commit/dc3b67b4cad5c18a687edfabd50779c3c656c620


sudonatalie (approved):
LGTM!

tools/clang/test/DXC/Passes/SimplifyCFG/simplifycfg-uaf-select-condition.hlsl
```diff
@@ -0,0 +1,59 @@
+// RUN: %dxc -T cs_6_6 %s | FileCheck %s
+
+// This was resulting in an ASAN use-after-free in SimplifyCFG
+// during SimplifyTerminatorOnSelect because it was deleting
+// a PHI node with an input value that the pass later emits
+// (the select condition value).
+// We simply make sure this compiles here, but see
+// simplifycfg-uaf-select-condition.ll for the IR-level test.
```
llvm-beanz, line 8: Instead of adding a test that doesn't really verify correctness of the output, can we put this source in a comment in the IR test?

amaiorano: Done.

llvm-beanz (approved):

Issue #6616, opened 2024-05-13
[SPIR-V] Address "unsupported" tests


sudonatalie:
As part of the Lit migration, a number of SPIR-V backend tests were found that were not actually being run, so they were placed in `tools/clang/test/CodeGenSPIRV/unsupported`. We should address these tests individually and either fix them up to be run and passing or remove them if they're not relevant.

PR #6632, opened 2024-05-16
[SPIR-V] Remove bad test

sudonatalie:
This test is marked as unsupported (ignored), but it currently fails to compile both for SPIR-V and DXIL with the same Sema error: "cannot implicitly convert from 'SecondStruct' to 'FirstStruct'". If it would have succeeded for SPIR-V at some point in the past it's no longer valid anyways, so removing it.

Related to #6616


Keenuts (approved):

s-perron (approved):
Looking at the history a little, v1.6 allowed flat conversions between two struct types, but that was removed in v1.7. It is fair to remove this test. When I rewrote flat conversion in spir-v, I added tests to cover as many cases as I could think of. I see no value in trying to update this test.

PR #6635, opened 2024-05-17
Fix aligment for empty structs

### LINKED ISSUE #2882: [SPIRV] Empty struct in cbuffer causes compilation to fail
If an empty struct is declared in a cbuffer (or a nested struct inside) the compilation seems to break trying to align members properly to their alignment requirements around that struct:

```
struct empty_struct { };

cbuffer cb
{
	float3 a;
	empty_struct b;
	float2 c;

	float4 test;
};

float4 main() : SV_TARGET
{
	return test;
}
```

When compiling with latest dxc (`dxcompiler.dll: 1.6 - 1.5.0.2653 (f0d4419f)`) I get this error:

```
fatal error: generated SPIR-V is invalid: Structure id 4 decorated as Block for variable in Uniform storage class must follow relaxed uniform buffer layout rules: member 1 at offset 12 is not aligned to 16
  %type_cb = OpTypeStruct %v3float %empty_struct %v2float %v4float

```

With the offset of 12 coming wrongly from the `float2 c` trying to pack directly after `float3 a`. If I allow scalar layout with `-fvk-use-scalar-layout` then it doesn't complain since a tight packing is valid.

s-perron:
We have a special case to that the the size and alignment for an empty
struct is `{1,0}`. However that is not correct. See
https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#interfaces-alignment-requirements.

> An empty structure has a base alignment equal to the size of the smallest
scalar type permitted by the capabilities declared in the SPIR-V module. (e.g.,
for a 1 byte aligned empty struct in the StorageBuffer storage class,
StorageBuffer8BitAccess or UniformAndStorageBuffer8BitAccess must be declared
in the SPIR-V module.

I'm not 100% sure how DXC handle this minimum alignment, but I figured I
would inialize the alignment to 1. If there are not members, then it
will remain 1, and I would let the rest of the logic happen. No special
case.

Fixes #2882

### DIFF
--- a/tools/clang/test/CodeGenSPIRV/empty_struct.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/empty_struct.hlsl
@@ -1,0 +1,30 @@
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv | FileCheck %s -check-prefix=RELAXED
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv -fvk-use-gl-layout | FileCheck %s -check-prefix=RELAXED
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv -fvk-use-scalar-layout | FileCheck %s -check-prefix=SCALAR
+
+struct empty_struct { };
+
+
+// RELAXED: OpMemberDecorate %type_cb 0 Offset 0
+// RELAXED: OpMemberDecorate %type_cb 1 Offset 16
+// RELAXED: OpMemberDecorate %type_cb 2 Offset 16
+// RELAXED: OpMemberDecorate %type_cb 3 Offset 32
+
+// SCALAR: OpMemberDecorate %type_cb 0 Offset 0
+// SCALAR: OpMemberDecorate %type_cb 1 Offset 12
+// SCALAR: OpMemberDecorate %type_cb 2 Offset 12
+// SCALAR: OpMemberDecorate %type_cb 3 Offset 20
+
+cbuffer cb
+{
+	float3 a;
+	empty_struct b;
+	float2 c;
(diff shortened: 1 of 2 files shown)


tools/clang/test/CodeGenSPIRV/empty_struct.hlsl
```diff
@@ -0,0 +1,30 @@
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv | FileCheck %s -check-prefix=RELAXED
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv -fvk-use-gl-layout | FileCheck %s -check-prefix=RELAXED
+// RUN: %dxc -T ps_6_0 -E main -fcgl  %s -spirv -fvk-use-scalar-layout | FileCheck %s -check-prefix=SCALAR
+
+struct empty_struct { };
+
+
+// RELAXED: OpMemberDecorate %type_cb 0 Offset 0
```
Keenuts, line 8: Shall the capabilities be checked?
Also, do you know if the capability trimming pass handles this correctly? (as in don't remove the mandatory capacity if one alignment requires it?)
(IIRC related capabilities are not trimmed, but in case)

Keenuts: I was referring to the 2 mentioned capabilities. DXC don't generate those, but SPIRV-Tools can handle the 16bit version. I tried your PR on BigWheels with some empty structs, 16bit types etc to force the use of those capabilities, and seems like no layer complained.
Thanks!

Keenuts (approved):

sudonatalie (approved):

PR #6639, opened 2024-05-21
Add error when pass not found by Extract IR script

pow2clk:
When ExtractIRForPassTest fails to find a pass due to a typo in the name or other reason, it would manifest as trying to join a None value. This wasn't too helpful to diagnose the problem. Now it tests for None and produces an error and a suggestion to identify the problem.

### DIFF
--- a/utils/hct/ExtractIRForPassTest.py
+++ b/utils/hct/ExtractIRForPassTest.py
@@ -104,7 +104,7 @@ def main(args):
         # 1. Gets the pass list for an HLSL compilation using -Odump
         cmd = ["dxc", "/Odump", args.hlsl_file] + args.compilation_options
         # print(cmd)
-        all_passes = subprocess.check_output(cmd, text=True)
+        all_passes = subprocess.check_output(cmd, universal_newlines=True)
         all_passes = all_passes.splitlines()
 
         # 2. Compiles HLSL with -fcgl and outputs to intermediate IR
@@ -124,6 +124,9 @@ def main(args):
         passes_before, passes_after = SplitAtPass(
             all_passes, args.desired_pass, args.invocation
         )
+        if passes_after is None:
+          sys.exit("\nPass {} not found!\nUse dxc with -Odump to see the available passes.\n".format(args.desired_pass))
+
         print(
             "\nPasses before: {}\n\nRemaining passes: {}".format(
                 " ".join(passes_before), " ".join(passes_after)


llvm-beanz (approved):

utils/hct/ExtractIRForPassTest.py
```diff
@@ -104,7 +104,7 @@ def main(args):
         # 1. Gets the pass list for an HLSL compilation using -Odump
         cmd = ["dxc", "/Odump", args.hlsl_file] + args.compilation_options
         # print(cmd)
-        all_passes = subprocess.check_output(cmd, text=True)
+        all_passes = subprocess.check_output(cmd, universal_newlines=True)
```
bob80905, line 107: Apparently text and universal_newlines are aliases of each other. Any specific reason you opted to switch to universal_newlines?

utils/hct/ExtractIRForPassTest.py
```diff
@@ -124,6 +124,9 @@ def main(args):
         passes_before, passes_after = SplitAtPass(
             all_passes, args.desired_pass, args.invocation
         )
+        if passes_after is None:
```
bob80905, line 127: an exception is raised is "after" is None, and it's checked inside the SplitAtPass function. So, if you wanted to accompany this exception with an error message, wouldn't it be more proper to surround the SplitAtPass invocation with a try catch block?

bob80905: Otherwise, you might as well move this sys.exit invocation into the SplitAtPass function itself.

utils/hct/ExtractIRForPassTest.py
pow2clk: I wasn't sure whether to leave this in and I did fail to mention it in the description. The reason is just that I happened to have an older version of python on a system I was using and the `text` alias was added later, so I thought I'd make it more available

utils/hct/ExtractIRForPassTest.py
pow2clk: I don't think I understand your first sentence here. I didn't want to add another try block. In practice, you will see the generic catch that says "something went wrong!" and then you'll see the specific error. That gave me all I wanted.

bob80905: Yeah, this is a nit. I was just noting that instead of `raise Exception(f"no such pass: {pass_name}")` and then going into your new if condition, you could replace the exception with `sys.exit("\nPass {} not found!\nUse dxc with -Odump to see the available passes.\n".format(args.desired_pass))`
Would that be unwanted?

tex3d: Dang, I already had an alternate change to throw a useful error in this case [here](https://github.com/tex3d/DirectXShaderCompiler/commit/302292df34b6ecb50b8c1c768c9c1bc5afbd5a79), but hadn't put up a PR.

tex3d: Wait, that's not true, I did have a PR here #6325.

tex3d: Looking at a prior change which introduced an exception when the pass wasn't found, it appears this change wouldn't be able to work, since `SplitAtPass` [would have raised an exception](https://github.com/microsoft/DirectXShaderCompiler/pull/6639/files/0b42af4a0bb9e00f3b7e5c65ced49ff4f9c36034#diff-eb56348534bb00cde78f36c136a36ba14c96a1af924f5ccb8dd93464c5a9b131R89-R90) before returning `None` for `passes_after`.

That change went in 2 weeks ago from this PR: #6519.

tex3d (changes requested):
I don't think this change is good as-is because of another change already made recently in main.  I'd like to merge my older PR #6325 instead, since it adds more useful error text.

Perhaps the `universal_newlines` change could be useful to someone though.  Does the version of python that required this change meet the DXC build requirements?

PR #6643, opened 2024-05-22
Fix LoopDeletion incorrectly updating PHI with multiple duplicate inputs

amaiorano:
LoopDeletion was incorrectly updating PHI nodes in the target block when it had duplicate input edges. This happens, for example, when deleting a loop that uses a switch with multiple cases that exit the same way.

After determining that this was the bug, I found this fix in LLVM: https://reviews.llvm.org/D34516 and applied it here.


amaiorano:
Here's a Compiler Explorer link that shows DXC segfaulting: https://godbolt.org/z/3ejsMxc4d

llvm-beanz (approved):

sudonatalie (approved):

PR #6648, opened 2024-05-27
fix compile error for newer VS toolchain

atetubou:
This is to fix compile error like

```
../../third_party/dxc/utils/TableGen/AsmWriterEmitter.cpp(974): error C2666: '`anonymous-namespace'::IAPrinter::operator ==': overloaded functions have similar conversions
../../third_party/dxc/utils/TableGen/AsmWriterEmitter.cpp(739): note: could be 'bool `anonymous-namespace'::IAPrinter::operator ==(const `anonymous-namespace'::IAPrinter &)'
../../third_party/dxc/utils/TableGen/AsmWriterEmitter.cpp(739): note: or 'bool `anonymous-namespace'::IAPrinter::operator ==(const `anonymous-namespace'::IAPrinter &)' [synthesized expression 'y == x']
../../third_party/dxc/utils/TableGen/AsmWriterEmitter.cpp(974): note: while trying to match the argument list '(`anonymous-namespace'::IAPrinter, `anonymous-namespace'::IAPrinter)'
```
in dawn project which is updating to newer VS toolchain.

This imports fix from https://github.com/llvm/llvm-project/commit/4ab57cd9abb9535149e62c18461b56b94b09285a.

more context is in https://issues.chromium.org/issues/341890053#comment2

### DIFF
--- a/utils/TableGen/AsmWriterEmitter.cpp
+++ b/utils/TableGen/AsmWriterEmitter.cpp
@@ -736,13 +736,14 @@ public:
     O.indent(4) << '}';
   }
 
-  bool operator==(const IAPrinter &RHS) {
+  bool operator==(const IAPrinter &RHS) const {
     if (Conds.size() != RHS.Conds.size())
       return false;
 
     unsigned Idx = 0;
-    for (std::vector<std::string>::iterator
-           I = Conds.begin(), E = Conds.end(); I != E; ++I)
+    for (std::vector<std::string>::const_iterator I = Conds.begin(),
+                                                  E = Conds.end();
+         I != E; ++I)
       if (*I != RHS.Conds[Idx++])
         return false;
 
@@ -1073,7 +1074,7 @@ void AsmWriterEmitter::EmitPrintAliasInstruction(raw_ostream &O) {
         << "    break;\n";
     }
     O << "  }\n";
-  }    
+  }
   O << "}\n\n";
 
   if (!MCOpPredicates.empty()) {
--- a/utils/TableGen/AsmWriterEmitter.cpp
+++ b/utils/TableGen/AsmWriterEmitter.cpp
@@ -736,6 +736,7 @@ public:
     O.indent(4) << '}';
   }
 
+  // HLSL Change begin
   bool operator==(const IAPrinter &RHS) const {
     if (Conds.size() != RHS.Conds.size())
       return false;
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Thanks!

For better tracking of HLSL-specific changes to LLVM code, could we add the following comment tags to mark the modified section:

```cpp
// HLSL Change begin
... (your modified HLSL code)
// HLSL Change end
```
?

This aligns with this repository existing practice:
https://github.com/search?q=repo%3Amicrosoft%2FDirectXShaderCompiler+%22HLSL+change%22&type=code

I am waiting on this patch to [roll new msvc toolchain version for Dawn.](https://dawn-review.googlesource.com/c/dawn/+/189780). Thanks for making it!

Non OWNER LGTM otherwise.

atetubou:
@ArthurSonzogni thank you for your review, I added the comment.

utils/TableGen/AsmWriterEmitter.cpp
```diff
@@ -736,18 +736,21 @@ class IAPrinter {
...
-  bool operator==(const IAPrinter &RHS) {
+  // HLSL Change begin
+  bool operator==(const IAPrinter &RHS) const {
     if (Conds.size() != RHS.Conds.size())
       return false;
 
     unsigned Idx = 0;
-    for (std::vector<std::string>::iterator
-           I = Conds.begin(), E = Conds.end(); I != E; ++I)
+    for (std::vector<std::string>::const_iterator I = Conds.begin(),
```
pow2clk, line 745: Much as I personally find the `auto` keyword annoying, I'd prefer that we keep things as consistent with upstream LLVM as possible and [this change](https://github.com/llvm/llvm-project/commit/4ab57cd9abb9535149e62c18461b56b94b09285a) opted to just used `auto` instead of specifying a const_iterator. I don't think it's necessary to take the other aspects of that change though. By making it consistent with upstream, you wouldn't need the "HLSL Change" comments either

pow2clk (approved):
Thanks for the contribution! Nothing wrong with this really, but I thought I'd make an alternative suggestion that would keep consistent with LLVM.

utils/TableGen/AsmWriterEmitter.cpp
atetubou: Thanks, I updated to align with LLVM change.

amaiorano (approved):

Issue #3950, opened 2021-09-15
[SPIR-V] assigning default value to an external global silently fails without warning


EeroMutka:
The following compiles successfully when generating spirv, but `groupshared uint testing = 0;` will NOT be always initialized to zero. If I switch the compile mode to DXIL, I will get the warning: `Initializer of external global will be ignored`

http://shader-playground.timjones.io/46e4c2aa215eef693bea5e82cc6c3f8a

```
groupshared uint testing = 0;

[numthreads(1, 1, 1)]
void cs_main(uint local_thread_id_flat : SV_GroupIndex) {
    if (testing > 0) {
        printf("aaa");
    }
}
```

jaebaek:
@Derzo Thank you for reporting this issue. We will take a look.

danginsburg:
I also just hit this bug, it works in the HLSL front-end of glslang.  For example:

```
groupshared uint gs_nVisibleLightCount = 0;

RWTexture2D<float> OutputTexture;
[numthreads(8,8, 1)]
void CSMain( uint3 vThreadId : SV_DispatchThreadID, uint3 vGroupThreadId : SV_GroupThreadID, uint3 vGroupId : SV_GroupID, uint nGroupIndex: SV_GroupIndex )
{
	OutputTexture[vThreadId.xy].r = gs_nVisibleLightCount;
}
```

will generate:
```
               OpStore %gs_nVisibleLightCount %uint_0
```

If I compile the shader with DXC it's identical if the initializer is present or not.  See example: https://shader-playground.timjones.io/28ae2cb61338a3c78fb9fb5c2657fea6

s-perron:
We should issue a warning as is done for DXIL. Generating a store in the entry point is more complex. To avoid a data race, the store will have to be atomic (with lots of contention) or we will have to pick a single active thread to do the store. In either case, we will have to add a barrier.

This is a lot of overhead that could be unexpected from the user. Since most users expect the DXIL behaviour from HLSL, we will do the same.

I think it is worth considering if this should become an error.

PR #6650, opened 2024-05-27
[SPIR-V] Add support for D3D12 descriptor heaps

### LINKED ISSUE #4255: [SPIR-V][Feature Request][SM6.6] ResourceDescriptorHeap support/emulation
Currently, attempting to assign a value indexed from `ResourceDescriptorHeap` results in a ICE (segfault) if `-spirv` is passed.

I'm writing this issue as sort of a placeholder to describe how this could be at least emulated in SPIR-V, at least until more direct support is available.

Consider a barebones shader like so:

```hlsl
[numthreads(32, 1, 1)]
void CSMain()
{
    Texture2D t = ResourceDescriptorHeap[0];
}
```

This shader fails to compile if `-spirv -E CSMain` are passed to `dxc`. However, one way this could be emulated is by creating "shadow" bindings for every resource type in a distinct space when `-spirv` is set. e.g. the compiler could transform the snippet above to:

```hlsl
[vk::binding(1000)]
Texture2D<float> __texture2d_float[];
[vk::binding(1000)]
Texture2D<float2> __texture2d_float2[];
[vk::binding(1000)]
Texture2D<float3> __texture2d_float3[];
// etc. for every resource type present in the shader. The set/binding for these emitted arrays can be set with a 
// command line argument but defaults to some high number
// The float and float2 texture types here are just illustrative.

[numthreads(32, 1, 1)]
void CSMain()
{
    Texture2D t = __texture2d_float2[0];
}
```

In the above transformation, the user could still write `ResourceDescriptorHeap` without thinking about which aliased descriptor array matched the resource type. The compiler does the following work:

- Analyzes which types are present in the shader
- Silently emits a binding at a deterministic set/binding slot corresponding to that type
- Mutates `ResourceDescriptorHeap` to reference the correctly typed descriptor array

To use the shader, the user is responsible for creating the descriptor set that binds the descriptors themselves. However, the user is still able to access all resource types in a single unified heap (and more importantly, author shaders that work with both DX12 + vulkan using the same binding style).

Keenuts:
This commit adds support for Sampler/Resource descriptor heaps in DXC. Support for those heaps on the SPIR-V side requires no other extension than SPV_EXT_descriptor_indexing.
On the Vulkan side, the VK_EXT_mutable_descriptor_type will be required as multiple descriptor types must be allowed on the same binding.

When loading a type from a heap, DXC generates a new OpRuntimeArray of the correct type, and binds it to `set=0, binding=<BindingNumberOfTheHeap>`.
This means multiple OpRuntimeArrays will share the same binding. This is why VK_EXT_mutable_descriptor_type is required.

This implementation uses at most 3 bindings:
 - N OpRuntimeArray as binding A for the ResourceDescriptorHeap
 - N OpRuntimeArray as binding B for the SamplerDescriptorHeap
 - 1 OpRuntimeArray %counter_type for the ResourceDescriptorHeap counters.

The bindings are only allocated if used. If only the SamplerDescriptorHeap is used, a single binding is required.

The binding allocation logic is:
 1. allocate bindings for every resources, excluding heaps.
 2. If ResourceDescriptorHeap is used, find the first unused binding in set=0 and use it.
 3. Same for the SamplerDescriptorHeap
 4. Same for the counters.

UAV counters are not always created, only if used.
When used, they are stored in an OpRuntimeArray. The index of a counter in that array
is equal to the index of the associated resource in its own OpRuntimeArray.

```hlsl
        RWStructuredBuffer a = ResourceDescriptorHeap[2];
        a.IncrementCounter();
        // buffer in descriptorSet 0, binding 0, OpRuntimeArray[index=2]
        // counter in descriptorSet 0, binding 1, OpRuntimeArray[index=2]
```

As-is, this PR doesn't allow resource heaps to alias regular resources, or to overlap.
A follow-up PR will add 3 flags to override each binding/set pairs:
 - 'fvk-bind-resource-heap <set> <binding>'
 - 'fvk-bind-sampler-heap <set> <binding>'
 - 'fvk-bind-counter-heap <set> <binding>'


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5899,6 +5927,32 @@ SpirvEmitter::doCXXOperatorCallExpr(const CXXOperatorCallExpr *expr,
...
 
+  SourceRange range =
+      (rangeOverride != SourceRange()) ? rangeOverride : expr->getSourceRange();
+
+  { // Handle ResourceDescriptorHeap and SamplerDescriptorHeap.
+    const Expr *baseExpr = nullptr;
+    const Expr *indexExpr = nullptr;
+    if (isDescriptorHeap(dyn_cast_or_null<CXXOperatorCallExpr>(expr), &baseExpr,
+                         &indexExpr)) {
+      const Expr *ParentExpr = cast<Expr>(parentMap->getParent(expr));
```
s-perron, line 5938: We should add an assert that the parent is a cast expression.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -7138,6 +7189,28 @@ bool SpirvEmitter::isBufferTextureIndexing(const CXXOperatorCallExpr *indexExpr,
   return false;
 }
 
+bool SpirvEmitter::isDescriptorHeap(const CXXOperatorCallExpr *indexExpr,
```
s-perron, line 7192: I prefer it when a function does 1 thing. I know the spir-v backend has lots of function like this, but could be split this into `isDescritorHeap`, which return true of false, as the name suggests. Then add another function to get the base and index.

I find it hard to read code when a function with a name like this has other return values.

docs/SPIR-V.rst
```diff
@@ -1923,6 +1940,65 @@ Example 3: (compiled with ``-fvk-bind-globals 2 1``)
...
+The SPIR-V backend supported SM6.6 resource heaps, using 2 extensions:
+- `SPV_EXT_descriptor_indexing`
+- `VK_EXT_mutable_descriptor_type`
+
+The descriptor 0 is used to address resources in both the sampler and resource
+heaps. The binding index is the index used to subscript in the heap.
+
+.. code:: hlsl
+   Texture2D tex = ResourceDescriptorHeap[10];
+   // tex is in the descriptor set 0, binding 10.
```
s-perron, line 1955: Vulkan binding is different than DX. In Vulkan, an array of resources do not take up a binding for each element of the array. As long as the option `-fspv-flatten-resource-arrays` is not used, the entire descriptor heap should only use a single binding in Vulkan. I don't know if that changes your ideas on how to make this work with implicit bindings.

docs/SPIR-V.rst
```diff
@@ -1923,6 +1940,65 @@ Example 3: (compiled with ``-fvk-bind-globals 2 1``)
...
 
+ResourceDescriptorHeaps & SamplerDescriptorHeaps
+------------------------------------------------
+
+The SPIR-V backend supported SM6.6 resource heaps, using 2 extensions:
+- `SPV_EXT_descriptor_indexing`
+- `VK_EXT_mutable_descriptor_type`
+
+The descriptor 0 is used to address resources in both the sampler and resource
+heaps. The binding index is the index used to subscript in the heap.
```
s-perron, line 1951: I don't think that the sampler and resource descriptor heaps should have the same binding. In DirectX, samples and other resources have to be in different heaps. If someone is trying to port their DX application, this will simply cause another change they will have to make to their code.

s-perron (commented):
The implementation looks pretty good. But we need to work on how the bindings and descriptor sets are assigned. There does seem to be some misunderstanding. We also need to consider the interaction with `-fspv-flatten-resource-arrays`.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -2396,6 +2425,29 @@ bool DeclResultIdMapper::decorateResourceBindings() {
...
+                  "implicit bindings is disallowed.",
+                  var.getSourceLocation());
+        return false;
+      }
+
+      const uint32_t set = bindingAttr->getSet();
+      const uint32_t binding = var.isCounter() ? counterAttr->getBinding()
+                                               : bindingAttr->getBinding();
+      spvBuilder.decorateDSetBinding(var.getSpirvInstr(), set, binding);
+    }
```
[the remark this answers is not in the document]
Keenuts: This code is now obsolete. Resolving for now.

Keenuts:
Resolving comments around bindings as the design changed since the last version.
Binding logic to implement now:

- All other resources follow the same rules as today (implicit & explicit bindings)
- Then, we find the 3 first unused bindings (not necessarily consecutive, starting at set 0, binding 0)
- Take 1 binding for resource heaps, 1 for sample heaps, and 1 for UAV counters.
- index in the binding for UAV counter matches the index of the associated UAV in the resource heap binding.
- 3 flags will be added: `-fvk-bind-[resource,sampler,counter]-heap <set> <binding>`, which allow the user to fix the binding/set pair, allowing aliasing resources.

Keenuts:
@s-perron I think this is ready for another review. I modified the PR description and the documentation to match the new binding allocation behavior.
This PR does not include the new flags as I find it already quite big.

s-perron (commented):
Looks good.

I have one more case to consider. I expect we would want to disallow passing the heaps as function parameters. As far as I know, it cannot be done without templates because we do not even know the type.

docs/SPIR-V.rst
```diff
@@ -1923,6 +1939,81 @@ Example 3: (compiled with ``-fvk-bind-globals 2 1``)
...
+
+   RWStructuredBuffer<int> tmp = ResourceDescriptorHeap[0];
+   tmp.IncrementCounter();
+   // tmp is in the descriptor set 0, binding 0.
+   // tmp.counter is in the descriptor set 0, binding 2
+
+.. code:: hlsl
+   RWStructuredBuffer buffer = ResourceDescriptorHeap[2];
+   // buffer is in the descriptor set 0, binding 2.
+   // the counter is in the descriptor set 1, binding 2.
```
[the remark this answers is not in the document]
Keenuts: Wrong example, leftover from the old version.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -2668,9 +2688,61 @@ bool DeclResultIdMapper::decorateResourceBindings() {
...
+      continue;
+
+    const bool isResourceHeap = isResourceDescriptorHeap(decl->getType());
+    const bool isSamplerHeap = isSamplerDescriptorHeap(decl->getType());
+
+    assert(!(var.isCounter() && isSamplerHeap));
+
+    hasResource |= isResourceHeap;
+    hasSamplers |= isSamplerHeap;
+    hasCounters |= (isResourceHeap || isSamplerHeap) && var.isCounter();
```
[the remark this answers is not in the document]
Keenuts: Yes, changed to only check resourceHeap && isCounter (the assert above checks the invalid case sampler & counter)

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -2668,9 +2688,61 @@ bool DeclResultIdMapper::decorateResourceBindings() {
...
   }
 
+  decorateResourceHeapsBindings(bindingSet);
   return true;
 }
 
+void DeclResultIdMapper::decorateResourceHeapsBindings(BindingSet &bindingSet) {
+  bool hasResource = false;
+  bool hasSamplers = false;
+  bool hasCounters = false;
```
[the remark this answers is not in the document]
Keenuts: There is more than 1 variable per heap: there is one variable per heap per type:
Texture2D + Texture3D -> 2 variables sharing the same binding.
I could add a vector and store the pointers, but do we really have that many resources to avoid iterating on those twice?

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -2668,9 +2688,61 @@ bool DeclResultIdMapper::decorateResourceBindings() {
...
+      continue;
+
+    if (isResourceDescriptorHeap(decl->getType()))
+      spvBuilder.decorateDSetBinding(var.getSpirvInstr(), /* set= */ 0,
+                                     var.isCounter() ? countersBinding
+                                                     : resourceBinding);
+    else if (isSamplerDescriptorHeap(decl->getType()))
+      spvBuilder.decorateDSetBinding(var.getSpirvInstr(), /* set= */ 0,
+                                     samplersBinding);
+  }
```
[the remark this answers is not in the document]
Keenuts: See above, there is potentially more than 1 variable per heap.

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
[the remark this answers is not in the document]
s-perron: Right, I forgot about that. This should be fine then. It might turn out to be a wash. What you currently have might be the simplest code. I'll leave it for you to decide what you want to do.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -7164,6 +7213,35 @@ bool SpirvEmitter::isBufferTextureIndexing(const CXXOperatorCallExpr *indexExpr,
...
 }
 
+bool SpirvEmitter::isDescriptorHeap(const Expr *expr) {
+  const CXXOperatorCallExpr *operatorExpr = dyn_cast<CXXOperatorCallExpr>(expr);
+  if (!operatorExpr)
+    return false;
+
+  // Must be operator[]
+  if (operatorExpr->getOperator() != OverloadedOperatorKind::OO_Subscript)
+    return false;
```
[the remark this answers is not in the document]
Keenuts: This one is quite irritating.
DXC handles that by generating a pointer, then forcing legalization knowing "this pointer can be removed from the parameters, and fetched from this builtin".
In the SPIR-V case, it breaks.
I'd definitely want to prevent using this type as template parameter, seems way simpler than hacking around trying to make this work in our codegen.

Keenuts:
Added the error + test, and rebased on main.

s-perron (approved):

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -1269,12 +1272,17 @@ SpirvVariable *DeclResultIdMapper::createExternVar(const VarDecl *var) {
   if (storageClass == spv::StorageClass::Workgroup)
     return varInstr;
 
-  const auto *regAttr = getResourceBinding(var);
   const auto *bindingAttr = var->getAttr<VKBindingAttr>();
-  const auto *counterBindingAttr = var->getAttr<VKCounterBindingAttr>();
-
-  resourceVars.emplace_back(varInstr, var, loc, regAttr, bindingAttr,
-                            counterBindingAttr);
+  // FIXME: handle the flags here?
```
sudonatalie, line 1276: Did you mean to leave this here?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5925,6 +5958,33 @@ SpirvEmitter::doCXXOperatorCallExpr(const CXXOperatorCallExpr *expr,
...
+  SourceRange range =
+      (rangeOverride != SourceRange()) ? rangeOverride : expr->getSourceRange();
+
+  { // Handle ResourceDescriptorHeap and SamplerDescriptorHeap.
+    if (isDescriptorHeap(expr)) {
+      const Expr *baseExpr = nullptr;
+      const Expr *indexExpr = nullptr;
+      getDescriptorHeapOperands(expr, &baseExpr, &indexExpr);
+
+      const Expr *ParentExpr = cast<CastExpr>(parentMap->getParent(expr));
```
sudonatalie, line 5970: nit: inconsistent var capitalization

docs/SPIR-V.rst
```diff
@@ -1923,6 +1939,81 @@ Example 3: (compiled with ``-fvk-bind-globals 2 1``)
 Note that if the developer chooses to use this command line option, it is their
 responsibility to provide proper numbers and avoid binding overlaps.
 
+ResourceDescriptorHeaps & SamplerDescriptorHeaps
+------------------------------------------------
+
+The SPIR-V backend supported SM6.6 resource heaps, using 2 extensions:
+- `SPV_EXT_descriptor_indexing`
```
sudonatalie, line 1946: I don't see this added anywhere explicitly in this PR, so just checking, is it added unconditionally by the capability visitor? Assuming it is, can you add some CHECKs for it in tests after capability stripping?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
sudonatalie, line 5970: Is it possible for the parent to be nullptr? It looks like it's legal for descriptor heaps to be used in global vars, can you add a test for that?

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
Keenuts: no thanks!

docs/SPIR-V.rst
Keenuts: Its added by DXC as soon as you have a resource of type OpRuntimeArray. So this PR isn't the first to use this extension, it's just another case where this extension is required.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts: Good point. Added a test and fix for static global variables.
For external globals, it silently ignore the warning (while DXIL fails).
Seems like there is an issue for this issue: https://github.com/microsoft/DirectXShaderCompiler/issues/3950
Thanks!

sudonatalie (approved):
LGTM!

PR #6654, opened 2024-05-27
Add option to set the max id

### LINKED ISSUE #6636: [SPIR-V] Hitting max ID bound
In my application I often exceed kDefaultMaxIdBound, how to best deal with this? My understanding is that this is just the minimum bound required by the specification, so how to override it? As a quick test if I put a higher value in this constant it avoids my problem, but since the value is given by the spec, this seems not the correct solution

s-perron:
Vulkan implementation can have different limits on the maximum value
used as an id in a SPIR-V binary. SPIRV-Tools generall assumes this
limit is 0x3FFFFF because all implementations must support at least that
value for an id.  Since many implementations can support larger values,
the tools allows an option that will set a different limit. This commit
add an option to DXC to do the same.

Fixes #6636


Keenuts (approved):

s-perron:
FYI: @cassiebeckley

cassiebeckley (approved):

PR #6655, opened 2024-05-29
NFC: Comment, refactor, and test hlsl::RemoveUnstructuredLoopExits

dneto0:
Add a pass to run hlsl::RemoveUnstructuredLoopExits in isolation

  Example: opt -dxil-r-u-l-e a.ll -S

Add some basic tests.

No functional change to the pass itself.


lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
```diff
@@ -625,3 +716,48 @@ bool hlsl::RemoveUnstructuredLoopExits(
...
+};
+} // namespace
+
+char DxilRemoveUnstructuredLoopExits::ID;
+
+Pass *llvm::createDxilRemoveUnstructuredLoopExitsPass() {
+  return new DxilRemoveUnstructuredLoopExits();
+}
+
+INITIALIZE_PASS_BEGIN(DxilRemoveUnstructuredLoopExits, "dxil-r-u-l-e",
```
dmpots, line 758: Missed opportunity to call this `dxil-rules` :)

A minor point, but I think actually spelling it out (i.e. `dxil-remove-unstructured-loop-exits`) would be a bit clearer to read. It shouldn't be something we have to type often so while it is unwieldly I don't think it would be a big burden.

tools/clang/test/DXC/Passes/DxilRemoveUnstructuredLoopExits/exit-via-false-propagate-exit-cond.ll
```diff
@@ -0,0 +1,83 @@
+; RUN: opt %s -analyze -loops | FileCheck -check-prefix=LOOPBEFORE %s
```
dmpots, line 1: I think we might want these tests to live in [test/HLSL/passes](https://github.com/microsoft/DirectXShaderCompiler/tree/main/test/HLSL/passes). The test organization is still unclear to me, but I think for `opt` tests we generally want them out of the `tools/clang` directory. 

I think the `clang/test/DXC/Passes` is more for things that either need `dxc` or `dxopt` to run correctly.

dmpots (approved):
LGTM

lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
dneto0: Pretty quickly I started writing shell scripts to run various things; I agree.  :-)

tools/clang/test/DXC/Passes/DxilRemoveUnstructuredLoopExits/exit-via-false-propagate-exit-cond.ll
dneto0: I've moved the tests as suggested.

lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
dneto0: Done.

adam-yang (approved):
LGTM. Thanks for the cleanup!

Issue #6659, opened 2024-05-30
Non-determinism in Reassociate caused by address coincidences


lizhengxing:
**Description**
The upstream change, Reassociate: add global reassociation algorithm (https://github.com/llvm/llvm-project/commit/b8a330c42ab43879119cd3a305756d28aefe9fe6), was landed in DXC by this PR
https://github.com/microsoft/DirectXShaderCompiler/pull/6598.

However, this change could cause non-determinism results due to address coincidences for some large shaders. It looks like that upstream guy ran into the same issue as well and here's the upstream fix (https://github.com/llvm/llvm-project/commit/ef8761fd3b0f0b24b7e9347ad4ae686fc9ed8f94) for this non-determinism bug.

*summary of the upstream fix*:
Between building the pair map and querying it there are a few places that erase and create Values. It's rare but the address of these newly created Values is occasionally the same as a just-erased Value that we already have in the pair map. These coincidences should be accounted for to avoid non-determinism.

Unfortunately, this upstream fix couldn't fix the non-determinism bug in DXC. Because the fix relies on the WeakVH and the WeakVH related code in DXC is out of date.

**Steps to Reproduce**
Compile the large failing hlsl code with the same command line 10 times. At least one time, the compile result will be different.

**Actual Behavior**
Compiled with the same command line, the hash value of the shader will be different.

**Environment**
DXC version: 1.8 - 1.8.0.14576 (main, 381750e25)
Host Operating System: Windows 11

PR #6662, opened 2024-05-31
Emulate TrackingVH using WeakVH

lizhengxing:
This PR pulls the upstream change, Emulate TrackingVH using WeakVH (https://github.com/llvm/llvm-project/commit/8a6238201f015729a47691c62808a23ec8525096), into DXC.

  This frees up one slot in the HandleBaseKind enum, which I will use later to add a new kind of value handle.  The size of the HandleBaseKind enum is important because we store a HandleBaseKind in
  the low two bits of a (in the worst case) 4 byte aligned pointer.

  Reviewers: davide, chandlerc

  Subscribers: mcrosier, llvm-commits

  Differential Revision: https://reviews.llvm.org/D32634

This is part 2 of the fix for #6659.

### DIFF
--- a/include/llvm/IR/ValueHandle.h
+++ b/include/llvm/IR/ValueHandle.h
@@ -48,7 +48,6 @@ protected:
   enum HandleBaseKind {
     Assert,
     Callback,
-    Tracking,
     Weak
   };
 
@@ -164,6 +163,10 @@ public:
 
   operator Value*() const {
     return getValPtr();
+  }
+
+  bool pointsToAliveValue() const {
+    return ValueHandleBase::isValid(getValPtr());
   }
 };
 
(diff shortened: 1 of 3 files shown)


include/llvm/IR/ValueHandle.h
```diff
@@ -275,46 +274,49 @@ struct isPodLike<AssertingVH<T> > {
...
+/// Assigning a value to a TrackingVH is always allowed, even if said TrackingVH
+/// no longer points to a valid value.
+template <typename ValueTy> class TrackingVH {
+  WeakVH InnerHandle;
 
-    // Null is always ok.
-    if (!VP) return;
+public:
+  ValueTy *getValPtr() const {
+    // HLSL Change begin
```
dmpots, line 298: Why is this only needed in our version and not in the original upstream change?

lizhengxing: I don't know. But during the test, the pointer could be Nullptr and look at the original TrackingVH implementation. Nullptr should be OK.

lizhengxing: @dmpots See my updated comment. This is needed because of different code base.

dmpots: Great, thanks for the investigation and update comment!

dmpots (approved):

include/llvm/IR/ValueHandle.h
```diff
@@ -275,46 +274,65 @@ struct isPodLike<AssertingVH<T> > {
...
+    // However, the llvm code that DXC forked has the implicit code like:
+    // TrackingVH V = nullptr;
+    //
+    // It will invoke setValPtr(nullptr) and then getValPtr(nullptr). So pull in
+    // the original upstream change in DXC will always assert here for debug
+    // build even this code is valid.
+    //
+    // The original upstream change works because of another upstream change
+    // https://github.com/llvm/llvm-project/commit/70a6051ddfd5f04777f2bc42503bb11bc8f1723a
+    // cleaned up the problematic code in DXC already.
```
tex3d, line 311: The other upstream change seems pretty minimal and safe.  If it could be merged to DXC without problems, wouldn't that be preferable to diverging in this change?

tex3d (approved):
Seems preferable to merge required simple upstream change to avoid divergence in this one.  But we decide not to do that, this looks good to me.

PR #6663, opened 2024-05-31
Rename WeakVH to WeakTrackingVH; NFC

lizhengxing:
This PR pulls the upstream change, Rename WeakVH to WeakTrackingVH; NFC (https://github.com/llvm/llvm-project/commit/e6bca0eecbd31d9dd65040b212326330c9cf2655), into DXC.

>   I plan to use WeakVH to mean "nulls itself out on deletion, but does not track RAUW" in a subsequent commit.
> 
>   Reviewers: dblaikie, davide
> 
>   Reviewed By: davide
> 
>   Subscribers: arsenm, mehdi_amini, mcrosier, mzolotukhin, jfb, llvm-commits, nhaehnle
> 
>   Differential Revision: https://reviews.llvm.org/D32266

This is part 3 of the fix for #6659.

### DIFF
--- a/unittests/IR/ValueHandleTest.cpp
+++ b/unittests/IR/ValueHandleTest.cpp
@@ -100,7 +100,6 @@ TEST_F(ValueHandle, WeakTrackingVH_NullOnDeletion) {
   EXPECT_EQ(null_value, WVH_Recreated);
 }
 
-
 TEST_F(ValueHandle, AssertingVH_BasicOperation) {
   AssertingVH<CastInst> AVH(BitcastV.get());
   CastInst *implicit_to_exact_type = AVH;
--- a/unittests/IR/ValueHandleTest.cpp
+++ b/unittests/IR/ValueHandleTest.cpp
@@ -34,8 +34,8 @@ public:
   ConcreteCallbackVH(Value *V) : CallbackVH(V) {}
 };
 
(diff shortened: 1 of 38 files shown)


adam-yang (approved):

unittests/IR/ValueHandleTest.cpp
```diff
@@ -34,8 +34,8 @@ class ConcreteCallbackVH : public CallbackVH {
   ConcreteCallbackVH(Value *V) : CallbackVH(V) {}
 };
 
-TEST_F(ValueHandle, WeakVH_BasicOperation) {
-  WeakVH WVH(BitcastV.get());
+TEST_F(ValueHandle, WeakTrackingVH_BasicOperation) {
```
dmpots, line 37: Do we run these tests as part of the dxc unit tests?

dmpots (approved):

unittests/IR/ValueHandleTest.cpp
adam-yang: Doesn't seem like it :/ The test seems like a very nice .exe that we can just run though.

adam-yang: They're all passing right now, which is good. I'll open a separate issue to run them.

PR #6666, opened 2024-05-31
[llc/opt] Add an option to run all passes twice

### LINKED ISSUE #6659: Non-determinism in Reassociate caused by address coincidences

lizhengxing:
This PR pulls the following upstream changes into DXC:

[llc/opt] Add an option to run all passes twice (https://github.com/llvm/llvm-project/commit/04464cf7313a7c2873f2b8a9011a19f1685b44c1)
>   Lately, I have submitted a number of patches to fix bugs that only occurred when using the same pass manager to compile
>   multiple modules (generally these bugs are failure to reset some persistent state).
> 
>   Unfortunately I don't think there is currently a way to test that from the command line. This adds a very simple flag to both
>   llc and opt, under which the tools will simply re-run their respective
>   pass pipelines using the same pass manager on (a clone of the same module). Additionally, we verify that both outputs are
>   bitwise the same.
> 
>   Reviewers: yaron.keren

[opt] Fix sanitizer complaints about r254774 (https://github.com/llvm/llvm-project/commit/38707c45be6ea1c6f3d382efbee960151d7301c8)
> `Out` can be null if no output is requested, so move any access
> to it inside the conditional. Thanks to Justin Bogner for finding
> this.

This is for the test of this change (https://github.com/llvm/llvm-project/commit/ef8761fd3b0f0b24b7e9347ad4ae686fc9ed8f94) to fix #6659.

### DIFF
--- a/test/MC/ELF/empty-twice.ll
+++ b/test/MC/ELF/empty-twice.ll
@@ -1,0 +1,6 @@
+; Check that there is no persistent state in the ELF emitter that crashes us
+; when we try to reuse the pass manager
+; RUN: llc -compile-twice -filetype=obj %s -o -
+
+target datalayout = "e-p:32:32:32-i1:8:8-i8:8:8-i16:16:16-i32:32:32-i64:32:64-f32:32:32-f64:32:64-v64:64:64-v128:128:128-a0:0:64-f80:32:32-n8:16:32"
+target triple = "i386-pc-linux-gnu"
--- a/test/Object/opt-twice.ll
+++ b/test/Object/opt-twice.ll
@@ -1,0 +1,14 @@
+; The pass here doesn't matter (we use deadargelim), but test
+; that the -run-twice options exists, generates output, and
+; doesn't crash
+; RUN: opt -run-twice -deadargelim -S < %s | FileCheck %s
+
+; CHECK: define internal void @test
+define internal {} @test() {
+  ret {} undef
+}
+
+define void @caller() {
+  call {} @test()
+  ret void
+}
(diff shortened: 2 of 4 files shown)


adam-yang (approved):

dmpots (approved):
LGTM

PR #6668, opened 2024-05-31
Fix RemoveUnstructuredLoopExits when an exiting edge jumps out multiple levels of loops.

dneto0:
Before doing any major surgery on an exit from loop L, ensure that if an exit edge from L goes to block X, then X is in L's parent loop or no loop at all.

Add test cases:
- a reduced test case where the exiting block does not dominate its own loop latch.
- a reduced test case where the exiting block is the latch for its own loop. This reproduces the assert triggered by the original HLSL.
- the original HLSL that triggered this bug fix.
- the intermediate module from the original HLSL, taken just before the attempt to remove unstructured loop exits.

### DIFF
--- a/lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
+++ b/lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
@@ -130,10 +130,12 @@
 #include "llvm/IR/Constant.h"
 #include "llvm/IR/Instructions.h"
 #include "llvm/IR/IntrinsicInst.h"
+#include "llvm/IR/Module.h"
 #include "llvm/IR/Verifier.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/raw_ostream.h"
 #include "llvm/Transforms/Scalar.h"
+#include "llvm/Transforms/Utils/BasicBlockUtils.h"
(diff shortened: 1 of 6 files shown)


dneto0:
yeah, I need to fix this.  I thought I was running with ASAN but I wasn't.

dneto0:
> LGTM, thanks for the fix!
> 
> I appreciate the comments and level of detail. A totally optional suggestion would be to add some ascii art to the test or comment to clarify the situation. Sometimes a picture can really help the understanding.

Indeed pictures are great. I've added ASCII art depictions of the CFGs for the two reduced test cases. The "regression" case is a mess, and doesn't provide extra insight.

llvm-beanz (approved):

dneto0:
The failures are weird; I can't explain them by this change.

An error is caused by this warning (with warnings-as-error):
```
2024-06-05T21:40:26.6478030Z D:\a\1\s\projects\dxilconv\include\ShaderBinary\ShaderBinary.h(381,10): warning C4201: nonstandard extension used: nameless struct/union [D:\a\1\projects\dxilconv\lib\ShaderBinary\ShaderBinary.vcxproj]
```
The source at that spot has a pragma that appears to guard the warning.

I'll try a no-change amend commit to kick off the bots again.

dneto0:
Trying again by rebasing against latest main.

PR #6670, opened 2024-06-03
Fix crash in scalarrepl-param-hlsl when dynamically indexing a GEP of a constant indexed GEP

amaiorano:
When processing global values to determine when to flatten vectors, this pass was only checking the immdiate users of the value for non-dynamic indexing of the vector. But this would fail in the case of a dynamic indexed GEP of a constant indexed GEP (e.g. h[0][a]) because the first level GEP was constant indexed, but not the second. We fix this by
    checking the full User tree of the value in `hasDynamicVectorIndexing`.

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -1645,6 +1645,10 @@ bool hasDynamicVectorIndexing(Value *V) {
         }
       }
     }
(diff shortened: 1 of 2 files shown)


amaiorano:
Compiler Explorer link that shows the segfault (would assert in a debug build, and result in a use-after-free with ASAN enabled): https://godbolt.org/z/q4sGossh1

llvm-beanz (approved):

sudonatalie (approved):

PR #6675, opened 2024-06-06
CMake: Use find_package(Python3)

dneto0:
Fixes the CMake CMP0148 warning at configure time.

find_package(Python3) is supported since CMake 3.12, and DXC requires at least CMake 3.17


hekota (approved):

llvm-beanz (approved):

dneto0:
Rebased against main.  This should kick off the bots again, and hopefully the MSVC images are ok this time.

PR #6676, opened 2024-06-06
Remove unstructured loop exits: Don't introduce loops

dneto0:
Previously, if the latch exit was reachable from a different exit block for the loop, then the pass would introduce a loop involving that exit block and the latch exit.  This is unwanted and unaccounted for.

- Add a test for shared exits.
- Add a test for non-dedicated latch exit

### DIFF
--- a/lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
+++ b/lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
@@ -478,6 +478,19 @@ static bool RemoveUnstructuredLoopExitsIteration(BasicBlock *exiting_block,
   // So prevent this from happening by ensuring the latch exit is
   // "dedicated": the only branches to it come from inside the
   // loop, and hence not from X.
+  //
+  // The latch_exit block could have *multiple* branches to it from
+  // outside the loop.
+  //
+  // When the edge from latch to latch_exit is split, the local picture is:
+  //
+  //    latch --> middle --> tail
+  //
+  // where:
+  //  - Branches that used to go to latch_exit, from outside the loop, now
+  //    point to 'tail'.
+  //  - 'middle' is now an exit block for the loop, and its only incoming
+  //    edge is from latch.
   for (auto *pred : predecessors(latch_exit)) {
     if (!L->contains(pred)) {
       SplitEdge(latch, latch_exit, DT, LI);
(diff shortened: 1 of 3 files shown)


dneto0:
cc: @amaiorano

lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
```diff
@@ -550,6 +550,26 @@ static bool RemoveUnstructuredLoopExitsIteration(BasicBlock *exiting_block,
...
+  //
+  // Suppose this iteration successfully moves an exit block X until
+  // after the latch block.  It will do so by rewiring the CFG so
+  // the latch *exit* block will branch to X.  If the latch exit
+  // block is already reachable from X, then the rewiring will
+  // create an unwanted loop.
+  // So prevent this from happening by ensuring the latch exit is
+  // "dedicated": the only branches to it come from inside the
+  // loop, and hence not from X.
+  for (auto *pred : predecessors(latch_exit)) {
```
dmpots, line 565: I was wondering about the case where the `latch_exit` has multiple precessors outside the loop (if that is even possible) and if we needed to split those edges as well.

But then I realized that this is splitting the edge between the `latch` and the `latch_exit` and the new block on the edge will become the new `latch_exit` when we recalculate the exit blocks. Is that correct? I think it wouldn't hurt to mention in the comment if so.

dmpots (approved):
LGTM overall.

llvm-beanz (approved):

lib/Transforms/Scalar/DxilRemoveUnstructuredLoopExits.cpp
dneto0: +1   This code does handle that case: latch-exit block has multiple predecessors from outside the loop.

You're correct, once the edge is split we have this picture:

       latch --> middle --> tail

The 'middle' block is the new latch exit block; its only predecessor is 'latch'.

The edges (from blocks outside the loop) that used to go to latch exit block now go to tail.

dneto0: > I was wondering about the case where the latch_exit has multiple precessors outside the loop (if that is even possible) 

That depends on the situations that this routine is called in.  I don't know if the standard flow can make this happen.
I hit this case when I was making manual test cases.

dneto0: Added a comment as requested.

dneto0: Also, the 'shared-exit.ll' test exercises the case you were wondering about.

PR #6679, opened 2024-06-07
Fix instcombine overflow check inserting inst at wrong place

amaiorano:
When optimizing an overflow check of an add followed by a compare, the new instruction was being inserted at the compare, and the add removed. This produced invalid IR in cases where there were other uses of the former add between it and the compare. This fix makes sure to insert the new instruction at the old add location, rather than at the compare.

Note that this was also fixed in LLVM:
https://github.com/llvm/llvm-project/commit/6f5dca70ed1c030957a45ad91bd295921f17b18d

### DIFF
--- a/lib/Transforms/InstCombine/InstCombineCompares.cpp
+++ b/lib/Transforms/InstCombine/InstCombineCompares.cpp
@@ -2145,6 +2145,12 @@ bool InstCombiner::OptimizeOverflowCheck(OverflowCheckFlavor OCF, Value *LHS,
     return true;
   };
 
+  // If the overflow check was an add followed by a compare, the insertion point
+  // may be pointing to the compare.  We want to insert the new instructions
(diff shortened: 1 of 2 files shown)


lib/Transforms/InstCombine/InstCombineCompares.cpp
```diff
@@ -2145,6 +2145,12 @@ bool InstCombiner::OptimizeOverflowCheck(OverflowCheckFlavor OCF, Value *LHS,
     return true;
   };
 
+  // If the overflow check was an add followed by a compare, the insertion point
```
sudonatalie, line 2148: nit: Add the `HLSL Change` comments

sudonatalie (approved):

amaiorano:
@llvm-beanz It looks like the CI failures are because of https://github.com/actions/runner-images/issues/10004. Can we forcibly merge this fix?

PR #6683, opened 2024-06-10
Workaround broken GitHub runner images

llvm-beanz:
This PR contains two changes:
1) Moves a pragma to disable a warning, which seems to be required by the new compiler.
2) Adds a preprocessor define to workaround the crashes caused by the runner image mismatching C++ runtime versions.

The second change we will want to revert once the runner images are fixed. The issue tracking the runner images is:

https://github.com/actions/runner-images/issues/10004

Related #6668

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,12 @@ if(WIN32 AND NOT (CMAKE_C_COMPILER_ARCHITECTURE_ID MATCHES "ARM.*"))
   add_link_options(/CETCOMPAT)
 endif()
 
+# This is a hack workaround for the broken azure runner images:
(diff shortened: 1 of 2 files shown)


damyanp:
> We do not want this change to stick around forever

I think that's true for the mutex/constexpr thing, but not for the moved pragma, right?

damyanp (approved):

llvm-beanz:
> > We do not want this change to stick around forever
> 
> I think that's true for the mutex/constexpr thing, but not for the moved pragma, right?

Probably right. I just added that change now because the debug build failed on my first change. I'll update the description.

hekota (approved):

PR #6686, opened 2024-06-11
[SPIR-V] Emit OpUndef for undefined values

### LINKED ISSUE #6653: [SPIR-V] OpConstantNull Result Type <id> '12[%type_2d_image]' cannot have a null value
**Description**
Invalid spirv generated when calling a function that returns a texture depending on a condition

**Steps to Reproduce**

[Shader Playground](https://shader-playground.timjones.io/a43e24ed089045576f91e58eee2fa5f0)

**Actual Behavior**
fatal error: generated SPIR-V is invalid: OpConstantNull Result Type <id> '12[%type_2d_image]' cannot have a null value.
  %13 = OpConstantNull %type_2d_image

**Environment**
DXC Trunk, no optimization

Keenuts:
Before this change, OpConstantNull was emitted when an undef value was required.
This causes an issue for some types which cannot have the OpConstantNull value.

In addition, it mixed well-defined values with undefined values, which prevents any kind of optimization/analysis later on.

Fixes #6653

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvInstruction.h
+++ b/tools/clang/include/clang/SPIRV/SpirvInstruction.h
@@ -66,6 +66,8 @@ public:
     IK_ConstantFloat,
     IK_ConstantComposite,
     IK_ConstantNull,
+
+    // OpUndef
     IK_Undef,
 
     // Function structure kinds
@@ -1182,7 +1184,8 @@ class SpirvConstant : public SpirvInstruction {
 public:
   // For LLVM-style RTTI
   static bool classof(const SpirvInstruction *inst) {
-    return inst->getKind() >= IK_ConstantBoolean && inst->getKind() <= IK_Undef;
+    return inst->getKind() >= IK_ConstantBoolean &&
+           inst->getKind() <= IK_ConstantNull;
   }
 
   bool operator==(const SpirvConstant &that) const;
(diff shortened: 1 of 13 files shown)


cassiebeckley (approved):

tools/clang/include/clang/SPIRV/SpirvInstruction.h
```diff
@@ -1302,6 +1302,22 @@ class SpirvConstantNull : public SpirvConstant {
   bool operator==(const SpirvConstantNull &that) const;
 };
 
+class SpirvUndef : public SpirvConstant {
```
s-perron, line 1305: This is awkward. None of the extra functions provided by SpirvConstant even make sense. For example, what happens if you call `setLiteral` on an undef? Is that meaning full?

However, because an undef could be a constituent of an OpCompositeComposite, you want to have parent type for all constants and undefs.

I don't think we should make major changes, but we should add comments to explain why an Undef is being treated as a constant, and have some type of error checking somewhere that an undef is not suppose to be a literal.

s-perron (changes requested):
Generally looks good. However, a little extra error checking and comments might be needed. There are some none obvious design choices.

tools/clang/include/clang/SPIRV/SpirvInstruction.h
Keenuts: Yes, I agree it's weird. I made the change to actually inherit from SpirvInstr, and it was fine.
The main reason is: we never generate OpUndef in an OpComposite. The only case it seems we can generate such instruction is with those dead branches, otherwise it's just an variable we haven't initialized. So it seems fine to split SpirvConstant and SpirvUndef.

s-perron (approved):

PR #6696, opened 2024-06-14
[SPIR-V] Re-enable rich debug instructions for objects

### LINKED ISSUE #6691: [SPIR-V] Re-enable object debug info
Now that SPV_KHR_relaxed_extended_instruction is published, we can fix issues like #5218 and #4911 for good.

```hlsl
// RUN: %dxc -T cs_6_0 %s -E main -spirv -fspv-target-env=vulkan1.1 -fspv-debug=vulkan-with-source | FileCheck %s

class A {
  void foo() {
  }
};

[numthreads(1, 1, 1)]
void main() {
  A a;
  a.foo();
}
```

```
warning: Member functions will not be linked to their class in the debug information. See https://github.com/KhronosGroup/SPIRV-Registry/issues/203
```

Keenuts:
This commit bumps SPIR-V tools version, and re-add support for objects debug instructions when using Vulkan's debug instructions.

Because OpenCL debug instructions are not a non-semantic set, the SPIR-V spec would need to be modified, as today it does not allows forward references.

Fixes #6691

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.h
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.h
@@ -1180,9 +1180,8 @@ private:
   /// Returns true on success and false otherwise.
   bool spirvToolsOptimize(std::vector<uint32_t> *mod, std::string *messages);
 
-  // \brief Calls SPIRV-Tools optimizer's, but only with the
-  // fix-opextinst-opcodes pass. This pass fixups
-  // OpExtInst/OpExtInstWithForwardRefsKHR opcodes to use the correct one
+  // \brief Calls SPIRV-Tools optimizer fix-opextinst-opcodes pass. This pass
+  // fixes OpExtInst/OpExtInstWithForwardRefsKHR opcodes to use the correct one
   // depending of the presence of forward references.
   bool spirvToolsFixupOpExtInst(std::vector<uint32_t> *mod,
                                 std::string *messages);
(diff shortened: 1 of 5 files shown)


cassiebeckley (approved):

tools/clang/lib/SPIRV/SpirvEmitter.h
```diff
@@ -1180,6 +1180,13 @@ class SpirvEmitter : public ASTConsumer {
   /// Returns true on success and false otherwise.
   bool spirvToolsOptimize(std::vector<uint32_t> *mod, std::string *messages);
 
+  // \brief Calls SPIRV-Tools optimizer's, but only with the
+  // fix-opextinst-opcodes pass. This pass fixups
+  // OpExtInst/OpExtInstWithForwardRefsKHR opcodes to use the correct one
+  // depending of the presence of forward references.
```
s-perron, line 1186:
```suggestion
  // \brief Calls SPIRV-Tools optimizer fix-opextinst-opcodes pass. This pass fixs
  // OpExtInst/OpExtInstWithForwardRefsKHR opcodes to use the correct one
  // depending of the presence of forward references.
```

s-perron: You might need to run clang format after accepting this suggestion.

s-perron (approved):
LGTM. Just a minor change in a comment.

PR #6700, opened 2024-06-18
Avoid adding types to default namespace

### LINKED ISSUE #6646: integral_constant is exposed to the global namespace in DXC (trunk)
**Description**
`integral_constant` is defined in the global namespace in DXC(1.8.2405.17). The primary template appears to be declared as follows:

```c++
template<typename T>
struct integral_constant;
```
I don't think this type should be defined in the global namespace.

**Steps to Reproduce**

This should not compile: https://godbolt.org/z/hKxdKhW7e.
DXC 1.8.2403 does not compile this code: https://godbolt.org/z/vha9GennT

**Environment**
- DXC version: 1.8.2405.17
- Host Operating System N/A

s-perron:
Some of the types that have been added to the vk namespace were being
added to the default namespace when compiling for DXIL. The if
conditions were such that they would fall through to a default case.

The solution is to explicitly add code that we should skip adding those
builtin types when the vk namespace is not defined.

Fixes #6646.


tools/clang/test/CodeGenSPIRV/vk.types.in.dxil.hlsl
s-perron, line 1: I put the new test in the SPIRV directory because I want it to be easier to find if we add new types to the vk namespace. However, they are tests that do not include `-spirv`, and only check diagnostics. Let me know if there is another directory you want this in.

Keenuts (approved):

s-perron:
@llvm-beanz Do you have any thoughts on where the tests should go?

tools/clang/test/CodeGenSPIRV/vk.types.in.dxil.hlsl
llvm-beanz: How about `SemaHLSL`? Since these are all Sema checks that makes the most sense to me.

s-perron: done.

s-perron:
@llvm-beanz Let me know if this seems reasonable.

llvm-beanz (approved):

Issue #6541, opened 2024-04-17
[SPIR-V] [Proposal 0011] SPIR-V Type for a PhysicalStorageBuffer class OpTypePointer fails to compile because of missing offsets for a struct


devshgraphicsprogramming:
**Description**
This was already solved for `vk::RawBufferLoad<Composite>` however it rears its head again when we start using `vk::SpirvOpaqueType` to declare the same poitner type `RawBufferLoad` uses under the hood.

Related to this probably: #4924 and #2411 

**Steps to Reproduce**
This Godbolt: https://godbolt.org/z/MjdfrGjPz

**Actual Behavior**
Every time an
```cpp
 vk::SpirvOpaqueType<
        /* OpTypePointer */ 32,
        /* PhysicalStorageBuffer */ vk::Literal<vk::integral_constant<uint,5349> >,
        T
    >;
```
gets instantiated where `T` is a composite, we get the following error
```
fatal error: generated SPIR-V is invalid: Structure id 7 decorated as Block for variable in PhysicalStorageBuffer storage class must follow relaxed storage buffer layout rules: member 0 is missing an Offset decoration
```

**Environment**
- DXC version : Latest trunk on Godbolt
- Host Operating System : Linux I guess?

devshgraphicsprogramming:
Its been over a year since I've checkd, but seems that still, contrary to the docs
```
struct Test
{
    // The vk::ext_decorate don't seem to work or emit `OpMemberDecorate` at all
    [[vk::ext_decorate(/*Offset*/ 35,0)]] float4 mem1;
    [[vk::ext_decorate(/*Offset*/ 35,16)]] float mem2;
    [[vk::ext_decorate(/*Offset*/ 35,20)]] int mem3;
};
```

sudonatalie:
Thanks for reporting @devshgraphicsprogramming. You've piqued my interest with https://github.com/microsoft/DirectXShaderCompiler/issues/6489#issuecomment-2058755134 so we'll try to prioritize this.

sudonatalie:
We'll want a dedicated implementation anyways, but I'm still keen to see full features done successfully in inline SPIR-V.

devshgraphicsprogramming:
The `vk::ext_decorate` seems really flaky, I'm trying to decorate  my pointer variable as `AliasedPointer` and seems like some SPIR-V Optimization rips it away

https://godbolt.org/z/9szf1nYq5

s-perron:
[...]

PR #6701, opened 2024-06-18
Set the layout rule explicitly for raw buffer operations

### LINKED ISSUE #6554: [SPIR-V] RawBufferStore still trips up on missing member offsets
**Description**
The usage of such a struct with `vk::RawBufferStore`
```cpp
struct emul_float64_t
{
    using storage_t = float32_t;

    emul_float64_t operator*(const emul_float64_t rhs)
    {
        emul_float64_t retval;
        retval.data = data*rhs.data;
        return retval;
    }

    storage_t data;
};
```
generates the following error
```
fatal error: generated SPIR-V is invalid: Structure id 3 decorated as Block for variable in PhysicalStorageBuffer storage class must follow relaxed storage buffer layout rules: member 0 is missing an Offset decoration
  %emul_float64_t = OpTypeStruct %float
```

**Steps to Reproduce**
This Godbolt: https://godbolt.org/z/de8xeaTeq

**Environment**
- DXC version: Latest Godbolt trunk
- Host Operating System: Linux I guess?

s-perron:
The first first fix in #5392 was not correct. It relied on the layout
rule for the address to be the correct layout rule, but that is not
always the case. The address is just an integer that could exist in any
storage class. The correct solution is to explicitly set the layout rule
for the BitCast operation when expanding the RawBuffer* functions. We
know that the result of the BitCast is a pointer to the physical storage
buffer storage class, so we know the layout need to be the storage
buffer layout.

Fixes #6554

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -14263,6 +14263,7 @@ SpirvEmitter::loadDataFromRawAddress(SpirvInstruction *addressInUInt64,
   SpirvUnaryOp *address = spvBuilder.createUnaryOp(
       spv::Op::OpBitcast, bufferPtrType, addressInUInt64, loc);
   address->setStorageClass(spv::StorageClass::PhysicalStorageBuffer);
+  address->setLayoutRule(spirvOptions.sBufferLayoutRule);
 
   SpirvLoad *loadInst =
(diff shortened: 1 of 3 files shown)


devshgraphicsprogramming:
Does this also fix #6541 ?

s-perron:
> Does this also fix #6541 ?

No, it does not.

Keenuts (approved):

sudonatalie (approved):

PR #6707, opened 2024-06-19
Fix a PIX pass's attempt to set the validator version

jeffnn:
This pass was attempting to compare different things. The return values of GetDxilVersion are not shader models, but... dxil version. Since the code is trying to upgrade the validator version, I changed this to GetValidatorVersion, to pair with SetValidatorVersion.
The previous code was breaking the nvidia driver on workgraphs.

### DIFF
--- a/lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
+++ b/lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
@@ -123,8 +123,8 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
   }
   unsigned int Major = 0;
   unsigned int Minor = 0;
-  m_DM->GetDxilVersion(Major, Minor);
-  if (Major < 6 || (Major == 6 && Minor <= 4)) {
+  m_DM->GetValidatorVersion(Major, Minor);
+  if (Major < 1 || (Major == 1 && Minor < 4)) {
     m_DM->SetValidatorVersion(1, 4);
   }
 
@@ -149,7 +149,7 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
     }
   }
 
-  for (auto *F : instrumentableFunctions) {
+  for (auto* F : instrumentableFunctions) {
     int InstructionRangeStart = InstNum;
     int InstructionRangeEnd = InstNum;
     for (auto &block : F->getBasicBlockList()) {
@@ -177,7 +177,6 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
                   << printableNameSubset << "\n";
     }
   }
-
   if (OSOverride != nullptr) {
     // Print a set of strings of the exemplary form "InstructionCount: <n>
     // <fnName>"
--- a/lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
+++ b/lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
@@ -124,7 +124,7 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
   unsigned int Major = 0;
   unsigned int Minor = 0;
   m_DM->GetValidatorVersion(Major, Minor);
-  if (Major < 1 || (Major == 1 && Minor < 4)) {
+  if (hlsl::DXIL::CompareVersions(Major, Minor, 1, 4) < 0) {
     m_DM->SetValidatorVersion(1, 4);
   }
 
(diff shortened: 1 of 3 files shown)


lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -123,8 +123,8 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
   }
   unsigned int Major = 0;
   unsigned int Minor = 0;
-  m_DM->GetDxilVersion(Major, Minor);
-  if (Major < 6 || (Major == 6 && Minor <= 4)) {
+  m_DM->GetValidatorVersion(Major, Minor);
+  if (Major < 1 || (Major == 1 && Minor < 4)) {
```
tex3d, line 127: If `Major < 1`, the validator version is probably 0.0, which means it's never meant to be validated, and shouldn't be changed to 1.4.  You should use `DXIL::CompareVersions` to compare, as it takes this into account.

```suggestion
  if (DXIL::CompareVersions(Major, Minor, 1, 4) < 0)) {
```

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -177,7 +177,6 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
                   << printableNameSubset << "\n";
     }
   }
-
```
tex3d, line 180: nit: is this intentional?

tools/clang/test/DXC/Passes/PIX/validatorUpgrade.ll
```diff
@@ -0,0 +1,64 @@
+; RUN: %dxopt %s -dxil-annotate-with-virtual-regs -hlsl-dxilemit -S | FileCheck %s
+
+; CHECK: !dx.valver = !{![[VALVER:.*]]}
+; CHECK-NOT: ![[VALVER]] = !{i32 1, i32 4}
```
tex3d, line 4: This ensures we don't downgrade the version, but is there already a test to make sure we upgrade the validator version in the appropriate scenario somewhere else?

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -123,8 +123,8 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
   }
   unsigned int Major = 0;
   unsigned int Minor = 0;
-  m_DM->GetDxilVersion(Major, Minor);
-  if (Major < 6 || (Major == 6 && Minor <= 4)) {
+  m_DM->GetValidatorVersion(Major, Minor);
```
tex3d, line 126: Wha's the original reason for upgrading the validator version to 1.4 specifically?

jeffnn: I wish I could ask jpporto that question... this pass does nothing but add some metadata, so maybe the metadata format changed in 1.4? I also wonder if it's here on behalf of the two other passes that are typically run with this one (value-to-delcare and shader debug instrumentation).
I guess I could hack this to assert 1.3 and see what breaks... I'd be pretty nervous about removing this though, even if nothing I try actually breaks.

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
jeffnn: Thanks!

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
jeffnn: Not

tools/clang/test/DXC/Passes/PIX/validatorUpgrade.ll
jeffnn: Didn't think of manually editing in an older valver. New test added.

bob80905 (approved):
Though I also don't understand the reasoning, the code and tests look correct. Although I would personally put a blurb on top of each test specifying why each test exists. (Like for example, in validatorNoDowngrade, "This test tests that a shader with dxil validator version of 1.8 does not downgrade to 1.4" and for validatorUpgrade "This test tests that a shader with dxil validator version of 1.3 gets upgraded to 1.4 after the annotate with virtual regs pass."

jeffnn:
> Though I also don't understand the reasoning, the code and tests look correct. Although I would personally put a blurb on top of each test specifying why each test exists. (Like for example, in validatorNoDowngrade, "This test tests that a shader with dxil validator version of 1.8 does not downgrade to 1.4" and for validatorUpgrade "This test tests that a shader with dxil validator version of 1.3 gets upgraded to 1.4 after the annotate with virtual regs pass.")

Good idea. Your comment suggestions used, almost verbatim

tex3d (approved):
The change looks fine, though I still don't know why the validator version was upgraded in the first place.

PR #6710, opened 2024-06-20
HLMatrixLower: allow exceptions to propagate out

dneto0:
If an exception is thrown, don't block it in the TempOverloadPool destructor. Allow it to propagate out as a user-visible error.

Explicitly clear the TempOverloadPool before returning from the HLMatrixLowerPass::runOnModule. In the normal case, when no exception is thrown, that will still verify that all the overloads actually have been lowered, and will assert out if they aren't.


dneto0:
cc: @amaiorano

tools/clang/test/HLSLFileCheck/passes/hl/hl_matrix_lower/dont_crash_on_invalid_cast.hlsl
```diff
@@ -0,0 +1,34 @@
+// RUN: %dxc -T cs_6_0 %s | xfail
+
+// The HL matrix lowering pass can sometimes throw an exception
+// due to an invalid LLVM-level cast<Ty> call.  Make sure that
+// propagates out to a user-level error.
+
+// Note: There is still a bug in the compiler here.  Not all matrix
+// lowerings are covered by the pass.
+
+// (I would like to use a FileCheck %s but that does not work.
```
dneto0, line 10: This can be tested better with the 'xfailcat' from #6711

amaiorano: You can use FileCheck, as I did here: https://github.com/microsoft/DirectXShaderCompiler/pull/6556
Change it to:
```
// RUN: not %dxc -T cs_6_0 %s 2>&1 | FileCheck %s
```

amaiorano (approved):
LGTM

tools/clang/test/HLSLFileCheck/passes/hl/hl_matrix_lower/dont_crash_on_invalid_cast.hlsl
llvm-beanz: Can we make this an IR test instead to run via lit using `dxopt`?

llvm-beanz (commented):
Is there a way that we can instead have an explicit `report_fatal_error` call for these error cases rather than failing because an exception is thrown due to a bad cast?

I hate DXC's reliance on exceptions, so it would be preferable to make the error explicit (even if report_fatal_error does just throw an exception in DXC).

dneto0:
> Is there a way that we can instead have an explicit `report_fatal_error` call for these error cases rather than failing because an exception is thrown due to a bad cast?
> 
> I hate DXC's reliance on exceptions, so it would be preferable to make the error explicit (even if report_fatal_error does just throw an exception in DXC).

In this case it comes from a failed checked cast. There are over a dozen of them in that pass:
```cpp
                 cast<FixedVectorType>(LoweredDstTy)->getNumElements(),
                 cast<FixedVectorType>(LoweredVec->getType())->getNumElements(),
      cast<FixedVectorType>(Result->getType())->getNumElements();
      cast<PointerType>(HLMatrixType::getLoweredType(Global->getType()));
      cast<PointerType>(HLMatrixType::getLoweredType(MatAlloca->getType()));
               cast<VectorType>(LoweredRhs->getType()),
  Constant *IdxVec = cast<Constant>(
      DXASSERT(cast<ConstantInt>(*GEP->idx_begin())->isZero(),
    DXASSERT(cast<FixedVectorType>(Result->getType())->getNumElements() ==
  DXASSERT(cast<VectorType>(LoweredLhs->getType()) &&
  Function *Func = cast<Function>(Module.getOrInsertFunction(MangledName, Ty));
    IRBuilder<> Builder(cast<Instruction>(Use.getUser()));
  return cast<Constant>(MatTy.emitLoweredRegToMem(Vec, DummyBuilder));
  VectorType *IdxVecTy = cast<VectorType>(IdxVec->getType());
  VectorType *VecTy = cast<VectorType>(LoweredVal->getType());
```
I don't know which one is triggered for the given test case.

Do you mean something like rewriting the code to use dyn_cast and then explicitly handle the null return case?  So the error is in apparent from the code?

The exception is thrown from https://github.com/microsoft/DirectXShaderCompiler/blame/b197bec653c11d50103c6971a9ccde7fa3cacc93/lib/Support/ErrorHandling.cpp#L143 
I noticed @pow2clk changed the policy around checked casts, to make them throw hlsl exceptions.  See https://github.com/microsoft/DirectXShaderCompiler/commit/6b44d611f737db42e7726d9597d7df76014a0f36

tools/clang/test/HLSLFileCheck/passes/hl/hl_matrix_lower/dont_crash_on_invalid_cast.hlsl
dneto0: Changed to a lit-based test.

dneto0:
I rebased, and moved the test so it's lit-based and uses dxopt.

llvm-beanz (approved):

PR #6711, opened 2024-06-20
HLSLTestLib: Support xfailcat: acts like xfail but propagates stdin,stdout     

dneto0:
Add support for 'xfailcat' in RUN: commands.
It requires the prior command to fail, and also propagates
stdout and stderr.

This is useful for cases for parsing error messages, example
    
       // RUN: %dxc -T cs_6_0 will_fail.hlsl %s | xfailcat | FileCheck %s
       // CHECK: error: something blah blah

This is an alternative to supporting 'not'.
    
Builds on #6710.
Also, modify the test from #6710 to check the error message


llvm-beanz:
But we do support `not` in the LIT-test flow, and we're trying to kill off the legacy test runners, so this seems like a step in the wrong direction.

PR #6717, opened 2024-06-21
Fix non-determinism in Reassociate caused by address coincidences

adam-yang:
Originally @lizhengxing's PR.  Retargeting main.

This PR pulls the upstream change, Fix non-determinism in Reassociate caused by address coincidences (https://github.com/llvm/llvm-project/commit/ef8761fd3b0f0b24b7e9347ad4ae686fc9ed8f94), into DXC.

  Between building the pair map and querying it there are a few places that erase and create Values. It's rare but the address of these newly created Values is occasionally the same as a
  just-erased Value that we already have in the pair map. These coincidences should be accounted for to avoid non-determinism.

  Thanks to Roman Tereshin for the test case.

This is part 6 (the last part) of the fix for #6659.


dmpots (approved):
Wohoo!

damyanp (approved):
LLAFPOTUCTM!

(Looks like a faithful port of the upstream commit to me)

PR #6722, opened 2024-06-25
 [DOC] Add doc about Module Flags Metadata in DXIL

python3kgae:
This patch adds documentation about Module Flags in DXIL.

The only module flags that are currently supported are: "Dwarf Version" and "Debug Info Version".

### DIFF
--- a/docs/DXIL.rst
+++ b/docs/DXIL.rst
@@ -267,6 +267,17 @@ The attribute 'waveops-include-helper-lanes' is utilized to indicate that wave o
 
 'fp32-denorm-mode' is employed to define the denorm mode for the function. The possible values for this attribute can be 'any', 'preserve', or 'ftz'.
 
+Module Flags
+------------
+
+DXIL uses `Module Flags Metadata <https://releases.llvm.org/3.7.1/docs/LangRef.html#module-flags-metadata>`_ to convey additional information about the module.
+
+The only module flags allowed in DXIL are "Dwarf Version", "Debug Info Version".
+
+The "Dwarf Version" module flag is used to specify the DWARF version to use for debug information.
+
+The "Debug Info Version" module flag is used to specify the debug info schema version.
+
 Identifiers
 -----------
 
--- a/docs/DXIL.rst
+++ b/docs/DXIL.rst
@@ -270,7 +270,7 @@ The attribute 'waveops-include-helper-lanes' is utilized to indicate that wave o
 Module Flags
 ------------
 
-DXIL uses `Module Flags Metadata <https://releases.llvm.org/3.7.1/docs/LangRef.html#module-flags-metadata>`_ to convey additional information about the module.
+DXIL uses `Module Flags Metadata <LangRef.html#module-flags-metadata>`_ to convey additional information about the module.
 
 The only module flags allowed in DXIL are "Dwarf Version", "Debug Info Version".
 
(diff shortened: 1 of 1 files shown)


docs/DXIL.rst
```diff
@@ -267,6 +267,17 @@ The attribute 'waveops-include-helper-lanes' is utilized to indicate that wave o
 
 'fp32-denorm-mode' is employed to define the denorm mode for the function. The possible values for this attribute can be 'any', 'preserve', or 'ftz'.
 
+Module Flags
+------------
+
+DXIL uses `Module Flags Metadata <https://releases.llvm.org/3.7.1/docs/LangRef.html#module-flags-metadata>`_ to convey additional information about the module.
```
llvm-beanz, line 273: Rather than linking to the external LLVM doc, can we just link to the copy of the correct version of the document that lives in DXC’s tree?

docs/DXIL.rst
```diff
@@ -267,6 +267,17 @@ The attribute 'waveops-include-helper-lanes' is utilized to indicate that wave o
...
+
+The only module flags allowed in DXIL are "Dwarf Version", "Debug Info Version".
```
llvm-beanz, line 275: What does the validator verify for these flags? Does it verify just the name, or does it rely on them having a specific structure?

python3kgae: We didn't check it at all in validation currently :(
Since we don't modify the Dwarf Version and Debug Info Version in dxc, shall we limit Dwarf Version always be 4 and Debug Info Version always be 3?

docs/DXIL.rst
python3kgae: Updated.

docs/DXIL.rst
```diff
@@ -267,6 +267,17 @@ The attribute 'waveops-include-helper-lanes' is utilized to indicate that wave o
 
 'fp32-denorm-mode' is employed to define the denorm mode for the function. The possible values for this attribute can be 'any', 'preserve', or 'ftz'.
 
+Module Flags
+------------
+
+DXIL uses `Module Flags Metadata <LangRef.html#module-flags-metadata>`_ to convey additional information about the module.
+
+The only module flags allowed in DXIL are "Dwarf Version" and "Debug Info Version".
```
damyanp, line 275: > The only module flags allowed in DXIL are ...

Implied in this is that any other module flags will need to be stripped out when lowering to DXIL.  May be worth calling out?

docs/DXIL.rst
```diff
@@ -267,6 +267,17 @@ The attribute 'waveops-include-helper-lanes' is utilized to indicate that wave o
...
+
+The "Dwarf Version" module flag is used to specify the DWARF version to use for debug information.
+
+The "Debug Info Version" module flag is used to specify the debug info schema version.
+
```
damyanp, line 280: For me, it'd help to see an example.  Are those names exactly how they'd appear in the metadata? Are these flags always i32's?

Something like:

```llvm
!0 = !{  i32 1, !"Dwarf Version", i32 1 }
```

Writing out that example made me realize that maybe we need to say something about the behaviors that someone consuming DXIL would expect to see here?

damyanp (approved):
LGTM, a couple of notes that I think would help me understand it better.

docs/DXIL.rst
python3kgae: Updated.

docs/DXIL.rst
python3kgae: Updated.

docs/DXIL.rst
llvm-beanz: What _does_ the validator check? If it doesn't check anything why are we claiming that these are the only valid flags?

python3kgae: Because those flags are only flags dxc will generated.

PR #6725, opened 2024-06-26
fix MultiByte <-> WideChar conversion return value on linux

exoticorn:
MultiByteToWideChar/WideCharToMultiByte are supposed to include/exclude null termination based on whether the source string buffer is null terminated.

This was done by comparing the input byte/wide character count to the output byte/wide character count excluding null termination. This only works correctly if the input only consists of ASCII characters. (ie. only characters that are single byte in UTF-8)

This commit changes the code to explicitly check for a null terminator in the input and add one to the output size if the input was terminated.

### DIFF
--- a/lib/DxcSupport/Unicode.cpp
+++ b/lib/DxcSupport/Unicode.cpp
@@ -54,7 +54,8 @@ int MultiByteToWideChar(uint32_t /*CodePage*/, uint32_t /*dwFlags*/,
   size_t rv;
   const char *prevLocale = setlocale(LC_ALL, nullptr);
   setlocale(LC_ALL, "en_US.UTF-8");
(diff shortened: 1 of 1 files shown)


exoticorn:
@microsoft-github-policy-service agree company="Keen Games GmbH"

exoticorn:
Without this fix, I ran into the assert
```
    DXASSERT(
        !utf8CharCount ||
            IsBufferNullTerminated(utf8NewCopy.m_pData, utf8CharCount, CP_UTF8),
        "otherwise, CodePageBufferToUtf8 failed to null-terminate buffer.");
```
in `DxcGetBlobAsUtf8` when loading a shader source that includes exactly one two-byte utf-8 character.

tex3d:
Thanks for the submission. However, when #7506 merges, it will include this fix.

PR #6729, opened 2024-06-27
Avoid NOMINMAX redefinition warnings

raedrizqie:
On MinGW, NOMINMAX is defined by default, so guard it to avoid redefinition warnings.


damyanp (approved):

hekota (approved):

raedrizqie:
thank you

PR #6731, opened 2024-06-27
Reintroduce bugpoint from LLVM 3.7

llvm-beanz:
This PR re-introduces LLVM's bugpoint tool from the LLVM 3.7 release which had previously been deleted in the DXC fork. This tool is extremely useful for reducing complex test cases to small reproductions.

I have not tested the tool to verify its degree of functionality, if we find it useful we may wish to invest in making it work better with the HLSL-specific changes in DXC's repository.


llvm-beanz:
Abandoning since it looks like there is more work to really make this useful. We can revive it if it is worth doing in the future.

PR #6741, opened 2024-07-01
For future reference: static linking across different platforms

Nielsbishere:
The following PR is meant more as a showcase of hacky solutions to get static linking to work across different platforms.
I'm not expecting this to get merged of course, it's just to demonstrate my findings and hopefully something can be improved internally to fix these issues in a better way down the road.
It demonstrates the following issues, that when solved can make static linking possible:
- DLL hooks should be made into DxcInitialize and DxcShutdown which the user needs to manually call if static linking is applied. 
- DXC_API_IMPORT becomes a NO-OP in static linking.
- LLVMHLSL, LLVMipo and LLVMScalarOpts have some weird circular dependency that makes linking impossible on non MSVC it seems like. In this branch, I merged them all in one, which is a terrible solution, but it works. Ideally, the circular dependency should be resolved instead. clangAnalysis and clangAST have the same problem, so I also merged those. 
- DXC CLI tools expect dynamic linking, so I turned them off (I only needed dxcompiler.lib/.a, etc.).

As an additional note, I hope the wchars can be phased out slowly too, as on linux/OSX these unexpectedly (for me at least) these are UTF32 rather than UTF16. UTF8 would be nice.


Nielsbishere:
Closing as main on my fork is now merging all sorts of things.

PR #6742, opened 2024-07-03
Allow cap and extention attribute on member variables and type aliases

s-perron:
This implements the changes proposed in
https://github.com/microsoft/hlsl-specs/pull/270.

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -4785,7 +4785,6 @@ void DeclResultIdMapper::registerCapabilitiesAndExtensionsForType(
     const TypedefType *type) {
   for (const auto *decl : typeAliasesWithAttributes) {
     if (type == decl->getTypeForDecl()) {
-      assert(decl->hasAttrs());
       for (auto *attribute : decl->specific_attrs<VKExtensionExtAttr>()) {
         clang::StringRef extensionName = attribute->getName();
         spvBuilder.requireExtension(extensionName, decl->getLocation());
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -968,13 +968,6 @@ DeclResultIdMapper::getDeclSpirvInfo(const ValueDecl *decl) const {
 SpirvInstruction *DeclResultIdMapper::getDeclEvalInfo(const ValueDecl *decl,
                                                       SourceLocation loc,
                                                       SourceRange range) {
-  if (decl->hasAttr<VKExtensionExtAttr>() ||
-      decl->hasAttr<VKCapabilityExtAttr>()) {
-    theEmitter.createSpirvIntrInstExt(decl->getAttrs(), QualType(),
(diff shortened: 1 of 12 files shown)


tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -4776,5 +4781,22 @@ void DeclResultIdMapper::storeOutStageVarsToStorage(
   }
 }
 
+void DeclResultIdMapper::registerCapabilitiesAndExtensionsForType(
+    const TypedefType *type) {
+  for (const auto *decl : typeAliasesWithAttributes) {
+    if (type == decl->getTypeForDecl()) {
+      assert(decl->hasAttrs());
```
Keenuts, line 4788: Is this assert useful?
(iterators should be fine if no attr are presents, and its presence doesn't guarantee we have at least one of them no?)

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -1696,6 +1697,26 @@ bool SpirvEmitter::validateVKAttributes(const NamedDecl *decl) {
   return success;
 }
 
+void SpirvEmitter::registerCapabilitiesAndExtensionsForVarDecl(
+    const VarDecl *varDecl) {
+  // First record any extensions that part of the actual variable declaration.
```
Keenuts, line 1702:
```suggestion
  // First record any extensions that are part of the actual variable declaration.
```

tools/clang/test/CodeGenSPIRV/spv.inline.extension.unused.hlsl
```diff
@@ -1,11 +1,21 @@
...
-// CHECK-NOT: OpCapability SampleMaskPostDepthCoverage
+// CHECK-NOT: OpExtension "ext_on_type"
+// CHECK-NOT: OpCapability Int8
+// CHECK-NOT: OpExtension "ext_on_field"
+// CHECK-NOT: OpCapability StorageBuffer8BitAccess
 
-[[vk::ext_extension("some_extension")]]
-[[vk::ext_capability(/* SampleMaskPostDepthCoverageCapability */ 4447)]]
-int val;
+// Test that the capabilities and extensions on an unused type are not added
```
Keenuts, line 8: If we have an unused function using those types, the capabilities/extensions are added no?
(I imagine this might be an issue for inline spirv headers if they add helper functions/classes methods using those annotated types).
(Don't recall if we only compile function we visit starting from the entrypoint or not)

Keenuts (approved):
LGTM, just some nits!

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
s-perron: I think I had a different implementation earlier that would assert if the decl has no attributes, so I added this assert to make sure nothing was added to typeAliasesWithAttributes that did not have any attributes.

However, that is not true with this implementation. I've removed it.

tools/clang/test/CodeGenSPIRV/spv.inline.extension.unused.hlsl
s-perron: We only compile functions that are reachable from the entry point, so it should not be a problem.

PR #6746, opened 2024-07-04
Iterate over the spir-v fields to handle bitfields

### LINKED ISSUE #6483: [SPIR-V] Bitfields in structs fail to compile to SPIR-V
**Description**

Structs with a specific setup of bitfields fail to compile to SPIR-V with the error 
```
fatal error: generated SPIR-V is invalid: Index is out of bounds, can not find index 1 in the structure <id> '2'. This structure has 1 members. Largest valid index is 0.
  %19 = OpCompositeExtract %uint %17 1
```

I looked through and there were quite a few other issues that seemed like they might be related, but they all seemed to be resolved/merged pre-March 2024 release, and this issue is still present in that release version.

**Steps to Reproduce**

A struct with uint bitfields fails to compile when one of the members of the struct is accessed via a method that is not a structured buffer.

Here is the smallest example I could make. The compile options are at the top, followed by the error message, followed by the non-working example. At the bottom is a piece of code that I think functionally does the same thing, but successfully compiles.
https://godbolt.org/z/hYd6faK1j

**Actual Behavior**
Should compile successfully to spir-v as it does to dxil.

**Environment**
- DXC versions:
    - Release 1.7.2308 (69e54e2) (August 2023)
    - Release 1.8.2403 (c9660a8) (March 2024)
    - Release 1.8.2403.1 (9c2b828 (March 2024 patch 1)
- Host Operating System: Windows 11 23H2 (Build 22631.3296)

s-perron:
The code that implements `RWByteAddressBuffer::Store` will iterate over
all of the fields in a struct to write each element in the struct.
However, it does not use the "Spir-V fields", which accounts for
multiple fields being packed into the same bitfield. This is fixed by
using the `forEachSpirvField` function to make sure that the bitfield
are correctly handled.

Fixes #6483

### DIFF
--- a/tools/clang/lib/SPIRV/RawBufferMethods.cpp
+++ b/tools/clang/lib/SPIRV/RawBufferMethods.cpp
@@ -285,37 +285,32 @@ SpirvInstruction *RawBufferHandler::processTemplatedLoadFromBuffer(
   // aligned like their field with the largest alignment.
   // As a result, there might exist some padding after some struct members.
   if (const auto *structType = targetType->getAs<RecordType>()) {
-    const auto *decl = structType->getDecl();
+    LowerTypeVisitor lowerTypeVisitor(astContext, theEmitter.getSpirvContext(),
+                                      theEmitter.getSpirvOptions(), spvBuilder);
+    const StructType *spvType = dyn_cast<StructType>(lowerTypeVisitor.lowerType(
+        targetType, theEmitter.getSpirvOptions().sBufferLayoutRule, llvm::None,
+        SourceLocation()));
     llvm::SmallVector<SpirvInstruction *, 4> loadedElems;
-    uint32_t fieldOffsetInBytes = 0;
-    uint32_t structAlignment = 0, structSize = 0, stride = 0;
-    std::tie(structAlignment, structSize) =
-        AlignmentSizeCalculator(astContext, theEmitter.getSpirvOptions())
-            .getAlignmentAndSize(targetType,
-                                 theEmitter.getSpirvOptions().sBufferLayoutRule,
-                                 llvm::None, &stride);
-    for (const auto *field : decl->fields()) {
-      AlignmentSizeCalculator alignmentCalc(astContext,
-                                            theEmitter.getSpirvOptions());
-      uint32_t fieldSize = 0, fieldAlignment = 0;
-      std::tie(fieldAlignment, fieldSize) = alignmentCalc.getAlignmentAndSize(
-          field->getType(), theEmitter.getSpirvOptions().sBufferLayoutRule,
-          /*isRowMajor*/ llvm::None, &stride);
-      fieldOffsetInBytes = roundToPow2(fieldOffsetInBytes, fieldAlignment);
-      auto *byteOffset = address.getByteAddress();
-      if (fieldOffsetInBytes != 0) {
-        byteOffset = spvBuilder.createBinaryOp(
-            spv::Op::OpIAdd, astContext.UnsignedIntTy, byteOffset,
-            spvBuilder.getConstantInt(astContext.UnsignedIntTy,
-                                      llvm::APInt(32, fieldOffsetInBytes)),
-            loc, range);
-      }
+    forEachSpirvField(
+        structType, spvType,
+        [this, &buffer, &address, range,
+         &loadedElems](size_t spirvFieldIndex, const QualType &fieldType,
+                       const auto &field) {
+          auto *baseOffset = address.getByteAddress();
+          if (field.offset.hasValue() && field.offset.getValue() != 0) {
+            const auto loc = buffer->getSourceLocation();
+            SpirvConstant *offset = spvBuilder.getConstantInt(
+                astContext.UnsignedIntTy,
+                llvm::APInt(32, field.offset.getValue()));
+            baseOffset = spvBuilder.createBinaryOp(
+                spv::Op::OpIAdd, astContext.UnsignedIntTy, baseOffset, offset,
+                loc, range);
+          }
 
-      loadedElems.push_back(processTemplatedLoadFromBuffer(
-          buffer, byteOffset, field->getType(), range));
-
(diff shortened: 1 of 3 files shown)


Keenuts (commented):
LGTM, but seems like the issue shouldn't be closed until the Load is also fixed:

```hlsl
// RUN: %dxc -T cs_6_6 %s -E main -Od -spirv | FileCheck %s

RWByteAddressBuffer outBuffer;

struct S {
  uint32_t x : 8;
  uint32_t y : 8;
};

[numthreads(1, 1, 1)]
void main() {
  S s = outBuffer.Load<S>(0);
}
```
```
fatal error: generated SPIR-V is invalid: Expected total number of Constituents to be equal to the number of members of Result Type struct
  %33 = OpCompositeConstruct %S %23 %29

```

s-perron:
> LGTM, but seems like the issue shouldn't be closed until the Load is also fixed:
> 
> ```hlsl
> // RUN: %dxc -T cs_6_6 %s -E main -Od -spirv | FileCheck %s
> 
> RWByteAddressBuffer outBuffer;
> 
> struct S {
>   uint32_t x : 8;
>   uint32_t y : 8;
> };
> 
> [numthreads(1, 1, 1)]
> void main() {
>   S s = outBuffer.Load<S>(0);
> }
> ```
> 
> ```
> fatal error: generated SPIR-V is invalid: Expected total number of Constituents to be equal to the number of members of Result Type struct
>   %33 = OpCompositeConstruct %S %23 %29
> 
> note: please file a bug report on https://github.com/Microsoft/DirectXShaderCompiler/issues with source code if possible
> ```

Done.

tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.store.hlsl
```diff
@@ -2,6 +2,12 @@
 
 RWByteAddressBuffer outBuffer;
 
+struct S
+{
+    uint32_t x:8;
+    uint32_t y:8;
+};
```
Keenuts, line 9: nit: shall we add a test checking we load the correct indices depending on the struct padding?
Ex:

```
struct S {
    uint32_t x : 8;
    uint64_t y : 8; // aligned on the 64bit boundary, hence 32bit uint padding inserted before.
};
```

Keenuts (approved):

tools/clang/test/CodeGenSPIRV/method.byte-address-buffer.store.hlsl
s-perron: There are other tests that check that the layout of the spir-v type is correct. See op.struct.access.bitfield.sized.hlsl. I don't want to repeat all of those tests for every way that bitfields can interact with other features.

tools/clang/lib/SPIRV/RawBufferMethods.cpp
```diff
@@ -284,44 +285,46 @@ SpirvInstruction *RawBufferHandler::processTemplatedLoadFromBuffer(
   // aligned like their field with the largest alignment.
   // As a result, there might exist some padding after some struct members.
   if (const auto *structType = targetType->getAs<RecordType>()) {
-    const auto *decl = structType->getDecl();
+    LowerTypeVisitor lowerTypeVisitor(astContext, theEmitter.getSpirvContext(),
+                                      theEmitter.getSpirvOptions(), spvBuilder);
+    const StructType *spvType = dyn_cast<StructType>(lowerTypeVisitor.lowerType(
+        targetType, theEmitter.getSpirvOptions().sBufferLayoutRule, llvm::None,
+        SourceLocation()));
```
sudonatalie, line 292: This should probably be the Load call's source location, right?

tools/clang/lib/SPIRV/RawBufferMethods.cpp
```diff
@@ -604,40 +607,34 @@ void RawBufferHandler::processTemplatedStoreToBuffer(SpirvInstruction *value,
...
-            spv::Op::OpIAdd, astContext.UnsignedIntTy, byteOffset,
-            spvBuilder.getConstantInt(astContext.UnsignedIntTy,
-                                      llvm::APInt(32, fieldOffsetInBytes)),
-            loc, range);
-      }
+    LowerTypeVisitor lowerTypeVisitor(astContext, theEmitter.getSpirvContext(),
+                                      theEmitter.getSpirvOptions(), spvBuilder);
+    const StructType *spvType = dyn_cast<StructType>(lowerTypeVisitor.lowerType(
+        valueType, theEmitter.getSpirvOptions().sBufferLayoutRule, llvm::None,
+        SourceLocation()));
```
sudonatalie, line 614: Same here

sudonatalie (approved):
Minor suggestion but I'll leave it up to you if there was a reason for wanting it unset

sudonatalie:
> Minor suggestion but I'll leave it up to you if there was a reason for wanting it unset

Actually looking closer at these functions, it's kind of odd that SourceRange of the Load call expr is passed around in the params, but then `const auto loc = buffer->getSourceLocation();` is used frequently throughout to get the location of (I think) the buffer var? Not something to address in this PR but might be worth double checking in a follow up whether diagnostics are ending up where they should.

tools/clang/lib/SPIRV/RawBufferMethods.cpp
s-perron: I'm not sure. I don't expect this to be the final source location when the spir-v is emitted because the result is used for iterating, but is not used or kept. I would expect the source location to be irrelevant to the final spir-v.

However, do not know that for sure, so I updated it. The source location of the load is not correct either, because the type is not defined at the position of the load. I used the source location of the tag decl associated with the struct. I believe one should always exist.

tools/clang/lib/SPIRV/RawBufferMethods.cpp
s-perron: Same as the other comment.

PR #6754, opened 2024-07-09
[SPIR-V] Fix invalid isnan codegen

### LINKED ISSUE #6712: [SPIR-V] Invalid SPIR-V operand with isnan()
**Description**
DXC generates an invalid SPIR-V

**Steps to Reproduce**

`dxc -T cs_6_0 -spirv path/to/file.hlsl`

HLSL code:

```
RWStructuredBuffer<float> Out;
[numthreads(1, 1, 1)]
void main(uint id : SV_DispatchThreadID)
{
        bool bar = isnan(Out[id]);
        Out[id] = (float)bar;
}
```

If we change the code as follows, the issue is not occurred.

```
        float baz = Out[id];
        Out[id] = (float)isnan(baz);
```

**Actual Behavior**

> fatal error: generated SPIR-V is invalid: Expected bool scalar or vector type as Result Type: IsNan
  %23 = OpIsNan %uint %22

**Environment**
- DXC version: libdxcompiler.dylib: 1.8(dev;4640-45018c75)
- Host Operating System: Windows 11 23H2 and macOS 14.5

Keenuts:
IsNan returns a boolean, even is the input-type is a float. This was working in most cases except:
 - if the layout was not Void
 - if the input type was not a matrix

The first bug is because a bool memory layout/representation is not specified, and shall never be exposed to externaly-accessible memory. Hence, if we saw a layout rule != Void, we converted it to a UINT. When calling isnan, the layout rule should not be propagated as we loose any layout info.

The second is because our codegen assumed matrix operations returned a matrix with the same type as the input parameters. In the case of isnan, this was just wrong.

Fixes #6712

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -6262,9 +6262,10 @@ SpirvInstruction *SpirvEmitter::doUnaryOperator(const UnaryOperator *expr) {
     if (isMxNMatrix(subType)) {
       // For matrices, we can only increment/decrement each vector of it.
       const auto actOnEachVec = [this, spvOp, one, expr,
-                                 range](uint32_t /*index*/, QualType vecType,
+                                 range](uint32_t /*index*/, QualType inType,
+                                        QualType outType,
                                         SpirvInstruction *lhsVec) {
-        auto *val = spvBuilder.createBinaryOp(spvOp, vecType, lhsVec, one,
+        auto *val = spvBuilder.createBinaryOp(spvOp, outType, lhsVec, one,
                                               expr->getOperatorLoc(), range);
         if (val)
           val->setRValue();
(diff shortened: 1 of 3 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -9042,6 +9056,15 @@ SpirvEmitter::processIntrinsicCallExpr(const CallExpr *callExpr) {
   case hlsl::IntrinsicOp::IOP_firstbitlow: {
     retVal = processIntrinsicFirstbit(callExpr, GLSLstd450::GLSLstd450FindILsb);
     break;
+  }
+  case hlsl::IntrinsicOp::IOP_isnan: {
```
s-perron, line 9060: Should this be done for `isinf` as well?

s-perron: Yes, `isinf` is a problem as well. Please open a new issues for `isinf`, no need to hold up this PR.

https://godbolt.org/z/fKfzaxE6E

s-perron (approved):
Looks good. We will need to follow up with other similar bugs.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
s-perron: I opened the follow up issue. https://github.com/microsoft/DirectXShaderCompiler/issues/6783

Keenuts: Oh right, thanks. Yes we might want to refactorize a bit this, several intrinsics might be impacted.

Keenuts:
@cassiebeckley in case you missed it.

sudonatalie (approved):
LGTM!

PR #6756, opened 2024-07-09
[SPIR-V] Fix flattened SPIR-V variable codegen.

### LINKED ISSUE #6718: [SPIR-V] DXC crashes when declaring a variable in multiple case block
**Description**
LLVM assertion failed.

**Steps to Reproduce**
`dxc -T cs_6_0 -spirv path/to/file.hlsl`

HLSL code:
```
uint Get(uint a)
{
        [branch]
        switch (a)
        {
        case 0: {
                uint d = a + 1;
                return a + 1;
        }
        case 1: {
                uint d1 = a + 2;
                return a + 2;
        }
        }
        return 0;
}

RWStructuredBuffer<uint> Out;
[numthreads(1, 1, 1)]
void main()
{
        uint a = Out[0];
        uint v = Get(a);
        Out[0] = v;
}
```

If change the code in one of the following ways, the issue is not occurred.

- Remove `[branch]`
- Remove `uint d = a + 1;` or `uint d1 = a + 2;` (in other words, only declaring one variable in the switch statement)

**Actual Behavior**
> Assertion failed: (astDecls[var].instr == nullptr), function createFnVar, file DeclResultIdMapper.cpp, line 1106.

**Environment**
- DXC version: libdxcompiler.dylib: 1.8(dev;4640-45018c75)
- Host Operating System: Windows 11 23H2 and macOS 14.5

Keenuts:
When the [branch] annotation is used, switches are converted into an if/else tree.
Issue arise when declaring a variable into a switch case:
 - when flattened, the same variable could be traversed twice in case of a case fall-through.

In addition, there was a small bug in the flattening logic: only breaks were stopping the handling of further statements, while early-returns were ignored. This was not an important bug as it only added more dead-code, but it was wrong.

Fixes #6718

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -1090,6 +1090,9 @@ void DeclResultIdMapper::createCounterVarForDecl(const DeclaratorDecl *decl) {
 SpirvVariable *
 DeclResultIdMapper::createFnVar(const VarDecl *var,
                                 llvm::Optional<SpirvInstruction *> init) {
+  if (astDecls[var].instr != nullptr)
+    return cast<SpirvVariable>(astDecls[var].instr);
+
   const auto type = getTypeOrFnRetType(var);
   const auto loc = var->getLocation();
   const auto name = var->getName();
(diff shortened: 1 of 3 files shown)


s-perron (approved):

Keenuts:
@cassiebeckley in case you missed it.

sudonatalie (approved):
LGTM!

PR #6758, opened 2024-07-09
[SPIRV] Generate DebugFunctionDefinition in real function instead of wrapper

SteveUrquhart:
This PR moves the DebugFunctionDefinition instruction from the SPIR-V wrapper to the real function.   This supplies debugger users with greater accuracy.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -1449,7 +1449,7 @@ void SpirvEmitter::doFunctionDecl(const FunctionDecl *decl) {
       isEntry = true;
       funcName = "src." + funcName;
       // Create wrapper for the entry function
-      if (!emitEntryFunctionWrapper(decl, func, debugFunction))
+      if (!emitEntryFunctionWrapper(decl, func))
         return;
       // Generate DebugEntryPoint if function definition
       if (spirvOptions.debugInfoVulkan && debugFunction) {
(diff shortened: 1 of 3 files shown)


SteveUrquhart:
The purpose of debug info is to connect the original source code to the corresponding machine code. In this case, the SPIR-V function that corresponds to HLSL's main function is %src_main, not %main. %main is just wrapper code inserted by dxc. By analogy: when debugging a C program the user wants to start at main, even though technically the program begins at an implementation-specific entry point like _start.

Furthermore, putting the DebugFunctionDefinition into %main breaks other debug info. For example, in the following shader:

%25 = OpExtInst %void %1 DebugFunction %23 %16 %18 %uint_1 %uint_1 %19 %24 %uint_3 %uint_2
%27 = OpExtInst %void %1 DebugLexicalBlock %18 %uint_2 %uint_1 %25
%29 = OpExtInst %void %1 DebugLocalVariable %28 %13 %18 %uint_3 %uint_11 %27 %uint_4
%33 = OpExtInst %void %1 DebugEntryPoint %25 %19 %31 %32
%main = OpFunction %void None %34
%35 = OpLabel
%36 = OpExtInst %void %1 DebugFunctionDefinition %25 %main

The real function will refer to DebugLocalVariable %29, which points to the DebugFunction %25, but the DebugFunctionDefinition that points to %25 is in %main.

s-perron (approved):

Keenuts (approved):

PR #6761, opened 2024-07-10
[NFC] Remove implicit `this` capture in lambdas

DKLoehr:
When declaring a lambda with a value-capture default [=, ...], the this pointer is implicitly captured by value as well. This results in potentially-unintuitive behavior and has been deprecated in C++20. It produces a warning in newer versions of clang (https://reviews.llvm.org/D142639).

This PR makes the implicit captures explicit, preventing the warning. It does not change the compiled code at all, since it's just removing some syntactic sugar.


DKLoehr:

damyanp (approved):
LGTM as a mechanical change to keep current behavior.

I can't help feeling that most of the cases here should really be `[&]` though. But that wouldn't be NFC.

coopp (approved):

Issue #6446, opened 2024-03-22
UBSAN "downcast of address" runtime error on `createSentinal` for `ilist`s


amaiorano:
ubsan-compiled dxc will fail with errors like:

```
/home/amaiorano/src/external/DirectXShaderCompiler/include/llvm/ADT/SparseBitVector.h:256:44: runtime error: downcast of address 0x7ffc07f29348 with insufficient space for an object of type 'llvm::ilist_traits<llvm::SparseBitVectorElement<128>>::Element' (aka 'SparseBitVectorElement<128U>')
0x7ffc07f29348: note: pointer points here
 00 00 00 00  00 00 00 00 00 00 00 00  18 ab 02 00 10 61 00 00  38 ab 02 00 10 61 00 00  68 ab 02 00
              ^ 
    #0 0x5572402d0e1d in createSentinel /home/amaiorano/src/external/DirectXShaderCompiler/include/llvm/ADT/SparseBitVector.h:256:44
    #1 0x5572402d0e1d in provideInitialHead /home/amaiorano/src/external/DirectXShaderCompiler/include/llvm/ADT/SparseBitVector.h:260:48
    #2 0x5572402d0e1d in iplist /home/amaiorano/src/external/DirectXShaderCompiler/include/llvm/ADT/ilist.h:350:25
    #3 0x5572402d0e1d in ilist /home/amaiorano/src/external/DirectXShaderCompiler/include/llvm/ADT/ilist.h:603:3
    #4 0x5572402d0e1d in SparseBitVector /home/amaiorano/src/external/DirectXShaderCompiler/include/llvm/ADT/SparseBitVector.h:437:3
    #5 0x5572402d0e1d in (anonymous namespace)::SparseBitVectorTest_TrivialOperation_Test::TestBody() /home/amaiorano/src/external/DirectXShaderCompiler/unittests/ADT/SparseBitVectorTest.cpp:18:21
<snip>
```

The reason this fails is that most implementations of `ilist_traits<T>::createSentinal` return a down-casted instance of a `ilist_half_node<T>` or `ilist_node<T>`, which although can be cast-down to T (since T derives from `ilist_node`), means we're returning a pointer to a sliced object of type T. And although we never dereference the pointer, this is still UB, which is why UBSAN flags it.

[...]

PR #6762, opened 2024-07-10
Use a real instruction for the sentinel in the Instruction intrusive …

dneto0:
…list

Instructions in a BasicBlock are maintained in a doubly-linked list. The links are "instrusive" in the sense that with inheritance tricks the Next and Prev nodes are embedded in the Instruction object itself.

The linked list uses a sentinel object to mark the tail of the list.

Previously, the sentinel was an ilist_half_node<Instruction> which just consists of a 'Instruction* Prev', but then it was cast to Instruction*. This is flagged by the undefined behaviour sanitizer in some cases. Also, it's weird and wrong.

Upstream LLVM has entirely reimplimented the intrusive list to avoid such problems.  But that's a massive change.

This change uses a real Instruction object as the sentinel. The instruction is only used for its Next and Prev properties. I used an UnreachableInst becuase it's small and simple.

Issue: #6446

### DIFF
--- a/include/llvm/IR/Instruction.h
+++ b/include/llvm/IR/Instruction.h
@@ -20,6 +20,8 @@
 #include "llvm/IR/DebugLoc.h"
 #include "llvm/IR/SymbolTableListTraits.h"
 #include "llvm/IR/User.h"
+#include <memory>  // HLSL Change
+#include <utility> // HLSL Change
 
 namespace llvm {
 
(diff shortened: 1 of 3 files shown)


dneto0:
I think the same technique would fix all the other cases in #6446 but I wanted to get this reviewed before putting in the work.

include/llvm/IR/Instruction.h
```diff
@@ -518,12 +529,7 @@ class Instruction : public User, public ilist_node<Instruction> {
 // HLSL Change Starts
 // Temporarily disable "downcast of address" UBSAN runtime error
 // https://github.com/microsoft/DirectXShaderCompiler/issues/6446
-#ifdef __has_feature
-#if __has_feature(undefined_behavior_sanitizer)
-__attribute__((no_sanitize("undefined")))
-#endif // __has_feature(address_sanitizer)
-#endif // defined(__has_feature)
-// HLSL Change Ends
+#if 0
```
dneto0, line 532: Alternatively we can delete this whole overload.  It's done inline above at line 43, and differently.
Not sure what you wanted.

dneto0:
I tested this locally with a Release build and a Debug + ASAN + UBSAN build.

include/llvm/IR/Instruction.h
llvm-beanz: I'd be fine with just deleting the code.

llvm-beanz (approved):
This all looks good to me. @bogner IIRC you reviewed at least some of Duncan's changes to fix all this in LLVM (4?). Do you have any feedback here?

bogner:
> This all looks good to me. @bogner IIRC you reviewed at least some of Duncan's changes to fix all this in LLVM (4?). Do you have any feedback here?

I agree it's probably not realistic to take in all of upstream's changes to fix the issues with intrusive lists, so a stop gap like this might be the most practical thing to do.

include/llvm/IR/Instruction.h
dneto0: I plan to delete the code on a followup when addressing the other cases.

dneto0:
> > This all looks good to me. @bogner IIRC you reviewed at least some of Duncan's changes to fix all this in LLVM (4?). Do you have any feedback here?
> 
> I agree it's probably not realistic to take in all of upstream's changes to fix the issues with intrusive lists, so a stop gap like this might be the most practical thing to do.

Thanks. I need a second approving review to land this. @amaiorano is OOO.

bogner (approved):

PR #6764, opened 2024-07-10
Add CMake option DXC_CODEGEN_EXCEPTIONS_TRAP

dneto0:
When enabled, any hlsl::Exception thrown during code generation and optimization will cause the process to trap.

---
edit: I've changed the implementation to generate a trap, instead of std::abort

### DIFF
--- a/tools/clang/lib/CodeGen/BackendUtil.cpp
+++ b/tools/clang/lib/CodeGen/BackendUtil.cpp
@@ -8,6 +8,10 @@
 //===----------------------------------------------------------------------===//
 
 #include "clang/CodeGen/BackendUtil.h"
+#include "dxc/HLSL/DxilGenerationPass.h" // HLSL Change
+#include "dxc/HLSL/HLMatrixLowerPass.h"  // HLSL Change
+#include "dxc/Support/Global.h"          // HLSL Change
+#include "dxc/config.h"                  // HLSL Change
 #include "clang/Basic/Diagnostic.h"
 #include "clang/Basic/LangOptions.h"
 #include "clang/Basic/TargetOptions.h"
@@ -41,10 +45,8 @@
 #include "llvm/Transforms/ObjCARC.h"
 #include "llvm/Transforms/Scalar.h"
 #include "llvm/Transforms/Utils/SymbolRewriter.h"
+#include <cstdio>
 #include <memory>
-#include "dxc/HLSL/DxilGenerationPass.h" // HLSL Change
-#include "dxc/HLSL/HLMatrixLowerPass.h"  // HLSL Change
-#include "dxc/Support/Global.h" // HLSL Change
 
 using namespace clang;
 using namespace llvm;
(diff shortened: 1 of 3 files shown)


tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -780,8 +783,14 @@ void clang::EmitBackendOutput(DiagnosticsEngine &Diags,
     // so that future passes can be skipped.
     AsmHelper.EmitAssembly(Action, OS);
   } catch (const ::hlsl::Exception &hlslException) {
+#if !defined(DXC_CODEGEN_EXCEPTIONS_ABORT)
     Diags.Report(Diags.getCustomDiagID(DiagnosticsEngine::Error, "%0\n"))
         << StringRef(hlslException.what());
+#else
+    fprintf(stdout, "internal error: %s\n", hlslException.what());
```
llvm-beanz, line 790: This should probably be writing to `llvm::errs()` rather than to `stdout`.

llvm-beanz (approved):
I _really_ don't like this change. It's an unfortunate hack around a problem rather than fixing it. That said, the problem isn't easy to fix and I can't really come up with a better way to address it that isn't insanely massive.

I wish we could just make exceptions an optional configuration, but that would be a huge change and even making our tests pass with exceptions replaced by fatal errors is a huge effort.

Exceptions are evil.
This code all makes me sad.
I refuse to say it "looks good" to me.
I will say it is an acceptable (although unfortunate) compromise.

One small change requested to use `llvm::errs()` instead of `fprintf`.

dneto0:
+1

I wish the exception-catching had not been inserted.

After posting I realized this patch has a big downside:  we don't get a stack trace.  Which means bug report deduplication will get too agressive and we'll likely miss things.

I converted to draft for now.

tools/clang/lib/CodeGen/BackendUtil.cpp
dneto0: It turns out llvm::errs() doesn't do anything in release builds on Linux in libdxcompiler.so,
nor when the dxc command line compiler traps here.

I filed #6774 to followup.

I do want the output in those cases though.

tools/clang/lib/CodeGen/BackendUtil.cpp
```diff
@@ -782,6 +784,13 @@ void clang::EmitBackendOutput(DiagnosticsEngine &Diags,
   } catch (const ::hlsl::Exception &hlslException) {
     Diags.Report(Diags.getCustomDiagID(DiagnosticsEngine::Error, "%0\n"))
         << StringRef(hlslException.what());
+#if defined(DXC_CODEGEN_EXCEPTIONS_TRAP)
+    // llvm::errs() doesn't work in release builds on Linux.
```
dneto0, line 788: ... in libdxcompiler.so which is mainly where I want to use it.

llvm-beanz: Ugh! I had a patch somewhere that fixed that, and I don't remember what the issue was. It was something _really_ silly.

sudonatalie (approved):

PR #6767, opened 2024-07-11
[SPIR-V] Don't flatten arrays during legalization

### LINKED ISSUE #6745: [SPIR-V] Global defined struct { Texture2D; SamplerState } causes irrelevant texture array to be compiled incorrectly
**Description**

A global struct that contains a Texture2D and a SamplerState causes texture array to be compiled into N discrete textures, instead of an array of N textures.

**Steps to Reproduce**

Code below is compiled into 4096 OpDecorates from 0 to 4095. Comment out `TexAndSampler _TS;` then it will be compiled into `%_arr_type_2d_image_uint_4096 = OpTypeArray %type_2d_image %uint_4096` correctly.

Minimum repro code:
```hlsl

// compile with -T ps_6_0 -E frag -spirv -O0

Texture2D<float4> _Textures[4096] : register(t0, space0);
int _TextureIndex;
SamplerState _MainSampler;

struct TexAndSampler {
    Texture2D t;
    SamplerState s;
};

TexAndSampler _TS; // <-------- HERE

float4 frag(float4 position : SV_Position) : SV_Target
{
    return _Textures[NonUniformResourceIndex(_TextureIndex)].Sample(_MainSampler, float2(0.5, 0.5));
}
```

Compiler Explorer link: https://godbolt.org/z/s9MM15zMP

**Actual Behavior**

**Environment**
- DXC version  1.8.2405
- Host Operating System  Windows 11

Keenuts:
When struct contains opaque resources, we must legalize the SPIR-V to move those resources out of the struct (Vulkan doesn't allow composite to store opaque resources).
Before this change, any resource array would also be flattened due to the pass flattening everything by default.

Patched SPIRV-Tools to allow some level of selection in what we flatten. Not perfect, as we would flatten all arrays or all composites and cannot pick only one variable to flatten, but for now this should be enough.

Fixes #6745

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -14698,13 +14698,19 @@ bool SpirvEmitter::spirvToolsLegalize(std::vector<uint32_t> *mod,
   }
   optimizer.RegisterLegalizationPasses(spirvOptions.preserveInterface);
   // Add flattening of resources if needed.
-  if (spirvOptions.flattenResourceArrays ||
-      declIdMapper.requiresFlatteningCompositeResources()) {
+  if (spirvOptions.flattenResourceArrays) {
     optimizer.RegisterPass(
         spvtools::CreateReplaceDescArrayAccessUsingVarIndexPass());
     optimizer.RegisterPass(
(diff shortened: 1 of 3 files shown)


s-perron (approved):

Keenuts:
@cassiebeckley in case you missed this one

sudonatalie (approved):

Issue #6769, opened 2024-07-11
DXIL Validator Hash


python3kgae:
This issue serves as a progress tracker for [DXIL Validator Hash](
https://github.com/microsoft/hlsl-specs/blob/main/proposals/infra/INF-0004-validator-hashing.md)

# Tasks

- [ ] Build the DXIL validator library both as a dynamic and static library to allow full flexibility.
- [ ] Open source the implementation of dxil.dll.
- [ ] Support bypassing Hash Verification.

damyanp:
Setting this to backlog - we're likely to do all the subtasks here, but not necessarily for the same release.

PR #6770, opened 2024-07-11
Open source dxil.dll

python3kgae:
This pull request introduces the open-source implementation of dxil.dll, including hashing and validation functionalities for DXIL containers.

DxilHash.cpp: Implements DXBC/DXIL container hashing functions.
dxcvalidatorp.cpp: Implements the DirectX Validator object, including signing support.
dxildll.cpp: Implements the DxcCreateInstance and DllMain functions for the dxil.dll module.
DxcSigningContainerBuilder.cpp: Implements the Dxil Container Builder.

This is first part for #6769


tools/clang/tools/dxildll/DxcSigningContainerBuilder.cpp
llvm-beanz, line 1: Can we rename this file to `DxcHashingContainerBuilder.cpp` and the class appropriately? It is hashing not signing.

tools/clang/tools/dxildll/dxcvalidatorp.cpp
```diff
@@ -0,0 +1,287 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// dxcvalidatorp.cpp                                                         //
+// Copyright (C) Microsoft Corporation. All rights reserved.                 //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+// Implements the DirectX Validator object, including signing support.       //
```
llvm-beanz, line 8:
```suggestion
// Implements the DirectX Validator object, including hashing support.       //
```

tools/clang/tools/dxildll/dxcvalidatorp.cpp
```diff
@@ -0,0 +1,287 @@
...
+#include "DxilHash.h"
+
+using namespace llvm;
+using namespace hlsl;
+
+class DxcValidator : public IDxcValidator2, public IDxcVersionInfo {
+private:
+  DXC_MICROCOM_TM_REF_FIELDS()
+
+  HRESULT RunValidation(_In_ IDxcBlob *pShader,
```
llvm-beanz, line 40: Can we get rid of the SAL annotations? We did from everywhere in DXC except the public API headers.

python3kgae: Fixed.

tools/clang/tools/dxildll/dxcvalidatorp.cpp
python3kgae: Fixed.

tools/clang/tools/dxildll/DxcSigningContainerBuilder.cpp
python3kgae: Fixed.

llvm-beanz: How about renaming the class?

python3kgae: Fixed.

llvm-beanz:
@python3kgae, I realize that you didn't write all this code and you're really just copying it from an internal fork into the public repo. That said, there are a lot of small things about this code that don't meet the LLVM coding standards. It might be worth spending a little bit of time cleaning up some of the low hanging fruit so that the code goes into the public repo in a cleaner state.

I pointed out a few places where the terminology "signing" is used to describe the hashing process. I think we need to correct all of those (there are still more). I know that historically this was called "signing", but it is not a signature and calling it a signature places an unmet expectation on what the code does.

python3kgae:
> @python3kgae, I realize that you didn't write all this code and you're really just copying it from an internal fork into the public repo. That said, there are a lot of small things about this code that don't meet the LLVM coding standards. It might be worth spending a little bit of time cleaning up some of the low hanging fruit so that the code goes into the public repo in a cleaner state.
> 
> I pointed out a few places where the terminology "signing" is used to describe the hashing process. I think we need to correct all of those (there are still more). I know that historically this was called "signing", but it is not a signature and calling it a signature places an unmet expectation on what the code does.

All the 'signning's are replaced with 'hashing' already.
I can update pSigned to pHashed next.
Do you have anything else in mind?

llvm-beanz:
> All the 'signning's are replaced with 'hashing' already. I can update pSigned to pHashed next. Do you have anything else in mind?

We spent months last year reviewing and cleaning up the SM 6.8 code to get closer to meeting LLVM's coding standards. I think spending a little time on this PR makes sense.

Some things to look for:
* `pSigned` should be `Hashed` or something similar. LLVM does not use Hungarian notation for naming  (see: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly). 
* There are several places where deep nesting could be simplified with early returns or just different control flow structures (see: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code).
* Windows-isms like `UINT` and `BYTE` could be replaced with C++-standard types.
* Lots of simple condition blocks have braces (see: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements)
* Look at the usage of DXC's macros (e.g. `IFR`, `IFRBOOL` etc) to see if they actually make sense.

I'm not asking that you rewrite this code, just spend a few minutes making sure that the code we introduce to DXC is a little nicer than what we had internally.

python3kgae:
> > All the 'signning's are replaced with 'hashing' already. I can update pSigned to pHashed next. Do you have anything else in mind?
> 
> We spent months last year reviewing and cleaning up the SM 6.8 code to get closer to meeting LLVM's coding standards. I think spending a little time on this PR makes sense.
> 
> Some things to look for:
> 
> * `pSigned` should be `Hashed` or something similar. LLVM does not use Hungarian notation for naming  (see: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly).
> * There are several places where deep nesting could be simplified with early returns or just different control flow structures (see: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code).
> * Windows-isms like `UINT` and `BYTE` could be replaced with C++-standard types.
> * Lots of simple condition blocks have braces (see: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements)
> * Look at the usage of DXC's macros (e.g. `IFR`, `IFRBOOL` etc) to see if they actually make sense.
> 
> I'm not asking that you rewrite this code, just spend a few minutes making sure that the code we introduce to DXC is a little nicer than what we had internally.

Updated Signed to Hashed.

For other issues, https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/tools/dxcompiler/dxcvalidator.cpp need to be updated as well.

Will do the job when work on [Build the DXIL validator library both as a dynamic and static library](https://github.com/microsoft/DirectXShaderCompiler/issues/6790) where these 2 validators could be merged into one.

damyanp:
> Will do the job when work on [Build the DXIL validator library both as a dynamic and static library](https://github.com/microsoft/DirectXShaderCompiler/issues/6790) where these 2 validators could be merged into one.

Can you help me understand the benefits of doing this later rather than now in this initial PR?  I assume that this change comes in with enough testing that we can make NFC refactors to the code without worrying about breaking it, and given that a bar for code style/quality has been set with SM6.8 work it isn't obvious to me why we'd change that for this change.

python3kgae:
> > Will do the job when work on [Build the DXIL validator library both as a dynamic and static library](https://github.com/microsoft/DirectXShaderCompiler/issues/6790) where these 2 validators could be merged into one.
> 
> Can you help me understand the benefits of doing this later rather than now in this initial PR? I assume that this change comes in with enough testing that we can make NFC refactors to the code without worrying about breaking it, and given that a bar for code style/quality has been set with SM6.8 work it isn't obvious to me why we'd change that for this change.

The code in
https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/tools/dxcompiler/dxcvalidator.cpp  and the validator code in this change are very similar and have similar issues.
 [Build the DXIL validator library both as a dynamic and static library](https://github.com/microsoft/DirectXShaderCompiler/issues/6790) will need to merge these 2 validators, where we only need to do the style change in one place.
If we do the style change only in the validator for current PR, when we merge the 2 validators, the code will get more difference just caused by style change.

damyanp:
> The code in https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/tools/dxcompiler/dxcvalidator.cpp and the validator code in this change are very similar and have similar issues. 

Oh, I didn't realize that this change was essentially adding a whole load of duplicate code. I think we should figure out how to deduplicate this before completing the PR, otherwise we're just adding more technical debt to the project.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -47,12 +48,50 @@ struct DiagRestore {
...
+      offsetof(struct DxilContainerHeader, Version);
+  const unsigned char *DataToHash =
+      (const unsigned char *)Container + DXBCHashStartOffset;
+  unsigned AmountToHash = Container->ContainerSizeInBytes - DXBCHashStartOffset;
+  ComputeHashRetail(DataToHash, AmountToHash, Container->Hash.Digest);
+}
+
+static void HashAndUpdateOrCopy(uint32_t Flags, IDxcBlob *Shader,
+                                IDxcBlob **Hashed) {
+  if (Flags & DxcValidatorFlags_InPlaceEdit) {
```
tex3d, line 64: If we were to add a flag to skip hashing and container modification called `DxcValidatorFlags_SkipHash`, we could implement it like so:

```suggestion
  if (Flags & DxcValidatorFlags_SkipHash) {
    *Hashed = Shader;
    Shader->AddRef();
  } else if (Flags & DxcValidatorFlags_InPlaceEdit) {
```

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -47,12 +48,50 @@ struct DiagRestore {
...
+    HashAndUpdate((DxilContainerHeader *)Shader->GetBufferPointer());
+    *Hashed = Shader;
+    Shader->AddRef();
+  } else {
+    // Possible gotcha: the blob allocated here is tied to this .dll, so the
+    // DLL shouldn't be unloaded before the blob is released.
```
tex3d, line 70: That's generally how returned blobs work in the DXC APIs.  Do we need to point this out as a comment here?  It probably belongs in the API documentation, and potentially as a comment on the API definition.

tools/clang/tools/dxrfallbackcompiler/dxcvalidator.cpp
```diff
@@ -64,7 +64,7 @@ HRESULT STDMETHODCALLTYPE DxcValidator::Validate(
     IDxcOperationResult *
         *ppResult // Validation output status, buffer, and errors
 ) {
-  return hlsl::validate(pShader, Flags, ppResult);
+  return hlsl::validate(pShader, Flags, true, ppResult);
```
tex3d, line 67: This should not be true, and I don't think we should have this implementation of IDxcValidator in dxrfallbackcompiler in the first place.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -155,24 +222,16 @@ runRootSignatureValidation(IDxcBlob *Shader,
 uint32_t hlsl::validate(
     IDxcBlob *Shader,            // Shader to validate.
     uint32_t Flags,              // Validation flags.
+    bool IsInternalValidator,    // Run internal validator.
```
tex3d, line 225: I don't think we should need this flag, ultimately.  But that will require a cleanup of how we do internal validation.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -47,12 +48,50 @@ struct DiagRestore {
...
+  }
+}
+
 static uint32_t runValidation(
     IDxcBlob *Shader,
     uint32_t Flags,            // Validation flags.
     llvm::Module *Module,      // Module to validate, if available.
     llvm::Module *DebugModule, // Debug module to validate, if available
-    AbstractMemoryStream *DiagMemStream) {
+    AbstractMemoryStream *DiagMemStream, IDxcBlob **Hashed) {
```
tex3d, line 94: This `Hashed` output pointer is not written to in this function, nor does it compute the hash.

This function is for the internal case, which I think needs an overhaul, especially now that we are integrating the hash here.

It's likely best if we did the overhaul as a separate change before integrating the hash changes.  It should greatly simplify the scenarios.

We should get rid of the special internal validator case where it accepts the main module as a `llvm::Module` and always validate the serialized bitcode in the container.  We will want to keep the debug module argument, so it may be passed along internally, avoiding extra deserialization, since it's only used for diagnostic locations, and not the thing being validated.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -191,32 +250,56 @@ uint32_t hlsl::validateWithDebug(
...
+    HR = CreateMemoryStream(TM.GetInstalledAllocator(), &DiagMemStream);
+    if (FAILED(HR))
+      throw hlsl::Exception(HR);
+
+    // Run validation may throw, but that indicates an inability to validate,
+    // not that the validation failed (eg out of memory).
+    if (Flags & DxcValidatorFlags_RootSignatureOnly)
+      ValidationStatus =
+          runRootSignatureValidation(Shader, DiagMemStream, Flags, &HashedBlob);
+    else if ((Flags & DxcValidatorFlags_ModuleOnly) && IsInternalValidator)
```
tex3d, line 268: `DxcValidatorFlags_ModuleOnly` should allow validation of raw bitcode, and disable hashing.  `IsInternalValidator` should not be required, and I don't think this code handles this case properly.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
tex3d: In fact, it appears that the only way this flag is used is to ignore `DxcValidatorFlags_ModuleOnly` if it's not set.  I don't think this is right.  See my comment on the use of this flag in `hlsl::validateWithDebug`.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -191,32 +250,56 @@ uint32_t hlsl::validateWithDebug(
...
-    if (FAILED(hr))
-      throw hlsl::Exception(hr);
-    raw_stream_ostream DiagStream(DiagMemStream);
-    llvm::DiagnosticPrinterRawOStream DiagPrinter(DiagStream);
-    PrintDiagnosticContext DiagContext(DiagPrinter);
-    Ctx.setDiagnosticHandler(PrintDiagnosticContext::PrintDiagnosticHandler,
-                             &DiagContext, true);
-    std::unique_ptr<llvm::Module> DebugModule;
-    if (OptDebugBitcode) {
-      hr = ValidateLoadModule((const char *)OptDebugBitcode->Ptr,
```
tex3d, line 209: Why has this been removed?  We should be passing the loaded debug module through to runValidation, right?

tex3d: Nevermind, it's just missing from the `DxcValidatorFlags_ModuleOnly` case, I'll make a separate comment.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -191,32 +250,56 @@ uint32_t hlsl::validateWithDebug(
...
+    if (FAILED(HR))
+      throw hlsl::Exception(HR);
+
+    // Run validation may throw, but that indicates an inability to validate,
+    // not that the validation failed (eg out of memory).
+    if (Flags & DxcValidatorFlags_RootSignatureOnly)
+      ValidationStatus =
+          runRootSignatureValidation(Shader, DiagMemStream, Flags, &HashedBlob);
+    else if ((Flags & DxcValidatorFlags_ModuleOnly) && IsInternalValidator)
+      ValidationStatus = runValidation(Shader, Flags, nullptr, nullptr,
```
tex3d, line 269: This should be passing the `OptDebugBitcode` along, though that will require deserialization in this overloaded `runValidation` path.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
tex3d: Ok, so this is really meant to be the inner function used by both code paths that have debug bitcode and the internal path that has a debug module.  We should probably use this in the other `runValidation` overload, and in the case of `DxcValidatorFlags_ModuleOnly` use a different function entirely to call this one which deserialized the debug module and does not have a `Hashed` output at all.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
python3kgae: Current dxil.dll doesn't support DxcValidatorFlags_ModuleOnly, just return DXC_E_CONTAINER_INVALID.
Shall we change that in this PR?

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
python3kgae: Removed.

tools/clang/tools/dxildll/dxcvalidatorp.cpp
tex3d, line 1: I don't think there's any need for the `p` at the end of the filename.  This is left over from the internal name.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
tex3d: Ok, I think this original structure should be restored so that a debug module may be passed down to a `validateWithOptModules` function that will apply the hash when applicable.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
python3kgae: Removed.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
python3kgae: Recovered debug module.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
python3kgae: Fixed.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -73,42 +110,69 @@ static uint32_t runValidation(
...
     return DXC_E_IR_VERIFICATION_FAILED;
-  }
 
   return S_OK;
 }
 
-static uint32_t
-runRootSignatureValidation(IDxcBlob *Shader,
-                           AbstractMemoryStream *DiagMemStream) {
+static uint32_t runValidation(IDxcBlob *Shader,
```
tex3d, line 144: I think it would be better if this was combined with the other overload, and the combined function handled whether `DxcValidatorFlags_ModuleOnly` was being performed, or if we have a full container to sign.  Right now it's confusing, and an internal path with a debug module can't go through the signing path.

tex3d: This can be accomplished by moving debug module serialization out, but it will also require a change to `ValidateDxilContainer` to take the `llvm::Module*` instead of the serialized bitcode.

tex3d: Hmm... perhaps this isn't sufficient, since it might need to load the debug module from the container instead.  So maybe we just need to add a debug `llvm::Module` argument to pass along if we already have one.

PR #6771, opened 2024-07-11
Add earlier check for invalid SV_Target[n]

### LINKED ISSUE #6115: Specifying an SV_Target<N> with <N> too large results in buffer overruns
**Steps to Reproduce**

See the following pixel shader: https://godbolt.org/z/YMMvvb8bG

With `SV_Target400`, DXC crashes.

With asserts enabled, `SV_Target32` or above results in this assert failing:
```
dxc: /home/amaiorano/src/external/DirectXShaderCompiler/lib/HLSL/ComputeViewIdStateBuilder.cpp:936: unsigned int (anonymous namespace)::DxilViewIdStateBuilder::GetLinearIndex(DxilSignatureElement &, int, unsigned int) const: Assertion `idx < kMaxSigScalars' failed.
Aborted
```

I debugged this, and the problem is that once SV_Target is 32 or more, DXC will end up copying memory beyond pre-allocated buffer sizes, specifically during `AssembleToContainer`. Once the value gets too large, like 400, the memcpy ends up stomping on data, and this results in crashes -- or with ASAN, memory leaks because, for example, a std::vector's begin pointer gets overwritten with 0.

Ultimately, the problem is one of timing. DXC compiles first with:
```cpp
        if (action.BeginSourceFile(compiler, file)) {
          action.Execute();
          action.EndSourceFile();
          compileOK = !compiler.getDiagnostics().hasErrorOccurred();
        }
```

And then validates later with:
```cpp
            valHR = dxcutil::ValidateAndAssembleToContainer(inputs);
```

But the memory stomping happens during compilation, so waiting until validation to report that the SV_Target value is too large (> 7) is too late. On that note, I'm not sure why there's a mismatch between the assertion's max size of 32 vs the validation's max size of 7.

A potential fix that seems to work for my test case is to modify `HLSignatureLower::ProcessArgument` so that it produces an error if the SV_Target is >= 32:

```cpp
  // Get Semantic interpretation, skipping if not in signature
  const Semantic *pSemantic = Semantic::GetByName(semanticStr);
  DXIL::SemanticInterpretationKind interpretation =
      SigPoint::GetInterpretation(pSemantic->GetKind(), sigPoint->GetKind(),
                                  pSM->GetMajor(), pSM->GetMinor());

// --------------------------- NEW CODE BEGIN -------------------------
  if (!paramAnnotation.GetSemanticIndexVec().empty()) {
    if (paramAnnotation.GetSemanticIndexVec()[0] >= 32u) {
          dxilutil::EmitErrorOnFunction(HLM.GetModule()->getContext(), func, "Semantic size too large");
          return;
    }
  }
// --------------------------- NEW CODE END -------------------------
```

This ensures that DXC bails before compiling. But I don't know if this is really the right place, or the right solution.

**Environment**
- DXC version: libdxcompiler.so: 1.8(dev;4386-a743e97f)
- Host Operating System: Linux (WSL)

sudonatalie:
According to the HLSL semantics documentation, the valid semantic indices for SV_Target[n] are 0 <= n <= 7:
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-semantics

A check for this already exists in DXIL validation, but for large values of n, crashes and/or buffer overruns may occur during compilation before validation, so an earlier check is needed.

Fixes #6115

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7963,7 +7963,7 @@ def warn_hlsl_legacy_integer_literal_signedness: Warning<
   "literal value is treated as signed in HLSL 2021 and earlier, and unsigned in later language versions">,
   InGroup<HLSLLegacyLiterals>, DefaultIgnore;
 def err_hlsl_unsupported_semantic_index: Error<
-  "'%0' is defined with semantic index %1, but only values %2 through %3 are supported">;
+  "'%0' is defined with semantic index %1, but only values 0 through %2 are supported">;
 // HLSL Change Ends
 
 // SPIRV Change Starts
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7962,6 +7962,8 @@ def warn_hlsl_barrier_no_mem_with_required_device_scope: Warning<
 def warn_hlsl_legacy_integer_literal_signedness: Warning<
   "literal value is treated as signed in HLSL 2021 and earlier, and unsigned in later language versions">,
   InGroup<HLSLLegacyLiterals>, DefaultIgnore;
+def err_hlsl_unsupported_semantic_index: Error<
+  "'%0' is defined with semantic index %1, but only values %2 through %3 are supported">;
 // HLSL Change Ends
 
 // SPIRV Change Starts
--- a/tools/clang/lib/Sema/SemaDecl.cpp
+++ b/tools/clang/lib/Sema/SemaDecl.cpp
@@ -11,8 +11,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "clang/Sema/SemaInternal.h"
 #include "TypeLocBuilder.h"
+#include "dxc/DXIL/DxilSemantic.h" // HLSL Change
 #include "clang/AST/ASTConsumer.h"
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/ASTLambda.h"
@@ -42,13 +42,14 @@
 #include "clang/Sema/ParsedTemplate.h"
 #include "clang/Sema/Scope.h"
 #include "clang/Sema/ScopeInfo.h"
+#include "clang/Sema/SemaHLSL.h" // HLSL Change
+#include "clang/Sema/SemaInternal.h"
 #include "clang/Sema/Template.h"
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/Triple.h"
 #include <algorithm>
 #include <cstring>
 #include <functional>
-#include "clang/Sema/SemaHLSL.h" // HLSL Change
 using namespace clang;
 using namespace sema;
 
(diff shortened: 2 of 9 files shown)


sudonatalie:
More detail about the possible failure modes in #6115, in particular, ASAN heap-buffer-overflows.

@llvm-beanz Let me know if you think there's a better place for this check to go.

llvm-beanz:
We should be able to do this somewhere in Sema right? Somewhere when we parse the attribute.

sudonatalie:
> We should be able to do this somewhere in Sema right? Somewhere when we parse the attribute.

I did briefly attempt that first, but I didn't see anywhere in the existing SemaHLSL logic that checks all semantics. There are some checks on the `SemanticDecl`s on an entry point but it seems like those don't catch decls on struct fields, and I wasn't confident I know all the cases to implement that traversal correctly. Am I overcomplicating it?

llvm-beanz:
All the `SemanticDecl`s are created in `Parser::MaybeParseHLSLAttributes`, but we don't really do much to validate them. I would add a function (`Sema::DiagnoseSemanticDecl(...)`) with a body in SemaHLSL that validates the `SemanticDecl`, and you could perform this check there. Then you could emit a diagnostic from Sema and test this more robustly with `-verify`.

dneto0:
+1  Generally user errors should be checked no later than Sema.  You still have the AST at that point and can emit good diagnostics.

sudonatalie:
Thanks for the code pointers! Updated to move it to Sema with `-verify` checks.

tools/clang/lib/Sema/SemaDecl.cpp
```diff
@@ -14629,3 +14630,21 @@ AvailabilityResult Sema::getCurContextAvailability() const {
...
+void Sema::DiagnoseSemanticDecl(hlsl::SemanticDecl *Decl) {
+  StringRef SemName = Decl->SemanticName;
+
+  StringRef BaseSemName; // The 'FOO' in 'FOO1'
+  uint32_t SemIndex;     // The '1' in 'FOO1'
+
+  // Split name and index.
+  hlsl::Semantic::DecomposeNameAndIndex(SemName, &BaseSemName, &SemIndex);
+
+  // The valid semantic indices for SV_Target[n] are 0 <= n <= 7.
```
dneto0, line 14643: nit: 7 is a bit magic.

It would be nice to have a named definition for this.
But it's not going to change any time soon.  I'm happy if Chris is happy.

dneto0 (approved):
LGTM.

Thanks!

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7962,6 +7962,8 @@ def warn_hlsl_barrier_no_mem_with_required_device_scope: Warning<
 def warn_hlsl_legacy_integer_literal_signedness: Warning<
   "literal value is treated as signed in HLSL 2021 and earlier, and unsigned in later language versions">,
   InGroup<HLSLLegacyLiterals>, DefaultIgnore;
+def err_hlsl_unsupported_semantic_index: Error<
+  "'%0' is defined with semantic index %1, but only values %2 through %3 are supported">;
```
llvm-beanz, line 7966: I assume the first value here is always `0` correct? If that's the case maybe it shouldn't be a variable. I realize that `%3` is currently only ever `7`, but we should absolutely add other verifications, but I think they would all be 0-?.

tools/clang/lib/Sema/SemaDecl.cpp
llvm-beanz: I think this is fine. 7 is a bit magic, but I tend to find named definitions hurt code readability, and there's really no way this changes anytime soon.

llvm-beanz (approved):
One minor nit, but I think this looks great.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
sudonatalie: Sounds right, done.

PR #6775, opened 2024-07-12
Fix debug info offsets for vectors with 16-bit types

gracezhang72:
This fixes a bug where the offsets for elements in vectors with 16-bit types doesn't take into account alignment bits and PIX wouldn't display vector element values correctly in the shader debugger. Eg. if `-enable-16bit-types` wasn't set, the offsets for a min16float4 would be 0, 16, 32, 48 instead of 0, 32, 64, 96.

Also removed the assert in PopulateAllocaMap_StructType that was checking whether the calculated aligned offset matches the packed offset (from SortedMembers) because it was false for members with sizes smaller than the alignment size.

### DIFF
--- a/lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
+++ b/lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
@@ -138,28 +138,8 @@ public:
     unsigned AlignMask = DescendTypeToGetAlignMask(Ty);
     if (AlignMask) {
       VALUE_TO_DECLARE_LOG("Aligning to %d", AlignMask);
-      // This is some magic arithmetic. Here's an example:
-      //
-      // Assume the natural alignment for Ty is 16 bits. Then
-      //
-      //     AlignMask = 0x0000000f(15)
-      //
-      // If the current aligned offset is
-      //
-      //     CurrentAlignedOffset = 0x00000048(72)
-      //
-      // Then
-      //
-      //     T = CurrentAlignOffset + AlignMask = 0x00000057(87)
-      //
-      // Which mean
-      //
-      //     T & ~AlignMask = 0x00000050(80)
-      //
-      // is the aligned offset where Ty should be placed.
-      AlignMask = AlignMask - 1;
       m_CurrentAlignedOffset =
-          (m_CurrentAlignedOffset + AlignMask) & ~AlignMask;
+          llvm::RoundUpToAlignment(m_CurrentAlignedOffset, AlignMask);
     } else {
       VALUE_TO_DECLARE_LOG("Failed to find alignment");
     }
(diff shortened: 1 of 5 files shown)


tools/clang/lib/CodeGen/CGDebugInfo.cpp
```diff
@@ -1043,14 +1043,37 @@ bool CGDebugInfo::TryCollectHLSLRecordElements(const RecordType *Ty,
...
     unsigned VecSize = hlsl::GetHLSLVecSize(QualTy);
     unsigned ElemSizeInBits = CGM.getContext().getTypeSize(ElemQualTy);
+    unsigned CurrentAlignedOffset = 0;
     for (unsigned ElemIdx = 0; ElemIdx < VecSize; ++ElemIdx) {
       StringRef FieldName = StringRef(&"xyzw"[ElemIdx], 1);
-      unsigned OffsetInBits = ElemSizeInBits * ElemIdx;
-      llvm::DIType *FieldType = createFieldType(FieldName, ElemQualTy, 0,
-        SourceLocation(), AccessSpecifier::AS_public, OffsetInBits,
-        /* tunit */ nullptr, DITy, Ty->getDecl());
+      // This matches the alignment calculation in OffsetManager in
```
adam-yang, line 1052: Does this achieve the same result as `llvm::RoundUpToAlignment`? Can we just call that instead?

adam-yang (approved):
LGTM!

adam-yang (approved):

tools/clang/test/DXC/min16vector_enabled.hlsl
```diff
@@ -0,0 +1,19 @@
+// Validate that the offsets for a min16float vector is correct if 16-bit types are enabled
+// RUN: %dxc -T cs_6_5 -Od -Zi -Qembed_debug -enable-16bit-types %s | FileCheck %s
+
+// CHECK-NOT: {{.*}}!DIDerivedType{{.*}}name: "x"{{.*}}offset: {{.*}}
```
jeffnn, line 4: Why's this one CHECK-NOT?

gracezhang72: The first element in the vector doesn't have an offset (is 0) so I'm double checking that there's no offset field!

lib/DxilPIXPasses/DxilDbgValueToDbgDeclare.cpp
```diff
@@ -1331,9 +1311,6 @@ void VariableRegisters::PopulateAllocaMap_StructType(
       // size would be lost
       PopulateAllocaMap(OffsetAndMember.second);
     } else {
-      assert(m_Offsets.GetCurrentAlignedOffset() ==
```
jeffnn, line 1334: I'm nervous about removing this assert. If GetCurrentAlignedOffset is wrong, can it be fixed to be right? I fear that if these are different, then we won't see any issues until data has made it all the way to the PIX UI.

gracezhang72: Synced up offline with Jeff, the aligned offsets are correct but are being compared to the packed offsets which don't match if the aligned bits are not the same as the member size. I re-added the assert back in and added a check for whether the size and alignment match instead of modifying SortMembers to return the aligned offsets, maybe we can revisit the aligned/packed code flow when we added more test coverage

jeffnn (approved):

PR #6777, opened 2024-07-12
Fix CB vector to scalar array translation generating invalid IR

### LINKED ISSUE #6510: Accessing "aggressively packed" array elements in a constant buffer via static cast results in ICE
As per https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-packing-rules#more-aggressive-packing

```glsl
#define MAX_TEXTURES 4
#define MAX_BUFFERS 4

cbuffer CResourceIndexRootConstants : register(b2)
{
    uint4 iTexs[MAX_TEXTURES / 4];
    uint4 iBufs[MAX_BUFFERS / 4];
}

float4 main(CPSInLighting pin) : SV_Target0
{
    static const uint iTexsArray[MAX_TEXTURES] = (uint[MAX_TEXTURES])iTexs;
    static const uint iBufsArray[MAX_BUFFERS] = (uint[MAX_BUFFERS])iBufs;

    uint iAlbedoTexture = iTexsArray[0];
    uint iNormalTexture = iTexsArray[1];
```

This results in:
```Internal compiler error: access violation. Attempted to read from address 0x0000000000000048```

`CResourceIndexRootConstants` is mapped via root constants.

Version: dxcompiler.dll: 1.8 - 1.8.0.14496 (main, 37c0487e9); dxil.dll: 1.8(101.8.2403.21)

### PR DESCRIPTION
**Pull request by tex3d**
In the special code to handle the memcpy pattern where a constant buffer contains a vector array that initializes a local (or static global) scalar array for use by the shader, an invalid assumption was made that if the memcpy dest was global, that the src is global as well.

This was not the case, and when expecting to generate constant expressions to index the src, these generated orphaned instructions instead, leading to invalid IR.

This fixes the issue by leveraging ReplaceConstantWithInst, and setting the insertion point for the Builder.  Now, replacement *could* fail, if src instructions don't dominate replacement uses, so bool for replaced all is returned from replaceScalarArrayWithVectorArray.

Another issue was that it would replace the dest for the original memcpy with src along the way.  Now, if we don't replace all uses, this turns the memcpy into a no-op and any remaining uses are no longer coming from src, but an undef dest instead.  This was also fixed to skip this replacement, then clean up this use if all other uses have been successfully replaced.

Fixes #6510

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -3427,7 +3427,7 @@ static void updateLifetimeForReplacement(Value *From, Value *To) {
 static bool DominateAllUsers(Instruction *I, Value *V, DominatorTree *DT);
 
 namespace {
-void replaceScalarArrayGEPWithVectorArrayGEP(User *GEP, Value *VectorArray,
+bool replaceScalarArrayGEPWithVectorArrayGEP(User *GEP, Value *VectorArray,
                                              IRBuilder<> &Builder,
                                              unsigned sizeInDwords) {
   gep_type_iterator GEPIt = gep_type_begin(GEP), E = gep_type_end(GEP);
@@ -3461,11 +3461,26 @@ void replaceScalarArrayGEPWithVectorArrayGEP(User *GEP, Value *VectorArray,
   Value *CompIdx = Builder.CreateAnd(ArrayIdx, mask);
   Value *NewGEP = Builder.CreateGEP(
       VecPtr, {ConstantInt::get(CompIdx->getType(), 0), CompIdx});
-  GEP->replaceAllUsesWith(NewGEP);
+
+  if (isa<ConstantExpr>(GEP) && isa<Instruction>(NewGEP)) {
+    if (!ReplaceConstantWithInst(cast<Constant>(GEP), NewGEP, Builder)) {
+      // If new instructions unable to be used, clean them up.
+      if (NewGEP->user_empty())
+        cast<Instruction>(NewGEP)->eraseFromParent();
+      if (VecPtr->user_empty())
+        cast<Instruction>(VecPtr)->eraseFromParent();
+      return false;
+    }
+    return true;
+  } else {
+    GEP->replaceAllUsesWith(NewGEP);
+  }
+  return true;
 }
 
-void replaceScalarArrayWithVectorArray(Value *ScalarArray, Value *VectorArray,
+bool replaceScalarArrayWithVectorArray(Value *ScalarArray, Value *VectorArray,
                                        MemCpyInst *MC, unsigned sizeInDwords) {
+  bool bReplacedAll = true;
   LLVMContext &Context = ScalarArray->getContext();
   // All users should be element type.
   // Replace users of AI or GV.
@@ -3474,24 +3489,31 @@ void replaceScalarArrayWithVectorArray(Value *ScalarArray, Value *VectorArray,
     if (U->user_empty())
       continue;
     if (BitCastInst *BCI = dyn_cast<BitCastInst>(U)) {
-      BCI->setOperand(0, VectorArray);
+      // Avoid replacing the dest of the memcpy to support partial replacement.
+      if (MC->getArgOperand(0) != BCI)
+        BCI->setOperand(0, VectorArray);
       continue;
     }
 
     if (ConstantExpr *CE = dyn_cast<ConstantExpr>(U)) {
       IRBuilder<> Builder(Context);
+      // If we need to replace the constant with an instruction, start at the
+      // memcpy, so we replace only users dominated by it.
+      if (isa<Instruction>(VectorArray))
+        Builder.SetInsertPoint(MC);
+
       if (GEPOperator *GEP = dyn_cast<GEPOperator>(U)) {
-        // NewGEP must be GEPOperator too.
-        // No instruction will be build.
-        replaceScalarArrayGEPWithVectorArrayGEP(U, VectorArray, Builder,
-                                                sizeInDwords);
+        if (!replaceScalarArrayGEPWithVectorArrayGEP(U, VectorArray, Builder,
+                                                sizeInDwords))
+          bReplacedAll = false;
       } else if (CE->getOpcode() == Instruction::AddrSpaceCast) {
         Value *NewAddrSpaceCast = Builder.CreateAddrSpaceCast(
             VectorArray,
             PointerType::get(VectorArray->getType()->getPointerElementType(),
                              CE->getType()->getPointerAddressSpace()));
-        replaceScalarArrayWithVectorArray(CE, NewAddrSpaceCast, MC,
-                                          sizeInDwords);
+        if (!replaceScalarArrayWithVectorArray(CE, NewAddrSpaceCast, MC,
+                                           sizeInDwords))
+          bReplacedAll = false;
       } else if (CE->hasOneUse() && CE->user_back() == MC) {
         continue;
       } else {
@@ -3499,13 +3521,16 @@ void replaceScalarArrayWithVectorArray(Value *ScalarArray, Value *VectorArray,
       }
     } else if (GetElementPtrInst *GEP = dyn_cast<GetElementPtrInst>(U)) {
       IRBuilder<> Builder(GEP);
-      replaceScalarArrayGEPWithVectorArrayGEP(U, VectorArray, Builder,
-                                              sizeInDwords);
-      GEP->eraseFromParent();
+      if (!replaceScalarArrayGEPWithVectorArrayGEP(U, VectorArray, Builder,
+                                               sizeInDwords))
+        bReplacedAll = false;
+      else
+        GEP->eraseFromParent();
     } else {
       DXASSERT(0, "not implemented");
     }
   }
+  return bReplacedAll;
 }
 
 // For pattern like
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp [hunk 1]
```diff
@@ -3521,8 +3546,23 @@ bool tryToReplaceCBVec4ArrayToScalarArray(Value *V, Type *TyV, Value *Src,
   Type *EltTy = AT->getElementType();
   unsigned sizeInBits = DL.getTypeSizeInBits(EltTy);
   // Convert array of float4 to array of float.
-  replaceScalarArrayWithVectorArray(V, Src, MC, sizeInBits >> 5);
-  return true;
+  if (replaceScalarArrayWithVectorArray(V, Src, MC, sizeInBits >> 5)) {
+    Value *DstBC = MC->getArgOperand(0);
+    MC->setArgOperand(0, UndefValue::get(MC->getArgOperand(0)->getType()));
+    DXASSERT(DstBC->user_empty(), "memcpy bitcast used by something else?");
+    if (DstBC->user_empty()) {
```
python3kgae, line 3553: Assert and then check the same thing feels strange.
Is it because DXASSERT only work for debug build?

tex3d: I wouldn't expect the bitcast to be used by anything else, but if it is and we are on a release build, we don't want to make the IR invalid.  Do you have a better pattern to suggest for this?

damyanp: Isn't the setArgOperand on line 3551 destructive, so that the IR is invalid here?

Wondering why this isn't an internal error?

tex3d: The setArgOperand clears the use in the replaced memcpy so it will be cleaned up as dead code later.  The idea is to get rid of any (now) dead users of `Dst` (or `V`) so we may delete it after the replacement.

We don't expect other users, but the presence of other users wouldn't represent invalid IR.  So, this asserts if the unexpected happens (other users), but still handles it properly (when asserts are disabled) by avoiding modification of the bitcast operand if there are other users, so it doesn't invalidate the IR for them.

If one of the asserts goes off, it's an indication that local assumptions may be wrong, but IR could still be perfectly valid.  Perhaps I should simply remove the asserts, but I'd be interested seeing the scenario which causes either these asserts to fire.

damyanp: I don't feel strongly enough to block on it.  But...my 2c is that if you really want to be sure you hear about any cases where the assumption is wrong then relying on asserts won't help.

damyanp (approved):

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
tex3d: Would this be more appropriate?
```diff
   // Convert array of float4 to array of float.
   if (replaceScalarArrayWithVectorArray(V, Src, MC, sizeInBits >> 5)) {
     Value *DstBC = MC->getArgOperand(0);
     MC->setArgOperand(0, UndefValue::get(MC->getArgOperand(0)->getType()));
-    DXASSERT(DstBC->user_empty(), "memcpy bitcast used by something else?");
     if (DstBC->user_empty()) {
       // Replacement won't include the memcpy dest.  Now remove that use.
       if (BitCastInst *BCI = dyn_cast<BitCastInst>(DstBC)) {
         Value *Dst = BCI->getOperand(0);
         Type *DstTy = Dst->getType();
-        DXASSERT(Dst == V, "otherwise unexpected dest");
         if (Dst == V)
           BCI->setOperand(0, UndefValue::get(DstTy));
+        else
+          llvm_unreachable("Unexpected dest of memcpy.")
       }
+    } else {
+      llvm_unreachable("Unexpected users of memcpy bitcast.");
     }
     return true;
   }
   return false;
```

tex3d: I've pushed that change, and will resolve this comment.  Please comment again if there are still concerns.

damyanp (approved):

python3kgae (approved):

python3kgae:
Would be nice to remove the debug locations in the tests.

tex3d:
> Would be nice to remove the debug locations in the tests.

That would be a pain unless we have a pass that does this, which I could incorporate into the `ExtractIRForPassTest.py` script as an option.  Plus, the line-only debug info doesn't add that much.

tools/clang/test/HLSLFileCheck/passes/hl/sroa_hlsl/array-from-cbvec-1.ll
```diff
@@ -0,0 +1,146 @@
...
+; scalar array, then that the scalar array is used for the result.
+
+; Generated using:
+; ExtractIRForPassTest.py -p scalarrepl-param-hlsl -o array-to-cbvec-1.ll array-to-cbvec-1.hlsl -- -T vs_6_0
+; uint4 VectorArray[2];
+;
+; uint2 main(int i : IN) : OUT {
+;     static const uint ScalarArray[8] = (uint[8])VectorArray;
+;     return uint2(ScalarArray[1], ScalarArray[6]);
+; }
```
pow2clk, line 19: Love this! I'll have to start doing that too.

pow2clk (commented):
Looks good. I have some minor questions about cleaning up generated instructions and clarification of the dest param of memcpy.

pow2clk:
Feel free to resolve my comments once you've responded to them. Assuming they don't make you think any further changes are needed, I won't object to committing as-is.

python3kgae:
> > Would be nice to remove the debug locations in the tests.
> 
> That would be a pain unless we have a pass that does this, which I could incorporate into the `ExtractIRForPassTest.py` script as an option. Plus, the line-only debug info doesn't add that much.

Will  -fdisable-loc-tracking help?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3461,11 +3461,26 @@ void replaceScalarArrayGEPWithVectorArrayGEP(User *GEP, Value *VectorArray,
...
       VecPtr, {ConstantInt::get(CompIdx->getType(), 0), CompIdx});
-  GEP->replaceAllUsesWith(NewGEP);
+
+  if (isa<ConstantExpr>(GEP) && isa<Instruction>(NewGEP)) {
+    if (!ReplaceConstantWithInst(cast<Constant>(GEP), NewGEP, Builder)) {
+      // If new instructions unable to be used, clean them up.
+      if (NewGEP->user_empty())
+        cast<Instruction>(NewGEP)->eraseFromParent();
+      if (VecPtr->user_empty())
+        cast<Instruction>(VecPtr)->eraseFromParent();
```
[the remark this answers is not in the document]
tex3d: Those will never be instructions (always constant expressions).  This is because we are replacing a GEP that used to be a constant expression using a constant pointer with one that needs to be an instruction because the new pointer is an instruction instead.  That means when we create the math ops, we actually just get constant expression evaluation with nothing we should worry about cleaning up at this point.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3474,38 +3489,48 @@ void replaceScalarArrayWithVectorArray(Value *ScalarArray, Value *VectorArray,
     if (U->user_empty())
       continue;
     if (BitCastInst *BCI = dyn_cast<BitCastInst>(U)) {
-      BCI->setOperand(0, VectorArray);
+      // Avoid replacing the dest of the memcpy to support partial replacement.
+      if (MC->getArgOperand(0) != BCI)
+        BCI->setOperand(0, VectorArray);
```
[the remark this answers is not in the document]
tex3d: AFAK, the dest of the memcpy inst will always be a bitcast instruction (bitcast to i8*), since that's the way it generates the memcpy.  I believe we have a lot of other code here that depends on that assumption.  I don't think there ever should be an addrspacecast between the dest pointer and the memcpy.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3474,38 +3489,48 @@ void replaceScalarArrayWithVectorArray(Value *ScalarArray, Value *VectorArray,
...
       } else if (CE->getOpcode() == Instruction::AddrSpaceCast) {
         Value *NewAddrSpaceCast = Builder.CreateAddrSpaceCast(
             VectorArray,
             PointerType::get(VectorArray->getType()->getPointerElementType(),
                              CE->getType()->getPointerAddressSpace()));
-        replaceScalarArrayWithVectorArray(CE, NewAddrSpaceCast, MC,
-                                          sizeInDwords);
+        if (!replaceScalarArrayWithVectorArray(CE, NewAddrSpaceCast, MC,
+                                               sizeInDwords))
+          bReplacedAll = false;
```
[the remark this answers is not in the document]
tex3d: We could if it's an instruction with no users in this branch.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3521,8 +3546,25 @@ bool tryToReplaceCBVec4ArrayToScalarArray(Value *V, Type *TyV, Value *Src,
   Type *EltTy = AT->getElementType();
   unsigned sizeInBits = DL.getTypeSizeInBits(EltTy);
   // Convert array of float4 to array of float.
-  replaceScalarArrayWithVectorArray(V, Src, MC, sizeInBits >> 5);
-  return true;
+  if (replaceScalarArrayWithVectorArray(V, Src, MC, sizeInBits >> 5)) {
+    Value *DstBC = MC->getArgOperand(0);
+    MC->setArgOperand(0, UndefValue::get(MC->getArgOperand(0)->getType()));
+    if (DstBC->user_empty()) {
+      // Replacement won't include the memcpy dest.  Now remove that use.
```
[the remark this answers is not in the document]
tex3d: Yes, we have a fair bit of code that relies on the assumption that a bitcast to i8* always feeds the memcpy.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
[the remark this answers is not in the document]
tex3d: Oh, hang on, that's not right.  They may not be constant expressions, I was thinking of the replace-constant-with-instruction path for some reason.

Ok, we can delete these if they are unused instructions, though they are less consequential due to not impacting uses of VectorArray and the ability to clean up other intermediates related to that, which could have other impacts before we get to DCE.

PR #6779, opened 2024-07-13
[Sema] Prevent unwanted conversions related to ResourceDescriptorHeap/SamplerDescriptorHeap

### LINKED ISSUE #6649: SamplerDescriptorHeap can be used to load textures.
**Description**

From [HLSL_SM_6_6_DynamicResources](https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_DynamicResources.html), it looks like SamplerDescriptorHeap must be used to access Samplers, not textures.
But if we use it to load a texture, DXC says OK.

**Steps to Reproduce**

```hlsl
float4 main() : SV_Target {
  SamplerState Sampler = SamplerDescriptorHeap[2];
  Texture2D Texture = SamplerDescriptorHeap[3];
  return Texture.Sample(Sampler, float2(0, 0));
}
```

```
dxc -T ps_6_6 repro.hlsl
```

**Actual Behavior**

DXC compiles. `dx.op.createHandleFromHeap` is created, loading a texture from a sampler heap.
As I'm adding this feature in the SPIR-V backend, I'm fine accepting this (since for us both heaps would be the same). But it looks like this should be illegal.

**Environment**
- DXC version: HEAD
- Host Operating System: linux

mateeeeeee:
Prevents unwanted conversions between ResourceDescriptorHeap elements and Sampler objects, and SamplerDescriptorHeap and other types of resources besides Samplers. Fixes #6649


mateeeeeee:
I've accidentally checked-out the branch from another branch that contained some changes from before. This pull request should be deleted. Sorry for inconvenience.

PR #6457, opened 2024-03-25
[Sema] Check FunctionDecl has identifier before getName. (#6439)

python3kgae:
Use identifier name without check the identifier exists will cause crash.

Fixes #6426

---------

Co-authored-by: Tex Riddell <texr@microsoft.com>
Co-authored-by: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
(cherry picked from commit 7581ff4d297d124b340d6715b709f1c423909ca5)

PR #6786, opened 2024-07-15
Merge needed march release followups into mesh nodes

### LINKED ISSUE #6426: Regression, SIGSEGV instead of diagnostics when encountering bool operator==(const T&, const T&)

pow2clk:
Merges the following changes into mesh nodes to fix a known bug that required a point release in the march release branch and also some build fixes for new visual studio diagnostics. 

9ca52f4e0 Add duplicate pragma (#6732)
7809c0bd3 Remove Windows C++ redist hack (#6692)
33277a123 Workaround broken GitHub runner images (#6683)
1b9a79643 [Sema] Check FunctionDecl has identifier before getName. (#6439) (#6457)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,12 +120,6 @@ if(WIN32 AND NOT (CMAKE_C_COMPILER_ARCHITECTURE_ID MATCHES "ARM.*"))
   add_link_options(/CETCOMPAT)
 endif()
 
-# This is a hack workaround for the broken azure runner images:
-# https://github.com/actions/runner-images/issues/10004
-if(WIN32)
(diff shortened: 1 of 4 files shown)


bob80905 (approved):

pow2clk:
The failure looks like a fluke. I can't see any actual failure message in the log. Rerunning

pow2clk:
This raft of cherry picks that collectively change just one line should be enough to fix the debug build and also keep the history consistent with main

damyanp (approved):

PR #6789, opened 2024-07-16
Release Note for OS-normalized include paths

### LINKED ISSUE #6669: New include path behaviour
**Description**

Due to #6317 include paths are now passed to `IDxcIncludeHandler::LoadSource` normalized to the operating system's path style. i.e. On Windows "/" is replaced with "\\". 

The issue with this is that it likely breaks some implementations of `IDxcIncludeHandler` because users have been expecting the argument of `IDxcIncludeHandler::LoadSource` to be exactly what was in the include directive in the shader source. There are not many programmers who would prefer to write "\\" over "/" in their header includes, so for some, upgrading to 1.8.2405.17 will have broken their include handler.

**Steps to Reproduce**

Write a shader that has an include such as `#include "Mylibrary/MyHeader.hlsli"` and compile it with DXC using a custom `IDxcIncludeHandler`.

**Actual Behavior**

Notice that the string passed to `IDxcIncludeHandler::LoadSource` is `".\\MyLibrary\\Myheader.hlsli"` which is different from the string used in the `#include` directive

**Environment**
- DXC version 1.8.2405.17
- Host Operating System Windows 10 Pro 22H2

pow2clk:
The behavior was changed with #6317 so that regardless of spelling in the shader, the include path will conform to the host OS style for the purposes of the include handler. This just adds a release note for that new behavior.

Fixes #6669

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -69,6 +69,9 @@ DX Compiler release for March 2024
   - Validation errors more accurately determine usage by the entry point
 - Improve debug info generation
 - Further improvements to Linux build quality
+- Include paths input to custom include handlers are now normalized to the OS path style
+  - This means that regardless of how they are denoted in the shader, the string passed to `IDxcIncludeHandler::LoadSource` will use OS-normalized paths
+  - This could break existing code
 
 ### Version 1.7.2308
 


python3kgae (approved):

damyanp:
@python3kgae - have you done the final merge of main -> release for July release?  If so then we'll need to cherry-pick this into release.

damyanp (approved):

tex3d (approved):

PR #6800, opened 2024-07-18
Fix CB vector to scalar array translation generating invalid IR (#6777)

### LINKED ISSUE #6510: Accessing "aggressively packed" array elements in a constant buffer via static cast results in ICE

python3kgae:
In the special code to handle the memcpy pattern where a constant buffer contains a vector array that initializes a local (or static global) scalar array for use by the shader, an invalid assumption was made that if the memcpy dest was global, that the src is global as well.

This was not the case, and when expecting to generate constant expressions to index the src, these generated orphaned instructions instead, leading to invalid IR.

This fixes the issue by leveraging ReplaceConstantWithInst, and setting the insertion point for the Builder. Now, replacement *could* fail, if src instructions don't dominate replacement uses, so bool for replaced all is returned from replaceScalarArrayWithVectorArray.

Another issue was that it would replace the dest for the original memcpy with src along the way. Now, if we don't replace all uses, this turns the memcpy into a no-op and any remaining uses are no longer coming from src, but an undef dest instead. This was also fixed to skip this replacement, then clean up this use if all other uses have been successfully replaced.

Fixes #6510

(cherry picked from commit 5cfefc7d0b2d4d5af93d5668dcd5c11a003ac666)

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -3427,7 +3427,7 @@ static void updateLifetimeForReplacement(Value *From, Value *To) {
 static bool DominateAllUsers(Instruction *I, Value *V, DominatorTree *DT);
 
 namespace {
-void replaceScalarArrayGEPWithVectorArrayGEP(User *GEP, Value *VectorArray,
+bool replaceScalarArrayGEPWithVectorArrayGEP(User *GEP, Value *VectorArray,
                                              IRBuilder<> &Builder,
                                              unsigned sizeInDwords) {
   gep_type_iterator GEPIt = gep_type_begin(GEP), E = gep_type_end(GEP);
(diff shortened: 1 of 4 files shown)


damyanp (approved):

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -3474,38 +3493,52 @@ void replaceScalarArrayWithVectorArray(Value *ScalarArray, Value *VectorArray,
     if (U->user_empty())
       continue;
     if (BitCastInst *BCI = dyn_cast<BitCastInst>(U)) {
-      BCI->setOperand(0, VectorArray);
+      // Avoid replacing the dest of the memcpy to support partial replacement.
+      if (MC->getArgOperand(0) != BCI)
```
bob80905, line 3497: Should there be a nullptr check for `MC` ?

bob80905:
Is there a way to shrink the sizes of the 3 tests?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
python3kgae: No.
The MC is from lowerMemcpy which start when have memcpy.

python3kgae:
> Is there a way to shrink the sizes of the 3 tests?

We could remove the debug info.
But this is a cherry-pick PR, we should need to change in main and then cherry-pick the test update to release branch.

bob80905 (approved):

PR #6805, opened 2024-07-18
[Metal] Add experimental Metal support

llvm-beanz:
This adds a new `-metal` flag to DXC which can be used to generate Metal's IR directly from DXC after compilation. There are some limitations in this flag which are worth noting:

1) It does not support library shaders (yet)
2) It does not support disassembly (yet)
3) It is _wildly_ under tested because wtihout (2) we can't do anything to really verify correct output (yay?)

### DIFF
--- a/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
+++ b/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
@@ -1176,18 +1176,18 @@ public:
                    "Library targets not supported for Metal (yet).");
             IRObjectGetMetalLibBinary(AIR, Stage, MetalLib);
             size_t MetalLibSize = IRMetalLibGetBytecodeSize(MetalLib);
-            uint8_t *MetalLibBytes = new uint8_t[MetalLibSize];
-            IRMetalLibGetBytecode(MetalLib, MetalLibBytes);
+            std::unique_ptr<uint8_t[]> MetalLibBytes =
+                std::unique_ptr<uint8_t[]>(new uint8_t[MetalLibSize]);
+            IRMetalLibGetBytecode(MetalLib, MetalLibBytes.get());
 
             // Store the metallib to custom format or disk, or use to create a
             // MTLLibrary.
 
             CComPtr<IDxcBlob> MetalBlob;
             IFT(hlsl::DxcCreateBlobOnHeapCopy(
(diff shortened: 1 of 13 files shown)


bogner (approved):

damyanp (approved):

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1107,7 +1115,86 @@ class DxcCompiler : public IDxcCompiler3,
...
+
+            // Store the metallib to custom format or disk, or use to create a
+            // MTLLibrary.
+
+            CComPtr<IDxcBlob> MetalBlob;
+            IFT(hlsl::DxcCreateBlobOnHeapCopy(
+                MetalLibBytes, (uint32_t)MetalLibSize, &MetalBlob));
+            std::swap(pOutputBlob, MetalBlob);
+
+            delete[] MetalLibBytes;
```
V-FEXrt, line 1190: I still have so much ownership context to untangle in my head :(

In this case, I'm assuming `DxcCreateBlobOnHeapCopy` does the `new` and passes  ownership back?

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1107,7 +1115,86 @@ class DxcCompiler : public IDxcCompiler3,
...
+              break;
+            case DXIL::ShaderKind::Compute:
+              Stage = IRShaderStageCompute;
+              break;
+            }
+            assert(Stage != IRShaderStageInvalid &&
+                   "Library targets not supported for Metal (yet).");
+            IRObjectGetMetalLibBinary(AIR, Stage, MetalLib);
+            size_t MetalLibSize = IRMetalLibGetBytecodeSize(MetalLib);
+            uint8_t *MetalLibBytes = new uint8_t[MetalLibSize];
```
V-FEXrt, line 1179: nit: Is there a particularly good reason to not use a smart pointer here?

V-FEXrt (approved):

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
llvm-beanz: No, that's a good suggestion. Will push an update.

PR #6807, opened 2024-07-19
Remove WaveMatrix

python3kgae:
Remove WaveMatrix from main branch.
The DXIL operations are changed to Depreated to avoid affect DXIL operation ID.


include/dxc/DXIL/DxilConstants.h
```diff
@@ -1278,18 +1274,6 @@ enum class OpCodeClass : unsigned {
...
-  WaveMatrix_Fill,
-  WaveMatrix_LoadGroupShared,
-  WaveMatrix_LoadRawBuf,
-  WaveMatrix_Multiply,
-  WaveMatrix_ScalarOp,
-  WaveMatrix_StoreGroupShared,
-  WaveMatrix_StoreRawBuf,
-
   // Work Graph intrinsics
   FinishedCrossGroupSharing,
```
damyanp, line 1278: Will these still have the right values?

include/dxc/DXIL/DxilInstructions.h
```diff
@@ -8042,498 +8042,234 @@ struct DxilInst_TextureStoreSample {
   void set_sampleIdx(llvm::Value *val) { Instr->setOperand(10, val); }
 };
 
-/// This instruction Annotate a wave matrix pointer with the type information
-struct DxilInst_WaveMatrix_Annotate {
+/// This instruction Deprecated
+struct DxilInst_Deprecated0 {
```
damyanp, line 8046: is there no way to do this that doesn't involve having a load of stub instruction structs?

lib/DXIL/DxilShaderFlags.cpp
```diff
@@ -721,20 +721,6 @@ ShaderFlags ShaderFlags::CollectShaderFlags(const Function *F,
...
-        case DXIL::OpCode::WaveMatrix_Fill:
-        case DXIL::OpCode::WaveMatrix_LoadGroupShared:
-        case DXIL::OpCode::WaveMatrix_LoadRawBuf:
-        case DXIL::OpCode::WaveMatrix_Multiply:
-        case DXIL::OpCode::WaveMatrix_MultiplyAccumulate:
-        case DXIL::OpCode::WaveMatrix_ScalarOp:
-        case DXIL::OpCode::WaveMatrix_StoreGroupShared:
-        case DXIL::OpCode::WaveMatrix_StoreRawBuf:
-        case DXIL::OpCode::WaveMatrix_SumAccumulate:
-          hasWaveMMA = true;
```
damyanp, line 736: is `hasWaveMMA` still needed?

include/dxc/DXIL/DxilConstants.h
python3kgae: Yes. The removed operations are changed into Deprecated.

lib/DXIL/DxilShaderFlags.cpp
python3kgae: Not sure if we want to remove feature flag about WaveMMA or not.
That part should be the same if WaveMatrix is needed in the future.

include/dxc/DXIL/DxilConstants.h
damyanp: Ah, so the numeric values of the OpCodeClass entries don't matter?  

(I was misreading the diff and thought these were in the OpCode enum)

lib/DXIL/DxilShaderFlags.cpp
damyanp: IMO we're either removing this feature entirely or we're not.  This maybe warrants further discussion outside of the PR?

include/dxc/DXIL/DxilConstants.h
python3kgae: Yes. OpCodeClass in DXIL is mapped to the Op function name.
The enum value for OpCodeClass doesn't matter.

lib/DXIL/DxilShaderFlags.cpp
python3kgae: Removed.

docs/DXIL.rst
```diff
@@ -2337,18 +2337,18 @@ ID  Name                                                  Description
...
-229 WaveMatrix_LoadRawBuf                                 Load wave matrix from raw buffer
-230 WaveMatrix_LoadGroupShared                            Load wave matrix from group shared array
-231 WaveMatrix_StoreRawBuf                                Store wave matrix to raw buffer
-232 WaveMatrix_StoreGroupShared                           Store wave matrix to group shared array
-233 WaveMatrix_Multiply                                   Mutiply left and right wave matrix and store in accumulator
-234 WaveMatrix_MultiplyAccumulate                         Mutiply left and right wave matrix and accumulate into accumulator
-235 WaveMatrix_ScalarOp                                   Perform scalar operation on each element of wave matrix
-236 WaveMatrix_SumAccumulate                              Sum rows or columns of an input matrix into an existing accumulator fragment matrix
-237 WaveMatrix_Add                                        Element-wise accumulate, or broadcast add of fragment into accumulator
+226 Deprecated0                                           Deprecated
```
damyanp, line 2340: Should we call these "reserved" rather than "deprecated"?

include/dxc/DXIL/DxilShaderFlags.h
```diff
@@ -204,6 +204,8 @@ class ShaderFlags {
   bool GetWriteableMSAATextures() const { return m_bWriteableMSAATextures; }
 
   // SM 6.8+
+  void setReserved(bool flag) { m_bReserved = flag; }
```
damyanp, line 207: Presumably `setReserved` is never / should never be called?  Any reason to keep it?

tools/dxexp/dxexp.cpp
```diff
@@ -147,18 +147,12 @@ typedef struct D3D12_FEATURE_DATA_D3D12_OPTIONS7 {
...
-} D3D12_WAVE_MMA_TIER;
-
 typedef struct D3D12_FEATURE_DATA_D3D12_OPTIONS9 {
   BOOL MeshShaderPipelineStatsSupported;
   BOOL MeshShaderSupportsFullRangeRenderTargetArrayIndex;
   BOOL AtomicInt64OnTypedResourceSupported;
   BOOL AtomicInt64OnGroupSharedSupported;
   BOOL DerivativesInMeshAndAmplificationShadersSupported;
-  D3D12_WAVE_MMA_TIER WaveMMATier;
 } D3D12_FEATURE_DATA_D3D12_OPTIONS9;
```
hekota, line 156: This might need to stay in, it is already shipped:
https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_feature_data_d3d12_options9

include/dxc/DXIL/DxilShaderFlags.h
python3kgae: To work around the warning on Linux build for unused private field.

docs/DXIL.rst
python3kgae: Updated.

include/dxc/DXIL/DxilShaderFlags.h
damyanp: Well, that warning is broken because it is still unused :)  

Is there any other way we can avoid the warning that doesn't involve adding dummy code?

tools/dxexp/dxexp.cpp
damyanp: Presumably these structs should match the ones in [d3d12.h](https://github.com/microsoft/DirectX-Headers/blob/9be295b3b81ce1d0ff2b44f18d0eb86ea54c5122/include/directx/d3d12.h#L2757)

python3kgae: Updated.

include/dxc/DXIL/DxilShaderFlags.h
python3kgae: Changed to diagnostic push/pop.

docs/DXIL.rst
python3kgae: Fixed. Forgot to enable doc build in cmake.

include/dxc/DXIL/DxilShaderFlags.h
damyanp: I'm surprised there's no better way to do this!

damyanp (approved):

include/dxc/DXIL/DxilConstants.h
```diff
@@ -474,6 +474,20 @@ inline bool IsFeedbackTexture(DXIL::ResourceKind ResourceKind) {
...
+  Reserved10 = 236, // Reserved
+  Reserved11 = 237, // Reserved
+  Reserved2 = 228,  // Reserved
+  Reserved3 = 229,  // Reserved
+  Reserved4 = 230,  // Reserved
+  Reserved5 = 231,  // Reserved
+  Reserved6 = 232,  // Reserved
+  Reserved7 = 233,  // Reserved
+  Reserved8 = 234,  // Reserved
+  Reserved9 = 235,  // Reserved
```
coopp, line 489: I don't see 236 or 237 in this list.

damyanp: I do.

coopp: Ah!  They are out of order!

coopp (approved):
Looks good to me.

Issue #5294, opened 2023-06-14
"Wrong type in array element initializer" assertion when asserts are enabled


amaiorano:
When building dxc with assertions enabled, attempting to compile the following shader asserts:

Shader `struct_in_array.hlsl`:
```
struct str {
  int i;
};

str func(inout str pointer) {
  return pointer;
}

static str P[4] = (str[4])0;

[numthreads(1, 1, 1)]
void main() {
  const str r = func(P[2]);
  return;
}
```

```shell
C:\src\external\hlsl.bin>.\Release\bin\dxc /T cs_6_0 C:\src\dawn\struct_in_array.hlsl 
Internal compiler error: LLVM Assert
```

When debugging with VS, the output we get is:

```
Error: assert(V[i]->getType() == Ty->getElementType() && "Wrong type in array element initializer")
File:
C:\src\external\DirectXShaderCompiler\lib\IR\Constants.cpp(885)
Func:	llvm::ConstantArray::getImpl
Exception thrown at 0x00007FF9356BCF19 (KernelBase.dll) in dxc.exe: 0xE0000001.
Unhandled exception at 0x00007FF9356BCF19 (KernelBase.dll) in dxc.exe: 0xE0000001.
```

It asserts here:

```
Constant *ConstantArray::getImpl(ArrayType *Ty, ArrayRef<Constant*> V) {
  // Empty arrays are canonicalized to ConstantAggregateZero.
  if (V.empty())
    return ConstantAggregateZero::get(Ty);

  for (unsigned i = 0, e = V.size(); i != e; ++i) {
    assert(V[i]->getType() == Ty->getElementType() &&
           "Wrong type in array element initializer");
  }
```

Note that without asserts enabled, DXC seems able to generate valid DXIL.

amaiorano:
I did some digging today, as this seems to be the only false positive assert we're hitting in our tests. 

So once again, here's the offending HLSL:
```cpp
struct str {
  int i;
};

str func(inout str pointer) {
  return pointer;
}

static str P[4] = (str[4])0;

[numthreads(1, 1, 1)]
void main() {
  const str r = func(P[2]);
  return;
}
```

When we hit the assertion, here's what the module IR looks like at this point:
```
; ModuleID = '/home/amaiorano/src/dawn/test/tint/ptr_ref/load/param/private/struct_in_array.wgsl.expected.dxc.hlsl'
[...]

PR #6812, opened 2024-07-22
Disable false-positive assert

amaiorano:
Until we resolve this bug, disable this assertion: https://github.com/microsoft/DirectXShaderCompiler/issues/5294


amaiorano:
Proper fix is up: https://github.com/microsoft/DirectXShaderCompiler/pull/6814

PR #6814, opened 2024-07-23
Fix assertion when compiling array of struct init

amaiorano:
In CGHLSLMSFinishCodeGen's BuildImmInit, when initializing an array, if the init value type doesn't match the array element type, we must bail and instead, have it inject a call to the global ctor. Without this, builds with asserts enabled would assert later with "Wrong type in array element initializer". In non-assert builds, this invalid IR would be removed, and valid code emitted.

See https://github.com/microsoft/DirectXShaderCompiler/issues/5294

### DIFF
--- a/tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
@@ -2966,7 +2966,6 @@ bool BuildImmInit(Function *Ctor) {
         allConst = false;
         break;
       }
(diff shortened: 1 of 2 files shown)


llvm-beanz (approved):

sudonatalie (approved):

dneto0 (approved):
LGTM.

I had to read the surrounding flow to understand what was going on.  The example in the linked bug was very helpful.
Bailing here means the fallback path will take effect, i.e. the constructor function will be called from the initialization code

PR #6815, opened 2024-07-24
Use llvm_assert/assert.h on all platforms, not just Windows

amaiorano:
This makes it possible to define how assert works on all platforms. The header was already being included by String.cpp, and was already designed to work for non-Windows platforms.

Also modify the non-Windows llvm_assert to emit the assertion message to stderr and trap. We cannot call standard assert as we are overriding it via include dirs, so there's no way to include the standard one and call it.

### DIFF
--- a/lib/Support/assert.cpp
+++ b/lib/Support/assert.cpp
@@ -14,10 +14,10 @@
 #include "dxc/Support/Global.h"
 #include "windows.h"
 
-void llvm_assert(const char *_Message, const char *_File, unsigned _Line,
-                 const char *_Function) {
-  OutputDebugFormatA("Error: assert(%s)\nFile:\n%s(%d)\nFunc:\t%s\n", _Message,
-                     _File, _Line, _Function);
+void llvm_assert(const char *Message, const char *File, unsigned Line,
+                 const char *Function) {
+  OutputDebugFormatA("Error: assert(%s)\nFile:\n%s(%d)\nFunc:\t%s\n", Message,
+                     File, Line, Function);
   RaiseException(STATUS_LLVM_ASSERT, 0, 0, 0);
 }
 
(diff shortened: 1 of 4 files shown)


lib/Support/assert.cpp
```diff
@@ -23,11 +23,15 @@ void llvm_assert(const char *_Message, const char *_File, unsigned _Line,
...
+#include "llvm/Support/Compiler.h"
+#include <cstdio>
 
-void llvm_assert(const char *message, const char *, unsigned,
+void llvm_assert(const char *_Message, const char *_File, unsigned _Line,
                  const char *_Function) {
-  assert(false && message);
+  fprintf(stderr, "Error: assert(%s)\nFile:\n%s(%d)\nFunc:\t%s\n", _Message,
+          _File, _Line, _Function);
+  fflush(stderr);
```
llvm-beanz, line 33: Is there a reason to use this instead of `llvm::errs()`?

lib/Support/assert.cpp
```diff
@@ -23,11 +23,14 @@ void llvm_assert(const char *_Message, const char *_File, unsigned _Line,
 
 #else
 
-#include <assert.h>
+#include "llvm/Support/Compiler.h"
+#include "llvm/Support/raw_ostream.h"
 
-void llvm_assert(const char *message, const char *, unsigned,
+void llvm_assert(const char *_Message, const char *_File, unsigned _Line,
                  const char *_Function) {
```
llvm-beanz, line 30: nit: Since we're touching this might as well match LLVM style.

```suggestion
void llvm_assert(const char *Message, const char *File, unsigned Line,
                 const char *Function) {
```

lib/Support/assert.cpp
```diff
@@ -23,11 +23,14 @@ void llvm_assert(const char *_Message, const char *_File, unsigned _Line,
...
-  assert(false && message);
+  llvm::errs() << "Error: assert(" << _Message << ")\nFile:\n"
+               << _File << "(" << _Line << ")\nFunc:\t" << _Function << "\n";
```
llvm-beanz, line 32: nit:
```suggestion
  llvm::errs() << "Error: assert(" << Message << ")\nFile:\n"
               << File << "(" << Line << ")\nFunc:\t" << Function << "\n";
```

llvm-beanz (approved):

lib/Support/assert.cpp
amaiorano: Done. Also updated the header file.

lib/Support/assert.cpp
amaiorano: Done.

sudonatalie (approved):

PR #6816, opened 2024-07-24
Add option to enable asserts in non-debug and to make asserts trap

amaiorano:
If LLVM_ENABLE_ASSERT_IN_NDEBUG is defined, asserts are enabled in NDEBUG builds. When asserts are enabled, macro ASSERTS_ENABLED is defined. This is used instead of NDEBUG to wrap assert-only code. Note that this change does not replace _all_ uses of NDEBUG with ASSERTS_ENABLED, only the ones required to get the build compiling. The rest will be handled in future changes.

If LLVM_ASSERT_TRAPS is defined, assert will always trap. Currently, assert calls RaiseException on Windows, and invokes the standard assert macro on non-Windows. This macro makes the behaviour well-defined and consistent across platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -325,6 +325,9 @@ else()
   option(LLVM_ENABLE_ASSERTIONS "Enable assertions" ON)
 endif()
 
+option(LLVM_ENABLE_ASSERT_IN_NDEBUG "Enable assertions in non-debug builds" OFF)
+option(LLVM_ASSERT_TRAPS "Force assertions to always trap, rather than the default unspecified behavior (e.g. RaiseException on Windows)" OFF)
+
 set(LLVM_ABI_BREAKING_CHECKS "WITH_ASSERTS" CACHE STRING
   "Enable abi-breaking checks.  Can be WITH_ASSERTS, FORCE_ON or FORCE_OFF.")
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -53,7 +53,7 @@ stages:
           configuration: Release
           CC: clang
           CXX: clang++
-          CMAKE_OPTS: -DLLVM_ENABLE_WERROR=On -DLLVM_USE_SANITIZER='Address;Undefined' -DLLVM_ENABLE_LIBCXX=On -DLLVM_USE_LINKER=lld
+          CMAKE_OPTS: -DLLVM_ENABLE_WERROR=On -DLLVM_USE_SANITIZER='Address;Undefined' -DLLVM_ENABLE_LIBCXX=On -DLLVM_USE_LINKER=lld -DLLVM_ENABLE_ASSERT_IN_NDEBUG=On -DLLVM_ASSERT_TRAPS=On
           CHECK_ALL_ENV: ASAN_OPTIONS=alloc_dealloc_mismatch=0
           OS: Linux
         Linux_Clang_Debug:
--- a/include/llvm/llvm_assert/assert.h
+++ b/include/llvm/llvm_assert/assert.h
@@ -22,11 +22,21 @@
 #define STATUS_LLVM_UNREACHABLE 0xE0000002
 #define STATUS_LLVM_FATAL 0xE0000003
 
-#ifdef NDEBUG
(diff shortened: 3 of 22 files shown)


amaiorano:
The idea behind this change is to allow assertions to be enabled in non-debug (`NDEBUG`) builds. This is something we want to for WebGPU as ~60% of security issues reported so far resulted in an assertion be triggered in debug builds. Leaving assertions in the release build downgrades these security issues to deterministic failures.

Note that without my changes, you can already force assertions to be enabled in `NDEBUG` builds by enabling the CMake flag `LLVM_ENABLE_ASSERTIONS`. When enabled, the CMake build forcibly removes `NDEBUG` from the list of compiler defines (see [here](https://github.com/microsoft/DirectXShaderCompiler/blob/e0fbce714da4746477639d82185ae76f6df4f472/cmake/modules/HandleLLVMOptions.cmake#L67-L80)). There are a couple of problems with using this strategy for us:

1. `NDEBUG` is used throughout the codebase for more than just assertion-related code. For instance, it's used to wrap all the `dump` code. We don't want to compile all this extra code in our release build of DXC.
2. We use GN to build Chrome, and because of its goal to make builds hermetic, there is no (easy) way to remove a flag like `NDEBUG` from a non-debug build.

So because of this, when `LLVM_ENABLE_ASSERT_IN_NDEBUG` is defined, we now define a new macro `ASSERTS_ENABLED`:

```cpp
// Use ASSERTS_ENABLED instead of NDEBUG to wrap assert-only code
#if !defined(NDEBUG) || defined(LLVM_ENABLE_ASSERT_IN_NDEBUG)
#define ASSERTS_ENABLED 1
#else
#define ASSERTS_ENABLED 0
#endif
```

`ASSERTS_ENABLED` is used instead of `NDEBUG` to wrap assert-related code (e.g. defining variables that are checked in assert). As part of this change, I did not replace _all_ uses of `NDEBUG` to use `ASSERTS_ENABLED` yet; I only replaced the ones required to make the project build successfully. The plan is to follow up this PR with future changes to do just that.

Finally, this change also introduces preprocessor flag `LLVM_ASSERT_TRAPS`. This is mainly to ensure that we get consistent behaviour across platforms. For Chrome, this means we'd consistently crash the process, and this would result in a crash report with a call stack. I considered making this the default behaviour for assertions, but I didn't want to change the Windows-specific `RaiseException` flow that was added at some point in the past, in case this is useful.

amaiorano:
Note that the build failures are because this change depends on https://github.com/microsoft/DirectXShaderCompiler/pull/6815, which hasn't landed yet. If we rebase this change on top of that one, the Linux builds should succeed.

llvm-beanz (changes requested):
I think this should be done using the `LLVM_ENABLE_ASSERTIONS` CMake option, not an extra option that sets a different preprocessor macro. This seems like a whole lot of code churn for no real reason.

llvm-beanz:
I had a separate conversation with @amaiorano about this, and I want to more completely summarize my thoughts here.

### Code churn

DXC's source repo has no contiguous history to LLVM. This means that we often resort to full directory diffing of DXC and LLVM 3.7 locally to identify differences in DXC vs LLVM. This is extremely useful because the bugs we encounter are often caused by DXC's modifications to LLVM and Clang code. Cascading lots of small changes across the inherited LLVM codebase will hurt that workflow, which is why we've generally avoided it in the past (see the existing assert hacks).

### Maintenance

As @amaiorano points out LLVM uses `NDEBUG` for a fair amount of code that isn't just asserts. It's also `dump` methods and other helpers used to drive assert verification. This is not likely to be an easy or clean pass to extricate, and I don't think we should be creating a new build configuration that is effectively "asserts without dump". If you want to enable asserts, you get dump too.

If that's a really big issue, we could consider a more targeted change that allows not marking dump methods with `__attribute__((__used__))`, which would allow them to be stripped during optimization & linking.

amaiorano:
Based on Chris's feedback, and some internal investigation, am going to drop this PR in favor of using `LLVM_ENABLE_ASSERTIONS`. I will follow up with a couple new PRs: one to add a macro to force asserts to trap (which was also part of this PR), and another to drop strings from the assert macro to reduce the compiled binary size.

PR #6818, opened 2024-07-24
Reserve some RDAT/DXIL enum values used by the mesh node preview

jenatali:
Let's ensure we don't accidentally stomp on these. If we do choose to break them, it should be intentional, not accidental.


include/dxc/DxilContainer/DxilRuntimeReflection.h
```diff
@@ -113,6 +115,8 @@ enum class RecordTableIndex : unsigned {
   IONodeTable,
   NodeShaderInfoTable,
 
+  Reserved_MeshNodesPreviewInfoTable,
```
damyanp, line 118: As this was added in the middle of the enum for the mesh nodes preview, I wonder if we really need to reserve this entry?

jenatali: I debated. The reason I added it is because the items below have never been consumable by any D3D runtime or driver that's ever been released publicly, where this entry does have runtimes in the wild that use it. Breaking that should be a higher bar.

damyanp (approved):

llvm-beanz (approved):

Issue #6824, opened 2024-07-25
DxilFixConstArrayInitializer can index OOB when const array init stores use negative index


dneto0:
**Description**
In a debug build, the compiler crashes on this compute shader:
```
groupshared float4 w;

[numthreads(1, 1, 1)]
void b() {
  int i = -1;
  w[i] = 0;
}
```

**Steps to Reproduce**
Compile via: `dxc -T cs_6_0 -E b a.hlsl`

**Actual Behavior**
The compiler should not crash.
The shader is doing an OOB access which is undefined behaviour, so it doesn't matter what the resulting shader does.

**Environment**
- DXC version: libdxcompiler.so: 1.8(dev;4686-ff5955a4) 
- Host Operating System: Linux x86-64

PR #6825, opened 2024-07-25
Avoid OOB index when fixing constant array-initialization

dneto0:
If the index on a constant store into an array is negative or out of bounds, that's an error, but shouldn't make the compiler index a vector out of bounds.

Fixed: #6824

### DIFF
--- a/tools/clang/test/DXC/Passes/DxilFixConstArrayInitializer/negative-index.ll
+++ b/tools/clang/test/DXC/Passes/DxilFixConstArrayInitializer/negative-index.ll
@@ -1,5 +1,14 @@
 ; RUN: %dxopt %s -hlsl-passes-resume -dxil-fix-array-init -S | FileCheck %s
 
+; The pass should not perform an out of bounds access when trying to determine
+; which elements of an array are coverd by stores.
+; The store instructions may have out of bounds accesses, including to negative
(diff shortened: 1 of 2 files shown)


amaiorano (approved):
LGTM

tools/clang/test/DXC/Passes/DxilFixConstArrayInitializer/negative-index.ll
```diff
@@ -0,0 +1,63 @@
...
+;
+; [numthreads(1, 1, 1)]
+; void b() {
+;   int i = -1;
+;   w[i] = 0;
+; }
+
+
+; CHECK: define void @b
+; CHECK: store float 0
```
llvm-beanz, line 16: Is this store still to a `-1` offset? If I'm reading the code right we just don't transform this so it doesn't cause the compiler to have an OOB access.

dneto0: > Is this store still to a `-1` offset? 

Yes.

> If I'm reading the code right we just don't transform this so it doesn't cause the compiler to have an OOB access.

Yes.

Will do.

llvm-beanz (approved):

PR #6826, opened 2024-07-25
Fix escape in Python string in lit.cfg

dneto0:
Change Python regexp ' clang\+\+ ' to ' clang\\+\\+' We're trying to match fixed strings like ` clang++ `, but `\+` is not a valid Python escape sequence.  Use `\\+` so the regexp machinery sees `\+`

### DIFF
--- a/tools/clang/test/lit.cfg
+++ b/tools/clang/test/lit.cfg
@@ -308,7 +308,7 @@ else:
 config.substitutions.append(
     (' clang ', """*** Do not use 'clang' in tests, use '%clang'. ***""") )
 config.substitutions.append(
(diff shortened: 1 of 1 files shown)


dneto0:
Here's an example of the warning I get:
```
/DirectXShaderCompiler/tools/clang/test/lit.cfg:311: SyntaxWarning: invalid escape sequence '\+'
  (' clang\+\+ ', """*** Do not use 'clang++' in tests, use '%clangxx'. ***"""))
```

amaiorano (approved):
Could also use raw strings, like `r' clang++ '`, but this is fine.

dneto0:
> Could also use raw strings, like `r' clang++ '`, but this is fine.

ha! I learned something new.

llvm-beanz (approved):

PR #6827, opened 2024-07-25
Added implementation for ID3D12FunctionReflection::GetDesc1

Nielsbishere:
Relates to https://github.com/microsoft/DirectX-Headers/pull/135.

I have tested this version in my own project and it seems to work. Only thing is that this requires the DirectX-Headers dep to be updated to ensure D3D12_FUNCTION_DESC1 and co can be found.

### DIFF
--- a/lib/HLSL/DxilContainerReflection.cpp
+++ b/lib/HLSL/DxilContainerReflection.cpp
@@ -724,6 +724,7 @@ CInvalidFunctionParameter g_InvalidFunctionParameter;
 
 class CInvalidFunction final : public ID3D12FunctionReflection {
   STDMETHOD(GetDesc)(D3D12_FUNCTION_DESC *pDesc) { return E_FAIL; }
+  STDMETHOD(GetDesc1)(D3D12_FUNCTION_DESC1 *pDesc) { return E_FAIL; }
 
   STDMETHOD_(ID3D12ShaderReflectionConstantBuffer *, GetConstantBufferByIndex)
   (UINT BufferIndex) { return &g_InvalidSRConstantBuffer; }
@@ -2814,6 +2815,7 @@ public:
 
   // ID3D12FunctionReflection
   STDMETHOD(GetDesc)(D3D12_FUNCTION_DESC *pDesc);
+  STDMETHOD(GetDesc1)(D3D12_FUNCTION_DESC1 *pDesc);
 
   // BufferIndex relative to used constant buffers here
   STDMETHOD_(ID3D12ShaderReflectionConstantBuffer *, GetConstantBufferByIndex)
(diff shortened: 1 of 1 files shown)


damyanp:
Hi - thank you for all your effort here!  

I just want to set appropriate expectations - this is a pretty large change adding a lot of new API surface area that we'll need to consider, taking into account future evolution of the language and compiler. As you may know we're hard at work adding [HLSL support to clang](https://clang.llvm.org/docs/HLSL/HLSLSupport.html), and so we need to consider the implications of a change that essentially exposes some amount of implementation details that may not be stable in the long term.  Also, as we're focusing on clang, this means we have limited bandwidth for reviews of this nature, especially as we're in the middle of summer vacation season.  So, be warned, it may take a while for us to find time to look at this properly.

In the meantime, I suggest you have a look at CONTRIBUTING.md to make sure that you have all your ducks in a row ready for a successful review.  For example, I think some tests may be required.

Thanks again!
- Damyan.

Nielsbishere:
Sounds reasonable, though I'm not sure if you're referring to unit tests or that the tests are failing on this PR.
Mainly the payloadSize/intersectionSize and some workgraph information is going to be useful for me. In my current parsers (also in production at work) I need to do custom work to parse struct sizes and other information that could hopefully be exposed even with a clang supported version of DXC. My goal there is to move as much as possible to a standardized DXC to avoid manual parser errors and missing functionality from my side. It is working in my PR, but it's tricky since this is dependent on the DirectX-Headers change. I updated the dependency, but still need to upstream some of the changes I had to make it build properly. If this PR is ever merged, it needs to change the external dependency's URL back from Oxsomi/DirectX-Headers to Microsoft/DirectX-Headers, since I have no other way to make the tests pass until that is merged.

damyanp:
> ...I'm not sure if you're referring to unit tests...

I'm referring to unit tests with the code.  From CONTRIBUTING.md:

> All changes that make functional or behavioral changes to the compiler whether by fixing bugs or adding features must include additional testing in the implementing pull request.

Nielsbishere:
Will take a look at Wednesday. Unexpectedly failing on remote. Also adding implementation for the reflection writer to know about these structs and output it as disassembly, along with a few unit tests.

Nielsbishere:
Since last time, added more reflection properties that were missing, got it to build across all platforms (except one more issue on windows) and added dxa support.
TODO: More testing, unit tests (specifically, adjusting current unit tests to include extra reflection data) and dxdisassembler support.

> There were some... issues with upgrading DXC to the latest DirectX-Headers, since it was behind by 2 years.

(See https://github.com/microsoft/DirectX-Headers/pull/135#issuecomment-2274300142).

This commit unifies wsl/WinAdapter.h and dxc/WinAdapter.h a bit. These two have very hard conflicts and I decided to resolve them by taking wsl's, since it seems more up to date and is far more logical. There are however cases when dxc/WinAdapter.h has more thorough definitions, so duplicate definitions in DirectX-Headers are now #ifdeffed. For example dxc/WinAdapter.h defines ULONG as unsigned long, but on linux (64-bit) that is not the same as a Windows ULONG (32-bit). I don't know enough about DXC in-depth to know what effect this would have, so definitely needs more testing / reviews.

damyanp:
I notice that this PR seems to be growing in scope.  You may have more success in getting it reviewed and accepted if you find a way to split out units of change that can be more trivially reviewed.  For example, I think you've updated some of the external headers and refactored how some of the adapter headers work?  These could be made into separate changes.

Nielsbishere:
Will do. It seems that DirectX-Headers wasn't being updated for 2 years, so quite some things broke in the meantime.

Nielsbishere:
This should now be ready for review. Unless support in the disassembler is required, though I'm not sure how useful it would be since dxa and the unit testing tool both provide this functionality through D3DReflectionDumper.

Nielsbishere:
Added an important function to ID3D12FunctionReflection1; which is GetRequiresFlags such as in ID3D12ShaderReflection. Without this, it is impossible to find out what features are being used by a lib file (I thought the desc Flags was used for this, but nope).

This is especially important now that some GPUs don't always support the base featureset, so I use the following to check feature support of the underlying GPUs:

```
		D3D_SHADER_REQUIRES_RAYTRACING_TIER_1_1,
		D3D_SHADER_REQUIRES_NATIVE_16BIT_OPS,
		D3D_SHADER_REQUIRES_INT64_OPS,
		D3D_SHADER_REQUIRES_VIEW_ID,
		D3D_SHADER_REQUIRES_DOUBLES,
		D3D_SHADER_REQUIRES_11_1_DOUBLE_EXTENSIONS,
		D3D_SHADER_REQUIRES_ATOMIC_INT64_ON_TYPED_RESOURCE,
		D3D_SHADER_REQUIRES_ATOMIC_INT64_ON_GROUP_SHARED,
		D3D_SHADER_REQUIRES_DERIVATIVES_IN_MESH_AND_AMPLIFICATION_SHADERS,
		D3D_SHADER_REQUIRES_WRITEABLE_MSAA_TEXTURES,
		D3D_SHADER_REQUIRES_WAVE_OPS
```

I use this reflection to know whether a shader is really using a feature. It's possible someone enables I64 to include something that requires it but never actually uses I64 in the final shader. This is then used to retroactively disable that feature to ensure we don't go checking features that aren't really relevant.

Issue #6829, opened 2024-07-26
assert in Reassociate::OptimizeAdd fires when it shouldn't due to miscounted multiplicative factors


dneto0:
**Description**
Assert fires on valid code, when asserts are enabled.

**Steps to Reproduce**
The compiler asserts, when asserts are on, in the following case.
Compile with `dxc -HV 2018 a.hlsl -T cs_6_0 -E m`

```
int2 tint_mod(int2 lhs, int rhs) {
  int2 r = int2((rhs).xx);
  int2 rhs_or_one = (((r == (0).xx) | ((lhs == (-2147483648).xx) & (r == (-1).xx))) ? (1).xx : r);
  if (any(((uint2((lhs | rhs_or_one)) & (2147483648u).xx) != (0u).xx))) {
    return (lhs - ((lhs / rhs_or_one) * rhs_or_one));
  } else {
    return (lhs % rhs_or_one);
  }
}

int2 tint_mod_1(int2 lhs, int2 rhs) {
  int2 rhs_or_one = (((rhs == (0).xx) | ((lhs == (-2147483648).xx) & (rhs == (-1).xx))) ? (1).xx : rhs);
  if (any(((uint2((lhs | rhs_or_one)) & (2147483648u).xx) != (0u).xx))) {
    return (lhs - ((lhs / rhs_or_one) * rhs_or_one));
  } else {
    return (lhs % rhs_or_one);
  }
}

[numthreads(1,1,1)]
void m() {
  float x = ddy(0.0f);
  bool2 y = bool2((x>0).xx); // simplified from origial case
  int2 w = ~(~(~(~(~(~(~(-(~(~(~(~(~(int2(y))))))))))))));
  int2 w6 = (w + (((((((~(-(~(~(~(~(~(~(w)))))))) - -(w)) - (-(w) * 2)) - w) - w) - ~(-(w))) - tint_mod(w, 4)) - 1));
  int2 w5 = tint_mod(-(w6), 2); // power of 2 is required here
  int2 w4 = tint_mod(w5, w6.x);
  int2 w3 = tint_mod(w4, 2); // power of 2 is reuquired here
  int2 w2 = tint_mod(w3, 5); // anything more than 1
  int2 w1 = tint_mod(w2, w6.x);
  w = tint_mod(w, w1.x);
  return;
}
```

Sorry, this is as reduced as I can make it.

[...]

PR #6830, opened 2024-07-26
Reassociation: fix counting of constant multiplicative factors

dneto0:
In the middle of rewriting expressions like (A*B + A*C + D) to pull common factor A out, the algorithm finds that there's actually only one A. This is unexpected, and it fires an assertion.

This can occur when A is a constant, and constant -A also appears in the terms somewhere else.

There is no harm in this situation, however, because the algorithm then creates an addition-tree, but with a single element, and that's still correct.

This bookkeeping issue was fixed later in LLVM, at https://github.com/llvm/llvm-project/commit/95abfa35d6c0b8a769399262496a3f692177ef12

Unfortunately the associated test doesn't translate cleanly to DXC-era LLVM. I've added test case reduced from our original case.

Fixed: #6829

### DIFF
--- a/test/HLSL/passes/reassociate/constant-counting.ll
+++ b/test/HLSL/passes/reassociate/constant-counting.ll
@@ -1,0 +1,189 @@
+; RUN: opt %s -reassociate -S | FileCheck %s
+
+; Issue: #6829
+
+; There are cases where Reassociation used to double-count a constant value
+; used as a multiplicative factor in an addition tree.
+; When rewriting an add-over-muls it asserts out when it has double-counted
+; those factors. The code generated is still correct because its add tree is
+; still has the same leaves.
+
+; In the input module, the troublesome tree of adds ends in %add.i0
+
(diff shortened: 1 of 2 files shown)


test/HLSL/passes/reassociate/constant-counting.ll
```diff
@@ -0,0 +1,189 @@
+; RUN: opt %s -reassociate -S | FileCheck %s
+
+; Issue: #6829
+
+; There are cases where Reassociation used to double-count a constant value
```
dneto0, line 5: I tried quite hard to reduce this test case, or to adapt the upstream test case.

Reassociation works iteratively, and it needs to generate certain intermediate results that then force the final issue.
It's very hard to follow. This is actually a cut down version of the original HLSL we got (and I put in the issue)

amaiorano (approved):

llvm-beanz (approved):

Issue #5739, opened 2023-09-18
DXC linker debug output isn't a valid PDB (and doesn't work with PIX)


EricLasotaRSE:
**Description**
Linking a compute shader with debug information will dump the raw ILDB part from the DXIL container instead of a valid PDB file.

**Steps to Reproduce**
Create a file named test.hlsl:
```Texture2D<uint> texResource : register(t900);
RWTexture2D<uint> rwTexResource[] : register(u0, space2400);
[numthreads(8, 8, 1)]
void main(uint3 dtid : SV_DispatchThreadID, uint3 gtid : SV_GroupThreadID, uint3 gid : SV_GroupID, uint gindex : SV_GroupIndex)
{
    rwTexResource[0][dtid.xy] = texResource.Load(dtid.xyz);
}
```

Run these commands:
```dxc.exe -T lib_6_3 -Zi -Qstrip_reflect -Qembed_debug -Fd testc.pdb -Fo test.lib test.hlsl
dxc.exe -link -T lib_6_3 -Zi -Qstrip_reflect -Qstrip_debug -Fd test.pdb -Fo test.bin test.lib
```

**Actual Behavior**
testc.pdb (the compiler output) has a valid PDB header.
test.pdb does not.

"PDBs" output this way will also not load in PIX.

**Environment**
- DXC version dxcompiler.dll: 1.7 - 1.7.2207.3 (e9137cd1d); dxil.dll: 1.7(101.7.2207.10)
- Host Operating System Windows 10

PR #6833, opened 2024-07-30
Fix -link -Qstrip_debug failing

### LINKED ISSUE #5737: Link fails when using -Fd with -Qstrip_debug
**Description**
Linking a compute shader with the command-line DXC will error out if using -Fd and -Qstrip_debug together.  This is not consistent with the compiler, which will output a PDB even if debug info is stripped from the object file.

**Steps to Reproduce**
Create a file named test.hlsl:
```Texture2D<uint> texResource : register(t900);
RWTexture2D<uint> rwTexResource[] : register(u0, space2400);
[numthreads(8, 8, 1)]
void main(uint3 dtid : SV_DispatchThreadID, uint3 gtid : SV_GroupThreadID, uint3 gid : SV_GroupID, uint gindex : SV_GroupIndex)
{
    rwTexResource[0][dtid.xy] = texResource.Load(dtid.xyz);
}```

Run these commands:
```dxc.exe -T lib_6_3 -Zi -Qstrip_reflect -Qembed_debug -Fd testc.pdb -Fo test.lib test.hlsl
dxc.exe -link -T lib_6_3 -Zi -Qstrip_reflect -Qstrip_debug -Fd test.pdb -Fo test.bin test.lib```

**Actual Behavior**
DXC fails with error:
`dxc failed : DXIL container does not contain the given part.`

**Environment**
- DXC version dxcompiler.dll: 1.7 - 1.7.2207.3 (e9137cd1d); dxil.dll: 1.7(101.7.2207.10)
- Host Operating System Windows 10

elasota:
Fixes #5737

Link doesn't include the ILDB chunk in the first place when -Qstrip_debug is used, so this fixes it failing in UpdatePart.

The compiler does the same thing, sort of, except the compiler's default behavior is different (it doesn't embed debug by default), so this omits the condition to reflect that.

This will cause -link -Fd to fail, but -link -Fd doesn't work properly anyway due to #5739 (which will be fixed in a separate forthcoming PR that doesn't conflict with this one)

### DIFF
--- a/tools/clang/tools/dxclib/dxc.cpp
+++ b/tools/clang/tools/dxclib/dxc.cpp
@@ -969,6 +969,10 @@ int DxcContext::Link() {
                     wpInputFiles.size(), args.data(), args.size(),
                     &pLinkResult));
 
+  // Link will not include any debug info if StripDebug is set, so clear it
+  // here so we don't attempt to remove it again in UpdatePart
+  m_Opts.StripDebug = false;
+
   HRESULT status;
   IFT(pLinkResult->GetStatus(&status));
   if (SUCCEEDED(status)) {


damyanp:
Hi - thank you for your contribution and apologies for the delay in responding to this.

We do ask that all changes and fixes like this are accompanied with tests that validate the fix.

Thanks!

PR #6837, opened 2024-07-30
Fix warning when use implicit launch type.

### LINKED ISSUE #6836: Got warning when use implicit launch type.
**Description**

By default, the launch type should be set to ‘Broadcast’ when diagnosing barriers. 
However, the current behavior sets the default launch type to ‘Invalid,’ resulting in warnings when the launch type is not explicitly specified as an attribute.
**Steps to Reproduce**

https://godbolt.org/z/MescEbMcq

**Actual Behavior**

Warning is reported which should not.

**Environment**
- DXC version 1.8.2407
- Host Operating System : Windows.

python3kgae:
By default, the launch type should be set to ‘Broadcast’ when diagnosing barriers. However, the current behavior sets the default launch type to ‘Invalid,’ resulting in warnings when the launch type is not explicitly specified as an attribute.

To address this issue, we’ll adjust the default setting to ‘Broadcast’ and thereby resolve the problem.

Fixes #6836

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
+++ b/tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
@@ -523,7 +523,7 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
     }
 
     DXIL::ShaderKind EntrySK = shaderModel->GetKind();
-    DXIL::NodeLaunchType NodeLaunchTy = DXIL::NodeLaunchType::Broadcasting;
+    DXIL::NodeLaunchType NodeLaunchTy = DXIL::NodeLaunchType::Invalid;
     if (EntrySK == DXIL::ShaderKind::Library) {
       // For library, check if the exported function is entry with shader
       // attribute.
@@ -533,6 +533,8 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
         if (const auto *pAttr = FDecl->getAttr<HLSLNodeLaunchAttr>())
           NodeLaunchTy =
               ShaderModel::NodeLaunchTypeFromName(pAttr->getLaunchType());
+        else
+          NodeLaunchTy = DXIL::NodeLaunchType::Broadcasting;
     }
     // Visit all visited functions in call graph to collect illegal intrinsic
     // calls.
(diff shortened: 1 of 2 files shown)


tools/clang/test/SemaHLSL/hlsl/intrinsics/barrier/barrier-node-errors.hlsl
```diff
@@ -192,3 +192,12 @@ void node02(RWThreadNodeInputRecord<RECORD> input,
   Barrier(threadRec, DEVICE_SCOPE);
   threadRec.OutputComplete();
 }
+
+// Make sure default broadingcast launch is supported when dignose barrier.
```
damyanp, line 196: As there's no expected errors here, maybe some more explanation would help about what this is testing and why? I think there's maybe something about the `Barrier` that means it'll only work if the NodeLaunch type is broadcasting?

python3kgae: Updated the comment.

tools/clang/test/SemaHLSL/hlsl/intrinsics/barrier/barrier-node-errors.hlsl
```diff
@@ -192,3 +192,15 @@ void node02(RWThreadNodeInputRecord<RECORD> input,
   Barrier(threadRec, DEVICE_SCOPE);
   threadRec.OutputComplete();
 }
+
+// Default launch type is broadingcast which has a visible group.
```
damyanp, line 196:
```suggestion
// Default launch type is broadcasting which has a visible group.
```

damyanp (approved):

bogner (approved):

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -523,7 +523,7 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
     }
 
     DXIL::ShaderKind EntrySK = shaderModel->GetKind();
-    DXIL::NodeLaunchType NodeLaunchTy = DXIL::NodeLaunchType::Invalid;
+    DXIL::NodeLaunchType NodeLaunchTy = DXIL::NodeLaunchType::Broadcasting;
```
tex3d, line 526: I think a better fix is to set this to Broadcasting below if `EntrySK == DXIL::ShaderKind::Node` and no `HLSLNodeLaunchAttr` is found.  Only if this is actually a node shader should this be set to something other than `Invalid`.

python3kgae: Updated.

PR #6838, opened 2024-07-30
Add llvm assertions trap and no strings

amaiorano:
Add option LLVM_ASSERTIONS_NO_STRINGS
    
    When defined, drop the stringized expression, __FILE__, and __FUNCTION__
    strings passed to llvm_assert. This dramatically reduces the binary
    size, which is useful when enabling assertions in a non-debug build.

   Add option LLVM_ASSERTIONS_TRAP
    
    When enabled, this forces asserts to always trap. Currently, on Windows
    asserts calls RaiseException, while they trap on non-Windows. This
    option makes the assertion behaviour consistent across platforms.


amaiorano:
@llvm-beanz this PR is made up of two commits. If you prefer I split them into separate PRs, let me know.

amaiorano:
Note that when `LLVM_ASSERTIONS_NO_STRINGS` is enabled, my release build of `dxcompiler.dll` goes from 27,875,840 bytes to 24,627,712, a savings of about 3.1 MiB.

llvm-beanz (approved):

sudonatalie (approved):

PR #6840, opened 2024-07-30
[SPIRV] Implement Gis option for spir-v

### LINKED ISSUE #3331: [SPIR-V] does not take -Gis (IEEE strict) argument into account
The `-Gis` command line argument should effectively mark all variables behave as if they had `precise` HLSL modifier in them (that's what #2861 seems to be doing in DXIL output).

But SPIR-V output seems to generate completely identical result no matter if `-Gis` is present or not. Only actually spelling out `precise` on variables makes it generate the `NoContraction` SPIR-V decorations.

s-perron:
Fixes #3331

### DIFF
--- a/tools/clang/lib/SPIRV/EmitVisitor.cpp
+++ b/tools/clang/lib/SPIRV/EmitVisitor.cpp
@@ -431,7 +431,9 @@ void EmitVisitor::initInstruction(SpirvInstruction *inst) {
                                spv::Decoration::RelaxedPrecision, {});
   }
   // Emit NoContraction decoration (if any).
-  if (inst->isPrecise() && inst->isArithmeticInstruction()) {
+  // STEVEN: Check if Gis is set here.
+  if ((spvOptions.IEEEStrict || inst->isPrecise()) &&
+      inst->isArithmeticInstruction()) {
     typeHandler.emitDecoration(getOrAssignResultId<SpirvInstruction>(inst),
                                spv::Decoration::NoContraction, {});
   }
--- a/tools/clang/lib/SPIRV/EmitVisitor.cpp
+++ b/tools/clang/lib/SPIRV/EmitVisitor.cpp
@@ -431,7 +431,6 @@ void EmitVisitor::initInstruction(SpirvInstruction *inst) {
                                spv::Decoration::RelaxedPrecision, {});
   }
   // Emit NoContraction decoration (if any).
-  // STEVEN: Check if Gis is set here.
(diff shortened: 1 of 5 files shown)


tools/clang/lib/SPIRV/EmitVisitor.cpp
```diff
@@ -431,7 +431,9 @@ void EmitVisitor::initInstruction(SpirvInstruction *inst) {
                                spv::Decoration::RelaxedPrecision, {});
   }
   // Emit NoContraction decoration (if any).
-  if (inst->isPrecise() && inst->isArithmeticInstruction()) {
+  // STEVEN: Check if Gis is set here.
```
Keenuts, line 434: Is this TODO a leftover?

tools/clang/test/CodeGenSPIRV/spirv.opt.gis.hlsl
```diff
@@ -1,5 +1,12 @@
...
 
-void main() {}
+// Make sure the no-contraction is added for the arithmetic operation.
+// CHECK: OpDecorate [[op:%[0-9]+]] NoContraction
+
+float4 v;
+float4 main(uint col : COLOR) : SV_Target0
+{
+  // [[op]] = OpVectorTimesScalar
+  return 3*v;
```
Keenuts, line 10: Shouldn't several operation types/basetypes be tested?

tools/clang/lib/SPIRV/EmitVisitor.cpp
s-perron: Yes tha was a marker to remind where to add the code after I added the new option. No longer needed.

tools/clang/test/CodeGenSPIRV/spirv.opt.gis.hlsl
s-perron: I thought about that. I'm not sure how many too do. This was meant to test whether the option is read and used correctly. Adding more types and operations is testing if `isArithmeticInstruction` works correctly, which gets covered by the decoration.no-contraction*.hlsl tests.

Let me know what you think I need to add if you do not the existing tests are good enough.

Keenuts: I agree. I was wondering if those test wouldn't catch changes in our spaghetti codegen shortcuts, but since this change only impacts the InitInstr function, seems fine.

cassiebeckley (approved):

PR #6848, opened 2024-08-06
[SPIRV] Add vulkan memory option to DXC

### LINKED ISSUE #5784: [Feature Request] Make DXC emit VulkanMemoryModel instead of Logical GLSL450
We would like to take advantage of the Vulkan Memory Model, for consistency and sanity reasons.

A compilation flag that would enable the [Vulkan memory model](https://www.khronos.org/blog/vulkan-has-just-become-the-worlds-first-graphics-api-with-a-formal-memory-model.-so-what-is-a-memory-model-and-why-should-i-care)

Use GLSL450 but we'd like to be able to reason about the memory ordering of our operations.
There is also the flag `-0config=--upgrade-memory-model` but this would require writing operations in glsl450 and rely on spirv-opt to upgrade them, whereas we'd like to use the proper VMM opcodes which are not necessarily expressible in glsl450.

**Additional context**
N/A

s-perron:
Some recent features require the vulkan memory model. We need a way to
support it in DXC. This commit adds an option to enable the Vulkan
memory model. The generated code will not change, but the compiler will
call the upgrade-memory-model pass in spirv-opt to generate the correct
code.

The expectation is that new code that requieres the vulkan memory model
can be written using the vulkan memory model, and then have the pass
update other references.

This change will undo the fix for #6066. The device scope in GLSL450 was
not a real device scope. It corresponds to QueueFamily scope in the
Vulkan memory model. We will start to use that scope for the atomic
operations in order to keep the behavoiur of the atomics the same
between the two memory models.

Haveing a true device scope atomic will not be possible with DXC. We
should be able to do something better when atomic operations are
implemented in clang.

Fixes #5784

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -770,9 +770,8 @@ public:
   /// the feature used at the given source location.
   inline void requireCapability(spv::Capability, SourceLocation loc = {});
 
-  /// \brief Adds the given capability to the module under construction due to
-  /// the feature used at the given source location.
-  inline bool requiresCapability(spv::Capability);
+  /// \brief Returns true if the module requires the given capability.
+  inline bool hasCapability(spv::Capability cap);
 
   /// \brief Adds an extension to the module under construction for translating
   /// the given target at the given source location.
@@ -900,7 +899,7 @@ void SpirvBuilder::requireCapability(spv::Capability cap, SourceLocation loc) {
   }
 }
 
-bool SpirvBuilder::requiresCapability(spv::Capability cap) {
+bool SpirvBuilder::hasCapability(spv::Capability cap) {
   SpirvCapability capability({}, cap);
   return mod->hasCapability(capability);
 }
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -770,6 +770,10 @@ public:
   /// the feature used at the given source location.
   inline void requireCapability(spv::Capability, SourceLocation loc = {});
 
+  /// \brief Adds the given capability to the module under construction due to
+  /// the feature used at the given source location.
+  inline bool requiresCapability(spv::Capability);
+
   /// \brief Adds an extension to the module under construction for translating
   /// the given target at the given source location.
   inline void requireExtension(llvm::StringRef extension, SourceLocation);
@@ -894,6 +898,11 @@ void SpirvBuilder::requireCapability(spv::Capability cap, SourceLocation loc) {
   } else {
     capability->releaseMemory();
   }
+}
+
+bool SpirvBuilder::requiresCapability(spv::Capability cap) {
+  SpirvCapability capability({}, cap);
+  return mod->hasCapability(capability);
 }
 
 void SpirvBuilder::requireExtension(llvm::StringRef ext, SourceLocation loc) {
--- a/tools/clang/include/clang/SPIRV/SpirvModule.h
+++ b/tools/clang/include/clang/SPIRV/SpirvModule.h
@@ -102,7 +102,7 @@ public:
   // Returns false otherwise (e.g. if the capability already existed).
   bool addCapability(SpirvCapability *cap);
 
-  // Returns true if the capability is in the modudle.
+  // Returns true if the capability is in the module.
   bool hasCapability(SpirvCapability &cap);
 
   // Set the memory model of the module.
--- a/tools/clang/include/clang/SPIRV/SpirvModule.h
+++ b/tools/clang/include/clang/SPIRV/SpirvModule.h
@@ -102,6 +102,9 @@ public:
   // Returns false otherwise (e.g. if the capability already existed).
   bool addCapability(SpirvCapability *cap);
 
+  // Returns true if the capability is in the modudle.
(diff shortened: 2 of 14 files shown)


tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -770,6 +770,10 @@ class SpirvBuilder {
   /// the feature used at the given source location.
   inline void requireCapability(spv::Capability, SourceLocation loc = {});
 
+  /// \brief Adds the given capability to the module under construction due to
+  /// the feature used at the given source location.
```
Keenuts, line 774:
```suggestion
  /// \brief Adds the given capability to the module under construction.
```

tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -896,6 +900,11 @@ void SpirvBuilder::requireCapability(spv::Capability cap, SourceLocation loc) {
   }
 }
 
+bool SpirvBuilder::requiresCapability(spv::Capability cap) {
+  SpirvCapability capability({}, cap);
+  return mod->hasCapability(capability);
+}
```
Keenuts, line 906: The function does not adds the capability, but checks if it has it already no?
Is the name/comment OK?

tools/clang/include/clang/SPIRV/SpirvModule.h
```diff
@@ -102,6 +102,9 @@ class SpirvModule {
   // Returns false otherwise (e.g. if the capability already existed).
   bool addCapability(SpirvCapability *cap);
 
+  // Returns true if the capability is in the modudle.
```
Keenuts, line 105:
```suggestion
  // Returns true if the capability is in the module.
```

tools/clang/lib/SPIRV/SpirvEmitter.h
```diff
@@ -1216,6 +1224,15 @@ class SpirvEmitter : public ASTConsumer {
   bool spirvToolsTrimCapabilities(std::vector<uint32_t> *mod,
                                   std::string *messages);
 
+  // \brief Runs the upgrade memory model pass using SPIRV-Tools's optimizer.
+  // This pass will modify the module so that it conforms to the Vulkan memory
+  // model instead of the pass. Removes unused capabilities from the given
+  // SPIR-V module |mod|, and returns info/warning/error messages via
+  // |messages|. This pass doesn't trim all capabilities. To see the list of
+  // supported capabilities, check the pass headers.
```
Keenuts, line 1232:
```suggestion
```

This is from the pass above, but doesn't apply to this one no?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -14709,33 +14714,28 @@ bool SpirvEmitter::spirvToolsFixupOpExtInst(std::vector<uint32_t> *mod,
...
-  options.set_preserve_bindings(spirvOptions.preserveBindings);
-  options.set_max_id_bound(spirvOptions.maxId);
-
-  optimizer.RegisterPass(spvtools::CreateTrimCapabilitiesPass());
+  spvtools::Optimizer::PassToken token = spvtools::CreateTrimCapabilitiesPass();
+  return spirvToolsRunPass(mod, std::move(token), messages);
+}
 
-  return optimizer.Run(mod->data(), mod->size(), mod, options);
+bool SpirvEmitter::spirvToolsUpgradeMemoryModel(std::vector<uint32_t> *mod,
```
Keenuts, line 14734: UpgradeToVulkanMemoryModel?
(I know this is the pass name, but I do find it a shame that the pass name doesn't say to what it upgrades)

tools/clang/include/clang/SPIRV/SpirvModule.h
```diff
@@ -102,6 +102,9 @@ class SpirvModule {
...
+  bool hasCapability(SpirvCapability &cap);
```
Keenuts, line 106: Where is this implemented?

Keenuts (changes requested):

tools/clang/include/clang/SPIRV/SpirvModule.h
s-perron: In SpirvModule.cpp

https://github.com/microsoft/DirectXShaderCompiler/pull/6848/files#diff-afc3a7f27324e1e5fc4f39ee02c1aca0f9a40ce18946d2bb51998402c825c971R252-R255

tools/clang/include/clang/SPIRV/SpirvBuilder.h
s-perron: I fixed it too many cut and paste error I did not fix up.

tools/clang/include/clang/SPIRV/SpirvBuilder.h
s-perron: I renamed the function to `hasCapability`.

tools/clang/lib/SPIRV/SpirvEmitter.h
s-perron: Fixed.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
s-perron: Renamed. That is a better name.

Keenuts (approved):

cassiebeckley (approved):

Issue #6817, opened 2024-07-24
Allow string table in the PSV0 part to have a different order than DXC during validation


python3kgae:
**Is your feature request related to a problem? Please describe.**
Currently, dxil validator will make sure PSV0 part matches the PSV0 part generated from dxc.
That requires all the names in PSV0 string table have same order and cause duplicate names.

**Describe the solution you'd like**
Sort the string table in validation.
This will only affect the validator. All other part of DXC will be the same.

**Describe alternatives you've considered**
Require string table to be sorted in PSV0 part.
    This will require update the Dxil Container builder.

**Additional context**
[Doc about string table in PSV0 part](https://github.com/llvm/llvm-project/blob/main/llvm/docs/DirectX/DXContainer.rst#id13)

llvm-beanz:
This impacts more than just PSV0. All of the container parts except the DXIL bitcode are validated this way. Correcting this will require design work and significant investment.

PR #6849, opened 2024-08-07
[NFC] Move DxilValidation to a dedicated folder.

python3kgae:
Move DxilValidation out of HLSL.
Also move code to validate dxil container into DxilContainerValidation.cpp from DxilValidation.cpp.

This is a preparatory step for #6817.


lib/DxilValidation/DxilValidationUtils.h
```diff
@@ -0,0 +1,149 @@
...
+  const Type *HandleTy;
+  const DataLayout &DL;
+  DebugLoc LastDebugLocEmit;
+  ValidationRule LastRuleEmit;
+  std::unordered_set<Function *> entryFuncCallSet;
+  std::unordered_set<Function *> patchConstFuncCallSet;
+  std::unordered_map<unsigned, bool> UavCounterIncMap;
+  std::unordered_map<Value *, unsigned> HandleResIndexMap;
+  // TODO: save resource map for each createHandle/createHandleForLib.
+  std::unordered_map<Value *, DxilResourceProperties> ResPropMap;
```
bob80905, line 80: Looks like `DxilResourceProperties` is an incomplete type, never gets defined after its declaration on line 51.

python3kgae: Fixed.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -0,0 +1,705 @@
...
+  const char *pName = nullptr;
+  switch (SigKind) {
+  case hlsl::DXIL::SignatureKind::Input:
+    pName = "Program Input Signature";
+    break;
+  case hlsl::DXIL::SignatureKind::Output:
+    pName = "Program Output Signature";
+    break;
+  case hlsl::DXIL::SignatureKind::PatchConstOrPrim:
+    if (ValCtx.DxilMod.GetShaderModel()->GetKind() == DXIL::ShaderKind::Mesh)
```
bob80905, line 131: Should we check for node shaders here? I think mesh nodes can have patch constant functions right?

python3kgae: We don't have a Signature part for node shaders.

bob80905 (approved):

damyanp:
Is this code really moved or is it copied from somewhere?  I see a lot of code being added, but not removed from anywhere.

python3kgae:
> Is this code really moved or is it copied from somewhere? I see a lot of code being added, but not removed from anywhere.

The code is moved around and spited into different files.
Many removing happened in DxilValidation.cpp and placed into DxilContainerValidation.cpp.

utils/hct/hctdb_instrhelp.py
```diff
@@ -1941,7 +1941,6 @@ def RunCodeTagUpdate(file_path):
             "lib/DXIL/DxilShaderModel.cpp",
             "include/dxc/DXIL/DxilConstants.h",
             "include/dxc/DXIL/DxilShaderModel.h",
-            "include/dxc/HLSL/DxilValidation.h",
```
damyanp, line 1944: This doesn't need to be updated to look at the new location?

damyanp (approved):

utils/hct/hctdb_instrhelp.py
python3kgae: There's no <py::lines left in DxilValidation.h after https://github.com/microsoft/DirectXShaderCompiler/commit/6cf0a6f7df67c72835d88659d410a43990307fc3

Issue #6808, opened 2024-07-19
[Feature Request] Open source the implementation of dxil.dll.


python3kgae:
**Is your feature request related to a problem? Please describe.**
Open source the implementation of dxil.dll for [open-sourcing-dxil-validator-hash](https://devblogs.microsoft.com/directx/open-sourcing-dxil-validator-hash/)

**Describe the solution you'd like**
Implement the solution in two steps:

Refactor the validator to a static library.
Open source dxil.dll with the static library validator.

**Describe alternatives you've considered**
An alternative approach is to open source dxil.dll first and then refactor. However, this would require refactoring the validator code twice.

**Context**
Open sourcing dxil.dll will introduce another validator to the repository, causing duplication with the existing validator. Refactoring the validator to a static library and sharing the code will eliminate this duplication. However, we cannot introduce new code with style issues into the repository. Therefore, refactoring should be done first, followed by open sourcing dxil.dll.

PR #6853, opened 2024-08-08
[Validator] enable validator hash by default.

### LINKED ISSUE #6857: [Feature Request] Integrate container hash functionality into internal validator
With validation hash open sourced at https://github.com/microsoft/DirectXShaderCompiler/commit/84f68ed1a40db3f3444a306162cab6ac0adfe842.
The internal validator could have the ability to do the validation hash work.
This will help cases where the compiling and hashing need to be in separate step when dxil.dll cannot find.

Internal validator and external validator will share the same code which do the validation hash.

Remove dxil.dll and perform only the validation hash in dxc.
This change will cause use cases that require only dxil.dll to stop working.

**Additional context**
dxil.dll will be part of dxc repo (#6808)

python3kgae:
The changes affect both the internal validator (used within the DXIL compiler) and external validation tools. Now, by default, validator hash is enabled for all validation processes.

#6863 was created for tracking the skip hash discussion.

This is second step for #6808.
Fixes #6857


lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -9,12 +9,13 @@
 //                                                                           //
 ///////////////////////////////////////////////////////////////////////////////
 
+#include "dxc/Support/Global.h"
+#include "dxc/Support/WinIncludes.h"
```
pow2clk, line 13: This move looks appropriate, but I don't see anything that would have made it necessary. Is it just an incidental cleanup?

lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -164,11 +169,62 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
...
+            HashAndUpdate((DxilContainerHeader *)PTR);
+        }
+      }
+    }
     return S_OK;
   }
   CATCH_CPP_RETURN_HRESULT();
 }
 
+void DxcContainerBuilder::FindHashFunctionFromSource(
```
pow2clk, line 192: I'm a fan of descriptive comments above any function, but this one in particular seems to need it. 

```suggestion
// Try hashing the source contained in ContainerHeader using retail and debug
// hashing functions. If either of them match the stored result, set the
// HashFunction to the matching variant. If neither match, set it to null.
void DxcContainerBuilder::FindHashFunctionFromSource(
```

lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -164,11 +169,62 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
...
+    const DxilContainerHeader *ContainerHeader) {
+  DXASSERT(ContainerHeader != nullptr &&
+               IsDxilContainerLike(ContainerHeader,
+                                   ContainerHeader->ContainerSizeInBytes),
+           "otherwise load function should have returned an error.");
+  static const uint32_t HashStartOffset =
+      offsetof(struct DxilContainerHeader, Version);
+  const BYTE *DataToHash = (const BYTE *)ContainerHeader + HashStartOffset;
+  UINT AmountToHash = ContainerHeader->ContainerSizeInBytes - HashStartOffset;
+  BYTE Result[DxilContainerHashSize];
```
pow2clk, line 202: It's pre-exiting, but there is no indication that `DxilContainerHashSize` is a constant. I know that some people cringe at Hungarian notation, but the practice in DXC has been to use the k* prefix.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1039,7 +1039,8 @@ class DxcCompiler : public IDxcCompiler3,
 
           dxcutil::AssembleInputs inputs(
               std::move(serializeModule), pOutputBlob, m_pMalloc,
-              SerializeFlags, pOutputStream, opts.GetPDBName(),
+              SerializeFlags, pOutputStream,
+              opts.SkipHash ? DxcValidatorFlags_SkipHash : 0, opts.GetPDBName(),
```
pow2clk, line 1043: If we want to maintain previous behavior, we may want to set the `skip_hash` flag when the internal validator is explicitly selected. I don't know why anyone would object to an unrequested hash, but the inclusion of the skip_hash flag suggests it's at least a possibility.

tools/clang/tools/dxcompiler/dxcutil.cpp
```diff
@@ -241,11 +230,13 @@ HRESULT ValidateAndAssembleToContainer(AssembleInputs &inputs) {
...
-                                         DxcValidatorFlags_InPlaceEdit,
+                                         DxcValidatorFlags_InPlaceEdit |
+                                             inputs.ValidationFlags,
                                          &debugModule, &pValResult));
     } else {
-      IFT(pValidator->Validate(inputs.pOutputContainerBlob,
-                               DxcValidatorFlags_InPlaceEdit, &pValResult));
+      IFT(pValidator->Validate(
+          inputs.pOutputContainerBlob,
+          DxcValidatorFlags_InPlaceEdit | inputs.ValidationFlags, &pValResult));
```
pow2clk, line 239: It might be slightly less intrusive if we just set the `InPlaceEdit` bit in `ValidationFlags` when `AssembleInputs` is created and could just pass in `inputs.ValidationFlags` here. I don't feel too strongly about it though if you prefer to leave it this way.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -31,6 +32,42 @@
...
+  if (Flags & DxcValidatorFlags_InPlaceEdit) {
+    HashAndUpdate((DxilContainerHeader *)Shader->GetBufferPointer());
+    *Hashed = Shader;
+    Shader->AddRef();
+  } else {
+    CComPtr<AbstractMemoryStream> HashedBlobStream;
+    uint32_t HR =
+        CreateMemoryStream(DxcGetThreadMallocNoRef(), &HashedBlobStream);
+    if (FAILED(HR))
+      throw hlsl::Exception(HR);
```
pow2clk, line 57: Just a note that you could have used the `IFT()` macro here and below since the return value isn't used later.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -180,18 +217,39 @@ uint32_t hlsl::validateWithOptDebugModule(
...
-        validationStatus, DXC_OUT_NONE,
-        {DxcOutputObject::ErrorOutput(
-            CP_UTF8, // TODO Support DefaultTextCodePage
-            (LPCSTR)pDiagBlob->GetBufferPointer(), pDiagBlob->GetBufferSize())},
-        Result);
-    if (FAILED(hr))
-      throw hlsl::Exception(hr);
+    if (Flags & (DxcValidatorFlags_ModuleOnly | DxcValidatorFlags_SkipHash)) {
+      // If we are validating a module only, we don't want to return the
+      // diagnostics stream.
```
pow2clk, line 222: It looks like it excludes the hash and not the diagnostics stream from modeuleonly.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -180,18 +217,39 @@ uint32_t hlsl::validateWithOptDebugModule(
...
+        throw hlsl::Exception(hr);
+      HashAndUpdateOrCopy(Flags, Shader, &HashedBlob);
+      hr = DxcResult::Create(
+          validationStatus, DXC_OUT_OBJECT,
+          {DxcOutputObject::DataOutput(DXC_OUT_OBJECT, HashedBlob),
+           DxcOutputObject::DataOutput(DXC_OUT_ERRORS, DiagBlobEnconding)},
+          Result);
+      if (FAILED(hr))
+        throw hlsl::Exception(hr);
+    }
```
pow2clk, line 252: I think we can test this behavior in a CompilerTest

lib/DxilContainer/DxcContainerBuilder.cpp
python3kgae: It is for DxilHash.h included in DxcContainerBuilder.h.
I've moved this change there.

lib/DxilContainer/DxcContainerBuilder.cpp
python3kgae: Updated.

lib/DxilContainer/DxcContainerBuilder.cpp
python3kgae: Do you mind I change it after this PR with another NFC PR?
Don't want to update other files in current PR.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
python3kgae: The idea here is that external validator will be the same as internal validator.
The only difference is that external validator is in dxil.dll which is smaller.
Then when user use dxc, there will be no different behavior when dxil.dll exist or not.

Maybe we don't need to have the concept of internal or external validator for dxc.
The external validator is only for things other than dxc to use?

tools/clang/tools/dxcompiler/dxcutil.cpp
python3kgae: Good idea.
Updated.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
python3kgae: Updated.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
python3kgae: Added test to ValidationTest.

lib/DxilContainer/DxcContainerBuilder.cpp
damyanp: There's multiple constants that this applies to in this file.  This seems like unnecessary churn to me.

include/dxc/Support/HLSLOptions.td
```diff
@@ -540,6 +540,8 @@ def all_resources_bound : Flag<["-", "/"], "all-resources-bound">, Flags<[CoreOp
   HelpText<"Enables agressive flattening">;
 def all_resources_bound_ : Flag<["-", "/"], "all_resources_bound">, Flags<[CoreOption, HelpHidden]>, Group<hlslcomp_Group>,
   HelpText<"Enables agressive flattening">;
+def skip_hash : Flag<["-", "/"], "skip-hash">, Flags<[CoreOption, HelpHidden]>, Group<hlslcomp_Group>,
```
damyanp, line 543: What's the use case for this?

python3kgae: Just in case for someone affected by behavior change has a work around.

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
pow2clk: As discussed, we may need to consider this more. In the meaintime, I think this is fine.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4071,6 +4072,43 @@ TEST_F(ValidationTest, ValidatePrintfNotAllowed) {
...
+  VERIFY_SUCCEEDED(pValidator->Validate(pProgram, Flags, &pResult));
+  // Make sure the validation was successful.
+  HRESULT status;
+  VERIFY_IS_NOT_NULL(pResult);
+  CComPtr<IDxcBlob> pValidationOutput;
+  pResult->GetStatus(&status);
+  VERIFY_SUCCEEDED(status);
+  pResult->GetResult(&pValidationOutput);
+  // Make sure the validation output is not null when hashing.
+  VERIFY_SUCCEEDED(pValidationOutput != nullptr);
```
pow2clk, line 4097: Can we not check that the hash value is reasonable?

pow2clk (approved):

include/dxc/Support/HLSLOptions.td
damyanp: We discussed offline - I'm keen to avoid adding new flags to the validator without any design review. I'd recommend removing the parts of this change around skipping hashes and filing an issue to follow up on it.

tools/clang/unittests/HLSL/ValidationTest.cpp
python3kgae: How about call DxilHash to get the golden for compare?
Because the input dxil module might change, we cannot get a stable value to compare.

include/dxc/Support/HLSLOptions.td
python3kgae: Removed.

tools/clang/unittests/HLSL/ValidationTest.cpp
python3kgae: Added hash check with DxilHash.

include/dxc/Support/HLSLOptions.h
```diff
@@ -223,6 +223,7 @@ class DxcOpts {
   unsigned long AutoBindingSpace = UINT_MAX; // OPT_auto_binding_space
   bool ExportShadersOnly = false;            // OPT_export_shaders_only
   bool ResMayAlias = false;                  // OPT_res_may_alias
+  bool SkipHash = false;                     // OPT_skip_hash
```
damyanp, line 226: should this be removed now?

python3kgae: Updated.

lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -164,11 +166,65 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
...
+      if (SUCCEEDED(HR)) {
+        CComPtr<IDxcBlob> pObject;
+        HR = (*ppResult)->GetResult(&pObject);
+        if (SUCCEEDED(HR)) {
+          LPVOID PTR = pObject->GetBufferPointer();
+          if (IsDxilContainerLike(PTR, pObject->GetBufferSize()))
+            HashAndUpdate((DxilContainerHeader *)PTR);
+        }
+      }
+    }
```
damyanp, line 183: We should deep nesting.  The rest of this function uses IFT rather than nested if (SUCCEEDED)'s.

```suggestion
    if (ppResult != nullptr && *ppResult != nullptr) {
      HRESULT HR;
      (*ppResult)->GetStatus(&HR);
      IFT(HR);

      CComPtr<IDxcBlob> pObject;
      IFT((*ppResult)->GetResult(&pObject));
      LPVOID PTR = pObject->GetBufferPointer();
      if (IsDxilContainerLike(PTR, pObject->GetBufferSize()))
        HashAndUpdate((DxilContainerHeader *)PTR);
    }
```

lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -164,11 +166,65 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
...
     return S_OK;
   }
   CATCH_CPP_RETURN_HRESULT();
 }
 
+// Try hashing the source contained in ContainerHeader using retail and debug
```
damyanp, line 189: I'm confused by the use of "source" here.  This doesn't refer to HLSL source code does it?  Also, it looks like it hashes the data stored in the container, not in the ContainerHeader.

The commend helps explain this function a bit more, but the name "FindHashFunctionFromSource" could be more descriptive.

damyanp: What if it were:

```c++
static HASH_FUNCTION_PROTO* DetermineHashFunctionFromContainerContents(
    const DxilContinerHeader* ContainerHeader) {
...
}
```

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
damyanp: There's very much an existing style in this file to use `IFT` here.  If we're going to continue to use exception based error handling in here then I think we should follow the convention that's here already.

If we're going to change the way error handling is done in this file then we might want to have a conversation about using exceptions at all here.

include/dxc/DxilContainer/DxcContainerBuilder.h
```diff
@@ -66,6 +69,12 @@ class DxcContainerBuilder : public IDxcContainerBuilder {
   const char *m_warning;
   bool m_RequireValidation;
   bool m_HasPrivateData;
+  // Function to compute hash when valid dxil container is built
+  // This is nullptr if loaded container has invalid hash
+  HASH_FUNCTION_PROTO *m_HashFunction;
```
damyanp, line 74: Why does this need to be stored?  Seems that it is only used in HashAndUpdate.  If there are ways of using DxcContainerBuilder that don't result in HashAndUpdate being called then this means we're always re-hashing existing content to determine the hash function even though we don't need it.

damyanp (changes requested):
See other comments.

include/dxc/DxilContainer/DxcContainerBuilder.h
damyanp: Ah, so the idea is that this is only set in DxcContainerBuilder::Load.  I assumed when I was looking that this must be the constructor because I can't find anywhere else that sets m_HashFunction - so what sets it to nullptr in the non-load cases?  I see that there's a DxcContainerBuilder::Init() function in the header that initializes the other members.

lib/DxilContainer/DxcContainerBuilder.cpp
python3kgae: Updated.

python3kgae: IFT will throw exception. The original code doesn't throw.

damyanp: Yeah, the pattern used throughout this function is that HRESULTs that we want to return as errors get thrown and then caught by CATCH_CPP_RETURN_HRESULT.  It's odd to have everything in SerializeContainer use the IFT approach for this and then suddenly switch to a different pattern 60 lines through it.

Although now I've read your latest version of this it is clear to me that the intent is to ignore any failures while trying to write out the hash?  Why's that?  Surely SerializeContainer has failed to do what it was asked to do if it couldn't add the hash.

damyanp: If there's a good reason for ignoring the errors it'd be good to have that explained in a comment here.

python3kgae: After remove IFT when verify root signature in dxc (that will allow dxc to check the error), return failed hr works.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
python3kgae: Updated.

lib/DxilContainer/DxcContainerBuilder.cpp
damyanp: I'm not sure we understand each other here, but it's at least a bit closer to the existing style in that function now.

damyanp (approved):

lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -164,9 +166,64 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
         {DxcOutputObject::DataOutput(DXC_OUT_OBJECT, pResult, DxcOutNoName),
          DxcOutputObject::DataOutput(DXC_OUT_ERRORS, pErrorBlob, DxcOutNoName)},
         ppResult));
-    return S_OK;
   }
   CATCH_CPP_RETURN_HRESULT();
+
+  if (ppResult == nullptr || *ppResult == nullptr)
+    return S_OK;
```
tex3d, line 173: We should have tested `ppResult` and returned `E_INVALIDARG` if nullptr at the beginning of the function.
For `*ppResult == nullptr`, that indicates a failure from `DxcResult::Create`, which should have been caught already.

With the current code, for either of these conditions, the `DxcResult::Create` should have failed causing a return under `CATCH_CPP_RETURN_HRESULT()` before getting here.

lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -164,9 +166,64 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
...
+
+  HRESULT HR;
+  (*ppResult)->GetStatus(&HR);
+  if (FAILED(HR))
+    return HR;
```
tex3d, line 178: This changes the protocol of the function.  The HRESULT returned from the function indicates whether an error so fatal has occurred that it is unable to even write error details to a `IDxcOperationResult **ppResult` object.  This is the pattern used in our APIs that use `IDxcOperationResult` or newer `IDxcResult` interfaces.  If this returns a failing HR, the caller shouldn't even attempt to look at `ppResult`.

This changes it to return a failing HR if the validator did, then `ppResult` would need to be looked at to see the validation failure details.

lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -164,9 +166,64 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
...
+
+  CComPtr<IDxcBlob> pObject;
+  IFR((*ppResult)->GetResult(&pObject));
+
+  // Add Hash.
+  LPVOID PTR = pObject->GetBufferPointer();
+  if (IsDxilContainerLike(PTR, pObject->GetBufferSize()))
+    HashAndUpdate((DxilContainerHeader *)PTR);
+  return S_OK;
```
tex3d, line 187: This code should be located within the try/catch, before the DxcResult construction, in case any fatal errors/exceptions were to occur in this process, and to allow error messages to be captured if necessary.  It should use `pResult` directly instead of loading `pObject` back out of `ppResult`.  If `IsContainerLike` were to fail, it indicates a malformed container result, which is a fatal error that should cause this to abort, not just skip container hashing.

tools/clang/tools/dxclib/dxc.cpp
```diff
@@ -644,7 +644,7 @@ int DxcContext::VerifyRootSignature() {
   IFT(pContainerBuilder->AddPart(hlsl::DxilFourCC::DFCC_RootSignature,
                                  pRootSignature));
   CComPtr<IDxcOperationResult> pOperationResult;
-  IFT(pContainerBuilder->SerializeContainer(&pOperationResult));
+  pContainerBuilder->SerializeContainer(&pOperationResult);
```
tex3d, line 647: We shouldn't have removed the `IFT` here.  Failing HRESULT indicates a serious error preventing `pOperationResult` construction.  We should not be accessing it in that case.

tex3d (commented):
I noticed some issues with the changes we already merged.  I was looking at this when trying to investigate potential behavior changes, since this merged while I was on vacation in the summer.

PR #6854, opened 2024-08-08
[SPIRV-V] Implement Shader Model 6.8 Expanded Comparison Sampling

cassiebeckley:
Compile `SampleCmpBias` using `OpImage*SampleDrefImplicitLod` and `SampleCmpGrad` using `OpImage*SampleDrefExplicitLod`.

The existing handlers for `CalculateLevelOfDetail` and `CalculateLevelOfDetailUnclamped` work for the
`SamplerComparisonSampler` overload, so no new code is needed other than tests.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -5474,6 +5474,35 @@ SpirvInstruction *SpirvEmitter::createImageSample(
   return retVal;
 }
 
+void SpirvEmitter::handleOptionalTextureSampleArgs(
+    const CXXMemberCallExpr *expr, uint32_t index,
+    SpirvInstruction **constOffset, SpirvInstruction **varOffset,
+    SpirvInstruction **clamp, SpirvInstruction **status) {
+  uint32_t numArgs = expr->getNumArgs();
+
+  bool hasOffsetArg = index < numArgs &&
+                      (expr->getArg(index)->getType()->isSignedIntegerType() ||
+                       hlsl::IsHLSLVecType(expr->getArg(index)->getType()));
+  if (hasOffsetArg) {
+    handleOffsetInMethodCall(expr, index, constOffset, varOffset);
+    index++;
+  }
+
+  bool hasClampArg =
+      index < numArgs && expr->getArg(index)->getType()->isFloatingType();
+  if (hasClampArg) {
+    *clamp = doExpr(expr->getArg(index));
+    index++;
+  }
+
+  const bool hasStatusArg =
+      index < numArgs &&
+      expr->getArg(index)->getType()->isUnsignedIntegerType();
+  if (hasStatusArg) {
+    *status = doExpr(expr->getArg(index));
+  }
+}
+
 SpirvInstruction *
 SpirvEmitter::processTextureSampleGather(const CXXMemberCallExpr *expr,
                                          const bool isSample) {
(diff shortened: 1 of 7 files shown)


tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5747,6 +5753,132 @@ SpirvEmitter::processTextureSampleCmp(const CXXMemberCallExpr *expr) {
...
+  const auto numArgs = expr->getNumArgs();
+  const bool hasStatusArg =
+      expr->getArg(numArgs - 1)->getType()->isUnsignedIntegerType();
+  auto *status = hasStatusArg ? doExpr(expr->getArg(numArgs - 1)) : nullptr;
+
+  SpirvInstruction *clamp = nullptr;
+  if (numArgs > 4 && expr->getArg(4)->getType()->isFloatingType())
+    clamp = doExpr(expr->getArg(4));
+  else if (numArgs > 5 && expr->getArg(5)->getType()->isFloatingType())
+    clamp = doExpr(expr->getArg(5));
```
s-perron, line 5790: Determining which argument is which by looking at the types is very hard to read. Could you do something with the type of `this`? You can get it from `expr->getImplicitObjectArgument()->getType()`.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5747,6 +5753,132 @@ SpirvEmitter::processTextureSampleCmp(const CXXMemberCallExpr *expr) {
...
+  //   float Bias,
+  //   float CompareValue
+  //   [, float Clamp]
+  //   [, out uint Status]
+  // );
+
+  const auto numArgs = expr->getNumArgs();
+  const bool hasStatusArg =
+      expr->getArg(numArgs - 1)->getType()->isUnsignedIntegerType();
+  auto *status = hasStatusArg ? doExpr(expr->getArg(numArgs - 1)) : nullptr;
```
s-perron, line 5784: Could you process the arguments in order? It would be easier to follow.

s-perron (changes requested):
Looks good. I like the amount of new tests. I think the code could be cleaned up a bit. I made comments in the cmp-bias function. I believe the same goes for the cmp-grad function.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: In this case the type of `this` doesn't give us much useful information. The only difference is whether or not `offset` may be present. I've refactored the code to try to match the optional arguments in order – I think this approach is more readable.

The new helper method could potentially be used in the other texture sampling methods as well.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: Done.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5751,6 +5786,108 @@ SpirvEmitter::processTextureSampleCmp(const CXXMemberCallExpr *expr) {
...
+
+  return createImageSample(
+      retType, imageType, image, sampler, coordinate, compareVal, bias,
+      /*lod*/ nullptr, std::make_pair(nullptr, nullptr), constOffset, varOffset,
+      /*constOffsets*/ nullptr, /*sampleNumber*/ nullptr, /*minLod*/ clamp,
+      status, expr->getCallee()->getLocStart(), expr->getSourceRange());
+}
+
+SpirvInstruction *
+SpirvEmitter::processTextureSampleCmpGrad(const CXXMemberCallExpr *expr) {
```
Keenuts, line 5844: Looks like those version are not supported?
```
SampleCmpGrad(SamplerState S, vector<float, CoordsCt> Location, float compareValue, vector<float, DimsCt> DDX, vector<float, DimsCt> DDY, int<dims> Offset)
SampleCmpGrad(SamplerState S, vector<float, CoordsCt> Location, float compareValue, vector<float, DimsCt> DDX, vector<float, DimsCt> DDY, int<dims> Offset, float Clamp)
SampleCmpGrad(SamplerState S, vector<float, CoordsCt> Location, float compareValue, vector<float, DimsCt> DDX, vector<float, DimsCt> DDY, int<dims> Offset, float Clamp, uint Status
```

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5751,6 +5786,108 @@ SpirvEmitter::processTextureSampleCmp(const CXXMemberCallExpr *expr) {
...
+
+  const auto *imageExpr = expr->getImplicitObjectArgument();
+  const QualType imageType = imageExpr->getType();
+  auto *image = loadIfGLValue(imageExpr);
+
+  auto *sampler = doExpr(expr->getArg(0));
+  auto *coordinate = doExpr(expr->getArg(1));
+  auto *compareVal = doExpr(expr->getArg(2));
+  auto *ddx = doExpr(expr->getArg(3));
+  auto *ddy = doExpr(expr->getArg(4));
```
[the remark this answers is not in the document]
cassiebeckley: Ah yeah, that's the inconsistency I raised during the meeting with MS. I've fixed the comment to reflect the actual parameters.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: I believe all three of these are supported?

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -5466,6 +5472,35 @@ SpirvInstruction *SpirvEmitter::createImageSample(
...
+    index++;
+  }
+
+  const bool hasStatusArg =
+      index < numArgs &&
+      expr->getArg(index)->getType()->isUnsignedIntegerType();
+  if (hasStatusArg) {
+    *status = doExpr(expr->getArg(index));
+  }
+}
```
s-perron, line 5502: I want to clarify something. Can a call get the status, but not the clamp argument? If not, then you do not have to check the types. You could do this:

```suggestion 
  if (index >= numArgs) 
    return;
    
  *clamp = doExpr(expr->getArg(index));
  index++;
  
  if (index >= numArgs) 
    return;

  *status = doExpr(expr->getArg(index));
  }
}
```

s-perron (commented):
LGTM, I just wonder why the optional argument parsing is more general than I expect. I could have the optional arguments understanding wrong.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: After rereading the proposal I believe you are correct:

> Note that these both have an additional overload that includes all the default parameters explicitly specified and a final output status parameter

I've updated to follow your suggestion.

s-perron (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
Keenuts: Ah right, sorry!

Keenuts (approved):

s-perron:
@cassiebeckley Can you rebase the PR and push it again? For some reason the tests are not running.

PR #6855, opened 2024-08-09
scalarrepl-param-hlsl: fix zero replacement in entry

amaiorano:
This is the same fix as was done in
https://github.com/microsoft/DirectXShaderCompiler/pull/6516 except that this is for replace uses of zero-init for instructions in the entry block.

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -3728,7 +3728,9 @@ static bool ReplaceUseOfZeroInitEntry(Instruction *I, Value *V) {
     // I is the last inst in the block after split.
     // Any inst in current block is before I.
     if (LoadInst *LI = dyn_cast<LoadInst>(UI)) {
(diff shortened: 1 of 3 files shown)


amaiorano:
Basically, in `ReplaceUseOfZeroInit()` we have:

```cpp
  bool bSuccess = false;
  if (&F->getEntryBlock() == I->getParent()) {
    bSuccess = ReplaceUseOfZeroInitEntry(I, GV);
  } else {
    DominatorTree DT;
    DT.recalculate(*F);
    SmallPtrSet<BasicBlock *, 8> Reachable;
    CollectReachableBBs(BB, Reachable);
    bSuccess = ReplaceUseOfZeroInit(I, GV, DT, Reachable);
  }
```

David's [fix](https://github.com/microsoft/DirectXShaderCompiler/pull/6516) covered the call to `ReplaceUseOfZeroInit` in the else branch. This fix does the same but in the call to `ReplaceUseOfZeroInitEntry` in the first branch.

llvm-beanz (approved):

sudonatalie (approved):

PR #6859, opened 2024-08-14
[Validator] Check content of PSV0 part in validation

python3kgae:
Replace the memcpy check for PSV0 part.
Build DxilPipelineStateValidation object from data in PSV0 part. Then compare the content with information from input DxilModule.

With this change, the order of signature elements and resources could be different between the container and the DxilModule.
And the StringTable could be in different order as well.

Second step for #6817

### DIFF
--- a/lib/DxilContainer/DxilPipelineStateValidation.cpp
+++ b/lib/DxilContainer/DxilPipelineStateValidation.cpp
@@ -360,14 +360,14 @@ void PSVResourceBindInfo1::Print(raw_ostream &OS) const {
 }
 
 void PSVSignatureElement::Print(raw_ostream &OS) const {
-  Print(OS, GetSemanticName());
+  Print(OS, GetSemanticName(), GetSemanticIndexes());
 }
 
-void PSVSignatureElement::Print(raw_ostream &OS, const char *Name) const {
+void PSVSignatureElement::Print(raw_ostream &OS, const char *Name,
+                                const uint32_t *SemanticIndexes) const {
   OS << "PSVSignatureElement:\n";
   OS << "  SemanticName: " << Name << "\n";
   OS << "  SemanticIndex: ";
-  const uint32_t *SemanticIndexes = GetSemanticIndexes();
   for (unsigned i = 0; i < GetRows(); ++i) {
     OS << *(SemanticIndexes + i) << " ";
   }
--- a/lib/DxilContainer/DxilPipelineStateValidation.cpp
+++ b/lib/DxilContainer/DxilPipelineStateValidation.cpp
@@ -834,6 +834,60 @@ void DxilPipelineStateValidation::PrintPSVRuntimeInfo(
       Comment);
 }
 
+void DxilPipelineStateValidation::PrintViewIDState(raw_ostream &OS) const {
+  unsigned NumStreams = IsGS() ? PSV_GS_MAX_STREAMS : 1;
+
+  if (m_pPSVRuntimeInfo1->UsesViewID) {
+    for (unsigned i = 0; i < NumStreams; ++i) {
+      OS << "Outputs affected by ViewID as a bitmask for stream " << i
+         << ":\n ";
+      uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigOutputVectors[i];
+      const PSVComponentMask ViewIDMask(m_pViewIDOutputMask[i], OutputVectors);
+      std::string OutputSetName = "Outputs";
+      OutputSetName += "[" + std::to_string(i) + "]";
+      ViewIDMask.Print(OS, "ViewID", OutputSetName.c_str());
+    }
+
+    if (IsHS() || IsMS()) {
+      OS << "PCOutputs affected by ViewID as a bitmask:\n";
+      uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigPatchConstOrPrimVectors;
+      const PSVComponentMask ViewIDMask(m_pViewIDPCOrPrimOutputMask,
+                                        OutputVectors);
+      ViewIDMask.Print(OS, "ViewID", "PCOutputs");
+    }
+  }
+
+  for (unsigned i = 0; i < NumStreams; ++i) {
+    OS << "Outputs affected by inputs as a table of bitmasks for stream " << i
+       << ":\n";
+    uint8_t InputVectors = m_pPSVRuntimeInfo1->SigInputVectors;
+    uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigOutputVectors[i];
+    const PSVDependencyTable Table(m_pInputToOutputTable[i], InputVectors,
+                                   OutputVectors);
+    std::string OutputSetName = "Outputs";
+    OutputSetName += "[" + std::to_string(i) + "]";
+    Table.Print(OS, "Inputs", OutputSetName.c_str());
+  }
+
+  if (IsHS()) {
+    OS << "Patch constant outputs affected by inputs as a table of "
+          "bitmasks:\n";
+    uint8_t InputVectors = m_pPSVRuntimeInfo1->SigInputVectors;
+    uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigPatchConstOrPrimVectors;
+    const PSVDependencyTable Table(m_pInputToPCOutputTable, InputVectors,
+                                   OutputVectors);
+    Table.Print(OS, "Inputs", "PatchConstantOutputs");
+  } else if (IsDS()) {
+    OS << "Outputs affected by patch constant inputs as a table of "
+          "bitmasks:\n";
+    uint8_t InputVectors = m_pPSVRuntimeInfo1->SigPatchConstOrPrimVectors;
+    uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigOutputVectors[0];
+    const PSVDependencyTable Table(m_pPCInputToOutputTable, InputVectors,
+                                   OutputVectors);
+    Table.Print(OS, "PatchConstantInputs", "Outputs");
+  }
+}
+
 void DxilPipelineStateValidation::Print(raw_ostream &OS,
                                         uint8_t ShaderKind) const {
   OS << "DxilPipelineStateValidation:\n";
@@ -878,57 +932,6 @@ void DxilPipelineStateValidation::Print(raw_ostream &OS,
       PSVSE.Print(OS);
     }
 
-    unsigned NumStreams = IsGS() ? PSV_GS_MAX_STREAMS : 1;
-
-    if (m_pPSVRuntimeInfo1->UsesViewID) {
-      for (unsigned i = 0; i < NumStreams; ++i) {
-        OS << "Outputs affected by ViewID as a bitmask for stream " << i
-           << ":\n ";
-        uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigOutputVectors[i];
-        const PSVComponentMask ViewIDMask(m_pViewIDOutputMask[i],
-                                          OutputVectors);
-        std::string OutputSetName = "Outputs";
-        OutputSetName += "[" + std::to_string(i) + "]";
-        ViewIDMask.Print(OS, "ViewID", OutputSetName.c_str());
-      }
-
-      if (IsHS() || IsMS()) {
-        OS << "PCOutputs affected by ViewID as a bitmask:\n";
-        uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigPatchConstOrPrimVectors;
-        const PSVComponentMask ViewIDMask(m_pViewIDPCOrPrimOutputMask,
-                                          OutputVectors);
-        ViewIDMask.Print(OS, "ViewID", "PCOutputs");
-      }
-    }
-
-    for (unsigned i = 0; i < NumStreams; ++i) {
-      OS << "Outputs affected by inputs as a table of bitmasks for stream " << i
-         << ":\n";
-      uint8_t InputVectors = m_pPSVRuntimeInfo1->SigInputVectors;
-      uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigOutputVectors[i];
-      const PSVDependencyTable Table(m_pInputToOutputTable[i], InputVectors,
-                                     OutputVectors);
-      std::string OutputSetName = "Outputs";
-      OutputSetName += "[" + std::to_string(i) + "]";
-      Table.Print(OS, "Inputs", OutputSetName.c_str());
-    }
-
-    if (IsHS()) {
-      OS << "Patch constant outputs affected by inputs as a table of "
-            "bitmasks:\n";
-      uint8_t InputVectors = m_pPSVRuntimeInfo1->SigInputVectors;
-      uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigPatchConstOrPrimVectors;
-      const PSVDependencyTable Table(m_pInputToPCOutputTable, InputVectors,
-                                     OutputVectors);
-      Table.Print(OS, "Inputs", "PatchConstantOutputs");
-    } else if (IsDS()) {
-      OS << "Outputs affected by patch constant inputs as a table of "
-            "bitmasks:\n";
-      uint8_t InputVectors = m_pPSVRuntimeInfo1->SigPatchConstOrPrimVectors;
-      uint8_t OutputVectors = m_pPSVRuntimeInfo1->SigOutputVectors[0];
-      const PSVDependencyTable Table(m_pPCInputToOutputTable, InputVectors,
-                                     OutputVectors);
-      Table.Print(OS, "PatchConstantInputs", "Outputs");
-    }
+    PrintViewIDState(OS);
   }
 }
--- a/lib/DxilContainer/DxilPipelineStateValidation.cpp
+++ b/lib/DxilContainer/DxilPipelineStateValidation.cpp
@@ -24,6 +24,18 @@
 using namespace hlsl;
 using namespace llvm;
 
+uint32_t hlsl::GetPSVVersion(uint32_t ValMajor, uint32_t ValMinor) {
+  unsigned PSVVersion = MAX_PSV_VERSION;
+  // Constraint PSVVersion based on validator version
+  if (DXIL::CompareVersions(ValMajor, ValMinor, 1, 1) < 0)
+    PSVVersion = 0;
+  else if (DXIL::CompareVersions(ValMajor, ValMinor, 1, 6) < 0)
+    PSVVersion = 1;
+  else if (DXIL::CompareVersions(ValMajor, ValMinor, 1, 8) < 0)
+    PSVVersion = 2;
+  return PSVVersion;
+}
+
 void hlsl::InitPSVResourceBinding(PSVResourceBindInfo0 *Bind0,
                                   PSVResourceBindInfo1 *Bind1,
                                   DxilResourceBase *Res) {
--- a/lib/DxilContainer/DxilPipelineStateValidation.cpp
+++ b/lib/DxilContainer/DxilPipelineStateValidation.cpp
@@ -360,8 +360,12 @@ void PSVResourceBindInfo1::Print(raw_ostream &OS) const {
 }
 
 void PSVSignatureElement::Print(raw_ostream &OS) const {
+  Print(OS, GetSemanticName());
+}
+
+void PSVSignatureElement::Print(raw_ostream &OS, const char *Name) const {
   OS << "PSVSignatureElement:\n";
-  OS << "  SemanticName: " << GetSemanticName() << "\n";
+  OS << "  SemanticName: " << Name << "\n";
   OS << "  SemanticIndex: ";
   const uint32_t *SemanticIndexes = GetSemanticIndexes();
   for (unsigned i = 0; i < GetRows(); ++i) {
@@ -518,13 +522,10 @@ void PSVDependencyTable::Print(raw_ostream &OS, const char *InputSetName,
   }
 }
 
-void DxilPipelineStateValidation::PrintPSVRuntimeInfo(
-    raw_ostream &OS, uint8_t ShaderKind, const char *Comment) const {
-  PSVRuntimeInfo0 *pInfo0 = m_pPSVRuntimeInfo0;
-  PSVRuntimeInfo1 *pInfo1 = m_pPSVRuntimeInfo1;
-  PSVRuntimeInfo2 *pInfo2 = m_pPSVRuntimeInfo2;
-  PSVRuntimeInfo3 *pInfo3 = m_pPSVRuntimeInfo3;
-
+void hlsl::PrintPSVRuntimeInfo(llvm::raw_ostream &OS, PSVRuntimeInfo0 *pInfo0,
+                               PSVRuntimeInfo1 *pInfo1, PSVRuntimeInfo2 *pInfo2,
+                               PSVRuntimeInfo3 *pInfo3, uint8_t ShaderKind,
+                               const char *EntryName, const char *Comment) {
   if (pInfo1 && pInfo1->ShaderStage != ShaderKind)
     ShaderKind = pInfo1->ShaderStage;
   OS << Comment << "PSVRuntimeInfo:\n";
@@ -817,9 +818,20 @@ void DxilPipelineStateValidation::PrintPSVRuntimeInfo(
     }
   }
   if (pInfo3)
-    OS << Comment
-       << " EntryFunctionName: " << m_StringTable.Get(pInfo3->EntryFunctionName)
-       << "\n";
+    OS << Comment << " EntryFunctionName: " << EntryName << "\n";
+}
+
+void DxilPipelineStateValidation::PrintPSVRuntimeInfo(
+    raw_ostream &OS, uint8_t ShaderKind, const char *Comment) const {
+  PSVRuntimeInfo0 *pInfo0 = m_pPSVRuntimeInfo0;
+  PSVRuntimeInfo1 *pInfo1 = m_pPSVRuntimeInfo1;
+  PSVRuntimeInfo2 *pInfo2 = m_pPSVRuntimeInfo2;
+  PSVRuntimeInfo3 *pInfo3 = m_pPSVRuntimeInfo3;
+
+  hlsl::PrintPSVRuntimeInfo(
+      OS, pInfo0, pInfo1, pInfo2, pInfo3, ShaderKind,
+      m_pPSVRuntimeInfo3 ? m_StringTable.Get(pInfo3->EntryFunctionName) : "",
+      Comment);
 }
 
 void DxilPipelineStateValidation::Print(raw_ostream &OS,
--- a/lib/DxilContainer/DxilPipelineStateValidation.cpp
+++ b/lib/DxilContainer/DxilPipelineStateValidation.cpp
@@ -110,9 +110,7 @@ void hlsl::InitPSVSignatureElement(PSVSignatureElement0 &E,
   E.DynamicMaskAndStream |= (SE.GetDynIdxCompMask()) & 0xF;
 }
 
-void hlsl::InitPSVRuntimeInfo(PSVRuntimeInfo0 *pInfo, PSVRuntimeInfo1 *pInfo1,
-                              PSVRuntimeInfo2 *pInfo2, PSVRuntimeInfo3 *pInfo3,
-                              const DxilModule &DM) {
+void hlsl::SetShaderProps(PSVRuntimeInfo0 *pInfo, const DxilModule &DM) {
   const ShaderModel *SM = DM.GetShaderModel();
   pInfo->MinimumExpectedWaveLaneCount = 0;
   pInfo->MaximumExpectedWaveLaneCount = (uint32_t)-1;
@@ -217,7 +215,6 @@ void hlsl::InitPSVRuntimeInfo(PSVRuntimeInfo0 *pInfo, PSVRuntimeInfo1 *pInfo1,
   case ShaderModel::Kind::Mesh: {
     pInfo->MS.MaxOutputVertices = (uint16_t)DM.GetMaxOutputVertices();
     pInfo->MS.MaxOutputPrimitives = (uint16_t)DM.GetMaxOutputPrimitives();
-    pInfo1->MS1.MeshOutputTopology = (uint8_t)DM.GetMeshOutputTopology();
     Module *mod = DM.GetModule();
     const DataLayout &DL = mod->getDataLayout();
     unsigned totalByteSize = 0;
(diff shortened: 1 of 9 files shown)


lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +76,672 @@ struct DiagRestore {
 static void emitDxilDiag(LLVMContext &Ctx, const char *str) {
   hlsl::dxilutil::EmitErrorOnContext(Ctx, str);
 }
+
+struct MemSpan {
+  uint32_t *Ptr = nullptr;
+  uint32_t Size = 0;
+  bool isEmpty() { return Size == 0; }
+};
```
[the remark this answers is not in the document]
python3kgae: Updated to ArrayRef.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +76,667 @@ struct DiagRestore {
...
+  NumInputSigScalars = Data[Offset++];
+  unsigned NumStreams =
+      SK == PSVShaderKind::Geometry ? DXIL::kNumOutputStreams : 1;
+  for (unsigned i = 0; i < NumStreams; i++) {
+    // #Outputs for stream StreamId.
+    unsigned OutputScalars = Data[Offset++];
+    NumOutputSigScalars[i] = OutputScalars;
+    unsigned MaskDwords = RoundUpToUINT(OutputScalars);
+    if (UsesViewID) {
+      ViewIDOutputMask[i] = {Data.data() + Offset, MaskDwords};
```
llvm-beanz, line 111: Can you do this with `ArrayRef::operator[]` instead? That gives us bounds checking.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +76,667 @@ struct DiagRestore {
...
+                             unsigned NumInputScalars,
+                             unsigned NumOutputScalars);
+  void VerifyEntryProperties(const ShaderModel *SM, PSVRuntimeInfo0 *PSV0,
+                             PSVRuntimeInfo1 *PSV1, PSVRuntimeInfo2 *PSV2);
+  void EmitFormatError(StringRef Name) {
+#ifdef _DEBUG
+    // Only emit detailed error for debugging.
+    std::string ErrorMsg = Name.str() + " does not match";
+    ValCtx.EmitFormatError(ValidationRule::ContainerPartMatches, {ErrorMsg});
+#endif
```
llvm-beanz, line 172: It seems weird to have validation messages that are assert-only enabled.

llvm-beanz (commented):
I feel like there is a lot of missing test coverage here. I don't see tests for the various errors. I also think the error messages don't seem to be particularly descriptive, so I'm not even sure how a compiler implementer would interpret the errors and know what was wrong.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +76,667 @@ struct DiagRestore {
...
+SimpleViewIDState::SimpleViewIDState(std::vector<uint32_t> &Data,
+                                     PSVShaderKind SK, bool UsesViewID) {
+  if (Data.empty())
+    return;
+
+  unsigned Offset = 0;
+  // #Inputs.
+  NumInputSigScalars = Data[Offset++];
+  unsigned NumStreams =
+      SK == PSVShaderKind::Geometry ? DXIL::kNumOutputStreams : 1;
```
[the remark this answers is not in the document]
python3kgae: The container code is here
https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/DxilContainer/DxilContainerAssembler.cpp#L507
This is ViewID related information, not the signature itself.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +76,667 @@ struct DiagRestore {
...
+  ArrayRef<uint32_t> InputToOutputTable[DXIL::kNumOutputStreams];
+  ArrayRef<uint32_t> InputToPCOutputTable;
+  ArrayRef<uint32_t> PCInputToOutputTable;
+  ArrayRef<uint32_t> ViewIDOutputMask[DXIL::kNumOutputStreams];
+  ArrayRef<uint32_t> ViewIDPCOutputMask;
+  bool IsValid = true;
+  SimpleViewIDState(std::vector<uint32_t> &Data, PSVShaderKind,
+                    bool UsesViewID);
+};
+static unsigned RoundUpToUINT(unsigned x) { return (x + 31) / 32; }
```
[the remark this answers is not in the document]
llvm-beanz: Why are we adding this function at all? This is what `llvm::RoundUpToAlignment` from `llvm/Support/MathExtras.h` does.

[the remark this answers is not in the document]
python3kgae: Updated.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: Something like this?
```
ViewIDOutputMask[i] = {&(Data[Offset]), MaskDwords};
```

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,610 @@ struct DiagRestore {
...
+                                         bool i1ToUnknownCompat) {
+  if (Count != Sig.GetElements().size()) {
+    EmitMismatchError(Name + "Elements", std::to_string(Count),
+                      std::to_string(Sig.GetElements().size()));
+    return;
+  }
+
+  ArrayRef<PSVSignatureElement0> Inputs(Base, Count);
+  // Build PSVSignatureSet with data in PSV.
+  PSVSignatureSet SigSet;
```
tex3d, line 425: This code allows signatures to differ in ordering between DxilModule and PSV0.  That should not be considered legal.  The ordering of signature elements must not differ.  What possible reason would you want to relax this?

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,610 @@ struct DiagRestore {
...
+  UINT uSRVs = DM.GetSRVs().size();
+  UINT uUAVs = DM.GetUAVs().size();
+  unsigned ResourceCount = uCBuffers + uSamplers + uSRVs + uUAVs;
+  if (PSV.GetBindCount() != ResourceCount) {
+    EmitMismatchError("ResourceCount", std::to_string(PSV.GetBindCount()),
+                      std::to_string(ResourceCount));
+    return;
+  }
+  // Build PSVResourceSet with data in PSV.
+  PSVResourceSet ResBindInfo0Set;
```
tex3d, line 549: What's the justification for relaxing the ordering requirement of resources defined in PSV0 in relation to DxilModule resources?  I don't think this should be allowed.

tex3d (changes requested):
This does two major things I don't think we should do:
- relaxes ordering requirement for signature elements relative to DxilModule (absolutely should not do this)
- relaxes ordering requirement for resources relative to DxilModule (I don't see any need or advantage to this)

I have other more minor comments, and I haven't completed reviewing some parts and the tests at this point.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+                  PSV.GetSigOutputElements(), "SigOutput", i1ToUnknownCompat);
+  // Verify patch constant signature
+  VerifySignature(DM.GetPatchConstOrPrimSignature(),
+                  PSV.GetPatchConstOrPrimElement0(0),
+                  PSV.GetSigPatchConstOrPrimElements(),
+                  "SigPatchConstantOrPrim", i1ToUnknownCompat);
+}
+
+void PSVContentVerifier::VerifySignature(const DxilSignature &Sig,
+                                         PSVSignatureElement0 *Base,
```
tex3d, line 370: Technically, with the PSV0 versioning scheme, this pattern that assumes this will always be an array of `PSVSignatureElement0` is not safe.  `DxilPipelineStateValidation::m_uPSVSignatureElementSize` must be used to index into the elements in the PSV0 data.

Furthermore, that stride (and all strides) should be checked for validity based on the version.  Loading will only check that `sizeof(PSVSignatureElement0) <= m_uPSVSignatureElementSize`.

The loader can handle a newer version of the format, but the validator needs to make sure the format is constrained to what it knows is valid.  This means the validator should verify that `sizeof(PSVSignatureElement0) == m_uPSVSignatureElementSize` for all versions of PSV0 defined so far.

I see that the record strides are not part of the public interface of the `DxilPipelineStateValidation` class though, so these may need exposure to enable both stride validation, and use of the `GetRecord<>()` accessor based on base pointer, stride, and element count.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+    OS1.flush();
+    EmitMismatchError("PSVRuntimeInfo", Str, Str1);
+  }
+}
+
+void PSVContentVerifier::Verify() {
+  unsigned ValMajor, ValMinor;
+  DM.GetValidatorVersion(ValMajor, ValMinor);
+  unsigned PSVVersion = MAX_PSV_VERSION;
+  // Constraint PSVVersion based on validator version
```
tex3d, line 534: Minor: Identifying `PSVVersion` from `ValMajor`/`ValMinor` should probably be in `DxilConstants.h` or some other shared location like that.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+  if (DXIL::CompareVersions(ValMajor, ValMinor, 1, 1) < 0)
+    PSVVersion = 0;
+  else if (DXIL::CompareVersions(ValMajor, ValMinor, 1, 6) < 0)
+    PSVVersion = 1;
+  else if (DXIL::CompareVersions(ValMajor, ValMinor, 1, 8) < 0)
+    PSVVersion = 2;
```
tex3d, line 540: After this point, we should verify that the strides in the loaded PSV are the same as retrieved from `PSVInitInfo(PSVVersion).<record>Size()`.  That will require exposing those strides publicly from `DxilPipelineStateValidation`, since it does not do so now.

tex3d: Note that this will make it unnecessary to check all the cases below for presence of specific `PSVRuntimeInfoN` against `PSVVersion`, since those are just based on this stride.

python3kgae: Updated.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: Updated.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: Switched to GetRecord instead of Base + i.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: Fixed.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: Fixed.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,610 @@ struct DiagRestore {
...
+    ValCtx.EmitFormatError(ValidationRule::ContainerContentDuplicates,
+                           {Name, DupContent});
+    PSVContentValid = false;
+  }
+  void EmitMissingError(StringRef Name, StringRef Content) {
+    ValCtx.EmitFormatError(ValidationRule::ContainerContentMissing,
+                           {Name, Content});
+    PSVContentValid = false;
+  }
+  void EmitMismatchError(StringRef Name, StringRef Expected, StringRef Actual) {
```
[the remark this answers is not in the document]
python3kgae: Fixed.

utils/hct/hctdb.py
```diff
@@ -6923,7 +6923,26 @@ def build_valrules(self):
...
+        )
+        self.add_valrule_msg(
+            "Container.ContentMissing",
+            "DXIL Container Content missing",
+            "Cannot find '%0' with content '%1' in dxil container.",
+        )
+        self.add_valrule_msg(
+            "Container.ContentInvalid",
+            "DXIL Container Content invalid",
+            "'%0' is invalid in dxil container.",
```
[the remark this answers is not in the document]
python3kgae: Fixed.

utils/hct/hctdb.py
```diff
@@ -6923,7 +6923,26 @@ def build_valrules(self):
...
+        )
+        self.add_valrule_msg(
+            "Container.ContentDuplicates",
+            "DXIL Container Content got duplication",
+            "'%0' with content '%1' duplicates.",
+        )
+        self.add_valrule_msg(
+            "Container.ContentMissing",
+            "DXIL Container Content missing",
+            "Cannot find '%0' with content '%1' in dxil container.",
```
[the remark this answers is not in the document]
python3kgae: Removed for not allow reorder.

utils/hct/hctdb.py
```diff
@@ -6923,7 +6923,26 @@ def build_valrules(self):
...
-
+        self.add_valrule_msg(
+            "Container.ContentMatches",
+            "DXIL Container Content must match Module",
+            "'%0' does not match, for container: '%1', for dxil module: '%2'.",
+        )
+        self.add_valrule_msg(
+            "Container.ContentDuplicates",
+            "DXIL Container Content got duplication",
+            "'%0' with content '%1' duplicates.",
```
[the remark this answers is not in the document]
python3kgae: Removed for not allow reorder.

utils/hct/hctdb.py
```diff
@@ -6923,7 +6923,26 @@ def build_valrules(self):
...
+        )
+        self.add_valrule_msg(
+            "Container.ContentMissing",
+            "DXIL Container Content missing",
+            "Cannot find '%0' with content '%1' in dxil container.",
+        )
+        self.add_valrule_msg(
+            "Container.ContentInvalid",
+            "DXIL Container Content invalid",
```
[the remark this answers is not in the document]
python3kgae: Updated.

utils/hct/hctdb.py
```diff
@@ -6923,7 +6923,26 @@ def build_valrules(self):
...
         )
-
+        self.add_valrule_msg(
+            "Container.ContentMatches",
+            "DXIL Container Content must match Module",
+            "'%0' does not match, for container: '%1', for dxil module: '%2'.",
+        )
+        self.add_valrule_msg(
+            "Container.ContentDuplicates",
+            "DXIL Container Content got duplication",
```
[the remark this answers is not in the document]
python3kgae: Removed for not allow reorder.

utils/hct/hctdb.py
```diff
@@ -6923,7 +6923,26 @@ def build_valrules(self):
             "Root Signature in DXIL Container must be compatible with shader",
             "Root Signature in DXIL container is not compatible with shader.",
         )
-
+        self.add_valrule_msg(
+            "Container.ContentMatches",
+            "DXIL Container Content must match Module",
+            "'%0' does not match, for container: '%1', for dxil module: '%2'.",
```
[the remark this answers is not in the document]
python3kgae: Fixed.

utils/hct/hctdb.py
```diff
@@ -6923,7 +6923,26 @@ def build_valrules(self):
...
+        )
+        self.add_valrule_msg(
+            "Container.ContentDuplicates",
+            "DXIL Container Content got duplication",
+            "'%0' with content '%1' duplicates.",
+        )
+        self.add_valrule_msg(
+            "Container.ContentMissing",
+            "DXIL Container Content missing",
```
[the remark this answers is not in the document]
python3kgae: Removed for not allow reorder.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,523 @@ struct DiagRestore {
...
+    if (((PSV.GetSigInputElements() + PSV.GetSigOutputElements() +
+          PSV.GetSigPatchConstOrPrimElements()) > 0) &&
+        PSV.GetSignatureElementSize() != PSVInfo.SignatureElementSize()) {
+      EmitMismatchError("SignatureElementSize",
+                        std::to_string(PSV.GetSignatureElementSize()),
+                        std::to_string(PSVInfo.SignatureElementSize()));
+      return;
+    }
+    uint8_t ShaderStage = static_cast<uint8_t>(SM->GetKind());
+    PSVRuntimeInfo1 *PSV1 = PSV.GetPSVRuntimeInfo1();
```
tex3d, line 570: We already have `PSV1` set above, (line 555) so don't need to shadow that variable here, or `VerifyEntryProperties` could just take `PSV` and get the versioned structs itself.

python3kgae: Removed.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+
+    VerifyViewIDDependence(PSV1);
+  }
+  if (PSVVersion > 1) {
+    // PSV2 only added NumThreadsX/Y/Z which verified in VerifyEntryProperties.
+  }
+  if (PSVVersion > 2) {
+    if (DM.GetEntryFunctionName() != PSV.GetEntryFunctionName())
+      EmitMismatchError("EntryFunctionName", DM.GetEntryFunctionName(),
+                        PSV.GetEntryFunctionName());
```
tex3d, line 589: Value from PSV0 and DxilModule were reversed.
```suggestion
      EmitMismatchError("EntryFunctionName", PSV.GetEntryFunctionName(),
                        DM.GetEntryFunctionName());
```

python3kgae: Good catch.
Fixed.

lib/DxilContainer/DxilPipelineStateValidation.cpp
```diff
@@ -817,9 +830,74 @@ void DxilPipelineStateValidation::PrintPSVRuntimeInfo(
...
+  PSVRuntimeInfo2 *pInfo2 = m_pPSVRuntimeInfo2;
+  PSVRuntimeInfo3 *pInfo3 = m_pPSVRuntimeInfo3;
+
+  hlsl::PrintPSVRuntimeInfo(
+      OS, pInfo0, pInfo1, pInfo2, pInfo3, ShaderKind,
+      m_pPSVRuntimeInfo3 ? m_StringTable.Get(pInfo3->EntryFunctionName) : "",
+      Comment);
+}
+
+void DxilPipelineStateValidation::PrintViewIDState(raw_ostream &OS) const {
```
pow2clk, line 849: This looks quite similar to the SimpleViewID::Print() function. Is there no way to merge these? Maybe similar to how PrintPSVRuntimeInfo can be called from DxilValidation now?

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
 static void emitDxilDiag(LLVMContext &Ctx, const char *str) {
   hlsl::dxilutil::EmitErrorOnContext(Ctx, str);
 }
+
+// Helper class which build PSV ViewID state from serialized view id state.
+struct SimpleViewIDState {
```
pow2clk, line 83: This shares an awful lot with `DxilViewIdState`. Can this be made a sublclass or share the common code some other way?

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+  unsigned NumInputSigScalars = 0;
+  unsigned NumOutputSigScalars[DXIL::kNumOutputStreams] = {0, 0, 0, 0};
+  unsigned NumPCOrPrimSigScalars = 0;
```
pow2clk, line 86: These are all three stored as scalars, but every one of them is aligned by 4 and converted to vectors every time they are retrieved. The only exception is the constructor where they are originally assigned where that original value is readily available. I think it would be cleaner to store these as vectors rather than perform that conversion every time.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+        ShaderStage == PSVShaderKind::Mesh) {
+      OS << "Patch constant outputs affected by inputs as a table of "
+            "bitmasks:\n";
+      uint8_t InputVectors =
+          llvm::RoundUpToAlignment(NumInputSigScalars, 4) / 4;
+      uint8_t OutputVectors =
+          llvm::RoundUpToAlignment(NumPCOrPrimSigScalars, 4) / 4;
+      const PSVDependencyTable Table(InputToPCOutputTable.data(), InputVectors,
+                                     OutputVectors);
+      Table.Print(OS, "Inputs", "PatchConstantOutputs");
```
pow2clk, line 149: Mesh shaders should refer to these as output primitives.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+          llvm::RoundUpToAlignment(NumPCOrPrimSigScalars, 4) / 4;
+      uint8_t OutputVectors =
+          llvm::RoundUpToAlignment(NumOutputSigScalars[0], 4) / 4;
+      const PSVDependencyTable Table(PCInputToOutputTable.data(), InputVectors,
+                                     OutputVectors);
+      Table.Print(OS, "PatchConstantInputs", "Outputs");
+    }
+  }
+};
+SimpleViewIDState::SimpleViewIDState(std::vector<uint32_t> &Data,
```
pow2clk, line 163: We now have largely the same code in three places: `DxilViewIdState::Serialize`, `ComputeSerializedViewIDStateSizeInUInts` (which suggests merging itself with the first) and here. Can we not collect this somehow? If nothing else, we should backport the switch to `llvm::RoundUpToAlignment` to those locations.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+                                     PSVShaderKind SK, bool UsesViewID) {
+  if (Data.empty())
+    return;
+
+  unsigned Offset = 0;
+  // #Inputs.
```
pow2clk, line 169: Why the hashes? (`#`)? It looks like they were copied that way from the code in one of the two places that have the same logic. Regardless, these are pretty unhelpful comments. Can we get full sentences?

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+  NumInputSigScalars = Data[Offset++];
```
pow2clk, line 170: `NumInputSigScalars` is used a few times in this function without conversion, but it could be stored to a temporary stack variable for those and stored in the class aligned and converted to vectors as that's how it's used everywhere else.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+    unsigned TabSize = MaskDwords * NumInputSigScalars;
+    if (TabSize == 0)
+      continue;
+    unsigned AlignedTabSize =
+        MaskDwords * llvm::RoundUpToAlignment(NumInputSigScalars, 4);
+    InputToOutputTable[i] = std::vector<uint32_t>(AlignedTabSize, 0);
+    memcpy(InputToOutputTable[i].data(), &(Data[Offset]), TabSize * 4);
+    Offset += TabSize;
+  }
+  if (SK == PSVShaderKind::Hull || SK == PSVShaderKind::Mesh) {
```
pow2clk, line 191: These Hull/Mesh and Domain cases are the only places where `NumPCOrPrimSigScalars and the contents of `NumOutputSigSAcalars` are used without converting to vectors, but the values are retrieved in this function and could be temporarily saved for use here.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+                         StringRef ModuleContent) {
+    ValCtx.EmitFormatError(ValidationRule::ContainerContentMatches,
+                           {Name, "PSV0", PartContent, ModuleContent});
+    PSVContentValid = false;
+  }
+  void EmitInvalidError(StringRef Name) {
+    ValCtx.EmitFormatError(ValidationRule::ContainerContentInvalid, {Name});
+    PSVContentValid = false;
+  }
+  template <typename Ty> static std::string GetDump(const Ty &T) {
```
pow2clk, line 262: Just a note, though this is a full template, it's only ever used to output BindInfo0 and its subclass BindInfo1

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+    if (PSVTable != DxilTable)
+      return true;
+  }
+
+  if (UsesViewID && NumOutputVectors > 0) {
+    PSVComponentMask DxilMask(ArrayMask.data(), NumOutputVectors);
+    if (PSVMask != DxilMask)
+      return true;
+  }
+  return false;
```
pow2clk, line 289: Can you explain this a bit? IN cases where there are only input vectors, or only output vectors and ViewID isn't used, there isn't any check for matching at all. Is that okay? If these situations are not expected to happen, I think this should still return that they mismatched and/or assert.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+    std::string Str;
+    raw_string_ostream OS(Str);
+    PSV.PrintViewIDState(OS);
+    OS.flush();
+    std::string Str1;
+    raw_string_ostream OS1(Str1);
+    ViewIDState.Print(OS1, static_cast<PSVShaderKind>(PSV1->ShaderStage),
+                      PSV1->UsesViewID);
+    OS1.flush();
+    EmitMismatchError("ViewIDState", Str, Str1);
```
pow2clk, line 349: Feels a bit weird to submit multiline strings as parameters to an error message, it would be easier to read if the specific mismatch were identified and printed. I'm not sure I'm ready to insist on that though.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+  else
+    Mismatch = true;
+
+  ModulePSVSE0.SemanticIndexes = PSVSE0->SemanticIndexes;
+
+  PSVSignatureElement PSVSE(StrTab, IndexTab, PSVSE0);
+  if (SE.IsArbitrary())
+    Mismatch |= strcmp(PSVSE.GetSemanticName(), SE.GetName());
+  else
+    Mismatch |= PSVSE0->SemanticKind != static_cast<uint8_t>(SE.GetKind());
```
pow2clk, line 414: Are both of these conditions tested? It looks like semantickind is not copied over.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+  Mismatch |= memcmp(&ModulePSVSE0, PSVSE0, sizeof(PSVSignatureElement0)) != 0;
+  if (Mismatch) {
+    PSVSignatureElement ModulePSVSE(StrTab, IndexTab, &ModulePSVSE0);
+    std::string ModuleStr;
+    raw_string_ostream OS(ModuleStr);
+    ModulePSVSE.Print(OS, SE.GetName(), SemanticIndexVec.data());
+    OS.flush();
+    std::string PartStr;
+    raw_string_ostream OS1(PartStr);
+    PSVSE.Print(OS1, PSVSE.GetSemanticName(), PSVSemanticIndexVec.data());
```
pow2clk, line 427: I see the introduction of this deconstructed print was added so that there could be just one error printing condition regardless of whether it was prompted by a difference in semantic indexes, semantic names, or something else. I think that's a lot of churn to prevent having custom fail cases above for the semantic index and name failure cases, dumping the entire Signature Element would be pretty confusing while we have more specific information.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+    OS1.flush();
+    EmitMismatchError(Name + "Element", PartStr, ModuleStr);
```
pow2clk, line 429: This is an even bigger multiline mismatch even when we have specific mismatch information above. The memcmp is valid, but even in that case, it would be cleaner if we found the specific mismatch and output it.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+  VerifyResourceTable(DM.GetUAVs(), ResIndex, PSVVersion);
+}
+
+void PSVContentVerifier::VerifyEntryProperties(const ShaderModel *SM,
+                                               PSVRuntimeInfo0 *PSV0,
+                                               PSVRuntimeInfo1 *PSV1,
+                                               PSVRuntimeInfo2 *PSV2) {
+  PSVRuntimeInfo3 DMPSV;
+  memset(&DMPSV, 0, sizeof(PSVRuntimeInfo3));
+  hlsl::InitPSVRuntimeInfo(&DMPSV, &DMPSV, &DMPSV, &DMPSV, DM);
```
pow2clk, line 490: This is a weird usage of a weird function whose previous use has become so mismatched as to be useless. 

First, this function does nothing with the pInfo3 parameter at all. Second, its existing usage in `DxilPSVWriter::Write` makes only one usage of any of the out params and that is to assert that `pInfo1` is not null, which this function doesn't affect. 

I suggest pulling out the old usage entirely and changing this function to your needs where it takes a single pInfo parameter and initializes all the things that are added after below

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+    hlsl::PrintPSVRuntimeInfo(OS, &DMPSV, &DMPSV, &DMPSV, &DMPSV,
+                              static_cast<uint8_t>(SM->GetKind()),
+                              DM.GetEntryFunctionName().c_str(), "");
+    OS.flush();
+    std::string Str1;
+    raw_string_ostream OS1(Str1);
+    PSV.PrintPSVRuntimeInfo(OS1, static_cast<uint8_t>(PSVShaderKind::Library),
+                            "");
+    OS1.flush();
+    EmitMismatchError("PSVRuntimeInfo", Str, Str1);
```
pow2clk, line 528: As before, it's a lot of spam to find what could be a single differing byte.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -73,6 +78,522 @@ struct DiagRestore {
...
+    }
+    if (PSV1->UsesViewID != DM.m_ShaderFlags.GetViewID())
+      EmitMismatchError("UsesViewID", std::to_string(PSV1->UsesViewID),
+                        std::to_string(DM.m_ShaderFlags.GetViewID()));
+
+    VerifySignatures(ValMajor, ValMinor);
+
+    VerifyViewIDDependence(PSV1);
+  }
+  if (PSVVersion > 1) {
```
pow2clk, line 583: The comment is fine. It will be optimized out of course, but I still think the conditional should be removed.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4523,3 +4548,1492 @@ TEST_F(ValidationTest, CacheInitWithLowPrec) {
...
+       "  Inputs[7] influencing Outputs[0] :  None",
+       "  Inputs[8] influencing Outputs[0] : 4 ",
+       "  Inputs[9] influencing Outputs[0] : 5 ",
+       "  Inputs[10] influencing Outputs[0] :  None",
+       "  Inputs[11] influencing Outputs[0] :  None",
+       "')",
+       "error: Container part 'Pipeline State Validation' does not match "
+       "expected for module.",
+       "Validation failed."},
+      /*maySucceedAnyway*/ false, /*bRegex*/ false);
```
pow2clk, line 5323: Yeah, that's a long error message

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: SemanticKind was initialized in InitPSVSignatureElement.
Only SemanticName and SemanticSemanticIndexes are copied because the data in PSVSignatureElement0 are just index.
PSVStringTableReorder will cover the semantic name path.
SemanticKind is not tested mismatch directly. But not mismatch was tested.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: These situations are not expected to happen.
If the situation happened, the loader of PSV or builder of ViewIDState should catch it. And we'll go the invalid path.
I'll add assert.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: Split into 3 functions.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: Removed.

lib/DxilValidation/DxilContainerValidation.cpp
python3kgae: For the error message, my feeling is more information is better because the input of validation is a container which is not easy to read. Having only a single piece of the message makes it difficult to understand the full scope of the problem.

pow2clk: I don't agree, but I don't think it's worth holding this up. Hopefully these errors will be seen infrequently if at all.

pow2clk (approved):
I appreciate the new code consolidation. I was glad to see how well `LoadViewIDStateFromPSV` subbed in for the SimpleViewID class. I still think the error messages are unwieldy, but I'm willing to defer to you there.

tex3d (approved):
One potential missing validation:
Where does it make sure the PSV part size is the expected size for the data for that PSV version based on the validator version?  Otherwise it could contain extra data tacked on to the end which would get ignored by the validator.

python3kgae:
> One potential missing validation: Where does it make sure the PSV part size is the expected size for the data for that PSV version based on the validator version? Otherwise it could contain extra data tacked on to the end which would get ignored by the validator.

Created #6924 to check PSV part size.

PR #6865, opened 2024-08-16
[SPIR-V] add -fvk-support-nonzero-base-vertex option

artamonovoleg:
There is already a workaround for the SV_BaseInstance inconsistency between HLSL and SPIR-V. The same issue applies to SV_BaseVertex, but no such option is available. This commit adds the -fvk-support-nonzero-base-vertex option, which behaves the same as the -fvk-support-nonzero-base-instance option.

### DIFF
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -3392,8 +3392,9 @@ SpirvVariable *DeclResultIdMapper::getInstanceIdFromIndexAndBase(
   return instanceIdVar;
 }
 
-SpirvVariable *DeclResultIdMapper::getVertexIdFromIndexAndBase(
-    SpirvVariable *vertexIndexVar, SpirvVariable *baseVertexVar) {
+SpirvVariable *
+DeclResultIdMapper::getVertexIdFromIndexAndBase(SpirvVariable *vertexIndexVar,
+                                                SpirvVariable *baseVertexVar) {
   QualType type = vertexIndexVar->getAstResultType();
   auto *vertexIdVar = spvBuilder.addFnVar(
       type, vertexIndexVar->getSourceLocation(), "SV_VertexID");
--- a/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
+++ b/tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
@@ -3417,7 +3417,7 @@ SpirvVariable *DeclResultIdMapper::getBaseInstanceVariable(
   auto *baseInstanceVar = spvBuilder.addStageBuiltinVar(
       type, spv::StorageClass::Input, spv::BuiltIn::BaseInstance, false,
       semantic->loc);
-  StageVar var(sigPoint, *semantic, nullptr, type,
+  StageVar var(sigPoint, {}, nullptr, type,
                getLocationAndComponentCount(astContext, type));
   var.setSpirvInstr(baseInstanceVar);
   var.setIsSpirvBuiltin();
@@ -3432,7 +3432,7 @@ SpirvVariable *DeclResultIdMapper::getBaseVertexVariable(
   auto *baseVertexVar = spvBuilder.addStageBuiltinVar(
       type, spv::StorageClass::Input, spv::BuiltIn::BaseVertex, false,
       semantic->loc);
-  StageVar var(sigPoint, *semantic, nullptr, type,
+  StageVar var(sigPoint, {}, nullptr, type,
                getLocationAndComponentCount(astContext, type));
   var.setSpirvInstr(baseVertexVar);
   var.setIsSpirvBuiltin();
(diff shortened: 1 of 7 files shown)


artamonovoleg:

tools/clang/lib/SPIRV/DeclResultIdMapper.h
```diff
@@ -791,6 +791,19 @@ class DeclResultIdMapper {
   SpirvVariable *getInstanceIdFromIndexAndBase(SpirvVariable *instanceIndexVar,
                                                SpirvVariable *baseInstanceVar);
 
+  // Creates a function scope variable to represent the "SV_VertexID"
+  // semantic, which it not immediately available in SPIR-V. Its value will be
```
Keenuts, line 795:
```suggestion
  // semantic, which is not immediately available in SPIR-V. Its value will be
```

tools/clang/lib/SPIRV/DeclResultIdMapper.h
```diff
@@ -791,6 +791,19 @@ class DeclResultIdMapper {
...
+  // set by subtracting the values of the given InstanceIndex and base instance
+  // variables.
+  //
+  // vertexIndexVar: The SPIR-V input variable that decorated with
```
Keenuts, line 799:
```suggestion
  // vertexIndexVar: The SPIR-V input variable decorated with
```

tools/clang/lib/SPIRV/DeclResultIdMapper.h
```diff
@@ -791,6 +791,19 @@ class DeclResultIdMapper {
...
+  // vertexIndex.
+  //
+  // baseVertexVar: The SPIR-V input variable that is decorated with
```
Keenuts, line 802:
```suggestion
  // baseVertexVar: The SPIR-V input variable decorated with
```

tools/clang/lib/SPIRV/DeclResultIdMapper.cpp
```diff
@@ -3407,6 +3425,21 @@ SpirvVariable *DeclResultIdMapper::getBaseInstanceVariable(
...
 }
 
+SpirvVariable *DeclResultIdMapper::getBaseVertexVariable(
+    SemanticInfo *semantic, const hlsl::SigPoint *sigPoint, QualType type) {
+  assert(type->isSpecificBuiltinType(BuiltinType::Kind::Int) ||
+         type->isSpecificBuiltinType(BuiltinType::Kind::UInt));
+  auto *baseVertexVar = spvBuilder.addStageBuiltinVar(
+      type, spv::StorageClass::Input, spv::BuiltIn::BaseVertex, false,
+      semantic->loc);
+  StageVar var(sigPoint, *semantic, nullptr, type,
```
Keenuts, line 3435:
```suggestion
  StageVar var(sigPoint, /* semaInfo= */, nullptr, type,
```

I wouldn't put the wrong semantic just to fill this.
Same for the location above, I'd probably use `{}` since this doesn't comes directly from a line of code.

Keenuts (changes requested):
Hello!
Thanks for the contribution!
Looks OK, just the part where you fake the semantic can be removed.
Also, you'll need to add some tests for this.
You can take a look at `tools/clang/test/CodeGenSPIRV/semantic.unsigned-nonzero-base-instance.vs.hlsl` and `tools/clang/test/CodeGenSPIRV/semantic.nonzero-base-instance.vs.hlsl` for inspiration.

artamonovoleg:
Hello! 
Resolved

tools/clang/lib/SPIRV/DeclResultIdMapper.h
```diff
@@ -810,6 +823,25 @@ class DeclResultIdMapper {
...
+  // variable is also added to the list of stage variable `this->stageVars`. Its
+  // type will be a 32-bit integer.
+  //
+  // The semantic is a lie. We currently give it the semantic for the
+  // VertexID. I'm not sure what would happen if we did not use a semantic, or
+  // tried to generate the correct one. I'm guessing there would be some issue
+  // with reflection.
+  //
+  // semantic: the semantic to attach to this variable
+  //
```
Keenuts, line 836: I don't think you need to do that. I'd remove the semantic parameter, and use `{}` as location in `getBaseVertexVariable`.
I know this comment was just pulled from the function above, but seems like it was added for no real reasons except a 'what if'.
Unless somebody has an example showing this is required, I'd suggest to:
 - removing this comment on both functions
 - removing the `semantic` parameter on both functions
 - replace `*semantic` with `{}`
 - replace `semantic->loc` with `{}`
 
 (Removing this doesn't break tests, nor makes reflection fail on what I've tried so I'd assume it's safe)

Keenuts (commented):
Thanks for those! LGTM, just one last bit

artamonovoleg:
@Keenuts, done

Keenuts (approved):
Thanks, LGTM!

PR #6866, opened 2024-08-16
Move dxil.dll to DXC repo.

### LINKED ISSUE #6769: DXIL Validator Hash

python3kgae:
Build dxil.dll in DXC repo.
It will shares same validator code with dxcompiler.dll.

Fixes #6769


tools/clang/tools/dxcvalidator/CMakeLists.txt
```diff
@@ -8,7 +8,9 @@ set( LLVM_LINK_COMPONENTS
   DXIL
   DxilContainer
   DxilHash
+  DxilRootSignature
   DxilValidation
+  HLSL       # for AreDxilResourcesDense and DxilSignatureAllocator which are used in DxilValidation
```
pow2clk, line 13: Given that there are no code changes in this directory, I don't know why any changes would be needed here.

tools/clang/tools/dxildll/dxccontainerbuilder.cpp
```diff
@@ -0,0 +1,24 @@
...
+  // Call dxil.dll's containerbuilder
+  *V = nullptr;
+  CComPtr<DxcContainerBuilder> Result(
+      DxcContainerBuilder::Alloc(DxcGetThreadMallocNoRef()));
+  if (nullptr == Result.p)
+    return E_OUTOFMEMORY;
+
+  Result->Init();
+  return Result->QueryInterface(RRID, V);
+}
```
pow2clk, line 24: I'll pass along the [feedback ](https://github.com/microsoft/DirectXShaderCompiler/pull/5952#discussion_r1382231012)I got from #5952 and suggest that this file be moved into dxildll.cpp

tools/clang/tools/dxildll/dxcvalidator.cpp
```diff
@@ -0,0 +1,105 @@
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// dxcvalidatorp.cpp                                                         //
```
pow2clk, line 3: nit: not what this file is called

tools/clang/tools/dxildll/dxcvalidator.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "dxc/Support/WinIncludes.h"
+
+#include "dxc/Support/Global.h"
+#include "dxc/Support/microcom.h"
+#include "dxc/dxcapi.h"
+
+using namespace llvm;
+using namespace hlsl;
+
+class DxcValidator : public IDxcValidator2, public IDxcVersionInfo {
```
pow2clk, line 23: I'm trying to think of a reason why we might want to enable querying the git commit info from this and coming up with nothing, but if you can think of something, maybe we should inherit from `IDxcVersionInfo2`

tools/clang/tools/dxildll/dxildll.cpp
```diff
@@ -0,0 +1,143 @@
...
+  HRESULT HR = (DxcInitThreadMalloc());
+  if (DXC_FAILED(HR))
+    goto Cleanup;
+  DxcSetThreadMallocToDefault();
+  MemSetup = true;
+  if (::llvm::sys::fs::SetupPerThreadFileSystem()) {
+    HR = E_FAIL;
+    goto Cleanup;
+  }
+Cleanup:
```
pow2clk, line 48: nit: I know it's not your code originally, but this seems a flimsy excuse to use a goto as one of them just falls through anyway and the other could very easily be a conditional block
```suggestion
  if (!DXC_FAILED(HR)) {
    DxcSetThreadMallocToDefault();
    MemSetup = true;
    if (::llvm::sys::fs::SetupPerThreadFileSystem())
      HR = E_FAIL;
  }
```

pow2clk (commented):
If not for the dxcvalidator/CMakeList.txt change I don't understand, I'd approve this. The other suggestions you can take or leave.

tools/clang/tools/dxildll/dxildll.cpp
python3kgae: Updated.

tools/clang/tools/dxildll/dxcvalidator.cpp
python3kgae: Unable to find a reason as well.

tools/clang/tools/dxildll/dxcvalidator.cpp
python3kgae: Updated.

tools/clang/tools/dxildll/dxccontainerbuilder.cpp
python3kgae: Updated.

tools/clang/tools/dxcvalidator/CMakeLists.txt
python3kgae: Moved to CMakeLists.txt for dxildll.

python3kgae: But gcc complains after remove change here.

pow2clk: Okay. I suspect I know what's going on here. I think that these dependencies were always there, but they were satisfied indirectly by other libraries by chance. The linux linker is really particular that way.

pow2clk:
GCC is still failing. Is the move of the container builder into dxildll.cpp the cause? I remember something like that with my similar change.

tools/clang/test/CMakeLists.txt
```diff
@@ -77,7 +77,11 @@ endif()
 set(CLANG_TEST_DEPS dxc dxa dxopt dxl dxv dxr dxcompiler clang-tblgen llvm-config opt FileCheck count not ClangUnitTests)
 if (WIN32)
 list(APPEND CLANG_TEST_DEPS
-     dxc_batch ExecHLSLTests
+     dxc_batch ExecHLSLTests dxildll
+    )
+elseif(NOT ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
```
llvm-beanz, line 82: I'm confused by this. Why is this needed?

python3kgae: Shall we enable libdxil to mac?
We only release dxildll for linux and windows now.

llvm-beanz: I think one of the biggest benefits of us open sourcing this is to allow people to build it on macOS where we don't currently have supported binary releases.

tools/clang/tools/dxcvalidator/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ set( LLVM_LINK_COMPONENTS
   DxilContainer
   DxilHash
   DxilValidation
+  DxilRootSignature
+  HLSL       # for gcc link error on AreDxilResourcesDense and DxilSignatureAllocator which are used in DxilValidation
```
tex3d, line 13: If these are the only dependencies on `HLSL.lib`, I think we could avoid this dependency with two things:
- `AreDxilResourcesDense`: this is only used to verify that resource IDs in metadata lists are sequential, starting at 0.  This could be locally checked in a simpler way for validation purposes.
- `DxilSignatureAllocator`: this is already implemented in a `.inl` file to make it easy to separate this dependency.

tools/clang/tools/dxildll/dxcvalidator.cpp
tex3d: The DXIL major/minor version info is not sufficient to identify different validator versions within the same release cycle.  Since this change introduces a build of `dxil.dll` into every build of the public DXC project, it seems more important than before to retain the additional version information going forward.

What do you think?

damyanp: > it seems more important than before to retain the additional version information going forward.

Totally agree with this.  I hope the answer isn't to use a commit hash though.  Ideally we'd get some kind minor-minor version number we can increment as appropriate? Maybe this is something we'll need to consider independently of this though

tools/clang/tools/dxcvalidator/CMakeLists.txt
python3kgae: AreDxilResourceDense should be easy to move to validation.
Where should we put DxilSignatureAllocator?
It is used in Validation and DxilPackSignatureElement.

tex3d: One option for `DxilSignatureAllocator.inl` would be to include it in a `.cpp` file under `dxildll` to provide an implementation when not linking in HLSL.lib.  Another solution might be to move the DxilPackSignatureElement stuff into the LLVMDXIL lib, or if we don't want to add it there, we could put it into a new lib instead.

python3kgae: Created issue https://github.com/microsoft/DirectXShaderCompiler/issues/6898 to track this.

pow2clk (approved):
Thanks!

tools/clang/tools/dxildll/dxcvalidator.cpp
python3kgae: We already have version number in dxc --version like
`dxcompiler.dll: 1.8 - 1.8.0.4710 (test_gcc_bld, 913ae2ce8); dxil.dll: 1.8(1.8.0.4710)`.

PR #6869, opened 2024-08-20
[NFC] add utils for PSV.

python3kgae:
Add init functions to help build PSVSignatureElement0/PSVRuntimeInfo/PSVResourceInfo.

Add print function for DxilPipelineStateValidation.

Add new option -dumppsv to dxa for testing.

For #6817.


include/dxc/DxilContainer/DxilPipelineStateValidation.h
```diff
@@ -16,6 +16,12 @@
 #include <cstring>
 #include <stdint.h>
 
+#include <assert.h>
```
damyanp, line 19: nit: why the line break between this and the other includes?

include/dxc/DxilContainer/DxilPipelineStateValidation.h
```diff
@@ -266,7 +274,8 @@ struct PSVComponentMask {
     if (ComponentIndex < NumVectors * 4)
       Mask[ComponentIndex >> 5] &= ~(1 << (ComponentIndex & 0x1F));
   }
-  bool IsValid() { return Mask != nullptr; }
+  bool IsValid() const { return Mask != nullptr; }
+  void print(llvm::raw_ostream &, const char *, const char *) const;
```
damyanp, line 278: Looks like there are two different naming conventions being used here.  It'd be good to try and figure out what one if most prevalent in this file and follow that one?  Based on a quick look, it seems like that would make this `Print`.

This applies to all the new methods in this PR, I won't call them all out, although if we agree we should follow this convention and you let me know you think you've updated them all I could do that if you like.

damyanp (approved):
It looks like quite a lot of code has moved around...I couldn't quite figure out if that's actually required to add the printing functionality, or it is was just a cleanup.  I haven't done a careful diff of the moved code, so I'm going to assume that there truly is NFC here.  I'm guessing that without the PSV dumping functionality added here that there's minimal existing test coverage of this.

include/dxc/DxilContainer/DxilPipelineStateValidation.h
python3kgae: Updated.

include/dxc/DxilContainer/DxilPipelineStateValidation.h
python3kgae: Fixed.

bob80905 (approved):
Nit: Several of the files don't end with a newline

PR #6872, opened 2024-08-21
Add license and readme for HLSL header library

llvm-beanz:
As we're starting to grow some new language features that are more convienent to write in headers, having a set of headers that are distributable without attribution requirement is a big win. These headers will all be contributed under the Apache 2.0 with LLVM exception license to align with upstream LLVM and provide apporpriate protections.

### DIFF
--- a/tools/clang/lib/Headers/hlsl/README.txt
+++ b/tools/clang/lib/Headers/hlsl/README.txt
@@ -1,0 +1,7 @@
+HLSL Standard Header Library
+============================
+
+The contents of this directory and subdirectory are the HLSL Standard Header
+library. These headers are open source software. You may freely distribute all
+or parts of these headers under the terms of the license agreement found in
+LICENSE.txt.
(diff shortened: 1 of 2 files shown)


damyanp (approved):

tools/clang/lib/Headers/hlsl/README.txt
```diff
@@ -0,0 +1,7 @@
+HLSL Standard Header Library
+============================
+
+The contents of this directory and subdirectory are the HLSL Standard Header
```
coopp, line 4: Is there only 1 subdirectory we care about?  Should this be 'The contents of this directory and _subdirectories_ are the HLSL Standard ....' ??

coopp (approved):
Looks good to me.  Just a little nit on subdirectory vs subdirectories in the final paragraph.

tools/clang/lib/Headers/hlsl/README.txt
llvm-beanz, line 4:
```suggestion
The contents of this directory and subdirectories are the HLSL Standard Header
```

s-perron (approved):

Issue #6578, opened 2024-05-02
[Feature Request] [SPIR-V] [Proposal 0011] `vk::addrof(v)` to obtain an `OpTypePointer` to a variable and deprecate `[vk::ext_reference]`


devshgraphicsprogramming:
**Is your feature request related to a problem? Please describe.**

When writing inline SPIR-V intrinsics in HLSL one needs to choose between making an input argument `T`, `[vk::ext_reference] T` or `vk::SpirvType`.

Its not well documented what `ext_reference` actually turns into but from my experience it seems to be and `OpTypePointer` to the actual argument invoked.

This makes it problematic to write an intrinsic signature thats supposed to work both with `T&` and arbitrary `vk::SpirvType`, for example `OpAtomicIAdd` .

If you write
```cpp
template<typename T>
[[vk::ext_instruction]]
T atomicIAdd([[vk::ext_reference]] T ptr, ...memory semantics... , T value);
```
then it will work on RWStructuredBuffer variables, but won't work on user-defined BDA pointers (or any pointers).

But if you write
```cpp
template<typename T, typename P>
[[vk::ext_instruction]]
T atomicIAdd(P ptr, ...memory semantics... , T value);
```
then it will only work on `vk::SpirvType` defined pointers.

Which wouldn't be too bad, except that you can't get a pointer/access-chain to any variable you like.

**Describe the solution you'd like**

A way to obtain an `OpTypePointer` for a given variable (basically a hidden `&`/`addrof`)

Also a way to obtain an `OpType` result id, because some instrinsics like `OpDecorate` or `OpMemberDecorate` need to be fed the type's `%result_id` as an argument, its unclear to me how to achieve that with the current mapping of.
```cpp
[[vk::ext_instruction(%opcode%)]]
%result_type% intrinsicHLSLName(arguments...);
```

**Describe alternatives you've considered**

Just generating an overload, but that makes the compiler confused with ambiguity.

**Additional context**

Trying to generate all Types and Operations from SPIR-V Headers.

Also implement Prop 0010.

devshgraphicsprogramming:
Here's what I tried in the meantime to return me a `vk::SpirvType<OpTypePointer,StorageClass,T>`

#### OpAccessChain with no Indices

- cannot be used for types that aren't composite
- fails with an ICE regarding a cast<X> https://godbolt.org/z/M3ra87ro8

#### OpBitcast with a `[[vk::ext_reference]]` input casting to identical type

[...]

PR #6873, opened 2024-08-21
Add specific handling for inline spirv pointer types

s-perron:
Add specific handling for inline spirv pointer types
    
We currently blindly create a new type for all vk::SpirvType* types.
This can cause problems when the type is suppose to match another type.
In this case, we want a spirv pointer type to match the pointer type of the
variable. The OpTypePointer for the variable is implicitly created when
declaring the variable, but, if we want to explicitly declare the
pointer as a SpirvType, we get two different OpTypePointer instructions
in the module. So technically the types do not match.
    
To fix this, I add special handling in the SPIR-V backend to be able to
merge the implicit pointer type created and the SpirvType when they
match.

This implements the SPIR-V Pointers in [HLSL spec proposal 0021](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0021-vk-coop-matrix.md#spir-v-pointers)

### DIFF
--- a/tools/clang/lib/Headers/hlsl/vk/spirv.h
+++ b/tools/clang/lib/Headers/hlsl/vk/spirv.h
@@ -1,0 +1,34 @@
+// Copyright (c) 2024 Google LLC
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef _HLSL_VK_SPIRV_H_
+#define _HLSL_VK_SPIRV_H_
+
+namespace vk {
+
+enum StorageClass {
+  StorageClassWorkgroup = 4,
+};
+
+// An opaque type to represent a Spir-V pointer to the workgroup storage class.
+// clang-format off
+template <typename PointeeType>
+using WorkgroupSpirvPointer = const vk::SpirvOpaqueType<
+    /* OpTypePointer */ 32,
+    vk::Literal<vk::integral_constant<uint, StorageClassWorkgroup> >,
+    PointeeType>;
+// clang-format on
+
+// Returns an opaque Spir-V pointer to v. The memory object v's storage class
+// modifier must be groupshared. If the incorrect storage class is used, then
+// there will be a validation error, and it will not show the correct
+template <typename T>
+[[vk::ext_instruction(/* OpCopyObject */ 83)]] WorkgroupSpirvPointer<T>
+GetGroupSharedAddress([[vk::ext_reference]] T v);
+
+} // namespace vk
+
+#endif // _HLSL_VK_SPIRV_H_
--- a/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
+++ b/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
@@ -732,6 +732,15 @@ const SpirvType *LowerTypeVisitor::lowerInlineSpirvType(
 
   auto args = specDecl->getTemplateArgs()[operandsIndex].getPackAsArray();
 
+  if (operandsIndex == 1 && args.size() == 2 &&
+      static_cast<spv::Op>(opcode) == spv::Op::OpTypePointer) {
+    const SpirvType *result =
+        getSpirvPointerFromInlineSpirvType(args, rule, isRowMajor, srcLoc);
+    if (result) {
+      return result;
+    }
+  }
+
   for (TemplateArgument arg : args) {
     switch (arg.getKind()) {
     case TemplateArgument::ArgKind::Type: {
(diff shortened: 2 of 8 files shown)


tools/clang/lib/Headers/hlsl/vk/spirv.h
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2024 Google LLC
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
s-perron, line 5: @llvm-beanz Let me know if this works with your change in https://github.com/microsoft/DirectXShaderCompiler/pull/6872

tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
```diff
@@ -1363,5 +1372,41 @@ LowerTypeVisitor::populateLayoutInformation(
   return result;
 }
 
+const SpirvType *LowerTypeVisitor::getSpirvPointerFromInlineSpirvType(
+    ArrayRef<TemplateArgument> args, SpirvLayoutRule rule,
+    Optional<bool> isRowMajor, SourceLocation location) {
+
+  assert(args.size() == 2 && "OpTypePoint requires exactly 2 arguments.");
```
Keenuts, line 1379:
```suggestion
  assert(args.size() == 2 && "OpTypePointer requires exactly 2 arguments.");
```

Keenuts (commented):
Are they places where we want to generate 2 distinct types for the same thing in DXC?
If not, why isn't this the default path for pointer type creation?

devshgraphicsprogramming:
Does this fix any of the #6541 , https://github.com/microsoft/DirectXShaderCompiler/issues/6578 or related issues with Access Chains crashing the compiler with a disallowed cast<X> ?

tools/clang/lib/Headers/hlsl/vk/spirv.h
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef _HLSL_VK_SPIRV_H_
+#define _HLSL_VK_SPIRV_H_
+
+namespace vk {
+
+enum StorageClass {
+  StorageClassWorkgroup = 4,
+};
```
devshgraphicsprogramming, line 14: why not pull the entire non-11 C++ spirv-headers like we do?

s-perron:
You should be able to do pointers for storage classes that do not require a layout. So function, private, and workgroup.

s-perron:
Some other issues might be small bugs that still need to be fixed separately from this one.

tools/clang/test/CodeGenSPIRV/workgroupspirvpointer.varpointer.hlsl
```diff
@@ -0,0 +1,23 @@
...
+
+[[vk::ext_instruction(/* OpLoad */ 61)]] int
+Load(vk::WorkgroupSpirvPointer<int> p);
+
+[[noinline]]
+int foo(vk::WorkgroupSpirvPointer<int> param) {
+  return Load(param);
+}
+
+[[vk::ext_capability(/* VariablePointersCapability */ 4442)]]
```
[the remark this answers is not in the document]
s-perron: In this case it is needed because the function will not be inlined. That makes it necessary.

devshgraphicsprogramming:
> Some other issues might be small bugs that still need to be fixed separately from this one. If you have problems let me know. Do we still have an issue opened for the access chain? I may have lost that issue.

AFAIK you just closed it, it was wrapped up with #6541

devshgraphicsprogramming:
> You should be able to do pointers for storage classes that do not require a layout. So function, private, and workgroup.

Later in https://github.com/microsoft/DirectXShaderCompiler/issues/6578 I did some tests with attempting to use `function` and `private` pointers, they won't work.

Above optimization level O0 (so O1 and above) stuff gets aggressively inlined so the variable never exists in the first place to take a pointer of.

P.S. When one uses `-fvk-use-scalar-layout` DXC's SPIR-V codegen makes all SSBO and UBOs use a scalar layout, so making a pointer with a Storage Class of PhysicalStorageBuffer, StorageBuffer and Uniform shouldn't need any deduction.

tools/clang/lib/Headers/hlsl/vk/spirv.h
Keenuts: This is something we are working on for those headers in general:
 - how to distribute them, and where shall they live
 - shall we add a new flavor of spirv headers files like we have for c++11 and non-11
We do plan on bringing the whole spirv-header file, this is more a temporary measure until we figure out the whole thing.

s-perron: This is the start of the standard headers, and I am working on adding cooperative matrices as an interesting proof of concept. I am doing the minimum I need to in order to do that. In the near future, we will decide how we want to handle the core functionality going forward.

s-perron:
I have opened https://github.com/microsoft/DirectXShaderCompiler/issues/6891 to follow up on the access chain issue. There was already an existing issues for the decoration attribute on fields.

s-perron:
> Are they places where we want to generate 2 distinct types for the same thing in DXC? 

For pointers, I don't think we ever want to generate two distinct types. We might want to generate different types for struts that are the "same" if the user gave them different names:

```
struct A { int i; };
struct B { int i; };
```

> If not, why isn't this the default path for pointer type creation?

I believe it is. The code that creates pointers calls `SpirvContext::getPointerType`. We cannot use the exact same function because how you extract the storage class and pointee type is different. That is different for every place that requires a pointer type. See below for one example:

https://github.com/microsoft/DirectXShaderCompiler/blob/bb3a0c3641f699fc42ac7541b1aa5ecae306c101/tools/clang/lib/SPIRV/LowerTypeVisitor.cpp#L225-L231

Keenuts (approved):

Keenuts (changes requested):
I need to check something.

Keenuts (approved):

tools/clang/lib/SPIRV/LowerTypeVisitor.cpp
```diff
@@ -732,6 +732,15 @@ const SpirvType *LowerTypeVisitor::lowerInlineSpirvType(
 
   auto args = specDecl->getTemplateArgs()[operandsIndex].getPackAsArray();
 
+  if (operandsIndex == 1 && args.size() == 2 &&
+      static_cast<spv::Op>(opcode) == spv::Op::OpTypePointer) {
+    const SpirvType *result =
+        getSpirvPointerFromInlineSpirvType(args, rule, isRowMajor, srcLoc);
```
cassiebeckley, line 738: Could you maybe add a short explanation of why we need a special case to handle pointers in a comment?

cassiebeckley (approved):
LGTM

s-perron:
> am I not understanding something or was this PR supposed to fix issues like needing the `-Vd` flag in https://godbolt.org/z/3vq3xWY7v ?

It should fix the issue with using access chains.

It only for storage classes that do not require a layout (Function, Private, workgroup). I don't think we will be able to fix the layout problem in DXC.

PR #6874, opened 2024-08-21
Add SPIRV version macros to DXC

s-perron:
With inline spir-v, it becomes important for users to know which version
of SPIR-V is being targeted. They may need to generate different code
depending on the version.

This commit add these `__SPIRV_MAJOR_VERSION__` and
`__SPIRV_MINOR_VERSION__` to the compiler.

### DIFF
--- a/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
+++ b/tools/clang/tools/dxcompiler/dxcompilerobj.cpp
@@ -63,6 +63,7 @@
 // SPIRV change starts
 #ifdef ENABLE_SPIRV_CODEGEN
 #include "clang/SPIRV/EmitSpirvAction.h"
+#include "clang/SPIRV/FeatureManager.h"
 #endif
 // SPIRV change ends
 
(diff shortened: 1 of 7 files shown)


cassiebeckley (approved):

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
```diff
@@ -1451,6 +1452,22 @@ class DxcCompiler : public IDxcCompiler3,
 // SPIRV change starts
 #ifdef ENABLE_SPIRV_CODEGEN
     compiler.getLangOpts().SPIRV = Opts.GenSPIRV;
+    llvm::Optional<spv_target_env> spirvTargetEnv =
+        spirv::FeatureManager::stringToSpvEnvironment(
+            Opts.SpirvOptions.targetEnv);
+
+    // If we do not have a valid target environment, the error will be handled
+    // later.
```
Keenuts, line 1460: I'm assuming there is already a test to cover this?

Keenuts (approved):

tools/clang/tools/dxcompiler/dxcompilerobj.cpp
s-perron: Yes there is.

PR #6876, opened 2024-08-22
[Build] Update to clang-18 for linux clang pipeline

python3kgae:
This is to work around LeakSanitizer issue 'LeakSanitizer has encountered a fatal error.'

For #6769

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -99,7 +99,7 @@ stages:
         versionSpec: '3.x'
 
     - bash: |
-        sudo apt-get install ninja-build
+        sudo apt-get install ninja-build libstdc++-12-dev
         wget https://apt.llvm.org/llvm.sh
         chmod u+x llvm.sh
         sudo ./llvm.sh 18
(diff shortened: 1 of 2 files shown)


pow2clk (approved):
Looks good. Let's try to remember to revisit this when the ubuntu images get updated and we can probably get this through apt at least.

python3kgae:
> Looks good. Let's try to remember to revisit this when the ubuntu images get updated and we can probably get this through apt at least.

The best I can get from apt is 15 now.
Not sure how long it will take to get 18.

damyanp (approved):
Is this a "workaround" or a "fix"?  ie does clang-18 actually fix something, or has this just rearranged things so that the bug no longer shows up?

python3kgae:
> Is this a "workaround" or a "fix"? ie does clang-18 actually fix something, or has this just rearranged things so that the bug no longer shows up?

./llvm.sh 18 will install clang18 and libclang-rt.
The libclang-rt included LeakSanitizer.
It is hard to say its clang18 fixed the issue or updated LeakSanitizer fixed the issue, or it is just rearranged things, so the bug no longer shows up.

What we know is that the failed tests didn't leak memory, and update to clang18 (which will update LeakSanitizer the same time) make the test not fail anymore. Without clang18 update, it is reported as pass in gtest but LeakSanitizer crash make it fail.

PR #6878, opened 2024-08-23
Support Mesh shader in GetInputToPCOutputTable.

python3kgae:
Mesh shader will have input to primitive output dependency table.
GetInputToPCOutputTable should return m_pInputToPCOutputTable instead of empty table for mesh shader.

Also fix the dump for mesh shader.

For #6817


lib/DxilContainer/DxilPipelineStateValidation.cpp
```diff
@@ -901,7 +901,7 @@ void DxilPipelineStateValidation::Print(raw_ostream &OS,
       Table.Print(OS, "Inputs", OutputSetName.c_str());
     }
 
-    if (IsHS()) {
+    if (IsHS() || IsMS()) {
```
tex3d, line 904: MS doesn't need or have this table.

PR #6879, opened 2024-08-23
Process subrecord decls in spirv backend

### LINKED ISSUE #5916: Static const members in nested structs don't compile to SPIR-V
**Description**
Accessing static const members in nested structs results in "found unregistered decl". This currently prevents us from implementing some compile-time performance optimizations via metaprogramming, since this issue persists if the nested struct is a template.

**Steps to Reproduce**
Reproducer can be found here: https://godbolt.org/z/brnGYKddj
```
struct A {
    static const uint value = 6u;
    struct B { static const uint value = 6u; };
};

[[vk::binding(0,4)]] RWStructuredBuffer<uint> a;

// Define some dummy entry point
[shader("raygeneration")]
void main() {
    a[0] = A::value;
    a[0] = A::B::value;
}
```
cannot be compiled with `dxc -T lib_6_4 -spirv -fspv-target-env=vulkan1.2 -enable-16bit-types -HV 2021`

Removing the line `a[0] = A::B::value;` results in successful compilation.

**Actual Behavior**
```
<source>:3:34: fatal error: found unregistered decl
    struct B { static const uint value = 6u; };
                                 ^

Compiler returned: 5
```

**Environment**
- DXC version: libdxcompiler.so: 1.7(dev;1-481090a8)
- Host Operating System: Compiler Explorer

s-perron:
The spir-v backend does not recurse into structs defined in struct in
doRecoredDecl. This causes us to miss some declarations.

Fixes #5916

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -1801,6 +1801,8 @@ void SpirvEmitter::doRecordDecl(const RecordDecl *recordDecl) {
         doVarDecl(varDecl);
     } else if (auto *enumDecl = dyn_cast<EnumDecl>(subDecl)) {
       doEnumDecl(enumDecl);
+    } else if (auto recordDecl = dyn_cast<RecordDecl>(subDecl)) {
+      doRecordDecl(recordDecl);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


Keenuts (approved):
Thanks! Surprised this wasn't caught before!

cassiebeckley (approved):

PR #6883, opened 2024-08-28
PIX: Rationalize UAV generation

jeffnn:
The impetus for these changes was unexplained crashes in a display driver while attempting to create a debug-instrumented shader for PIX. The heart of it is the new test in pixtest.cpp: use the compiler to generate a raw UAV, and then compare the generated DXIL with what PIX generates for the same purpose.
Some of the PIX passes need only one UAV for a module, but some need two or more. In the latter case, the previous code was a bit loose about what it was doing with respect to adding the UAV resource, and creating its handles for each interested function. Most of the actual changes herein are to do with that. 
Lastly, the PIX UAV is raw at the D3D API level, but the instrumentation had been doing non-raw writes. No driver seemed to care, but I've fixed it anyway.


damyanp (approved):

adam-yang (approved):
LGTM!

PR #106, opened 2017-02-24
Support Unsigned for shorthand vectors/matrices

youngkim93:
1. enable unsigned keyword for int vectors/matrices
   (for example, "unsigned int2" will be converted to  "vector<unsigned int, 2>")
2. disable unsigned keyword as a complete type specifier to be consistent with old compiler.

PR #6885, opened 2024-08-28
Allow unsigned as a full type

pow2clk:
#106 disabled the unsigned keyword as a complete type specifier among other changes. While it gave no rationale for that decision nor any tests, it was likely because FXC doesn't allow it. It's harmless and sure not to break any code. It also represents a discrepancy with the developing clang HLSL support that I'm not inclined to change there. 

This restores the ability to promote a variable with just the `unsigned` keyword as an `unsigned int` and includes tests that affirm it's availability in a number of contexts and compatibility with `uint`

### DIFF
--- a/tools/clang/include/clang/Sema/DeclSpec.h
+++ b/tools/clang/include/clang/Sema/DeclSpec.h
@@ -619,9 +619,9 @@ public:
   /// \brief Return true if any type-specifier has been found.
   bool hasTypeSpecifier() const {
       return getTypeSpecType() != DeclSpec::TST_unspecified ||
-            getTypeSpecWidth() != DeclSpec::TSW_unspecified ||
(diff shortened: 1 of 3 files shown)


pow2clk:
Oh that's a bit odd. The commit message contains less info than the description on github which says explicitly that it was to match FXC. So there's that.

python3kgae (approved):

llvm-beanz (commented):
I think there are a bunch of test cases that need to be written for this, particularly for sema cases.

Things like:
* `is_same<unsigned, uint>`
* `is_same<unsigned int, uint>`
* `is_same<unsigned int64_t, uint64_t>`

What about `long` or `short`? 

I had assumed we would actually remove the C keywords from Clang, so if we're not going to do that maybe we should discuss which ones we want to support in DXC.

PR #6886, opened 2024-08-28
[ViewID] build viewid mask for primitive output

python3kgae:
Currently, viewid mask for primitive output is always empty. This change will fix it by treat entry as PC entry and skip output when it is not what is checking.

For #6817

### DIFF
--- a/lib/HLSL/ComputeViewIdStateBuilder.cpp
+++ b/lib/HLSL/ComputeViewIdStateBuilder.cpp
@@ -141,7 +141,8 @@ private:
                                     llvm::CallGraphNode *pNode,
                                     FunctionSetType &FuncSet);
   void AnalyzeFunctions(EntryInfo &Entry);
-  void CollectValuesContributingToOutputs(EntryInfo &Entry, bool IsForPatchConstant);
+  void CollectValuesContributingToOutputs(EntryInfo &Entry,
+                                          bool IsForPatchConstant);
   void CollectValuesContributingToOutputRec(
       EntryInfo &Entry, llvm::Value *pContributingValue,
       InstructionSetType &ContributingInstructions);
--- a/lib/HLSL/ComputeViewIdStateBuilder.cpp
+++ b/lib/HLSL/ComputeViewIdStateBuilder.cpp
@@ -111,13 +111,11 @@ private:
     // Contributing instructions per output.
     std::unordered_map<unsigned, InstructionSetType>
         ContributingInstructions[kNumStreams];
-    bool IsForPatchConstantOrPrimitive;
-    EntryInfo(bool ForPC) : IsForPatchConstantOrPrimitive(ForPC) {}
     void Clear();
   };
 
-  EntryInfo m_Entry = EntryInfo{false};
-  EntryInfo m_PCEntry = EntryInfo{true};
+  EntryInfo m_Entry;
+  EntryInfo m_PCEntry;
 
   // Information per function.
   using FunctionReturnSet = std::unordered_set<llvm::ReturnInst *>;
@@ -143,7 +141,7 @@ private:
                                     llvm::CallGraphNode *pNode,
                                     FunctionSetType &FuncSet);
   void AnalyzeFunctions(EntryInfo &Entry);
-  void CollectValuesContributingToOutputs(EntryInfo &Entry);
+  void CollectValuesContributingToOutputs(EntryInfo &Entry, bool IsForPatchConstant);
   void CollectValuesContributingToOutputRec(
       EntryInfo &Entry, llvm::Value *pContributingValue,
       InstructionSetType &ContributingInstructions);
@@ -209,10 +207,11 @@ void DxilViewIdStateBuilder::Compute() {
   }
 
   // 4. Collect sets of values contributing to outputs.
-  CollectValuesContributingToOutputs(m_Entry);
-  if (m_PCEntry.pEntryFunc) {
-    CollectValuesContributingToOutputs(m_PCEntry);
-  }
+  CollectValuesContributingToOutputs(m_Entry,
+                                     /*IsForPatchConstantOrPrimitive*/ false);
+  if (m_PCEntry.pEntryFunc)
+    CollectValuesContributingToOutputs(m_PCEntry,
+                                       /*IsForPatchConstantOrPrimitive*/ true);
 
   // 5. Construct dependency sets.
   for (unsigned StreamId = 0; StreamId < (pSM->IsGS() ? kNumStreams : 1u);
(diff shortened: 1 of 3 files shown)


tex3d (approved):
This works, even if it feels slightly odd.

lib/HLSL/ComputeViewIdStateBuilder.cpp
```diff
@@ -111,12 +111,13 @@ class DxilViewIdStateBuilder {
     // Contributing instructions per output.
     std::unordered_map<unsigned, InstructionSetType>
         ContributingInstructions[kNumStreams];
-
+    bool IsForPatchConstantOrPrimitive;
```
damyanp, line 114: I don't understand why this needs to be a member of EntryInfo.  It's only checked in CollectValuesContributingToOutputs, and the caller of that function knows whether or not it is for the patch constant or primitive.

lib/HLSL/ComputeViewIdStateBuilder.cpp
```diff
@@ -481,17 +486,23 @@ void DxilViewIdStateBuilder::CollectValuesContributingToOutputs(
...
       pContributingValue = SPC.get_value();
       GetUnsignedVal(SPC.get_outputSigID(), &id);
       GetUnsignedVal(SPC.get_row(), (uint32_t *)&startRow);
       GetUnsignedVal(SPC.get_col(), &col);
+      IsForPatchConstantOrPrimitive = true;
     } else {
       IFT(DXC_E_GENERAL_INTERNAL_ERROR);
     }
 
+    // Skip vertex outputs for MS when check primitive outputs.
```
damyanp, line 502: I don't quite understand why these flags being different is the way to determine if we're checking for primitive outputs or not?

Is there not a more direct "if (CheckingPrimitiveOutputs)" type thing we could be saying here?

lib/HLSL/ComputeViewIdStateBuilder.cpp
python3kgae: Switched to send it to CollectValuesContributingToOutputs.

lib/HLSL/ComputeViewIdStateBuilder.cpp
python3kgae: Switched to check IsForPatchConstantOrPrimitive.

tools/clang/test/DXC/dumpPSV_MS.hlsl
```diff
@@ -189,7 +189,7 @@ void main(
       op.normal = mpl.normal;
       op.malnor = gsMem[tig / 3 + 1];
       op.alnorm = mpl.alnorm;
-      op.ormaln = mpl.ormaln;
+      op.ormaln = mpl.ormaln + vid;
```
damyanp, line 192: I see many changes inside conditionals above.  Are they really all covered by this one test change?

lib/HLSL/ComputeViewIdStateBuilder.cpp
```diff
@@ -454,35 +458,44 @@ void DxilViewIdStateBuilder::AnalyzeFunctions(EntryInfo &Entry) {
...
+        continue;
       pDxilSig = &m_pModule->GetPatchConstOrPrimSignature();
       pContributingValue = SPO.get_value();
       GetUnsignedVal(SPO.get_outputSigId(), &id);
       GetUnsignedVal(SPO.get_colIndex(), &col);
       GetUnsignedVal(SPO.get_rowIndex(), (uint32_t *)&startRow);
+      IsForPatchConstantOrPrimitive = true;
     } else if (DxilInst_StorePatchConstant SPC =
                    DxilInst_StorePatchConstant(CI)) {
+      if (!IsForPatchConstantOrPrimitive)
```
damyanp, line 497: I see that `IsForPatchConstantOrPrimitive` is set to true if it is a `DxilInst_StorePrimitiveOutput`.  Why doesn't it also get set to true for `DxilInst_StorePatchConstant`?

damyanp: Or alternatively, why does it get set?  It seems that it puts this function into a different mode. Does this mean that the order of `Entry.Outputs` is significant here?

python3kgae: We have 2 modes in CollectValuesContributingToOutputs.
And we'll call CollectValuesContributingToOutputs with different FuncInfo, one for check normal output and vertex output (outputs saved in OutputSignature), one for check patch constant output and primitive output (outputs saved in PatchConstOrPrimSignature).
The order doesn't important here, we just scan all the outputs collected in early pass.

Without this change, we put vertex output and primitive output in one table for mesh shader, which is wrong.
Patch constant doesn't have this issue because patch constant has its own function.

But mesh shader only has one entry, so when in vertex mode, we only scan vertex output, in primitive mode, we only scan primitive.

tools/clang/test/DXC/dumpPSV_MS.hlsl
python3kgae: This issue only affects mesh shader primitive output.
The result for other tests will be the same.

lib/HLSL/ComputeViewIdStateBuilder.cpp
damyanp: I understand that there are two modes. What I don't understand is why it changes mode when a `DxilInst_StorePrimitiveOutput` is encountered, on line 494. That very much makes it seem like the order of outputs matters, since after the first `DxilInst_StorePrimitiveOutput` is encountered all remaining outputs will be ignored.

python3kgae: Got it. Missed dead code when update the code :(
Fixed.

damyanp: If the tests weren't capturing this, it strongly suggests to me that the testing coverage for this change needs to be improved.

damyanp (changes requested):
Testing coverage for this needs to be improved since some clearly incorrect code managed to slip through.

lib/HLSL/ComputeViewIdStateBuilder.cpp
python3kgae: It is impossible to capture it, since it is dead code :(
If the code can reach line 494, the `IsForPatchConstantOrPrimitive` must be true.
So `IsForPatchConstantOrPrimitive = true;` will not change anything.

damyanp: Ah, I didn't notice that the various tests on `IsForPatchConstantOrPrimitive` weren't all the same.

python3kgae:
> Testing coverage for this needs to be improved since some clearly incorrect code managed to slip through.

It is dead code which cannot make any different output.

damyanp (approved):

PR #6887, opened 2024-08-28
[NFC] update print functions for PSV

python3kgae:
Update print functions to share with validation.
This change is part of a big change for #6817 where the print function is shared between the DxilPipelineStateValidation::Print and the psv content validation.

Move this change out of the big change to help review.

For #6817

### DIFF
--- a/lib/DxilContainer/DxilPipelineStateValidation.cpp
+++ b/lib/DxilContainer/DxilPipelineStateValidation.cpp
@@ -360,8 +360,12 @@ void PSVResourceBindInfo1::Print(raw_ostream &OS) const {
 }
 
 void PSVSignatureElement::Print(raw_ostream &OS) const {
+  Print(OS, GetSemanticName());
+}
+
+void PSVSignatureElement::Print(raw_ostream &OS, const char *Name) const {
   OS << "PSVSignatureElement:\n";
-  OS << "  SemanticName: " << GetSemanticName() << "\n";
+  OS << "  SemanticName: " << Name << "\n";
   OS << "  SemanticIndex: ";
   const uint32_t *SemanticIndexes = GetSemanticIndexes();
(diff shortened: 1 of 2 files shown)


damyanp:
I don't understand why this change is being made.  Can you add some explanation please to the description?  Is this just refactoring?  If so, why?  Or is there actually some functional change in here?

lib/DxilContainer/DxilPipelineStateValidation.cpp
```diff
@@ -360,8 +360,12 @@ void PSVResourceBindInfo1::Print(raw_ostream &OS) const {
 }
 
 void PSVSignatureElement::Print(raw_ostream &OS) const {
+  Print(OS, GetSemanticName());
+}
+
+void PSVSignatureElement::Print(raw_ostream &OS, const char *Name) const {
```
damyanp, line 366: I'm trying to figure out the motivation for this change?  Is there some place now where we want a name other than what `GetSemanticName()` returns to be displayed?

python3kgae: It is for validation to print Signature Element where SemanticName is not in StringTable.

damyanp: I guess I don't have enough context to understand what that means.  So sometimes PSVSignatureElement doesn't contain the data it needs?

python3kgae: It is for emit error message on a PSVSignatureElement generated from DxilSiginatureElement which doesn't have SemanticName setup. The SemanticName will be get from the DxilSiginatureElement.

damyanp: This may be a case of PR being sliced too far away from the code that makes it possible to understand the change.

PR #6888, opened 2024-08-29
Replace reachable llvm_unreachable with errors in HLSignatureLower

tex3d:
Also, constrain IA signature allocation to legal bounds.


lib/HLSL/DxilPackSignatureElement.cpp
```diff
@@ -73,8 +73,10 @@ unsigned PackDxilSignature(DxilSignature &sig, DXIL::PackingStrategy packing) {
     // incrementally assign each element that belongs in the signature to the
     // start of the next free row
     for (auto &SE : packElements) {
-      SE.SetLocation(rowsUsed, 0);
-      rowsUsed += SE.GetRows();
+      if (SE.GetRows() <= (32 - rowsUsed)) {
```
python3kgae, line 76: Should we report error when out of bound?

PR #6892, opened 2024-09-03
Disable the aggressive reassociation optimization by default

lizhengxing:
The aggressive reassociation optimization for this common factor issue (https://github.com/microsoft/DirectXShaderCompiler/issues/6593) can reduce redundant calculations obviously.

But it also exposed some issues for the final codegen and caused perf regressions for some shaders in runtime tests.

Fixing those codegen issues is non-trival, as a workaround, this PR disables the aggressive reassociation optimization by default and will get the perf loss back.


damyanp:
Is this reverting a recently made change?

lizhengxing:
> Is this reverting a recently made change?

No. It just turns off  the change by default.

PR #6893, opened 2024-09-03
Cherry-Pick: Fix debug info offsets for vectors with 16-bit types (#6775)

jeffnn:
This fixes a bug where the offsets for elements in vectors with 16-bit types doesn't take into account alignment bits and PIX wouldn't display vector element values correctly in the shader debugger. Eg. if `-enable-16bit-types` wasn't set, the offsets for a min16float4 would be 0, 16, 32, 48 instead of 0, 32, 64, 96.

Also removed the assert in PopulateAllocaMap_StructType that was checking whether the calculated aligned offset matches the packed offset (from SortedMembers) because it was false for members with sizes smaller than the alignment size.

(cherry picked from commit 84c0a095576c63d8103ffffc56bb165b3cf9ddf3)


damyanp (approved):
Approved for checking in to the release branch.

adam-yang (approved):

PR #6901, opened 2024-09-05
[Test] Add validator version check for tests

python3kgae:
This change adds validation version check for the tests.

This is for fix tests fail when running on different version of validator.


hekota (approved):

tex3d (approved):
LGTM.

Issue #5440, opened 2023-07-20
[SPIRV] Handle arrays of RWStructuredBuffers, AppendStructuredBuffers, and ConsumeStructuredBuffers


s-perron:
# Summary

Many individuals have [requested an array of RWStructuredBuffers](https://github.com/microsoft/DirectXShaderCompiler/issues/3281). The main design consideration is how to handle the counter buffer and how to assign the sets and bindings.

The solution will be to handle a RWStructuredBuffer as if it were two arrays of resources: one for the buffer and one for the counter. We can assign set and binding in the same way we do other resource arrays, such as arrays of images and samplers.

# Design difficulties

As [Ehsan noted](https://github.com/microsoft/DirectXShaderCompiler/issues/3281#issuecomment-741846195), there are some design decisions that need to be made regarding the bindings and how to flatten the arrays. I believe that this comment was interpreted in the most pessimistic way, which led to [an early design where we attempted to add arrays of RWStructuredBuffers, but only in the case where counters are not needed](https://github.com/microsoft/DirectXShaderCompiler/pull/4663). This is not an ideal solution because it creates a special case in the compiler: There could be a counter for the RWStructuredBuffer, but not always. This also limits what users will be able to do.

A better solution would be to permit arrays of RWStructuredBuffers, regardless of whether or not counters are required. This would simplify the compiler and give users more flexibility. When I read Ehsan's comment, I do not believe the issues raised are insurmountable. Rather, they are issues that must be addressed. In general, we already have the tools and precedent to solve them.

# Step to implement

The code to implement each of these case will be guarded by a command line option until the feature is complete

- [x] Add option `-fvk-allow-rwstructuredbuffer-arrays` to enable arrays of RWStructuredBuffers and the like. (#5405)
- [x] Fix sized arrays with no counter (#5405)
- [x] Fix sized arrays with a counter (https://github.com/microsoft/DirectXShaderCompiler/pull/5415)
- [x] Passing an element of the array to a function without a counter (https://github.com/microsoft/DirectXShaderCompiler/pull/5447)
- [x] Passing an element of the array to a function with a counter (https://github.com/microsoft/DirectXShaderCompiler/pull/5447)
- [x] Creating a local alias to the global array without a counter (https://github.com/microsoft/DirectXShaderCompiler/pull/5523)
- [x] Creating a local alias to the global array with a counter (https://github.com/microsoft/DirectXShaderCompiler/pull/5523)
[...]

PR #6903, opened 2024-09-06
Enable vk-allow-rwstructuredbuffer-arrays, and remove option

s-perron:
We have done all of the development on this feature that we will do.

We cannot handle Multidimentional arrays of these buffers because Vulkan
does not allow it.

We cannot handle arrays that are in a struct if the counter variable is
used. We have tests cases to identify those cases.

Closes https://github.com/microsoft/DirectXShaderCompiler/issues/5440


Keenuts (approved):

s-perron:
FYI: @EpicJeanNoeMorissette

cassiebeckley (approved):

PR #6910, opened 2024-09-11
Fix ASAN UAF in DxilConditionalMem2Reg

amaiorano:
ScalarizePreciseVectorAlloca would iterate over all instructions, then for each instruction use, would iterate and potentially erase the instruction. If the erased instruction was the immediate next instruction after the alloca, this would invalidate the outer instruction iterator. Fixed by collecting the allocas in a vector first.

### DIFF
--- a/lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
+++ b/lib/Transforms/Scalar/DxilConditionalMem2Reg.cpp
@@ -270,7 +270,7 @@ public:
   static bool ScalarizePreciseVectorAlloca(Function &F) {
     BasicBlock *Entry = &*F.begin();
 
-    bool Changed = false;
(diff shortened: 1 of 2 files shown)


amaiorano:
As noted in the test file:
```
// Unfortunately, we cannot create an IR test for this because dxil-cond-mem2reg
// executes between scalarrepl-param-hlsl and hlsl-dxil-precise, and the former
// temporarily marks empty functions as 'precise' while the latter pass uses this
// information, and then deletes these functions. But splitting the passes in between
// these two fails validation because empty functions cannot have attributes on them.
// So we use a full HLSL test for this.
```

python3kgae (approved):

llvm-beanz (approved):

PR #6912, opened 2024-09-12
Fix CMake gen failing when HLSL_BUILD_DXILCONV=0

amaiorano:
Target dxildll depends on DxcRuntimeEtw, so make sure to add this target, even when HLSL_BUILD_DXILCONV is false.


amaiorano:
I think this is the preferable fix over https://github.com/microsoft/DirectXShaderCompiler/pull/6911. I'm happy to land this if you are.

python3kgae (approved):

llvm-beanz (approved):

PR #6920, opened 2024-09-18
sema: fix string literal allowed in templates

### LINKED ISSUE #6877: [SPIR-V] DXC crash with buffer assignment + template function
**Description**
We are seeing a crash in our shader debug code which uses template functions to output strings to a `RWByteAddressBuffer`.

Only SPIR-V compilation crashes.  DXIL compilation works fine.

**Steps to Reproduce**
This is the most minimal shader that I have found to reproduce the problem:
```
RWByteAddressBuffer u_OutputBuffer : register(u0);

template<typename T, uint N> uint StrLen(T str[N])
{
    // Includes the null terminator
    return N;
}

[numthreads(1, 1, 1)]
void main()
{
    // commenting out either of these lines makes compilation succeed
    // calling StrLen() on a non-string array instead also appears to compile successfully
    const uint unusedStrLen = StrLen("test_string");
    RWByteAddressBuffer unusedBuffer = u_OutputBuffer;
}
```
DXC command line used to reproduce:
```
dxc.exe repro.hlsl -T cs_6_5 -E main -spirv
```

**Actual Behavior**
Compilation crashes with error
```
Internal compiler error: access violation. Attempted to read from address 0x0000000000000000
```

**Environment**
- Reproduced with DXC July 2024 release `dxcompiler.dll: 1.8 - 1.8.2407.7 (416fab6b5); dxil.dll: 1.8(101.8.2407.12)`
- Reproduced with DXC from Vulkan SDK 1.3.290.0 `dxcompiler.dll: 1.8 - 1.8.0.4650 (49d1ae9a0); dxil.dll: 1.5(10.0.19041.1001)`
- Host Operating System: Windows 10 19045.4046

Keenuts:
String literal are not supported in HLSL, except in the special printf case.
Most cases were handled, but not the templated function case.

Fixes #6877

### DIFF
--- a/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
+++ b/tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
@@ -7663,8 +7663,8 @@ def err_hlsl_unsupported_nested_typedef : Error<
 def err_hlsl_unsupported_operator : Error<
   "operator is not supported">;
 // This error is unused because the HLSL parser will catch all cases.
-//def err_hlsl_unsupported_string_literal : Error<
-//  "unsupported style of string literal - use a char-based literal">;
+def err_hlsl_unsupported_string_literal : Error<
+  "string literals parameters are not supported in HLSL">;
 def err_hlsl_unsupported_struct_op: Error<
   "operator cannot be used with user-defined type %0">;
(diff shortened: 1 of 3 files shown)


s-perron (approved):
LGTM

llvm-beanz (approved):

PR #6921, opened 2024-09-19
Remove assert.h include from DxilPipelineStateValidation.h

### LINKED ISSUE #6922: Include <assert.h> in DxilPipelineStateValidation.h caused build failures
**Description**
Some projects outside DXC include DxilPipelineStateValidation.h. Their own custom assert might be overwritten by standard assert if  <assert.h> is included in DxilPipelineStateValidation.h.  

**Actual Behavior**
The compilation will report "too many arguments for function-like macro invocation" if the custom assert is defined has multiple arguments.

**Environment**
- DXC version:  dxcompiler.dll: 1.8 - 1.8.0.14576 (main, 381750e25); dxil.dll: 1.8(1.8.0.14576)
- Host Operating System : Windows

lizhengxing:
DxilPipelineStateValidation.h is included from multiple environments. Don't include assert.h in it, otherwise it might overwrite the custom assert in the code that includes DxilPipelineStateValidation.h.

Fixes #6922

### DIFF
--- a/include/dxc/DxilContainer/DxilPipelineStateValidation.h
+++ b/include/dxc/DxilContainer/DxilPipelineStateValidation.h
@@ -12,19 +12,18 @@
 #ifndef __DXIL_PIPELINE_STATE_VALIDATION__H__
 #define __DXIL_PIPELINE_STATE_VALIDATION__H__
 
+// Don't include assert.h here.
+// Since this header is included from multiple environments,
+// it is necessary to define assert before this header is included.
+// #include <assert.h>
+
 #include "dxc/WinAdapter.h"
-#include <assert.h>
 #include <cstring>
 #include <stdint.h>
 
 namespace llvm {
 class raw_ostream;
 }
-
-// Don't include assert.h here.
-// Since this header is included from multiple environments,
-// it is necessary to define assert before this header is included.
-// #include <assert.h>
 
 #ifndef UINT_MAX
 #define UINT_MAX 0xffffffff
--- a/lib/DxilContainer/DxilPipelineStateValidation.cpp
+++ b/lib/DxilContainer/DxilPipelineStateValidation.cpp
@@ -9,6 +9,8 @@
 //                                                                           //
 ///////////////////////////////////////////////////////////////////////////////
 
+#include <assert.h>  // Needed for DxilPipelineStateValidation.h
+
 #include "dxc/DxilContainer/DxilPipelineStateValidation.h"
 #include "dxc/DXIL/DxilModule.h"
 #include "dxc/DXIL/DxilResource.h"
--- a/lib/DxilContainer/DxilPipelineStateValidation.cpp
+++ b/lib/DxilContainer/DxilPipelineStateValidation.cpp
@@ -9,7 +9,7 @@
 //                                                                           //
 ///////////////////////////////////////////////////////////////////////////////
 
-#include <assert.h>  // Needed for DxilPipelineStateValidation.h
+#include <assert.h> // Needed for DxilPipelineStateValidation.h
 
 #include "dxc/DxilContainer/DxilPipelineStateValidation.h"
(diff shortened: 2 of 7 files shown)


damyanp (commented):
It's unfortunate that we have to do this, but I believe that this is the pragmatic change to make at this point.

Can you file an issue please to cover cleaning this up properly?

lib/DxilContainer/DxilPipelineStateValidation.cpp
```diff
@@ -20,6 +20,7 @@
 #include "llvm/IR/DataLayout.h"
 #include "llvm/IR/Module.h"
 #include "llvm/Support/raw_ostream.h"
+#include <assert.h> // Needed for DxilPipelineStateValidation.h
```
tex3d, line 23: clang-format autopublish moved this assert here, but it needs to be included before `DxilPipelineStateValidation.h`.  So I believe it needs to be separated by a blank line, right?

lib/DxilContainer/DxilRDATBuilder.cpp
```diff
@@ -2,6 +2,7 @@
 #include "dxc/DxilContainer/DxilPipelineStateValidation.h"
 #include "dxc/Support/FileIOHelper.h"
 #include "dxc/Support/Global.h"
+#include <assert.h> // Needed for DxilPipelineStateValidation.h
```
tex3d, line 5: clang-format autopublish moved this assert here, but it needs to be included before `DxilPipelineStateValidation.h`.  So I believe it needs to be separated by a blank line, right?

lib/DxilContainer/DxilPipelineStateValidation.cpp
lizhengxing: @tex3d  Yes, you're right. Added a blank line to fix it.

lib/DxilContainer/DxilRDATBuilder.cpp
lizhengxing: @tex3d Yes, you're right. Added a blank line to fix it.

include/dxc/DxilContainer/DxilPipelineStateValidation.h
```diff
@@ -12,20 +12,19 @@
 #ifndef __DXIL_PIPELINE_STATE_VALIDATION__H__
 #define __DXIL_PIPELINE_STATE_VALIDATION__H__
 
+// Don't include assert.h here.
+// Since this header is included from multiple environments,
+// it is necessary to define assert before this header is included.
+// #include <assert.h>
```
python3kgae, line 18: Shall we remove `// #include <assert.h>`?

lizhengxing:
> It's unfortunate that we have to do this, but I believe that this is the pragmatic change to make at this point.
> 
> Can you file an issue please to cover cleaning this up properly?

Sure, I've filed an issue here #6922

include/dxc/DxilContainer/DxilPipelineStateValidation.h
lizhengxing: I think it's fine. It's the original comment and use `// #include <assert.h>` here probably can catch more attention.

lizhengxing:
> It's unfortunate that we have to do this, but I believe that this is the pragmatic change to make at this point.
> 
> Can you file an issue please to cover cleaning this up properly?

@damyanp Sure, I've filed an issue here https://github.com/microsoft/DirectXShaderCompiler/issues/6922

python3kgae (approved):

damyanp (changes requested):
Can you investigate whether we could do something like:

```
#ifndef assert
#include <assert.h>
#endif
```

So that we can remove all the modifications to the other files, but still have this header worked when used in the other environments please?

lizhengxing:
Close this PR since we will use this fix https://github.com/microsoft/DirectXShaderCompiler/pull/6925 .

PR #6923, opened 2024-09-19
Fix mismatched memory allocation and deallocation.

s-perron:
In NestedNameSpecifier.cpp, the Append function allocates data using `new`:

https://github.com/microsoft/DirectXShaderCompiler/blob/e1bb926f63ba44119577c845e9d7a1a10c0f50e9/tools/clang/lib/AST/NestedNameSpecifier.cpp#L449-L450

However, if it deallocated using `free`:

https://github.com/microsoft/DirectXShaderCompiler/blob/e1bb926f63ba44119577c845e9d7a1a10c0f50e9/tools/clang/lib/AST/NestedNameSpecifier.cpp#L514-L516

This can cause problems. I'm changing the `free` to `delete`.

### DIFF
--- a/tools/clang/lib/AST/NestedNameSpecifier.cpp
+++ b/tools/clang/lib/AST/NestedNameSpecifier.cpp
@@ -512,7 +512,7 @@ operator=(const NestedNameSpecifierLocBuilder &Other) {
   
   // Free our storage, if we have any.
   if (BufferCapacity) {
-    free(Buffer);
+    delete[] Buffer;
     BufferCapacity = 0;
   }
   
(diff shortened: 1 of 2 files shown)


Keenuts (approved):

tools/clang/lib/AST/NestedNameSpecifier.cpp
```diff
@@ -512,7 +512,7 @@ operator=(const NestedNameSpecifierLocBuilder &Other) {
   
   // Free our storage, if we have any.
   if (BufferCapacity) {
-    free(Buffer);
+    delete[] Buffer;
```
damyanp, line 515: Should this (and the other change on line 650) also have the `// HLSL Change: Use overridable operator delete` comment?

damyanp (approved):

tools/clang/lib/AST/NestedNameSpecifier.cpp
s-perron: Done.

PR #6924, opened 2024-09-19
[Validator] Check size of PSV.

### LINKED ISSUE #6817: Allow string table in the PSV0 part to have a different order than DXC during validation

python3kgae:
Check size of PSV part matches the PSVVersion.
Updated DxilPipelineStateValidation::ReadOrWrite to read based on initInfo.PSVVersion.
And return fail when size mismatch in RWMode::Read.

Fixes #6817

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -23,6 +23,7 @@ The included licenses apply to the following files:
 Place release notes for the upcoming release below this line and remove this line upon naming this release.
 
 - The incomplete WaveMatrix implementation has been removed.
+- The DXIL validator supports the string table and semantic index table in any order.
 
 ### Version 1.8.2407
 
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -23,7 +23,7 @@ The included licenses apply to the following files:
 Place release notes for the upcoming release below this line and remove this line upon naming this release.
 
 - The incomplete WaveMatrix implementation has been removed.
-- The DXIL validator supports the string table and semantic index table in any order.
+- DXIL container validation for PSV0 part allows any content ordering inside string and semantic index tables.
 
 ### Version 1.8.2407
 
--- a/include/dxc/DxilContainer/DxilPipelineStateValidation.h
+++ b/include/dxc/DxilContainer/DxilPipelineStateValidation.h
@@ -596,9 +596,12 @@ private:
                    const PSVInitInfo &initInfo = PSVInitInfo(MAX_PSV_VERSION));
 
 public:
-  bool InitFromPSV0(const void *pBits, uint32_t size,
-                    uint32_t PSVVersion = MAX_PSV_VERSION) {
-    return ReadOrWrite(pBits, &size, RWMode::Read, PSVInitInfo(PSVVersion));
+  bool InitFromPSV0(const void *pBits, uint32_t size) {
+    return ReadOrWrite(pBits, &size, RWMode::Read);
+  }
+
+  bool InitFromPSV0(const void *pBits, uint32_t *pSize, uint32_t PSVVersion) {
+    return ReadOrWrite(pBits, pSize, RWMode::Read, PSVInitInfo(PSVVersion));
   }
 
   // Initialize a new buffer
@@ -1084,8 +1087,7 @@ DxilPipelineStateValidation::ReadOrWrite(const void *pBits, uint32_t *pSize,
     *pSize = rw.GetSize();
     m_pPSVRuntimeInfo1 = nullptr; // clear ptr to tempRuntimeInfo
   } else if (mode == RWMode::Read) {
-    if (rw.GetSize() != rw.GetOffset())
-      return false;
+    *pSize = rw.GetOffset();
   }
   return true;
 }
--- a/include/dxc/DxilContainer/DxilPipelineStateValidation.h
+++ b/include/dxc/DxilContainer/DxilPipelineStateValidation.h
@@ -554,6 +554,7 @@ public:
 
     uint32_t GetSize() { return Size; }
     RWMode GetMode() { return Mode; }
+    uint32_t GetOffset() { return Offset; }
 
     // Return true if size fits in remaing buffer.
     bool CheckBounds(size_t size);
@@ -595,8 +596,9 @@ private:
                    const PSVInitInfo &initInfo = PSVInitInfo(MAX_PSV_VERSION));
 
 public:
-  bool InitFromPSV0(const void *pBits, uint32_t size) {
-    return ReadOrWrite(pBits, &size, RWMode::Read);
+  bool InitFromPSV0(const void *pBits, uint32_t size,
+                    uint32_t PSVVersion = MAX_PSV_VERSION) {
+    return ReadOrWrite(pBits, &size, RWMode::Read, PSVInitInfo(PSVVersion));
   }
 
   // Initialize a new buffer
@@ -937,12 +939,15 @@ DxilPipelineStateValidation::ReadOrWrite(const void *pBits, uint32_t *pSize,
 
   PSV_RETB(rw.MapValue(&m_uPSVRuntimeInfoSize, initInfo.RuntimeInfoSize()));
   PSV_RETB(rw.MapArray(&m_pPSVRuntimeInfo0, 1, m_uPSVRuntimeInfoSize));
-  AssignDerived(&m_pPSVRuntimeInfo1, m_pPSVRuntimeInfo0,
-                m_uPSVRuntimeInfoSize); // failure ok
-  AssignDerived(&m_pPSVRuntimeInfo2, m_pPSVRuntimeInfo0,
-                m_uPSVRuntimeInfoSize); // failure ok
-  AssignDerived(&m_pPSVRuntimeInfo3, m_pPSVRuntimeInfo0,
-                m_uPSVRuntimeInfoSize); // failure ok
+  if (initInfo.PSVVersion > 0)
+    AssignDerived(&m_pPSVRuntimeInfo1, m_pPSVRuntimeInfo0,
+                  m_uPSVRuntimeInfoSize); // failure ok
+  if (initInfo.PSVVersion > 1)
+    AssignDerived(&m_pPSVRuntimeInfo2, m_pPSVRuntimeInfo0,
+                  m_uPSVRuntimeInfoSize); // failure ok
+  if (initInfo.PSVVersion > 2)
+    AssignDerived(&m_pPSVRuntimeInfo3, m_pPSVRuntimeInfo0,
+                  m_uPSVRuntimeInfoSize); // failure ok
 
   // In RWMode::CalcSize, use temp runtime info to hold needed values from
   // initInfo
@@ -1078,6 +1083,9 @@ DxilPipelineStateValidation::ReadOrWrite(const void *pBits, uint32_t *pSize,
   if (mode == RWMode::CalcSize) {
     *pSize = rw.GetSize();
     m_pPSVRuntimeInfo1 = nullptr; // clear ptr to tempRuntimeInfo
+  } else if (mode == RWMode::Read) {
+    if (rw.GetSize() != rw.GetOffset())
+      return false;
   }
   return true;
 }
--- a/include/dxc/DxilContainer/DxilPipelineStateValidation.h
+++ b/include/dxc/DxilContainer/DxilPipelineStateValidation.h
@@ -223,7 +223,8 @@ struct PSVStringTable {
   PSVStringTable() : Table(nullptr), Size(0) {}
   PSVStringTable(const char *table, uint32_t size) : Table(table), Size(size) {}
   const char *Get(uint32_t offset) const {
-    assert(offset < Size && Table && Table[Size - 1] == '\0');
+    if (!(offset < Size && Table && Table[Size - 1] == '\0'))
+      return nullptr;
     return Table + offset;
   }
 };
@@ -341,7 +342,8 @@ struct PSVSemanticIndexTable {
   PSVSemanticIndexTable(const uint32_t *table, uint32_t entries)
       : Table(table), Entries(entries) {}
   const uint32_t *Get(uint32_t offset) const {
-    assert(offset < Entries && Table);
+    if (!(offset < Entries && Table))
+      return nullptr;
     return Table + offset;
   }
 };
@@ -554,7 +556,6 @@ public:
 
     uint32_t GetSize() { return Size; }
     RWMode GetMode() { return Mode; }
-    uint32_t GetOffset() { return Offset; }
 
     // Return true if size fits in remaing buffer.
     bool CheckBounds(size_t size);
@@ -600,10 +601,6 @@ public:
     return ReadOrWrite(pBits, &size, RWMode::Read);
   }
 
-  bool InitFromPSV0(const void *pBits, uint32_t *pSize, uint32_t PSVVersion) {
-    return ReadOrWrite(pBits, pSize, RWMode::Read, PSVInitInfo(PSVVersion));
-  }
-
   // Initialize a new buffer
   // call with null pBuffer to get required size
 
@@ -640,7 +637,8 @@ public:
   _T *GetRecord(void *pRecords, uint32_t recordSize, uint32_t numRecords,
                 uint32_t index) const {
     if (pRecords && index < numRecords && sizeof(_T) <= recordSize) {
-      assert((size_t)index * (size_t)recordSize <= UINT_MAX);
+      if (!((size_t)index * (size_t)recordSize <= UINT_MAX))
+        return nullptr;
       return reinterpret_cast<_T *>(reinterpret_cast<uint8_t *>(pRecords) +
                                     (index * recordSize));
     }
@@ -1086,8 +1084,6 @@ DxilPipelineStateValidation::ReadOrWrite(const void *pBits, uint32_t *pSize,
   if (mode == RWMode::CalcSize) {
     *pSize = rw.GetSize();
     m_pPSVRuntimeInfo1 = nullptr; // clear ptr to tempRuntimeInfo
-  } else if (mode == RWMode::Read) {
-    *pSize = rw.GetOffset();
   }
   return true;
 }
@@ -1132,6 +1128,10 @@ void InitPSVResourceBinding(PSVResourceBindInfo0 *, PSVResourceBindInfo1 *,
 void InitPSVSignatureElement(PSVSignatureElement0 &E,
                              const DxilSignatureElement &SE,
                              bool i1ToUnknownCompat);
+
+// Setup PSVInitInfo with DxilModule.
+// Note that the StringTable and PSVSemanticIndexTable are not done.
+void SetupPSVInitInfo(PSVInitInfo &InitInfo, const DxilModule &DM);
 
 // Setup shader properties for PSVRuntimeInfo* with DxilModule.
 void SetShaderProps(PSVRuntimeInfo0 *pInfo, const DxilModule &DM);
--- a/include/dxc/DxilContainer/DxilPipelineStateValidation.h
+++ b/include/dxc/DxilContainer/DxilPipelineStateValidation.h
@@ -940,15 +940,12 @@ DxilPipelineStateValidation::ReadOrWrite(const void *pBits, uint32_t *pSize,
 
   PSV_RETB(rw.MapValue(&m_uPSVRuntimeInfoSize, initInfo.RuntimeInfoSize()));
   PSV_RETB(rw.MapArray(&m_pPSVRuntimeInfo0, 1, m_uPSVRuntimeInfoSize));
-  if (initInfo.PSVVersion > 0)
-    AssignDerived(&m_pPSVRuntimeInfo1, m_pPSVRuntimeInfo0,
-                  m_uPSVRuntimeInfoSize); // failure ok
-  if (initInfo.PSVVersion > 1)
-    AssignDerived(&m_pPSVRuntimeInfo2, m_pPSVRuntimeInfo0,
-                  m_uPSVRuntimeInfoSize); // failure ok
-  if (initInfo.PSVVersion > 2)
-    AssignDerived(&m_pPSVRuntimeInfo3, m_pPSVRuntimeInfo0,
-                  m_uPSVRuntimeInfoSize); // failure ok
+  AssignDerived(&m_pPSVRuntimeInfo1, m_pPSVRuntimeInfo0,
+                m_uPSVRuntimeInfoSize); // failure ok
+  AssignDerived(&m_pPSVRuntimeInfo2, m_pPSVRuntimeInfo0,
+                m_uPSVRuntimeInfoSize); // failure ok
+  AssignDerived(&m_pPSVRuntimeInfo3, m_pPSVRuntimeInfo0,
+                m_uPSVRuntimeInfoSize); // failure ok
 
   // In RWMode::CalcSize, use temp runtime info to hold needed values from
   // initInfo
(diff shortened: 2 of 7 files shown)


include/dxc/DxilContainer/DxilPipelineStateValidation.h
```diff
@@ -1078,6 +1083,9 @@ DxilPipelineStateValidation::ReadOrWrite(const void *pBits, uint32_t *pSize,
   if (mode == RWMode::CalcSize) {
     *pSize = rw.GetSize();
     m_pPSVRuntimeInfo1 = nullptr; // clear ptr to tempRuntimeInfo
+  } else if (mode == RWMode::Read) {
+    if (rw.GetSize() != rw.GetOffset())
+      return false;
```
tex3d, line 1088: We don't want to fail the load when the size isn't equal to the version, otherwise, the forward compatibility will be broken.  We only want to verify this in the validator, to ensure that we only validate known versions of PSV.

tex3d (changes requested):
Should not block loading of newer PSV versions.  Need to do this only in the validator code.

include/dxc/DxilContainer/DxilPipelineStateValidation.h
python3kgae: Changed to return rw.GetOffset() and do the compare in validator.

python3kgae: We'll fail the load when PSVVersion > MAX_PSV_VERSION anyway.
So, fail the load when size not match should be fine as well.

python3kgae: PSVVersion could send as default. So fail the load will not work :(

include/dxc/DxilContainer/DxilPipelineStateValidation.h
```diff
@@ -937,12 +942,15 @@ DxilPipelineStateValidation::ReadOrWrite(const void *pBits, uint32_t *pSize,
...
-                m_uPSVRuntimeInfoSize); // failure ok
-  AssignDerived(&m_pPSVRuntimeInfo3, m_pPSVRuntimeInfo0,
-                m_uPSVRuntimeInfoSize); // failure ok
+  if (initInfo.PSVVersion > 0)
+    AssignDerived(&m_pPSVRuntimeInfo1, m_pPSVRuntimeInfo0,
+                  m_uPSVRuntimeInfoSize); // failure ok
+  if (initInfo.PSVVersion > 1)
+    AssignDerived(&m_pPSVRuntimeInfo2, m_pPSVRuntimeInfo0,
+                  m_uPSVRuntimeInfoSize); // failure ok
+  if (initInfo.PSVVersion > 2)
```
tex3d, line 951: These should be unnecessary.  The scenarios are:
- CalcSize or Write mode: `m_uPSVRuntimeInfoSize` based on `PSVVersion` in `initInfo`, `AssignDerived` will only assign when struct fits in size.
- Read mode: `m_uPSVRuntimeInfoSize` determines the struct version available, PSVVersion in initInfo is not used for this, and there is no independent source for PSVVersion for PSV readers.

Validator will want to validate that `m_uPSVRuntimeInfoSize` matches the one computed by `initInfo.RuntimeInfoSize()` where `initInfo.PSVVersion` has been set according to validator version.  It's probably best if this is done manually by the validator before invoking `InitFromPSV0`, since an incorrect size will run that off the rails in a way that might just result in failure to load (malformed), or maybe not any error depending on the data and sizes, rather than a more specific error that we could emit in the validator.  If this size is wrong, there's no point in trying to read anything else.

include/dxc/DxilContainer/DxilPipelineStateValidation.h
```diff
@@ -1078,6 +1086,8 @@ DxilPipelineStateValidation::ReadOrWrite(const void *pBits, uint32_t *pSize,
   if (mode == RWMode::CalcSize) {
     *pSize = rw.GetSize();
     m_pPSVRuntimeInfo1 = nullptr; // clear ptr to tempRuntimeInfo
+  } else if (mode == RWMode::Read) {
+    *pSize = rw.GetOffset();
```
tex3d, line 1090: We should not need this, as the passed in size when reading is the size of the container part.
CalcSize mode should be sufficient for validating the container part size is that calculated size, aligned, and any padding for alignment is filled with zero.

tex3d: Reverting this eliminates the need to add `GetOffset()` as well.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -509,14 +506,24 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
 
 static void VerifyPSVMatches(ValidationContext &ValCtx, const void *pPSVData,
                              uint32_t PSVSize) {
+  unsigned ValMajor, ValMinor;
+  ValCtx.DxilMod.GetValidatorVersion(ValMajor, ValMinor);
+  unsigned PSVVersion = hlsl::GetPSVVersion(ValMajor, ValMinor);
   DxilPipelineStateValidation PSV;
-  if (!PSV.InitFromPSV0(pPSVData, PSVSize)) {
+  uint32_t PSVSizeRead = PSVSize;
+  if (!PSV.InitFromPSV0(pPSVData, &PSVSizeRead, PSVVersion)) {
```
tex3d, line 514: Before we do this, we should check that `m_uPSVRuntimeInfoSize` (first 4 bytes of data) matches expected size for PSVVersion (`PSVInitInfo(PSVVersion).RuntimeInfoSize()`), reporting error and aborting if not.

Then, there's no need to pass PSVVersion to `InitFromPSV0`, and that overload can be removed/reverted.

tex3d (changes requested):
In addition to the comments, we should be checking the string and index tables more thoroughly:
- check that string table is well formed - size is aligned, buffer ends with `\0`
- check that accesses to string table entries are in-bounds, before using the value.
- check that accesses to index table entries are in-bounds, and that the (index+1)+size is within dword size of index table, before using the index array.

lib/DxilValidation/DxilContainerValidation.cpp
tex3d: We should also probably do some additional manual validation of sizes/strides based on the format before diving into the full `InitFromPSV0`:
- `size_t offset = 4`
- start by checking that the `offset + PSVRuntimeInfo_size <= PSVSizeRead`
- map the appropriate PSV structures fitting in size manually to `pPSVData + offset`. `offset += PSVRuntimeInfo_size`
- if ResourceCount > 0:
  - next 4 bytes is PSVResourceBindInfo_size, which must be ` == PSVInitInfo(PSVVersion).ResourceBindInfoSize()`. `offset += 4`
  - check that `offset + PSVResourceBindInfo_size * ResourceCount <= PSVSizeRead`. `offset += PSVResourceBindInfo_size * ResourceCount`
- if `pPSVRuntimeInfo1`:
  - `offset + 4 <= PSVSizeRead`.  Read `uint32_t StringTableSize`. `offset += 4`
    - if `StringTableSize` non-zero:
      - `StringTableSize` must be 4-aligned, `offset + StringTableSize <= PSVSizeRead`, `pPSVData + offset + StringTableSize - 1` must be null char `\0`. `offset += StringTableSize`
    - `offset + 4 <= PSVSizeRead`.  Read `uint32_t SemanticIndexTableEntries`. `offset += 4`
    - `offset + (SemanticIndexTableEntries * 4) <= PSVSizeRead`. `offset += SemanticIndexTableEntries * 4`
  - If `SigInputElements || SigOutputElements || SigPatchConstOrPrimElements`:
    - `offset + 4 <= PSVSizeRead`.  Read `uint32_t PSVSignatureElement_size`. `offset += 4`
    - `PSVSignatureElement_size == PSVInitInfo(PSVVersion).SignatureElementSize()`
    - `offset + PSVSignatureElement_size * (SigInputElements + SigOutputElements  + SigPatchConstOrPrimElements) <= PSVSizeRead`

After this, it's possible to manually go through the ViewID data calculation, but not much you can report other than it being malformed if the overall size isn't correct, since there are no more specific size values in this part of the data that can be independently verified.  Since `InitFromPSV0` will already do that, you can just move on I think.

If anything fails in here, we should not attempt to continue with more checks or load the PSV0 data.

It's a lot, I know, but doing this is necessary to prevent incorrect sizes for records, which can allow either extra unvalidated embedded data, or result in general malformed error when we could be more helpful.

Note how the `ReadOrWrite` method uses the `CheckedReaderWriter` to ease all these calculations and checks.  It would probably be best to write some helper methods similar to that, but with proper error reporting for the validator.  For instance, it's useful to have a method that checks the remaining size before loading some value from the current offset, then increments the offset by the size of that value.  You could simply use the CheckedReaderWriter, but there will be issues with the way errors are handled that might make it hard to report the appropriate error.

python3kgae: Working on this.

python3kgae: Updated.

docs/ReleaseNotes.md
```diff
@@ -23,6 +23,7 @@ The included licenses apply to the following files:
 Place release notes for the upcoming release below this line and remove this line upon naming this release.
 
 - The incomplete WaveMatrix implementation has been removed.
+- The DXIL validator supports the string table and semantic index table in any order.
```
tex3d, line 26: It feels like this relnote needs a bit more context.  How about:
> DXIL container validation for `PSV0` part allows any content ordering inside string and semantic index tables.

python3kgae: Updated.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -79,17 +80,95 @@ static void emitDxilDiag(LLVMContext &Ctx, const char *str) {
...
+        OffsetToUseCountMap[Start] = 0;
+        Start = i + 1;
+      }
+    }
+    if (Table.Size >= 4) {
+      // Remove the '\0's at the end of the table added for padding.
+      for (unsigned i = Table.Size - 1; i > Table.Size - 4; --i) {
+        if (Table.Table[i] != '\0')
+          break;
+        OffsetToUseCountMap.erase(i);
```
tex3d, line 102: Won't this potentially attempt to erase a non-existent entry pointing to the null terminating character of the last string in the table?

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -79,17 +80,95 @@ static void emitDxilDiag(LLVMContext &Ctx, const char *str) {
...
+
+      ValCtx.EmitFormatError(ValidationRule::ContainerUnusedItemInTable,
+                             {"StringTable", Table.Get(Offset)});
+    }
+  }
+};
+
+class SemanticIndexTableVerifier {
+  const PSVSemanticIndexTable &Table;
+  std::vector<bool> UseMask;
```
tex3d, line 129: Why not use `llvm;:BitVector`?

python3kgae: Good idea.
Updated.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +597,217 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+    return *(const uint32_t *)GetPtrAtOffset(BasePtr, Offset);
+  }
+};
+
 static void VerifyPSVMatches(ValidationContext &ValCtx, const void *pPSVData,
                              uint32_t PSVSize) {
+  SimplePSV SimplePSV(pPSVData, PSVSize);
+  if (!SimplePSV.IsValid) {
+    ValCtx.EmitFormatError(ValidationRule::ContainerPartMatches,
+                           {"Pipeline State Validation"});
```
tex3d, line 764: So, this is the generic error that the container part must match the module.  If this is invalid though, I think the error could be more specific, indicating that the PSV0 part is malformed.

python3kgae: Updated.

tex3d (commented):
Mainly, I think we need to do away with the extra size check based on module facts, since it would only hit this, after other checks passed, when it would be better to report a more specific mismatch later.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +597,217 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+          RuntimeInfo1->SigOutputVectors[0] &&
+          RuntimeInfo1->SigPatchConstOrPrimVectors) {
+        uint32_t TableSizeInBytes =
+            sizeof(uint32_t) * PSVComputeInputOutputTableDwords(
+                                   RuntimeInfo1->SigPatchConstOrPrimVectors,
+                                   RuntimeInfo1->SigOutputVectors[0]);
+        INCREMENT_POS(TableSizeInBytes);
+      }
+    }
+    IsValid = PSVSize == Offset;
```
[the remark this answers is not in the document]
python3kgae: Removed.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +597,217 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+      PSVStringTable(SimplePSV.StringTable, SimplePSV.StringTableSize);
+  PSVInfo.SemanticIndexTable = PSVSemanticIndexTable(
+      SimplePSV.SemanticIndexTable, SimplePSV.SemanticIndexTableEntries);
+  uint32_t ExpectedSize = 0;
+  DxilPipelineStateValidation SizePSV;
+  if (!SizePSV.InitNew(PSVInfo, nullptr, &ExpectedSize)) {
+    ValCtx.EmitFormatError(ValidationRule::ContainerPartMatches,
+                           {"Pipeline State Validation"});
+    return;
+  }
```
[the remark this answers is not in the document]
python3kgae: Removed.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +597,217 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+    return (const char *)BasePtr + Offset;
+  }
+  uint32_t GetUint32AtOffset(const void *BasePtr, uint32_t Offset) const {
+    return *(const uint32_t *)GetPtrAtOffset(BasePtr, Offset);
+  }
+};
+
 static void VerifyPSVMatches(ValidationContext &ValCtx, const void *pPSVData,
                              uint32_t PSVSize) {
+  SimplePSV SimplePSV(pPSVData, PSVSize);
```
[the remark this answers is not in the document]
python3kgae: Updated.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +597,217 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+  if (!SimplePSV.IsValid) {
+    ValCtx.EmitFormatError(ValidationRule::ContainerPartMatches,
+                           {"Pipeline State Validation"});
+    return;
+  }
+
+  unsigned ValMajor, ValMinor;
+  ValCtx.DxilMod.GetValidatorVersion(ValMajor, ValMinor);
+  unsigned PSVVersion = hlsl::GetPSVVersion(ValMajor, ValMinor);
```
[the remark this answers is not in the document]
python3kgae: Updated.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +597,217 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+
+  PSVInitInfo PSVInfo(PSVVersion);
+  hlsl::SetupPSVInitInfo(PSVInfo, ValCtx.DxilMod);
```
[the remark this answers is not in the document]
python3kgae: Updated with the table comment removed.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +597,217 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+
+  if (!SimplePSV.ValidatePSVInit(PSVInfo, ValCtx))
```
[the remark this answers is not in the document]
python3kgae: Updated.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +597,217 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+    return;
+
+  if (SimplePSV.StringTable &&
+      SimplePSV.StringTable[SimplePSV.StringTableSize - 1] != '\0') {
+    ValCtx.EmitFormatError(ValidationRule::ContainerContentInvalid,
+                           {"PSV part StringTable"});
+    return;
+  }
```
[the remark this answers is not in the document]
python3kgae: Updated.

tex3d (approved):
Looks good!

lib/DxilValidation/DxilContainerValidation.cpp
tex3d: I think my concern was this:
Say the string table is: "12\0\0"
There will be an `OffsetToUseCountMap` entry pointing at the last '\0', but not the second to last one that is the null-termination character for the string "12".  This loop will still iterate to this character though, and `Table.Table[i] == '\0'`, so it will attempt to erase an entry from `OffsetToUseCountMap` that doesn't exist.

include/dxc/DxilContainer/DxilPipelineStateValidation.h
```diff
@@ -635,7 +637,8 @@ class DxilPipelineStateValidation {
   _T *GetRecord(void *pRecords, uint32_t recordSize, uint32_t numRecords,
                 uint32_t index) const {
     if (pRecords && index < numRecords && sizeof(_T) <= recordSize) {
-      assert((size_t)index * (size_t)recordSize <= UINT_MAX);
+      if (!((size_t)index * (size_t)recordSize <= UINT_MAX))
```
bob80905, line 640: This would always be false right? the product of two uints will always be at most UINT_MAX, otherwise there would be an overflow, so how could it ever exceed UINT_MAX?

python3kgae: size_t is 64bits for x64.
UINT_MAX is 32bits.
x86 will always be false.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -507,16 +598,204 @@ bool VerifySignatureMatches(llvm::Module *pModule, DXIL::SignatureKind SigKind,
...
+        PSVSignatureElementSize = GetUint32AtOffset(pPSVData, Offset);
+        INCREMENT_POS(4);
+        uint32_t PSVNumSignatures = RuntimeInfo1->SigInputElements +
+                                    RuntimeInfo1->SigOutputElements +
+                                    RuntimeInfo1->SigPatchConstOrPrimElements;
+        INCREMENT_POS(PSVNumSignatures * PSVSignatureElementSize);
+      }
+      if (RuntimeInfo1->UsesViewID) {
+        for (unsigned i = 0; i < DXIL::kNumOutputStreams; i++) {
+          uint32_t SigOutputVectors = RuntimeInfo1->SigOutputVectors[i];
```
bob80905, line 668: Do we also need to account for incrementing the position for SigInputVectors? Are they in the string table? I don't see any place where the position is incremented based on the presence of SigInputVectors like it's done here for SigOutputVectors.

python3kgae: SigOutputVectors is an array of 4 for the case Geometry Shader might have 4 streams, each stream has its own output.
SigInputVectors is not an array, so not need to index for SigInputVectors.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4679,6 +4693,8 @@ TEST_F(ValidationTest, PSVStringTableReorder) {
        "')",
        "error: DXIL container mismatch for 'EntryFunctionName' between 'PSV0' "
        "part:('ain') and DXIL module:('main')",
+       "error: In 'StringTable', 'A' is not used",
```
bob80905, line 4696: Looking at your verify function at `lib/DxilValidation/DxilContainerValidation.cpp:117`, there should be an error here for `B` not being used, right?

bob80905: Presumably because `tools/clang/unittests/HLSL/ValidationTest.cpp::4625` has the start of the string table as 'B', 0.

bob80905:
Couple other questions that may not need resolution:
The validation procedure changes in certain cases:
- what if there are no resources?
-what if string table size is 0? Is that possible?
-what if there are no SemanticIndexTableEntries? Is that possible?

Are these cases worth testing?

python3kgae:
> Couple other questions that may not need resolution: The validation procedure changes in certain cases:
> 
> * what if there are no resources?
> * what if string table size is 0? Is that possible?
> * what if there are no SemanticIndexTableEntries? Is that possible?
> 
> Are these cases worth testing?

Added test to force these values to 0.

No resources happen a lot.
String table size will be 0 when no signature and PSVVersion <=2 (PSVVersion 3 added entry name which makes string table size always > 0).
SemanticIndexTableEntries will be 0 when no signature.

bob80905:
> > Couple other questions that may not need resolution: The validation procedure changes in certain cases:
> > 
> > * what if there are no resources?
> > * what if string table size is 0? Is that possible?
> > * what if there are no SemanticIndexTableEntries? Is that possible?
> > 
> > Are these cases worth testing?
> 
> Added test to force these values to 0.
> 
> No resources happen a lot. String table size will be 0 when no signature and PSVVersion <=2 (PSVVersion 3 added entry name which makes string table size always > 0). SemanticIndexTableEntries will be 0 when no signature.

Thanks, this is good, though what I meant to ask for at first was a test that ensures that the PSV is valid when there are no resources / 0 table size / no semantic index table entries. It would ensure that the PSV won't contain the table size as a part if the table size is 0. But these tests seem to be another way to validate that.

bob80905 (approved):

tools/clang/unittests/HLSL/ValidationTest.cpp
bob80905: Hey Xiang, can you respond to this? I think an error should be emitted that is not being emitted.

python3kgae: Sorry, I thought you've explained it.

Here's more detail.
The OrigStringTable is {0,   'A', 0,   'B', 0, 'm', 'a', 'i', 'n', 0,   0, 0};
'', 'A', 'B' and 'main' are used which map to index 0, 1, 3, 5

The UpdatedStringTable is {'B', 0,   'A', 0, 'm', 'a',  'i', 'n', 0,   0, 0,   0};
the index 0, 1, 3, 5 now map to 'B', '', '', 'ain'.
So 'B' is used like you said above.

bob80905: Ah I think I get it. The OffsetToUseCountMap is not being reset, but the Table is.

PR #6930, opened 2024-09-24
Fix assertion on splat of groupshared scalar

amaiorano:
When splatting a groupshared scalar, we would trip an "Invalid constantexpr cast!" assertion. This would happen while evaluating the ImplicitCastExpr to turn the groupshared scalar into a vector because the scalar expression was in a different address space (groupshared) vs the target vector (local). The fix is to ensure that when looking up the vector member expression, insert an lvalue-to-rvalue cast if necessary; i.e. when a swizzle contains duplicate elements.

### DIFF
--- a/tools/clang/test/CodeGenHLSL/SplatGroupSharedScalar.hlsl
+++ b/tools/clang/test/CodeGenHLSL/SplatGroupSharedScalar.hlsl
@@ -1,0 +1,8 @@
+// RUN: %dxc -E main -T cs_6_0 %s
+
+groupshared int a;
+[numthreads(64, 1, 1)]
+void main() {
+  a = 123;
+  int4 x = (a).xxxx;
+}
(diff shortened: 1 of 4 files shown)


python3kgae (approved):

tools/clang/test/CodeGenHLSL/SplatGroupSharedScalar.hlsl
```diff
@@ -0,0 +1,8 @@
+// RUN: %dxc -E main -T cs_6_0 %s
+
+groupshared int a;
+[numthreads(64, 1, 1)]
+void main() {
+  a = 123;
+  int4 x = (a).xxxx;
```
llvm-beanz, line 7: This test should probably verify something... As-is we verify that it doesn't crash, but it doesn't actually verify correctness of any output.

amaiorano: Done.

llvm-beanz (approved):

PR #6931, opened 2024-09-24
Support SV_DispatchGrid semantic in a nested record

### LINKED ISSUE #6928: SV_DispatchGrid semantic in a nested record
**Description**
The SV_DispatchGrid DXIL metadata for a node input record is not generated in cases where:
- the field with the SV_DispatchGrid semantic is in a nested record
- the field with the SV_DispatchGrid semantic is in a record field
- the field with the SV_DispatchGrid semantic is inherited from a base record 
Note that combinations of these scenarios are also possible.
 
**Steps to Reproduce**
https://godbolt.org/z/s5PdMbK95

**Actual Behavior**
The node input metadata record does not include a tag or pointer for a SV_DispatchGrid metadata record. 

**Environment**
- DXC version : (libdxcompiler.so: 1.8(dev;1-97af068e); libdxil.so: 1.8)
- Host Operating System :  Any (reproducible on both Windows and Linux)

tcorringham:
The SV_DispatchGrid DXIL metadata for a node input record was not generated in cases where:
- the field with the SV_DispatchGrid semantic was in a nested record
- the field with the SV_DispatchGrid semantic was in a record field
- the field with the SV_DispatchGrid semantic was inherited from a base record
- in any combinations of the above

Added FindDispatchGridSemantic() to be used by the AddHLSLNodeRecordTypeInfo() function, and added a test case.

Fixes #6928

### DIFF
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -2572,24 +2572,15 @@ bool CGMSHLSLRuntime::FindDispatchGridSemantic(const CXXRecordDecl *RD,
                                                CharUnits Offset) {
   const ASTRecordLayout &Layout = CGM.getContext().getASTRecordLayout(RD);
 
-  // Collect any bases.
-  SmallVector<const CXXRecordDecl *, 4> Bases;
+  // Check (non-virtual) bases
   for (const CXXBaseSpecifier &Base : RD->bases()) {
-    if (!Base.getType()->isDependentType())
-      Bases.push_back(Base.getType()->getAsCXXRecordDecl());
-  }
-
-  // Sort bases by offset.
-  std::stable_sort(Bases.begin(), Bases.end(),
-                   [&](const CXXRecordDecl *L, const CXXRecordDecl *R) {
-                     return Layout.getBaseClassOffset(L) <
-                            Layout.getBaseClassOffset(R);
-                   });
-
-  // Check bases in order
-  for (const CXXRecordDecl *Base : Bases) {
-    CharUnits BaseOffset = Offset + Layout.getBaseClassOffset(Base);
-    if (FindDispatchGridSemantic(Base, SDGRec, BaseOffset))
+    DXASSERT(!Base.getType()->isDependentType(),
+             "Node Record with dependent base class not caught by Sema");
+    if (Base.isVirtual() || Base.getType()->isDependentType())
+      continue;
+    CXXRecordDecl *BaseDecl = Base.getType()->getAsCXXRecordDecl();
+    CharUnits BaseOffset = Offset + Layout.getBaseClassOffset(BaseDecl);
+    if (FindDispatchGridSemantic(BaseDecl, SDGRec, BaseOffset))
       return true;
   }
 
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -2572,10 +2572,10 @@ bool CGMSHLSLRuntime::FindDispatchGridSemantic(const CXXRecordDecl *RD,
                                                CharUnits Offset) {
   const ASTRecordLayout &Layout = CGM.getContext().getASTRecordLayout(RD);
 
-  // Collect any non-virtual bases.
+  // Collect any bases.
   SmallVector<const CXXRecordDecl *, 4> Bases;
   for (const CXXBaseSpecifier &Base : RD->bases()) {
-    if (!Base.isVirtual() && !Base.getType()->isDependentType())
+    if (!Base.getType()->isDependentType())
       Bases.push_back(Base.getType()->getAsCXXRecordDecl());
   }
 
@@ -2586,7 +2586,7 @@ bool CGMSHLSLRuntime::FindDispatchGridSemantic(const CXXRecordDecl *RD,
                             Layout.getBaseClassOffset(R);
                    });
 
-  // Check (non-virtual) bases
+  // Check bases in order
   for (const CXXRecordDecl *Base : Bases) {
     CharUnits BaseOffset = Offset + Layout.getBaseClassOffset(Base);
     if (FindDispatchGridSemantic(Base, SDGRec, BaseOffset))
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -2576,7 +2576,7 @@ bool CGMSHLSLRuntime::FindDispatchGridSemantic(const CXXRecordDecl *RD,
   for (const CXXBaseSpecifier &Base : RD->bases()) {
     DXASSERT(!Base.getType()->isDependentType(),
              "Node Record with dependent base class not caught by Sema");
-    if (Base.isVirtual() || Base.getType()->isDependentType())
+    if (Base.getType()->isDependentType())
       continue;
     CXXRecordDecl *BaseDecl = Base.getType()->getAsCXXRecordDecl();
     CharUnits BaseOffset = Offset + Layout.getBaseClassOffset(BaseDecl);
--- a/tools/clang/lib/CodeGen/CGHLSLMS.cpp
+++ b/tools/clang/lib/CodeGen/CGHLSLMS.cpp
@@ -288,6 +288,9 @@ public:
                                            llvm::Value *DestPtr,
                                            clang::QualType DestTy) override;
   void AddHLSLFunctionInfo(llvm::Function *, const FunctionDecl *FD) override;
+  bool FindDispatchGridSemantic(const CXXRecordDecl *RD,
+                                hlsl::SVDispatchGrid &SDGRec,
+                                CharUnits Offset = CharUnits());
   void AddHLSLNodeRecordTypeInfo(const clang::ParmVarDecl *parmDecl,
                                  hlsl::NodeIOProperties &node);
   void EmitHLSLFunctionProlog(llvm::Function *,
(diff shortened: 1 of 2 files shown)


tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2558,6 +2561,75 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+// We return true if we have populated the SV_DispatchGrid values.
+bool CGMSHLSLRuntime::FindDispatchGridSemantic(const CXXRecordDecl *RD,
+                                               hlsl::SVDispatchGrid &SDGRec,
+                                               CharUnits Offset) {
+  const ASTRecordLayout &Layout = CGM.getContext().getASTRecordLayout(RD);
+
+  // Collect any non-virtual bases.
+  SmallVector<const CXXRecordDecl *, 4> Bases;
+  for (const CXXBaseSpecifier &Base : RD->bases()) {
+    if (!Base.isVirtual() && !Base.getType()->isDependentType())
```
llvm-beanz, line 2576: I assume it isn't actually possible in HLSL to have virtual base classes right? The language certainly shouldn't allow them, so I think this check is probably unnecessary.

llvm-beanz (commented):
We should also make sure we have error detection in Sema for the case of multiple fields in a single or nested structure definition having SV_DispatchGrid applied. Since the CodeGen can only handle a single declaration, we should error if a structure has more than one annotated field.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tcorringham: No, virtual base classes aren't possible - but the dependent type is, so we do need that check here. This condition just follows the style and layout of similar checks elsewhere - the isVirtual() check could safely be removed.

tcorringham:
The presence of more than one SV_DispatchGrid field is already detected in Sema, and covers the cases of base classes and nested records.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
llvm-beanz: Yea, I think we should remove the `isVirtual` check since that will always be false and I can't imagine any world where we'd be adding virtual inheritance to HLSL...

tools/clang/test/HLSLFileCheck/hlsl/workgraph/nested_sv_dispatchgrid.hlsl
```diff
@@ -0,0 +1,78 @@
...
+  Record3 r;
+};
+
+[Shader("node")]
+[NodeLaunch("broadcasting")]
+[NodeMaxDispatchGrid(32,16,1)]
+[NumThreads(32,1,1)]
+void node5(DispatchNodeInputRecord<Record5> input) {}
+// CHECK: , i32 1, ![[SVDG_5:[0-9]+]]
+// CHECK: [[SVDG_5]] = !{i32 20, i32 5, i32 2}
```
llvm-beanz, line 78: What about some test cases with templates?

Something like:

```c++
template <typename T>
struct Base {
  T DG : SV_DispatchGrid;
};

struct Derived1 : Base<uint3> {
  int4 x;
};

[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeMaxDispatchGrid(32,16,1)]
[NumThreads(32,1,1)]
void node6(DispatchNodeInputRecord<Derived1 > input) {}

template <typename T>
struct Derived2 : Base<T> {
  T Y;
};

[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeMaxDispatchGrid(32,16,1)]
[NumThreads(32,1,1)]
void node7(DispatchNodeInputRecord<Derived2<uint2> > input) {}


template <typename T>
struct Derived3 {
  Derived2<T> V;
};

[Shader("node")]
[NodeLaunch("broadcasting")]
[NodeMaxDispatchGrid(32,16,1)]
[NumThreads(32,1,1)]
void node8(DispatchNodeInputRecord< Derived3 <uint3> > input) {}
```

tcorringham: Great idea! I've updated the test to include these cases.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tcorringham: Done.

damyanp:
@llvm-beanz - could you have a look at this with some urgency please (we want it in the upcoming release) when you're back?

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2558,6 +2561,75 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+    for (const hlsl::UnusualAnnotation *it : Field->getUnusualAnnotations()) {
+      if (it->getKind() == hlsl::UnusualAnnotation::UA_SemanticDecl) {
+        const hlsl::SemanticDecl *sd = cast<hlsl::SemanticDecl>(it);
+        if (sd->SemanticName.equals("SV_DispatchGrid")) {
+          const llvm::Type *FTy = CGM.getTypes().ConvertType(Field->getType());
+          const llvm::Type *ElTy = FTy;
+          SDGRec.NumComponents = 1;
+          SDGRec.ByteOffset = (unsigned)FieldOffset.getQuantity();
+          if (const llvm::VectorType *VT = dyn_cast<llvm::VectorType>(FTy)) {
+            SDGRec.NumComponents = VT->getNumElements();
```
pow2clk, line 2615: These used to assert it was an integer. was that invalid?

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2558,6 +2561,75 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+    }
+    // Otherwise check this field for the SV_DispatchGrid semantic annotation
+    for (const hlsl::UnusualAnnotation *it : Field->getUnusualAnnotations()) {
+      if (it->getKind() == hlsl::UnusualAnnotation::UA_SemanticDecl) {
+        const hlsl::SemanticDecl *sd = cast<hlsl::SemanticDecl>(it);
+        if (sd->SemanticName.equals("SV_DispatchGrid")) {
+          const llvm::Type *FTy = CGM.getTypes().ConvertType(Field->getType());
+          const llvm::Type *ElTy = FTy;
+          SDGRec.NumComponents = 1;
+          SDGRec.ByteOffset = (unsigned)FieldOffset.getQuantity();
```
pow2clk, line 2613: The removal of the struct check is because of the above that iterates into any records?

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2558,6 +2561,75 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+    if (!Base.getType()->isDependentType())
+      Bases.push_back(Base.getType()->getAsCXXRecordDecl());
+  }
+
+  // Sort bases by offset.
+  std::stable_sort(Bases.begin(), Bases.end(),
+                   [&](const CXXRecordDecl *L, const CXXRecordDecl *R) {
+                     return Layout.getBaseClassOffset(L) <
+                            Layout.getBaseClassOffset(R);
+                   });
```
pow2clk, line 2585: What does ordering the bases give us? I could imagine this might be intended to prevent duplicate testing, but If there are multiple levels of bases, it seems this results in checking the highest parent base once for each descendent.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2558,6 +2561,75 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+// expect valid size and types, and use the first occurance found.
+// We return true if we have populated the SV_DispatchGrid values.
+bool CGMSHLSLRuntime::FindDispatchGridSemantic(const CXXRecordDecl *RD,
+                                               hlsl::SVDispatchGrid &SDGRec,
+                                               CharUnits Offset) {
+  const ASTRecordLayout &Layout = CGM.getContext().getASTRecordLayout(RD);
+
+  // Collect any bases.
+  SmallVector<const CXXRecordDecl *, 4> Bases;
+  for (const CXXBaseSpecifier &Base : RD->bases()) {
```
pow2clk, line 2575: If I understand this right, parent classes will be checked for every descendent and every struct that has a common ancestor. Maybe that's hard to avoid?

pow2clk (commented):
Looks good! I think the parts I'm most concerned about are the repeated checks of parent classes and that I might not understand the motive for sorting base offsets. The other stuff is minor.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tcorringham: The types are checked in Sema so we won't get this far if the type is wrong.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tcorringham: Yes, we iterate into records above.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tcorringham: We need to order the bases so we can keep track of the offset from the start of the input record.
If a record type appears more than once then it can't contain a field with SV_DispatchGrid as that would be errored in Sema, so we could track which records we have already checked and skip checking the fields of any we have seen before. However, I'm not sure that such cases would occur often enough to make the extra housekeeping worthwhile - so currently we may check records more than once.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
tcorringham: Yes, as per my comment above I don't expect such cases to occur often, so I believe the extra housekeeping probably isn't worthwhile.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2558,6 +2561,75 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+                     return Layout.getBaseClassOffset(L) <
+                            Layout.getBaseClassOffset(R);
+                   });
+
+  // Check bases in order
+  for (const CXXRecordDecl *Base : Bases) {
+    CharUnits BaseOffset = Offset + Layout.getBaseClassOffset(Base);
+    if (FindDispatchGridSemantic(Base, SDGRec, BaseOffset))
+      return true;
+  }
```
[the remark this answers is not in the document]
tcorringham: 1. Actually I've rethought this - I didn't allow for the HLSL restriction of only one concrete base class, so I now think my code is overly complex, and bases don't need to be sorted by offset at all.
2. I think so, but couldn't convince myself 100% so left the check in.
3. I didn't know that.
I think your suggestion is actually the right thing here.

pow2clk (approved):
I don't mind the unneeded virtual check, but perhaps @llvm-beanz feels differently.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2570,24 +2570,15 @@ bool CGMSHLSLRuntime::FindDispatchGridSemantic(const CXXRecordDecl *RD,
...
-                            Layout.getBaseClassOffset(R);
-                   });
-
-  // Check bases in order
-  for (const CXXRecordDecl *Base : Bases) {
-    CharUnits BaseOffset = Offset + Layout.getBaseClassOffset(Base);
-    if (FindDispatchGridSemantic(Base, SDGRec, BaseOffset))
+    DXASSERT(!Base.getType()->isDependentType(),
+             "Node Record with dependent base class not caught by Sema");
+    if (Base.isVirtual() || Base.getType()->isDependentType())
```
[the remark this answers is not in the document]
tcorringham: I had removed this unnecessary check, but it looks as if it was accidentally reinstated with a subsequent commit. I have removed it again.

tools/clang/lib/CodeGen/CGHLSLMS.cpp
```diff
@@ -2560,6 +2563,66 @@ void CGMSHLSLRuntime::AddHLSLFunctionInfo(Function *F, const FunctionDecl *FD) {
...
+    // If this field is a record check its fields
+    if (const CXXRecordDecl *D = Field->getType()->getAsCXXRecordDecl()) {
+      if (FindDispatchGridSemantic(D, SDGRec, FieldOffset))
+        return true;
+    }
+    // Otherwise check this field for the SV_DispatchGrid semantic annotation
+    for (const hlsl::UnusualAnnotation *it : Field->getUnusualAnnotations()) {
+      if (it->getKind() == hlsl::UnusualAnnotation::UA_SemanticDecl) {
+        const hlsl::SemanticDecl *sd = cast<hlsl::SemanticDecl>(it);
+        if (sd->SemanticName.equals("SV_DispatchGrid")) {
```
tex3d, line 2602: Let's follow coding standard for naming.  Here's a suggested minimal change:
```suggestion
    for (const hlsl::UnusualAnnotation *UA : Field->getUnusualAnnotations()) {
      if (it->getKind() == hlsl::UnusualAnnotation::UA_SemanticDecl) {
        const hlsl::SemanticDecl *SD = cast<hlsl::SemanticDecl>(UA);
        if (SD->SemanticName.equals("SV_DispatchGrid")) {
```

tex3d (approved):
Looks good.  Only one minor nit on new variable names not following coding standard.

damyanp (approved):
I see two approvals that have been made stale by a change to some variable names.  So I'm going to reapprove and merge this.

Issue #6927, opened 2024-09-23
[Feature Request][SPIR-V] HLSL style binding assignment for array of resources


manas-kulkarni:
**Is your feature request related to a problem? Please describe.**
We have feature in engine that depends on hlsl style binding assignment for resource arrays. This doesn't work with the Vulkan style assignment where array resources takes only one binding

**Describe the solution you'd like**
-fspv-hlsl-resource-array-bindings
- Resource array will take up [array-size] number of bindings when doing implicit binding assignment

s-perron:
I think we already have a feature like that:

>   -fspv-flatten-resource-arrays
                          Flatten arrays of resources so each array element takes one binding number

Does that work for you? If there are bug fixes to that feature, we might be able help fix those. If you require something very different, we are very unlikely to implement it in DXC.

s-perron:
A workaround was identified in #6932. You can use `-fspv-flatten-resource-arrays` with `-fcgl`. Then you can run the legalization and optimization passes separately. If you are using the command line, it would look somethink like this:

```
dxc -T cs_6_6 -spirv t.hlsl -fspv-flatten-resource-arrays -fcgl -Fo t.spv && \
spirv-opt t.spv -o o.spv --trim-capabilities --legalize-hlsl --descriptor-composite-scalar-replacement -O
```

Please post any change if the optimization passes do not work for you.

PR #6932, opened 2024-09-25
[SPIR-V] HLSL style register assignment for resource arrays

manas-kulkarni:
- New spirv flag -fspv-hlsl-resource-arrays
- When the flag is specified, resource arrays will use one binding per array element
  - We need this change to support array textures in our engine
- Matches with hlsl style register where arrays take up one register per array element
- Added test shader for the clang tests

```
Texture2D MyTextures[5];
Texture2D AnotherTexture;
```
MyTextures will take 5 bindings, AnotherTexture will be at binding 5

Closes https://github.com/microsoft/DirectXShaderCompiler/issues/6927

### DIFF
--- a/include/dxc/Support/HLSLOptions.td
+++ b/include/dxc/Support/HLSLOptions.td
@@ -390,8 +390,8 @@ def fspv_target_env_EQ : Joined<["-"], "fspv-target-env=">, Group<spirv_Group>, 
   HelpText<"Specify the target environment: vulkan1.0 (default), vulkan1.1, vulkan1.1spirv1.4, vulkan1.2, vulkan1.3, or universal1.5">;
 def fspv_flatten_resource_arrays: Flag<["-"], "fspv-flatten-resource-arrays">, Group<spirv_Group>, Flags<[CoreOption, DriverOption]>,
   HelpText<"Flatten arrays of resources so each array element takes one binding number">;
-def fspv_implicit_hlsl_resource_arrays: Flag<["-"], "fspv-implicit-hlsl-resource-arrays">, Group<spirv_Group>, Flags<[CoreOption, DriverOption]>,
-  HelpText<"Resource arrays with implicit register assignment use one binding per array element to match hlsl style">;
+def fspv_hlsl_resource_arrays: Flag<["-"], "fspv-hlsl-resource-arrays">, Group<spirv_Group>, Flags<[CoreOption, DriverOption]>,
+  HelpText<"Resource arrays use one binding per array element to match hlsl style">;
 def fspv_reduce_load_size: Flag<["-"], "fspv-reduce-load-size">, Group<spirv_Group>, Flags<[CoreOption, DriverOption]>,
   HelpText<"Replaces loads of composite objects to reduce memory pressure for the loads">;
 def fspv_fix_func_call_arguments: Flag<["-"], "fspv-fix-func-call-arguments">, Group<spirv_Group>, Flags<[CoreOption, DriverOption, HelpHidden]>,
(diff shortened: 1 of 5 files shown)


manas-kulkarni:

include/dxc/Support/HLSLOptions.td
```diff
@@ -390,6 +390,8 @@ def fspv_target_env_EQ : Joined<["-"], "fspv-target-env=">, Group<spirv_Group>,
   HelpText<"Specify the target environment: vulkan1.0 (default), vulkan1.1, vulkan1.1spirv1.4, vulkan1.2, vulkan1.3, or universal1.5">;
 def fspv_flatten_resource_arrays: Flag<["-"], "fspv-flatten-resource-arrays">, Group<spirv_Group>, Flags<[CoreOption, DriverOption]>,
   HelpText<"Flatten arrays of resources so each array element takes one binding number">;
+def fspv_hlsl_resource_arrays: Flag<["-"], "fspv-hlsl-resource-arrays">, Group<spirv_Group>, Flags<[CoreOption, DriverOption]>,
+  HelpText<"Resource arrays use one binding per array element to match hlsl style">;
```
s-perron, line 394: This is not an accurate description of what is happening. In Vulkan, an array of resources take just a single binding. In the test below, `MyTextures[5]` will still use a single binding. And then binding 1-4 are left unused.

Also the name of option is not very descriptive. What is an HLSL resource array? You should reference something in DirectX, you seem to want to try to match DX's binding model in someway. Even that I believe is very difficult because   Vulkan's binding model is very different.

s-perron (commented):
This looks like it is adding a new feature to DXC, and not a fix to an existing feature. We may need to discuss this more.

The first issue is that I'm concerned this does not do what you think it does.

Second, I want to make sure a solution does not already exist for what you want to do. Without knowing you exact workflow, I do not know what to suggest. Here is my guess.

Some developers use `-fspv-flatten-resource-arrays` to get resource arrays where each element actually has a different binding. If that is want is important you, that is how you do it. I believe this came from a time when [descriptor indexing](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_descriptor_indexing.html) was not commonly available. 

Other use [explicit binding](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#explicit-binding-number-assignment), so they know where each resource will be placed.

The last possibility is to use [spir-v reflect](https://github.com/KhronosGroup/SPIRV-Reflect). This is what I think you want to do. I'm totally guessing, but I'm guessing you use implicit binding for DX. You get the register numbers from reflection, and you want reusing the same numbers for the Vulkan bindings. If that is the case, I would suggest not using the register numbers from the DXIL, but to run the SPIR-V through spir-reflect to get the binding numbers that are Vulkan specific. The binding is different enough that you will keep running into problem using the DXIL register numbers. For example, in Vulkan we have to have a different binding for the counter buffers that correspond to RWStructuredBuffers. You cannot match the dxil binding numbers in these cases, so you will still have to do reflection on the SPIR-V.

Let me know you work flow, and we can try to figure out a solution.

manas-kulkarni:
We were able to get same functionality with -fspv-flatten-resource-arrays and -fcgl. Passing -fcgl skips the pass where separate bindings are created per array element

manas-kulkarni:
So if the -fcgl in future also creates separate resources per array element, we would need this separate flag though. In current version, only the binding numbers are used per array element and the resources themselves stay as arrays.
If this behavior change in future, we would need the -fspv-dx/hlsl-resource-arrays option

s-perron:
> So if the -fcgl in future also creates separate resources per array element, we would need this separate flag though. In current version with ```-fspv-flatten-resource-arrays + -fcgl```, only the binding numbers are used per array element and the resources themselves stay as arrays.
> If this behavior change in future, we would need the -fspv-dx/hlsl-resource-arrays option


Fcgl turns off all optimizations so that we can test what the spirv is before spirv-opt does anything. You may need to run the optimizer separately to generate valid Spir-V.

manas-kulkarni:
Yes. We run spirv-opt in separate pass. Were wondering if the behavior of Fcgl will ever change and it does include the spirv passes that create new variables per array element, we would need a different flag to perform the current behavior of ```fspv-flatten-resource-arrays + fcgl```

s-perron:
`-fcgl` will not change.

PR #6942, opened 2024-10-01
Fix DxilPayloadFieldAnnotation::GetPayloadFieldQualifier

### LINKED ISSUE #6941: DXIL PAQ metadata helpers report wrong information
**Description**
Attempting to use `DxilPayloadFieldAnnotation::GetPayloadFieldQualifier` returns `ReadWrite` for a field which was not annotated to be read *or* written by a given shader stage.

**Steps to Reproduce**
In the (still pending) PAQ test https://github.com/microsoft/DirectXShaderCompiler/pull/5487, the `color_write_AH` field is not annotated to be read or written by the `closesthit` shader stage. However, after loading that shader and using `GetPayloadFieldQualifier` on field 4 of the payload structure with the `PayloadAccessShaderStage::Closesthit` value, the result is `ReadWrite`.

The bug is clear:
https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/DXIL/DxilTypeSystem.cpp#L176
```c++
  // default type is always ReadWrite
  DXIL::PayloadAccessQualifier accessType =
      DXIL::PayloadAccessQualifier::ReadWrite;

  const unsigned readBit =
      static_cast<unsigned>(DXIL::PayloadAccessQualifier::Read);
  const unsigned writeBit =
      static_cast<unsigned>(DXIL::PayloadAccessQualifier::Write);

  unsigned accessBits = m_bitmask >> bitOffset;
  if (accessBits & readBit) {
    // set Read if the first bit is set
    accessType = DXIL::PayloadAccessQualifier::Read;
  }
  if (accessBits & writeBit) {

    // set Write only if the second bit set, if both are set set to ReadWrite
    accessType = accessType == DXIL::PayloadAccessQualifier::ReadWrite
                     ? DXIL::PayloadAccessQualifier::Write
                     : DXIL::PayloadAccessQualifier::ReadWrite;
  }
  return accessType;
```

If neither `readBit` nor `writeBit` are set, the 'default' of `ReadWrite` is returned, instead of the correct default of `NoAccess`. I presume that this is to handle a field which has no annotations at all. However that should be treated as a special case, perhaps with an early return.

A much simpler implementation that correctly handles `NoAccess` would look like:
```c++
  if (!HasAnnotations())
    return DXIL::PayloadAccessQualifier::ReadWrite;

  int bitOffset = GetBitOffsetForShaderStage(shaderStage);
  unsigned accessBits = m_bitmask >> bitOffset;
  return (DXIL::PayloadAccessQualifier)(accessBits & DXIL::PayloadAccessQualifierValidMaskPerStage);
```

jenatali:
Fixes #6941

### DIFF
--- a/lib/DXIL/DxilTypeSystem.cpp
+++ b/lib/DXIL/DxilTypeSystem.cpp
@@ -175,31 +175,13 @@ void DxilPayloadFieldAnnotation::AddPayloadFieldQualifier(
 DXIL::PayloadAccessQualifier
 DxilPayloadFieldAnnotation::GetPayloadFieldQualifier(
     DXIL::PayloadAccessShaderStage shaderStage) const {
+  if (!HasAnnotations())
+    return DXIL::PayloadAccessQualifier::ReadWrite;
 
   int bitOffset = GetBitOffsetForShaderStage(shaderStage);
-
-  // default type is always ReadWrite
-  DXIL::PayloadAccessQualifier accessType =
-      DXIL::PayloadAccessQualifier::ReadWrite;
-
-  const unsigned readBit =
-      static_cast<unsigned>(DXIL::PayloadAccessQualifier::Read);
-  const unsigned writeBit =
-      static_cast<unsigned>(DXIL::PayloadAccessQualifier::Write);
-
   unsigned accessBits = m_bitmask >> bitOffset;
-  if (accessBits & readBit) {
-    // set Read if the first bit is set
-    accessType = DXIL::PayloadAccessQualifier::Read;
-  }
-  if (accessBits & writeBit) {
-
-    // set Write only if the second bit set, if both are set set to ReadWrite
-    accessType = accessType == DXIL::PayloadAccessQualifier::ReadWrite
-                     ? DXIL::PayloadAccessQualifier::Write
-                     : DXIL::PayloadAccessQualifier::ReadWrite;
-  }
-  return accessType;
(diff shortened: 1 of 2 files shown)


damyanp (approved):
LGTM - I can't say I fully understand exactly why the simplified code is equivalent.  Have you confirmed that there's any existing tests at all that are validating that it at least sometimes did what it was supposed to do?

jenatali:
The simplified code is *not* equivalent, this isn't just a simplification, it's also a bugfix.

According to the [DXR spec](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#semantics):
`<none>` means "The payload field is neither available for reading nor are its contents modified by the given stage."

That would map to the `NoAccess` value. The existing test for this metadata helper only had one payload field, which had every stage represented as either read, write, or read-write. There was no stage which was completely omitted from the annotations. By adding the second payload field which omits closesthit/anyhit, I can verify that the old code *incorrectly* returned `ReadWrite` for those stages.

The old code was correct as long as the stage had either a read or write bit set, or both, it was only incorrect in that it returned `ReadWrite` instead of `NoAccess`. It was just... very verbose and therefore confusing in the way that it calculated the return value.

damyanp:
Oops, I misspoke by "equivalent"!  The nuance I was trying to get at was that presumably there are some cases where the new code would be expected to return the same results as before, and that there is some test coverage that continues to validate that.

tex3d (approved):
Looks good.

PR #6943, opened 2024-10-02
Unwrap top-level array for OutVertices when flattenArgument.

### LINKED ISSUE #6940: Regression when using SV_Position with a mesh shader
**Description**

This regression was introduced fairly recently, but unsure when

When using SV_Position as the only output for vertices, it will fail compilation with:
error: validation errors
error: DXIL container mismatch for 'SigOutputElement' between 'PSV0' part:('PSVSignatureElement:

If the output is changed to a struct with SV_Position as one of the members, it works fine.

**Steps to Reproduce**

https://godbolt.org/z/7vEoEjzMf

**Actual Behavior**

Crash in validation

**Environment**
- DXC version (latest)
- Host Operating System (any)

python3kgae:
For primitives and vertices output of mesh shader, we need to unwrap the top-level array to get correct semantic index.

Fixes #6940

### DIFF
--- a/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
+++ b/lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
@@ -5300,7 +5300,9 @@ void SROA_Parameter_HLSL::flattenArgument(
     // Unwrap top-level array if primitive
     if (inputQual == DxilParamInputQual::InputPatch ||
         inputQual == DxilParamInputQual::OutputPatch ||
-        inputQual == DxilParamInputQual::InputPrimitive) {
+        inputQual == DxilParamInputQual::InputPrimitive ||
+        inputQual == DxilParamInputQual::OutPrimitives ||
+        inputQual == DxilParamInputQual::OutVertices) {
       Type *Ty = Arg->getType();
       if (Ty->isPointerTy())
         Ty = Ty->getPointerElementType();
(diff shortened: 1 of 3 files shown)


damyanp (commented):
We should look for a review from someone more familiar with this part of the compiler than I am.

Is this actually a pre-existing bug that the updated validation code now catches?

Or is this fixing a regression introduced as part of updating the validation code?

python3kgae:
> We should look for a review from someone more familiar with this part of the compiler than I am.
> 
> Is this actually a pre-existing bug that the updated validation code now catches?
> 
> Or is this fixing a regression introduced as part of updating the validation code?

This is pre-existing bug that the updated validation code now catches.
The shader will hit assert in debug build before validation.

bob80905 (approved):
LGTM, other than `tools/clang/test/HLSLFileCheck/shader_targets/mesh/semantic_on_parameter.hlsl` needs a new line at EOF

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -5300,7 +5300,9 @@ void SROA_Parameter_HLSL::flattenArgument(
     // Unwrap top-level array if primitive
     if (inputQual == DxilParamInputQual::InputPatch ||
         inputQual == DxilParamInputQual::OutputPatch ||
-        inputQual == DxilParamInputQual::InputPrimitive) {
+        inputQual == DxilParamInputQual::InputPrimitive ||
```
bfavela, line 5303: are you missing a testcase for this qualifier?

bfavela (approved):
LGTM after adding missing testcase (if needed)

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
python3kgae: Added.

PR #6945, opened 2024-10-04
[SPIR-V] Implement EvaluateAttribute* functions

### LINKED ISSUE #3649: [SPIR-V] Support EvaluateAttribute* intrinsics
Currently, SPIR-V backend does not support `EvaluateAttribute*` intrinsics.

According to the discussion in https://github.com/KhronosGroup/glslang/issues/2584, `OpInterpolate*` needs a pointer operand rather than `OpLoad`. Based on the offline discussion with @greg-lunarg we need a spirv-opt pass to replace `OpLoad` operand of `OpInterpolate*` to the pointer loaded by the `OpLoad`.

When a spirv-opt pass is prepared, we can simply let DXC emit `OpInterpolate*` for `EvaluateAttribute*`. Since it will just follow typical code generation routine, it will use `OpLoad` for its operand. We can rely on the spirv-opt pass to make it correct (as a part of the legalization pass).

cassiebeckley:
Emits SPIR-V InterpolateAt* instructions from the GLSL.std.450 extended instruction set for EvaluateAttribute*. Relies on the optimizer's copy propagate passes to ensure that these instructions are passed valid pointers in the Input storage class after legalization.

Fixes #3649


tools/clang/test/CodeGenSPIRV/intrinsics.evaluate-attribute-at-sample.hlsl
```diff
@@ -0,0 +1,44 @@
...
+	a[4] = float4(EvaluateAttributeAtSample(gl_PointCoord, 4), 0, 0);
+}
+
+// CHECK: %setValue = OpFunction
+void setValue(float4 input) {
+// CHECK: [[val:%[0-9]+]] = OpLoad %v4float %input
+// CHECK:                   OpStore %temp_var_v4float_1 [[val]]
+// CHECK:                   OpExtInst %v4float [[instSet]] InterpolateAtSample %temp_var_v4float_1 %uint_1
+	a[1] = EvaluateAttributeAtSample(input, 1);
+}
```
s-perron, line 44: Missing new line at the end of the file.

tools/clang/test/CodeGenSPIRV/intrinsics.evaluate-attribute-centroid.hlsl
```diff
@@ -0,0 +1,45 @@
...
+}
+
+
+// CHECK: %setValue = OpFunction
+void setValue(float4 input) {
+// CHECK: [[val:%[0-9]+]] = OpLoad %v4float %input
+// CHECK:                   OpStore %temp_var_v4float_1 [[val]]
+// CHECK:                   OpExtInst %v4float [[instSet]] InterpolateAtCentroid %temp_var_v4float_1
+	a[1] = EvaluateAttributeCentroid(input);
+}
```
s-perron, line 45: New line at the end of the file.

tools/clang/test/CodeGenSPIRV/intrinsics.evaluate-attribute-snapped.hlsl
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// CHECK: %setValue = OpFunction
+void setValue(float4 input) {
+// CHECK:    [[val:%[0-9]+]] = OpLoad %v4float %input
+// CHECK:                      OpStore %temp_var_v4float_1 [[val]]
+// CHECK: [[offset:%[0-9]+]] = OpConvertSToF %v2float [[v2int_1_1]]
+// CHECK:                      OpExtInst %v4float [[instSet]] InterpolateAtOffset %temp_var_v4float_1 [[offset]]
+	a[1] = EvaluateAttributeSnapped(input, int2(1,1));
+}
```
s-perron, line 55: new line at the end of the file.

s-perron (approved):
LGTM. You are missing the end-of-line at the end of a few files.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11939,6 +11945,44 @@ SpirvInstruction *SpirvEmitter::processIntrinsicUsingGLSLInst(
   return nullptr;
 }
 
+SpirvInstruction *SpirvEmitter::processEvaluateAttributeAt(
+    const CallExpr *callExpr, hlsl::IntrinsicOp opcode, SourceLocation loc,
+    SourceRange range) {
```
amaiorano, line 11950: Should `loc` and `range` should be const refs? So like: `const SourceLocation& loc, const SourceRange& range`

s-perron: The standard idiom in LLVM is to pass them by value. SourceLocation is a class with just a single member, so that can be passed in a register. SourceRange is two SourceLocations, so I believe it can be passed in a register as well. However, that depends on the ABI.

amaiorano: ACK.

amaiorano (approved):

PR #6950, opened 2024-10-08
PIX: Implement shader access tracking for descriptor-heap-indexed TLAS for TraceRay

jeffnn:
Simple missing case, should have implemented it in the first place.
Also noticed that from a recent change to move to raw buffer writes, the debug output UAV offset was being set to a non-dword aligned offset. No drivers/hardware seem to care, but it's a concerning thing to leave as-is.


bob80905 (approved):

tex3d (approved):
Looks good.

PR #6951, opened 2024-10-08
[SPIR-V] Allow built-in object types as arguments to SpirvType

### LINKED ISSUE #6498: [SPIR-V] Fix bug using `Texture2D` with alias template for `SpirvType`
**Description**
For the (currently unmerged) `SpirvType` PR, `Texture2D` (and presumably other built-in types) cannot be used as an argument to an alias template for `SpirvType`

**Steps to Reproduce**

Compile

```hlsl
template<typename SomeType, uint Length>
using Array = vk::SpirvOpaqueType</* OpTypeArray */ 28, SomeType, vk::integral_constant<uint, Length> >;

void main() {
  Array<Texture2D, 4> image;
}

```

With the command

```
dxc -T ps_6_0 -E main -spirv
```

**Actual Behavior**
Error:

```
foo.hlsl:2:1: error: 'Texture2D<vector<float, 4> >' is an object and cannot be used as a type parameter
using Array = vk::SpirvOpaqueType</* OpTypeArray */ 28, SomeType, vk::integral_constant<uint, Length> >;
^
foo.hlsl:5:3: note: in instantiation of template type alias 'Array' requested here
  Array<Texture2D, 4> image;
  ^
```

**Environment**
- DXC version `libdxcompiler.so: 1.8(dev;4545-a97b48a5)` (development version of #6156)
- Host Operating System Linux

cassiebeckley:
There has been a validation error for passing in a built-in type to a template since the first commit to DXC. This check exists to prevent instances like this:

```
Texture2D<SamplerState> image;
```

When DXC was first created, users could not create their own templates, and therefore having a generic check for this did not cause any problems. However, now that HLSL 2021 allows users to create their own templates, and inline SPIR-V lets users pass types in to the `SpirvType` and `SpirvOpaqueType` templates in order to reference them, this is causing problems.

This PR allows the `SpirvType` and `SpirvOpaqueType` templates to take HLSL built-in types as arguments.

Fixes #6498

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -4904,7 +4904,7 @@ public:
   }
 
   bool IsValidTemplateArgumentType(SourceLocation argLoc, const QualType &type,
-                                   bool requireScalar) {
+                                   bool requireScalar, bool allowObject) {
     if (type.isNull()) {
       return false;
     }
@@ -4929,7 +4929,7 @@ public:
       if (qt->isArrayType()) {
         const ArrayType *arrayType = qt->getAsArrayTypeUnsafe();
         return IsValidTemplateArgumentType(argLoc, arrayType->getElementType(),
-                                           false);
+                                           false, allowObject);
       } else if (objectKind == AR_TOBJ_VECTOR) {
         bool valid = true;
         if (!IsValidVectorSize(GetHLSLVecSize(type))) {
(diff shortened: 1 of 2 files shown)


llvm-beanz (commented):
I think we need to have some discussion about this.

Texture types should really only take scalar or vectors. The fact that they ever allowed anything else is a bug.

They do actually work with template objects today (see: https://godbolt.org/z/zPq1T4q3b). The change you're making here allows them to work with HLSL's built-in objects (resources and the like), which absolutely should not be supported.

I'd like to understand better the case where these types should be allowed so that we can better craft language rules around them.

devshgraphicsprogramming:
For me it would just be enough to allow `SpirvType` to take another `SpirvType` AND be able to obtain the `SpirvType` (or some link) to a compound object like `groupshared T` or `RWStructuredBuffer<T>`

Stuffing SpirTypes into HLSL templates is probably a road to many bugs, and the most important feature is being able to build SpirvTypes from Native **and SpirV Types** because that way we can build any type we want/need.

s-perron:
> Texture types should really only take scalar or vectors. The fact that they ever allowed anything else is a bug.

This should be be allowing more type as templates into the standard Texture types. If it does, that is unintentional.

We want to allow more types as templates to the `vk::Spirv*Type`. We were thinking we had a use case for allowing textures as a template to be able to define combined texture samplers, but that will not work. We have other issues trying to to implement combined texture samplers as inline spir-v.

However, we should still be able to use a SpirvType as a template to a SpirvType: https://godbolt.org/z/Pqh6nYchf.

PR #6953, opened 2024-10-09
[SPIR-V] Handle vectors passed to asuint

### LINKED ISSUE #6735: [SPIR-V] asuint fails on bit cast
The following code works when targeting DXIL and fails to compile when targeting SPIRV
SPIRV:
```bash
DXC_Debug_BUILD/bin/dxc scratch/asuint_spirv_test.hlsl -T lib_6_8 -enable-16bit-types -spirv -fspv-target-env=universal1.5 -O0 
```
```hlsl
export uint4 fn(double4 p1, uint4 p2, uint4 p3) {
    asuint(p1, p2, p3);
    return p2;
}
```
error
```bash
fatal error: generated SPIR-V is invalid: Expected input to have the same total bit width as Result Type: Bitcast
  %16 = OpBitcast %v2uint %14
```

DXIL:
```bash
DXC_Debug_BUILD/bin/dxc scratch/asuint_spirv_test.hlsl -T lib_6_8 -enable-16bit-types -O0
```

```llvm
target datalayout = "e-m:e-p:32:32-i1:32-i8:8-i16:16-i32:32-i64:64-f16:16-f32:32-f64:64-n8:16:32:64"
target triple = "dxil-ms-dx"

%dx.types.splitdouble = type { i32, i32 }

@dx.nothing.a = internal constant [1 x i32] zeroinitializer

; Function Attrs: nounwind
define <4 x i32> @"\01?fn@@YA?AV?$vector@I$03@@V?$vector@N$03@@V1@1@Z"(<4 x double> %p1, <4 x i32> %p2, <4 x i32> %p3) #0 {
  %1 = extractelement <4 x double> %p1, i64 0
  %2 = call %dx.types.splitdouble @dx.op.splitDouble.f64(i32 102, double %1)  ; SplitDouble(value)
  %3 = extractvalue %dx.types.splitdouble %2, 0
  %4 = insertelement <4 x i32> undef, i32 %3, i64 0
  %5 = extractelement <4 x double> %p1, i64 1
  %6 = call %dx.types.splitdouble @dx.op.splitDouble.f64(i32 102, double %5)  ; SplitDouble(value)
  %7 = extractvalue %dx.types.splitdouble %6, 0
  %8 = insertelement <4 x i32> %4, i32 %7, i64 1
  %9 = extractelement <4 x double> %p1, i64 2
  %10 = call %dx.types.splitdouble @dx.op.splitDouble.f64(i32 102, double %9)  ; SplitDouble(value)
  %11 = extractvalue %dx.types.splitdouble %10, 0
  %12 = insertelement <4 x i32> %8, i32 %11, i64 2
  %13 = extractelement <4 x double> %p1, i64 3
  %14 = call %dx.types.splitdouble @dx.op.splitDouble.f64(i32 102, double %13)  ; SplitDouble(value)
  %15 = extractvalue %dx.types.splitdouble %14, 0
  %16 = insertelement <4 x i32> %12, i32 %15, i64 3
  %17 = load i32, i32* getelementptr inbounds ([1 x i32], [1 x i32]* @dx.nothing.a, i32 0, i32 0)
  ret <4 x i32> %16
}

; Function Attrs: nounwind readnone
declare %dx.types.splitdouble @dx.op.splitDouble.f64(i32, double) #1

attributes #0 = { nounwind }
attributes #1 = { nounwind readnone }

!llvm.ident = !{!0}
!dx.version = !{!1}
!dx.valver = !{!1}
!dx.shaderModel = !{!2}
!dx.entryPoints = !{!3}

!0 = !{!"dxc(private) 1.8.0.14649 (main, 206133c9e)"}
!1 = !{i32 1, i32 8}
!2 = !{!"lib", i32 6, i32 8}
!3 = !{null, !"", null, null, !4}
!4 = !{i32 0, i64 8388612}
```

cassiebeckley:
`asuint` should be able to take vectors in addition to scalar values. Previously, it would be lowered as a bitcast from the input value to a vector of uints with a width of 2, which is not large enough if the input value is larger than a scalar value. In order to handle, for example, an input value that is a `double4`, we instead perform a component-wise bitcast.

Fixes #6735

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -11415,6 +11415,79 @@ SpirvEmitter::processIntrinsicAllOrAny(const CallExpr *callExpr,
   return nullptr;
 }
 
+void SpirvEmitter::splitDouble(SpirvInstruction *value,
+                               SpirvInstruction *&lowbits,
+                               SpirvInstruction *&highbits, SourceLocation loc,
+                               SourceRange range) {
+  const QualType uintType = astContext.UnsignedIntTy;
+  const QualType uintVec2Type = astContext.getExtVectorType(uintType, 2);
+
+  SpirvInstruction *uints = spvBuilder.createUnaryOp(
+      spv::Op::OpBitcast, uintVec2Type, value, loc, range);
+
+  lowbits = spvBuilder.createCompositeExtract(uintType, uints, {0}, loc, range);
+  highbits =
+      spvBuilder.createCompositeExtract(uintType, uints, {1}, loc, range);
+}
+
+void SpirvEmitter::splitDoubleVector(QualType elemType, uint32_t count,
+                                     QualType outputType,
+                                     SpirvInstruction *value,
+                                     SpirvInstruction *&lowbits,
+                                     SpirvInstruction *&highbits,
+                                     SourceLocation loc, SourceRange range) {
+  llvm::SmallVector<SpirvInstruction *, 4> lowElems;
+  llvm::SmallVector<SpirvInstruction *, 4> highElems;
+
+  for (uint32_t i = 0; i < count; ++i) {
+    SpirvInstruction *elem =
+        spvBuilder.createCompositeExtract(elemType, value, {i}, loc, range);
+    SpirvInstruction *lowbitsResult = nullptr;
+    SpirvInstruction *highbitsResult = nullptr;
+    splitDouble(elem, lowbitsResult, highbitsResult, loc, range);
+    lowElems.push_back(lowbitsResult);
+    highElems.push_back(highbitsResult);
+  }
+
+  lowbits =
+      spvBuilder.createCompositeConstruct(outputType, lowElems, loc, range);
+  highbits =
+      spvBuilder.createCompositeConstruct(outputType, highElems, loc, range);
+}
+
+void SpirvEmitter::splitDoubleMatrix(QualType elemType, uint32_t rowCount,
+                                     uint32_t colCount, QualType outputType,
+                                     SpirvInstruction *value,
+                                     SpirvInstruction *&lowbits,
+                                     SpirvInstruction *&highbits,
+                                     SourceLocation loc, SourceRange range) {
+
+  llvm::SmallVector<SpirvInstruction *, 4> lowElems;
+  llvm::SmallVector<SpirvInstruction *, 4> highElems;
+
+  QualType colType = astContext.getExtVectorType(elemType, rowCount);
+
+  const QualType uintType = astContext.UnsignedIntTy;
+  const QualType outputColType =
+      astContext.getExtVectorType(uintType, rowCount);
+
+  for (uint32_t i = 0; i < colCount; ++i) {
+    SpirvInstruction *column =
+        spvBuilder.createCompositeExtract(colType, value, {i}, loc, range);
+    SpirvInstruction *lowbitsResult = nullptr;
+    SpirvInstruction *highbitsResult = nullptr;
+    splitDoubleVector(elemType, colCount, outputColType, column, lowbitsResult,
+                      highbitsResult, loc, range);
+    lowElems.push_back(lowbitsResult);
+    highElems.push_back(highbitsResult);
+  }
+
+  lowbits =
+      spvBuilder.createCompositeConstruct(outputType, lowElems, loc, range);
+  highbits =
+      spvBuilder.createCompositeConstruct(outputType, highElems, loc, range);
+}
+
 SpirvInstruction *
 SpirvEmitter::processIntrinsicAsType(const CallExpr *callExpr) {
   // This function handles the following intrinsics:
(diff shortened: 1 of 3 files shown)


cassiebeckley:
@s-perron lmk if this approach looks correct to you. If it does, I think I'll need to update dead code elimination again, since the case where a temporary variable is being used is having the `OpStore` removed.

tools/clang/test/CodeGenSPIRV/intrinsics.asuint.hlsl
```diff
@@ -76,11 +76,34 @@ void main() {
     double value;
     uint lowbits;
     uint highbits;
-// CHECK-NEXT:      [[value:%[0-9]+]] = OpLoad %double %value
-// CHECK-NEXT:  [[resultVec:%[0-9]+]] = OpBitcast %v2uint [[value]]
-// CHECK-NEXT: [[resultVec0:%[0-9]+]] = OpCompositeExtract %uint [[resultVec]] 0
+// CHECK-NEXT:  [[resultArr:%[0-9]+]] = OpBitcast %_ptr_Function__arr_uint_uint_2 %value
```
s-perron, line 79: You can't do this. A bitcast from one pointer type to another is not allowed in the logical addressing mode.

See https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_universal_validation_rules. In particular, the list of valid opcodes for which a pointer can be an operand.

cassiebeckley: Hm, I think in that case I can maybe do component-wise bitcasts for each member of the vector.

Is there a reason the validator doesn't catch this?

s-perron (commented):
Looks good. I would tests some more cases. I think they will work, but I'm not sure.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11519,23 +11592,40 @@ SpirvEmitter::processIntrinsicAsType(const CallExpr *callExpr) {
...
+    } else if (isVectorType(argType, &elemType, &rowCount)) {
+      splitDoubleVector(elemType, rowCount, arg1->getType(), value,
+                        lowbitsResult, highbitsResult, loc, range);
+    } else if (isMxNMatrix(argType, &elemType, &rowCount, &colCount)) {
+      splitDoubleMatrix(elemType, rowCount, colCount, arg1->getType(), value,
+                        lowbitsResult, highbitsResult, loc, range);
+    }
+
+    spvBuilder.createStore(lowbits, lowbitsResult, arg1loc, arg1range);
+    spvBuilder.createStore(highbits, highbitsResult, arg2loc, arg2range);
```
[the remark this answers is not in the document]
cassiebeckley: For `asuint(d, x, y)`, `arg1loc` is the location of the `x` argument, and `arg2loc` is the location of `y`. I'm using them because I was thinking of the `OpStore` for each argument as being part of the code to handle that argument, though come to think of it, that's not how we normally treat `out` parameters. I'll replace these with `loc` and `range`.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11519,23 +11592,40 @@ SpirvEmitter::processIntrinsicAsType(const CallExpr *callExpr) {
...
+
+    // TODO: handle matrices
```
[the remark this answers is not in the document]
cassiebeckley: Not anymore! Thanks for catching that.

tools/clang/lib/SPIRV/SpirvEmitter.h
```diff
@@ -1312,6 +1312,22 @@ class SpirvEmitter : public ASTConsumer {
   /// the Vulkan memory model capability has been added to the module.
   bool UpgradeToVulkanMemoryModelIfNeeded(std::vector<uint32_t> *module);
 
+  // TODO: docs
```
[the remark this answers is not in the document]
cassiebeckley: Done.

tools/clang/test/CodeGenSPIRV/intrinsics.asuint.hlsl
```diff
@@ -79,8 +79,105 @@ void main() {
...
+// CHECK-NEXT:      [[high2:%[0-9]+]] = OpCompositeExtract %uint [[resultVec2]] 1
+// CHECK-NEXT:        [[low:%[0-9]+]] = OpCompositeConstruct %v3uint [[low0]] [[low1]] [[low2]]
+// CHECK-NEXT:       [[high:%[0-9]+]] = OpCompositeConstruct %v3uint [[high0]] [[high1]] [[high2]]
+// CHECK-NEXT:                          OpStore %lowbits3 [[low]]
+// CHECK-NEXT:                          OpStore %highbits3 [[high]]
+    asuint(value3, lowbits3, highbits3);
+
+    double2x2 value2x2;
+    uint2x2 lowbits2x2;
+    uint2x2 highbits2x2;
```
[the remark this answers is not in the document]
cassiebeckley: Done.

s-perron (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11415,6 +11415,79 @@ SpirvEmitter::processIntrinsicAllOrAny(const CallExpr *callExpr,
   return nullptr;
 }
 
+void SpirvEmitter::splitDouble(SpirvInstruction *value,
+                               SpirvInstruction *&lowbits,
+                               SpirvInstruction *&highbits, SourceLocation loc,
+                               SourceRange range) {
```
Keenuts, line 11421: nit: output parameters should be last

tools/clang/lib/SPIRV/SpirvEmitter.cpp
```diff
@@ -11519,23 +11592,33 @@ SpirvEmitter::processIntrinsicAsType(const CallExpr *callExpr) {
...
+
+    if (isScalarType(argType)) {
+      splitDouble(value, lowbitsResult, highbitsResult, loc, range);
+    } else if (isVectorType(argType, &elemType, &rowCount)) {
+      splitDoubleVector(elemType, rowCount, arg1->getType(), value,
+                        lowbitsResult, highbitsResult, loc, range);
+    } else if (isMxNMatrix(argType, &elemType, &rowCount, &colCount)) {
+      splitDoubleMatrix(elemType, rowCount, colCount, arg1->getType(), value,
+                        lowbitsResult, highbitsResult, loc, range);
+    }
```
Keenuts, line 11617: shouldn't we have an unreachable/assert in an `else` clause ? Or assert that lowBitsResults & highBitsResult are not null?

Keenuts (approved):

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: Done.

tools/clang/lib/SPIRV/SpirvEmitter.cpp
cassiebeckley: Sure, I added `llvm_unreachable`.

PR #6954, opened 2024-10-10
Enable resource directory for DXC.

s-perron:
This change enables the resource directory for DXC. This will allow the HLSL standard headers to be accessed without using the `-I` option.


s-perron:
Things that still need to be done:

- [ ] Modify the cmake install to copy the header to the resource directory.
- [ ] Change the default path to something that makes sense for dxc.
- [ ] Manually test on a few different platforms to make sure it works.

s-perron:
@pow2clk @llvm-beanz 

Do you have any suggestions on how to implement this? I'm not sure how to proceed.  I don't know how to pass information from the driver into the library without changing the public interface.

PR #6965, opened 2024-10-14
Both blocks of the if statement are identical.

smurav:
Simplification of the code due to the fact that both blocks of the if statement are identical.


damyanp (approved):

bogner (approved):

smurav:
The check didn't fail, it was cancelled.

damyanp:
> The check didn't fail, it was cancelled.

I think it timed out...I'm re-running it.

smurav:
photo_2024-10-15_21-19-11

Issue #6939, opened 2024-09-29
[SPIR-V] Non semantic shader information issue (-fspv-debug=vulkan-with-source)


Goshido:
**Description**
I'm using _Vulkan + HLSL_ workflow. After editing shader in _RenderDoc_ and press compile there is a compilation error:

```txt
Running "D:/Development/VulkanSDK/1.3.290.0/Bin/dxc.exe" -T ps_6_0 -E PS -spirv -fspv-target-env=vulkan1.1 -Fo C:/Users/Goshido/AppData/Local/Temp/shader_output C:/Users/Goshido/AppData/Local/Temp/shader_input
error: missing entry point definition


Process crashed with code -2147467259.
Output file is 0 bytes
```

I also noticed that _SPIR-V_ blobs produced by _DXC v1.8.2407.10013_ `a2a220bc616ca28848ec0af77ad33a09f92fa43a` showed exactly the same compilation flag information in _RenderDoc_:

```bash
-T {hlsl_stage2}_6_0 -E {entry_point} -spirv -fspv-target-env={vulkan_ver}  -E PS -T ps_6_8 -HV 2021 -spirv 
-fvk-use-dx-layout -fspv-reduce-load-size -fspv-target-env=vulkan1.1 -enable-16bit-types -WX -I tools\editor\hlsl
-I tools\editor\include -I app\src\main\hlsl -I app\src\main\cpp\include\pbr -Od -fspv-debug=vulkan-with-source
-Fo tools\editor\editor-assets\shaders\blit.ps.spv -Qembed_debug
```

Now _SPIR-V_ blobs compiled by recent _DXC v1.8.2407.10091_ `75ff50caa046a054747ae15b5c1910a4c8aa1917` shows less information:

```bash
-T {hlsl_stage2}_6_0 -E {entry_point} -spirv -fspv-target-env={vulkan_ver}
```

**Steps to Reproduce**
1. Take the following shader code:

`blit_program.ps`:

```cpp
#include "blit_program.inc"
#include "color_space.ps"
#include "srgb_program.inc"


// 1.0H / 2.4H
#define INVERSE_GAMMA           4.16667e-1F

[[vk::constant_id ( CONST_INVERSE_GAMMA )]]
float const                     g_inverseGamma = INVERSE_GAMMA;

[[vk::binding ( BIND_IMAGE, SET_BLIT )]]
Texture2D<float32_t4>           g_image:        register ( t0 );

[...]

PR #6966, opened 2024-10-14
[SPIRV] Emit DebugFunction/Definition for both wrapper and real functions

SteveUrquhart:
https://github.com/microsoft/DirectXShaderCompiler/pull/6758 unfortunately caused a regression for shaders compiled without -fcgl.  This PR extends the original fix and corrects the regression.  We now emit DebugFunction and DebugFunctionDefinition pairs for both the wrapper and the main functions, and a DebugEntrypoint that points at the wrapper function.  This survives inlining.

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -1442,37 +1442,16 @@ void SpirvEmitter::doFunctionDecl(const FunctionDecl *decl) {
   // This will allow the entry-point name to be something like
   // myNamespace::myEntrypointFunc.
   std::string funcName = getFnName(decl);
-  std::string debugFuncName = funcName;
-
+  std::string srcFuncName = "src." + funcName;
   SpirvFunction *func = declIdMapper.getOrRegisterFn(decl);
+  SpirvFunction *srcFunc = nullptr;
 
   auto loc = decl->getLocStart();
   auto range = decl->getSourceRange();
   RichDebugInfo *info = nullptr;
   SpirvDebugFunction *debugFunction = nullptr;
+  SpirvDebugFunction *srcDebugFunction = nullptr;
   SpirvDebugInstruction *outer_scope = spvContext.getCurrentLexicalScope();
-  const auto &sm = astContext.getSourceManager();
-  if (spirvOptions.debugInfoRich && decl->hasBody()) {
-    const uint32_t line = sm.getPresumedLineNumber(loc);
-    const uint32_t column = sm.getPresumedColumnNumber(loc);
-    info = getOrCreateRichDebugInfo(loc);
-
-    auto *source = info->source;
-    // Note that info->scopeStack.back() is a lexical scope of the function
-    // caller.
-    auto *parentScope = info->compilationUnit;
-    // TODO: figure out the proper flag based on the function decl.
-    // using FlagIsPublic for now.
-    uint32_t flags = 3u;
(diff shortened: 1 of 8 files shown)


s-perron (approved):

s-perron:
Lots of tests are failing. Try running the spir-v tests: `ninja check-clang-codegenspirv`.

SteveUrquhart:
I'm sorry, I have one test left to fix.

SteveUrquhart:
@s-perron, the current failure in CI seems to be coming from an optimizer pass.  I've attached the spvasm if we use fcgl or O0, and we see there are now 2 DebugFunctions, one for the wrapper and one for src.main.  debug_info_manager.cpp:117 in the optimizer asserts.  
    assert(
        fn_id_to_dbg_fn_.find(fn_id) == fn_id_to_dbg_fn_.end() &&
        "Register DebugFunction for a function that already has DebugFunction");
I appreciate your advice on this.
[spirv.debug.opline.end.of.shader.O0.spvasm.txt](https://github.com/user-attachments/files/17423224/spirv.debug.opline.end.of.shader.O0.spvasm.txt)
[spirv.debug.opline.end.of.shader.fcgl.spvasm.txt](https://github.com/user-attachments/files/17423225/spirv.debug.opline.end.of.shader.fcgl.spvasm.txt)

s-perron:
@SteveUrquhart The `DebugFunction` instructions do not look correct. Both of them have `%src_main` for their declaration, which is incorrect for two reasons. First they represent different functions, so they should be different. Second, that value should be the result of a `DebugFunctionDeclaration`, and not the result of the `OpFunction`.

```
   %src_main = OpFunction %void None %15
         %10 = OpExtInst %void %1 DebugFunction %8 %5 %6 3 1 %7 %9 FlagIsProtected|FlagIsPrivate 3 %src_main
         %14 = OpExtInst %void %1 DebugFunction %13 %5 %6 3 1 %7 %9 FlagIsProtected|FlagIsPrivate 3 %src_main
```

SteveUrquhart:
I've found the NonSemantic DebugScope for the entry function is incorrectly referencing the DebugFunction for the wrapper.  I will enhance the testing to catch this and push again.

SteveUrquhart:
I've added testing for NonSemantic DebugScope, and cleaned up the names the debug user sees for the wrapper and the original function.  While debugging src_main, the user will want to see "main" from the HLSL.

s-perron:
This looks good to me, but I'm not setup to test it in renderdoc. @Goshido Does this fix #6939?

SteveUrquhart:
I've found one more problem with the logic.  The new convenience function emitDebugFunction() was the wrong place to pushCurrentLexicalScope().  I have a fix for this and am adding test logic for it now.

Goshido:
I gonna check _PR_ when you are ready.

SteveUrquhart:
@Goshido, please check it when you are ready.

Goshido:
I checked branch `fix-ns-debugfunctiondef`, commit `e1a4719d7b53ef6fddffa3f18422b6526f65a79b`. _DXC_ version 1.8.2407.93. It's working as expected.
image

Awesome! 👍

_P.S._ Quick question. What last digit of _DXC_ version means?

```txt
            minor   ???
              |      |
              ▼      ▼
DXC version 1.8.2407.93
            ^     ^
            |     |
          major   |
                  |
             day and month
```

s-perron (approved):

cassiebeckley (approved):

PR #6985, opened 2024-10-24
[SPIR-V] Fix bad OpLoad on loaded texture

### LINKED ISSUE #4136: [SPIR-V] Invalid SPIR-V generated when using RW resource returned from function
The following shader generates invalid SPIR-V:

```hlsl
RWTexture2D<float2> textures[];

cbuffer CB {
	uint index;
}

RWTexture2D<float2> GetTexture() {
	return textures[index];
}

[numthreads(1, 1, 1)]
void main(uint3 tid : SV_DispatchThreadID)
{
	GetTexture()[tid.xy] = float2(1.0, 0.0);
	// workaround:
	// RWTexture2D<float2> tex = GetTexture(); tex[tid.xy] = float2(1.0, 0.0);
}
```

Compiled using latest release (v1.6.2112) with command line `dxc.exe -E main -T cs_6_0 -Fo uav_indexing.comp.spv -spirv -fspv-target-env=vulkan1.1 uav_indexing.comp.hlsl`

As commented in the code, using the returned value from the function and storing it in a resource variable works and has the intended effect.

Keenuts:
This code path always added an OpLoad since it expected either an OpAccessChain or a variable.
This is wrong if we get the texture from a called function.

There is one thing I do find weird is the OpAccessChain result is considered to be an L-value. But I'd expect this to be an r-value.

Fixes #4136

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -7900,8 +7900,12 @@ SpirvInstruction *SpirvEmitter::tryToAssignToRWBufferRWTexture(
       beginInvocationInterlock(baseExpr->getExprLoc(), range);
     }
 
-    auto *image = spvBuilder.createLoad(imageType, baseInfo,
-                                        baseExpr->getExprLoc(), range);
+    SpirvInstruction *image = baseInfo;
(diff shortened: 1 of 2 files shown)


cassiebeckley (approved):
It makes sense to me that the `OpAccessChain` result is treated as an l-value – `OpAccessChain` always returns a pointer, which can be assigned to.

s-perron (approved):

PR #6990, opened 2024-10-29
PIX: split vector alloca writes to enable debug instrumentation

jeffnn:
Bit of background: PIX's debug passes add new allocas, stores to which tie debug info to DXIL Values. 
In the case of a preexisting alloca, PIX will still add its debug writes, but the codegen may have emitted vector-valued stores. Concretely, this happens for the ray payload and RayDesc structs in DXR. Those vector-values stores were being treated incorrectly, resulting in missing values in the PIX shader debugger.

This change splits vector-valued stores to such allocas into extractelement/scalar-stores, which enables the rest of the PIX instrumentation to function as expected. 

(Worth noting that this change only applies to the PIX shader debugging pass.)

### DIFF
--- a/lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
+++ b/lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
@@ -76,7 +76,7 @@ public:
 
 private:
   void AnnotateValues(llvm::Instruction *pI);
-  void AnnotateStore(llvm::Instruction *pI);
+  void AnnotateStore(hlsl::OP *HlslOP, llvm::Instruction *pI);
   bool IsAllocaRegisterWrite(llvm::Value *V, llvm::AllocaInst **pAI,
                              llvm::Value **pIdx);
   void AnnotateAlloca(llvm::AllocaInst *pAlloca);
@@ -141,13 +141,17 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
     }
   }
 
+  // Make a vector of the instructions since we may delete some of them as we go
+  std::vector<llvm::Instruction *> InstrumentableInstructions;
   for (auto *F : instrumentableFunctions) {
     for (auto &block : F->getBasicBlockList()) {
       for (llvm::Instruction &I : block.getInstList()) {
-        AnnotateStore(&I);
+        InstrumentableInstructions.push_back(&I);
       }
     }
   }
+  for (llvm::Instruction *I : InstrumentableInstructions)
+    AnnotateStore(m_DM->GetOP(), I);
 
   for (auto *F : instrumentableFunctions) {
     int InstructionRangeStart = InstNum;
@@ -200,7 +204,8 @@ void DxilAnnotateWithVirtualRegister::AnnotateValues(llvm::Instruction *pI) {
   }
 }
 
-void DxilAnnotateWithVirtualRegister::AnnotateStore(llvm::Instruction *pI) {
+void DxilAnnotateWithVirtualRegister::AnnotateStore(hlsl::OP *HlslOP,
+                                                    llvm::Instruction *pI) {
   auto *pSt = llvm::dyn_cast<llvm::StoreInst>(pI);
   if (pSt == nullptr) {
     return;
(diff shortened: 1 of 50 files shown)


damyanp (commented):
LGTM, but we should get a review from someone more familiar with all of this. @adam-yang perhaps?

Is there a reason that the testing for this isn't filecheck based?

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -133,6 +134,18 @@ bool DxilAnnotateWithVirtualRegister::runOnModule(llvm::Module &M) {
   auto instrumentableFunctions =
       PIXPassHelpers::GetAllInstrumentableFunctions(*m_DM);
 
+  // Make a vector of the instructions since we may delete some of them as we go
```
adam-yang, line 137: There's a way to iterate safely while deleting:
```
  for (auto *F : instrumentableFunctions) {
    for (auto &block : F->getBasicBlockList()) {
      for (auto it = block.begin(); it != block.end();) {
        llvm::Instruction &I = *(it++);
        SplitVectorStores(m_DM->GetOP(), I);
      }
    }
  }
```

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -433,6 +474,39 @@ void DxilAnnotateWithVirtualRegister::AssignNewAllocaRegister(
...
+
+  llvm::AllocaInst *Alloca;
+  llvm::Value *Index;
+  if (!IsAllocaRegisterWrite(pSt->getPointerOperand(), &Alloca, &Index)) {
+    return;
+  }
+
+  llvm::Type *SourceType = pSt->getValueOperand()->getType();
+  const llvm::Type::TypeID ID = SourceType->getTypeID();
+  if (ID == llvm::Type::TypeID::VectorTyID) {
```
adam-yang, line 492: Can just call `isVectorTy()`

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -433,6 +474,39 @@ void DxilAnnotateWithVirtualRegister::AssignNewAllocaRegister(
...
+    if (auto *constIntIIndex = llvm::cast<llvm::ConstantInt>(Index)) {
+      // break vector alloca stores up into individual stores
+      llvm::IRBuilder<> B(pSt);
+      auto *source = llvm::dyn_cast<llvm::VectorType>(SourceType);
```
adam-yang, line 496: No need to dyn_cast it, just call `SourceType->getVectorNumElements()`. Gets rid of the shadowing of the name `source` too.

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -297,15 +310,41 @@ bool DxilAnnotateWithVirtualRegister::IsAllocaRegisterWrite(
...
+
+        // And the source pointer may be a vector (floatn) type,
+        // and if so, that's another offset to consider.
+        llvm::Type *DestType = pGEP->getPointerOperand()->getType();
+        const llvm::Type::TypeID ID = DestType->getTypeID();
+        // We expect this to be a pointer type (it's a GEP after all):
+        if (ID == llvm::Type::TypeID::PointerTyID) {
+          llvm::Type *PointedType =
+              llvm::cast<llvm::PointerType>(DestType)->getElementType();
+          const llvm::Type::TypeID PointedID = PointedType->getTypeID();
```
adam-yang, line 334: Can just be `PointedType->isPointerTy()`.

adam-yang (changes requested):
Only big one is the leak.

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
jeffnn: Thanks. I guess both approaches implicitly rely on the fact that the container doesn't reallocate elements...

lib/DxilPIXPasses/DxilAnnotateWithVirtualRegister.cpp
```diff
@@ -433,6 +474,39 @@ void DxilAnnotateWithVirtualRegister::AssignNewAllocaRegister(
...
+      llvm::IRBuilder<> B(pSt);
+      auto *source = llvm::dyn_cast<llvm::VectorType>(SourceType);
+      for (uint64_t el = 0; el < source->getVectorNumElements(); ++el) {
+        llvm::Value *destPointer = B.CreateGEP(pSt->getPointerOperand(),
+                                               {B.getInt32(0), B.getInt32(el)});
+        llvm::Value *source =
+            B.CreateExtractElement(pSt->getValueOperand(), el);
+        B.CreateStore(source, destPointer);
+      }
+      pI->removeFromParent();
```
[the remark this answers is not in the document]
jeffnn: Thanks!

jeffnn:
> LGTM, but we should get a review from someone more familiar with all of this. @adam-yang perhaps?
> 
> Is there a reason that the testing for this isn't filecheck based?

Yes: you can't get the textual pass "printf" output via filecheck.

damyanp (approved):

damyanp:
Still LGTM - changing my review into an approval!

PR #7010, opened 2024-11-14
PIX: Make shader access tracking pass return whether or not it modified anything

jeffnn:
The SAT pass will now report a string when it finds that no resource access was performed by a module.

This solves two issues.
-The SAT pass, when operating on a shader that has no resource access, will add a resource declaration for the PIX output UAV. This becomes a problem for PIX in terms of juggling root signatures and/or root parameters. 
-SAT is a notoriously time-consuming operation in PIX, so if the SAT pass can report that its results do not need to be re-packaged into a new container for passing to the d3d API, then PIX can avoid the overhead of doing so.

But of course we need to be careful that the SAT pass doesn't erroneously report that it didn't modify the module when in fact it did, cuz then PIX would fail to notice some dynamic resource accesses. That's what the tests are for.

### DIFF
--- a/tools/clang/unittests/HLSL/PixTest.cpp
+++ b/tools/clang/unittests/HLSL/PixTest.cpp
@@ -589,7 +589,8 @@ PassOutput PixTest::RunShaderAccessTrackingPass(IDxcBlob *blob) {
       m_dllSupport.CreateInstance(CLSID_DxcOptimizer, &pOptimizer));
   std::vector<LPCWSTR> Options;
   Options.push_back(L"-opt-mod-passes");
-  Options.push_back(L"-hlsl-dxil-pix-shader-access-instrumentation,config=U0:0:10i0;U0:1:2i0;.0;0;0.");
+  Options.push_back(L"-hlsl-dxil-pix-shader-access-instrumentation,config=U0:0:"
+                    L"10i0;U0:1:2i0;.0;0;0.");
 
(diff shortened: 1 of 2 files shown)


tools/clang/unittests/HLSL/PixTest.cpp
```diff
@@ -576,13 +583,14 @@ TEST_F(PixTest, CompileDebugDisasmPDB) {
...
-CComPtr<IDxcBlob> PixTest::RunShaderAccessTrackingPass(IDxcBlob *blob) {
+PassOutput PixTest::RunShaderAccessTrackingPass(IDxcBlob *blob) {
   CComPtr<IDxcOptimizer> pOptimizer;
   VERIFY_SUCCEEDED(
       m_dllSupport.CreateInstance(CLSID_DxcOptimizer, &pOptimizer));
   std::vector<LPCWSTR> Options;
   Options.push_back(L"-opt-mod-passes");
-  Options.push_back(L"-hlsl-dxil-pix-shader-access-instrumentation,config=");
+  Options.push_back(L"-hlsl-dxil-pix-shader-access-instrumentation,config=U0:0:"
+                    L"10i0;U0:1:2i0;.0;0;0.");
```
bob80905, line 593: Why does this need to change?

jeffnn: Without the config info, the SAT pass won't actually make any changes.

adam-yang (approved):

PR #7037, opened 2024-12-16
Move hlsl_header definition from lit.site.cfg.in to lit.cfg

bob80905:
Internal pipelines fail because a definition that was added to lit.site.cfg.in isn't being recognized. This PR moves the definition over to the relevant lit.cfg file, which leaves this autogenerated file unedited. The new definition is picked up and recognized, which prevents failures.

### DIFF
--- a/tools/clang/test/lit.cfg
+++ b/tools/clang/test/lit.cfg
@@ -210,6 +210,7 @@ if has_plugins and config.llvm_plugin_ext:
 
 config.substitutions.append( ('%llvmshlibdir', config.llvm_shlib_dir) )
 config.substitutions.append( ('%pluginext', config.llvm_plugin_ext) )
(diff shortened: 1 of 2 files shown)


damyanp (commented):
Superficially, this is surprising to me - why is hlsl_headers_dir different to all the other things after `config.` in the .in file?

bob80905:
> Superficially, this is surprising to me - why is hlsl_headers_dir different to all the other things after `config.` in the .in file?

It was manually added in this PR:
[Add specific handling for inline spirv pointer types by s-perron · Pull Request #6873 · microsoft/DirectXShaderCompiler](https://github.com/microsoft/DirectXShaderCompiler/pull/6873/files)

bob80905:
Dxc doesn't need to be modified it appears.

Issue #7034, opened 2024-12-12
Stack overflow when compiling for cs_6_6


damyanp:
See internal bug #55479829 for details until we're able to put an IP-free repro case publicly.

We have a shader that compiles successfully with cs_6_5 but hits a stackoverflow when compiled with cs_6_6 or higher.

damyanp:
@llvm-beanz found this commit that I was able to hackily port to DXC enough to verify that it does fix the issue:

https://github.com/llvm/llvm-project/commit/9d6d24c25056c17db56cf1ef5124f82eb18afc2c#diff-2d5b286ee59dc3de7d78cbba82a57b919689dddd618fea35ae43b812fe036f4d

PR #7048, opened 2025-01-06
[DXIL] Avoid infinite loop / stack overflow on invalid instruction

bob80905:
The simplify instruction pass has a function called `findScalarElement`.
This function will possibly recurse infinitely into the first operand of the instruction that it is passed. If this instruction, for example, is  `%2 = insertelement <2 x i32> %2, i32 %sub, i32 0`
Then the pass will recurse into the first operand of the insertelement instruction, which is the same insertelement instruction, and so on. This PR prevents infinite recursion and stack overflow by bailing out of the function early if there is a detection of the same value in the first operand.
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7034

### DIFF
--- a/lib/Analysis/VectorUtils2.cpp
+++ b/lib/Analysis/VectorUtils2.cpp
@@ -42,7 +42,7 @@ llvm::Value *llvm::findScalarElement(llvm::Value *V, unsigned EltNo) {
     if (EltNo == IIElt)
       return III->getOperand(1);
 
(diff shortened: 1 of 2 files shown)


tex3d (approved):
Looks good.  I still worry about the root cause of a malformed `insertelement` instruction, but this is a good guard against the downstream effect.

llvm-beanz (approved):

PR #7053, opened 2025-01-09
Add Preview Hashing sentinel value to dxil validator

bob80905:
When the dxil validator detects that the target shader model is a preview shader model that hasn't officially been released, it should sign the dxil container with a special sentinel hash: 02020202020202020202020202020202

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7083


utils/hct/hctdb_instrhelp.py
```diff
@@ -1535,11 +1537,22 @@ def get_interpretation_table():
     gen = db_sigpoint_gen(db)
     return run_with_stdout(lambda: gen.print_interpretation_table())
 
+# highest recognized is different than highest released,
+# since there can be prerelease versions that are higher
+# than the last released version
+highest_recognized_major = 6
+highest_recognized_minor = 9
```
tex3d, line 1544: I don't think we needed to rename these to add "recognized".  When I was talking to you about "highest recognized shader model", that's what these variables already defined.  By renaming this variable at the same time as introducing the new one, it makes it harder to tell exactly what's different.

utils/hct/hctdb_instrhelp.py
```diff
@@ -1535,11 +1537,22 @@ def get_interpretation_table():
...
 
+# fetch the last released version from latest-released.json
+json_path = os.path.dirname(os.path.dirname(__file__)) + "/version/latest-release.json"
+with open(json_path, 'r') as file:
+    json_data = json.load(file)
+
+highest_released_minor = int(json_data["version"]["minor"])
+
+
+highest_shader_models = {4: 1, 5: 1, 6: highest_recognized_minor}
```
tex3d, line 1555: I don't think `highest_shader_models` needed to be moved here.

lib/DXIL/DxilMetadataHelper.cpp
```diff
@@ -284,13 +284,20 @@ void DxilMDHelper::LoadDxilShaderModel(const ShaderModel *&pSM) {
...
       (Minor == ShaderModel::kOfflineMinor ? "x" : std::to_string(Minor));
   pSM = ShaderModel::GetByName(ShaderModelName.c_str());
-  if (!pSM->IsValidForDxil()) {
-    char ErrorMsgTxt[40];
-    StringCchPrintfA(ErrorMsgTxt, _countof(ErrorMsgTxt),
-                     "Unknown shader model '%s'", ShaderModelName.c_str());
-    string ErrorMsg(ErrorMsgTxt);
-    throw hlsl::Exception(DXC_E_INCORRECT_DXIL_METADATA, ErrorMsg);
+  // check to see if the provided shader model is the pre-release version
+  if (ShaderModel::IsPreReleaseShaderModel(Major, Minor)) {
```
tex3d, line 288: Why is this change necessary?  I don't think any special case is necessary here for preview shader models.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -32,7 +33,13 @@
 using namespace llvm;
 using namespace hlsl;
 
-static void HashAndUpdate(DxilContainerHeader *Container) {
+static void HashAndUpdate(DxilContainerHeader *Container, bool isPreRelease) {
+  if (isPreRelease) {
+    // If preview bypass is enabled, use the preview hash.
+    UINT previewHash[4] = {0x02020202, 0x02020202, 0x02020202, 0x02020202};
```
tex3d, line 39: Shouldn't we define the preview hash pattern in `DxilContainer.h`?

tools/clang/test/HLSLFileCheck/validation/prerelase-hash.ll
```diff
@@ -0,0 +1,39 @@
+; RUN: %dxilver 1.9 | %dxv %s | FileCheck %s
```
tex3d, line 1: I don't see any CHECK lines.  How does this test work?

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -45,8 +52,25 @@ static void HashAndUpdate(DxilContainerHeader *Container) {
...
+  bool isPreRelease = false;
+  const DxilContainerHeader *DxilContainer =
+      IsDxilContainerLike(Shader->GetBufferPointer(), Shader->GetBufferSize());
+  if (!DxilContainer)
+    return;
+
+  const DxilProgramHeader *ProgramHeader =
+      GetDxilProgramHeader(DxilContainer, DFCC_DXIL);
+
+  if (ProgramHeader) {
```
tex3d, line 64: Might be useful to note the valid scenario where ProgramHeader is null here: when hashing a root signature container.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -45,8 +52,25 @@ static void HashAndUpdate(DxilContainerHeader *Container) {
...
+    int PV = ProgramHeader->ProgramVersion;
+    int major = (PV >> 4) & 0xF; // Extract the major version (next 4 bits)
+    int minor = PV & 0xF;        // Extract the minor version (lowest 4 bits)
+    isPreRelease = ShaderModel::IsPreReleaseShaderModel(major, minor);
```
tex3d, line 68: Just had a thought: in the validator, do we check that `ProgramHeader->ProgramVersion` matches the shader model in the DxilModule somehow?  If not, we need to do that as well.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4149,6 +4159,95 @@ TEST_F(ValidationTest, ValidateWithHash) {
   VERIFY_ARE_EQUAL(memcmp(Result, pHeader->Hash.Digest, sizeof(Result)), 0);
 }
 
+// verify that containers that are not valid DXIL do not
+// get assigned a hash.
+TEST_F(ValidationTest, ValidatePreviewBypassHash) {
+  if (m_ver.SkipDxilVersion(1, 8))
```
tex3d, line 4165: I predict subtle issues:

This test will potentially be run with a prior validator externally available.  When that's the case, this is meant to skip the test if it doesn't support testing of the full versions required by this code.  In this case, it skips unless we have compiler and validator version 1.8 or above, but we compile to both the highest released, and a potential experimental shader model version.  These could both potentially fail when they should be skipped instead.

I also noticed that the comment before the test method does not describe the current test.

My recommendation is to split this test into two tests:
1. Check that the hash is applied on the highest released version.  Use `SkipDxilVersion(1, ShaderModel::kHighestReleasedMinor)` instead here.  After CompileSource, I do not see why you need to re-validate using the validator interface.  Validation is not disabled, so the hash should be set by the appropriate validator after CompileSource.
2. Check for preview hash value.  Use `SkipDxilVersion(1, ShaderModel::kHighestMinor)` instead here, which will prevent it from running on down-level released validators.  After CompileSource, there's no need to call the validator again, since that should validate and hash the shader.

include/dxc/DXIL/DxilShaderModel.h
```diff
@@ -86,6 +94,9 @@ class ShaderModel {
   static const ShaderModel *Get(Kind Kind, unsigned Major, unsigned Minor);
   static const ShaderModel *GetByName(llvm::StringRef Name);
   static const char *GetKindName(Kind kind);
+  static bool IsPreReleaseShaderModel(int Major, int Minor);
+  static const ShaderModel *GetPreReleaseShaderModel(llvm::StringRef name);
```
tex3d, line 98: `GetPreReleaseShaderModel` doesn't appear to be defined or used anywhere.

tex3d (changes requested):
See prior comments.

tools/clang/unittests/HLSL/ValidationTest.cpp
bob80905: 1. has already been done with the test "ValidateWithHash".
I'll discard 1 and make 2.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4149,6 +4159,90 @@ TEST_F(ValidationTest, ValidateWithHash) {
...
+      (hlsl::DxilContainerHeader *)pProgram->GetBufferPointer();
+
+  // Should be equal, this proves the hash is set to the preview bypass hash
+  // when a prerelease version is used
+  VERIFY_ARE_EQUAL(memcmp(&hlsl::PreviewByPassHash, pHeader->Hash.Digest,
+                          sizeof(hlsl::PreviewByPassHash)),
+                   0);
+
+  // test that when the program version differs from the dxil module shader
+  // model version, the validator fails
```
tex3d, line 4200: I'm not sure why this test is part of the ValidatePreviewBypassHash.  One concrete disadvantage is that this test is skipped when there is no preview shader model.  I think it can just be a separate test and check `m_ver.SkipDxilVersion(1, 8)` for skipping older validators.

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -1298,6 +1298,39 @@ HRESULT ValidateLoadModuleFromContainerLazy(
...
+    int major = (PV >> 4) & 0xF; // Extract the major version (next 4 bits)
+    int minor = PV & 0xF;        // Extract the minor version (lowest 4 bits)
+    DxilModule *pDxilModule = DxilModule::TryGetDxilModule(pModule);
+    if (!pDxilModule) {
+      return DXC_E_IR_VERIFICATION_FAILED;
+    }
+    ValidationContext ValCtx(*pModule, pDebugModule, *pDxilModule);
+    int moduleMajor = ValCtx.DxilMod.GetShaderModel()->GetMajor();
+    int moduleMinor = ValCtx.DxilMod.GetShaderModel()->GetMinor();
+    if (moduleMajor != major || moduleMinor != minor) {
```
tex3d, line 1323: I think this validation would be better suited under `ValidateDxilContainerParts`.  This check can simply be triggered under both of these cases that are currently skipped:

https://github.com/microsoft/DirectXShaderCompiler/blob/25faa88053437c30e599b9e5323c94addc2397a1/lib/DxilValidation/DxilContainerValidation.cpp#L1036-L1037

You'll already have the `pPart`, `ValCtx`, and `pDxilModule`.

Plus, you can `GetShaderModel()` directly from `pDxilModule` instead of going through `ValCtx.DxilMod`.

tools/clang/tools/dxcvalidator/dxcvalidator.cpp
```diff
@@ -45,8 +52,26 @@ static void HashAndUpdate(DxilContainerHeader *Container) {
...
+  bool isPreRelease = false;
+  const DxilContainerHeader *DxilContainer =
+      IsDxilContainerLike(Shader->GetBufferPointer(), Shader->GetBufferSize());
+  if (!DxilContainer)
+    return;
+
+  const DxilProgramHeader *ProgramHeader =
+      GetDxilProgramHeader(DxilContainer, DFCC_DXIL);
+
+  // ProgramHeader may be null here, when hasing a root signature container
```
tex3d, line 64:
```suggestion
  // ProgramHeader may be null here, when hashing a root signature container
```

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -4149,6 +4159,90 @@ TEST_F(ValidationTest, ValidateWithHash) {
...
+  int oldMinor = 0;
+  int newMajor = 0;
+  int newMinor = 0;
+  VERIFY_IS_NOT_NULL(pMutableProgramHeader);
+  uint32_t &PV = pMutableProgramHeader->ProgramVersion;
+  oldMajor = (PV >> 4) & 0xF; // Extract the major version (next 4 bits)
+  oldMinor = PV & 0xF;        // Extract the minor version (lowest 4 bits)
+
+  // flip the bit of the shader model version minor
+  PV ^= 1;
```
tex3d, line 4215: This bit flip could increase the version above `ShaderModel::kHighestMinor`, depending on what that is.  I think that is a bad scenario where the validator should avoid even loading the module from the container, though that's not what it does today.

I think when extracting this test from the prior one, we can simply compile a shader to fixed shader model, like SM 6.0, and modify the ProgramPart to SM 6.1 or something like that.  No need for a floating version.

include/dxc/DXIL/DxilShaderModel.h
```diff
@@ -26,7 +26,7 @@ class ShaderModel {
 public:
   using Kind = DXIL::ShaderKind;
 
-  // Major/Minor version of highest shader model
+  // Major/Minor version of highest released shader model
```
tex3d, line 29: This change isn't accurate now.  Perhaps add the comment with "released" added before the block generated by `get_highest_released_shader_model()` instead.

tex3d (approved):
Looks good!

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -193,11 +193,23 @@ const ShaderModel *ShaderModel::Get(Kind Kind, unsigned Major, unsigned Minor) {
   // VALRULE-TEXT:END
 }
 
-const ShaderModel *ShaderModel::GetByName(llvm::StringRef Name) {
-  // [ps|vs|gs|hs|ds|cs|ms|as]_[major]_[minor]
+bool ShaderModel::IsPreReleaseShaderModel(int major, int minor) {
+  if (DXIL::CompareVersions(major, minor, kHighestReleasedMajor,
+                            kHighestReleasedMinor) <= 0) {
+    return false;
+  }
```
llvm-beanz, line 200: nit:
```suggestion
                            kHighestReleasedMinor) <= 0)
    return false;
```

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -193,11 +193,23 @@ const ShaderModel *ShaderModel::Get(Kind Kind, unsigned Major, unsigned Minor) {
...
+
+  // now compare against highest recognized
+  if (DXIL::CompareVersions(major, minor, kHighestMajor, kHighestMinor) <= 0) {
+    return true;
+  }
```
llvm-beanz, line 205: nit:
```suggestion
  if (DXIL::CompareVersions(major, minor, kHighestMajor, kHighestMinor) <= 0)
    return true;
```

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -1033,8 +1033,28 @@ HRESULT ValidateDxilContainerParts(llvm::Module *pModule,
     case DFCC_ResourceDef:
     case DFCC_ShaderStatistics:
     case DFCC_PrivateData:
+      break;
     case DFCC_DXIL:
-    case DFCC_ShaderDebugInfoDXIL:
+    case DFCC_ShaderDebugInfoDXIL: {
+      const DxilProgramHeader *pProgramHeader =
+          reinterpret_cast<const DxilProgramHeader *>(GetDxilPartData(pPart));
+      if (pProgramHeader) {
```
llvm-beanz, line 1041: nit: you could continue on `!pProgramHeader` to reduce nesting.
https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

llvm-beanz (approved):
Some coding style nits, feel free to ignore them. I know DXC's added code doesn't really conform with LLVM Coding Standards, but I do think consistency in how we write new code is nice.

Overall I think this change looks good.

Issue #7051, opened 2025-01-08
Regression in ContainerBuilder error handling code introduced by #6853


tex3d:
**Description**
PR #6853 changed error handling for `DxcContainerBuilder::SerializeContainer` in a way that violates the DXC API pattern for HRESULT returns on functions that use `IDxcOperationResult` to capture error outputs.  The pattern is to return `S_OK` if it was possible to produce the `IDxcOperationResult` object which captures the output object, error code, and messages.  If the return value is not `S_OK`, the error is so fatal that it is not safe for the calling code to even dereference that interface pointer (which should be `nullptr` anyway).

With the change, the interface function will now return a failing HRESULT when a validation failure occurs and it constructs a valid `DxcResult` object.  Calling code in `dxc.cpp` was also updated to ignore the HRESULT returned from the API, which is also incorrect.  I suspect this was done to retrieve the error messages now that the call fails under normal validation error circumstances due to the first change.  Doing so could lead to dereferencing a `nullptr` or accessing an invalid object if the error was truly fatal as the method's return value is supposed to indicate.

Additional related error handling issues with the change in this function:
- Returns `S_OK` if `ppResult` is `nullptr` or points to `nullptr`.  The first case should result in `E_INVALIDARG` at the beginning of the function and the second case would be caught by a failing `DxcResult::Create` already.  `S_OK` doesn't make sense here in any case.
- Adds some code after the try/catch block to compute and apply the hash, but this code should be placed inside the try/catch before the construction of the final `DxcResult` object.
- When computing the hash inside the try/catch, the `IDxcBlob pResult` should be used directly, rather than waiting until after the `DxcResult` is created and reading it back from there.
- `IsDxilContainerLike` returning false shouldn't be ignored, that's a fatal internal error indicating a corrupt container.

All the details are outlined in my comments on the PR here: https://github.com/microsoft/DirectXShaderCompiler/pull/6853#pullrequestreview-2538392678

**Steps to Reproduce**
This is a regression in fatal error handling code (such as hitting OOM) found by code inspection, for which a repro can't easily be constructed.

**Environment**
- DXC version: latest main
- Host Operating System: any

PR #7056, opened 2025-01-11
[Error Handling] Fix DxcContainerBuilder error handling

bob80905:
Some error handling changes were incorrectly made to the container assembler, as described in the issue below. 
This is bad because the API protocol for error handling must remain consistent across all functions. This PR
aims to restore the correct semantic meaning of returning bad HR values. 
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7051

### DIFF
--- a/lib/DxilContainer/DxcContainerBuilder.cpp
+++ b/lib/DxilContainer/DxcContainerBuilder.cpp
@@ -105,7 +105,7 @@ HRESULT STDMETHODCALLTYPE DxcContainerBuilder::RemovePart(UINT32 fourCC) {
 HRESULT STDMETHODCALLTYPE
 DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
   DxcThreadMalloc TM(m_pMalloc);
-  if (ppResult == nullptr || *ppResult == nullptr)
+  if (ppResult == nullptr)
     return E_INVALIDARG;
 
   try {
--- a/lib/DxilContainer/DxcContainerBuilder.cpp
+++ b/lib/DxilContainer/DxcContainerBuilder.cpp
@@ -105,6 +105,9 @@ HRESULT STDMETHODCALLTYPE DxcContainerBuilder::RemovePart(UINT32 fourCC) {
 HRESULT STDMETHODCALLTYPE
 DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
   DxcThreadMalloc TM(m_pMalloc);
+  if (ppResult == nullptr || *ppResult == nullptr)
+    return E_INVALIDARG;
+
   try {
     // Allocate memory for new dxil container.
     uint32_t ContainerSize = ComputeContainerSize();
@@ -161,6 +164,13 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
       errorHeap.Detach();
     }
 
+    // Add Hash.
+    LPVOID PTR = pResult->GetBufferPointer();
+    if (!IsDxilContainerLike(PTR, pResult->GetBufferSize()))
+      return E_FAIL;
+
+    HashAndUpdate((DxilContainerHeader *)PTR);
+
     IFT(DxcResult::Create(
         valHR, DXC_OUT_OBJECT,
         {DxcOutputObject::DataOutput(DXC_OUT_OBJECT, pResult, DxcOutNoName),
@@ -169,21 +179,6 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
   }
   CATCH_CPP_RETURN_HRESULT();
 
-  if (ppResult == nullptr || *ppResult == nullptr)
-    return S_OK;
-
-  HRESULT HR;
-  (*ppResult)->GetStatus(&HR);
-  if (FAILED(HR))
-    return HR;
-
-  CComPtr<IDxcBlob> pObject;
-  IFR((*ppResult)->GetResult(&pObject));
-
-  // Add Hash.
-  LPVOID PTR = pObject->GetBufferPointer();
-  if (IsDxilContainerLike(PTR, pObject->GetBufferSize()))
-    HashAndUpdate((DxilContainerHeader *)PTR);
   return S_OK;
 }
 
--- a/lib/DxilContainer/DxcContainerBuilder.cpp
+++ b/lib/DxilContainer/DxcContainerBuilder.cpp
@@ -104,9 +104,10 @@ HRESULT STDMETHODCALLTYPE DxcContainerBuilder::RemovePart(UINT32 fourCC) {
 
 HRESULT STDMETHODCALLTYPE
 DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
-  DxcThreadMalloc TM(m_pMalloc);
   if (ppResult == nullptr)
     return E_INVALIDARG;
+
+  DxcThreadMalloc TM(m_pMalloc);
 
   try {
     // Allocate memory for new dxil container.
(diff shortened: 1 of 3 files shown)


lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -105,6 +105,9 @@ HRESULT STDMETHODCALLTYPE DxcContainerBuilder::RemovePart(UINT32 fourCC) {
 HRESULT STDMETHODCALLTYPE
 DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
   DxcThreadMalloc TM(m_pMalloc);
+  if (ppResult == nullptr)
+    return E_INVALIDARG;
+
```
tex3d, line 110:
```suggestion
  if (ppResult == nullptr)
    return E_INVALIDARG;

  DxcThreadMalloc TM(m_pMalloc);
```

lib/DxilContainer/DxcContainerBuilder.cpp
```diff
@@ -161,6 +164,13 @@ DxcContainerBuilder::SerializeContainer(IDxcOperationResult **ppResult) {
       errorHeap.Detach();
     }
 
+    // Add Hash.
+    LPVOID PTR = pResult->GetBufferPointer();
+    if (!IsDxilContainerLike(PTR, pResult->GetBufferSize()))
+      return E_FAIL;
+
+    HashAndUpdate((DxilContainerHeader *)PTR);
```
tex3d, line 172: Actually, IsDxilContainerLike should never fail here, and it returns the `DxilContainerHeader *` upon success.  Plus, we need to make sure validation succeeded before hashing:
```suggestion
    if (SUCCEEDED(valHR))
      HashAndUpdate(IsDxilContainerLike(pResult->GetBufferPointer(),
                                        pResult->GetBufferSize()));
```

tex3d (changes requested):
If this succeeded testing without the suggested change to check `valHR` before hashing, we are missing a test that makes sure it doesn't hash the shader even when validation fails.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1178,6 +1179,57 @@ TEST_F(ValidationTest, UavBarrierFail) {
 TEST_F(ValidationTest, UndefValueFail) {
   TestCheck(L"..\\CodeGenHLSL\\UndefValue.hlsl");
 }
+// verify that containers that are not valid DXIL do not
+// get assigned a hash.
+TEST_F(ValidationTest, ValidationFailNoHash) {
+  if (m_ver.SkipDxilVersion(1, 8))
+    return;
+  CComPtr<IDxcBlob> pProgram;
+  LPCSTR pSource = "float main(snorm float b : B) : SV_DEPTH \
```
tex3d, line 1188: It would be useful to have a comment explaining how this shader is expected to fail validation but not compilation (due to missing diagnostics for reading uninitialized values).

It's generally best to start with IL rather than HLSL to reliably produce a validator failure, since we could theoretically fix the missing diagnostics and inadvertently invalidate this test.

Additionally, you could use a raw string instead of escaped newlines for better readability/editability.  Like:
```c++
LPCSTR pSource = R"(
    float main(snorm float b : B) : SV_DEPTH
    {
        float a;
        return b + a;
    }
)";
```

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1178,6 +1179,57 @@ TEST_F(ValidationTest, UavBarrierFail) {
...
+      (hlsl::DxilContainerHeader *)pProgram->GetBufferPointer();
+  // Validate the hash.
+  constexpr uint32_t HashStartOffset =
+      offsetof(struct DxilContainerHeader, Version);
+  auto *DataToHash = (const BYTE *)pHeader + HashStartOffset;
+  UINT AmountToHash = pHeader->ContainerSizeInBytes - HashStartOffset;
+  BYTE Result[DxilContainerHashSize];
+  ComputeHashRetail(DataToHash, AmountToHash, Result);
+  // Should be unequal, this proves the hash isn't written when validation fails
+  VERIFY_ARE_NOT_EQUAL(memcmp(Result, pHeader->Hash.Digest, sizeof(Result)), 0);
```
tex3d, line 1231: I think we should instead simply make sure the hash is all zeros.  We shouldn't need to compute the hash ourselves.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1178,6 +1179,57 @@ TEST_F(ValidationTest, UavBarrierFail) {
...
+  pResult->GetStatus(&status);
+
+  // expect validation to fail
+  VERIFY_FAILED(status);
+  pResult->GetResult(&pValidationOutput);
+  // Make sure the validation output is not null when hashing.
+  VERIFY_SUCCEEDED(pValidationOutput != nullptr);
+
+  hlsl::DxilContainerHeader *pHeader =
+      (hlsl::DxilContainerHeader *)pProgram->GetBufferPointer();
```
tex3d, line 1222: We should use `IsDxilContainerLike` for this translation to ` hlsl::DxilContainerHeader *` rather than a cast like this.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1178,6 +1179,57 @@ TEST_F(ValidationTest, UavBarrierFail) {
...
+  VERIFY_SUCCEEDED(pValidator->Validate(pProgram, Flags, &pResult));
+  HRESULT status;
+  VERIFY_IS_NOT_NULL(pResult);
+  CComPtr<IDxcBlob> pValidationOutput;
+  pResult->GetStatus(&status);
+
+  // expect validation to fail
+  VERIFY_FAILED(status);
+  pResult->GetResult(&pValidationOutput);
+  // Make sure the validation output is not null when hashing.
```
tex3d, line 1218:
```suggestion
  // Make sure the validation output is not null even when validation fails.
```

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1178,6 +1179,59 @@ TEST_F(ValidationTest, UavBarrierFail) {
...
   TestCheck(L"..\\CodeGenHLSL\\UndefValue.hlsl");
 }
+// verify that containers that are not valid DXIL do not
+// get assigned a hash.
+TEST_F(ValidationTest, ValidationFailNoHash) {
+  if (m_ver.SkipDxilVersion(1, 8))
+    return;
+  CComPtr<IDxcBlob> pProgram;
+
+  // we need any shader that will pass compilation but fail validation
```
tex3d, line 1189: Nit: the comment could say why the shader is expected to fail validation, otherwise it might not be obvious.  That was the initial issue that tripped me up when reading the test code.

I still would have preferred we assembled IR instead of compiled HLSL, since that's what we should be using in validation tests.

```suggestion
  // We need any shader that will pass compilation but fail validation.
  // This shader reads from uninitialized 'float a', which works for now.
```

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1178,6 +1179,59 @@ TEST_F(ValidationTest, UavBarrierFail) {
...
+  VERIFY_FAILED(status);
+  pResult->GetResult(&pValidationOutput);
+  // Make sure the validation output is not null even when validation fails
+  VERIFY_SUCCEEDED(pValidationOutput != nullptr);
+
+  hlsl::DxilContainerHeader *pHeader = IsDxilContainerLike(
+      pProgram->GetBufferPointer(), pProgram->GetBufferSize());
+  VERIFY_IS_NOT_NULL(pHeader);
+
+  BYTE Result[DxilContainerHashSize] = {0, 0, 0, 0, 0, 0, 0, 0,
```
tex3d, line 1229: Nit: I think `Result` should be renamed to be clear that it's zeros, like `ZeroHash`.  That way if you had a failure in the VERIFY macro, you would see that it's comparing with `ZeroHash` and understand what was expected without having to read this source code.

tools/clang/unittests/HLSL/ValidationTest.cpp
```diff
@@ -1178,6 +1179,59 @@ TEST_F(ValidationTest, UavBarrierFail) {
...
+                                        0, 0, 0, 0, 0, 0, 0, 0};
+  // ComputeHashRetail(DataToHash, AmountToHash, Result);
+  // Should be unequal, this proves the hash isn't written when validation fails
```
tex3d, line 1232: Nit: Comment out of date.
```suggestion
  // Should be equal, this proves the hash isn't written when validation fails
```

tex3d (approved):
Looks good.

damyanp (approved):

PR #7059, opened 2025-01-13
Use UMad for dot products on uint vectors

### LINKED ISSUE #7058: Dot products on unsigned integer vectors do not use umad
**Description**
When using the `dot` intrinsic to calculate a dot product, the generated code will use a signed mad operator instead of the umad operator that the `mul` intrinsic uses when it amounts to a dot product.

**Steps to Reproduce**

Just pass unsigned integer vectors of 2 or more elements to the `dot` intrinsic.

Switching the defined FUNC between `dot` and `mul` in this godbolt link will show the difference in output: https://godbolt.org/z/Pz6a5a38r

**Actual Behavior**

You can see that it will change between calling IMad with opcode 48 and UMad with opcode 49. 

**Environment**
- DXC version 1.8.2407
- Host Operating System macos 14.7.2

pow2clk:
Dot products on uint vectors were not using the unsigned mad operations when they were formed using the dot intrinsic though they did when it was the effect of the mul intrinsic. While such operations aren't very common and less common still are the cases where the unsigned variant makes a difference, this was an internal inconsistency as well as with the clang implementation. This corrects the problem using a dummy op for udot similar to how other intrinsics address similar problems to pass the unsigned information to operation lowering.

Incidentally adds unsigned notations to all the mul intrinsic entries. This is a non-functional change as any one overload that specifies an unsigned variant will apply for all of them, but it is less confusing than having half of them do it and the other half not.

The test verifies the output of a few dot operations with sint, float, and uint types for both dot and mul intrinsics

Fixes #7058


farzonl (approved):

V-FEXrt (approved):

tools/clang/test/CodeGenHLSL/dot.hlsl
```diff
@@ -0,0 +1,135 @@
+// RUN: %dxc -T vs_6_0 -DFUNC=dot %s | FileCheck %s
+// RUN: %dxc -T vs_6_0 -DFUNC=mul %s | FileCheck %s
+
+// Verifies correct implementation of dot and mul with vectors for various sizes and types.
+
+// Partially pilfered from SPIRV's intrinsic.dot.hlsl
+
+float4 main(int1 i1[2] : IO, int2 i2[2] : IT, int3 i3[2] : IH, int4 i4[2] : IF,
+            float1 f1[2] : FO, float2 f2[2] : FT, float3 f3[2] : FH, float4 f4[2] : FF,
+            uint1 u1[2] : UO, uint2 u2[2] : UT, uint3 u3[2] : UH, uint4 u4[2] : UF) : SV_Position {
```
tex3d, line 10: I think this might have been a bit simpler with separate functions in a library target for each case, but this works too.

tex3d (approved):
Looks good.

tools/clang/test/CodeGenHLSL/dot.hlsl
tex3d: Actually, perhaps separate tests for IRGen (`-fcgl`), `HLOperationLower`, and SPIR-V gen would have been in keeping with our preferred approach now.

pow2clk: The SPIRV generator makes no use of `UMulExtended` and SPIRV doesn't have an explicit mad/fma operator that I can see. I don't think this is applicable to SPIR-V.

I don't mind adding a fcgl variant. I need to do something to prompt the previously glitched tests from running anyway.

Issue #6995, opened 2024-11-01
[SPIR-V] Invalid generated SPIR-V when nesting HLSL 2021 `and` intrinsic.


Mr-Wiseguy:
**Description**
In an attempt to minimize branching in our HLSL 2021 shader code, we have been converting boolean operators to use the `and` intrinsic from HLSL 2021 that doesn't short circuit. However, when nesting two calls to `and`, we get the following error:
`fatal error: generated SPIR-V is invalid: Expected both operands to be of Result Type: LogicalAnd`

**Steps to Reproduce**
https://shader-playground.timjones.io/a2afd625eb79aac130756befae206296
https://godbolt.org/z/vfn3Kxc7W

**Environment**
- Tested latest July 2024 release and latest trunk on godbolt (5704c474)

s-perron:
This is incorrect even with just a single `and` (https://godbolt.org/z/Te48c9zzz). The AST for the call to `and` is:


```
  |         |-ImplicitCastExpr 0x56546c5b52c8 <col:12> 'bool (*)(int, int)' <FunctionToPointerDecay>
  |         | `-DeclRefExpr 0x56546c5b5260 <col:12> 'bool (int, int)' lvalue Function 0x56546c5b5048 'and' 'bool (int, int)'
```

It treats the parameters as integers. However, the spir-v instruction takes bools. Should be an easy fix.

PR #7060, opened 2025-01-13
Fix and/or intrinsics with non-int parameters

### LINKED ISSUE #7057: non-integer parameters to or/and intrinsics crash the compiler
**Description**
Providing floating point values to and/or intrinsics results in a compiler crash. 

**Steps to Reproduce**

Passing floating values, whether scalars or vectors, from any source to either `or` or `and`.
https://godbolt.org/z/aa8v4T6sd

**Actual Behavior**
 Results in an internal compiler error.

**Environment**
- DXC version: 1.8.2407
- Host Operating System macos

pow2clk:
And/or intrinsics were set to allow any parameters, which is consistent with the behavior of the && and || operators they were meant to replace, however this meant that if they were passed floating point values, those values would be applied to the binary and/or operands, which isn't allowed. Instead, they should be converted to booleans to be consistent with the behavior of && and ||. This can be done simply by restricting the parameters to booleans which forces the appropriate conversions. Adds tests for for using bools, ints, and floats in scalars, different sized vectors, and matrices as parameters to or and and.

Fixes: #7057
FIxes: #6995

### DIFF
--- a/tools/clang/test/CodeGenHLSL/vector-and.hlsl
+++ b/tools/clang/test/CodeGenHLSL/vector-and.hlsl
@@ -1,0 +1,141 @@
+// RUN: %dxc -T ps_6_0 -HV 2021 -DTYPE=bool %s | FileCheck %s --check-prefixes=CHECK,I32
+// RUN: %dxc -T ps_6_0 -HV 2018 -DTYPE=bool %s | FileCheck %s --check-prefixes=CHECK,I32
+// RUN: %dxc -T ps_6_0 -HV 2021 -DTYPE=int %s | FileCheck %s --check-prefixes=CHECK,I32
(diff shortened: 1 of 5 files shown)


s-perron:
This might mean that https://github.com/microsoft/DirectXShaderCompiler/pull/7055 is not need anymore.

pow2clk:
> This might mean that #7055 is not need anymore.

I think it should. That wasn't my intent! I caught this in an independent investigation, but I'm glad we can resolve something globally.

s-perron (commented):
Would it be possible to add SPIR-V tests? This seems to work for SPIR-V as well. However, I did find another bug with matrices, but that is a different issue.

pow2clk:
> Would it be possible to add SPIR-V tests? This seems to work for SPIR-V as well. However, I did find another bug with matrices, but that is a different issue.

If you don't mind me stealing the tests from #7055, I can do that easily!

s-perron (approved):

farzonl (approved):

PR #7065, opened 2025-01-13
Update version to 1.8.2502

bob80905:
Update version to 1.8.2502


docs/ReleaseNotes.md
```diff
@@ -21,6 +21,8 @@ The included licenses apply to the following files:
 
 Place release notes for the upcoming release below this line and remove this line upon naming this release.
 
+### Version 1.8.2502
+
```
tex3d, line 25: Should we do a survey for any additional notable changes that missed being mentioned here?

bob80905: If we intend to branch today, I think we should perhaps cherry pick a change later, instead of blocking on that now.

llvm-beanz (approved):

tex3d (approved):

damyanp:
nudge!

PR #7071, opened 2025-01-15
[Pipelines] Extend time and add PR trigger

bob80905:
When the pipeline builds and tests in a windows environment, the time limit of 90 minutes reached. There is an insufficient amount of time available to run the pipeline in the windows environment. So, this PR increases the limit to 120 minutes.
Here is an example of insufficient time constraints:
https://github.com/microsoft/DirectXShaderCompiler/pull/7065
Additionally, this PR explicitly adds a trigger when a PR is created, so that the pipeline is explicitly run when that event happens. This allows the pipeline to be run with stricter security rules.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,7 +44,7 @@ stages:
       displayName: 'DXIL Tests'
 
   - job: Nix
-    timeoutInMinutes: 90
+    timeoutInMinutes: 120
 
     variables:
       macOS: macOS-latest
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,6 +1,10 @@
 trigger:
   - main
   - release*
+
+pr: 
+  - main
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -9,7 +13,7 @@ stages:
 - stage: Build
   jobs:
   - job: Windows
-    timeoutInMinutes: 90
+    timeoutInMinutes: 120
```
farzonl, line 16: this seems fine, is there anything that can be done to improve build times?

farzonl (approved):

damyanp (approved):

azure-pipelines.yml
bob80905: I am unsure, but created an issue:
https://github.com/microsoft/DirectXShaderCompiler/issues/7090

Issue #7004, opened 2024-11-07
Add Shader Model 6.9 to main


llvm-beanz:
We are beginning development of SM 6.9 and working with partners. To facilitate this we should add the shader model support to `main`.

damyanp:
Need to also figure out validator / hashing to support PREVIEW appropriately.

damyanp:
Also consider making it so that preview / incomplete features don't appear in the help text, or something like that to make it clear that these features aren't actually ready for production use yet.

bob80905:
Closed by #7075

PR #7075, opened 2025-01-16
Bump SM version to 6.9

bob80905:
Bumping SM version to 6.9. Though 6.9 is a while away from release, we intend to use 6.9 as a temporary marker for a pre-release version.


pow2clk (approved):
Funny how I can approve even though I wrote the commit and I expect Josh can't because he submitted it? Maybe not, but either way one of us wrote it and the other approves of it.

I thought this had to go in with the preview validator hash per this comment: https://github.com/microsoft/DirectXShaderCompiler/issues/7004#issuecomment-2474378517

hekota (approved):

PR #7091, opened 2025-01-27
Fix and/or intrinsics with non-int parameters (cp to 1.8.2502)

pow2clk:
Cherry-pick of #7060 to release-1.8.2502
    
And/or intrinsics were set to allow any parameters, which is consistent
with the behavior of the && and || operators they were meant to replace,
however this meant that if they were passed floating point values, those
values would be applied to the binary and/or operands, which isn't
allowed. Instead, they should be converted to booleans to be consistent
with the behavior of && and ||. This can be done simply by restricting
the parameters to booleans which forces the appropriate conversions.
Adds tests for for using bools, ints, and floats in scalars, different
sized vectors, and matrices as parameters to or and and.
    
Fixes: #7057
Fixes: #6995
(cherry picked from commit 25faa88053437c30e599b9e5323c94addc2397a1)

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -22,6 +22,7 @@ The included licenses apply to the following files:
 - The incomplete WaveMatrix implementation has been removed.
 - DXIL Validator Hash is open sourced.
 - DXIL container validation for PSV0 part allows any content ordering inside string and semantic index tables.
+- The and() and or() intrinsics will now accept non-integer parameters by casting them to bools.
 
 ### Version 1.8.2407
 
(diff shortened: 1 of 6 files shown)


bob80905 (approved):

docs/ReleaseNotes.md
```diff
@@ -22,6 +22,7 @@ The included licenses apply to the following files:
 - The incomplete WaveMatrix implementation has been removed.
 - DXIL Validator Hash is open sourced.
 - DXIL container validation for PSV0 part allows any content ordering inside string and semantic index tables.
+- The and() and or() intrinsics will now accept non-integer parameters by casting them to bools.
```
damyanp, line 25: :)

Was looking for this, and found it!

damyanp (approved):

tex3d (approved):

PR #7094, opened 2025-01-28
Shader Execution Reordering implementation

simoll:
Shader Execution Reordering (SER) implementation

Specification PR: https://github.com/microsoft/hlsl-specs/pull/277


simoll:

utils/hct/hctdb.py
```diff
@@ -696,6 +696,23 @@ def populate_categories_and_models(self):
             self.name_idx[i].category = "Extended Command Information"
             self.name_idx[i].shader_stages = ("vertex",)
             self.name_idx[i].shader_model = 6, 8
+        for i in (
+            "HitObject_TraceRay,HitObject_FromRayQuery,HitObject_FromRayQueryWithAttrs,HitObject_MakeMiss,HitObject_MakeNop,HitObject_Invoke,ReorderThread,HitObject_IsMiss,HitObject_IsHit,HitObject_IsNop,HitObject_RayFlags,HitObject_RayTMin,HitObject_RayTCurrent,HitObject_WorldRayOrigin,HitObject_WorldRayDirection,HitObject_ObjectRayOrigin,HitObject_ObjectRayDirection,HitObject_ObjectToWorld3x4,HitObject_ObjectToWorld4x3,HitObject_WorldToObject3x4,HitObject_WorldToObject4x3,HitObject_GeometryIndex,HitObject_InstanceIndex,HitObject_InstanceID,HitObject_PrimitiveIndex,HitObject_HitKind,HitObject_ShaderTableIndex,HitObject_SetShaderTableIndex,HitObject_LoadLocalRootTableConstant,HitObject_Attributes"
+        ).split(","):
+            self.name_idx[i].category = "Shader Execution Reordering"
+            self.name_idx[i].shader_model = 6, 8
```
damyanp, line 703: Presumably this should be `6, 9`?  @bob80905 might be able to comment on the right way to do this here?

simoll: This is fixed with the latest commit

simoll:
@pow2clk, 

> Thanks for this! It's clearly a lot of work. That is why I have to make a request. Could you divide this up a bit into separate PRs?

Yes. I've divided the change in this PR into three patches:
1. HLSL HitObject type and default construction
2. Remaining HLSL intrinsics
3. reodercoherent and REORDER_SCOPE

The first PR (https://github.com/microsoft/DirectXShaderCompiler/pull/7097) contains most of the novelties for DXC (static member functions for builtins, HitObject scalar type).

I suggest we keep this PR as an on-going record of the changes and open separate PRs for the patches.

> I also see that there are no checks for shader model version. I would expect to see at least a few checks for IsSM69Plus to prevent declaring the new objects in older shader models etc. I haven't gotten deep enough into the change to say exactly where, but we want to limit this feature to 6.9.

There were checks: The EnableShaderExecutionReordering option is only set for SM6.9+
That said, I've made changes to only enable SER types and intrinsics in SM6.9+. This means you can have a 'class HitObject' and a function 'ReorderThread(..)' in pre-SM6.9 user code.

simoll:
This has been replaced by #7214 and the various `[SER]` PRs . Closing as obsolete.

PR #7097, opened 2025-01-29
[SER] Patch 1: HitObject type lowering and SM 6.9 enablement

simoll:
Reduction of the complete SER implementation to just the HitObject type and its default constructor.

This has most of the infrastructure changes in DXC to support SER, eg static member functions for builtins, HitObject scalar type.

Specification PR: https://github.com/microsoft/hlsl-specs/pull/277


simoll:

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-entry-errors.hlsl
```diff
@@ -0,0 +1,57 @@
...
+{
+    float elem
+          : write(caller,anyhit,closesthit,miss)
+          : read(caller,anyhit,closesthit,miss);
+};
+
+struct Attribs { float2 barys; };
+
+void UseHitObject() {
+// expected-error@+4{{Shader kind 'compute' incompatible with shader execution reordering (has to be raygeneration, closesthit or miss)}}
```
bob80905, line 13: Does this test work if you move these expected errors into the corresponding entry functions? If not, would you re-order the entry functions below so that you can move these into the entry functions? Just a design preference, so that it's obvious the error is that it's the call site being in the context of a bad shader kind, not the function definition itself.

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-in-buffer.hlsl
```diff
@@ -0,0 +1,8 @@
+// RUN: %dxc -T lib_6_9 %s -verify
+
+// expected-error@+1{{'HitObject' is an object and cannot be used as a type parameter}}
+RWStructuredBuffer<HitObject> InvalidBuffer;
+
+[shader("raygeneration")] void main() {
```
bob80905, line 6: Can this function be removed?

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -613,6 +613,22 @@ bool IsHLSLResourceType(clang::QualType type) {
   return false;
 }
 
+bool IsHLSLHitObjectType(QualType type) {
```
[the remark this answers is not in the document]
simoll: The patch uses the same naming and implementation approach for `HitObject` as was done  for `NodeOutput` and `RayQuery`. The name `IsHLSLHitObjectType` says what the function does not how it is implemented, so i don't find that argument very convincing.
Do we really want to change this?

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-entry-errors.hlsl
simoll: Moved into entry functions in [57c861a](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/57c861aab39debbbe5ba53dbb2a9bb4a7c40a822).

tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
```diff
@@ -2794,6 +2794,56 @@ unsigned AlignBufferOffsetInLegacy(unsigned offset, unsigned size,
...
+
+    // Match the HitObject constructor signature. It should be impossible to
+    // achieve the same signature from HLSL.
+    if (!F.getName().startswith("\01??0HitObject@@"))
+      continue;
+    DXASSERT(F.arg_size() == 1 && Ty == F.arg_begin()->getType(),
+             "Wrong signature for apparent HitObject constructor");
+
+    DXASSERT(!HitObjectCtor,
+             "Multiple candidate functions that quality as HitObject "
```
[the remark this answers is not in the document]
simoll: Fixed in [57c861a](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/57c861aab39debbbe5ba53dbb2a9bb4a7c40a822)

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-in-buffer.hlsl
simoll: Removed in [57c861a](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/57c861aab39debbbe5ba53dbb2a9bb4a7c40a822)

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -3967,6 +3999,13 @@ class HLSLExternalSource : public ExternalSemaSource {
     return IsRayQueryBasicKind(GetTypeElementKind(type));
   }
 
+  bool IsHitObjectBasicKind(ArBasicKind kind) {
```
llvm-beanz, line 4002: This function seems unhelpful, so I don't really think we should have it.

tools/clang/lib/AST/HlslTypes.cpp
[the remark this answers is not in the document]
llvm-beanz: We should not string match names. We have done this in the past, it was an unfortunate decision.

With the NodeRecord types we demonstrated a new solution, where we add an unspellable attribute to the type that identifies it as a specific object. We should do something like that here as well.

If you take a look at how `GetNodeIOType` is implemented that will give you a model for how this should be done.

tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
```diff
@@ -2794,6 +2794,56 @@ unsigned AlignBufferOffsetInLegacy(unsigned offset, unsigned size,
   return offset;
 }
 
+// Translate HitObject constructor to dx.op.hitObject_MakeNop
+void TranslateHitObjectConstructor(HLModule &HLM) {
```
llvm-beanz, line 2798: This concerns me. CGHLSLMS is a giant hack that is nothing but a problem.

Is there a reason we can't have the AST represent a sane constructor that we can codegen directly?

tools/clang/lib/SPIRV/AstTypeProbe.cpp
```diff
@@ -1101,6 +1101,13 @@ bool isOpaqueType(QualType type) {
 
     if (name == "SubpassInput")
       return true;
+
+    if (name == "HitObject") {
+      const CXXRecordDecl *typeRecordDecl = type->getAsCXXRecordDecl();
+      return typeRecordDecl &&
+             typeRecordDecl
+                 ->isImplicit(); // This is only our builtin, if SER is enabled
```
llvm-beanz, line 1109:
```suggestion
      // This is only our builtin, if SER is enabled
      return typeRecordDecl && typeRecordDecl->isImplicit(); 
```

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -1161,6 +1163,31 @@ CXXRecordDecl *hlsl::DeclareRayQueryType(ASTContext &context) {
   return typeDeclBuilder.getRecordDecl();
 }
 
+CXXRecordDecl *hlsl::DeclareHitObjectType(ASTContext &Context) {
+  // HitObject { ... }
+  BuiltinTypeDeclBuilder TypeDeclBuilder(Context.getTranslationUnitDecl(),
```
llvm-beanz, line 1168: Since we are not providing a Vulkan implementation of this, we should ensure that all the SER API additions are added to the `dx` namespace.

tools/clang/lib/CodeGen/CGCall.cpp
```diff
@@ -1902,7 +1904,7 @@ void CodeGenFunction::EmitFunctionProlog(const CGFunctionInfo &FI,
     case ABIArgInfo::Extend:
     case ABIArgInfo::Direct: {
       // HLSL Change Begins
-      if (hlsl::IsHLSLMatType(Ty)) {
+      if (hlsl::IsHLSLMatType(Ty) || hlsl::IsHLSLHitObjectType(Ty)) {
```
llvm-beanz, line 1907: This special handling seems odd. Why is this needed?

tools/clang/lib/CodeGen/CGExprScalar.cpp
```diff
@@ -1101,6 +1103,17 @@ void ScalarExprEmitter::EmitBinOpCheck(
 //                            Visitor Methods
 //===----------------------------------------------------------------------===//
 
+Value *ScalarExprEmitter::VisitCXXConstructExpr(CXXConstructExpr *E) {
```
llvm-beanz, line 1106: This is very wrong. Scalar objects shouldn't have constructors.

tools/clang/lib/CodeGen/CGCall.cpp
simoll: This is necessary because otherwise call lowering will flatten the HitObject struct.

tools/clang/lib/Sema/SemaHLSL.cpp
simoll: Removed in bf3400a99

tools/clang/lib/AST/HlslTypes.cpp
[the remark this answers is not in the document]
simoll: Fyi @bob80905 , commit bf3400a99 changes this function to rely on the `HLSLHitObjectAttr` instead.

tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
simoll: > With the NodeRecord types we demonstrated a new solution, where we add an unspellable attribute to the type that identifies it as a specific object. We should do something like that here as well.

bf3400a99 follows your suggestion and adds the `HLSLHitObjectAttr` attribute to model the HitObject type analogously to how it's done with `GetNodeIOType`

> This concerns me. CGHLSLMS is a giant hack that is nothing but a problem.
> Is there a reason we can't have the AST represent a sane constructor that we can codegen directly?

Luckily, `EmitHLSLScalarObjectDefaultConstructor` directly emits a call to `HitObject_MakeNop`. This change in CGHLSLMS was never needed and i removed it in 29c8705e5.

tools/clang/lib/SPIRV/AstTypeProbe.cpp
simoll: Accepted formatting suggestion in 0aa26b00a

tools/clang/lib/CodeGen/CGExprScalar.cpp
simoll: `HitObject` is modelled as a `TEK_Scalar` for codegen as it's done for the HLSL matrix type. This means `HitObject` flows as an opaque value and we only get trivial load/store/alloca of `HitObject` directly from codegen, no type replacement etc needed, exactly as we want it.
This is why you are seeing this constructor call visited by the `ScalarExprEmitter`.


Details on HitObject initialization:
- `HitObject` is declared with a default constructor. 
- When Clang builds an `InitializationSequence` for `HitObject`-typed things, `InitializeInitSequenceForHLSL(..)` adds the constructor call for an `SK_ConstructorInitialization` step.
- Since `HitObject` is `TEK_Scalar`, `ScalarExprEmitter` visits the constructor calls for `HitObject`. It directly emits `HitObject_MakeNop` through `EmitHLSLScalarObjectDefaultConstructor`.

llvm-beanz: This and the handling in CGCall are wrong.

The way matrices are implemented in DXC are a gigantic hack that has caused a never-ending stream of bugs. We should not be following that pattern.

Instead we `HitObjects` should be treated like resource types, which are types that contain handles that we can safely treat as aggregates.

tools/clang/lib/CodeGen/CGCall.cpp
llvm-beanz: See my other comment, but we should not be matching the pattern set by matrices because it is an endless source of bugs.

simoll: Removed in 358942da6 . HitObject is now an aggregate.

tools/clang/lib/AST/ASTContextHLSL.cpp
simoll: This is in line with all other type declarations living in the hlsl namespace (and there is no dx namespace). The spec makes HitObject part of HLSL, not of the DX backend. Please elaborate what you mean here.

tools/clang/lib/CodeGen/CGExprScalar.cpp
simoll: 358942da6 changes HitObject to be handled like a resource type. I've removed the scalar special handling cases.

tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
simoll: This is obsolete, 358942da6 removes the special handling for the HitObject type in CGHLSLMS and elsewhere.

tools/clang/lib/AST/ASTContextHLSL.cpp
llvm-beanz: Thank you for pointing out the issue with the spec. I've filed this issue tracking updating the spec: https://github.com/microsoft/hlsl-specs/issues/395

The original feedback stands.

tools/clang/lib/AST/HlslTypes.cpp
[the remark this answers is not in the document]
simoll: With  b028aaf6b this is using DXHitObjectAttr and no string matching. Marking as resolved.

tools/clang/lib/AST/ASTContextHLSL.cpp
simoll: > With the NodeRecord types we demonstrated a new solution, where we add an unspellable attribute to the type that identifies it as a specific object. We should do something like that here as well.
> 
> If you take a look at how GetNodeIOType is implemented that will give you a model for how this should be done.

HitObject is now modelled after the NodeRecord types as you suggested.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -537,6 +562,19 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
           NodeLaunchTy = DXIL::NodeLaunchType::Broadcasting;
       }
     }
+
+    // Shader Execution Reordering
+    // ReorderThread(..) handled elsewhere.
```
damyanp, line 567:
```suggestion
    // MaybeReorderThread(..) handled elsewhere.
```

tools/clang/test/SemaHLSL/hlsl/intrinsics/ser/reorder-unclaimed-pre-sm69.hlsl
```diff
@@ -0,0 +1,13 @@
+// RUN: %dxc -T lib_6_6 %s -verify
+// RUN: %dxc -T lib_6_8 %s -verify
+
+// Check that intrinsic names of Shader Execution Reordering are unclaimed pre SM 6.9.
+// expected-no-diagnostics
+
+void ReorderThread(uint CoherenceHint, uint NumCoherenceHintBitsFromLSB) {
```
damyanp, line 7:
```suggestion
void MaybeReorderThread(uint CoherenceHint, uint NumCoherenceHintBitsFromLSB) {
```

tools/clang/test/SemaHLSL/hlsl/intrinsics/ser/reorder-unclaimed-pre-sm69.hlsl
```diff
@@ -0,0 +1,13 @@
...
+}
+
+[shader("raygeneration")]
+void main() {
+  ReorderThread(15u, 4u);
```
damyanp, line 12:
```suggestion
  MaybeReorderThread(15u, 4u);
```

damyanp (changes requested):
Note, as per https://github.com/microsoft/hlsl-specs/pull/378, `ReorderThread` needs to be named `MaybeReorderThread`.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
simoll: Renamed to MaybeReorderThread in [36c1470](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/36c1470742bbb7677eed566e39e96aa1b4a54e00).  Resolved.

tools/clang/test/SemaHLSL/hlsl/intrinsics/ser/reorder-unclaimed-pre-sm69.hlsl
simoll: Renamed to MaybeReorderThread in [36c1470](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/36c1470742bbb7677eed566e39e96aa1b4a54e00).  Resolved.

tools/clang/lib/AST/ASTContextHLSL.cpp
llvm-beanz: > [b028aaf](https://github.com/microsoft/DirectXShaderCompiler/commit/b028aaf6b849d285f307dd7077e68e8bc46a4a56) moves the SER implementation to the `dx` namespace. @llvm-beanz

This change does the wrong thing. You've moved the implementation to a namespace which is not what was requested. The `HitObject` type should be emitted into the AST under the `dx` namespace.

simoll: [cefafa6](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/cefafa6a7d78fcff5488bac60cc848266c98e265)  moves the SER HitObject type and builtins to the `dx` namespace

.cache/clangd/index/dxcmain.cpp.6E359381028004C7.idx
llvm-beanz, line 1: Please remove the clangd indices.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7966,6 +7966,16 @@ def err_hlsl_unsupported_semantic_index: Error<
   "'%0' is defined with semantic index %1, but only values 0 through %2 are supported">;
 // HLSL Change Ends
 
+// SER Change Starts
+// Shader Execution Reordering
+def err_hlsl_ser_invalid_shader_kind : Error<
+   "Shader kind '%0' incompatible with shader execution reordering (has to be raygeneration, closesthit or miss)">;
+def err_hlsl_reorder_invalid_shader_kind : Error<
+   "Shader kind '%0' incompatible with MaybeReorderThread intrinsic (only in raygeneration)">;
```
llvm-beanz, line 7974: We should conditionalize this error to reduce how many new warnings we add. Changing the warning text to something like this would allow using a single diagnostic for diagnosing disallowed uses of `HitObject` and `MaybeReorderThread`.

```suggestion
   "%0 is unavailable in shader stage '%1 (requires %2)">;
```

tools/clang/lib/CodeGen/CodeGenTypes.cpp
```diff
@@ -365,8 +367,9 @@ llvm::Type *CodeGenTypes::ConvertType(QualType T) {
                  .getConstantArrayType(eltTy, llvm::APInt(32, count),
                                        ArrayType::ArraySizeModifier::Normal, 0)
                  .getTypePtr();
-      }
-      else
+      } else if (hlsl::IsHLSLHitObjectType(T)) {
+        return hlsl::dxilutil::GetHLSLHitObjectType(&TheModule);
+      } else
```
llvm-beanz, line 372: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
      } else if (hlsl::IsHLSLHitObjectType(T))
        return hlsl::dxilutil::GetHLSLHitObjectType(&TheModule);
      else
```

tools/clang/lib/Sema/SemaDecl.cpp
```diff
@@ -776,11 +776,18 @@ Sema::ClassifyName(Scope *S, CXXScopeSpec &SS, IdentifierInfo *&Name,
   bool IsFilteredTemplateName = false;
   
 Corrected:
+  // Allow namespace ::dx to obtain a ULE, which will be used to construct the
+  // actual intrinsic declaration in AddOverloadedCallCandidates.
+  NamespaceDecl *nsDecl =
+      SS.isValid() ? SS.getScopeRep()->getAsNamespace() : nullptr;
+  bool isDxNamespace = nsDecl && nsDecl->getIdentifier() &&
```
llvm-beanz, line 783: This code change is concerning. Needing custom name classification rules for the `dx` namespace seems wrong.

What declaration are you hitting that got you here?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -4809,11 +4996,17 @@ class HLSLExternalSource : public ExternalSemaSource {
           m_usedIntrinsics.insert(UsedIntrinsic(pIntrinsic, functionArgTypes));
       bool insertedNewValue = insertResult.second;
       if (insertedNewValue) {
+        NamespaceDecl *nsDecl = m_hlslNSDecl;
+        if (isVkNamespace)
+          nsDecl = m_vkNSDecl;
+        else if (isDxNamespace) {
+          nsDecl = m_dxNSDecl;
+        }
```
llvm-beanz, line 5004: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
        else if (isDxNamespace)
          nsDecl = m_dxNSDecl;
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -4757,10 +4922,17 @@ class HLSLExternalSource : public ExternalSemaSource {
...
     // Intrinsics live in the global namespace, so references to their names
     // should be either unqualified or '::'-prefixed.
-    // Exception: Vulkan-specific intrinsics live in the 'vk::' namespace.
-    if (isQualified && !isGlobalNamespace && !isVkNamespace) {
+    // Exceptions:
+    // - Vulkan-specific intrinsics live in the 'vk::' namespace.
+    // - DirectX-specific intrinsics live in the 'dx::' namespace.
+    if (isQualified && !isGlobalNamespace && !isVkNamespace && !isDxNamespace) {
       return false;
     }
```
llvm-beanz, line 4937: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
    if (isQualified && !isGlobalNamespace && !isVkNamespace && !isDxNamespace)
      return false;
```

tools/clang/lib/Sema/SemaOverload.cpp
```diff
@@ -10936,7 +10936,13 @@ bool Sema::buildOverloadedCallSet(Scope *S, Expr *Fn,
...
         ULE->getQualifier()->getAsNamespace()->getName() == "vk";
 
-    assert((!ULE->getQualifier() || isVkNamespace) && "non-vk qualified name with ADL");
+    bool isDxNamespace =
+        ULE->getQualifier() &&
+        ULE->getQualifier()->getKind() == NestedNameSpecifier::Namespace &&
+        ULE->getQualifier()->getAsNamespace()->getName() == "dx";
+
+    assert((!ULE->getQualifier() || isVkNamespace || isDxNamespace) &&
+           "non-vk qualified name with ADL");
```
llvm-beanz, line 10945:
```suggestion
           "non-namespace qualified name with ADL");
```

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
simoll: Applied the suggestion in [ba11d4f](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/ba11d4f3ffde25c5945852099caf9e983b2deb58)

tools/clang/lib/CodeGen/CodeGenTypes.cpp
simoll: Applied in [ba11d4f](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/ba11d4f3ffde25c5945852099caf9e983b2deb58)

tools/clang/lib/Sema/SemaHLSL.cpp
simoll: Applied in [ba11d4f](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/ba11d4f3ffde25c5945852099caf9e983b2deb58)

tools/clang/lib/Sema/SemaHLSL.cpp
simoll: Applied in [ba11d4f](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/ba11d4f3ffde25c5945852099caf9e983b2deb58)

tools/clang/lib/Sema/SemaOverload.cpp
simoll: Applied in [ba11d4f](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/ba11d4f3ffde25c5945852099caf9e983b2deb58)

.cache/clangd/index/dxcmain.cpp.6E359381028004C7.idx
simoll: Removed

tools/clang/lib/Sema/SemaDecl.cpp
simoll: Removed in [2bc823b](https://github.com/microsoft/DirectXShaderCompiler/pull/7097/commits/2bc823b0df962d343ea43e8be1a6246c67474f70). 'dx' builtins now implemented the same as 'vk' ones.

simoll:
> Note, as per [microsoft/hlsl-specs#378](https://github.com/microsoft/hlsl-specs/pull/378), `ReorderThread` needs to be named `MaybeReorderThread`.

Right, `ReorderThread` is now renamed to `MaybeReorderThread` everywhere

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7970,6 +7970,14 @@ def err_hlsl_unsupported_semantic_index: Error<
   "'%0' is defined with semantic index %1, but only values 0 through %2 are supported">;
 // HLSL Change Ends
 
+// SER Change Starts
+// Shader Execution Reordering
```
llvm-beanz, line 7974:
```suggestion
// HLSL Change Starts - Shader Execution Reordering
```

Alternatively you could remove the HLSL Change Ends comment above.

tools/clang/lib/CodeGen/CodeGenTypes.cpp
```diff
@@ -14,21 +14,23 @@
 #include "CodeGenTypes.h"
 #include "CGCXXABI.h"
 #include "CGCall.h"
+#include "CGHLSLRuntime.h" // HLSL Change
 #include "CGOpenCLRuntime.h"
 #include "CGRecordLayout.h"
+#include "CodeGenModule.h" // HLSL Change
 #include "TargetInfo.h"
+#include "dxc/DXIL/DxilUtil.h"
```
llvm-beanz, line 22:
```suggestion
#include "dxc/DXIL/DxilUtil.h" // HLSL Change
```

tools/clang/lib/CodeGen/CodeGenTypes.cpp
```diff
@@ -14,21 +14,23 @@
...
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/DeclCXX.h"
-#include "clang/AST/DeclTemplate.h"
 #include "clang/AST/DeclObjC.h"
+#include "clang/AST/DeclTemplate.h"
```
llvm-beanz, line 26: I know this isn't really a change so much as clang-format, but marking it makes comparing to Clang 3.7 easier.

```suggestion
#include "clang/AST/DeclTemplate.h" // HLSL Change - clang-format
```

tools/clang/lib/CodeGen/CodeGenTypes.cpp
```diff
@@ -14,21 +14,23 @@
...
 #include "clang/AST/Expr.h"
+#include "clang/AST/HlslTypes.h"
```
llvm-beanz, line 28:
```suggestion
#include "clang/AST/HlslTypes.h" // HLSL Change
```

tools/clang/lib/CodeGen/CodeGenTypes.cpp
```diff
@@ -365,7 +367,8 @@ llvm::Type *CodeGenTypes::ConvertType(QualType T) {
                  .getConstantArrayType(eltTy, llvm::APInt(32, count),
                                        ArrayType::ArraySizeModifier::Normal, 0)
                  .getTypePtr();
-      }
+      } else if (hlsl::IsHLSLHitObjectType(T))
+        return hlsl::dxilutil::GetHLSLHitObjectType(&TheModule);
```
llvm-beanz, line 371:
```suggestion
      } else if (hlsl::IsHLSLHitObjectType(T)) // HLSL Change
        return hlsl::dxilutil::GetHLSLHitObjectType(&TheModule);
```

tools/clang/lib/CodeGen/TargetInfo.cpp
```diff
@@ -17,6 +17,7 @@
 #include "CGCXXABI.h"
 #include "CGValue.h"
 #include "CodeGenFunction.h"
+#include "clang/AST/HlslTypes.h"
```
llvm-beanz, line 20:
```suggestion
```

tools/clang/lib/Sema/SemaCXXScopeSpec.cpp
```diff
@@ -11,15 +11,16 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "clang/Sema/SemaInternal.h"
 #include "TypeLocBuilder.h"
 #include "clang/AST/ASTContext.h"
 #include "clang/AST/DeclTemplate.h"
 #include "clang/AST/ExprCXX.h"
+#include "clang/AST/HlslTypes.h"
```
llvm-beanz, line 18:
```suggestion
#include "clang/AST/HlslTypes.h" // HLSL Change
```

tools/clang/lib/Sema/SemaCXXScopeSpec.cpp
```diff
@@ -11,15 +11,16 @@
...
 #include "clang/AST/NestedNameSpecifier.h"
 #include "clang/Basic/PartialDiagnostic.h"
 #include "clang/Sema/DeclSpec.h"
 #include "clang/Sema/Lookup.h"
+#include "clang/Sema/SemaInternal.h"
```
llvm-beanz, line 23:
```suggestion
#include "clang/Sema/SemaInternal.h" // HLSL Change - clang-format
```

tools/clang/lib/Sema/SemaCXXScopeSpec.cpp
```diff
@@ -200,9 +201,13 @@ bool Sema::RequireCompleteDeclContext(CXXScopeSpec &SS,
...
   // If we're currently defining this type, then lookup into the
   // type is okay: don't complain that it isn't complete yet.
+  // HLSL: If this is a builtin type, add all method definitions and complete
+  // the type.
   QualType type = Context.getTypeDeclType(tag);
   const TagType *tagType = type->getAs<TagType>();
-  if (tagType && tagType->isBeingDefined())
+  if (tagType && tagType->isBeingDefined() &&
+      (hlsl::IsUserDefinedRecordType(type) ||
+       hlsl::IsHLSLObjectWithImplicitMemberAccess(type)))
```
llvm-beanz, line 210: This change doesn't make sense to me. Can you explain why this is required?

If we get here and `tagType` a builtin type that is not yet completed, we should go down into `RequireCompleteType` below which will call into the external sema source to complete it. I'm confused why we need additional early-outs here, and the comment above makes it more confusing.

Also changes in this file should be wrapped in `// HLSL Change Begin|End` comments.

tools/clang/lib/Sema/SemaDecl.cpp
```diff
@@ -804,7 +804,7 @@ Sema::ClassifyName(Scope *S, CXXScopeSpec &SS, IdentifierInfo *&Name,
         return BuildDeclarationNameExpr(SS, Result, /*ADL=*/false);
       }
     }
-
```
llvm-beanz, line 807: Please revert this whitespace only change.

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-target-profile-error.hlsl
```diff
@@ -0,0 +1,9 @@
+// RUN: not %dxc -T lib_6_8 %s 2>&1 | FileCheck %s
+
+namespace dx {}
```
llvm-beanz, line 3: What happens if you remove this? That would be the more common situation.

simoll:
> @simoll - do you know why all the builds are failing for this PR?

It's because i've rebased the PR onto tot and tot was failing at that time:
`  FAIL: Clang :: CodeGenDXIL/hlsl/intrinsics/dot.hlsl (4 of 2488)`

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11529,6 +11746,60 @@ static bool isStringLiteral(QualType type) {
...
+    break;
+  case DXIL::ShaderKind::RayGeneration:
+  case DXIL::ShaderKind::ClosestHit:
+  case DXIL::ShaderKind::Miss:
+    ValidEntryForHitObject = true;
+    break;
+  }
+
+  // Return for anything that is not a HitObject intrinsic.
+  std::string FeatureStr = "Shader Execution Reordering";
```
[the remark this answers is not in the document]
simoll: Using %select in latest version

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
simoll: Removed the HLSL Change end marker in latest version

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-target-profile-error.hlsl
simoll: Kept this test - but added another one that specifically checks that a namespace 'dx' is not predeclared pre-SM6.9

lib/HLSL/HLOperationLower.cpp
```diff
@@ -6064,6 +6064,24 @@ Value *TranslateUnpack(CallInst *CI, IntrinsicOp IOP, OP::OpCode opcode,
 
 } // namespace
 
+// Shader Execution Reordering.
+namespace {
+Value *TranslateHitObjectMake(CallInst *CI, IntrinsicOp IOP, OP::OpCode opcode,
+                              HLOperationLowerHelper &helper,
+                              HLObjectOperationLowerHelper *pObjHelper,
+                              bool &Translated) {
+  return UndefValue::get(CI->getType()); // TODO: Merge SER DXIL patches
```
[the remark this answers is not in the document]
simoll: Added test stubs in latest revision for MaybeReorderThread, MakeNop and the HitObject type

tools/clang/lib/Sema/SemaCXXScopeSpec.cpp
simoll: The current HLSL-specific modifications to scoped lookup do not support 'static'. I've changed my approach and eagerly complete the HitObject type (including its static functions) instead. This means, no special handling for HitObject is any longer necessary.
This specific change is gone now - considering this resolved.

simoll:
> Missing tests:

I've added the requestedtests:

> 
> 1. IR test for the HLOperationLowering changes
- tools\clang\test\HLSLFileCheck\hlsl\objects\HitObject\maybereorder.hlsl
- tools\clang\test\HLSLFileCheck\hlsl\objects\HitObject\hitobject_make.hlsl

> 2. HitObject and new APIs used without `dx` namespace (should error)
tools\clang\test\SemaHLSL\hlsl\objects\HitObject\maybereorderthread-without-namespace.hlsl
- tools\clang\test\SemaHLSL\hlsl\objects\HitObject\hitobject-without-namespace.hlsl

> 3. HitObject and new APIs used with `using namespace dx` instead of explicit namespace
- tools\clang\test\SemaHLSL\hlsl\objects\HitObject\hitobject-using-namespace.hlsl

llvm-beanz:
> > Missing tests:
> 
> I've added the requested tests:
> 
> > 1. IR test for the HLOperationLowering changes
> 
> * tools\clang\test\HLSLFileCheck\hlsl\objects\HitObject\maybereorder.hlsl
> * tools\clang\test\HLSLFileCheck\hlsl\objects\HitObject\hitobject_make.hlsl

These are HLSL tests, we should have IR tests for IR passes. IR tests start as IR and verify transformations in isolation.

Please see the tests here as examples: https://github.com/microsoft/DirectXShaderCompiler/tree/main/tools/clang/test/DXC/Passes

llvm-beanz:
@simoll the bots are failing on one of your tests:
> D:\a\1\s\tools\clang\test\HLSL\..\HLSLFileCheck\hlsl\objects\HitObject\maybereorder.hlsl [Failed]

simoll:
> 
> These are HLSL tests, we should have IR tests for IR passes. IR tests start as IR and verify transformations in isolation.
> 
> Please see the tests here as examples: https://github.com/microsoft/DirectXShaderCompiler/tree/main/tools/clang/test/DXC/Passes

Added dxilgen pass tests that cover all HitObject types and intrinsics in this PR:
* tools\clang\test\DXC\Passes\DxilGen\maybereorder_dxilgen.ll
* tools\clang\test\DXC\Passes\DxilGen\hitobject_dxilgen.ll

> @simoll the bots are failing on one of your tests:

Fixed in latest version. Output mismatch between release and debug build leading to CHECK: clauses not matching in release

@llvm-beanz - back to you

llvm-beanz (approved):
I'm fine with this as it is now. We'll need to shore up the IR testing as subsequent PRs come in.

@tex3d or @pow2clk, any feedback?

damyanp:
@simoll - while we're waiting on @tex3d or @pow2clk's review, I notice that there's a merge conflict that needs resolving.

tex3d (changes requested):
Partial review, but there are a couple blocking issues.
- A breaking change to HLSL_INTRINSICS should be separated and made by MSFT to synchronize internal dependencies, see #7198
- I have requested changing to availability attributes for shader model version check which will require additional changes described by issue #7198 and incoming changes to check the availability attributes from #7156. This should eliminate the need for the special `EnableShaderExecutionReordering` flag added to various option classes as well.
- I'm curious why this has broken with the pattern for object CodeGen and later lowering to DXIL types used with other object types.

utils/hct/hctdb_instrhelp.py
```diff
@@ -971,7 +972,7 @@ def get_hlsl_intrinsics():
         if last_ns != i.ns:
             last_ns = i.ns
             id_prefix = (
-                "IOP" if last_ns == "Intrinsics" or last_ns == "VkIntrinsics" else "MOP"
+                "IOP" if last_ns.endswith("Intrinsics") else "MOP"
```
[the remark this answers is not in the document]
simoll: Removed in df1e26fd4

tools/clang/include/clang/Basic/LangOptions.h
```diff
@@ -162,6 +162,7 @@ class LangOptions : public LangOptionsBase {
   bool EnableDX9CompatMode = false;
   bool EnableFXCCompatMode = false;
   bool EnablePayloadAccessQualifiers = false;
+  bool EnableShaderExecutionReordering = false;
```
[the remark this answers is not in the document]
simoll: Removed SER opts in c390f311e

tools/clang/include/clang/Frontend/CodeGenOptions.h
```diff
@@ -244,6 +244,8 @@ class CodeGenOptions : public CodeGenOptionsBase {
   bool HLSLEnableLifetimeMarkers = false;
   /// Put shader sources and options in the module
   bool HLSLEmbedSourcesInModule = false;
+  /// Enable Shader Execution Reordering.
+  bool HLSLEnableShaderExecutionReordering = false;
```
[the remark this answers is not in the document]
simoll: Using SM instead of opts in c390f311e

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -3591,6 +3749,12 @@ class HLSLExternalSource : public ExternalSemaSource {
         recordDecl = DeclareConstantBufferViewType(*m_context, /*bTBuf*/ true);
       } else if (kind == AR_OBJECT_RAY_QUERY) {
         recordDecl = DeclareRayQueryType(*m_context);
+      } else if (kind == AR_OBJECT_HIT_OBJECT) {
+        // Declare 'HitObject' in '::dx' extension namespace.
+        if (EnableSER) {
```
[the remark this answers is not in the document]
simoll: Blocked until https://github.com/microsoft/DirectXShaderCompiler/pull/7156 has panded

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -3894,6 +4071,9 @@ class HLSLExternalSource : public ExternalSemaSource {
       AddIntrinsicTableMethods(intrinsic);
     }
 
+    if (SM->IsSM69Plus())
```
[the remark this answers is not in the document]
simoll: Blocked until https://github.com/microsoft/DirectXShaderCompiler/pull/7156 has landed

damyanp:
@simoll - just to let you know, #7156 was resolved via #7199.  I think #7156 also has some parts that might be of interest to you wrt availability attributes.  @tex3d can probably fill in more details.

simoll:
> @simoll - just to let you know, #7156 has landed. #7199 also has some parts that might be of interest to you wrt availability attributes. @tex3d can probably fill in more details.

Great! I'll rebase and make changes to use the availability attributes

include/dxc/dxcapi.internal.h
```diff
@@ -165,6 +167,7 @@ struct HLSL_INTRINSIC {
   BOOL bReadOnly;          // Only read memory
   BOOL bReadNone;          // Not read memory
   BOOL bIsWave;            // Is a wave-sensitive op
+  BOOL bStaticMember;      // HLSL static member function
```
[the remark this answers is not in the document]
simoll: I made the following changes after the PR was merged:
- introduced a new `INTRIN_FLAG_STATIC_MEMBER` to represent static  member functions
- added min_sm=6.9 attribute to the SER intrinsic declarations

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
simoll: Adding the intrinsic unconditionally with an AvailabilityAttribute now. Considering this resolved.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
simoll: Same as for the other instance: always declaring the type and using availability attributes instead.

tools/clang/lib/CodeGen/CodeGenTypes.cpp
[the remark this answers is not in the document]
simoll: Is the suggestion to use the standard clang type codegen here and perform type replacement in the hlsl sroa pass?
If we remove this change, we assert in the ScalarReplAggregatesHLSL.cpp :: RewriteCall special case for RayQuery rewriting.

include/dxc/DXIL/DxilUtil.h
```diff
@@ -162,6 +162,8 @@ GetHLSLResourceProperties(llvm::Type *Ty);
 bool IsHLSLResourceType(llvm::Type *Ty);
 bool IsHLSLObjectType(llvm::Type *Ty);
 bool IsHLSLRayQueryType(llvm::Type *Ty);
+llvm::Type *GetHLSLHitObjectType(llvm::Module *M);
+bool IsHLSLHitObjectType(llvm::Type *Ty);
```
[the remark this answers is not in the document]
simoll: This is used by IsHLSLObjectType . I commented below on HitObject type lowering

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -1243,6 +1243,10 @@ CXXRecordDecl *hlsl::DeclareHitObjectType(NamespaceDecl &NSDecl) {
       static_cast<int>(hlsl::IntrinsicOp::MOP_DxHitObject_MakeNop)));
   pConstructorDecl->addAttr(HLSLCXXOverloadAttr::CreateImplicit(Context));
 
+  // Shader Execution Reordering requires SM6.9
+  VersionTuple VT69 = VersionTuple(6, 9);
+  pConstructorDecl->addAttr(ConstructAvailabilityAttribute(Context, VT69));
```
bob80905, line 1248: Would you be able to add an ast-dump test to verify this new attribute in the AST? Same with the attribute you add in SemaHLSL.cpp now.
Separately, I'm not sure how this attribute is useful. Does `dx::HitObject::MakeNop();` create a constructor Decl? If not, I don't see where this attribute is actually generating an error when SM < 6.9, and I suspect the other attribute you added in SemaHLSL is actually doing the work.
For example, if you were to change this VT69 to, say, VT63, would any of the errors in your tests go away? Which ones would change?

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-unavailable-pre-sm69.hlsl
```diff
@@ -1,20 +1,9 @@
...
-    HitObject hit;
-    hit.notTheSM69HitObject = 1;
-    return hit;
-  }
-};
-}
 
 [shader("raygeneration")]
 void main() {
+  // expected-warning@+1{{potential misuse of built-in function 'dx::HitObject::MakeNop' in shader model lib_6_8; introduced in shader model 6.9}}
```
bob80905, line 7: Not sure if HitObject::Miss is in yet or not, but I imagine a similar test would need to be added in that case. Just a comment FYI, no resolution needed.

tools/clang/lib/AST/ASTContextHLSL.cpp
simoll: I've removed the AvailabilityAttribute from the constructor.. it was unused.

I've added ast tests for `HitObject::MakeNop` and `MaybeReorderThread` to check for the AvailabilityAttribute as you suggested.

simoll:
> > @simoll - just to let you know, #7156 has landed. #7199 also has some parts that might be of interest to you wrt availability attributes. @tex3d can probably fill in more details.
> 
> Great! I'll rebase and make changes to use the availability attributes

This is using AvailabilityAttributes now. There are ast tests to check for the presence of the attribute and there are tests to check for the availability diagnostic when SER is used pre-SM6.9.

tools/clang/lib/CodeGen/CodeGenTypes.cpp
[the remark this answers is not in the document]
damyanp: @simoll - can you file an issue on this and we can address this later.  Thanks!

include/dxc/DXIL/DxilUtil.h
```diff
@@ -162,6 +162,8 @@ GetHLSLResourceProperties(llvm::Type *Ty);
 bool IsHLSLResourceType(llvm::Type *Ty);
 bool IsHLSLObjectType(llvm::Type *Ty);
 bool IsHLSLRayQueryType(llvm::Type *Ty);
+llvm::Type *GetHLSLHitObjectType(llvm::Module *M);
```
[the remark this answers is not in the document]
damyanp: This will be resolved as part of https://github.com/microsoft/DirectXShaderCompiler/pull/7097#discussion_r2003791199

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11554,6 +11845,8 @@ void Sema::DiagnoseReachableHLSLCall(CallExpr *CE, const hlsl::ShaderModel *SM,
   if (!IsBuiltinTable(IntrinsicAttr->getGroup()))
     return;
 
+  DiagnoseReachableCallForSER(CE, EntrySK, EntryDecl, SM);
```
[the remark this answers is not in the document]
damyanp: @simoll  - we discussed this today, will make this change.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -430,13 +433,27 @@ class HLSLCallDiagnoseVisitor // Could rename to HLSLReachableDiagnoseVisitor
...
+      return;
+
+    // TBD: Determine best way to distinguish between builtin constant decls
+    // and other decls.
+    enum class DiagKind { Constant = 0, Function = 1 };
+    DiagKind Kind;
+    switch (ND->getKind()) {
+    case Decl::Function:
+    case Decl::FunctionTemplate:
+    case Decl::CXXMethod:
```
tex3d, line 446: I think `isa<FunctionDecl>` is all we need.

tools/clang/test/DXC/Passes/DxilGen/hitobject_dxilgen.ll
```diff
@@ -0,0 +1,100 @@
+; RUN: %dxopt %s -hlsl-passes-resume -dxilgen -S | FileCheck %s
```
tex3d, line 1:
```suggestion
; RUN: %dxopt %s -hlsl-passes-resume -dxilgen -S | FileCheck %s
; REQUIRES: dxil-1-9
```

tools/clang/test/DXC/Passes/DxilGen/maybereorder_dxilgen.ll
```diff
@@ -0,0 +1,105 @@
+; RUN: %dxopt %s -hlsl-passes-resume -dxilgen -S | FileCheck %s
```
tex3d, line 1:
```suggestion
; RUN: %dxopt %s -hlsl-passes-resume -dxilgen -S | FileCheck %s
; REQUIRES: dxil-1-9
```

tools/clang/test/HLSLFileCheck/hlsl/objects/HitObject/hitobject_make.hlsl
```diff
@@ -0,0 +1,11 @@
+// RUN: %dxc -T lib_6_9 -E main %s | FileCheck %s
```
tex3d, line 1: All these tests currently under `HLSLFileCheck` can be moved under the `CodeGenDXIL` path to make them ordinary lit shell tests.  This is something we are generally advising for new tests that aren't required to run under this path due to special tooling only available here still.

When doing so, you need to add `// REQUIRES: dxil-1-9`

```suggestion
// RUN: %dxc -T lib_6_9 -E main %s | FileCheck %s
// REQUIRES: dxil-1-9
```

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-in-buffer.hlsl
```diff
@@ -0,0 +1,4 @@
+// RUN: %dxc -T lib_6_9 %s -verify
+
+// expected-error@+1{{'dx::HitObject' is an object and cannot be used as a type parameter}}
+RWStructuredBuffer<dx::HitObject> InvalidBuffer;
```
tex3d, line 4: Should probably also block it from being used as a global constant buffer variable.

simoll: Yes, but i see no infrastructure / pre-existing diagnostic in DXC to do this. I suggest we open an issue for this to address later.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7662,8 +7662,8 @@ def warn_hlsl_rayquery_flags_conflict : Warning<
   " must have RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set.">, DefaultError, InGroup<HLSLRayQueryFlags>;
 def err_hlsl_unsupported_builtin_op: Error<
   "operator cannot be used with built-in type %0">;
-def warn_hlsl_builtin_constant_unavailable: Warning<
-  "potential misuse of built-in constant %0 in shader model %1; introduced"
+def warn_hlsl_builtin_unavailable: Warning<
```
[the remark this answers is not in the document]
simoll: This uses `warn_hlsl_intrinsic_in_wrong_shader_model` now . Removed the ser-specific diagnostic.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -1816,15 +1941,14 @@ static void AddHLSLIntrinsicAttr(FunctionDecl *FD, ASTContext &context,
...
-  //      clang::VersionTuple(), clang::VersionTuple(), false,
-  //      "HLSL Intrinsic availability limited by shader model."));
-  //}
+  if (pIntrinsic->MinShaderModel) {
+    unsigned Major = pIntrinsic->MinShaderModel >> 4;
+    unsigned Minor = pIntrinsic->MinShaderModel & 0xF;
+    FD->addAttr(AvailabilityAttr::CreateImplicit(
+        context, &context.Idents.get(""), clang::VersionTuple(Major, Minor),
+        clang::VersionTuple(), clang::VersionTuple(), false,
+        "HLSL Intrinsic availability limited by shader model."));
```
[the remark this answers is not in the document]
simoll: > This results in less code and complexity than the current approach, I think.

Simplified following the latest suggestion.

> It turns out this string will be unused, so we can leave it empty ("").

Replaced with "" and updated tests.

Resolved.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7991,6 +7991,12 @@ def warn_hlsl_legacy_integer_literal_signedness: Warning<
   InGroup<HLSLLegacyLiterals>, DefaultIgnore;
 def err_hlsl_unsupported_semantic_index: Error<
   "'%0' is defined with semantic index %1, but only values 0 through %2 are supported">;
+
+// Shader Execution Reordering
+def err_hlsl_ser_unsupported : Error<
+   "%select{dx::HitObject|dx::MaybeReorderThread}0 is unavailable in shader stage '%1' (requires 'raygeneration'%select{|, 'closesthit' or 'miss'}2)">;
```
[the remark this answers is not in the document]
simoll: There are two diagnostics now as suggested.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -430,13 +433,27 @@ class HLSLCallDiagnoseVisitor // Could rename to HLSLReachableDiagnoseVisitor
...
+      Kind = DiagKind::Function;
+      break;
+    default:
+      Kind = DiagKind::Constant;
+      break;
     }
+
+    sema->Diag(Loc, diag::warn_hlsl_builtin_unavailable)
```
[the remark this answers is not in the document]
simoll: Using that diagnostic now and removed the SER-specific diagnostic.

tools/clang/lib/CodeGen/CodeGenTypes.cpp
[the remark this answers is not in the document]
simoll: Yep, tracking this in issue #7233 - [SER] Do not short-cut to %dx.types.HitObject DXIL type in HL from AST codegen

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-in-buffer.hlsl
simoll: @tex3d - Tracking this change request in #7234 . Can we consider this non-blocking for this PR?

simoll:
@tex3d  - you made the suggestion to move some SER tests out of HLSLFileCheck to CodeGenDXIL (the exact comment got lost AFAICT). Is this necessary?

tex3d:
> @tex3d - you made the suggestion to move some SER tests out of HLSLFileCheck to CodeGenDXIL (the exact comment got lost AFAICT). Is this necessary?

No, not absolutely necessary.  Generally, we are trying to avoid adding new tests under HLSLFileCheck unless necessary for testing infrastructure reasons (D3DReflect testing, for instance).  We could create a follow-up issue for bulk test reorg later if it helps the flow of other commits on top of this one.

One thing that will help with that move is the addition of the `REQUIRES: dxil-1-9`.  Under HLSLFileCheck, there is automatic detection of dxc compilation shader target arguments, which causes skips if, for instance, the output is tested against a DXIL validator that doesn't meet the requirement.  IR opt tests should not require this, `-fcgl` tests should not require this, but ones that  compile to a shader target with `dxc` do.

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject-in-buffer.hlsl
tex3d: Yes, sounds good.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7665,6 +7665,9 @@ def err_hlsl_unsupported_builtin_op: Error<
 def warn_hlsl_builtin_constant_unavailable: Warning<
   "potential misuse of built-in constant %0 in shader model %1; introduced"
   " in shader model %2">, InGroup<HLSLAvailabilityConstant>;
+def warn_hlsl_builtin_type_unavailable: Warning<
+  "potential misuse of built-in type %0 in shader model %1; introduced"
+  " in shader model %2">, InGroup<HLSLAvailability>;
```
tex3d, line 7670: Should we add `DefaultError` to make this an error by default?

tex3d (approved):
Looks good!

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
simoll: > Should we add DefaultError to make this an error by default?

I don't think so given that the type may be used on dead code paths in HLSL and optimized out before it is caught by the DXIL validator. The DXIL validator change we track in #7234

simoll: Turned this into a `DefaultError` - we do the same for the other diags of this kinds (except bui;ltin constants).Resolved,

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
simoll: Resolved, implemented as shown here

Issue #7079, opened 2025-01-22
Load/Store intrinsics for Raw and Typed buffers are inconsistent in allowing vec1 indexing


pow2clk:
**Description**
Equivalent intrinsics on different buffers and sometimes within the same buffer kind will accept or reject single-element vectors as indices with no consistency.

Since the clang implementation will implicitly truncate such vectors to scalars in all cases, the most forward compatible solution is to allow these in all the cases.

**Steps to Reproduce**
https://godbolt.org/z/Y7cYo5vPs includes a long list of buffers with their full list of load/store operations used with int1 indices, which can be toggled to scalars with the change of a single `#if 0` on line 9.

These are pretty simple to reproduce though, any shader that passes in a vec1, however trivial to eliminate will fail if the method in question isn't among those that happen to be supported.

**Actual Behavior**
Structured buffers and typed buffers accept vec1s for subscripts whether used for retrieval or assignment. ByteAddress and Structured buffers both reject them for any Load or Store method.
Typed buffers accept vec1s in all cases except for RWBuffer Loads. This is simply because of how they are defined in gen_intrin_main.txt.

**Environment**
- DXC version 1.8.2407
- Host Operating System MacOs 14.7.2 Also Linux something(dot)something. Yes, that sounds right.

pow2clk:
Sidenote, this was an obstacle to getting sensible testing for long vectors or it wouldn't have been important enough to pay attention to in DXC.

PR #7114, opened 2025-02-03
Revise overload resolution for splats/truncations

pow2clk:
Allow truncations when matching arguments for intrinsic overloads. This eliminates the need for explicit scalar extractions from vectors for arguments that are scalar by nature. This encompasses any vectors passed for scalars, allowing the truncation, but emitting a warning the same as is done for other assignments of vectors to scalars.

This maintains splats as the preferred transformations and promotes perfect matches to be preferred over that. This has the effect of removing the need to carefully order intrinsics to ensure that the right variant gets matched first before another one incorrectly takes its place with a faulty cast.

Allowing truncations causes a problems with a small subset of intrinsics that have explicit overloads for various matrix,vector, scalar combinations. Namely the mul overloads. These could be simplified to accept a new range of template types except the dimensions need to be matched in unconventional ways.

For these, the notion of uncastable or "ONLY" variants of the template/layout types are introduced. These are indicated with a trailing "!" after the parameter typename in gen_intrin_main, which directs them to an array that contains a NOCAST enum that, when encountered, will skip the attempts to splat or truncate.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7079


pow2clk:
Alternate, but inferior approach can be viewed here #7115.

include/dxc/dxcapi.internal.h
```diff
@@ -44,10 +44,13 @@ enum LEGAL_INTRINSIC_TEMPLATES {
   LITEMPLATE_MATRIX = 3, // Matrix types (eg. float3x3).
   LITEMPLATE_ANY =
       4, // Any one of scalar, vector or matrix types (but not object).
-  LITEMPLATE_OBJECT = 5, // Object types.
-  LITEMPLATE_ARRAY = 6,  // Scalar array.
+  LITEMPLATE_OBJECT = 5,      // Object types.
+  LITEMPLATE_ARRAY = 6,       // Scalar array.
+  LITEMPLATE_SCALAR_ONLY = 7, // Uncastable scalar types.
+  LITEMPLATE_VECTOR_ONLY = 8, // Uncastable vector types (eg. float3).
+  LITEMPLATE_MATRIX_ONLY = 9, // Uncastable matrix types (eg. float3x3).
```
pow2clk, line 51: Castability here refers to the ability to truncate or splat something to change it's "shape" (or template or layout, we use a few terms) and not the castability of the contents.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -4213,22 +4211,6 @@ void TranslateLoad(ResLoadHelper &helper, HLResource::Kind RK,
...
-
-      loadArgs.emplace_back(scalarOffset); // offset
-    } else {
-      // TODO: calculate the real address based on opcode
-
-      loadArgs.emplace_back(helper.addr); // offset
-    }
-  }
-  // offset 0
-  if (opcode == OP::OpCode::TextureLoad) {
```
pow2clk, line 4231: This was some simplification that came naturally with the removal of the extract operator here. The opcode is determined completely by the resource kind so a test for one is the same as a test for the other.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -4420,12 +4400,7 @@ void TranslateStore(DxilResource::Kind RK, Value *handle, Value *val,
   if (RK == DxilResource::Kind::RawBuffer ||
       RK == DxilResource::Kind::TypedBuffer) {
     // Offset 0
-    if (offset->getType()->isVectorTy()) {
-      Value *scalarOffset = Builder.CreateExtractElement(offset, (uint64_t)0);
-      storeArgs.emplace_back(scalarOffset); // offset
-    } else {
-      storeArgs.emplace_back(offset); // offset
-    }
+    storeArgs.emplace_back(offset); // offset
```
pow2clk, line 4403: One of several explicit "casts" in the form of a vector to scalar truncation that are now done in the default code paths instead of relying on special cases.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -989,9 +990,18 @@ static const ArTypeObjectKind g_NullTT[] = {AR_TOBJ_VOID, AR_TOBJ_UNKNOWN};
...
+    AR_TOBJ_VECTOR, AR_TOBJ_NOCAST, AR_TOBJ_UNKNOWN};
+
+static const ArTypeObjectKind g_MatrixOnlyTT[] = {
+    AR_TOBJ_MATRIX, AR_TOBJ_NOCAST, AR_TOBJ_UNKNOWN};
+
 const ArTypeObjectKind *g_LegalIntrinsicTemplates[] = {
-    g_NullTT, g_ScalarTT, g_VectorTT, g_MatrixTT,
-    g_AnyTT,  g_ObjectTT, g_ArrayTT,
+    g_NullTT,   g_ScalarTT, g_VectorTT,     g_MatrixTT,     g_AnyTT,
+    g_ObjectTT, g_ArrayTT,  g_ScalarOnlyTT, g_VectorOnlyTT, g_MatrixOnlyTT,
```
pow2clk, line 1004: Automatically generated code for intrinsics will assign indexes into this array that will use the `Only` variants that have a `NOCAST` entry which will stop iterating through possibilities in a way that makes it clear that it failed due to requiring a cast where none is allowed.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6113,7 +6123,7 @@ bool HLSLExternalSource::MatchArguments(
   ArBasicKind
       ComponentType[MaxIntrinsicArgs]; // Component type for each argument,
                                        // AR_BASIC_UNKNOWN if unspecified.
-  UINT uSpecialSize[IA_SPECIAL_SLOTS]; // row/col matching types, UNUSED_INDEX32
+  UINT uSpecialSize[IA_SPECIAL_SLOTS]; // row/col matching types, UnusedSize
```
pow2clk, line 6126: Incidental, just an incorrect and misleading comment.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6292,6 +6305,11 @@ bool HLSLExternalSource::MatchArguments(
       }
     }
 
+    // If the intrinsic parameter has variable rows or columns but must match
+    // other argument dimensions, it will be specified in pIntrinsicArg with
+    // a special value indicating that the dimension depends on passed values.
+    // uSpecialSize stores the dimensions of the actual passed type.
```
pow2clk, line 6311: Here and below, I just try to better document what is actually happening here for my own and others' benefit.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6398,18 +6416,39 @@ bool HLSLExternalSource::MatchArguments(
...
+            Template[i] = *pTT;
             break;
+          } else if (AR_TOBJ_VECTOR == Template[i] && AR_TOBJ_SCALAR == *pTT) {
+            // If a vector was passed in and the expected value was scalar
+            // convert to the template type for a truncation.
+            // Only applicable to ScalarTT,
+            // since the scalar has to be first in the list.
+            Template[i] = AR_TOBJ_SCALAR;
+            break;
+          }
```
pow2clk, line 6446: This is the crux of the change. Instead of checking for splats first, we check for perfect matches first. If that fails due to running out of options, we'll try splats AND truncations (which, in spite of a comment suggesting otherwise, were not allowed here). However, if the ending enum is `NOCAST`, then we don't allow the splat or truncation and the match fails.

The reordering of the splat check won't change any previous behavior directly as it only kicked in when only a single template type was allowed in the first place. It was necessary for the new behavior allowing NOCAST though.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+
+template<typename T, int N>
+struct OffVector {
+  float4 pad1;
+  double pad2;
+  vector<T, N> v;
+  OffVector operator+(OffVector vec) {
+    OffVector ret;
+    ret.pad1 = 0.0;
+    ret.pad2 = 0.0;
```
pow2clk, line 75: By only using the final vector, I don't have to check for as many loads for this padding, which is just there to force an offset requirement.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+
+  // CHECK-DAG: [[HDLCON:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 4, i32 4, i32 0, i8 1 }, i32 4, i1 false)
+  // CHECK-DAG: [[HDLAPP:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 5, i32 5, i32 0, i8 1 }, i32 5, i1 false)
+
+  // CHECK-DAG: [[IX0:%.*]] = call i32 @dx.op.loadInput.i32(i32 4, i32 {{[0-9]*}}, i32 [[BOFF:0]]
+  // CHECK-DAG: [[RIX0:%.*]] = call i32 @dx.op.loadInput.i32(i32 4, i32 {{[0-9]*}}, i32 [[BOFF]]
+  
+
+  // CHECK: [[ANHDLRWBY:%.*]] = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle [[HDLRWBY]]
+  // OFF: [[RIX0:%.*]] = add i32 [[IX0]], [[BOFF:[0-9]+]]
```
pow2clk, line 127: RIX is the "raw" index for byteaddress buffers, which aren't allowed to use any offset, so they have to add it to the index.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+  // CHECK: call %dx.types.ResRet.[[TY]] @dx.op.rawBufferLoad.[[TY]](i32 139, %dx.types.Handle [[ANHDLRWBY]], i32 [[RIX0]]
+  // MAT: [[IX0p4:%.*]] = add i32 [[RIX0]], [[p4:[0-9]+]]
+  // MAT: call %dx.types.ResRet.[[TY]] @dx.op.rawBufferLoad.[[TY]](i32 139, %dx.types.Handle [[ANHDLRWBY]], i32 [[IX0p4]]
+  // MAT: [[IX0p8:%.*]] = add i32 [[RIX0]], [[p8:[0-9]+]]
+  // MAT: call %dx.types.ResRet.[[TY]] @dx.op.rawBufferLoad.[[TY]](i32 139, %dx.types.Handle [[ANHDLRWBY]], i32 [[IX0p8]]
```
pow2clk, line 132: The larger matrix types require multiple loads  because raw buffer loads are limited to 4 elements at a time. (for now). This checks for them.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+  // I1: icmp ne i32 %{{.*}}, 0
+  // I1: icmp ne i32 %{{.*}}, 0
+  DECL(TYPE,NUM,babElt2) = RoByBuf.Load< TYNAME(TYPE,NUM) >(IX[0]);
+
+  // I1: zext i1 %{{.*}} to i32
+  // I1: zext i1 %{{.*}} to i32
+  // I1: zext i1 %{{.*}} to i32
+  // I1: zext i1 %{{.*}} to i32
+  // OFF: call void @dx.op.rawBufferStore.f32(i32 140, %dx.types.Handle [[ANHDLRWBY]], i32 {{%.*}}, i32 undef, float 0.0
+  // OFF: call void @dx.op.rawBufferStore.f64(i32 140, %dx.types.Handle [[ANHDLRWBY]], i32 {{%.*}}, i32 undef, double 0.0
```
pow2clk, line 154: We don't load the padding because we don't use it, but we do store it with zeroes.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+  // CHECK: call void @dx.op.rawBufferStore.[[TY]](i32 140, %dx.types.Handle [[ANHDLRWBY]], i32 [[RIX0]]
+  // MAT: call void @dx.op.rawBufferStore.[[TY]](i32 140, %dx.types.Handle [[ANHDLRWBY]], i32 [[IX0p4]]
+  // MAT: call void @dx.op.rawBufferStore.[[TY]](i32 140, %dx.types.Handle [[ANHDLRWBY]], i32 [[IX0p8]]
+  RwByBuf.Store< TYNAME(TYPE,NUM) >(IX[0], Add(babElt1, babElt2));
+
+  // StructuredBuffer Tests
+  // CHECK: [[ANHDLRWST:%.*]] = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle [[HDLRWST]]
+  // CHECK: call %dx.types.ResRet.[[TY]] @dx.op.rawBufferLoad.[[TY]](i32 139, %dx.types.Handle [[ANHDLRWST]], i32 [[IX0]], i32 [[BOFF]]
```
pow2clk, line 162: Unlike the above BABuffers, Structured buffers are allowed to use the offset parameter and it's used both to skip the padding in the case of structure elements and also to iterate over multiple 4-element loads. `BOFF` is the base offset that puts you just after the padding.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+  // CHECK: call void @dx.op.rawBufferStore.[[TY]](i32 140, %dx.types.Handle [[ANHDLRWST]], i32 [[IX0]], i32 [[BOFF]]
+  // MAT: call void @dx.op.rawBufferStore.[[TY]](i32 140, %dx.types.Handle [[ANHDLRWST]], i32 [[IX0]], i32 [[p4]]
+  // MAT: call void @dx.op.rawBufferStore.[[TY]](i32 140, %dx.types.Handle [[ANHDLRWST]], i32 [[IX0]], i32 [[p8]]
+  RwStBuf[IX[0]] = Add(stbElt1, stbElt2, stbElt3, stbElt4);
+
+  // {Append/Consume}StructuredBuffer Tests
+  // CHECK: [[ANHDLCON:%.*]] = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle [[HDLCON]]
+  // CHECK: [[CONIX:%.*]] = call i32 @dx.op.bufferUpdateCounter(i32 70, %dx.types.Handle [[ANHDLCON]], i8 -1) 
+  // OFF: call %dx.types.ResRet.f32 @dx.op.rawBufferLoad.f32(i32 139, %dx.types.Handle [[ANHDLCON]], i32 [[CONIX]], i32 
+  // OFF: call %dx.types.ResRet.f64 @dx.op.rawBufferLoad.f64(i32 139, %dx.types.Handle [[ANHDLCON]], i32 [[CONIX]], i32 16
```
pow2clk, line 209: Since there is no addition operator to null out the padding, its loading is significant. I didn't think it was worth assigning zeroes explicitly, so I check for it here.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+#elif defined(OFF)
+#define TYNAME(T,N) OffVector< T, N >
+#define DECL(T,N,V) OffVector< T, N > V
+#else
+#define TYNAME(T,N) Vector< T, N >
+#define DECL(T,N,V) Vector< T, N > V
+#endif
+
+template<typename T, int N>
+struct Vector {
```
pow2clk, line 58: Just using a struct or matrix forces us down a different code path. This is a completely useless wrapper around a vector, but it allows comparing that behavior to that of native vectors.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+  vector<T, N> v;
+  Vector operator+(Vector vec) {
+    Vector ret;
+    ret.v = v + vec.v;
+    return ret;
+  }
+};
+
+template<typename T, int N>
+struct OffVector {
```
pow2clk, line 68: A struct with an offset we have to get past to access the vector we're interested in. Since this offset is handled differently for ByteAddress and Structured buffers, testing it here is important.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+};
+
+  ByteAddressBuffer RoByBuf : register(t1);
+RWByteAddressBuffer RwByBuf : register(u1);
+
+  StructuredBuffer< TYNAME(TYPE,NUM) > RoStBuf : register(t2);
+RWStructuredBuffer< TYNAME(TYPE,NUM) > RwStBuf : register(u2);
+
+ConsumeStructuredBuffer< TYNAME(TYPE,NUM) > CnStBuf : register(u4);
+AppendStructuredBuffer< TYNAME(TYPE,NUM) > ApStBuf  : register(u5);
```
pow2clk, line 88: Each of these buffer types take slightly or significantly different cod paths when generating loads and stores.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+
+TYPE Add(TYPE f1[NUM], TYPE f2[NUM])[NUM] {
```
pow2clk, line 90: Arrays take a unique code path.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+  // CHECK-DAG: [[HDLRWBY:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 1, i32 1, i32 0, i8 1 }, i32 1, i1 false)
+
+  // CHECK-DAG: [[HDLROST:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 2, i32 2, i32 0, i8 0 }, i32 2, i1 false)
+  // CHECK-DAG: [[HDLRWST:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 2, i32 2, i32 0, i8 1 }, i32 2, i1 false)
+
+  // CHECK-DAG: [[HDLCON:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 4, i32 4, i32 0, i8 1 }, i32 4, i1 false)
+  // CHECK-DAG: [[HDLAPP:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 5, i32 5, i32 0, i8 1 }, i32 5, i1 false)
+
+  // CHECK-DAG: [[IX0:%.*]] = call i32 @dx.op.loadInput.i32(i32 4, i32 {{[0-9]*}}, i32 [[BOFF:0]]
+  // CHECK-DAG: [[RIX0:%.*]] = call i32 @dx.op.loadInput.i32(i32 4, i32 {{[0-9]*}}, i32 [[BOFF]]
```
pow2clk, line 123: Double capture for the non-offset cases where IX0 == RIX0

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-load-stores.hlsl
```diff
@@ -0,0 +1,213 @@
+// RUN: %dxc -DTYPE=float4    -DIX=SIx -T vs_6_6 %s | FileCheck %s
+// RUiN: %dxc -DTYPE=bool4     -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,I1
+// RUN: %dxc -DTYPE=uint64_t2 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,I64
+// RUN: %dxc -DTYPE=double2   -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,F64
+
+// RUN: %dxc -DTYPE=float4    -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX
+// RUiN: %dxc -DTYPE=bool4     -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX,I1
```
pow2clk, line 7: Boolean cases have some additional conversions that will be fixed with a separate change.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-load-stores.hlsl
```diff
@@ -0,0 +1,213 @@
+// RUN: %dxc -DTYPE=float4    -DIX=SIx -T vs_6_6 %s | FileCheck %s
```
pow2clk, line 1: This test sticks to non-aggregate types as LLVM defines them, which doesn't include vectors. This allows using typed buffers, which have a different code path from raw buffers.

pow2clk (commented):
Review guidance comments.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: I'm guessing truncation of matrices aren't allowed? I'd expect that to take place in the else if condition. Should there be some diagnostic if we encounter a matrix in this case?

utils/hct/hctdb.py
```diff
@@ -8381,6 +8381,7 @@ def process_arg(desc, idx, done_args, intrinsic_name):
             component_list = "LICOMPTYPE_ANY"
             rows = "1"
             cols = "1"
+            only = ""
```
bob80905, line 8384: I understand why you would choose only for a variable name, but I'm wondering if there could be a better way to name this.

PR #7115, opened 2025-02-03
Handle coords in load helper

pow2clk:
Load operations previously split out the addr into separate components in individual locations where they were required. This was error prone as it had a lot of places that were missing. This changes the contents of resload and atomic helpers to contain the broken down coordinates instead of the individual Value. This catches a number of new locations, but it requires the intrinsic overloads to be defined as vec1s to handle that case and it still involves a lot of fiddly places for the code that don't produce truncation warnings when longer vectors are used. Not recommended, but functional.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7079


lib/HLSL/HLOperationLower.cpp
```diff
@@ -4597,11 +4617,13 @@ namespace {
...
 struct AtomicHelper {
   AtomicHelper(CallInst *CI, OP::OpCode op, Value *h, Type *opType = nullptr);
-  AtomicHelper(CallInst *CI, OP::OpCode op, Value *h, Value *bufIdx,
-               Value *baseOffset, Type *opType = nullptr);
+  AtomicHelper(CallInst *CI, OP::OpCode op, Value *h, unsigned coordDimensions,
+               Value *coords[], Value *baseOffset, Type *opType = nullptr);
   OP::OpCode opcode;
   Value *handle;
-  Value *addr;
+  static const unsigned kMaxCoordDimensions = 3;
```
pow2clk, line 4624: None of the users of this constructor will have more than one coordinate, but atomic ops can have up to 3.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -8301,58 +8321,58 @@ void TranslateStructBufSubscriptUser(Instruction *user, Value *handle,
...
-                            handle, bufIdx, baseOffset);
+                            handle, 1, &bufIdx, baseOffset);
         TranslateAtomicCmpXChg(helper, Builder, OP);
       } break;
       case IntrinsicOp::IOP_InterlockedCompareStoreFloatBitwise:
       case IntrinsicOp::IOP_InterlockedCompareExchangeFloatBitwise: {
         Type *i32Ty = Type::getInt32Ty(userCall->getContext());
         AtomicHelper helper(userCall, DXIL::OpCode::AtomicCompareExchange,
-                            handle, bufIdx, baseOffset, i32Ty);
+                            handle, 1, &bufIdx, baseOffset, i32Ty);
```
pow2clk, line 8375: Since these are all for structured buffers, they only have one coord and an offset that's handled separately. Should this approach be revived, I'll probably add more atomic testing.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
+// RUN: %dxc -DARR -DTYPE=float    -DNUM=4 -DIX=SIx -T vs_6_6 %s | FileCheck %s
```
pow2clk, line 1: These test are identical to those in the preferred solution #7114. See that for more comments on them.

PR #7116, opened 2025-02-03
Set the resource kind as u32 for boolean resources

pow2clk:
It was producing a validation error previously because they are represented as u32 and i1 is not valid for output resources. This preserves the cmp conversions for values read from the resource, but satisfies the validation requirement. Performs some incidental consolidation of types that are represented as u32.

Includes a couple incidental NFC fixes as well:

Moves lit filecheck tests into the correct directory. I got CodeGenHLSL and CodeGenDXIL mixed up.

Remove dead array handing code. The array type checks were for a type extracted from a vector. Arrays cannot be elements of vectors. This is leftover from when the type came from more sources. The code shows as never executed in coverage.

Related to #7079


tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+  // CHECK-DAG: [[HDLRWBY:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 1, i32 1, i32 0, i8 1 }, i32 1, i1 false)
+
+  // CHECK-DAG: [[HDLROST:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 2, i32 2, i32 0, i8 0 }, i32 2, i1 false)
+  // CHECK-DAG: [[HDLRWST:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 2, i32 2, i32 0, i8 1 }, i32 2, i1 false)
+
+  // CHECK-DAG: [[HDLCON:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 4, i32 4, i32 0, i8 1 }, i32 4, i1 false)
+  // CHECK-DAG: [[HDLAPP:%.*]] = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 5, i32 5, i32 0, i8 1 }, i32 5, i1 false)
+
+  // CHECK-DAG: [[IX0:%.*]] = call i32 @dx.op.loadInput.i32(i32 4, i32 {{[0-9]*}}, i32 [[BOFF:0]]
+  // CHECK-DAG: [[RIX0:%.*]] = call i32 @dx.op.loadInput.i32(i32 4, i32 {{[0-9]*}}, i32 [[BOFF]]
```
tex3d, line 123: Another pattern of CHECK-DAG where BOFF is defined by the first one and used by the second, and they would both match exactly the same IR (unless I'm missing something), are these mixed in with the create handle ops above them?  If they could match the same thing, how do you know if IX0 or RIX0 are what you want elsewhere?

This test seems pretty complicated and might be difficult to update or maintain if that was necessary.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+
+// RUN: %dxc -DOFF -DTYPE=float    -DNUM=4 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,OFF
+// RUN: %dxc -DOFF -DTYPE=bool     -DNUM=4 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,OFF
+// RUN: %dxc -DOFF -DTYPE=uint64_t -DNUM=2 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,OFF
+// RUN: %dxc -DOFF -DTYPE=double   -DNUM=2 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,OFF
+
+// RUiN: %dxc -DARR -DTYPE=float    -DNUM=4 -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX
+// RUiN: %dxc -DARR -DTYPE=bool     -DNUM=4 -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX
+// RUiN: %dxc -DARR -DTYPE=uint64_t -DNUM=2 -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX
+// RUiN: %dxc -DARR -DTYPE=double   -DNUM=2 -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX
```
tex3d, line 27: I don't see any corresponding `VIX:` check lines.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-load-stores.hlsl
```diff
@@ -0,0 +1,213 @@
+// RUN: %dxc -DTYPE=float4    -DIX=SIx -T vs_6_6 %s | FileCheck %s
+// RUN: %dxc -DTYPE=bool4     -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,I1
+// RUN: %dxc -DTYPE=uint64_t2 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,I64
+// RUN: %dxc -DTYPE=double2   -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,F64
+
+// RUiN: %dxc -DTYPE=float4    -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX
+// RUiN: %dxc -DTYPE=bool4     -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX,I1
+// RUiN: %dxc -DTYPE=uint64_t2 -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX,I64
+// RUiN: %dxc -DTYPE=double2   -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX,F64
```
llvm-beanz, line 9: Are you trying to `RUiN` something again? 😨

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
```diff
@@ -0,0 +1,231 @@
...
+// RUN: %dxc -DTYPE=bool     -DNUM=4 -DIX=SIx -T vs_6_6 %s | FileCheck %s
+// RUN: %dxc -DTYPE=uint64_t -DNUM=2 -DIX=SIx -T vs_6_6 %s | FileCheck %s
+// RUN: %dxc -DTYPE=double   -DNUM=2 -DIX=SIx -T vs_6_6 %s | FileCheck %s
+
+// RUN: %dxc -DOFF -DTYPE=float    -DNUM=4 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,OFF
+// RUN: %dxc -DOFF -DTYPE=bool     -DNUM=4 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,OFF
+// RUN: %dxc -DOFF -DTYPE=uint64_t -DNUM=2 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,OFF
+// RUN: %dxc -DOFF -DTYPE=double   -DNUM=2 -DIX=SIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,OFF
+
+// RUiN: %dxc -DARR -DTYPE=float    -DNUM=4 -DIX=VIx -T vs_6_6 %s | FileCheck %s --check-prefixes=CHECK,VIX
```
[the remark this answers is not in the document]
pow2clk: They depend on #7114. The intent was whichever went in second would get their RUiNs changed to RUN. I find the lowercase `i` makes it stand out more to be sure it was caught. Since that change went over like a lead balloon, I'll remove the outright.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
pow2clk: Yeah, I originally thought I might need them, but forgot to remove them when I didn't. These lines are getting removed per the above.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
pow2clk: That's correct. The intent is to initially populate RIX0 with the same thing as IX0 for the cases where there is no offset, so it won't get set to anything else. None of this would be necessary with modern FileCheck that allows defining variables through an invocation flag. Barring that, this is what I can do.

I'm not sure how to respond to the concern about the test being complicated. I could add some comments to try to make the function clearer. I suppose much of the complication comes from the weird way arrays are used. I could pare that part off, but all the checks would be the same and any updating would need to be in both places, which is also a maintenance burden that often gets overlooked.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-load-stores.hlsl
```diff
@@ -0,0 +1,213 @@
...
+// Test codegen for various load and store operations and conversions
+//  for different scalar/vector buffer types and indices.
+///////////////////////////////////////////////////////////////////////
+
+
+
+// CHECK-DAG: %dx.types.ResRet.[[TY:[a-z][0-9][0-9]]] = type { [[TYPE:[a-z0-9]*]],
+// CHECK-DAG: %dx.types.ResRet.[[TY32:[a-z][0-9][0-9]]] = type { [[TYPE]],
+// I64: %dx.types.ResRet.[[TY32:i32]]
+// F64: %dx.types.ResRet.[[TY32:i32]]
```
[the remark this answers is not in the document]
pow2clk: We cannot use `-D`. This is precisely to compensate for that lack. 

I mentioned it briefly above, but I'll go into more detail here. The intent is that these match the same line to initially assign the same capture to two different variables. TY32 can be `i32` or `f32` depending in <64 cases, it will just match TY. In the cases where TY is `i64` or `f64`, this needs to be `i32` because that's how they are stored. In such cases, there is another type defined that accounts for this. They are order dependent, but the DAG won't proceed past the second resret definition in the cases it exists because it matches perfectly.

pow2clk (commented):
A harbinger to a test update

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl
pow2clk: I simplified the defines and macros used in this test. Hopefully that will make this more maintainable?

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-load-stores.hlsl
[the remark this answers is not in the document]
tex3d: > We cannot use -D.

Oh yeah, we added that to our internal FileCheckerTest wrapper which is used under `HLSLFileCheck`, but that doesn't work for the lit tests.

I still think this would be confusing if I had to come back to it later.  I suppose it would be less confusing with a comment and some slight adjustments that indicate the constraints on expectations, like:

```suggestion
// TY matches the i32, i64, f32, or f64 resource return type expected for Raw
// and Structured buffer loads, and should be declared first.
// TY32 is used for the typed buffer/texture resource return type.
// We use CHECK-DAG to match the same type declaration for TY32 by default,
// but this is overridden for I64 and F64 where each 64-bit value uses two i32
// components for typed buffers and textures.
// TYPE will be the same as TY, and is used just to get TY32 to match the same
// type declaration by default.
// CHECK-DAG: %dx.types.ResRet.[[TY:[if](32|64)]] = type { [[TYPE:[if](32|64)]],
// CHECK-DAG: %dx.types.ResRet.[[TY32:[if](32|64)]] = type { [[TYPE]],
// I64: %dx.types.ResRet.[[TY32:i32]]
// F64: %dx.types.ResRet.[[TY32:i32]]
```

Please correct me if I got anything wrong here, as it reflects my current understanding of these checks.

tex3d (approved):
Looks good.

Issue #7113, opened 2025-02-03
[OMM] Add AllocateRayQuery2 DXIL Op


bob80905:
**Is your feature request related to a problem? Please describe.**
According to the spec here:
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0024-opacity-micromaps.md
There is a new DXIL Op that needs to be added as part of the OMM feature. `AllocateRayQuery2` is a variant of the allocate ray query DXIL Op, but it accepts a new argument, as part of the new optional template argument that may be provided to RayQuery objects upon construction in HLSL.
We need a new DXIL Op, but it doesn't exist yet. 

**Describe the solution you'd like**
Create AllocateRayQuery2, and run tests on DXIL files containing it. Test validation errors too.

> Note this issue doesn't include HLK testing

PR #7124, opened 2025-02-05
[OMM] Add AllocateRayQuery2 DXIL Op, add validation

bob80905:
This PR adds AllocateRayQuery2 as a new DXIL Op, which is intended to be the DXIL Op that any rayquery object initialized with the 2nd optional template argument will be lowered to. Additionally, according to the spec here: https://github.com/microsoft/hlsl-specs/blob/main/proposals/0024-opacity-micromaps.md
there are new validation rules that should be run against the allocateRayQuery* DXIL Ops. This PR implements these validation rules, and tests them.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7113


lib/DxilValidation/DxilValidation.cpp
```diff
@@ -2756,6 +2756,44 @@ static void ValidateFunctionBody(Function *F, ValidationContext &ValCtx) {
...
+          if (dxilOpcode == DXIL::OpCode::AllocateRayQuery) {
+            // validate flags are immediate and compatible
+            llvm::Value *constRayFlag = CI->getOperand(1);
+            if (!llvm::isa<llvm::Constant>(constRayFlag)) {
+              ValCtx.EmitInstrError(
+                  &I, ValidationRule::
+                          DeclNonConstFlagsUnsupportedForAllocateRayQuery);
+            }
+          }
+          if (dxilOpcode == DXIL::OpCode::AllocateRayQuery2) {
```
alsepkow, line 2769: >if (dxilOp [](http://example.com/codeflow?start=10&length=10)

nit: formatting looks like there should be a blank line before every if

bob80905: It does, but looks like our code formatting bot doesn't like it for some reason, reverting.

simoll:
I'll pull the new dxil opcode `AllocateRayQuery2 `into the opcode block reserved by:https://github.com/microsoft/DirectXShaderCompiler/pull/7125

The opcode 258 is currently claimed by both PRs. This will allow us to merge more easily.

utils/hct/hctdb.py
```diff
@@ -5508,9 +5511,37 @@ def UFI(name, **mappings):
...
+                ),
+            ],
+        )
+        next_op_idx += 1
+
         # End of DXIL 1.8 opcodes.
         self.set_op_count_for_version(1, 8, next_op_idx)
-        assert next_op_idx == 258, (
+        assert next_op_idx == 259, (
             "258 is expected next operation index but encountered %d and thus opcodes are broken"
```
damyanp, line 5545: Maybe?

```suggestion
            "259 is expected next operation index but encountered %d and thus opcodes are broken"
```

utils/hct/hctdb.py
```diff
@@ -5508,9 +5511,37 @@ def UFI(name, **mappings):
...
+            "AllocateRayQuery2",
+            "allocates space for RayQuery and return handle",
+            "v",
+            "",
+            [
+                db_dxil_param(0, "i32", "", "handle to RayQuery state"),
+                db_dxil_param(
+                    2,
+                    "u32",
+                    "constRayFlags",
```
damyanp, line 5527: Is this an existing convention to have a "const" prefix for parameter names like "constRayFlags" here?

damyanp (approved):
LGTM - some questions, maybe something you want to change as well.

You really should get a review from someone familiar with the DXC codebase.

utils/hct/hctdb.py
bob80905: It is not an existing convention, this seems to be the only parameter with that prefix.

utils/hct/hctdb.py
bob80905: 258 has been reserved by Simon for AllocateRayQuery2, so that opcode should be all set, no change needed.

damyanp: I don't understand why the `258` in the message doesn't also need to be updated to `259`, like in the assertion expression.

bob80905: Ah I get it, you're right.

utils/hct/hctdb.py
tex3d: The reason for constRayFlags was that additional non-const "RayFlags" are provided on the "TraceRayInline" which are combined (OR'd) with these flags, so this naming differentiates them.

utils/hct/hctdb.py
```diff
@@ -5508,10 +5511,38 @@ def UFI(name, **mappings):
...
+                    "u32",
+                    "constRayQueryFlags",
+                    "Valid combination of RAYQUERY_FLAGS",
+                    is_const=True,
+                ),
+            ],
+        )
+        next_op_idx += 1
+
         # End of DXIL 1.8 opcodes.
```
tex3d, line 5542: This intrinsic should have been added after this marker.  This is a DXIL 1.9 opcode.

utils/hct/hctdb.py
```diff
@@ -629,6 +629,9 @@ def populate_categories_and_models(self):
         ).split(","):
             self.name_idx[i].category = "Inline Ray Query"
             self.name_idx[i].shader_model = 6, 5
+        for i in "AllocateRayQuery2".split(","):
+            self.name_idx[i].category = "Inline Ray Query"
+            self.name_idx[i].shader_model = 6, 8
```
tex3d, line 634: This should be `6, 9`, not `6, 8`

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -3938,7 +3977,7 @@ static void ValidateShaderFlags(ValidationContext &ValCtx) {
   unsigned valMajor, valMinor;
   ValCtx.DxilMod.GetValidatorVersion(valMajor, valMinor);
   if (DXIL::CompareVersions(valMajor, valMinor, 1, 5) >= 0 &&
-      DXIL::CompareVersions(valMajor, valMinor, 1, 8) < 0 &&
+      DXIL::CompareVersions(valMajor, valMinor, 1, 8) != 0 &&
```
tex3d, line 3980: I don't think this change is correct.  This is a WAR for a bug fixed in DXIL 1.8.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -2756,6 +2756,45 @@ static void ValidateFunctionBody(Function *F, ValidationContext &ValCtx) {
...
+            // as a RayQueryFlag argument
+            llvm::ConstantInt *Arg1 =
+                llvm::cast<llvm::ConstantInt>(constRayFlag);
+            llvm::ConstantInt *Arg2 =
+                llvm::cast<llvm::ConstantInt>(RayQueryFlag);
+            if (Arg1->getValue().getSExtValue() ==
+                    (unsigned)DXIL::RayFlag::ForceOMM2State &&
+                Arg2->getValue().getSExtValue() !=
+                    (unsigned)DXIL::RAYQUERY_FLAG::
+                        RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS) {
```
tex3d, line 2792: These should not be testing for equality but whether the corresponding bits are set.
```suggestion
            if ((Arg1->getValue().getSExtValue() &
                    (unsigned)DXIL::RayFlag::ForceOMM2State) &&
                (Arg2->getValue().getSExtValue() &
                    (unsigned)DXIL::RAYQUERY_FLAG::
                        RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS) == 0) {
```

include/dxc/DxilContainer/RDAT_SubobjectTypes.inl
```diff
@@ -66,12 +66,13 @@ RDAT_ENUM_END()
 RDAT_DXIL_ENUM_START(hlsl::DXIL::RaytracingPipelineFlags, uint32_t)
   RDAT_ENUM_VALUE_NODEF(None)
   RDAT_ENUM_VALUE_NODEF(SkipTriangles)
-  RDAT_ENUM_VALUE_NODEF(SkipProceduralPrimitives)
+  RDAT_ENUM_VALUE_NODEF(SkipProceduralPrimitives)
```
tex3d, line 69: nit: added trailing whitespace?

include/dxc/DXIL/DxilConstants.h
```diff
@@ -1774,8 +1776,17 @@ enum class RayFlag : uint32_t {
   CullNonOpaque = 0x80,
   SkipTriangles = 0x100,
   SkipProceduralPrimitives = 0x200,
+  ForceOMM2State = 0x400, // Force 2-state in Opacity Micromaps
 };
 
+enum class RAYQUERY_FLAG : uint32_t {
+  RAYQUERY_FLAG_NONE = 0x00, // default
+  RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS = 0x01,
+};
```
tex3d, line 1785: This shouldn't be defined here.  These names are defined for HLSL and correspond to values defined by `RayQueryFlag` here.

utils/hct/hctdb.py
simoll: fyi, https://github.com/microsoft/DirectXShaderCompiler/pull/7125 has been updated to reserve the AllocateRayQuery2 opcode in DXIL 1.9

include/dxc/DXIL/DxilConstants.h
bob80905: Removed entirely, I think it makes most sense to introduce this with the frontend change.

bob80905: On further investigation, it looks like this enum belongs in an external header, defined by the D3D team. D3D12_RAY_FLAGS is defined in `external\DirectX-Headers\include\directx\d3d12.h`, so I would think this enum would belong in that same file.

utils/hct/hctdb.py
```diff
@@ -5509,9 +5512,38 @@ def UFI(name, **mappings):
...
+                ),
+            ],
+        )
+        next_op_idx += 1
+
         self.set_op_count_for_version(1, 8, next_op_idx)
-        assert next_op_idx == 258, (
-            "258 is expected next operation index but encountered %d and thus opcodes are broken"
+        assert next_op_idx == 259, (
+            "259 is expected next operation index but encountered %d and thus opcodes are broken"
```
tex3d, line 5546: I think you misunderstood my previous comment. This change should not be made.  This block of code should remain with the comment `End of DXIL 1.8 opcodes.`, and the new opcode should appear below this assert.

Otherwise, the new opcode is considered part of DXIL 1.8.  The point of this block of code between each DXIL version is to prevent changing opcodes for an already-shipped DXIL version.

tex3d (changes requested):
See comments for details:
- new op should not be in DXIL 1.8
- wrong location for DXIL op call checking code
- Some minor suggestions on validation rule naming

utils/hct/hctdb.py
```diff
@@ -5508,13 +5511,43 @@ def UFI(name, **mappings):
         )
         next_op_idx += 1
 
-        # End of DXIL 1.8 opcodes.
-        self.set_op_count_for_version(1, 8, next_op_idx)
+        self.set_op_count_for_version(1, 9, next_op_idx)
```
tex3d, line 5514: There should be no change to these lines at all.  This is the end-of-DXIL 1.8 marker which follows the same pattern as prior DXIL versions.  After DXIL 1.8, you add new opcodes below this without changing the marker.

utils/hct/hctdb.py
```diff
@@ -5527,6 +5555,8 @@ def UFI(name, **mappings):
         # Reserved block C
         next_op_idx = self.reserve_dxil_op_range("ReservedC", next_op_idx, 10)
 
+        self.set_op_count_for_version(1, 9, next_op_idx)
```
tex3d, line 5558: We should not set this yet.  DXIL 1.9 is not final.

tex3d (approved):
Looks good!

PR #7126, opened 2025-02-06
Add error diagnostic for semantic with parameters (#6487)

### LINKED ISSUE #6487: DXC crashes on bad shader register input
**Description**

```
StructuredBuffer<uint> foo : FOO(123);
void main() {} // unneeded
```

This crashes on latest dxc.

tcorringham:
A non-null parenthesized list following a semantic was mis-parsed as a direct initializer. When the semantic actions for a direct initializer are performed the type of the initializer is dereferenced - however as a semantic doesn't have a type this resulted in a NULLPTR being dereferenced which caused a crash.

As semantics don't accept parameters any parenthesized list following a semantic is an error. We now generate an error diagnostic if a l-paren follows a semantic.

Fixes #6487


tcorringham:
This issue has been fixed by the alternative PR https://github.com/microsoft/DirectXShaderCompiler/pull/7729, so I am closing this PR.

Issue #7095, opened 2025-01-28
[SPIR-V] -fspv-debug=vulkan-with-source does not support functions with out value and vector component parameter (NonSemantic.Shader.DebugInfo.100 DebugDeclare: expected operand Variable must be a result id of OpVariable or OpFunctionParameter)


webez:
When generating code with -spirv and -fspv-debug=vulkan-with-source, if there exists a function with an out parameter and it is called passing a vector component, an error message is shown

NonSemantic.Shader.DebugInfo.100 DebugDeclare: expected operand Variable must be a result id of OpVariable or OpFunctionParameter
  %91 = OpExtInst %void %1 DebugDeclare %37 %89 %21

**Steps to Reproduce**
This code does not generate spirv (  https://godbolt.org/z/ecsean1nc )
`struct PSInput
{
    float4 position : SV_Position;
    float4 color    : COLOR0;
};

void ModifyFunction(out float output_val)
{
    output_val = 10;
}

float4 PSMain(PSInput input) : SV_Target0
{
    ModifyFunction(input.color.w);
    return input.color;
}`

**Actual Behavior**
`fatal error: generated SPIR-V is invalid: NonSemantic.Shader.DebugInfo.100 DebugDeclare: expected operand Variable must be a result id of OpVariable or OpFunctionParameter
  %91 = OpExtInst %void %1 DebugDeclare %37 %89 %21

note: please file a bug report on https://github.com/Microsoft/DirectXShaderCompiler/issues with source code if possible

Compiler returned: 5`

**Environment**
- DXC trunk

Issue #5191, opened 2023-05-08
[ SPIR-V ] NonSemantic.Shader.DebugInfo.100 DebugDeclare: expected operand Variable must be a result id of OpVariable or OpFunctionParameter


danginsburg:
We encountered the following error with one of our shaders when using `-fspv-debug=vulkan-with-source` to debug the shader in RenderDoc:

```
fatal error: generated SPIR-V is invalid: NonSemantic.Shader.DebugInfo.100 DebugDeclare: expected operand Variable must be a result id of OpVariable or OpFunctionParameter
  %94 = OpExtInst %void %1 DebugDeclare %36 %92 %20
```

@jeremy-lunarg helped strip it down to a test case we could share (attached).  You can reproduce the error with this shader and the following command-line:

```
dxc.exe -spirv -T vs_6_0 -E MainVs -fspv-debug=vulkan-with-source shaderdebug_simple.hlsl
fatal error: generated SPIR-V is invalid: NonSemantic.Shader.DebugInfo.100 DebugDeclare: expected operand Variable must be a result id of OpVariable or OpFunctionParameter
  %94 = OpExtInst %void %1 DebugDeclare %36 %92 %20

note: please file a bug report on https://github.com/Microsoft/DirectXShaderCompiler/issues with source code if possible
```

[shaderdebug_simple.zip](https://github.com/microsoft/DirectXShaderCompiler/files/11422065/shaderdebug_simple.zip)

jeremy-lunarg:
I've done some investigating and it appears to be coming from the exhaustive inlining pass.

jeremy-lunarg:
I'm working on fix to the optimizer to convert access chain function call arguments to temporary OpVariables so the inlining pass can handle DebubDeclare correctly.

jeremy-lunarg:
After discussing with @s-perron and @greg-lunarg, I believe [the work](https://github.com/microsoft/DirectXShaderCompiler/compare/main...llvm-beanz:DirectXShaderCompiler:cbieneman/rewrite-out-params) that @llvm-beanz is doing will completely avoid this problem and it is the ideal way to fix this issue.

s-perron:
It will avoid it until we add reference to the language. We can wait until references are implemented to fix it. Otherwise we won't be able to test the fix.

ChristianReinbold:
[...]

Issue #7015, opened 2024-11-20
[SPIR-V] 'vk::SpirvOpaqueType<NNNN>' is an object and cannot be used as a type parameter` a.k.a. SpirvType as `T` of SpirvType


devshgraphicsprogramming:
**Description**

Follow up to https://github.com/microsoft/DirectXShaderCompiler/issues/5092 as requested by @s-perron 

**Steps to Reproduce**

Declare one SpirvOpaqueType (or even SpirvType) and shove it in another opaque type, examples:
- image type, inside combined sampler image
```cpp
template<typename SampledType, spv::Dim dim, uint32_t Depth, uint32_t Arrayed, uint32_t MS, uint32_t Sampled, spv::ImageFormat Format=spv::ImageFormatUnknown, spv::AccessQualifier AccessQualifier=spv::AccessQualifierMax>
using image = vk::SpirvOpaqueType<
    /*spv::OpTypeImage*/ 25,
    SampledType,
    vk::Literal<vk::integral_constant<uint,dim> >,
    vk::Literal<vk::integral_constant<uint,Depth> >,
    vk::Literal<vk::integral_constant<uint,Arrayed> >,
    vk::Literal<vk::integral_constant<uint,MS> >,
    vk::Literal<vk::integral_constant<uint,Sampled> >,
    vk::Literal<vk::integral_constant<uint,Format> >
// TODO: use `nbl::hlsl::conditional_t<AccessQualifier==Max` to choose between two SpirvOpaqueTypes
>;

template<typename Image>
using sampledImage = vk::SpirvOpaqueType</*spv::OpTypeSampledImage*/ 27,Image>;

// %type_2d_image = OpTypeImage %float 2D 2 0 0 1 Unknown
// %type_sampled_image = OpTypeSampledImage %type_2d_image
// %_ptr_UniformConstant_type_2d_image_0 = OpTypePointer UniformConstant %type_2d_image_0
using sampler2D = spirv::pointer<spv::StorageClassUniformConstant,spirv::sampledImage<spirv::image<float,spv::Dim2D,2,0,0,1> > >;
```
- Acceleration Structure Type inside pointer type
```cpp
using pointer_t = vk::SpirvOpaqueType<spv::OpTypePointer,vk::Literal<vk::integral_constant<uint32_t,StorageClass> >,T>;

using RayQueryKHR = vk::SpirvOpaqueType<spv::OpTypeRayQueryKHR>;

[...]

PR #7127, opened 2025-02-06
[SPIRV] Use copy-in/copy-out for non-declaration

s-perron:
DXC tries to avoid the copy-in-copy-out for some parameters when it
thinks it is possible. However, the SPIR-V spec says that if a parameter
is a pointer, it must point to a memory object declarion or a pointer to
an element in a sampler or image array.

This PR enforces this. If the pointer for the parameter is not an
OpVariable (a memory object declaration), then we do not elide the
copy-in copy-out.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7095
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/5191

### DIFF
--- a/tools/clang/lib/SPIRV/SpirvEmitter.cpp
+++ b/tools/clang/lib/SPIRV/SpirvEmitter.cpp
@@ -3104,6 +3104,12 @@ SpirvInstruction *SpirvEmitter::processCall(const CallExpr *callExpr) {
         argInfo && argInfo->getStorageClass() != spv::StorageClass::Function &&
         isResourceType(paramType);
 
+    // HLSL requires that the parameters be copied in and out from temporaries.
+    // This looks for cases where the copy can be elided. To generate valid
+    // SPIR-V, the argument must be a memory declaration.
+    //
+    //
+
     // If argInfo is nullptr and argInst is a rvalue, we do not have a proper
     // pointer to pass to the function. we need a temporary variable in that
     // case.
@@ -3112,7 +3118,7 @@ SpirvInstruction *SpirvEmitter::processCall(const CallExpr *callExpr) {
     // create a temporary variable for it because the function definition
     // expects are point-to-pointer argument for resources, which will be
     // resolved by legalization.
-    if ((argInfo || (argInst && !argInst->isRValue())) &&
+    if ((argInfo || (argInst && argInst->getopcode() == spv::Op::OpVariable)) &&
         canActAsOutParmVar(param) && !isArgGlobalVarWithResourceType &&
         paramTypeMatchesArgType(paramType, arg->getType())) {
       // Based on SPIR-V spec, function parameter must be always Function
--- a/tools/clang/test/CodeGenSPIRV/cs.groupshared.function-param.out.hlsl
+++ b/tools/clang/test/CodeGenSPIRV/cs.groupshared.function-param.out.hlsl
@@ -28,10 +28,14 @@ groupshared S D;
 [numthreads(1,1,1)]
 void main() {
 // CHECK: %E = OpVariable %_ptr_Function_int Function
+// CHECK-NEXT: [[TempVar:%[a-zA-Z0-9_]+]] = OpVariable %_ptr_Function_int Function
+
   int E;
 
 // CHECK:        [[A:%[0-9]+]] = OpAccessChain %_ptr_Uniform_int %A %int_0 %uint_0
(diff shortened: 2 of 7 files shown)


cassiebeckley (approved):

devshgraphicsprogramming:
How badly does this affect our `inout` is `restrict &` assuming hacks?

llvm-beanz:
> How badly does this affect our `inout` is `restrict &` assuming hacks?

Don't assume this. This is never valid to assume and can cause lots of behavior differences.

Keenuts (approved):

s-perron:
> > > How badly does this affect our `inout` is `restrict &` assuming hacks?
> > 
> > 
> > Don't assume this. This is never valid to assume and can cause lots of behavior differences.
> 
> The problem is that only Inline SPIR-V functions can use `vk::ext_reference`
> 
> https://github.com/Devsh-Graphics-Programming/Nabla/blob/6efd7e9b1a4dd3e006b2958165a65e587336c8c9/include/nbl/builtin/hlsl/spirv_intrinsics/core.hlsl#L63-79
> 
> So if I want to wrap an intrinsic in another function, `inout` is needed
> 
> https://github.com/Devsh-Graphics-Programming/Nabla/blob/6efd7e9b1a4dd3e006b2958165a65e587336c8c9/include/nbl/builtin/hlsl/glsl_compat/core.hlsl#L49-58

I made minimal changes to fix the issue with the debug information. If this causes you problem, you will need to find something else. As Chris mentioned, your code is not standard, so there is no guarantee that it will work. I do not want to keep non-standard code working at the expense of compliant code not working.If this bothers you, sorry about that, but it is necessary.

devshgraphicsprogramming:
Seems that my stuff still works https://godbolt.org/z/Khrr1GfGc

However current Godbolt DXC (trunk) is still one commit behind this one

devshgraphicsprogramming:
> Seems that my stuff still works https://godbolt.org/z/Khrr1GfGc
> 
> However current Godbolt DXC (trunk) is still one commit behind this one

Update: my stuff still works, I'm not a Clang/SPIR-V expert but I thought that the things I passed were not OpVariables (they were OpAccessChains to OpVariables though).

You have a test thats similar with RWStructuredBuffer and `A[0]`, and that gets made into a tempvar. Not sure why my stuff still works, does `out` get treated differently to `inout` ?

s-perron:
> > Seems that my stuff still works https://godbolt.org/z/Khrr1GfGc
> > However current Godbolt DXC (trunk) is still one commit behind this one
> 
> Update: my stuff still works, I'm not a Clang/SPIR-V expert but I thought that the things I passed were not OpVariables (they were OpAccessChains to OpVariables though).
> 
> You have a test thats similar with RWStructuredBuffer and `A[0]`, and that gets made into a tempvar. Not sure why my stuff still works, does `out` get treated differently to `inout` ?

I just checked, and your code does not work anymore. This is an difficult because this is the exact situation that the copy-in and copy-out were added. All parameters should be in the function storage class.

I see two way around this.

1. Make the parameter a `vk::SpirvType` representing a pointer to the storage class. This could hit other issue I'm not aware of.
2. Create a version, where you index the buffer in the function instead. That that is bad for code reuse. Sorry that HLSL is not expressive enough to be able to do what you want. We are trying to work towards it, but we are not there yet.

Since there are still more issues with the debug info, I'll see if I can fix that problem another way.

devshgraphicsprogramming:
> Make the parameter a vk::SpirvType representing a pointer to the storage class. This could hit other issue I'm not aware of.

This would involve me having to call `OpCopyObject` intrinsic to "get" the pointer to the argument, and do the following
```cpp
atomicAdd(get_ref<AddressSpace>(buf[1]),5);
```

Then I have the problem that I need to explicitly provide the address space as it can't be deduced, also I'll run into a number of SPIR-V issues:
1. The SPIR-V layout of a type `T` in the case of other functions, not atomics 
2. Inability to construct certain SPIR-V pointers because they require a SPIR-V type to be constructed from #7015 
3. Spirv-opt or your legalization getting completely thrown off my leaked references/pointers https://github.com/microsoft/DirectXShaderCompiler/issues/6578#issuecomment-2297181671

For now I just implemented a crude version of #5675 in my DXC fork which allows `[[vk::ext_reference]] inout T` on non-inline-SPIR-V-intrinsic functions.

PR #7129, opened 2025-02-07
[NFC] Remove unused functions in ExecutionTest.cpp and assert.cpp

bob80905:
Some functions need to be removed / changed to unblock some internal pipelines. They are emitting warnings that they are unused.
Specifically, `CompareOutputWithExpectedValueFloat`, `CompareOutputWithExpectedValueHalf`, and `CompareOutputWithExpectedValueFloat` need to be removed because they are unused.
Additionally, there is a case in assert.cpp where `llvm_assert_trap` is left unused depending on what's been defined and how the preprocessor directives execute. This needs to be remedied so that the function is only defined when it is used.

### DIFF
--- a/lib/Support/assert.cpp
+++ b/lib/Support/assert.cpp
@@ -11,6 +11,17 @@
 #include "llvm/Support/Compiler.h"
 #include "llvm/Support/raw_ostream.h"
 
+#if defined(LLVM_ASSERTIONS_TRAP) || !defined(WIN32)
+namespace {
+void llvm_assert_trap(const char *_Message, const char *_File, unsigned _Line,
+                      const char *_Function) {
+  llvm::errs() << "Error: assert(" << _Message << ")\nFile:\n"
+               << _File << "(" << _Line << ")\nFunc:\t" << _Function << "\n";
+  LLVM_BUILTIN_TRAP;
+}
+} // namespace
+#endif
+
 #ifdef _WIN32
 #include "dxc/Support/Global.h"
 #include "windows.h"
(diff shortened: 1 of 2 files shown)


hekota (approved):
LGTM!

alsepkow:
LGTM

lib/Support/assert.cpp
```diff
@@ -10,6 +10,8 @@
 #include "assert.h"
 #include "llvm/Support/Compiler.h"
 #include "llvm/Support/raw_ostream.h"
+
+#if defined(LLVM_ASSERTIONS_TRAP) || !defined(WIN32)
```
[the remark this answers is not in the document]
bob80905: It should be `||`, because if it were `&&`, the needed function would not be defined. Consider the case where WIN32 is not defined, the bottom else case.
We need the function defined when WIN32 is not defined. Within this if condition, we don't necessarily need LLVM_ASSERTIONS_TRAP to be true or false, it doesn't matter. But if it's false, the `&&` will prevent the function from being defined. So I think `||` is appropriate.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -6954,29 +6923,6 @@ static bool VerifyOutputWithExpectedValueHalf(uint16_t output, uint16_t ref,
       CompareOutputWithExpectedValueHalf(output, ref, type, tolerance));
 }
 
-template <typename T>
-static bool CompareOutputWithExpectedValue(T output, T ref,
```
tex3d, line 6958: Can we just make these non-static?  This was infrastructure used by WaveMatrix tests that were removed, but I expect that it would be useful again in upcoming tests.

tex3d (commented):
Can we more clearly describe the changes in the change description, rather than just the justification?  Doesn't have to be long, but naming the functions removed from ExecutionTest.cpp, and clarifying that llvm_assert_trap is placed under the preprocessor condition, rather than removed, would be better when reading the description to know what changes are being made.

I also don't really like that these two are being changed in the same PR, and I think the ExecutionTest functions are potentially useful for upcoming testing work, so maybe these can be made non-static?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
hekota: We can always add it back when needed.

hekota (approved):
LGTM!

bob80905:
> Can we more clearly describe the changes in the change description, rather than just the justification? Doesn't have to be long, but naming the functions removed from ExecutionTest.cpp, and clarifying that llvm_assert_trap is placed under the preprocessor condition, rather than removed, would be better when reading the description to know what changes are being made.
> 
> I also don't really like that these two are being changed in the same PR, and I think the ExecutionTest functions are potentially useful for upcoming testing work, so maybe these can be made non-static?

I've improved the description.
Unsure if making them non-static will solve the problem, if somehow it doesn't this will delay release. I know for certain deleting them will. In agreement with Helena, we can add it back if needed later.

tex3d (approved):

PR #7130, opened 2025-02-07
Disallow structs, arrays, and overwide vectors from typed buffers

### LINKED ISSUE #7080: Disallow anything but scalars and vectors in Typed Buffers
**Description**
Typed buffers can be declared with structs, matrices, and arrays. However, the code that is generated is incorrect in these cases and in the case of arrays, won't even pass validation. Rather than expend effort to fix this feature of limited utility, we've opted to remove it and produce errors when such types are used in Typed Buffers.

**Steps to Reproduce**
The most visible example is using `float[4]` for Stores: https://godbolt.org/z/hnxY7WhTn 

**Actual Behavior**
The validation errors produced include errors because the compiler gets very confused and tries to output raw buffer stores. Even when that is resolved, the array lowering results in a load or store operation for each element, which is read as not assigning the complete entry for a Typed Buffer, which is another validation error. 

**Environment**
- DXC version 1.8.2407
- Host Operating System MacOs 14.7.2 And Plan 9 Fourth Edition (why not?)

pow2clk:
These never reliably generated the right results, so they are being disallowed. The check takes place in SemaHLSL and replaces a few places that checked for such things in special cases. To facilitate the check, the definition of the texture bit was expanded to include RW textures. This is consistent with the description and since the bit was never actually used before, has no other consequences.

Some code that executed after Sema that either produced errors or processed resources with these now forbidden types was removed. Most of this is from CodeGen and HLOperationLower where code was generated or expanded for load/store operations on such resources.

Additionally moved the check for overlarge elements in typed resources to enable keeping some tests around that tested these side by side as well as adding a new one. I changed the nature of the check a bit to limit the types to four elements and also that they be no larger than 4 32-bit elements. This means that 64-bit elements are limited to two. While there wasn't an error that caught this before, there was an assert that fired.

Incidentally fixes an issue with error reporting for modified types after the first mismatch of that intrinsic is encountered. Because the insertion code changed them in the temp array and the stored function type, but printed from the temp array and the later encounters didn't add the modifiers to the temp array, later errors would incorrectly leave off the reference. By retrieving the type from the function type, the reference is always preserved.

This required changing a fair number of tests that relied on this behavior including removing test cases and entire test files that existed solely for such tests.

Fixes #7080


pow2clk:
The first four commits are intended to facilitate reviewing of mostly independent parts:
* 61d8196d1db6946a736586f3f03d7c74a7a0a1c2 is incidental, but saved me from having to make a nonsensical change to atomics-on-bitfields.hlsl because the first encounter of the atomic ops would be removed, making the second the first which correctly had the references. Now they all correctly do
* c1d03f89d874370962d68c2febdd64125200c06b is the meat of the change. This is where the new checks are added along with the tests targeted to this behavior.
* c1a1661767054ec67b6b1af04d39d77dd9c1b5ee is a cleanup that the last commit allows. Because the errors are caught in Sema, later processing that accounted for these sorts of resources can be removed
* 4633d6bc51cec0efeeab053290402b0cd48c13dc groups all the messy test cleanups into one change. Most of these are removing tests that focused on the disallowed behavior.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14178,6 +14180,35 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
 
   ArBasicKind basicKind = hlslSource->GetTypeElementKind(qt);
 
+  // Check for invalid template params in typed buffer/texture declarations.
```
tex3d, line 14183: I think this checking needs to be done under `CheckTemplateArgumentListForHLSL`, like by updating `IsValidTemplateArgumentType`, rather than on the `DiagnoseHLSLDecl`.

pow2clk: I moved it to `CheckTemplateArgumentListForHLSL`. I was tempted to move it to `IsValidTemplateArgumentType`, but it would need to add some params to that and then would just be an else there instead of outside where we can already identify textures (though I'm not thrilled with how we need to identify them there either)

tools/clang/test/CodeGenSPIRV/type.rwbuffer.struct.error.hlsl
```diff
@@ -1,16 +1,16 @@
...
   float b;
 };
 
-// CHECK: error: cannot instantiate RWBuffer with struct type 'S'
-RWBuffer<S> MyRWBuffer;
-
-// CHECK: error: cannot instantiate RasterizerOrderedBuffer with struct type 'S'
-RasterizerOrderedBuffer<S> MyROVBuffer;
+// BUF: error: Elements of typed buffers and textures must be scalars or vectors
+// ROV: error: cannot instantiate RasterizerOrderedBuffer with struct type 'S'
```
tex3d, line 10: Shouldn't we produce the same error for ROV as for RWBuffer?

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -938,11 +938,9 @@ CXXRecordDecl *hlsl::DeclareTemplateTypeWithHandleInDeclContext(
...
       /*templateDepth*/ 0, 0, ParameterPackFalse, elementTemplateParamDecl);
 
-  if (templateArgCount > 1 &&
-      // Only need array type for inputpatch and outputpatch.
-      // Avoid Texture2DMS which may use 0 count.
-      // TODO: use hlsl types to do the check.
-      !name.startswith("Texture") && !name.startswith("RWTexture")) {
+  if (templateArgCount > 1 && (!Attr || !isa<HLSLTypedResourceAttr>(Attr))) {
+    // Only need array type for inputpatch and outputpatch.
+    // isTyped check avoids Texture2DMS which may use 0 count.
```
tex3d, line 943: This is super ugly, assuming that we can determine whether this is an InputPatch or OutputPatch based on whether it has two or more template arguments and is not a typed resource.

In PR #7143, you add the `HLSLTessPatch` attribute [here](https://github.com/microsoft/DirectXShaderCompiler/pull/7143/files#diff-ece93d307d53026d6a66bb7372fe5af375c97df5c5913a10bf591d61ae3a3bcaR1001).

It would be much cleaner to depend on that attribute for positive identification instead of what it does now.

Could the required attributes be added in an earlier NFC change that these PRs depend on?

tex3d: In fact, the refactor of these functions to accept and apply the attribute could be included with this earlier NFC change.

tools/clang/include/clang/Basic/Attr.td
```diff
@@ -992,6 +992,12 @@ def HLSLNodeTrackRWInputSharing : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
+def HLSLTypedResource : InheritableAttr {
+  let Spellings = []; // No spellings!
+  let Subjects = SubjectList<[CXXRecord]>;
+  let Documentation = [Undocumented];
```
tex3d, line 998: While this minimally supports the use case at the moment, I think it would be far better to have a more general resource attribute that captures at least `DXIL::ResourceClass`, `DXIL::ResourceKind`, and `DXIL::SamplerKind`, which could be used for all resource objects.  There are already useful inline helper functions in `DxilConstants.h` for checking subsets of resource kinds like `IsTyped`.

tools/clang/test/CodeGenSPIRV/type.rwbuffer.struct.error.hlsl
pow2clk: Per the checks that you cited in https://github.com/microsoft/DirectXShaderCompiler/blob/d5951b74d76351c17626bf90daf6684ed8598d3d/include/dxc/DXIL/DxilConstants.h#L454, ROV buffers are not considered typed. To convert to using ResKind to determined the typedness, I'd either need to add a check or changed the behavior of IsTyped, which could have implications elsewhere that are not at all related to the intent of this change. I don't mind particularly either way, but I'm not familiar enough with its usage to confidently assert that an ROV buffer is a typed buffer for this or any other purpose.

tools/clang/include/clang/Basic/Attr.td
tex3d: > I don't think they should share an attribute with other resources.

That's fine by me.

tools/clang/test/HLSL/cpp-errors-hv2015.hlsl
```diff
@@ -576,8 +571,8 @@ void expressions()
   internal->fn();                 // expected-error {{operator is not supported}}
   local_i = (int3) { 1, 2, 3 };   // expected-error {{compound literal is unsupported in HLSL}}
 
-  // `class` ok, but component count should be checked earlier (1 to 4 uniform components):
-  Texture2D<::c_outer_fn> local_texture;
+  // ok, but component count should be checked earlier (1 to 4 uniform components):
+  Texture2D<::c_outer_td> local_texture;
```
tex3d, line 575: I feel like the comment is out of place now.  Before, it was there because the class used had no members (was empty), and it was a comment that there was no Sema error checking this (DXIL backend checked in CGHLSLMS).  Now that it's just a normal float4, what does the comment mean to a future reader of this line?

Maybe we should remove the comment now?

tools/clang/test/HLSL/cpp-errors.hlsl
```diff
@@ -572,8 +567,8 @@ void expressions()
   internal->fn();                 // expected-error {{operator is not supported}}
   local_i = (int3) { 1, 2, 3 };   // expected-error {{compound literal is unsupported in HLSL}}
 
-  // `class` ok, but component count should be checked earlier (1 to 4 uniform components):
-  Texture2D<::c_outer_fn> local_texture;
+  // ok, but component count should be checked earlier (1 to 4 uniform components):
+  Texture2D<::c_outer_td> local_texture;
```
tex3d, line 571: Comment seems out of place now, see previous comment about the same change.

tools/clang/test/SemaHLSL/atomics-on-bitfields.hlsl
```diff
@@ -13,20 +13,16 @@ struct TexCoords {
...
 void main( uint2 tid : SV_DispatchThreadID) {
 
-  InterlockedOr(buf[tid.y].q, 2);                           /* expected-error {{no matching function for call to 'InterlockedOr'}} expected-note {{candidate function not viable: no known conversion from 'uint' to 'unsigned int &' for 1st argument}} expected-note {{candidate function not viable: no known conversion from 'uint' to 'unsigned long long &' for 1st argument}} */
-  InterlockedCompareStore(buf[tid.y].q, 3, 1);              /* expected-error {{no matching function for call to 'InterlockedCompareStore'}} expected-note {{candidate function not viable: no known conversion from 'uint' to 'unsigned int &' for 1st argument}} expected-note {{candidate function not viable: no known conversion from 'uint' to 'unsigned long long &' for 1st argument}} */
+  InterlockedOr(str[tid.y].q, 2);                           /* expected-error {{no matching function for call to 'InterlockedOr'}} expected-note {{candidate function not viable: no known conversion from 'uint' to 'unsigned int &' for 1st argument}} expected-note {{candidate function not viable: no known conversion from 'uint' to 'unsigned long long &' for 1st argument}} */
```
tex3d, line 23: Note on existing issue: It's a shame we don't have a better message for the bitfield case, since the error message doesn't seem very helpful.
(nothing to do for this here, just a note)

tools/clang/test/SemaHLSL/atomics-on-bitfields.hlsl
```diff
@@ -13,20 +13,16 @@ struct TexCoords {
...
-  InterlockedCompareStore(gs.q, 3, 1);                      /* expected-error {{no matching function for call to 'InterlockedCompareStore'}} expected-note {{candidate function not viable: 1st argument ('__attribute__((address_space(3))) uint') is in address space 3, but parameter must be in address space 0}} expected-note {{candidate function not viable: no known conversion from '__attribute__((address_space(3))) uint' to 'unsigned long long' for 1st argument}} */
+  InterlockedOr(gs.q, 2);                                   /* expected-error {{no matching function for call to 'InterlockedOr'}} expected-note {{candidate function not viable: 1st argument ('__attribute__((address_space(3))) uint') is in address space 3, but parameter must be in address space 0}} expected-note {{candidate function not viable: no known conversion from '__attribute__((address_space(3))) uint' to 'unsigned long long &' for 1st argument}} */
```
tex3d, line 26: Another existing issue note: This error seems even worse, since TGSM address space should be ok, and it's the fact that it's a bitfield that's the problem, right?

tools/clang/test/SemaHLSL/invalid-rovbuffer-elements.hlsl
```diff
@@ -0,0 +1,31 @@
+// RUN: %dxc -Tlib_6_3 -verify %s
+
+// RasterizerOrderedBuffer has suficiently different errors that it requires its own verifiction test.
```
tex3d, line 3: I think this is no longer the case, right?  Can this be rolled back into the other test then?

tools/clang/test/SemaHLSL/template-checks.hlsl
```diff
@@ -13,19 +13,9 @@ matrix<bool, 1, 2> m_bool;
...
-RWBuffer<s_float2_float3> rwb_struct;           /* fxc-error {{X3037: elements of typed buffers and textures must fit in four 32-bit quantities}} */
+RWBuffer<double3> rwb_struct;    /* expected-error {{Elements of typed buffers and textures must fit in four 32-bit quantities}} fxc-error {{X3037: elements of typed buffers and textures must fit in four 32-bit quantities}} */
 
-struct s_float4_sampler
-{
-  float4 f_float4;
-  SamplerState f_sampler;
-};
-RWBuffer<s_float4_sampler> rwb_struct_objs; /* expected-error {{'SamplerState' is an object and cannot be used as a type parameter}} expected-note {{usage of 'SamplerState' found in field 'f_sampler' of type 's_float4_sampler'}} fxc-pass {{}} */
+RWBuffer<SamplerState> rwb_struct_objs; /* expected-error {{'SamplerState' is an object and cannot be used as a type parameter}} fxc-pass {{}} */
```
tex3d, line 18: Unless you tested FXC, we should probably remove the `fxc-pass {{}}` part.
```suggestion
RWBuffer<SamplerState> rwb_struct_objs; /* expected-error {{'SamplerState' is an object and cannot be used as a type parameter}} */
```

tex3d (approved):
Approved with some very minor comments.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7596,6 +7596,10 @@ def err_hlsl_unsupported_buffer_packoffset : Error<
   "packoffset is only allowed within a constant buffer, not on the constant buffer declaration">;
 def err_hlsl_unsupported_buffer_slot_target_specific : Error<
   "user defined constant buffer slots cannot be target specific">;
+def err_hlsl_unsupported_typedbuffer_template_parameter : Error<
+  "Elements of typed buffers and textures must be scalars or vectors">;
```
llvm-beanz, line 7600:
```suggestion
  "elements of typed buffers and textures must be scalars or vectors">;
```

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7596,6 +7596,10 @@ def err_hlsl_unsupported_buffer_packoffset : Error<
...
+def err_hlsl_unsupported_typedbuffer_template_parameter_size : Error<
+  "Elements of typed buffers and textures must fit in four 32-bit quantities">;
```
llvm-beanz, line 7602:
```suggestion
  "elements of typed buffers and textures must fit in four 32-bit quantities">;
```

llvm-beanz (approved):
A few suggestions to apply, otherwise LGTM.

tools/clang/test/SemaHLSL/invalid-rovbuffer-elements.hlsl
pow2clk: No, this is still the case. The issue is the absence of `RWRasterizerOrderedBuffer` and the different behavior regarding no template arguments. Essentially it's a RW buffer without the RW prefix. It only affects the first few lines, but I value those lines as I didn't find any tests for them elsewhere and I couldn't come up with macro skullduggery to cram this round peg into that square hole. I do enjoy macro skullduggery, but I think it has bothered reviewers in the past where I might have more clearly just created a new test with some redundancy as I've done here.

tools/clang/test/SemaHLSL/template-checks.hlsl
pow2clk: It was pre-existing. I'll remove it.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
pow2clk: Applied in new commit along with all the test changes.

tools/clang/test/SemaHLSL/template-checks.hlsl
pow2clk: Applied in new commit.

tools/clang/test/SemaHLSL/atomics-on-bitfields.hlsl
pow2clk: I created #7158 to document these flaws.

tools/clang/test/HLSL/cpp-errors.hlsl
pow2clk: Removed in latest commit

tools/clang/test/CodeGenSPIRV/type.rwbuffer.struct.error.hlsl
pow2clk: These errors are being produced now.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -4618,6 +4627,136 @@ class HLSLExternalSource : public ExternalSemaSource {
...
+      ResKind = DXIL::ResourceKind::StructuredBuffer;
+      ResClass = DXIL::ResourceClass::UAV;
+      return true;
+    case AR_OBJECT_CONSTANT_BUFFER:
+      ResKind = DXIL::ResourceKind::CBuffer;
+      ResClass = DXIL::ResourceClass::CBuffer;
+      return true;
+    case AR_OBJECT_TEXTURE_BUFFER:
+      ResKind = DXIL::ResourceKind::TBuffer;
+      ResClass = DXIL::ResourceClass::CBuffer;
```
tex3d, line 4744: Technically, `ResClass` for `TextureBuffer` should be `DXIL::ResourceClass::SRV`, not `CBuffer`.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -4618,6 +4627,136 @@ class HLSLExternalSource : public ExternalSemaSource {
...
+      return true;
+    case AR_OBJECT_FEEDBACKTEXTURE2D:
+      ResKind = DXIL::ResourceKind::FeedbackTexture2D;
+      ResClass = DXIL::ResourceClass::SRV;
```
tex3d, line 4748: Unfortunately, this is wrong too.  FeedbackTextures are UAVs not SRVs.  Again, `GetHLSLResourceProperties` got this wrong, but that bug has no impact.  It turns out that `GetHLSLResourceProperties` is a bad reference to use to fill in this function!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -4618,6 +4627,136 @@ class HLSLExternalSource : public ExternalSemaSource {
...
+      return true;
+    case AR_OBJECT_FEEDBACKTEXTURE2D_ARRAY:
+      ResKind = DXIL::ResourceKind::FeedbackTexture2DArray;
+      ResClass = DXIL::ResourceClass::SRV;
```
tex3d, line 4752: FeedbackTextures are UAVs not SRVs.

PR #7132, opened 2025-02-07
[NFC] Remove unused functions in ExecutionTest.cpp and assert.cpp (#7129)

bob80905:
Some functions need to be removed / changed to unblock some internal pipelines. They are emitting warnings that they are unused. Specifically, `CompareOutputWithExpectedValueFloat`, `CompareOutputWithExpectedValueHalf`, and
`CompareOutputWithExpectedValueFloat` need to be removed because they are unused.
Additionally, there is a case in assert.cpp where `llvm_assert_trap` is left unused depending on what's been defined and how the preprocessor directives execute. This needs to be remedied so that the function is only defined when it is used.

### DIFF
--- a/lib/Support/assert.cpp
+++ b/lib/Support/assert.cpp
@@ -10,6 +10,8 @@
 #include "assert.h"
 #include "llvm/Support/Compiler.h"
 #include "llvm/Support/raw_ostream.h"
+
+#if defined(LLVM_ASSERTIONS_TRAP) || !defined(WIN32)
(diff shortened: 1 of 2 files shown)


damyanp:
Is this a cherry-pick?

damyanp (approved):

bob80905:
> Is this a cherry-pick?

Yes

tex3d (approved):

Issue #4195, opened 2022-01-18
[SPIR-V] enable `[[vk::ext_decorate/_id/_string]]` for struct/class fields or function parameter


jaebaek:
As a part of HLSL version of GL_EXT_spirv_intrinsics (#3919), we have to enable `[[vk::ext_decorate/_id/_string]]` for struct/class fields or function parameter.

PR #7137, opened 2025-02-11
[SPIRV] Fix decoration attributes on members.

s-perron:
This commit implements adding the `vk::ext_decorate` on a field. It also
adds an error in sema if `vk::ext_decorate_id` or
`vk::ext_decorate_string` are used on members.

For `vk::ext_decorate_id`, there is no `OpMemberDecorateId` instruction,
so we cannot add apply these decorations to members.

For `vk::ext_decorate_string`, there is only one decoration that uses
OpDecorateString or OpMemberDecorateString, UserSemantic. However, that
decoration could have used OpDecorate or OpMemberDecorate because those
instructions accept string literals. I do not expect any new decorations
to use OpDecorateString or OpMemberDecorateString. I may eventually want
to deprecate `vk::ext_decorate_string`.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/4195

### DIFF
--- a/tools/clang/include/clang/Basic/Attr.td
+++ b/tools/clang/include/clang/Basic/Attr.td
@@ -1341,7 +1341,7 @@ def VKDecorateExt : InheritableAttr {
 
 def VKDecorateIdExt : InheritableAttr {
   let Spellings = [CXX11<"vk", "ext_decorate_id">];
-  let Subjects = SubjectList<[Function, Var, ParmVar, Field, TypedefName], ErrorDiag>;
+  let Subjects = SubjectList<[Function, Var, ParmVar, TypedefName], ErrorDiag>;
   let Args = [UnsignedArgument<"decorate">, VariadicExprArgument<"arguments">];
   let LangOpts = [SPIRV];
   let Documentation = [Undocumented];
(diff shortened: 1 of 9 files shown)


Keenuts (approved):

cassiebeckley (approved):

bob80905:
@s-perron @damyanp 
I've encountered an error where `tools\clang\test\SemaHLSL\attributes\spv.inline.decorate.member.hlsl` only fails when the build env doesn't have spirv enabled. I believe we want all tests to pass whether the build has spirv enabled or not.
Is there a better spot to move this file, or maybe change some lit configs so this test doesn't fail without spirv enabled in the build environment?

s-perron:
I'll let @damyanp decide how to handle this. I just don't want to remove test coverage.

tex3d:
> @s-perron @damyanp I've encountered an error where `tools\clang\test\SemaHLSL\attributes\spv.inline.decorate.member.hlsl` only fails when the build env doesn't have spirv enabled. I believe we want all tests to pass whether the build has spirv enabled or not. Is there a better spot to move this file, or maybe change some lit configs so this test doesn't fail / run without spirv enabled in the build environment?

You should be able to add a line to the test:

```hlsl
// REQUIRES: spirv
```

webez:
Shouldn't vk::ext_decorate_string support Fields if there exists OpMemberDecorateString? Our code is currently broken because of this change because vk::ext_decorate_string does not support Fields anymore and vk::ext_decorate does not support strings

s-perron:
> Shouldn't vk::ext_decorate_string support Fields if there exists OpMemberDecorateString? Our code is currently broken because of this change because vk::ext_decorate_string does not support Fields anymore and vk::ext_decorate does not support strings

Your code was probably not doing what you thought it did. Before this change, the decoration attributes on members were ignored. See https://godbolt.org/z/EMsKMMj1E. Note that there is no OpMemberDecorateString. Please open a new issue with details about what you are trying to do, and why you feel this is a regression.

Issue #7117, opened 2025-02-04
[SM6.9] Support long vector declaration, casting, and assignment


pow2clk:
Long vectors should be supported in 
* Elements of arrays, structs, StructuredBuffers, and ByteAddressBuffers.
* Parameters and return types of non-entry functions.
* Stored in groupshared memory.
* Static global variables.
* Local function scoped variables.

This task is to allow declaring vectors of size greater than 4 using the template declaration in all of the above. In addition, they should be assignable, initializable, and castable in ways consistent with existing vectors using splatted scalars, initializer lists, constructors, and casted arrays. The best way to test this is probably a library shader since that allows exporting non-entry functions which allows parameters to be used to verify the assignment operations.

Some example assignments:

```hlsl
vector<uint, 5> InitList = {1, 2, 3, 4, 5};
vector<uint, 6> Construct = vector<uint, 6>(6, 7, 8, 9, 0, 0);
uint4 initval = {0, 0, 0, 0};
vector<uint, 8> VecVec = {uint2(coord.xy), vecB};
vector<uint, 6> Assigned = vecB;
float arr[5];
vector<float, 5> CastArr = (vector<float, 5>)arr;
vector<float, 6> ArrScal = {arr, 7.9};
vector<float, 10> ArrArr = {arr, arr};
vector<float, 15> Scal = 4.2;
```

PR #7143, opened 2025-02-18
[SM6.9] Allow native vectors longer than 4

pow2clk:
Remove errors in Sema diagnostics for vectors longer than 4 in 6.9.
Test for failures using long vectors in unspported contexts and for correct codegen in
supported contexts. Verify errors persist in pre-6.9 shader models

The type buffer cache expects a max vector size of 4. By just skipping the cache for longer vectors, we don't overrun and store float7 vectors in the double3 slot or retrieve the double3 in place of float7.

Testing is for acceptance, mangling and basic copying that takes place
at the high level to ensure they are being accepted and recognized
correctly. The intent is not to tully test the passing of data as that
requires enabling vector operations to do properly. This test is used to
verify that these same constructs are disallowed in 6.8 and earlier.

A separate test verifies that disallowed contexts produce the
appropriate errors

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7117


tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -903,18 +903,19 @@ void hlsl::AddStdIsEqualImplementation(clang::ASTContext &context,
 /// <parm name="templateArgCount">Number of template arguments (one or
 /// two).</param> <parm name="defaultTypeArgValue">If assigned, the default
 /// argument for the element template.</param>
-CXXRecordDecl *
-hlsl::DeclareTemplateTypeWithHandle(ASTContext &context, StringRef name,
-                                    uint8_t templateArgCount,
-                                    TypeSourceInfo *defaultTypeArgValue) {
+CXXRecordDecl *hlsl::DeclareTemplateTypeWithHandle(
+    ASTContext &context, StringRef name, uint8_t templateArgCount,
+    TypeSourceInfo *defaultTypeArgValue, InheritableAttr *Attr) {
```
alsepkow, line 908: minor nit: Why the difference in signature formatting in this file? See 'DeclareRecordTypeWithHandleAndNoMemberFunctions'. Most look look like they match the formatting you swapped to.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11882,6 +11909,33 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
   return false;
 }
 
+bool hlsl::HasLongVecs(const QualType &qt) {
+  if (qt.isNull()) {
+    return false;
+  }
+
+  if (IsHLSLVecType(qt)) {
+    if (GetHLSLVecSize(qt) > 4)
```
alsepkow, line 11918: >4 [](http://example.com/codeflow?start=29&length=1)

Would this be better defined as a constant somewhere?

llvm-beanz: I agree we should reduce the number of places we put the magic value. Can we make this a language option though? Then we can query the Language Options for something like `maxSupportedVectorSize()`?

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7843,6 +7841,8 @@ def err_hlsl_load_from_mesh_out_arrays: Error<
    "output arrays of a mesh shader can not be read from">;
 def err_hlsl_out_indices_array_incorrect_access: Error<
    "a vector in out indices array must be accessed as a whole">;
+def err_hlsl_unsupported_long_vector: Error<
+   "Vectors of over 4 elements in %0 are not supported">;
```
llvm-beanz, line 7845: Seems like the element count could be a format argument too so that we can error on vectors too large for the large-vectors proposal too?

tools/clang/include/clang/Sema/SemaHLSL.h
```diff
@@ -128,6 +128,8 @@ unsigned CaculateInitListArraySizeForHLSL(clang::Sema *sema,
                                           const clang::InitListExpr *InitList,
                                           const clang::QualType EltTy);
 
+bool HasLongVecs(const clang::QualType &qt);
```
llvm-beanz, line 131: nit: IIUC, this is really checking if the type passed in contains a sub-object of long vector type?
```suggestion
bool ContainsLongVecSubObject(const clang::QualType &);
```

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -1131,6 +1135,9 @@ hlsl::DeclareConstantBufferViewType(clang::ASTContext &context, bool bTBuf) {
   typeDeclBuilder.addField(
       "h", context.UnsignedIntTy); // Add an 'h' field to hold the handle.
 
+  typeDeclBuilder.getRecordDecl()->addAttr(
+      HLSLCBufferAttr::CreateImplicit(context));
```
llvm-beanz, line 1139: I really appreciate this cleanup. Thank you for doing this.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5055,7 +5053,10 @@ class HLSLExternalSource : public ExternalSemaSource {
 
   bool CheckRangedTemplateArgument(SourceLocation diagLoc,
                                    llvm::APSInt &sintValue) {
-    if (!sintValue.isStrictlyPositive() || sintValue.getLimitedValue() > 4) {
+    const auto *SM =
+        hlsl::ShaderModel::GetByName(m_sema->getLangOpts().HLSLProfile.c_str());
+    if (!sintValue.isStrictlyPositive() ||
+        (sintValue.getLimitedValue() > 4 && !SM->IsSM69Plus())) {
```
llvm-beanz, line 5059: I thought we were not requiring SM 6.9 for long vectors. Am I misreading this or misremembering?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11882,6 +11909,33 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
...
+      return true;
+  } else if (qt->isArrayType()) {
+    const ArrayType *arrayType = qt->getAsArrayTypeUnsafe();
+    return HasLongVecs(arrayType->getElementType());
+  } else if (qt->isStructureOrClassType()) {
+    const RecordType *recordType = qt->getAs<RecordType>();
+    const RecordDecl *recordDecl = recordType->getDecl();
+    if (recordDecl->isInvalidDecl())
+      return false;
+    RecordDecl::field_iterator begin = recordDecl->field_begin();
```
llvm-beanz, line 11928: What about base classes?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11882,6 +11909,33 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
   return false;
 }
 
+bool hlsl::HasLongVecs(const QualType &qt) {
+  if (qt.isNull()) {
+    return false;
+  }
```
llvm-beanz, line 11915: nit:
```suggestion
  if (qt.isNull())
    return false;
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11882,6 +11909,33 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
...
+    return HasLongVecs(arrayType->getElementType());
+  } else if (qt->isStructureOrClassType()) {
+    const RecordType *recordType = qt->getAs<RecordType>();
+    const RecordDecl *recordDecl = recordType->getDecl();
+    if (recordDecl->isInvalidDecl())
+      return false;
+    RecordDecl::field_iterator begin = recordDecl->field_begin();
+    RecordDecl::field_iterator end = recordDecl->field_end();
+    for (; begin != end; begin++) {
+      const FieldDecl *fieldDecl = *begin;
```
tex3d, line 11931: Could this use the `fields()`, like so?
```suggestion
    for (auto field : recordDecl->fields()) {
      const FieldDecl *fieldDecl = *field;
```

tex3d: Also, does field iteration cover base class fields as well? I think we need to iterate `CXXRecordDecl::bases()` as well.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11882,6 +11909,33 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
...
+  }
+
+  if (IsHLSLVecType(qt)) {
+    if (GetHLSLVecSize(qt) > 4)
+      return true;
+  } else if (qt->isArrayType()) {
+    const ArrayType *arrayType = qt->getAsArrayTypeUnsafe();
+    return HasLongVecs(arrayType->getElementType());
+  } else if (qt->isStructureOrClassType()) {
+    const RecordType *recordType = qt->getAs<RecordType>();
```
llvm-beanz, line 11924: We could potentially avoid doing the record decl traversal at all if we take a bit from `DefinitionData` and cache it as we add members and base classes to the decl.

tools/clang/lib/Sema/SemaHLSL.cpp
tex3d, line 5059: I'll leave my objection here:
This changes what can be parsed globally, not just based on what's used by an entry point being compiled. In my opinion, switching global HLSL language availability based on shader model is ugly and unexpected.

tex3d: One point on this code though:
Shouldn't we have something like a `MaximumVectorLength` value, which is checked here, and can be used with an updated diagnostic message?

tools/clang/include/clang/Basic/Attr.td
```diff
@@ -992,6 +992,18 @@ def HLSLNodeTrackRWInputSharing : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
+def HLSLCBuffer : InheritableAttr {
+  let Spellings = []; // No spellings!
+  let Subjects = SubjectList<[CXXRecord]>;
+  let Documentation = [Undocumented];
+}
+
+def HLSLTessPatch : InheritableAttr {
```
tex3d, line 1001: What about the geometry shader stream out objects (PointStream, LineStream, TriangleStream)?  These should also prevent long vectors in their template type argument, since that defines the signature outputs for the geometry shader.

tools/clang/include/clang/Sema/SemaHLSL.h
pow2clk: I didn't spend too much time considering this name. With other changes, I see this taking a max length param which makes the "long" in "longvec" subjective, but I'll name it something similar to your suggestion.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: It's an interesting idea. I didn't like the recursion and throwing in base classes just adds to the overhead.

tools/clang/test/SemaHLSL/hlsl/types/invalid_longvec_decls.hlsl
```diff
@@ -0,0 +1,132 @@
+// RUN: %dxc  -DTYPE=float -DNUM=7 -T ps_6_9 -verify %s
```
pow2clk, line 1: Reminder to self to add tests for vectors longer than 1024 as well as structs with base classes containing long vectors in all the forbidden locations.

pow2clk (commented):
Thanks for the feedback! Followup commit soon.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: This is requiring a lot more additional code than I am inclined to put in here. In particular, various template parameters are not instantiated until after the checks are done now, so the DefintionData is set way too late. We have some forcing of instantiation, but only if the parameter is a template itself. @llvm-beanz Will you still consider approving the change without this?

tools/clang/include/clang/Sema/SemaHLSL.h
llvm-beanz: I don't think it should take a maximum length parameter, isn't a "long" vector anything over 4?

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: Okay, that seems like a reasonable rationale. Separately maybe we should consider if supporting shorter long vectors in old shader models is reasonable. I was looking at some compute shaders last week that bent over backwards to create 8-element vector types which might benefit from a smaller relaxing of vector size limitations.

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: I'm a bit surprised by this, DefinitionData tends to get updated as members get added into records, so I would have expected it to be sufficiently easy to update this in the `addMember` function.

I don't consider this a blocker if it is more complicated than I expected.

tools/clang/include/clang/Basic/Attr.td
tex3d: Yes, a separate attribute for GS stream seems best.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: It's mainly the ConstantBufer<LonVec> usage where the template parameter is already weird and doesn't get addressed in the usual order of things. Other usages might be easier, but that was the one I started with.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -520,6 +520,18 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
+          self->Diag(param->getLocation(),
+                     diag::err_hlsl_unsupported_long_vector)
+              << DXIL::kDefaultMaxVectorLength
+              << "patch constant function parameters";
+
+      if (ContainsLongVector(pPatchFnDecl->getReturnType()))
+        self->Diag(pPatchFnDecl->getLocation(),
+                   diag::err_hlsl_unsupported_long_vector)
+            << DXIL::kDefaultMaxVectorLength
+            << "patch constant function return type";
```
alsepkow, line 534: curious, why can't long vectors be used in patch functions? I tried to find a reference in the spec and didn't see anything. Am I missing something? Just trying to connect some dots to help learning.

tex3d (commented):
Finished with code review part.  Haven't thoroughly reviewed all the tests yet.

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -1134,6 +1147,10 @@ hlsl::DeclareConstantBufferViewType(clang::ASTContext &context, bool bTBuf) {
   typeDeclBuilder.addField(
       "h", context.UnsignedIntTy); // Add an 'h' field to hold the handle.
 
+  typeDeclBuilder.getRecordDecl()->addAttr(HLSLResourceAttr::CreateImplicit(
+      context, (unsigned)DXIL::ResourceKind::CBuffer,
+      (unsigned)DXIL::ResourceClass::CBuffer));
```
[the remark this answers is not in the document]
pow2clk: I see the rationale for this. I also take from it that this line is wrong: https://github.com/microsoft/DirectXShaderCompiler/blob/0a1143572d107c8b6980df092b84a79190ec1fbd/tools/clang/lib/CodeGen/CGHLSLMS.cpp#L3024

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7853,6 +7851,8 @@ def err_hlsl_load_from_mesh_out_arrays: Error<
    "output arrays of a mesh shader can not be read from">;
 def err_hlsl_out_indices_array_incorrect_access: Error<
    "a vector in out indices array must be accessed as a whole">;
+def err_hlsl_unsupported_long_vector
+    : Error<"vectors of over 4 elements in %0 are not supported">;
```
llvm-beanz, line 7855: DXC has not historically done this well, but it would be better if our diagnostics followed Clang convention:
https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/docs/InternalsManual.rst#the-format-string

Specifically, arguments that are just going to be constant strings are discouraged, instead you should use `%select`

tools/clang/lib/AST/DeclCXX.cpp
```diff
@@ -48,34 +48,31 @@ void LazyASTUnresolvedSet::getFromExternalSource(ASTContext &C) const {
...
+      DefaultedDefaultConstructorIsConstexpr(true),
+      HasConstexprDefaultConstructor(false),
+      HasNonLiteralTypeFieldsOrBases(false), ComputedVisibleConversions(false),
+      UserProvidedDefaultConstructor(false), DeclaredSpecialMembers(0),
+      ImplicitCopyConstructorHasConstParam(true),
+      ImplicitCopyAssignmentHasConstParam(true),
+      HasDeclaredCopyConstructorWithConstParam(false),
+      HasDeclaredCopyAssignmentWithConstParam(false), IsLambda(false),
+      IsParsingBaseSpecifiers(false), HasHLSLLongVector(false), NumBases(0),
+      NumVBases(0), Bases(), VBases(), Definition(D), FirstFriend() {}
```
llvm-beanz, line 75:
```suggestion
    // HLSL Change Begin - Add HasLongVector and clang-format
    : UserDeclaredConstructor(false), UserDeclaredSpecialMembers(0),
      Aggregate(true), PlainOldData(true), Empty(true), Polymorphic(false),
      Abstract(false), IsStandardLayout(true), HasNoNonEmptyBases(true),
      HasPrivateFields(false), HasProtectedFields(false),
      HasPublicFields(false), HasMutableFields(false), HasVariantMembers(false),
      HasOnlyCMembers(true), HasInClassInitializer(false),
      HasUninitializedReferenceMember(false),
      NeedOverloadResolutionForMoveConstructor(false),
      NeedOverloadResolutionForMoveAssignment(false),
      NeedOverloadResolutionForDestructor(false),
      DefaultedMoveConstructorIsDeleted(false),
      DefaultedMoveAssignmentIsDeleted(false),
      DefaultedDestructorIsDeleted(false), HasTrivialSpecialMembers(SMF_All),
      DeclaredNonTrivialSpecialMembers(0), HasIrrelevantDestructor(true),
      HasConstexprNonCopyMoveConstructor(false),
      DefaultedDefaultConstructorIsConstexpr(true),
      HasConstexprDefaultConstructor(false),
      HasNonLiteralTypeFieldsOrBases(false), ComputedVisibleConversions(false),
      UserProvidedDefaultConstructor(false), DeclaredSpecialMembers(0),
      ImplicitCopyConstructorHasConstParam(true),
      ImplicitCopyAssignmentHasConstParam(true),
      HasDeclaredCopyConstructorWithConstParam(false),
      HasDeclaredCopyAssignmentWithConstParam(false), IsLambda(false),
      IsParsingBaseSpecifiers(false), HasHLSLLongVector(false), NumBases(0),
      NumVBases(0), Bases(), VBases(), Definition(D), FirstFriend() {}
      // HLSL Change End
```

tools/clang/lib/AST/DeclCXX.cpp
```diff
@@ -204,6 +201,10 @@ CXXRecordDecl::setBases(CXXBaseSpecifier const * const *Bases,
     if (!BaseClassDecl->isStandardLayout())
       data().IsStandardLayout = false;
 
+    // Propagate presence of long vector to child classes.
+    if (BaseClassDecl->hasHLSLLongVector())
+      data().HasHLSLLongVector = true;
+
```
llvm-beanz, line 207:
```suggestion
    // HLSL Change Begin - Long vector support
    // Propagate presence of long vector to child classes.
    if (BaseClassDecl->hasHLSLLongVector())
      data().HasHLSLLongVector = true;
    // HLSL Change End - Long vector support
```

tools/clang/lib/AST/DeclCXX.cpp
```diff
@@ -385,6 +386,9 @@ void CXXRecordDecl::addedClassSubobject(CXXRecordDecl *Subobj) {
     data().NeedOverloadResolutionForMoveConstructor = true;
     data().NeedOverloadResolutionForDestructor = true;
   }
+
+  if (Subobj->hasHLSLLongVector())
+    data().HasHLSLLongVector = true;
```
llvm-beanz, line 391:
```suggestion
  // HLSL Change Begin - Long vector support
  if (Subobj->hasHLSLLongVector())
    data().HasHLSLLongVector = true;
  // HLSL Change End - Long vector support
```

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -53,44 +53,44 @@ ConvertHLSLVecMatTypeToExtVectorType(const clang::ASTContext &context,
...
+  type = type.getCanonicalType();
+  if (const RecordType *RT = type->getAs<RecordType>()) {
+    if (const auto *Spec =
+            dyn_cast<ClassTemplateSpecializationDecl>(RT->getDecl()))
+      if (const auto *Template =
+              dyn_cast<ClassTemplateDecl>(Spec->getSpecializedTemplate()))
+        return Template->getTemplatedDecl()->getAttr<HLSLMatrixAttr>() ||
+               Template->getTemplatedDecl()->getAttr<HLSLVectorAttr>();
+    if (const auto *Decl = dyn_cast<CXXRecordDecl>(RT->getDecl()))
+      return Decl->getAttr<HLSLMatrixAttr>() || Decl->getAttr<HLSLVectorAttr>();
```
llvm-beanz, line 80: Doesn't this just do the same thing as the `getAttr` helper above but for two types in parallel?

I know the extra RTTI queries seem wasteful, but it might be a lot easier to read and maintain this code if it used the helper.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -474,160 +474,52 @@ clang::QualType GetHLSLMatElementType(clang::QualType type) {
...
-    if (name == "SamplerState" || name == "SamplerComparisonState")
-      return true;
 
-    if (name == "ConstantBuffer" || name == "TextureBuffer")
-      return true;
-
-    if (name == "RaytracingAccelerationStructure")
-      return true;
-  }
+bool IsHLSLResourceType(clang::QualType type) {
```
llvm-beanz, line 515: Cannot tell you how much I _love_ this cleanup!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -4741,7 +4755,7 @@ class HLSLExternalSource : public ExternalSemaSource {
       return true;
     case AR_OBJECT_TEXTURE_BUFFER:
```
tex3d, line 4756: You fixed issues (presumably will fix Consume too) I pointed out in `GetHLSLResourceProperties()`, could you fix the Append/Consume cases above (both are UAVs) while you're at it?

tools/clang/include/clang/Sema/SemaHLSL.h
```diff
@@ -128,7 +128,7 @@ unsigned CaculateInitListArraySizeForHLSL(clang::Sema *sema,
                                           const clang::InitListExpr *InitList,
                                           const clang::QualType EltTy);
 
-bool ContainsLongVector(clang::QualType qt);
+bool containsLongVector(clang::QualType qt);
```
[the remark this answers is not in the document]
pow2clk: Chris requested it.

lib/DXIL/DxilUtil.cpp
```diff
@@ -426,35 +426,37 @@ GetHLSLResourceProperties(llvm::Type *Ty) {
...
+      return RetType(true,
+                     MakeResourceProperties(
+                         hlsl::DXIL::ResourceKind::StructuredBuffer,
+                         /*UAV*/ true, /*ROV*/ false, /*cmp or counter*/ true));
 
     if (name.startswith("ConsumeStructuredBuffer<"))
       return RetType(true, MakeResourceProperties(
                                hlsl::DXIL::ResourceKind::StructuredBuffer,
-                               false, false, /*cmp or counter*/ true));
+                               /*UAV*/ false, /*ROV*/ false,
```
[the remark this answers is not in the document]
pow2clk: Yeah I acknowledge that. The code was wrong before and I missed it when I started identifying the arguments.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
pow2clk: I don't know if you are requesting a change or validating the change I made before

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Just want to make sure I understand. The changes are to `ConsumeStructuredBuffer` in `GetHLSLResourceProperties` in DxilUtil.cpp and to the cases for `AR_OBJECT_CONSUME_STRUCTURED_BUFFER`  `AR_OBJECT_APPEND_STRUCTURED_BUFFER` here?

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Handled through definitionData

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Done.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Done

tools/clang/test/SemaHLSL/hlsl/types/invalid_longvec_decls.hlsl
pow2clk: Add in toolong_vectors.hlsl

tools/clang/lib/AST/DeclCXX.cpp
pow2clk: Done

tools/clang/lib/AST/DeclCXX.cpp
pow2clk: Done

tools/clang/lib/AST/DeclCXX.cpp
pow2clk: Done

tools/clang/lib/AST/HlslTypes.cpp
pow2clk: Done.

lib/DXIL/DxilUtil.cpp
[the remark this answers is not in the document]
pow2clk: Done.

tools/clang/include/clang/Sema/SemaHLSL.h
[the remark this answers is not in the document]
llvm-beanz: LLVM style guidelines: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
llvm-beanz: I'm requesting that you use `%select` to select a substring, instead of passing string arguments to `Diag`.

pow2clk: I'll bend to the guidelines, but having `<< 0` instead of `<< "ConstantBuffers or TextureBuffers` doesn't seem more readable to me.

llvm-beanz (approved):

tools/clang/test/SemaHLSL/hlsl/types/toolong-vectors.hlsl
```diff
@@ -0,0 +1,116 @@
+// RUN: %dxc -T lib_6_9 -DTYPE=float -DNUM=1025 -verify %s
+// RUN: %dxc -T ps_6_9  -DTYPE=float -DNUM=1025 -verify %s
+
+// A test to verify that declarations of longvecs are permitted in all the accepted places.
+// Only tests for acceptance, most codegen is ignored for now.
```
alsepkow, line 5: When you say "most codegen is ignored for now" do you mean until long vecs are implemented? Do we intend to use this hlsl file to verify codegen at some point? Just wondering if this comment is going to become stale.

PR #7155, opened 2025-02-20
Allow native vectors for LLVM operations

### LINKED ISSUE #7123: [SM6.9] Generate native vector variants of elementwise operators
In 6.9, elementwise vector operators should produce operators that work on the native vectors instead of scalarizing them. These operations include:
* arithmetic operations: `+`, `-`(unary and binary), `*`, `/`, `%`, `+=`, `-=`, `*=`, `/=`
* bitwise operations: `|`, `&`, `^`, `~`, `|=`, `&=`, `^=`
* equality operations: `==`, `!=`, `<`, `>`, `<=`, `>=`
* logic operations: `||`, `&&`, `?:`

Note that logic operations are disallowed for vectors as of HLSL2021 in favor of `and`, `or`, and `select`

This will largely be accomplished by disabling the scalarization passes that previously converted vectors to scalar variants. Testing should verify that the DXIL output preserves the vectors as arguments for all of these operators.

pow2clk:
Disables various forms of scalarization and vector elimination to permit
vectors to pass through to final DXIL when used in native LLVM
operations and loading/storing.

Introduces a few vector manipulation llvm instructions to DXIL allowing
for them to appear in output DXIL.

Skips passes for 6.9 that scalarize, convert to arrays, or otherwise eliminate vectors.
This eliminates the element-by-element loading of the vectors
In many cases, this required plumbing the shader model information to
passes that didn't have it before.

Many changes were needed for the MatrixBitcastLower pass related to
linking to avoid converting matrix vectors, but also to perform the
conversion if a shader was compiled for 6.9+, but then linked to a
earlier target.
This now adapts to the linker target to either preserve vectors for 6.9 or arrays for previous versions.
This requires running the DynamicIndexing VectorToArray pass during linking since 6_x and 6_9+ will fail to run this in the initial compile, but will still need to lower vectors to arrays.

Ternary conditional/select operators were element extracted in codegen.
Removing this allows 6.9 to preserve the vectors, but also maintains
behavior for previous shader models because the operations get
scalarized later anyway.

Keep groupshared variables as vectors for 6.9. They are no longer represented as indivual groupshared scalars.

Adds extensive tests for these operations using different types and
sizes and testing them appropriately. Booleans produce significantly
different code, so they get their own test.

Fixes #7123


pow2clk:
The first two commits are dependencies that belong to a different PR. Only the last is relevant to this: 68a284b49eeb7f38bd1f963da55a7067da40d3de

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7843,6 +7841,8 @@ def err_hlsl_load_from_mesh_out_arrays: Error<
    "output arrays of a mesh shader can not be read from">;
 def err_hlsl_out_indices_array_incorrect_access: Error<
    "a vector in out indices array must be accessed as a whole">;
+def err_hlsl_unsupported_long_vector: Error<
+   "Vectors of over 4 elements in %0 are not supported">;
```
llvm-beanz, line 7845:
```suggestion
   "vectors of over 4 elements in %0 are not supported">;
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11882,6 +11913,33 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
   return false;
 }
 
+bool hlsl::HasLongVecs(const QualType &qt) {
```
llvm-beanz, line 11916: nit: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11882,6 +11913,33 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
...
+  if (qt.isNull()) {
+    return false;
+  }
```
llvm-beanz, line 11919: nit: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
  if (qt.isNull())
    return false;
```

llvm-beanz (commented):
This PR has changes to multiple passes that do not have IR tests. We should have tests that verify correctness of the changes in the passes separately from the Clang-level changes.

lib/HLSL/HLMatrixBitcastLowerPass.cpp
```diff
@@ -193,31 +194,42 @@ void MatrixBitcastLowerPass::lowerMatrix(Instruction *M, Value *A) {
         SmallVector<Value *, 2> idxList(GEP->idx_begin(), GEP->idx_end());
         DXASSERT(idxList.size() == 2,
                  "else not one dim matrix array index to matrix");
-
-        HLMatrixType MatTy = HLMatrixType::cast(EltTy);
-        Value *matSize = Builder.getInt32(MatTy.getNumElements());
-        idxList.back() = Builder.CreateMul(idxList.back(), matSize);
+        if (!DM.GetShaderModel()->IsSM69Plus()) {
```
tex3d, line 197: Can we base the low-level decisions on a descriptively named boolean field in the `MatrixBitcastLowerPass` class, initialized based on the shader model, instead of passing the DxilModule down and checking the shader model each time?

tex3d: Maybe even something about scalarizing operations, rather than anything to do with the native DXIL vector support?

lib/HLSL/HLMatrixBitcastLowerPass.cpp
```diff
@@ -193,31 +194,42 @@ void MatrixBitcastLowerPass::lowerMatrix(Instruction *M, Value *A) {
...
+          NewVec = NewLI;
+        } else {
+          Value *zeroIdx = Builder.getInt32(0);
+          unsigned vecSize = Ty->getNumElements();
+          NewVec = UndefValue::get(LI->getType());
+          for (unsigned i = 0; i < vecSize; i++) {
+            Value *GEP = CreateEltGEP(A, i, zeroIdx, Builder);
+            Value *Elt = Builder.CreateLoad(GEP);
+            NewVec = Builder.CreateInsertElement(NewVec, Elt, i);
+          }
```
tex3d, line 232: I noticed that below this point, there's:
```cpp
    } else if (StoreInst *ST = dyn_cast<StoreInst>(U)) {
```
where it still scalarizes the store for the vector.

Did you mean to leave that scalarization in?

lib/Transforms/Scalar/DxilEliminateVector.cpp
```diff
@@ -151,6 +153,10 @@ bool DxilEliminateVector::TryRewriteDebugInfoForVector(InsertElementInst *IE) {
 
 bool DxilEliminateVector::runOnFunction(Function &F) {
 
+  if (F.getParent()->HasDxilModule())
+    if (F.getParent()->GetDxilModule().GetShaderModel()->IsSM69Plus())
+      return false;
```
tex3d, line 158: So, this is where we presumably will still want to do something different for vec1, right?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -1869,7 +1869,8 @@ bool SROAGlobalAndAllocas(HLModule &HLM, bool bHasDbgInfo) {
       // if
       // all its users can be transformed, then split up the aggregate into its
       // separate elements.
-      if (ShouldAttemptScalarRepl(AI) && isSafeAllocaToScalarRepl(AI)) {
+      if (!HLM.GetShaderModel()->IsSM69Plus() && ShouldAttemptScalarRepl(AI) &&
```
tex3d, line 1872: This pass is so complicated as it is, can we capture a bool for DXIL vector support earlier and use that instead of the individual SM 6.9 checks here as well?

lib/Transforms/Scalar/Scalarizer.cpp
```diff
@@ -290,6 +292,10 @@ bool Scalarizer::doInitialization(Module &M) {
 }
 
 bool Scalarizer::runOnFunction(Function &F) {
+  if (F.getParent()->HasDxilModule())
+    if (F.getParent()->GetDxilModule().GetShaderModel()->IsSM69Plus())
+      return false;
```
tex3d, line 297: Do we actually want to turn off scalarization entirely?  Maybe this is one place where we need to preserve it for vec1 only?

tools/clang/lib/CodeGen/CGExprScalar.cpp
```diff
@@ -3713,20 +3713,7 @@ VisitAbstractConditionalOperator(const AbstractConditionalOperator *E) {
...
       llvm::Value *RHS = Visit(rhsExpr);
-      if (llvm::VectorType *VT = dyn_cast<llvm::VectorType>(CondV->getType())) {
-        llvm::VectorType *ResultVT = cast<llvm::VectorType>(LHS->getType());
-        llvm::Value *result = llvm::UndefValue::get(ResultVT);
-        for (unsigned i = 0; i < VT->getNumElements(); i++) {
-          llvm::Value *EltCond = Builder.CreateExtractElement(CondV, i);
-          llvm::Value *EltL = Builder.CreateExtractElement(LHS, i);
-          llvm::Value *EltR = Builder.CreateExtractElement(RHS, i);
-          llvm::Value *EltSelect = Builder.CreateSelect(EltCond, EltL, EltR);
-          result = Builder.CreateInsertElement(result, EltSelect, i);
```
tex3d, line 3724: I guess nothing depended on this early scalarization of select?  I wonder why it was added in the first place.  I could imagine subtle regressions by removing this code if some optimization before scalarization depended on it.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6487,6 +6487,9 @@ bool HLSLExternalSource::MatchArguments(
     }
   }
 
+  std::string profile = m_sema->getLangOpts().HLSLProfile;
+  const ShaderModel *SM = hlsl::ShaderModel::GetByName(profile.c_str());
```
tex3d, line 6491: I wonder if we couldn't have a LangOpt for HLSLLongVectorSupport, initialized based on the shader model instead?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6657,8 +6660,9 @@ bool HLSLExternalSource::MatchArguments(
       }
 
       // Verify that the final results are in bounds.
-      CAB(uCols > 0 && uCols <= MaxVectorSize && uRows > 0 &&
-              uRows <= MaxVectorSize,
+      CAB((uCols > 0 && uRows > 0 &&
+           ((uCols <= MaxVectorSize && uRows <= MaxVectorSize) ||
+            (SM->IsSM69Plus() && uRows == 1))),
```
tex3d, line 6665: How is this change relevant to allowing native LLVM vectors?  Shouldn't this be part of #7143 instead?

tools/clang/unittests/HLSL/LinkerTest.cpp
```diff
@@ -526,6 +526,11 @@ TEST_F(LinkerTest, RunLinkMatArrayParam) {
   Link(L"main", L"ps_6_0", pLinker, {libName, libName2},
        {"alloca [24 x float]", "getelementptr [12 x float], [12 x float]*"},
        {});
+
+  Link(L"main", L"ps_6_9", pLinker, {libName, libName2},
+       {"alloca [2 x <12 x float>]",
+        "getelementptr [12 x float], [12 x float]*"},
```
tex3d, line 532: I'm not sure how to tell if this is right.  Shouldn't this be dealing with a vector rather than an array here?  Can I see the IR?

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Not part of this PR. I've responded to it here: https://github.com/microsoft/DirectXShaderCompiler/pull/7143/commits/50d1af5b645651b7ee4d4ef063bdc88c7d6790d4

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
pow2clk: This change is not part of this PR, but a prerequisite as I tried to explain in this comment: https://github.com/microsoft/DirectXShaderCompiler/pull/7155#issuecomment-2672803480 I have fixed it in the relevant PR: #7143

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: As above, this is not part of this PR, but part of the prerequisite #7143. I renamed it inline with your suggestion on that change, which I'm afraid doesn't match the coding standards you cited. https://github.com/microsoft/DirectXShaderCompiler/pull/7143/commits/e010223f74f9c2d96d51349849cd5f5e99542e99#r1960826249

pow2clk:
I have fixed the merge conflicts and responded to some of the feedback locally, but the way the dependencies are set up, if I push those changes before #7143 is merged, it's going to make distinguishing the prerequisite changes and the changes that are the subject of this PR even harder.

joaosaffran (approved):
Code looks good to me. Not an expert though

lib/Transforms/Scalar/Scalarizer.cpp
pow2clk: Proper support for vec1 was missing from this change. I've added it now.

tools/clang/unittests/HLSL/LinkerTest.cpp
pow2clk: Well I was ready to write something here about how this was compiled as 6.3 because that's what the file says, but it's actually 6_x. The 12 element array comes from HLMatrixLower, which I hadn't touched.

pow2clk: Okay, it's just scratch space used to move the elements of the matrix around. They use vectors too. I don't know why the test cares about it so much, but it doesn't need changing. I'll take a look at leveraging long vectors for matrices later, but this works perfectly fine for now.

include/dxc/DXIL/DxilInstructions.h
```diff
@@ -645,6 +645,42 @@ struct LlvmInst_VAArg {
...
+/// This instruction extracts from vector
+struct LlvmInst_ExtractElement {
+  llvm::Instruction *Instr;
+  // Construction and identification
+  LlvmInst_ExtractElement(llvm::Instruction *pInstr) : Instr(pInstr) {}
+  operator bool() const {
+    return Instr->getOpcode() == llvm::Instruction::ExtractElement;
+  }
+  // Validation support
+  bool isAllowed() const { return true; }
```
farzonl, line 657: A little suprised to see `ExtractElement`\`InsertElement`\`ShuffleVector` weren't already  defined since I expected at least the first two as required for scalarization which DXIL already supports.

If these are new for vectorization, should` isAllowed()` be checking `IsSM69Plus()` instead of always returning true?

lib/HLSL/HLMatrixBitcastLowerPass.cpp
```diff
@@ -193,31 +197,42 @@ void MatrixBitcastLowerPass::lowerMatrix(Instruction *M, Value *A) {
         SmallVector<Value *, 2> idxList(GEP->idx_begin(), GEP->idx_end());
         DXASSERT(idxList.size() == 2,
                  "else not one dim matrix array index to matrix");
-
-        HLMatrixType MatTy = HLMatrixType::cast(EltTy);
-        Value *matSize = Builder.getInt32(MatTy.getNumElements());
-        idxList.back() = Builder.CreateMul(idxList.back(), matSize);
+        unsigned NumElts = HLMatrixType::cast(EltTy).getNumElements();
```
farzonl, line 200: I've gotten in a habit where if I have to do `isa<>` and then `cast<>` to instead just do dyn_cast<>? Does that make sense in DXC? In other words, is there a way to instead do something similar to a `dyn_cast` in the conditional like if (HLMatrixType* MatTy = HLMatrixType::dyncast(EltTy))  for line 192.

lib/Transforms/Scalar/Scalarizer.cpp
```diff
@@ -290,6 +293,10 @@ bool Scalarizer::doInitialization(Module &M) {
 }
 
 bool Scalarizer::runOnFunction(Function &F) {
+  if (F.getParent()->HasDxilModule())
+    if (F.getParent()->GetDxilModule().GetShaderModel()->IsSM69Plus())
```
farzonl, line 297:
```suggestion
  auto *Parent = F.getParent();
  if (Parent->HasDxilModule())
    if (Parent->GetDxilModule().GetShaderModel()->IsSM69Plus())
```

tools/clang/test/HLSLFileCheck/passes/dxil/lower_type/vec_array_param.ll
```diff
@@ -31,3 +31,25 @@ declare float @"\01?foo@@YAMY02V?$vector@M$02@@@Z"([3 x <3 x float>]*)
...
+!dx.entryPoints = !{!12}
+!dx.fnprops = !{}
+!dx.options = !{!13, !14}
+
+!0 = !{i32 2, !"Debug Info Version", i32 3}
+!1 = !{!"hlsl-hlemit", !"hlsl-hlensure"}
+!3 = !{i32 1, i32 6}
+!4 = !{i32 1, i32 6}
+!5 = !{!"lib", i32 6, i32 6}
+!6 = !{i32 1, float ([3 x <3 x float>]*)* @"\01?bar@@YAMY02V?$vector@M$02@@@Z", !7, float ([3 x <3 x float>]*)* @"\01?foo@@YAMY02V?$vector@M$02@@@Z", !7}
```
farzonl, line 47: what happens if we have a `[3 x [3 x <3 x float>]]`?

utils/hct/hctdb.py
```diff
@@ -1175,6 +1175,37 @@ def populate_llvm_instructions(self):
         self.add_llvm_instr(
             "OTHER", 53, "VAArg", "VAArgInst", "vaarg instruction", "", []
         )
+
+        self.add_llvm_instr(
```
farzonl, line 1179: What does it mean to `add_llvm_instr` in this case? I don't think it means anything about validity because i've seen these instructions emitted by DXC before?

Does this means there is a way to print out all the llvm instructions DXC supports via querying the `DirectXShaderCompiler.utils.hct.hctdb_instrhelp` package? 

I'm curious for my own reasons because I want to build a  [Set difference](https://en.wikipedia.org/wiki/Set_difference) of DXC to upstream llvm and know which instructions need to be legalized.

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-operators-int.hlsl
```diff
@@ -0,0 +1,73 @@
...
+// RUN: %dxc -T lib_6_9   -DTYPE=uint16_t -DNUM=9 -enable-16bit-types %s | FileCheck %s --check-prefixes=CHECK,UNSIG
+
+// Test bitwise operators on an assortment vector sizes and integer types with 6.9 native vectors.
+
+// Test bit twiddling operators.
+// CHECK-LABEL: define void @"\01?bittwiddlers
+// CHECK-SAME: ([11 x <[[NUM:[0-9][0-9]*]] x [[TYPE:[a-z0-9]*]]>]*
+export void bittwiddlers(inout vector<TYPE, NUM> things[11]) {
+  // CHECK: [[adr1:%[0-9]*]] = getelementptr inbounds [11 x <[[NUM]] x [[TYPE]]>], [11 x <[[NUM]] x [[TYPE]]>]* %things, i32 0, i32 1
+  // CHECK: [[vec1:%[0-9]*]] = load <[[NUM]] x [[TYPE]]>, <[[NUM]] x [[TYPE]]>* [[adr1]]
```
farzonl, line 12: would it make sense to make most of the `CHECK-NEXT`?

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -1945,7 +1949,7 @@ bool SROAGlobalAndAllocas(HLModule &HLM, bool bHasDbgInfo) {
         continue;
       }
 
-      // Flat Global vector if no dynamic vector indexing.
+      // Flatten Global vector if no dynamic vector indexing.
```
llvm-beanz, line 1952: nit: If we're fixing a comment, might as well make it a complete sentence.
```suggestion
      // Flatten global vector if no dynamic vector indexing occurs.
```

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-operators-bool.hlsl
```diff
@@ -0,0 +1,464 @@
...
+
+  // CHECK: [[add4:%[0-9]*]] = getelementptr inbounds [10 x <[[NUM]] x i32>], [10 x <[[NUM]] x i32>]* %truth, i32 0, i32 4
+  // CHECK: [[vec4:%[0-9]*]] = load <[[NUM]] x i32>, <[[NUM]] x i32>* [[add4]]
+  // CHECK: [[bvec4:%[0-9]*]] = icmp ne <[[NUM]] x i32> [[vec4]], zeroinitializer
+  // CHECK: [[add5:%[0-9]*]] = getelementptr inbounds [10 x <[[NUM]] x i32>], [10 x <[[NUM]] x i32>]* %truth, i32 0, i32 5
+  // CHECK: [[vec5:%[0-9]*]] = load <[[NUM]] x i32>, <[[NUM]] x i32>* [[add5]]
+  // CHECK: [[bvec5:%[0-9]*]] = icmp ne <[[NUM]] x i32> [[vec5]], zeroinitializer
+  // MORE STUFF
+
+  res[3] = truth[3] ? truth[4] : truth[5];
```
llvm-beanz, line 284: Can we have a HLSL 2018 test case something like:
```
export float4 main(bool4 B, float4 X, float4 Y) {
    return B ? ++X : Y;
}
```

What I want to verify is that the side-effect of the LHS always occurs and no branch is generated under HLSL 2018.

lib/Transforms/Scalar/LowerTypePasses.cpp
```diff
@@ -207,6 +211,21 @@ class DynamicIndexingVectorToArray : public LowerTypePass {
   void ReplaceAddrSpaceCast(ConstantExpr *CE, Value *A, IRBuilder<> &Builder);
 };
 
+void DynamicIndexingVectorToArray::initialize(Module &M) {
+  // Can be invoked in a few places:
+  //  - From standard compile before dxilgen.
+  //  - When linking, where dxmodule is available.
+  //  - In isolated dxopt, where the module will need to be created.
+  // Since HL module can't be created when linking, check for that first.
+  // Otherwise, either retrieve or generate the HL module.
```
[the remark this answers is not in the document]
pow2clk: No, we actually can create it when linking. Maybe I'll replace it with getorcreatedxilmodule just in case. We can't create an hlmodule while linking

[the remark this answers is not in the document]
pow2clk: I see why it's confusing now though. the "it" was supposed to be a dxil module, but that is not at all clear from the wording.

llvm-beanz (approved):

lib/HLSL/HLMatrixBitcastLowerPass.cpp
```diff
@@ -76,13 +76,17 @@ Type *TryLowerMatTy(Type *Ty) {
...
+  bool SupportsVectors = false;
 
 public:
   static char ID; // Pass identification, replacement for typeid
   explicit MatrixBitcastLowerPass() : FunctionPass(ID) {}
 
   StringRef getPassName() const override { return "Matrix Bitcast lower"; }
   bool runOnFunction(Function &F) override {
+    DxilModule &DM = F.getParent()->GetOrCreateDxilModule();
+    SupportsVectors = DM.GetShaderModel()->IsSM69Plus();
```
tex3d, line 88: If you want to avoid full DxilModule dependency, I think it would be reasonable to have something that can look up the DXIL version in the module without loading full metadata, whether or not it's HLM or DM.

lib/Transforms/Scalar/LowerTypePasses.cpp
```diff
@@ -207,6 +211,21 @@ class DynamicIndexingVectorToArray : public LowerTypePass {
...
+  if (M.HasDxilModule()) {
+    SupportsVectors = M.GetDxilModule().GetShaderModel()->IsSM69Plus();
+  } else {
+    HLModule &HLM = M.GetOrCreateHLModule();
+    SupportsVectors = HLM.GetShaderModel()->IsSM69Plus();
```
tex3d, line 225: I mentioned earlier, it might be better to get the DXIL version from metadata without relying on full HLModule/DxilModule metadata loading.  Maybe we can add a `dxilutil::ReadDxilVersion` function?

But the fixes to the HLModule loading are appreciated!

lib/HLSL/HLMatrixBitcastLowerPass.cpp
pow2clk: I like this idea. I know I mentioned it somewhere before. I'd prefer it not block this change though.

pow2clk:
> Generally looks good to me, though it'd be good to get explicit approval from one of the folks who've done a few rounds of review on it.
> 
> Some of the LLVM IR tests have a few bits of unreferenced or unnecessary metadata that it'd be nice to clean up before committing.

Just as an informational note, I discovered that the `!paseresume` line that I think was identified as unneeded at some point actually turns out to be in some places. Without it, we don't get the shader model, so we fall back to default pre-6.9 behavior. That's okay for the vec1 tests because they should be scalarized, but they weren't actually confirming that the scalarization still took place in 6.9, which was their intent.

lib/HLSL/HLMatrixBitcastLowerPass.cpp
```diff
@@ -180,44 +183,56 @@ Value *CreateEltGEP(Value *A, unsigned i, Value *zeroIdx,
 }
 } // namespace
 
-void MatrixBitcastLowerPass::lowerMatrix(Instruction *M, Value *A) {
+void MatrixBitcastLowerPass::lowerMatrix(DxilModule &DM, Instruction *M,
```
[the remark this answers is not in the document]
pow2clk: Yeah, it's a relic from when I was getting the version from it.

lib/Transforms/Scalar/Scalarizer.cpp
```diff
@@ -290,6 +293,10 @@ bool Scalarizer::doInitialization(Module &M) {
 }
 
 bool Scalarizer::runOnFunction(Function &F) {
+  Module *M = F.getParent();
+  if (M->HasDxilModule() && M->GetDxilModule().GetShaderModel()->IsSM69Plus())
```
tex3d, line 297: I think the independent lookup of DXIL version would help here, since it won't do something different when there's no created `DxilModule` already present.

pow2clk: I could also replace it with getorcreate. I was a bit nervous about that becuase of how it failed for hlmodules, but I've since learned that it's a lot more robust and I've made the hlmodule variant similarly so.

PR #7156, opened 2025-02-24
[OMM] Implement front end diagnostics for OMM, including on TraceRayInline, and add Availability Attributes

### LINKED ISSUE #7145: [OMM] Add OMM Feature Front end, with D3D12_RAY_FLAGS
The OMM feature requires a new way to construct RayQuery objects, now with a second optional template parameter, `RayQueryFlags`.
D3D12_RAY_FLAGS is an enum of possible `RayQueryFlags` values that needs to be defined in HLSL.
This, along with front end diagnostics need to be added. These front end diagnostics are defined in the spec: 
https://github.com/microsoft/hlsl-specs/blob/main/proposals/0024-opacity-micromaps.md

Code needs to be added to implement the front-end part of the OMM feature. Diagnostics should be added, tests should test the diagnostics.

**Additional context**

bob80905:
This PR addresses the front end part of OMM, defining the new flags defined in the spec, and implementing the relevant diagnostics should the flags be incompatible. It also adds the second template argument to the RayQuery object, which is set to have a default value of 0 if no explicit template argument is provided.
Fixes #7145


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14308,6 +14313,91 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
...
+                    (unsigned)DXIL::RayQueryFlag::AllowOpacityMicromaps)) {
+                Diag(D.getLocStart(), diag::warn_hlsl_unexpected_rayquery_flag);
+                result = false;
+              }
+            }
+          }
+        }
+      }
+    }
+  }
```
damyanp, line 14399: Nesting this deep makes me wonder if there's some extraction or early returns that could be done here to try and make this code easier to follow.  I notice that this is the deepest nesting I could find on a quick survey through this file.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +325,96 @@ class HLSLCallDiagnoseVisitor
...
+
+    if (ME) {
+      const Decl *MD = ME->getMemberDecl();
+
+      if (const CXXMethodDecl *methodDecl = dyn_cast<CXXMethodDecl>(MD)) {
+        if (methodDecl->getNameAsString() == "TraceRayInline") {
+          // validate that if the ray flag parameter is
+          // RAY_FLAG_FORCE_OMM_2_STATE, the ray query object
+          // has the RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS
+          // flag set, otherwise emit a diagnostic
```
damyanp, line 355: Is this the "normal" way of doing this?  I'm a bit surprised to see string comparisons here, but maybe that's just how things are done?

damyanp (commented):
Couple of questions, but otherwise looks good to my untrained eye.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14161,6 +14161,97 @@ static bool IsUsageAttributeCompatible(AttributeList::Kind kind, bool &usageIn,
...
 }
 
+bool DiagnoseRayQueryObject(QualType qt, Declarator &D, Sema &S) {
+  // now we know that qt is a TemplateSpecializationType
+  // handle diagnostics that relate to enumtypedecl vs literal first
+  const TemplateSpecializationType *TST =
+      dyn_cast<TemplateSpecializationType>(qt);
+  if (!TST) {
+    return false;
+  }
```
damyanp, line 14171:
```suggestion
  if (!TST) 
    return false;
```

https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

(repeated throughout this code)

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14161,6 +14161,97 @@ static bool IsUsageAttributeCompatible(AttributeList::Kind kind, bool &usageIn,
   return true;
 }
 
+bool DiagnoseRayQueryObject(QualType qt, Declarator &D, Sema &S) {
```
damyanp, line 14164:
```suggestion
static bool DiagnoseRayQueryObject(QualType qt, Declarator &D, Sema &S) {
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14161,6 +14161,97 @@ static bool IsUsageAttributeCompatible(AttributeList::Kind kind, bool &usageIn,
...
+    }
+    return true;
+  }
+
+  // we're guaranteed 2 args now
+
+  // ensure that if the second template argument has a non-zero value,
+  // the shader model is at least 6.9
+  const TemplateArgument *Args = TST->getArgs();
+  auto Arg2 = Args[1];
```
damyanp, line 14205: Do you intend to create a copy of the template argument?

Maybe this should be:

```suggestion
  const TemplateArgument &Arg2 = Args[1];
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14161,6 +14161,97 @@ static bool IsUsageAttributeCompatible(AttributeList::Kind kind, bool &usageIn,
...
+  llvm::APSInt Arg2val = SpecDecl->getTemplateArgs()[1].getAsIntegral();
+  std::string profile = S.getLangOpts().HLSLProfile;
+  const ShaderModel *SM = hlsl::ShaderModel::GetByName(profile.c_str());
```
damyanp, line 14208: This makes an unnecessary copy of the HLSLProfile string.

Might as well just do this inline and not need to worry about the copy:

```suggestion
  const ShaderModel *SM = hlsl::ShaderModel::GetByName(S.getLangOpts().HLSLProfile.c_str());
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -14161,6 +14161,97 @@ static bool IsUsageAttributeCompatible(AttributeList::Kind kind, bool &usageIn,
...
+  // handle diagnostics that relate to enumtypedecl vs literal first
+  const TemplateSpecializationType *TST =
+      dyn_cast<TemplateSpecializationType>(qt);
+  if (!TST) {
+    return false;
+  }
+
+  auto typeRecordDecl = qt->getAsCXXRecordDecl();
+  ClassTemplateSpecializationDecl *SpecDecl =
+      llvm::dyn_cast<ClassTemplateSpecializationDecl>(typeRecordDecl);
```
damyanp, line 14175: It's not obvious to me what the type of `typeRecordDecl` is, and so I can't tell by reading just this code whether or not it is making an unnecessary copy or not.  An explicit `auto *` would help make that clearer, but I think really this could just be done inline (as I see it done in other places):

```suggestion
  ClassTemplateSpecializationDecl *SpecDecl =
      llvm::dyn_cast<ClassTemplateSpecializationDecl>(qt->getAsCXXRecordDecl());
```

Although in this case I think using auto would be ok because of the dyn_cast:


```suggestion
  auto *SpecDecl =
      llvm::dyn_cast<ClassTemplateSpecializationDecl>(qt->getAsCXXRecordDecl());
```

damyanp (commented):
Looking good - a few style nits and I think I spotted a few places where there's some unwanted copies happening.

tools/clang/lib/Sema/SemaHLSL.cpp
bob80905: I've done some early returns to help make this more flat.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
bob80905: I've found a way to compare opcodes instead of string comparing.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +325,136 @@ class HLSLCallDiagnoseVisitor
...
+    }
+    return false;
+  }
+
+  void DiagnoseTraceRayInline(const MemberExpr *ME,
+                              CXXMemberCallExpr *callExpr) {
+    // validate that if the ray flag parameter is
+    // RAY_FLAG_FORCE_OMM_2_STATE, the ray query object
+    // has the RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS
+    // flag set, otherwise emit a diagnostic
```
tex3d, line 382: Not "if the ray flag parameter is RAY_FLAG_FORCE_OMM_2_STATE", but "if the ray flag parameter has the RAY_FLAG_FORCE_OMM_2_STATE flag set".
```suggestion
    // If the RayFlag parameter has RAY_FLAG_FORCE_OMM_2_STATE set,
    // the RayQuery decl must have RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set,
    // otherwise emit a diagnostic.
```

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +325,136 @@ class HLSLCallDiagnoseVisitor
...
+          }
+        }
+      } else if (const IntegerLiteral *intLit =
+                     dyn_cast<IntegerLiteral>(RayFlagArg->getSubExpr())) {
+        unsigned int rayFlagValue = intLit->getValue().getZExtValue();
+        if (rayFlagValue & (unsigned int)DXIL::RayFlag::ForceOMM2State) {
+          IsRayFlagForceOMM2State = true;
+        }
+      }
+    }
```
tex3d, line 410: We should just use `isIntegerConstantExpr` to get the value from the ICE if possible.

```suggestion
    Expr *Expr1 = CE->getArg(1);
    llvm::APSInt constantResult;
    bool IsRayFlagForceOMM2State =
        Expr1->isIntegerConstantExpr(constantResult, sema->getASTContext()) &&
        (constantResult.getLimitedValue() &
         (uint64_t)DXIL::RayFlag::ForceOMM2State) != 0;
```

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +325,136 @@ class HLSLCallDiagnoseVisitor
...
+
+    if (!ME)
+      return false;
+
+    const Decl *MD = ME->getMemberDecl();
+
+    if (const CXXMethodDecl *methodDecl = dyn_cast<CXXMethodDecl>(MD)) {
+      auto *attr = methodDecl->getAttr<HLSLIntrinsicAttr>();
+      if (!attr)
+        return true;
```
tex3d, line 450: Need to check group before assuming you can cast and compare the opcode.

```suggestion
      auto *attr = methodDecl->getAttr<HLSLIntrinsicAttr>();
      if (!attr)
        return true;
      StringRef OpcodeGroup = GetHLOpcodeGroupName(HLOpcodeGroup::HLIntrinsic);
      if (attr->getGroup() != OpcodeGroup)
        return true;
```

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +325,136 @@ class HLSLCallDiagnoseVisitor
...
+        unsigned int rayFlagValue = intLit->getValue().getZExtValue();
+        if (rayFlagValue & (unsigned int)DXIL::RayFlag::ForceOMM2State) {
+          IsRayFlagForceOMM2State = true;
+        }
+      }
+    }
+    if (IsRayFlagForceOMM2State) {
+      const DeclRefExpr *DRE = dyn_cast<DeclRefExpr>(ME->getBase());
+      assert(DRE);
+      const VarDecl *VD = cast<VarDecl>(DRE->getDecl());
```
tex3d, line 414: I don't think this is the right approach.  I think you want to call `getImplicitObjectArgument()` on the `CXXMemberCallExpr` instead.  Then use the expression's type to drill into the template argument.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +325,136 @@ class HLSLCallDiagnoseVisitor
...
+        }
+      }
+    } else if (const IntegerLiteral *intLit =
+                   dyn_cast<IntegerLiteral>(RayFlagArg->getSubExpr())) {
+      unsigned int rayFlagValue = intLit->getValue().getZExtValue();
+      if (rayFlagValue & (unsigned int)DXIL::RayFlag::ForceOMM2State)
+        return true;
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
tex3d: Wait, after writing this, I don't think this function is even used.  It should be deleted then.

tex3d (commented):
Partial review, see comments so far.

lib/DxcSupport/FileIOHelper.cpp
```diff
@@ -1369,6 +1369,8 @@ class ReadOnlyBlobStream : public IStream {
       return E_POINTER;
     ULONG cbLeft = m_size - m_offset;
     *pcbRead = std::min(cb, cbLeft);
+    if (m_pMemory + m_offset == 0)
+      return S_FALSE;
```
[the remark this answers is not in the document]
bob80905: I believe there was some CompileTest that was failing, because the addition resulted in 0, so it was being treated as a nullptr when passed to memcpy. But I am no longer seeing this. I'll undo these 2 lines, and if it pops back up again I'll put up a separate PR.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7652,8 +7652,17 @@ def err_payload_fields_is_payload_and_overqualified : Error<
   "payload field '%0' is a payload struct. Payload access qualifiers are not allowed on payload types.">;
 def warn_hlsl_payload_qualifer_dropped : Warning<
   "payload access qualifiers ignored. These are only supported for lib_6_7+ targets and lib_6_6 with with the -enable-payload-qualifiers flag.">, InGroup<HLSLPayloadAccessQualifer>;
+def warn_hlsl_unexpected_value_for_ray_query_flags_template_arg : Warning<
+  "A non-zero value for the RayQueryFlags template argument requires shader model 6.9 or above.">, DefaultError;
```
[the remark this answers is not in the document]
bob80905: Ok, I've put the two warnings of interest into the `HLSLAvailabilityConstant` group

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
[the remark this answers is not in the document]
bob80905: Made use of the function and your suggestion 👍

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +326,101 @@ class HLSLCallDiagnoseVisitor
...
+    // Validate if the RayFlag parameter has RAY_FLAG_FORCE_OMM_2_STATE set,
+    // the RayQuery decl must have RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set,
+    // otherwise emit a diagnostic.
+    bool IsRayFlagForceOMM2State = IsRayFlagForceOMM2StateSet(callExpr);
+    if (IsRayFlagForceOMM2State) {
+      const DeclRefExpr *DRE =
+          dyn_cast<DeclRefExpr>(callExpr->getImplicitObjectArgument());
+      assert(DRE);
+      QualType QT = DRE->getType();
+      auto typeRecordDecl = QT->getAsCXXRecordDecl();
```
damyanp, line 376:
```suggestion
      auto *typeRecordDecl = QT->getAsCXXRecordDecl();
```

damyanp (approved):
Approved, assuming the test failures are fixed and Tex is happy you've addressed his feedback.

lib/DxcSupport/FileIOHelper.cpp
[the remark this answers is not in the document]
bob80905: Ok, it appeared again. Here's the output for the NIX Linux Clang release pipeline:
```
FAIL: Clang-Unit :: HLSL/ClangHLSLTests/CompilerTest.CodeGenHashStabilityHLSL (1674 of 4075)
******************** TEST 'Clang-Unit :: HLSL/ClangHLSLTests/CompilerTest.CodeGenHashStabilityHLSL' FAILED ********************
Note: Google Test filter = CompilerTest.CodeGenHashStabilityHLSL
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from CompilerTest
/home/vsts/work/1/s/lib/DxcSupport/FileIOHelper.cpp:1372:16: runtime error: null pointer passed as argument 2, which is declared to never be null
/usr/include/string.h:44:28: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/vsts/work/1/s/lib/DxcSupport/FileIOHelper.cpp:1372:16 

********************
```

This points to the fact that the sum ends up as 0, and is being treated as a nullptr.
I don't see a straightforward way to test this, as this is deep deep into the call stack, and deals with a private member variable of a stream object type. 
Also, I don't think setting pcbRead to 0 will solve the issue, since m_pMemory + m_offset still sums to 0, and the test is failing because the function specified that it should never receive a nullptr as the second argument. 
I'll reintroduce the original addition check, and return S_FALSE, though switching to an E_FAIL may be the right move later on. I can do some more investigating as to why this specific change serves as a unit test for this if condition.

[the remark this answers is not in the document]
bob80905: Further detail, here is the call stack:

```
 	hlsl::ReadOnlyBlobStream::Read(void * pv, unsigned long cb, unsigned long * pcbRead) Line 1373	C++
 	ReadAllBytes(IStream * pStream, void * pDst, unsigned __int64 uSize) Line 82	C++
 	PDBReader::ReadSuperblock(MSF_SuperBlock * pSB) Line 375	C++
 	PDBReader::PDBReader(IMalloc * pMalloc, IStream * pStream) Line 365	C++
 	hlsl::pdb::LoadDataFromStream(IMalloc * pMalloc, IStream * pIStream, IDxcBlob * * ppHash, IDxcBlob * * ppContainer) Line 494	C++
 	hlsl::pdb::LoadDataFromStream(IMalloc * pMalloc, IStream * pIStream, IDxcBlob * * ppContainer) Line 534	C++
 	DxilContainerReflection::Load(IDxcBlob * pContainer) Line 409	C++
 	CompileForHash(hlsl::options::DxcOpts & opts, const wchar_t * CommandFileName, dxc::DxcDllSupport & DllSupport, std::vector<wchar_t const *,std::allocator<wchar_t const *>> & flags, IDxcBlob * * ppHashBlob, std::string & output) Line 345	C++
 	FileRunCommandPart::RunDxcHashTest(dxc::DxcDllSupport & DllSupport) Line 410	C++
 	FileRunCommandPart::RunHashTests(dxc::DxcDllSupport & DllSupport) Line 64	C++
 	FileRunTestResultImpl::RunHashTestFromCommands(const char * commands, const wchar_t * fileName) Line 1280	C++
 	FileRunTestResultImpl::RunHashTestFromFileCommands(const wchar_t * fileName) Line 1375	C++
 	FileRunTestResult::RunHashTestFromFileCommands(const wchar_t * fileName) Line 1385	C++
>	CompilerTest::CodeGenTestHashFullPath(const wchar_t * fullPath) Line 403	C++
```

The `tools\clang\test\HLSLFileCheck\hlsl\objects\RayQuery\rayquery-trace-ray-inline-errors.hlsl` is causing the error. Seems like something related to not initializing the PDBReader object correctly, which should stem from the incorrect pStream initialiation on `IFR(hlsl::CreateReadOnlyBlobStream(pContainer, &pStream));`, located at
`D:\DXC\lib\HLSL\DxilContainerReflection.cpp::408`
Turns out it was initialized incorrectly because in `FileCheckerTest.cpp`, on line 329, `IFT(pCompiler2->CompileWithDebug(pSource, CommandFileName, entry.c_str(),
                                 profile.c_str(), flags.data(), flags.size(),
                                 nullptr, 0, pIncludeHandler, &pResult,
                                 &pDebugName, &pPDBBlob));`
Didn't produce any output (because the -verify flag was given, no output was produced by the compiler)
So, with no output, `pCompiledBlob` is empty / has no memory. This emptiness is the nullptr that gets passed along to the point where the nullptr is detected. 

TLDR: the -verify file `rayquery-trace-ray-inline-errors.hlsl` doesn't belong in the file-check directory, it needs to go somewhere where the CompilerTest doesn't check, not in a directory where all files are expected to have compiler output. Oops!

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +326,130 @@ class HLSLCallDiagnoseVisitor
...
+    // check if this declrefexpr refers to another one
+    // and drill into it if possible.
+    VarDecl *VD = dyn_cast<VarDecl>(DRE->getDecl());
+    if (!VD) {
+      return true;
+    }
+
+    ImplicitCastExpr *ICE = nullptr;
+    if (VD->getInit()) {
+      if ((ICE = dyn_cast<ImplicitCastExpr>(VD->getInit()))) {
```
tex3d, line 342: I do not think we want to manually try to unwrap things to find availability attributes.  This method won't scale well.

One thing we will need to do is collect all the global variables with initializers referenced by any `DeclRefExpr` visited here.  Then we will need to run `TraverseDecl` on each of these global variables.  It will probably be helpful to collect some path to the location of the first `DeclRefExpr` for a useful note if an error is encountered in that second pass.  By "path" I mean the location of the `DeclRefExpr` and the entry/export function leading to that usage.

I mocked up some changes to HLSLCallDiagnoseVisitor as a rough start for the high-level idea of how this visitor pattern could work.  `DiagnosedGlobalInit` allows skipping of shader model version diagnostics, just like `DiagnosedCalls` does, since the shader model version is the same for all entries we traverse.  `GlobalInitFound` tracks the global decls with initializers referenced by DeclRefExprs.  Since each set of `GlobalInitFound` is constructed for a particular entry or export function, the current `EntryDecl` already tracks the entry function part of the path to the use.  One drawback of the below approach is that it only captures one level of DeclRefExpr, so you could have a global using another global in an init expression, and the first init expression will be the DeclRefExpr location rather than a location in a function referring to the first global.  But this is a problem with these checks in general.  We may not even need to track the DeclRefExpr that used the global if the entry function is sufficient.  I also moved the `HLSLCallDiagnoseVisitor Vistor` declaration outside the visited function loop, since I didn't see any advantage to it being recreated at that frequency (I could have missed something).

Alternatively, we might just visit _all_ global init expressions, whether the global is used by an active entry or not.  This would be more in line with C++ global initialization.  We might need a discussion about the implications of one approach vs. the other.  One issue is that global initialization really is run under each entry function, so we might need to re-check each global init under the context (shader kind) of each entry function.

```diff
diff --git a/tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp b/tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
index cf5d74154..1b61949b2 100644
--- a/tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
+++ b/tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
@@ -18,6 +18,8 @@
 #include "clang/AST/RecursiveASTVisitor.h"
 #include "clang/Sema/SemaDiagnostic.h"
 #include "clang/Sema/SemaHLSL.h"
+#include "llvm/ADT/MapVector.h"
+#include "llvm/ADT/SmallPtrSet.h"
 #include "llvm/Support/Debug.h"
 #include <optional>

@@ -305,12 +307,18 @@ clang::FunctionDecl *ValidateNoRecursion(CallGraphWithRecurseGuard &callGraph,
 class HLSLCallDiagnoseVisitor
     : public RecursiveASTVisitor<HLSLCallDiagnoseVisitor> {
 public:
+  typedef llvm::SmallMapVector<VarDecl *, DeclRefExpr *, 8> GlobalInitMap;
+
   explicit HLSLCallDiagnoseVisitor(
       Sema *S, const hlsl::ShaderModel *SM, DXIL::ShaderKind EntrySK,
       DXIL::NodeLaunchType NodeLaunchTy, const FunctionDecl *EntryDecl,
-      llvm::SmallPtrSetImpl<CallExpr *> &DiagnosedCalls)
+      llvm::SmallPtrSetImpl<CallExpr *> &DiagnosedCalls,
+      GlobalInitMap &GlobalInitFound,
+      llvm::SmallPtrSetImpl<VarDecl *> &DiagnosedGlobalInit)
       : sema(S), SM(SM), EntrySK(EntrySK), NodeLaunchTy(NodeLaunchTy),
-        EntryDecl(EntryDecl), DiagnosedCalls(DiagnosedCalls) {}
+        EntryDecl(EntryDecl), DiagnosedCalls(DiagnosedCalls),
+        GlobalInitFound(GlobalInitFound),
+        DiagnosedGlobalInit(DiagnosedGlobalInit) {}

   bool VisitCallExpr(CallExpr *CE) {
     // Set flag if already diagnosed from another entry, allowing some
@@ -325,6 +333,8 @@ public:
     return true;
   }

+  void setDeclRefExpr(DeclRefExpr *Expr) { CurDeclRefExpr = Expr; }
+
   clang::Sema *getSema() { return sema; }

 private:
@@ -334,6 +344,11 @@ private:
   DXIL::NodeLaunchType NodeLaunchTy;
   const FunctionDecl *EntryDecl;
   llvm::SmallPtrSetImpl<CallExpr *> &DiagnosedCalls;
+  GlobalInitMap &GlobalInitFound;
+  // When traversing global decls, this tracks where an active use was found to
+  // lead from an active entry or export function.
+  DeclRefExpr *CurDeclRefExpr;
+  llvm::SmallPtrSetImpl<VarDecl *> &DiagnosedGlobalInit;
 };

 std::optional<uint32_t>
@@ -430,6 +445,7 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {

   std::set<FunctionDecl *> DiagnosedDecls;
   llvm::SmallPtrSet<CallExpr *, 16> DiagnosedCalls;
+  llvm::SmallPtrSet<VarDecl *, 8> DiagnosedGlobalInit;
   // for each FDecl, check for recursion
   for (FunctionDecl *FDecl : FDeclsToCheck) {
     CallGraphWithRecurseGuard callGraph;
@@ -539,10 +555,19 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
     }
     // Visit all visited functions in call graph to collect illegal intrinsic
     // calls.
+    HLSLCallDiagnoseVisitor::GlobalInitMap GlobalInitFound;
+    HLSLCallDiagnoseVisitor Visitor(self, shaderModel, EntrySK, NodeLaunchTy,
+                                    FDecl, DiagnosedCalls, GlobalInitFound,
+                                    DiagnosedGlobalInit);
     for (FunctionDecl *FD : callGraph.GetVisitedFunctions()) {
-      HLSLCallDiagnoseVisitor Visitor(self, shaderModel, EntrySK, NodeLaunchTy,
-                                      FDecl, DiagnosedCalls);
       Visitor.TraverseDecl(FD);
     }
+    // Initializers might refer to other globals, adding those to
+    // GlobalInitFound, so this iteration method accounts for this.
+    for (unsigned Idx = 0; Idx < GlobalInitFound.size(); ++Idx) {
+      auto Found = GlobalInitFound.begin()[Idx];
+      Visitor.setDeclRefExpr(Found.second);
+      Visitor.TraverseDecl(Found.first);
+    }
   }
 }

```

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +326,130 @@ class HLSLCallDiagnoseVisitor
...
+            (uint64_t)DXIL::RayFlag::ForceOMM2State) != 0;
+  }
+
+  void DiagnoseTraceRayInline(const MemberExpr *ME,
+                              CXXMemberCallExpr *callExpr) {
+    // Validate if the RayFlag parameter has RAY_FLAG_FORCE_OMM_2_STATE set,
+    // the RayQuery decl must have RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set,
+    // otherwise emit a diagnostic.
+    bool IsRayFlagForceOMM2State = IsRayFlagForceOMM2StateSet(callExpr);
+    if (IsRayFlagForceOMM2State) {
```
tex3d, line 400: No point keeping a variable around if it's only used in the if expression.
```suggestion
    if (IsRayFlagForceOMM2StateSet(callExpr)) {
```

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +326,130 @@ class HLSLCallDiagnoseVisitor
...
+
+    if (const CXXMethodDecl *methodDecl = dyn_cast<CXXMethodDecl>(MD)) {
+      auto *attr = methodDecl->getAttr<HLSLIntrinsicAttr>();
+      if (!attr)
+        return true;
+      StringRef OpcodeGroup = GetHLOpcodeGroupName(HLOpcodeGroup::HLIntrinsic);
+      if (attr->getGroup() != OpcodeGroup)
+        return true;
+      if ((hlsl::IntrinsicOp)attr->getOpcode() ==
+          hlsl::IntrinsicOp::MOP_TraceRayInline)
```
tex3d, line 447: Actually, this should simply be added as a case to `DiagnoseReachableHLSLCall`.  Member calls should already reach that function.  I don't think we need to override `VisitCXXMemberCallExpr` at all in that case.

tex3d (commented):
Major points:
- we should only report on incompatible flags in `CheckTemplateArgumentListForHLSL`
- we need to leverage the recursive visitor, looping in global initializers, to enforce availability attributes in a general way
- We may use `HLSLCallDiagnoseVisitor` to diagnose reachable RayQuery VarDecl (add `VisitVarDecl`?) with non-zero RayQueryFlags for `warn_hlsl_rayquery_flags_disallowed`
- Should add a case to `DiagnoseReachableHLSLCall`, instead of adding `VisitCXXMemberCallExpr`

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7652,8 +7652,15 @@ def err_payload_fields_is_payload_and_overqualified : Error<
   "payload field '%0' is a payload struct. Payload access qualifiers are not allowed on payload types.">;
 def warn_hlsl_payload_qualifer_dropped : Warning<
   "payload access qualifiers ignored. These are only supported for lib_6_7+ targets and lib_6_6 with with the -enable-payload-qualifiers flag.">, InGroup<HLSLPayloadAccessQualifer>;
+def warn_hlsl_rayquery_flags_disallowed : Warning<
+  "A non-zero value for the RayQueryFlags template argument requires shader model 6.9 or above.">, DefaultError, InGroup<HLSLAvailabilityConstant>;
+def warn_hlsl_rayquery_flags_conflict  : Warning<
+  "When using 'RAY_FLAG_FORCE_OMM_2_STATE' in RayFlags, RayQueryFlags must have RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set.">, DefaultError, InGroup<HLSLAvailabilityConstant>;
```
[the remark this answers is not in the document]
bob80905: I'll remove it from any group, but to keep with the spec, I'll leave it as a warning.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
bob80905: Thanks this was mega useful!

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -570,6 +573,7 @@ static void AddConstUInt(clang::ASTContext &context, StringRef name,
 struct Enumerant {
   StringRef name;
   unsigned value;
+  AvailabilityAttr *avail = nullptr;
```
tex3d, line 576: To avoid potential leaks on failure, maybe we should make this a `unique_ptr`?

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -599,10 +609,8 @@ void hlsl::AddRaytracingConstants(ASTContext &context) {
 
   // Create an availability attribute for
   // shader model 6.9 for the RAYQUERY_FLAG enum
-  AvailabilityAttr *AAttr_6_9 = AvailabilityAttr::CreateImplicit(
-      context, &context.Idents.get(""), clang::VersionTuple(6, 9),
-      clang::VersionTuple(), clang::VersionTuple(), false,
-      "potential misuse of built-in constant introduced in shader model 6.9");
+  VersionTuple VT69 = VersionTuple(6,9);
+  ConstructAvailabilityAttribute(context, VT69);
```
tex3d, line 612: Don't want to construct an attribute here and throw away the pointer.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +359,89 @@ class HLSLCallDiagnoseVisitor
...
+        }
+      }
+    }
+    return true;
+  }
+
+  bool VisitDeclRefExpr(DeclRefExpr *DRE) {
+    // Diagnose availability for referenced decl.
+    // Skip redundant availability diagnostics for the same Decl.
+    if (DiagnosedDecls.insert(DRE).second)
```
tex3d, line 407: The code between `VisitDeclRefExpr`, `GetAttrAndStoreDecl`, and `DiagnoseAvailability` looks a bit confused.
I think the intent was to put this `if (DiagnosedDecls.insert(DRE).second)` check inside `GetAttrAndStoreDecl`, then you don't need to condition the call `DiagnoseAvailability` here.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +359,89 @@ class HLSLCallDiagnoseVisitor
...
+    if (!ND)
+      return nullptr;
+    AvailabilityAttr *AAttr = ND->getAttr<AvailabilityAttr>();
+    if (!AAttr)
+      return nullptr;
+
+    return AAttr;
+  }
+
+  void DiagnoseAvailability(DeclRefExpr *DRE, SourceLocation Loc) {
```
tex3d, line 424: Don't need `SourceLocation Loc` parameter because now it's just `DRE->getLocation()`.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -325,6 +359,89 @@ class HLSLCallDiagnoseVisitor
...
+  bool VisitDeclRefExpr(DeclRefExpr *DRE) {
+    // Diagnose availability for referenced decl.
+    // Skip redundant availability diagnostics for the same Decl.
+    if (DiagnosedDecls.insert(DRE).second)
+      DiagnoseAvailability(DRE, DRE->getLocation());
+
+    return true;
+  }
+
+  AvailabilityAttr *GetAttrAndStoreDecl(DeclRefExpr *DRE) {
```
tex3d, line 413: I think I'd prefer a name like `GetAvailabilityAttrOnce`.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -334,6 +451,7 @@ class HLSLCallDiagnoseVisitor
   DXIL::NodeLaunchType NodeLaunchTy;
   const FunctionDecl *EntryDecl;
   llvm::SmallPtrSetImpl<CallExpr *> &DiagnosedCalls;
+  llvm::SmallPtrSetImpl<DeclRefExpr *> &DiagnosedDecls;
```
tex3d, line 454: I think `DiagnosedDecls` isn't exactly appropriate anymore.  Maybe `DeclAvailabilityChecked`?

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -482,14 +608,14 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
       // disconnected with respect to the call graph.
       // Only check this if neither function decl is recursive
       if (!result && !patchResult) {
-        CallGraphWithRecurseGuard CG;
-        CG.BuildForEntry(pPatchFnDecl);
+        CallGraphWithRecurseGuard CG; // Why not use already-built 'callGraph'?
```
tex3d, line 611: Forgot to replace `CG` with use of existing `callGraph`.

tools/clang/test/SemaHLSL/rayquery-decls-errors-TU-multiple.hlsl
```diff
@@ -0,0 +1,42 @@
+// RUN: %dxc -T vs_6_5 -E main -verify %s
+
+// tests that RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS usage will emit
+// exactly one warning when multiple incompatible 
+// availability attribute decls appear, and the compilation target
+// is less than shader model 6.9.
```
tex3d, line 6: Not anymore.

tools/clang/test/SemaHLSL/rayquery-decls-errors-TU-export.hlsl
```diff
@@ -0,0 +1,6 @@
+// RUN: %dxc -T lib_6_5 -verify %s
+
+export float4 MyExportedFunction(float4 color) {
```
tex3d, line 3: Should have the globals in this file as well to verify that they are not picked up by the export function for diagnostics.

tools/clang/test/SemaHLSL/rayquery-trace-ray-inline-errors.hlsl
```diff
@@ -0,0 +1,20 @@
...
+// Test that at the call site of any TraceRayInline call, a default error
+// warning is emitted that indicates the ray query object has the
+// RAY_FLAG_FORCE_OMM_2_STATE set, but doesn't have 
+// RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set
+
+RaytracingAccelerationStructure RTAS;
+void main(RayDesc rayDesc : RAYDESC) : OUT {
+  // expected-note@+2{{RayQueryFlags declared here}}
+  // expected-note@+1{{RayQueryFlags declared here}}
+  RayQuery<0> rayQuery; // implicitly, the second arg is 0.
```
tex3d, line 12: Need another that shows a use of non-zero RayQueryFlags requires SM 6.9, without use of the `RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS` identifier:
```hlsl
  RayQuery<0, 1> rayQueryInvalid;
```

tools/clang/test/SemaHLSL/rayquery-trace-ray-inline-errors.hlsl
```diff
@@ -0,0 +1,23 @@
...
+
+// Test that at the call site of any TraceRayInline call, a default error
+// warning is emitted that indicates the ray query object has the
+// RAY_FLAG_FORCE_OMM_2_STATE set, but doesn't have 
+// RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set
+
+RaytracingAccelerationStructure RTAS;
+void main(RayDesc rayDesc : RAYDESC) : OUT {
+  // expected-note@+2{{RayQueryFlags declared here}}
+  // expected-note@+1{{RayQueryFlags declared here}}
```
tex3d, line 11: These can be combined like:
```suggestion
  // expected-note@+1 2 {{RayQueryFlags declared here}}
```

tools/clang/test/SemaHLSL/rayquery-decls-errors-TU-multiple.hlsl
```diff
@@ -0,0 +1,41 @@
...
+  // expected-warning@+1{{potential misuse of built-in constant 'RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS' in shader model vs_6_5; introduced in shader model 6.9}}
+  static const int badVar = RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS;
+}
+
+// expected-warning@+1{{potential misuse of built-in constant 'RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS' in shader model vs_6_5; introduced in shader model 6.9}}
+groupshared const int otherBadVar = RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS;
+
+int retNum(){
+  // expected-warning@+1{{potential misuse of built-in constant 'RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS' in shader model vs_6_5; introduced in shader model 6.9}}
+  return RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS;
```
tex3d, line 17: Maybe one more copy of this function that isn't called to verify that it doesn't have a diagnostic?

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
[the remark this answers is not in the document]
tex3d: If you're not giving it a group at all, then a `DefaultError` warning might as well be an error, since that turns it into an error with no way to override or silence it.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7652,8 +7652,17 @@ def err_payload_fields_is_payload_and_overqualified : Error<
   "payload field '%0' is a payload struct. Payload access qualifiers are not allowed on payload types.">;
 def warn_hlsl_payload_qualifer_dropped : Warning<
   "payload access qualifiers ignored. These are only supported for lib_6_7+ targets and lib_6_6 with with the -enable-payload-qualifiers flag.">, InGroup<HLSLPayloadAccessQualifer>;
+def warn_hlsl_rayquery_flags_disallowed : Warning<
+  "A non-zero value for the RayQueryFlags template argument requires"
+  " shader model 6.9 or above.">, DefaultError, InGroup<HLSLAvailability>;
+def warn_hlsl_rayquery_flags_conflict : Warning<
+  "When using 'RAY_FLAG_FORCE_OMM_2_STATE' in RayFlags, RayQueryFlags"
+  " must have RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set.">, DefaultError;
```
tex3d, line 7660: I don't think there's any point in keeping this a warning with `DefaultError` and no group, since it will always be turned into an error, with no way to override the behavior.  I think we need to either give it a group, or turn it into an error.

tools/clang/test/SemaHLSL/rayquery-omm-type-diag.hlsl
```diff
@@ -0,0 +1,12 @@
+// RUN: %dxc -T vs_6_9 -E RayQueryTests -verify %s
+// RUN: %dxc -T vs_6_5 -E RayQueryTests2 -verify %s
+
+// validate 2nd template argument flags
+// expected-error@+1{{When using 'RAY_FLAG_FORCE_OMM_2_STATE' in RayFlags, RayQueryFlags must have RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set.}}
+typedef RayQuery<RAY_FLAG_FORCE_OMM_2_STATE> BadRayQuery;
+// expected-error@+1{{When using 'RAY_FLAG_FORCE_OMM_2_STATE' in RayFlags, RayQueryFlags must have RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS set.}}
+typedef RayQuery<RAY_FLAG_FORCE_OMM_2_STATE, 0> BadRayQuery2;
+
+BadRayQuery rayQuery0a;
```
tex3d, line 10: Oh, we should probably have prevented declaring RayQuery variables as constant buffer values, that's a bug that there's no diagnostic here.  Maybe for this change you can add "static" and we can file an issue for that bug?

bob80905: Created an issue here:
https://github.com/microsoft/DirectXShaderCompiler/issues/7202

tex3d: Filed issue: #7203.

tex3d (approved):

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -482,15 +605,14 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
       // disconnected with respect to the call graph.
       // Only check this if neither function decl is recursive
       if (!result && !patchResult) {
-        CallGraphWithRecurseGuard CG;
-        CG.BuildForEntry(pPatchFnDecl);
-        if (CG.CheckReachability(pPatchFnDecl, FDecl)) {
+        callGraph.BuildForEntry(pPatchFnDecl, GlobalsWithInit);
```
tex3d, line 608: Don't need to call BuildForEntry again here.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -482,15 +605,14 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
+        if (callGraph.CheckReachability(pPatchFnDecl, FDecl)) {
           self->Diag(FDecl->getSourceRange().getBegin(),
                      diag::err_hlsl_patch_reachability_not_allowed)
               << 1 << FDecl->getName() << 0 << pPatchFnDecl->getName();
         }
-        CG.BuildForEntry(FDecl);
-        if (CG.CheckReachability(FDecl, pPatchFnDecl)) {
+        callGraph.BuildForEntry(FDecl, GlobalsWithInit);
```
tex3d, line 614: Don't need to call BuildForEntry again here.

tex3d (approved):

PR #7168, opened 2025-02-28
Make unsigned dot test more robust

pow2clk:
When built without spirv, the CGL tests were offset by 5 because of the missing opcodes. This captures the preliminary opcode and confirms that it matches everywhere it should.


damyanp (approved):

bob80905 (approved):

tex3d (approved):
LGTM!

PR #7191, opened 2025-03-11
Resolve some default error warnings

bob80905:
DxilContainerValidation.cpp has some int / bool comparisons that cause default error warnings in some of the private builds.
This needs to be addressed. This PR changes the comparisons by converting the numerical expressions into the appropriate boolean, then comparing the booleans.

### DIFF
--- a/lib/DxilValidation/DxilContainerValidation.cpp
+++ b/lib/DxilValidation/DxilContainerValidation.cpp
@@ -337,9 +337,12 @@ void PSVContentVerifier::VerifySignatureElement(
 
   PSVSignatureElement PSVSE(StrTab, IndexTab, PSVSE0);
   if (SE.IsArbitrary())
-    Mismatch |= strcmp(PSVSE.GetSemanticName(), SE.GetName());
+    Mismatch |=
+        strcmp(PSVSE.GetSemanticName(), SE.GetName()) == 0 ? false : true;
   else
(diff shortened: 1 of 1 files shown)


lizhengxing (approved):

lib/DxilValidation/DxilContainerValidation.cpp
```diff
@@ -337,9 +337,12 @@ void PSVContentVerifier::VerifySignatureElement(
 
   PSVSignatureElement PSVSE(StrTab, IndexTab, PSVSE0);
   if (SE.IsArbitrary())
-    Mismatch |= strcmp(PSVSE.GetSemanticName(), SE.GetName());
+    Mismatch |=
+        strcmp(PSVSE.GetSemanticName(), SE.GetName()) == 0 ? false : true;
   else
-    Mismatch |= PSVSE0->SemanticKind != static_cast<uint8_t>(SE.GetKind());
+    Mismatch |= PSVSE0->SemanticKind != static_cast<uint8_t>(SE.GetKind()) == 0
```
tex3d, line 343: This change was unnecessary, plus it scares me a bit how we would be dependent on a left to right order of evaluating the comparison operations without parentheses to enforce the correct order (and the final comparison is basically `(bool) == (int)`).

Also, we could use `!= 0` everywhere instead of `== 0 ? false : true`.  That would be keeping with most of the other similar code elsewhere in this function, and it's easier to read.

PR #7199, opened 2025-03-13
NFC: Update HLSL_INTRINSIC struct for Flags and MinShaderModel fields

tex3d:
HLSL_INTRINSIC will need to be updated for SM 6.9, specifically:
- to add a new flag
- to encode minimum shader model version for an availability attribute

Changing this structure is a breaking change to the internal intrinsic table protocol, which is used for the extension mechanism.

This change separates out the breaking change with no functional changes for simpler review and testing.

For the new flag, this change switches to using a UINT Flags field to make flags extensible without breaking the table format. For the version, a UINT MinShaderModel will be the encoded version format used elsewhere: (Major << 4) | (Minor & 0xF)

Commented code for using the MinShaderModel is provided for when a subsequent change will implement the availability attribute checks.

### DIFF
--- a/include/dxc/dxcapi.internal.h
+++ b/include/dxc/dxcapi.internal.h
@@ -161,9 +161,9 @@ struct HLSL_INTRINSIC_ARGUMENT {
 };
 
 // HLSL_INTRINSIC flags
-static const UINT INTRIN_FLAG_READ_ONLY = 0x1;
-static const UINT INTRIN_FLAG_READ_NONE = 0x2;
-static const UINT INTRIN_FLAG_IS_WAVE = 0x4;
+static const UINT INTRIN_FLAG_READ_ONLY = 1U << 0;
+static const UINT INTRIN_FLAG_READ_NONE = 1U << 1;
+static const UINT INTRIN_FLAG_IS_WAVE = 1U << 2;
 
 struct HLSL_INTRINSIC {
   UINT Op;                 // Intrinsic Op ID
--- a/include/dxc/dxcapi.internal.h
+++ b/include/dxc/dxcapi.internal.h
@@ -160,11 +160,16 @@ struct HLSL_INTRINSIC_ARGUMENT {
               // matching input constraints.
 };
 
+// HLSL_INTRINSIC flags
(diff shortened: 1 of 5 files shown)


tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -1810,12 +1810,21 @@ static void AddHLSLIntrinsicAttr(FunctionDecl *FD, ASTContext &context,
   }
   FD->addAttr(
       HLSLIntrinsicAttr::CreateImplicit(context, tableName, lowering, opcode));
-  if (pIntrinsic->bReadNone)
+  if (pIntrinsic->Flags & INTRIN_FLAG_READ_ONLY)
     FD->addAttr(ConstAttr::CreateImplicit(context));
-  if (pIntrinsic->bReadOnly)
+  if (pIntrinsic->Flags & INTRIN_FLAG_READ_NONE)
     FD->addAttr(PureAttr::CreateImplicit(context));
```
damyanp, line 1816: I think you may have got these two the wrong way round?

damyanp (approved):
LGTM (assuming the automated tests are all green - looks like the python formatting check at least isn't happy)

include/dxc/dxcapi.internal.h
```diff
@@ -160,11 +160,16 @@ struct HLSL_INTRINSIC_ARGUMENT {
               // matching input constraints.
 };
 
+// HLSL_INTRINSIC flags
+static const UINT INTRIN_FLAG_READ_ONLY = 0x1;
+static const UINT INTRIN_FLAG_READ_NONE = 0x2;
+static const UINT INTRIN_FLAG_IS_WAVE = 0x4;
```
bogner, line 166: Personal preference really, but I find defining flags using the bitshift to be clearer once you get to the larger numbers:
```suggestion
static const UINT INTRIN_FLAG_READ_ONLY = 1U << 0;
static const UINT INTRIN_FLAG_READ_NONE = 1U << 1;
static const UINT INTRIN_FLAG_IS_WAVE = 1U << 2;
```

bogner (approved):

adam-yang (approved):

PR #7205, opened 2025-03-14
Lower RayQuery constructor to allocateRayQuery2

bob80905:
This PR connects the front end change to the back end change, by taking the existing rayquery constructor translation, and augmenting it so that allocaterayquery2 can be emitted as an opcode if there are 2 template arguments. 
It is independent of the shader model. If 2 template args are detected, and the 2nd template argument has a non-zero value, it just emits allocateRayQuery2.
A test was added to make sure that when targeting shader model 6.9, using 2 template args where the 2nd arg is non-zero in a rayquery declaration will produce an allocateRayQuery2 opcode.

Fixes [#7136](https://github.com/microsoft/DirectXShaderCompiler/issues/7136)


include/dxc/HlslIntrinsicOp.h
```diff
@@ -10,6 +10,7 @@ enum class IntrinsicOp {
   IOP_AllMemoryBarrier,
   IOP_AllMemoryBarrierWithGroupSync,
   IOP_AllocateRayQuery,
+  IOP_AllocateRayQuery2,
```
damyanp, line 13: Maybe silly question, but I'll learn something from it.  Does the numeric values of these enum entries matter? Presumably this change increments all the values after this one.

damyanp (approved):
LGTM - I have a question for my own education.

I did spent a little while trying to see if there were any more test cases that should be added, but nothing stood out to me.

include/dxc/HlslIntrinsicOp.h
bob80905: I don't think so. I think the important thing is that the order is consistent.
Note that IOP_AllocateRayQuery and IOP_Barrier come before and after, and that in HLOperationLower.cpp, the entries in gLowerTable go:
```
{IntrinsicOp::IOP_AllocateRayQuery, TranslateAllocateRayQuery,
     DXIL::OpCode::AllocateRayQuery},
    {IntrinsicOp::IOP_AllocateRayQuery2, TranslateAllocateRayQuery2,
     DXIL::OpCode::AllocateRayQuery2},
    {IntrinsicOp::IOP_Barrier, TranslateBarrier, DXIL::OpCode::NumOpCodes}
```
So, allocaterayquery2 go between these two opcodes. This won't break the access to gLowerTable at HLOperationLower.cpp:6829, where the translation function is fetched.

If, for example, I were to place the IOP anywhere else in the enum, then I would fetch the wrong translation function when accessing gLowerTable, and things would be messed up. So the value doesn't matter in and of itself, but the ordering between the enum and the gLowerTable array need to be in sync.

tex3d: The high-level operations in HlslIntrinsicOp.h are not version stable.  One or more may change whenever we add a high level op (new HLSL intrinsic).  They are only supposed to be used internally.

The enum value placement is automatic - generated from the hctdb scripts.  The table in HLOperationLower is manually updated, but the array indices must match the operations in this enum.

Important Note: this fact makes writing version-stable `dxilgen`-pass-only IR tests impossible.  In other words, those tests can break randomly when new operations are added, and you would need to go back and remap the high level operation numbers in the first argument of every HL call in any tests based on the diff of this enum.

Come to think of it, this is probably the main the reason we don't have any of those for any of the prior features.  We do have one test referencing the high-level `abs` intrinsic IIRC, but that one breaks now and again and has to be updated.

utils/hct/gen_intrin_main.txt
```diff
@@ -364,6 +364,9 @@ bool [[ro]] IsHelperLane();
 // HL Op for allocating ray query object that default constructor uses
 uint [[hidden]] AllocateRayQuery(in uint flags);
 
+// HL Op for allocating ray query object when non-zero 2nd template argument is given
+uint [[hidden]] AllocateRayQuery2(in uint flags, in uint rayqueryflags);
```
tex3d, line 368: I actually think we could have gotten away with using the same hidden intrinsic, with an argument update.  Then, the high level op is always the same, with two arguments, and the lowering just lowers to the prior DXIL op when the second arg is zero.

include/dxc/HlslIntrinsicOp.h
tex3d: Actually, this is the reason for the pipeline failures.  The new vector DXIL lowering tests no longer have the correct HL opcode.

Strangely, the one we used to always run into (`abs`) is absent, so I wonder if we no longer run that test.  Oh wait, now it's coming back to me: that might have been a `-fcgl` test where we replaced the explicit HL opcode with a regex to stop it from failing and having to update the expected text.  We can't do that sort of thing for a lowering test, since the opcode must be explicitly provided in the test IR.

bob80905: Yes, the relevant change is here:
https://github.com/microsoft/DirectXShaderCompiler/pull/7168/files

tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
```diff
@@ -2824,7 +2834,25 @@ void TranslateRayQueryConstructor(HLModule &HLM) {
         llvm::ConstantInt::get(i32Ty, (uint64_t)0, false);
     llvm::FunctionType *funcTy =
         llvm::FunctionType::get(i32Ty, {i32Ty, i32Ty}, false);
+
     unsigned opcode = (unsigned)IntrinsicOp::IOP_AllocateRayQuery;
+    // reset opcode to IOP_AllocateRayQuery2 if there are 2 template args
```
tex3d, line 2839: There is no IOP_AllocateRayQuery2 now.  Also, I believe this would be simpler if we always generated a single signature for the HL op, supplying both flag arguments, whether or not RayQueryFlags are provided in the template arguments or equal to zero.

Then, HLOperationLower can choose the DXIL op based on whether RayQueryFlags are zero at that point.

tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
```diff
@@ -2795,10 +2795,12 @@ unsigned AlignBufferOffsetInLegacy(unsigned offset, unsigned size,
 }
 
 // Translate RayQuery constructor.  From:
-//  %call = call %"RayQuery<flags>" @<constructor>(%"RayQuery<flags>" %ptr)
+//  %call = call %"RayQuery<flags, constrayqueryflags<optional rayquery flags>>"
+//  @<constructor>(%"RayQuery<flags>" %ptr)
 // To:
-//  i32 %handle = AllocateRayQuery(i32 <IntrinsicOp::IOP_AllocateRayQuery>, i32
-//  %flags) %gep = GEP %"RayQuery<flags>" %ptr, 0, 0 store i32* %gep, i32
+//  i32 %handle = AllocateRayQuery2(i32 <IntrinsicOp::IOP_AllocateRayQuery>, i32
```
tex3d, line 2801: I think we need a `-fcgl` test to verify this translation.  Be sure to make the HL OpCode argument value a regex in the CHECK to avoid test fragility.

tools/clang/lib/CodeGen/CGHLSLMSFinishCodeGen.cpp
```diff
@@ -2848,8 +2854,13 @@ void TranslateRayQueryConstructor(HLModule &HLM) {
       llvm::IRBuilder<> Builder(CI);
       llvm::Value *rayFlags =
           Builder.getInt32(SA->GetTemplateArgAnnotation(0).GetIntegral());
-      llvm::Value *Call =
-          Builder.CreateCall(opFunc, {opVal, rayFlags}, pThis->getName());
+      // this could be a nullptr assignment if there is no 2nd template arg
```
tex3d, line 2857: Is this possible with the current template definition?  Or will the default argument always be here?

tex3d: Comments and code above appear to assume there could be 1 or 2 template arguments, but this code would crash if that's the case, right?  It seems like some adjustment is still necessary for the code and comments to be fully consistent with what's possible at this point.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -5670,7 +5670,20 @@ Value *TranslateAllocateRayQuery(CallInst *CI, IntrinsicOp IOP,
                                  HLObjectOperationLowerHelper *pObjHelper,
                                  bool &Translated) {
   hlsl::OP *hlslOP = &helper.hlslOP;
-  Value *refArgs[] = {nullptr, CI->getOperand(1)};
+  // upgrade to allocateRayQuery2 if there is a non-zero 2nd template arg
```
tex3d, line 5673: In this location, you might say if RayQueryFlags are non-zero, then add indices to `HLOperations.h` and use them, like:
```cpp
// AllocateRayQuery
const unsigned kAllocateRayQueryRayFlagsIdx = 1;
const unsigned kAllocateRayQueryRayQueryFlagsIdx = 2;
```

lib/HLSL/HLOperationLower.cpp
```diff
@@ -5670,7 +5670,20 @@ Value *TranslateAllocateRayQuery(CallInst *CI, IntrinsicOp IOP,
...
+  DXASSERT(CI->getNumArgOperands() == 3,
+           "hlopcode for allocaterayquery always expects 3 arguments");
+
+  llvm::Value *Arg = CI->getOperand(2);
```
tex3d, line 5677: You would use `HLOperandIndex::kAllocateRayQueryRayQueryFlagsIdx` here.  Also, I think you want `getArgOperand`, don't you?
```suggestion
  llvm::Value *Arg =
      CI->getArgOperand(HLOperandIndex::kAllocateRayQueryRayQueryFlagsIdx);
```

tex3d: I guess they return the same operand, but we generally use `getArgOperand` when getting call args in this code.

tools/clang/test/CodeGenHLSL/allocateRayQuery2.hlsl
```diff
@@ -0,0 +1,16 @@
+// RUN: %dxc -T lib_6_9 %s | FileCheck %s 
+
+RaytracingAccelerationStructure RTAS;
+// DXR entry point to ensure RDAT flags match during validation.
+[shader("vertex")]
+void main(RayDesc rayDesc : RAYDESC) {
+
+  // CHECK: call i32 @dx.op.allocateRayQuery2(i32 258, i32 1024, i32 1)  ; AllocateRayQuery2(constRayFlags,constRayQueryFlags)
```
tex3d, line 8: nit: I don't think we need to keep the comment added by our disassembler here:
```suggestion
  // CHECK: call i32 @dx.op.allocateRayQuery2(i32 258, i32 1024, i32 1)
```

tools/clang/test/CodeGenHLSL/allocateRayQuery2.hlsl
```diff
@@ -0,0 +1,16 @@
...
+  RayQuery<RAY_FLAG_FORCE_OMM_2_STATE, RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS> rayQuery1;
+
+  rayQuery1.TraceRayInline(RTAS, RAY_FLAG_FORCE_OMM_2_STATE, 2, rayDesc);
+
+  // CHECK: call i32 @dx.op.allocateRayQuery(i32 178, i32 1)  ; AllocateRayQuery(constRayFlags)
```
tex3d, line 13: Or here:
```suggestion
  // CHECK: call i32 @dx.op.allocateRayQuery(i32 178, i32 1)
```

tools/clang/test/CodeGenHLSL/allocateRayQuery2.hlsl
```diff
@@ -0,0 +1,15 @@
+// RUN: %dxc -T lib_6_9 %s | FileCheck %s
```
tex3d, line 1: I think we just need another RUN line with `-fcgl` and additional checks to verify the initial RayQuery constructor translation in `CGHLSLMSFinishCodeGen.cpp`.

tex3d (approved):

tools/clang/test/DXC/Passes/DxilGenerationPass/LowerAllocateRayQuery2.ll
tex3d, line 1: It looks like there are other PRs that are using a slightly different path for this with `DxilGen` instead of `DxilGenerationPass` as the subfolder.  If you have a chance before merging, maybe you can change this to match, which will be consistent with other incoming changes.

PR #7221, opened 2025-03-18
Always enable SPIR-V

### LINKED ISSUE #7169: [Feature Request] Always enable SPIR-V
The time when SPIR-V was an ancillary part of the compiler is past. We have enabled SPIR-V with our official releases for over a year. We have announced our intent to switch to SPIR-V. 

The ability to disable SPIR-V gives us a different build configuration that we don't test in automation, which can result in committing changes that disrupt developers who haven't enabled SPIR-V.

Remove the ability to disable SPIR-V. This will simplify a lot of configurations and give us a consistent build target. Additionally, those who disable it locally won't find breakages only when they upload the code.

We could add a spirv-disabled build to automation.

pow2clk:
disables the ability to disable SPIR-V. This allows intermediate opcodes to be consistent at least across the same compiler build regardless of build environment. This is a requirement for IR tests that depend on these opcodes.

Given the embrace of SPIR-V and the stability the generation target has gained, there's no reason not to do this.

Fixes #7169


pow2clk:
Note that I've enabled this by force rather than just enabling it by default to address existing installs. This does mean that some will have to get the SPIRV subdirs to get what they need, but the alternative is seemingly failing IR tests

pow2clk:
format checker is refusing to find he latest hash. null change incoming

pow2clk:
Well, there seems to be no satisfying the formatting bot for the present.

farzonl (approved):

damyanp (approved):
LGTM - would it also be good to get a review from some of the developers working on SPIR-V as well?

Is there any documentation in the repo that needs updating?  READMEs etc?

pow2clk:
I assure you this was clang-formatted. The bot is having a bad week. Bypassing.

pow2clk:
> LGTM - would it also be good to get a review from some of the developers working on SPIR-V as well?
> 
> Is there any documentation in the repo that needs updating? READMEs etc?

I didn't see this comment before my last comment for security reasons. These are good suggestions.

pow2clk:
Given that we have another way to solve the problem of different configurations with the same build failing IR test due to instable opcode values with #7231, suddenly requiring the SPIRV submodules upon syncing to the latest is no longer the lesser evil as tests will no longer suddenly fail on SPIRV disabled builds. As such, we are still determined to change the default to enabling SPIRV, but it will be just a few lines changing defaults in cmake and batch files. 

This change doesn't bear sufficient resemblance to that eventual simple, though impactful change, so I'm closing it.

PR #7222, opened 2025-03-18
Update clang-format-checker.yml

alsepkow:
The github action for clang-format-checker started failing with "Error: Similar commit hashes detected: previous sha: 3ddf29bd4384cd2b81a6b04c71ca9e8f3160714f is equivalent to the current sha: 3ddf29bd4384cd2b81a6b04c71ca9e8f3160714f."

### DIFF
--- a/.github/workflows/clang-format-checker.yml
+++ b/.github/workflows/clang-format-checker.yml
@@ -15,14 +15,15 @@ jobs:
       - name: Fetch LLVM sources
         uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          fetch-depth: 0 # Fetch all history
 
       - name: Get changed files
         id: changed-files
         uses: tj-actions/changed-files@v41
         with:
           separator: ","
-          fetch_depth: 100 # Fetches only the last 10 commits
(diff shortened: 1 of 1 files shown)


.github/workflows/clang-format-checker.yml
```diff
@@ -15,14 +15,16 @@ jobs:
...
         uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          fetch-depth: 0 # Fetch all history
 
       - name: Get changed files
         id: changed-files
-        uses: step-security/changed-files@3dbe17c78367e7d60f00d78ae6781a35be47b4a1 # v45.0.1
+        uses: tj-actions/changed-files@v41
```
llvm-beanz, line 23: This change was deliberate in https://github.com/microsoft/DirectXShaderCompiler/pull/7217 because the tj-actions version of the action had a significant security vulnerability.

I'm not sure what the correct fix is, but we should not revert back to the tj-actions version.

alsepkow: Definitely was not intending to check this in. Initially didn't think everyone would see the PR if it was marked as a draft.

PR #7235, opened 2025-03-20
Use Wide String variants explicitly for Windows API calls

bob80905:
This PR changes some code in ExecutionTests.cpp to use the wide string variants of Windows API calls explicitly. This is because some internal builds will get confused about which overload to resolve the GetModuleHandle function to. By being explicit, this should eliminate the error that an arg can't be converted to LPCWSTR.

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -820,10 +820,10 @@ public:
         return false;
       }
 
-      if (GetModuleHandle("d3d10warp.dll") != NULL) {
-        CHAR szFullModuleFilePath[MAX_PATH] = "";
-        GetModuleFileName(GetModuleHandle("d3d10warp.dll"),
-                          szFullModuleFilePath, sizeof(szFullModuleFilePath));
+      if (GetModuleHandleW(L"d3d10warp.dll") != NULL) {
+        WCHAR szFullModuleFilePath[MAX_PATH] = L"";
+        GetModuleFileNameW(GetModuleHandleW(L"d3d10warp.dll"),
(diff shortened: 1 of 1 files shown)


tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -820,10 +820,10 @@ class ExecutionTest {
         return false;
       }
 
-      if (GetModuleHandle("d3d10warp.dll") != NULL) {
-        CHAR szFullModuleFilePath[MAX_PATH] = "";
-        GetModuleFileName(GetModuleHandle("d3d10warp.dll"),
-                          szFullModuleFilePath, sizeof(szFullModuleFilePath));
+      if (GetModuleHandleW(L"d3d10warp.dll") != NULL) {
```
farzonl, line 823: This change looks good. However, I've seen projects where the use of the W api function vs the char one was controlled via a compile time flag. Would there be any reason to do that here? 

This is just a test case so i'm assuming no. But was curious what if we wanted to cross compile a windows build on linux? what if we wanted to use wine  to run test cases or some other windows emulation layer? Do WCHARs preclude us from this?

farzonl (approved):

damyanp (approved):

tex3d (approved):

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
tex3d: By the way, this is the reason for all the forced conversions in the DXC API.  The only Unicode API style for Windows (and thus COM APIs) is the wide string APIs, yet we internally use char strings representing Unicode in UTF-8 encoding.

Issue #7190, opened 2025-03-10
[OMM] Create d3dreflect tests for OMM functionality


bob80905:
**Is your feature request related to a problem? Please describe.**
The RayTracingPipelineConfig1 subobject needs to represent the new `RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS` flag. Additionally, shaders that use the new HLSL OMM flags (`RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS` or `RAY_FLAG_FORCE_OMM_2_STATE`) should have the minimum shader target be 6.9

**Describe the solution you'd like**
`RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS` should be added as a new flag to HLSL. Tests should be written to reflect the new flag, and the minimum shader target of 6.9.

PR #7239, opened 2025-03-20
[OMM] Add D3D Flag RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS, and run d3dreflect tests

bob80905:
This PR adds a new flag, RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS, according to the spec.
It can be used with `RaytracingPipelineConfig1` subobjects. We expect this new flag to be represented in the output.
Additionally, d3dreflect tests are run to ensure that when a rayquery object is using the new OMM enablement flags, that
the minimum shader model target is 6.9.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7190


damyanp (approved):
Looks Good To My Untrained Eye

tools/clang/test/HLSLFileCheck/d3dreflect/rayquery-reflection.hlsl
```diff
@@ -0,0 +1,14 @@
+// RUN: %dxilver 1.9 | %dxc -T lib_6_9 -validator-version 1.9 %s | FileCheck -check-prefix=NOD3D %s
+// RUN: %dxilver 1.9 | %dxc -T lib_6_9 -validator-version 1.9 %s | %D3DReflect %s | FileCheck -check-prefix=D3D %s
+
+// NOD3D: ; RaytracingPipelineConfig1 rpc = { MaxTraceRecursionDepth = 32, Flags = RAYTRACING_PIPELINE_FLAG_SKIP_TRIANGLES | RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS };
+// D3D: MinShaderTarget: 0x10069
```
tex3d, line 5: Note: this `MinShaderTarget: 0x10069` is based only on the use of the `AllocateRayQuery2` DXIL op in the shader function.  It should have nothing to do with the subobject flag being introduced in this PR.

tools/clang/test/HLSLFileCheck/d3dreflect/rayquery-reflection.hlsl
```diff
@@ -0,0 +1,14 @@
...
+RaytracingPipelineConfig1 rpc = { 32, RAYTRACING_PIPELINE_FLAG_SKIP_TRIANGLES | RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS };
+
+RaytracingAccelerationStructure RTAS;
+// DXR entry point to ensure RDAT flags match during validation.
```
tex3d, line 9: This isn't a DXR entry point.

It wouldn't be checked to match with the subobject even if it was, because how the subobject is used depends on runtime state.  So the D3D12 Runtime will validate this kind of consistency, at least as much as it can.  Since we don't expose the use of `RAY_FLAG_FORCE_OMM_2_STATE` in a DXR entry point to the runtime, it can't actually validate this interaction.

Ultimately, we don't need an entry point in this test, as it doesn't interact with the subobject and the flag.

tools/clang/lib/AST/ASTContextHLSL.cpp
```diff
@@ -698,6 +698,10 @@ void hlsl::AddRaytracingConstants(ASTContext &context) {
   AddConstUInt(
       context, StringRef("RAYTRACING_PIPELINE_FLAG_SKIP_PROCEDURAL_PRIMITIVES"),
       (unsigned)DXIL::RaytracingPipelineFlags::SkipProceduralPrimitives);
+  AddConstUInt(context, context.getTranslationUnitDecl(),
+               StringRef("RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS"),
+               (unsigned)DXIL::RaytracingPipelineFlags::AllowOpacityMicromaps,
+               ConstructAvailabilityAttribute(context, VT69));
```
tex3d, line 704: You will want a test that checks the availability attribute.  However, you'll need some kind of change to ensure that we pick it up.  Normally, we collect all the initializers for globals that could be used by any entry points.  Subobjects are not visible to entry points, so they would have to be collected independently to check availability attributes on constants like this one.  So I believe there will need to be additional work in that code to hook this up.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -733,4 +732,20 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
     for (FunctionDecl *FD : callGraph.GetVisitedFunctions())
       Visitor.TraverseDecl(FD);
   }
+
+  if (shaderModel->GetKind() == DXIL::ShaderKind::Library) {
+    for (VarDecl *VD : GlobalsWithInit) {
```
tex3d, line 737: We need to place subobjects into a separate collection so we do not include them with other globals that entry points initialize.  Here's my suggested change to `GatherGlobalsWithInitializers` for this:
```cpp
void GatherGlobalsWithInitializers(
    DeclContext *DC, llvm::SmallVectorImpl<VarDecl *> &GlobalsWithInit,
    llvm::SmallVectorImpl<VarDecl *> &SubobjectGlobals) {
  for (auto *D : DC->decls()) {
    // Skip built-ins and function decls.
    if (D->isImplicit() || isa<FunctionDecl>(D))
      continue;
    if (auto *VD = dyn_cast<VarDecl>(D)) {
      // Add if user-defined static or groupshared global with initializer.
      if (VD->hasInit() && VD->hasGlobalStorage() &&
          (VD->getStorageClass() == SC_Static ||
           VD->hasAttr<HLSLGroupSharedAttr>())) {
        // Place subobjects in a separate collection.
        if (const RecordType *RT = VD->getType()->getAs<RecordType>()) {
          if (RD->hasAttr<HLSLSubObjectAttr>()) {
            SubobjectGlobals.push_back(VD);
            continue;
          }
        }
        GlobalsWithInit.push_back(VD);
      }
    } else if (auto *DC = dyn_cast<DeclContext>(D)) {
      // Recurse into DeclContexts like namespace, cbuffer, class/struct, etc.
      GatherGlobalsWithInitializers(DC, GlobalsWithInit, SubobjectGlobals);
    }
  }
}
```

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -310,11 +311,19 @@ void GatherGlobalsWithInitializers(
...
           (VD->getStorageClass() == SC_Static ||
-           VD->hasAttr<HLSLGroupSharedAttr>()))
-        GlobalsWithInit.push_back(VD);
+           VD->hasAttr<HLSLGroupSharedAttr>())) {
+        // Place subobjects in a separate collection.
+        QualType QT = VD->getType();
+        if (const RecordType *RT = QT->getAs<RecordType>()) {
+          RecordDecl *RD = RT->getDecl();
+          if (RD->hasAttr<HLSLSubObjectAttr>())
+            SubObjects.push_back(VD);
```
tex3d, line 320: If `RD` lacks `HLSLSubObjectAttr`, VD will not be added to `GlobalsWithInit`.  See my suggested change.  It uses `continue;`, instead of an `else` on the first branch.

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -592,14 +601,25 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
+
+  if (shaderModel->GetKind() == DXIL::ShaderKind::Library) {
+    for (VarDecl *VD : SubObjects) {
+      DXIL::NodeLaunchType NodeLaunchTy = DXIL::NodeLaunchType::Invalid;
+      HLSLReachableDiagnoseVisitor Visitor(
+          self, shaderModel, shaderModel->GetKind(), NodeLaunchTy, nullptr,
+          DiagnosedCalls, DeclAvailabilityChecked, DiagnosedTypeLocs);
+      Visitor.TraverseDecl(VD);
+    }
+  }
```
tex3d, line 621: Visitor can/should be initialized outside of the loop:
```suggestion
  if (shaderModel->GetKind() == DXIL::ShaderKind::Library) {
    DXIL::NodeLaunchType NodeLaunchTy = DXIL::NodeLaunchType::Invalid;
    HLSLReachableDiagnoseVisitor Visitor(
        self, shaderModel, shaderModel->GetKind(), NodeLaunchTy, nullptr,
        DiagnosedCalls, DeclAvailabilityChecked, DiagnosedTypeLocs);
    for (VarDecl *VD : SubObjects)
      Visitor.TraverseDecl(VD);
  }
```

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -592,14 +601,25 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
+                                GlobalsWithInit, SubObjects);
+
+  if (shaderModel->GetKind() == DXIL::ShaderKind::Library) {
+    DXIL::NodeLaunchType NodeLaunchTy = DXIL::NodeLaunchType::Invalid;
+    HLSLReachableDiagnoseVisitor Visitor(
+        self, shaderModel, shaderModel->GetKind(), NodeLaunchTy, nullptr,
+        DiagnosedCalls, DeclAvailabilityChecked, DiagnosedTypeLocs);
+    for (VarDecl *VD : SubObjects) {
+      Visitor.TraverseDecl(VD);
+    }
```
tex3d, line 620: nit: [brace style](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements).
```suggestion
    for (VarDecl *VD : SubObjects)
      Visitor.TraverseDecl(VD);
```

tex3d (approved):

PR #7240, opened 2025-03-20
Fix compatibility with newer versions of CMake

### LINKED ISSUE #7237: Build failure due to use of old CMAKE version
**Description**
We're trying to vendor and build DirectXShaderCompiler but the version of CMake used by the included version of clang is causing the build to fail with CMake 4.0.0 which has removed support for CMake versions < 3.5.
The current version of clang requires a minimum version of 3.20.0 while the version used in DXC is still using 2.8.8

**Steps to Reproduce**
Attempt to create build with CMake >=4.0.0

**Actual Behavior**
[cmake] CMake Error at DirectXShaderCompiler/tools/clang/CMakeLists.txt:1 (cmake_minimum_required):
[cmake]   Compatibility with CMake < 3.5 has been removed from CMake.
[cmake] 
[cmake]   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
[cmake]   to tell CMake that the project requires at least <min> but has been updated
[cmake]   to work with policies introduced by <max> or earlier.
[cmake] 
[cmake]   Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
[cmake] 
[cmake] 
[cmake] -- Configuring incomplete, errors occurred!

**Environment**
- DXC version Latest

JaimeHW:
These changes allow the use of newer versions of CMake while maintaining compatibility with old versions.

Fixes #7237

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,13 +17,16 @@ if(POLICY CMP0022)
   cmake_policy(SET CMP0022 NEW) # automatic when 2.8.12 is required
 endif()
 
-if (POLICY CMP0051)
-  # CMake 3.1 and higher include generator expressions of the form
-  # $<TARGETLIB:obj> in the SOURCES property.  These need to be
-  # stripped everywhere that access the SOURCES property, so we just
-  # defer to the OLD behavior of not including generator expressions
-  # in the output for now.
-  cmake_policy(SET CMP0051 OLD)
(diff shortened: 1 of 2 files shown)


JaimeHW:

llvm-beanz:
@JaimeHW, apologies for not noticing this PR until just now. We have fixed this issue in #7287 by just removing the outdated CMake since we don't actually support versions that old anyways.

PR #7254, opened 2025-03-25
Update DXIL.rst

alsepkow:
Minor grammar fixes

### DIFF
--- a/docs/DXIL.rst
+++ b/docs/DXIL.rst
@@ -3294,9 +3294,12 @@ Modules and Linking
 ===================
 
 HLSL has linking capabilities to enable third-party libraries. The linking step happens before shader DXIL is given to the driver compilers.
-Experimental library generation is added in DXIL1.1. A library could be created by compile with lib_6_1 profile.
-A library is a dxil container like the compile result of other shader profiles. The difference is library will keep information for linking like resource link info and entry function signatures.
-Library support is not part of DXIL spec. Only requirement is linked shader must be valid DXIL.
+Experimental library generation is added in DXIL1.1. A library could be created
+by compiling with the lib_6_1 profile.
+A library is a dxil container like the compile result of other shader profiles.
+The difference is a library will keep information for linking like resource link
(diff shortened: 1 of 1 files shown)


docs/DXIL.rst
```diff
@@ -3294,9 +3294,12 @@ Modules and Linking
...
 HLSL has linking capabilities to enable third-party libraries. The linking step happens before shader DXIL is given to the driver compilers.
-Experimental library generation is added in DXIL1.1. A library could be created by compile with lib_6_1 profile.
-A library is a dxil container like the compile result of other shader profiles. The difference is library will keep information for linking like resource link info and entry function signatures.
-Library support is not part of DXIL spec. Only requirement is linked shader must be valid DXIL.
+Experimental library generation is added in DXIL1.1. A library could be created
+by compiling with the lib_6_1 profile.
+A library is a dxil container like the compile result of other shader profiles.
+The difference is a library will keep information for linking like resource link
+info and entry function signatures.
+Library support is not part of the DXIL spec. The only requirement is linked shader must be valid DXIL.
```
bogner, line 3302: Is this reflow intentional/helpful? I approve of 80-column but the rest of the doc doesn't do that (not even the first sentence of this paragraph...) and it seems odd to wrap at periods.

bogner (approved):

docs/DXIL.rst
pow2clk: Do you mean wrap outside of periods? I don't mind sentences having their own lines to limit the sidescroll where relevant, but I think that since this is mostly consumed in plain text, it's better to not break up sentences.

bob80905 (approved):

docs/DXIL.rst
alsepkow: Bah, that's from my editor config. I have it set to show me everything as 80 columns now, but I also hit the hotkey to format this after I edited it. Will restore the original formatting.

bob80905 (approved):

PR #7255, opened 2025-03-25
[NFC] containsLongVector -> ContainsLongVector

llvm-beanz:
I provided feedback during code review that this function should be named following LLVM conventions. That feedback did not account for the fact that SemaHLSL is otherwise consistent using CamelCase instead of camelCase naming.

This corrects my error by renaming to match the consistent style in SemaHLSL.h.

I've also updated the parameter naming in the source file to conform to LLVM style since I was in the area anyways.

### DIFF
--- a/tools/clang/lib/Sema/SemaHLSL.cpp
+++ b/tools/clang/lib/Sema/SemaHLSL.cpp
@@ -12184,7 +12184,7 @@ bool hlsl::ContainsLongVector(QualType QT) {
     return false;
 
   while (const ArrayType *Arr = QT->getAsArrayTypeUnsafe())
-  QT = Arr->getElementType();
+    QT = Arr->getElementType();
 
   if (CXXRecordDecl *Decl = QT->getAsCXXRecordDecl()) {
     if (!Decl->isCompleteDefinition())


tex3d:
> I've also updated the parameter naming in the source file to conform to LLVM style since I was in the area anyways.

Did I miss something?  I don't see this part of the change.

tex3d (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12179,14 +12179,14 @@ bool hlsl::ShouldSkipNRVO(clang::Sema &sema, clang::QualType returnType,
...
-bool hlsl::containsLongVector(QualType qt) {
-  if (qt.isNull() || qt->isDependentType())
+bool hlsl::ContainsLongVector(QualType QT) {
+  if (QT.isNull() || QT->isDependentType())
     return false;
 
-  while (const ArrayType *Arr = qt->getAsArrayTypeUnsafe())
-    qt = Arr->getElementType();
+  while (const ArrayType *Arr = QT->getAsArrayTypeUnsafe())
+  QT = Arr->getElementType();
```
bogner, line 12187: Looks like an accidental reformat mistake here.

bogner (dismissed):

damyanp (approved):

Issue #7214, opened 2025-03-17
SER implementation tracker


simoll:
Umbrella bug for tracking the SER implementation

##

- [ ] HitObject type and SM6.9 enablement

### DXIL opcodes + check-pass validation tests

DXIL opcode definitions and DXV tests showing wellformed dxil passing

- [ ] MakeNop|Miss
- [ ] Accessors
- [ ] FromRayQuery
- [ ] Trace
- [ ] Invoke
- [ ] MaybeReorderThread
- [ ] reordercoherent attribute
- [ ] REORDER_SCOPE

### DXIL validation + check-fail validation tests

DXIL validation rules, in so far necessary, plus DXV tests with illformed dxil showing those validation failures

- [ ] Unsupported uses of HitObject type (buffers, groupshared mem, payloads, intersection attributes)
- [ ] MakeMiss
- [ ] Accessors
- [ ] FromRayQuery
- [ ] Trace
- [ ] Invoke
- [ ] MaybeReorderThread
- [ ] reordercoherent attribute
- [ ] REORDER_SCOPE

### HLSL -> HL lowering + sema

HLSL gen_intrin_main.txt definitions, HLSL -> DXIL lowering. HLSL + HL IR tests

- [ ] MakeNop|Miss
- [ ] Accessors
- [ ] FromRayQuery
- [ ] Trace
- [ ] Invoke
- [ ] MaybeReorderThread
- [ ] reordercoherent attribute
- [ ] REORDER_SCOPE

simoll:
Updated table with tracking bugs #7233 and #7234

PR #7256, opened 2025-03-25
[SER] MaybeReorderThread DXIL opcode and validation

simoll:
- DXIL opcodes for MaybeReorderThread
- Validator rules
- DXV validation test (passing & expected failures)

Specification: https://github.com/microsoft/hlsl-specs/blob/main/proposals/0027-shader-execution-reordering.md

DXC SER implementation tracker: #7214


simoll:
Changed dxil validation rules in acc with https://github.com/microsoft/hlsl-specs/issues/485
- `HitObject` must not be `undef`
- `NumCoherenceBits` must not be `undef`
- `CoherenceBits` must not be undef when `NumCoherenceBits != 0`

also updated passing and failing validation test cover those cases

llvm-beanz (approved):

utils/hct/hctdb.py
```diff
@@ -866,6 +866,11 @@ def populate_categories_and_models(self):
                 "closesthit",
                 "miss",
             )
+        for i in ("MaybeReorderThread").split(","):
+            self.name_idx[i].shader_stages = (
```
tex3d, line 870: Forgot category and shader model here:

```suggestion
        for i in ("MaybeReorderThread").split(","):
            self.name_idx[i].category = "Shader Execution Reordering"
            self.name_idx[i].shader_model = 6, 9
            self.name_idx[i].shader_stages = (
```

Note: generated files will need updating after this, so it should be applied offline.

tex3d (commented):
DXIL op is missing category and shader model, otherwise looks good.

utils/hct/hctdb.py
tex3d: I just did this, since it was easy for me to do at this time.

tex3d (approved):
Looks good now, but I'm the last one who pushed changes, so it needs a new independent review.

llvm-beanz (approved):

Issue #7257, opened 2025-03-25
[Feature Request] Add SubObject attribute


bob80905:
**Is your feature request related to a problem? Please describe.**
Subobjects may exist in library shaders without any function declarations. If subobjects exist in such a library shader, they won't be semantically diagnosed in the translation unit semantic analysis. This means if subobjects use a value that has an availability attribute in its initializer, it won't be caught.

**Describe the solution you'd like**
We need a way to detect that a given decl is a subobject decl, to know whether or not to investigate further and check if its initializer contains any availability attributes. A proposed solution is to add a subobject hlsl attribute, that will be attached to all subobject decls.

This will help with the OMM implementation, since RayTracingPipelineConfig1 subobjects may be initialized with values that have availability attributes.

PR #7258, opened 2025-03-25
[Sema] Add and test new Subobject Attribute

bob80905:
This PR adds and tests a new subobject attribute. It will be useful for checking if a given decl is a subobject decl. This functionality will be used in https://github.com/microsoft/DirectXShaderCompiler/pull/7239
We need an attribute in order to determine whether to check its initializer for availability attributes or not.

Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7257


tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -687,6 +687,11 @@ bool GetHLSLSubobjectKind(clang::QualType type,
   hgType = (DXIL::HitGroupType)(-1);
   type = type.getCanonicalType();
   if (const RecordType *RT = type->getAs<RecordType>()) {
+    RecordDecl *RD = RT->getDecl();
+    if (!RD->hasAttr<HLSLSubObjectAttr>()) {
+      return false;
+    }
```
pow2clk, line 693: Given that you've added the attribute now, it would be pretty easy to avoid the string compares by adding a parameter to it containing an integer representation of the Subobjectkind similar to how HLSLResourceAttr does. The callers of , CreateSubobjectStateObjectConfig and friends that call StartSubObjectDecl has the AR, which indicates which DXIL:SubobjectKind they need. It would be pretty easy to plumb it down and apply it to the attr.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -692,60 +693,38 @@ bool GetHLSLSubobjectKind(clang::QualType type,
...
-    case 27:
-      if (name == "ProceduralPrimitiveHitGroup") {
-        subobjectKind = DXIL::SubobjectKind::HitGroup;
-        hgType = DXIL::HitGroupType::ProceduralPrimitive;
-        return true;
-      }
-      return false;
+    HLSLSubObjectAttr *Attr = RD->getAttr<HLSLSubObjectAttr>();
+
+    switch (Attr->getSubObjKindUint()) {
```
pow2clk, line 698: Just to make sure of clarity here. I expected the argument to be the DXIL::SubobjectKind enum rather than the AR_OBJECTs

pow2clk (approved):
Thanks for indulging me!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -2785,13 +2785,17 @@ AddBuiltInTriangleIntersectionAttributes(ASTContext &context,
 //
 // Subobjects
 
-static CXXRecordDecl *StartSubobjectDecl(ASTContext &context,
-                                         const char *name) {
+static CXXRecordDecl *
+StartSubobjectDecl(ASTContext &context, const char *name,
+                   DXIL::SubobjectKind kind,
+                   DXIL::HitGroupType hgtype = (DXIL::HitGroupType)(-1)) {
```
tex3d, line 2791: I don't think we should cast an in valid enum value (-1) to the enum type.

Recommend using `DXIL::HitGroupType::LastEntry` instead.

You can use `IsValidHitGroupType()` to check if it has been supplied with a valid value when interpreting it later.

In the attribute definition, we do not need a default value, since it is always created with an explicit value.  This solves the problem of keeping the int representation in the `.td` file for the default value in sync with `LastEntry`.

Let's also name our new parameters based on our coding standard.

```suggestion
                   DXIL::SubobjectKind Kind,
                   DXIL::HitGroupType HGT = DXIL::HitGroupType::LastEntry) {
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -2785,13 +2785,17 @@ AddBuiltInTriangleIntersectionAttributes(ASTContext &context,
...
   IdentifierInfo &id =
       context.Idents.get(StringRef(name), tok::TokenKind::identifier);
   CXXRecordDecl *decl = CXXRecordDecl::Create(
       context, TagTypeKind::TTK_Struct, context.getTranslationUnitDecl(), NoLoc,
       NoLoc, &id, nullptr, DelayTypeCreationTrue);
+  decl->addAttr(
+      HLSLSubObjectAttr::CreateImplicit(context, (unsigned)kind, (int)hgtype));
```
tex3d, line 2798: Let's use `static_cast` for casts, according to the coding standards.

```suggestion
  decl->addAttr(HLSLSubObjectAttr::CreateImplicit(
      context, static_cast<unsigned>(Kind), static_cast<int>(HGT)));
```

tools/clang/include/clang/Basic/Attr.td
```diff
@@ -1157,6 +1157,13 @@ def HLSLRayQueryObject : InheritableAttr {
   let Documentation = [Undocumented];
 }
 
+def HLSLSubObject : InheritableAttr {
+  let Spellings = []; // No spellings!
+  let Subjects = SubjectList<[CXXRecord]>;
+  let Documentation = [Undocumented];
+  let Args = [UnsignedArgument<"SubObjKindUint">, DefaultIntArgument<"HitGroupType", -1>];
```
tex3d, line 1164: Let's remove the default argument and make sure it's always a defined `DXIL::HitGroupType` enum value, which means this could also be unsigned.
```suggestion
  let Args = [UnsignedArgument<"SubObjKindUint">, UnsignedArgument<"HitGroupType">];
```

tools/clang/test/HLSLFileCheck/shader_targets/raytracing/subobjects_attr.hlsl
```diff
@@ -0,0 +1,69 @@
+// RUN: %dxilver 1.9 | %dxc -T lib_6_9  -ast-dump-implicit %s | FileCheck -check-prefix=ASTIMPL %s
+// The HLSL source is just a copy of 
+// tools\clang\test\HLSLFileCheck\shader_targets\raytracing\subobjects_raytracingPipelineConfig1.hlsl
```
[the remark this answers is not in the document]
bob80905: Yes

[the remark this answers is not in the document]
tex3d: Ok, so it appears to be true, but why have two copies, unless you put this under SemaHLSL instead, since it's only testing AST.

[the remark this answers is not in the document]
tex3d: Also, is it even necessary to declare any subobjects to check all of the built-in subobject AST?  I don't think so.

tex3d (approved):
LGTM!

PR #7262, opened 2025-03-26
[SER] MaybeReorderThread + Make(Nop|Miss) HLSL -> DXIL lowering and tests

simoll:
- HLSL -> DXIL lowering
- ast, hlsl->dxil, dxilgen, and ScalarReplAggregatesHLSL tests

SER implementation tracker (#7214)


lib/HLSL/HLOperationLower.cpp
```diff
@@ -6068,7 +6068,53 @@ Value *TranslateHitObjectMake(CallInst *CI, IntrinsicOp IOP, OP::OpCode opcode,
                               HLOperationLowerHelper &helper,
                               HLObjectOperationLowerHelper *pObjHelper,
                               bool &Translated) {
```
llvm-beanz, line 6070: Normally I wouldn't ask you to clean up pre-existing code, but in this case you added the code and the change here modifies the whole function, so please update the parameter names and all the variables throughout the body to conform to LLVM Coding Standards (https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly)

simoll: Reworked the variable names to follow the coding standards

tools/clang/test/DXC/Passes/ScalarReplHLSL/hitobject_make_scalarrepl.ll
```diff
@@ -0,0 +1,130 @@
+; RUN: %dxopt %s -hlsl-passes-resume -scalarrepl-param-hlsl -S | FileCheck %s
+; REQUIRES: dxil-1-9
```
llvm-beanz, line 2: The IR tests don't run the validator and don't run with different versions of the compiler, so we don't need these REQUIRES lines.

tools/clang/test/DXC/Passes/ScalarReplHLSL/hitobject_make_scalarrepl.ll
```diff
@@ -0,0 +1,130 @@
...
+; Function Attrs: nounwind
+define void @"\01?main@@YAXXZ"() #0 {
+entry:
+  %hit = alloca %dx.types.HitObject, align 4
+  %tmp = alloca %dx.types.HitObject, align 4
+  %ray = alloca %struct.RayDesc, align 4
+  %tmp2 = alloca %dx.types.HitObject, align 4
+  %0 = bitcast %dx.types.HitObject* %hit to i8*, !dbg !23 ; line:42 col:3
+  call void @llvm.lifetime.start(i64 4, i8* %0) #0, !dbg !23 ; line:42 col:3
+; CHECK:  %{{[^ ]+}} = call %dx.types.HitObject* @"dx.hl.op..%dx.types.HitObject* (i32, %dx.types.HitObject*)"(i32 358, %dx.types.HitObject* %{{[^ ]+}})
```
llvm-beanz, line 40: This match would succeed even if the IR is unchanged from with is in the test case now, so we probably need some more specific matches, maybe a variable match in order to verify that this IR is actually being transformed.

llvm-beanz (changes requested):
All these test cases need reworking. It looks like many of the matches that you have in the tests would match on the input file so they aren't actually testing that the IR was changed.

tools/clang/test/DXC/Passes/ScalarReplHLSL/hitobject_make_scalarrepl.ll
simoll: That is because this call to `HitObject_MakeNop` is not expected to change. SroaHLSL must only flatten the `RayDesc` param in the `DxHitObject_MakeMiss` call below and the test is checking for exactly that.
@llvm-beanz If that is not what we want, could you provide more specific guidance of what you want to see changed?

tools/clang/test/DXC/Passes/ScalarReplHLSL/hitobject_make_scalarrepl.ll
simoll: Removed the `REQUIRES` line

tools/clang/test/DXC/Passes/ScalarReplHLSL/hitobject_make_scalarrepl.ll
llvm-beanz: If the goal is to make sure this doesn't get scalarized, we should have some variable matches to make sure that the values move as expected through the operations.

tools/clang/test/CodeGenDXIL/hlsl/objects/HitObject/hitobject_make.hlsl
```diff
@@ -0,0 +1,75 @@
...
+
+// AST: | | |-FunctionTemplateDecl {{[^ ]+}} <<invalid sloc>> <invalid sloc> MakeNop
+// AST-NEXT: | | | |-TemplateTypeParmDecl {{[^ ]+}} <<invalid sloc>> <invalid sloc> class TResult
+// AST-NEXT: | | | |-CXXMethodDecl {{[^ ]+}} <<invalid sloc>> <invalid sloc> implicit MakeNop 'TResult () const' static
+// AST-NEXT: | | | `-CXXMethodDecl {{[^ ]+}} <<invalid sloc>> <invalid sloc> used MakeNop 'dx::HitObject ()' static
+// AST-NEXT: | | |   |-TemplateArgument type 'dx::HitObject'
+// AST-NEXT: | | |   |-HLSLIntrinsicAttr {{[^ ]+}} <<invalid sloc>> Implicit "op" "" 358
+// AST-NEXT: | | |   `-AvailabilityAttr {{[^ ]+}} <<invalid sloc>> Implicit  6.9 0 0 ""
+
+// FCGL: %{{[^ ]+}} = call %dx.types.HitObject* @"dx.hl.op..%dx.types.HitObject* (i32, %dx.types.HitObject*)"(i32 358, %dx.types.HitObject* %{{[^ ]+}})
```
llvm-beanz, line 42: Just as a pointer, even in the `-fcgl` this SSA value is never used, so we really don't seem to need it.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -6183,19 +6184,114 @@ Value *TranslateUnpack(CallInst *CI, IntrinsicOp IOP, OP::OpCode opcode,
...
+  Value *HitObjectPtr = CI->getArgOperand(SrcIdx++);
+  if (Opcode == OP::OpCode::HitObject_MakeNop) {
+    Value *HitObject = TrivialDxilOperation(
+        Opcode, {nullptr}, Type::getVoidTy(CI->getContext()), CI, HlslOP);
+    Builder.CreateStore(HitObject, HitObjectPtr);
+    // Passthru the 'this' pointer when this calls 'HitObject_MakeNop' as the
+    // default constructor
+    if (!CI->getType()->isVoidTy())
+      return HitObjectPtr;
+    return nullptr;
```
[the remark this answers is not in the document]
simoll: Can we address this in a followup? Changing the default ctor is kind of orthogonal to the lowering we do here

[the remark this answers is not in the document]
simoll: Tracking this as a followup in #7354

tools/clang/test/CodeGenDXIL/hlsl/objects/HitObject/hitobject_make.hlsl
simoll: Tracking this as a followup in #7354

tools/clang/test/DXC/Passes/ScalarReplHLSL/hitobject_make_scalarrepl.ll
simoll: Added explicit `CHECK`s that `RayDesc` gets scalarized and all other params remain in place. Also checking that the ptr from the default ctr remains unused.
@llvm-beans does this resolve this request?

tools/clang/test/CodeGenDXIL/hlsl/objects/HitObject/hitobject_make.hlsl
```diff
@@ -1,3 +1,4 @@
+// REQUIRES: dxil-1-9
```
llvm-beanz, line 1: I don't think this is needed. This is a compiler test not a validator test. It is never run without the latest validator.

simoll: Deleted that line again

tools/clang/test/DXC/Passes/ScalarReplHLSL/hitobject_make_scalarrepl.ll
simoll: Asserting that `CI->use_empty()` as discussed offline. Resolved

llvm-beanz (approved):
LGTM.

tools/clang/test/CodeGenDXIL/hlsl/objects/HitObject/hitobject_make.hlsl
```diff
@@ -0,0 +1,75 @@
+// RUN: %dxc -T lib_6_9 -E main %s | FileCheck %s --check-prefix DXIL
+// RUN: %dxc -T lib_6_9 -E main %s -fcgl | FileCheck %s --check-prefix FCGL
+// RUN: %dxc -T lib_6_9 -E main %s -ast-dump-implicit | FileCheck %s --check-prefix AST
```
tex3d, line 3: I think AST tests belong under SemaHLSL.  We could probably address this later for these tests, but in the future, we should be putting them there.  For -fcgl and full compilation, that should go here, but we also need to add a `lit.local.cfg` under `CodeGenDXIL/hlsl/objects/HitObject` to require `dxil-1-9`.  This is not a requirement for SemaHLSL tests.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/maybereorder.hlsl
```diff
@@ -0,0 +1,37 @@
+// RUN: %dxc -T lib_6_9 -E main %s | FileCheck %s --check-prefix DXIL
+// RUN: %dxc -T lib_6_9 -E main %s -fcgl | FileCheck %s --check-prefix FCGL
+// RUN: %dxc -T lib_6_9 -E main %s -ast-dump-implicit | FileCheck %s --check-prefix AST
```
tex3d, line 3: I think AST tests belong under SemaHLSL.  We could probably address this later for these tests, but in the future, we should be putting them there.  For -fcgl and full compilation, that should go here, but for tests not under a directory already excluded from unsupported DXIL versions, we need a `// REQUIRES: dxil-1-9` in the test.

tex3d (approved):
Minor comments.  Could address later.
In the future, let's put AST tests under `SemaHLSL`, not under `CodeGenDXIL`.

tools/clang/test/CodeGenDXIL/hlsl/objects/HitObject/hitobject_make.hlsl
simoll: Tracking this in #7359

simoll: Added a `lit.local.cfg` to the `CodeGenDXIL/../HitObject` dir

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/maybereorder.hlsl
simoll: Tracking this in #7359

lib/HLSL/HLOperationLower.cpp
```diff
@@ -6183,11 +6183,57 @@ Value *TranslateUnpack(CallInst *CI, IntrinsicOp IOP, OP::OpCode opcode,
...
+  }
+
+  DXASSERT_NOMSG(CI->getNumArgOperands() ==
+                 HLOperandIndex::kHitObjectMakeMiss_NumOp);
+  Value *RayFlags = CI->getArgOperand(SrcIdx++);
+  Value *MissShaderIdx = CI->getArgOperand(SrcIdx++);
+  Value *RayDescOrigin = CI->getArgOperand(SrcIdx++);
```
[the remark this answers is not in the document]
simoll: Added the `assert`

tex3d (approved):

PR #7265, opened 2025-03-26
Update DXC's CONTRIBUTING file

llvm-beanz:
This change seeks to address some recent questions about how the LLLVM Coding Standards are applied in DXC.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,6 +44,28 @@ The coding, style, and general engineering guidelines follow those described in 
 
 DXC has adopted a clang-format requirement for all incoming changes to C and C++ files. PRs to DXC should have the *changed code* clang formatted to the LLVM style, and leave the remaining portions of the file unchanged. This can be done using the `git-clang-format` tool or IDE driven workflows. A GitHub action will run on all PRs to validate that the change is properly formatted.
 
+#### Applying LLVM Standards
+
+All new code contributed to DXC should follow the LLVM coding standards.
+
+Note that the LLVM Coding Standards have a golden rule:
+
+> **If you are extending, enhancing, or bug fixing already implemented code, use the style that is already being used so that the source is uniform and easy to follow.**
+
+The golden rule should continue to be applied to places where DXC is self-consistent. A good example is DXC's common use of `CamelCase` instead of `camelCase` for APIs in some parts of the HLSL implementation. In any place where DXC is not self-consistent new code should follow the LLVM Coding Standard.
+
+A good secondary rule to follow is:
+
+> **When in doubt, follow LLVM.**
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -44,6 +44,28 @@ The coding, style, and general engineering guidelines follow those described in
...
 
+#### Applying LLVM Standards
+
+All new code contributed to DXC should follow the LLVM coding standards.
+
+Note that the LLVM Coding Standards have a golden rule:
+
+> **If you are extending, enhancing, or bug fixing already implemented code, use the style that is already being used so that the source is uniform and easy to follow.**
+
+The golden rule should continue to be applied to places where DXC is self-consistent. A good example is DXC's common use of `CamelCase` instead of `camelCase` for APIs in some parts of the HLSL implementation. In any place where DXC is not self-consistent new code should follow the LLVM Coding Standard.
```
V-FEXrt, line 55: nit:

```suggestion
The golden rule should continue to be applied to places where DXC is self-consistent. A good example is DXC's common use of `PascalCase` instead of `camelCase` for APIs in some parts of the HLSL implementation. In any place where DXC is not self-consistent new code should follow the LLVM Coding Standard.
```

damyanp (approved):
Should we have a link to the LLVM coding standards from here as well?

V-FEXrt (approved):

bob80905 (approved):

llvm-beanz:
> Should we have a link to the LLVM coding standards from here as well?

There was a broken link in the docs a few lines above (DXC has always referred to the LLVM Coding Standards in its CONTRIBUTING file). I fixed that link. Do you think it warrants a second link, or is the one in the first sentence of the containing section sufficient?

Icohedron (approved):

damyanp:
> There was a broken link in the docs a few lines above (DXC has always referred to the LLVM Coding Standards in its CONTRIBUTING file). I fixed that link. Do you think it warrants a second link, or is the one in the first sentence of the containing section sufficient?

Oh, I missed that - I was looking for a link to https://llvm.org/... so didn't spot the link to a local file.  If the contents of the linked file is correct then I don't think we need another one.

llvm-beanz:
> If the contents of the linked file is correct then I don't think we need another one.

I think they're correct enough. While the LLVM coding standards have evolved over the ~10 years since DXC forked, the core of the standards have remained largely the same. If we encounter issues where the DXC document is inappropriately stale I'd suggest we update it.

bogner (approved):

inbelic (approved):

bharadwajy (approved):

joaosaffran (approved):

PR #7272, opened 2025-03-28
[PIX] Add a pass for PIX to log missing NonUniformResourceIndex usage into a UAV

nopandbrk:
This is a pass to add instructions to determine missing usage of the NonUniformResourceIndex qualifier when dynamically indexing resources. The instruction numbers will be written out to a UAV for later ingestion by PIX to present a view of the output.


lib/DxilPIXPasses/PixPassHelpers.cpp
```diff
@@ -500,6 +500,90 @@ unsigned int FindOrAddSV_Position(hlsl::DxilModule &DM,
...
+void ForEachDynamicallyIndexedResource(
+    hlsl::DxilModule &DM,
+    const std::function<bool(bool, Instruction *, Value *)> &Visitor) {
+  OP *HlslOP = DM.GetOP();
+  LLVMContext &Ctx = DM.GetModule()->getContext();
+
+  for (llvm::Function &F : DM.GetModule()->functions()) {
+    if (F.isDeclaration() && !F.use_empty() && OP::IsDxilOpFunc(&F)) {
+      if (F.hasName()) {
+        if (F.getName().find("createHandleForLib") != StringRef::npos) {
```
jeffnn, line 512: There's probably a good reason that I've forgotten , but why is createHandleForLib treated differently from the other three in this function?

lib/DxilPIXPasses/DxilNonUniformResourceIndexInstrumentation.cpp
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        if (!PixUavResource) {
+          PixUavResource =
+              PIXPassHelpers::CreateGlobalUAVResource(DM, 0, "PixUavResource");
+        }
+
+        IRBuilder<> Builder(CreateHandle);
+
+        CallInst *HandleForUAV = PIXPassHelpers::CreateHandleForResource(
```
jeffnn, line 80: This is creating too many UAVs!  CreateUAVOnceForModule should probably be deleted (or at least renamed!)- there aren't many situations in which is applicable. 

The pattern is intended to be : 
-Call CreateGlobalUAVResource once near the start of runOnModule
-For every function in the module in which you want to access the UAV, create a handle to that UAV via CreateHandleForResource, and use that handle inside that function. (For [noinline], you should be able to create/use a new handle for that function. TBH I've still not followed up on that issue for the other passes.)

jeffnn: (I've used a map from Function * to UAV handle in other passes, FWIW.)

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: Oh, cuz it got pasted from the access tracking pass! Might be a good opportunity to rationalize the four cases, then.

lib/DxilPIXPasses/DxilShaderAccessTracking.cpp
```diff
@@ -795,87 +795,6 @@ DxilShaderAccessTracking::GetResourceFromHandle(Value *resHandle,
   return ret;
 }
 
-static bool CheckForDynamicIndexing(OP *HlslOP, LLVMContext &Ctx,
-                                    DxilModule &DM) {
```
jeffnn, line 799: Maybe this shouldn't be deleted in this PR cuz it's unrelated to the goals of the PR?

adam-yang (dismissed):
lgtm!

jeffnn (dismissed):
LGTM!

lib/DxilPIXPasses/PixPassHelpers.cpp
jeffnn: (Decided that would be for another day)

jeffnn (approved):

PR #7285, opened 2025-04-01
Add staging branches to azure pipeline testing

pow2clk:
The name matches for azure pipelines to run excluded staging branches. This adds them in.

Not sure this is desired, but it's here if we want it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,10 +1,12 @@
 trigger:
   - main
   - release*
+  - staging*
 
 pr: 
   - main
   - release*
+  - staging*
   
 resources:
 - repo: self


damyanp:
Are you sure that this shouldn't be targeting main?

damyanp:
> Are you sure that this shouldn't be targeting main?

Well, I guess the fact that this PR is running the CI tests is pretty strong evidence that this has worked!

pow2clk:
> Are you sure that this shouldn't be targeting main?

Depends if we want a permanent policy shift. Main is fine with me

damyanp (approved):

bob80905 (approved):

PR #7286, opened 2025-04-01
Add UUID compiler extension check on Clang

### LINKED ISSUE #7248: WinAdapter.h is not compilable under Clang on Linux
**Description**
Getting a bunch of errors like this:
```cpp
In file included from .../dxc/dxcapi.h:36:
.../dxc/WinAdapter.h:572:33: error: expected parameter declarator
  572 | CROSS_PLATFORM_UUIDOF(IUnknown, "00000000-0000-0000-C000-000000000046")
      |                                 ^
.../dxc/WinAdapter.h:572:33: error: expected ')'
.../dxc/WinAdapter.h:572:1: note: to match this '('
  572 | CROSS_PLATFORM_UUIDOF(IUnknown, "00000000-0000-0000-C000-000000000046")
      | ^
.../dxc/WinAdapter.h:552:25: note: expanded from macro 'CROSS_PLATFORM_UUIDOF'
  552 |   struct __declspec(uuid(spec)) interface;
      |                         ^
.../dxc/WinAdapter.h:572:23: error: expected function body after function declarator
  572 | CROSS_PLATFORM_UUIDOF(IUnknown, "00000000-0000-0000-C000-000000000046")
      |                       ^
.../dxc/WinAdapter.h:579:36: error: 'Q' does not refer to a value
  579 |     return QueryInterface(__uuidof(Q), (void **)pp);
      |                                    ^
.../dxc/WinAdapter.h:578:19: note: declared here
  578 |   template <class Q> HRESULT QueryInterface(Q **pp) {
      |                   ^
```

This part of the header is incorrect:
```cpp
// If it is GCC, there is no UUID support and we must emulate it.
#ifndef __clang__
#define __EMULATE_UUID 1
#endif // __clang__
```

Maybe UUID-related extensions were accepted in old versions of Clang even on Linux without `-fms-extensions`, but with recent ones they are not. This check must be removed to always emulate UUID on Linux.

**Steps to Reproduce**
[One-line reproducer on godbolt](https://gcc.godbolt.org/z/EP9j1d9n4)

**Environment**
- DXC version: any
- Host Operating System: any Linux with Clang compiler

urshanselmann:
Fixes #7248

Fix Clang Compilation on Linux without Microsoft extensions enabled.

## Rationale
Clang support depends on the `-fms-extensions` compiler flag. [[1]](https://clang.llvm.org/docs/UsersManual.html#microsoft-extensions)
If enabled, the `_MSC_EXTENSIONS` macro is defined. [[2]](https://github.com/llvm/llvm-project/blob/19a319667b567a26a20f9829a0ae7e6a5c259cba/clang/lib/Basic/Targets/OSTargets.cpp#L248)

### DIFF
--- a/include/dxc/WinAdapter.h
+++ b/include/dxc/WinAdapter.h
@@ -51,7 +51,8 @@
 #define _countof(a) (sizeof(a) / sizeof(*(a)))
 
 // If it is GCC, there is no UUID support and we must emulate it.
-#ifndef __clang__
+// Clang support depends on the -fms-extensions compiler flag.
+#if !defined(__clang__) || !defined(_MSC_EXTENSIONS)
 #define __EMULATE_UUID 1
 #endif // __clang__
 


llvm-beanz (approved):
This change LGTM, and should allow clients of DXC to use Clang without the MS extensions enabled.

damyanp (approved):

PR #7287, opened 2025-04-01
Require CMake 3.17, remove CMP0051

llvm-beanz:
Hopefully this works and gets us able to build with CMake 4+.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,15 +17,6 @@ if(POLICY CMP0022)
   cmake_policy(SET CMP0022 NEW) # automatic when 2.8.12 is required
 endif()
 
-if (POLICY CMP0051)
-  # CMake 3.1 and higher include generator expressions of the form
-  # $<TARGETLIB:obj> in the SOURCES property.  These need to be
-  # stripped everywhere that access the SOURCES property, so we just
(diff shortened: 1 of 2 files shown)


damyanp (approved):

pow2clk (approved):
This was inherited from LLVM 3.7. There's no reason to think we need it here if everything builds without it.

CMakeLists.txt
```diff
@@ -17,15 +17,6 @@ if(POLICY CMP0022)
...
-if (POLICY CMP0051)
-  # CMake 3.1 and higher include generator expressions of the form
-  # $<TARGETLIB:obj> in the SOURCES property.  These need to be
-  # stripped everywhere that access the SOURCES property, so we just
-  # defer to the OLD behavior of not including generator expressions
-  # in the output for now.
-  cmake_policy(SET CMP0051 OLD)
-endif()
-
 if(CMAKE_VERSION VERSION_LESS 3.1.20141117)
```
tex3d, line 20: Existing issue note: is it possible for `if(CMAKE_VERSION VERSION_LESS 3.1.20141117)` to be true with the `cmake_minimum_required(VERSION 3.17.2)` set at the top of this file?

tex3d (approved):

PR #7289, opened 2025-04-01
Fix assert due to unreachable discard

amaiorano:
When emitting discard in an unreachable code context (e.g. after an infinite loop), DXC would assert (if asserts enabled), or trigger a UBSAN failure because the discard instruction would have no parent. When an infinite loop is emitted during CodeGen, the InsertPt is cleared, thus subsequent discard instructions would be created, but no parent set. We skip emitting discard in this case, which follows the same pattern as is done for EmitIfStmt, and EmitSwitchStmt.

### DIFF
--- a/tools/clang/lib/CodeGen/CGStmt.cpp
+++ b/tools/clang/lib/CodeGen/CGStmt.cpp
@@ -525,6 +525,10 @@ void CodeGenFunction::EmitGotoStmt(const GotoStmt &S) {
 
 // HLSL Change Begins.
 void CodeGenFunction::EmitDiscardStmt(const DiscardStmt &S) {
+  // Skip unreachable discard.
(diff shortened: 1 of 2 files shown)


amaiorano:
An example of the problem can be seen on compiler explorer: https://godbolt.org/z/c4fq7eaTz

Because CE uses a release build without asserts and without ubsan, the extra discard is needed before the while loop to showcase the failure. I didn't dig into why exactly, but without it, DXC behaves "correctly". But with asserts or UBSAN enabled, the fact that the second discard has a nullptr parent results in a crash.

llvm-beanz (dismissed):

tex3d (approved):

llvm-beanz (approved):

damyanp (approved):

PR #7290, opened 2025-04-01
Implementation of the CoopVec Inference and Training builtin intrinisics

anupamachandra:
Implements
HLSL:
__builtin_MatVecMul
__builtin_MatVecMulAdd
__builtin_OuterProductAccumulate
__builtin_VectorAccumulate

Lowered to
DXIL:
@dx.op.matVecMul
@dx.op.matVecMulAdd
 @dx.op.outerProductAccumulate
 @dx.op.vectorAccumulate


utils/hct/hctdb.py
```diff
@@ -5848,9 +5848,97 @@ def UFI(name, **mappings):
...
+              "MatVecMul",
+              next_op_idx,
+              "MatVecMul",
+              "Matrix-Vector Multiply",
+              "<hfwi,<hfwi",
+              "",
+              [
+                  db_dxil_param(0, "$x0", "outputVector", "output vector"),
+                  db_dxil_param(2, "$x1", "inputVector", "input vector"),
+                  db_dxil_param(3, "i1", "isInputSigned", "input signed op kind"),
```
damyanp, line 5861:
```suggestion
                  db_dxil_param(3, "i1", "isInputUnsigned", "input signed op kind"),
```

damyanp:
This applies to all the builtins I've tried so far, but the VectorAccumulate example is quite minimal.  Given this code:

```c++
export void TruncatedVector(vector<half, 254> Input254, vector<half, 255> Input255) {
  __builtin_VectorAccumulate(Input254, RWBuf, 0);
  __builtin_VectorAccumulate(Input255, RWBuf, 0);
}```

This generates:

```llvm
; Function Attrs: nounwind
define void @"\01?TruncatedVector@@YAXV?$vector@$halff@$0PO@@@V?$vector@$halff@$0PP@@@@Z"(<254 x float> %Input254, <255 x float> %Input255) #0 {
  %1 = load %dx.types.Handle, %dx.types.Handle* @"\01?RWBuf@@3URWByteAddressBuffer@@A", align 4
  %2 = call %dx.types.Handle @dx.op.createHandleForLib.dx.types.Handle(i32 160, %dx.types.Handle %1)  ; CreateHandleForLib(Resource)
  %3 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %2, %dx.types.ResourceProperties { i32 4107, i32 0 })  ; AnnotateHandle(res,props)  resource: RWByteAddressBuffer
  call void @dx.op.vectorAccumulate.v254f32(i32 308, <254 x float> %Input254, %dx.types.Handle %3, i32 0)  ; VectorAccumulate(inputVector,arrayBuffer,arrayOffset)
  %4 = shufflevector <255 x float> %Input255, <255 x float> undef, <1 x i32> zeroinitializer
  %5 = call %dx.types.Handle @dx.op.createHandleForLib.dx.types.Handle(i32 160, %dx.types.Handle %1)  ; CreateHandleForLib(Resource)
  %6 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %5, %dx.types.ResourceProperties { i32 4107, i32 0 })  ; AnnotateHandle(res,props)  resource: RWByteAddressBuffer
  call void @dx.op.vectorAccumulate.v1f32(i32 308, <1 x float> %4, %dx.types.Handle %6, i32 0)  ; VectorAccumulate(inputVector,arrayBuffer,arrayOffset)
  ret void
}
```

Note how Input255 is explicitly truncated to 1xfloat before the vectorAccumulate is called.  Input254 is not truncated.

utils/hct/gen_intrin_main.txt
```diff
@@ -380,6 +380,14 @@ void [[]] Barrier(in NodeRecordOrUAV o, in uint SemanticFlags);
...
 
+void [[]] __builtin_MatVecMul(out numeric<c> output_vector, in bool isOutputSigned, in numeric<c2> input_vector, in bool isInputUnsigned, in uint input_interpretation, in ByteAddressBuffer matrix_buffer, in uint matrix_offset, in uint matrix_interpretation, in uint M, in uint K, in uint matrix_layout, in bool matrix_is_transposed, in uint matrix_stride);
+ 
+void [[]] __builtin_MatVecMulAdd(out numeric<c> output_vector, in bool isOutputUnsigned, in numeric<c2> input_vector, in bool isInputUnsigned, in uint input_interpretation, in ByteAddressBuffer matrix_buffer, in uint matrix_offset, in uint matrix_interpretation, in uint M, in uint K, in uint matrix_layout, in bool matrix_is_transposed, in uint matrix_stride, in ByteAddressBuffer bias_vector, in uint bias_offset, in uint bias_interpretation);
+ 
+void [[]] __builtin_OuterProductAccumulate(in numeric<c> input_vector1, in numeric<c2> input_vector2, in RWByteAddressBuffer matrix_buffer, in uint matrix_offset, in uint matrix_interpretation, in uint matrix_layout, in uint matrix_stride);
+ 
+void [[]] __builtin_VectorAccumulate(in numeric<c> input_vector, in RWByteAddressBuffer matrix_buffer, in uint matrix_offset);
+
```
damyanp, line 390: These aren't actually in the DX namespace.

See `MaybeReorderThread` (I think) for an example of how to do this.  This also has the `min_sm` stuff that would be needed to prevent these from showing up in older shader models.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -6300,6 +6300,150 @@ Value *TranslateSelect(CallInst *CI, IntrinsicOp IOP, OP::OpCode opcode,
   }
   return Builder.CreateSelect(cond, t, f);
 }
+
+Value *TranslateMatVecMul(CallInst *CI, IntrinsicOp IOP, OP::OpCode opcode,
+                          HLOperationLowerHelper &helper,
+                          HLObjectOperationLowerHelper *pObjHelper,
+                          bool &Translated) {
+
+  hlsl::OP *hlslOP = &helper.hlslOP;
```
llvm-beanz, line 6309: My biggest immediate reaction here is that all new code contributed to DXC should conform to the LLVM coding standards. While we haven't always been great about enforcing that has always been the documented standard:

https://github.com/microsoft/DirectXShaderCompiler/blob/main/CONTRIBUTING.md#coding-guidelines

This all needs clang-format applied, and also should have all the variable naming updated to follow the standards:

lib/DXIL/DxilShaderModel.cpp
```diff
@@ -79,114 +79,111 @@ const ShaderModel *ShaderModel::Get(Kind Kind, unsigned Major, unsigned Minor) {
   /* <py::lines('VALRULE-TEXT')>hctdb_instrhelp.get_shader_model_get()</py>*/
   // VALRULE-TEXT:BEGIN
   const static std::pair<unsigned, unsigned> hashToIdxMap[] = {
-      {1024, 0},    // ps_4_0
```
llvm-beanz, line 82: A lot of the diffs here are because the generated code wasn't run through clang-format, so this should clean up and dramatically simplify the patch once it is formatted correctly.

utils/hct/gen_intrin_main.txt
anupamachandra: Oh right! That wasn't checked in yet when I made these changes. Looks like it is there now.

lib/DXIL/DxilShaderModel.cpp
anupamachandra: Earlier I was able to apply codeformatting through github-actions and it used the right clang-format version, didn't work this time I guess due to the diff size. Anyway did it locally and pushed.

lib/HLSL/HLOperationLower.cpp
anupamachandra: I tried to mimic the style of the file. Anyway updated to LLVM style.

utils/hct/hctdb.py
anupamachandra: Should I make is `isInputUnsigned" and keep it consistent with the builtin name? And the values `1` and `0` are self explanatory?

damyanp: That sounds good to me - I'll get the proposal updated to match once the dust has settled.

utils/hct/hctdb.py
```diff
@@ -5848,9 +5848,97 @@ def UFI(name, **mappings):
...
+                  db_dxil_param(3, "res", "arrayBuffer", "output array resource"),
+                  db_dxil_param(4, "i32", "arrayOffset", "output array offset"),
+              ],
+          )
+        next_op_idx += 1
+
         self.set_op_count_for_version(1, 9, next_op_idx)
-        assert next_op_idx == 305, (
+        assert next_op_idx == 309, (
             "305 is expected next operation index but encountered %d and thus opcodes are broken"
```
damyanp, line 5942:
```suggestion
            "309 is expected next operation index but encountered %d and thus opcodes are broken"
```

utils/hct/gen_intrin_main.txt
damyanp: See https://github.com/microsoft/DirectXShaderCompiler/issues/7301 - we'll leave this as it is for the preview and will address the namespace for the new intrinsics later.

utils/hct/hctdb.py
```diff
@@ -5848,10 +5848,98 @@ def UFI(name, **mappings):
...
+                  db_dxil_param(16, "i1", "isOutputUnsigned", "output signed op kind"),
+              ],
+          )
+        next_op_idx += 1
+
+        self.add_dxil_op(
+              "OuterProductAccumulate",
+              next_op_idx,
+              "OuterProductAccumulate",
+              "Outer Product Accumulate",
```
tex3d, line 5908: nit: It would be nice to have better one-line descriptions for these operations.

tex3d (commented):
Besides the generated content, I think this looks good.  Just one small nit regarding DXIL Op descriptions.

I believe the generated content is out of date/incorrect, since I noticed some deleted operations and a missing `.json` file update.  In any case, generated files will need to be updated before the final PR is ready for merging.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,175 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
   }
 }
 
+static bool CheckFromRegisterInterpretations(uint32_t Ri) {
+  std::set<DXIL::ComponentType> ValidSet = {
```
damyanp, line 974: It would almost certainly be more efficient to implement this by having a static array of component types and then doing a linear search through it to see if Ri is in that array.

As it is written now (and for all of the `std::set` based ones, this is building up dynamically allocated complex data structures and throwing them away for each call.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,175 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+      DXIL::ComponentType::F8_E4M3, DXIL::ComponentType::F8_E5M2};
+
+  if (ValidSet.find(static_cast<DXIL::ComponentType>(Mi)) != ValidSet.end()) {
+    return true;
+  }
+  return false;
+}
+
+static bool CheckMatrixLayout(uint32_t Ml) {
+  std::set<DXIL::DXILMatrixLayout> ValidSet = {
```
damyanp, line 1003: For this one, since the DXILMatrixLayout values are contiguous this could just be `return Ml <= DXIL::DXILMatrixLayout::OuterProductOptimal;`

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,175 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+      DXIL::DXILMatrixLayout::OuterProductOptimal};
+
+  if (ValidSet.find(static_cast<DXIL::DXILMatrixLayout>(Ml)) !=
+      ValidSet.end()) {
+    return true;
+  }
+  return false;
+}
+
+static void ValidateImmOperandsForMatVecOps(CallInst *CI, DXIL::OpCode opcode,
```
damyanp, line 1015: Note - we should also do these sorts of checks in the builtins, not just in the validator.

lib/DxilValidation/DxilValidation.cpp
anupamachandra: yeah! I hadn't considered that even for static initialization the memory is dynamically allocated.

lib/DxilValidation/DxilValidation.cpp
anupamachandra: Yep!

lib/DxilValidation/DxilValidation.cpp
anupamachandra: Yeah, can you open an issue for it, I'll handle it in a followup.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linang-builins.hlsl
```diff
@@ -0,0 +1,55 @@
+// dxc.exe -T cs_6_9 -E cs_main %s
```
llvm-beanz, line 1: This isn't a valid `RUN` which will cause LIT to report a failure since it can't identify the test command to run. There are also no `CHECKS`, which makes me think this isn't supposed to be a test. Can you instead put this source into comments in the `ll` file so that it is tracked as being connected but won't be detected as a test file?

anupamachandra: Yeah, this was meant as a source file, but I figured I can check for the HLSL intrinisics here before lowering here. Left the HL opcodes here as they are meant to be stable.

utils/hct/hctdb.py
simoll: Catching up. This has been addressed.

lib/DxilValidation/DxilValidation.cpp
simoll: Catching up. This has been resolved.

lib/DxilValidation/DxilValidation.cpp
simoll: Catching up. This has been resolved.

simoll:
Azure pipelines are failing during cmake config:
```
CMake Error at tools/clang/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

```

tools/clang/lib/CodeGen/CGDebugInfo.cpp
```diff
@@ -1047,8 +1047,15 @@ bool CGDebugInfo::TryCollectHLSLRecordElements(const RecordType *Ty,
     unsigned VecSize = hlsl::GetHLSLVecSize(QualTy);
     unsigned ElemSizeInBits = CGM.getContext().getTypeSize(ElemQualTy);
     unsigned CurrentAlignedOffset = 0;
+    std::string FieldNameBuf;
     for (unsigned ElemIdx = 0; ElemIdx < VecSize; ++ElemIdx) {
-      StringRef FieldName = StringRef(&"xyzw"[ElemIdx], 1);
+      StringRef FieldName;
+      if (VecSize <= 4) {
```
simoll, line 1053: This fixes OOB accesses into the constant string for long vectors. Asan was reporting an error for the tools\clang\test\CodeGenDXIL\hlsl\intrinsics\linalg-builtins.hlsl test:
```
==214895==ERROR: AddressSanitizer: global-buffer-overflow on address 0x7f9d9299a905 at pc 0x7f9d8fd110be bp 0x7ffcc6fe4b40 sp 0x7ffcc6fe4b38
READ of size 1 at 0x7f9d9299a905 thread T0
```

lib/DXIL/DxilShaderModel.cpp
simoll: Resolved

lib/HLSL/HLOperationLower.cpp
simoll: Resolved

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linang-builins.hlsl
simoll: Fixed the RUN line and there are CHECKs. This is a proper test now

utils/hct/hctdb.py
damyanp:
```suggestion
                  db_dxil_param(3, "i1", "isInputSigned", "is input signed"),
```

This is repeated for other ops, also need to update the isOutputSigned descriptions as well.

damyanp: Oh, this hasn't been addressed.  It needs to be named `isInputUnsigned` and the description updated to match. Same for the output ones.

lib/DxilValidation/DxilValidation.cpp
damyanp: #7329

tools/clang/test/HLSLFileCheck/hlsl/linalg_builtins/mat-vec-mul_multioverload.hlsl
```diff
@@ -0,0 +1,33 @@
...
+// Test use of __builtin_MatVecMulAdd in compute shader
+// CHECK: define void @main()
+// CHECK:   call <4 x i32> @dx.op.matVecMul.v4i32.v8f32(i32 {{[0-9]+}}
+
+
+[NumThreads(1,1,1)]
+void main()
+{	
+	vector<uint32_t, 4> output_vector;
+	static const uint is_output_unsigned = 0;
```
tex3d, line 17: Shouldn't this be true?  Here's a case where we probably should have emitted a diagnostic.

And shouldn't we be checking that each of the values passed through faithfully to the HL IR, then the DXIL, in cases other than the one with a single set of parameters, where many of them use the same value as other similarly typed parameters?

I don't think we need exhaustive combinations, but enough to make it highly unlikely that a bug where one parameter came from wrong place could be lurking.

There are some clever tricks you can use to capture macro values for use in CHECK lines like [this example](https://github.com/microsoft/DirectXShaderCompiler/pull/7324/files#diff-0503a36fea3580600344b820fc40846d7144b8457764092592bfe905f7351108R23-R26).  That way you can just have a set of RUN lines that explore the desired variations.

It would likely help readability to have a set of defined values for enum inputs (like interpretations and matrix layout).  Then the RUN lines can have define arguments assigned to enum symbols instead of numbers, like `-D I_INTERP=F32 -D M_INTERP=F16 -D M_LAYOUT=RowMajor`.  These can be captured into FileCheck variables by writing the assigned values into specific buffer locations and matching them, like in the example.

tools/clang/test/HLSLFileCheck/hlsl/linalg_builtins/mat-vec-mul_multioverload.hlsl
```diff
@@ -0,0 +1,33 @@
+// RUN: %dxc -T cs_6_9 %s | FileCheck %s
```
tex3d, line 1: Shouldn't we check the `-fcgl` results too?

tools/clang/test/HLSLFileCheck/hlsl/linalg_builtins/mat-vec-mul_multioverload.hlsl
```diff
@@ -0,0 +1,33 @@
...
+	
+	const uint matrix_offset = 0;
+	const uint matrix_interpretation = 9; /*F32*/
+	const uint matrix_dimM = 8;
+	const uint matrix_dimK = 8;
+	const uint matrix_layout = 0; /*RowMajor*/
+	const bool matrix_is_transposed = false; 
+	const uint matrix_stride = 64;
+
+	__builtin_MatVecMul(output_vector, is_output_unsigned, input_vector, is_input_unsigned, input_interpretation, matrix_buffer, matrix_offset, matrix_interpretation,
```
tex3d, line 31: Shouldn't we also test MatVecMulAdd with multi-overloads?  That could happen in this same test to reuse common parameters and variation code.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+  llvm::Value *MatrixK =
+      CI->getOperand(DXIL::OperandIndex::kMatVecMulMatrixKIdx);
+  llvm::Value *MatrixLayout =
+      CI->getOperand(DXIL::OperandIndex::kMatVecMulMatrixLayoutIdx);
+  llvm::Value *MatrixTranspose =
+      CI->getOperand(DXIL::OperandIndex::kMatVecMulMatrixTransposeIdx);
+
+  if (!llvm::isa<llvm::Constant>(InputIsUnsigned)) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrMatVecOpIsUnsignedFlagsAreConst);
```
tex3d, line 1030: We're missing validation testing for each of these.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
   }
 }
 
+static bool CheckFromRegisterInterpretations(uint32_t input) {
```
tex3d, line 973: We should conform to coding standards.
```suggestion
static bool CheckFromRegisterInterpretations(uint32_t Input) {
```
Same with others below, `Input`, `OpCode`.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+  DXIL::ComponentType ValidSet[] = {
+      DXIL::ComponentType::I16,         DXIL::ComponentType::U16,
+      DXIL::ComponentType::I32,         DXIL::ComponentType::U32,
+      DXIL::ComponentType::F16,         DXIL::ComponentType::F32,
+      DXIL::ComponentType::PackedS8x32, DXIL::ComponentType::PackedU8x32,
+      DXIL::ComponentType::U8,          DXIL::ComponentType::I8,
+      DXIL::ComponentType::F8_E4M3,     DXIL::ComponentType::F8_E5M2};
+
+  for (auto Val : ValidSet) {
+    if (Val == static_cast<DXIL::ComponentType>(input))
```
tex3d, line 983:
```suggestion
    if (Val == static_cast<DXIL::ComponentType>(Input))
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+      return true;
+  }
+  return false;
+}
+
+static bool CheckInMemoryInterpretations(uint32_t input) {
```
tex3d, line 989:
```suggestion
static bool CheckInMemoryInterpretations(uint32_t Input) {
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+  DXIL::ComponentType ValidSet[] = {
+      DXIL::ComponentType::I16,     DXIL::ComponentType::U16,
+      DXIL::ComponentType::I32,     DXIL::ComponentType::U32,
+      DXIL::ComponentType::F16,     DXIL::ComponentType::F32,
+      DXIL::ComponentType::U8,      DXIL::ComponentType::I8,
+      DXIL::ComponentType::F8_E4M3, DXIL::ComponentType::F8_E5M2};
+
+  for (auto Val : ValidSet) {
+    if (Val == static_cast<DXIL::ComponentType>(input))
```
tex3d, line 998:
```suggestion
    if (Val == static_cast<DXIL::ComponentType>(Input))
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+      return true;
+  }
+  return false;
+}
+
+static bool CheckMatrixLayout(unsigned input) {
+  return (input <=
```
tex3d, line 1005:
```suggestion
static bool CheckMatrixLayout(unsigned Input) {
  return (Input <=
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+          static_cast<unsigned>(DXIL::DXILMatrixLayout::OuterProductOptimal));
+}
+
+static void ValidateImmOperandsForMatVecOps(CallInst *CI, DXIL::OpCode opcode,
```
tex3d, line 1009:
```suggestion
static void ValidateImmOperandsForMatVecOps(CallInst *CI, DXIL::OpCode OpCode,
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+  if (!llvm::isa<llvm::Constant>(InputInterpretation) ||
+      !llvm::isa<llvm::Constant>(MatrixInterpretation)) {
+    ValCtx.EmitInstrError(
+        CI, ValidationRule::InstrLinalgInterpretationParamAreConst);
+  }
+
+  // Check if InputInterpretation and MatrixInterpretation are valid
+  ConstantInt *Ii = cast<ConstantInt>(InputInterpretation);
+  auto IiValue = Ii->getLimitedValue();
+  if (!CheckFromRegisterInterpretations(IiValue)) {
```
tex3d, line 1042: Coding standards:
Since these are initials, they should both be capitalized.
```suggestion
  ConstantInt *II = cast<ConstantInt>(InputInterpretation);
  auto IIValue = II->getLimitedValue();
  if (!CheckFromRegisterInterpretations(IIValue)) {
```
Same with others below - `MI`, `ML`, `BI`.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+    ValCtx.EmitInstrError(
+        CI, ValidationRule::InstrLinalgInvalidRegisterInterpValue);
+  }
+
+  ConstantInt *Mi = cast<ConstantInt>(MatrixInterpretation);
+  auto MiValue = Mi->getLimitedValue();
+  if (!CheckInMemoryInterpretations(MiValue)) {
```
tex3d, line 1049:
```suggestion
  ConstantInt *MI = cast<ConstantInt>(MatrixInterpretation);
  auto MIValue = MI->getLimitedValue();
  if (!CheckInMemoryInterpretations(MIValue)) {
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+      !llvm::isa<llvm::Constant>(MatrixK) ||
+      !llvm::isa<llvm::Constant>(MatrixLayout) ||
+      !llvm::isa<llvm::Constant>(MatrixTranspose)) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgMatrixShapeParamsAreConst);
+  }
+
+  ConstantInt *Ml = cast<ConstantInt>(MatrixLayout);
+  auto MlValue = Ml->getLimitedValue();
+  if (!CheckMatrixLayout(MlValue)) {
```
tex3d, line 1064:
```suggestion
  ConstantInt *ML = cast<ConstantInt>(MatrixLayout);
  auto MLValue = ML->getLimitedValue();
  if (!CheckMatrixLayout(MLValue)) {
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+    }
+    llvm::Value *BiasInterpretation =
+        CI->getOperand(DXIL::OperandIndex::kMatVecMulAddBiasInterpretation);
+    if (!llvm::isa<llvm::Constant>(BiasInterpretation)) {
+      ValCtx.EmitInstrError(
+          CI, ValidationRule::InstrLinalgInterpretationParamAreConst);
+    }
+    ConstantInt *Bi = cast<ConstantInt>(BiasInterpretation);
+    auto BiValue = Bi->getLimitedValue();
+    if (!CheckInMemoryInterpretations(BiValue)) {
```
tex3d, line 1094:
```suggestion
    ConstantInt *BI = cast<ConstantInt>(BiasInterpretation);
    auto BIValue = BI->getLimitedValue();
    if (!CheckInMemoryInterpretations(BIValue)) {
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+  llvm::Value *MatrixLayout =
+      CI->getOperand(DXIL::OperandIndex::kOuterProdAccMatrixLayout);
+
+  if (!llvm::isa<llvm::Constant>(MatrixInterpretation)) {
+    ValCtx.EmitInstrError(
+        CI, ValidationRule::InstrLinalgInterpretationParamAreConst);
+  }
+  ConstantInt *Mi = cast<ConstantInt>(MatrixInterpretation);
+  auto MiValue = Mi->getLimitedValue();
+  if (!CheckInMemoryInterpretations(MiValue)) {
```
tex3d, line 1119:
```suggestion
  ConstantInt *MI = cast<ConstantInt>(MatrixInterpretation);
  auto MIValue = MI->getLimitedValue();
  if (!CheckInMemoryInterpretations(MIValue)) {
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+                          ValidationRule::InstrLinalgInvalidMemoryInterpValue);
+  }
+
+  if (!llvm::isa<llvm::Constant>(MatrixLayout)) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgMatrixShapeParamsAreConst);
+  }
+  ConstantInt *Ml = cast<ConstantInt>(MatrixLayout);
+  auto MlValue = Ml->getLimitedValue();
+  if (!CheckMatrixLayout(MlValue)) {
```
tex3d, line 1130:
```suggestion
  ConstantInt *ML = cast<ConstantInt>(MatrixLayout);
  auto MLValue = ML->getLimitedValue();
  if (!CheckMatrixLayout(MLValue)) {
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,169 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+    }
+  } break;
+  default:
+    break;
+  }
+}
+
+static void ValidateImmOperandsForOuterProdAcc(CallInst *CI,
+                                               DXIL::OpCode opcode,
+                                               ValidationContext &ValCtx) {
```
tex3d, line 1106: I believe `opcode` is unused, right?
```suggestion
static void ValidateImmOperandsForOuterProdAcc(CallInst *CI,
                                               ValidationContext &ValCtx) {
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -1942,6 +2105,16 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
                                  GetLaunchTypeStr(nodeLaunchType)});
 
     break;
+  case DXIL::OpCode::MatVecMul:
+  case DXIL::OpCode::MatVecMulAdd:
+    ValidateImmOperandsForMatVecOps(CI, opcode, ValCtx);
+    break;
+  case DXIL::OpCode::OuterProductAccumulate:
+    ValidateImmOperandsForOuterProdAcc(CI, opcode, ValCtx);
```
tex3d, line 2113:
```suggestion
    ValidateImmOperandsForOuterProdAcc(CI, ValCtx);
```

include/dxc/DXIL/DxilConstants.h
```diff
@@ -163,23 +163,31 @@ const unsigned kDxilMaxOloadDims = 2;
...
   Invalid = 0,
   I1,
-  I16,
-  U16,
-  I32,
-  U32,
+  I16, // = 2
+  U16, // = 3
+  I32, // = 4
+  U32, // = 5
```
tex3d, line 169: Why don't we just explicitly assign all the values in this enum to make it clear that it must remain stable.

simoll: Make those numbers explicit. Resolved.

lib/DxilValidation/DxilValidation.cpp
simoll: Wasn't able to commit this suggestions but incorporated the change manually. Resolved.

utils/hct/hctdb.py
simoll: Changed param names and descriptions as suggested (could not apply from github). Resolved.

tools/clang/lib/CodeGen/CGDebugInfo.cpp
simoll: Pulled out this fix into #7332 and reverted the change here.
This means this PR will see get asan test failures in the Azure Pipelines until #7332 lands

tools/clang/test/HLSLFileCheck/hlsl/linalg_builtins/check-shader-stages.hlsl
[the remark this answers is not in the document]
simoll: Moved tests as suggested and added `lit.local.cfg`. Resolved.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg-builtins.ll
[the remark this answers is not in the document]
simoll: Moved test. Resolved.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg-builtins.hlsl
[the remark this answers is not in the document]
simoll: Did so. Resolved.

tools/clang/test/HLSLFileCheck/hlsl/linalg_builtins/check-shader-stages.hlsl
```diff
@@ -0,0 +1,298 @@
...
+	
+	const uint matrix_offset = 0;
+	const uint matrix_interpretation = 9; /*F32*/
+	const uint matrix_dimM = 4;
+	const uint matrix_dimK = 4;
+	const uint matrix_layout = 0; /*RowMajor*/
+	const bool matrix_is_transposed = false; 
+	const uint matrix_stride = 64;
+
+	__builtin_MatVecMul(output_vector, is_output_unsigned, input_vector,
```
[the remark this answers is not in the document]
simoll: Moved common code into `UseCoopVec` function. Resolved.

tools/clang/test/HLSLFileCheck/hlsl/linalg_builtins/mat-vec-mul_multioverload.hlsl
simoll: Added variations. I used the minimum supported set as a starting point and added some unsigned variants. Resolved.

tools/clang/test/HLSLFileCheck/hlsl/linalg_builtins/mat-vec-mul_multioverload.hlsl
simoll: Added the same test for matVecMuladd, including variations, hlops and unsigned versions. Resolved.

tools/clang/test/HLSLFileCheck/hlsl/linalg_builtins/mat-vec-mul_multioverload.hlsl
```diff
@@ -0,0 +1,33 @@
...
+
+// Test use of __builtin_MatVecMulAdd in compute shader
+// CHECK: define void @main()
+// CHECK:   call <4 x i32> @dx.op.matVecMul.v4i32.v8f32(i32 {{[0-9]+}}
+
+
+[NumThreads(1,1,1)]
+void main()
+{	
+	vector<uint32_t, 4> output_vector;
```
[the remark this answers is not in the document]
simoll: Yes, if i read the spec correctly, Data is transformed in and out of the internal representation, which is what is actually used for the operation. So you can have f32 -> u8 -> i16 for example

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg-builtins.hlsl
```diff
@@ -0,0 +1,59 @@
...
+	
+	const uint matrix_offset = 0;
+	const uint matrix_interpretation = 9; /*F32*/
+	const uint matrix_dimM = 4;
+	const uint matrix_dimK = 4;
+	const uint matrix_layout = 0; /*RowMajor*/
+	const bool matrix_is_transposed = false; 
+	const uint matrix_stride = 64;
+
+	// CHECK: call void @"dx.hl.op..void (i32, <4 x float>*, i1, <4 x float>, i1, i32, %dx.types.Handle, i32, i32, i32, i32, i32, i1, i32)"(i32 363,
```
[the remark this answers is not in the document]
tex3d: I think this comment got lost when the file was updated.  I'll add a new one.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/linalg-builtins.hlsl
```diff
@@ -0,0 +1,59 @@
...
+	
+	const uint matrix_offset = 0;
+	const uint matrix_interpretation = 9; /*F32*/
+	const uint matrix_dimM = 4;
+	const uint matrix_dimK = 4;
+	const uint matrix_layout = 0; /*RowMajor*/
+	const bool matrix_is_transposed = false; 
+	const uint matrix_stride = 64;
+
+	// CHECK: call void @"dx.hl.op..void (i32, <4 x float>*, i1, <4 x float>, i1, i32, %dx.types.Handle, i32, i32, i32, i32, i32, i1, i32)"(i32 363,
```
tex3d, line 26: Shouldn't these CHECKs check the arguments?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11941,6 +11941,18 @@ void Sema::DiagnoseReachableHLSLCall(CallExpr *CE, const hlsl::ShaderModel *SM,
...
     DiagnoseReachableSERCall(*this, CE, EntrySK, EntryDecl, true);
+    break;
+  case hlsl::IntrinsicOp::IOP___builtin_MatVecMul:
+  case hlsl::IntrinsicOp::IOP___builtin_MatVecMulAdd:
+  case hlsl::IntrinsicOp::IOP___builtin_OuterProductAccumulate:
+  case hlsl::IntrinsicOp::IOP___builtin_VectorAccumulate:
+    if (!SM->IsSM69Plus()) {
+      Diags.Report(CE->getExprLoc(),
+                   diag::warn_hlsl_intrinsic_in_wrong_shader_model)
+          << FD->getNameAsString() << EntryDecl->getNameAsString() << "6.9";
```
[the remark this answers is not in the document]
tex3d: If this is planned as a follow up, you can link the issue here and resolve the comment.

lib/DxilValidation/DxilValidation.cpp
tex3d: Go ahead and resolve this comment after you link the follow-up issue.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
simoll: Tracking task to create CoopVec dxil tests with validation errors in #7335
Resolved.

lib/DxilValidation/DxilValidation.cpp
simoll: Tracking task to create CoopVec dxil tests with validation errors in #7335
Resolved.

simoll:
#7336 tracks the outstanding task to diagnose coop vec issues on hlsl level. Will address in a followup. Request for sema/diagnostics resolved for this PR.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+
+static bool CheckUnsignedFlag(Type *VecTy, bool IsUnsigned) {
+  Type *ElemTy = VecTy->getScalarType();
+  if (ElemTy->isFloatingPointTy())
+    return !IsUnsigned;
+
+  return true;
+}
+
+static Value *getMatVecOpIsOutputUnsigned(CallInst *CI, DXIL::OpCode OpCode) {
```
llvm-beanz, line 1029: nit: All the other functions you've added are PascalCase, this one is camelCase. The coding conventions are camelCase, but if the rest of the file is all PascalCase we should match.
(see: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly)
```suggestion
static Value *GetMatVecOpIsOutputUnsigned(CallInst *CI, DXIL::OpCode OpCode) {
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+  ConstantInt *II = dyn_cast<ConstantInt>(InputInterpretation);
+  ConstantInt *MI = dyn_cast<ConstantInt>(MatrixInterpretation);
+  if (!II || !MI) {
+    ValCtx.EmitInstrError(
+        CI, ValidationRule::InstrLinalgInterpretationParamAreConst);
+    return;
+  }
+
+  // Check if InputInterpretation and MatrixInterpretation are valid
+  auto IIValue = II->getLimitedValue();
```
llvm-beanz, line 1087: nit: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable
```suggestion
  uint64_t IIValue = II->getLimitedValue();
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+  if (!CheckFromRegisterInterpretations(IIValue)) {
+    ValCtx.EmitInstrError(
+        CI, ValidationRule::InstrLinalgInvalidRegisterInterpValue);
+    return;
+  }
+
+  auto MIValue = MI->getLimitedValue();
```
llvm-beanz, line 1094: nit: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable
```suggestion
  uint64_t MIValue = MI->getLimitedValue();
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+  ConstantInt *MatrixLayoutConst = dyn_cast<ConstantInt>(MatrixLayout);
+  if (!llvm::isa<llvm::Constant>(MatrixM) ||
+      !llvm::isa<llvm::Constant>(MatrixK) || !MatrixLayoutConst ||
+      !MatrixTransposeConst) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgMatrixShapeParamsAreConst);
+    return;
+  }
+
+  auto MLValue = MatrixLayoutConst->getLimitedValue();
```
llvm-beanz, line 1111: nit: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable
```suggestion
  uint64_t MLValue = MatrixLayoutConst->getLimitedValue();
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+  case DXIL::OpCode::MatVecMulAdd: {
+    llvm::Value *BiasInterpretation =
+        CI->getOperand(DXIL::OperandIndex::kMatVecMulAddBiasInterpretation);
+    if (!llvm::isa<llvm::Constant>(BiasInterpretation)) {
+      ValCtx.EmitInstrError(
+          CI, ValidationRule::InstrLinalgInterpretationParamAreConst);
+      return;
+    }
+    ConstantInt *BI = cast<ConstantInt>(BiasInterpretation);
+    auto BIValue = BI->getLimitedValue();
```
llvm-beanz, line 1150: nit: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable
```suggestion
    uint64_t BIValue = BI->getLimitedValue();
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+
+  llvm::Value *MatrixInterpretation =
+      CI->getOperand(DXIL::OperandIndex::kOuterProdAccMatrixInterpretation);
+  llvm::Value *MatrixLayout =
+      CI->getOperand(DXIL::OperandIndex::kOuterProdAccMatrixLayout);
+
+  if (!llvm::isa<llvm::Constant>(MatrixInterpretation)) {
+    ValCtx.EmitInstrError(
+        CI, ValidationRule::InstrLinalgInterpretationParamAreConst);
+  }
```
llvm-beanz, line 1173: nit: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
  if (!llvm::isa<llvm::Constant>(MatrixInterpretation))
    ValCtx.EmitInstrError(
        CI, ValidationRule::InstrLinalgInterpretationParamAreConst);
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+  ConstantInt *MI = cast<ConstantInt>(MatrixInterpretation);
+  auto MIValue = MI->getLimitedValue();
+  if (!CheckInMemoryInterpretations(MIValue)) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgInvalidMemoryInterpValue);
+  }
```
llvm-beanz, line 1179: nit: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
  if (!CheckInMemoryInterpretations(MIValue))
    ValCtx.EmitInstrError(CI,
                          ValidationRule::InstrLinalgInvalidMemoryInterpValue);
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+
+  if (!llvm::isa<llvm::Constant>(MatrixLayout)) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgMatrixShapeParamsAreConst);
+  }
```
llvm-beanz, line 1184: nit: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
  if (!llvm::isa<llvm::Constant>(MatrixLayout))
    ValCtx.EmitInstrError(CI,
                          ValidationRule::InstrLinalgMatrixShapeParamsAreConst);
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+      CI->getOperand(DXIL::OperandIndex::kOuterProdAccMatrixInterpretation);
+  llvm::Value *MatrixLayout =
+      CI->getOperand(DXIL::OperandIndex::kOuterProdAccMatrixLayout);
+
+  if (!llvm::isa<llvm::Constant>(MatrixInterpretation)) {
+    ValCtx.EmitInstrError(
+        CI, ValidationRule::InstrLinalgInterpretationParamAreConst);
+  }
+  ConstantInt *MI = cast<ConstantInt>(MatrixInterpretation);
+  auto MIValue = MI->getLimitedValue();
```
llvm-beanz, line 1175: nit: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable
```suggestion
  uint64_t MIValue = MI->getLimitedValue();
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgInvalidMemoryInterpValue);
+  }
+
+  if (!llvm::isa<llvm::Constant>(MatrixLayout)) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgMatrixShapeParamsAreConst);
+  }
+  ConstantInt *ML = cast<ConstantInt>(MatrixLayout);
+  auto MLValue = ML->getLimitedValue();
```
llvm-beanz, line 1186: nit: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable
```suggestion
  uint64_t MLValue = ML->getLimitedValue();
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,226 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+  if (!CheckMatrixLayout(MLValue)) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgInvalidMatrixLayoutValue);
+  }
```
llvm-beanz, line 1190: nit: https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
```suggestion
  if (!CheckMatrixLayout(MLValue))
    ValCtx.EmitInstrError(CI,
                          ValidationRule::InstrLinalgInvalidMatrixLayoutValue);
```

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/check-shader-stages.hlsl
```diff
@@ -0,0 +1,131 @@
+// RUN: %dxc -T lib_6_9 %s | FileCheck %s
+ 
+ByteAddressBuffer matrix_buffer;
+ByteAddressBuffer bias_buffer;
+RWByteAddressBuffer rw_matrix_buffer;
+
+void UseCoopVec() {
+	vector<float, 4> output_vector;
```
llvm-beanz, line 8: OMG these tests have _tabs_!
_gasp_
/me: clutching pearls

Seriously though, can you do a replace all `/t` with `  `?

We don't really apply the coding standards to tests, but if we're not actually testing the ability of the compiler to parse odd things it is nice to have at least some consistency of style.

llvm-beanz (commented):
Some style nits (and shock and awe), but otherwise I think this is mostly good.

I usually don't like separating implementation from testing, but if @tex3d is okay having the validator tests come in a separate PR I'm fine with it too.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/check-shader-stages.hlsl
simoll: Yep, replaced all "`\t`" with "`    `". Resolved.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/linalg-builtins.hlsl
```diff
@@ -0,0 +1,59 @@
...
+	const uint is_input_unsigned = 0;
+	const uint input_interpretation = 9; /*F32*/
+	
+	const uint matrix_offset = 0;
+	const uint matrix_interpretation = 9; /*F32*/
+	const uint matrix_dimM = 4;
+	const uint matrix_dimK = 4;
+	const uint matrix_layout = 0; /*RowMajor*/
+	const bool matrix_is_transposed = false; 
+	const uint matrix_stride = 64;
```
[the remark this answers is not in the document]
simoll: `:%s/\t/    /g`

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/mat-vec-mul-add.hlsl
```diff
@@ -0,0 +1,35 @@
...
+
+// Test use of __builtin_MatVecMulAdd in compute shader
+// CHECK: define void @main()
+// CHECK:  call <4 x float> @dx.op.matVecMulAdd.v4f32.v4f32(i32 {{[0-9]+}}, <4 x float> {{%[0-9]+}}, i1 false, i32 9, %dx.types.Handle {{%[0-9]+}}, i32 0, i32 9, i32 4, i32 4, i32 0, i1 false, i32 64, %dx.types.Handle {{%[0-9]+}}, i32 0, i32 9, i1 false)
+
+
+[NumThreads(1,1,1)]
+void main()
+{	
+	vector<float, 4> output_vector;
```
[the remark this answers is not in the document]
simoll: `:%s/\t/    /g`

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/mat-vec-mul-add_multioverload.hlsl
```diff
@@ -0,0 +1,106 @@
...
+  RowMajor = 0,
+  ColumnMajor = 1,
+  MulOptimal = 2,
+  OuterProductOptimal = 3,
+};
+
+[NumThreads(1,1,1)]
+void main()
+{	
+	vector<OTY, 4> output_vector;
```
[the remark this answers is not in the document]
simoll: `:%s/\t/    /g`

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/mat-vec-mul_multioverload.hlsl
```diff
@@ -0,0 +1,102 @@
...
+  RowMajor = 0,
+  ColumnMajor = 1,
+  MulOptimal = 2,
+  OuterProductOptimal = 3,
+};
+
+[NumThreads(1,1,1)]
+void main()
+{	
+	vector<OTY, 4> output_vector;
```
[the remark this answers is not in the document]
simoll: `:%s/\t/    /g`

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/outer-product-accumulate-multioverload.hlsl
```diff
@@ -0,0 +1,21 @@
...
+
+RWByteAddressBuffer matrix_buffer;
+
+// CHECK: define void @main()
+// CHECK: call void @dx.op.outerProductAccumulate.v2i32.v4i32(i32 {{[0-9]+}}
+
+[Numthreads(1,1,1)]
+void main()
+{
+	vector<uint, 2> input_vector1 = 1;
```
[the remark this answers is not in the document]
simoll: `:%s/\t/    /g`

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/outer-product-accumulate.hlsl
```diff
@@ -0,0 +1,21 @@
...
+
+RWByteAddressBuffer matrix_buffer;
+
+// CHECK: define void @main()
+// CHECK:  call void @dx.op.outerProductAccumulate.v2i32.v2i32(i32 {{[0-9]+}}, <2 x i32> <i32 1, i32 1>, <2 x i32> <i32 2, i32 2>, %dx.types.Handle {{%[0-9]+}}, i32 0, i32 5, i32 0, i32 64)
+
+[Numthreads(1,1,1)]
+void main()
+{
+	vector<uint, 2> input_vector1 = 1;
```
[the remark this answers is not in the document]
simoll: `:%s/\t/    /g`

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/vector-accumulate.hlsl
```diff
@@ -0,0 +1,16 @@
...
+RWByteAddressBuffer matrix_buffer;
+
+// Test use of __builtin_VectorAccumulate in compute shader
+// CHECK: define void @main()
+// CHECK: call void @dx.op.vectorAccumulate.v2i32(i32 {{[0-9]+}}, <2 x i32> <i32 5, i32 5>, %dx.types.Handle {{%[0-9]+}}, i32 0)
+
+[NumThreads(1,1,1)]
+void main()
+{
+	vector<uint, 2> input_vector1 = 5;
```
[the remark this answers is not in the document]
simoll: `:%s/\t/    /g`

tools/clang/test/SemaHLSL/hlsl/linalg/unavailable-pre-sm69.hlsl
```diff
@@ -0,0 +1,59 @@
...
+ 
+ByteAddressBuffer matrix_buffer;
+ByteAddressBuffer bias_buffer;
+RWByteAddressBuffer rw_matrix_buffer;
+
+[Shader("compute")]
+[Numthreads(1,1,1)]
+void cs_main()
+{	
+	vector<float, 4> output_vector;
```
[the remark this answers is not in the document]
simoll: `:%s/\t/    /g`

simoll:
I hope i replaced all the `\t` this time around

> This still needs to be addressed. (Couldn't reply on https://github.com/microsoft/DirectXShaderCompiler/pull/7290?notification_referrer_id=NT_kwDOALxkNrQxNTYzMzc5ODc1OToxMjM0NjQyMg#discussion_r2042624415 directly

Added explicit parameter checks, in particular for the handles.
Back to @llvm-beanz

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -1955,6 +2171,16 @@ static void ValidateDxilOperationCallInProfile(CallInst *CI,
...
     break;
+  case DXIL::OpCode::MatVecMul:
+  case DXIL::OpCode::MatVecMulAdd:
+    ValidateImmOperandsForMatVecOps(CI, Opcode, ValCtx);
+    break;
+  case DXIL::OpCode::OuterProductAccumulate:
+    ValidateImmOperandsForOuterProdAcc(CI, ValCtx);
+    break;
+  case DXIL::OpCode::VectorAccumulate:
+
```
tex3d, line 2182: Is it TBD to validate the resource handle type, or that required dimensions are consistent with argument values, or anything else?

One thing I noticed is that the spec doesn't have any details under the `Validation Changes` section, but does have points for validating just the input/matrix/bias interpretation args for each op.  It doesn't include some additional argument validation already implemented here.

Is there a spec update coming to fill out more validation plans?  Is there an issue tracking this?

utils/hct/hctdb.py
```diff
@@ -5850,6 +5855,97 @@ def UFI(name, **mappings):
         )
         next_op_idx += 1
 
+        self.add_dxil_op(
+            "MatVecMul",
+            next_op_idx,
+            "MatVecMul",
+            "Multiplies a MxK dimension matrix and a K sized input vector",
+            "<hfwi,<hfwi",
+            "",
```
tex3d, line 5864: Isn't this operation read-only with respect to memory referenced through the resource?
```suggestion
            "ro",
```

utils/hct/hctdb.py
```diff
@@ -5850,6 +5855,97 @@ def UFI(name, **mappings):
...
+        )
+        next_op_idx += 1
+
+        self.add_dxil_op(
+            "MatVecMulAdd",
+            next_op_idx,
+            "MatVecMulAdd",
+            "multiplies a MxK dimension matrix and a K sized input vector and adds an M-sized bias vector",
+            "<hfwi,<hfwi",
+            "",
```
tex3d, line 5889: Isn't this operation read-only with respect to memory referenced through the resource?
```suggestion
            "ro",
```

utils/hct/hctdb.py
```diff
@@ -7803,6 +7899,47 @@ def build_valrules(self):
...
+        )
+
+        self.add_valrule(
+            "Instr.LinalgMatrixLayoutNotTransposable",
+            "Matrix Layout not transposable",
+        )
+
+        self.add_valrule(
+            "Instr.LinalgNotAnUnsignedType",
+            "Unsigned flag set for signed type",
```
tex3d, line 7940: I think the validation descriptions/messages aren't ideal.  We could update these in a follow-up maybe.

Ideally, we pass two strings in place of this one.  The first is the description of the rule, usually stated in a positive way, like: `"Unsigned flag corresponding to float typed parameter must be false"`

The next string is used as an error string by validation, and can have arguments, like:
`"Unsigned flag is true when corresponding %select{input|output}0 vector is float"`

If the second string is not supplied (like for these cases), the error message defaults to the description string.

tex3d (commented):
A few comments and questions.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/linalg-builtins.hlsl
```diff
@@ -0,0 +1,74 @@
...
+    const uint matrix_dimM = 4;
+    const uint matrix_dimK = 4;
+    const uint matrix_layout = 0; /*RowMajor*/
+    const bool matrix_is_transposed = false; 
+    const uint matrix_stride = 64;
+
+    // CHECK: %[[MLD0:[^ ]+]] = load %struct.ByteAddressBuffer, %struct.ByteAddressBuffer* @"\01?matrix_buffer@@3UByteAddressBuffer@@A"
+    // CHECK: %[[MCH0:[^ ]+]] = call %dx.types.Handle @"dx.hl.createhandle..%dx.types.Handle (i32, %struct.ByteAddressBuffer)"(i32 0, %struct.ByteAddressBuffer %[[MLD0]])
+    // CHECK: %[[MAH0:[^ ]+]] = call %dx.types.Handle @"dx.hl.annotatehandle..%dx.types.Handle (i32, %dx.types.Handle, %dx.types.ResourceProperties, %struct.ByteAddressBuffer)"(i32 14, %dx.types.Handle %[[MCH0]], %dx.types.ResourceProperties { i32 11, i32 0 }, %struct.ByteAddressBuffer undef)
+    // CHECK: call void @"dx.hl.op..void (i32, <4 x float>*, i1, <4 x float>, i1, i32, %dx.types.Handle, i32, i32, i32, i32, i32, i1, i32)"(i32 363, <4 x float>* %{{[^ ]+}}, i1 false, <4 x float> %{{[^ ]+}}, i1 false, i32 9, %dx.types.Handle %[[MAH0]], i32 0, i32 9, i32 4, i32 4, i32 0, i1 false, i32 64)
```
tex3d, line 29: Ultimately, I think the `undef` input vector will need to be updated to something valid, since we would probably want Sema diagnostics for an uninitialized vector input.

We should also validate undef vector input in DxilValidation.

We should just be sure to capture these points into the proposal or an issue before resolving this comment.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/linalg-builtins.hlsl
simoll: This has been addressed, see https://github.com/microsoft/DirectXShaderCompiler/pull/7290/files#diff-ade942e9326b532aaa48b48b208b236d6de1227c542da1010b67f46a957d0df4
(Clicking on the file name in the thread title got me to a stale version of that file)

tools/clang/test/DXC/Passes/DxilGen/linalg-builtins.ll
```diff
@@ -0,0 +1,127 @@
...
+define void @cs_main() #0 {
+entry:
+  %output_vector = alloca <4 x float>, align 4
+  %tmp = bitcast <4 x float>* %output_vector to i8*, !dbg !18 ; line:9 col:2
+  call void @llvm.lifetime.start(i64 16, i8* %tmp) #0, !dbg !18 ; line:9 col:2
+  %tmp1 = load %struct.ByteAddressBuffer, %struct.ByteAddressBuffer* @"\01?matrix_buffer@@3UByteAddressBuffer@@A", !dbg !22 ; line:24 col:2
+  %tmp2 = call %dx.types.Handle @"dx.hl.createhandle..%dx.types.Handle (i32, %struct.ByteAddressBuffer)"(i32 0, %struct.ByteAddressBuffer %tmp1), !dbg !22 ; line:24 col:2
+  %tmp3 = call %dx.types.Handle @"dx.hl.annotatehandle..%dx.types.Handle (i32, %dx.types.Handle, %dx.types.ResourceProperties, %struct.ByteAddressBuffer)"(i32 14, %dx.types.Handle %tmp2, %dx.types.ResourceProperties { i32 11, i32 0 }, %struct.ByteAddressBuffer zeroinitializer), !dbg !22 ; line:24 col:2
+
+  ;CHECK: call <4 x float> @dx.op.matVecMul.v4f32.v4f32(i32 305, <4 x float> undef, i1 false, i32 9, %dx.types.Handle %1, i32 0, i32 9, i32 4, i32 4, i32 0, i1 false, i32 64, i1 false)
```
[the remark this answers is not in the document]
llvm-beanz: This still needs to be addressed.

tools/clang/test/DXC/Passes/DxilGen/linalg-builtins.ll
```diff
@@ -0,0 +1,127 @@
...
+
+  call void @"dx.hl.op..void (i32, <4 x float>*, i1, <4 x float>, i1, i32, %dx.types.Handle, i32, i32, i32, i32, i32, i1, i32)"(i32 363, <4 x float>* %output_vector, i1 false, <4 x float> undef, i1 false, i32 9, %dx.types.Handle %tmp3, i32 0, i32 9, i32 4, i32 4, i32 0, i1 false, i32 64), !dbg !22 ; line:24 col:2
```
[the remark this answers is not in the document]
llvm-beanz: This one too.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -970,6 +970,222 @@ static void ValidateImmOperandForMathDxilOp(CallInst *CI, DXIL::OpCode Opcode,
...
+  if (!llvm::isa<llvm::Constant>(MatrixM) ||
+      !llvm::isa<llvm::Constant>(MatrixK) || !MatrixLayoutConst ||
+      !MatrixTransposeConst) {
+    ValCtx.EmitInstrError(CI,
+                          ValidationRule::InstrLinalgMatrixShapeParamsAreConst);
+    return;
+  }
+
+  uint64_t MLValue = MatrixLayoutConst->getLimitedValue();
+  if (!CheckMatrixLayout(MLValue)) {
```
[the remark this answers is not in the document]
anupamachandra: The MatVecMul* ops can accept both the Optimal layouts, OuterProductAccumulate can only accept TrainingOptimal. Updating the Validator to enforce that.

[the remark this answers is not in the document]
damyanp: Ohhhh.....that maybe means we need to reconsider this change to the spec: https://github.com/microsoft/hlsl-specs/pull/468

The overly generic "TrainingOptimal" and "InferenceOptimal" names were confusing people, so we renamed them to correspond to the operations they work with.  It seems that the spec wasn't clear about this, and we've now got names that are confusing in a different way.

lib/DxilValidation/DxilValidation.cpp
anupamachandra: Created an issue: https://github.com/microsoft/hlsl-specs/issues/491
Let me know if you had any specific ones in mind. I'll also go through the spec and update the issue with more specifics.

utils/hct/hctdb.py
anupamachandra: For this specific case, just to stay consistent with the error messages for SignedInt and float, should we just have the checking in sema and skip the DXIL Validation?

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/linalg-builtins.hlsl
anupamachandra: Updated in Issue: https://github.com/microsoft/hlsl-specs/issues/491

utils/hct/hctdb.py
tex3d: As we discussed, we should still have the validation.

anupamachandra: Yep, left it there and updated per feedback.

utils/hct/hctdb.py
```diff
@@ -8303,6 +8405,55 @@ def build_valrules(self):
...
+        self.add_valrule_msg(
+            "Instr.LinalgInvalidMatrixLayoutValueForMatVecOps",
+            "Matrix Layout for Linalg Mul/MulAdd operation must be valid.",
+            "Matrix Layout value '%0' is not valid. Must be between [%1 - %2].",
+        )
+
+        self.add_valrule_msg(
+            "Instr.LinalgMatrixLayoutNotTransposable",
+            "Row Major and Column Major matrix layouts are not transposable.",
+            "'%0' matrix layout is not transposable.",
```
tex3d, line 8448: While I don't want to hold up this PR, we normally expect a `%select{...}0` for the string layout here instead of using a function to select the string in the cpp code.

tex3d: Unfortunately the diagnostic system we use in validation can't handle this, so I resolved this.

tex3d (approved):

tex3d:
Just an FYI:
Our gcc pipelines started failing today because of a docker image update that updates Ubuntu ([will be ultimately to Ubunutu v24.4 by 5/9](https://github.com/actions/runner-images/issues/11766)), which will require us to bump our gcc version a few.  A fix is in the works, but in the meantime, I think we should override this failure and merge if there are no other failures, once the other pipelines are complete.

tex3d (approved):

PR #7292, opened 2025-04-01
[SM6.9] Native vector load/store lowering

pow2clk:
Enables the declaration of long vector types for raw buffers, the lowering of those
and traditional vectors in loads and stores maintaining the native types with
new dxil ops along with validation and testing support of the same.

Allow declaring long vector rawbuffer resources.
Previously disallowed along with other global types, this provides a
mechanism for indicating which buffers are raw and allowing them to
contain long vectors, continuing to produce an error for other resource
types verified by existing tests

Introduce native vector DXIL load/store intrinsics.
Add new raw buffer vector load/store intrinsics using the new vector overload types.
Include them in validation associated with similar load/stores

Lower native vector raw buffers load/stores into new ops.
When the loaded/stored type is a vector of more than 1 element, the
shader model is 6.9 or higher, and the operation is on a raw buffer,
enable the generation of a native vector raw buffer load or store.

Incidental removal of unused parameter in load translation and some refactoring
of the lowering to flow better with the new resret types.

add validation and compute shader tests

Vector to scalar raw buffer load lowering pass
Native vector loads and stores are generated for 6.9 targets and above.
This includes the 6.x target used when compiling to libraries. This adds
a pass run when linking that will lower the vector operations to scalar
operations for shader models that don't have native vector support. This
allows libraries compiled for supportive shader models to be linked to
targets without support.


pow2clk:
REVIEW GUIDANCE! The first four individual commits were designed to be reviewed individually and might make the review more pleasant. That said, I don't expect the full review would be too difficult either.

pow2clk:
I clang-formatted this to within an inch of its life. No idea why it didn't fail until now.

bogner (dismissed):
A couple of minor things but this LGTM

utils/hct/hctdb.py
```diff
@@ -5776,6 +5778,83 @@ def UFI(name, **mappings):
...
+            counters=("tex_store",),
+        )
+        next_op_idx += 1
+
+        # End of DXIL 1.9 opcodes.
+        self.set_op_count_for_version(1, 9, next_op_idx)
+        assert next_op_idx == 305, (
+            "305 is expected next operation index but encountered %d and thus opcodes are broken"
+            % next_op_idx
+        )
```
tex3d, line 5856: Can we please comment this out until DXIL 1.9 is finalized?
```suggestion
        # End of DXIL 1.9 opcodes.
        # Update and uncomment when DXIL 1.9 opcodes are finalized:
        # self.set_op_count_for_version(1, 9, next_op_idx)
        # assert next_op_idx == NNN, (
        #     "NNN is expected next operation index but encountered %d and thus opcodes are broken"
        #     % next_op_idx
        # )
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15071,7 +15083,8 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
   }
 
   // Disallow long vecs from $Global cbuffers.
-  if (isGlobal && !isStatic && !isGroupShared) {
+  if (isGlobal && !isStatic && !isGroupShared &&
+      !IS_BASIC_RAWBUFFER(basicKind)) {
```
tex3d, line 15087: The fact that you need this makes me think something else isn't detecting correctly.  We should not be looking at any objects that don't go into the constant buffer at all.  I think the property look up to use is `IS_BASIC_OBJECT`.

Note: when looking at the property table, I noticed some erroneous entries (not from your change).  I filed: #7293 for that follow-up.
```suggestion
      !IS_BASIC_OBJECT(basicKind)) {
```

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+public:
+  static char ID; // Pass identification, replacement for typeid
+  explicit DxilScalarizeVectorLoadStores() : ModulePass(ID) {}
+
+  StringRef getPassName() const override {
+    return "DXIL scalarize vector load/stores";
+  }
+
+  bool runOnModule(Module &M) override {
```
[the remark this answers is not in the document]
pow2clk: the member dxilmodule was actually unnecessary and I'd meant to remove it. Most of the simplification of runOnModule you've added here would just push the iteration down to the scalarization function. I don't mind where it goes, but the benefit, small though it is, of iterating in runOnModule is I only have to iterate through the functions once. 

I'm perfectly willing to consider it, but I think I remain unclear on the benefit of having separate classes.

tools/clang/lib/Sema/SemaHLSL.cpp
pow2clk: Sure enough! This simplifies things.

pow2clk: I did see a lot of incorrect bits that I refrained from addressing for the time being. Many of them are never checked.

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
```diff
@@ -0,0 +1,239 @@
...
+  SmallVector<Value *, 10> Args;
+  Args.emplace_back(opArg);                     // opcode @0.
+  Args.emplace_back(VecLd.get_buf());           // Resource handle @1.
+  Args.emplace_back(VecLd.get_index());         // Index @2.
+  Args.emplace_back(VecLd.get_elementOffset()); // Offset @3.
+  Args.emplace_back(nullptr);                   // Mask to be set later @4.
+  Args.emplace_back(VecLd.get_alignment());     // Alignment @5.
+
+  // Set offset to increment depending on whether the real offset is defined.
+  unsigned OffsetIdx = 0;
```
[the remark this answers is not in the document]
pow2clk: Makes sense.

utils/hct/hctdb.py
pow2clk: Done

lib/HLSL/HLOperationLower.cpp
```diff
@@ -4162,18 +4179,19 @@ Value *TranslateBufLoad(ResLoadHelper &helper, HLResource::Kind RK,
...
 
-  // 64-bit types are stored as int32 pairs in typed buffers.
+  // Adjust number of components as needed.
   if (is64 && isTyped) {
+    // 64-bit types are stored as int32 pairs in typed buffers.
     DXASSERT(NumComponents <= 2, "Typed buffers only allow 4 dwords.");
     NumComponents *= 2;
+  } else if (opcode == OP::OpCode::RawBufferVectorLoad) {
+    // Native vector loads only have a single vector element in ResRet.
+    EltTy = VectorType::get(EltTy, NumComponents);
```
tex3d, line 4189: This has the effect of subtly changing the alignment of vector load/store operations.  I do not think this is intended.

`GetBufLoadArgs` uses `min(alignment, LdSize)` for the alignment, where `LdSize` is now initialized to the size of the vector, instead of the size of the vector element.

tex3d: Here are a couple example vectors where the alignment would change:
- `float16_t2` or `float16_t3` in `ByteAddressBuffer` will align by 4 instead of 2.
- `float16_t4` in `StructuredBuffer` will align by 8 instead of 2.
- `float2` in `StructuredBuffer` will align by 8 instead of 4.

Test appears to have evidence of this change when using the new vector op.

tex3d: BTW: I think the store may be ok in this regard.  This only appears to impact the load.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-load-stores-sm69.hlsl
```diff
@@ -0,0 +1,91 @@
...
+// RUN: %dxc -DTYPE=double   -DNUM=32 -T vs_6_9 %s | FileCheck %s
+
+///////////////////////////////////////////////////////////////////////
+// Test codegen for various load and store operations and conversions
+//  for different scalar/vector buffer types and indices.
+///////////////////////////////////////////////////////////////////////
+
+// CHECK: %dx.types.ResRet.[[VTY:v[0-9]*[a-z][0-9][0-9]]] = type { <[[NUM:[0-9]*]] x [[TYPE:[a-z_0-9]*]]>, i32 }
+
+  ByteAddressBuffer RoByBuf : register(t1);
```
llvm-beanz, line 18: take or leave: the leading whitespace bothers me greatly... I know we don't really have a policy on formatting HLSL but this reminds me too much of the objective-c tendency to align colons that I've always found grating.

tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-load-stores-sm69.hlsl
```diff
@@ -0,0 +1,91 @@
...
+RWByteAddressBuffer RwByBuf : register(u1);
+
+StructuredBuffer< vector<TYPE, NUM> > RoStBuf : register(t2);
+RWStructuredBuffer< vector<TYPE, NUM>  > RwStBuf : register(u2);
```
llvm-beanz, line 22: also take or leave: I also find some of the double-spaces odd... I know we have have C++03-ish templates so you need a space, but extra spaces seem to just make lines longer.

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-operators-cs.hlsl
```diff
@@ -0,0 +1,708 @@
+// RUN: %dxc -HV 2018 -T cs_6_9 -DTYPE=float -DNUM=2 %s | FileCheck %s --check-prefixes=CHECK,NODBL,NOINT
```
llvm-beanz, line 1: Should we also be testing `vector<T,1>` here? Or potentially elsewhere?

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-operators.hlsl
```diff
@@ -48,24 +48,6 @@ struct Interface {
   TYPE scales[10];
 };
 
-#if 0
-// Requires vector loading support. Enable when available.
```
tex3d, line 52: Since this is the change which introduces said support, why is this deleted instead of used?

tools/clang/test/CodeGenDXIL/passes/longvec-load-stores-scalarizevecldst.ll
```diff
@@ -0,0 +1,478 @@
...
+  ; CHECK: [[vec8:%.*]] = insertelement <17 x float> [[vec7]], float [[val8]], i64 8
+  ; CHECK: [[vec9:%.*]] = insertelement <17 x float> [[vec8]], float [[val9]], i64 9
+  ; CHECK: [[vec10:%.*]] = insertelement <17 x float> [[vec9]], float [[val10]], i64 10
+  ; CHECK: [[vec11:%.*]] = insertelement <17 x float> [[vec10]], float [[val11]], i64 11
+  ; CHECK: [[vec12:%.*]] = insertelement <17 x float> [[vec11]], float [[val12]], i64 12
+  ; CHECK: [[vec13:%.*]] = insertelement <17 x float> [[vec12]], float [[val13]], i64 13
+  ; CHECK: [[vec14:%.*]] = insertelement <17 x float> [[vec13]], float [[val14]], i64 14
+  ; CHECK: [[vec15:%.*]] = insertelement <17 x float> [[vec14]], float [[val15]], i64 15
+  ; CHECK: [[vec16:%.*]] = insertelement <17 x float> [[vec15]], float [[val16]], i64 16
+  %tmp18 = call %dx.types.ResRet.v17f32 @dx.op.rawBufferVectorLoad.v17f32(i32 303, %dx.types.Handle %tmp17, i32 %tmp6, i32 0, i32 8)
```
tex3d, line 203: This starts with an alignment of 8, due to the bug noted in `HLOperationLower.cpp`.  You might want to adjust this source IR after fixing the bug.

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
```diff
@@ -0,0 +1,239 @@
...
+  static char ID; // Pass identification, replacement for typeid
+  explicit DxilScalarizeVectorLoadStores() : ModulePass(ID) {}
+
+  StringRef getPassName() const override {
+    return "DXIL scalarize vector load/stores";
+  }
+
+  bool runOnModule(Module &M) override {
+    DxilModule &DM = M.GetOrCreateDxilModule();
+    m_DM = &DM;
```
tex3d: It doesn't look like `m_DM` is used anywhere else in the pass, so why have it?

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    // Shader Model 6.9 allows native vectors and doesn't need this pass.
+    if (DM.GetShaderModel()->IsSM69Plus())
+      return false;
+
+    bool Changed = false;
+
+    hlsl::OP *HlslOP = DM.GetOP();
+    for (auto F = M.functions().begin(), E = M.functions().end(); F != E;) {
```
tex3d, line 51: You can iterate the overloads for specific ops you want instead with `HlslOP::GetOpFuncList`.  For instance:
https://github.com/microsoft/DirectXShaderCompiler/blob/2f357a9d625eaaa982ce1fac513e5f77a7d81900/lib/HLSL/DxilPreparePasses.cpp#L398

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
```diff
@@ -0,0 +1,239 @@
...
+      Function *Func = &*(F++);
+      DXIL::OpCodeClass OpClass;
+      if (HlslOP->GetOpCodeClass(Func, OpClass)) {
+        if (OpClass == DXIL::OpCodeClass::RawBufferVectorLoad) {
+          for (auto U = Func->user_begin(), UE = Func->user_end(); U != UE;) {
+            CallInst *CI = cast<CallInst>(*(U++));
+            scalarizeVectorLoad(HlslOP, M.getDataLayout(), CI);
```
tex3d, line 58: It's always a good form when iterating DXIL op class overload function users to check the opcode (even though you know in this case it should be the only one for this OpCodeClass):
```suggestion
            if (HlslOP->GetDxilOpFuncCallInst(CI) !=
                hlsl::OP::OpCode::RawBufferVectorLoad)
              continue;
            scalarizeVectorLoad(HlslOP, M.getDataLayout(), CI);
```

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
```diff
@@ -0,0 +1,239 @@
...
+          }
+          Func->eraseFromParent();
+        }
+      }
+    }
+    return Changed;
+  }
+};
+
+static unsigned GetRawBufferMask(unsigned NumComponents) {
```
tex3d, line 76: Of course, this could have been just `(1 << NumComponents) - 1` (with assert for `NumComponents <= 4`).  But oh well.

tex3d (commented):
Looking good.  One correctness issue with the alignment for the load op, and a few minor comments.

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    // Shader Model 6.9 allows native vectors and doesn't need this pass.
+    if (DM.GetShaderModel()->IsSM69Plus())
+      return false;
+
+    bool Changed = false;
+
+    hlsl::OP *HlslOP = DM.GetOP();
+    for (auto F = M.functions().begin(), E = M.functions().end(); F != E;) {
+      Function *Func = &*(F++);
```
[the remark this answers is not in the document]
pow2clk: Yeah, and their version of filecheck.

pow2clk (commented):
Thanks for all the feedback! I'll respond to most of it in the next commit. The vec1 test will have its own commit as it'll take just a bit longer.

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
tex3d: Oh, for one thing, erasing a function during iteration of this list will erase the entry from the list.  Since this is a SmallMapVector, the iterator points into the vector, which will be invalidated by any modification.  That doesn't seem like it should result specifically in duplicate entries and double-delete, but it is in UB territory.

If the original list contains duplicate entries, that is a cause for concern.  We shouldn't allow that.

lib/HLSL/DxilScalarizeVectorLoadStores.cpp
tex3d: It turns out that you *do* technically have to check, even with `GetOpFuncList`.  That's because this returns the list of overload functions for the `OpCodeClass` corresponding to the `OpCode`.  Basically, that's all it could do, since different DXIL operations use the same function declaration when they share the same `OpCodeClass`.

I think the best way is to get the `OpCode` using: `hlsl::OP::GetDxilOpFuncCallInst(CI) == hlsl::OP::OpCode::RawBufferVectorLoad` (this avoids a redundant `IsDxilOpFuncCallInst` check on `NDEBUG` at least).

I looked at other code using this function and basically *nobody* is checking the `OpCode` of the `CallInst` users.  I'm considering filing an issue about this, but I don't think there are any real bugs caused by this at the moment.  Basically, you're safe if there's a 1:1 mapping between `OpCode` and `OpCodeClass`, which happens to be true in all use cases.  If not, since this always lists all the overloads for the `OpCodeClass` that corresponds to the `OpCode` passed in, you could be iterating `CallInst` users that aren't calling with the same `OpCode` you're looking for.  I found one use of `GetOpFuncList` that looks like it could potentially have this problem in `OP::IsDxilOpUsed`, but this is only called with `OpCode`s that are safe, so I don't think it doesn't cause any real bugs in practice (yet).

I think it would probably be better to make `GetOpFuncList` take the `OpCodeClass` instead of the `OpCode`.  Then you're forced to `HlslOp->GetOpFuncList(hlsl::OP::GetOpCodeClass(MyOpCode))` and it's more obvious that you need to check the `OpCode` for each `CallInst` user.

tex3d (approved):

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -1548,16 +1538,32 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
-    unsigned alignSize = 0;
-    if (!isa<ConstantInt>(align)) {
-      ValCtx.EmitInstrError(CI,
-                            ValidationRule::InstrCoordinateCountForRawTypedBuf);
-    } else {
-      alignSize = bufSt.get_alignment_val();
+    if (resClass != DXIL::ResourceClass::UAV)
+      ValCtx.EmitInstrError(CI, ValidationRule::InstrResourceClassForUAVStore);
+
+    unsigned alignIdx = DXIL::OperandIndex::kRawBufferStoreAlignmentOpIdx;
```
llvm-beanz, line 1554:
```suggestion
    unsigned AlignIdx = DXIL::OperandIndex::kRawBufferStoreAlignmentOpIdx;
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -1548,16 +1538,32 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+    if (DXIL::OpCode::RawBufferVectorStore == opcode) {
+      alignIdx = DXIL::OperandIndex::kRawBufferVectorStoreAlignmentOpIdx;
+      unsigned valueIx = DXIL::OperandIndex::kRawBufferVectorStoreValOpIdx;
```
llvm-beanz, line 1557:
```suggestion
      unsigned ValueIx = DXIL::OperandIndex::kRawBufferVectorStoreValOpIdx;
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -1475,34 +1475,34 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
-    Value *offset = bufLd.get_elementOffset();
-    Value *align = bufLd.get_alignment();
-    unsigned alignSize = 0;
-    if (!isa<ConstantInt>(align)) {
-      ValCtx.EmitInstrError(CI,
-                            ValidationRule::InstrCoordinateCountForRawTypedBuf);
-    } else {
-      alignSize = bufLd.get_alignment_val();
-    }
+    unsigned alignIdx = DXIL::OperandIndex::kRawBufferLoadAlignmentOpIdx;
```
llvm-beanz, line 1498:
```suggestion
    unsigned AlignIdx = DXIL::OperandIndex::kRawBufferLoadAlignmentOpIdx;
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -1475,34 +1475,34 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
...
+    if (DXIL::OpCode::RawBufferVectorLoad == opcode)
+      alignIdx = DXIL::OperandIndex::kRawBufferVectorLoadAlignmentOpIdx;
+    if (!isa<ConstantInt>(CI->getOperand(alignIdx)))
+      ValCtx.EmitInstrError(CI, ValidationRule::InstrConstAlignForRawBuf);
+
+    Value *offset =
```
llvm-beanz, line 1504:
```suggestion
    Value *Offset =
```

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -1548,16 +1538,32 @@ static void ValidateResourceDxilOp(CallInst *CI, DXIL::OpCode opcode,
     if (!ValidateStorageMasks(CI, opcode, mask, stValMask, false /*isTyped*/,
                               ValCtx))
       return;
+  }
+    LLVM_FALLTHROUGH;
+  case DXIL::OpCode::RawBufferVectorStore: {
+    Value *handle =
```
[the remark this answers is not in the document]
llvm-beanz: Really?

* [m_](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/DxilValidation/DxilValidation.cpp#L69)
* [Member PascalCase not m_](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/DxilValidation/DxilValidation.cpp#L94)
* [Double Capital and camelCase](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/DxilValidation/DxilValidation.cpp#L100)
* On the next line [camelCase and PascalCase](https://github.com/microsoft/DirectXShaderCompiler/blob/main/lib/DxilValidation/DxilValidation.cpp#L101)

That's just up to line 101.

tools/clang/test/DXILValidation/vector-validation.hlsl
```diff
@@ -0,0 +1,16 @@
+// RUN: %dxc -T vs_6_9 %s -Od | FileCheck %s
```
tex3d, line 1: Like with the other one, I think the RUN line makes it look like it's run as a test when it isn't.

tex3d (approved):
Looks good!

Just a minor comment about RUN lines in hlsl files under `DXILValidation` where these are not run as tests.  I know other files have similar issues, but it's been misleading.  Sometimes, other contributors will place tests here, thinking that it's the right location for validation shell tests, and they are simply ignored.

lib/HLSL/DxilLinker.cpp
```diff
@@ -1272,6 +1276,7 @@ void DxilLinkJob::RunPreparePass(Module &M) {
 
   // Clean up vectors, and run mem2reg again
   PM.add(createScalarizerPass());
+
```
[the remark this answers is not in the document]
damyanp: Looks like this has been resolved.

tools/clang/test/DXILValidation/load-store-validation.hlsl
```diff
@@ -0,0 +1,76 @@
+// RUN: %dxc -T ps_6_9 %s -Od | FileCheck %s
+
+// HLSL source for validation of various invalid load/store parameters.
+// See LitDxilValidation/load-store-validation.ll.
+// Output is modified to trigger various validation errors.
```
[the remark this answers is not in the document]
pow2clk: The -Od wasn't actually used to generate the file. I don't remember why I put it here.

pow2clk:
Well, I wanted to see if resolving conflicts online would still require a new review and it does

tex3d (approved):

toji:
It seems like this may have broken our ability to roll the shader compiler into Dawn for Chromium (WebGPU).

Attempting to update to ToT with this change included starts producing the following compiler errors similar to the following for us:

```
FAILED: 71219ff8-ffe6-4725-9dbc-3e16cf060fd2 "./libdxcompiler.so" SOLINK ./libdxcompiler.so
err: exit=1
"python3" "../../build/toolchain/gcc_solink_wrapper.py" --readelf="../../third_party/llvm-build/Release+Asserts/bin/llvm-readelf" --nm="../../third_party/llvm-build/Release+Asserts/bin/llvm-nm"  --sofile="./libdxcompiler.so" --tocfile="./libdxcompiler.so.TOC" --output="./libdxcompiler.so" -- ../../third_party/llvm-build/Release+Asserts/bin/clang++ -shared -Wl,-soname="libdxcompiler.so" -fuse-ld=lld -Wl,--fatal-warnings -Wl,--build-id=fast -fPIC -Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now -Wl,--color-diagnostics -Wl,--undefined-version -Wl,--no-call-graph-profile-sort -m64 -no-canonical-prefixes -Wl,--gdb-index -Wl,-z,defs -Wl,--as-needed -nostdlib++ --sysroot=../../build/linux/debian_bullseye_amd64-sysroot -Werror -o "./libdxcompiler.so" @"./libdxcompiler.so.rsp"  
build step: solink "./libdxcompiler.so"
siso_rule: clang/solink
stderr:
ld.lld: error: lib_hlsl/DxilLinker.o: DWARF unit at offset 0x00000000 has unsupported version 0, supported are 2-5
ld.lld: error: undefined symbol: llvm::createDxilScalarizeVectorLoadStoresPass()
>>> referenced by DxilLinker.cpp
>>>               lib_hlsl/DxilLinker.o:((anonymous namespace)::DxilLinkJob::RunPreparePass(llvm::Module&)) in archive obj/third_party/gn/dxc/lib_hlsl.a

ld.lld: error: lib_hlsl/DxcOptimizer.o: DWARF unit at offset 0x00000000 has unsupported version 0, supported are 2-5
ld.lld: error: undefined symbol: llvm::initializeDxilScalarizeVectorLoadStoresPass(llvm::PassRegistry&)
>>> referenced by DxcOptimizer.cpp
>>>               lib_hlsl/DxcOptimizer.o:(hlsl::SetupRegistryPassForHLSL()) in archive obj/third_party/gn/dxc/lib_hlsl.a
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

Given that it can't resolve `createDxilScalarizeVectorLoadStoresPass` it definitely appears related to this PR.

dneto0:
> It seems like this may have broken our ability to roll the shader compiler into Dawn for Chromium (WebGPU).
> 

The fix was on our end: we had to add the new .cpp file to our BUILD.gn file.

https://dawn-review.googlesource.com/c/dawn/+/235214

PR #7295, opened 2025-04-02
[NFC] Update Shader Flags section of DXIL.rst

### LINKED ISSUE #7294: Update shader flag documentation in DXIL.rst
The Shader Flags section of DXIL.rst is currently incomplete and lacks documentation regarding the conditions under which the shader flags are/need to be set.

ACs:
- Update DXIL.rst to list and describe all shader flags
- Document the conditions or criteria under which each shader flag is set

Icohedron:
Fixes #7294

This PR updates DXIL.rst to list and describe all current shader flags in DXC (from `DxilShaderFlags.h`).
Also adds extra columns to the table of shader flags to document the minimum shader model version and the conditions or criteria under which each shader flag is set (as seen in `DxilShaderFlags.cpp` and `DxilModule.cpp`).


docs/DXIL.rst
```diff
@@ -515,33 +515,52 @@ Additional shader properties are specified via tag-value pair list, which is the
...
+Bit SM   Description                                                                              Criteria to set shader flag
+=== ==== ======================================================================================== ==========================================================================================================================================================================================================================================================================================================================================================
+0        Disable shader optimizations                                                             Command-line flag ``-disable-llvm-optzns`` provided to DXC's internal frontend
+1        Disable math refactoring                                                                 Not set
```
inbelic, line 524: It is not immediately clear what 'Not set' means in this context? Same for the others with the same criteria.

Is it that there is no set criteria, or, just if the flag is not set?

Icohedron: It just means the flag is not set

docs/DXIL.rst
```diff
@@ -515,33 +515,52 @@ Additional shader properties are specified via tag-value pair list, which is the
...
+
+=== ==== ======================================================================================== ==========================================================================================================================================================================================================================================================================================================================================================
+Bit SM   Description                                                                              Criteria to set shader flag
+=== ==== ======================================================================================== ==========================================================================================================================================================================================================================================================================================================================================================
+0        Disable shader optimizations                                                             Command-line flag ``-disable-llvm-optzns`` provided to DXC's internal frontend
```
llvm-beanz, line 523: `-dixable-llvm-optzns` isn't really supported by DXC, I believe the flag is `-Od`.

docs/DXIL.rst
llvm-beanz: I believe this is an older shader model flag. Maybe something more like: "Removed in Shader Model 6.0" is appropriate. @tex3d any thoughts?

Icohedron: Though I suppose the wording is a bit awkward because I think the intention for some of these flags is that you can set it if you wish to have its effects? Like if you want to disable math refactoring, you can set the bit for it?

The criteria is more like describing the conditions under which the flag is required to be set and the validator will complain about it if you use a capability without setting its corresponding bit.

docs/DXIL.rst
```diff
@@ -515,33 +515,52 @@ Additional shader properties are specified via tag-value pair list, which is the
...
+10       Inner coverage                                                                           Use of the InnerCoverage instruction, or the InnerCoverage semantic is present in the output signature of a pixel shader
+11       Stencil reference value                                                                  The StencilRef semantic is present in the output signature of a pixel shader
+12       Tiled resources                                                                          Use of the CheckAccessFullyMapped instruction or, if DXIL validator version >= 1.8, the use of LodClamp in intrinsics SampleGrad, SampleCmpGrad, Sample, SampleBias, SampleCmp, or SampleCmpBias
+13       Typed UAV load additional formats                                                        Use of TextureLoad or BufferLoad on a UAV with a multi-component data type. If DXIL validator version == 1.0 then this flag is set when a TextureLoad or BufferLoad is used on any UAV, regardless of its data type
+14       Comparison filtering for feature level 9                                                 Not set
```
llvm-beanz, line 537: It would be good to know the context in which this was valid, but I suspect this was also removed in SM 6.0.

docs/DXIL.rst
```diff
@@ -515,33 +515,52 @@ Additional shader properties are specified via tag-value pair list, which is the
...
+15       64 UAV slots                                                                             The presence of more than 8 UAVs. If Shader Model >= 6.6, then UAV ranges count as multiple UAVs
+16       UAVs at every shader stage                                                               The shader has UAVs and: if Shader Model >= 6.8, the shader is a vertex, hull, domain, or geometry shader; else (Shader Model < 6.8) the shader is not a compute or pixel shader
+17       CS4 raw and structured buffers                                                           Use of RawBuffer or StructuedBuffer in a Shader Model 4.x compute shader
+18       Rasterizer order views                                                                   Not set
```
llvm-beanz, line 541: Is this really not set in DXC?
@tex3d, do we not set this flag on ROV usage?

Icohedron: Oh yea it is. I just missed it. It's set in [DxilModule.cpp:L315](https://github.com/microsoft/DirectXShaderCompiler/blob/2b1c2e640dae09adf1cb2dd52bc5ce860d73b02b/lib/DXIL/DxilModule.cpp#L315). I will update the flag criteria

docs/DXIL.rst
Icohedron: I was able to confirm that `-disable-llvm-optzns` is tied to the setting of this flag, but I wasn't sure about `-Od`.

edit: nevermind, found the code path that sets it!

inbelic:
What is the status of this pr?

Icohedron:
> What is the status of this pr?

The PR needs input from @tex3d to address the comments @llvm-beanz made.

However, this PR doesn't seem urgent at the moment.

docs/DXIL.rst
tex3d: This flag is used in the DXBC shader converter to capture the inverse of the global flag `D3D10_SB_GLOBAL_FLAG_REFACTORING_ALLOWED`, which may be present in the DXBC.
The closest equivalent would be `-Gis`, which is partly what would have driven this flag in FXC (the other part depends on the shader profile).

DXC doesn't set this flag for shaders compiled from HLSL to DXIL, since using `-Gis` will control this property at the operation level through fast math flags and the precise metadata.

So perhaps the more appropriate wording would be:
"Only used by DxilConv; corresponds to ``~D3D10_SB_GLOBAL_FLAG_REFACTORING_ALLOWED`` in DXBC"

docs/DXIL.rst
tex3d: This maps a feature flag that is not used for DXIL shaders. We could have gotten away with omitting the flag entirely here originally, but it's here now.

I might say "Not set; Unused legacy feature requirement flag" if you wanted more context here.

docs/DXIL.rst
```diff
@@ -515,33 +515,52 @@ Additional shader properties are specified via tag-value pair list, which is the
...
+20       64-Bit integers                                                                          Use of i64 data types
+21  6.1+ View Instancing                                                                          Use of the ViewID instruction
+22  6.1+ Barycentrics                                                                             Use of the AttributeAtVertex instruction, or the Barcentrics semantic is present in the shader input signature. If DXIL validator version < 1.6 then this flag must not be set at all
+23  6.2+ Enable native low-precision data types                                                   Command-line flag ``-enable-16bit-types`` is provided to DXC
+24  6.4+ ShadingRate                                                                              The ShadingRate semantic is present in the shader input or output signature
+25  6.5+ Raytracing tier 1.1 features                                                             Use of the AllocateRayQuery, AllocateRayQuery2, or GeometryIndex instruction
+
+26  6.5+ Sampler feedback                                                                         Not set
```
tex3d: That's worrisome. This one should have been set based on when any sampler feedback operations (identified with `OP::IsDxilOpFeedback(OpCode C)`) are used. It looks like the code doesn't do that!

tex3d: Specific ops are: `WriteSamplerFeedback`, `WriteSamplerFeedbackBias`, `WriteSamplerFeedbackLevel`, and `WriteSamplerFeedbackGrad`.

tex3d: This should be the fix (minus the test), if we were able to get this into the next point release for 1.9:

```diff
diff --cc lib/DXIL/DxilShaderFlags.cpp
index 71c7305b6,71c7305b6..cbe62e0e2
--- a/lib/DXIL/DxilShaderFlags.cpp
+++ b/lib/DXIL/DxilShaderFlags.cpp
@@@ -586,6 -586,6 +586,8 @@@ ShaderFlags ShaderFlags::CollectShaderF
            continue;
          if (hlsl::OP::IsDxilOpWave(dxilOp))
            hasWaveOps = true;
++        if (hlsl::OP::IsDxilOpFeedback(dxilOp))
++          hasSamplerFeedback = true;
          switch (dxilOp) {
          case DXIL::OpCode::CheckAccessFullyMapped:
            hasCheckAccessFully = true;
@@@ -821,6 -821,6 +823,10 @@@
      // Before validator version 1.8, RequiresGroup flag did not exist.
      requiresGroup = false;
    }
++  if (hasSamplerFeedback && DXIL::CompareVersions(valMajor, valMinor, 1, 9) < 0) {
++    // Before validator version 1.9, SamplerFeedback flag was not set.
++    hasSamplerFeedback = false;
++  }
  
    flag.SetEnableDoublePrecision(hasDouble);
    flag.SetStencilRef(hasStencilRef);
```

docs/DXIL.rst
```diff
@@ -515,33 +515,52 @@ Additional shader properties are specified via tag-value pair list, which is the
...
+29  6.6+ Derivatives in mesh and amplification shaders                                            Use of instructions DerivFineX, DerivFineY, DerivCoarseX, DerivCoarseY, CalculateLOD, Sample, SampleBias, SampleCmp, or SampleCmpBias in a mesh or amplification shader
+30  6.6+ Resource descriptor heap indexing                                                        Use of the CreateHandleFromHeap instruction on a resource descriptor heap
+31  6.6+ Sampler descriptor heap indexing                                                         Use of the CreateHandleFromHeap instruction on a sampler descriptor heap
+32  6.6+ 64-bit Atomics on Heap Resources                                                         Use of i64 AtomicBinOp or AtomicCompareExchange instructions on a descriptor heap resource
+33  6.7+ Any UAV may not alias any other UAV                                                      If DXIL validator version >= 1.8: use of a UAV in any function. If DXIL validator version < 1.8: global presence of a UAV. This flag is not set if the ``-res-may-alias`` command-line flag is provided to DXC or if DXIL version < 1.7
```
tex3d: I don't think we should comment on the validator version logic. Just the latest meaning is what we should put here. The version logic is to make it possible to compile and validate with a previously released validator, especially when the current version isn't final/released yet.

tex3d: Same goes for other places where validator version is referred to for other bits.

docs/DXIL.rst
Icohedron: Will fix this in a separate non-NFC PR, tracked by new issue: https://github.com/microsoft/DirectXShaderCompiler/issues/8533

docs/DXIL.rst
```diff
@@ -515,33 +515,52 @@ Additional shader properties are specified via tag-value pair list, which is the
...
+Bit SM   Description                                                                              Criteria to set shader flag
+=== ==== ======================================================================================== ==========================================================================================================================================================================================================================================================================================================================================================
+0        Disable shader optimizations                                                             Command-line flag ``/Od`` is provided to DXC
+1        Disable math refactoring                                                                 Only used in Dxilconv; corresponds to `~D3D10_SB_GLOBAL_FLAG_REFACTORING_ALLOWED` in DXBC
```
tex3d: Do the back-quote characters need to be doubled for `.rst`, like they are for all the others?

```suggestion
1        Disable math refactoring                                                                 Only used in Dxilconv; corresponds to ``~D3D10_SB_GLOBAL_FLAG_REFACTORING_ALLOWED`` in DXBC
```

tex3d (approved):

tex3d (approved):

inbelic (approved):

PR #7305, opened 2025-04-03
Fix vk::BufferPointer for non-SPIR-V builds

dneto0:
PR #7163 added support for vk::BufferPointer

It did not build when building for non-SPIR-V scenarios.

Bug: crbug.com/408215487


amaiorano:
Actually, the author of the original change put up a similar fix: https://github.com/microsoft/DirectXShaderCompiler/pull/7306
We'll go with that one to keep ownership of the the changes with one person.

PR #7307, opened 2025-04-03
[NFC] Standardize DxilValidation variable capitalization

pow2clk:
Capitalize all the variables and rename a few in DxilValidation.cpp in keeping with https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

As this file was easily mistaken for applying to the golden rule: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly it is at serious risk of receiving changes that will get hung up by requirements to follow the LLVM coding guidelines. This brings the cases where variable capitalization is not in line with the coding standards to avoid such pitfalls in the future.

### DIFF
--- a/lib/DxilValidation/DxilValidation.cpp
+++ b/lib/DxilValidation/DxilValidation.cpp
@@ -97,68 +97,68 @@ struct PSExecutionInfo {
 };
 
 static unsigned ValidateSignatureRowCol(Instruction *I,
-                                        DxilSignatureElement &SE, Value *rowVal,
-                                        Value *colVal, EntryStatus &Status,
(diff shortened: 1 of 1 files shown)


damyanp (approved):
This should help avoid ambiguity in the future.

We may want to consider waiting until some other changes in flight are complete (or we may not - we should check anyway!)

damyanp (approved):

tex3d (approved):
While I don't normally like wholesale reformatting changes like this, if it alleviates conflicts regarding the ambiguity of coding standards to apply for future changes, it's probably the lesser of evils.

PR #7309, opened 2025-04-04
Use internal validator for execution test that requires SM 6.8

bob80905:
This PR modifies WaveSizeRange test which depends on shader model 6.8.
The compiler needs -select-validator internal.
This will allow the tests to be run in different testing environments when an external validator that isn't sufficient is available.


damyanp (approved):
Is this all that's needed?

jenatali (approved):

jenatali:
> Is this all that's needed?

Technically all tests could be modified to use `-select-validator internal` to not require an external validator anywhere on the system, but the SM6.8 tests are the biggest problem right now since the latest released Windows SDK includes a validator that only supports SM6.7. So, yeah this is sufficient for the problems that we're seeing right now.

PR #7312, opened 2025-04-04
[Cherry Pick] Use internal validator for execution test that requires SM 6.8 

bob80905:
This PR modifies WaveSizeRange test which depends on shader model 6.8.
The compiler needs -select-validator internal.
This will allow the tests to be run in different testing environments when an external validator that isn't sufficient is available.

Cherry pick of https://github.com/microsoft/DirectXShaderCompiler/pull/7309

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -11641,20 +11641,23 @@ void ExecuteWaveSizeRangeInstance(UINT minWaveSize, UINT maxWaveSize,
       })";
 
   // format compiler args
-  char compilerOptions[64];
+  char compilerOptions[70];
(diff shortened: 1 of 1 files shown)


damyanp:
Do we need this in the short term?  I wonder if the correct long term fix here is to make the internal validator the default, which is what the plan was all along.

bob80905:
> Do we need this in the short term? I wonder if the correct long term fix here is to make the internal validator the default, which is what the plan was all along.

Unneeded in the short term as far as I'm concerned. But I imagine this PR would prevent similar issues if we decide to make a release before changing the internal validator default behavior.

bob80905:
We're going to opt to change validation behavior entirely to resolve this. 
According to https://github.com/microsoft/DirectXShaderCompiler/issues/7389

Issue #7260, opened 2025-03-25
Implement priority long vector execution tests


alsepkow:
Implement the high priority long vector exec tests as listed in [this pr](https://github.com/microsoft/hlsl-specs/pull/421)

* Initializing a vector with another.
* Multiply all components of a vector with a scalar value
* Add all components of a vector with a scalar value
* Clamp all components of a vector to the range [c, t]
* Component wise minimum between 2 vectors
* Component wise maximum between 2 vectors
* Component wise multiply between 2 vectors
* Component wise add between 2 vectors
* Subscript access, vec[i] = c

alsepkow:
Work completed in staging branch.

PR #7320, opened 2025-04-07
Basic implementation of priority long vector exec tests.

alsepkow:
This PR is a basic implementation of the priority long vector execution tests #7260.


tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -501,6 +501,16 @@ class ExecutionTest {
                        L"Table:ShaderOpArithTable.xml#PackUnpackOpTable")
   END_TEST_METHOD()
 
+  BEGIN_TEST_METHOD(LongVector_BinaryOpTest)
```
alsepkow, line 504: Note:  I did not yet include additional types/vector sizes. But the infrastructure to do so is mostly there. I wanted to further discuss possible ways of doing that in this PR and will include the adjustment to implement this in a follow up commit on this PR.

I'm conflicted on a clean way of doing that. I don't want to copy/paste large vectors in the XML. Too clunky and error prone. I'm currently considering adding 'interesting' values to the <value> tags in the XML and using those to generate larger arrays in a deterministic way. Perhaps something as simple as repeating the comma separated list of values until the array is the desired test size.

I'm also conflicted on the best/cleanest way to define test cases across data types. I don't really want to copy base the XML test cases. I'm leaning towards keeping a small number of XML rows and again generating test cases based off of that. But it many be a little clunky when values are data type specific, i.e maybe we want to include small values for floats like '0.001'. 

Additionally, the way in which we define the test cases is relevant for the OS HLK infra. We're free to generate as many tests and GUIDs for those test cases. This may be worth a design discussion.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -710,6 +720,14 @@ class ExecutionTest {
                                const char *pShaderModelStr, const char *pShader,
                                Ty *pInputDataPairs, unsigned inputDataCount);
 
+  template <typename T, std::size_t N = 4>
+  //typename = std::enable_if_t<std::is_arithmetic_v<T>>>
```
alsepkow, line 724: Will rectify this commented out code, my intension was to enforce that the T was of an arithmetic type but I was hitting some compilation errors on this when trying to use the template. I'm not sure it was doing what I expected.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11135,402 @@ TEST_F(ExecutionTest, PackUnpackTest) {
   }
 }
 
+// TODOLongVec : Need to change the members to vectors. But when I did that the
```
alsepkow, line 11138: Stale comment. Will fix.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11135,402 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+struct SLongVectorBinaryOp {
+  T scalarInput;
+  std::array<T, N> vecInput1;
+  std::array<T, N> vecInput2;
+  std::array<T, N> vecOutput;
+};
+
+// vec1 == Expected vector
+// vec2 == Actual vector
+template <typename T, typename = std::enable_if_t<std::is_arithmetic_v<T>>>
```
alsepkow, line 11154: Whoops, I intended to either add an index parameter to this function or have the callers also log the index of the array they are testing. They're arrays of arrays.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11135,402 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+  LongVectorBinaryOpTestBase<float, 4>();
+}
+
+template <typename T, std::size_t N>
+//typename = std::enable_if_t<std::is_arithmetic_v<T>>>
+void ExecutionTest::LongVectorBinaryOpTestBase() {
+  WEX::TestExecution::SetVerifyOutput verifySettings(
+      WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
+
+  // TODOLONGVEC: This seems frivolous as the device we target is created
```
alsepkow, line 11259: Will look at the rest of the test code and update this.
We may also want to consider a member variable based on whether or not this is running in the HLK vs as an exec test. Skipping should never be allowed in the HLK for any reason.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11135,402 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+    // TODOLONGVEC: This should fail. HLK requirements should select the test.
+    // But for exectests we can just return and skip. Will need to add a
+    // mechanism to faciliate that behavior in a good way that enforces it to
+    // always be on in OS builds.
+    WEX::Logging::Log::Comment("Device does not support SM 6.9. Can't run these tests.");
+    WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+    return;
+  }
+
+  // TableParameterHandler is a helper that parses the xml file linked
```
alsepkow, line 11274: I'm considering refactoring this into a helper function that parses the args into a struct based off of the applicable 'LongVector*Parameters' table definition for easier use amongst test cases.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11135,402 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+
+  const size_t inputVectorCount = vecInput1.size();
+  const bool hasScalarInput = scalarInputs.size() > 0;
+  const bool hasVecInput2 = vecInput2.size() > 0;
+  VERIFY_IS_FALSE(hasScalarInput == hasVecInput2, L"Expecting either a vector and vector op or a vector and scalar op");
+
+  // The two operand vectors must be the same size.
+  VERIFY_IS_TRUE(inputVectorCount == vecInput2.size() || inputVectorCount == scalarInputs.size());
+
+  // Additionally all elements (also vectors) in the vector must be the same
```
alsepkow, line 11316: Will add.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11135,402 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+
+TEST_F(ExecutionTest, LongVector_UnaryOpTest) {
+  WEX::TestExecution::SetVerifyOutput verifySettings(
+      WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
+    LongVectorUnaryOpTestBase<float, 4>();
+}
+
+template <typename T, std::size_t N>
+//typename = std::enable_if_t<std::is_arithmetic_v<T>>>
+void ExecutionTest::LongVectorUnaryOpTestBase() {
```
alsepkow, line 11408: This code is extremely similar to BinaryTestOpBase, consider factoring out common logic into helpers.

tools/clang/unittests/HLSLExec/ShaderOpTest.cpp
```diff
@@ -861,6 +861,11 @@ void ShaderOpTest::CreateShaders() {
       CHECK_HR(pLibrary->CreateBlobWithEncodingFromPinned(
           pText, (UINT32)strlen(pText), CP_UTF8, &pTextBlob));
       CHECK_HR(m_pDxcSupport->CreateInstance(CLSID_DxcCompiler, &pCompiler));
+      WEX::Logging::Log::Comment(L"Compiling shader:");
```
alsepkow, line 864: Added this for convenience. Was handy when validating.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -501,6 +502,100 @@ class ExecutionTest {
...
 
+  BEGIN_TEST_METHOD(LongVector_BinaryOpTest_bool)
+  TEST_METHOD_PROPERTY(L"DataSource",
+                       L"Table:ShaderOpArithTable.xml#LongVector_BinaryOpTable")
+  END_TEST_METHOD()
+
+  //BEGIN_TEST_METHOD(LongVector_BinaryOpTest_float16)
+  //TEST_METHOD_PROPERTY(L"DataSource",
+  //                     L"Table:ShaderOpArithTable.xml#LongVector_BinaryOpTable")
+  //END_TEST_METHOD()
```
damyanp, line 513: What's up with float16?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
damyanp: Is that necessary? Presumably if it wasn't an arithmetic type you'd pretty soon get errors about it.

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
```diff
@@ -3750,4 +3750,24 @@ void MSMain(uint GID : SV_GroupIndex,
     </Shader>
   </ShaderOp>
 
+  <ShaderOp Name="LongVectorBinaryOp" CS="CS">
+    <RootSignature>RootFlags(0), UAV(u0)</RootSignature>
+    <Resource Name="SLongVectorBinaryOp&lt;T&gt;" Dimension="BUFFER" Width="1024" Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST" TransitionTo="UNORDERED_ACCESS" Init="ByName" ReadBack="true" />
```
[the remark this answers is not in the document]
damyanp: Why?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11224,605 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+        for (size_t index : mismatchedIndexes) {
+          std::wstringstream wss(L"");
+          wss << L"Mismatch at Index: " << index;
+          wss << L" Expected Value:" << vec1[index];
+          wss << L" Actual Value:" << vec2[index];
+          WEX::Logging::Log::Error(wss.str().c_str());
+        }
+    }
+
+    return vectorsMatch;
```
damyanp, line 11286: nit: `vectorsMatch` isn't needed, you can just use `mismatchedIndexes.empty()`

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11224,605 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+  const bool hasVecInput2 = vecInput2.size() > 0;
+  VERIFY_IS_FALSE(hasScalarInput == hasVecInput2, L"Expecting either a vector and vector op or a vector and scalar op");
+
+  // The two operand vectors must be the same size.
+  VERIFY_IS_TRUE(inputVectorCount == vecInput2.size() || inputVectorCount == scalarInputs.size());
+
+  // Additionally all elements (also vectors) in the vector must be the same
+  // size.
+  // TODO: Add check.
+  // Keep filling the vectors with 'copies' of themselves until we hit N
```
damyanp, line 11491: Why?  Is this trying to make up for potential bugs in the XML?

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
```diff
@@ -3750,4 +3750,24 @@ void MSMain(uint GID : SV_GroupIndex,
...
+  </ShaderOp>
+
+  <ShaderOp Name="LongVectorUnaryOp" CS="CS">
+    <RootSignature>RootFlags(0), UAV(u0)</RootSignature>
+    <Resource Name="SLongVectorUnaryOp&lt;T&gt;" Dimension="BUFFER" Width="1024" Flags="ALLOW_UNORDERED_ACCESS" InitialResourceState="COPY_DEST" TransitionTo="UNORDERED_ACCESS" Init="ByName" ReadBack="true" />
+    <RootValues>
+      <RootValue Index="0" ResName="SLongVectorUnaryOp&lt;T&gt;" />
+    </RootValues>
+    <Shader Name="CS" Target="cs_6_0">
+    </Shader>
```
damyanp, line 3770: Is this just ignored?

tools/clang/unittests/HLSLExec/ShaderOpArithTable.xml
```diff
@@ -7524,4 +7524,308 @@
...
+        <Parameter Name="ShaderOp.Text">
+            <![CDATA[
+
+            template<typename T>
+            struct SLongVectorBinaryOp{
+                T scalarInput;
+                vector<T, NUM> vecInput1;
+                vector<T, NUM> vecInput2;
+                vector<T, NUM> vecOutput;
+            };
```
damyanp, line 7549: Wondering why `T` as a template parameter (instead of just using `TYPE`), but using the `NUM` macro.

tools/clang/unittests/HLSLExec/ShaderOpArithTable.xml
```diff
@@ -7524,4 +7524,308 @@
             </Parameter>
         </Row>
     </Table>
+    <Table Id="LongVector_BinaryOpTable">
+    <ParameterTypes>
```
damyanp, line 7528: Presumably everything under the `<Table>` element should be indented?

tools/clang/unittests/HLSLExec/ShaderOpArithTable.xml
```diff
@@ -7524,4 +7524,308 @@
...
+                SLongVectorBinaryOp<TYPE> l = g_buf[GI];
+                l.vecOutput = l.vecInput1 * l.scalarInput;
+                g_buf[GI] = l;
+            };
+
+            ]]>
+        </Parameter>
+        <Parameter Name="ShaderOp.Target">cs_6_9</Parameter>
+        <Parameter Name="Validation.VecInput1">
+            <Value>1,1,1,1</Value>
```
damyanp, line 7562: Are all these operations elementwise that we're testing?  I wonder if there's a better way to slice it - I'm worried about what'll happen when we come to wanting to test hundreds of elements in a vector.  Also, I'm not sure having every element of the vector is a good test.

I wonder instead if we collect some sample values for each operation and then have the testing infrastructure test different combinations of these with different sized vectors.

damyanp (changes requested):
My comments here are mainly observations.  The main (pun unintended) change I'm requesting is that this to the is retargeted to the staging branch.  I think there's quite a lot of work we'll want to do before this is ready to go to main.

tools/clang/unittests/HLSLExec/ShaderOpArithTable.xml
alsepkow: Yah, I think we will need to get creative here. The test cases have to be reproducible and defining large vectors by hand just isn't going to scale.

My current thoughts:

1. 'Random' generation with the same seed values so it becomes deterministic.
2. 'Fill with sequence' so we can define sequences containing values that we think are interesting, perhaps max/min or whatever in some smaller vector and then fill the test vector with that sequence repeatedly until its filled to the test size.

tools/clang/unittests/HLSLExec/ShaderOpArithTable.xml
alsepkow: Good point, the template isn't needed. I'll update to just use macro values.

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
alsepkow: Not 'ignored', initially the test parses this as the target but it's overridden by the lambda thats passed into RunShaderOpTest. ' pShaderOp->Shaders.at(0).Target = Target.m_psz'.  I copied the pattern from other test cases before I realized what was going on.

This infra needs to be cleaned up.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Yah, this is another reason I'm leaning towards not defining things in the XML. Too easy to mess this up.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11228,609 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+  } else if constexpr (std::is_same_v<T, float>) {
+    return std::stof(str);
+  } else if constexpr (std::is_same_v<T, double>) {
+    return std::stod(str);
+  } else if constexpr (std::is_same_v<T, bool>) {
+    bool boolVal = std::stod(str) > 0.0 ? true : false;
+    return boolVal ;
+  } else {
+    VERIFY_IS_TRUE(false, L"Unsupported type for parsing string to number.");
+  }
```
llvm-beanz, line 11318: nit: applying two LLVM coding conventions: 
* [don't use `else` after a `return`](https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return)
* [don't use braces for single statement bodies](https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements)
```suggestion
  if constexpr (std::is_same_v<T, uint16_t>)
    return static_cast<uint16_t>(std::stoul(str));
   if constexpr (std::is_same_v<T, uint32_t>)
    return std::stoul(str);
  if constexpr (std::is_same_v<T, uint64_t>)
    return std::stoull(str);
  if constexpr (std::is_same_v<T, int16_t>)
    return static_cast<int16_t>(std::stol(str));
  if constexpr (std::is_same_v<T, int32_t>)
    return std::stol(str);
  if constexpr (std::is_same_v<T, int64_t>)
    return std::stoll(str);
  if constexpr (std::is_same_v<T, float>)
    return std::stof(str);
  if constexpr (std::is_same_v<T, double>)
    return std::stod(str);
  if constexpr (std::is_same_v<T, bool>) {
    bool boolVal = std::stod(str) > 0.0 ? true : false;
    return boolVal ;
  }
  VERIFY_IS_TRUE(false, L"Unsupported type for parsing string to number.");
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11228,609 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+        for (size_t index : mismatchedIndexes) {
+          std::wstringstream wss(L"");
+          wss << L"Mismatch at Index: " << index;
+          wss << L" Expected Value:" << vec1[index];
+          wss << L" Actual Value:" << vec2[index];
+          WEX::Logging::Log::Error(wss.str().c_str());
+        }
+    }
+
+    return vectorsMatch;
```
llvm-beanz, line 11292: nit: Slightly different way of structuring. It combines Damyan's suggestion with applying the coding conventions for early exits (https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code).

```suggestion
    if (mismatchedIndexes.empty())
      return true;
      
    for (size_t index : mismatchedIndexes) {
      std::wstringstream wss(L"");
      wss << L"Mismatch at Index: " << index;
      wss << L" Expected Value:" << vec1[index];
      wss << L" Actual Value:" << vec2[index];
      WEX::Logging::Log::Error(wss.str().c_str());
    }

    return false;
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11228,609 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+// would parse to 3 vectors of 4 elements each.
+// Or three strings ["1" , "2", "3"] will parse to 3 single elements vectors.
+// Vectors for a single element is a little weird, but it allows this helper
+// function to be very flexible.
+template <typename T, typename = std::enable_if_t<std::is_arithmetic_v<T>>>
+std::vector<std::vector<T>> parseStringsToNumbers(const std::vector<WEX::Common::String>* input) {
+  std::vector<std::vector<T>> result = {};
+  if (input == nullptr || input->empty()) {
+      return result;
+  }
```
llvm-beanz, line 11335: Looks like your editor likes 4-space indents. The format checker is complaining about this too. Also no braces required 😄.
```suggestion
  if (input == nullptr || input->empty())
    return result;
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11228,609 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+  for (const auto& wexStr : *input) {
+
+    std::wstring wStr(wexStr);
+    std::wstringstream wss(wStr);
+    std::wstring number;
+    std::vector<T> vec;
+
+    while (std::getline(wss, number, L',')) {
+        vec.push_back(parseStringToNumber<T>(number));
+    }
```
llvm-beanz, line 11346:
```suggestion
    while (std::getline(wss, number, L','))
        vec.push_back(parseStringToNumber<T>(number));
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11228,609 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+  auto scalarInputs = parseStringsToNumbers<T>(Scalar_Inputs);
+
+  // double is plenty of room for tolerance range
+  const double Validation_Tolerance =
+  handler.GetTableParamByName(L"Validation.Tolerance")->m_double;
+
+  const size_t inputVectorCount = vecInput1.size();
+  const bool hasScalarInput = scalarInputs.size() > 0;
+  const bool hasVecInput2 = vecInput2.size() > 0;
+  VERIFY_IS_FALSE(hasScalarInput == hasVecInput2, L"Expecting either a vector and vector op or a vector and scalar op");
```
llvm-beanz, line 11489: 80-col or bust! 🤪

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11228,609 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+      pShaderOp->Shaders.at(0).Arguments = compilerOptions;
+    }
+  );
+
+  // Map the data from GPU to CPU memory so we can verify our expectations.
+  MappedData data;
+  test->Test->GetReadBackData("SLongVectorUnaryOp<T>", &data);
+
+  // Cast the buffer back into an array of the structs we expect.
+  SLongVectorUnaryOp<T, N> *pPrimitive = (SLongVectorUnaryOp<T, N>*)data.data();
```
llvm-beanz, line 11823: There's a lot of new code being added here in a mix of style conventions. Ideally new code should conform to the LLVM conventions (see: [the variable naming conventions](https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly)), unless there is a consistent style in the existing code.

For this you have a lot of new code which is itself inconsistent. We should avoid that.

Also I have a particular dislike of Hungarian notation, which is why I left this comment on this line specifically.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: The errors were easier to parse for a human with these asserts. But you're right, you'll still get errors. Opting to remove the std::enable_if's.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Went a different route with code cleanup.

tools/clang/unittests/HLSLExec/ShaderOpArithTable.xml
alsepkow: Got rid of using the XML for these tests.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: The coding styles in these tests are all over the place. Auto-complete in my editor was having a bad time because of it. I intended to come back on a formatting pass. 

They're also already using Hungarian notation in most places.

damyanp: Yeah, when we're adding new code in a file like this that's inconsistent we follow llvm naming conventions.  Although the llvm doc doesn't explictly say "no hungarian", I think it is pretty clear that hungarian is incompatible with the naming conventions described.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Needed to add logic because C++ doesn't have a native type for it. Well, C++23 does.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -35,6 +35,8 @@
 #include <algorithm>
 #include <type_traits>
 #include <bitset>
+#include <random>
```
alsepkow, line 38: Both of these were added for the DeterministicNumberGenerator I added.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -501,11 +512,122 @@ class ExecutionTest {
                        L"Table:ShaderOpArithTable.xml#PackUnpackOpTable")
   END_TEST_METHOD()
 
+  // bool binary ops
```
alsepkow, line 515: There are a bajillion entry points because the HLK infra is clunky. Generally, each test case should have an entry point so it can have its own GUID. If folks really don't like this we can combine more tests cases into an entry point - but we'll lose the ability to filter them in the HLK.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -501,11 +512,122 @@ class ExecutionTest {
...
+  // uint64 unary ops
+  TEST_METHOD(LongVector_Clamp_uint64);
+  TEST_METHOD(LongVector_Initialize_uint64);
+
   dxc::DxcDllSupport m_support;
 
   bool m_D3DInitCompleted = false;
   bool m_ExperimentalModeEnabled = false;
   bool m_AgilitySDKEnabled = false;
+  bool m_HLKModeEnabled = false; // Prevent skip logic when running HLK tests.
```
alsepkow, line 630: There is no differentiation between a TAEF skip or pass in the HLK infra. Skipping when running under the HLK infra is bad. Device capabilities are meant to be queried by the HLK gatherers and tests are only meant to be run if the target device reported support.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11238,1067 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+      Uint32Dist = std::uniform_int_distribution<T>(std::numeric_limits<T>::min(), std::numeric_limits<T>::max());
+    else if constexpr (std::is_same_v<T, uint64_t>)
+      Uint64Dist = std::uniform_int_distribution<T>(std::numeric_limits<T>::min(), std::numeric_limits<T>::max());
+    else if constexpr (std::is_same_v<T, float>)
+      // TODO before PR completed ::min is the lowest postivie value. Passing
+      // ::lowest was causing an assert in the distribution. Need to look
+      // further into why
+      FloatDist = std::uniform_real_distribution<T>(std::numeric_limits<T>::min(), std::numeric_limits<T>::max());
+    else if constexpr (std::is_same_v<T, double>)
+      // TODO before PR completed ::min is the lowest postivie value. Passing
```
alsepkow, line 11517: will fix

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11238,1067 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+    return false;
+  }
+
+  // Stash mismatched indexes for easy failure logging later
+  std::vector<size_t> MismatchedIndexes;
+  for (size_t Index = 0; Index < VecInput.size(); ++Index) {
+    if constexpr (std::is_same_v<T, HLSLBool_t>) {
+      // Compiler was very picky and wanted an explicit case for any T that
+      // doesn't implement the operators in the below else. ( > and -). It
+      // wouldn't accept putting this constexpr as an or case in the above if.
```
alsepkow, line 11581: stale comment, will fix

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11238,1067 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+    while(ClampArgC >= ClampArgT)
+    {
+      // Generate a new value for ClampArgC. It needs to be smaller than
+      // ClampArgT.
+      ClampArgT = NumberGenerator.generate();
+    }
+  }
+
+  std::vector<T> VecExpected = {};
+  for (size_t Index = 0; Index < VecInput1.size(); Index++) {
```
alsepkow, line 12160: This logic is working for now. But I was going to discuss further with the team about adding logic to compute the scalarized expected values in the shader or add reasonable tolerance values for comparing expected.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11238,1067 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+    CompilerOptions<< (TestConfig.IsScalarOp ? "LVOp.ScalarInput" : "LVOp.VecInput2");
+  }
+  CompilerOptions << " -DFUNC=";
+  CompilerOptions << TestConfig.IntrinsicString;
+  CompilerOptions << " -DCLAMP_ARGC=";
+  CompilerOptions << ClampArgC;
+  CompilerOptions << " -DCLAMP_ARGT=";
+  CompilerOptions << ClampArgT;
+
+  // We have to construct the string outside of the lambda. Otherwise it's
```
alsepkow, line 12246: The helper class really should be copying the string so this can't happen.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11238,1067 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+      LogCommentFmt(L"RunShaderOpTest CallBack. Shader Name: %S", Name);
+
+      // Sanity. Helps diagnose test test configuration issues.
+      VERIFY_IS_TRUE(0 == _stricmp(Name, "LongVectorOp"));
+
+      const size_t DataSize = sizeof(LongVectorOp<T, N>);
+      ShaderData.resize(DataSize);
+
+      // Treat the data as an array of LongVectorOp<T, N> structs.
+      LongVectorOp<T, N> *ShaderStruct = reinterpret_cast<LongVectorOp<T, N> *>(ShaderData.data());
```
alsepkow, line 12270: This casting logic isn't robust for a struct, nor is the copy logic below robust when copying HLSLBool_t or HLSLHalf_t. It's working fine for me for now. But I intend to swap to ByteAddressBuffers for input/output. I'll make this more robust when I do that. I'll do that for this PR.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1067 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+template <typename T, std::size_t N>
+void ExecutionTest::LongVectorOpTestBase(
+    LongVectorOpTestConfig<T> &TestConfig) {
+  WEX::TestExecution::SetVerifyOutput verifySettings(
+      WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
+
+  LogCommentFmt(L"Running LongVectorOpTestBase<%S, %zu>", typeid(T).name(), N);
+
+  CComPtr<ID3D12Device> D3DDevice;
+  if ((!m_HLKModeEnabled && !m_ExperimentalModeEnabled) &&
```
alsepkow, line 12111: This logic is wrong, will fix.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -6033,6 +6168,11 @@ class TableParameterHandler {
     return nullptr;
   }
 
+  template <typename T> std::vector<T> GetTableParamByName(LPCWSTR name) {
+    std::vector<WEX::Common::String> *table = GetTableParamByName(name);
+    return parseStringsToNumbers<T>(GetTableParamByName(name));
```
damyanp, line 6173:
```suggestion
    return parseStringsToNumbers<T>(GetTableParamByName(name));
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+
+    WEX::Logging::Log::Comment(
+        "Device does not support SM 6.9. Can't run these tests.");
+    WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+    return;
+  }
+
+  DeterministicNumberGenerator<T> NumberGenerator(1337);
+  std::vector<T> VecInput1 = {};
+  std::vector<T> VecInput2 = {};
```
damyanp, line 12127: Since the sizes of these are known at compile time you could use `std::array<T, N>`.  Then you could avoid the runtime checks that validate the vectors are the same size.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+
+  // We have to construct the string outside of the lambda. Otherwise it's
+  // cleaned up when the lambda finishes executing but before the shader runs.
+  std::string CompilerOptionsString = CompilerOptions.str();
+
+  // ShaderOpArith.xml defines some boiler plate for the shader code.
+  CComPtr<IStream> TestXML;
+  ReadHlslDataIntoNewStream(L"ShaderOpArith.xml", &TestXML);
+
+  // RunShaderOpTest is a helper function that does all device/resource creation
```
damyanp, line 12257: Ah, and that's pre-existing, so there's not much point digging into why it returns a shared_ptr rather than a unique_ptr.

damyanp (commented):
First round of thoughts.

I can't find the actual shader code.  Did none get added for this change?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Sort of a loaded question. I only saw HLK requirements checking for a given shader model when I was poking around. Assuming we have HLK tests that have optional support or support that isn't tied to a shader model, then we might be relying on skip logic in the HLK today. I'd need to audit the test cases we actually register with the HLK to determine if any of them should have more granular selection logic.

Skips probably could be leveraged in the HLK. But the worrisome part is that it would be very easy for a bug in the skip logic to go unnoticed.

That being said...adding this logic for the long vector test cases isn't really signing us up to fix existing tests.

alsepkow:
> First round of thoughts.
> 
> I can't find the actual shader code. Did none get added for this change?

It's just a raw string inside of LongVectorOpTestBase now. Line ~12200.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+  vector<TYPE, NUM> testInitialize(vector<TYPE, NUM> Vec)
+  {
+    vector<TYPE, NUM> VecCopy = Vec;
+    return VecCopy;
+  }
+
+  vector<TYPE, NUM> testClamp(vector<TYPE, NUM> Vec)
+  {
+    return clamp(Vec, TYPE(CLAMP_ARGC), TYPE(CLAMP_ARGT));
+  }
```
damyanp, line 12213: As these aren't always used, suggest arranging for them to only be included in the shader if they're needed.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
damyanp: You wrote that code though, so why not make it do that yourself? :)

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+    } else if (Tolerance == 0 && VecInput[Index] != VecExpected[Index]) {
+      MismatchedIndexes.push_back(Index);
+    } else if constexpr (std::is_same_v<T, HLSLBool_t>) {
+      // Compiler was very picky and wanted an explicit case for any T that
+      // doesn't implement the operators in the below else. ( > and -). It
+      // wouldn't accept putting this constexpr as an or case with other
+      // statements.
+      if (VecInput[Index] != VecExpected[Index]) {
+        MismatchedIndexes.push_back(i);
+      }
```
damyanp, line 11585: On line 11584, it pushes back `i`, but there is no variable called `i`.  Wondering how this compiles?  Maybe this code is dead?

damyanp: Ahh - so there's two HLSLBool_t arms here, so only the first one is probably being looked at.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+    LongVectorOpTestConfig<T> &TestConfig) {
+  WEX::TestExecution::SetVerifyOutput verifySettings(
+      WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
+
+  LogCommentFmt(L"Running LongVectorOpTestBase<%S, %zu>", typeid(T).name(), N);
+
+  CComPtr<ID3D12Device> D3DDevice;
+  if (!CreateDevice(&D3DDevice, D3D_SHADER_MODEL_6_9)
+      && !m_ExperimentalModeEnabled) {
+    if(m_HLKModeEnabled)
```
bob80905, line 12115: This seems unintentional, but 12119-12122 is dead code because the return isn't under scope in 12115

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+      LogErrorFmt(L"Device does not support SM 6.9. Can't run these tests.");
+      return;
+
+    WEX::Logging::Log::Comment(
+        "Device does not support SM 6.9. Can't run these tests.");
+    WEX::Logging::Log::Result(WEX::Logging::TestResults::Skipped);
+    return;
+  }
+
+  DeterministicNumberGenerator<T> NumberGenerator(1337);
```
bob80905, line 12125: Very leet 😎

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+  }
+
+  // We pass these values into the shader and they're requried to compile. So
+  // they need to set to something.
+  T ClampArgC = 0;
+  T ClampArgT = 0;
+  if (TestConfig.IntrinsicString == "testClamp") {
+    ClampArgC = NumberGenerator.generate();
+    ClampArgT = NumberGenerator.generate();
+    while (ClampArgC >= ClampArgT) {
```
bob80905, line 12148: Could this infinitely loop if T is a bool? maybe the 1337 seed works out so this never happens. Might want a case where if its bool, then just directly force the values. Unless you're excluding clamp for bools.

Separately, nit, the comment should be smaller or equal, or you should adjust the condition so its ClampArgC > ClampArgT

tools/clang/unittests/HLSLExec/ShaderOpTest.cpp
bob80905: Thanks this is great

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+      else
+        LogErrorFmt(L"Unrecognized intrinsic string: %s",
+                    TestConfig.IntrinsicString.c_str());
+    }
+  }
+
+  // 1,2, and 3 are required to compile the shader.
+  // 1. -DOPERATOR : "*" "+" "," "-" "/" etc.
+  // 2. -DOPERAND2: : l.VecInput2 or scalarInput, depending on the test.
+  // 3. -DCLAMP_ARGC, -DCLAMP_ARG2 : clamp arg 1 and 2. Required to compile.
```
bob80905, line 12194: Just a curious thought, up to you if you want to do this.
You might try ifdefing the testClamp function in the shader if the two defs don't exist, and then you can conditionally add the clamp args in 12245. This way there are less options to the compiler that aren't used. This works too. Then 3. isn't required to compile.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Discussed this, adding tolerances is acceptable. But, the tests are passing now. I will add additional tolerance logic in a subsequent change when adding the rest of the coverage.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Fixed in latest revision.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -549,6 +671,15 @@ class ExecutionTest {
       } else {
         LogCommentFmt(L"Debug layer enabled.");
       }
+
+      bool HLKModeEnabled = false;
```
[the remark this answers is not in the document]
alsepkow: It WAS intentional because I thought there was a pattern that required a local variable be used when getting TAEF runtime parameters. But thinking about it, I can't see why it would. Looked at some older test code im familiar with and I dont see that pattern there. 

Removing the local.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11236,1074 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+           DirectX::PackedVector::XMConvertHalfToFloat(other.val);
+  }
+
+  bool operator>(const HLSLHalf_t &other) const {
+    return DirectX::PackedVector::XMConvertHalfToFloat(val) >
+           DirectX::PackedVector::XMConvertHalfToFloat(other.val);
+  }
+
+  // Used by tolerance checks in the tests.
+  bool operator>(float d) const {
```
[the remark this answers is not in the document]
alsepkow: Not needed. We'd get a compiler error if something tried to use it. For completeness I may as well add it though.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Yah, that was a mistake. Fixed.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Good catch. In theory there is probably a seed when T is bool that could lead to an infinite loop if the condition was only >. I think it should be fine if we stick with the >= though?

Also, looking at this. I broke this logic because I changed the string to "TestClamp" instead of "testClamp". I'll swap to comparing the OpType.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Adjusted in latest revision.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Yup, that second else if constexpr shouldn't be there anymore. It must be taken out by the compiler because this same code referencing 'i' is still present and building locally for me. 

Fixed.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
bob80905: Assuming the only valid values for bools are 0 and 1, 
(because of
else if constexpr (std::is_same_v<T, HLSLBool_t>)
      // Anything non-zero, including negative values, is true as a bool. So 0
      // and 1 are all we need.
      Uint16Dist = std::uniform_int_distribution<uint16_t>(0, 1);
)
and assuming that ClampArgC is set to 1 initially,
it will never be the case that you could generate a value for ClampArgT that is strictly greater than ClampArgC.
You'll be generating values forever.
Whether the condition is > or >=, both will infinitely loop forever if ClampArgC gets 1.
And IMO clamping doesn't make much sense for bools, so you might consider just forcing these values in the bool case.

bob80905 (approved):

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11096,6 +11230,1107 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+    InputVector1[Index] = NumberGenerator.generate();
+
+    if (IsVectorBinaryOp)
+      InputVector2[Index] = NumberGenerator.generate();
+  }
+
+  // We pass these values into the shader and they're requried to compile. So
+  // they need to set to something.
+  T ClampArgC = 0;
+  T ClampArgT = 0;
```
damyanp, line 12180: Where do the "C" and "T" bits of these names come from?

alsepkow: Good question. I thought I had seen these in the documentation for clamp, but I checked and thats not where I saw it. Must have been some example code somewhere. I'll change this to 'ClampArgMin' and 'ClampArgMax'

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
tex3d: Splitting tests at this granularity has been a major PITA for working with the playlists and HLK infrastructure.  I don't think it's necessary.

Each of these methods has to be manually expanded to something like this with a new GUID and updated name/description/specification strings when integrating into the HLK, which is quite a pain:
```
  BEGIN_TEST_METHOD(SaturateTest)
    TEST_METHOD_PROPERTY(L"Kits.Description", L"Verifies the saturate instruction.")
    TEST_METHOD_PROPERTY(L"Kits.TestId", L"f1d6cd38-d89c-47d2-bfb4-2e17b94b4d0f")
    TEST_METHOD_PROPERTY(L"Kits.TestName", L"D3D12 - DXIL Core Test - Saturate instruction test")
    TEST_METHOD_PROPERTY(L"Kits.Specification", L"Device.Graphics.WDDM22.AdapterRender.D3D12.DXILCore.ShaderModel60.CoreRequirement")
  END_TEST_METHOD()
```

There's a more streamlined (scripted) process for adding this info into the "Row"s defined in the `ShaderOpArithTable.xml`, which allows TEST_METHOD entries to be copied, unmodified, into the HLK cpp.  But even with this scripted process for helping add/update these in `ShaderOpArithTable.xml`, it was still a pain to manage all of these in HLK database.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
tex3d: You can use `#ifdef _HLK_CONF` to check if code is under HLK build.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Thats why I split them like this. In my past experience with the HLK without this level of granularity we ran into issues if a shipped HLK test had a problem on a particular IHV. The process to rectify that is to grant them a waiver which is basically a regex that checks the test output to match the failure. Then a 'pass' is granted for that scenario for WHQL purposes.

The part that concerns me about using TAEF data driven tests to simplify the test process is that you end up with a gap around your ability to waiver failures in the shipping HLK. We could probably do our best to always run each individual test case...but there are no guarantees around that with a failure.

The other pain point with the taef data table was parsing numbers from strings. I flip flopped on a bunch of these points already and opted to implement something for now and follow up with more discussion. 

I'm obviously open to whatever the team has done historically, and feels is the right balance.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: The st::ShaderOp should own the strings lifetime. But I didn't want to muck with that class. So instead, associating the strings lifetime with the test method seemed like the better option.

tools/clang/unittests/HLSLExec/ShaderOpArith.xml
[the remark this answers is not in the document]
alsepkow: He mentioned that all of the names used in the xml were meaningless without it. But after a few typos I discovered that wasn't true. So, I'm not sure that it really matters.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Discussed a little more offline. Will leave this as is for now. We can discuss a different design further for future iterations.

include/dxc/Test/HlslTestUtils.h
```diff
@@ -515,6 +528,31 @@ inline bool isnanFloat16(uint16_t val) {
 uint16_t ConvertFloat32ToFloat16(float val) throw();
 float ConvertFloat16ToFloat32(uint16_t val) throw();
 
+inline bool CompareDoubleULP(
+    const double &fsrc, const double &fref, int64_t ULPTolerance,
+    hlsl::DXIL::Float32DenormMode mode = hlsl::DXIL::Float32DenormMode::Any) {
```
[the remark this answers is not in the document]
damyanp: style: naming

Also, it's probably better to pass these by value rather than reference.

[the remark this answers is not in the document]
damyanp: Local variables in this function also need to be updated to match naming conventions.

tools/clang/unittests/HLSLExec/CMakeLists.txt
```diff
@@ -39,3 +39,10 @@ endif()
 file(TO_NATIVE_PATH "${CMAKE_CURRENT_SOURCE_DIR}" DOS_STYLE_SOURCE_DIR)
 file(TO_NATIVE_PATH "${TAEF_BIN_DIR}" DOS_TAEF_BIN_DIR)
 configure_file(ExecHLSLTests.vcxproj.user.txt ExecHLSLTests.vcxproj.user)
+
+# Copy the ShaoderOpArith.xml file to the output directory. It's used by the
+# exec tests and it's convient to have it copied here if you want to easily copy
+# the tests to another machine after building.
```
damyanp, line 45:
```suggestion
# Copy the ShaderOpArith.xml file to the output directory. It's used by the
# exec tests and it's convenient to have it copied here if you want to easily copy
# the tests to another machine after building.
```

damyanp (commented):
Some more notes

include/dxc/Test/HlslTestUtils.h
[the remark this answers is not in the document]
alsepkow: That was intentional. I added very little to this file and this function is largely a copy of the one right below. I almost templated it.

So, I think it makes sense to match naming conventions of the file?

include/dxc/Test/HlslTestUtils.h
```diff
@@ -459,15 +470,17 @@ inline bool GetTestParamUseWARP(bool defaultVal) {
 
 #ifdef FP_SUBNORMAL
 
-inline bool isdenorm(float f) { return FP_SUBNORMAL == std::fpclassify(f); }
+template <typename T = float> inline bool isdenorm(T f) {
```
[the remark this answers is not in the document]
alsepkow: I was concerned about changing the signature in a way that breaks other callers. I wasn't sure if it might get used by someone that consumes the source vpack. I could have done more due diligence to figure that out.

But I wanted to make sure that a caller doing:
double d;
isdenorm(d) 

would still compile.

Maybe it would be better to just define:
inline bool isdenorm(double d)
and
inline bool isdenorm(float f)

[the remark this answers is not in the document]
alsepkow: And it made more sense when I had initially templated Compare*ULP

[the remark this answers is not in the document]
damyanp: `template <typename T> inline bool isdenorm(T f) {` would be fine

include/dxc/Test/HlslTestUtils.h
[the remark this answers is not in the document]
damyanp: So generally our policy to coding conventions is that new code, in files that are not consistent, follow the new naming conventions.

In particular the "f" prefix of `fsrc` and `fdst` I think are some kind of hungarian for "float", so they're actually misnamed according to the rough set of conventions that the original version used.

At this point we should just use llvm conventions rather than spend energy figuring out what these would have been called according to the previous (probably ad-hoc) conventions used in other parts of the file.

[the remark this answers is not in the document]
alsepkow: Updated, and updated the types. Diff was wrong.

damyanp (approved):
I think this is ready to go into staging to help unblock people who need some execution tests to work from.

If you've created issues for follow up can you paste here?  Could consider just an umbrella one for general improvements to the long vector tests for the preview would be enough to get started.

alsepkow:
> I think this is ready to go into staging to help unblock people who need some execution tests to work from.
> 
> If you've created issues for follow up can you paste here? Could consider just an umbrella one for general improvements to the long vector tests for the preview would be enough to get started.

[7363]  https://github.com/microsoft/DirectXShaderCompiler/issues/7363

Issue #7120, opened 2025-02-04
generating native vectors operations for elementwise intrinsics


pow2clk:

farzonl:
Needs one final test to do a simple shader which each intrinsic to confirm implementation boundary.

V-FEXrt:
- https://godbolt.org/z/qxKKa75hY `f16tof32`/`f32tof16` doesn't currently vectorize, since it is backed by `legacyFXXtoFYY` which was decided here that is shouldn't vectorize https://github.com/microsoft/hlsl-specs/pull/584

PR #7321, opened 2025-04-07
[SM6.9] Enable trivial native vector Dxil Operations plus a few

pow2clk:
REVIEWER GUIDANCE: I recommend initially just looking at this commit: 2bdf33e357e699d4fa088128760ebae0cebd9b7a The rest are NFC formatting and style changes or autogenerated code changes.  << Remove before submitting.

This enables the generation of native vector DXIL Operations
that are "trivial", meaning they take only a single DXOp Call
instruction to implement as well as a few others that either only took
such a call and some llvm operations or were of particular interest for
other reasons.

This involves allowing the overloads by adding the vector indication in
hctdb, altering the lowering to maintain the vectors instead of
scalarizing them, and a few sundry changes to fix issues along the way.

The "trivial" dxil operations that return a different value from the
overload type had to be moved out of the way and given their own
lowering function so that the main function could generate vectors
conditional on the version and vector type. These will be added in a
later change.

While the long vector supporting intrinsics that weren't given this
treatment will continue to generate scalarized operations, some of them
needed some work as well. The dot product for float vectors longer than
4 had to take the integer fallback path, which required some small
modifications and a rename.
Additionally, a heuristic for pow that malfunctioned with too many
elements had to have a limit placed on it.

Since the or()/and()/select() intrinsics translate directly to LLVM ops,
they can have their lowering scalarization removed and what future
scalarization might be needed by the current version can be done by
later passes as with other LLVM operators.

An issue with a special value used to represent unassigned dimensions had
to be addressed since new dimensions can exceed that value. It's now
MAX_INT.

Contributes to https://github.com/microsoft/DirectXShaderCompiler/issues/7120, but I'd prefer to leave it open until all
intrinsics are covered

### DIFF
--- a/lib/HLSL/HLOperationLower.cpp
+++ b/lib/HLSL/HLOperationLower.cpp
@@ -2510,7 +2510,11 @@ Value *ExpandDot(Value *Arg0, Value *Arg1, unsigned VecSize, hlsl::OP *OP,
                  DXIL::OpCode MadOpCode = DXIL::OpCode::IMad) {
   Value *Elt0 = Builder.CreateExtractElement(Arg0, (uint64_t)0);
   Value *Elt1 = Builder.CreateExtractElement(Arg1, (uint64_t)0);
-  Value *Result = Builder.CreateMul(Elt0, Elt1);
+  Value *Result;
+  if (Elt0->getType()->isFloatingPointTy())
+    Result = Builder.CreateFMul(Elt0, Elt1);
+  else
+    Result = Builder.CreateMul(Elt0, Elt1);
   for (unsigned Elt = 1; Elt < VecSize; ++Elt) {
     Elt0 = Builder.CreateExtractElement(Arg0, Elt);
     Elt1 = Builder.CreateExtractElement(Arg1, Elt);
--- a/lib/HLSL/HLOperationLower.cpp
+++ b/lib/HLSL/HLOperationLower.cpp
@@ -474,8 +474,8 @@ Value *TrivialDxilOperation(Function *dxilFunc, OP::OpCode opcode,
 // Utility objects `HlslOp` and `Builder` are used to create a call to the given
 // `DxilFunc` with `RefArgs` arguments.
 Value *TrivialDxilVectorOperation(Function *Func, OP::OpCode Opcode,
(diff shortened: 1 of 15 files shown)


pow2clk:
REVIEWER GUIDANCE: I recommend initially just looking at this commit: 2bdf33e357e699d4fa088128760ebae0cebd9b7a The rest are NFC formatting and style changes or autogenerated code changes.

pow2clk:
Note that an IR test is incoming.

llvm-beanz (changes requested):
This is a mess to review. HLOperationLower.cpp has a massive volume of unrelated changes to code style. Pointing reviewers at the first commit that occurs before the reformatting is extremely unhelpful.

As I go through the first commit, GitHub won't let me comment on out-of-date lines. This means as a reviewer I need to review both he single commit and the whole change in parallel to see if what I'm commenting on is fixed and to put a comment in the appropriate place.

If you want to do a mass reformat, please put it in a separate PR.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -459,17 +467,42 @@ Value *TrivialDxilOperation(Function *dxilFunc, OP::OpCode opcode,
...
+}
+
+// Generates a DXIL operation with the overloaded type based on `Ty` and return
+// type `RetTy`. When Ty is a vector, it will either generate per-element calls
+// for each vector element and reconstruct the vector type from those results or
+// operate on and return native vectors depending on vector size and the value
+// of `SupportsVectors`, which is deteremined by version and opcode support.
+Value *TrivialDxilOperation(OP::OpCode Opcode, ArrayRef<Value *> Args, Type *Ty,
+                            Type *RetTy, OP *OP, IRBuilder<> &Builder,
+                            bool SupportsVectors = false) {
```
tex3d, line 492: Instead of the caller supplying `SupportsVectors`, couldn't this be looked up by the operation and the module?  Something like:

```cpp
if (Ty->isVectorTy() && Ty->getVectorNumElements() > 1 &&
    OP->GetModule()->GetHLModule()->GetShaderModel()->IsSM69Plus() &&
    OP::IsOverloadLegal(Opcode, Ty)) {
  // ...
}
```

PR #7324, opened 2025-04-08
[SM6.9] Enable trivial native vector Dxil Operations plus a few

### LINKED ISSUE #7297: vectors longer than 254 are truncated to 1 when passed to intrinsics
**Description**

```c++
template<int N>
int VectorLength(vector<float, N>) { return N; }

export int Test1(vector<float, 254> Input) {
    return VectorLength(mul(Input, 1));
    // ret i32 254
}

export int Test2(vector<float, 255> Input) {
    return VectorLength(mul(Input, 1));
    // EXPECTED: ret i32 255
    // ACTUAL  : ret i32 1
}
```

**Steps to Reproduce**

Compile the above shader with -Tlib_6_9 to see the repro.

llvm-beanz:
This enables the generation of native vector DXIL Operations
that are "trivial", meaning they take only a single DXOp Call
instruction to implement as well as a few others that either only took
such a call and some llvm operations or were of particular interest for
other reasons.

This involves allowing the overloads by adding the vector indication in
hctdb, altering the lowering to maintain the vectors instead of
scalarizing them, and a few sundry changes to fix issues along the way.

The "trivial" dxil operations that return a different value from the
overload type had to be moved out of the way and given their own
lowering function so that the main function could generate vectors
conditional on the version and vector type. These will be added in a
later change.

While the long vector supporting intrinsics that weren't given this
treatment will continue to generate scalarized operations, some of them
needed some work as well. The dot product for float vectors longer than
4 had to take the integer fallback path, which required some small
modificaitons and a rename.
Additionally, a heuristic for pow that malfunctioned with too many
elements had to have a limit placed on it.

Since the or()/and()/select() intrinsics translate directly to LLVM ops,
they can have their lowering scalarization removed and what future
scalarization might be needed by the current version can be done by
later passes as with other LLVM operators.

An issue with a special value used to represent unassined dimensions had
to be addressed since new dimensions can exceed that value. It's now
MAX_INT.

Contributes to #7120, but I'd prefer to leave it open until all
intrinsics are covered

Primary work by @pow2clk

Fixes #7297 & #7120

### DIFF
--- a/lib/HLSL/HLOperationLower.cpp
+++ b/lib/HLSL/HLOperationLower.cpp
@@ -525,7 +525,7 @@ Value *TrivialDxilOperation(OP::OpCode opcode, ArrayRef<Value *> refArgs,
 Value *TrivialUnaryOperationRet(CallInst *CI, IntrinsicOp IOP,
                                 OP::OpCode OpCode,
                                 HLOperationLowerHelper &Helper,
-                                HLObjectOperationLowerHelper *pObjHelper,
+                                HLObjectOperationLowerHelper *,
                                 bool &Translated) {
   Value *Src = CI->getArgOperand(HLOperandIndex::kUnaryOpSrc0Idx);
   Type *Ty = Src->getType();
@@ -539,13 +539,13 @@ Value *TrivialUnaryOperationRet(CallInst *CI, IntrinsicOp IOP,
   return TrivialDxilOperation(OpCode, Args, Ty, RetTy, OP, Builder);
 }
 
-Value *TrivialDxilUnaryOperation(OP::OpCode OpCode, Value *src,
+Value *TrivialDxilUnaryOperation(OP::OpCode OpCode, Value *Src,
                                  hlsl::OP *Op, IRBuilder<> &Builder,
                                  bool SupportsVectors = false) {
   Type *Ty = src->getType();
 
   Constant *OpArg = Op->GetU32Const((unsigned)OpCode);
-  Value *Args[] = {OpArg, src};
+  Value *Args[] = {OpArg, Src};
 
   return TrivialDxilOperation(OpCode, Args, Ty, Ty, Op, Builder,
                               SupportsVectors);
--- a/lib/HLSL/HLOperationLower.cpp
+++ b/lib/HLSL/HLOperationLower.cpp
@@ -484,16 +484,17 @@ Value *TrivialDxilVectorOperation(Function *Func, OP::OpCode Opcode,
 // Generates a DXIL operation with the overloaded type based on `Ty` and return
 // type `RetTy`. When Ty is a vector, it will either generate per-element calls
 // for each vector element and reconstruct the vector type from those results or
-// operate on and return native vectors depending on vector size and the value
-// of `SupportsVectors`, which is deteremined by version and opcode support.
+// operate on and return native vectors depending on vector size and the
+// legality of the vector overload.
 Value *TrivialDxilOperation(OP::OpCode opcode, ArrayRef<Value *> refArgs,
                             Type *Ty, Type *RetTy, OP *hlslOP,
-                            IRBuilder<> &Builder,
-                            bool SupportsVectors = false) {
+                            IRBuilder<> &Builder) {
 
   // If supported and the overload type is a vector with more than 1 element,
   // create a native vector operation.
-  if (SupportsVectors && Ty->isVectorTy() && Ty->getVectorNumElements() > 1) {
+  if (Ty->isVectorTy() && Ty->getVectorNumElements() > 1 &&
+      hlslOP->GetModule()->GetHLModule().GetShaderModel()->IsSM69Plus() &&
+      OP::IsOverloadLegal(opcode, Ty)) {
     Function *dxilFunc = hlslOP->GetOpFunc(opcode, Ty);
     return TrivialDxilVectorOperation(dxilFunc, opcode, refArgs, Ty, hlslOP,
                                       Builder);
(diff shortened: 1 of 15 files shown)


lib/HLSL/HLOperationLower.cpp
```diff
@@ -459,12 +467,40 @@ Value *TrivialDxilOperation(Function *dxilFunc, OP::OpCode opcode,
...
+// type `Ty`.
+// Utility objects `HlslOp` and `Builder` are used to create a call to the given
+// `DxilFunc` with `RefArgs` arguments.
+Value *TrivialDxilVectorOperation(Function *Func, OP::OpCode Opcode,
+                                  ArrayRef<Value *> Args, Type *Ty, OP *OP,
+                                  IRBuilder<> &Builder) {
+  if (!Ty->isVoidTy())
+    return Builder.CreateCall(Func, Args, OP->GetOpCodeName(Opcode));
+  else
+    return Builder.CreateCall(Func, Args); // Cannot add name to void.
```
llvm-beanz, line 482: see: https://llvm.org/docs/CodingStandards.html#don-t-use-else-after-a-return

```suggestion
  return Builder.CreateCall(Func, Args); // Cannot add name to void.
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -6606,7 +6606,7 @@ bool HLSLExternalSource::MatchArguments(
   argTypes.clear();
   const bool isVariadic = IsVariadicIntrinsicFunction(pIntrinsic);
 
-  static const UINT UnusedSize = 0xFF;
+  static const UINT UnusedSize = UINT_MAX;
```
damyanp, line 6609: Is this a fix for #7297?

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-scalarized-intrinsics.hlsl
```diff
@@ -0,0 +1,147 @@
...
+  vec1 = dot(vec1, vec2);
+
+  vector<bool, 8> bvec = b;
+  // CHECK: or i1
+  // CHECK: or i1
+  // CHECK: or i1
+  // CHECK: or i1
+  // CHECK: or i1
+  // CHECK: or i1
+  // CHECK: or i1
```
damyanp, line 125:
```suggestion
  // CHECK-COUNT-7: or i1
```

Which does make me wonder why it isn't 8 though.

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-trivial-scalarized-intrinsics.hlsl
```diff
@@ -0,0 +1,87 @@
...
+#define CALLARGS(x,y,z) x, i
+// ARITY 5 is used for 1 vec + uint4 mask for wavemultiprefix*
+#elif ARITY == 5
+#define CALLARGS(x,y,z) x, m
+#endif
+
+StructuredBuffer< vector<TYPE, 8> > buf;
+ByteAddressBuffer rbuf;
+
+float4 main(uint i : SV_PrimitiveID, uint4 m : M) : SV_Target {
```
damyanp, line 65: @alsepkow - the approach used in this file might be interesting to think about for the HLKs you're working on

lib/HLSL/HLOperationLower.cpp
```diff
@@ -484,55 +519,81 @@ Value *TrivialDxilOperation(OP::OpCode opcode, ArrayRef<Value *> refArgs,
...
 }
 
-Value *TrivialDxilUnaryOperation(OP::OpCode opcode, Value *src,
-                                 hlsl::OP *hlslOP, IRBuilder<> &Builder) {
-  return TrivialDxilUnaryOperationRet(opcode, src, src->getType(), hlslOP,
-                                      Builder);
+Value *TrivialDxilUnaryOperation(OP::OpCode OpCode, Value *Src, hlsl::OP *Op,
+                                 IRBuilder<> &Builder,
+                                 bool SupportsVectors = false) {
+  Type *Ty = src->getType();
```
damyanp, line 545:
```suggestion
  Type *Ty = Src->getType();
```

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: I believe so. I have a small update to this line coming to not use the Windows typedefs and use C++ limits.

lib/HLSL/HLOperationLower.cpp
llvm-beanz: This is in my latest change.

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-scalarized-intrinsics.hlsl
llvm-beanz: Sadly DXC's FileCheck doesn't support `CHECK-COUNT`, that's a newer feature:
https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/CommandGuide/FileCheck.rst

`CHECK-COUNT` was added in 2018 with this patch: https://reviews.llvm.org/D54336

llvm-beanz: As to why this is 7. The IR is basically:
```
vec[0] & vec[1] & vec[2] & vec[3] & vec[4] & vec[5] & vec[6] & vec[7]
```
Eight elements, 7 operations.

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-intrinsics.hlsl
```diff
@@ -0,0 +1,391 @@
+// RUN: %dxc -T cs_6_9 -enable-16bit-types -DNUM=7   %s | FileCheck %s
+// RUN: %dxc -T cs_6_9 -enable-16bit-types -DNUM=125 %s | FileCheck %s
```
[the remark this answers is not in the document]
llvm-beanz: Yea, I'm also going to throw 2 into the mix, because the short vectors should also be vectorizing now too.

lib/HLSL/HLOperationLower.cpp
```diff
@@ -459,12 +467,39 @@ Value *TrivialDxilOperation(Function *dxilFunc, OP::OpCode opcode,
...
+// of `SupportsVectors`, which is deteremined by version and opcode support.
 Value *TrivialDxilOperation(OP::OpCode opcode, ArrayRef<Value *> refArgs,
                             Type *Ty, Type *RetTy, OP *hlslOP,
-                            IRBuilder<> &Builder) {
+                            IRBuilder<> &Builder,
+                            bool SupportsVectors = false) {
+
+  // If supported and the overload type is a vector with more than 1 element,
+  // create a native vector operation.
+  if (SupportsVectors && Ty->isVectorTy() && Ty->getVectorNumElements() > 1) {
```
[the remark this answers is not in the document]
llvm-beanz: Great catch @tex3d!

Fixing this actually corrects at least some of the missing intrinsics you cited in your comment above. I'll roll this change through and do a pass at removing spurious diffs too. I think this is a fantastic simplification of the change.

llvm-beanz:
After adapting to @tex3d's feedback this really simplified and cleaned up a lot. Please let me know if there are other changes I should make, but I think I got everything that has been suggested.

damyanp (approved):
FWIW, LGTM.

damyanp (approved):

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-scalarized-intrinsics.hlsl
tex3d, line 1: Nit: Given that only `WaveMatch` is truly "scalarized", the name of the test isn't great.  Perhaps "expanded" instead.  But I don't think this justifies an update on its own.  This can be cleaned up later.

tex3d (approved):
Looks good!

PR #7332, opened 2025-04-10
Fix field names in long vector DICompositeType

simoll:
Fix OOB accesses for debug info vector field names for vectors of length >4.

### DIFF
--- a/tools/clang/lib/CodeGen/CGDebugInfo.cpp
+++ b/tools/clang/lib/CodeGen/CGDebugInfo.cpp
@@ -1047,8 +1047,15 @@ bool CGDebugInfo::TryCollectHLSLRecordElements(const RecordType *Ty,
     unsigned VecSize = hlsl::GetHLSLVecSize(QualTy);
     unsigned ElemSizeInBits = CGM.getContext().getTypeSize(ElemQualTy);
     unsigned CurrentAlignedOffset = 0;
+    std::string FieldNameBuf;
     for (unsigned ElemIdx = 0; ElemIdx < VecSize; ++ElemIdx) {
-      StringRef FieldName = StringRef(&"xyzw"[ElemIdx], 1);
+      StringRef FieldName;
+      if (VecSize <= 4) {
+        FieldName = StringRef(&"xyzw"[ElemIdx], 1);
+      } else {
+        FieldNameBuf = "c" + std::to_string(ElemIdx);
+        FieldName = FieldNameBuf;
+      }
       CurrentAlignedOffset =
           llvm::RoundUpToAlignment(CurrentAlignedOffset, AlignBits);
       llvm::DIType *FieldType =
--- a/tools/clang/lib/CodeGen/CGDebugInfo.cpp
+++ b/tools/clang/lib/CodeGen/CGDebugInfo.cpp
@@ -1047,14 +1047,16 @@ bool CGDebugInfo::TryCollectHLSLRecordElements(const RecordType *Ty,
     unsigned VecSize = hlsl::GetHLSLVecSize(QualTy);
     unsigned ElemSizeInBits = CGM.getContext().getTypeSize(ElemQualTy);
     unsigned CurrentAlignedOffset = 0;
-    std::string FieldNameBuf;
+    SmallString<8> FieldNameBuf;
     for (unsigned ElemIdx = 0; ElemIdx < VecSize; ++ElemIdx) {
       StringRef FieldName;
       if (VecSize <= 4) {
         FieldName = StringRef(&"xyzw"[ElemIdx], 1);
       } else {
-        FieldNameBuf = "c" + std::to_string(ElemIdx);
-        FieldName = FieldNameBuf;
+        FieldNameBuf.clear();
+        llvm::raw_svector_ostream OS(FieldNameBuf);
+        OS << 'c' << ElemIdx;
+        FieldName = OS.str();
       }
       CurrentAlignedOffset =
           llvm::RoundUpToAlignment(CurrentAlignedOffset, AlignBits);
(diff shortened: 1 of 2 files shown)


tools/clang/lib/CodeGen/CGDebugInfo.cpp
```diff
@@ -1047,8 +1047,15 @@ bool CGDebugInfo::TryCollectHLSLRecordElements(const RecordType *Ty,
...
     unsigned ElemSizeInBits = CGM.getContext().getTypeSize(ElemQualTy);
     unsigned CurrentAlignedOffset = 0;
+    std::string FieldNameBuf;
     for (unsigned ElemIdx = 0; ElemIdx < VecSize; ++ElemIdx) {
-      StringRef FieldName = StringRef(&"xyzw"[ElemIdx], 1);
+      StringRef FieldName;
+      if (VecSize <= 4) {
+        FieldName = StringRef(&"xyzw"[ElemIdx], 1);
+      } else {
+        FieldNameBuf = "c" + std::to_string(ElemIdx);
```
damyanp, line 1056: Is this "c" prefix naming convention used elsewhere?

@jeffnn - will this end up going directly through to PIX?

jeffnn: Schweet

jeffnn (approved):

tools/clang/lib/CodeGen/CGDebugInfo.cpp
```diff
@@ -1047,8 +1047,15 @@ bool CGDebugInfo::TryCollectHLSLRecordElements(const RecordType *Ty,
     unsigned VecSize = hlsl::GetHLSLVecSize(QualTy);
     unsigned ElemSizeInBits = CGM.getContext().getTypeSize(ElemQualTy);
     unsigned CurrentAlignedOffset = 0;
+    std::string FieldNameBuf;
```
llvm-beanz, line 1050: This way we don't have dynamic allocations until we get pretty big.
```suggestion
    SmallString<16> FieldNameBuf;
```

tools/clang/lib/CodeGen/CGDebugInfo.cpp
```diff
@@ -1047,8 +1047,15 @@ bool CGDebugInfo::TryCollectHLSLRecordElements(const RecordType *Ty,
...
     for (unsigned ElemIdx = 0; ElemIdx < VecSize; ++ElemIdx) {
-      StringRef FieldName = StringRef(&"xyzw"[ElemIdx], 1);
+      StringRef FieldName;
+      if (VecSize <= 4) {
+        FieldName = StringRef(&"xyzw"[ElemIdx], 1);
+      } else {
+        FieldNameBuf = "c" + std::to_string(ElemIdx);
+        FieldName = FieldNameBuf;
```
llvm-beanz, line 1057: This is at least one, but probably two calls to `new`.

There are a couple different approaches we could take to reduce this. One would be a `Twine`, the other would be to have a `raw_svector_ostream` print directly into `FieldNameBuf`.
That would be something like:
```suggestion
        FieldNameBuf.clear();
        raw_svector_ostream OS(FieldNameBuf);
        OS << "c" << ElemIdx;
        FieldName = OS.str();
```

tools/clang/lib/CodeGen/CGDebugInfo.cpp
tex3d: I agree with using `SmallString`, though `16` seems even larger than necessary.  I think `"c1023"` is the largest string we can get here, so I think `8` should be sufficient, don't you?

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-field-di.hlsl
```diff
@@ -0,0 +1,33 @@
+// RUN: %dxc -Zi -Qembed_debug -T lib_6_9 %s -DNUM=8 | FileCheck %s  --check-prefix=CHECK-LONG
+// RUN: %dxc -Zi -Qembed_debug -T lib_6_9 %s -DNUM=4 | FileCheck %s  --check-prefix=CHECK-SHORT
+
+// Test debug info for short and long vector types
+
+RWByteAddressBuffer buf;
+
+export vector<float, NUM> lv_global_arr_ret() { // expected-error{{invalid value, valid range is between 1 and 1024 inclusive}}
```
tex3d, line 8: Keeping `// expected-error{{invalid value, valid range is between 1 and 1024 inclusive}}` seems misleading, doesn't it?

tex3d (commented):
Looks mostly good with some good suggestions/comments from others.

tools/clang/lib/CodeGen/CGDebugInfo.cpp
llvm-beanz: Sure. It’s also only 8 bytes of stack space but we probably don’t need more than 5 bytes (8 is the practical minimum since the structure has 8-byte alignment requirement).

simoll: Updated to use `SmallString<8>`

tools/clang/lib/CodeGen/CGDebugInfo.cpp
simoll: Using `raw_svector_ostream` now

tools/clang/test/CodeGenDXIL/hlsl/types/longvec-field-di.hlsl
simoll: Yep, stripped that leftover from the test

llvm-beanz (approved):

tools/clang/lib/CodeGen/CGDebugInfo.cpp
simoll: This thread is still open. Is `c` an acceptable prefix for those fields?

damyanp: Nobody else seems to have an opinion.  We can always change if we get feedback on it while people are debugging shaders with it.

I assume the "c" stands for "component"?

jeffnn: c for component makes sense to me. I asked Dwayne, who has extensive experience in ML scenarios, also and he's OK with it.

tex3d (approved):

PR #7337, opened 2025-04-11
[SER] Basic execution test to build up DXR test infra

simoll:
- Adds `CreateDXRDevice` and `RunDXRTest` helpers for triangle dxr tests. Will extend this to procedural in followups.
- Adds a trivial `SERBasicTest`. As SER functionality is still missing in DXC, this only uses the `dx::HitObject` type.


simoll:
Not pursuing those as dxc execution tests

PR #6715, opened 2024-06-21
CMake: Add initial support for MinGW-w64

raedrizqie:

PR #7346, opened 2025-04-15
Add HctGen of DXIL.rst back to build without LLVM_BUILD_DOCS required

tex3d:
HctGen of DXIL.rst should happen on every ordinary build, and be updated with other HctGen modified files.

This isn't about building the doc, it's about updating it when definitions change in hctdb.py.

We've been missing updates to DXIL.rst for quite a while due to this issue, introduced [here](https://github.com/microsoft/DirectXShaderCompiler/pull/6715/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR768-R770).

This also brings DXIL.rst up to date.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -762,9 +762,7 @@ if (LLVM_INCLUDE_DOCS)
   add_subdirectory(docs)
 endif()
 
-if (LLVM_BUILD_DOCS)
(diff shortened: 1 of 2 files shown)


damyanp (approved):
Seems like this is something we'd want to have happen as a matter of course.  Nice catch!

llvm-beanz (approved):

Issue #7304, opened 2025-04-03
Implement preview cooperative vectors API


damyanp:
The API described in [Proposal 0031](https://github.com/microsoft/hlsl-specs/blob/main/proposals/0031-hlsl-vector-matrix-operations.md) needs to have an implementation added, with tests.

A prototype can be found here: https://github.com/damyanp/DirectXShaderCompiler/tree/cv_api_staging/tools/clang/test/CodeGenDXIL/hlsl/linalg

### Distributing the header

For the preview it is fine to require developers to do something like `#include <linalg.h>`, but we still need to figure out how package this header file up with the preview package.

Issue #7336, opened 2025-04-11
CoopVec Diagnose and test invalid use of CoopVec builtins in HLSL


simoll:
Diagnose problems specific to CoopVec during sema and test for those diagnostics.

Current CoopVec implementation has no CoopVec-specific diagnostics on HLSL level. Invalid use of the builtins is rejected only by the DXIl validator.

damyanp:
See https://devblogs.microsoft.com/directx/shader-model-6-9-and-the-future-of-cooperative-vector/

PR #7350, opened 2025-04-15
[CoopVec] Add Linear Algebra common header with tests

bob80905:
This PR introduces the linear algebra header file, and places it in a location that is by default included in all HLSL compilation.
The builtins in the API aren't yet defined, and depend on the #7290 PR merging first.
The tests that have been added have temporary diagnostic messages while 7290 is in progress. They will need to be updated.
Open to feedback on better / suggested error messages, or whether there shouldn't be any sema-level validation for these errors.

Fixes [#7304](https://github.com/microsoft/DirectXShaderCompiler/issues/7304)


tools/clang/lib/Headers/hlsl/linalg.h
```diff
@@ -0,0 +1,172 @@
+// Header for linear algebra APIs.
```
damyanp, line 1: Should this be in `tools/clang/lib/Headers/hlsl/dx` since it is dx specific?

tools/clang/test/CodeGenDXIL/hlsl/linalg/make-interp-vec-errors.hlsl
```diff
@@ -0,0 +1,38 @@
+// RUN: not %dxc -T lib_6_9 %s 2>&1 | FileCheck %s
+
+#include "linalg.h"
```
damyanp, line 3:
```suggestion
#include <linalg.h>
```

tools/clang/lib/Headers/hlsl/linalg.h
bob80905: Moved.

tools/clang/test/CodeGenDXIL/hlsl/linalg/make-interp-vec-errors.hlsl
bob80905: I've just removed the line entirely since it's unneeded, the file is included by default when compiling a shader.

damyanp: Oh wow, so just putting in that directory makes it get included by default?  That's awesome!

tools/clang/lib/Headers/hlsl/dx/linalg.h
```diff
@@ -0,0 +1,172 @@
+// Header for linear algebra APIs.
```
damyanp, line 1: Wonder if we should use ifdef's to exclude this if language version is before 2021 and SM < 6.9?

tools/clang/test/CodeGenDXIL/hlsl/linalg/mat-vec-mul-transpose-errors.hlsl
```diff
@@ -0,0 +1,29 @@
...
+ByteAddressBuffer Buf;
+
+export float4 Test1(vector<float, 4> Input) {
+  using namespace dx::linalg;
+
+  MatrixRef<DATA_TYPE_FLOAT16, 4, 4, MATRIX_LAYOUT_ROW_MAJOR, true> Matrix = {
+      Buf, 0, 0};
+
+  // clang-format off
+  // CHECK: error: something about transposing not supported for rowmajor / colmajor layouts
```
damyanp, line 12: presumably this check fails?

bob80905: Yes, it's a placeholder for now. There should be other placeholders too. Seeking feedback on whether this should be an error at the sema-level, or whether this should be left as a validation error. If Sema, I can come up with a better message, and change the code to emit the modified error message. If left to the validator, I'd break this test out to a new file and test for the validator error.

damyanp: I don't think you're on the hook for implementing these diagnostics at all at the moment.  They should come from the underlying builtin.  The current PR has this just done in DXIL validation, but we also have this issue: https://github.com/microsoft/DirectXShaderCompiler/issues/7336

bob80905: I see. Well, I've written the extra tests, so I can merge them in and leave them as XFAIL for simon to modify as he sees fit.
Yes, I thought I was on the hook for 7336. I'll set the new tests with placeholdeer diagnostics to XFAIL.

tools/clang/test/CodeGenDXIL/hlsl/linalg/.clang-format
```diff
@@ -0,0 +1,2 @@
+BasedOnStyle: LLVM
```
damyanp, line 1: I think we usually turn off clang-format for tests?  I had it enabled while I was working on the proposal so examples would be formatted nicely for the document, but it might be preferable to just not use it on these test files so you can avoid having to turn it off/on again around CHECK lines.

I don't mind what you do here, but just wanted to be sure you weren't suffering with a choice I made now that the requirements have changed.

tools/clang/test/CodeGenDXIL/hlsl/linalg/mat-vec-muladd.hlsl
```diff
@@ -0,0 +1,96 @@
...
+      biasVector1);
+  layer1 = max(layer1, 0);
+
+  vector<float, 3> output = MulAdd<float>(
+      matrix2, MakeInterpretedVector<DATA_TYPE_FLOAT8_E4M3>(layer1),
+      biasVector2);
+  output = exp(output);
+
+  return output;
+}
```
damyanp, line 94: This isn't testing anything and is left over from work I was doing for the proposal.  Maybe it's worth keeping around so we can be confident that the examples in the proposal work as expected?

tools/clang/test/CodeGenDXIL/hlsl/linalg/outerproductaccumulate.hlsl
```diff
@@ -0,0 +1,19 @@
...
+
+export void Test4(vector<half, 128> Input1, vector<half, 64> Input2) {
+  using namespace dx::linalg;
+
+  RWMatrixRef<DATA_TYPE_FLOAT16, 128, 64, MATRIX_LAYOUT_OUTER_PRODUCT_OPTIMAL>
+      matrix = {RWBuf, 0, 0};
+
+  // clang-format off
+  // CHECK: call void @dx.op.outerProductAccumulate.v128f16.v64f16(i32 307, <128 x half> %{{.+}}, <64 x half> %{{.+}}, %dx.types.Handle %{{.+}}, i32 0, i32 8, i32 3, i32 0)
+  // __builtin_OuterProductAccumulate(Input1, Input2, RWBuf, 0, DATA_TYPE_FLOAT16, MATRIX_LAYOUT_OUTER_PRODUCT_OPTIMAL, 0);
```
damyanp, line 15: Should commented-out the call to builtin be removed?

```suggestion
```

tools/clang/test/CodeGenDXIL/hlsl/linalg/vector-matrix-proposal.hlsl
```diff
@@ -0,0 +1,143 @@
+// RUN: %dxc -I %hlsl_headers -T lib_6_9 -enable-16bit-types %s
```
damyanp, line 1: These are the code samples that appear in the proposal.  Do you think they should stay, since they're not really testing anything?

tools/clang/test/SemaHLSL/hlsl/linalg/outerproductaccumulate-errors.hlsl
```diff
@@ -0,0 +1,53 @@
...
+export void Test4(vector<half, 128> Input1, vector<half, 64> Input2) {
+  using namespace dx::linalg;
+
+  RWMatrixRef<DATA_TYPE_FLOAT16, 128, 64, MATRIX_LAYOUT_OUTER_PRODUCT_OPTIMAL, true>
+      matrix = {RWBuf, 0, 0};
+
+  // clang-format off
+  // expected-error@+5{{no matching function for call to 'OuterProductAccumulate'}}
+  // expected-note@dx/linalg.h:157{{candidate template ignored: could not match 0 against 1}}
+  // __builtin_OuterProductAccumulate(Input1, Input2, RWBuf, 0, DATA_TYPE_FLOAT16, MATRIX_LAYOUT_OUTER_PRODUCT_OPTIMAL, 0);
```
damyanp, line 17:
```suggestion
```

damyanp: I see this repeated in other tests so I'll stop calling them out. Is it deliberate?

damyanp (commented):
A few comments / questions I'd like to get cleared up before approving.

tools/clang/lib/Headers/hlsl/dx/linalg.h
```diff
@@ -0,0 +1,178 @@
...
+
+#if ((__SHADER_TARGET_MAJOR > 6) ||                                            \
+     (__SHADER_TARGET_MAJOR == 6 && __SHADER_TARGET_MINOR >= 9)) &&            \
+    (__HLSL_VERSION >= 2021)
```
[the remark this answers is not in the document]
bob80905: The values are in `HLSLVersion.h`:

```
enum class LangStd : unsigned long {
  vUnset = 0,
  vError = 1,
  v2015 = 2015,
  v2016 = 2016,
  v2017 = 2017,
  v2018 = 2018,
  v2021 = 2021,
  v202x = 2029,
  vLatest = v2021
};
```
It is set to `2029`

tools/clang/test/CodeGenDXIL/hlsl/linalg/.clang-format
bob80905: Yep just a copy paste, unsure if it was needed or not. I'll remove clang-format.

tools/clang/test/CodeGenDXIL/hlsl/linalg/mat-vec-muladd.hlsl
bob80905: Yes, I like that idea. Something handy and tangible to compile like a Hello World program.

tools/clang/test/CodeGenDXIL/hlsl/linalg/vector-matrix-proposal.hlsl
bob80905: I do think they should stay, yes.

tools/clang/test/SemaHLSL/hlsl/linalg/outerproductaccumulate-errors.hlsl
bob80905: I'll remove all commented out builtin calls.

tools/clang/test/CodeGenDXIL/hlsl/linalg/vector-matrix-proposal.hlsl
bob80905: Though perhaps their location could be improved. I'm thinking DXC/examples/LinAlg/vector-matrix-proposal.hlsl

examples/LinAlg/vector-matrix-proposal.hlsl
```diff
@@ -0,0 +1,143 @@
+// RUN: %dxc -I %hlsl_headers -T lib_6_9 -enable-16bit-types %s
```
damyanp, line 1: Does this actually get executed during the build?

damyanp (approved):
LGTM, assuming the test runs succeed.

If the proposal example doesn't get executed often then it'll rot.  Is there anything we can do to move it somewhere where it runs?  If not, then we've got the example in the proposal itself so I don't know if there's much value keeping another copy around in the DXC repo.

examples/LinAlg/vector-matrix-proposal.hlsl
bob80905: This location won't run it as a test. I do think it's kind of awkward to keep it given there's a copy in the proposal itself, so I'll just delete it.

tools/clang/test/CodeGenDXIL/hlsl/linalg/vectoraccumulate-errors.hlsl
llvm-beanz, line 1: This is a sema test right?

tools/clang/lib/Headers/hlsl/dx/linalg.h
```diff
@@ -0,0 +1,178 @@
+// Header for linear algebra APIs.
+
+#if ((__SHADER_TARGET_MAJOR > 6) ||                                            \
```
llvm-beanz, line 3: Should we make this header error if it is included when targeting older shader models?
What about targeting SPIRV?

tools/clang/test/SemaHLSL/hlsl/linalg/make-interp-vec-errors.hlsl
```diff
@@ -0,0 +1,33 @@
...
+export float4 Test2(vector<float, 4> Input) {
+  using namespace dx::linalg;
+
+  MatrixRef<DATA_TYPE_UINT16, 4, 4, MATRIX_LAYOUT_MUL_OPTIMAL, true> Matrix = {
+      Buf, 0, 0};
+
+  // expected-error@+3{{no matching function for call to 'MakeInterpretedVector'}}
+  // expected-note@dx/linalg.h:93{{candidate template ignored: invalid explicitly-specified argument for template parameter 'DT'}}
+  return Mul<float>(    
+      Matrix, MakeInterpretedVector<DATA_TYPE_InvalidType>(Input));
```
llvm-beanz, line 31: I think we should be careful going overboard about testing diagnostic emission from header-implementations. We would not want these tests in Clang because template instantiation and overload resolution are pretty well tested indipendently.

I'm kinda okay with this in DXC because I don't think DXC's template instantiation code is well tested (and all the C++ tests are disabled and don't run).

damyanp: I added this test because it is meant to test that the API generates good diagnostics when this mistake is made.  Earlier versions of this API were written in such a way that coding errors like this generated confusing diagnostics further away from where the error actually was.

tools/clang/test/CodeGenDXIL/hlsl/linalg/vectoraccumulate-errors.hlsl
bob80905: Right

tools/clang/lib/Headers/hlsl/dx/linalg.h
llvm-beanz: The SPIRV path doesn't generate IR in DXC, it goes straight from the AST to SPIRV, which would likely cause the compiler either to crash or to emit a less-than helpful diagnostic.

llvm-beanz: Maybe we could add something like this to the top of the file?

```hlsl
#if __spirv__
#error "Cooperative matrix not (yet) supported for SPIRV"
#endif
```

damyanp: s/matrix/vector

tools/clang/lib/Headers/hlsl/dx/linalg.h
```diff
@@ -0,0 +1,182 @@
+// Header for linear algebra APIs.
+
+#if __spirv__
+#error "Cooperative matrix/vectors not (yet) supported for SPIRV"
```
llvm-beanz, line 4:
```suggestion
#error "Cooperative vectors not (yet) supported for SPIRV"
```
Sorry for my bad suggestion. Cooperative matrix actually is supported in SPIRV, they have a separate header for it.

tools/clang/test/SemaHLSL/hlsl/linalg/outerproductaccumulate-spirv-errors.hlsl
```diff
@@ -0,0 +1,19 @@
+// RUN: %dxc -I %hlsl_headers -T lib_6_9 -enable-16bit-types -spirv %s -verify
+
+// Tests that the header file cannot be included for spirv compilations
+// This is a copy of \tools\clang\test\CodeGenDXIL\hlsl\linalg\outerproductaccumulate.hlsl
+// except that spirv is targeted
+
+// expected-error@dx/linalg.h:4{{Cooperative matrix/vectors not (yet) supported for SPIRV}}
```
llvm-beanz, line 7:
```suggestion
// expected-error@dx/linalg.h:4{{Cooperative vectors not (yet) supported for SPIRV}}
```

tools/clang/test/SemaHLSL/hlsl/linalg/outerproductaccumulate-spirv-errors.i
llvm-beanz, line 1: This file doesn't seem to belong here.

llvm-beanz (approved):

PR #7357, opened 2025-04-16
Moved HANDLE management to Test Class Constructor and Destructor

alecardoso-tech:
Test class `ShaderOpTest` had a `HANDLE` not closed after `RunShaderOp` test ran. 

Created a constructor and destructor for the Test Class to manage the `HANDLE`.


alecardoso-tech:
@microsoft-github-policy-service agree company="Microsoft"

tools/clang/unittests/HLSLExec/ShaderOpTest.h
```diff
@@ -312,7 +313,7 @@ class ShaderOpTest {
   CComPtr<ID3D12Resource> m_pQueryBuffer;
   dxc::DxcDllSupport *m_pDxcSupport = nullptr;
   CommandListRefs m_CommandList;
-  HANDLE m_hFence;
+  HhEvent m_hFence;
```
damyanp, line 316: This seems like a lot of churn just to avoid adding a destructor to ShaderOpTest with a CloseHandle() in it.

alecardoso-tech: I think using RAII classes for managing objects like HANDLEs is a good practice in general. Developers often forget to close them after use. The act of moving all the RAII helper classes would be a good starting point for devs to start implementing them in their code and not have to worry about closing the managed objects.

If this change is too disruptive, I can undo most of the changes and just apply a `CloseHandle()` at the end of `ShaderOpTest::RunShaderOp(ShaderOp *pShaderOp)`.

damyanp: I think you're more likely to get a speedy PR approval with a smaller, more targeted, change.

tools/clang/unittests/HLSLExec/ShaderOpTest.cpp
```diff
@@ -1067,6 +1067,7 @@ void ShaderOpTest::RunShaderOp(ShaderOp *pShaderOp) {
   CreateCommandList();
   RunCommandList();
   CopyBackResources();
+  CloseHandle(m_hFence);
```
damyanp, line 1070: Should initialize this in the constructor (or where it's declared in the header file).

damyanp (approved):
One comment that should probably get fixed - otherwise LGTM.

Also the PR title and description needs updating.

tools/clang/unittests/HLSLExec/ShaderOpTest.cpp
```diff
@@ -258,6 +258,16 @@ void CommandListRefs::CreateForDevice(ID3D12Device *pDevice, bool compute) {
                                       IID_PPV_ARGS(&List)));
 }
 
+ShaderOpTest::ShaderOpTest()
+{
+  m_hFence = CreateEvent(nullptr, FALSE, FALSE, nullptr);
```
damyanp, line 262: Does each ShaderOpTest instance get used for just one test?  Or is it used multiple times?  If it is used multiple times, have you verified that this still works if it also reuses the same event for each one?

damyanp (approved):
Approved, assuming the answer to the comment below is that everything still works ok.

tools/clang/unittests/HLSLExec/ShaderOpTest.cpp
alecardoso-tech: I've ran all the tests after applying all the changes and everything works fine. Each instance uses just one `HANDLE` for running the `RunShaderOp` test. It can be reused for multiple test runs.

alsepkow (approved):

PR #7372, opened 2025-04-23
[SER] Validate MakeMiss

simoll:
Validate:
 HitObject_MakeMiss

Rules:
 No undef MissShaderIndex or RayFlags

SER implementation tracker: #7214

### DIFF
--- a/tools/clang/test/LitDXILValidation/ser_hitobject_make_failing.ll
+++ b/tools/clang/test/LitDXILValidation/ser_hitobject_make_failing.ll
@@ -1,0 +1,44 @@
+; REQUIRES: dxil-1-9
+; RUN: not %dxv %s 2>&1 | FileCheck %s
+
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+%dx.types.HitObject = type { i8* }
+
+; CHECK: Function: ?main@@YAXXZ: error: Instructions should not read uninitialized value.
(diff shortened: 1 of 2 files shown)


tools/clang/test/LitDXILValidation/ser_hitobject_make_failing.ll
```diff
@@ -0,0 +1,44 @@
+; REQUIRES: dxil-1-9
+; RUN: not %dxv %s 2>&1 | FileCheck %s
+
+target datalayout = "e-m:e-p:32:32-i1:32-i8:32-i16:32-i32:32-i64:64-f16:32-f32:32-f64:64-n8:16:32:64"
+target triple = "dxil-ms-dx"
+
+%dx.types.HitObject = type { i8* }
+
+; CHECK: Function: ?main@@YAXXZ: error: Instructions should not read uninitialized value.
+; CHECK: note: at '%r265_udmiss = call %dx.types.HitObject @dx.op.hitObject_MakeMiss(i32 265, i32 4, i32 undef, float 0.000000e+00, float 0.000000e+00, float 0.000000e+00, float 0.000000e+00, float 1.000000e+00, float 0.000000e+00, float 0.000000e+00, float 9.999000e+03)' in block '#0' of function '?main@@YAXXZ'.
```
tex3d, line 10: How about the rest of the float values?  Should undef be invalid there too?

tex3d (approved):

tools/clang/test/LitDXILValidation/ser_hitobject_make_failing.ll
simoll: I don't think it matters for those `float` values. It is relevant for the `MissShaderIndex` and `RayFlags`  as those may result in memory corruption by any `Invoke` call down the line, something we can stop here at least for the case of a blunt `undef`s

tex3d: I filed #7380 to track this.

damyanp (approved):

PR #7376, opened 2025-04-24
[SER] Diagnose HitObject in unsupported declaration contexts

### LINKED ISSUE #7234: [SER] Diagnose and validate illegal use of HitObject in unsupported contexts
Tracking bug to implement HLSL and DXIL diagnostics/validation for HitObject in unsupported contexts:

- [ ] Add HLSL error diagnostics for HitObject in `cbuffer` and global declarations
- [ ] Add DXIL validation rules to block HitObject in memory buffers, in groupshared memory, in ray payloads, or in intersection attributes.

simoll:
- Generalize long vector diagnostics code to HitObjects.
- Diagnose unsupported use of HitObject in globals, entry params/returns and various other shader-kind-specific contexts.
- Create HitObject variants from the invalid-longvec-decls*.hlsl tests to make sure all cases are covered.

Specification: https://github.com/microsoft/hlsl-specs/blob/main/proposals/0027-shader-execution-reordering.md

Closes #7234 [SER] Diagnose and validate illegal use of HitObject in unsupported contexts (discussed offline)


tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7878,14 +7878,14 @@ def err_hlsl_load_from_mesh_out_arrays: Error<
    "output arrays of a mesh shader can not be read from">;
 def err_hlsl_out_indices_array_incorrect_access: Error<
    "a vector in out indices array must be accessed as a whole">;
-def err_hlsl_unsupported_long_vector
-    : Error<"vectors of over 4 elements in "
+def err_hlsl_unsupported_type
+    : Error<"%select{vectors of over 4 elements|HitObjects}0 in "
```
tex3d, line 7882: The set of contexts that disallow HitObjects is not the same as those that disallow vectors of over 4 elements.  For HitObjects, we should start with the same rules (and diagnostic) as for various other built-in HLSL object types (like resources) that aren't supported in a variety of locations.  For instance, missing from the below set wrt. object types are the StructuredBuffer record type, ByteAddress buffer templated Load/Store template type, and groupshared memory, where long vectors *are* allowed.  There may be more I'm forgetting too.

So I'm not sure about extending this particular diagnostic just for HitObject, unless we were to expand this to be used for other object types generally, and expand the list of contexts to include some other locations where these objects are disallowed.

I noticed at least one more related diagnostic that could be folded in as well: `err_hlsl_node_record_object`.  I think that the approach of `"object '%0' is not allowed in %select(...)1"` where the object type is printed in `%0` might be more appropriate for the *object not allowed in context* class of error.

I also noticed `err_hlsl_objectintemplateargument` as another related object diagnostic.

tools/clang/lib/AST/DeclCXX.cpp
```diff
@@ -72,8 +72,9 @@ CXXRecordDecl::DefinitionData::DefinitionData(CXXRecordDecl *D)
       ImplicitCopyAssignmentHasConstParam(true),
       HasDeclaredCopyConstructorWithConstParam(false),
       HasDeclaredCopyAssignmentWithConstParam(false), IsLambda(false),
-      IsParsingBaseSpecifiers(false), HasHLSLLongVector(false), NumBases(0),
-      NumVBases(0), Bases(), VBases(), Definition(D), FirstFriend() {}
+      IsParsingBaseSpecifiers(false), HasHLSLLongVector(false),
+      HasHLSLHitObject(false), NumBases(0), NumVBases(0), Bases(), VBases(),
```
tex3d, line 76: I think `HasHLSLHitObject` is overly specific.  I would prefer a flag that maps to our concept of an intangible type that we are using in upstream.  This could be used for any decl that contains any special HLSL intangible object type, such as resource objects, RayQuery, HitObject, and so on.

tools/clang/lib/AST/HlslTypes.cpp
```diff
@@ -124,6 +125,8 @@ bool IsHLSLNumericUserDefinedType(clang::QualType type) {
 // which can't be annotated. But includes UDTs of trivially copyable data and
 // the builtin trivially copyable raytracing structs.
 bool IsHLSLCopyableAnnotatableRecord(clang::QualType QT) {
+  if (ContainsHitObject(QT))
+    return false;
```
tex3d, line 129: I don't know why this special case code is necessary.  `IsHLSLNumericUserDefinedType` should return false for anything containing a HitObject.  If it doesn't, that's a bug that should be fixed there.

tex3d (changes requested):
Mainly, I don't think we should be adding as much special-case code for HitObject, where the code should be the same for other object types, such as detecting whether we have an intangible HLSL object type inside another type decl.  HitObject should be considered one of a variety of intangible HLSL object types that are handled uniformly in a bunch of contexts.

But also, combining the HitObject diagnostic with the long vector one is a bit confusing, as the contexts that disallow HitObject, and other intangible objects generally, are different than the contexts that disallow long vectors.  This is conceptually a different rule.

tools/clang/lib/AST/HlslTypes.cpp
simoll: Removed that special case

tools/clang/lib/AST/DeclCXX.cpp
simoll: Removed the tracking bit, recursing into the type instead to reach all fields (to identify things like disallowed field types in subobjects of arrays of subobjects of the object that is being declared)

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7886,6 +7884,15 @@ def err_hlsl_unsupported_long_vector
     "entry function parameters|entry function return type|"
     "patch constant function parameters|patch constant function return type|"
     "payload parameters|attributes}0 are not supported">;
+// %select options must be compatible with err_hlsl_unsupported_long_vector (same index used)
+def err_hlsl_unsupported_object_context
+    : Error<"object '%0' is not allowed in "
```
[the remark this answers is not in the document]
simoll: Removed them

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -7886,6 +7884,15 @@ def err_hlsl_unsupported_long_vector
...
+    "%select{ConstantBuffers or TextureBuffers|"
+    "tessellation patches|geometry streams|node records|"
+    "cbuffers or tbuffers|user-defined struct parameter|"
+    "entry function parameters|entry function return type|"
+    "patch constant function parameters|patch constant function return type|"
+    "payload parameters|attributes|type parameters|structured buffers|global variables|groupshared variables}1">;
```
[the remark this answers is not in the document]
simoll: Changed this to `builtin template parameters` for now. Please mark as resolved if that works for you

tools/clang/lib/Sema/SemaHLSLDiagnoseTU.cpp
```diff
@@ -709,20 +709,25 @@ void hlsl::DiagnoseTranslationUnit(clang::Sema *self) {
...
       }
-      for (const auto *param : pPatchFnDecl->params())
-        if (ContainsLongVector(param->getType())) {
-          const unsigned PatchConstantFunctionParametersIdx = 8;
+      for (const auto *param : pPatchFnDecl->params()) {
+        const unsigned PatchConstantFunctionParametersIdx = 8;
+        if (ContainsLongVector(param->getType()))
           self->Diag(param->getLocation(),
                      diag::err_hlsl_unsupported_long_vector)
               << PatchConstantFunctionParametersIdx;
```
pow2clk, line 717:
```suggestion
        const unsigned PatchConstantFunctionParametersIdx = 8;
        if (ContainsLongVector(param->getType()))
          self->Diag(param->getLocation(),
                     diag::err_hlsl_unsupported_long_vector)
              << TypeDiagContext::PatchConstantFunctionParameters;
```

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject_traceinvoke_payload_udt.hlsl
```diff
@@ -16,7 +16,8 @@ struct Attribs
 [shader("raygeneration")]
 void RayGen()
 {
-  // expected-error@+1{{payload parameter 'payload_in_rg' must be a user-defined type composed of only numeric types}}
+  // expected-error@+2{{object 'dx::HitObject' is not allowed in payload parameters}}
+  // expected-note@8{{'dx::HitObject' field declared here}}
```
pow2clk, line 20: For this and those below, I have a mild preference that the notes be placed on the lines they reference instead of hard coding the line number. It would be more robust to test additions that way.

tools/clang/test/SemaHLSL/hlsl/types/invalid-hitobject-decls-struct.hlsl
```diff
@@ -0,0 +1,340 @@
...
+    inout LineStream<TYPE> OutputStream0)
+    // expected-error@-1{{object 'dx::HitObject' is not allowed in geometry streams}}
+    // expected-note@16{{'dx::HitObject' field declared here}}
+{}
+
+
+[shader("geometry")]
+[maxvertexcount(12)]
+void gs_line(
+    line TYPE a,
```
pow2clk, line 137: This is a flaw copied from my own work in incomplete-type.hlsl and invalid-longvec-decls.hlsl, but this should have been for triangles
```suggestion
void gs_tri(
    triangle TYPE a,
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12164,25 +12216,51 @@ bool hlsl::DiagnoseNodeStructArgument(Sema *self, TemplateArgumentLoc ArgLoc,
     return true;
   case AR_TOBJ_COMPOUND: {
     bool ErrorFound = false;
-    const RecordDecl *RD = ArgTy->getAs<RecordType>()->getDecl();
+    const RecordDecl *RD = Ty->getAs<RecordType>()->getDecl();
+    // Never recurse infinitely into related subtypes
```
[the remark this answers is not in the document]
simoll: I saw an infinite descend in `DirectXShaderCompiler\tools\clang\test\SemaHLSL\invalid-decl-template-arg.hlsl` without this guard

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12128,34 +12155,59 @@ void Sema::DiagnoseReachableHLSLCall(CallExpr *CE, const hlsl::ShaderModel *SM,
...
+    // TODO: This is only here because DiagnoseNodeStructArgument got folded
+    // into this function. Could fold all context-dependent long vector checks
+    // into this function.
+    if (CheckLongVec && GetHLSLVecSize(Ty) > DXIL::kDefaultMaxVectorLength) {
+      S.Diag(Loc, diag::err_hlsl_unsupported_long_vector) << DiagContextIdx;
       Empty = false;
       return false;
     }
     LLVM_FALLTHROUGH;
-  case AR_TOBJ_ARRAY:
```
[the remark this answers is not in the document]
simoll: We peel off arrays just before the switch and don't expect to see them here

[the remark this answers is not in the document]
simoll: We only need to cover recursive types in the switch and the element types we are checking for. I've added a comment for this in the default case (and this sets `isEmpty = false` now)

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12128,34 +12155,59 @@ void Sema::DiagnoseReachableHLSLCall(CallExpr *CE, const hlsl::ShaderModel *SM,
...
   case AR_TOBJ_VECTOR:
-    if (GetHLSLVecSize(ArgTy) > DXIL::kDefaultMaxVectorLength) {
-      const unsigned NodeRecordsIdx = 3;
-      self->Diag(ArgLoc.getLocation(), diag::err_hlsl_unsupported_long_vector)
-          << NodeRecordsIdx;
+    // TODO: This is only here because DiagnoseNodeStructArgument got folded
+    // into this function. Could fold all context-dependent long vector checks
+    // into this function.
+    if (CheckLongVec && GetHLSLVecSize(Ty) > DXIL::kDefaultMaxVectorLength) {
+      S.Diag(Loc, diag::err_hlsl_unsupported_long_vector) << DiagContextIdx;
```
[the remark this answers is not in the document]
simoll: Added an assertion

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15295,19 +15373,42 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
...
+    // Suppress actual emitting of errors for incompletable types here
+    // They are redundant to those produced in ActOnUninitializedDecl.
+    struct SilentDiagnoser : public TypeDiagnoser {
+      SilentDiagnoser() : TypeDiagnoser(true) {}
+      virtual void diagnose(Sema &S, SourceLocation Loc, QualType T) {}
+    } SD;
+    RequireCompleteType(D.getLocStart(), qt, SD);
+    TypeDiagContext DiagContext = TypeDiagContext::CBuffersOrTBuffers;
+    if (isStatic)
+      DiagContext = TypeDiagContext::GlobalVariables;
```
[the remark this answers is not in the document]
simoll: Folded this in with the `isGlobal` case as suggested

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject_traceinvoke_payload_udt.hlsl
simoll: I don't have a strong opinion on this. Some things i considered when i decided to group the notes with their error:
- more test code may be added below without moving the lines that the notes refer to
- the grouping makes clear what error caused the note to be emitted
- it is trivial to update line numbers even if prior lines need to be changed

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12164,25 +12216,51 @@ bool hlsl::DiagnoseNodeStructArgument(Sema *self, TemplateArgumentLoc ArgLoc,
...
+                                 DiagContext, CheckedDecls, nullptr);
     return ErrorFound;
   }
   default:
-    DXASSERT(false, "unreachable");
     return false;
   }
 }
 
+bool hlsl::DiagnoseTypeElements(Sema &S, SourceLocation Loc, QualType Ty,
```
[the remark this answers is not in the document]
simoll: Folded CheckLongVec

tools/clang/include/clang/AST/DeclCXX.h
```diff
@@ -465,10 +465,6 @@ class CXXRecordDecl : public RecordDecl {
     /// \brief Whether we are currently parsing base specifiers.
     bool IsParsingBaseSpecifiers : 1;
 
-    /// \brief Whether this class contains at least one member or base
-    ///  class containing an HLSL vector longer than 4 elements.
-    bool HasHLSLLongVector : 1;
```
pow2clk, line 470: I confess I didn't anticpate that merging the longvector check with the other type checks would result in removing this bit, but I see how it follows and I don't object.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -5659,15 +5656,9 @@ class HLSLExternalSource : public ExternalSemaSource {
...
-        const unsigned TessellationPatchesIDx = 1;
-        m_sema->Diag(argLoc.getLocation(),
-                     diag::err_hlsl_unsupported_long_vector)
-            << TessellationPatchesIDx;
-        return true;
-      }
+      const TypeDiagContext DiagContext = TypeDiagContext::TessellationPatches;
       if (DiagnoseTypeElements(*m_sema, argLoc.getLocation(), arg.getAsType(),
-                               TypeDiagContext::TessellationPatches))
+                               DiagContext, DiagContext))
```
pow2clk, line 5661: As frequently as the pattern of both objects and long vectors having the same context occurs, it may make sense to have an overload of `DiagnoseTypeElements` that takes a single context that is propagated to both for the longer overload, but I won't insist on it if there are no other commits to be made to this PR that change might be packaged with.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -12202,10 +12201,10 @@ DiagnoseElementTypes(Sema &S, SourceLocation Loc, QualType Ty, bool &Empty,
     return false;
   case AR_TOBJ_OBJECT:
     Empty = false;
-    if (AllowObjectInContext(Ty, DiagContext))
+    if (!CheckObjects || AllowObjectInContext(Ty, ObjDiagContext))
```
pow2clk, line 12204: I didn't expect and after review didn't observe any cases where the call to this would result in `CheckObjects` being false. There's nothing wrong with handling the condition in the interest of making it more robust, but I wanted to confirm I had that right.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -15395,22 +15370,22 @@ bool Sema::DiagnoseHLSLDecl(Declarator &D, DeclContext *DC, Expr *BitWidth,
...
+    if (isGroupShared)
+      ObjDiagContext = TypeDiagContext::GroupShared;
+    else if (isStatic)
+      ObjDiagContext = TypeDiagContext::GlobalVariables;
 
+    TypeDiagContext LongVecDiagContext = TypeDiagContext::Valid;
+
+    // Disallow long vecs from $Global cbuffers.
+    if (!isStatic && !isGroupShared && !IS_BASIC_OBJECT(basicKind))
+      LongVecDiagContext = TypeDiagContext::CBuffersOrTBuffers;
```
pow2clk, line 15385: I was imagining `else if (!IS_BASIC_OBJECT(basicKind))` as the final else if after the `else if (isStatic)` on line 15378 which would result in something more concise, but I won't insist. If you make another change to the PR, you might consider that consolidation.

pow2clk (approved):
Looks good! I have just a couple of nits you might consider if you end up making another iteration of this. I'm not sure why it says a merge resolution is required. When I tried to inspect the conflict, it showed zero conflicts.

tools/clang/test/SemaHLSL/hlsl/objects/HitObject/hitobject_traceinvoke_payload_udt.hlsl
simoll: no objection since.. assuming this is accepted

tex3d (approved):

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -10942,6 +10949,11 @@ HLSLExternalSource::DeduceTemplateArgumentsForHLSL(
               << intrinsicName;
           return Sema::TemplateDeductionResult::TDK_Invalid;
         }
+        if (DiagnoseTypeElements(
+                *getSema(), Loc, functionTemplateTypeArg,
+                TypeDiagContext::TypeParameter /*ObjDiagContext*/,
+                TypeDiagContext::Valid /*LongVecDiagContext*/))
+          return Sema::TemplateDeductionResult::TDK_Invalid;
```
tex3d, line 10956: We should remove the prior diagnostic if the new is going to catch invalid cases now.

tex3d: Also, `IsHLSLNumericOrAggregateOfNumericType` hits a bug in `IsHLSLCopyableAnnotatableRecord` where it doesn't traverse bases.  We should fix that too.

pow2clk (approved):
Looks good! Better diagnostics overall!

damyanp (approved):
Approving based on previous approvals before the most recent change.

Issue #5960, opened 2023-11-01
[SPIRV] Work graphs


danbrown-amd:
The SM 6.8 staging branch lacks support for work graphs for the SPIR-V target.

danbrown-amd:
(I'm already working on this task.)

s-perron:
Thanks for opening this.

Issue #3092, opened 2020-08-19
[SPIR-V] Allow thread group size to be specified with specialization constants


SaschaWillems:
In glsl it's possible to set the thread group / work group size for compute shaders via specialization constants. This allows an application to use different sizes depending e.g. on the hardware limits, without the need for different shaders.

The `GL_ARB_gl_spirv` extension added explicit qualifiers for this:

> "The built-in constant vector gl_WorkGroupSize can be specialized using
> the local_size_{xyz}_id qualifiers, to individually give the components
> an id. For example:
> 
>
>          layout(local_size_x_id = 18, local_size_z_id = 19) in;

There doesn't seem to be an equivalent of this for HLSL to SPIR-V right now.

ehsannas:
Thanks for reporting @SaschaWillems

shangjiaxuan:
Related: #4032 #2191
Seems like specialization constant logic in dxc doesn't propagate constants that can be evaluated early-on?
(earlier shader models seems to recognize #2191)

shangjiaxuan:
One suggestion would be detecting `[[xxx]]` attributes in `[numthreads(4, 4, 1)]`, and use the actual numbers as the default number (no current code will be affected). Then the following may be ok:
```
    [numthreads([[vk::constant_id(1)]] 3, [[vk::constant_id(2)]] 4, [[vk::constant_id(3)]] 1)]
```
Seems a bit hacky though...

shangjiaxuan:
Sample assembly from `glslc` compilation of following code:
```
layout(local_size_x_id = 1, local_size_y_id = 2, local_size_z = 3) in;
void main(void)
{
}
```
Assembly:
```
; SPIR-V
; Version: 1.5
; Generator: Google Shaderc over Glslang; 10
; Bound: 12
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionMode %main LocalSize 1 1 3
               OpSource GLSL 450
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %main "main"
               OpDecorate %7 SpecId 1
               OpDecorate %8 SpecId 2
               OpDecorate %gl_WorkGroupSize BuiltIn WorkgroupSize
       %void = OpTypeVoid
[...]

PR #7378, opened 2025-04-25
[SPIRV] Refactor OpExecutionModeId

s-perron:
The current implementation of OpExecutionModeId assumes that the
operands must be OpConstants. However, they could be the id of a
spec constant. The first step in allowing OpExecutionModeId is to modify
the internal representation of the instruction to hold general spirv
instructions.

Part of https://github.com/microsoft/DirectXShaderCompiler/issues/5960
and https://github.com/microsoft/DirectXShaderCompiler/issues/3092.

### DIFF
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -997,9 +997,6 @@ SpirvInstruction *SpirvBuilder::addExecutionModeId(
   SpirvExecutionModeId *mode = nullptr;
   SpirvExecutionModeBase *existingInstruction =
       mod->findExecutionMode(entryPoint, em);
-  assert(!existingInstruction || existingInstruction->getKind() ==
-                                     SpirvInstruction::IK_ExecutionModeId);
-
   if (!existingInstruction) {
     mode = new (context) SpirvExecutionModeId(loc, entryPoint, em, params);
     mod->addExecutionMode(mode);
@@ -1009,7 +1006,7 @@ SpirvInstruction *SpirvBuilder::addExecutionModeId(
     // `addExecutionMode` should have been called with `em` or the existing
     // instruction is wrong.
     assert(existingInstruction->getKind() ==
-           SpirvInstruction::IK_ExecutionMode);
+           SpirvInstruction::IK_ExecutionModeId);
     mode = cast<SpirvExecutionModeId>(existingInstruction);
   }
 
--- a/tools/clang/include/clang/SPIRV/SpirvBuilder.h
+++ b/tools/clang/include/clang/SPIRV/SpirvBuilder.h
@@ -615,8 +615,15 @@ public:
   inline SpirvInstruction *addExecutionMode(SpirvFunction *entryPoint,
                                             spv::ExecutionMode em,
                                             llvm::ArrayRef<uint32_t> params,
-                                            SourceLocation,
-                                            bool useIdParams = false);
+                                            SourceLocation);
+
+  /// \brief Adds an execution mode to the module under construction if it does
+  /// not already exist. Return the newly added instruction or the existing
+  /// instruction, if one already exists.
+  inline SpirvInstruction *
+  addExecutionModeId(SpirvFunction *entryPoint, spv::ExecutionMode em,
+                     llvm::ArrayRef<SpirvInstruction *> params,
+                     SourceLocation loc);
 
   /// \brief Adds an OpModuleProcessed instruction to the module under
   /// construction.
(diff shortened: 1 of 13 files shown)


s-perron:
FYI: @danbrown-amd

tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -963,17 +970,47 @@ SpirvBuilder::setDebugSource(uint32_t major, uint32_t minor,
...
+}
+
+SpirvInstruction *SpirvBuilder::addExecutionModeId(
+    SpirvFunction *entryPoint, spv::ExecutionMode em,
+    llvm::ArrayRef<SpirvInstruction *> params, SourceLocation loc) {
+  SpirvExecutionModeId *mode = nullptr;
+  SpirvExecutionModeBase *existingInstruction =
       mod->findExecutionMode(entryPoint, em);
+  assert(!existingInstruction || existingInstruction->getKind() ==
+                                     SpirvInstruction::IK_ExecutionModeId);
```
cassiebeckley, line 1001: Do you need both this assert and the one on line 1011?

tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -963,17 +970,47 @@ SpirvBuilder::setDebugSource(uint32_t major, uint32_t minor,
...
+    mode = new (context) SpirvExecutionModeId(loc, entryPoint, em, params);
     mod->addExecutionMode(mode);
   } else {
-    mode = existingInstruction;
+    // No execution mode can be used with both OpExecutionMode and
+    // OpExecutionModeId. If this assert is triggered, then either this
+    // `addExecutionMode` should have been called with `em` or the existing
+    // instruction is wrong.
+    assert(existingInstruction->getKind() ==
+           SpirvInstruction::IK_ExecutionMode);
```
Keenuts, line 1012: Shouldn't the opcode be ExecutionModeID?

tools/clang/include/clang/SPIRV/SpirvBuilder.h
```diff
@@ -963,17 +970,47 @@ SpirvBuilder::setDebugSource(uint32_t major, uint32_t minor,
 SpirvInstruction *
 SpirvBuilder::addExecutionMode(SpirvFunction *entryPoint, spv::ExecutionMode em,
                                llvm::ArrayRef<uint32_t> params,
-                               SourceLocation loc, bool useIdParams) {
+                               SourceLocation loc) {
   SpirvExecutionMode *mode = nullptr;
-  SpirvExecutionMode *existingInstruction =
+  SpirvExecutionModeBase *existingInstruction =
+      mod->findExecutionMode(entryPoint, em);
```
Keenuts, line 976: Find doesn't check the params are the same. Should this check all the instruction operands are the same before returning the existing one?

tools/clang/lib/SPIRV/EmitVisitor.cpp
```diff
@@ -617,19 +617,21 @@ bool EmitVisitor::visit(SpirvEntryPoint *inst) {
...
-          /*isSpecConst */ false));
+  if (auto *exeModeId = dyn_cast<SpirvExecutionModeId>(inst)) {
+    for (SpirvInstruction *param : exeModeId->getParams()) {
+      if (auto *ConstantInst = dyn_cast<SpirvConstant>(param))
+        typeHandler.getOrCreateConstant(ConstantInst);
+      curInst.push_back(getOrAssignResultId<SpirvInstruction>(param));
     }
+  } else {
+    auto *exeMode = dyn_cast<SpirvExecutionMode>(inst);
+    assert(exeMode);
```
Keenuts, line 632:
```suggestion
    auto *exeMode = cast<SpirvExecutionMode>(inst);
```

Keenuts (commented):
Some bits, but overall LGTM

tools/clang/lib/SPIRV/EmitVisitor.cpp
s-perron: I had done that at first and it did not work. I retried, and found the issue. There is a `cast` namespace, so it resolves to that namespace. I had to do `llvm::cast`. I have it working now.

tools/clang/include/clang/SPIRV/SpirvBuilder.h
s-perron: I'm keeping the existing behaviour. You are probably right. I'll open up a follow up issue for that. I see that as outside the scope of this PR, which is about refactoring existing behaviour.

Keenuts (approved):

PR #7379, opened 2025-04-25
[SER] Basic execution tests

simoll:
- All trivial scalar/vector/matrix getters
- HitObject::FromRayQuery with procedural hit
- HitObject::GetAttributes<T> with custom attributes and procedural hit

SER implementation tracker: #7214

### DIFF
--- a/tools/clang/unittests/HLSLExec/DXRUtil.h
+++ b/tools/clang/unittests/HLSLExec/DXRUtil.h
@@ -1,0 +1,226 @@
+//===------------ DXRUtil.h - DXR Utility Functions ------------*- C++ -*-===//
+///////////////////////////////////////////////////////////////////////////////
+//                                                                           //
+// DXRUtil.h                                                                 //
+// Copyright (C) Nvidia Corporation. All rights reserved.                    //
+// This file is distributed under the University of Illinois Open Source     //
+// License. See LICENSE.TXT for details.                                     //
+//                                                                           //
+//  This file contains the utility functions for DXR execution tests.        //
+//                                                                           //
+///////////////////////////////////////////////////////////////////////////////
+
+//= DXR Utility
+//============================================================================
+#define SHADER_ID_SIZE_IN_BYTES 32
+
+#ifndef ROUND_UP
+#define ROUND_UP(v, powerOf2Alignment)                                         \
+  (((v) + (powerOf2Alignment)-1) & ~((powerOf2Alignment)-1))
+#endif
+struct SceneConsts {
+  DirectX::XMFLOAT4 eye;
+  DirectX::XMFLOAT4 U;
+  DirectX::XMFLOAT4 V;
+  DirectX::XMFLOAT4 W;
+  float sceneScale;
+  unsigned windowSize[2];
+  int rayFlags;
+};
+
+struct Instance {
+  D3D12_RAYTRACING_GEOMETRY_TYPE type;
+  DirectX::XMFLOAT4X4 matrix;
+  UINT geometryCount;
+  UINT bottomASIdx;
+  UINT instanceID;
+  UINT mask;
+  UINT flags;
+};
+
+class ShaderTable {
+public:
+  void Init(ID3D12Device *device, int raygenCount, int missCount,
+            int hitGroupCount, int rayTypeCount, int rootTableDwords) {
+    m_rayTypeCount = rayTypeCount;
+    m_raygenCount = raygenCount;
+    m_missCount = missCount * rayTypeCount;
+    m_hitGroupCount = hitGroupCount * rayTypeCount;
+    m_rootTableSizeInBytes = rootTableDwords * 4;
+    m_shaderRecordSizeInBytes =
+        ROUND_UP(m_rootTableSizeInBytes + SHADER_ID_SIZE_IN_BYTES,
+                 D3D12_RAYTRACING_SHADER_RECORD_BYTE_ALIGNMENT);
+    m_missStartIdx = m_raygenCount;
+    m_hitGroupStartIdx = m_missStartIdx + m_missCount;
+
+    const int m_totalSizeInBytes =
+        (m_raygenCount + m_missCount + m_hitGroupCount) *
+        m_shaderRecordSizeInBytes;
+
+    D3D12_RESOURCE_DESC desc = CD3DX12_RESOURCE_DESC::Buffer(
+        m_totalSizeInBytes, D3D12_RESOURCE_FLAG_NONE,
+        std::max(D3D12_RAYTRACING_SHADER_RECORD_BYTE_ALIGNMENT,
+                 D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT));
+    CD3DX12_HEAP_PROPERTIES heap =
+        CD3DX12_HEAP_PROPERTIES(D3D12_HEAP_TYPE_DEFAULT);
+    VERIFY_SUCCEEDED(device->CreateCommittedResource(
+        &heap, D3D12_HEAP_FLAG_NONE, &desc,
+        D3D12_RESOURCE_STATE_NON_PIXEL_SHADER_RESOURCE, nullptr,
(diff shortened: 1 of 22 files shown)


tools/clang/unittests/HLSLExec/DXRUtil.h
```diff
@@ -0,0 +1,226 @@
...
+#endif
+struct SceneConsts {
+  DirectX::XMFLOAT4 eye;
+  DirectX::XMFLOAT4 U;
+  DirectX::XMFLOAT4 V;
+  DirectX::XMFLOAT4 W;
+  float sceneScale;
+  unsigned windowSize[2];
+  int rayFlags;
+};
```
damyanp, line 29: This needs a pass through to update to follow LLVM naming conventions.

```suggestion
struct SceneConsts {
  DirectX::XMFLOAT4 Eye;
  DirectX::XMFLOAT4 U;
  DirectX::XMFLOAT4 V;
  DirectX::XMFLOAT4 W;
  float SceneScale;
  unsigned WindowSize[2];
  int RayFlags;
};
```

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -292,6 +294,15 @@ class ExecutionTest {
   TEST_METHOD(SaturateTest);
   TEST_METHOD(SignTest);
   TEST_METHOD(Int64Test);
+  TEST_METHOD(SERBasicTest);
```
damyanp, line 297: Wondering how it was decided to insert these test methods here, rather than after the other test methods?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -2078,6 +2095,707 @@ void ExecutionTest::RunRWByteBufferComputeTest(ID3D12Device *pDevice,
   WaitForSignal(pCommandQueue, FO);
 }
 
+CComPtr<ID3D12Resource> ExecutionTest::RunDXRTest(
```
damyanp, line 2098: Any reason this all has to be in this file?

damyanp (changes requested):
There's a lot of code here to review and I've not had a chance to go through it all in detail.

However, I'll note that before this can go into main we'll need to get it updated to follow the LLVM coding standards.

For the short term maybe this should go into the staging branch?

clandrew:
Thanks! Taking a look at this now. It might be that we can simply ingest the test code based on this staging branch and it does not necessarily have to be merged.

simoll:
> There's a lot of code here to review and I've not had a chance to go through it all in detail.
> 
> However, I'll note that before this can go into main we'll need to get it updated to follow the LLVM coding standards.
> 
> For the short term maybe this should go into the staging branch?

Should close this PR and re-open against `staging-sm6.9`?

simoll:
There are merge conflicts with the sm69 staging branch just because this was originally based on `main`. Will reopen as  a new pR.

tools/clang/unittests/HLSLExec/DXRUtil.h
simoll: Updated to follow coding standards

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
simoll: Moved all SER test methods to the end of the list

tools/clang/unittests/HLSLExec/ExecutionTest_SER.h
```diff
@@ -0,0 +1,2554 @@
...
+{
+    payload.visited |= 1U;
+}
+
+// Triangles
+[shader("anyhit")]
+void anyhit(inout PerRayData payload, in Attrs attrs)
+{
+    payload.visited |= 2U;
+    AcceptHitAndEndSearch();
```
jenatali, line 981: There is a test bug here:

The test attempts to very that threads fall into 3 buckets of work:

* Threads that hit nothing (payload ends up as 0)
* Threads that hit the AABB, which doesn't commit the hit, and that's all (payload ends up as 8)
* Threads that hit the AABB and the triangle (payload ends up as 2 | 8 | 16 = 26)

For that last bucket, there's no requirement on the ordering between the hits of the AABB and the triangle. If the triangle is hit first, then this function call causes the AABB to not be hit, and therefore the 8 is missing from the final payload (value 18). WARP hits this case 100% of the time.

However, even without this, if the triangle is hit first and is in front of the procedural primitive (which it apparently is, half of the time, according to what I'm seeing with WARP), then accepting the triangle will cause the procedural primitive or its any-hit to be skipped.

tools/clang/unittests/HLSLExec/DXRUtil.h
```diff
@@ -0,0 +1,223 @@
...
+  UINT InstanceID;
+  UINT Mask;
+  UINT Flags;
+};
+
+class ShaderTable {
+public:
+  void Init(ID3D12Device *Device, int RaygenCount, int MissCount,
+            int HitGroupCount, int RayTypeCount, int RootTableDwords) {
+    RayTypeCount = RayTypeCount;
```
jenatali, line 47: These initializers are no-ops because the left hand side and right hand side refer to the same argument variable. If this was an initializer list then it would work, or if the left hand side was accessed with `this->`.

tools/clang/unittests/HLSLExec/ExecutionTest_SER.h
```diff
@@ -0,0 +1,2554 @@
...
+        hitKind = 1;
+    hitKind *= 2;
+    if (relHitPos.x >= 0.0f)
+        hitKind += 1;
+    hitKind *= 2;
+    if (relHitPos.z >= 0.0f)
+        hitKind += 1;
+
+    CustomAttrs attrs;
+    attrs.dist = length(relHitPos);
```
jenatali, line 1569: It seems something in here is too dependent on floating point precision. If I allow FMA fusing in WARP (which literally just was hooked up a few weeks ago) then the test passes, but if I disable it then I get a bucketing which is off-by-one from the test expectations.

tools/clang/unittests/HLSLExec/ExecutionTest_SER.h
```diff
@@ -0,0 +1,2554 @@
...
+void miss(inout PerRayData payload)
+{
+    // UNUSED
+}
+
+// Triangles
+[shader("anyhit")]
+void anyhit(inout PerRayData payload, in Attrs attrs)
+{
+    AcceptHitAndEndSearch();
```
jenatali, line 2345: This test appears to suffer from the same issue I described for `SERShaderTableIndexTest`, where the order of triangle vs procedural hits in the acceleration structure determines the overall bucketing.

tools/clang/unittests/HLSLExec/ExecutionTest_SER.h
simoll: I changed the test to no longer rely on anyhit/intersections. Instead, we hit or miss the triangle (aabbs ignored by ray flag), then mutate to an aabb hit. Only testing for correct invocation of the miss/closesthit/chAABB shaders

tools/clang/unittests/HLSLExec/DXRUtil.h
simoll: Made that a proper ctor with initializer list

tools/clang/unittests/HLSLExec/ExecutionTest_SER.h
simoll: Made the hitkind more robust by replacing the fp comparison and with launchIndex-based comparisons to generate a meaningful hitKind. Please test on your end, there may still be fp precision issues with the aabb intersection logic itself.

tools/clang/unittests/HLSLExec/ExecutionTest_SER.h
simoll: Hit order issue should be resolved. Keeping this open until confirmed.

Issue #7389, opened 2025-04-28
[Validator] Use the internal validator by default


bob80905:
**Is your feature request related to a problem? Please describe.**
DXC will often be unable to validate a shader targeted towards a preview shader model because there is no accompanying dxv.exe that is correctly versioned. 

**Describe the solution you'd like**
DXC should use the internal validator by default, as described in https://github.com/microsoft/hlsl-specs/blob/main/proposals/infra/INF-0004-validator-hashing.md.

**Additional context**
This will prevent other repos that ingest DXC from failing tests when there are no valid DXIL.dlls that are suitable for validation for the preview shader model.

PR #7392, opened 2025-04-28
Change default validation behavior

bob80905:
This PR changes the default validation behavior to use the internal validator by default.
If no options are specified, the internal validator will be used, and if it fails, then compilation fails.
The external validator can still be run but must be explicitly chosen.
Specifying internal works just as before.

There is plenty of testing and infrastructure that needs to be added to verify this change, but that needs to be added in a separate change. This change is step 1.

Addresses https://github.com/microsoft/DirectXShaderCompiler/issues/7389

### DIFF
--- a/include/dxc/Support/HLSLOptions.h
+++ b/include/dxc/Support/HLSLOptions.h
@@ -115,7 +115,7 @@ struct RewriterOpts {
 };
 
 enum class ValidatorSelection : int {
-  Auto,        // Try internal validator; fallback to DXIL.dll
+  Auto,        // Force internal validator (even if DXIL.dll is present)
   Internal,    // Force internal validator (even if DXIL.dll is present)
   External,    // Use DXIL.dll, failing compilation if not available
   Invalid = -1 // Invalid
--- a/include/dxc/Support/HLSLOptions.h
+++ b/include/dxc/Support/HLSLOptions.h
@@ -115,7 +115,7 @@ struct RewriterOpts {
 };
 
 enum class ValidatorSelection : int {
-  Auto,        // Try DXIL.dll; fallback to internal validator
+  Auto,        // Try internal validator; fallback to DXIL.dll
   Internal,    // Force internal validator (even if DXIL.dll is present)
   External,    // Use DXIL.dll, failing compilation if not available
   Invalid = -1 // Invalid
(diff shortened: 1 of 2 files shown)


tools/clang/tools/dxcompiler/dxcutil.cpp
```diff
@@ -52,19 +52,22 @@ namespace {
...
 
-  bool bInternalValidator = false;
-  if (pValidator == nullptr) {
-    IFTBOOL(!bExternal, DXC_E_VALIDATOR_MISSING);
+  if (!bExternal) {
     IFT(CreateDxcValidator(IID_PPV_ARGS(&pValidator)));
     bInternalValidator = true;
   }
+
+  if (pValidator == nullptr && DxilLibIsEnabled()) {
```
damyanp, line 66: hmm...it looks to me that this function could now exit, without throwing, with pValidator set to nullptr.  Previously this wasn't possible.

I think both the original, and this version, are pretty hard to follow.  Some comments, or maybe even refactoring, could help here.

damyanp: The new version is a lot clearer!

include/dxc/Support/HLSLOptions.h
```diff
@@ -115,7 +115,7 @@ struct RewriterOpts {
 };
 
 enum class ValidatorSelection : int {
-  Auto,        // Try DXIL.dll; fallback to internal validator
+  Auto,        // Try internal validator; fallback to DXIL.dll
```
damyanp, line 118: Does it really fall back?  Or is the fall back logic outside of the code modified here?

llvm-beanz:
Is there any reasonable way to test this?

include/dxc/Support/HLSLOptions.h
bob80905: Good point, I looked outside the code and don't think there is any fallback logic. 
Auto -> Internal

tex3d (changes requested):
Please hold merging for further review.

I think there are other changes required and design considerations to address.  I will follow up with more soon.

tex3d (approved):
Ok, approved, with planned future changes and testing in this area.

damyanp (approved):
Can you file an issue for the followup and tag it in this PR please?

(That can happen after it's merged if you want)

bob80905:
> Can you file an issue for the followup and tag it in this PR please?
> 
> (That can happen after it's merged if you want)

https://github.com/microsoft/DirectXShaderCompiler/issues/7397

PR #7393, opened 2025-04-28
Update LongVector Execution tests to now use XML 

alsepkow:
This change is mainly focused on swapping to use the XML and statically defined values. Opting to stage this for review in smaller chunk while I iterate on adding the additional test cases.

1. Update the LongVector exec tests to use the ShaderOpArithTable.xml TEAF table file for test entry points. This aligns with existing HLK tests.
2. Some light code cleanup.
3. Hard coded value sets in LongVectorTestData.h. Value sets give us a simple way to generate larger arrays from a smaller set of statically defined values. At a later point we can add logic to produce value sets at build time in this same header by consuming from a YAML file used in the offload test suite.

### DIFF
--- a/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
+++ b/tools/clang/unittests/HLSLExec/ExecutionTest.cpp
@@ -279,6 +279,7 @@ typedef struct D3D12_FEATURE_DATA_D3D12_OPTIONS4 {
 
 // Virtual class to compute the expected result given a set of inputs
 struct TableParameter;
+class TableParameterHandler; // Forward declaration
 
 class ExecutionTest {
 public:
@@ -507,115 +508,15 @@ public:
                        L"Table:ShaderOpArithTable.xml#PackUnpackOpTable")
   END_TEST_METHOD()
 
-  // bool binary ops
-  TEST_METHOD(LongVector_ScalarAdd_bool)
-  TEST_METHOD(LongVector_ScalarMultiply_bool)
-  TEST_METHOD(LongVector_Multiply_bool)
-  TEST_METHOD(LongVector_Add_bool)
-  TEST_METHOD(LongVector_Min_bool)
-  TEST_METHOD(LongVector_Max_bool)
-  // bool unary ops
-  // Note that clamp doesn't make sense for bools.
-  TEST_METHOD(LongVector_Initialize_bool);
+  BEGIN_TEST_METHOD(LongVector_BinaryOpTest)
+  TEST_METHOD_PROPERTY(L"DataSource",
+                       L"Table:ShaderOpArithTable.xml#LongVectorBinaryOpTable")
+  END_TEST_METHOD()
 
-  // float16 (half) binary ops
-  TEST_METHOD(LongVector_ScalarAdd_float16)
-  TEST_METHOD(LongVector_ScalarMultiply_float16)
-  TEST_METHOD(LongVector_Multiply_float16)
-  TEST_METHOD(LongVector_Add_float16)
-  TEST_METHOD(LongVector_Min_float16)
-  TEST_METHOD(LongVector_Max_float16)
-  // float16 (half) unary ops
-  TEST_METHOD(LongVector_Clamp_float16);
-  TEST_METHOD(LongVector_Initialize_float16);
-
-  // float32 binary ops
-  TEST_METHOD(LongVector_ScalarAdd_float32)
-  TEST_METHOD(LongVector_ScalarMultiply_float32)
-  TEST_METHOD(LongVector_Multiply_float32)
-  TEST_METHOD(LongVector_Add_float32)
-  TEST_METHOD(LongVector_Min_float32)
-  TEST_METHOD(LongVector_Max_float32)
-  // float32 unary ops
-  TEST_METHOD(LongVector_Clamp_float32);
-  TEST_METHOD(LongVector_Initialize_float32);
-
-  // float64 binary ops
-  TEST_METHOD(LongVector_ScalarAdd_float64)
-  TEST_METHOD(LongVector_ScalarMultiply_float64)
-  TEST_METHOD(LongVector_Multiply_float64)
-  TEST_METHOD(LongVector_Add_float64)
-  TEST_METHOD(LongVector_Min_float64)
-  TEST_METHOD(LongVector_Max_float64)
-  // float64 unary ops
-  TEST_METHOD(LongVector_Clamp_float64);
-  TEST_METHOD(LongVector_Initialize_float64);
-
-  // int16 binary ops
-  TEST_METHOD(LongVector_ScalarAdd_int16)
-  TEST_METHOD(LongVector_ScalarMultiply_int16)
-  TEST_METHOD(LongVector_Multiply_int16)
-  TEST_METHOD(LongVector_Add_int16)
-  TEST_METHOD(LongVector_Min_int16)
-  TEST_METHOD(LongVector_Max_int16)
-  // int16 unary ops
-  TEST_METHOD(LongVector_Clamp_int16);
-  TEST_METHOD(LongVector_Initialize_int16);
-
-  // int32 binary ops
-  TEST_METHOD(LongVector_ScalarAdd_int32)
-  TEST_METHOD(LongVector_ScalarMultiply_int32)
-  TEST_METHOD(LongVector_Multiply_int32)
-  TEST_METHOD(LongVector_Add_int32)
-  TEST_METHOD(LongVector_Min_int32)
-  TEST_METHOD(LongVector_Max_int32)
-  // int32 unary ops
-  TEST_METHOD(LongVector_Clamp_int32);
-  TEST_METHOD(LongVector_Initialize_int32);
-
-  // int64 binary ops
-  TEST_METHOD(LongVector_ScalarAdd_int64)
-  TEST_METHOD(LongVector_ScalarMultiply_int64)
-  TEST_METHOD(LongVector_Multiply_int64)
-  TEST_METHOD(LongVector_Add_int64)
-  TEST_METHOD(LongVector_Min_int64)
-  TEST_METHOD(LongVector_Max_int64)
-  // int64 unary ops
-  TEST_METHOD(LongVector_Clamp_int64);
-  TEST_METHOD(LongVector_Initialize_int64);
-
-  // uint16 binary ops
-  TEST_METHOD(LongVector_ScalarAdd_uint16)
-  TEST_METHOD(LongVector_ScalarMultiply_uint16)
-  TEST_METHOD(LongVector_Multiply_uint16)
-  TEST_METHOD(LongVector_Add_uint16)
-  TEST_METHOD(LongVector_Min_uint16)
-  TEST_METHOD(LongVector_Max_uint16)
-  // uint16 unary ops
-  TEST_METHOD(LongVector_Clamp_uint16);
-  TEST_METHOD(LongVector_Initialize_uint16);
-
-  // uint32 binary ops
-  TEST_METHOD(LongVector_ScalarAdd_uint32)
-  TEST_METHOD(LongVector_ScalarMultiply_uint32)
-  TEST_METHOD(LongVector_Multiply_uint32)
-  TEST_METHOD(LongVector_Add_uint32)
-  TEST_METHOD(LongVector_Min_uint32)
-  TEST_METHOD(LongVector_Max_uint32)
-  // uint32 unary ops
-  TEST_METHOD(LongVector_Clamp_uint32);
-  TEST_METHOD(LongVector_Initialize_uint32);
-
-  // uint64 binary ops
-  TEST_METHOD(LongVector_ScalarAdd_uint64)
-  TEST_METHOD(LongVector_ScalarMultiply_uint64)
-  TEST_METHOD(LongVector_Multiply_uint64)
-  TEST_METHOD(LongVector_Add_uint64)
-  TEST_METHOD(LongVector_Min_uint64)
-  TEST_METHOD(LongVector_Max_uint64)
-  // uint64 unary ops
-  TEST_METHOD(LongVector_Clamp_uint64);
-  TEST_METHOD(LongVector_Initialize_uint64);
+  BEGIN_TEST_METHOD(LongVector_UnaryOpTest)
+  TEST_METHOD_PROPERTY(L"DataSource",
+                       L"Table:ShaderOpArithTable.xml#LongVectorUnaryOpTable")
+  END_TEST_METHOD()
 
   dxc::DxcDllSupport m_support;
 
@@ -826,9 +727,14 @@ public:
                                const char *pShaderModelStr, const char *pShader,
                                Ty *pInputDataPairs, unsigned inputDataCount);
 
+  template <typename T>
+  void LongVectorOpTestDispatchByDataType(T OpType, std::wstring DataType, TableParameterHandler &Handler);
+
+  template <typename T, typename U> 
+  void LongVectorOpTestDispatchByVectorSize(U OpType, TableParameterHandler &Handler);
+
   template <typename T, std::size_t N>
   void LongVectorOpTestBase(LongVectorOpTestConfig<T> &TestConfig);
-  template <typename T> void LongVectorOpTestBase(LongVectorOpType OpType);
 
   template <class Ty> const wchar_t *BasicShaderModelTest_GetFormatString();
 
@@ -6580,6 +6486,20 @@ static TableParameter PackUnpackOpParameters[] = {
     {L"Validation.Input", TableParameter::UINT32_TABLE, true},
 };
 
+static TableParameter LongVectorBinaryOpParameters[] = {
+  {L"DataType", TableParameter::STRING, true},
+  {L"OpTypeEnum", TableParameter::STRING, true},
+  {L"InputValueSetName1", TableParameter::STRING, false},
+  {L"InputValueSetName2", TableParameter::STRING, false},
+};
+
+static TableParameter LongVectorUnaryOpParameters[] = {
+  {L"DataType", TableParameter::STRING, true},
+  {L"OpTypeEnum", TableParameter::STRING, true},
+  {L"InputValueSetName1", TableParameter::STRING, false},
+  {L"InputArgsName", TableParameter::STRING, false},
+};
+
 static bool IsHexString(PCWSTR str, uint16_t *value) {
   std::wstring wString(str);
   wString.erase(std::remove(wString.begin(), wString.end(), L' '),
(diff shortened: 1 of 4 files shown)


damyanp:
I'm guessing that the submodule changes aren't meant to be a part of this change?

alsepkow:
> I'm guessing that the submodule changes aren't meant to be a part of this change?

They are not. Accidentally grabbed those when I added the clang-format. Will fix.

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -199,62 +206,189 @@ void FillLongVectorDataFromShaderBuffer(MappedData &ShaderBuffer,
...
+              "a new enum value?");
+
+LongVectorUnaryOpType
+GetLongVectorUnaryOpType(const std::wstring &OpTypeString) {
+  return GetLongVectorOpType<LongVectorUnaryOpType>(
+      LongVectorUnaryOpTypeStringToEnumMap, OpTypeString,
+      std::size(LongVectorUnaryOpTypeStringToEnumMap));
+}
+
+template <typename T> std::vector<T> GetInputValueSetByKey(std::wstring Key) {
```
damyanp, line 293:
```suggestion
template <typename T> std::vector<T> GetInputValueSetByKey(const std::wstring &Key) {
```

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -199,62 +206,189 @@ void FillLongVectorDataFromShaderBuffer(MappedData &ShaderBuffer,
...
+
+  if constexpr (std::is_same_v<T, HLSLBool_t>)
+    return std::vector<T>(DefaultInputValueSet_bool.at(Key).begin(),
+                          DefaultInputValueSet_bool.at(Key).end());
```
damyanp, line 297: This seems convoluted to me.  If there's a good reason for it to be this way then a comment would be good.

The value type of these sets is already a vector of the right type?

```suggestion
  if constexpr (std::is_same_v<T, HLSLBool_t>)
    return DefaultInputValueSet_bool.at(Key);
```

I think you'd write it this way if the element types didn't match and you wanted to reconstruct each element...but why not just have them be the right type in the map to start with so they don't need to be converted again here?

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -293,108 +456,210 @@ template <typename T> struct LongVectorOpTestConfig {
...
+  std::string OperatorString_;
+  // To be used for the value of -DFUNC
+  std::string IntrinsicString_;
+  // Optional, can be used to override shader code.
+  float Tolerance_ = 0.0;
+  LongVectorBinaryOpType BinaryOpType_ = LongVectorBinaryOpType_EnumValueCount;
+  LongVectorUnaryOpType UnaryOpType_ = LongVectorUnaryOpType_EnumValueCount;
+  std::wstring InputValueSetName1_ = L"DefaultInputValueSet1";
+  std::wstring InputValueSetName2_ = L"DefaultInputValueSet2";
+  std::wstring InputArgsArrayName_ = L""; // No default args array
```
damyanp, line 628: I'm not sure that underscore suffixes is compatible with the llvm code standards.

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -199,62 +206,189 @@ void FillLongVectorDataFromShaderBuffer(MappedData &ShaderBuffer,
...
+
+template <typename T> std::vector<T> GetInputValueSetByKey(std::wstring Key) {
+
+  if constexpr (std::is_same_v<T, HLSLBool_t>)
+    return std::vector<T>(DefaultInputValueSet_bool.at(Key).begin(),
+                          DefaultInputValueSet_bool.at(Key).end());
+  else if constexpr (std::is_same_v<T, HLSLHalf_t>)
+    return std::vector<T>(DefaultInputValueSet_float16.at(Key).begin(),
+                          DefaultInputValueSet_float16.at(Key).end());
+  else if constexpr (std::is_same_v<T, float>)
```
damyanp, line 301: This pattern of using `if constexpr (std::is_same_v<T, float>)` I think might be leading you into some directions that result in more complicated / verbose code than necessary.

Here's an approach that uses more traditional template specialization: https://godbolt.org/z/15qjobn9j

tools/clang/unittests/HLSLExec/LongVectors.h
alsepkow: I had asked co-pilot and it linked me to LLVM documentation saying to add that. But manually navigating to the coding standards documentation and asking co-pilot again doesn't bring it up. Not sure what it showed me...

tools/clang/unittests/HLSLExec/LongVectors.h
alsepkow: You're right. I was hitting an issue with the implicit copy. But that's only an issue for the HLSL* types. I'll clean that up by making the map the right type to begin with.

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -412,3 +510,89 @@ bool DoArraysMatch(const std::array<T, N> &ActualValues,
...
+
+  for (size_t i = 0; i < N; ++i) {
+    ExpectedValues[i] = Config.ComputeExpectedValue(InputVector1[i]);
+  }
+
+  return ExpectedValues;
+}
+
+template <typename T, std::size_t N>
+void LogLongVector(const std::array<T, N> &Values, std::wstring Name) {
```
alsepkow, line 571: Name should be a const reference. Here and below. Will fix on next commit.

tools/clang/unittests/HLSLExec/LongVectorTestData.h
```diff
@@ -0,0 +1,241 @@
+#pragma once
```
alsepkow, line 1: Should I be using ifdef include guards instead of #pragma once?

damyanp: I guess so :(  https://llvm.org/docs/CodingStandards.html#header-guard

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11716,200 +11253,153 @@ void ExecutionTest::LongVectorOpTestBase(
...
+
+  if (LogInputs) {
+    LogLongVector<T, N>(InputVector1, L"InputVector1");
+
+    if (IsVectorBinaryOp)
+      LogLongVector<T, N>(InputVector2, L"InputVector2");
+    else if (TestConfig.IsScalarOp())
+      LogLongVector<T, 1>(ScalarInput, L"ScalarInput");
+
+    if (TestConfig.GetUnaryOpType() == LongVectorUnaryOpType_Clamp) {
```
alsepkow, line 11322: This should be more generic. I'll update it when there are more test cases with input args.

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -199,62 +72,156 @@ void FillLongVectorDataFromShaderBuffer(MappedData &ShaderBuffer,
...
-  LongVectorOpType_ScalarAdd,
-  LongVectorOpType_ScalarMultiply,
-  LongVectorOpType_Multiply,
-  LongVectorOpType_Add,
-  LongVectorOpType_Min,
-  LongVectorOpType_Max,
-  LongVectorOpType_Clamp,
-  LongVectorOpType_Initialize,
-  LongVectorOpType_UnInitialized
+enum LongVectorBinaryOpType {
```
alsepkow, line 75: Just want to note here that for subsequent PR's that add more tests cases I'm going to split these enums up by category instead of unary vs binary.
HLSLOperators, trigonometric, math, etc. And then I'll also have some child classes that inherit from LongVectorTestConfig for these various categories to help split out the logic by category and avoid large switch statements.

include/dxc/Test/HlslTestUtils.h
```diff
@@ -553,6 +563,27 @@ inline bool CompareDoubleULP(
   return AbsoluteDiff <= (uint64_t)ULPTolerance;
 }
 
+inline bool CompareDoubleEpsilon(
+    const double &Src, const double &Ref, float Epsilon,
+    hlsl::DXIL::Float32DenormMode Mode = hlsl::DXIL::Float32DenormMode::Any) {
+  if (Src== Ref) {
+    return true;
+  }
+  if (std::isnan(Src)) {
```
bob80905, line 574: What happens if Src is not nan and Ref is nan?

include/dxc/Test/HlslTestUtils.h
```diff
@@ -604,8 +635,11 @@ inline bool CompareFloatRelativeEpsilon(
 
 inline bool CompareHalfULP(const uint16_t &fsrc, const uint16_t &fref,
                            float ULPTolerance) {
-  if (fsrc == fref)
+  // Compare as floats to handle when we have '0 == -0' which should be true
+  // but won't be true if we compare as uint16_t.
+  if (ConvertFloat16ToFloat32(fsrc) == ConvertFloat16ToFloat32(fref))
```
bob80905, line 641: Just curious, is there a ConvertUInt16ToFloat32 variant? Should that be created, or is the implementation of this conversion function identical to what would be done to convert uint16s?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -6580,6 +6494,20 @@ static TableParameter PackUnpackOpParameters[] = {
...
+    {L"OpTypeEnum", TableParameter::STRING, true},
+    {L"InputValueSetName1", TableParameter::STRING, false},
+    {L"InputValueSetName2", TableParameter::STRING, false},
+};
+
+static TableParameter LongVectorUnaryOpParameters[] = {
+    {L"DataType", TableParameter::STRING, true},
+    {L"OpTypeEnum", TableParameter::STRING, true},
+    {L"InputValueSetName1", TableParameter::STRING, false},
+    {L"InputArgsName", TableParameter::STRING, false},
```
bob80905, line 6508: Surprised BinaryOpParameters doesn't need something like InputArgsName.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11216,486 +11144,116 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
+    LongVectorOpTestDispatchByVectorSize<uint64_t>(OpType, Handler);
+  else if (DataType == L"float16")
+    LongVectorOpTestDispatchByVectorSize<HLSLHalf_t>(OpType, Handler);
+  else if (DataType == L"float32")
+    LongVectorOpTestDispatchByVectorSize<float>(OpType, Handler);
+  else if (DataType == L"float64")
+    LongVectorOpTestDispatchByVectorSize<double>(OpType, Handler);
+  else
+    VERIFY_FAIL(
+        String().Format(L"DataType: %s is not recognized.", DataType.c_str()));
```
bob80905, line 11221: would this need to use %ls for wide strings or something of the sort?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11216,486 +11144,116 @@ TEST_F(ExecutionTest, PackUnpackTest) {
...
-template <typename T>
-void ExecutionTest::LongVectorOpTestBase(LongVectorOpType opType) {
-  WEX::TestExecution::SetVerifyOutput verifySettings(
-      WEX::TestExecution::VerifyOutputSettings::LogOnlyFailures);
+  // InputValueSetName2 is optional. So the string may be empty. An empty
+  // string will result in the default value set for this DataType being used.
+  if (TestConfig.IsBinaryOp()) {
+    std::wstring InputValueSet2(
+        Handler.GetTableParamByName(L"InputValueSetName2")->m_str);
+    if (!InputValueSet1.empty()) {
```
bob80905, line 11245: InputValueSet2 right?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11708,208 +11266,169 @@ void ExecutionTest::LongVectorOpTestBase(LongVectorOpType opType) {
...
+        // Process the callback for the InputVector2 resource.
+        if (0 == _stricmp(Name, "InputVector2")) {
+          if (IsVectorBinaryOp) {
+            FillShaderBufferFromLongVectorData<T, N>(ShaderData, InputVector2);
+          }
+          return;
+        }
+
+        LOG_ERROR_FMT_THROW(
+            L"RunShaderOpTest CallBack. Unexpected Resource Name: %S", Name);
```
bob80905, line 11425: Is %ls needed?

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -293,108 +494,309 @@ template <typename T> struct LongVectorOpTestConfig {
...
+  T Diff = A > B ? A - B : B - A;
+  return Diff > Tolerance;
+}
 
-template <typename T> class DeterministicNumberGenerator {
-  // Mersenne Twister 'random' number generator. Generated numbers are based
-  // on the seed value and are deterministic for any given seed.
-  std::mt19937 Generator;
+inline bool DoValuesMatch(HLSLBool_t A, HLSLBool_t B, [[maybe_unused]] float Tolerance, [[maybe_unused]] LongVector::ValidationType ValidationType) { 
+  return A == B;
```
bob80905, line 737: Why doesn't this use the Tolerance and ValidationType args? Is it possible to just remove those 2 args?

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -412,3 +815,91 @@ bool DoArraysMatch(const std::array<T, N> &ActualValues,
...
+    ExpectedValues[i] = Config.ComputeExpectedValue(InputVector1[i]);
+  }
+
+  return ExpectedValues;
+}
+
+template <typename T, std::size_t N>
+void LogLongVector(const std::array<T, N> &Values, const std::wstring &Name) {
+  WEX::Logging::Log::Comment(
+      WEX::Common::String().Format(L"LongVector Name: %s", Name.c_str()));
```
bob80905, line 876: %ls?

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -412,3 +815,91 @@ bool DoArraysMatch(const std::array<T, N> &ActualValues,
...
+      Wss << L", ";
+  }
+  Wss << L" ]";
+
+  WEX::Logging::Log::Comment(Wss.str().c_str());
+}
+
+template <typename T> void LogScalar(const T &Value, const std::wstring &Name) {
+  WEX::Logging::Log::Comment(
+      WEX::Common::String().Format(L"Scalar Name: %s", Name.c_str()));
```
bob80905, line 897: %ls?

include/dxc/Test/HlslTestUtils.h
alsepkow: Good question :). I coped the pattern from other functions here but I'm not sure what would happen. I'll debug that case to see and report back.

include/dxc/Test/HlslTestUtils.h
alsepkow: That helper doesn't exist in this code.  There is no float16 type in c++ (well, there actually is in c++23, but we don't use that yet). So float16 was implemented in a DX header in the Windows SDK. 

So, this function name is a little confusing IMHO. 

A little more context......
The two Convert* functions defined in HlslTestUtils reference their respective conversion functions in that DX SDK header. 'XMConvertHalfToFloat' and 'XMConvertFloatToHalf'.  DirectX::PackedVector::HALF is aliased as a uint16_t in that header. And that uint16_t is used to store the bits of a 16-bit float. The DX header provides the functions to convert those bits into a floating-point value via the conversion functions, but because C++ had no float16 we use a float but the values are effectively clamped to the range of a 16 bit float.

The take-away is that we store and pass around float16 values as uint16_t's, but when it comes time to do comparisons we use the helper functions to interpret that uint16_t bit pattern as a float value.

Here's a link to the mentioned header if you're curious: [DirectXPackedVector](https://github.com/microsoft/DirectXMath/blob/main/Inc/DirectXPackedVector.h )

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: It will eventually. Non of the binary intrinsics added up to this point have required input arguments. Although, there might be a better name to use here. The 'InputArgNames' are used for additional arguments like the min/max values in clamp. Distinct from the unary, binary, or ternary operands.

tools/clang/unittests/HLSLExec/LongVectors.h
alsepkow: This is an artifact of the templated functions. Theres a compiler warning enabled as a failure for unused function parameters. If I modify the calling location to only include these arguments based on the type it doesn't really solve much. It's called by DoArraysMatch a few lines below.

If you have a cleaner suggestion I'm open!

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
alsepkow: Name is an explicit LPCSTR (long pointer to a char). And we're formatting that into a wide character string. So, in this case %S is what we want. 

This gets confusing because there are functions in Windows that can compile down to either a narrow (char*) or wide (wchar*) depending on the environment (ANSI vs Unicode). The format specifier characters can have different meanings based on which variant they end up in. This is helpful for writing portable code where the type of the string (wide or narrow) and the type of the string expected by the function could change based on the environment. When the types could change you want to use %S and %s 

This is why windows API functions have things like 'CreateFile', 'CreateFileA', and 'CreateFileW'. 

%ls will interpret the argument as a wide character string in both printf and wprintf.
%s is a char* in printf, but a wchar_t* in wprintf
%S is a wchar_t* in printf, but a char* in wprintf 

Strings are 'fun' :).

tools/clang/unittests/HLSLExec/LongVectors.h
damyanp: Not necessarily better, but this is another way I think to silence the warning:

```c++
DoValuesMatch(HLSLBool_t A, HLSLBool_t B, float, LongVector::ValidationType) {
```

include/dxc/Test/HlslTestUtils.h
tex3d: In that case the comparison should fail at the end `(Src - NaN) = NaN`, `fabs(NaN) = NaN`, `NaN < Eps = false`.

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -199,68 +73,375 @@ void FillLongVectorDataFromShaderBuffer(MappedData &ShaderBuffer,
...
+  BinaryOpType_Modulus,
+  // Below this line to be moved to HLSLMathOpType (min, max etc)
+  BinaryOpType_Min,
+  BinaryOpType_Max,
+  BinaryOpType_ScalarMin,
+  BinaryOpType_ScalarMax,
+  BinaryOpType_EnumValueCount
+};
+
+static const LongVectorOpTypeStringToEnumValue BinaryOpTypeStringToEnumMap[] = {
```
tex3d, line 135: Couldn't these be a `std::map`?

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -412,3 +813,93 @@ bool DoArraysMatch(const std::array<T, N> &ActualValues,
...
+  WEX::Logging::Log::Comment(
+      WEX::Common::String().Format(L"Scalar Name: %s", Name.c_str()));
+
+  std::wstringstream Wss(L"");
+  Wss << L"Scalar Value: ";
+  Wss << Value;
+  WEX::Logging::Log::Comment(Wss.str().c_str());
+}
+
+#endif // LONGVECTORS_H
```
bob80905, line 905: newline at eof

bob80905 (approved):
Just some nits, take em or leave em

include/dxc/Test/HlslTestUtils.h
```diff
@@ -604,8 +636,11 @@ inline bool CompareFloatRelativeEpsilon(
...
```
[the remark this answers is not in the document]
alsepkow: You're right :). Not needed here, will revert this.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -826,9 +733,16 @@ class ExecutionTest {
...
-  template <typename T> void LongVectorOpTestBase(LongVectorOpType OpType);
+  template <typename T>
+  void LongVectorOpTestDispatchByDataType(T OpType, std::wstring DataType,
+                                          TableParameterHandler &Handler);
+
+  template <typename T, typename U>
+  void LongVectorOpTestDispatchByVectorSize(U OpType,
+                                            TableParameterHandler &Handler);
+
+  template <typename T, std::size_t N, typename U>
```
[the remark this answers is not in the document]
alsepkow: Yeah, I can update the names to be more descriptive.

It is the vector length. I could probably set the size in the TestConfig and update to using vectors so the length can be determined when needed. I don't see a good reason to keep these as std::arrays anymore. I forget why I picked a std::array over a vector when I started writing these tests, but whatever it was I don't think it was the right assumption. Or it's not needed anymore.

Do you think it would be cleaner/more readable to remote the std::size_t and set a size in the config. Other code can check the vector size if it needs it or use iterators.

tools/clang/unittests/HLSLExec/LongVectors.h
alsepkow: I wanted to have a static assert on the size to have a compile error if you forget to add to this enum value when adding an *OpType. Even declaring the std::map as static const still wouldn't allow a static assert on the size.

Open to ideas if you think some other pattern is cleaner though.

tools/clang/unittests/HLSLExec/CMakeLists.txt
```diff
@@ -3,6 +3,8 @@
 find_package(TAEF REQUIRED)
 find_package(D3D12 REQUIRED) # Used for ExecutionTest.cpp.
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
```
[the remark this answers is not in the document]
alsepkow: I'll take a look into doing that. Originally this code was more coupled with other helper code in ExecutionTest.cpp. 

A few thoughts come to mind though:
  1. How much time to we want to spend trying to clean this up right now vs getting long vector exec tests in.
  2. We might do a refactoring/re-writing of a bunch of this in the not-too-distant future.
  3. "We'll do this later" often just turns into never 😔.

But I'll still take a closer look and see what I could easily break up. I think it will make reviews easier.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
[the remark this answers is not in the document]
alsepkow: Removed the need for the size_t

tools/clang/unittests/HLSLExec/CMakeLists.txt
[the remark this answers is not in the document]
alsepkow: I'd need to factor out some of the ShaderOpTest* code starting on line 3531 with the ShaderOpTestResultStruct. Seems reasonable to put that into ShaderOpTest.h/cpp. And the TableParameter/TableParameterHandler stuff would need to be factored out. That could probably also go in its own appropriately named header/cpp file.

Both seem like reasonable updates I could make to help clean this file up a little. Think its worth doing that?

include/dxc/Test/HlslTestUtils.h
```diff
@@ -604,12 +631,23 @@ inline bool CompareFloatRelativeEpsilon(
 
 inline bool CompareHalfULP(const uint16_t &fsrc, const uint16_t &fref,
                            float ULPTolerance) {
+  // Treat +0 and -0 as equal
+  if ((fsrc & ~FLOAT16_BIT_SIGN) == 0 && (fref & ~FLOAT16_BIT_SIGN) == 0)
+    return true;
   if (fsrc == fref)
     return true;
   if (isnanFloat16(fsrc))
     return isnanFloat16(fref);
```
tex3d, line 640: I actually think the following would be more robust/correct for this case:
```suggestion
  bool nanRef = isnanFloat16(fref);
  bool nanSrc = isnanFloat16(fsrc);
  if (nanSrc || nanRef)
      return (nanSrc && nanRef);
```

Basically, we should have a reference value (fref) of NaN if we expect NaN, and if so, only NaN is valid for fsrc, otherwise, fsrc should not be NaN.

Without this change, it's possible we have fref=NaN, but fsrc is zero or a denorm close to zero, and just because of the particular NaN's bit pattern and the monotonic ordering, we end up with an adjacent value for the ULP comparison, which would be an unintended pass.

tools/clang/unittests/HLSLExec/LongVectors.tpp
```diff
@@ -0,0 +1,688 @@
...
+}
+
+template <typename DataTypeT>
+bool LongVector::DoValuesMatch(DataTypeT A, DataTypeT B, float Tolerance,
+                   LongVector::ValidationType) {
+  if (Tolerance == 0.0f)
+    return A == B;
+
+  DataTypeT Diff = A > B ? A - B : B - A;
+  return Diff > Tolerance;
```
alsepkow: This is wrong, should be Diff <= Tolerance.

utils/version/latest-release.json
tex3d, line 1: Shouldn't modify this file.

tools/clang/unittests/HLSLExec/ShaderOpArithTable.xml
```diff
@@ -7524,4 +7524,4 @@
             </Parameter>
         </Row>
     </Table>
-</Data>
+</Data>
```
tex3d: I think this change was probably unintentional.

tools/clang/unittests/HLSLExec/LongVectors.tpp
tex3d, line 1: I don't think I've seen the `.tpp` extension used before.  Why not use `.cpp` here?

tex3d: I realize now there already is a `LongVectors.cpp`, and this file is just included at the end of `LongVectors.h`.  Could these not just be implemented in the `LongVectors.cpp` file?

tools/clang/unittests/HLSLExec/LongVectors.h
```diff
@@ -12,403 +12,351 @@
...
+  BinaryOpType_ScalarMultiply,
+  BinaryOpType_ScalarSubtract,
+  BinaryOpType_ScalarDivide,
+  BinaryOpType_ScalarModulus,
+  BinaryOpType_Multiply,
+  BinaryOpType_Add,
+  BinaryOpType_Subtract,
+  BinaryOpType_Divide,
+  BinaryOpType_Modulus,
+  // Below this line to be moved to HLSLMathOpType (min, max etc)
```
tex3d: I'm not sure what to make of these comments.  Is this a TODO?

tex3d (approved):
There are some things I think we should do when taking these changes to main:
- Break up the changes into smaller parts: refactors of existing code could be reviewed and merged ahead of the code that depends on it, and make subsequent reviews much easier.
- Make sure these changes will work with the HLK content porting script (understanding changes necessary there).
- If there's a way to simplify the design to make it easier to understand, add, and maintain tests, take that into consideration.

tools/clang/unittests/HLSLExec/LongVectors.h
alsepkow: Yeah, essentially. I was making some notes while working should have probably just removed this.

In a subsequent PR when adding some of the math specific intrinsics I was considering adding a MathOpType_*.  And then they'd have a LongVector::BasicOpType::BasicOpType_Binary property set.

alsepkow: Removing the comment and the one above it.

tools/clang/unittests/HLSLExec/LongVectors.tpp
alsepkow: This is an artifact from when the LongVector test cases were still in ExecutionTests.cpp. If they were declared in a header but defined in LongVector.cpp they wouldn't be available in the translation unit when linking. You get linker errors. I actually expected this to work but from some googling and co-piloting I learned it doesn't work.

Now that the test cases themselves are in LongVector.cpp, the definitions could also be moved to LongVectors.cpp — and I agree that makes more sense.

I'll move that over in a subsequent PR tomorrow.

Issue #7395, opened 2025-04-29
DXC crashes on simple HLSL input


jimblandy:
When invoked as `dxc -HV 2018 -T cs_6_7 -E main naga-crashes-dxc/reduced.hlsl`, the following input makes DXC crash:

```hlsl
void main()
{
    int i = asuint(int(1));
}
```

I'm running DXC 34b6d0f91 on Linux, built as follows:

```
cmake ..                                        \
  -C ../cmake/caches/PredefinedParams.cmake     \
  -DCMAKE_BUILD_TYPE=Debug                \
  -DHLSL_INCLUDE_TESTS=OFF                      \
  -DCLANG_INCLUDE_TESTS=OFF                     \
  -DLLVM_INCLUDE_TESTS=OFF                      \
  -DHLSL_BUILD_DXILCONV=OFF                     \
  -DSPIRV_WERROR=OFF                            \
  -DENABLE_SPIRV_CODEGEN=OFF                    \
  -DLLVM_ENABLE_ASSERTIONS=ON                   \
  -DLLVM_ASSERTIONS_NO_STRINGS=ON               \
  -DLLVM_ASSERTIONS_TRAP=ON                     \
  -DDXC_CODEGEN_EXCEPTIONS_TRAP=ON              \
  -DDXC_DISABLE_ALLOCATOR_OVERRIDES=ON          \
  -G Ninja
```

The call stack is as follows:

<details>
<summary>Call stack at crash</summary>
<pre>
#0  (anonymous namespace)::llvm_assert_trap (_Message=0x7ffff7051ce0 "", _File=0x7ffff7051ce0 "", _Line=0, _Function=0x7ffff7051ce0 "")
    at /home/jimb/DirectXShaderCompiler/lib/Support/assert.cpp:20
#1  0x00007ffff518fb81 in llvm_assert (Message=0x7ffff7051ce0 "", File=0x7ffff7051ce0 "", Line=0, Function=0x7ffff7051ce0 "") at /home/jimb/DirectXShaderCompiler/lib/Support/assert.cpp:44
#2  0x00007ffff5f3426d in (anonymous namespace)::IntExprEvaluator::Success (this=0x7fffffff5190, SI=..., E=0x8ed080, Result=...)
    at /home/jimb/DirectXShaderCompiler/tools/clang/lib/AST/ExprConstant.cpp:6201
[...]

PR #7396, opened 2025-04-29
Fix `asuint` operator when applied to `int` values.

jimblandy:
Bother to mark the APSInt result as unsigned, to avoid tripping IntExprEvaluator::Success's assertion that SI.isSigned() matches E's type.


jimblandy:
@amaiorano Hi! @Kangz suggested that I 'at' you on this PR. I'm working on Firefox's WebGPU implementation, and #7395 is a blocker for us transitioning from the SDK's DXC to our own builds. If you have time, could you take a look?

I was also surprised that Chrome hadn't run into this problem itself, and was a bit worried that I had missed some fixes that hadn't been upstreamed. Corentin says that, as far as he knows, all of Google's DXC fixes have been upstreamed here; is that your understanding as well?

amaiorano:
> @amaiorano Hi! @Kangz suggested that I 'at' you on this PR. I'm working on Firefox's WebGPU implementation, and #7395 is a blocker for us transitioning from the SDK's DXC to our own builds. If you have time, could you take a look?
> 
> I was also surprised that Chrome hadn't run into this problem itself, and was a bit worried that I had missed some fixes that hadn't been upstreamed. Corentin says that, as far as he knows, all of Google's DXC fixes have been upstreamed here; is that your understanding as well?

All our fixes are indeed upstreamed. Just to be clear, this isn't actually a "crash", but an assertion, right? It seems to compile fine in a non-assert build: https://godbolt.org/z/a8q8zxMza

Having said that, we do enable asserts in our release build, as this has helped tremendously to reduce the number of security issues (better to trap than to crash in some random location). If you're interested in doing this for Firefox, btw, you'll probably want to take advantage of some of the changes I landed to reduce the binary size of dxcompiler (see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/third_party/gn/dxc/BUILD.gn;l=596-608?q=dxc%2Fbuild.gn)).

Anyway, back to this bug, I tried to write WGSL that would produce the same output as in your example, but I'm not able to. The closest I came was with this WGSL:
```
@compute @workgroup_size(1)
fn main() {
  // var i = u32(i32(1));
  // var i = u32(1i);
  let a = 1i;
  var i = u32(a);
}
```
With tint, we produce the following HLSL:
```
[numthreads(1, 1, 1)]
void main() {
  int a = int(1);
  uint i = uint(a);
}
```

Not quite the same, and this seems to compile. There may be a way to craft WGSL to make Tint produce something similar, but I haven't been able to in the past few minutes. I guess this is why we haven't run into this one yet!

jimblandy:
> Just to be clear, this isn't actually a "crash", but an assertion, right?

Yes - sorry. It is an assertion failure. Following the Dawn team's advice, Mozilla is indeed doing our own build of DXC with assertions enabled, for the security benefits.

> If you're interested in doing this for Firefox, btw, you'll probably want to take advantage of some of the changes I landed to reduce the binary size of dxcompiler (see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/third_party/gn/dxc/BUILD.gn;l=596-608?q=dxc%2Fbuild.gn)).

I think we've managed to [get all that](https://searchfox.org/mozilla-central/rev/d57e641fd8748a0551f5fe150df27d789cb4931d/taskcluster/scripts/misc/build-dxc.sh#43-45) in our build script - thank you!

> I tried to write WGSL that would produce the same output as in your example, but I'm not able to.

It may be that Naga simply leans on `asuint` whereas Tint prefers `uint` casts. I am not sure why we're using `asuint`.

Are you able to review the PR itself?

amaiorano:
I'll on my phone, so I'll review tomorrow; however, a couple on things right off the bat:

- Since the change you made is in Clang/LLVM code, it will need the HLSL change begin/end comment tags (I forget the exact text, but it should be easy enough to find). Having said that, usually if a change is made to Clang/LLVM code, you should check what was done in upstream LLVM. If they have a similar patch, then it can be taken verbatim without the need for tags. This is preferable.

- Add at least a test that would have asserted before, but passes now. I'll point out some good examples for you tomorrow (or take a look at some my of recent patches).

jimblandy:
@microsoft-github-policy-service agree company="Mozilla"

amaiorano (approved):

jimblandy:
@amaiorano Okay, I've added a test. Please take a look.

amaiorano:
Hi Jim, you can hit the "Squash and merge" when you're ready.

jimblandy:
@llvm-beanz Would you be able to take a look at this? Firefox's WebGPU is blocked by #7395, and we'd prefer to get the fixes approved before we include them in Firefox's DXC build.

jimblandy:
@llvm-beanz Ping!

tools/clang/test/DXC/asuint-constant-eval.hlsl
llvm-beanz, line 1: We should move this test to `tools/clang/test/CodeGenDXIL/hlsl/intrinsics/`

The `DXC` tests are really intended just as basic smoke tests of the command line interface.

jimblandy:
@llvm-beanz Sorry for the delay. The revised PR should address your review comments. PTAL.

jimblandy:
@llvm-beanz Review ping!

PR #7398, opened 2025-04-29
Update version to 1.8.2505

bob80905:
Update version to 1.8.2505


damyanp (approved):

tex3d (approved):
Presumably, we will follow up with a change to allow globally selecting an external validator if desired before we release.

PR #7408, opened 2025-04-30
Add DXIL REQUIRES to certain tests that require a minimum validation versoin

bob80905:
Internal testing using older validators reveals some failures on tests that were intended to be run on newer validators. This PR changes the tests to require a minimum validation version to run.

### DIFF
--- a/tools/clang/test/LitDXILValidation/load-store-validation.ll
+++ b/tools/clang/test/LitDXILValidation/load-store-validation.ll
@@ -1,3 +1,4 @@
+; REQUIRES: dxil-1-9
 ; RUN: not %dxv %s 2>&1 | FileCheck %s
 
 ; Ensure proper validation errors are produced for invalid parameters to load and store operations.
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -3712,7 +3712,7 @@ TEST_F(CompilerTest, CompileWhenODumpThenOptimizerMatch) {
     LPCWSTR Target = L"ps_6_0";
     CreateBlobFromText("float4 main() : SV_Target { return 0; }", &pSource);
 
-    LPCWSTR Args[3] = {OptLevel, L"-select-validater external", L"/Odump"};
+    LPCWSTR Args[3] = {OptLevel, L"-select-validator external", L"/Odump"};
 
     // Get the passes for this optimization level.
     VERIFY_SUCCEEDED(pCompiler->Compile(pSource, L"source.hlsl", L"main",
--- a/tools/clang/unittests/HLSL/CompilerTest.cpp
+++ b/tools/clang/unittests/HLSL/CompilerTest.cpp
@@ -3712,7 +3712,7 @@ TEST_F(CompilerTest, CompileWhenODumpThenOptimizerMatch) {
     LPCWSTR Target = L"ps_6_0";
     CreateBlobFromText("float4 main() : SV_Target { return 0; }", &pSource);
 
-    LPCWSTR Args[3] = {OptLevel, L"-select-validator external", L"/Odump"};
+    LPCWSTR Args[2] = {OptLevel, L"/Odump"};
 
     // Get the passes for this optimization level.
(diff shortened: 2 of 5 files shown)


damyanp (approved):
Please file a follow up task for us to look at these tests again.

We discussed offline - the tests that needed this change weren't expecting to run against the validator that was just built, these are tests that require an older validator to pass.

tools/clang/unittests/HLSL/CompilerTest.cpp
```diff
@@ -3712,7 +3712,7 @@ TEST_F(CompilerTest, CompileWhenODumpThenOptimizerMatch) {
     LPCWSTR Target = L"ps_6_0";
     CreateBlobFromText("float4 main() : SV_Target { return 0; }", &pSource);
 
-    LPCWSTR Args[2] = {OptLevel, L"/Odump"};
+    LPCWSTR Args[3] = {OptLevel, L"-select-validator external", L"/Odump"};
```
tex3d, line 3715: `/Odump` should not run validation.  I don't understand why this option would be necessary.

tools/clang/test/LitDXILValidation/load-store-validation.ll
```diff
@@ -1,3 +1,4 @@
+; REQUIRES: dxil-1-9
```
tex3d, line 1: I think we should start a convention where we create sub-folders under `LitDXILValidation` for specific validator version requirements.  All new tests would go into the latest folder, and we would have a lit.local.cfg that implements this requirement, so that none of the tests are required to individually state this requirement.

tools/clang/unittests/HLSL/DxilModuleTest.cpp
```diff
@@ -514,51 +514,67 @@ TEST_F(DxilModuleTest, SetValidatorVersion) {
     return;
 
   // Current version
-  VerifyValidatorVersionMatches(L"ps_6_2", {});
-  VerifyValidatorVersionMatches(L"lib_6_3", {});
+  VerifyValidatorVersionMatches(L"ps_6_2", {L"-select-validator external"});
+  VerifyValidatorVersionMatches(L"lib_6_3", {L"-select-validator external"});
```
tex3d, line 518: I don't understand why we need any of this for any of these tests.  If these tests fail without this added command line option, I think we have a bug.  We should fix that bug, not add this option to the test.

tex3d (changes requested):
I don't think any of the added `-select-validator external` should have been necessary.  I think this is an indication of a bug somewhere else.

On the change to require dxil-1-9 on a couple new tests, this is going to keep happening.  I recommended that we place these tests under a subdirectory which blocks for specific validator version.  That should make it easier to do the right thing in the first place.

tools/clang/unittests/HLSL/CompilerTest.cpp
bob80905: I can confirm that this change does fix this test failure. Previously there was a failure when running `RunOptimizer`.

tex3d:
I suspect the real problem is that this code was not changed when we defaulted to using the internal validator:
https://github.com/microsoft/DirectXShaderCompiler/blob/6d67e4a0af4710d18515ccaf0153922c172d415d/tools/clang/tools/dxcompiler/dxcapi.cpp#L90-L94

This is code used by `dxcompiler.dll` to pass through any `IDxcValidator` creation to an external validator (`DXIL.dll`) if one is present.  We don't want to do that anymore, since we have a perfectly good signing validator in `dxcompiler.dll` now.

This can cause test failures because compilation is performed with the internal validator, which sets the validator version metadata to the latest version, then subsequent validation is performed on a validator created through the `DxcCreateInstance` API, which goes through this code path and selects the external validator that's present during the test run.  That validator cannot validate the IR that has a newer validator version, so it fails.

tools/clang/test/LitDXILValidation/load-store-validation.ll
bob80905: https://github.com/microsoft/DirectXShaderCompiler/issues/7413

PR #7409, opened 2025-05-01
Don't pass through to DXIL.dll in DxcCreateInstance for CLSID_DxcValidator

tex3d:
This change removes the code meant to pass through to DXIL.dll when creating a CLSID_DxcValidator object with DxcCreateInstance.

Since the internal validator is now the default, and it will sign shaders, there is no reason to pass through to DXIL.dll when using DxcCreateInstance on DxCompiler.dll.

Testing will come with planned work to include the ability to globally override the default validator to external.


damyanp (approved):

bob80905 (approved):

bob80905:
Thanks Tex!!

PR #7412, opened 2025-05-01
For OuterProductAccumulate, matrix layout must be outerproductoptimal and matrix stride must be zero 

anupamachandra:
Implements the DXIL portion of https://github.com/microsoft/hlsl-specs/pull/494, The HLSL checks will be a part of the HLSL validation checks, did not add it to this PR due to shared infrastructure.


tools/clang/test/CodeGenDXIL/hlsl/intrinsics/linalg_builtins/outer-product-accumulate-matrix-layout.hlsl
```diff
@@ -0,0 +1,68 @@
...
+  F8_E4M3 = 21,
+  F8_E5M2 = 22,
+};
+
+enum MatLayout {
+  RowMajor = 0,
+  ColumnMajor = 1,
+  MulOptimal = 2,
+  OuterProductOptimal = 3,
+};
```
damyanp: I'm not super keen on having the full enums duplicated here.  I may have missed a discussion about it though...

Since only one value is could we instead do something like:

```c++
  const uint matrix_interpretation = 8; // ComponentType::F16;
```

That way grepping through the source code will find the right things.

anupamachandra:
Opened a new PR targetting master branch: https://github.com/microsoft/DirectXShaderCompiler/pull/7417

PR #7415, opened 2025-05-02
[SPIRV] Fix constant value in function

GinShio:
We will get a crash when use spirv intrinsic to create a constant value.
```fundamental
fatal error: generated SPIR-V is invalid: Constant cannot appear in a function declaration
  %spirvIntrinsicType_42 = OpConstant %spirvIntrinsicType 42
```


s-perron (approved):

GinShio:
@microsoft-github-policy-service agree company="AMD"

PR #7417, opened 2025-05-02
[0029] [Main] For OuterProductAccumulate, matrix layout must be outerproductoptimal and matrix stride must be zero 

anupamachandra:
Retargets https://github.com/microsoft/DirectXShaderCompiler/pull/7412 to main branch


tools/clang/test/CodeGenHLSL/linalg/outer-product-accumulate-matrix-layout.hlsl
```diff
@@ -0,0 +1,65 @@
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=RowMajor -DSTRIDE=64 2>&1| FileCheck %s --check-prefixes DXIL-0
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=ColumnMajor -DSTRIDE=64 2>&1 | FileCheck %s --check-prefixes DXIL-1
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=MulOptimal -DSTRIDE=64 2>&1 | FileCheck %s --check-prefixes DXIL-2
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=OuterProductOptimal -DSTRIDE=64 2>&1 | FileCheck %s --check-prefixes DXIL-3
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=OuterProductOptimal -DSTRIDE=0 2>&1 | FileCheck %s --check-prefixes DXIL-4
+ByteAddressBuffer input_vector_buffer;
+ByteAddressBuffer input_vector_buffer2;
+RWByteAddressBuffer matrix_buffer;
+
+enum CompType {
```
anupamachandra, line 10: Addressing comment https://github.com/microsoft/DirectXShaderCompiler/pull/7412#discussion_r2070712329

I followed the pattern in `outer-product-accumulate-multioverload.hlsl` for consistency. I am fine changing to your suggestion if you prefer it.

docs/DXIL.rst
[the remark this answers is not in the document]
damyanp: Yeah, I've seen this happen before. It's unfortunate!

damyanp (approved):

tools/clang/test/CodeGenHLSL/linalg/outer-product-accumulate-matrix-layout.hlsl
tex3d: These tests predated the high-level API header.  I think we can switch to defined values from the high-level API now.

tools/clang/test/CodeGenHLSL/linalg/outer-product-accumulate-matrix-layout.hlsl
```diff
@@ -0,0 +1,65 @@
...
+  ColumnMajor = 1,
+  MulOptimal = 2,
+  OuterProductOptimal = 3,
+};
+
+// DXIL-0: error: matrix layout value 'RowMajor' is not valid for outerproductaccumulate, must be 'OuterProductOptimal'
+// DXIL-1: error: matrix layout value 'ColumnMajor' is not valid for outerproductaccumulate, must be 'OuterProductOptimal' 
+// DXIL-2: error: matrix layout value 'MulOptimal' is not valid for outerproductaccumulate, must be 'OuterProductOptimal' 
+// DXIL-3-NOT: error: matrix layout value 'OuterProductOptimal' is not valid for outerproductaccumulate, must be 'OuterProductOptimal' 
+// DXIL-3: error: matrix stride must be zero for optimal layouts
```
tex3d, line 49: Since these cases (`DXIL-0` through `DXIL-3`) are DxilValidation tests, they need to be in IR form under LitDXILValidation instead.  That's because, ideally, we should be checking some of these things in Sema as well, which will prevent testing DxilValidation starting with invalid HLSL.  You should be able to base the validation test on IR produced in the valid case (`DXIL-4`).

Since you can't vary tests with defines for IR, you'll need to have multiple calls for checking multiple validation errors from the same IR test.

tex3d (changes requested):
I'm requesting changes due to the DxilValidation test being in the wrong form in the wrong place.

tools/clang/test/CodeGenHLSL/linalg/outer-product-accumulate-matrix-layout.hlsl
anupamachandra: Before I remove this, wanted to check if we want to stash the source of the IR in some place.

tools/clang/test/CodeGenHLSL/linalg/outer-product-accumulate-matrix-layout.hlsl
```diff
@@ -0,0 +1,65 @@
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=RowMajor -DSTRIDE=64 2>&1| FileCheck %s --check-prefixes DXIL-0
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=ColumnMajor -DSTRIDE=64 2>&1 | FileCheck %s --check-prefixes DXIL-1
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=MulOptimal -DSTRIDE=64 2>&1 | FileCheck %s --check-prefixes DXIL-2
+// RUN: %dxc -T cs_6_9 %s -enable-16bit-types -DML=OuterProductOptimal -DSTRIDE=64 2>&1 | FileCheck %s --check-prefixes DXIL-3
```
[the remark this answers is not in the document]
anupamachandra: We don't have an error like.."the stride is a multiple of X". I was thinking about whether it was benefitial to add it just for the cases where stride is a compile time constant. But decided to skip it. What do you think, it's useful to have something like ``` if(StrideIsCompileTimeConstant && !MatrixLayoutIsOptimal) CheckIfStrideIsAMultipleOf16() ```?

damyanp:
This LGTM still.  @Tex3d - please merge when you're ready to approve.

tex3d (approved):
Looks good.

lib/DxilValidation/DxilValidation.cpp
```diff
@@ -1229,6 +1229,32 @@ static void ValidateImmOperandsForOuterProdAcc(CallInst *CI,
...
+            InstrLinalgInvalidMatrixLayoutValueForOuterProductAccumulate,
+        {GetMatrixLayoutStr(MLValue),
+         GetMatrixLayoutStr(static_cast<unsigned>(
+             DXIL::LinalgMatrixLayout::OuterProductOptimal))});
+
+  llvm::Value *MatrixStride =
+      CI->getOperand(DXIL::OperandIndex::kOuterProdAccMatrixStride);
+  if (!llvm::isa<llvm::Constant>(MatrixStride)) {
+    ValCtx.EmitInstrError(
+        CI, ValidationRule::InstrLinalgMatrixStrideZeroForOptimalLayouts);
```
bob80905, line 1248: Shouldn't this be a separate error? One error that triggers when the value isn't 0, and one when the value isn't constant?
Using the same diagnostic could work, but perhaps an improvement to the message would be to specify that the 0 has to be a constant.

anupamachandra: So matrix stride is in general not expected to be a constant value, but for optimal formats it has to be a constant zero value, hence I structured the error this way. I can change the error to specify that the value should be cosntant zero for optimal layouts if you prefer. Also for this and the comment below do you want it in a separate PR once this is merged?

bob80905 (approved):

tex3d (approved):

damyanp (approved):
Approving post-merge.

Issue #7419, opened 2025-05-02
[HLSLOptions] Deprecate `-select-validator` option


bob80905:
This issue tracks the deprecation of the select-validator option. The validator will be internal by default. The only way to override this is to provide an environment variable when invoking dxc.exe. There is no longer any use of this option, and it is not officially documented anywhere. 
This is compliant with the soon-to-be design proposal here: 
https://github.com/microsoft/hlsl-specs/pull/499/files#diff-37d48932bf2b3d27214cfe1f7e46168df5309cdb9a56cf16583b7a78e106f600

PR #7423, opened 2025-05-02
[HLSL Options] Remove select-validator option

bob80905:
This PR removes the select-validator option. It is being deprecated, and it wasn't ever officially documented.
Fixes https://github.com/microsoft/DirectXShaderCompiler/issues/7419


tex3d (commented):
We need a way to update tests with an option to prevent downlevel validator testing.  Some of these tests should be revisited more carefully for their intended purpose after adding the external validator path option and global environment variable override.

tools/clang/test/HLSLFileCheck/infra/auto-dxilver.hlsl
```diff
@@ -21,7 +21,7 @@
 // Catch any other unexpected error cases.
 // CHECK-NOT: error
 
-// RUN: %dxc -T vs_6_8 -select-validator internal %s | FileCheck %s
+// RUN: %dxc -T vs_6_8 %s | FileCheck %s
```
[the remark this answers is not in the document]
bob80905: Understood, removed that Run Line. Cleaned up the comments too.
Fully intend to use the proposed dxil_dll_path option to augment these tests when I get around to that.

tools/clang/test/Parser/deprecated-select-validator.hlsl
tex3d, line 1: I just realized that this is in the wrong location.  This is a folder of clang c/cpp parser tests, not for testing the argument parsing.  This whole folder is disabled for the DXC project, so this test will not run.

Check `lit.local.cfg` under Parser for:
```
config.suffixes = [] # HLSL Change - Disable lit suites.
```

This should be moved under the `DXC` folder instead.

tex3d (changes requested):
The option test needs to be moved under the `DXC` path instead of `Parser`.

tex3d (approved):

tools/clang/test/DXC/deprecated-select-validator.hlsl
```diff
@@ -0,0 +1,14 @@
+// Test that the deprecated option, select-validator, doesn't work.
+// RUN: not %dxc -E main -T vs_6_7 -select-validator internal %s 2>&1 | FileCheck %s
+
+// CHECK: dxc failed : Unknown argument: '-select-validator'
```
pow2clk, line 4: I wasn't going to suggest we might include a bespoke error message since the option was hidden all along. For some reason, this test makes me think to at least mention the possibility of a warning that the flag is ignored and a mention of the replacement environment variable. I don't mind either way.

pow2clk (approved):
I have a suggestion, but you can take or leave it.

PR #7424, opened 2025-05-02
[CoopVec] Initial CoopVec ExecutionTest support for Mul[Add] and OuterProductAccumulate

jholewinski-nv:
ExecutionTest::CoopVec_MulAdd:

Functional verification for the Mul() and MulAdd() HLSL APIs. The driver
matrix conversion API is tested as well. These tests should be considered
as work-in-progress as this point. They include coverage primarily for SINT8,
FLOAT16, FLOAT_E4M3, and FLOAT_E5M2. The test queries the driver for all
supported configurations and runs each one, with a filtering mechanism to
limit the set of tests to the minimal feature set. The set of tests can be
further filtered by the following TE parameters:

  CoopVecMatrixInterp: SINT8, FLOAT16, FLOAT_E4M3, ...
  CoopVecMatrixLayout: ROW_MAJOR, COLUMN_MAJOR, MUL_OPTIMAL, OUTER_PRODUCT_OPTIMAL
  CoopVecBiasInterp: SINT32, FLOAT16, FLOAT_E4M3, ...
  CoopVecInputInterp: SINT8, FLOAT16, FLOAT_E4M3, ...
  CoopVecInputType: SINT8, UINT8, SINT16, UINT16, SINT32, UINT32, FLOAT16, FLOAT32, ...
  CoopVecOutputType: SINT32, UINT32, FLOAT16, FLOAT32, ...

Filter example:
  $ TE.exe ... -p:CoopVecMatrixInterp=FLOAT16 -p:CoopVecMatrixLayout=MUL_OPTIMAL

Precision coverage is minimal at this point, using an all-ones input matrix and
test vector with ones in the first two components. This is enough to test basic
functionality, but more comprehensive tests are needed.

ExecutionTest::CoopVec_OuterProduct:

Functional verification for the OuterProductAccumulate() HLSL API. This test queries
the driver for all supported configurations and runs each one. No filtering is
currently implemented.


jholewinski-nv:

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -752,7 +758,7 @@ class ExecutionTest {
 #endif
   }
 
-  bool UseDebugIfaces() { return true; }
+  bool UseDebugIfaces() { return false; }
```
jholewinski-nv, line 761: This is a WAR for the debug layer crashes. We'll need a better solution.

jholewinski-nv:
The code formatting check is failing on `CoopVecAPIExtensions.h`, which are the extensions for `d3d12.h`. Since this is external code, should be exclude it from the clang-format check?

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11912,10 +11961,1120 @@ VERIFY_SUCCEEDED(DoArraysMatch<T>(OutputVector, ExpectedVector,
...
+// Filter example:
+//   TE.exe ... -p:CoopVecMatrixInterp=FLOAT16 -p:CoopVecMatrixLayout=MUL_OPTIMAL
+//
+// The current implementation will always write the final output data as float.
+void ExecutionTest::RunCoopVecMulTest() {
+  // Create device and verify coopvec support
+  CComPtr<ID3D12Device> pD3DDevice;
+  if (!CreateDevice(&pD3DDevice, D3D_SHADER_MODEL_6_9)) {
+    WEX::Logging::Log::Comment(
+        "Device does not support SM 6.9. Skipping.");
```
damyanp: This isn't the only reason that CreateDevice could fail.  It looks like CreateDevice already has logging to report details of why it failed.  It's probably best not to confidently state reasons for failure that may not be the case.  Otherwise you might send people down a long rabbit hole debugging only to discover that there's some other reason that the creation is failing.  Ask me how I know!

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11912,6 +11971,1378 @@ VERIFY_SUCCEEDED(DoArraysMatch<T>(OutputVector, ExpectedVector,
...
+                inputVector1.size());
+
+    std::vector<float> inputVector2FP32(inputVector2.size() / sizeof(float));
+    std::memcpy(inputVector2FP32.data(), inputVector2.data(),
+                inputVector2.size());
+
+    for (int threadIdx = 0; threadIdx < Config.NumThreads; ++threadIdx) {
+      for (int m = 0; m < Config.DimM; ++m) {
+        for (int n = 0; n < Config.DimN; ++n) {
+          float acc = inputVector1FP32[m] * inputVector2FP32[n];
```
jenatali, line 13113: I see that `threadIdx` isn't used here when determining what set of input data to look at, but in the shader, `SV_GroupThreadID` is used to determine the offset of the vector data to load. This seems to be producing different results than what the shader would produce.

tools/clang/unittests/HLSLExec/CoopVec.h
```diff
@@ -0,0 +1,330 @@
...
+  static std::vector<uint8_t> CreateAllOnesInputMatrix(uint32_t Width, uint32_t Height) {
+    std::vector<EltTy> inputMatrix(Width * Height);
+    for (uint32_t i = 0; i < Width * Height; i++) {
+      if constexpr (std::is_same_v<EltTy, uint8_t> || std::is_same_v<EltTy, int8_t>) {
+        inputMatrix[i] = 1;
+      } else if constexpr (std::is_same_v<EltTy, DirectX::PackedVector::HALF>) {
+        inputMatrix[i] = ConvertFloat32ToFloat16(1.0f);
+      } else if constexpr (std::is_same_v<EltTy, float>) {
+        inputMatrix[i] = 1.0f;
+      }
```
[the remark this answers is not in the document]
jholewinski-nv: Fixed

tools/clang/unittests/HLSLExec/CoopVec.h
```diff
@@ -0,0 +1,330 @@
...
+                                            void **ppvObject) override {
+// FIXME: This is a workaround for a warning-as-error about unused parameters.
+#pragma warning(push)
+#pragma warning(disable: 4100)
+    return DoBasicQueryInterface<IDxcIncludeHandler>(this, iid, ppvObject);
+#pragma warning(pop)
+  }
+};
+
+struct CoopVecHelpers {
```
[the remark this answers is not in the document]
jholewinski-nv: No, changed to namespace.

tools/clang/unittests/HLSLExec/CoopVec.h
```diff
@@ -0,0 +1,330 @@
...
+        return 4;
+      default:
+        return 1;
+    }
+  }
+
+  // This type is used in generated HLSL source to represent the vector type
+  // for the given data type.
+  static std::wstring GetHlslDataTypeForDataType(D3D12_LINEAR_ALGEBRA_DATATYPE DataType, D3D12_LINEAR_ALGEBRA_DATATYPE InputInterpretation) {
+    UNREFERENCED_PARAMETER(InputInterpretation);
```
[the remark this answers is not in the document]
jholewinski-nv: No, it's a hold out from an earlier implementation. I removed it.

tools/clang/unittests/HLSLExec/CoopVecAPIExtensions.h
```diff
@@ -0,0 +1,165 @@
+#pragma once
+
+#if D3D12_PREVIEW_SDK_VERSION < 717
+
+// This file contains the definitions of the D3D12 cooperative vector API extensions.
+// It is used to test the cooperative vector API extensions on older SDKs.
```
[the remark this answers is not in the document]
jholewinski-nv: Fixed

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
jholewinski-nv: Yup, that's fair. I dropped the error strings.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
jholewinski-nv: Good catch, this currently works because the input vectors are the same across threads but that's clearly not the intended future behavior. Fixed.

jenatali: Oh, I see. Then the bug is this in the shader instead:
```hlsl
  uint inputOffset1 = (DIM_M * threadIdx);
  inputOffset1 = (inputOffset1 + 3) & ~3; // Align to 4 bytes
  vector<INPUT_DATA_TYPE, DIM_M / INPUT_DIVISOR> input1 = InputVector1.Load<vector<INPUT_DATA_TYPE, DIM_M / INPUT_DIVISOR> >(inputOffset1);
```

`DIM_M` is an element count, so the result of `inputOffset1` is an element offset, but it's passed into the ByteAddressBuffer as a byte offset.

While testing a WARP implementation I'm seeing the different threads loading different data, which it sounds like is unexpected from your comment. I'm still pretty sure I'm looking at a test bug here though.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11912,6 +11971,1356 @@ VERIFY_SUCCEEDED(DoArraysMatch<T>(OutputVector, ExpectedVector,
...
+RWByteAddressBuffer OutputBuffer: register(u0);
+
+[shader("compute")]
+[numthreads(NUM_THREADS, 1, 1)]
+void main(uint threadIdx : SV_GroupThreadID)
+{
+  using namespace dx::linalg;
+
+  // Ensure 4-byte alignment for vector loads
+  uint inputOffset = (INPUT_PER_THREAD * threadIdx);
```
jenatali, line 12244: The same bug is here. This is an element offset, not a byte offset, it needs a `* sizeof(INPUT_DATA_TYPE)` here.

damyanp:
What's the plan for fixing these build errors?

https://dev.azure.com/DirectXShaderCompiler/public/_build/results?buildId=8102&view=logs&j=768863b7-b94f-55cd-5a82-1cc425bff7c1&t=c1057c0c-eb6e-5a40-81b8-f53da0e21902&l=2095

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
jholewinski-nv: Fixed

jholewinski-nv:
> What's the plan for fixing these build errors?
> 
> https://dev.azure.com/DirectXShaderCompiler/public/_build/results?buildId=8102&view=logs&j=768863b7-b94f-55cd-5a82-1cc425bff7c1&t=c1057c0c-eb6e-5a40-81b8-f53da0e21902&l=2095

I've been debating about this one. We can pull in `ID3D12GraphicsCommandList10`, but that brings along even more dependencies on structs/enums. How far down the rabbit hole do we go?

jholewinski-nv:

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -11912,6 +11981,1374 @@ VERIFY_SUCCEEDED(DoArraysMatch<T>(OutputVector, ExpectedVector,
...
+
+  VERIFY_SUCCEEDED(D3DDevice->CheckFeatureSupport(
+      (D3D12_FEATURE)D3D12_FEATURE_COOPERATIVE_VECTOR, &DevOptions,
+      sizeof(DevOptions)));
+
+  // Test each supported data type and matrix layout
+  for (auto MulAddConfig : MulAddProps) {
+    // Filter on preview test support
+    bool PreviewConfig = false;
+    if (MulAddConfig.InputType == D3D12_LINEAR_ALGEBRA_DATATYPE_FLOAT16 &&
```
tex3d, line 12045: Just a minor comment, no change really necessary:
Wouldn't a nicer way to do this be to have an array of preview configs to compare against?

tex3d (approved):
I think this is fine for the staging branch.

I have some reservations for the test structure for when we need merge to main and port to the HLK, such as the ability to separate test scenarios into ones supported by different device feature support driven by the gatherer instead of running every possible combination under one test method.  Additionally, there is a *lot* of code in some of these functions, which might be better if broken up a bit more into units of functionality, even if creating a class to manage shared state is necessary.

damyanp:
I'm going to go ahead and merge this - the license/cla check doesn't seem to be working, but I see that the CLA has been accepted in the comments and we've discussed offline about taking this in.

PR #7427, opened 2025-05-03
Use WARP from nuget by default

damyanp:
Previously using WARP from nuget was something that had to be opted in to.  This change makes it the default behvior.

This allows some simplifications.  The nuget package is installed at configure time. If the user wants to override the nuget installation behavior then they can pass in extra parameters, rather than having an alternative mechanism for describing which package to use.

The actual DLL to load is specified through a TAEF parameter and specifies the full path to the DLL.


tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -806,6 +806,21 @@ class ExecutionTest {
...
+
+      // The WARP_DLL runtime parameter can be used to specify a specific DLL to
+      // load.  To force this to be used, we make sure that this DLL is loaded
+      // before attempting to create the device.
+      HMODULE ExplicitlyLoadedWarpDll = NULL;
+      WEX::Common::String WarpDllPath;
+      if (SUCCEEDED(WEX::TestExecution::RuntimeParameters::TryGetValue(
+              L"WARP_DLL", WarpDllPath))) {
+        WEX::Logging::Log::Comment(WEX::Common::String().Format(
+            L"WARP_DLL requested: %ls", (const wchar_t *)WarpDllPath));
```
alsepkow, line 818: >WarpDllPath [](http://example.com/codeflow?start=57&length=11)

You can call GetBuffer on a WEX::Common::String instead of casting.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -806,6 +806,21 @@ class ExecutionTest {
 
     VERIFY_SUCCEEDED(CreateDXGIFactory1(IID_PPV_ARGS(&factory)));
     if (GetTestParamUseWARP(UseWarpByDefault())) {
+
+      // The WARP_DLL runtime parameter can be used to specify a specific DLL to
+      // load.  To force this to be used, we make sure that this DLL is loaded
+      // before attempting to create the device.
+      HMODULE ExplicitlyLoadedWarpDll = NULL;
```
alsepkow, line 813: >HMODULE ExplicitlyLoadedWarpDll = NULL; [](http://example.com/codeflow?start=6&length=39)

You can use a wil::unqiue_hmdoule here instead and avoid needing the code that free's it below. Also good because it avoids leaking the handle if the verify throws as the wil stuff is cleaned up on scope exit.

#include <wil/resource.h>

wil::unique_hmodule ExplicitlyLoadedWarpDLL(LoabLibraryExW(WarpDllPath, NULL, 0);

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -806,6 +806,21 @@ class ExecutionTest {
...
+      WEX::Common::String WarpDllPath;
+      if (SUCCEEDED(WEX::TestExecution::RuntimeParameters::TryGetValue(
+              L"WARP_DLL", WarpDllPath))) {
+        WEX::Logging::Log::Comment(WEX::Common::String().Format(
+            L"WARP_DLL requested: %ls", (const wchar_t *)WarpDllPath));
+        ExplicitlyLoadedWarpDll = LoadLibraryExW(WarpDllPath, NULL, 0);
```
alsepkow, line 819: >NULL [](http://example.com/codeflow?start=62&length=4)

should 'nullptr' be used for new code?

tools/clang/test/taef_exec/DownloadWarp.cmake
```diff
@@ -0,0 +1,97 @@
...
+# This script is run at configure time to install the latest WARP.
+
+if(NOT WIN32)
+    # WARP is only a thing on Win32
+    return()
+endif()
+
+
+#
+# At runtime, the Execution Tests looks at the WARP_DLL runtime parameter to
```
alsepkow, line 12: >the Execution Tests looks  [](http://example.com/codeflow?start=14&length=26)

nit: grammar 'The Execution Tests look'

tools/clang/test/taef_exec/DownloadWarp.cmake
```diff
@@ -0,0 +1,97 @@
...
+# decide which DLL to use. When USE_NUGET_WARP is set to true, this will
+# configure the tests by default to use the DLL from the installed nuget
+# package.
+
+set(USE_NUGET_WARP TRUE CACHE BOOL
```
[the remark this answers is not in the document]
damyanp: It is: this is the escape hatch to allow you to opt out of it if it doesn't work for some reason.  eg you want to build without an internet connection.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
damyanp: Yup.  I prefer this though, since it's actually what GetBuffer does implicitly.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
damyanp: I don't want to add a dependency on wil here though, but good catch on it not being exception safe. I've put a small RAII helper in here to handle that.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
damyanp: For "proper" pointers I would, but I try and use NULL when interacting with Win32 API's that explicitly state "NULL" as having a meaning to the API.

tools/clang/test/taef_exec/DownloadWarp.cmake
damyanp: fixed

alsepkow (approved):
LGTM, including the minor cast suggestion. Can re-approve if you change it.

tools/clang/unittests/HLSLExec/ExecutionTest.cpp
```diff
@@ -806,6 +806,35 @@ class ExecutionTest {
...
+          }
+        }
+      };
+
+      WarpDll ExplicitlyLoadedWarpDll;
+      WEX::Common::String WarpDllPath;
+      if (SUCCEEDED(WEX::TestExecution::RuntimeParameters::TryGetValue(
+              L"WARP_DLL", WarpDllPath))) {
+        WEX::Logging::Log::Comment(WEX::Common::String().Format(
+            L"WARP_DLL requested: %ls", (const wchar_t *)WarpDllPath));
```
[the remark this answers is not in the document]
damyanp: Personally, I don't prefer it in this case :)

tools/clang/test/taef_exec/DownloadWarp.cmake
tex3d, line 1: I was slightly surprised you didn't place this file under `cmake/modules`, not that I know much about cmake conventions.  I was also thinking we should get the AgilitySDK from nuget as well, so I wonder if defining a general function for installing nugets would be even better.

tex3d: Ok, I now see that the original contained the general FetchNuGetPackage function, plus warp-specific code.  Why wouldn't we want a general function defined under `cmake/modules`?

damyanp: Mainly because I don't want to spend time trying to design and build a generic function for installing nuget packages when we only have a need for this one.  If we find a need for another, then we might be able to figure out what the common requirements between the two are and we can make an appropriate decision then.

damyanp:
Two approvals and a review ... merging now and I'll keep a close eye on everything!

PR #7430, opened 2025-05-05
Implement HLSL Diagnostics for LinAlg operations

anupamachandra:
This PR implements checks to validate the linalg builtin functions: __builtin_MatVecMul, __builtin_MatVecMulAdd, __builtin_OuterProductAccumulate and __builtin_VectorAccumulate. This includes:
- verify valid types for input and output vectors
- const checks for compile-time const parameters
- value checks for interpretation and layout (enum) parameters
- min/max checks for matrix dimensions
- verify input-output vector are the right dimensions for the given matrix (dimM and dimK) : packed and unpacked cases
- verify matrix layout, transpose and stride rules
- incorrect shader model warning

Adds tests for all the above error checks.

Implements https://github.com/microsoft/DirectXShaderCompiler/issues/7336


tools/clang/test/SemaHLSL/hlsl/linalg/builtins/mul_add_invalid.hlsl
```diff
@@ -0,0 +1,239 @@
+// RUN: %dxc -T lib_6_9 -enable-16bit-types %s -verify
+
+enum CompType {
```
anupamachandra, line 3: Will update these to use the linalg headers.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11663,6 +11663,552 @@ static bool CheckBarrierCall(Sema &S, FunctionDecl *FD, CallExpr *CE,
...
+static const unsigned kMatVecMulMatrixLayoutIdx = 10;
+static const unsigned kMatVecMulMatrixTransposeIdx = 11;
+static const unsigned kMatVecMulMatrixStrideIdx = 12;
+static const unsigned kMatVecMulIsOutputUnsignedIdx = 13;
+
+// MatVecAdd
+const unsigned kMatVecMulAddBiasInterpretation = 15;
+const unsigned kMatVecMulAddIsOutputUnsignedIdx = 16;
+
+enum MatrixLayout {
```
anupamachandra, line 11686: Will update the matrixlayout and comptype to use the linalg headers.

anupamachandra:
Couple of items left to do here:
1. Update the DXIL tests with IR, since the HLSL diagnostics prevent the invalid IR generation we were relying on.
2. Packed Format input vector validations. Will discuss this offline about the integer to float cast issue I am running into.

anupamachandra:
> @anupamachandra - given that there's build errors and you have some more work planned, is it worth reviewing at this point?

Yes please. I'm working on the build errors and missed item is marked with a //FIXME. It think it'll be good to get some eyes on it to check if the PR meets expectations in terms of scope and if any big issues/gaps stand out.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8015,6 +8015,24 @@ def err_hlsl_hitobject_unsupported_stage : Error<
    "dx::HitObject is unavailable in shader stage '%0' (requires 'raygeneration', 'closesthit' or 'miss')">;
 // HLSL Change Ends
 
+// Linear Algebra Operations
```
anupamachandra, line 8018: eeps, set column width to 80. Note to self.

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11663,6 +11663,560 @@ static bool CheckBarrierCall(Sema &S, FunctionDecl *FD, CallExpr *CE,
...
 }
 
+// MatVec Ops
+static const unsigned kMatVecMulOutputVectorIdx = 0;
+static const unsigned kMatVecMulOutputIsUnsignedIdx = 1;
+static const unsigned kMatVecMulInputVectorIdx = 2;
+static const unsigned kMatVecMulIsInputUnsignedIdx = 3;
+static const unsigned kMatVecMulInputInterpretationIdx = 4;
+// static const unsigned kMatVecMulMatrixBufferIdx = 5;
+// static const unsigned kMatVecMulMatrixOffsetIdx = 6;
```
damyanp, line 11673: Is there a reason these are commented out?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11663,6 +11663,560 @@ static bool CheckBarrierCall(Sema &S, FunctionDecl *FD, CallExpr *CE,
...
+const unsigned kMatVecMulAddBiasInterpretation = 15;
+
+enum MatrixLayout {
+  MATRIX_LAYOUT_ROW_MAJOR = 0,
+  MATRIX_LAYOUT_COLUMN_MAJOR = 1,
+  MATRIX_LAYOUT_MUL_OPTIMAL = 2,
+  MATRIX_LAYOUT_OUTER_PRODUCT_OPTIMAL = 3
+};
+
+bool IsValidMatrixLayoutForMulandMulAddOps(unsigned Layout) {
```
damyanp, line 11691: If these functions aren't used outside of this cpp file it'd be good to mark them as static:

* IsValidMatrixLayoutForMulandMulAddOps
* IsOptimalTypeMatrixLayout
* IsValidTransposeForMatrixLayout
* IsPackedType

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11663,6 +11663,560 @@ static bool CheckBarrierCall(Sema &S, FunctionDecl *FD, CallExpr *CE,
...
+  case MatrixLayout::MATRIX_LAYOUT_ROW_MAJOR:
+  case MatrixLayout::MATRIX_LAYOUT_COLUMN_MAJOR:
+    return !Transposed;
+
+  default:
+    return true;
+  }
+}
+
+enum DataType {
```
damyanp, line 11714: Is there a reason this can't use the dxilconstants.h enum?  There may well be a good one, just wanted to check.

tools/clang/lib/Headers/hlsl/dx/linalg.h
```diff
@@ -43,14 +43,24 @@ enum MatrixLayout {
 // Helper for signedness
 //
 namespace details {
-template <typename T> bool IsUnsigned() { return false; }
 
-#ifdef __HLSL_ENABLE_16_BIT
-template <> bool IsUnsigned<uint16_t>() { return true; }
-#endif
+template <typename T> struct IsUnsigned {
+  static const bool value = false;
```
damyanp, line 48:
```suggestion
  static const bool Value = false;
```

tools/clang/lib/Sema/SemaHLSL.cpp
anupamachandra: They are not needed, but I kept them there to as a useful comment for anyone examining the callexpr.

tools/clang/lib/Sema/SemaHLSL.cpp
anupamachandra: Yep this was a TBD, I wanted to check if we wanted to us the dxilconstants one or the linalg headers before updating. The linalg update will need a cmake change to add the path to include the headers.

tools/clang/lib/Headers/hlsl/dx/linalg.h
anupamachandra: Phew, no line number change.. the tests needn't be updated!

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11663,6 +11664,535 @@ static bool CheckBarrierCall(Sema &S, FunctionDecl *FD, CallExpr *CE,
...
+static const unsigned kMatVecMulMatrixMIdx = 8;
+static const unsigned kMatVecMulMatrixKIdx = 9;
+static const unsigned kMatVecMulMatrixLayoutIdx = 10;
+static const unsigned kMatVecMulMatrixTransposeIdx = 11;
+static const unsigned kMatVecMulMatrixStrideIdx = 12;
+
+// MatVecAdd
+const unsigned kMatVecMulAddBiasInterpretation = 15;
+
+static bool IsValidMatrixLayoutForMulandMulAddOps(unsigned Layout) {
```
damyanp, line 11685: nit:

```suggestion
static bool IsValidMatrixLayoutForMulAndMulAddOps(unsigned Layout) {
```

damyanp (approved):

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
damyanp: ok to resolve?

damyanp (approved):

tools/clang/test/SemaHLSL/hlsl/linalg/builtins/mul_invalid.hlsl
```diff
@@ -0,0 +1,1153 @@
...
+  const uint input_interpretation_11 = 23;
+
+  // expected-error@+2 {{23 is an invalid Register Interpretation value}}
+  __builtin_MatVecMul(output_vector, is_output_unsigned, input_vector,
+                      is_input_unsigned, input_interpretation_11, matrix_buffer,   
+                      matrix_offset, matrix_interpretation, matrix_dimM,
+                      matrix_dimK, matrix_layout, matrix_is_transposed,
+                      matrix_stride);
+
+  const uint input_interpretation_12 = 100;
```
bob80905, line 700: Same as my previous comment

tools/clang/test/SemaHLSL/hlsl/linalg/builtins/mul_invalid.hlsl
```diff
@@ -0,0 +1,1153 @@
...
+
+  const uint matrix_interpretation_11 = 23;
+  // expected-error@+3 {{23 is an invalid Memory Interpretation value}}
+  __builtin_MatVecMul(output_vector, is_output_unsigned, input_vector,
+                      is_input_unsigned, input_interpretation, matrix_buffer,
+                      matrix_offset, matrix_interpretation_11, matrix_dimM,
+                      matrix_dimK, matrix_layout, matrix_is_transposed,
+                      matrix_stride);
+
+  const uint matrix_interpretation_12 = 100;
```
bob80905, line 995: Same as previous comment

tools/clang/test/SemaHLSL/hlsl/linalg/builtins/outer_product_accumulate_invalid.hlsl
```diff
@@ -0,0 +1,256 @@
...
+
+#include <dx/linalg.h>
+
+using namespace dx::linalg;
+
+ByteAddressBuffer input_vector_buffer;
+RWByteAddressBuffer accumulate_buffer;
+ByteAddressBuffer constants_buffer;
+
+// Check ofr input vectors aren't the same component type
```
bob80905, line 11: nit: Check if?

bob80905 (commented):
A few nits / questions

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
    "dx::MaybeReorderThread is unavailable in shader stage '%0' (requires 'raygeneration')">;
 def err_hlsl_hitobject_unsupported_stage : Error<
    "dx::HitObject is unavailable in shader stage '%0' (requires 'raygeneration', 'closesthit' or 'miss')">;
+
+// Linear Algebra Operations
+def err_hlsl_linalg_param_must_be_const : Error<
+  "'%0' must be a constant parameter">;
```
llvm-beanz, line 8019: Maybe instead use `err_expr_not_ice`.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
...
+def err_hlsl_linalg_vector_incorrect_type : Error<
```
llvm-beanz, line 8020: Maybe instead use `err_typecheck_cond_expect_int_float`

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
...
+  "%0 is incorrect type, must be 16-bit or 32-bit 'unsigned int', "
+  "'signed int' or 'float'">;
+def err_hlsl_linalg_isunsigned_incorrect_for_given_type : Error<
+  "%0 must be %select{false|true}1 for a %2 vector type">;
```
llvm-beanz, line 8024: Generally we should avoid strings as inputs to diagnostic formatters:
```suggestion
  "%0 must be %select{false|true}1 for vector of `%select{floating point|signed integer|unsigned integer}2 type">;
```

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
...
+def err_hlsl_linalg_interpretation_value_incorrect : Error<
+  "%0 is an invalid %select{Memory|Register}1 Interpretation value">;
```
llvm-beanz, line 8026:
```suggestion
  "%0 is an invalid %select{memory|register}1 interpretation value">;
```

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
...
+def err_hlsl_linalg_matrix_layout_is_not_transposable : Error<
+  "RowMajor and ColumnMajor matrices are not transposable">;
+def err_hlsl_linalg_optimal_matrix_layout_matrix_stride_must_be_zero : Error<
+  "for optimal matrix layout, matrix stride must be zero">;
```
llvm-beanz, line 8030:
```suggestion
  "for optimal matrix layout, matrix stride must be greater than 0">;
```

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
...
+def err_hlsl_linalg_matrix_dim_must_not_be_zero: Error<
+  "matrix dimension must not be zero">;
```
llvm-beanz, line 8032:
```suggestion
  "matrix dimension must be greater than 0">;
```

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
...
+def err_hlsl_linalg_matrix_layout_invalid : Error<
+  "matrix layout %0 is not valid, must be in the range %1 - %2">;
```
llvm-beanz, line 8034:
```suggestion
  "matrix layout %0 is not valid, must be in the range [%1, %2]">;
```

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
...
+def err_hlsl_linalg_function_requires_shader_model_6_9_or_above : Error<
```
llvm-beanz, line 8035: This should use `warn_hlsl_intrinsic_in_wrong_shader_model` (which is a default-error warning)

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
```diff
@@ -8013,6 +8013,54 @@ def err_hlsl_reorder_unsupported_stage : Error<
...
+  "packed input vector type must be a 32-bit unsigned int in linalg mul/muladd operations, "
+  "packed formats uint8_t4_packed and sint8_t4_packed are not supported currently">;
+def err_hlsl_linalg_mul_muladd_exceeds_max_matrix_dim_m: Error<
+  "matrix dimension M must be less than %0, in a linalg Mul/MulAdd operation">;
+def err_hlsl_linalg_mul_muladd_exceeds_max_matrix_dim_k_unpacked_input: Error<
+  "matrix dimension K when using unpacked input vectors must be less than %0, in a "
+  "linalg Mul/MulAdd operation">;
+def err_hlsl_linalg_mul_muladd_exceeds_max_matrix_dim_k_packed_input: Error<
+  "matrix dimension K when using packed input vectors must be less than %0, in a "
+  "linalg Mul/MulAdd operation">;
```
llvm-beanz, line 8057: We can collapse these down to a single error (this makes for more compact binary representation.
```suggestion
def err_hlsl_linalg_mul_muladd_invalid_dim: Error<
  "matrix dimension %select{M|K when using unpacked input vectors|K "
  "when using packed input vectors}0 must be less than %1, in a linalg "
  "Mul/MulAdd operation">;
```

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11663,6 +11664,581 @@ static bool CheckBarrierCall(Sema &S, FunctionDecl *FD, CallExpr *CE,
...
+                                             unsigned MatrixK, unsigned MatrixM,
+                                             bool isInputPacked) {
+  // Check if output vector size equals to matrix dimension M
+  if (OutputVectorSize != MatrixM) {
+    Expr *OutputVector = CE->getArg(kMatVecMulOutputVectorIdx);
+    S.Diags.Report(
+        OutputVector->getExprLoc(),
+        diag::
+            err_hlsl_linalg_mul_muladd_output_vector_size_not_equal_to_matrix_M);
+    return false;
```
llvm-beanz, line 11747: Should we be returning on each of these? That prevents emitting the full suite of diagnostics if there are multiple errors on the same call.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
anupamachandra: this one must be zero. (it's a requirement for optimal layouts)

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
anupamachandra: yeah, much better.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
anupamachandra: warnings can be overriden?

tools/clang/lib/Sema/SemaHLSL.cpp
```diff
@@ -11710,6 +12286,45 @@ void Sema::CheckHLSLFunctionCall(FunctionDecl *FDecl, CallExpr *TheCall,
...
+    }
+    CheckOuterProductAccumulateCall(*this, FDecl, TheCall);
+    break;
+  case hlsl::IntrinsicOp::IOP___builtin_VectorAccumulate:
+    if (!SM->IsSM69Plus()) {
+      Diags.Report(
+          TheCall->getExprLoc(),
+          diag::err_hlsl_linalg_function_requires_shader_model_6_9_or_above)
+          << FDecl->getNameAsString();
+      return;
```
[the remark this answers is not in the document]
anupamachandra: Other than the SM6.9 testing, I don't think there is any HLSL validation needed.
```
void __builtin_VectorAccumulate(vector<TY, NUM> InputVector,
                                RES OutputArrayResource,
                                uint OutputArrayOffset);
```
what do you think?

tools/clang/test/SemaHLSL/hlsl/linalg/builtins/mul_add_invalid.hlsl
```diff
@@ -0,0 +1,1395 @@
...
+  const uint matrix_offset = 0;
+  const uint matrix_dimM = 4;
+  const uint matrix_dimK = 4;
+  const uint matrix_layout = MatrixLayout::MATRIX_LAYOUT_ROW_MAJOR;
+  const bool matrix_is_transposed = false;
+  const uint matrix_stride = 64;
+  const uint bias_offset = 0;
+  const uint bias_interpretation = DataType::DATA_TYPE_FLOAT32;
+
+  const uint matrix_interpretation_0 = 0;
```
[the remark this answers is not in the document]
anupamachandra: It makes sure that the right error is triggered for all the invalid cases, so it checks if the valid lists for each interpretation type is correct.

tools/clang/lib/Sema/SemaHLSL.cpp
anupamachandra: So the errors have dependencies with previous errors and so I structured it this way. For example, the value testing of certain parameters (intepretation, isunsigned, matrix dimensions) etc depends on the is constant expression test. We boundary check the matrix dimensions before checking if the input/output vetctor sizes are valid for the matrix shape etc. So, I stop the further checking for a call in the presence on an error. What do you think?

tools/clang/test/SemaHLSL/hlsl/linalg/builtins/mul_add_invalid.hlsl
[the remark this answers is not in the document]
anupamachandra: The specific values were chosen per this rationale: Check for all the values from the below enum that are not allowed and check one value outside the enum.
enum class ComponentType : uint32_t {
  Invalid = 0,
  I1,
  I16, // = 2
  U16, // = 3
  I32, // = 4
  U32, // = 5
  I64,
  U64,
  F16, // = 8
  F32, // = 9
  F64,
  SNormF16,
  UNormF16,
  SNormF32,
  UNormF32,
  SNormF64,
  UNormF64,
  PackedS8x32, // = 17
  PackedU8x32, // = 18

  // BEGIN NEW FOR SM 6.9
  U8,      // = 19
  I8,      // = 20
  F8_E4M3, // = 21  
  F8_E5M2, // = 22
  // END     

  LastEntry
};

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
llvm-beanz: Doh. I misread it, but we should use `0` not `zero` to avoid localization issues.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
llvm-beanz: yes, and for all the availability warnings we allow them to be overridden because they _might_ be valid if the compiler fully optimizes them away.

tools/clang/lib/Sema/SemaHLSL.cpp
llvm-beanz: We should try to do as many errors as we can without failing and aborting so that we don't force compile->error->retry->... cycles.

If there are errors that are dependent (like the constant expr test), that's fine to stop early, but we should try to be as complete as we can in a single pass.

tools/clang/lib/Sema/SemaHLSL.cpp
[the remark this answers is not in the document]
bob80905: Yeah, doesn't seem to be anything, resolved.

tools/clang/lib/Sema/SemaHLSL.cpp
anupamachandra: hmm.. I agree with the reasoning but this would require structuring the whole validation check code to allow this (unless I can use goto-s ;)). Is it ok to do this in a follow-up?

daviddemarco135: We will file issue and do this post preview due to this estimated to take 1 - 2 weeks. Anupama will create issue within comment.

tools/clang/lib/Headers/hlsl/dx/linalg.h
```diff
@@ -43,14 +43,24 @@ enum MatrixLayout {
 // Helper for signedness
 //
 namespace details {
-template <typename T> bool IsUnsigned() { return false; }
 
-#ifdef __HLSL_ENABLE_16_BIT
-template <> bool IsUnsigned<uint16_t>() { return true; }
-#endif
+template <typename T> struct IsUnsigned {
+  static const bool Value = false;
```
tex3d, line 48: I feel like the default template shouldn't have any definition for `value`, if possible.  That way, unsupported types trigger an error here instead of defaulting to signed.  That would require specializations for float types and all supported signed int types, of course, but it would also prevent use of unexpected types.

tools/clang/lib/Headers/hlsl/dx/linalg.h
```diff
@@ -43,14 +43,24 @@ enum MatrixLayout {
...
-template <> bool IsUnsigned<uint32_t>() { return true; }
-template <> bool IsUnsigned<uint64_t>() { return true; }
+template <> struct IsUnsigned<uint64_t> {
+  static const bool value = true;
+};
+
+#ifdef __HLSL_ENABLE_16_BIT
+template <> struct IsUnsigned<uint16_t> {
+  static const bool value = true;
+};
```
tex3d, line 62: These specializations use a different casing for `Value`.
```suggestion
template <> struct IsUnsigned<uint64_t> {
  static const bool Value = true;
};

#ifdef __HLSL_ENABLE_16_BIT
template <> struct IsUnsigned<uint16_t> {
  static const bool Value = true;
};
```

tex3d (commented):
A couple comments on `linalg.h` API changes.

tools/clang/lib/Headers/hlsl/dx/linalg.h
anupamachandra: Yeah, this is the issue Joshua mentioned in his review that I mentioned at today's morning meeting was a bug. I am adding high-level API tests that would have caught this.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
anupamachandra: I remembered as I was modifying this per your feedback that I had originally tried to use `warn_hlsl_intrinsic_in_wrong_shader_model` but that error prints out the entry function which is not available here. In my earlier PR I had added check to the `DiagnoseReachableHLSLCall` in SemaHLSL.cpp, but that gets called after these intrinsics are processed. Is there a better place to move this diagnostic?

tools/clang/lib/Headers/hlsl/dx/linalg.h
anupamachandra: Do we still want this, given the builtins are now updated to use only the allowed types?

tools/clang/lib/Sema/SemaHLSL.cpp
anupamachandra: https://github.com/microsoft/DirectXShaderCompiler/issues/7456

tools/clang/lib/Headers/hlsl/dx/linalg.h
anupamachandra: I added tests for uint and uint16_t, removed the uint64_t case since that will fail the builtin overload check.

tools/clang/include/clang/Basic/DiagnosticSemaKinds.td
llvm-beanz: The API availability diagnostics should all be emitted from `DiagnoseReachableHLSLCall`. You can't process it in the intrinsic handling because if it isn't reachable we shouldn't emit the diagnostic.

tex3d: @llvm-beanz
I pushed a change (after your comment) that fixed this by removing this diagnostic, the extra shader model checks, and just relies on the availability diagnostic by adding the `min_sm` attribute to the intrinsics.

This means that other parameter checking diagnostics will fire during parse time, before the shader model is checked.  I think this is what we want.  If so, please resolve, if not, please speak up.  Thanks!

tex3d (approved):

bob80905 (approved):

PR #7432, opened 2025-05-06
[SM6.9] Disable native vec deriv ops and expand testing

pow2clk:
Several intrinsics that were enabled for native vectors late got their testing removed as it expected scalarized forms. This adds tests for pow, modf, and abs in their native vector intrinsic forms. It removes native vector intrinsics for derivative operations as they require more scalarization removal due to their convergent markers and restores the scalarized testing for them. The 1024 size was removed from longvec-intrinsics as the verbose way that constant vectors are represented in the disassembly made the test take significantly longer.


tex3d (approved):
Yes please, looks good.  Let's take this change into preview.

damyanp (approved):
Do the existing issues we have track the work required to re-enable them all again?

PR #7435, opened 2025-05-07
[SER] Transform RayDesc in ScalarReplHLSL for robust lowering

### LINKED ISSUE #7434: Unflattened RayDesc breaking HL->DXIL lowering
ScalarReplHLSL is missing transforms for some RayDesc users, making the RayDesc flattening unreliable. However, the RayDesc has to be flattened in the HL op parameter lists or the DXIL opcode translation will fail.

**Steps to Reproduce**
HLSL shaders (raygen) with the following snippet will fail. The `HitObject::TraceRay` prevents the flattening of the `RayDesc`.
```
    dx::HitObject hit1 = dx::HitObject::MakeMiss(RAY_FLAG_NONE, 0, ray);
    dx::MaybeReorderThread(hit1);
    dx::HitObject hit2 = dx::HitObject::TraceRay(topObject, RAY_FLAG_SKIP_TRIANGLES, 0xFF, 0, 0, 0, ray, payload);
    dx::MaybeReorderThread(hit2);
```

**Actual Behavior**
Assert/crash in `TranslateHitObjectMake` as the `RayDesc` isn't flattened in the MakeMiss HL op.

simoll:
ScalarReplHLSL will refuse to faltten RayDesc if some users are incompatible. This is at odds with HLOperationLower, which always expects a specific flattened form.
This implements explicit RayDesc flattening in ScalarReplHLSL to make this more reliable. One issue with RayDesc-from-cbuffer in RayQuery remains and is documented with a test.

Implement ScalarReplHLSL handling for RayDesc param in HL ops:
- HitObject::MakeMiss [RewriteCallArg]
- HitObject::TraceRay [RewriteCallArg]
- TraceRay            [RewriteCallArg]
- RayQuery::TraceRayInline [unchanged - flatten to RayDesc elements]

In the process:
- Fix copyout code in RewriteCallArg (ScalarReplHLSL)
- Remove obsolete tests that expected a RayDesc alloca (where RayDesc is now SROA'd entirely)
- Document RayQuery lowering bug when RayDesc is taken directly from cbuffer.

Closes #7434  Unflattened RayDesc breaking HL->DXIL lowering #7434


lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -2780,21 +2781,10 @@ void SROA_Helper::RewriteCall(CallInst *CI) {
       case IntrinsicOp::MOP_DxHitObject_MakeMiss: {
         if (OldVal ==
             CI->getArgOperand(HLOperandIndex::kHitObjectMakeMissRayDescOpIdx)) {
-          RewriteWithFlattenedHLIntrinsicCall(CI, OldVal, NewElts,
-                                              /*loadElts*/ true);
-          DeadInsts.push_back(CI);
+          RewriteCallArg(CI, HLOperandIndex::kHitObjectMakeMissRayDescOpIdx,
```
tex3d, line 2784: I don't understand why we are no longer flattening RayDesc for MakeMiss.  This doesn't seem consistent with your description of the solution: "This implements explicit RayDesc flattening in ScalarReplHLSL to make this more reliable."

I see that we did inconsistent things with existing operations for RayDesc parameters, when I think it should have always been flattened for all operations that take this input instead.  Only the payload argument should remain unflattened and use RewriteCallArg.

What happens if you change the TraceRay operations to flatten RayDesc instead?  Won't that be more consistent?

tex3d: Of course a complication is that the flattening of the TraceRay RayDesc arg would change the subsequent payload parameter to a new index, where RewriteCallArg is needed, and may come before or after the flattening, due to the way RewriteCall is invoked on users of an aggregate pointer.  It may be better to do all necessary RewriteCallArg operations (by iterating users of the specific intrinsics that need it) before DoScalarReplacement.

tex3d: Oh, that won't be easy to do it appears, since `RewriteCallArg` uses `RewriteMemIntrin`, which depends on the value-specific instance of SROA_Helper.

I spent some time experimenting with this.  I have a draft PR that refactors this out of RewriteCall to allow for flattening without disrupting UDT args.  It seems a bit cleaner, but I haven't switched to flattening or fixed up all the tests sensitive to the change yet.

See #7440.

simoll: > I have a draft PR 

Same. Here is an alternative draft that uses RayQuery's load-and-flatten approach for all RayDesc ops: #7441

> This doesn't seem consistent with your description of the solution: "This implements explicit RayDesc flattening in ScalarReplHLSL to make this more reliable."

True, the term 'flattening' is confusing.. i meant something like 'treatment'. All RayDesc users have to be treated by scalarreplhlsl or RayDesc flattening for some calls may fail and we get the crash.

> What happens if you change the TraceRay operations to flatten RayDesc instead? Won't that be more consistent?

Yes but it means we have to update a lot of tests, so i opted for this solution, which leaves existing `TraceRay` unchanged.

> Of course a complication is that the flattening of the TraceRay RayDesc arg would change the subsequent payload parameter to a new index,

That's fine, we know which hl ops we are dealing with and the payload parameter is always the last one (eg, see #7441 which exploits that).

simoll:
Pipelines failing because the `RaygenAllocaStructAlignment` pixtest has to be updated. Will do that once we settle for a solution.

pow2clk (commented):
I just have one question.

I think we've decided to move forward with this change for the time being. I think @tex3d had a conversation with Jeff Noyle and confirmed that the pix test changes are fine. Since the alternatives would, at least, require significantly more testing, it's preferrable to take this solution while those get completed.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
```diff
@@ -2687,7 +2685,7 @@ void SROA_Helper::RewriteCallArg(CallInst *CI, unsigned ArgIdx, bool bIn,
     Builder.SetInsertPoint(CI->getNextNode());
     MemCpyInst *cpy = cast<MemCpyInst>(Builder.CreateMemCpy(
         userTyV, Alloca, DL.getTypeAllocSize(userTyElt), false));
-    RewriteMemIntrin(cpy, cpy->getRawSource());
+    RewriteMemIntrin(cpy, cpy->getRawDest());
```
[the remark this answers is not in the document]
simoll: This was erroneously triggering the memcpy-undef-to-undef-optimization  in`RewriteMemIntrin` for the `out`-copy. This change makes the `in` and `out` `RewriteMemIntrin` calls symmetric and resolves the mis-optimization.

lib/Transforms/Scalar/ScalarReplAggregatesHLSL.cpp
simoll: Considering this resolved as we've agreed to go ahead with this PR as a WAR. A proper flatten-all-RayDescs solution will be put in place later.

damyanp:
We're closing this one in favor of #7440.

